PR #257492, opened 2025-07-23
race prompt response and terminal/task idle, hide prompt part when terminal/task becomes idle or prompt is replied to

meganrogge:
When the terminal or task becomes idle, we want to hide the prompt if it hasn't been replied to by then. This PR accomplishes that.

It adds `hide` and `onDidRequestHide` to `ChatElicitationRequestPart` so that we can hide the prompt if it has not been replied to by the time the terminal or task has become idle or when the prompt is replied to since it's no longer relevant. cc @connor4312 

It also hides the prompt once the user has taken either action.

https://github.com/user-attachments/assets/e06a3967-95dd-4888-8955-391dd24ea5dc

### DIFF
--- a/src/vs/workbench/contrib/chat/browser/chatElicitationRequestPart.ts
+++ b/src/vs/workbench/contrib/chat/browser/chatElicitationRequestPart.ts
@@ -13,8 +13,8 @@ export class ChatElicitationRequestPart extends Disposable implements IChatElici
 	public state: 'pending' | 'accepted' | 'rejected' = 'pending';
 	public acceptedResult?: Record<string, unknown>;
 
-	private _onDidRequestRemovePart = this._register(new Emitter<void>());
-	public readonly onDidRequestRemovePart = this._onDidRequestRemovePart.event;
+	private _onDidRequestHide = this._register(new Emitter<void>());
+	public readonly onDidRequestHide = this._onDidRequestHide.event;
 
 	constructor(
 		public readonly title: string | IMarkdownString,
@@ -29,7 +29,7 @@ export class ChatElicitationRequestPart extends Disposable implements IChatElici
 	}
 
 	removePart(): void {
-		this._onDidRequestRemovePart.fire();
+		this._onDidRequestHide.fire();
 	}
 
 	public toJSON() {
--- a/src/vs/workbench/contrib/chat/browser/chatElicitationRequestPart.ts
+++ b/src/vs/workbench/contrib/chat/browser/chatElicitationRequestPart.ts
@@ -28,7 +28,7 @@ export class ChatElicitationRequestPart extends Disposable implements IChatElici
 		super();
 	}
 
-	removePart(): void {
+	hide(): void {
 		this._onDidRequestHide.fire();
 	}
 
--- a/src/vs/workbench/contrib/chat/browser/chatElicitationRequestPart.ts
+++ b/src/vs/workbench/contrib/chat/browser/chatElicitationRequestPart.ts
@@ -3,13 +3,18 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
+import { Emitter } from '../../../../base/common/event.js';
 import { IMarkdownString } from '../../../../base/common/htmlContent.js';
+import { Disposable } from '../../../../base/common/lifecycle.js';
 import { IChatElicitationRequest } from '../common/chatService.js';
 
-export class ChatElicitationRequestPart implements IChatElicitationRequest {
+export class ChatElicitationRequestPart extends Disposable implements IChatElicitationRequest {
 	public readonly kind = 'elicitation';
 	public state: 'pending' | 'accepted' | 'rejected' = 'pending';
 	public acceptedResult?: Record<string, unknown>;
+
+	private _onDidRequestRemovePart = this._register(new Emitter<void>());
+	public readonly onDidRequestRemovePart = this._onDidRequestRemovePart.event;
 
 	constructor(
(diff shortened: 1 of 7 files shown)


src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/bufferOutputPolling.ts
```diff
@@ -102,31 +102,34 @@ export async function pollForOutputAndIdle(
...
+			let part: ChatElicitationRequestPart | undefined = undefined;
+			const promise = new Promise<boolean>(resolve => {
+				part = new ChatElicitationRequestPart(
 					new MarkdownString(localize('poll.terminal.waiting', "Continue waiting for `{0}` to finish?", command)),
 					new MarkdownString(localize('poll.terminal.polling', "Copilot will continue to poll for output to determine when the terminal becomes idle for up to 2 minutes.")),
 					'',
 					localize('poll.terminal.accept', 'Yes'),
 					localize('poll.terminal.reject', 'No'),
 					async () => {
+						part!.state = 'accepted';
```
Tyriar: I don't have a nicer pattern for this, but you can avoid the ! by doing this:

```
const thePart = part = new ...
```

TS isn't sure about the `let` since it's used in a callback and could be undefined as a result.

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/runInTerminalTool.ts
```diff
@@ -293,26 +293,41 @@ export class RunInTerminalTool extends Disposable implements IToolImpl {
 				}
 				const execution = new BackgroundTerminalExecution(toolTerminal.instance, xterm, command);
 				RunInTerminalTool._backgroundExecutions.set(termId, execution);
-
 				outputAndIdle = await pollForOutputAndIdle(execution, false, token, this._languageModelsService);
```
Tyriar, line 296: Can we move everything from this line up until before `resultText` is set into a private function?

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/runInTerminalTool.ts
```diff
@@ -293,26 +293,41 @@ export class RunInTerminalTool extends Disposable implements IToolImpl {
...
 					const raceResult = await Promise.race([
-						pollPromise.then(result => ({ type: 'poll', result })),
-						promptPromise.then(result => ({ type: 'prompt', result }))
+						pollPromiseWrapped,
+						promptPromiseWrapped
 					]);
 					if (raceResult.type === 'poll') {
-						// Only assign if result is a poll result
 						outputAndIdle = raceResult.result as typeof outputAndIdle;
+						await part?.removePart();
```
Tyriar: A comment here about how the part could be showing still and we want to automatically handle it for the user would be helpful

Tyriar (dismissed):

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/runInTerminalTool.ts
meganrogge: just did that?

meganrogge: did you see https://github.com/microsoft/vscode/pull/257492/commits/0188a48be4d9f8b03a161ab3cd8af64bf4064f27?

src/vs/workbench/contrib/chat/browser/chatElicitationRequestPart.ts
```diff
@@ -3,14 +3,19 @@
...
 import { IChatElicitationRequest } from '../common/chatService.js';
 
-export class ChatElicitationRequestPart implements IChatElicitationRequest {
+export class ChatElicitationRequestPart extends Disposable implements IChatElicitationRequest {
 	public readonly kind = 'elicitation';
 	public state: 'pending' | 'accepted' | 'rejected' = 'pending';
 	public acceptedResult?: Record<string, unknown>;
 
+	private _onDidRequestHide = this._register(new Emitter<void>());
+	public readonly onDidRequestHide = this._onDidRequestHide.event;
```
Tyriar, line 17: fyi @connor4312

Tyriar (dismissed):

src/vs/workbench/contrib/terminalContrib/chatAgentTools/test/browser/bufferOutputPolling.test.ts
```diff
@@ -0,0 +1,117 @@
...
+		const result = await racePollingOrPrompt(args.pollFn, args.promptFn, args.originalResult, args.token, args.languageModelsService, args.execution);
+		assert.deepEqual(result, args.originalResult);
+	});
+
+	test('should poll again if prompt is accepted', async () => {
+		let extraPollCount = 0;
+		const args = getArgs({
+			pollFn: async () => {
+				extraPollCount++;
+				return { terminalExecutionIdleBeforeTimeout: false, output: 'output', pollDurationMs: 100 };
```
Tyriar: Is it possible to run this with faked timers like this?

https://github.com/microsoft/vscode/blob/0a7fbc80a6369de26e3ef003350cd75806da5bc4/src/vs/base/test/common/async.test.ts#L671

If not, we should rename this to `.integrationTest.ts` as unit tests are expected to execute very fast. The faster the test is the better as this will directly increase CI time, waiting for 100*20 for example isn't ideal.

Tyriar (dismissed):

Tyriar (approved):

PR #257498, opened 2025-07-23
send telemetry when lines are accepted or rejected

amunger:
fix https://github.com/microsoft/vscode-internalbacklog/issues/5680

### DIFF
--- a/src/vs/workbench/contrib/chat/browser/chatEditing/chatEditingTextModelChangeService.ts
+++ b/src/vs/workbench/contrib/chat/browser/chatEditing/chatEditingTextModelChangeService.ts
@@ -101,6 +101,12 @@ export class ChatEditingTextModelChangeService extends Disposable {
 	private readonly _didAcceptOrRejectLines = this._register(new Emitter<acceptedOrRejectedLines>());
 	public readonly onDidAcceptOrRejectLines = this._didAcceptOrRejectLines.event;
 
+	private notifyHunkAction(state: 'accepted' | 'rejected', lineCount: number, hasRemainingEdits: boolean) {
+		if (lineCount > 0) {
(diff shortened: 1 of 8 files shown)


amunger:
It looks like some actions that go through this pipe are sent as telemetry in the core (as in this PR), while others trigger telemetry events in the chat extension [here](https://github.com/microsoft/vscode-copilot-chat/blob/2c81ebc5a58c717a0862fd8f729cf9ea1127611e/src/extension/conversation/vscode-node/userActions.ts#L51). 

All events, including the one added here, are sent over to the EH, so it seems pretty straight forward to shift the telemetry event for this over to the extension if needed.

src/vs/workbench/contrib/chat/browser/chatEditing/chatEditingTextModelChangeService.ts
```diff
@@ -117,6 +123,17 @@ export class ChatEditingTextModelChangeService extends Disposable {
 		this._register(toDisposable(() => {
 			this.clearCurrentEditLineDecoration();
 		}));
+
+		this._register(autorun(r => this.updateLineChangeCount(this._diffInfo.read(r))));
+	}
+
+	private updateLineChangeCount(diff: IDocumentDiff) {
+		this.lineChangeCount = 0;
+		for (const change of diff.changes) {
```
amunger, line 138: I don't completely understand how `movedLines` in the diff relates to `changes`, but I couldn't manage to get a diff calculation result with `movedLines`, so I just didn't worry about them.

hediet: Moved lines will not be set in `diff`, so you don't have to worry about that.

roblourens (approved):

PR #257510, opened 2025-07-23
Remove expandable hover setting

mjbvz:
Seems like there's no reason people would want to disable this at the language level


justschen (approved):

justschen:
@mjbvz looks like ci failure!

Issue #255396, opened 2025-07-11
Investigate run in terminal tool hanging issues


Tyriar:
This is a meta issue for the plan to investigate all issues around hanging in the terminal. For the June release one major case was fixed related to conpty being off https://github.com/microsoft/vscode/issues/252489, but we still get many reports for all OS's.

Some of these will likely be helped with the move of the tool from core as we don't need to rely on the extension API and can touch terminal internals https://github.com/microsoft/vscode/issues/252650

Tyriar:
Previous details from manual testing in https://github.com/microsoft/vscode-copilot/issues/14065#issuecomment-2741375544 back in March

Test matrix below.

- Success means that the output ("run ls in the terminal") was correct exactly in the tool call for rich calls, and correct as expected for basic/none (may include surrounding lines).
- SI sequences are what shows in the terminal hover after running the test

| OS | Shell | Prompt | SI injection | SI quality | SI sequences | Success?
|---|---|---|---|---|---|---|
| macOS | zsh | default | | rich | `P, D, A, B, E, C` | ✅ 
| macOS | zsh | default | disabled | none | N/A | ✅ 
| macOS | zsh | oh-my-zsh | | rich | `P, 7;file://..., D, A, B, E, C` | ✅ 
| macOS | zsh | powerlevel10k | | basic  | `A, B, C, D` | ✅<sup>1</sup>
| macOS | zsh | starship | | rich | `P, D, A, B, H, I, E, C` | ✅ 
| macOS | bash | default | | rich | `P, A, B, E, C, D` | ✅ 
| macOS | bash | default | disabled | none | N/A | ✅ 
| macOS | bash | [WT SI](https://learn.microsoft.com/en-us/windows/terminal/tutorials/shell-integration#bash) | disabled | basic | `D, A, B` | ✅ 
| macOS | bash | starship | | rich | `P, A, B, E, C, D` | ✅ 
| macOS | fish | default | disabled | none | N/A | ✅ 
| macOS | fish | default | | rich | `P, E, C, D, A, B` | ✅<sup>2</sup> 
| macOS | pwsh | default | | rich | `P, A, B, E, C, D` | ✅ 
| macOS | pwsh | default | disabled | none | N/A | ✅ 
| macOS | pwsh | oh-my-posh | | rich | `P, A, B, D, E, C` | ❌<sup>3</sup>
[...]

PR #257517, opened 2025-07-23
Add recorded tests for command detection

Tyriar:
Part of #255396


src/vs/workbench/contrib/terminal/test/browser/xterm/shellIntegrationAddon.integrationTest.ts
```diff
@@ -0,0 +1,240 @@
...
+	},
+	{
+		name: 'basic_macos_zsh_p10k_ls_one_time',
+		events: basic_macos_zsh_p10k_ls_one_time as any as RecordedSessionEvent[],
+		finalAssertions: (commandDetection: ICommandDetectionCapability | undefined) => {
+			// Prompt input model doesn't work for p10k yet
+			// Assert a single command has completed
+			deepStrictEqual(commandDetection!.commands.map(e => e.command), ['']);
+		}
+	},
```
Tyriar, line 107: @meganrogge these tests will be handy to make sure we don't regress in command detection as the data events will replay exactly as they appear in the real world and we can then assert whatever we want on `ICommandDetectionCapability`

lszomoru (approved):

PR #257564, opened 2025-07-24
mcp: implement autorun, rework trust

### LINKED ISSUE #248010: "New tools available" button needs UI/UX polish
Image

There are two UI issues I see with it:

1. The current rendering of a partial button with one regular action and another primary action looks like screen cheese. Also, if we have a primary action in that row of actions it should always be the `Send Message` one.
2. Turns out it's not two actions at all, but just one. 😕 

There's also major UX issue:

- It's not clear to the user why this shows up.
- It's not clear to the user what happens on click.
- Once I click it, it's gone (after an async refresh) and I'm left confused as to what I have just done.

connor4312:
This adds an on-by-default MCP autorun setting, which will start any new
or changed MCP servers when a chat request is sent. When the setting is
on, we will no longer show the infamous 'refresh' indicator. We also
now respond to changes in existing mcp.jsons server definitions as a
signal that we need to refresh the tools.

This also means we need a new take on trust. We don't want a
prompt-injected model to be able to add an MCP server in the workspace
that's silently run on the next chat request. For user level settings,
there's no change -- these are outside the workspace and editing by the
agent is generally disallowed.

For workspace-level servers, users are asked to trust the server the
first time they run it or whenever its definition changes. If they
decline, we won't prompt them again, but they can still manually choose
to run it later on. We also nicely group servers together to avoid a
flurry of prompts. This is what that looks like. In the case of multiple
servers, users can pick ones they wish to trust, or not.

image

This gets most of the way there but I need to do more testing so I will
not merge this yet.

Closes https://github.com/microsoft/vscode/issues/248010

### DIFF
--- a/src/vs/platform/mcp/common/mcpManagement.ts
+++ b/src/vs/platform/mcp/common/mcpManagement.ts
@@ -208,3 +208,10 @@ export interface IAllowedMcpServersService {
 
 export const mcpEnabledConfig = 'chat.mcp.enabled';
 export const mcpGalleryServiceUrlConfig = 'chat.mcp.gallery.serviceUrl';
+export const mcpAutoStartConfig = 'chat.mcp.autostart';
+
+export const enum McpAutoStartValue {
+	Never = 'never',
+	OnlyNew = 'onlyNew',
+	NewAndOutdated = 'newAndOutdated',
+}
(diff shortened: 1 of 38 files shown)


DonJayamanne (dismissed):

DonJayamanne:
@connor4312 you have failures

Tyriar (approved):

PR #257593, opened 2025-07-24
Create by default, but give dropdown for Preview on Github

### LINKED ISSUE #257529: Issue Reporter no longer opens a draft issue
Type: <b>Bug</b>

In older versions of VS Code, the in-product issue reporter (Help > Report Issue) opened a _draft issue_, but now this flow opens just a regular issue, not in a draft state where you can edit before posting. The previous behavior was great because it allowed folks to preview and edit their issue before submitting. The instructions in the in-product issue reporter also state:

> You will be able to edit your issue and add screenshots when we preview it on GitHub.

This is no longer true, as the button no longer opens a draft issue.

VS Code version: Code 1.102.2 (c306e94f98122556ca081f527b466015e1bc37b0, 2025-07-22T12:15:48.520Z)
OS version: Darwin arm64 24.5.0
Modes:

<details>

|Item|Value|
|---|---|
|CPUs|Apple M1 (8 x 2400)|
|GPU Status|2d_canvas: enabled<br>canvas_oop_rasterization: enabled_on<br>direct_rendering_display_compositor: disabled_off_ok<br>gpu_compositing: enabled<br>multiple_raster_threads: enabled_on<br>opengl: enabled_on<br>rasterization: enabled<br>raw_draw: disabled_off_ok<br>skia_graphite: enabled_on<br>video_decode: enabled<br>video_encode: enabled<br>webgl: enabled<br>webgl2: enabled<br>webgpu: enabled<br>webnn: disabled_off|
|Load (avg)|2, 3, 5|
|Memory (System)|16.00GB (0.38GB free)|
|Process Argv|--crash-reporter-id 525192d5-314f-459b-ac86-bdf65193e863|
|Screen Reader|no|
|VM|0%|
</details><details><summary>Extensions (44)</summary>

---|---|---
ruff|cha|2025.24.0
vscode-eslint|dba|3.0.10
docker|doc|0.13.0
python-environment-manager|don|1.2.7
gitlens|eam|17.3.2
EditorConfig|Edi|0.17.4
prettier-vscode|esb|11.0.0
copilot|Git|1.346.0
copilot-chat|Git|0.29.1
vscode-github-actions|git|0.27.2
vscode-pull-request-github|Git|0.114.3
vscode-pr-pinger|jri|0.0.6
vscode-format-context-menu|lac|1.0.4
vscode-containers|ms-|2.1.0
vscode-docker|ms-|2.0.0
pyright|ms-|1.1.403
black-formatter|ms-|2025.2.0
debugpy|ms-|2025.10.0
flake8|ms-|2025.2.0
isort|ms-|2025.0.0
python|ms-|2025.10.1
vscode-pylance|ms-|2025.6.2
vscode-python-envs|ms-|1.0.0
datawrangler|ms-|1.22.0
jupyter|ms-|2025.6.0
jupyter-keymap|ms-|1.1.2
jupyter-renderers|ms-|1.3.0
vscode-jupyter-cell-tags|ms-|0.1.9
vscode-jupyter-slideshow|ms-|0.1.6
remote-containers|ms-|0.422.1
extension-test-runner|ms-|0.0.12
vscode-github-issue-notebooks|ms-|0.0.133
vscode-selfhost-test-provider|ms-|0.3.25
indent-rainbow|ode|8.3.1
air-vscode|Pos|0.14.0
shiny|Pos|1.3.0
quarto|qua|1.123.0
r|REd|2.8.6
r-syntax|REd|0.1.3
rust-analyzer|rus|0.3.2547
even-better-toml|tam|0.21.2
vscode-lldb|vad|1.11.5
vscode-selfhost-test-provider|ms-|0.4.0
vscode-selfhost-import-aid|ms-|0.0.1

(2 theme extensions excluded)

</details><details>

```
vsliv368cf:30146710
vswsl492cf:30256860
binariesv615:30325510
14424c2-chatv6:31345185
2e7ec940:31000449
nativeloc1:31344060
dwcopilot:31170013
6074i472:31201624
dwoutputs:31242946
hdaa2157:31222309
copilot_t_ci:31333650
e5gg6876:31282496
pythoneinst12:31285622
c7cif404:31314491
pythonpulldiag:31343502
996jf627:31283433
pythonrdcb7:31342333
usemplatestapi:31297334
0aa6g176:31307128
747dc170:31275177
6518g693:31334701
aj953862:31281341
generatesymbolt:31295002
convertfstringf:31295003
pylancequickfixf:31350060
9d2cg352:31346308
usemarketplace:31343026
nesew2to5:31336538
agentclaude:31335814
nes-diff-11:31337487
replacestringexc:31350595
nes-set-on:31340697
6abeh943:31336334
envsactivate1:31353494
yijiwantestdri0626-c:31336931
0927b901:31350571
ji9b5146:31348712
45650338:31351948
0cj2b977:31352657

```

</details>

### LINKED ISSUE #260182: Issue Reporter -- Add Create/Preview option + repo routing choices
Looking into enhancements for the native issue reporter within vscode. There have been changes within the last ~6mo that changed the default behavior from "Preview" to "Create", leading to more issues created at a lower quality bar. While the classifier bot can take care of a portion of this load, there is still feedback that asks explicitly for "Preview" as the default behavior for ease of adding screenshots etc.

- [x] 🏃 Add dropdown option for issue reporter to allow users to choose create vs preview. (default: preview) https://github.com/microsoft/vscode/issues/257529
  - https://github.com/microsoft/vscode/pull/257593 
- [ ] 🏃 Add optional routing UX if specific users need to file an issue against a private repository
  - https://github.com/microsoft/vscode/tree/milively/internal-reporting

--- 
Notes:
- Look into closed issues labeled verified from prior couple milestones, identifying whether or not them come from issue reporter or gh website. Should be identifiable based on body content at the end, and markdown comments from the template.
  - Should give an indication for where the more "useful" issues come from
  - Maybe look at closed issues as not planned and/or `spam` to maybe see where users are sending those from
- Reach out to raymond and harald about potential issues re: org id's and repo destinations

justschen:
fix https://github.com/microsoft/vscode/issues/257529

cc. @Yoyokrazy


Yoyokrazy:
@bpasero would appreciate a review on this one, since I had to make a small change or two to the contextmenuService. The issue was that without changing how we computed the anchor position, it would anchor itself well below the dropdown button when zoomed in, regardless of whether or not it was on screen.

The change was 
```
const elementPosition = dom.getDomNodePagePosition(anchor);
``` 
to 
```
const clientRect = anchor.getBoundingClientRect();
const elementPosition = { left: clientRect.left, top: clientRect.top, width: clientRect.width, height: clientRect.height };
```

I've checked as many context menus throughout the product as I could find, but not sure if there could be an edge case I'm missing.

bpasero (dismissed):
@justschen @Yoyokrazy one thing to note is that we have a context menu implementation for `browser` that implements a similar logic, and now with this change the two implementations differ. The one for `browser` is:

https://github.com/microsoft/vscode/blob/4d917a1ab503ea29c857e6ab05e3c931e1604c70/src/vs/base/browser/ui/contextview/contextview.ts#L278

It is also possible for a user on desktop to enable custom context menus (something @benibenj added recently).

Can we see for a simpler fix if possible specifically for the zooming issue (which seems to be the real issue here) and/or understand if it impacts `browser` as well or differently?

Yoyokrazy:
> Can we see for a simpler fix if possible specifically for the zooming issue (which seems to be the real issue here) and/or understand if it impacts browser as well or differently?

@bpasero I'll see if there's any implications with browser, however I do want to clarify that it's actually not an issue with zoom. It's just an issue with where the button is positioned in the DOM vs the viewport. Zooming into the issue reporter just makes the issue more apparent since it pushes the button off the form. 

To test this, I reverted all changes to the contextMenu service, and then changed the issue reporter window size from `700x800` to `700x300`. Here's what we end up getting with how the context menu anchor is currently computed:

image

and then when I simply expand the issue reporter window manually, we can see that it was "correctly" anchored, just on the initial DOM position. 

image

I'm not an expert with the front end work involved in a lot of the DOM structure and the differences with this and the browser implementation but I'll see if it's effected. Could be that we can bring this logic over to it as well, unsure. The clipping logic seems like something we could/should use regardless of how we want to get the anchor position (via either `DOM.getDomNodePagePosition(anchor)` or my proposed `anchor.getBoundingClientRect();` + rect position/width/height)

Maybe the deeper issue is that the we don't recompute the anchor position as the window resizes. I know a lot of this code has been around for quite a while, so not sure how much relies on the nuance of the anchor hueristics we have right now.

I did notice that I couldn't really find any buttons with dropdowns that could be scrolled off the screen, so I wonder if this could've just been us exposing an edge case that we've never really had any reason to see until now.

Yoyokrazy:
Tested all reporting action flows (preview + create) on native and custom context menus, windows + macOS as well. Changes to contextView and contextMenuService necessary now that we have a case where we can scroll a dropdown context menu off screen (calculations hadn't been built out to do viewport relative calculations and scroll offsets)

rebornix (approved):

PR #257691, opened 2025-07-24
add basic regex auto-reply support

meganrogge:
fixes https://github.com/microsoft/vscode-copilot/issues/16430

Have a new issue  #257726 to track:
- multiple prompts (include this in output polling instead of after)
- support for foreground terminals
- delayed prompts
- ask the llm to determine options vs using regex
- consider automatically replying

### DIFF
--- a/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/bufferOutputPolling.ts
+++ b/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/bufferOutputPolling.ts
@@ -184,6 +184,40 @@ export function promptForMorePolling(command: string, context: IToolInvocationCo
 	return { promise: Promise.resolve(false) };
 }
 
+
+export function promptForYesNo(context: IToolInvocationContext, chatService: IChatService): { promise: Promise<boolean>; part?: ChatElicitationRequestPart } {
+	const chatModel = chatService.getSession(context.sessionId);
+	if (chatModel instanceof ChatModel) {
+		const request = chatModel.getRequests().at(-1);
+		if (request) {
+			let part: ChatElicitationRequestPart | undefined = undefined;
+			const promise = new Promise<boolean>(resolve => {
+				const thePart = part = new ChatElicitationRequestPart(
+					new MarkdownString(localize('poll.terminal.yes', 'Respond yes in the terminal?')),
+					new MarkdownString(localize('poll.terminal.yesNo', 'Copilot will run the reply in the terminal.')),
+					'',
+					localize('poll.terminal.accept', 'Yes'),
+					localize('poll.terminal.reject', 'No'),
+					async () => {
+						thePart.state = 'accepted';
+						thePart.hide();
+						resolve(true);
+					},
+					async () => {
+						thePart.state = 'rejected';
+						thePart.hide();
+						resolve(false);
+					}
+				);
+				chatModel.acceptResponseProgress(request, thePart);
+			});
+			return { promise, part };
+		}
+	}
+	return { promise: Promise.resolve(false) };
+}
+
+
 export async function assessOutputForErrors(buffer: string, token: CancellationToken, languageModelsService: ILanguageModelsService): Promise<string> {
 	const models = await languageModelsService.selectLanguageModels({ vendor: 'copilot', family: 'gpt-4o-mini' });
 	if (!models.length) {
@@ -228,24 +262,22 @@ export function getExpectedUserInputKind(output: string): string | undefined {
 	if (!output) {
 		return undefined;
 	}
-	const lines = output.split(/\r?\n/);
-	const lastLine = lines[lines.length - 1].trim();
 
 	const patterns: { regex: RegExp; kind: string }[] = [
-		{ regex: /\bdo you want to continue\b.*\(y\/n\)/i, kind: 'y/n' },
-		{ regex: /\bcontinue\? \(y\/n\)/i, kind: 'y/n' },
-		{ regex: /\bproceed\? \(y\/n\)/i, kind: 'y/n' },
-		{ regex: /\btype yes or no\b/i, kind: 'yes/no' },
-		{ regex: /\b\(yes\/no\)/i, kind: 'yes/no' },
-		{ regex: /\b\[y\/n\]/i, kind: 'y/n' },
-		{ regex: /\benter your choice\b/i, kind: 'choice' },
-		{ regex: /\bselect an option\b/i, kind: 'choice' },
-		{ regex: /\bplease respond\b/i, kind: 'response' },
-		{ regex: /\bpress (enter|return|any key)\b/i, kind: 'press key' }
+		{ regex: /\bdo you want to continue\b.*\(y\/n\)/ig, kind: 'y/n' },
+		{ regex: /\bcontinue\? \(y\/n\)/ig, kind: 'y/n' },
+		{ regex: /\bproceed\? \(y\/n\)/ig, kind: 'y/n' },
+		{ regex: /\btype yes or no\b/ig, kind: 'yes/no' },
+		{ regex: /\b\(yes\/no\)/ig, kind: 'yes/no' },
+		{ regex: /\b\[y\/n\]/ig, kind: 'y/n' },
+		{ regex: /\benter your choice\b/ig, kind: 'choice' },
+		{ regex: /\bselect an option\b/ig, kind: 'choice' },
+		{ regex: /\bplease respond\b/ig, kind: 'response' },
+		{ regex: /\bpress (enter|return|any key)\b/ig, kind: 'press key' }
 	];
 
 	for (const { regex, kind } of patterns) {
-		if (regex.test(lastLine)) {
+		if (regex.test(output)) {
 			return kind;
 		}
 	}
--- a/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/bufferOutputPolling.ts
+++ b/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/bufferOutputPolling.ts
@@ -11,7 +11,7 @@ import { ExtensionIdentifier } from '../../../../../platform/extensions/common/e
 import { ChatElicitationRequestPart } from '../../../chat/browser/chatElicitationRequestPart.js';
 import { IChatService } from '../../../chat/common/chatService.js';
 import { ChatMessageRole, ILanguageModelsService } from '../../../chat/common/languageModels.js';
-import { IToolInvocationContext } from '../../../chat/common/languageModelToolsService.js';
+import { IToolInvocationContext, IToolResult } from '../../../chat/common/languageModelToolsService.js';
 import { promptForYesNo } from '../../../elicitation/browser/elicitation.js';
 import { ITerminalInstance } from '../../../terminal/browser/terminal.js';
 import type { IMarker as IXtermMarker } from '@xterm/xterm';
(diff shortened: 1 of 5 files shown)


src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/bufferOutputPolling.ts
```diff
@@ -215,3 +240,37 @@ export async function assessOutputForErrors(buffer: string, token: CancellationT
...
+		{ regex: /\bcontinue\? \(y\/n\)/ig, kind: 'y/n' },
+		{ regex: /\bproceed\? \(y\/n\)/ig, kind: 'y/n' },
+		{ regex: /\btype yes or no\b/ig, kind: 'yes/no' },
+		{ regex: /\b\(yes\/no\)/ig, kind: 'yes/no' },
+		{ regex: /\b\[y\/n\]/ig, kind: 'y/n' },
+		{ regex: /\benter your choice\b/ig, kind: 'choice' },
+		{ regex: /\bselect an option\b/ig, kind: 'choice' },
+		{ regex: /\bplease respond\b/ig, kind: 'response' },
+		{ regex: /\bpress (enter|return|any key)\b/ig, kind: 'press key' }
+	];
```
Tyriar: Did you have programs that output these? We should include the programs in comments if so, so we know the source.

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/bufferOutputPolling.ts
```diff
@@ -215,3 +240,37 @@ export async function assessOutputForErrors(buffer: string, token: CancellationT
...
+ * Returns the kind of input expected by the last line of the output if it appears to be a user prompt.
+ * For example, returns 'y/n', 'yes/no', 'press key', 'choice', etc., or undefined if not a prompt.
+ */
+export function getExpectedUserInputKind(output: string): string | undefined {
+	if (!output) {
+		return undefined;
+	}
+
+	const patterns: { regex: RegExp; kind: string }[] = [
+		{ regex: /\bdo you want to continue\b.*\(y\/n\)/ig, kind: 'y/n' },
```
Tyriar: If we idle briefly with one of these matches, we could do more general ones like '/\(y\/n\):?\s*$/', checking the end of the line is important too.

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/bufferOutputPolling.ts
```diff
@@ -215,3 +240,37 @@ export async function assessOutputForErrors(buffer: string, token: CancellationT
...
+/**
+ * Returns the kind of input expected by the last line of the output if it appears to be a user prompt.
+ * For example, returns 'y/n', 'yes/no', 'press key', 'choice', etc., or undefined if not a prompt.
+ */
+export function getExpectedUserInputKind(output: string): string | undefined {
+	if (!output) {
+		return undefined;
+	}
+
+	const patterns: { regex: RegExp; kind: string }[] = [
```
Tyriar: This is pwsh's confirm impact one, not sure if there are slight variants but this was from trying to `remove-item folder`.

```
[Y] Yes  [A] Yes to All  [N] No  [L] No to All  [S] Suspend  [?] Help (default is "Y"): 
```

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/bufferOutputPolling.ts
```diff
@@ -184,6 +190,25 @@ export function promptForMorePolling(command: string, context: IToolInvocationCo
...
+}
+
+export function promptForYesNo(context: IToolInvocationContext, chatService: IChatService): { promise: Promise<boolean>; part?: ChatElicitationRequestPart } {
+	return createYesNoPrompt(
+		context,
+		chatService,
+		localize('poll.terminal.yes', 'Respond yes in the terminal?'),
+		localize('poll.terminal.yesNo', 'Copilot will run the reply in the terminal.')
+	);
+}
```
Tyriar: Should these live in ./elicitation.ts?

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/bufferOutputPolling.ts
```diff
@@ -184,6 +190,25 @@ export function promptForMorePolling(command: string, context: IToolInvocationCo
...
+		title,
+		message
+	);
+}
+
+export function promptForYesNo(context: IToolInvocationContext, chatService: IChatService): { promise: Promise<boolean>; part?: ChatElicitationRequestPart } {
+	return createYesNoPrompt(
+		context,
+		chatService,
+		localize('poll.terminal.yes', 'Respond yes in the terminal?'),
```
Tyriar: We could be more specific:

- Respond "yes" in the terminal?
- Respond "y" in the terminal?

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/runInTerminalTool.ts
```diff
@@ -297,14 +297,35 @@ export class RunInTerminalTool extends Disposable implements IToolImpl {
...
 						outputAndIdle,
 						token,
 						this._languageModelsService,
 						execution
 					);
 				}
-
+				const userInputKind = await getExpectedUserInputKind(outputAndIdle.output);
+				if (userInputKind) {
+					if (userInputKind !== 'choice' && userInputKind !== 'key') {
```
Tyriar: Can this be reused between tools?

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/runInTerminalTool.ts
```diff
@@ -297,14 +297,35 @@ export class RunInTerminalTool extends Disposable implements IToolImpl {
...
-						() => promptForMorePolling(command, invocation.context!, this._chatService),
+						() => promptForMorePolling(localize('poll.terminal.waiting', "Continue waiting for `{0}` to finish?", command), localize('poll.terminal.polling', "Copilot will continue to poll for output to determine when the terminal becomes idle for up to 2 minutes."), invocation.context!, this._chatService),
 						outputAndIdle,
 						token,
 						this._languageModelsService,
 						execution
 					);
 				}
-
+				const userInputKind = await getExpectedUserInputKind(outputAndIdle.output);
```
Tyriar: This could happen multiple times and also maybe only after some period of time has passed, I think it needs to live inside the polling loop which should be doing general evaluation of the state instead of just detecting idle.

Tyriar: We want this for foreground terminals too, currently I think they'll just hang and keep spinning until the user responds.

Tyriar (changes requested):

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/runInTerminalTool.ts
meganrogge: yes, pushed that

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/task/createAndRunTaskTool.ts
```diff
@@ -115,13 +115,30 @@ export class CreateAndRunTaskTool implements IToolImpl {
...
+				async () => await pollForOutputAndIdle({ getOutput: () => getOutput(terminal), isActive: () => this._isTaskActive(task) }, true, token, this._languageModelsService),
+			);
+			if (handleResult.handled) {
+				if (handleResult.outputAndIdle) {
+					outputAndIdle = handleResult.outputAndIdle;
+				}
+			} else {
+				return { content: [{ kind: 'text', value: `The task is still running and requires user input of ${userInputKind}.` }], toolResultMessage: new MarkdownString(localize('copilotChat.taskRequiresUserInput', 'The task `{0}` is still running and requires user input. {1}', task._label, userInputKind)) };
+			}
+		}
```
Tyriar: Sharing this between the 3 tools would be good. This could live in elicitation.ts with the prompt calls too?

Tyriar (dismissed):

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/runInTerminalTool.ts
meganrogge: Deferring

meganrogge: Deferring

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/task/createAndRunTaskTool.ts
meganrogge: can't put that in elicitation.ts as there are very specific terminal and task things in this function like outputAndIdle, but moved to bufferOutputPolling

meganrogge: done though in terms of consolidating them

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/bufferOutputPolling.ts
```diff
@@ -152,38 +152,43 @@ export async function pollForOutputAndIdle(
...
+export async function handleTerminalUserInputPrompt(
+	outputAndIdle: { output: string; terminalExecutionIdleBeforeTimeout: boolean; pollDurationMs?: number; modelOutputEvalResponse?: string },
+	invocation: any,
+	chatService: IChatService,
+	terminal: ITerminalInstance,
+	executableLabel: string,
+	token: CancellationToken,
+	languageModelsService: ILanguageModelsService
+): Promise<{ handled: boolean; userResponse?: 'accepted' | 'rejected'; outputAndIdle: { output: string; terminalExecutionIdleBeforeTimeout: boolean; pollDurationMs?: number; modelOutputEvalResponse?: string }; message?: IToolResult }> {
+	const userInputKind = await getExpectedUserInputKind(outputAndIdle.output);
```
Tyriar, line 164:
```suggestion
	const userInputKind = getExpectedUserInputKind(outputAndIdle.output);
```

Tyriar (changes requested):
Pushed https://github.com/microsoft/vscode/pull/257691/commits/98e45b3c6539870bbf2042fcd60a2b0c7df44454 to add the events we want to collect for this. From my testing the `userResponse` variable doesn't actually work currently.

When I pulled the code I also found it pretty hard to follow. I think it's a little late to push something so large that has the potential to hang the tool due to async logic, especially without any tests.

Since it's an edge case (bg terminals), doesn't handle multiple questions and just does basic regex, I recommend waiting and implementing it in a more robust way next month. It's not worth the risk considering how little this feature actually covers in its current state. What we need is:

- Some output monitoring class that `RunInTerminalTool` hands ownership of the execution to and then it has a loop that does all the idle/question polling and handling in a central place (not inside `RunInTerminalTool.invoke`)
- This component must have a good set of unit tests
- It should return a well structured return object as `handleTerminalUserInputPrompt`'s return result is quite messy right now.

If anything the hint with an a11y alert as I suggested earlier is the safer approach as it's just adding a low risk hint to tell the user the terminal is awaiting input.

meganrogge:
too many conflicts, starting from scratch

PR #257785, opened 2025-07-25
make sure to resize images coming from tool calls

justschen:
ref https://github.com/microsoft/vscode-copilot-issues/issues/113


src/vs/platform/imageResize/electron-main/imageResizeMainService.ts
```diff
@@ -0,0 +1,70 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { IImageResizeService } from '../common/imageResizeService.js';
+import { nativeImage } from 'electron';
+
+export class ImageResizeMainService implements IImageResizeService {
```
connor4312: Curious why we're doing this in electron sandbox rather than the browser as we already have a `resizeImage` implementation that runs in the browser.

If we implement it like this we'll still need an additional implementation in the browser layer to not regress MCP functionality. If we have it in a browser, I'm not sure we need an electron version of it unless the electron version does things the browser version doesn't.

connor4312 (changes requested):

src/vs/workbench/contrib/chat/browser/chatContentParts/chatToolInputOutputContentPart.ts
```diff
@@ -199,29 +200,32 @@ export class ChatCollapsibleInputOutputContentPart extends Disposable {
...
 		const el = dom.h('.chat-collapsible-io-resource-group', [
 			dom.h('.chat-collapsible-io-resource-items@items'),
 			dom.h('.chat-collapsible-io-resource-actions@actions'),
 		]);
 
-		const entries = parts.map((part): IChatRequestVariableEntry => {
+		const entries = parts.map(async (part): Promise<IChatRequestVariableEntry> => {
 			if (part.mimeType && getAttachableImageExtension(part.mimeType)) {
-				return { kind: 'image', id: generateUuid(), name: basename(part.uri), value: part.value, mimeType: part.mimeType, isURL: false, references: [{ kind: 'reference', reference: part.uri }] };
+				const resized = part.value && await resizeImage(part.value).catch(() => part.value);
```
connor4312: The stuff here is purely display. Do we still need to resize at this point?

src/vs/workbench/contrib/mcp/common/mcpLanguageModelToolContribution.ts
```diff
@@ -222,14 +224,34 @@ class McpToolImplementation implements IToolImpl {
 				}
 			}
 
-			// Rewrite image rsources to images so they are inlined nicely
-			const addAsInlineData = (mimeType: string, value: string, uri?: URI) => {
+			// Rewrite image resources to images so they are inlined nicely
+			const addAsInlineData = async (mimeType: string, value: string, uri?: URI): Promise<VSBuffer | void> => {
```
connor4312, line 228: The usages of this need to be awaited

src/vs/platform/imageResize/browser/imageResizeService.ts
```diff
@@ -0,0 +1,170 @@
...
+				URL.revokeObjectURL(url);
+				reject(error);
+			};
+		});
+	}
+
+	convertStringToUInt8Array(data: string): Uint8Array {
+		const base64Data = data.includes(',') ? data.split(',')[1] : data;
+		if (this.isValidBase64(base64Data)) {
+			return Uint8Array.from(atob(base64Data), char => char.charCodeAt(0));
```
connor4312: The `decodeBase64` function we have is faster here

src/vs/workbench/contrib/mcp/common/mcpLanguageModelToolContribution.ts
```diff
@@ -222,14 +224,34 @@ class McpToolImplementation implements IToolImpl {
...
 				details.output.push({ type: 'embed', mimeType, value, uri });
 				if (isForModel) {
-					result.content.push({
-						kind: 'data',
-						value: { mimeType, data: decodeBase64(value) }
-					});
+					let finalData: VSBuffer;
```
connor4312, line 231: The `value` here is always going to be a base64 string, not a URI, so you can probably pass it directly into `this._imageResizeService.resizeImage(decodeBase64(value).buffer, mimeType)`

src/vs/platform/imageResize/browser/imageResizeService.ts
```diff
@@ -0,0 +1,170 @@
...
+		}
+		return new TextEncoder().encode(data);
+	}
+
+	// Only used for URLs
+	convertUint8ArrayToString(data: Uint8Array): string {
```
connor4312, line 97: Seems unused

src/vs/platform/imageResize/browser/imageResizeService.ts
```diff
@@ -0,0 +1,170 @@
...
+		} catch {
+			return '';
+		}
+	}
+
+	isValidBase64(str: string): boolean {
+		// checks if the string is a valid base64 string that is NOT encoded
+		return /^[A-Za-z0-9+/]*={0,2}$/.test(str) && (() => {
+			try {
+				atob(str);
```
connor4312: We decode data twice here. If you use `decodeBase64`, it'll throw for you when if it sees an invalid character

src/vs/platform/imageResize/browser/imageResizeService.ts
```diff
@@ -0,0 +1,170 @@
...
+				return true;
+			} catch {
+				return false;
+			}
+		})();
+	}
+
+	async createFileForMedia(fileService: IFileService, imagesFolder: URI, dataTransfer: Uint8Array, mimeType: string): Promise<URI | undefined> {
```
connor4312, line 116: This method and the three below seem unused -- are they just there for the future?

src/vs/platform/imageResize/browser/imageResizeService.ts
justschen: leaving this and the below for future proof ! (when we use this to replace imageUtils, we'll need this)

src/vs/workbench/contrib/mcp/common/mcpLanguageModelToolContribution.ts
```diff
@@ -222,14 +224,27 @@ class McpToolImplementation implements IToolImpl {
...
+					try {
+						const serializedData = await this._imageResizeService.resizeImage(decodeBase64(value).buffer, mimeType);
+						finalData = decodeBase64(value);
+						if (serializedData instanceof Uint8Array) {
+							const resizedData = new Uint8Array(Object.values(serializedData));
+							const resized = VSBuffer.wrap(resizedData);
+							if (resized.byteLength > 0) {
+								finalData = resized;
+							}
+						}
```
connor4312: `resizeImage` always returns an Uint8Array, so you should just be able to 

```suggestion
						const resized = await this._imageResizeService.resizeImage(decodeBase64(value).buffer, mimeType);
						finalData = VSBuffer.wrap(resized);
```

src/vs/workbench/contrib/chat/browser/chatContentParts/chatToolInputOutputContentPart.ts
```diff
@@ -199,29 +199,31 @@ export class ChatCollapsibleInputOutputContentPart extends Disposable {
 		return contents.root;
 	}
 
-	private addResourceGroup(parts: IChatCollapsibleIODataPart[], container: HTMLElement) {
+	private async addResourceGroup(parts: IChatCollapsibleIODataPart[], container: HTMLElement) {
```
connor4312: I think you no longer need changes in this file, `git checkout main src/vs/workbench/contrib/chat/browser/chatContentParts/chatToolInputOutputContentPart.ts` to reset it

justschen: forgot to revert this yeah 😶‍🌫️

src/vs/workbench/contrib/mcp/common/mcpLanguageModelToolContribution.ts
justschen: updated, but wrapped it in try/catch instead 👍🏻

connor4312 (approved):

PR #257892, opened 2025-07-25
Implement copilot token flag for agent mode

roblourens:

### DIFF
--- a/src/vs/workbench/contrib/chat/browser/chat.contribution.ts
+++ b/src/vs/workbench/contrib/chat/browser/chat.contribution.ts
@@ -385,6 +385,7 @@ configurationRegistry.registerConfiguration({
 			policy: {
 				name: 'ChatAgentMode',
 				minimumVersion: '1.99',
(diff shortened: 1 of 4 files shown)


src/vs/workbench/contrib/chat/browser/chat.contribution.ts
```diff
@@ -385,6 +385,7 @@ configurationRegistry.registerConfiguration({
 			policy: {
 				name: 'ChatAgentMode',
 				minimumVersion: '1.99',
```
joshspicer, line 387: This changes looks good and will work 👍 

If you wanted the policy to set a value other than `false`, you could use `defaultValue`. 

`false` is the default though, so this works fine as-is:

https://github.com/microsoft/vscode/blob/01b30db2571c523839c30dfa2d437c6c1a78f749/src/vs/workbench/services/policies/common/accountPolicyService.ts#L64

rzhao271 (approved):

PR #257970, opened 2025-07-26
fix: The drag range of the debugToolBar has been changed to be applic…

huyunan:
The dragging range of the debugToolBar is very limited and uncomfortable. For instance, when searching for content, it is difficult to move it away as it cannot be done. So I changed it so that it can be dragged within the VSCode window as well. Setting a limit will not bring up a window. The processing has been done for the initial loading of the window, its closing, and changes in window size. And the test has been conducted and there were no issues found. Only tested in the Windows environment. I assume there will be no issues in other environments. Hope to be able to do it.
Before the modification:
PixPin_2025-07-26_20-45-57
After the modification:
PixPin_2025-07-26_20-47-21


huyunan:

connor4312:
Thank you for the contribution! However, this is a UI pattern that I don't think we want to change at the moment. We've not heard this as a problematic behavior from other users, and restricting the free-dragging range of the debug toolbar avoids it getting 'lost' amongst the UI. 

To avoid wasting effort in the future, it is often better to create an issue where we can discuss or respond first prior to creating a PR. Again, we appreciate you taking the time to make this contribution and hope to see you in the future 🙂

PR #258119, opened 2025-07-28
fix(git): prevent deleting currently active worktree

MengAiDev:
- Add a check to prevent users from deleting the worktree they are currently in
- Display an error message if the user tries to delete the active worktree
- Suggest switching to another worktree or the main repository before deleting


nguyenchristy:
Thank you for your contribution! I believe that the fix should be within the `deleteWorktree` command rather than `createWorktree` command . Also, we don't have to search for a match but we can just check if the main repo exists within the `deleteWorktree` command and send that error message. https://github.com/microsoft/vscode/pull/258297

PR #258129, opened 2025-07-28
Fix action view items being too long

sshockwave:
The filter input textbox might take up large spaces despite the screen being wide enough.
If we hit "PROBLEMS" then "OUTPUT", the behavior is normal:
problems_to_output
However, if we hit "DEBUG CONSOLE" first and then "OUTPUT", all action buttons are hidden in the ellipsis:
debug_to_output
This is an inconsistent behavior that may reduce productivity.

The cause of this issue is that the available width of the left toolbar depends on the width of the right toolbar, which is then computed from `clientWidth`, i.e., the current width. In "DEBUG CONSOLE", the filter input box is stretching (`flex: 1`) to occupy the remaining space, and this space is not freed when switching to the "OUTPUT" tab.

The proposed change is to add `getMinWidth()` to supported components like textboxes and select dropdowns. When calculating their widths in layout, use their minimum width first so that other elements can take up more space. After that, they can grow to their desired length in the final rendering.


sshockwave:

Issue #249479, opened 2025-05-21
Support multiple LSP based completion provider


anthonykim1:
With https://github.com/microsoft/vscode/pull/243528 providing support for Python LSP provider in terminal suggest, we should continue making effort towards supporting more lsp completion provider.

Easiest would be starting from pwsh, then potentially for node, julia, etc.

anthonykim1:
Moving this out of milestone until https://github.com/microsoft/vscode/pull/258137#pullrequestreview-3098177976

anderseknert:
Hi there! I was playing with this yesterday and I like it a lot 😃 Would love if there was a way for our [language server](https://github.com/open-policy-agent/regal) to help provide completions in the terminal!

PR #258137, opened 2025-07-28
Remove hard-coded Python for lsp in terminal

### LINKED ISSUE #257343: Remove hard-coded Python REPL related code from lsp completion
Remove hard-coded things that are added for Python and REPL. So that lsp completion related files are more generic and ready for https://github.com/microsoft/vscode/issues/249479

anthonykim1:
Resolves: https://github.com/microsoft/vscode/issues/257343
in order to get ready for 
https://github.com/microsoft/vscode/issues/249479

This PR intends to achieve: 

Debts:
- Need to remove bunch of hard-coded Python related stuff for lsp in terminal

---------------------
Above is needed so that we can support pwsh completion via lspCompletionProviderAddon https://github.com/microsoft/vscode/issues/249479

Items below should be on separate PR(s): 
- Figure out way to disable pwsh completion provider if lsp is CHANGED TO TRUE. If both are enabled, weird behavior happens currently. 
- lsp Should support pwsh, so we can eventually get rid of pwshCompletionAddon.ts
- Need to handle `rawCompletionToITerminalCompletion` in lsp. Probably differs with `createCompletionItemPython`
- Make sure Trigger character is appropriate for that provider.


anthonykim1:
Should I be defaulting lsp providers to true or false? @Tyriar 

Future code pointer: https://github.com/microsoft/vscode/blob/2c804a60266ce4f499599b1171b245a7c4386c88/src/vs/workbench/contrib/terminalContrib/suggest/common/terminalSuggestConfiguration.ts#L220-L225

src/vs/workbench/contrib/terminalContrib/suggest/browser/lspTerminalUtil.ts
```diff
@@ -3,6 +3,49 @@
...
+export const LspSupportedShellTypes: TerminalShellType[] = [
+	GeneralShellType.Python,
+	GeneralShellType.PowerShell
+];
+
+export const LspSupportedProvider: string[] = [
+	PYLANCE_DEBUG_DISPLAY_NAME,
+	NEW_PYLANCE_DEBUG_DISPLAY_NAME,
+	PWSH_DEBUG_DISPLAY_NAME
+];
```
Tyriar: Is there a downside to just open this up to everything?

Tyriar (changes requested):

Tyriar:
> Should I be defaulting lsp providers to true or false? @Tyriar

We should make it true if insiders until we verify it doesn't have major impacts. For example is the LSP provider was significantly slower than the others and managed to affect regular shells (pwsh or others), it would delay completions for all others.

src/vs/workbench/contrib/terminalContrib/suggest/browser/lspTerminalUtil.ts
anthonykim1: Good point, we could open up for all.

However, 
my concern was unexpected future providers failing to be mapped since at least for Python vs. Pwsh the mapping of completion item to terminal completion seemed pretty different:

comparing: https://github.com/microsoft/vscode/blob/16992b21c02dad23a3748022559aeed1f05e80cf/src/vs/workbench/contrib/terminalContrib/suggest/browser/lspCompletionProviderAddon.ts#L86 
with 
https://github.com/microsoft/vscode/blob/16992b21c02dad23a3748022559aeed1f05e80cf/src/vs/workbench/contrib/terminalContrib/suggest/browser/pwshCompletionProviderAddon.ts#L237

src/vs/workbench/contrib/terminalContrib/suggest/browser/terminal.suggest.contribution.ts
```diff
@@ -54,13 +54,14 @@ class TerminalSuggestContribution extends DisposableStore implements ITerminalCo
 
 	private readonly _addon: MutableDisposable<SuggestAddon> = new MutableDisposable();
 	private readonly _pwshAddon: MutableDisposable<PwshCompletionProviderAddon> = new MutableDisposable();
-	private readonly _lspAddon: MutableDisposable<LspCompletionProviderAddon> = new MutableDisposable(); // TODO: Support multiple lspAddons in the future.
+	private readonly _lspAddons: MutableDisposable<DisposableStore> = new MutableDisposable();
```
anthonykim1: I feel like I might be doing this wrong. How should I be tracking multiple lspAddons? @Tyriar 

Not sure if any of `_lspAddons` and `lspAddonInstances` are correct and/or need both🧐

```
private readonly _lspAddons: MutableDisposable<DisposableStore> = new MutableDisposable();
private _lspAddonInstances: LspCompletionProviderAddon[] = [];
```

What I'd want to achieve is have an array of mutable disposables, where each mutuable disposable is a single lspCompletionProviderAddon.

Tyriar: You should only need one. `MutableDisposable` is nice as it makes sure it gets cleaned up when `this` is disposed, though I notice we're not registering it which defeats the purpose:

```ts
this._register(new MutableDisposable());
```

Try this instead:

```ts
private readonly _lspAddons: DisposableMap<string, LspCompletionProviderAddon> = this.register(new DisposableMap());
```

Then access via `this._lspAddons.values()`. Doing the register like this means you don't need to this:

```ts
			for (const lspAddon of this._lspAddonInstances) {
				lspAddon.dispose();
			}
```

Tyriar (changes requested):

src/vs/workbench/contrib/terminalContrib/suggest/browser/terminal.suggest.contribution.ts
anthonykim1: Thanks! I'm now storing this disposable map of provider id, and lspCompletionProviderAddon.

But now have some disposable leaks when I toggle suggest on/off:


https://github.com/user-attachments/assets/4e5f47e6-2a5b-4451-9b9f-fbcf408497d0


Maybe this is because Im using `clearAndDisposeAll` since `clear` wasn't available on DisposableMap? 
```
if (!completionsEnabled) {
	this._addon.clear();
	this._pwshAddon.clear();
	this._lspAddons.clearAndDisposeAll();
}
```
Tried to use `.dispose` instead of `.clearAndDisposeAll()` but that didn't help either.

Tyriar: Since you use `.add`, it should call dispose on the map which disposes all entries when the _contribution_ is disposed. I'm not actually sure if this can happen in practice.

Those leaks seem unrelated, you sure they don't happen regardless of your changes?

src/vs/workbench/contrib/terminalContrib/suggest/common/terminalSuggestConfiguration.ts
```diff
@@ -212,6 +214,15 @@ export function registerTerminalSuggestProvidersConfiguration(availableProviders
 		default: false
 	};
 
+	// Add LSP provider configuration when not in stable
+	if (product.quality !== 'stable') {
```
anthonykim1: Had to import `product.quality` instead of `productservice.quality` to avoid code layer violation.

I think this should be fine? looking at: 

Screenshot 2025-08-06 at 10 46 20 AM

src/vs/workbench/contrib/terminalContrib/suggest/common/terminalSuggestConfiguration.ts
```diff
@@ -201,8 +201,10 @@ export function registerTerminalSuggestProvidersConfiguration(availableProviders
 	const providersProperties: IStringDictionary<IConfigurationPropertySchema> = {};
 
 	const corePwshProviderId = 'core:pwsh-shell-integration';
+	const lspProviderId = 'lsp';
```
anthonykim1, line 204: Wondering if we should attach provider unique id (language extension's debugDisplayName) after `lsp`, so it looks something like: `lsp:ms-python.python(.["');` in user's settings.json:

```
 "terminal.integrated.suggest.providers": {
    
    "core:pwsh-shell-integration": false,
    "terminal-suggest": true,
   
// Option #1 Consolidate all lsp providers with "lsp" for user's setting
"lsp": true 
OR 
// Option #2 Multiple providers with lsp prefix
"lsp:ms-python.python(.["')"
"lsp: ms-vscode.powershell(.-:\\$ );"
... other future lsp
  },
```
The downside here is that the debugDisplayName for these providers can look ugly like ``ms-python.python(.["')`;` 

Then, we can set default true on insiders for all of the providers that start with lsp.

src/vs/workbench/contrib/terminalContrib/suggest/browser/terminal.suggest.contribution.ts
anthonykim1: nvm, was able to repro on latest main 😅

src/vs/workbench/contrib/terminalContrib/suggest/common/terminalSuggestConfiguration.ts
Tyriar: How about `lsp:<extensionId>` instead?

src/vs/workbench/contrib/terminalContrib/suggest/common/terminalSuggestConfiguration.ts
Tyriar: It's ok, I think we have to do it this way currently for settings as we can't depend on services that early in workbench init.

Tyriar: We want to add it for stable, just `default` should be `product.quality !== 'stable'`

src/vs/workbench/contrib/terminalContrib/suggest/browser/lspTerminalModelContentProvider.ts
```diff
@@ -148,7 +147,7 @@ export class LspTerminalModelContentProvider extends Disposable implements ILspT
 			if (!languageId) {
 				switch (extension) {
 					case 'py': languageId = PYTHON_LANGUAGE_ID; break;
-					// case 'ps1': languageId = 'powershell'; break;
+					case 'ps1': languageId = PWSH_LANGUAGE_ID; break;
 					// case 'js': languageId = 'javascript'; break;
 					// case 'ts': languageId = 'typescript'; break; etc...
 				}
```
Tyriar: There must be some service available to do this file extension to language type mapping.

Tyriar (changes requested):

src/vs/workbench/contrib/terminalContrib/suggest/common/terminalSuggestConfiguration.ts
anthonykim1: I like that idea, except we dont have access to extensionId from provider: https://github.com/microsoft/vscode/blob/3cef167a29bb4501247c72c77dff14ca1dd8c195/src/vs/editor/common/languages.ts#L698

We do get access to extensionID later when we get lsp completion items, but that would be too late and way after the constructor is called :/

So it may have to be the debugDisplay name, or I can get little hacky and build out regex to extract out words like `ms-python.python` and `ms-vscode.powershell`?  The latter won't always be accurate though.
* [f449213](https://github.com/microsoft/vscode/pull/258137/commits/f449213b101e3a537237f21f0b57c67e85266bd3) would be with the debugDisplay name

src/vs/workbench/contrib/terminalContrib/suggest/browser/lspTerminalModelContentProvider.ts
anthonykim1: I love it, this code  block just became a one liner
```
const languageId = this._languageService.guessLanguageIdByFilepathOrFirstLine(resource);
```
[c4bc0db](https://github.com/microsoft/vscode/pull/258137/commits/c4bc0db77d2bd4aa8229d9b9af923d466e2d01e9)

src/vs/workbench/contrib/terminalContrib/suggest/common/terminalSuggestConfiguration.ts
```diff
@@ -212,11 +212,20 @@ export function registerTerminalSuggestProvidersConfiguration(availableProviders
...
 				continue;
+			} else if (providerId.startsWith('lsp')) {
+				providersProperties[providerId] ??= {
+					type: 'boolean',
+					description: localize(`suggest.provider.lsp.${providerId}.description`, "Enable or disable the Language Server Protocol (LSP) based provider. This provides enhanced completion support through language servers."),
+					default: product.quality !== 'stable'
+				};
+				continue;
 			}
+
```
Tyriar: @andyleejordan I'm trying to test this in the powershell extension, does the extension not analyze all schemes? We need the `vscode-terminal` scheme to support completions in LSP for this feature.

@anthonykim1 or am I doing something wrong here? There's no `lsp` provider here after installing the pwsh ext:

image

anthonykim1: @Tyriar Can you try opening pwsh terminal after activating powershell extension (I think the easiest way is opening ps1 file), and then the pwsh extension would show up as a provider:
Screenshot 2025-08-07 at 8 02 20 AM

Maybe we'd want to do https://github.com/microsoft/vscode/issues/259915 on pwsh extension too to allow users to get provider name, suggestions without having to activate pwsh extension themselves.

Tyriar (changes requested):
Since pwsh doesn't work that well due to conflicts and not showing suggestions when we want them to show (eg. empty prompt, content after the cursor). Let's centralize the supported language like this:

```ts
export const terminalLspSupportedLanguages = new Set<{ shellType: string; languageId: string; extension: string }>([
	{
		shellType: 'python',
		languageId: 'python',
		extension: 'py'
	}
]);
```

Then just move back to the `lsp` setting.

src/vs/workbench/contrib/terminalContrib/suggest/browser/terminal.suggest.contribution.ts
```diff
@@ -75,7 +74,7 @@ class TerminalSuggestContribution extends DisposableStore implements ITerminalCo
 		this.add(toDisposable(() => {
 			this._addon?.dispose();
 			this._pwshAddon?.dispose();
-			this._lspAddon?.dispose();
+			this._lspAddons.dispose();
```
Tyriar: Since you call `.add` on the disposable map you shouldn't need to do this.

src/vs/workbench/contrib/terminalContrib/suggest/common/terminalSuggestConfiguration.ts
Tyriar, line 204: Because of the `provider: `lsp:${item.extensionId}`` change I don't think this will work anymore? We can hard code the extension name in this config section if needed.

Tyriar (changes requested):
Very close!

src/vs/workbench/contrib/terminalContrib/suggest/common/terminalSuggestConfiguration.ts
anthonykim1: Screenshot 2025-08-08 at 7 54 55 AM
Hmm.. I think its still working since we have `readonly id = 'lsp'` in 

https://github.com/microsoft/vscode/blob/bebc70d37093ff410900484193704aec23d1cfd5/src/vs/workbench/contrib/terminalContrib/suggest/browser/lspCompletionProviderAddon.ts#L18

Should we just hard code this to match  `provider: lsp: extensionId ` in https://github.com/microsoft/vscode/blob/bebc70d37093ff410900484193704aec23d1cfd5/src/vs/workbench/contrib/terminalContrib/suggest/browser/lspCompletionProviderAddon.ts#L64

or leave as `lsp`?

Tyriar (approved):

PR #258184, opened 2025-07-28
Computing full diff across a turn

### LINKED ISSUE #258076: Changes 25 files when only 4 changed
It seems to be counting file change events or something, not individual files?

Image

It was a pretty long session

Image

Version: 1.103.0-insider (user setup)
Commit: d8361fecca3b53536de8f7a949b107fb7451e5e6
Date: 2025-07-25T05:03:31.984Z
Electron: 37.2.3
ElectronBuildId: 11993778
Chromium: 138.0.7204.100
Node.js: 22.17.0
V8: 13.8.500258-electron.0
OS: Windows_NT x64 10.0.26100

aiday-mar:
fixes https://github.com/microsoft/vscode/issues/258076


src/vs/workbench/contrib/chat/browser/chatEditing/chatEditingTimeline.ts
```diff
@@ -40,8 +40,10 @@ export class ChatEditingTimeline {
 
 	private readonly _linearHistory = observableValue<readonly IChatEditingSessionSnapshot[]>(this, []);
 	private readonly _linearHistoryIndex = observableValue<number>(this, 0);
+	private readonly _requestToLinearHistoryMap = new Map<string, number>();
```
connor4312: I would prefer we didn't have this unless you noticed a performance issue here. But if we do, this should be `derived` from the `_linearHistory` to avoid it ever diverging/us having to manually manage state

src/vs/workbench/contrib/chat/browser/chatEditing/chatEditingTimeline.ts
```diff
@@ -418,6 +416,94 @@ export class ChatEditingTimeline {
 			return observable;
 		}
 	}
+
+	public getEntryDiffBetweenRequests(uri: URI, startRequestId: string, stopRequestId: string): IObservable<IEditSessionEntryDiff | undefined> | undefined {
+		const key = `${uri}\0${startRequestId}\0${stopRequestId}`;
+		let observable = this._diffsBetweenRequests.get(key);
```
connor4312: Do you actually need the diff cached in `_diffsBetweenRequests`? I have it for going between stops because of the kind of weird way we render those pills, but it's been on my list to try and clean that up for a while. If you don't actually caching for the between-requests diff, I would try to avoid this since it is a little bit of a memory leak (until the session is ended)

src/vs/workbench/contrib/chat/browser/chatEditing/chatEditingTimeline.ts
```diff
@@ -418,6 +416,94 @@ export class ChatEditingTimeline {
...
+		if (!observable) {
+			observable = this._createDiffBetweenRequestsObservable(uri, startRequestId, stopRequestId);
+			this._diffsBetweenRequests.set(key, observable);
+		}
+		return observable;
+	}
+
+	private _createDiffBetweenRequestsObservable(uri: URI, startRequestId: string, stopRequestId: string): IObservable<IEditSessionEntryDiff | undefined> {
+		const modelRefs = derived((reader) => {
+			const firstSnapshotUri = this._getFirstSnapshotForUriAfterRequest(uri, startRequestId, true).read(reader);
```
connor4312: I suggest moving the calls to `this._getFirstSnapshotForUriAfterRequest` out of the `derived` block -- they're creating new observables each time which is doubling the work

src/vs/workbench/contrib/chat/browser/chatEditing/chatEditingTimeline.ts
```diff
@@ -418,6 +416,94 @@ export class ChatEditingTimeline {
...
+			const lastSnapshotUri = this._getFirstSnapshotForUriAfterRequest(uri, stopRequestId, false).read(reader);
+			if (!firstSnapshotUri || !lastSnapshotUri) {
+				return;
+			}
+			const store = new DisposableStore();
+			reader.store.add(store);
+			const referencesPromise = Promise.all([firstSnapshotUri, lastSnapshotUri].map(u => this._textModelService.createModelReference(u))).then(refs => {
```
connor4312, line 419: I would make a new intermediate observable for the URIs like we do for `modelUrisObservable`. As-is, this `derive` block will run any time the history changes (as it reads from `_linearHistory`) but unless the URIs change you don't need to get new references (and thus recompute the diff down the line)

connor4312 (changes requested):

src/vs/workbench/contrib/chat/browser/chatEditing/chatEditingTimeline.ts
aiday-mar: Do you mean instead instead we could iterate from the beginning of the linear history until we reach the relevant request id ?

src/vs/workbench/contrib/chat/browser/chatEditing/chatEditingTimeline.ts
aiday-mar: I have initially tried without the caching and noticed that there was some sort of infinite loop created due to how the deriveds were chained that was making the whole thing be recomputed over and over again. Using the cache allowed to break that infinite loop. I initially thought that was the reason you used a cache.

src/vs/workbench/contrib/chat/browser/chatEditing/chatEditingTimeline.ts
aiday-mar: So do you mean adding them into another derived but separate block?

src/vs/workbench/contrib/chat/browser/chatEditing/chatEditingTimeline.ts
aiday-mar: I made the other changes you mentioned in this PR. When I removed the cache, I saw the infinite loop happen again, I am not quite sure what is causing it.

src/vs/workbench/contrib/chat/browser/chatEditing/chatEditingTimeline.ts
connor4312: Yea -- or make `_requestToLinearHistoryMap = this._linearHistory.map(arr => new Map(...))` so that it is derived that way

src/vs/workbench/contrib/chat/browser/chatEditing/chatEditingTimeline.ts
connor4312: Does `getEntryDiffBetweenRequests` end up getting called multiple times and, if so, is there a way to avoid doing that or cache it within the consumer of that call?

src/vs/workbench/contrib/chat/browser/chatEditing/chatEditingTimeline.ts
connor4312: yep what you did is good

connor4312: I mean to the top level in the function. `_getFirstSnapshotForUriAfterRequest` internally creates a derived observable. So what happens is:

1. The first `modelRefs` function is called, all good.
2. The history changes. This causes `_getFirstSnapshotForUriAfterRequest` to need to recompute. It does so and then will trigger the `modelRefs` function to recompute.
3. But now the `modelRefs` gets called again. And instead of using the computation that `_getFirstSnapshotForUriAfterRequest` just did, it calls `_getFirstSnapshotForUriAfterRequest` again and computes it for a third time when it creates the new observable within that function.

If the calls to `this._getFirstSnapshotForUriAfterRequest` were moved outside the `derived` block, they would be reused and not do the third computation ^

src/vs/workbench/contrib/chat/browser/chatEditing/chatEditingTimeline.ts
aiday-mar: I changed the code so as to iterate in a linear fashion on it.

src/vs/workbench/contrib/chat/browser/chatEditing/chatEditingTimeline.ts
aiday-mar: Thanks that makes sense. I made the relevant change.

src/vs/workbench/contrib/chat/browser/chatEditing/chatEditingTimeline.ts
aiday-mar: The method `getEntryDiffBetweenRequests` is called for every single file change at least once, and then it is called when the constituents inside change.

aiday-mar: So I am wondering if I can cache it in the consumer of the call. I have as arguments, the reference, the first request id and the last request id. The corresponding snapshot URI can change as the linear history changes, and I don't have access to that information on the consumer side, so I am wondering if I can cache it there. Let me know if I misunderstood your message.

connor4312: You would cache the observable itself, not the value of the observable.

aiday-mar: Thanks I made the required change. Could you review the PR again?

aiday-mar:
Hi thanks for the comments. I implemented the review changes. Could you perhaps rereview?

connor4312 (dismissed):

connor4312 (approved):

lszomoru (approved):

PR #258206, opened 2025-07-28
fix: memory leak in context key

SimonSiefke:
This fixes a memory leak related to context keys. In `contextkeys.ts` there is this function: 

```ts
this._register(Event.runAndSubscribe(onDidRegisterWindow, ({ window, disposables }) => disposables.add(addDisposableListener(window, EventType.FOCUS_IN, () => this.updateInputContextKeys(window.document, disposables), true)), { window: mainWindow, disposables: this._store }));
```
When a new window is registered, for each window it adds a `focusin` event listener for updating the context keys. 

<br>

The memory leak problem seems to be that it passes the `disposables` of this window to the `updateInputContextKeys` function. 

```ts
private updateInputContextKeys(ownerDocument: Document, disposables: DisposableStore): void {

 /* ... */

if (isInputFocused) {
	const tracker = disposables.add(trackFocus(ownerDocument.activeElement as HTMLElement));
	Event.once(tracker.onDidBlur)(() => {

	  /* ... */

		tracker.dispose();
	}, undefined, disposables);
}
```

Here the `FocusTracker` disposable is added to `disposables` each time an input element in the window is being focused, making `disposables` grow each time.

For example: 
1. A window is open
2.  The quickpick is being opened
3.  The quickpick input is being focused
4.  A focus tracker is created and added to the `disposables` array of this window
5. Close the quickpick
6. Repeat the steps 2-5. Each time a `FocusTracker` is being added to the `disposables` array

### Fix

The fix uses a `MutableDisposable` to dispose and overwrite the previous focus trackers. 

```ts
private updateInputContextKeys(ownerDocument: Document, mutableDisposable: MutableDisposable<any>): void {

 /* ... */

mutableDisposable.value = store;

if (isInputFocused) {
	const tracker = store.add(trackFocus(ownerDocument.activeElement as HTMLElement));
	Event.once(tracker.onDidBlur)(() => {

	  /* ... */

		tracker.dispose();
	}, undefined, store);
}
```

This way when an input element in the window is focused, the previous `FocusTracker` will be unreferenced by setting `mutableDisposable.value` with the new `DisposableStore`.

### Before

When showing and hiding the quickpick 31 times, the number of `FocusTracker` instances increases each time from 63 to 125:

named-instance-count

### After

When showing and hiding the quickpick 31 times, the number of `FocusTracker` stays constant at 63:

named-instance-count

### DIFF
--- a/src/vs/workbench/browser/contextkeys.ts
+++ b/src/vs/workbench/browser/contextkeys.ts
@@ -4,7 +4,7 @@
  *--------------------------------------------------------------------------------------------*/
 
 import { Event } from '../../base/common/event.js';
-import { Disposable, DisposableStore } from '../../base/common/lifecycle.js';
+import { Disposable, DisposableStore, MutableDisposable } from '../../base/common/lifecycle.js';
 import { IContextKeyService, IContextKey, setConstant as setConstantContextKey } from '../../platform/contextkey/common/contextkey.js';
 import { InputFocusedContext, IsMacContext, IsLinuxContext, IsWindowsContext, IsWebContext, IsMacNativeContext, IsDevelopmentContext, IsIOSContext, ProductQualityContext, IsMobileContext } from '../../platform/contextkey/common/contextkeys.js';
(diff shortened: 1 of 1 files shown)


deepak1556 (approved):

DonJayamanne (approved):

bpasero:
@SimonSiefke thanks, I am pushing https://github.com/microsoft/vscode/pull/259818 as refinement over your work. Notable:
* avoid use of `any`
* make the `MutableDisposable` a thing at the controller level and not within the function itself

Please speak up if that change does not make sense to you.

fyi @deepak1556

PR #258245, opened 2025-07-28
Add note about shell integration when not detected

### LINKED ISSUE #258230: Run in terminal: Add warning about lack of shell integration
Some issues are hard to fix without shell integration like https://github.com/microsoft/vscode/issues/255932, perhaps we should just show a hint to the user to guide them towards fixing their shell integration problem?

Tyriar:
Fixes #258230

image


meganrogge (approved):
Nice 👏🏼

PR #258265, opened 2025-07-28
Codespace organic meme 697gx4r96xx436xp

gtggy890m-spec:


gtggy890m-spec:
value = "[Grab this power-up!](https://marketplace.visualstudio.com/items?itemName=mtxr.sqltools)👍"

We want the URL inside the parentheses.

We can find the index of '(' and then the next ')'.

Alternatively, we can split:

Let s = value
We know it starts with '[' and then has some text until ']', then '(' and then the URL until ')'.

We can do:

    # Find the first '(' after the ']'
    start_index = value.find('(') + 1
    end_index = value.find(')', start_index)
    url = value[start_index:end_index]

Then we have the URL.

PR #258346, opened 2025-07-28
Removing chat session provider label

osortega:
Removing chat session provider label and using the extension point contribution instead


connor4312 (approved):

src/vs/workbench/contrib/chat/common/chatSessionsService.ts
```diff
@@ -14,6 +14,7 @@ import { IChatAgentRequest } from './chatAgents.js';
 import { IRelaxedExtensionDescription } from '../../../../platform/extensions/common/extensions.js';
 
 export interface IChatSessionsExtensionPoint {
+	readonly id: string;
```
mjbvz, line 17: I think that we should only have `type` now. It was previously called `id` but I just changed this and having both will be confusing

Issue #257872, opened 2025-07-25
leaked disposables when closing notebook inline chat


amunger:
Steps to Reproduce:

1. open inline chat v2 in a notebook and use it to make a change
2. click undo 

:bug: 

```
Error: Trying to add a disposable to a DisposableStore that has already been disposed of. The added object will be leaked!
    at DisposableStore.add (vscode-file://vscode-app/d:/src/vscode/out/vs/base/common/lifecycle.js:344:30)
    at ChatEditingModifiedNotebookEntry._register (vscode-file://vscode-app/d:/src/vscode/out/vs/base/common/lifecycle.js:407:28)
    at ChatEditingModifiedNotebookEntry.resolveCellModel (vscode-file://vscode-app/d:/src/vscode/out/vs/workbench/contrib/chat/browser/chatEditing/chatEditingModifiedNotebookEntry.js:876:67)
lifecycle.ts:453
  ERR Model is disposed!: Error: Model is disposed!
    at TextModel._assertNotDisposed (vscode-file://vscode-app/d:/src/vscode/out/vs/editor/common/model/textModel.js:305:19)
    at TextModel.getValue (vscode-file://vscode-app/d:/src/vscode/out/vs/editor/common/model/textModel.js:569:14)
    at new ChatEditingNotebookCellEntry (vscode-file://vscode-app/d:/src/vscode/out/vs/workbench/contrib/chat/browser/chatEditing/notebook/chatEditingNotebookCellEntry.js:51:50)
    at InstantiationService._createInstance (vscode-file://vscode-app/d:/src/vscode/out/vs/platform/instantiation/common/instantiationService.js:130:24)
    at InstantiationService.createInstance (vscode-file://vscode-app/d:/src/vscode/out/vs/platform/instantiation/common/instantiationService.js:101:27)
    at ChatEditingModifiedNotebookEntry.getOrCreateModifiedTextFileEntryForCell (vscode-file://vscode-app/d:/src/vscode/out/vs/workbench/contrib/chat/browser/chatEditing/chatEditingModifiedNotebookEntry.js:886:63)
    at vscode-file://vscode-app/d:/src/vscode/out/vs/workbench/contrib/chat/browser/chatEditing/chatEditingModifiedNotebookEntry.js:638:18
log.ts:460
[LEAKED DISPOSABLE] Error: CREATED via:
[...]

PR #258379, opened 2025-07-29
Fix possible leaks in notebook chat Editing component

DonJayamanne:
For https://github.com/microsoft/vscode/issues/257872#event-18865319863

@amunger /cc (I think this might fix it)


amunger:
still has one leak: 

```
Error: Trying to add a disposable to a DisposableStore that has already been disposed of. The added object will be leaked!
    at DisposableStore.add (lifecycle.ts:453:18)
    at ChatEditingModifiedNotebookEntry._register (lifecycle.ts:525:22)
    at ChatEditingModifiedNotebookEntry.resolveCellModel (chatEditingModifiedNotebookEntry.ts:969:61)
```

image

The repro may also require the new setting `"inlineChat.notebookAgent": true` - though when I turn that off and try a similar edit, I see the error `Notebook edits not generated as notebook already has cells` and inline chat closes and won't open again.
Need to play around a bit more

TylerLeonhardt (approved):

PR #258454, opened 2025-07-29
Only change the line height of the view line

### LINKED ISSUE #249967: Apply decoration lineHeight on all affected lines after line wrapping
In #194609 I implemented support for dynamic line heights using decorations. This works perfectly for us, but there are some unhandled things that were not relevant to us. Due to a change of priorities, this PR got abandoned. Now I have to get back to it.

Recently #241218 was merged. This also adds support for dynamic line heights using decorations. It was tested against some cases I never tested against, which is great. However, the behaviour is quite different.

There are two major differences:
1. The implementation in #194609 calculates the line heights of the lines after text wrapping. The implementation in #241218 calculates the line height of the unwrapped line, and applies this height to display lines that display the wrapped real line.
2. The implementation in #194609 applies the line height to all lines that are covered by the decoration range. The implementation in #241218 only applies to the first line of the decoration range.

I think these behaviours from #194609 are better. They match with what LibreOffice, Google Docs, and Microsoft Word do.

Would a PR be welcome to change this behaviour?

### LINKED ISSUE #295179: Custom line heights should only affect the view line

- VS Code Version: 1.109.3
- OS Version: Pop!_OS 22.04

Steps to Reproduce:

1. Use the following settings:
   ```jsonc
   {
       "textMateRules": [
         {
           "scope": "entity.name.type.class",
           "settings": {
             "fontSize": 3,
             "lineHeight": 2
           }
         }
       ]
     },
     "[typescript]": {
       "editor.wordWrap": "bounded",
       "editor.wordWrapColumn": 50
     },
  }
   ```
2. Create a TypeScript file with the following content:
   ```ts
   class A extends VeryLongBaseClassNameUsedToTriggerLineWrapping {}
   ```

Observe how the class name `A` is bigger than the rest. The line height is adjusted as well. However, this line height expands even after the line wrapping.

Other editors that support variable font sizes, only make the actual view line bigger. After line wrapping, the font size of one bigger word doesn’t increase the line height of any of the wrapped view lines. For example, this screenshot shows the behaviour in Google Docs.

Internally in VSCode, this uses Monaco decorations. Monaco decorations have the property `isWholeLine`. If this is set to true, the decoration is applied to the entire line instead of just the selected range. This can be used to determine if the custom line height should be applied to the entire line as well, or just to the unwrapped view line.

remcohaszing:
Previously, if a decoration specified a custom line height, this affected all viewlines that are covered by model line. Now the line height is only applied to the view lines that render the decoration.

It is still possible to change the line height of an entire view line by setting the decoration option `isWholeLine`.

image

Closes #249967
Closes #295179

cc @aiday-mar

### DIFF
--- a/src/vs/editor/common/model/decorationProvider.ts
+++ b/src/vs/editor/common/model/decorationProvider.ts
@@ -29,14 +29,15 @@ export interface DecorationProvider {
 export class LineHeightChangingDecoration {
 
 	public static toKey(obj: LineHeightChangingDecoration): string {
-		return `${obj.ownerId};${obj.decorationId};${obj.lineNumber}`;
+		return `${obj.ownerId};${obj.decorationId}`;
 	}
 
 	constructor(
 		public readonly ownerId: number,
 		public readonly decorationId: string,
-		public readonly lineNumber: number,
-		public readonly lineHeight: number | null
+		public readonly range: Range,
+		public readonly lineHeight: number | null,
+		public readonly isWholeLine: boolean
 	) { }
 }
 
--- a/src/vs/editor/common/model/textModel.ts
+++ b/src/vs/editor/common/model/textModel.ts
@@ -1654,7 +1654,7 @@ export class TextModel extends Disposable implements model.ITextModel, IDecorati
 	private _fireOnDidChangeLineHeight(affectedLineHeights: Set<LineHeightChangingDecoration> | null): void {
 		if (affectedLineHeights && affectedLineHeights.size > 0) {
 			const affectedLines = Array.from(affectedLineHeights);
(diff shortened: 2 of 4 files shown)


remcohaszing:
Apparently this works, but the line heights are not updated if the lines are affected by changes to line wrapping. I’m feeling a bit stuck on this and would really appreciate some help or feedback.

aiday-mar:
Hi @remcohaszing thank you for this PR. I am not sure we should apply the line height only to the view line because as you mentioned if line wrapping is turned on, and you change the size of the window, then the decorations will change position and you would have to change the size of the line heights. This means the line heights would dynamically change as you resize the screen. This adds a whole layer of complexity to the work. Could we start off by opening an issue about this, and discussing it within the issue, how important of an issue this is, if its urgent, before writing this code and merging it?

PR #258463, opened 2025-07-29
Fix testFailure stringify

Skn0tt:
Follow-up to https://github.com/microsoft/vscode/issues/257350. The code is currently joining an array of objects, so the result is a soup of `[object Object]`.


connor4312 (approved):
thanks!

Tyriar (approved):

PR #258664, opened 2025-07-29
Avoid registering disposables when Notebook Edit session is disposed

DonJayamanne:
For https://github.com/microsoft/vscode/issues/257872

@amunger another fix, not sure this will fix it, as I cannot repro.
Either way, thanks for the screenshot with stack traces in previous PR


amunger:
tested - fixes the disposables issue and I couldn't find any new issues

Tyriar (approved):

PR #258709, opened 2025-07-30
fix default path bug related to branch prefix

### LINKED ISSUE #258623: default worktree path changed
Test for #257910

I somehow got the default path for a worktree to be prefixed with `don/` when creating a new worktree for a new branch:

Image

I had previously created a worktree from Don's branch `don/rural-dinosaur` and now trying to create a worktree from that branch seems to auto-select the name to the branch name and goes straight to selecting the path:

Image

nguyenchristy:
Fixes https://github.com/microsoft/vscode/issues/258623#issue-3274659521

The default worktree name will be the branch name with the user's branch prefix removed. If there are still '/', then replace with '-' to prevent nested worktree paths.

### DIFF
--- a/extensions/git/src/commands.ts
+++ b/extensions/git/src/commands.ts
@@ -3400,6 +3400,7 @@ export class CommandCenter {
 
 	private async _createWorktree(repository: Repository): Promise<void> {
 		const config = workspace.getConfiguration('git');
+		const branchPrefix = config.get<string>('branchPrefix')!;
 		const showRefDetails = config.get<boolean>('showReferenceDetails') === true;
 
 		const createBranch = new CreateBranchItem();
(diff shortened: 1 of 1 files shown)


extensions/git/src/commands.ts
```diff
@@ -3451,10 +3451,10 @@ export class CommandCenter {
 			commitish = choice.refName;
 		}
 
-
-		const worktreeName = (branch ?? commitish).startsWith(branchPrefix)
+		const worktreeName = ((branch ?? commitish).startsWith(branchPrefix)
 			? (branch ?? commitish).substring(branchPrefix.length)
-			: (branch ?? commitish);
+			: (branch ?? commitish)
+		).split('/').pop() || (branch ?? commitish);
```
lszomoru: I think that it might make more sense to replace `/` with `-`.

lszomoru (approved):

PR #131213, opened 2021-08-19
Add prototype of mac open command

rzhao271:
This PR fixes https://github.com/microsoft/vscode/issues/60579 and fixes https://github.com/microsoft/vscode/issues/102975.

It switches the spawn call on macOS to use the `open` command, so that the application opens as though it were launched from the dock. In turn, launching the application from the console should now not result in as many bugs such as duplicated recently opened icons.

To test on macOS:
1. Build the OSS project
2. Run `code-cli.sh` in the terminal, e.g. `./scripts/code-cli.sh <your-flags-here>`.

PR #258824, opened 2025-07-30
Use open -a in code.sh on mac

JBlitzar:
In reference to #258821

Uses the `open` command rather than launching via Electron directly.
Test by replacing the contents of /usr/local/bin/code with the proposed changes (or rebuilding).

### DIFF
--- a/resources/darwin/bin/code.sh
+++ b/resources/darwin/bin/code.sh
@@ -30,6 +30,3 @@ if [ -z "$APP_PATH" ]; then
 fi
 open -a "$APP_PATH" --args "$@"
 exit $?
-unset NODE_REPL_EXTERNAL_MODULE
-ELECTRON_RUN_AS_NODE=1 "$ELECTRON" "$CLI" "$@"
-exit $?
--- a/resources/darwin/bin/code.sh
+++ b/resources/darwin/bin/code.sh
@@ -28,12 +28,8 @@ if [ -z "$APP_PATH" ]; then
 	echo "Unable to determine app path from symlink : ${BASH_SOURCE[0]}"
 	exit 1
 fi
-CONTENTS="$APP_PATH/Contents"
-ELECTRON="$CONTENTS/MacOS/Electron"
-CLI="$CONTENTS/Resources/app/out/cli.js"
-export VSCODE_NODE_OPTIONS=$NODE_OPTIONS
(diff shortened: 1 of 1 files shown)


deepak1556 (changes requested):
The shell script calls into a cli script in headless mode, the cli inturn opens the actual application based on the arguments via `open` https://github.com/microsoft/vscode/blob/b234a54b54591f23df239de3ad9c3bef36a0732f/src/vs/code/node/cli.ts#L493-L567

I don't see how this makes it any different than what we have today. Also we cannot skip the cli script since it takes of various argument processing, some don't require actual application to be opened.

JBlitzar:
Ah I see. Then it's mostly a design choice. Why are we using -n? This appears to be intentional to create a new instance. Don't most people want it to just open in the same existing instance?

deepak1556:
We have our own implementation of singleton instance https://github.com/microsoft/vscode/blob/ce135b94c0f34aa4ea4ae52d6f95462366d71914/src/vs/code/electron-main/main.ts#L292-L420, `-n` is used to route to our logic.

JBlitzar:
Then I guess this is a separate question. What it ends up doing visually is that it creates a new instance that instantly quits and transitions focus to the old instance, but then there's an extra VSC in my dock.

JBlitzar:
Thanks for all the clarification

deepak1556:
> What it ends up doing visually is that it creates a new instance that instantly quits and transitions focus to the old instance, but then there's an extra VSC in my dock

Yup thats correct, the extra icon is something I can confirm as well but I tend to rule this out as macOS expectation. The icon takes the space of recently launched application, unless you have VSCode pinned to the dock then a different application launch should replace the extra icon. If there is an open option without changing our cli behavior, happy to try that.

JBlitzar:
Hm I see. I'll try to see if I can think of something.

JBlitzar:
Funnily enough, I don't seem to be able to reproduce it appearing as a recently quit app in the dock anymore. Either way, perhaps inserting `app.dock.hide()` would benefit

deepak1556:
Remembered we had a discussion around revisiting the use of `dock.hide` when transitioning to use `open` command https://github.com/microsoft/vscode/pull/131213#issuecomment-912251877 but we never got to it. Feel free to give it a try in this PR.

JBlitzar:
all right, I'll give it a try

Issue #258619, opened 2025-07-29
reusing existing worktree name has a warning but then fails silently


amunger:
Test for #257910

1. create a worktree from a branch and remember the name used
2. right click the worktree item and "close repository"
3. create the worktree again and type the same name previously used
    1. a warning appears in the quickpick
4. press enter anyway, and press enter again for the path

:bug: nothing happens

rzhao271:
\closedWith 1297e26

osortega:
It still lets you go through the flow and shows an error at the end? Should we just reopen the worktree repository instead?

<img width="1840" height="1191" alt="Image" src="https://github.com/user-attachments/assets/e5b1bcd8-e43c-4556-9c5c-e0ef538b1cce" />

nguyenchristy:
\closedWith https://github.com/microsoft/vscode/commit/f58a56ea034f5866a9c5f99c69dc04dc6e9e38e7

PR #258867, opened 2025-07-30
Handle worktree errors

### LINKED ISSUE #258557: attempting to create worktree on branch open on worktree- unclear result
Testing #257910

doing steps for: `Create worktree for a branch that already has a worktree`, (see video below as I go through the steps). If I select a branch that already is checkout on a worktree, it prompts me to create a new branch. I got confused when this happened because there was no messaging about how my prior action couldn't work. I would also want to be presented with the option to choose a different existing on or to create a new branch after selecting one that is already checked out. 

To understand it better I did the same command from the terminal and got this error message which would be nice to surface to users in some way to keep them better in the loop

Image

https://github.com/user-attachments/assets/5022f2a8-30b4-4bbd-93b3-f6645e886363

nguyenchristy:
Fixes https://github.com/microsoft/vscode/issues/258557#event-18883432995 and https://github.com/microsoft/vscode/issues/258619#event-18884722720

This PR handles 2 separate kinds of worktree errors: (1) when a user wants to create a worktree at a path where a worktree already exists and (2) if a user attempts to check out a branch that already has a worktree. 
image

Shows:
1) Checking out branch that is currently used by worktree
2) Creating worktree at a path where a worktree already exists
3) Checking out a branch that already has a worktree
4) Creating a worktree by checking out a branch you're currently on 

https://github.com/user-attachments/assets/26cf37b9-efb1-4516-a65a-39763652bfab

### DIFF
--- a/extensions/git/src/api/git.d.ts
+++ b/extensions/git/src/api/git.d.ts
@@ -442,5 +442,6 @@ export const enum GitErrorCodes {
 	CherryPickEmpty = 'CherryPickEmpty',
 	CherryPickConflict = 'CherryPickConflict',
 	WorktreeContainsChanges = 'WorktreeContainsChanges',
-	WorktreeAlreadyExists = 'WorktreeAlreadyExists'
+	WorktreeAlreadyExists = 'WorktreeAlreadyExists',
+	WorktreeBranchAlreadyUsed = 'WorktreeBranchAlreadyUsed'
 }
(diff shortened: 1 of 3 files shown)


lszomoru:
I think that when it comes to the flow to create a new worktree we need to provide feedback to the user as soon as possible: 
1) as soon as they pick a branch that we know is checked out in another worktree we should show a dialog 
2) in the worktree path input we can validate the path before the user attempts to create a new worktree

I have pushed some changes to the pull request accordingly. Please take a look at them and test them.

rebornix (approved):

PR #258891, opened 2025-07-30
make sure to resize image to the same mimetype

justschen:


src/vs/workbench/contrib/chat/browser/imageUtils.ts
```diff
@@ -66,7 +66,7 @@ export async function resizeImage(data: Uint8Array | string, mimeType?: string):
 					} else {
 						reject(new Error('Failed to create blob from canvas'));
 					}
-				}, 'image/png');
+				}, mimeType || 'image/png');
```
connor4312: we may want to pick only png or jpg. webp images result in 500 errors in Sonnet https://github.com/microsoft/vscode/issues/251450

TylerLeonhardt (approved):

Issue #255218, opened 2025-07-10
Support pinning client version when running via `serve-web`


justin39:
Currently, `code serve-web` seems to always [grab the latest release](https://github.com/microsoft/vscode/blob/main/cli/src/commands/serve_web.rs#L621) for the client. This is convenient, but occasionally there are problems with the latest client using `code serve-web` leaves no way to downgrade or pin to a working client version. A flag like `--commit-id` would be a convenient escape hatch for these situations where we just need a way to temporarily avoid the latest release.

(This request was in particular inspired by #251731 as it would cause the client to crash upon opening certain notebooks)

rzhao271:
Adding a commit ID seems to result in the browser not downloading any version at all.

Example command: 
`code-insiders serve-web --commit-id f5db653369bc87e56162be11ed3af62746347558`

justin39:
Ah yeah - I think I see the issue; should still check for updates once if a commit is passed to actually grab the version that was requested. I’ll try and get a fix out soon!

PR #258904, opened 2025-07-30
Fix --commit-id flag for code serve-web

justin39:
Fixes a bug where passing --commit-id during `code serve-web` would prevent serve-web from downloading a new version of the client at all.

See [this comment](https://github.com/microsoft/vscode/issues/255218#issuecomment-3137695656) from #255218 for additional context.

### DIFF
--- a/cli/src/commands/serve_web.rs
+++ b/cli/src/commands/serve_web.rs
@@ -91,6 +91,12 @@ pub async fn serve_web(ctx: CommandContext, mut args: ServeWebArgs) -> Result<i3
 	if args.commit_id.is_none() {
 		cm.clone()
 			.start_update_checker(Duration::from_secs(update_check_interval));
+	} else {
+		// If a commit was provided, invoke get_latest_release() once to ensure we're using that exact version;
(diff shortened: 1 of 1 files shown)


cli/src/commands/serve_web.rs
```diff
@@ -91,6 +91,12 @@ pub async fn serve_web(ctx: CommandContext, mut args: ServeWebArgs) -> Result<i3
 	if args.commit_id.is_none() {
 		cm.clone()
 			.start_update_checker(Duration::from_secs(update_check_interval));
+	} else {
+		// If a commit was provided, invoke get_latest_release() once to ensure we're using that exact version;
+		// get_latest_release() will short-circuit to args.commit_id.
+		if let Err(e) = cm.clone().get_latest_release().await {
```
connor4312: You should not need a clone here

```suggestion
		if let Err(e) = cm.get_latest_release().await {
```

rebornix (approved):

rzhao271 (approved):

PR #258931, opened 2025-07-31
Todo items with description tooltip

digitarald:
Description otherwise isn't shown


src/vs/workbench/contrib/chat/browser/media/chat.css
```diff
@@ -2416,6 +2416,17 @@ display: none;
 	padding: 2px 0;
 }
 
+.chat-todo-list-widget .todo-item[title] {
+	cursor: help;
```
bhavyaus: Should this be a question mark? Could we instead use the default (hand) icon to keep it consistent with the other hover icons?

digitarald: 🤖 Conducted comprehensive analysis of tooltip and cursor usage across VS Code codebase:

**Key Findings:**
- `cursor: help` appears **nowhere else** in the entire VS Code codebase
- Standard pattern is no special cursor for tooltips
- Hover states consistently use `var(--vscode-list-hoverBackground)`
- Examples: Settings view, tree views, progress lists all follow this pattern

bhavyaus (approved):

PR #258951, opened 2025-07-31
added summarizeReleseNotes with chat query and tied to hide copilot command

eli-w-king:


src/vs/workbench/contrib/update/browser/update.ts
```diff
@@ -516,6 +516,15 @@ export class UpdateContribution extends Disposable implements IWorkbenchContribu
 		CommandsRegistry.registerCommand('_update.state', () => {
 			return this.state;
 		});
+
+		// Register secure command to summarize release notes without exposing arbitrary prompts
+		CommandsRegistry.registerCommand('summarize.release.notes', (accessor: ServicesAccessor) => {
```
bhavyaus, line 521: This should be in update.contribution.ts

bhavyaus, line 521: command id usually starts with workbench.action.

PR #259033, opened 2025-07-31
Fix remote and multi-root issues with task tools

### LINKED ISSUE #258720: Task not found error when the task exists
Testing #258358

tool call
```
id   : call_sJufapG6WYDJHmUG6otndWCz__vscode-1753824250347
tool : run_task
args : {
  "id": "Run Bingo API",
  "workspaceFolder": "/home/kanadig/GIT/TPI/2025-07/bingo2"
}
```

task.json:
```
{
  "version": "2.0.0",
  "tasks": [
    {
      "label": "Run Bingo API",
      "type": "shell",
      "command": "python -m uvicorn main:app --reload",
      "problemMatcher": []
    }
  ]
}

```

### LINKED ISSUE #258723: All `get_terminal_output` failed, when using the pid passed in context
Testing #258358

Image

It does not find the task terminal:
```
id   : call_8a4YKZbe6SQS9lNi3Ok0UeAW__vscode-1753824250351
tool : get_terminal_output
args : {
  "id": "18556"
}

id   : call_Dre3OXsHoJUcbquuea7oGchU__vscode-1753824250354
tool : get_terminal_output
args : {
  "id": "14944"
}

id   : call_6f9ziWZQpr2klIpspEPEe25e__vscode-1753824250353
tool : get_terminal_output
args : {
  "id": "18556"
}
```

same result in all three calls:
```
ERROR while calling tool: Invalid terminal ID
Please check your input and try again.
```

It seems to not be able to find the running task:
```
id   : call_Z573lugQhwlyuhhUnfknvUZu__vscode-1753824250349
tool : get_task_output
args : {
  "id": "Run Bingo API",
  "workspaceFolder": "/home/kanadig/GIT/TPI/2025-07/bingo2"
}

Task not found: Run Bingo API
```

task.json:
```
{
  "version": "2.0.0",
  "tasks": [
    {
      "label": "Run Bingo API",
      "type": "shell",
      "command": ".venv/bin/uvicorn main:app --reload",
      "problemMatcher": []
    }
  ]
}
```

Image

this is the context it has:
```
<context>
The current date is July 29, 2025.
Tasks:
Task: Run Bingo API (, is running, type: shell, command: .venv/bin/uvicorn main:app --reload)
Terminals:
Terminal: bash
Output: {Use get_terminal_output for terminal with ID: 14944.}
Terminal: Run Bingo API
Output: {Use get_terminal_output for terminal with ID: 18556.}
User created Terminals:
Terminal: bash
Output: {Use get_terminal_output for terminal with pid: 14944.}
Terminal: Run Bingo API
Output: {Use get_terminal_output for terminal with pid: 18556.}

</context>
```

### LINKED ISSUE #259014: task tools do not work for multi-root workspaces
They never have AFAIK

> In the vscode/chat multi-workspace, Copilot could not find tasks:
> 
> Image 

 _Originally posted by @digitarald in [#258358](https://github.com/microsoft/vscode/issues/258358#issuecomment-3138247894)_

meganrogge:
1) Task tools weren't working for remotes

It can take longer in a remote window for the tasks service to update after a config change, such as when a task is added to the `tasks.json`. Before, we had a static timeout of 200 ms. Now, we poll for up to 5 seconds to wait for the task to be discovered. 

The key for tasks in a remote window is different from the local case, so we have to parse the URI to check for the workspace folder 
Screenshot 2025-07-31 at 11 39 44 AM
Screenshot 2025-07-31 at 11 37 30 AM

2) Task tools weren't working for mult-root repos, just needed to loop over all folders to find task config

fixes #258720
fixes #258723
fixes https://github.com/microsoft/vscode/issues/259014

### DIFF
--- a/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/task/createAndRunTaskTool.ts
+++ b/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/task/createAndRunTaskTool.ts
@@ -84,7 +84,6 @@ export class CreateAndRunTaskTool implements IToolImpl {
 		if (!exists) {
 			await this._fileService.createFile(tasksJsonUri, VSBuffer.fromString(tasksJsonContent), { overwrite: true });
 			_progress.report({ message: 'Created tasks.json file' });
-			await timeout(200);
 		} else {
 			// add to the existing tasks.json file
 			const content = await this._fileService.readFile(tasksJsonUri);
@@ -92,10 +91,18 @@ export class CreateAndRunTaskTool implements IToolImpl {
 			tasksJson.tasks.push(newTask);
 			await this._fileService.writeFile(tasksJsonUri, VSBuffer.fromString(JSON.stringify(tasksJson, null, '\t')));
 			_progress.report({ message: 'Updated tasks.json file' });
-			await timeout(200);
 		}
 		_progress.report({ message: new MarkdownString(localize('copilotChat.fetchingTask', 'Resolving the task')) });
-		const task = (await this._tasksService.tasks())?.find(t => t._label === args.task.label);
+
+		let task: Task | undefined;
+		const start = Date.now();
+		while (Date.now() - start < 5000 && !token.isCancellationRequested) {
+			task = (await this._tasksService.tasks())?.find(t => t._label === args.task.label);
+			if (task) {
+				break;
+			}
+			await timeout(100);
+		}
 		if (!task) {
 			return { content: [{ kind: 'text', value: `Task not found: ${args.task.label}` }], toolResultMessage: new MarkdownString(localize('copilotChat.taskNotFound', 'Task not found: `{0}`', args.task.label)) };
 		}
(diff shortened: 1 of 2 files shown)


src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/task/createAndRunTaskTool.ts
```diff
@@ -84,18 +84,25 @@ export class CreateAndRunTaskTool implements IToolImpl {
...
+
+		let task: Task | undefined;
+		const start = Date.now();
+		while (Date.now() - start < 5000 && !token.isCancellationRequested) {
+			task = (await this._tasksService.tasks())?.find(t => t._label === args.task.label);
+			if (task) {
+				break;
+			}
+			await timeout(100);
+		}
```
Tyriar, line 105: If there's an event for adding tasks that would be a better approach, we can then do:

```ts
const found = await raceTimeout(new Promise<boolean>(r => {
  // Register and listen to event, resolve only if it matches
}, 5000);
// Handle cancel
```

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/task/taskHelpers.ts
```diff
@@ -56,7 +63,18 @@ export async function getTaskForTool(id: string | undefined, taskDefinition: { t
 	if (!task) {
 		return;
 	}
-	const configuringTasks: IStringDictionary<ConfiguringTask> | undefined = (await taskService.getWorkspaceTasks())?.get(URI.file(workspaceFolder).toString())?.configurations?.byIdentifier;
+
+	let tasksForWorkspace;
+	for (const [folder, tasks] of workspaceFolderToTaskMap) {
+		if (URI.parse(folder).path === workspaceFolder) {
```
Tyriar, line 69: Having `workspaceFolder` as a string here isn't ideal, comparing 2 URIs should be safer

Tyriar (approved):

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/task/taskHelpers.ts
meganrogge: We get workspace folder from the model as an arg. Don't think we can ask it for a URI

Issue #252105, opened 2025-06-22
GitHub Copilot "Copy" icon truncates text


thdoan:
Does this issue occur when all extensions are disabled?: Yes

Version: 1.101.1 (user setup)
Commit: 18e3a1ec544e6907be1e944a94c496e302073435
Date: 2025-06-18T13:35:12.605Z
Electron: 35.5.1
ElectronBuildId: 11727614
Chromium: 134.0.6998.205
Node.js: 22.15.1
V8: 13.4.114.21-electron.0
OS: Windows_NT x64 10.0.26100

Steps to Reproduce:

1. Get the model to respond with one long line of code
2. Copy the text by pressing the "Copy" icon
3. Paste the text somewhere
4. Observe issue: the text is truncated

Here's a recording:

https://github.com/user-attachments/assets/b0964f31-b385-402e-9991-a2990628a92f

joshspicer:
Interesting, this happens consistently for you?

roblourens:
Seems that there must be a race somewhere in updating the codeblock text model but I can't repro it

thdoan:
Thank you for looking into this. Since I can no longer reproduce this in the latest version of VS Code, I'm closing this. If I encounter this again, I'll open a new issue and link it to this.

roblourens:
I've had multiple reports of this issue, so I would believe that it's still there :)

u84six:
This is still an issue with the latest version of vscode (1.102.3). I can reproduce this issue all the time. I've disabled any suspected extension that I have running (e.g. Prettier, Prettier-TypeScript, Better Comments, etc.) I have nothing else enabled that's associated with the editor and copilot (other than the copilot extensions of course). I'm curious as to why my setup is seeing this issue vs everyone else, because it's an extremely annoying problem, almost as bad as if Windows itself had a copy&paste defect :). So I guess it's just a small set of users who are experiencing this. It's very strange. The only other unique difference in my setup than most, is that I use Git Bash for Windows with vscode and all my tools. I don't use Windows Subsystem because I get all the commands that I need with git bash (no need for the overkill).

My original issue was closed because it was marked as a dupe: [#256129](https://github.com/microsoft/vscode/issues/256129) but I've added more information there since then

u84six:
This is a bit blurring because I had to reduce the resolution to github's max attachment size, but you can still see it happening

![Image](https://github.com/user-attachments/assets/efef5f76-31a4-42ae-9164-4c4c701c7d5c)

[...]

PR #259040, opened 2025-07-31
Update code block actions context when model changes

mjbvz:
Possible fix for #252105 which I've seen a few reports of 

I'm having trouble repoing #252105 but one possible cause is that the text model changes outside of the codeBlockPart getting updated. Assigning to August since I'm not super confident in the fix


roblourens:
Nice, let's try. I was able to repro previously, the trick was having multiple codeblocks involved. I'll try later.

roblourens:
I was able to repro with this prompt `Now give me a python sample and the equivalent in ts`. Then I checked out your PR and tried several times and could not repro. Good enough for me

roblourens (approved):

Issue #259038, opened 2025-07-31
Issues with switching to new coding agent chat editor (perf, chat title)


bamurtaugh:
Latest VS Code Insiders and GHPR pre-release

Version: 1.103.0-insider (user setup)
Commit: 724eea79bcad9a344a22eaecf949bca2f5f8f8bc
Date: 2025-07-31T05:03:35.073Z
Electron: 37.2.3
ElectronBuildId: 12035395
Chromium: 138.0.7204.100
Node.js: 22.17.0
V8: 13.8.500258-electron.0
OS: Windows_NT arm64 10.0.26100

**Steps to Reproduce:**

1. Kick off a new coding agent session. In the recording below, it's `[WIP] Update Link for "Learn about...`
2. In the chat sessions view, switch to an already complete coding agent session. In the recording below, the complete session is `Add Cancel Review button to...`
3. Run into a few issues when switching:
- As I've been switching between opening these sessions, the selection highlighting is incorrect (even though I only have the WIP session open at the beginning, the highlighting looks like I have the `Add Cancel Review button...` session open)
     - <img width="500" height="171" alt="Image" src="https://github.com/user-attachments/assets/6e0735c7-47b9-487d-a6b6-25f40f407e90" />
- Switching to another chat editor takes a while (and due to this, it almost looks like the content getting streamed to the WIP session is part of the second session)
     - Potentially related to https://github.com/microsoft/vscode/issues/258616 
- When I open the file diff, the title of the second chat becomes `----`
     - Potentially related to https://github.com/microsoft/vscode/issues/258596
     - <img width="1668" height="127" alt="Image" src="https://github.com/user-attachments/assets/4119644c-5798-4e74-bd76-e27e22c339d8" />

![Image](https://github.com/user-attachments/assets/fd8586ea-8814-4d7e-b471-c82fc4e76694)

bamurtaugh:
Things have evolved since I last reported - here's an updated look at what I still encounter.

- Opening additional coding agent sessions results in titles like `Chat 2`:

<img width="1122" height="108" alt="Image" src="https://github.com/user-attachments/assets/4076139b-7df7-466a-82a6-ec237170b19b" />

- Chat session titles become `---` when I continue to open further sessions, and this is also represented in the local chat sessions history

<img width="1371" height="135" alt="Image" src="https://github.com/user-attachments/assets/800a161d-98d1-466b-90ca-530ef2723d6f" />

[...]

PR #259108, opened 2025-07-31
Set title for contributed chat session

### LINKED ISSUE #258596: Title of coding agent chat editor is inconsistent
Testing #258320

After clicking `Add shellIntegrationNonce...` to open the editor it shows Chat 2:

Image

Then clicking Allow tasks to be run via terminal suggest, Chat 2 becomes `Initial implementation` for some reason:

Image

It should say `Add shellIntegration...`

joshspicer:
closes https://github.com/microsoft/vscode/issues/258596
ref https://github.com/microsoft/vscode/issues/259038

image

### DIFF
--- a/src/vs/platform/editor/common/editor.ts
+++ b/src/vs/platform/editor/common/editor.ts
@@ -45,11 +45,6 @@ export interface IBaseUntypedEditorInput {
 	 * Description to show for the input.
 	 */
 	readonly description?: string;
-
-	/**
-	 * The preferred title for the input.
-	 */
-	readonly preferredTitle?: string;
 }
 
 export interface IBaseResourceEditorInput extends IBaseUntypedEditorInput {
--- a/src/vs/platform/editor/common/editor.ts
+++ b/src/vs/platform/editor/common/editor.ts
@@ -45,6 +45,11 @@ export interface IBaseUntypedEditorInput {
 	 * Description to show for the input.
 	 */
 	readonly description?: string;
+
+	/**
+	 * The preferred title for the input.
+	 */
+	readonly preferredTitle?: string;
 }
 
 export interface IBaseResourceEditorInput extends IBaseUntypedEditorInput {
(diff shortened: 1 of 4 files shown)


src/vs/platform/editor/common/editor.ts
```diff
@@ -320,6 +325,11 @@ export interface IEditorOptions {
 		 */
 		alwaysOnTop?: boolean;
 	};
+
+	/**
+	 * A hint of a possible title for the input
+	 */
+	readonly preferredTitle?: string;
```
joshspicer: I had to add this to get the types to line up here:

https://github.com/microsoft/vscode/blob/301883d9b87f9d0b0201ab37d998c6728ebb11ef/src/vs/workbench/contrib/chat/browser/chatSessions.ts#L915

src/vs/workbench/contrib/chat/common/chatModel.ts
```diff
@@ -1556,8 +1563,9 @@ export class ChatModel extends Disposable implements IChatModel {
 		return request;
 	}
 
-	setCustomTitle(title: string): void {
+	public setCustomTitle(title: string): void {
 		this._customTitle = title;
+		this._onDidChange.fire({ kind: 'setCustomTitle', title });
```
roblourens, line 1568: Do you need this event?

src/vs/platform/editor/common/editor.ts
roblourens: I don't understand why you need it on this interface?

src/vs/workbench/contrib/chat/common/chatModel.ts
joshspicer: I don't see the update until I swap tabs if I do not fire the event

src/vs/platform/editor/common/editor.ts
joshspicer: ah, just TS oddness. Moved out to const per your suggestion and was able to remove that

https://github.com/microsoft/vscode/pull/259108/commits/79ee76df2259b3c7576a7ee61a4b0beb2d8e360e

osortega (approved):

PR #259124, opened 2025-08-01
fix get task uri comparison for windows

meganrogge:
addresses #259014 in main

### DIFF
--- a/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/task/taskHelpers.ts
+++ b/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/task/taskHelpers.ts
@@ -4,6 +4,7 @@
  *--------------------------------------------------------------------------------------------*/
 
 import { IStringDictionary } from '../../../../../../base/common/collections.js';
+import { isEqual } from '../../../../../../base/common/resources.js';
 import { URI } from '../../../../../../base/common/uri.js';
 import { IConfigurationService } from '../../../../../../platform/configuration/common/configuration.js';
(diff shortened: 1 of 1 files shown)


src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/task/taskHelpers.ts
```diff
@@ -66,7 +67,8 @@ export async function getTaskForTool(id: string | undefined, taskDefinition: { t
 
 	let tasksForWorkspace;
 	for (const [folder, tasks] of workspaceFolderToTaskMap) {
-		if (URI.parse(folder).path === workspaceFolder) {
+		// Use isEqual to compare URIs for cross-platform compatibility
```
meganrogge, line 70:
```suggestion
```

DonJayamanne (approved):

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/task/taskHelpers.ts
```diff
@@ -66,7 +67,8 @@ export async function getTaskForTool(id: string | undefined, taskDefinition: { t
...
+		if (isEqual(URI.parse(folder), URI.file(workspaceFolder))) {
```
bpasero, line 71: @meganrogge is this a safe use of `URI.file`? What if this is running from web with a remote URI?

meganrogge: yes there was this followup https://github.com/microsoft/vscode/pull/259231/files

PR #259178, opened 2025-08-01
Adding workbench hover to the `View All File Changes` button and making it focusable

### LINKED ISSUE #258530: no hover on checkpoint button
Testing #257830

when I hover my cursor over this button, I am unable to get a text description to pop up explaining what the button does like a lot of other hovers. Would be nice as I check this sometimes to figure out exactly what a button does. It is also potentially an accessibility item if there isn't extra info on this button or if there is an overlay between the bar to collapse the files changes conflicts with the button to show the checkpoint but I am not sure how to evaluate that correctly.

Image

aiday-mar:
fixes https://github.com/microsoft/vscode/issues/258530

### DIFF
--- a/src/vs/workbench/contrib/chat/browser/chatContentParts/chatChangesSummaryPart.ts
+++ b/src/vs/workbench/contrib/chat/browser/chatContentParts/chatChangesSummaryPart.ts
@@ -30,6 +30,7 @@ import { MultiDiffEditorItem } from '../../../multiDiffEditor/browser/multiDiffS
 import { IEditorGroupsService } from '../../../../services/editor/common/editorGroupsService.js';
 import { Emitter } from '../../../../../base/common/event.js';
 import { IHoverService } from '../../../../../platform/hover/browser/hover.js';
+import { localize2 } from '../../../../../nls.js';
 
 export class ChatCheckpointFileChangesSummaryContentPart extends Disposable implements IChatContentPart {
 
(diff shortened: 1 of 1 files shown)


src/vs/workbench/contrib/chat/browser/chatContentParts/chatChangesSummaryPart.ts
```diff
@@ -130,7 +132,11 @@ export class ChatCheckpointFileChangesSummaryContentPart extends Disposable impl
 
 	private renderViewAllFileChangesButton(container: HTMLElement): IDisposable {
 		const button = container.appendChild($('.chat-view-changes-icon'));
+		this.hoverService.setupDelayedHover(button, () => ({
+			content: `View All File Changes`
```
Tyriar: Needs to be localized

src/vs/workbench/contrib/chat/browser/chatContentParts/chatChangesSummaryPart.ts
```diff
@@ -130,7 +132,11 @@ export class ChatCheckpointFileChangesSummaryContentPart extends Disposable impl
...
+		}));
 		button.classList.add(...ThemeIcon.asClassNameArray(Codicon.diffMultiple));
+		button.tabIndex = 0;
```
Tyriar, line 141: If this is actually a button, it should be `button` or use the aria role button. Otherwise it just looks like focusable text to screen readers.

Tyriar (dismissed):

src/vs/workbench/contrib/chat/browser/chatContentParts/chatChangesSummaryPart.ts
aiday-mar: tracked in https://github.com/microsoft/vscode/issues/262103

src/vs/workbench/contrib/chat/browser/chatContentParts/chatChangesSummaryPart.ts
aiday-mar: tracked in https://github.com/microsoft/vscode/issues/262103

Tyriar:
@aiday-mar these are bugs not debt, can't they just be fixed in this PR? It's just a 2 line change AFAICT

aiday-mar:
Hi @Tyriar yes good point, I mistakenly thought the PR was already merged. I'll make the changes.

bpasero (approved):

PR #259248, opened 2025-08-01
fix: settings link formatting

rzhao271:


src/vs/editor/common/config/editorOptions.ts
```diff
@@ -6084,7 +6084,7 @@ export const EditorOptions = {
 	renderRichScreenReaderContent: register(new EditorBooleanOption(
 		EditorOption.renderRichScreenReaderContent, 'renderRichScreenReaderContent', false,
 		{
-			description: nls.localize('renderRichScreenReaderContent', "Whether to render rich screen reader content when the `editor.editContext` is enabled."),
+			description: nls.localize('renderRichScreenReaderContent', "Whether to render rich screen reader content when `#editor.editContext#` is enabled."),
```
Tyriar: This won't work unless it's markdown description?

Tyriar (changes requested):

Issue #259125, opened 2025-08-01
Editor is not locked to correct chat session on reload


joshspicer:
Does this issue occur when all extensions are disabled?: Yes/No

- VS Code Version: 
- OS Version: 

Steps to Reproduce:

1. Open a chat editor linked to a [chat session](https://github.com/microsoft/vscode/issues/258320)
2. Reload your page
3. When the editor reloads, the content will appear but the input box will not be linked to the session (broken state)

PR #259126, opened 2025-08-01
wait for extension activation in getChatSessionContributions()

joshspicer:
ref https://github.com/microsoft/vscode/issues/259125

Fixes reloading a window with an open chat session editor

PR #259264, opened 2025-08-01
Improve waiting for session contributions

joshspicer:
ref https://github.com/microsoft/vscode/issues/259125


joshspicer:
hm, doesn't seem to work.  Contribution is not loaded yet
image

joshspicer:
There's [readExtensionPointContributions<T>(_extPoint: IExtensionPoint<T>): Promise<ExtensionPointContribution<T>[]> { return Promise.resolve(Object.create(null)); }](https://github.com/microsoft/vscode/blob/b893c30e4cf02d20e7ce7fbb4c8d7234c842716e/src/vs/workbench/services/extensions/common/extensions.ts#L608-L609) which may be correct to use here, but it's also async. The goal of this change vs https://github.com/microsoft/vscode/pull/259126 is to make access contributing sync (not certain if the correct approach)

PR #259349, opened 2025-08-02
fix: tiny memory leak in debugToolBar.ts

alexkozy:
`dispose(IDisposable[])` call disposes every element of the array but does not clear the array itself - but it returns an empty array. As a result `debugViewTitleItems` array accumulates all items and keep references to them forever.

There are two potential solutions:
- `debugViewTitleItems = dispose(debugViewTitleItems)`
- use DisposableStore

This commit implements the second one.


alexkozy:
cc @deepak1556

connor4312 (dismissed):
thanks!

DonJayamanne (dismissed):

alexkozy:
@microsoft-github-policy-service agree company="Anysphere"

alexkozy:
@connor4312 thanks for the review - i removed unused imports to address the hygiene concerns, so it looks like i need another approval.

connor4312 (approved):

dbaeumer (approved):

src/vs/workbench/contrib/debug/browser/debugToolBar.ts
```diff
@@ -392,7 +392,7 @@ export function createDisconnectMenuItemAction(action: MenuItemAction, disposabl
 
 // Debug toolbar
 
-const debugViewTitleItems: IDisposable[] = [];
+const debugViewTitleItems = new DisposableStore();
```
bpasero, line 395: Small nit: this should be marked as singleton as well because it never gets disposed.

PR #259352, opened 2025-08-02
Use app.dock.hide when connecting to an existing instance

JBlitzar:
As discussed in https://github.com/microsoft/vscode/issues/258824#issuecomment-3146091562

(and previously https://github.com/microsoft/vscode/pull/131213#issuecomment-912251877)

Calls app.dock.hide() after another running instance is detected. 

@deepak1556

### DIFF
--- a/src/vs/code/electron-main/main.ts
+++ b/src/vs/code/electron-main/main.ts
@@ -314,6 +314,11 @@ class CodeMain {
 			}
 
 			// there's a running instance, let's connect to it
+			// Hide the dock icon since we're connecting to an existing instance
+			if (app.dock) {
+				app.dock.hide();
+			}
+
 			let client: NodeIPCClient<string>;
 			try {
 				client = await nodeIPCConnect(environmentMainService.mainIPCHandle, 'main');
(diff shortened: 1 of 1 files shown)


deepak1556:
Calling hide at this point might be late we could have a flash of the dock icon. Can we try to start the application with dock hidden and if instance can be claimed then call `dock.show` ?

bpasero:
I like the idea of trying this again, but shouldn't we just restore how it used to be before we removed it? I think I was calling it at a very early point.

bpasero:
Thanks, but I do not feel good restoring anything but the solution we had in place before we reverted it: https://github.com/microsoft/vscode/commit/f8f0c0ec5499cf5d210b12672d71784e93d42f0d

Specifically I think optimistically hiding the dock right on startup could lead to unexpected issues, unless macOS guidelines tell us that this is the official way of doing it.

This might still bring back the issue of multiple dock icons and in that case we will have to revert again, but lets test this in insiders.

Continues in https://github.com/microsoft/vscode/pull/259883

PR #259424, opened 2025-08-03
Add new terminal to tab well and empty editor menus

Tyriar:
This increases discoverability of terminal editors and by extension terminals in aux windows.

Fixes #259002

FYI @bpasero 

image
image


Magdalena497:
852117 Message ID: 0PYGGQiwfmU

DonJayamanne (approved):

bpasero:
@Tyriar I think thats ok, since the menu is otherwise relatively short. Should there be a separator between the first 2 entries that are about files and the terminal?

bpasero:
It is also interesting that we somehow missed out on having "New File" with the picker that asks for which editor to use in the empty case:

image

PR #259501, opened 2025-08-04
Adds doc comments to IStatusbarEntry.content

hediet:


bpasero (changes requested):
Can we document better what the implications are when using this? E.g. are any of the other properties then not working, like kind where colors are set?

Also, what does it mean for showing a custom hover?

bpasero (approved):
Lets get this in at least.

Issue #211083, opened 2024-04-23
Terminal Icon picker weird placement


benibenj:
The terminal icon picker is positioned at the far top of the window. This looks a bit weird as the top border of the widget is not visible. I would have expected the widget to be placed bellow the command center.

<img width="594" alt="image" src="https://github.com/microsoft/vscode/assets/44439583/996bc79b-40ab-4386-b485-72a4961f825a">

Tyriar:
Code pointer: https://github.com/microsoft/vscode/blob/191a248afd908b018bbf05087be8898498e1eec6/src/vs/workbench/contrib/terminal/browser/terminalIconPicker.ts#L60

Tyriar:
@sossost ideally we would position it below the command center when it's triggered via the command palette, or near the terminal tab if it's triggered from there.

tisilent:
Wow.

Narendherraj:
@Tyriar Like @abhijit-chikane has expressed his concern in the PR raised should we probably change the position of the terminal icon colour change tab as well if not it might look weird where one shows the position on bottom and the other on top . 
![image](https://github.com/microsoft/vscode/assets/44496526/bedae4a0-1be8-4357-94e9-a2443554ef1e)

bricefriha:
@Tyriar I'm working on this one if you don't mind. 
I took over the PR but I had to make some changes to it because the icon picker was still at the top when there was only one terminal instance.

I'm also changing the color picker so we have it positioned like the icons picker

ZaakirDev:
I want to work on this issue as my first ever proper issue. Can I be assigned to this?

ZaakirDev:
How do I test this code?

Tyriar:
@benibenj FYI I don't think we should do the placement change for the command center as the quick pick puts itself over it. It does however hide the command center which we should do instead to remain consistent.

The icon hover attached to the tab should still happen though.

Svea90:
/| Header | Header | Header | Header | Header |

Testing

Tyriar:
The widget now has an arrow on the bottom of it:

![Image](https://github.com/user-attachments/assets/29633b0b-6ec8-475a-a12d-5f1316e85e7c)

Destroyer795:
Hello @benibenj, I see that the issue is still open and labeled as good first issue. I would like to contribute to this issue as my first open-source contribution. Could you please confirm if its still available?

benibenj:
@Destroyer795 yes it is

[...]

PR #259513, opened 2025-08-04
fix: position terminal icon picker dynamically based on trigger source

sanchitkhthpalia:
This PR fixes the placement issue with the terminal icon picker widget, as discussed in #211083.

✅ **Changes made:**
- Updated `TerminalIconPicker` so it now accepts a `source` parameter (`'commandPalette'` or `'terminalTab'`).
- The picker is now dynamically positioned:
  - **When triggered via the command palette**: appears centered below the command center.
  - **When triggered via the terminal tab**: appears near the terminal tab area.
- Adjusted `TerminalInstance.changeIcon` to call the picker with the correct source.

📌 **Why:**
Previously, the icon picker always appeared near the top center, which could look odd or be partially hidden.
This makes the UX more intuitive by placing the picker close to where it was triggered.

✅ **How to test:**
- Trigger the terminal icon picker via the terminal tab ➜ it should appear near the tab area.
- Trigger via the command palette ➜ it should appear centered below the command center.

---

**Fixes:** #211083  
✔️ Allows edits by maintainers

### DIFF
--- a/src/vs/workbench/contrib/terminal/browser/terminalIconPicker.ts
+++ b/src/vs/workbench/contrib/terminal/browser/terminalIconPicker.ts
@@ -51,7 +51,7 @@ export class TerminalIconPicker extends Disposable {
 		});
 	}
 
-	async pickIcons(): Promise<ThemeIcon | undefined> {
+	async pickIcons(source: 'commandPalette' | 'terminalTab'): Promise<ThemeIcon | undefined> {
 		const dimension = new Dimension(486, 260);
 		return new Promise<ThemeIcon | undefined>(resolve => {
 			this._register(this._iconSelectBox.onDidSelect(e => {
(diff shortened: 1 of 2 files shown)


sanchitkhthpalia:

src/vs/workbench/contrib/terminal/browser/terminalIconPicker.ts
```diff
@@ -51,22 +51,34 @@ export class TerminalIconPicker extends Disposable {
...
+
+			let targetX = bodyRect.left + (bodyRect.width - dimension.width) / 2;
+			let targetY = bodyRect.top + this._layoutService.activeContainerOffset.quickPickTop - 2;
+
+			if (source === 'terminalTab') {
+				// Example positioning: near top-left where terminal tabs usually are
+				targetX = bodyRect.left + 50; // adjust as needed
+				targetY = bodyRect.top + 100;  // adjust as needed
+			}
+			// else keep default centered position for commandPalette
```
Tyriar, line 74: This doesn't solve the problem at all and just positions it randomly based on arbitrary values. Getting the real values is literally everything for this issue. (50, 100) puts it randomly in the top left corner 🤔

PR #259633, opened 2025-08-04
Clean up markdownDocumentRenderer api

mjbvz:
Similar changes to those made for markdown renderer:

- Clarify api and how override vs augment works
- Move most defaults into callers

@bhavyaus Let me know if you notice any issues with welcome flows since these seemed to use the customization most heavily

### DIFF
--- a/src/vs/base/browser/markdownRenderer.ts
+++ b/src/vs/base/browser/markdownRenderer.ts
@@ -434,7 +434,6 @@ export const allowedMarkdownHtmlAttributes = [
 	'autoplay',
 	'alt',
 	'checked',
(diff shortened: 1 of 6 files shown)


src/vs/base/browser/markdownRenderer.ts
```diff
@@ -434,7 +434,6 @@ export const allowedMarkdownHtmlAttributes = [
 	'autoplay',
 	'alt',
 	'checked',
-	'class',
```
mjbvz, line 437: Was already included below

lramos15 (approved):

Issue #259100, opened 2025-07-31
[NuGet MCP] Add proposal API for mapping MCP server.json to mcp.json


joelverhagen:
It appears there is no way for extensions to invoke this routing via the VS Code plugin API:
https://github.com/microsoft/vscode/blob/3eaa65007ec41a0426295bea00914814bd5c0650/src/vs/platform/mcp/common/mcpManagementService.ts#L235

I am working on the Copilot Chat extension and have NuGet packages which contain an embedded `.mcp/server.json` file. I would like to map this to VS Code MCP JSON using the `toScannedMcpServerAndInputs` method.

That way, it can be flowed through the assisted installation flow.

I propose a new `enabledApiProposals` is added that exposes this mapping behavior.

Alternatively, the contract of `github.copilot.chat.mcp.setup.flow` could be changed to allow `server.json` format to be returned while still supporting VS Code's JSON format also -- perhaps with an enum to differentiate, thus allowing VS Code internally to do the mapping.

Related to https://github.com/microsoft/vscode/issues/257679

connor4312:
Kind of related to potential DXT support as well.

sandy081:
@joelverhagen I am probably missing some or ton of context here after coming back from vacation. Can you please explain in detail what is the purpose of this request and what it is trying to achieve?

joelverhagen:
Sure, the goal here is for there to be a way _somehow_ to allow VS Code extensions to invoke the logic that maps `server.json` to VS Code MCP configuration. This allows, for example, the AI-assisted MCP installation flow to reliably handle a server.json found inside of a package during installation. MCP servers in NuGet packages have embedded server.json files which can be used to improve the installation experience since they have a well defined schema and have a known mapping to VS Code.

This has currently been achieved via https://github.com/microsoft/vscode/pull/259634/files/b5695319e680c9cdddc6f9458b9303a0a41610bb#diff-21f5e7877ee70a81c338dfa19a7268e5588393d8dd9c5f56ec40f2917d472a85.
In short, when uses `MCP: Add Server...` > `NuGet package` > provide some package ID, the chat extension will check the NuGet package for a server.json and return it to VS Code. VS Code will use `toScannedMcpServerAndInputs` to produce the VS Code MCP JSON and then fall into the same flow as before. The alternative is having the VS Code MCP JSON generated by an AI-assisted tool loop based on README Markdown content. Happy to chat on Teams or a call if that helps.

connor4312:
[...]

PR #259634, opened 2025-08-04
Support server.json being returned by assisted MCP install

joelverhagen:
This works around the lack of https://github.com/microsoft/vscode/issues/259100 and is progress on https://github.com/microsoft/vscode/issues/257679. There will be a separate PR on the chat extension to support server.json in NuGet packages.

It enhances the `github.copilot.chat.mcp.setup.flow` command to have an additional `type` property. When this type property is set to `"server.json"`, the result is treated as an MCP Registry server.json and mapped to the VS Code configuration format. If the `type` property is falsey set or set to `"vscode"`, the existing codepath is used.

Additionally, I updated the `toScannedMcpServerAndInputs` to be a public static method so it can be called after the `github.copilot.chat.mcp.setup.flow` command returns. 

I noticed the resulting `args` and `env` placeholder values had the format `{input:foo}`, instead of format `${input:foo}`. Based on https://code.visualstudio.com/docs/reference/variables-reference#_input-variables, it seems like this was a bug in the original implementation, but I may be wrong. @sandy081 - could you confirm?

Verified with:
- npm: `@azure/mcp` (old flow, no server.json)
- Python: `mcp-server-fetch` (old flow, no server.json)
- Docker: `mcp/node-code-sandbox` (old flow, no server.json)
- NuGet: `NuGet.Mcp.Server` (new flow), `Knacode.SampleMcpServer` (new flow, this one has an input)

### DIFF
--- a/src/vs/platform/mcp/common/mcpManagementService.ts
+++ b/src/vs/platform/mcp/common/mcpManagementService.ts
@@ -325,10 +325,10 @@ export abstract class AbstractMcpResourceManagementService extends Disposable {
 			}
 			else if (serverPackage.registry_name === PackageType.NUGET) {
 				args.push(serverPackage.version ? `${serverPackage.name}@${serverPackage.version}` : serverPackage.name);
-			}
-
-			if (serverPackage.package_arguments && serverPackage.registry_name === PackageType.NUGET) {
-				args.push('--');
+				args.push('--yes'); // installation is confirmed by the UI, so --yes is appropriate here
+				if (serverPackage.package_arguments?.length) {
+					args.push('--');
+				}
 			}
 
 			for (const arg of serverPackage.package_arguments ?? []) {
--- a/src/vs/platform/mcp/common/mcpManagementService.ts
+++ b/src/vs/platform/mcp/common/mcpManagementService.ts
@@ -232,7 +232,7 @@ export abstract class AbstractMcpResourceManagementService extends Disposable {
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


connor4312 (approved):

roblourens (approved):

src/vs/platform/mcp/common/mcpManagementService.ts
```diff
@@ -232,7 +232,7 @@ export abstract class AbstractMcpResourceManagementService extends Disposable {
 		}
 	}
 
-	protected toScannedMcpServerAndInputs(manifest: IMcpServerManifest, packageType?: PackageType): { config: IMcpServerConfiguration; inputs?: IMcpServerVariable[] } {
+	static toScannedMcpServerAndInputs(manifest: IMcpServerManifest, packageType?: PackageType): { config: IMcpServerConfiguration; inputs?: IMcpServerVariable[] } {
```
sandy081, line 235: @joelverhagen wondering why this is made public and static? This is an internal implementation of handling gallery mcp server installations. Now it is being reused for other installation flows which will make difficult to do any changes/enhancements for gallery installation flows.

sandy081: Our general guideline is to use expose APIs(methods) on services instead of such static functions.

joelverhagen: Hey @sandy081, apologies that I didn't follow that guideline. I was looking for a way to perform the mapping from server.json to VS Code mcp.json and this seemed like the simplest way to get it done. It sounds like what you're saying is that this metho should be made public but not static, and some implementation of `AbstractMcpResourceManagementService` is injected via DI instead? Or am I misunderstanding?

PR #259654, opened 2025-08-04
return file Operation error rather than throwing

### LINKED ISSUE #259618: fileOperations errors are not serialized across EH - Renderer boundary
Steps to Reproduce:

1. enable `"notebook.experimental.remoteSave": true`
2. run a cell that keeps updating the output

```
import time
for i in range(10):
    print("Iteration", i)
    time.sleep(1)
```

2. While that's running, modify the same notebook doc in another editor and save
3. You should get a `File modified since` error

:bug: the error is treated as generic instead of handling that case specifically: https://insiders.vscode.dev/github/microsoft/vscode/blob/copilot/fix-258908/src/vs/workbench/services/workingCopy/common/storedFileWorkingCopy.ts#L1148

Expected error widget options:

Image

the correct error is thrown [here](https://insiders.vscode.dev/github/microsoft/vscode/blob/copilot/fix-258908/src/vs/workbench/api/common/extHostNotebook.ts#L570), but the error type and extra fields are lost in serialization

amunger:
fix https://github.com/microsoft/vscode/issues/259618

### DIFF
--- a/src/vs/workbench/api/common/extHostNotebook.ts
+++ b/src/vs/workbench/api/common/extHostNotebook.ts
@@ -313,83 +313,91 @@ export class ExtHostNotebookController implements ExtHostNotebookShape {
 		return VSBuffer.wrap(bytes);
 	}
 
-	async $saveNotebook(handle: number, uriComponents: UriComponents, versionId: number, options: files.IWriteFileOptions, token: CancellationToken): Promise<INotebookPartialFileStatsWithMetadata> {
+	async $saveNotebook(handle: number, uriComponents: UriComponents, versionId: number, options: files.IWriteFileOptions, token: CancellationToken): Promise<INotebookPartialFileStatsWithMetadata | files.FileOperationError> {
 		const uri = URI.revive(uriComponents);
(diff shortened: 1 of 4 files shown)


src/vs/workbench/api/common/extHostNotebook.ts
```diff
@@ -313,83 +313,91 @@ export class ExtHostNotebookController implements ExtHostNotebookShape {
...
+				etag: files.etag({ mtime: stat.mtime, size: stat.size }),
+				children: undefined
+			};
 
-		this.trace(`exit saveNotebook(versionId: ${versionId}, ${uri.toString()})`);
-		return fileStats;
+			this.trace(`exit saveNotebook(versionId: ${versionId}, ${uri.toString()})`);
+			return fileStats;
+		} catch (error) {
+			// return fileOperationsErrors to keep the whole object across serialization, these errors are handled specially by the WCS
```
roblourens, line 395: Cool, but I wish our IPC would make this easier for you so you don't have to do all this manually

roblourens (approved):

PR #259691, opened 2025-08-05
Expand the run-tests tool description

digitarald:
Based on recent flows where the agent kept using `npm test` over run tests; self-diagnosed by the agent based on the tool descriptions.


src/vs/workbench/contrib/testing/common/testingChatAgentTool.ts
```diff
@@ -142,8 +139,15 @@ class RunTestTool extends Disposable implements IToolImpl {
...
+		const summary = this._makeModelTestResults(result);
+		const failures = result.counts[TestResultState.Errored] + result.counts[TestResultState.Failed];
+		const content = [{ kind: 'text', value: summary } as const];
+		if (failures > 0) {
+			content.push({ kind: 'text', value: 'Some tests failed. Use the test_failure tool to inspect structured failure details before debugging.' } as const);
+		}
+
 		return {
-			content: [{ kind: 'text', value: this._makeModelTestResults(result) }],
+			content: content as any, // satisfy readonly typing
```
connor4312: Can you avoid the any? E.g. with `Mutable<T>`?

src/vs/workbench/contrib/testing/common/testingChatAgentTool.ts
```diff
@@ -142,8 +139,15 @@ class RunTestTool extends Disposable implements IToolImpl {
 			};
 		}
 
+		const summary = this._makeModelTestResults(result);
+		const failures = result.counts[TestResultState.Errored] + result.counts[TestResultState.Failed];
+		const content = [{ kind: 'text', value: summary } as const];
+		if (failures > 0) {
+			content.push({ kind: 'text', value: 'Some tests failed. Use the test_failure tool to inspect structured failure details before debugging.' } as const);
```
connor4312: We should provide any necessary information here and not instruct the model to call another tool.

src/vs/workbench/contrib/testing/common/testingChatAgentTool.ts
```diff
@@ -71,23 +71,19 @@ class RunTestTool extends Disposable implements IToolImpl {
...
 			type: 'object',
 			properties: {
 				files: {
 					type: 'array',
-					items: {
-						type: 'string',
-					},
-					description: 'Absolute paths to the test files to run. If not provided, all test files will be run.',
+					items: { type: 'string' },
+					description: 'Array of absolute or workspace-relative paths to test files (or directories containing tests). Narrow selection to reduce run time. Omit to run all discovered tests in the workspace.',
```
connor4312: We should keep demanding absolute paths even if we accept absolute paths. This matches other tools and how we instruct the model.

src/vs/workbench/contrib/testing/common/testingChatAgentTool.ts
```diff
@@ -71,23 +71,19 @@ class RunTestTool extends Disposable implements IToolImpl {
 		canBeReferencedInPrompt: true,
 		when: TestingContextKeys.hasRunnableTests,
 		displayName: 'Run tests',
-		modelDescription: 'Runs unit tests in files. Use this tool if the user asks to run tests or when you want to validate changes using unit tests. When possible, always try to provide `files` paths containing the relevant unit tests in order to avoid unnecessarily long test runs.',
+		modelDescription: 'Run workspace tests through the VS Code Testing API to quickly validate code changes and re-run known failing tests. Accepts test file paths (absolute or workspace-relative) and/or substrings of test or suite names to narrow what runs. Returns a structured summary of pass/fail counts and failure messages (not full raw runner stdout). If you need custom test runner CLI flags (e.g., jest --testNamePattern, mocha -g, pytest -k) or full console output for deep debugging, prefer using a terminal command (npm test / yarn test / cargo test, etc.). Always provide the narrowest set of files and/or test names you can to save time.',
```
connor4312: - The "If you need custom test runner CLI flags" line is a little confusing, the testNamePattern and `-g` flag on mocha are both things that could be expressed with a tool call and _should_ go through this tool, not the terminal
- "full console output" --  this can be provided via the testing API, a need for "full console output" should not cause the model to use the terminal tool
- Is "through the VS Code Testing API" something the model needs to know about / something that would influence its behavior?

connor4312 (changes requested):

src/vs/workbench/contrib/testing/common/testingChatAgentTool.ts
digitarald: Added full failure details

src/vs/workbench/contrib/testing/common/testingChatAgentTool.ts
```diff
@@ -72,24 +72,20 @@ class RunTestTool implements IToolImpl {
...
+					description: 'Array of absolute or workspace-relative paths to test files (or directories containing tests). Narrow selection to reduce run time. Omit to run all discovered tests in the workspace.',
 				},
 				testNames: {
 					type: 'array',
-					items: {
-						type: 'string',
-					},
-					description: 'An array of test names to run. Depending on the context, test names defined in code may be strings or the names of functions or classes containing the test cases. If not provided, all tests in the files will be run.',
+					items: { type: 'string' },
+					description: 'Array of case-insensitive substrings matched against test item labels (suite/class/case). Use to run only specific tests within the provided files. Omit to run all tests in those files.',
```
connor4312: I don't think we need to include that they're case insensitive for the model description. We can accept case insensitivity, but that doesn't change how the model calls this tool

connor4312 (commented):
There were some comments from my last review still pending

connor4312:
Still pending some comments up there ^ btw 🙂

digitarald:
> Still pending some comments up there ^ btw 🙂

Sorry, this has been lingering for too long with active feedback.

connor4312 (approved):

PR #259729, opened 2025-08-05
Add instructions for new APIs

rebornix:
Provide detailed guidelines for adding new proposed APIs to VS Code, covering the necessary layers, communication protocols, and implementation steps.

### DIFF
--- a/.github/instructions/api.instructions.md
+++ b/.github/instructions/api.instructions.md
@@ -1,0 +1,170 @@
+---
+applyTo: 'src/vscode-dts/**/*.d.ts,src/vs/workbench/api/**/*.ts'
+---
+
+# VS Code API Development
+
+<!-- TODO: Add more details about the API development process -->
+
+## Adding a New Proposed API
+
+When adding a new proposed API to VS Code, implement two main layers that communicate via RPC:
+
+- **Extension Host API layer** (`extHost*.ts`): The implementation of the API that extensions call.
+- **Renderer Process API counterparts** (`mainThread*.ts`): These run in the renderer process and are the counterparts to the `extHost` APIs. They interact with shared VS Code services.
+- **Shared Services**: Central services that manage state and functionality, coordinating contributions from different `mainThread` instances.
+- **Communication protocol** (`extHost.protocol.ts`): The RPC contract between the two layers.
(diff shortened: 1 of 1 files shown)


.github/instructions/api.instructions.md
```diff
@@ -0,0 +1,170 @@
...
+export const ExtHostContext = {
+    ExtHostYourFeature: createProxyIdentifier<ExtHostYourFeatureShape>('ExtHostYourFeature'),
+};
+```
+
+**Key Requirements:**
+- Methods that cross the process boundary (RPC calls) must be prefixed with `$` and return a `Promise` or `Thenable`.
+- Complex objects are serialized when passed between processes, which means they lose their methods. You may need to create converters for them.
+
+### Implement the Extension Host part
```
alexr00, line 70: This is often where the proposal check occurrs.

.github/instructions/api.instructions.md
```diff
@@ -0,0 +1,170 @@
+---
+applyTo: 'src/vscode-dts/**/*.d.ts,src/vs/workbench/api/**/*.ts'
+---
+
+# VS Code API Development
+
+<!-- TODO: Add more details about the API development process -->
```
alexr00, line 9: API guidelines are here: https://github.com/microsoft/vscode/wiki/Extension-API-guidelines

alexr00 (commented):
Will having API instructions encourage Copilot to add more API? In general, we try to not have unintentionally infinite proposals, so having Copilot create more proposals goes against that. However, there are certainly times where it would be nice if it would correctly add a new small proposal.

rebornix:
@alexr00 @mjbvz the reason I'm creating an instruction file for the api architecture is more about explaining to both the team members and coding agent on how API works in general, not necessarily using the coding agent to craft new proposed APIs. What I learnt last week while working with the team on bringing @mjbvz 's API proposal is there are quite some triable knowledge how data transfer, why we have ExtHost/Mainthread counterparts and how they relate to multi extension hosts. This draft is my initial attempt to close the gap and it would be awesome if you can push directly to this branch and get this merged in your preferred way (I'll be out thus I'll leave it to you).

PR #259753, opened 2025-08-05
Fix memory reference handling in Watch window

DrSergei:
This patch fixes a small problem when the `View Binary Data` button showed in the `Watch` window in case of an error returned from the debug adapter (`memoryReference` field is not updated).

Before:
memory-reference-before

After:

memory-reference-after


connor4312 (approved):

connor4312:
thanks!

Tyriar (approved):

PR #259839, opened 2025-08-05
Fix theme not being synchronized with external windows on firefox

### LINKED ISSUE #259835: Firefox: changes to global stylesheets are not applied live

- VS Code Version: 1.103.0-insider
- OS Version: Ubuntu 24.04

Steps to Reproduce:

1. Go to https://vscode.dev on the latest version of Firefox
2. Create a new unamed test file
3. Right click on the tab title, then `Copy in new window`
4. Open the command picker, type `color theme` then select `Preferences: Color Theme`
5. While the external window is still visible, change the theme to a white theme
6. In the external window, the theme is partially applied and looks completely broken

CGNonofr:
fix #259835

Firefox doesn't allow to watch style element content using a MutationObserver. A MutationObserver is used to synchronize main windows styles with external windows. As a consequence, the theme is not properly updated on external windows on Firefox.

As a workaround, also add a data-hash attribute on the style elements, and watch it as well in the MutationObserver

### DIFF
--- a/src/vs/workbench/services/themes/browser/workbenchThemeService.ts
+++ b/src/vs/workbench/services/themes/browser/workbenchThemeService.ts
@@ -33,7 +33,7 @@ import { updateColorThemeConfigurationSchemas, updateFileIconThemeConfigurationS
 import { ProductIconThemeData, DEFAULT_PRODUCT_ICON_THEME_ID } from './productIconThemeData.js';
 import { registerProductIconThemeSchemas } from '../common/productIconThemeSchema.js';
 import { ILogService } from '../../../../platform/log/common/log.js';
-import { isWeb } from '../../../../base/common/platform.js';
+import { isFirefox, isWeb } from '../../../../base/common/platform.js';
 import { ColorScheme, ThemeTypeSelector } from '../../../../platform/theme/common/theme.js';
 import { IHostColorSchemeService } from '../common/hostColorSchemeService.js';
 import { RunOnceScheduler, Sequencer } from '../../../../base/common/async.js';
@@ -42,6 +42,7 @@ import { getIconsStyleSheet } from '../../../../platform/theme/browser/iconsStyl
 import { asCssVariableName, getColorRegistry } from '../../../../platform/theme/common/colorRegistry.js';
 import { ILanguageService } from '../../../../editor/common/languages/language.js';
 import { mainWindow } from '../../../../base/browser/window.js';
+import { generateUuid } from '../../../../base/common/uuid.js';
 
 // implementation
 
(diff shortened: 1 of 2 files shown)


bpasero (changes requested):
On a glance the direction seems legit, but the implementation is not working for me:
* we need to scope this to only apply for firefox (via `isFirefox`) and not always
* computing the SHA seems overkill, I would do a pragmatic approach and set some UUID or even easier a counter that just increments, this may produce false positives but I think is fair enough for Firefox

CGNonofr:
> On a glance the direction seems legit, but the implementation is not working for me:

by **not working for you** you mean that you tried it and it didn't work, or are you just unhappy with the way it's implemented?

> * we need to scope this to only apply for firefox (via `isFirefox`) and not always

I was mainly for the sake of simplicity, and also because I was anticipating it may also impact some other browsers I'm unable to test  (like safari?)

> * computing the SHA seems overkill, I would do a pragmatic approach and set some UUID or even easier a counter that just increments, this may produce false positives but I think is fair enough for Firefox

I had multiple ideas on how to implement it :thinking: I've started with using a simple counter, an UUID, and at the end I was thinking using a hash was the most elegant way to implement it (Also I'm not sure how to name the attribute if using an UUID :thinking:) I'm aware of the additional computation required, but computing a hash is pretty light, and also it's only done when switching the theme so...

Btw it's really not a big deal, I just want that issue fixed and wanted to help do it quickly by suggesting a way to implement it, feel free to refactor it the way you want. I can also do it with clean indications (what attribute name for the UUID? Are you sure to scope it for Firefox only?)

bpasero:
It works fine for me in all chrome based browsers and safari, so I think its an isolated issue with firefox only. I can do a fix as well but would probably start fresh then.

I am having trouble actually verifying that your fix works running out of sources because somehow NO CSS loads for me in this case. I need to create a build with the fix to verify it.

CGNonofr:
What about now?

bpasero:
@CGNonofr better, but I made some changes on top, though I cannot push them back to your branch. I created a new PR at https://github.com/microsoft/vscode/pull/260037

CGNonofr:
> @CGNonofr better, but I made some changes on top, though I cannot push them back to your branch. I created a new PR at #260037

Sorry, forgot about the UUID tools in VSCode. I've replaced my commit with yours. Feel free to merge either one or the other

bpasero (approved):
Thank you, this is not a trivial contribution, great work. I confirmed it seems fixed with the change on insiders.vscode.dev

CGNonofr:
> I confirmed it seems fixed with the change on insiders.vscode.dev

I've just tried and it doesn't seem to be deployed :thinking: is there some kind of gradual rollout? I'm still on commit c661d0ee523e3aaad18d273087ab1eff0c30b91d

bpasero:
@CGNonofr try with https://insiders.vscode.dev/?vscode-version=67afbe24e877cd0d0769d32de3014821a9fae306

CGNonofr:
Oh thanks, nice feature :)

src/vs/workbench/services/themes/browser/workbenchThemeService.ts
```diff
@@ -793,13 +794,21 @@ class ThemeFileWatcher {
...
-		elStyle.textContent = styleSheetContent;
-	} else {
-		(<HTMLStyleElement>themeStyles[0]).textContent = styleSheetContent;
+	let themeStyle = mainWindow.document.head.getElementsByClassName(rulesClassName).item(0);
+	if (!themeStyle) {
+		themeStyle = createStyleSheet();
+		themeStyle.className = rulesClassName;
+	}
+
+	if (themeStyle.textContent !== styleSheetContent) {
```
aeschli, line 803: @bpasero Should we put that code in a `applyStyleSheetContent` utility in `src/vs/base/browser/domStylesheets.ts`, so that ''data-version'' is an implementation-detail there?
Also I wonder if there are other places where we set the stylesheet content?

aeschli (changes requested):

src/vs/workbench/services/themes/browser/workbenchThemeService.ts
bpasero: @aeschli I can clean this up after the PR landed, its quite hard for me currently to push changes into the remote branch.

aeschli (approved):

Issue #259434, opened 2025-08-03
Compare active file with clipboard editor order


antondevson:
When you run the command `File: Compare Active File with Clipboard` the clipboard is in the left pane of the diff editor, making it more like reverse `File: Compare Clipboard with Active File` in my opinion. The solution would be to plainly change the order and either add another function that does the reverse (currently implemented) order, or add a settings entry where you chose the default behavior.

<img width="516" height="160" alt="Image" src="https://github.com/user-attachments/assets/b5a61c7a-aa86-45d5-af44-e90c513d8425" />

A secondary request would also be to make sure the swap sides button is available for this function (it's not present for clipboard comparison), but that seems to be because of the hacky way the `clipboardCompare{int}` scheme is implemented, and thus I think that could be considered a bug rather than a feature request.

<img width="276" height="108" alt="Image" src="https://github.com/user-attachments/assets/0eac7569-424f-4a9f-b0b0-09b7eef72d79" />

I find myself using this function a lot these days when getting code written by external tools such as Gemini Canvas, and want to quickly see what the actual changes are for each iteration. I've seen this request brought up elsewhere before, with the solution being offered is using the swapdiff extension by shalimski, but that is still a bit buggy (opens a new editor when swapping) and requires an extra step.

I haven't contributed here before so I'm not aware how things work around here, but I could take this upon myself if needed and just implement the first solution, which would be good enough for my current requirements. I don't know how fast things move around here though, some quick insight would be appreciated.

Relevant code: [fileActions.ts#L848](https://github.com/microsoft/vscode/blob/ec0f0be4a71892d46a136811f5a5c160f032f129/src/vs/workbench/contrib/files/browser/fileActions.ts#L827)

```Version: 1.102.2 (user setup)
Commit: c306e94f98122556ca081f527b466015e1bc37b0
Date: 2025-07-22T12:15:48.520Z
Electron: 35.6.0
ElectronBuildId: 11847422
Chromium: 134.0.6998.205
Node.js: 22.15.1
V8: 13.4.114.21-electron.0
OS: Windows_NT x64 10.0.26100

antondevson:
Actually, looking at all the other compare commands I found that they all follow the logical order (first thing mentioned being on the left), so maybe this is to be considered a bug after all.

<img width="348" height="170" alt="Image" src="https://github.com/user-attachments/assets/88920f5e-35ce-4fca-b41f-3cf182427735" />

bpasero:
[...]

PR #259846, opened 2025-08-05
Reversed diff editor order for comparing to clipboard

antondevson:
Changes the editor pane order for comparing active file to clipboard, so that the clipboard is on the right (modified) instead of left (original) side.

Image

This is to reflect the behavior of all the other `Compare...` commands where the active file is on the left.

Image

This is the simplest fix for issue https://github.com/microsoft/vscode/issues/259434 if it is to be considered a bug, which is indicated by the behavior of the other similar commands.

Regarding changing the localization key, I'm not sure if this is needed or not as I'm unaware of how such a change in the message prompts updates for translations, please correct if not needed.

### DIFF
--- a/src/vs/workbench/contrib/files/browser/fileActions.ts
+++ b/src/vs/workbench/contrib/files/browser/fileActions.ts
@@ -860,11 +860,11 @@ export class CompareWithClipboardAction extends Action2 {
 			}
 
 			const name = resources.basename(resource);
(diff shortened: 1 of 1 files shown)


bpasero:
I am not convinced, given the nice flow you get today in https://github.com/microsoft/vscode/issues/259434#issuecomment-3156188594

PR #259851, opened 2025-08-05
Add more safe commands to auto approve

### LINKED ISSUE #259668: Add a set of safe and common readonly commands to terminal auto approve
We should add some `true` defaults to this setting:

https://github.com/microsoft/vscode/blob/1796164e45829c75e93bc583409c3f69db0b7217/src/vs/workbench/contrib/terminalContrib/chatAgentTools/common/terminalChatAgentToolsConfiguration.ts#L88-L101

Tyriar:
Fixes #259668


agreaves-ms (approved):
It will be great to have these additional commands, thank you for the update

mjbvz (approved):

PR #259856, opened 2025-08-05
Fix auto activation of chat providers

### LINKED ISSUE #259759: Extension that provides ChatProvider is not activated when user opens GitHub Copilot Manage Models quick pick
- Copilot Chat Extension Version:   0.30.2025080501
- VS Code Version:
```
Version: 1.103.0-insider (user setup)
Commit: 34769c5d17df269aed44407eb1a1c0e6362e2d79
Date: 2025-08-04T21:43:35.566Z
Electron: 37.2.3
ElectronBuildId: 12035395
Chromium: 138.0.7204.100
Node.js: 22.17.0
V8: 13.8.500258-electron.0
OS: Windows_NT x64 10.0.26100
```
- OS Version: 
- Feature (e.g. agent/edit/ask mode): Manage models
- Selected model (e.g. GPT 4.1, Claude 3.7 Sonnet): n/a
- Logs:

Steps to Reproduce:

1. My extension implements the LanguageModelChatProvider2 interface by following [this example](https://github.com/microsoft/vscode-extension-samples/blob/main/chat-model-provider-sample/package.json)
2. But our extension only activates on files
```
workspaceContains:some_files
```
3. When I some extensions list the models provided by our extension, for example, opening GitHub Copilot Chat => Manage Models quick pick, at this time, if our extension is not automatically activated, so it won't be able to show models provided by our extension.

However, if I manually activate my extension and then click Manage Models, it works well.

Is there an activationEvent when other extension uses our chat provider? Is there any other workaround than setting activationEvents to "*".

lramos15:
Fix #259759

### DIFF
--- a/src/vs/workbench/services/extensions/common/extensionsRegistry.ts
+++ b/src/vs/workbench/services/extensions/common/extensionsRegistry.ts
@@ -394,6 +394,11 @@ export const schema: IJSONSchema = {
 						description: nls.localize('vscode.extension.activationEvents.onChatParticipant', 'An activation event emitted when the specified chat participant is invoked.'),
 					},
 					{
+						label: 'onLanguageModelChat',
+						body: 'onLanguageModelChat:${1:vendor}',
+						description: nls.localize('vscode.extension.activationEvents.onLanguageModelChat', 'An activation event emitted when a chat model provider for the given vendor is requested.'),
+					},
+					{
 						label: 'onLanguageModelTool',
 						body: 'onLanguageModelTool:${1:toolId}',
 						description: nls.localize('vscode.extension.activationEvents.onLanguageModelTool', 'An activation event emitted when the specified language model tool is invoked.'),
--- a/src/vs/workbench/contrib/chat/common/languageModels.ts
+++ b/src/vs/workbench/contrib/chat/common/languageModels.ts
@@ -408,6 +408,9 @@ export class LanguageModelsService implements ILanguageModelsService {
 		if (typeof vendors === 'string') {
 			vendors = [vendors];
 		}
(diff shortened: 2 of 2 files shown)


src/vs/workbench/services/extensions/common/extensionsRegistry.ts
```diff
@@ -393,6 +393,11 @@ export const schema: IJSONSchema = {
 						body: 'onChatParticipant:${1:participantId}',
 						description: nls.localize('vscode.extension.activationEvents.onChatParticipant', 'An activation event emitted when the specified chat participant is invoked.'),
 					},
+					{
+						label: 'onLanguageModelChat',
+						body: 'onLanguageModelChat:${1:vendor}',
+						description: nls.localize('vscode.extension.activationEvents.onLanguageModelChat', 'An activation event emitted when a chat model provider for the given vendor is requested.'),
+					},
```
Tyriar, line 400: Is this a new activation event? They need to go through the API call.

Tyriar (changes requested):

src/vs/workbench/services/extensions/common/extensionsRegistry.ts
lramos15: This is an existing activation event, that is just havine intellisense added to it. You can see the activation event has existed awhile in [main](https://github.com/microsoft/vscode/blob/7e6267da4ea06c526b2e2d249db3448b51229e15/src/vs/workbench/contrib/chat/common/languageModels.ts#L286-L291) and has already been discussed in the API call

Tyriar (approved):

PR #259883, opened 2025-08-05
macOS - restore use of `hide` and `show` for `app.dock` second instances

bpasero:
cc @deepak1556 @rzhao271

### DIFF
--- a/src/vs/code/electron-main/main.ts
+++ b/src/vs/code/electron-main/main.ts
@@ -313,6 +313,11 @@ class CodeMain {
 				throw error;
 			}
 
+			// Since we are the second instance, we do not want to show the dock
+			if (isMacintosh) {
+				app.dock?.hide();
+			}
+
 			// there's a running instance, let's connect to it
 			let client: NodeIPCClient<string>;
 			try {
@@ -410,6 +415,11 @@ class CodeMain {
 			console.log(localize('statusWarning', "Warning: The --status argument can only be used if {0} is already running. Please run it again after {0} has started.", productService.nameShort));
 
 			throw new ExpectedError('Terminating...');
+		}
+
+		// dock might be hidden at this case due to a retry
+		if (isMacintosh) {
+			app.dock?.show();
 		}
 
 		// Set the VSCODE_PID variable here when we are sure we are the first


Tyriar (approved):

JBlitzar:
Hey, just a quick note: this implementation appears to be identical to the original version of my PR #259352 before changes were requested. I would appreciate attribution or at least a mention that the original work came from there.

bpasero:
@JBlitzar the change is a revert of https://github.com/microsoft/vscode/commit/f8f0c0ec5499cf5d210b12672d71784e93d42f0d, what am I missing?

JBlitzar:
Would be nice to be acknowledged for the work done in #259352 is all

bpasero:
@JBlitzar that PR was not merged. Again what am I missing? You can have credits for bringing up the idea of restoring the original code, but we typically do not call that out in our release notes.

JBlitzar:
Ok then. Just so we're on the same page, here's my current understanding of the situation:

I suggested something with open -a (v1)
I learned about the singleton instance argument passing etc, and  made a solution (v2)
Deepak suggested hiding first and showing if it's the only instance (v3)

You merge what essentially was v2 except independently made previously. (v0)

Since I didn't write v0, I shouldn't be credited. At the same time, I wrote v2 (which seems to be the same except for maybe line ordering and comment wording) independently, but that didn't get merged. I obviously brought up the conversation, but that isn't credited. 

So this all makes logical sense. Still, from my perspective it feels like I wrote a solution, revised on feedback, and then the same thing was self merged by a maintainer without credit. For me, the main incentive of contributing to OSS is crediting, so it felt kind of discouraging, even if technically fine

bpasero:
@JBlitzar no worries, I can add your original PR to our "Thank you" section of the upcoming release notes, which is normally automatically generated for each accepted PR, pretending as if your PR was accepted. How does that sound?

For the inconvenience, I enabled your GH handle to receive some swag from us, if you log in at: https://app.brilliantmade.com/receive/VS%20Code/microatckbfdy

Now I am looking forward to more PRs from you though 👍

JBlitzar:
Wow, amazing! Thanks so much, I really appreciate it.

PR #260131, opened 2025-08-06
engineering: use tar+zstd for win32 node_modules cache

tmm1:
this is significantly faster in my tests, down to 60s vs 120-180s before in the Extract node_modules phase

GHA and Azure Pipelines images have tar.exe and zstandard bundled since a few years ago

(will need to bump cachesalt if merging, to invalidate old cache)

cc @lszomoru @deepak1556


build/azure-pipelines/win32/product-build-win32.yml
```diff
@@ -73,7 +73,8 @@ steps:
       cacheHitVar: NODE_MODULES_RESTORED
     displayName: Restore node_modules cache
 
-  - powershell: 7z.exe x .build/node_modules_cache/cache.7z -aoa
+  - powershell: |
+      & "C:\Program Files\Git\usr\bin\tar.exe" -xf .build/node_modules_cache/cache.tzst -P -C . --force-local --use-compress-program "zstd -d"
```
deepak1556: Is this folder not on PATH ?

deepak1556 (commented):
Thanks! should also mirror this in `.github/workflows/pr-node-modules.yml` and `.github/workflows/pr-win32-test.yml` that is used for pr builds

build/azure-pipelines/win32/product-build-win32.yml
tmm1: It may work without the path, I didn't try. This invocation is copied from what github actions/cache does on Windows.

I know win32 ships their own tar.exe but I think its bsd not gnutar and the options may not match.

https://techcommunity.microsoft.com/blog/containers/tar-and-curl-come-to-windows/382409

tmm1:
> should also mirror this in `.github/workflows/pr-node-modules.yml` and `.github/workflows/pr-win32-test.yml` that is used for pr builds

Those paths already use tar but may see added speedups from zstd. For some reason whatever 7z does in the azure-pipeline is very slow.

deepak1556:
> Those paths already use tar but may see added speedups from zstd.

The windows jobs use 7z https://github.com/microsoft/vscode/blob/d5e090535a16ab04d9a92c4f42e12a1b04dbf721/.github/workflows/pr-win32-test.yml#L91 and https://github.com/microsoft/vscode/blob/d5e090535a16ab04d9a92c4f42e12a1b04dbf721/.github/workflows/pr-node-modules.yml#L293

deepak1556:
@joaomoreno actually wrote the archive cache steps, I couldn't find from commit history why 7z was preferred for windows, maybe it was one of the reasons you had pointed out but lets wait for his review before pushing this forward. He is oof this week :)

> I know win32 ships their own tar.exe but I think its bsd not gnutar and the options may not match.

PR #260138, opened 2025-08-06
Cover other cases for legacy terminal toolSpecificData

### LINKED ISSUE #258542: Chat fails to render old terminal commands
- Open an old chat session with a terminal command in it from history
- See this error about rendering the content

Remember that chat content is persisted across versions and we need to either be backwards compatible or migrate data forward.

Image

I'm actually curious what you think about this because it's not an easy problem and I've done it in a sort of ad-hoc way so far. I sort of have a pattern for this but it may not be fully fleshed out enough for you https://github.com/microsoft/vscode/blob/5b7867967324732ae16387f3d403deb822dfa10b/src/vs/workbench/contrib/chat/common/chatModel.ts#L1032

Probably what's missing is that `ISerializableChatRequestData` is not versioned like `ISerializableChatData` is. You can probably find an easy fix and maybe I can help build a better system for handling this going forward.

Tyriar:
Fixes #258542


roblourens (commented):
I see these usages for `terminalData.commandLine` 

Image

roblourens (dismissed):

justschen (approved):

PR #260178, opened 2025-08-06
Use ts-native to compile our build scripts

mjbvz:
There are a few minor differences in output, mostly around formatting/comments (see https://github.com/microsoft/typescript-go/issues/1526). Const enums are also not inlined (https://github.com/microsoft/typescript-go/issues/1216) so converting those over

### DIFF
--- a/build/azure-pipelines/common/checkForArtifact.js
+++ b/build/azure-pipelines/common/checkForArtifact.js
@@ -1,9 +1,9 @@
 "use strict";
+Object.defineProperty(exports, "__esModule", { value: true });
 /*---------------------------------------------------------------------------------------------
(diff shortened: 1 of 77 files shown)


DanielRosenwasser:
Chatted with @jakebailey - we're glad this found some emit differences, but it might be a good idea to hold off on this until we stabilize there.

aeschli (approved):

jakebailey:
Er, I guess it was merged anyway?

mjbvz:
Thanks for following up. I'm ok using tsgo for these scripts because we check in the js output too. That makes it easy to review any changes in emit. All the changes I noticed are pretty minor so I decided it was fine to go ahead before these issues are fixed

PR #260215, opened 2025-08-06
Add support for a help link when MCP assisted installation fails

joelverhagen:
This is in support of https://github.com/microsoft/vscode/issues/257679.

This will also support https://github.com/microsoft/vscode/issues/259901 if, for example, we move to using `npm` for the "does the package exist" check and npm is not available.

I am working in using the server.json contained in NuGet packages to improve MCP installation. To do these, I need to invoke `dotnet tool install` commands to download the package. If the user does not have `dotnet` installed, we should let them known. This is prior to the MCP server being installed so the existing notification flow will not kick in yet.

The URL and label are optionally sent back as part of the error. I believe this is a useful additional to the contract of `github.copilot.chat.mcp.setup.validatePackage`.

It looks like this:

image

Also, added `dotnet` to the list of recognized missing commands in case `dotnet tool exec` is used.
Also, fix a misnamed editor service dependency.

### DIFF
--- a/src/vs/workbench/contrib/mcp/browser/mcpCommandsAddConfiguration.ts
+++ b/src/vs/workbench/contrib/mcp/browser/mcpCommandsAddConfiguration.ts
@@ -20,6 +20,7 @@ import { IFileService } from '../../../../platform/files/common/files.js';
 import { ILabelService } from '../../../../platform/label/common/label.js';
 import { IMcpRemoteServerConfiguration, IMcpServerConfiguration, IMcpServerVariable, IMcpStdioServerConfiguration, McpServerType } from '../../../../platform/mcp/common/mcpPlatformTypes.js';
 import { INotificationService } from '../../../../platform/notification/common/notification.js';
+import { IOpenerService } from '../../../../platform/opener/common/opener.js';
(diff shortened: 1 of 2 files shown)


src/vs/workbench/contrib/mcp/browser/mcpCommandsAddConfiguration.ts
```diff
@@ -319,7 +321,23 @@ export class McpAddConfigurationCommand {
...
+						id: LoadAction.OpenUri,
+						label: result.helpUriLabel ?? localize('mcp.error.openHelpUri', 'Open help URL'),
+						helpUri: URI.parse(result.helpUri),
+					});
+				}
+
+				items.push(...[
+					{ id: LoadAction.Retry, label: localize('mcp.error.retry', 'Try a different package') },
+					{ id: LoadAction.Cancel, label: localize('cancel', 'Cancel') },
+				]);
```
connor4312:
```suggestion
				items.push(
					{ id: LoadAction.Retry, label: localize('mcp.error.retry', 'Try a different package') },
					{ id: LoadAction.Cancel, label: localize('cancel', 'Cancel') },
				);
```

connor4312 (approved):

Tyriar (approved):

Issue #259, opened 2015-11-19
Shortcut `use regular expression` mode in Find


so0k:
Coming from Sublime Text, trying to use Visual Studio Code.

How can I do the following: Find & select all instances of the `\w*` regular pattern for example..

In Sublime Text (default shortcuts):

1. <kbd>CTRL</kbd>+<kbd>F</kbd>
1. <kbd>ALT</kbd>+<kbd>R</kbd>
1. Enter: `\w*`
1. <kbd>ALT</kbd>+<kbd>ENTER</kbd>

In Visual Studio Code (default shortcuts):

1. <kbd>CTRL</kbd>+<kbd>F</kbd>
1. ????
1. Enter: `\w*`
1. <kbd>CTRL</kbd>+<kbd>SHIFT</kbd>+<kbd>L</kbd>
1. <kbd>ESC</kbd>

Tyriar:
A shortcut for match case would be good too.

alexandrudima:
@so0k I've pushed:
* `toggleFindCaseSensitive`: <kbd>ALT</kbd>+<kbd>C</kbd>, <kbd>CMD</kbd>+<kbd>ALT</kbd>+<kbd>C</kbd>
* `toggleFindWholeWord`: <kbd>ALT</kbd>+<kbd>W</kbd>, <kbd>CMD</kbd>+<kbd>ALT</kbd>+<kbd>W</kbd>
* `toggleFindRegex`: <kbd>ALT</kbd>+<kbd>R</kbd>, <kbd>CMD</kbd>+<kbd>ALT</kbd>+<kbd>R</kbd>

These will also work when the find widget is not visible.

Tyriar:
Can the keybindings be customized in keybindings.json?

so0k:
Thank you very much. Will test later & confirm

alexandrudima:
@Tyriar yes

PR #260290, opened 2025-08-07
fix: use readable name for SCM status bar entries on GitHub.dev (#259…

hihry:
Fixes #259649

### Problem
On GitHub.dev, the SCM status bar entries showed internal identifiers like `status.scm.0` instead of readable labels.

### Solution
Ensured the `name` field of each `IStatusbarEntry` is properly set to a readable string like `Source Control: Sync` or `Source Control: SwitchToBranch`, falling back to the command title if no repo-agnostic name is found.

This ensures GitHub.dev's status bar menu does not fall back to internal IDs as visible labels.


lszomoru:
Thanks for your contribution. Closing the PR in favour of https://github.com/microsoft/vscode/pull/263236

PR #260354, opened 2025-08-07
fix: prevent crash when abandoning variable input task

### LINKED ISSUE #255503: Unable to abandon task with ${input} variable type command

- VS Code Version:  1.101.0
- OS Version: Win 11

The handling of the result of an `${input}` variable type `command` in a `tasks.json` file has changed.

```json
  "inputs": [
    {
      "id": "pickOption",
      "type": "command",
      "command": "myextension.pickOption",
      "args": {
        "description": "Pick an option"
      }
    }
  ]
```

It is no longer possible to abandon the task by returning `undefined` from the command.
If I use an `${input}` variable type `pickString` in a `tasks.json` and I press `Esc` the task is abandoned.

```json
  "inputs": [
    {
      "id": "pickOptionNative",
      "type": "pickString",
      "description": "Pick an option (Native)",
      "options": ["Option A", "Option B", "Option C", "Option D"]
    }
  ]
```

Looking at the source code: [`baseConfigurationResolverService.ts`](https://github.com/microsoft/vscode/tree/main/src/vs/workbench/services/configurationResolver/browser)

The code for `pickString` returns `undefined` if the user abandons the quickPick, (line 322-329)

The code for `command` **always** returns: `return { value: result, input: info };`
This is correct if the return value of the command is a string, but not if it is `undefined` or `null`.

In commit 81775e7 (2025-03-11) it was:

```ts
						if (typeof result === 'string' || Types.isUndefinedOrNull(result)) {
							return result;
						}
```

In commit  [4c95d54](https://github.com/microsoft/vscode/commit/4c95d54dd5885d786952750c8f986cd80e9b3b4a)  (2025-03-18) it has changed to:

```ts
						if (typeof result === 'string' || Types.isUndefinedOrNull(result)) {
							return { value: result, input: info };
						}
```

Now an `undefined` result is treated as a valid result.
Better would this be to change it to:
```ts
						if (typeof result === 'string') {
							return { value: result, input: info };
						}
						if (Types.isUndefinedOrNull(result)) {
							return undefined;
						}
```

hihry:
fixes #255503


rioj7:
@hihry It looks like this will not fix #255503. There the problem is not related to the problem matcher. See the issue what lines will fix the problem

anthonykim1:
Closing as this doesnt seem to resolve https://github.com/microsoft/vscode/pull/256043

PR #260439, opened 2025-08-08
Convert extension webpack scripts to modules

mjbvz:
Should be nicer to write and work with, plus a good step towards converting them to ts


DonJayamanne (dismissed):

justschen (approved):

mjbvz:
Tested with a full build too just to confirm all extensions are still packaged correctly

PR #260499, opened 2025-08-08
Adding logs to determine focus state of NativeEditContext

aiday-mar:
Adding logs to determine focus state of NativeEditContext

### DIFF
--- a/src/vs/editor/browser/controller/editContext/native/nativeEditContext.ts
+++ b/src/vs/editor/browser/controller/editContext/native/nativeEditContext.ts
@@ -5,7 +5,7 @@
 
 import './nativeEditContext.css';
 import { isFirefox } from '../../../../../base/browser/browser.js';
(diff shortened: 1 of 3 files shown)


src/vs/editor/browser/controller/editContext/native/nativeEditContext.ts
```diff
@@ -310,6 +311,11 @@ export class NativeEditContext extends AbstractEditContext {
 		this._screenReaderSupport.onWillPaste();
 	}
 
+	public onWillCopy(): void {
+		this.logService.trace('NativeEditContext#onWillCopy');
+		this.logService.trace('NativeEditContext#isFocused : ', this.domNode.domNode === getActiveElement());
```
deepak1556, line 316: Do you also want to log the current focused element, would that help to know what was focused when the editcontext was not or is that information not so useful ?

deepak1556 (approved):

PR #260533, opened 2025-08-08
remove pause action from chat view

isidorn:
fixes https://github.com/microsoft/vscode/issues/252271


isidorn:
Tried it out of OSS, and seems to work

Screenshot 2025-08-08 at 11 58 07

connor4312:
We can also get rid of the `ChatContextKeys.isRequestPaused` context key and the downstream methods used here

```
			this.lastRequest.response.setPaused(pausedValue);
			this.chatAgentService.setRequestPaused(this.lastRequest.response.agent.id, this.lastRequest.id, pausedValue);
```

isidorn:
@connor4312 thank you for your review. I have addressed your feedback and updated the PR.
A lot of code got removed - which is always good. Let me know if this is too much - I was just unsure about the `ChatWorkingProgressContentPart` class - in the end I removed it.

src/vs/workbench/contrib/chat/browser/chatContentParts/chatProgressContentPart.ts
```diff
@@ -80,35 +78,6 @@ function shouldShowSpinner(followingContent: IChatRendererContent[], element: Ch
 	return isResponseVM(element) && !element.isComplete && followingContent.length === 0;
 }
 
-export class ChatWorkingProgressContentPart extends ChatProgressContentPart implements IChatContentPart {
```
connor4312, line 83: I think we still need this, we can just remove the `isPaused` check

isidorn: Thank you, makes sense. I brought it back

isidorn:
Updated the PR and brought back the class me and agent mode over-aggresively removed.
Let me know what you think.

src/vs/workbench/contrib/chat/browser/chatListRenderer.ts
```diff
@@ -1056,10 +1072,6 @@ export class ChatListItemRenderer extends Disposable implements ITreeRenderer<Ch
 			element.renderData = { lastRenderTime: Date.now(), renderedWordCount: newRenderedWordCount, renderedParts: partsToRender };
 		}
 
-		if (this.shouldShowWorkingProgress(element, partsToRender)) {
```
connor4312, line 1059: Was moving this check [up](https://github.com/microsoft/vscode/pull/260533/files/#diff-0fdbf88ff23fe938772a7a268e691003f663411f83be1a6b6c8506e68f7b858fR720) to earlier in the renderer intentional?

src/vs/workbench/contrib/chat/browser/chatListRenderer.ts
```diff
@@ -732,6 +731,23 @@ export class ChatListItemRenderer extends Disposable implements ITreeRenderer<Ch
 		this.updateItemHeightOnRender(element, templateData);
 	}
 
+	private shouldShowWorkingProgress(element: IChatResponseViewModel, partsToRender: IChatRendererContent[]): boolean {
```
connor4312, line 732: This is different than what `shouldShowWorkingProgress` did before, is that right?

isidorn: My mistake. Pushing fix for this so it is aligned with how it was before.

isidorn:
Thanks for review. Updated PR based on your comments.

connor4312 (approved):

justschen:
looks like this broke the working spinner - investigating to see if i see any obvious culprits

https://github.com/user-attachments/assets/fd7ed560-8085-46a1-93c3-e7f6f3339ea5

Issue #152324, opened 2022-06-16
Suggestion box has the wrong height in case of a small window size


vdavid:
Does this issue occur when all extensions are disabled?: Yes

- VS Code Version: 1.67.1
- OS Version: macOS 12.3.1

Steps to Reproduce:

1. Open a code file
2. Vertically resize the VS Code window to make it about 200px high
3. Start typing a term that's valid for the language of the open file, to make suggestions appear.
4. Observe how the suggestions box has the wrong height.

Or see this [2-min video](https://www.loom.com/share/e152bce924c746d993d66ec29089f58b) where I demo the behavior.

VSCodeTriageBot:
Thanks for creating this issue! It looks like you may be using an old version of VS Code, the latest stable release is 1.68.1. Please try upgrading to the latest version and checking whether this issue remains.

Happy Coding!

vdavid:
It's still an issue in version 1.68.1, I've confirmed it.

jrieken:
> Observe how the suggestions box has the wrong height

Not sure what that is supposed to mean. When a custom size doesn't fit a default is taken (instead of overriding your size) until more space is available

vdavid:
@jrieken See the video above. There is _clearly_ a huge amount of space available:

<img width="590" alt="CleanShot 2022-12-05 at 12 16 11@2x" src="https://user-images.githubusercontent.com/2552265/205624332-fb014c18-0822-4d8e-b489-e78f174c9c47.png">

The bug is that the component checks for the space available **above** the field rather than **below** it. (And, yeah, there is indeed not much space available above.)

VSCodeTriageBot:
Hey @jrieken, this issue might need further attention.

@vdavid, you can help us out by closing this issue if the problem no longer exists, or adding more information.

vdavid:
@VSCodeTriageBot @jrieken What further info do you need? I attached a screenshot, a video, and steps to reproduce. The bug is quite obvious to me, and I can easily reproduce it.
I understand if it's not a priority for your team, that's your job to decide, but confirming the issue's existence is pretty straightforward to me.

andr8928:
This is having a big impact on my use case for the monaco editor. To help visualize the issue:

Normal space:

<img width="742" height="563" alt="Image" src="https://github.com/user-attachments/assets/e0585925-c3d9-4044-89da-ac7688b731cf" />    

    
BUG (less space):   
[...]

PR #260583, opened 2025-08-08
Suggest widget: Bug fix - when widget is too tall, ensure larger of above and below space is used.

andr8928:
**Issue addressed:**

https://github.com/microsoft/vscode/issues/152324

**Description:**
Currently, when the suggest widget has enough space below, everything is fine:
monaco-test-suggest-OK

    

When the suggest widget is larger than the area below the current line, the current code always calculates the height and position of the widget using the space **above** the line (even if the space below is larger):

With this change, if the suggest box is larger than the space below, it will now choose the largest of "above" and "below" to use in its calculations (unless force-above is enabled):


andr8928:
@mjbvz Is there additional people that can be assigned to this?
It seems @jrieken is away and hasn't been active for a couple of months. This bug is causing problems for my team and product.

jrieken (approved):
lgtm

lszomoru (approved):

PR #260683, opened 2025-08-08
check the shadow root

### LINKED ISSUE #16660: TextInput and TextAreaInput from bokeh triggers notebook shortcuts while writing text
### Applies To

- [x] Notebooks (.ipynb files)

### What happened?

When using either TextInput or TextAreaInput from bokeh the active cell is sensitive to the written letters and this causes that letters and strings like 'a', 'b' or 'dd' will trigger the shortcut. Turns out that the notebook adds and removes cells, which is confusing.

Image

Using the same code in jupyter lab in a browser works fine, so I think this behavior is is caused on the VSCode side. I can't reproduce this with ipywidgets.

```python
from bokeh.models import TextInput
from bokeh.io import output_notebook, show
output_notebook()

ti = TextInput()
show(ti)
```

### VS Code Version

1.100.2

### Jupyter Extension Version

2025.4.0

### Jupyter logs

```shell
None
```

Python 3.12.4

_No response_

_No response_

Local

amunger:
fix https://github.com/microsoft/vscode-jupyter/issues/16660

### DIFF
--- a/src/vs/workbench/contrib/notebook/browser/view/renderers/webviewPreloads.ts
+++ b/src/vs/workbench/contrib/notebook/browser/view/renderers/webviewPreloads.ts
@@ -188,9 +188,15 @@ async function webviewPreloads(ctx: PreloadContext) {
 		}, 0);
 	};
 
-	const isEditableElement = (element: Element) => {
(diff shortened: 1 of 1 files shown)


anthonykim1 (approved):

krassowski (commented):
Tested locally, works well!

PR #260770, opened 2025-08-09
Update confirmations to use new proposed terminal style

### LINKED ISSUE #259088: Run in terminal: Confirmation edited command disappears when you scroll it out of view and back in
Repro:

1. Ask `run ls in terminal`
2. Click into the input and edit it
3. Scroll up so it's out of view
4. Scroll down, 🐛 it's the original string

Image

### LINKED ISSUE #260705: Disappearing terminal confirmation editors
- get a terminal tool confirmation
- click into it
- It disappears

Image

This is weird- what happened is that the editor's model was disposed, and the editor was removed from the DOM. That happened when the wordHighlighter contrib disposed its reference and there were no remaining refs. I don't know what changed, and I can't repro reliably.

I'm a little unsure about the proper flow, I think we are using `createModel` wrong somehow. https://github.com/microsoft/vscode/blob/7ca850c73f7fa37e1fc80090f14268b0b6b504bb/src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolSubPart.ts#L118-L123

In some places like https://github.com/microsoft/vscode/blob/7ca850c73f7fa37e1fc80090f14268b0b6b504bb/src/vs/workbench/contrib/bulkEdit/browser/preview/bulkEditPreview.ts#L443-L448 I see them call `createModelReference` after `createModel`, maybe we are supposed to do that.

@Tyriar you've touched this code recently, any idea? I haven't tried to bisect yet.

Tyriar:
Part of #257468
Fixes #260705
Fixes #260776
Fixes #259088

This adds `ChatCustomConfirmationWidget2` that aims to be this new style for all tools.

I did not test the following:

- `InstallExtensionsTool`/`ExtensionsInstallConfirmationWidgetSubPart` as it's not hooked up. Adoption for others didn't have any problems though.
- `kind: 'confirmation` types as defined in [here](https://github.com/microsoft/vscode/blob/b4b07a0f221229f1ba58dc491a6e5ab965027445/src/vs/workbench/contrib/chat/common/chatService.ts#L250-L257), not sure where they show up

Initial state:

image

Dropdown:

image

Delayed hover on command:

image

---

Terminal output polling:

image

Extension install ([from copilot ext](https://github.com/microsoft/vscode-copilot-chat/blob/13157e08b3bb517f622704786b1baa714b57711c/src/extension/tools/node/installExtensionTool.tsx#L94)):

image

Task run:

image

Test run:

image

MCP (disclaimer wrapping is existing issue https://github.com/microsoft/vscode/issues/260806):

image


src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolConfirmationSubPart.ts
```diff
@@ -114,19 +119,26 @@ export class TerminalConfirmationWidgetSubPart extends BaseChatToolInvocationSub
...
-		const langId = this.languageService.getLanguageIdByLanguageName(terminalData.language ?? 'sh') ?? 'shellscript';
+		const languageId = this.languageService.getLanguageIdByLanguageName(terminalData.language ?? 'sh') ?? 'shellscript';
 		const model = this.modelService.createModel(
 			terminalData.commandLine.toolEdited ?? terminalData.commandLine.original,
-			this.languageService.createById(langId),
+			this.languageService.createById(languageId),
 			this._getUniqueCodeBlockUri(),
 			true
 		);
+		textModelService.createModelReference(model.uri).then(ref => {
```
roblourens: See my helper `thenifNotDisposed`

src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolConfirmationSubPart.ts
```diff
@@ -114,19 +119,26 @@ export class TerminalConfirmationWidgetSubPart extends BaseChatToolInvocationSub
 				ariaLabel: typeof title === 'string' ? title : title.value
 			}
 		};
-		const langId = this.languageService.getLanguageIdByLanguageName(terminalData.language ?? 'sh') ?? 'shellscript';
+		const languageId = this.languageService.getLanguageIdByLanguageName(terminalData.language ?? 'sh') ?? 'shellscript';
 		const model = this.modelService.createModel(
```
roblourens: Do we need to `this._register` the model?

src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolConfirmationSubPart.ts
roblourens: The fact that we have to do this is confusing.

Tyriar:
Todo:

- fix tests
- make all bottom corners rounded
- Action comments

connor4312:
THe last screenshot has odd word wrap 

image

Tyriar:
@connor4312 mentioned that above, actually not a regression https://github.com/microsoft/vscode/issues/260806

src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolConfirmationSubPart.ts
Tyriar: I don't think I can use that since I need to call `dispose` on it to free the reference when the chat part is disposed?

connor4312:
Wrapping in this header is a bit ugly in smaller widths

image

Maybe [below a certain width](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_containment/Container_queries), the text keeps 100% width and the buttons stay right-aligned but get pushed down below the title text line.

connor4312:
1. For elicitation, the response is not added to the response, but instead the query is printed again, and the "respond" button no longer disappears / changes to a success state. Hitting "Cancel" has the same effect

image

2. If there's a tool call right after an elicitation, the elicitation is missing some bottom margin

image

lgtm otherwise

src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolConfirmationSubPart.ts
roblourens: Good point. There could probably be a helper that covers what you want to do

Tyriar:
> Wrapping in this header is a bit ugly in smaller widths

@jo-oikawa is rethinking this to maybe show the buttons on the bottom. This can come through in a follow up PR though, don't want merge conflicts to keep stacking up.

> For elicitation, the response is not added to the response, but instead the query is printed again, and the "respond" button no longer disappears / changes to a success state. Hitting "Cancel" has the same effect

Can you let me know how to reproduce this? I don't know what it looked like before.

> If there's a tool call right after an elicitation, the elicitation is missing some bottom margin

Will fix.

src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolConfirmationSubPart.ts
Tyriar: Added helper, created https://github.com/microsoft/vscode/issues/261211 to adopt after this PR is merged.

connor4312 (approved):

mjbvz (approved):

PR #260910, opened 2025-08-10
fix overlay when editing previous requests

### LINKED ISSUE #256002: Chat: UI controls are still operational even though they look disabled
1. Send a message to Copilot
2. Edit the message, entire view below goes grey

🐛 I can still open the mode/model pickers or open the tools quick pick, etc...

https://github.com/user-attachments/assets/6df9df0a-f055-4a07-a1d8-fe4f90719f5f

justschen:
fix https://github.com/microsoft/vscode/issues/256002

### DIFF
--- a/src/vs/workbench/contrib/chat/browser/chatInputPart.ts
+++ b/src/vs/workbench/contrib/chat/browser/chatInputPart.ts
@@ -220,7 +220,7 @@ export class ChatInputPart extends Disposable implements IHistoryNavigationWidge
 	private attachmentsContainer!: HTMLElement;
 
 	private chatInputOverlay!: HTMLElement;
-	private _overlayClickListener: IDisposable | undefined;
+	private readonly _overlayClickListener = this._register(new MutableDisposable<IDisposable>());
 
 	private attachedContextContainer!: HTMLElement;
 	private readonly attachedContextDisposables: MutableDisposable<DisposableStore>;
@@ -1308,15 +1308,13 @@ export class ChatInputPart extends Disposable implements IHistoryNavigationWidge
 	public toggleChatInputOverlay(editing: boolean): void {
 		this.chatInputOverlay.classList.toggle('disabled', editing);
 		if (editing) {
-			this._overlayClickListener?.dispose();
-			this._overlayClickListener = this._register(dom.addStandardDisposableListener(this.chatInputOverlay, dom.EventType.CLICK, e => {
+			this._overlayClickListener.value = dom.addStandardDisposableListener(this.chatInputOverlay, dom.EventType.CLICK, e => {
 				e.preventDefault();
 				e.stopPropagation();
 				this._onDidClickOverlay.fire();
-			}));
+			});
 		} else {
-			this._overlayClickListener?.dispose();
-			this._overlayClickListener = undefined;
+			this._overlayClickListener.clear();
 		}
 	}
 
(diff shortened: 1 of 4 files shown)


src/vs/workbench/contrib/chat/browser/chatInputPart.ts
```diff
@@ -217,6 +220,7 @@ export class ChatInputPart extends Disposable implements IHistoryNavigationWidge
 	private attachmentsContainer!: HTMLElement;
 
 	private chatInputOverlay!: HTMLElement;
+	private _overlayClickListener: IDisposable | undefined;
```
Tyriar: You're better off using a Mutable Disposable here, register that on class init and you don't need to worry about disposing it when recreated or when the class is disposed.

src/vs/workbench/contrib/chat/browser/chatInputPart.ts
```diff
@@ -1303,6 +1307,17 @@ export class ChatInputPart extends Disposable implements IHistoryNavigationWidge
 
 	public toggleChatInputOverlay(editing: boolean): void {
 		this.chatInputOverlay.classList.toggle('disabled', editing);
+		if (editing) {
+			this._overlayClickListener?.dispose();
+			this._overlayClickListener = this._register(dom.addStandardDisposableListener(this.chatInputOverlay, dom.EventType.CLICK, e => {
```
Tyriar: You're registering it to the class and tracking the listener. This will have the class hold on to it after the listener is disposed.

src/vs/workbench/contrib/chat/browser/media/chat.css
```diff
@@ -2342,11 +2342,11 @@ have to be updated for changes to the rules above, or to support more deeply nes
 	height: 100%;
 	background-color: var(--vscode-sideBar-background);
 	opacity: 0.6;
-	pointer-events: none;
 	display: flex;
-	cursor: default;
 	z-index: 101;
 	user-select: none;
+	pointer-events: auto;
```
Tyriar: Can these rules be removed?

Tyriar (changes requested):

src/vs/workbench/contrib/chat/browser/chatInputPart.ts
```diff
@@ -158,6 +158,9 @@ export class ChatInputPart extends Disposable implements IHistoryNavigationWidge
 	private _onDidAcceptFollowup: Emitter<{ followup: IChatFollowup; response: IChatResponseViewModel | undefined }>;
 	readonly onDidAcceptFollowup: Event<{ followup: IChatFollowup; response: IChatResponseViewModel | undefined }>;
 
+	private _onDidClickOverlay: Emitter<void> = new Emitter<void>();
```
Tyriar: Missing register

src/vs/workbench/contrib/chat/browser/chatInputPart.ts
```diff
@@ -1303,6 +1308,17 @@ export class ChatInputPart extends Disposable implements IHistoryNavigationWidge
 
 	public toggleChatInputOverlay(editing: boolean): void {
 		this.chatInputOverlay.classList.toggle('disabled', editing);
+		if (editing) {
+			const store = new DisposableStore();
```
Tyriar: Can make it a MutableDisposable<IDisposable> to avoid creating the store

Tyriar (changes requested):

src/vs/workbench/contrib/chat/browser/chatInputPart.ts
justschen: i had agent mode set this up and it's interesting that there are a bunch of patterns of emitters like this in `chatInputPart`, but it chose to do it in a different way still :/

DonJayamanne (approved):

Tyriar (approved):

PR #261013, opened 2025-08-11
add remote debugging port to command line switches

hediet:


deepak1556 (commented):
Can you add it to https://github.com/microsoft/vscode/blob/dd7867e5f1e6d80ae0ff08032c1930de7b71e107/src/main.ts#L356 and https://github.com/microsoft/vscode/blob/dd7867e5f1e6d80ae0ff08032c1930de7b71e107/src/vs/workbench/electron-browser/desktop.contribution.ts#L396

hediet:
@copilot can you look into deepak1556's suggestions?

deepak1556 (approved):

Issue #260703, opened 2025-08-09
Address user issues collected from socials


sbatten:
We have collected a good amount of user feedback in our socials and channels and need to align these with work items existing or new. This is a meta issue tracking work to both read through our collected feedback in channels to prioritize them and highlight our action items.

Items will be added to this list as we review the feedback:
- [x] Message Editing: Allow editing of chat messages after sending @justschen 
- [x] Agent: "Restore to Checkpoint"/Robust Undo: Implement reliable undo/rollback for agent changes, including reverting file additions/deletions @justschen 
- [ ] Support for Thinking Tokens [vscode#257104](https://github.com/microsoft/vscode/issues/257104) @justschen 
- [x] Make the changed files UX more lightweight [vscode#260175](https://github.com/microsoft/vscode/issues/260175) @justschen 
- [ ] Export Chat History [vscode#256366](https://github.com/microsoft/vscode/issues/256366) @eli-w-king @hawkticehurst 
- [x] Show History in Empty state [vscode#251187](https://github.com/microsoft/vscode/issues/251187) @eli-w-king 
- [ ] Smooth Undo Pop-ups: Make undo actions unobtrusive and avoid jarring pop-ups. [https://github.com/microsoft/vscode/issues/270250] @hawkticehurst

bpasero:
👏 that we plan this!

I am not sure if team feedback also counts, but there is probably overlap with issues we had already collected internally:

Here is a project board (now a bit outdated) that Isi and me compiled over the last months: https://github.com/orgs/microsoft/projects/1716/views/1

And then, I have recently filed many papercut issues using the newer `workbench-copilot` label: https://github.com/microsoft/vscode/issues?q=is%3Aissue%20state%3Aopen%20label%3Aworkbench-copilot

//cc @isidorn

justschen:
cc @bpasero @isidorn ref https://github.com/microsoft/vscode/issues/260175 one thing we chatted about and that i'd like to play around with as well is when we "keep" files, we remove them from the working set entirely. 

i know @connor4312 has some work with removing "keep", but not sure where that is atm. we can coordinate on that tho!

bpasero:
I guess this would allow us to drop the bold vs. non-bold semantic, because then a file would disappear when kept and then re-appear when new changes are made? Sounds interesting, would have to play with it.

isidorn:
[...]

PR #261085, opened 2025-08-11
remove item from working set when kept

justschen:
ref https://github.com/microsoft/vscode/issues/260703#issuecomment-3172476909


Tyriar (approved):
@justschen make sure this gets covered during testing

justschen:
yep! the plan is to have all the working set UX under one TPI + any other UX work under different ones

Issue #260478, opened 2025-08-08
Explorer View no longer syncs to active source file in editor


matusthemostlygreat:
Type: <b>Bug</b>

Whenever I switched tabs in the editor, the current file would also be highlighted in the Explorer pane. Now it is not, and event manually trying to use Reveal Active File In Explorer fails to work. This is a recent usability regression.

VS Code version: Code 1.103.0 (e3550cfac4b63ca4eafca7b601f0d2885817fd1f, 2025-08-06T21:40:10.271Z)
OS version: Windows_NT x64 10.0.26100
Modes:
Remote OS version: Linux x64 6.6.87.2-microsoft-standard-WSL2

<details>
<summary>System Info</summary>

|Item|Value|
|---|---|
|CPUs|13th Gen Intel(R) Core(TM) i7-13800H (20 x 2918)|
|GPU Status|2d_canvas: enabled<br>direct_rendering_display_compositor: disabled_off_ok<br>gpu_compositing: enabled<br>multiple_raster_threads: enabled_on<br>opengl: enabled_on<br>rasterization: enabled<br>raw_draw: disabled_off_ok<br>skia_graphite: disabled_off<br>trees_in_viz: disabled_off<br>video_decode: enabled<br>video_encode: enabled<br>vulkan: disabled_off<br>webgl: enabled<br>webgl2: enabled<br>webgpu: enabled<br>webnn: disabled_off|
|Load (avg)|undefined|
|Memory (System)|31.83GB (5.32GB free)|
|Process Argv|--folder-uri=vscode-remote://wsl+Ubuntu-20.04/home/matusl/AVP --remote=wsl+Ubuntu-20.04 --crash-reporter-id 0837dbd2-c6a3-4642-bdde-fa58b4825238|
|Screen Reader|no|
|VM|0%|

|Item|Value|
|---|---|
|Remote|WSL: Ubuntu-20.04|
|OS|Linux x64 6.6.87.2-microsoft-standard-WSL2|
|CPUs|13th Gen Intel(R) Core(TM) i7-13800H (20 x 0)|
|Memory (System)|23.47GB (19.41GB free)|
|VM|0%|
</details><details><summary>Extensions (23)</summary>

Extension|Author (truncated)|Version
---|---|---
togglehs|bbe|0.1.2
githistory|don|0.6.20
gitlens|eam|17.3.3
filter-lines|ear|1.1.0
copilot|Git|1.350.0
copilot-chat|Git|0.30.0
[...]

PR #261091, opened 2025-08-11
Fix for setInput canceling all refresh promises

osortega:
Fix for https://github.com/microsoft/vscode/issues/260478

### DIFF
--- a/src/vs/base/browser/ui/tree/asyncDataTree.ts
+++ b/src/vs/base/browser/ui/tree/asyncDataTree.ts
@@ -769,7 +769,7 @@ export class AsyncDataTree<TInput, T, TFilterData = void> implements IDisposable
 	}
 
 	async setInput(input: TInput, viewState?: IAsyncDataTreeViewState): Promise<void> {
-		this.cancelAllRefreshPromises()
+		this.cancelAllRefreshPromises();
 
 		this.root.element = input!;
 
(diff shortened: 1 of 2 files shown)


src/vs/base/browser/ui/tree/asyncDataTree.ts
```diff
@@ -769,7 +769,8 @@ export class AsyncDataTree<TInput, T, TFilterData = void> implements IDisposable
 	}
 
 	async setInput(input: TInput, viewState?: IAsyncDataTreeViewState): Promise<void> {
-		this.cancelAllRefreshPromises();
+		this.refreshPromises.forEach(promise => promise.cancel());
+		this.refreshPromises.clear();
```
Tyriar: Can you pull this out of `cancelAllRefreshPromises` and have a private helper that both places call into?

Tyriar (changes requested):

src/vs/base/browser/ui/tree/asyncDataTree.ts
osortega: Changed this so we call this with an arg to decide whether to cancel subtrees or not

Tyriar (approved):

src/vs/base/browser/ui/tree/asyncDataTree.ts
eungintha: e

PR #261094, opened 2025-08-11
Add Python activate hooks into shell integration scripts 

anthonykim1:
Resolves: https://github.com/microsoft/vscode/issues/259637 

Important things to note: 
* Update all four shell integration scripts (pwsh, zsh, bash, fish) ---> Done
* Sanitize in process.ts and terminalSuggest.main.ts ---> Done
* If activate returns non-zero (failed), we should prevent crash, catch exit code, and print to terminal 
   ---> Findings: Try catch was not available for zsh, bash so using negate to prevent crash. Fish doesn't crash on eval 
   ---> Try catch was available in pwsh
   ---> Done
* Block other extension from modifying Python specific activate (Important -> Should do this as activate var is leading to code execution)  ---> Done
* Have `*` icon similar to terminal restored message. ---> Done
* Sanitize and remove hard-coded vsc python env var in terminalEnv ---> Done
* Clean-ups ---> Done?

### DIFF
--- a/extensions/terminal-suggest/src/terminalSuggestMain.ts
+++ b/extensions/terminal-suggest/src/terminalSuggestMain.ts
@@ -521,6 +521,7 @@ export function sanitizeProcessEnvironment(env: Record<string, string>, ...prese
 	const keysToRemove = [
 		/^ELECTRON_.$/,
 		/^VSCODE_(?!(PORTABLE|SHELL_LOGIN|ENV_REPLACE|ENV_APPEND|ENV_PREPEND)).$/,
+		/^VSCODE_(PWSH|ZSH|BASH)_ACTIVATE$/,
 		/^SNAP(|_.*)$/,
 		/^GDK_PIXBUF_.$/,
 	];
--- a/extensions/terminal-suggest/src/terminalSuggestMain.ts
+++ b/extensions/terminal-suggest/src/terminalSuggestMain.ts
@@ -521,7 +521,7 @@ export function sanitizeProcessEnvironment(env: Record<string, string>, ...prese
 	const keysToRemove = [
 		/^ELECTRON_.$/,
 		/^VSCODE_(?!(PORTABLE|SHELL_LOGIN|ENV_REPLACE|ENV_APPEND|ENV_PREPEND)).$/,
-		/^VSCODE_(PWSH|ZSH|BASH)_ACTIVATE$/,
+		/^VSCODE_PYTHON_(PWSH|ZSH|BASH)_ACTIVATE/,
 		/^SNAP(|_.*)$/,
 		/^GDK_PIXBUF_.$/,
 	];
--- a/extensions/terminal-suggest/src/terminalSuggestMain.ts
+++ b/extensions/terminal-suggest/src/terminalSuggestMain.ts
@@ -520,8 +520,7 @@ export function sanitizeProcessEnvironment(env: Record<string, string>, ...prese
 	}, {});
 	const keysToRemove = [
 		/^ELECTRON_.$/,
-		/^VSCODE_(?!(PORTABLE|SHELL_LOGIN|ENV_REPLACE|ENV_APPEND|ENV_PREPEND)).$/,
-		/^VSCODE_PYTHON_(PWSH|ZSH|BASH|FISH)_ACTIVATE/,
+		/^VSCODE_(?!(PORTABLE|SHELL_LOGIN|ENV_REPLACE|ENV_APPEND|ENV_PREPEND)).+$/,
 		/^SNAP(|_.*)$/,
 		/^GDK_PIXBUF_.$/,
 	];
--- a/extensions/terminal-suggest/src/terminalSuggestMain.ts
+++ b/extensions/terminal-suggest/src/terminalSuggestMain.ts
@@ -521,7 +521,7 @@ export function sanitizeProcessEnvironment(env: Record<string, string>, ...prese
 	const keysToRemove = [
 		/^ELECTRON_.$/,
 		/^VSCODE_(?!(PORTABLE|SHELL_LOGIN|ENV_REPLACE|ENV_APPEND|ENV_PREPEND)).$/,
-		/^VSCODE_PYTHON_(PWSH|ZSH|BASH)_ACTIVATE/,
+		/^VSCODE_PYTHON_(PWSH|ZSH|BASH|FISH)_ACTIVATE/,
 		/^SNAP(|_.*)$/,
 		/^GDK_PIXBUF_.$/,
 	];
--- a/src/vs/base/common/processes.ts
+++ b/src/vs/base/common/processes.ts
@@ -109,7 +109,7 @@ export function sanitizeProcessEnvironment(env: IProcessEnvironment, ...preserve
 	const keysToRemove = [
 		/^ELECTRON_.+$/,
 		/^VSCODE_(?!(PORTABLE|SHELL_LOGIN|ENV_REPLACE|ENV_APPEND|ENV_PREPEND)).+$/,
-		/^VSCODE_PYTHON_(PWSH|ZSH|BASH)_ACTIVATE/,
+		/^VSCODE_PYTHON_(PWSH|ZSH|BASH|FISH)_ACTIVATE/,
 		/^SNAP(|_.*)$/,
 		/^GDK_PIXBUF_.+$/,
 	];
--- a/src/vs/base/common/processes.ts
+++ b/src/vs/base/common/processes.ts
@@ -109,7 +109,6 @@ export function sanitizeProcessEnvironment(env: IProcessEnvironment, ...preserve
 	const keysToRemove = [
 		/^ELECTRON_.+$/,
 		/^VSCODE_(?!(PORTABLE|SHELL_LOGIN|ENV_REPLACE|ENV_APPEND|ENV_PREPEND)).+$/,
-		/^VSCODE_PYTHON_(PWSH|ZSH|BASH|FISH)_ACTIVATE/,
 		/^SNAP(|_.*)$/,
 		/^GDK_PIXBUF_.+$/,
 	];
--- a/src/vs/base/common/processes.ts
+++ b/src/vs/base/common/processes.ts
@@ -109,7 +109,7 @@ export function sanitizeProcessEnvironment(env: IProcessEnvironment, ...preserve
 	const keysToRemove = [
 		/^ELECTRON_.+$/,
 		/^VSCODE_(?!(PORTABLE|SHELL_LOGIN|ENV_REPLACE|ENV_APPEND|ENV_PREPEND)).+$/,
-		/^VSCODE_(PWSH|ZSH|BASH)_ACTIVATE$/,
+		/^VSCODE_PYTHON_(PWSH|ZSH|BASH)_ACTIVATE/,
 		/^SNAP(|_.*)$/,
 		/^GDK_PIXBUF_.+$/,
 	];
--- a/src/vs/base/common/processes.ts
+++ b/src/vs/base/common/processes.ts
@@ -109,6 +109,7 @@ export function sanitizeProcessEnvironment(env: IProcessEnvironment, ...preserve
 	const keysToRemove = [
 		/^ELECTRON_.+$/,
 		/^VSCODE_(?!(PORTABLE|SHELL_LOGIN|ENV_REPLACE|ENV_APPEND|ENV_PREPEND)).+$/,
+		/^VSCODE_(PWSH|ZSH|BASH)_ACTIVATE$/,
 		/^SNAP(|_.*)$/,
 		/^GDK_PIXBUF_.+$/,
 	];
--- a/src/vs/base/test/common/processes.test.ts
+++ b/src/vs/base/test/common/processes.test.ts
@@ -42,7 +42,6 @@ suite('Processes', () => {
 		assert.strictEqual(env['VSCODE_PYTHON_ZSH_ACTIVATE'], undefined);
 		assert.strictEqual(env['VSCODE_PYTHON_PWSH_ACTIVATE'], undefined);
 		assert.strictEqual(env['VSCODE_PYTHON_FISH_ACTIVATE'], undefined);
-
 		assert.strictEqual(Object.keys(env).length, 3);
 	});
 });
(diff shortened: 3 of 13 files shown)


src/vs/platform/terminal/common/environmentVariableCollection.ts
```diff
@@ -13,13 +13,17 @@ const mutatorTypeToLabelMap: Map<EnvironmentVariableMutatorType, string> = new M
 	[EnvironmentVariableMutatorType.Prepend, 'PREPEND'],
 	[EnvironmentVariableMutatorType.Replace, 'REPLACE']
 ]);
+// TODO: key was --> 'VSCODE_BASH_ACTIVATE:::file:///Users/anthonykim/Desktop/Skeleton'
+const PYTHON_ACTIVATION_VARS_PATTERN = /^VSCODE_(PWSH|ZSH|BASH)_ACTIVATE/;
```
anthonykim1: I didn't include the trailing $, because key (`VSCODE_BASH_ACTIVATE`) was attached with `:::<cwd>`

Although variable name was just `VSCODE_BASH_ACTIVATE` by itself NOT with the `:::<cwd>`

Screenshot 2025-08-11 at 3 46 09 PM


`'VSCODE_BASH_ACTIVATE:::file:///Users/anthonykim/Desktop/Skeleton'` instead of `VSCODE_BASH_ACTIVATE`.


But in `applyAtProcessCreation` you would just see `VSCODE_BASH_ACTIVATE` by itself: 
Screenshot 2025-08-11 at 3 53 34 PM


Other variables dont seem to do this. 
Edit: I'm guessing `:::` might be just internal for map in constructor??

src/vs/base/common/processes.ts
```diff
@@ -109,6 +109,7 @@ export function sanitizeProcessEnvironment(env: IProcessEnvironment, ...preserve
 	const keysToRemove = [
 		/^ELECTRON_.+$/,
 		/^VSCODE_(?!(PORTABLE|SHELL_LOGIN|ENV_REPLACE|ENV_APPEND|ENV_PREPEND)).+$/,
+		/^VSCODE_(PWSH|ZSH|BASH)_ACTIVATE$/,
```
anthonykim1: Just tested/debugged this, and I dont think this does anything. 
The envs from the parameter didn't even contain these VSCODE_ACTIVATE var. 

I do see them show up later in `src/vs/platform/terminal/common/environmentVariableCollection.ts` though. So maybe this is expected behavior?

src/vs/workbench/contrib/terminal/common/scripts/shellIntegration-bash.sh
```diff
@@ -91,6 +91,16 @@ if [ -n "${VSCODE_ENV_APPEND:-}" ]; then
 	builtin unset VSCODE_ENV_APPEND
 fi
 
+# Register Python shell activate hooks
+if [ -n "$VSCODE_BASH_ACTIVATE" ] && [ "$TERM_PROGRAM" = "vscode" ]; then
+	# Prevent crashing by negating exit code
+	if ! builtin eval "$VSCODE_BASH_ACTIVATE"; then
+		__vsc_activation_status=$?
+		# TODO: Figure out how to add 'star' icon similar to terminal restored.
```
anthonykim1: I was hoping I can do the star icon next to the message like the `History restored` message when restoring terminal.

Was thinking something like https://github.com/microsoft/vscode/blob/f4c18a6849f25fa97b6ff7e26f9105152a2ac9e3/src/vs/workbench/contrib/terminal/browser/terminalInstance.ts#L1689 but not sure how to get this on the shell script side🧐

extensions/terminal-suggest/src/terminalSuggestMain.ts
```diff
@@ -521,6 +521,7 @@ export function sanitizeProcessEnvironment(env: Record<string, string>, ...prese
 	const keysToRemove = [
 		/^ELECTRON_.$/,
 		/^VSCODE_(?!(PORTABLE|SHELL_LOGIN|ENV_REPLACE|ENV_APPEND|ENV_PREPEND)).$/,
+		/^VSCODE_(PWSH|ZSH|BASH)_ACTIVATE$/,
```
anthonykim1: nit: `/^VSCODE_(PWSH|ZSH|BASH)_ACTIVATE$/`

Subjective) Would it be better to attach PYTHON after the word VSCODE? 
Would have to make sure to update both here and in env extension though.

src/vs/platform/terminal/common/environmentVariableCollection.ts
```diff
@@ -28,6 +33,15 @@ export class MergedEnvironmentVariableCollection implements IMergedEnvironmentVa
 			while (!next.done) {
 				const mutator = next.value[1];
 				const key = next.value[0];
+
+				// Only Python env extension can modify Python activate env var.
+				if (PYTHON_ACTIVATION_VARS_PATTERN.test(key) &&
+					PYTHON_ENV_EXTENSION_ID !== extensionIdentifier) {
+					this._logService.warn(`Extension '${extensionIdentifier}' attempted to modify Python activation variable '${key}' but was blocked`);
```
anthonykim1: ended up touching so many files, to have logs here to denote in case other extension where trying to modify Python activate vars.

Honestly not sure if this is worth all the changes.

anthonykim1:
After this PR, will make an additional PR in env extension to make sure they have fallback to still activate in case env extension can't locate hooks from vscode shell integration scripts.

src/vs/base/common/processes.ts
Tyriar: This function is called by a bunch of things, just adding the regex like this should be sufficient.

extensions/terminal-suggest/src/terminalSuggestMain.ts
Tyriar: Yes I'd call is `VSCODE_PYTHON_...`

src/vs/platform/terminal/common/environmentVariableCollection.ts
Tyriar: I don't remember that but I guess that's how it works. Kartik may have added that part 🤔

src/vs/platform/terminal/common/environmentVariableCollection.ts
```diff
@@ -70,6 +84,13 @@ export class MergedEnvironmentVariableCollection implements IMergedEnvironmentVa
 			const actualVariable = isWindows ? lowerToActualVariableNames![variable.toLowerCase()] || variable : variable;
 			for (const mutator of mutators) {
 				const value = variableResolver ? await variableResolver(mutator.value) : mutator.value;
+
+				// Only Python env extension can modify Python activate env var.
+				if (PYTHON_ACTIVATION_VARS_PATTERN.test(mutator.variable) && PYTHON_ENV_EXTENSION_ID !== mutator.extensionIdentifier) {
+					this._logService.warn(`Extension '${mutator.extensionIdentifier}' attempted to modify Python activation variable '${mutator.variable}' but was blocked`);
+					continue;
```
Tyriar: I see this twice so let's pull into a helper function

src/vs/workbench/contrib/terminal/common/environmentVariableService.ts
```diff
@@ -105,7 +107,7 @@ export class EnvironmentVariableService extends Disposable implements IEnvironme
 	}
 
 	private _resolveMergedCollection(): IMergedEnvironmentVariableCollection {
-		return new MergedEnvironmentVariableCollection(this.collections);
+		return new MergedEnvironmentVariableCollection(this.collections, this._logService);
```
Tyriar: Since this is a ctor you want to use dependency injection; make sure you use the `@ITerminalLogService` decorator in the constructor impl and then use:

```
instantiationService.createInstance(MergedEnvironmentVariableCollection, this.collections)
```

src/vs/workbench/contrib/terminal/common/scripts/shellIntegration-bash.sh
Tyriar: For that message you'll need to write the raw escape sequences. Try printing these:

```
\x1b[0m\x1b[7m * \x1b[0;103m <message>\x1b[0m
\x1b[0m\x1b[7m * \x1b[0m <message>
```

src/vs/workbench/contrib/terminal/common/scripts/shellIntegration.ps1
```diff
@@ -68,6 +68,21 @@ if ($env:VSCODE_ENV_APPEND) {
...
+if ($env:VSCODE_PWSH_ACTIVATE -and $env:TERM_PROGRAM -eq 'vscode') {
+	$activateScript = $env:VSCODE_PWSH_ACTIVATE
+	Remove-Item Env:VSCODE_PWSH_ACTIVATE
+
+	try {
+		Invoke-Expression $activateScript
+	}
+	catch {
+		$activationError = $_
+		# TODO: Figure out how to add 'star' icon similar to terminal restored.
```
Tyriar: For pwsh use `e instead of \x1b

src/vs/workbench/contrib/terminal/test/common/environmentVariableCollection.test.ts
```diff
@@ -37,7 +38,7 @@ suite('EnvironmentVariable - MergedEnvironmentVariableCollection', () => {
 						['A-key', { value: 'a4', type: EnvironmentVariableMutatorType.Append, variable: 'A', options: { applyAtProcessCreation: true, applyAtShellIntegration: true } }]
 					])
 				}]
-			]));
+			]), new NullLogService());
```
Tyriar: Switching these to use instantiation service would be best. Initialize it like this in tests to get `NullLogService` created for you:

https://github.com/microsoft/vscode/blob/57235e35e5be1ea9db32da9b57862c9ea1880a7f/src/vs/workbench/contrib/terminalContrib/chatAgentTools/test/browser/runInTerminalTool.test.ts#L57-L59

You don't need the config service part for your use case.

Tyriar (changes requested):

src/vs/workbench/contrib/terminal/common/environmentVariableService.ts
anthonykim1: InstantiationService, was fine here, then I went to use instantiation service here: https://github.com/microsoft/vscode/blob/0e00a15ef0c59d39aff227c63fb3c55b1b25f9ac/src/vs/platform/terminal/node/terminalEnvironment.ts#L303 

Journey to get instantiation service for other files: 
Needing instantiation service here: `addEnvMixinPathPrefix`in terminalEnvironment.ts -> `getShellIntegrationInjection` in terminalEnvironment.ts-> `start()` in terminalProcess.ts --> `createProcess and  newTerminalProcess` in ptyService.ts needs instantiationService...

Basically lead me to `ptyHostMain.ts` https://github.com/microsoft/vscode/blob/0e00a15ef0c59d39aff227c63fb3c55b1b25f9ac/src/vs/platform/terminal/node/ptyHostMain.ts#L83 and `ptyService` https://github.com/microsoft/vscode/blob/0e00a15ef0c59d39aff227c63fb3c55b1b25f9ac/src/vs/platform/terminal/node/ptyService.ts#L136-L137
Not quite sure how and if we should get instantiationService from ptyService..

 Feels like too much pain just to log `we blocked non-python extension from modifying activate var`.

Should I just skip log?

anthonykim1:
Here is how the * shows up for example (ofc, exit code will be different when it actually fails, just did this to show what UI would look like): 
Screenshot 2025-08-12 at 11 42 33 PM

extensions/terminal-suggest/src/terminalSuggestMain.ts
```diff
@@ -521,7 +521,7 @@ export function sanitizeProcessEnvironment(env: Record<string, string>, ...prese
 	const keysToRemove = [
 		/^ELECTRON_.$/,
 		/^VSCODE_(?!(PORTABLE|SHELL_LOGIN|ENV_REPLACE|ENV_APPEND|ENV_PREPEND)).$/,
-		/^VSCODE_(PWSH|ZSH|BASH)_ACTIVATE$/,
+		/^VSCODE_PYTHON_(PWSH|ZSH|BASH)_ACTIVATE/,
```
Tyriar: The top one is actually a negative lookahead, so the python keys would already get removed from it

src/vs/base/common/processes.ts
```diff
@@ -109,7 +109,7 @@ export function sanitizeProcessEnvironment(env: IProcessEnvironment, ...preserve
 	const keysToRemove = [
 		/^ELECTRON_.+$/,
 		/^VSCODE_(?!(PORTABLE|SHELL_LOGIN|ENV_REPLACE|ENV_APPEND|ENV_PREPEND)).+$/,
```
Tyriar: The `.` seems like a bug 🤔 @anthonykim1 can you verify just this deleted all variables except for those explicitly mentioned in the lookahead (`?!`)?

```suggestion
		/^VSCODE_(?!(PORTABLE|SHELL_LOGIN|ENV_REPLACE|ENV_APPEND|ENV_PREPEND))$/,
```

Can you add the `_PYTHON` cases to the test?

Tyriar (changes requested):

src/vs/base/common/processes.ts
anthonykim1: @Tyriar I just tested this by hand (the .+), and actually we need the `.+`.
Otherwise the other VSCODE variables does not get filtered.

Also I found out we did not have the `.+` in the terminalSuggest code, so I've added them as part of: [0084fe1](https://github.com/microsoft/vscode/pull/261094/commits/0084fe13c00c661047c6c8955a7982ae7390a4da)


With .+ -->
RmPy +Good


Without the .+ (variables dont get filtered as we desired)--> 

RmPyNo +Badc

src/vs/base/test/common/processes.test.ts
```diff
@@ -37,3 +37,35 @@ suite('Processes', () => {
 		assert.strictEqual(Object.keys(env).length, 3);
 	});
 });
+
+
+test('sanitizeProcessEnvironment removes Python activation variables', () => {
```
Tyriar: Should be moved into the suite above

Tyriar (changes requested):

Tyriar (dismissed):

Tyriar (approved):

PR #261139, opened 2025-08-12
Adding actions to local chats toolbar

osortega:
Adding actions to local chat sessions toolbar

### DIFF
--- a/src/vs/workbench/contrib/chat/browser/actions/chatActions.ts
+++ b/src/vs/workbench/contrib/chat/browser/actions/chatActions.ts
@@ -927,35 +927,6 @@ export function registerChatActions() {
 				resource: ChatEditorInput.getNewEditorUri(),
 				options: {
 					pinned: true,
-					auxiliary: { compact: true, bounds: { width: 640, height: 640 } }
-				} satisfies IChatEditorOptions
-			}, AUX_WINDOW_GROUP);
-		}
-	});
-
-	registerAction2(class OpenChatEditorInNewMaximizedWindowAction extends Action2 {
-		constructor() {
-			super({
-				id: `workbench.action.chat.newChatInNewMaximizedWindow`,
-				title: localize2('chatSessions.openNewChatInNewMaximizedWindow', 'Open New Chat in Maximized Window'),
-				f1: false,
-				category: CHAT_CATEGORY,
-				precondition: ChatContextKeys.enabled,
-				menu: {
-					id: MenuId.ViewTitle,
-					group: 'submenu',
-					order: 1,
-					when: ContextKeyExpr.equals('view', `${VIEWLET_ID}.local`),
-				}
-			});
-		}
-
-		async run(accessor: ServicesAccessor) {
-			const editorService = accessor.get(IEditorService);
-			await editorService.openEditor({
-				resource: ChatEditorInput.getNewEditorUri(),
-				options: {
-					pinned: true,
 					auxiliary: { compact: false }
 				} satisfies IChatEditorOptions
 			}, AUX_WINDOW_GROUP);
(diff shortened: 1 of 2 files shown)


src/vs/workbench/contrib/chat/browser/actions/chatActions.ts
```diff
@@ -903,6 +903,82 @@ export function registerChatActions() {
...
+				} satisfies IChatEditorOptions
+			}, AUX_WINDOW_GROUP);
+		}
+	});
+
+	registerAction2(class NewChatInSideBarAction extends Action2 {
+		constructor() {
+			super({
+				id: `workbench.action.chat.newChatInSideBar`,
+				title: localize2('chatSessions.openNewChatInSideBar', 'Open New Chat in Side Bar'),
```
roblourens: How is this different from the normal new chat action?

src/vs/workbench/contrib/chat/browser/chatSessions.ts
```diff
@@ -962,3 +962,43 @@ MenuRegistry.appendMenuItem(MenuId.ViewTitle, {
 	when: ContextKeyExpr.equals('view', `${VIEWLET_ID}.local`),
 });
 
+MenuRegistry.appendMenuItem(MenuId.ViewTitle, {
```
roblourens: You could set up these menu registrations in the same call as `registerAction2` with `menus:`? But if you want to do it this way, please share some const with the action ID instead of copying strings around, this is a pain for tracking where things are used.

src/vs/workbench/contrib/chat/browser/actions/chatActions.ts
osortega: This command clears the chat widget as well

src/vs/workbench/contrib/chat/browser/chatSessions.ts
osortega: Ahh good suggestion, I move them into their `registerAction2` method

src/vs/workbench/contrib/chat/browser/actions/chatActions.ts
roblourens: "new chat" does that too, that's what starting a new chat means

src/vs/workbench/contrib/chat/browser/actions/chatActions.ts
```diff
@@ -903,6 +904,133 @@ export function registerChatActions() {
...
+				chatWidget.focusInput();
+			}
+		}
+	});
+
+	registerAction2(class OpenChatInNewEditorGroupAction extends Action2 {
+		constructor() {
+			super({
+				id: 'workbench.action.chat.openInNewEditorGroup',
+				title: localize2('chat.openInNewEditorGroup.label', "Open Chat in New Editor Group"),
```
roblourens: We usually say "to the side" 

Image

DonJayamanne (dismissed):

src/vs/workbench/contrib/chat/browser/actions/chatActions.ts
osortega: @roblourens Sorry I meant that it clears and reveals the new chat. I believe "new chat" only creates a new one but doesn't bring it up

src/vs/workbench/contrib/chat/browser/actions/chatActions.ts
osortega: Ahh nice! I like that better

lramos15 (dismissed):

src/vs/workbench/contrib/chat/browser/actions/chatActions.ts
roblourens: We could change it to do that. I don't know why it wouldn't

osortega: @roblourens we might still want a different title, so it's not confused with the new chat action. In this case it should be clear for user that they will clear and open the widget as opposed to creating a new chat entry in the local chat sessions menu

connor4312 (approved):

PR #254285, opened 2025-07-06
feat: Include/ Exclude file types from file search

kplates:
This pull request introduces two new actions: ExcludeFileTypeFromSearchAction and IncludeFileTypeInSearchAction. These actions allow users to modify search patterns by excluding or including specific file types directly from the search result file context menu.

image

**How to test**

- Create a new repo or directory
- Add a few files with with different file types but some common value, we can also have files such as .spec.ts or further dot notation such as some.file.test.spec.ts
- Open search explorer and search for common value
- In the results list right click and select either Include or Exclude File Type From Search

PR #261231, opened 2025-08-12
Add quick file-type filter in global search

### LINKED ISSUE #261212: Add Quick File-Type Filter in Global Search ( e.g, *.ts, *.md ) Without editing "File to Include"
**Feature Request:** Quick file-type filter in search

Right now, if I want to search only in certain file types, I have to go to the *Files to include* field and type something like `*.ts` or `*.md` every time.  
It works, but it’s not very fast if you often switch between file types in a big project.

What I think would help:
- A small dropdown next to the search box with common file types (`*.ts`, `*.js`, `*.md`) and maybe a "Custom…" option.
- Or a quick inline shortcut in the search bar, like `:ts` or `:md`, that instantly applies the filter.

Why it would be useful:
- Less typing, faster searches.
- Easier to jump between file types without touching *Files to include*.
- Should be possible as a small UI enhancement without major changes to search logic.

Example:
1. Search: `myFunction` → select `*.ts` → only TypeScript results show up.
2. Search: `docs :md` → only Markdown results show up.

Similar feature exists in JetBrains IDEs, and it’s very convenient. Would love to see something like this in VS Code.

scrollDynasty:
- Add dropdown with common file types (*.ts, *.js, *.md)
- Add inline shortcuts support (:ts, :md)
- Fixes #261212


scrollDynasty:

scrollDynasty:
@osortega can you check my pull request pls and accept if you think it is necessary because I think it is very useful

osortega:
@scrollDynasty this change is a bit too disruptive to the current search view.
Would the change introduced [here](https://github.com/microsoft/vscode/pull/254285) be enough for a quicker way of filtering out to a particular file type?

scrollDynasty:
@osortega Thanks for the feedback! Understood that this is too disruptive to the current search view. The change in #254285 helps when I already have results. My original goal was a very fast way to pre-filter by file type before running a search and via keyboard only (e.g., :ts expanding to *.ts into “files to include”). If that sounds reasonable, I can pivot this to a minimal change: just support the “:ts, :md, …” inline shortcuts that insert the corresponding glob into “files to include,” reusing the existing mechanisms and without adding new UI elements. If you’d still prefer not to add that, happy to defer to #254285 alone.

scrollDynasty:
@osortega I've implemented a new feature that introduces inline shortcuts for file type filtering directly within the global search input. This enhancement is aimed at power users, allowing for quicker and more efficient file type-based searches without any changes to the existing UI.
Inline Shortcuts: Users can now type shortcuts like :ts directly into the search field to instantly filter results by file type.
No UI Changes: The feature operates seamlessly in the background, maintaining a clean and uncluttered search interface.
New: src/vs/workbench/contrib/search/browser/searchShortcuts.ts – Core utility for processing search shortcuts.
Modified: src/vs/workbench/contrib/search/browser/searchView.ts – Integration of shortcut processing into the search logic within _onQueryChanged (approximately lines 1560–1570).
How It Works:
User Input: A user types a shortcut (e.g., :ts) into the search field.
Automatic Conversion: The shortcut is automatically converted into a file filter (e.g., **/*.ts) and applied to the hidden "files to include" field.
Invisible Application: The UI remains unchanged; no new inputs or visible modifications appear.
Single Active Shortcut: By default, only one shortcut is active at a time, unless multiple are explicitly specified.

PR #261265, opened 2025-08-12
show desktop about details even when using custom dialogs

sbatten:
Historically the about dialogs created their details themselves meaning they could only display details available in their layer. Now, about takes the detail information and we can calculate desktop details for both custom(html) and native dialogs.

image

This was always a pet peeve of mine since I use custom and I often use about to check our current node and electron version.


bpasero (approved):
Nice, how did I never notice this 🤯

bpasero:
@sbatten maybe make an issue with verification needed to cover this during endgame.

PR #261272, opened 2025-08-12
cap retrieved terminal chars at 16k

### LINKED ISSUE #260791: Really large tool result from get_task_output
When `get_task_output` was in the extension, we had this code

`const buffer = this.terminalService.getBufferForTerminal(terminal, Math.min(options.input.maxCharsToRetrieve ?? 16000, 16000));`

Now that it's in core, I see the tool returning extremely long results, wasting tokens and doing very frequent summarization. And I don't see where it's output would be limited. Why did we remove that?

```
# Tool Call - call_yjpOvYmciC9cA2N08iciyCMo__vscode-1754686273723

## Request
~~~
id   : call_yjpOvYmciC9cA2N08iciyCMo__vscode-1754686273723
tool : get_task_output
args : {
  "id": "npm: Core - Build",
  "workspaceFolder": "/Users/roblou/code/vscode"
}
~~~
## Response
~~~

[5000 lines of output]
```

Image

meganrogge:
fix #260791

Also fix start line

### DIFF
--- a/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/bufferOutputPolling.ts
+++ b/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/bufferOutputPolling.ts
@@ -75,18 +75,24 @@ export async function racePollingOrPrompt(
 
 
 export function getOutput(instance: ITerminalInstance, startMarker?: IXtermMarker): string {
-	if (!instance.xterm || !instance.xterm.raw) {
+	if (!instance.xterm?.raw) {
 		return '';
 	}
-	const lines: string[] = [];
-	for (let y = Math.min(startMarker?.line ?? 0, 0); y < instance.xterm!.raw.buffer.active.length; y++) {
-		const line = instance.xterm!.raw.buffer.active.getLine(y);
-		if (!line) {
-			continue;
-		}
-		lines.push(line.translateToString(true));
+	const buffer = instance.xterm.raw.buffer.active;
+	const startLine = Math.max(startMarker?.line ?? 0, 0);
+	const endLine = buffer.length;
+	const lines: string[] = new Array(endLine - startLine);
+
+	for (let y = startLine; y < endLine; y++) {
+		const line = buffer.getLine(y);
+		lines[y - startLine] = line ? line.translateToString(true) : '';
 	}
-	return lines.join('\n');
+
+	let output = lines.join('\n');
+	if (output.length > 16000) {
+		output = output.slice(-16000);
+	}
+	return output;
 }
 
 export async function pollForOutputAndIdle(
--- a/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/bufferOutputPolling.ts
+++ b/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/bufferOutputPolling.ts
@@ -13,8 +13,7 @@ import { ChatModel } from '../../../chat/common/chatModel.js';
 import { IChatService } from '../../../chat/common/chatService.js';
 import { ChatMessageRole, ILanguageModelsService } from '../../../chat/common/languageModels.js';
 import { IToolInvocationContext } from '../../../chat/common/languageModelToolsService.js';
-import { ITerminalInstance } from '../../../terminal/browser/terminal.js';
-import type { IMarker as IXtermMarker } from '@xterm/xterm';
+import type { Terminal as RawXtermTerminal, IMarker as IXtermMarker } from '@xterm/xterm';
 
 export const enum PollingConsts {
 	MinNoDataEvents = 2, // Minimum number of no data checks before considering the terminal idle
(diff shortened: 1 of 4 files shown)


src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/bufferOutputPolling.ts
```diff
@@ -75,18 +75,24 @@ export async function racePollingOrPrompt(
...
+	const lines: string[] = new Array(endLine - startLine);
+
+	for (let y = startLine; y < endLine; y++) {
+		const line = buffer.getLine(y);
+		lines[y - startLine] = line ? line.translateToString(true) : '';
+	}
+
+	let output = lines.join('\n');
+	if (output.length > 16000) {
+		output = output.slice(-16000);
```
Tyriar, line 92: Let's add a test that enforces this

Tyriar (dismissed):

src/vs/workbench/contrib/terminalContrib/chatAgentTools/test/browser/bufferOutputPolling.test.ts
```diff
@@ -18,6 +19,19 @@ suite('racePollingOrPrompt', () => {
 	const defaultLanguageModelsService = {} as any;
 	const defaultExecution = { getOutput: () => 'output' };
 
+	test('getOutput enforces 16000 character limit', () => {
+		const longString = 'A'.repeat(17000);
+		const fakeBuffer = {
+			length: 1,
+			getLine: () => ({ translateToString: () => longString }),
+		};
+		const fakeXterm = { raw: { buffer: { active: fakeBuffer } } };
```
Tyriar: It's easy to init actual xterm for tests, the more unnecessary fake stuff the harder it is to maintain. You could pass in `xterm: XtermTerminal | undefined` instead of instance to simplify the call

Tyriar (dismissed):

src/vs/workbench/contrib/terminalContrib/chatAgentTools/test/browser/bufferOutputPolling.test.ts
meganrogge: it's a lot of lines, but did that 👍🏼

src/vs/workbench/contrib/terminalContrib/chatAgentTools/test/browser/bufferOutputPolling.test.ts
```diff
@@ -5,19 +5,76 @@
...
+			terminal: {
+				integrated: defaultTerminalConfig
+			}
+		});
+
+		instantiationService = workbenchInstantiationService({
+			configurationService: () => configurationService
+		}, store);
+
+		XTermBaseCtor = (await importAMDNodeModule<typeof import('@xterm/xterm')>('@xterm/xterm', 'lib/xterm.js')).Terminal;
```
Tyriar: You can avoid most of this ceremony if you just work off the raw object from xterm, so this in the bufferOutputPolling.ts file:

```
import type { Terminal as RawXtermTerminal } from '@xterm/xterm';
```

Tyriar (dismissed):

src/vs/workbench/contrib/terminalContrib/chatAgentTools/test/browser/bufferOutputPolling.test.ts
```diff
@@ -18,6 +19,38 @@ suite('racePollingOrPrompt', () => {
...
+			}
+		};
+		const terminal: Pick<RawXtermTerminal, 'buffer'> = {
+			buffer: {
+				active: buffer,
+				normal: buffer,
+				alternate: buffer,
+				onBufferChange: new Emitter<IBuffer>().event
+			},
+		};
```
Tyriar: Instead of faking all this now it should just be a matter of creating an actual xterm.js terminal (not `XtermTerminal`) and writing `longString` to it.

Tyriar (changes requested):

src/vs/workbench/contrib/terminalContrib/chatAgentTools/test/browser/bufferOutputPolling.test.ts
meganrogge: I tried that originally but was getting this error. I've suppressed that error for now. 

Screenshot 2025-08-12 at 4 01 39 PM

Tyriar: This is how you do it in tests for now:

https://github.com/microsoft/vscode/blob/634b6dcfa890fb81e71f862d05d36cad5a7c8198/src/vs/workbench/contrib/terminal/test/browser/xterm/xtermTerminal.test.ts#L120

src/vs/workbench/contrib/terminalContrib/chatAgentTools/test/browser/bufferOutputPolling.test.ts
```diff
@@ -18,6 +20,15 @@ suite('racePollingOrPrompt', () => {
...
 	const defaultExecution = { getOutput: () => 'output' };
 
+	test('getOutput enforces 16000 character limit', () => {
+		const longString = 'A'.repeat(17000);
+		const terminal = new RawXtermTerminal();
+		terminal.write(longString, () => {
+			const output = getOutput(terminal);
+			assert.strictEqual(output, longString.slice(-16000));
+		});
+	});
```
Tyriar: Since it's a callback, you need an `async () =>` test. Use this helper for write:

```ts
	function write(data: string): Promise<void> {
		return new Promise<void>((resolve) => {
			xterm.write(data, resolve);
		});
	}
```

Tyriar (dismissed):

DonJayamanne (approved):

PR #261318, opened 2025-08-12
Fix manage models being improperly shown

lramos15:
Fix https://github.com/microsoft/vscode/issues/260955

### DIFF
--- a/src/vs/workbench/contrib/chat/common/chatEntitlementService.ts
+++ b/src/vs/workbench/contrib/chat/common/chatEntitlementService.ts
@@ -101,6 +101,7 @@ export interface IChatEntitlementService {
 	readonly onDidChangeEntitlement: Event<void>;
 
 	readonly entitlement: ChatEntitlement;
+	readonly isInternal: boolean;
 
 	readonly onDidChangeQuotaExceeded: Event<void>;
 	readonly onDidChangeQuotaRemaining: Event<void>;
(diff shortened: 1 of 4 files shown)


benibenj (approved):

src/vs/workbench/contrib/chat/common/chatEntitlementService.ts
```diff
@@ -1004,9 +1025,10 @@ export class ChatEntitlementContext extends Disposable {
 	update(context: { installed: boolean; disabled: boolean; untrusted: boolean }): Promise<void>;
 	update(context: { hidden: boolean }): Promise<void>;
 	update(context: { later: boolean }): Promise<void>;
-	update(context: { entitlement: ChatEntitlement }): Promise<void>;
-	update(context: { installed?: boolean; disabled?: boolean; untrusted?: boolean; hidden?: boolean; later?: boolean; entitlement?: ChatEntitlement }): Promise<void> {
+	update(context: { entitlement: ChatEntitlement; isInternal?: boolean }): Promise<void>;
+	update(context: { installed?: boolean; disabled?: boolean; untrusted?: boolean; hidden?: boolean; later?: boolean; entitlement?: ChatEntitlement; isInternal?: boolean }): Promise<void> {
 		this.logService.trace(`[chat entitlement context] update(): ${JSON.stringify(context)}`);
+		this.state.isInternal = !!context.isInternal;
```
bpasero, line 1031: @lramos15 this is not good, there are calls into this method without the `isInternal` flag and thus they would overwrite this flag by accident. Please take a look at my follow up change in https://github.com/microsoft/vscode/pull/261596

Issue #255704, opened 2025-07-14
Improve next edit suggestions for notebooks


DonJayamanne:
Treat the whole notebook as a single text document.
The expectation is NES will get more context than just the current cell (e.g. variables from previous cell & the like)

ulugbekna:
There's `VSCodeWorkspace` ([ref](https://github.com/microsoft/vscode-copilot-chat/blob/b7251b1ce4e2d9d90c13ac31388048b1a2a106a7/src/extension/inlineEdits/vscode-node/parts/vscodeWorkspace.ts#L30)) that listen to document open/close/change and make observable values from those. 

There's `NesXtabHistoryTracker` ([ref](https://github.com/microsoft/vscode-copilot-chat/blob/b7251b1ce4e2d9d90c13ac31388048b1a2a106a7/src/platform/inlineEdits/common/workspaceEditTracker/nesXtabHistoryTracker.ts#L49)) that subscribe document change observables and put those edits into an array that represents edit/selection-change history. 

I'm thinking that one of the entities above (probably the latter) should use a single document to represent the whole notebook. When it receives a text document change event, it should translate a text document (cell change) event to the whole-notebook-document change, ie map `/foo/bar/nb.ipynb#cellID1` -> `/foo/bar/nb.ipynb` and translate edit range to the one in the string that represents `/foo/bar/nb.ipynb`. 

I think doing this in the latter component may be nicer, but I don't know how to go from "I received a change event for a cell" to "I have the contents of the whole notebook" in a nice way that wouldn't require vscode API (because we don't use vscode API at this abstraction layer)

DonJayamanne:
@hediet Perhaps you can help.
The plan is to represent the notebook as a single virtual string/text document and use that for NES.
Looking at `IObservableDocument`, it seems feasible.
However before we go down that path, wanted to check if you had any thoughts/concerns or alternatives

DonJayamanne:
>  "I received a change event for a cell" to "I have the contents of the whole notebook" in a nice way that wouldn't require vscode API (because we don't use vscode API at this abstraction layer)

Yeah, this won't be pretty, but i guess this means we'd need to bake in the concept of notebooks into the recorder.
As thats where these events are handled. 
Again perhaps @hediet can help.

hediet:
`VSCodeWorkspace` could represent a notebook document as a single IObservableDocument that represents the concatenated cells (with a separator).

[...]

PR #261383, opened 2025-08-13
Display NES for notebook cells on focus

DonJayamanne:
For https://github.com/microsoft/vscode/issues/255704

### DIFF
--- a/src/vs/editor/contrib/inlineCompletions/browser/controller/inlineCompletionsController.ts
+++ b/src/vs/editor/contrib/inlineCompletions/browser/controller/inlineCompletionsController.ts
@@ -8,8 +8,11 @@ import { timeout } from '../../../../../base/common/async.js';
 import { cancelOnDispose } from '../../../../../base/common/cancellation.js';
 import { createHotClass } from '../../../../../base/common/hotReloadHelpers.js';
 import { Disposable, toDisposable } from '../../../../../base/common/lifecycle.js';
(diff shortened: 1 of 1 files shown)


hediet (changes requested):
I believe this PR is about making this use-case work:
1. The user edits cell1 and gets a suggestion to accept an edit in cell2
2. The user presses tab and jumps to cell2
3. Now cell2 shows the suggestion

...by using "onFocus" as signal to detect end of step 2 and then calling ".trigger".

Howeve, I think the command that is invoked when the user presses tab in step 2 to jump to  cell2 should also call trigger, instead of the editor waiting for the onfocus event.

Otherwise, random cell focus changes will trigger NES.

DonJayamanne:
> Otherwise, random cell focus changes will trigger NES.

yes, That is the idea, for notebooks, 
i.e there's no need to wait for NES to be triggered via commands & set focus

As its similar to a regular single text document, i.e. user moves around the text document we get NES getting computed.

hediet:
>yes, That is the idea, for notebooks,
>i.e there's no need to wait for NES to be triggered via commands & set focus
>As its similar to a regular single text document, i.e. user moves around the text document we get NES getting computed.

In that case this should be implemented together with that logic here:

https://github.com/microsoft/vscode-copilot-chat/blob/75afad3ef60da47f70f63c26268574e6bcdc9437/src/extension/inlineEdits/vscode-node/inlineEditModel.ts#L96

PR #261384, opened 2025-08-13
Avoid colorizing inline completion display label

DonJayamanne:

### DIFF
--- a/src/vs/editor/contrib/inlineCompletions/browser/view/inlineEdits/inlineEditsView.ts
+++ b/src/vs/editor/contrib/inlineCompletions/browser/view/inlineEdits/inlineEditsView.ts
@@ -36,6 +36,7 @@ import { IOriginalEditorInlineDiffViewState, OriginalEditorInlineDiffView } from
 import { applyEditToModifiedRangeMappings, createReindentEdit } from './utils/utils.js';
 import './view.css';
 import { $ } from '../../../../../../base/browser/dom.js';
+import { isEqual } from '../../../../../../base/common/resources.js';
 
 
 export class InlineEditsView extends Disposable {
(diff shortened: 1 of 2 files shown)


src/vs/editor/contrib/inlineCompletions/browser/view/inlineEdits/inlineEditsView.ts
```diff
@@ -254,6 +255,7 @@ export class InlineEditsView extends Disposable {
 		this._customView = this._register(this._instantiationService.createInstance(InlineEditsCustomView,
 			this._editor,
 			this._model.map((m, reader) => this._uiState.read(reader)?.state?.kind === 'custom' ? m?.displayLocation : undefined),
+			this._model.map((m, reader) => this._uiState.read(reader)?.state?.kind !== 'custom' || !m?.uri || isEqual(m.uri, this._editor.getModel()?.uri)),
```
DonJayamanne, line 258: This doesn't feel like the right solution.
I think a better fix is for `InlineCompletionDisplayLocation` to contain some indicator whether the `label` is a block of code or just a regular label.

E.g. 
```typescript
export type InlineCompletionDisplayLocation = {
		range: Range;
		label: string;
	} | {
		range: Range;
		code: string;
	}
```

I.e. convert to union having either `label` or `code`
And labels are displayed as is without any tokenization and colorization..

DonJayamanne: FYI: Today I'm returning the text `Go To Next Cell` from chat extension, however this gets tokenized as python code.

@hediet /cc

hediet: > This doesn't feel like the right solution.

Yes, this is not good. Instead, this should happen where you return "custom" view in case the uri does not match the current document.

The union type looks good!

hediet (changes requested):
Instead of using `labelContainsCode`, it should be label: `{ kind: 'code', value: string } | { kind: 'text', value: string }`.

PR #261390, opened 2025-08-13
view diffs between worktree and current workspace

nguyenchristy:
PR experiments with viewing diffs and staging files across repositories. Users can access these commands by the 'arrow-up' icon.
stage
stageall


extensions/git/package.json
```diff
@@ -1967,6 +1977,11 @@
           "when": "scmProvider == git && scmResourceGroup == index",
           "group": "2_view@2"
         },
+        {
+          "command": "git.compareWithWorkspace",
```
bhavyaus, line 2005: should the command name be `git.openWorktreeChange` ? To keep it inline with the existing `git.openChange`

bhavyaus: or `git.compareWorktreeChange` ? The current workspace is implied.

bhavyaus (changes requested):

extensions/git/src/repository.ts
```diff
@@ -507,9 +512,37 @@ class ResourceCommandResolver {
 	}
 
 	resolveChangeCommand(resource: Resource): Command {
+		return this.resolveComparisonCommand(resource);
```
bhavyaus: Could you undo the changes to resolveChangeCommand?
You should perform all of the work necessary for the worktree resource resolution in `resolveCompareWithWorkspaceCommand` and then call into `resolveChangeCommand`

bhavyaus: Could we also change the name of the function as we rename the command?

extensions/git/package.json
nguyenchristy: I choose Compare with Workspace because if we say Compare Worktree Change, how does that make it clear that Compare Worktree Change is different than Open Change in the context menu of a worktree?

Issue #261073, opened 2025-08-11
Don't offer auto approve options for subshell executions


Tyriar:
Since this i running `powershell -Command`, it shouldn't offer sub-command approval since we don't understand them currently:

<img width="358" height="102" alt="Image" src="https://github.com/user-attachments/assets/c3d777de-d752-48ea-af39-7b0e892d5c50" />

Tyriar:
Python, bash, etc.

Tyriar:
Detecting what bash/powershell will do would be better, maybe disallow if it's messing with execution policy though.

Tyriar:
This is a little difficult to do without an actual command line parser due to quoting being involved. Let's put on hold for now.

Tyriar:
This is done

PR #261510, opened 2025-08-13
Detect sub-shell executions

Tyriar:
Part of #261073


Tyriar:
See https://github.com/microsoft/vscode/issues/261073#issuecomment-3185207623

PR #261521, opened 2025-08-13
Fix coordinate typos in hoverWidget

LeftPhalange:
This addresses issue #261519 by renaming functions `computeXCordinate` and `computeYCordinate` to `computeXCoordinate` and `computeYCoordinate` respectively.


Tyriar (commented):
Sorry, we don't accept non-user facing typo corrections

PR #261539, opened 2025-08-13
When a task with problem matchers becomes idle, return those instead of using the LLM to evaluate the output

meganrogge:
This improves accuracy and speed of response. Also added a test.

Fixes https://github.com/microsoft/vscode/issues/261516

### DIFF
--- a/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/bufferOutputPolling.ts
+++ b/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/bufferOutputPolling.ts
@@ -159,7 +159,7 @@ export async function pollForOutputAndIdle(
 		terminalExecutionIdleBeforeTimeout = true;
 		if (execution.task) {
 			const problems = await getProblemsForTasks(execution.task, markerService, execution.dependencyTasks, knownMatchers);
-			if (problems.size) {
+			if (problems) {
 				// Problem matchers exist for this task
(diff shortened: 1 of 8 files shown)


src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/bufferOutputPolling.ts
```diff
@@ -149,10 +155,33 @@ export async function pollForOutputAndIdle(
...
+		if (execution.task) {
+			const problems = await getProblemsForTasks(execution.task, markerService, execution.dependencyTasks, knownMatchers);
+			if (problems) {
+				// Problem matchers exist for this task
+				const problemList: string[] = [];
+				for (const [owner, problemArray] of problems.entries()) {
+					if (problemArray.length) {
+						problemList.push(`Problems for task "${owner ?? 'unknown'}":`);
+						for (const p of problemArray) {
+							problemList.push(`Problem: ${p.message} (Code: ${p.code}, Severity: ${p.severity}, Range: [${p.startLineNumber},${p.startColumn}]-[${p.endLineNumber},${p.endColumn}], Target: ${typeof p.code === 'string' ? p.code : p.code?.value ?? 'unknown'})`);
```
meganrogge: To defer: improve this so it links to the file range

TylerLeonhardt (dismissed):

roblourens (approved):

Issue #255962, opened 2025-07-15
Let's revisit AI actions submenu


isidorn:
<img width="1014" height="464" alt="Image" src="https://github.com/user-attachments/assets/6eafe981-e89e-421d-aa80-be17392a94ad" />

To make AI not look like it is built on top, I suggest to:
1) Remove this submenu, and move relevant actions to the top editor context menu
2) Here's the usage by unique user in last 7 days

<img width="489" height="225" alt="Image" src="https://github.com/user-attachments/assets/794d90ad-8514-43e0-acce-57d62c1fba8b" />

Based on the usage I suggest
1) Remove generate docs and Generate tests Review and comment
2) Add selection to chat only shows up when selection exists
3) Move all others top level -> Editor Inline Chat SEPARATOR Add File to Chat, Explain, Fix

isidorn:
On a higher level all these Fix / Explain go into @workspace slash command experiences. I am not convinced those are better than just asking agent mode to do it with the right prompt.

Also all of those use participants which take the user into Ask mode. @roblourens is working on fixing this, so at least the Agent mode remains.

bpasero:
As I am out soon, moving to August. @roblourens if you wanna do it earlier, feel free.

roblourens:
Hey @aeschli is there any plan for something like built-in prompt files? I agree with Isidor that literally just asking agent mode "explain this code" with the selection attached would probably be a better experience. So the only question is whether to hard-code some prompt string or if we have considered a built-in `/explain` sort of thing. I don't know that there's a need for that, just wondering.

aeschli:
One idea is that we have a contribution point so that extensions can contribute prompt files

eli-w-king:
@bpasero @pierceboggan, pivoting from the proposal in https://github.com/microsoft/vscode/issues/255341. From the comments of this issue and [slack thread](https://vscodeteam.slack.com/archives/C1C2M1KEZ/p1752231016264169), the idea is to move forward with the following highlighted state:   

<img width="780" height="634" alt="Image" src="https://github.com/user-attachments/assets/ab008f1d-4c58-443b-a91f-d3e62f4b29be" />

digitarald:
Review shows up with significant usage in our dashboards.

I hope the mockup doesn't imply that we level menus with `@workspace /fix` or expect users to find the slash commands? Anything that doesn't have a UI entry point doesn't exist to the majority of users.

eli-w-king:
@digitarald
[...]

PR #261547, opened 2025-08-13
Revised context menu

eli-w-king:
Addresses: https://github.com/microsoft/vscode/issues/255962 with the following changes:
1. Renamed 'Inline Editor Chat' and 'Inline Terminal Chat' to 'Open Inline Chat'.
2. 'Add Selection to Chat' only shows when text is highlighted in editor, this follows the terminal behavior.
3. Added 'Explain' from nested menu to inline item.

Changes made in vscode-copilot chat: https://github.com/microsoft/vscode-copilot-chat/pull/590
1. Renamed 'Review and Comment' to 'Code Review' to match change made in SCM panel.
2. Removed 'Explain' from nested menu.


src/vs/workbench/contrib/chat/browser/actions/chatContextActions.ts
```diff
@@ -244,6 +245,7 @@ class AttachSelectionToChatAction extends Action2 {
 				order: 1,
 				when: ContextKeyExpr.and(
 					ChatContextKeys.enabled,
+					EditorContextKeys.hasNonEmptySelection,
```
bpasero, line 248: The consequence of this is that you cannot just right click and add the current cursor location as context, is that intentional? Even without a selection, there might still be value to add the location of the cursor into the chat?

cc @isidorn

Tyriar (changes requested):
Something that needs to be improved here is that both of these are in the inline chat category, so they'll end up looking identical:
 
Screenshot 2025-08-14 at 5 21 13 am

Easy fix would be changing this category to be the standard terminal one so it shows as `Terminal: Open Inline Chat`

src/vs/workbench/contrib/chat/browser/actions/chatActions.ts
```diff
@@ -1577,12 +1577,25 @@ const menuContext = ContextKeyExpr.and(
...
+		id: 'github.copilot.chat.explain',
+		title: localize('explain', 'Explain')
+	},
+	group: '1_chat',
+	order: 4,
+	when: ContextKeyExpr.and(
+		menuContext,
+		ContextKeyExpr.not('github.copilot.interactiveSession.disabled')
+	)
+});
```
bpasero, line 1593: We unfortunately cannot do that without more work: if the Copilot extension is not installed, this will throw an error because the command is not registered.
If we want the same context menu to be present that is present if you have Copilot setup, then we need to provide implementations of these commands that take care of the setup ad-hoc and then forward to the extension when its done. I would do that as a separate feature request / pr as its likely a lot more work.

src/vs/workbench/contrib/inlineChat/browser/inlineChatActions.ts
```diff
@@ -69,13 +69,21 @@ export class StartSessionAction extends Action2 {
...
+				// Place below "Add Selection to Chat" (order 1) and "Add File to Chat" (order 2)
+				// which both use the '1_chat' group in the editor context menu
+				group: '1_chat',
 				order: 3,
+				when: ContextKeyExpr.and(
+					ContextKeyExpr.or(CTX_INLINE_CHAT_HAS_AGENT, CTX_INLINE_CHAT_HAS_AGENT2),
+					CTX_INLINE_CHAT_POSSIBLE,
+					EditorContextKeys.writable,
+					EditorContextKeys.editorSimpleInput.negate()
+				)
```
bpasero, line 82: Why the extra `when` condition that was not present before in `MenuId.ChatTextEditorMenu`?

bpasero (changes requested):

bpasero:
Continues in https://github.com/microsoft/vscode/pull/262902

PR #261552, opened 2025-08-13
replaced copilot with chat sparkle icon in command center

eli-w-king:
Addresses: https://github.com/microsoft/vscode-internalbacklog/issues/5713


bpasero (commented):
@eli-w-king if you do a search of `Codicon.copilot` in the file `src/vs/workbench/contrib/chat/browser/actions/chatActions.ts`, you will notice 7 matches, including those that have a decoration (`Codicon.copilotNotConnected` and `Codicon.copilotWarning`). Are we considering to preserve those because they relate more to GitHub Copilot or should we have specific decorated variants of Chat?

There is also still a verbatim use in:

https://github.com/microsoft/vscode/blob/2d6f48ea7fb5dfd0aa623e7444f1f4f49c6e6817/src/vs/workbench/contrib/chat/browser/actions/chatActions.ts#L128

bpasero:
continues in https://github.com/microsoft/vscode/pull/262889

Issue #261300, opened 2025-08-12
Allow to visually hide thinking tokens


kieferrm:
About https://github.com/microsoft/vscode/issues/257104

1) enable chat.agent.showThinking
2) run a session with a model that supports it
-> Thinking tokens take up a large part of the response. I'd like to be able to visually hide them, and only reveal them if I want to see them

<img width="548" height="549" alt="Image" src="https://github.com/user-attachments/assets/4c58806a-4668-4789-b22a-30bf3996554c" />

justschen:
sneak peek into what we're looking to do here:
https://github.com/user-attachments/assets/d46d63de-5d72-48fa-a148-a8213b8563c7

PR #261600, opened 2025-08-14
make thinking tokens collapsible INLINE

justschen:
first fix for https://github.com/microsoft/vscode/issues/261300

likely temporary, for the time being while we continue with more UI updates


ifox777:
Good idea

aeschli (approved):

PR #261619, opened 2025-08-14
Support registering views to the secondary side bar

### LINKED ISSUE #151681: Allow extensions to add views to the secondary side bar
Now that the secondary sidebar has fully landed, it would be great if extensions could add views directly to it.

I am hoping to use the secondary sidebar for a new "Commit Details" view in GitLens, that could possibly track the current line (like the line history) and also for showing commit details when we add the ability for having a full commit graph in an editor tab,

/cc @sbatten

benibenj:
```Copilot Generated Description:``` Introduce support for contributing views to the auxiliary bar, allowing extensions to register custom view containers in the secondary side bar. This includes updates to the views extension handler and the addition of a placeholder for the auxiliary bar contribution point.

This was inspired by https://github.com/microsoft/vscode/pull/156526

closes https://github.com/microsoft/vscode/issues/151681

### DIFF
--- a/src/vs/workbench/api/browser/viewsExtensionPoint.ts
+++ b/src/vs/workbench/api/browser/viewsExtensionPoint.ts
@@ -71,8 +71,8 @@ export const viewsContainersContribution: IJSONSchema = {
 			type: 'array',
 			items: viewsContainerSchema
 		},
-		'auxiliarybar': {
-			description: localize('views.container.auxiliarybar', "Contribute views containers to Secondary Side Bar"),
+		'secondarySidebar': {
+			description: localize('views.container.secondarySidebar', "Contribute views containers to Secondary Side Bar"),
 			type: 'array',
 			items: viewsContainerSchema
 		}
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Comment by bpasero** [comment]
Not seeing any red flags, but I think @sandy081 should give the final 👍 

I wonder if we should leak the technical term "auxiliary bar" into the API, why not rather just secondary bar, or secondary side bar.

**Review by benibenj — COMMENTED** [review]
That's a good point. I'm not to sure what we should use as the primary side bar is called `activityBar` and so in theory this would be called `secondaryActivityBar` but that sounds weird. I changed it to `secondarySideBar` which seems reasonable to me.

File: src/vs/workbench/api/browser/viewsExtensionPoint.ts [hunk 1]
```diff
@@ -70,6 +70,11 @@ export const viewsContainersContribution: IJSONSchema = {
 			description: localize('views.container.panel', "Contribute views containers to Panel"),
 			type: 'array',
 			items: viewsContainerSchema
+		},
+		'auxiliarybar': {
```
sandy081: This should be called `secondarySidebar`

sandy081 (approved):

PR #261645, opened 2025-08-14
Modify LM api based on feedback

lramos15:


src/vscode-dts/vscode.proposed.chatProvider.d.ts
```diff
@@ -102,6 +102,27 @@ declare module 'vscode' {
...
 	}
 
+	/**
+	 * The provider version of @link {LanguageModelChatMessage}.
+	 */
+	export interface LanguageModelChatRequestMessage {
+		/**
+			* The role of this message.
+			*/
+		role: LanguageModelChatMessageRole;
```
mjbvz: Since this is all being passed to the provider, I think we should mark the fields as readonly

mjbvz (dismissed):
API changes look good. Just a few nits and comment on another part of the api

mjbvz (approved):

PR #261707, opened 2025-08-14
add count of problems to chat response progress

meganrogge:
Screenshot 2025-08-14 at 4 31 22 PM

### DIFF
--- a/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/task/runTaskTool.ts
+++ b/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/task/runTaskTool.ts
@@ -99,26 +99,28 @@ export class RunTaskTool implements IToolImpl {
 			});
 		}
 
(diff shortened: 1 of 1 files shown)


src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/task/runTaskTool.ts
```diff
@@ -99,26 +99,28 @@ export class RunTaskTool implements IToolImpl {
 			});
 		}
 
+		const details = terminalResults.map(r => `Terminal: ${r.name}\nOutput:\n${r.output}`);
+		const uniqueDetails = Array.from(new Set(details)).join('\n\n');
+		const toolResultDetails = Array.from(new Map(
+			terminalResults
+				.flatMap(r => r.resources?.filter(res => res.uri && res.range).map(res => ({ uri: res.uri, range: res.range })) ?? [])
+				.map(item => [`${item.uri.toString()}-${item.range.toString()}`, item])
```
Tyriar, line 107: Can problems be detected without ranges? We should and count them too if so

Tyriar (approved):

PR #261777, opened 2025-08-15
remove microphone icon from chat input when voice is not installed

isidorn:
I tested and it seems to work good.
Also when I install voice extension - all seems to be back.

Screenshot 2025-08-15 at 11 15 00

Screenshot 2025-08-15 at 11 15 05

fyi @digitarald


bpasero (approved):
Thanks!

PR #261815, opened 2025-08-15
Fix OAuth2 resource parameter compliance with RFC 8707

### LINKED ISSUE #261364: VS Code OAuth2 Flow Violates RFC 8707: Missing Resource Parameter in Token Exchange Request

- VS Code Version:  Version: 1.104.0-insider (Universal)
- OS Version:  15.6 (24G84) MAC

**Problem**
VS Code's external authorization code request includes a resource indicator which allows AuthZ server to construct the token narrowly targeted for the protected resource (MCP server). However, VSCode doesn't include the same resource indicator into subsequent token request (exchange code for tokens) as mandated by rfc8707 (https://datatracker.ietf.org/doc/html/rfc8707#token-endpoint-example-ac). When resource indicator is present on the authorize request but is missing from the token request, some implementations of IDPs will fail to create an audience claim on the resulting token resulting in token validation failures on the protected resource (MCP server).

**Root Cause**
Missing check for the resource indicator leading to a missing resource indicator parameter in the exchange code for token request.

**Impact**
Token validation failures on the protected resource due to missing audience.

**Code Reference**
This code perhaps may be the cause
https://github.com/microsoft/vscode/blob/0e00a15ef0c59d39aff227c63fb3c55b1b25f9ac/src/vs/workbench/api/common/extHostAuthentication.ts#L599

bluedog13:
## Summary

This PR fixes VS Code's OAuth2 flow to comply with RFC 8707 by including the resource parameter in token exchange requests.

## Problem

The current implementation includes the resource indicator in the authorization request but omits it from both the authorization code and refresh token exchange requests. This violates RFC 8707 and can cause token validation failures on protected resources that require the resource parameter to be consistently present throughout the OAuth2 flow.

## Solution

Added the resource parameter to both token exchange methods in the `DynamicAuthProvider` class:
- `exchangeCodeForToken()` - now includes resource parameter when exchanging authorization code for tokens
- `exchangeRefreshTokenForToken()` - now includes resource parameter when refreshing tokens

The fix ensures that when `_resourceMetadata?.resource` is available, it is consistently included in all OAuth2 requests throughout the authentication flow.

## Test Plan

- [x] Verified the fix addresses the specific code location mentioned in #261364 (line 599 in extHostAuthentication.ts)
- [x] Ensured the resource parameter is added in both token exchange scenarios
- [x] Confirmed the implementation follows the same pattern as the existing authorization request code

Fixes #261364


TylerLeonhardt (approved):
This LGTM. We also need to do this here https://github.com/bluedog13/vscode/blob/80c2dec96d860b41e7001a86b93203dafe9eeedb/src/vs/workbench/api/node/extHostAuthentication.ts#L250-L258

but that can be a separate PR since it would be an addition

Tyriar (approved):

Issue #261794, opened 2025-08-15
Use tree sitter for terminal auto approve command parsing


Tyriar:
The current logic for parsing a full command line into sub-commands is a naive string splitting algorithm using the following list:

https://github.com/microsoft/vscode/blob/14978d7cc92765e69d18cbdbe02bfa8faf84aee7/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/subCommands.ts#L17-L48

This should use a proper parser like tree-sitter to parse the command. This would fix cases like `echo "a|b|c"` being incorrectly treated as sub commands `['echo', 'a', 'b', 'c']` for example.

Tyriar:
Example from https://github.com/microsoft/vscode/issues/262317

```
jq '.[] | to_entries' lint.json and head -n 40
```

```
2025-08-19 14:56:26.013 [info] autoApprove: Command line NOT auto-approved []
2025-08-19 14:56:26.013 [info] - Command 'jq '.[]' is approved by allow list rule: jq []
2025-08-19 14:56:26.013 [info] - Command 'to_entries' lint.json' has no matching auto approve entries []
2025-08-19 14:56:26.013 [info] - Command 'head -n 40' is approved by allow list rule: head []
2025-08-19 14:56:26.013 [info] - Command line 'jq '.[] | to_entries' lint.json | head -n 40' has no matching auto approve entries []
```

rseymour:
I've also seen this split on a \| in quotes like `grep "something\|otherthing"` in a weird way.

PR #261817, opened 2025-08-15
Review safe auto approve commands, remove inline command system, add tests for defaults

Tyriar:
Part of https://github.com/microsoft/vscode/issues/259668


bpasero (approved):

houmark:
Hey there @Tyriar!

I've been hit by some of these new rules in the recent days (using Insider and Pre-Release CoPilot/CoPilot Chat).

I think some of this is allowing CoPilot to be too opinionated and make it (AFAIK) impossible (or at least very hard with a long range of complex approving regex'es) for me to approve a simple "scan/get" command like this AWS DDB scan:

`aws dynamodb scan --table-name Table-Name-dev --filter-expression "siteId = :siteId AND published = :pub" --expression-attribute-values '{":siteId": {"S": "myvalue"}, ":pub": {"BOOL": true}}' --select COUNT`

This is being auto-rejected per rule `/\{.+\}/` which is hardcoded in the codebase and thus not possible to even remove in the User Settings. I have not yet tested if I am able to make an allow rules that actually will allow this, but I have a feeling (without having reviewed this code) that deny will always take precedence.

I think it's cool you are trying to protect users from very dangerous commands, but having so wide regex'es like the one above is way, way too wide and overreaching imho. This list should really be super limited to the very worst potential offenders.

Tyriar:
@houmark I agree it's a shame we need to include those, but we need https://github.com/microsoft/vscode/issues/261794 first in order to remove those broad rules in a safe way.

If you're willing to accept the risk\* you can always "unset" them by setting to null:

```
"/\{.+\}/": null
```

This should remove that from the deny list and allow anything that passes to succeed. Is there anything that could have made this more clear as it's called out in the setting description:

image

\* The risk here will probably be fine as long as you take precautions to protect yourself from prompt injection, mainly being very careful around any tool (like web fetch, GH MCP, etc.) that could return content from the internet.

PR #261829, opened 2025-08-15
render token usage

### LINKED ISSUE #248860: Show context size for each Copilot model
Copilot should show the user, in some way, and in the documentation, the context window available for each model.
There is a huge difference between models currently (Gemini Pro 2.5 having around 64k and o4-mini having around 100k) and this is hidden from the user.

When picking a model, this information is really important.

Cursor does this this way: https://docs.cursor.com/models

@isidorn

isidorn:
Fixes https://github.com/microsoft/vscode/issues/248860

This is a first stab at rendering token usage in Chat. Feedback is very much welcome. Open questinos:
1) Do we already have a utility that summs up all the tokens in the session (I could not find it, so I am manully implementing it in the renderer). I would love to simplify this.
2) I am not a fan of "(preview)(0x)" rendering - double brackets. Ideas on what better to do are welcome

Next steps are to colour this in some warning colour once we are above 85%. @roblourens I think that is the limit you use for summarization 

I was not convinced where to render the token usage, but after much delibaration I think showing it only on the last request actually works nicely. E.g. then it is more married to the session than to the actual request (hopefully).

Justin if you do not like the implementation let me know and I can start over - I tried to keep it minimal.

https://github.com/user-attachments/assets/12ba51df-0d42-4028-9a3e-7972ebd818a0

fyi @jo-oikawa @pierceboggan

### DIFF
--- a/src/vs/workbench/contrib/chat/browser/chatListRenderer.ts
+++ b/src/vs/workbench/contrib/chat/browser/chatListRenderer.ts
@@ -87,6 +87,8 @@ import { ChatMarkdownDecorationsRenderer } from './chatMarkdownDecorationsRender
 import { ChatMarkdownRenderer } from './chatMarkdownRenderer.js';
 import { ChatEditorOptions } from './chatOptions.js';
 import { ChatCodeBlockContentProvider, CodeBlockPart } from './codeBlockPart.js';
+import { ILanguageModelsService } from '../common/languageModels.js';
+import { CancellationToken } from '../../../../base/common/cancellation.js';
 
 const $ = dom.$;
 
@@ -203,6 +205,7 @@ export class ChatListItemRenderer extends Disposable implements ITreeRenderer<Ch
 		private viewModel: IChatViewModel | undefined,
 		@IInstantiationService private readonly instantiationService: IInstantiationService,
 		@IConfigurationService private readonly configService: IConfigurationService,
+		@ILanguageModelsService private readonly lmService: ILanguageModelsService,
 		@ILogService private readonly logService: ILogService,
 		@IContextKeyService private readonly contextKeyService: IContextKeyService,
 		@IThemeService private readonly themeService: IThemeService,
@@ -544,9 +547,8 @@ export class ChatListItemRenderer extends Disposable implements ITreeRenderer<Ch
 		templateData.footerToolbar.context = element;
 
 		// Render result details in footer if available
-		if (isResponseVM(element) && element.result?.details) {
-			templateData.footerDetailsContainer.textContent = element.result.details;
-			templateData.footerDetailsContainer.classList.remove('hidden');
+		if (isResponseVM(element)) {
+			this.renderFooterDetailsWithTokens(element, index, templateData);
 		} else {
 			templateData.footerDetailsContainer.classList.add('hidden');
 		}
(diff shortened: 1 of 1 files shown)


isidorn:
Alternative approach would be to show the token use information below the input box (in a small font). Here's how LM studio does it

Screenshot 2025-08-15 at 17 44 54

digitarald:
As one consideration for putting it on a message or underneath; we got feedback from a Claude Code user that seeing the live counter increasing for % tokens as streams return gave important context to _"Working …"_.

Meanwhile seeing it in the end of a message gives no signal to potentially stop unnecessary tool iterations.

isidorn:
@digitarald good point. So what you want is to also give a feeling of progress - automatically updating the token count as they are streamed. I think this UI would be powerful.

Not sure what others think - if they like it, I can try to do it.

The missing piece is probably that there is no central place I can listen to for streaming token changes. Because I need a listener for every incoming token, such that I can update the token counter live below the input box. @roblourens ?

Or as agent mode suggest the best I can do is:

Start from IChatService: when a request is sent (listen to onDidSubmitRequest or track sendRequest), grab the session via getSession(sessionId). The session’s selected model (ILanguageModelChatMetadata) has capabilities.agentMode you can check. For each response in that session, subscribe to IChatResponseModel.onDidChange (or consume the low-level stream returned by ILanguageModelsService if you sent the request yourself).

Ristovski:
Just my five cents: having the token/context usage info at the bottom below the chat input as done by LM Studio and suggested as the alternative by @isidorn would be wasteful space-wise imo.

A better approach could be adding it on the same line as the model selector at the bottom of the chat input (i.e. between the currently selected model name and the "Configure tools..." button). That way it can be persistent when scrolling the chat, without wasting any vertical space.

isidorn:
Feedback from Zurich standup:
1. Show the full context window, just by showing 109K the users will get a wrong impression. So it should be 128K (what we get from service). And if we have a part used for tokens/tools it should be shown as part of used
2. We could show the global indicator as well - the global token indicator would update live as the user is adding context
3. We could show actual tokens used, and hover explains the token usage, and explains the state once we hit 85% and summarization is kicking in
4. Looks decent enough to already ship

src/vs/workbench/contrib/chat/browser/chatListRenderer.ts
```diff
@@ -650,6 +652,111 @@ export class ChatListItemRenderer extends Disposable implements ITreeRenderer<Ch
...
+					if (modelIdForCount && it.messageText) {
+						sum += await this.lmService.computeTokenLength(modelIdForCount, it.messageText, CancellationToken.None);
+					}
+				} else if (isResponseVM(it)) {
+					// Count response tokens by concatenating markdown content
+					const respModelId = (it as any).modelId ?? currentModelId;
+					let respText = '';
+					try {
+						for (const part of it.response.value) {
+							if ((part as any).kind === 'markdownContent' && (part as any).content?.value) {
```
roblourens, line 713: It's not really possible for vscode to count up tokens, I believe this needs to come from the actual agent. You need to know how a request was rendered, when it was summarized, the tool results, etc. 

Side note, ideally we would rely more on real token counts from the endpoint vs our own token counting (which is a rough estimate). But prompt-tsx doesn't quite work that way

isidorn: @roblourens Thanks for your feedback.
After thinking more about this, I do not think any of this is needed. I just need to look at the actual response token count from the last response. Since you already send the full conversation as part of the latest request. Right?

roblourens: Yeah, it's just that the count in the response is out of sync with the estimate that we actually use for stuff like summarization, so the results might seem odd. (But I'm not sure whether it would be off by enough to matter much). My second paragraph above is saying that I'd like to use that actual count consistently.

fhemmerrelias:
Bump. Would love to see some movement in this area. Meanwhile you can use the "Show Chat Debug View" option as shown in the screenshots to get a good idea about token consumption as you move along your conversations and tool interactions.
image 
image

PR #261851, opened 2025-08-15
Hide non user selectable models from picker when provider specifies user selectable

lramos15:


rzhao271 (approved):

bscaballo:
why would you use shitty MSFT products?

PR #261999, opened 2025-08-17
storage: optimize sqlite insert with upsert syntax

tmm1:
fixes #260807

cc @bpasero

### DIFF
--- a/src/vs/base/parts/storage/node/storage.ts
+++ b/src/vs/base/parts/storage/node/storage.ts
@@ -104,7 +104,7 @@ export class SQLiteStorageDatabase implements IStorageDatabase {
 				});
 
 				keysValuesChunks.forEach(keysValuesChunk => {
-					this.prepare(connection, `INSERT INTO ItemTable VALUES ${new Array(keysValuesChunk.length / 2).fill('(?,?)').join(',')}`, stmt => stmt.run(keysValuesChunk), () => {
+					this.prepare(connection, `INSERT INTO ItemTable VALUES ${new Array(keysValuesChunk.length / 2).fill('(?,?)').join(',')} ON CONFLICT (key) DO UPDATE SET value = excluded.value WHERE value != excluded.value`, stmt => stmt.run(keysValuesChunk), () => {
 						const keys: string[] = [];
 						let length = 0;
 						toInsert.forEach((value, key) => {
--- a/src/vs/base/parts/storage/node/storage.ts
+++ b/src/vs/base/parts/storage/node/storage.ts
@@ -328,7 +328,7 @@ export class SQLiteStorageDatabase implements IStorageDatabase {
 						// - validate that the DB is not corrupt (the open() call does not throw otherwise)
 						return this.exec(connection, [
 							'PRAGMA user_version = 1;',
-							'CREATE TABLE IF NOT EXISTS ItemTable (key TEXT UNIQUE, value BLOB)'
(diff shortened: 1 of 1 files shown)


src/vs/base/parts/storage/node/storage.ts
```diff
@@ -104,7 +104,7 @@ export class SQLiteStorageDatabase implements IStorageDatabase {
 				});
 
 				keysValuesChunks.forEach(keysValuesChunk => {
-					this.prepare(connection, `INSERT INTO ItemTable VALUES ${new Array(keysValuesChunk.length / 2).fill('(?,?)').join(',')}`, stmt => stmt.run(keysValuesChunk), () => {
+					this.prepare(connection, `INSERT INTO ItemTable VALUES ${new Array(keysValuesChunk.length / 2).fill('(?,?)').join(',')} ON CONFLICT (key) DO UPDATE SET value = excluded.value WHERE value != excluded.value`, stmt => stmt.run(keysValuesChunk), () => {
```
bpasero, line 107: @tmm1 curious, the table already defines what to do on conflict (`ON CONFLICT REPLACE`):

```sql
CREATE TABLE IF NOT EXISTS ItemTable (key TEXT UNIQUE ON CONFLICT REPLACE, value BLOB)
```

Are we not now basically duplicating the work, or does the `INSERT...` statement override the table one?

tmm1: It overrides the table one as shown in my benchmarks on the issue.

bpasero: @tmm1 can we not just set it then on the table statement instead of the insert statement?

tmm1: Not afaik. They are two separate SQLite features

bpasero: It seems redundant and unclear to me to have the conflict condition on the table:

```sql
CREATE TABLE IF NOT EXISTS ItemTable (key TEXT UNIQUE ON CONFLICT REPLACE, value BLOB)
...
INSERT INTO ItemTable VALUES ... ON CONFLICT (key) DO UPDATE SET value = excluded.value WHERE value != excluded.value
```

Why would we not do:

```sql
CREATE TABLE IF NOT EXISTS ItemTable (key TEXT UNIQUE, value BLOB)
...
INSERT INTO ItemTable VALUES ... ON CONFLICT (key) DO UPDATE SET value = excluded.value WHERE value != excluded.value
```

tmm1: That's fine but would require some kind of migration?

Unless you want to change the schema only for new installs?

bpasero: Yeah, for existing DBs it would not make a difference and I would not do any migration, but keep in mind that we create a new DB even for existing users for every new workspace that is opened for the first time to store workspace state, so it does apply often, even for recurring users.

tmm1: Gotcha, I understand now. Updated PR as discussed.

bpasero (approved):
Actually, lets keep the `CREATE` statement as it was, my perf testing does not show any improvements changing that and I feel we should not have a mix of old and new schema here. Besides, the `CREATE` statement as it was seems better to explain the intent.

I also played around with setting the `key` to `PRIMARY KEY` but that does not seem to yield significant improvements.

Thanks for reporting and fixing this. This kind of contribution requires some good understanding of our code 👏

lszomoru (approved):

PR #262150, opened 2025-08-18
Avoid recreating deriveds

### LINKED ISSUE #262136: Avoid recreating deriveds
Deriveds themselves are already lazy (the compute function is only called on demand). If each consumer gets their own derived, they will not reuse a shared cache.

https://github.com/microsoft/vscode/blob/8fe15b993cd4cfb16f1c3fbd62fd3030b7dafae4/src/vs/workbench/contrib/chat/browser/chatSelectedTools.ts#L80

Instead, do this:
```
readonly entriesMap = derived(r => { ...
```

aeschli:
Fixes https://github.com/microsoft/vscode/issues/262136


bpasero (approved):

bpasero:
@aeschli Red CI

PR #262226, opened 2025-08-18
Add term auto approve enablement setting and policy

Tyriar:
Part of #255789

Keeping the issue open as there's some clean up that needs to happen here. This needs to go in today though.

cc @joshspicer


DonJayamanne (approved):

watkinslisa435:
Pull

PR #262234, opened 2025-08-18
Don't skip sanitize for system extensions readmes

mjbvz:
I'd really like all of our markdown rendering to have some sanitization. Originally sanitization was skipped for #211920. Instead of disabling all sanitization, I'd like to loosen just the one rule we seem to need

I tested the release notes link after this change and everything seems to work


mjbvz:
Going to try merging this to get testing. @sandy081 Let me know if you notice any issues. Hopefully we can just relax a few options as needed instead of disabling sanitization again

Tyriar (approved):

PR #262239, opened 2025-08-19
handle thinking part serialization

### LINKED ISSUE #261298: Restored session with thinking tokens doesn't render correctly
About https://github.com/microsoft/vscode/issues/257104

1) enable `chat.agent.showThinking`
2) run a session with a model that supports it
3) started a new session
4) restore session from step 2
-> Thinking token are rendered as regular response tokens prefixed with `<|im_sep|>****`

Image

justschen:
fix https://github.com/microsoft/vscode/issues/261298

### DIFF
--- a/src/vs/workbench/contrib/chat/common/annotations.ts
+++ b/src/vs/workbench/contrib/chat/common/annotations.ts
@@ -17,6 +17,9 @@ export function annotateSpecialMarkdownContent(response: Iterable<IChatProgressR
 
 	const result: IChatProgressRenderableResponseContent[] = [];
 	for (const item of response) {
+		if (item.kind === 'thinking' && !item.value) {
+			continue;
+		}
 		const previousItemIndex = findLastIdx(result, p => p.kind !== 'textEditGroup' && p.kind !== 'undoStop');
 		const previousItem = result[previousItemIndex];
 		if (item.kind === 'inlineReference') {
--- a/src/vs/workbench/contrib/chat/common/annotations.ts
+++ b/src/vs/workbench/contrib/chat/common/annotations.ts
@@ -17,9 +17,6 @@ export function annotateSpecialMarkdownContent(response: Iterable<IChatProgressR
 
 	const result: IChatProgressRenderableResponseContent[] = [];
 	for (const item of response) {
-		if (item.kind === 'thinking' && !item.value) {
-			continue;
-		}
 		const previousItemIndex = findLastIdx(result, p => p.kind !== 'textEditGroup' && p.kind !== 'undoStop');
 		const previousItem = result[previousItemIndex];
 		if (item.kind === 'inlineReference') {
--- a/src/vs/workbench/contrib/chat/common/chatModel.ts
+++ b/src/vs/workbench/contrib/chat/common/chatModel.ts
@@ -316,8 +316,9 @@ class AbstractResponse implements IResponse {
 	 */
 	protected _markdownContent = '';
 
+	// Ignore empty thinking parts.
 	get value(): IChatProgressResponseContent[] {
-		return this._responseParts;
+		return this._responseParts.filter(p => !(p.kind === 'thinking' && !p.value));
 	}
 
 	constructor(value: IChatProgressResponseContent[]) {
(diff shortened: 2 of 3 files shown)


anthonykim1 (dismissed):

src/vs/workbench/contrib/chat/common/chatModel.ts
```diff
@@ -1401,7 +1403,15 @@ export class ChatModel extends Disposable implements IChatModel {
 						// eslint-disable-next-line local/code-no-dangerous-type-assertions
 						{ errorDetails: raw.responseErrorDetails } as IChatAgentResult : raw.result;
 					request.response = new ChatResponseModel({
-						responseContent: raw.response ?? [new MarkdownString(raw.response)],
+						responseContent: raw.response ? raw.response.map(part => {
+							if (isMarkdownString(part)) {
+								return part;
+							} else if ('kind' in part && part.kind === 'thinking') {
+								return { kind: 'thinking', value: part.value, id: part.id, metadata: part.metadata };
```
roblourens: Why do you have to do this, just for thinking parts?

src/vs/workbench/contrib/chat/common/chatModel.ts
```diff
@@ -463,10 +463,12 @@ export class Response extends AbstractResponse implements IDisposable {
 	private _citations: IChatCodeCitation[] = [];
 
 
-	constructor(value: IMarkdownString | ReadonlyArray<IMarkdownString | IChatResponseProgressFileTreeData | IChatContentInlineReference | IChatAgentMarkdownContentWithVulnerability | IChatResponseCodeblockUriPart>) {
-		super(asArray(value).map((v) => (isMarkdownString(v) ?
-			{ content: v, kind: 'markdownContent' } satisfies IChatMarkdownContent :
-			'kind' in v ? v : { kind: 'treeData', treeData: v })));
+	constructor(value: IMarkdownString | ReadonlyArray<IMarkdownString | IChatResponseProgressFileTreeData | IChatContentInlineReference | IChatAgentMarkdownContentWithVulnerability | IChatResponseCodeblockUriPart | IChatThinkingPart>) {
+		super(asArray(value).map((v) => (
```
roblourens, line 467: This code is so sad, I don't know how we ended up storing `IMarkdownString` instead of markdown content parts, I should fix it up some time

src/vs/workbench/contrib/chat/common/annotations.ts
```diff
@@ -17,6 +17,9 @@ export function annotateSpecialMarkdownContent(response: Iterable<IChatProgressR
 
 	const result: IChatProgressRenderableResponseContent[] = [];
 	for (const item of response) {
+		if (item.kind === 'thinking' && !item.value) {
```
roblourens: I thought we weren't sending empty thinking parts anymore? Is this still an issue?

justschen: we fixed the issue where it was getting sent all the time, but the main thing is that it is possible to send something like this ⬇️  once a thinking part with text is sent

```
{text: '', id: 'cot_id_2736427342', 'metadata': 'asldjfhasldf'}
```

what i think is happening is a thinking part with text being sent, and then a corrsponding ID being sent with any additional metadata, but no text

cc @karthiknadig

src/vs/workbench/contrib/chat/common/chatModel.ts
justschen: 😢

src/vs/workbench/contrib/chat/common/annotations.ts
karthiknadig: Yes, this is a possibility. You could get a delta that can contain empty text with id.

karthiknadig (dismissed):

src/vs/workbench/contrib/chat/common/annotations.ts
roblourens: It's fine, but I think this function isn't really the right place to decide how a thinking part is rendered. That decision seems unrelated to `annotateSpecialMarkdownContent`.

bpasero (dismissed):

karthiknadig (approved):

PR #262326, opened 2025-08-19
Refactor runInTerminalTool to simplify prepare call

Tyriar:


lramos15 (approved):

danowar2:
That's the place in the center

PR #262348, opened 2025-08-19
Prevent sticky scroll flash for pagers

anthonykimseattle:


Tyriar (commented):
@anthonykim1 I don't see an issue linked, but I think the problem could be reproduced via `git log`? I can't reproduce anymore, did we fix this some other way?

PR #262414, opened 2025-08-20
 Git: Fix worktree checkout default actions

### LINKED ISSUE #262143: Git: Fix worktree checkout default actions
1. Create a new git repo
2. Create a new worktree
3. Try to checkout main in the new worktree

You are greeted with this warning:

Image

These two actions are no-ops, since the original worktree is already open as a workspace in the current window

nguyenchristy:
Fixes https://github.com/microsoft/vscode/issues/262143

When a user attempts to checkout a branch in a worktree, already in use by main repository, they are presented with an option to create a new branch off of the head of the main repository or cancel.

### DIFF
--- a/extensions/git/src/commands.ts
+++ b/extensions/git/src/commands.ts
@@ -2944,38 +2944,29 @@ export class CommandCenter {
 					}
 
 					const mainRepository = this.model.getRepository(path.dirname(repository.dotGit.commonPath));
-					if (mainRepository && item.refName && item.refName === mainRepository.HEAD?.name) {
+					if (mainRepository && item.refName && item.refName.replace(`${item.refRemote}/`, '') === mainRepository.HEAD?.name) {
 						const message = l10n.t('Branch "{0}" is already checked out in the current repository.', item.refName);
-						const createBranch = l10n.t('Create New Branch');
-						const pick = await window.showWarningMessage(message, { modal: true }, createBranch);
-
-						if (pick === createBranch) {
-							await this._branch(repository, undefined, false, mainRepository.HEAD?.name);
-						} else {
-							return false;
-						}
-					} else {
-						this.handleWorktreeBranchAlreadyUsed(err);
+						await window.showErrorMessage(message, { modal: true });
 						return false;
 					}
+					this.handleWorktreeBranchAlreadyUsed(err);
+					return false;
 				}
 
-				if (err.gitErrorCode === GitErrorCodes.DirtyWorkTree) {
-					const stash = l10n.t('Stash & Checkout');
-					const migrate = l10n.t('Migrate Changes');
-					const force = l10n.t('Force Checkout');
-					const choice = await window.showWarningMessage(l10n.t('Your local changes would be overwritten by checkout.'), { modal: true }, stash, migrate, force);
+				const stash = l10n.t('Stash & Checkout');
+				const migrate = l10n.t('Migrate Changes');
+				const force = l10n.t('Force Checkout');
+				const choice = await window.showWarningMessage(l10n.t('Your local changes would be overwritten by checkout.'), { modal: true }, stash, migrate, force);
 
-					if (choice === force) {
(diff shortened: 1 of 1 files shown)


lszomoru:
@nguyenchristy, sorry for not getting back to you on this until now. In this particualar case the flow that triggers the dialog is switching branches. I do not believe that as part of switching branches we should guide users to create a new worktree since all they wanted to do is switch a branch. 

I believe that we should check if the branch that the user is trying to switch to is checked out in an existing worktree. If it is, check if the worktree is the main repository, we should only show the "Cancel" action. Otherwise we should show the current dialog with the tree actions. Does that make sense?

nguyenchristy:
> @nguyenchristy, sorry for not getting back to you on this until now. In this particualar case the flow that triggers the dialog is switching branches. I do not believe that as part of switching branches we should guide users to create a new worktree since all they wanted to do is switch a branch.
> 
> I believe that we should check if the branch that the user is trying to switch to is checked out in an existing worktree. If it is, check if the worktree is the main repository, we should only show the "Cancel" action. Otherwise we should show the current dialog with the tree actions. Does that make sense?

Yeah, I made it so that they could create a branch (not a worktree) off of the branch they initially wanted to check out if the worktree is the main repository, but I agree that just the 'Cancel' action is simpler and less confusing. Also, I considered the remote branch scenario as well in this updated PR.

extensions/git/src/commands.ts
```diff
@@ -2938,7 +2938,18 @@ export class CommandCenter {
 				}
 
 				if (err.gitErrorCode === GitErrorCodes.WorktreeBranchAlreadyUsed) {
-					this.handleWorktreeBranchAlreadyUsed(err);
+					if (!repository.dotGit.commonPath) {
+						await this.handleWorktreeBranchAlreadyUsed(err);
+						return false;
+					}
+
+					const mainRepository = this.model.getRepository(path.dirname(repository.dotGit.commonPath));
```
lszomoru: Given that the actions in the "normal" dialog refer to "window" (ex: Open Worktree in a New Window, Open Worktree in the Current Window) I think that here we should be doing the same. We should check whether the "commonPath" matches any of the workspace folders in the current workspace and if that is the case, we should say "Branch ___ is already checked out in the current window". Does that make sense?

lszomoru (changes requested):

extensions/git/src/commands.ts
nguyenchristy: I'm not completely understanding. Do you mean something like: 
`const worktree = path.dirname(repository.dotGit.commonPath);`
`const isInCurrentWindow = workspace.workspaceFolders.some(folder => pathEquals(folder.uri.fsPath, worktree));` 

`repository.dotGit.commonPath` returns something like `C:\\Users\\t-nguyenchr\\vscode\\.git` (which is why I did path.dirname). It does not give me info about the specific branch a worktree is on, so I don't completely understand why I am seeing if the worktree's 'commonPath' matches with a folder (eg. vscode). Also, I'm not seeing that a worktree is considered a workspace folder. 

The main issue was to check if the branch we want to check out in a worktree is already used by the main repository's currently checked out branch. We previously have already implemented this check: 'if the branch that the user is trying to switch to is checked out in an existing worktree' which is why I just added `item.refName = mainRepoHead`.

lszomoru: Correct. That is what I meant. 

Let's take a step back and look at the two actions that are in the usual dialog: "Open in New Window", "Open in Current Window". The bug that was filed was about these actions being no-ops when one picks the branch in the workspace that is current opened. This is the condition that we need to detect and not show the actions.

nguyenchristy: Ok, I think I understand now. I was missing the **extra** check of the commonPath matching the workspace folders. I updated it so we check for that + if the selected branch matches the workspace's currently checked out branch.

lszomoru (approved):

PR #262420, opened 2025-08-20
support model switching in automation

rwoll:
Programmatically allow setting models via `workbench.action.chat.open` for automation scenarios.

`setCurrentLanguageModel` implementation visibility is changed from `private` to `public`, but it's implementation remains unchanged. Re-located to be alongside the other `public` methods.

Manually tested the command:

Screenshot 2025-08-20 at 10 13 56 AM

### DIFF
--- a/src/vs/workbench/contrib/chat/browser/actions/chatActions.ts
+++ b/src/vs/workbench/contrib/chat/browser/actions/chatActions.ts
@@ -72,9 +72,8 @@ import { VIEWLET_ID } from '../chatSessions.js';
 import { ChatViewPane } from '../chatViewPane.js';
 import { convertBufferToScreenshotVariable } from '../contrib/screenshot.js';
 import { clearChatEditor } from './chatClear.js';
-import { ILanguageModelsService } from '../../common/languageModels.js';
+import { ILanguageModelChatSelector, ILanguageModelsService } from '../../common/languageModels.js';
 import { IChatResponseModel } from '../../common/chatModel.js';
-import { LanguageModelChatSelector } from 'vscode';
 
 export const CHAT_CATEGORY = localize2('chat.category', 'Chat');
 
@@ -127,7 +126,7 @@ export interface IChatViewOpenOptions {
 	 * }
 	 * ```
 	 */
-	modelSelector?: LanguageModelChatSelector;
+	modelSelector?: ILanguageModelChatSelector;
 
 	/**
 	 * Wait to resolve the command until the chat response reaches a terminal state (complete, error, or pending user confirmation, etc.).
--- a/src/vs/workbench/contrib/chat/browser/actions/chatActions.ts
+++ b/src/vs/workbench/contrib/chat/browser/actions/chatActions.ts
@@ -74,6 +74,7 @@ import { convertBufferToScreenshotVariable } from '../contrib/screenshot.js';
 import { clearChatEditor } from './chatClear.js';
 import { ILanguageModelsService } from '../../common/languageModels.js';
 import { IChatResponseModel } from '../../common/chatModel.js';
+import { LanguageModelChatSelector } from 'vscode';
 
 export const CHAT_CATEGORY = localize2('chat.category', 'Chat');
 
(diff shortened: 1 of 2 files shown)


src/vs/workbench/contrib/chat/browser/actions/chatActions.ts
```diff
@@ -166,6 +173,17 @@ abstract class OpenChatGlobalAction extends Action2 {
 			await this.handleSwitchToMode(switchToMode, chatWidget, instaService, commandService);
 		}
 
+		if (opts?.modelId) {
+			await languageModelService.selectLanguageModels({}, false);
```
roblourens: You have to do this before calling `lookupLanguageModel`? cc @lramos15 for what the right usage is.

roblourens: It seems like if we want to wait for it, then `select` is what we should use, maybe we should just take input in the form of a selector object? Or add a method to let us get a model by identifier but also wait for provider activation.

roblourens (dismissed):

src/vs/workbench/contrib/chat/browser/actions/chatActions.ts
connor4312: `selectLanguageModels` does the activation of relevant extensions, `lookupLanguageModel` does not. I would have the command take the selector and use the first model ID returned from this method as the one to look up in the next line.

connor4312 (dismissed):

src/vs/workbench/contrib/chat/browser/actions/chatActions.ts
rwoll: Updated! Thank you for the suggestion!

connor4312 (dismissed):

src/vs/workbench/contrib/chat/browser/actions/chatActions.ts
roblourens: Can we actually have the command take a `ILanguageModelChatSelector`, so that we can include the vendor? Because `selectLanguageModels` will activate extensions unnecessarily if we call it without a vendor. (this probably doesn't matter for the test scenario but I've told many people about this command, it will be in the product, I'd like it to follow the best practice)

rwoll: Updating momentarily. I also think we should throw if the the select is ambiguous (i.e. more than one match returned). Taking the first seems arb. and unpreditable.

mjbvz (dismissed):

rwoll:
Converting to draft while I sort out a hygiene error:

```
[10:42:43] /Users/rosswollman/code/vsc/vscode/src/vs/workbench/contrib/chat/browser/actions/chatActions.ts: line 77, col 43, Warning - Imports violates 'vs/base/{common,browser}/** or vs/base/parts/*/{common,browser}/** or vs/platform/*/{common,browser}/** or vs/editor/{common,browser}/** or vs/editor/contrib/*/{common,browser}/** or vs/workbench/{common,browser}/** or vs/workbench/services/*/{common,browser}/** or vs/workbench/contrib/*/{common,browser}/** or vs/workbench/contrib/terminal/terminalContribChatExports* or vs/workbench/contrib/terminal/terminalContribExports* or vscode-notebook-renderer or @vscode/tree-sitter-wasm or @xterm/xterm or @xterm/addon-* or vscode-textmate or vs/nls.js or vs/amdX.js' restrictions. See https://github.com/microsoft/vscode/wiki/Source-Code-Organization (local/code-import-patterns)
```

connor4312 (dismissed):

src/vs/workbench/contrib/chat/browser/actions/chatActions.ts
roblourens: I think the selector was meant to be flexible, you can use it to say "use some claude model" or "use some built-in (vendor: copilot) model" if you don't care about being more specific. I'd say that taking the first match is ok

rzhao271 (dismissed):

src/vs/workbench/contrib/chat/browser/actions/chatActions.ts
rwoll: Updated! I was too focused on the repro'able automation case and not on the general utility of this command for others.

In the future, we could consider adding a `strict: true` flag in the selector to enforce the non-ambiguous behavior.

connor4312 (approved):

rzhao271 (approved):

Issue #251187, opened 2025-06-11
Re-think the Chat view welcome


isidorn:
When a user opens a new Chat view in agent mode they see the following:

![Image](https://github.com/user-attachments/assets/691e6228-f49e-49ea-b7be-1614f7591570)

I think this view is a testiment to how we evolved the feature (it was first chat then agent). We need to re-think this as the future will be agentic first and only. Some comments on the current look:

1) There is no need for the Copilot Frog icon. I think we can survive without any icon at all. AI is key to the experience, no need to have a special simbol for it
2) "Edit with Copilot" is completely not important. It is agent, not "copilot". We should not push the Copilot brand
3) Explanation paragraph - hopefully not needed, users just start and figure it out. It's not like we have an explanation in the editor how to use it
4) "Powered by AI mistakes are possible" - I can discuss with CELA / Responsible AI folk that we put this somewhere else (make it smaller)
5) "Internal user..." only internal users see this, not important
6) @ and # are not important in agent mode. Agent figures out all by itself, and we have "@ Add Context" affordance already for this

Thus looking at the current content, I can not say any of it is useful.

Zed shows and empty view and at the bottom recent chat sessions.

![Image](https://github.com/user-attachments/assets/f3f3b7c2-6706-492a-8798-460007888819)

Cursor similar, but they start with the input at the top, recent sessions at bottom, and then once chat starts they move input to the bottom.
Windsurf similar, but they show the input in middle, and recent sessions at bottom, once chat starts the move input to the bottom.

![Image](https://github.com/user-attachments/assets/60cd1d08-8ef9-4a34-98b8-3cd245e76213)

ChatGPT mac app just shows an empty space (lol, but I like it).

How can we improve the Agent welcome experience? Industry seems to be aligning on the pattern it seems, should we follow?

fyi @roblourens @kkbrooks @bpasero @pierceboggan

kkbrooks:
Notion's AI welcome state:

![Image](https://github.com/user-attachments/assets/343c06c1-d34a-4b7d-b112-e43f6aedc879)

jo-oikawa:
BTW @eli-w-king and I spoke about this yesterday, Eli is going to drive this as an extension of the welcome UX work to make sure it fits into a cohesive story!

justschen:
ref https://github.com/microsoft/vscode/issues/253864 - a nice ui flow would be having previous conversations in the empty state? or making that `Show Chats..` button more discoverable.

isidorn:
[...]

PR #262430, opened 2025-08-20
addressed review comments

eli-w-king:
Added chat history to empty state under setting `chat.emptyState.history.enabled`. Addresses https://github.com/microsoft/vscode/issues/251187

Screenshot 2025-08-19 at 9 27 00 PM


src/vs/workbench/contrib/chat/browser/media/chatViewWelcome.css
```diff
@@ -29,6 +29,45 @@
 	overflow: hidden;
 }
 
+/* When recent history is shown, stack it above the welcome view */
+.interactive-session .chat-welcome-view-container:has(.chat-welcome-history-root) {
```
justschen: you can nest css like this for readability:


```css
.interactive-session .chat-welcome-view-container:has(.chat-welcome-history-root) {
   
    flex-direction: column;
	justify-content: flex-start;

    .chat-welcome-view > .chat-welcome-view-title {
           display: none;
    }

    .....

}
```

src/vs/workbench/contrib/chat/browser/media/chatViewWelcome.css
```diff
@@ -29,6 +29,45 @@
...
+}
+
+/* Hide the welcome section title when history is displayed */
+.interactive-session .chat-welcome-view-container:has(.chat-welcome-history-root) div.chat-welcome-view > .chat-welcome-view-title {
+	display: none;
+}
+
+/* Vertically center the icon without changing its size */
+.interactive-session .chat-welcome-view-container:has(.chat-welcome-history-root) .chat-welcome-view-icon .codicon {
+	line-height: 48px; /* matches icon container min-height for vertical centering */
```
justschen: i think we lowkey don't need comments like this (ik copilot likes adding them, but we probably don't need anything this detailed, unless there is something really weird we are adding)

src/vs/workbench/contrib/chat/browser/media/chatViewWelcome.css
```diff
@@ -29,6 +29,45 @@
...
+}
+
+/* Vertically center the entire welcome view within the container (history remains at top) */
+.interactive-session .chat-welcome-view-container:has(.chat-welcome-history-root) div.chat-welcome-view {
```
justschen: is there a non-div `chat-welcome-view` element? otherwise, can just do `.chat-welcome-view` instead

src/vs/workbench/contrib/chat/browser/chatWidget.ts
```diff
@@ -818,6 +827,96 @@ export class ChatWidget extends Disposable implements IChatWidget {
 		}
 	}
 
+	private async renderWelcomeHistorySection(): Promise<void> {
+		try {
+			const historyRoot = dom.append(this.welcomeMessageContainer, $('.chat-welcome-history-root'));
```
justschen, line 945: can we maybe toggle `chat-welcome-view-container` here with an additional class when we want to render history so we don't have to do css like `.chat-welcome-view-container:has(.chat-welcome-history-root)`? this way we don't have to do `has` everywhere.

src/vs/workbench/contrib/chat/browser/chatWidget.ts
```diff
@@ -818,6 +827,96 @@ export class ChatWidget extends Disposable implements IChatWidget {
...
+						await this.openHistorySession(item.sessionId);
+					}
+				}));
+			}
+
+			// Show more link
+			const more = dom.append(container, $('.chat-welcome-history-more'));
+			const moreLink = dom.append(more, $('a')) as HTMLAnchorElement;
+			moreLink.textContent = localize('chat.history.showMoreButton', 'Show chat history...');
+			moreLink.href = '#';
```
justschen: don't think you need the `href` set specifically here, but not 100% sure

src/vs/workbench/contrib/chat/browser/chatWidget.ts
```diff
@@ -818,6 +827,96 @@ export class ChatWidget extends Disposable implements IChatWidget {
...
+		} catch (err) {
+			this.logService.error('Failed to render welcome history', err);
+		}
+	}
+
+	private async openHistorySession(sessionId: string): Promise<void> {
+		try {
+			const viewsService = this.instantiationService.invokeFunction(accessor => accessor.get(IViewsService));
+			const chatView = await viewsService.openView<any>(ChatViewId);
+			await (chatView as any)?.loadSession?.(sessionId);
```
justschen:
```suggestion
			const viewsService = this.instantiationService.invokeFunction(accessor => accessor.get(IViewsService));
			const chatView = await viewsService.openView<ChatViewPane>(ChatViewId);
			await chatView?.loadSession?.(sessionId);
```

src/vs/workbench/contrib/chat/browser/chatWidget.ts
```diff
@@ -818,6 +827,96 @@ export class ChatWidget extends Disposable implements IChatWidget {
...
+			this._register(dom.addDisposableListener(moreLink, dom.EventType.CLICK, async (e: MouseEvent) => {
+				e.preventDefault();
+				e.stopPropagation();
+				// Defer to let the click settling finish before showing quick pick
+				setTimeout(() => {
+					this.instantiationService.invokeFunction(accessor => accessor.get(ICommandService).executeCommand('workbench.action.chat.history'));
+				}, 0);
+			}));
+			this._register(dom.addDisposableListener(moreLink, dom.EventType.KEY_DOWN, async (e: KeyboardEvent) => {
+				if (e.key === 'Enter' || e.key === ' ') {
```
justschen: i think we've transitioned to using `dom.addStandardDisposableListener` and keycodes instead (see `terminalTabLists.ts` as an example), which is a little cleaner since you can do `e.equals(KeyCode.Space)`

src/vs/workbench/contrib/chat/browser/media/chatViewWelcome.css
```diff
@@ -184,16 +249,84 @@ div.chat-welcome-view {
...
+	padding-left: 12px;
+}
+
+.chat-welcome-history-more a {
+	color: var(--vscode-textLink-foreground);
+	cursor: pointer;
+}
+
+/* Hide the "Show chat history..." link when the history list is empty */
+.chat-welcome-history-root:not(:has(.chat-welcome-history-item)) .chat-welcome-history-more {
```
justschen: same thing here, if there is a good way to toggle `.chat-welcome-history-root` when we know it's hidden, we won't have to do `has`

src/vs/workbench/contrib/chat/browser/chatWidget.ts
```diff
@@ -801,6 +804,12 @@ export class ChatWidget extends Disposable implements IChatWidget {
 				welcomeContent = this.getWelcomeViewContent(additionalMessage);
 				welcomeContent.tips = tips;
 			}
+			// Optional: recent chat history above welcome content when enabled
+			const showHistory = this.configurationService.getValue<boolean>(ChatConfiguration.EmptyStateHistoryEnabled) === true;
```
justschen: don't need the `=== true` since it is a `boolean` already

Agingvollkkaa99 (commented):
Leave a review

src/vs/workbench/contrib/chat/browser/chatWidget.ts
```diff
@@ -818,6 +827,96 @@ export class ChatWidget extends Disposable implements IChatWidget {
...
+					if (e.key === 'Enter' || e.key === ' ') {
+						e.preventDefault();
+						await this.openHistorySession(item.sessionId);
+					}
+				}));
+			}
+
+			// Show more link
+			const more = dom.append(container, $('.chat-welcome-history-more'));
+			const moreLink = dom.append(more, $('a')) as HTMLAnchorElement;
```
justschen: i think the the show more button is missing a tab index as well, i wasn't able to tab into it!

bpasero:
Are we still working on the UX? Somehow it is not immediately clear to me that the top thing is about previous chats.

src/vs/workbench/contrib/chat/browser/media/chatViewWelcome.css
```diff
@@ -175,7 +240,7 @@ div.chat-welcome-view {
 			width: fit-content;
 			margin: 0 4px;
 
-			> .chat-welcome-view-suggested-prompt-icon {
+			& > .chat-welcome-view-suggested-prompt-icon {
```
justschen, line 217: im not 100% sure if we need the & here or not, or if it adds some specificity we think we need?

justschen (changes requested):

justschen:
> Are we still working on the UX? Somehow it is not immediately clear to me that the top thing is about previous chats.

@bpasero i think this was what landed, but can maybe discuss some more in ux sync if we think it's not clear enough

bpasero:
Yes please.

src/vs/workbench/contrib/chat/browser/chatWidget.ts
```diff
@@ -818,6 +843,119 @@ export class ChatWidget extends Disposable implements IChatWidget {
...
+			headerTitle.textContent = localize('chat.history.title', 'Chat History');
+			const headerActions = dom.append(header, $('.chat-welcome-history-header-actions'));
+
+			const items = await this.chatService.getHistory();
+			const filtered = items
+				.filter(i => !i.isActive)
+				.sort((a, b) => (b.lastMessageDate ?? 0) - (a.lastMessageDate ?? 0))
+				.slice(0, 3);
+
+			const showAllButton = dom.append(headerActions, $('.chat-welcome-history-show-all')) as HTMLElement;
```
justschen: don't think you need the `as HTMLElement` here. in TS, it is already the inferred type here i believe.

src/vs/workbench/contrib/chat/browser/chatWidget.ts
```diff
@@ -818,6 +843,119 @@ export class ChatWidget extends Disposable implements IChatWidget {
...
+			}
+
+			// Compute today's midnight once for label decisions
+			const todayMidnight = new Date();
+			todayMidnight.setHours(0, 0, 0, 0);
+			const todayMidnightMs = todayMidnight.getTime();
+
+			for (const item of filtered) {
+				const row = dom.append(listEl, $('.chat-welcome-history-item'));
+				const title = dom.append(row, $('.chat-welcome-history-title'));
```
justschen: i'm not a big fan of this. this is kind of hacky since we're manually appending things with dom

could we try using the list service/widget or workbench lists? we'd get a lot of styling for free that way as well

justschen (changes requested):

src/vs/workbench/contrib/chat/browser/chatWidget.ts
eli-w-king: removed === true

src/vs/workbench/contrib/chat/browser/chatWidget.ts
eli-w-king: removed

PR #262436, opened 2025-08-20
Fix terminal sticky scroll flash when pager 

### LINKED ISSUE #222075: Terminal sticky scroll can show up for 1 frame when using page down in a pager
Repro:

1. On Windows
2. Enable sticky scroll
3. Open pwsh
4. Run `git diff` with more than a page of changes
5. Page up+down, 🐛 sticky scroll will show up for 1 frame every time page down is used

It rarely shows up in the gif due to low recording framerate

Recording 2024-07-18 at 04 46 30

anthonykim1:
Resolves: https://github.com/microsoft/vscode/issues/222075

### DIFF
--- a/src/vs/workbench/contrib/terminalContrib/stickyScroll/browser/terminalStickyScrollOverlay.ts
+++ b/src/vs/workbench/contrib/terminalContrib/stickyScroll/browser/terminalStickyScrollOverlay.ts
@@ -8,7 +8,7 @@ import type { WebglAddon as WebglAddonType } from '@xterm/addon-webgl';
 import type { LigaturesAddon as LigaturesAddonType } from '@xterm/addon-ligatures';
 import type { IBufferLine, IMarker, ITerminalOptions, ITheme, Terminal as RawXtermTerminal, Terminal as XTermTerminal } from '@xterm/xterm';
 import { $, addDisposableListener, addStandardDisposableListener, getWindow } from '../../../../../base/browser/dom.js';
-import { throttle } from '../../../../../base/common/decorators.js';
+import { debounce, throttle } from '../../../../../base/common/decorators.js';
 import { Event } from '../../../../../base/common/event.js';
 import { Disposable, MutableDisposable, combinedDisposable, toDisposable } from '../../../../../base/common/lifecycle.js';
 import { removeAnsiEscapeCodes } from '../../../../../base/common/strings.js';
@@ -59,7 +59,6 @@ export class TerminalStickyScrollOverlay extends Disposable {
 	private _contextMenu: IMenu;
 
 	private readonly _refreshListeners = this._register(new MutableDisposable());
-	private readonly _refreshTimeout = this._register(new MutableDisposable());
 
 	private _state: OverlayState = OverlayState.Off;
 	private _isRefreshQueued = false;
@@ -186,6 +185,7 @@ export class TerminalStickyScrollOverlay extends Disposable {
 		this._refreshListeners.clear();
 	}
 
+	@debounce(100)
 	private _setVisible(isVisible: boolean) {
 		if (isVisible) {
 			this._ensureElement();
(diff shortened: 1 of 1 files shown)


src/vs/workbench/contrib/terminalContrib/stickyScroll/browser/terminalStickyScrollOverlay.ts
```diff
@@ -193,6 +194,19 @@ export class TerminalStickyScrollOverlay extends Disposable {
 	}
 
 	private _refresh(): void {
+		this._refreshTimeout.clear();
+
+		// Always debounce by 100ms to prevent flashing during rapid scroll/enter events
```
anthonykim1: I've tried wrapping this timeout conditionally for only pager related scenarios, using https://github.com/microsoft/vscode/blob/a3208e218c788bc19d9c915c4612d37ae2650096/src/vs/workbench/contrib/terminalContrib/stickyScroll/browser/terminalStickyScrollOverlay.ts#L280-L284 , but found out the current code (Pre-PR) doesn't always detect we are in pager as I enter through the `git diff`.

Hence made this debounce generic(?) here. Curious what you think @Tyriar

src/vs/workbench/contrib/terminalContrib/stickyScroll/browser/terminalStickyScrollOverlay.ts
```diff
@@ -193,6 +194,19 @@ export class TerminalStickyScrollOverlay extends Disposable {
...
+		const timeoutId = window.setTimeout(() => {
+			this._executeRefresh();
+		}, 100);
+
+		this._refreshTimeout.value = toDisposable(() => {
```
anthonykim1: Is this how I properly use disposable for timeout?

src/vs/workbench/contrib/terminalContrib/stickyScroll/browser/terminalStickyScrollOverlay.ts
Tyriar: `_refresh` actually already has some debouncing but it's at the microtask level:

https://github.com/microsoft/vscode/blob/fb3ad8dfeb2118c3dbf5154c89be35cdf8cb9e58/src/vs/workbench/contrib/terminalContrib/stickyScroll/browser/terminalStickyScrollOverlay.ts#L195-L204

You want to debounce the hide -> show transition specifically. And for that you should use the `@debounce` helper

Tyriar (changes requested):

src/vs/workbench/contrib/terminalContrib/stickyScroll/browser/terminalStickyScrollOverlay.ts
anthonykim1: Sounds good! @Tyriar 
I think `_setVisible` might be the one we want.
And it came out to be one liner.. hahah

src/vs/workbench/contrib/terminalContrib/stickyScroll/browser/terminalStickyScrollOverlay.ts
```diff
@@ -185,6 +185,7 @@ export class TerminalStickyScrollOverlay extends Disposable {
 		this._refreshListeners.clear();
 	}
 
+	@debounce(100)
 	private _setVisible(isVisible: boolean) {
 		if (isVisible) {
 			this._ensureElement();
```
Tyriar: Almost, I think we need to be more specific here. We want to make sure the debounce only applies going from hidden -> visible. Visible -> hidden should be immediate. So I think we need to split this into 2 functions. Not doing this means stale would stick around a little longer, that's not the end of the world but an unnecessary 100ms delay can make apps feel sluggish.

Tyriar (changes requested):

src/vs/workbench/contrib/terminalContrib/stickyScroll/browser/terminalStickyScrollOverlay.ts
anthonykim1: You're right. Hiding should be immediate, and we should only debounce when showing to prevent flash.
What do you think of https://github.com/microsoft/vscode/pull/262436/commits/718442ddc4896762856616ff78347488258b0e8e ?

Tyriar (approved):

PR #262462, opened 2025-08-20
Treat ellipsis character as search wildcard

Skn0tt:
Some tools, like the Playwright test runner CLI, ellipsize filenames with the `…` character:

 Screenshot 2025-08-20 at 10 57 03

At the moment, the ellipsized path isn't clickable because VS Code's fuzzy matching doesn't handle that character well. This PR adds `…` as an alternative wildcard. This makes both the terminal path matching and the file quick picker find the file.


Tyriar (approved):
Makes sense to me. FYI @jrieken

bpasero (approved):

bpasero:
Follow up simplification: https://github.com/microsoft/vscode/pull/265987

PR #262545, opened 2025-08-20
Update Beginner walkthrough

bhavyaus:


src/vs/workbench/contrib/welcomeGettingStarted/common/gettingStartedContent.ts
```diff
@@ -556,6 +475,16 @@ export const walkthroughs: GettingStartedWalkthroughContent = [
 		content: {
 			type: 'steps',
 			steps: [
+				{
+					id: 'settingsAndSync',
+					title: localize('gettingStarted.settings.title', "Tune your settings"),
+					description: localize('gettingStarted.settingsAndSync.description.interpolated', "Customize every aspect of VS Code and your extensions to your liking. [Back up and sync](command:workbench.userDataSync.actions.turnOn) your essential customizations across all your devices.\n{0}", Button(localize('tweakSettings', "Open Settings"), 'command:toSide:workbench.action.openSettings')),
```
cwebster-99: It would be great to shorten this description and include an entry point and call out for the `code` command (`workbench.action.installCommandLine`). 
```suggestion
					description: localize('gettingStarted.settingsAndSync.description.interpolated', "Customize every aspect of VS Code and [sync](command:workbench.userDataSync.actions.turnOn) customizations across devices.\n{0}", Button(localize('tweakSettings', "Open Settings"), 'command:toSide:workbench.action.openSettings')),
```

Could we show a pre-checked box with the description "Use `code` to launch VS Code from the command line". We may need to play around with the layout of the tile to do so.

DonJayamanne (approved):

PR #262546, opened 2025-08-20
debug: convert the renderer attacher to a standard websocket

connor4312:
Allows better interoperability with other tools. `rendererDebugPort`
is not `rendererDebugAddr` which is a `ws://` address.


sandy081 (approved):

src/vs/base/parts/ipc/node/ipc.net.ts
```diff
@@ -3,10 +3,13 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
+import * as crypto from 'crypto';
```
bpasero, line 6: @connor4312 you import `crypto` twice

PR #262558, opened 2025-08-20
Explore installing npm package when installing mcp server

sandy081:


src/vs/platform/mcp/node/npmPackageService.ts
```diff
@@ -0,0 +1,102 @@
...
+export class NpmPackageService implements INpmPackageManagementService {
+	readonly _serviceBrand: undefined;
+
+	constructor(
+		@ILogService private readonly logService: ILogService,
+	) { }
+
+	async getLatestPackageVersion(packageName: string): Promise<string> {
+		return new Promise<string>((resolve, reject) => {
+			const npmProcess = spawn('npm', ['view', packageName, 'version'], {
```
connor4312, line 28: This is unfortunately going to be a bit of pain for users who have nvm/fnm/etc, especially on Windows where we don't do shell resolution. It may be better to run this in a terminal rather than just a subprocess so that you get the full environment.

src/vs/platform/mcp/node/mcpManagementService.ts
```diff
@@ -21,10 +25,117 @@ export class McpUserResourceManagementService extends CommonMcpUserResourceManag
...
+						return this.uriIdentityService.extUri.joinPath(nodeModulesPath, binName as string);
+					}
+				}
+			}
+
+			if (packageJson.main) {
+				return this.uriIdentityService.extUri.joinPath(nodeModulesPath, packageJson.main);
+			}
+
+			return this.uriIdentityService.extUri.joinPath(nodeModulesPath, 'index.js');
```
connor4312, line 131: I suggest checking `packageJson.exports?.['.']` too for packages using export maps

connor4312 (approved):

PR #262638, opened 2025-08-21
Update instructions for builds

digitarald:
Noticed how much agent runs random tsx and typecheck commands (likely to fail), even while the tasks are running, either working and reporting obvious errors.

Starting to add more instructions, making agent better at telemetry events.


bpasero (changes requested):
This seems to duplicate what `.github/instructions/typescript.instructions.md` already does 🤔

bpasero (changes requested):
Why can we not have separate instructions per language, why must everything be inlined into the `copilot-instructions.md`? Do you have a scenario where the instructions were ignored?

digitarald:
With the fixes, tasks are a lot more reliably checked.

In a project with 95%+ TS having a separate file for TS is not adding much domain separation (cc @aeschli). As most vscode development workflows need benefit from those steps, having it in the main file makes more sense and will make them more reliable. The main instructions already have TS-related style and quality guidance, so this isn't a big shift.

bpasero (approved):
I see. I am not sure about this change in general because I think we have instruction files for a reason and it would be good for us to selfhost, but I do not have strong enough objections to block this.

PR #262648, opened 2025-08-21
Use context for line breaks computation 

aiday-mar:


src/vs/editor/common/viewModel/viewModelImpl.ts
```diff
@@ -305,143 +307,135 @@ export class ViewModel extends Disposable implements IViewModel {
...
+					case textModelEvents.RawContentChangedType.LinesInserted: {
+						for (let lineNumber = change.newFromLineNumber; lineNumber <= change.newToLineNumber; lineNumber++) {
+							lineBreaksComputer.addRequest(lineNumber, null);
 						}
+						break;
+					}
+					case textModelEvents.RawContentChangedType.LineChanged: {
+						lineBreaksComputer.addRequest(change.newLineNumber, null);
+						break;
 					}
```
alexdima, line 343: I don't believe this will work correctly. The principle of keeping the view model in sync with the text model via the data contained inside the event is violated. The problem is an execution order like the following:
* the model is in state s0 and mutates to state s1 and emits event e1
* the model is in state s1 and mutates to state s2 and emits event e2
* the view model receives event e1
When the view model receives event e1, the model is already in state s2, so it is incorrect for the view model to access the model while processing event e1.

A more concrete example:
* the text model contains the text `Hello\nworld` (s0)
* an exclamation mark is added at the end, the change event is line number 2, the new content is `Hello\nworld!` (s1)
* the newline is deleted, the change event is line number 2 deleted, the new content is `Helloworld!` (s2)
* the view model receives a change event for line number 2, when the view model attempts to read information about line number 2, like its content or its injected text, the model will throw because the model is already in state s2 `Helloworld!` and line number 2 doesn't exist.

alexdima (changes requested):

src/vs/editor/common/viewModel/viewModelImpl.ts
aiday-mar: Hi Alex. So the code you are referring is inside of the method `onDidChangeContentOrInjectedText` which is called inside of the `_emitContentChangedEvent` and inside of the `handleBeforeFireDecorationsChangedEvent` method. 

The `_emitContentChangedEvent` method is called from `_setValueFromTextBuffer`, `setEOL` and from `_doApplyEdits`, meaning it is directly called after the text buffer is changed without passing through a listener/event. In this case, the view model should be updated directly after the model change event. 

As for the `handleBeforeFireDecorationsChangedEvent` method, this method is called only when `_onDidChangeDecorations` is fired so `onDidChangeContentOrInjectedText` will be called in an asynchronous manner. For this case, I have rewritten the code so as to add a new method `onDidChangeInjectedText` which is called directly on the decorations change in the methods `_changeDecorationImpl` and `_deltaDecorationsImpl`. 

In the following screen recording you can see that the `onDidChangeContentOrInjectedText` is directly called after the `_doApplyEdits` method is called every single time, before the next time the latter method is called.

https://github.com/user-attachments/assets/92a71bfa-f464-468a-8ee0-2da1eb4ab805

src/vs/editor/browser/widget/diffEditor/components/diffEditorViewZones/diffEditorViewZones.ts
```diff
@@ -176,7 +176,7 @@ export class DiffEditorViewZones extends Disposable {
 							if (i > originalModel.getLineCount()) {
 								return { orig: origViewZones, mod: modViewZones };
 							}
-							deletedCodeLineBreaksComputer?.addRequest(originalModel.getLineContent(i), null, null);
+							deletedCodeLineBreaksComputer?.addRequest(i, null);
```
alexdima, line 187: This needs to be created with a different context, pointing to the original lines

alexdima (changes requested):

src/vs/editor/common/viewModel/viewModelImpl.ts
alexdima: :+1: With the order this is correct! 🎉

aiday-mar:
So I made the two changes we discussed:

1. Adopt the function `withEventCapturing` in the file `model.test.ts`
2. Adapt the method `createLineBreaksComputer` to the diffEditorViewZones.ts file by using the line breaks computer on the original view model not the modified view model.

Issue #260708, opened 2025-08-09
Support previewing and merging work tree changes into the current window


sbatten:
Given a git repo opened in VS Code on a work tree (e.g. the main work tree) with working changes in other work trees, we should make it simple to both view the diff between the 2 work trees and apply those changes directly to the current window much like a stash apply. We should also make it simple to cleanup the work trees after merging.

0ptim:
Love to see this. Really need locally running parallel Copilot instances to work nicely within VS Code. I think work trees are the best way to implement this.

nguyenchristy:
Different merging cases to consider:

- Edited in worktree and/or workspace in existing file (resolve merge conflicts if any)
- New file in worktree (doesn't exist in workspace)
- Delete file in worktree (exists in workspace)
- History diverged: workspace pulled/committed after worktree branched
- Renames in worktree 
- Worktree has staged changes in file1 and unstaged changes in file1 -> staged changes are prioritized in migration

nguyenchristy:
Approaches:

Viewing diffs:
- Open diff editor. (Drawback: creates confusing UI where 'revert block' can either make changes to worktree or parent repository, depending on which repository was modified)
- Open merge editor shows clearer picture of diffs but does not align with the sole goal to view diffs 

Applying changes from worktree to workspace
- Overwrite the worktree files into the workspace (drawback: makes the assumption that the worktree is exactly how you want the workspace file to be and local changes on workspace is overwritten)
- Open merge editor and accept all changes from right to left (drawback: flashes merge editor, may not be the right approach for finding diffs and applying patches) 
- Related to merge editor, using underlying logic of generating diff patches to apply them (drawback: does not address Git nuances, seems more related to just changes in what's shown in text editor, inaccessible in Git extension without copying merging algorithms over)
- git merge file (has shown conflict markers even when there's no conflict)
- git diff, git apply (overwrites local changes from parent repo file, does not merge cleanly)

nguyenchristy:
We should be relying on the current Git flow of committing, pulling, and merging those changes and not making our own merging logic, similar to the GHPR extension.

nguyenchristy:
Make migrate changes command more visible https://github.com/microsoft/vscode/issues/264625

sandy081:
@nguyenchristy Is this can be closed?

PR #262656, opened 2025-08-21
migrate worktree changes

nguyenchristy:
Migrate worktree changes to workspace. 

Addresses https://github.com/microsoft/vscode/issues/260708

### DIFF
--- a/extensions/git/src/commands.ts
+++ b/extensions/git/src/commands.ts
@@ -3428,13 +3428,15 @@ export class CommandCenter {
 			return;
 		}
 
-		if (worktreeRepository.workingTreeGroup.resourceStates.length === 0 &&
+		if (worktreeRepository.indexGroup.resourceStates.length === 0 &&
+			worktreeRepository.workingTreeGroup.resourceStates.length === 0 &&
 			worktreeRepository.untrackedGroup.resourceStates.length === 0) {
-			window.showInformationMessage(l10n.t('There are no changes in the selected worktree to migrate.'));
+			await window.showInformationMessage(l10n.t('There are no changes in the selected worktree to migrate.'));
 			return;
 		}
 
 		const worktreeChangedFilePaths = [
+			...worktreeRepository.indexGroup.resourceStates,
 			...worktreeRepository.workingTreeGroup.resourceStates,
 			...worktreeRepository.untrackedGroup.resourceStates
 		].map(resource => path.relative(worktreeRepository.root, resource.resourceUri.fsPath));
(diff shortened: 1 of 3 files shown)


extensions/git/src/commands.ts
```diff
@@ -3407,6 +3407,87 @@ export class CommandCenter {
...
+
+		if (worktreeRepository.indexGroup.resourceStates.length === 0 &&
+			worktreeRepository.workingTreeGroup.resourceStates.length === 0 &&
+			worktreeRepository.untrackedGroup.resourceStates.length === 0) {
+			await window.showInformationMessage(l10n.t('There are no changes in the selected worktree to migrate.'));
+			return;
+		}
+
+		const worktreeChangedFilePaths = [
+			...worktreeRepository.indexGroup.resourceStates,
```
lszomoru, line 3439: We get the resources from the index in the worktree but not in the main repository. Why?

extensions/git/src/commands.ts
```diff
@@ -3407,6 +3407,87 @@ export class CommandCenter {
...
+			...repository.workingTreeGroup.resourceStates,
+			...repository.untrackedGroup.resourceStates
+		].map(resource => path.relative(repository.root, resource.resourceUri.fsPath));
+
+		// Detect overlapping unstaged files in worktree stash and target repository
+		const conflicts = worktreeChangedFilePaths.filter(path => targetChangedFilePaths.includes(path));
+
+		// Check for 'LocalChangesOverwritten' error
+		if (conflicts.length > 0) {
+			const fileList = conflicts.join('\n ');
```
lszomoru, line 3454: Have we looked at how this dialog looks like with 2, or 5 files? I am a little worried that depending on the number of files the dialog layout will completely get out of hands. The message should also say that "your changes in the repository before migrating changes."

extensions/git/src/commands.ts
```diff
@@ -3407,6 +3407,87 @@ export class CommandCenter {
...
+			const message = l10n.t('Your local changes to the following files would be overwritten by merge:\n {0}\n\nPlease stage, commit, or stash your changes before migrating changes.', fileList);
+			await window.showErrorMessage(message, { modal: true });
+			return;
+		}
+
+		await worktreeRepository.createStash(undefined, true);
+		const stashes = await worktreeRepository.getStashes();
```
lszomoru, line 3461: Any reason why we are doing this. I would expect that if `createStash()` completes, the stash should be there.

extensions/git/src/commands.ts
```diff
@@ -3407,6 +3407,87 @@ export class CommandCenter {
...
+		} catch (err) {
+			if (err.gitErrorCode === GitErrorCodes.StashConflict) {
+				const message = l10n.t('There are merge conflicts from migrating changes. Please resolve them before committing.');
+				const show = l10n.t('Show Changes');
+				const choice = await window.showWarningMessage(message, show);
+				if (choice === show) {
+					await commands.executeCommand('workbench.view.scm');
+				}
+			} else if (err.gitErrorCode === GitErrorCodes.LocalChangesOverwritten) {
+				// Should not occur
```
lszomoru: Do we still need code here given that we have code to detect this error condition upfront?

extensions/git/src/commands.ts
```diff
@@ -3407,6 +3407,87 @@ export class CommandCenter {
...
+				const message = l10n.t('Your local changes would be overwritten by the stash apply. Please commit or stash them before continuing.');
+				await window.showWarningMessage(message, { modal: true });
+			} else {
+				throw err;
+			}
+		} finally {
+			await worktreeRepository.popStash();
```
lszomoru: Any reason why we are doing this in case the migration completes successfully? To me, "migrating changes" implies that changes are "moved" from the source worktree to the target repository.

lszomoru (changes requested):
Please take a look at the comments. Thanks!

extensions/git/src/commands.ts
nguyenchristy: On the worktree side, indexGroup.resourceStates is included because the stash has staged, unstaged, and untracked changes. In the main repository, we only check unstaged changes, since git stash apply can only overwrite those (error). There's **no** error when we apply a stash that includes a modified file1.txt to the main repo's file1.txt if it's unchanged or its changes are staged. There **is** an error if a worktree's file1.txt is staged, but the main repo's file1.txt has unstaged changes.

extensions/git/src/commands.ts
nguyenchristy: Here are some examples of varying number of files. What do we think?
Screenshot 2025-08-23 211726
Screenshot 2025-08-23 211858

lszomoru: Would be interesting to see how this looks on macOS. Still I think that it makes sense to limit the number or files that we like to N (maybe 5) with some kind of text if there are more files. You can get the changes in and then polish the dialog in a follow-up commit.

lszomoru (approved):

PR #262717, opened 2025-08-21
Using an inline decorations computer

aiday-mar:
Using an inline decorations computer to extract inline decorations computations to a separate class


src/vs/editor/common/viewModel/inlineDecorations.ts
```diff
@@ -20,20 +24,241 @@ export class InlineDecoration {
...
+				if (lineStartOffsetInInputWithInjections < injectedTextEndOffsetInInputWithInjections) {
+					// Injected text ends after or in this line (but also starts in or before this line).
+					const options = injectionOptions![currentInjectedOffset];
+					if (options.inlineClassName) {
+						const wrappedTextIndentLength = this.context.getWrappedTextIndentLength(modelLineNumber);
+						const offset = (outputLineIndex > 0 ? wrappedTextIndentLength : 0);
+						const start = offset + Math.max(injectedTextStartOffsetInInputWithInjections - lineStartOffsetInInputWithInjections, 0);
+						const end = offset + Math.min(injectedTextEndOffsetInInputWithInjections - lineStartOffsetInInputWithInjections, lineEndOffsetInInputWithInjections - lineStartOffsetInInputWithInjections);
+						if (start !== end) {
+							const range = new Range(modelLineNumber, start + 1, modelLineNumber, end + 1);
```
alexdima: [Copilot generated, pls check if it's a real problem]

---

I think this is a bug, using model line numbers instead of view line numbers here.

In `InjectedTextInlineDecorationsComputer.getInlineDecorations(modelLineNumber)` at inlineDecorations.ts:

```typescript
const range = new Range(modelLineNumber, start + 1, modelLineNumber, end + 1);
```

The `InlineDecoration` range uses the **model line number**. These decorations end up in `ViewLineData.inlineDecorations` and are then merged directly (without conversion) in viewModelImpl.ts:

```typescript
if (lineData.inlineDecorations) {
    inlineDecorations = [
        ...inlineDecorations,
        ...lineData.inlineDecorations  // no conversion!
    ];
}
```

**In the old code**, `SingleLineInlineDecoration` stored only offsets (no line number), and `.toInlineDecoration(lineNumber)` was called at this merge site with the **view line number**:

```typescript
...lineData.inlineDecorations.map(d => d.toInlineDecoration(lineNumber))
```

where `lineNumber` is the view line number passed to `_getViewLineRenderingData`.

**Why this matters**: Downstream, `LineDecoration.filter()` at lineDecorations.ts compares the decoration's range line numbers against the **view** `lineNumber`:

```typescript
if (range.endLineNumber < lineNumber || range.startLineNumber > lineNumber) {
    continue; // ← skips the decoration entirely!
}
```

When other lines above cause word wrapping, a model line number (e.g., 3) won't match the view line number (e.g., 5). The decoration will be **silently dropped**. The same problem exists in the GPU render path (viewportRenderStrategy.ts), which also compares `y` (view line) against `decoration.range.startLineNumber`.

**Repro scenario**: Any document with (1) word wrapping active on lines before the injected text, and (2) injected text that has an `inlineClassName` (e.g., inlay hints with custom styling). The injected text decoration will vanish because `modelLineNumber !== viewLineNumber`.

alexdima (changes requested):

src/vs/editor/common/viewModel/inlineDecorations.ts
aiday-mar: I believe this comment is not an issue. The inline decorations are retrieved from the method `getViewLinesData` which takes in as a parameter a model line number and an outputLinIdx, which is the index of the view line number we are seeking among the view line numbers corresponding to the model line number. The method `getInlineDecorations` returns an array of array of inline decorations for all the view line numbers corresponding to a model line number. Later in `_getViewLineData` we call:

`result[globalIndex] = this._getViewLineData(lineWithInjections, lineInlineDecorations ? lineInlineDecorations[outputLineIndex] : null, outputLineIndex);`

So we fetch the correct inline decorations for the view line number.

aiday-mar: After investigation I have made a fix for this issue by passing in the view line number into the range of the decoration. You can see the screen recording here:

https://github.com/user-attachments/assets/84aa5428-7637-42fe-b19b-4787809b4b0a

src/vs/editor/common/viewModel/inlineDecorations.ts
```diff
@@ -20,20 +24,246 @@ export class InlineDecoration {
...
+export class InjectedTextInlineDecorationsComputer implements IInlineDecorationsComputer {
+
+	constructor(private readonly context: IInjectedTextInlineDecorationsComputerContext) { }
+
+	public getInlineDecorations(modelLineNumber: number): InlineDecoration[][] {
+		const injectionOffsets = this.context.getInjectionOffsets(modelLineNumber);
+		if (!injectionOffsets) {
+			return [];
+		}
+		const lineInlineDecorations = [];
```
alexdima, line 218: Small nit: this is of type `any[]`

alexdima (approved):

PR #262740, opened 2025-08-21
Per-item truncation for Inlay Hints

jackrobertscott:
Added a new configuration option to the editor's inlay hints feature, allowing users to set a maximum length for each individual inlay hint item.

* Added a new `maximumItemLength` property to the `IEditorInlayHintsOptions` interface, allowing users to specify a maximum character length for each inlay hint item. Setting it to `0` disables per-item truncation.
* Modified the option parsing logic to support the new `maximumItemLength` property, ensuring it is clamped to a valid integer range.
* Updated the `InlayHintsController` to read the new `maximumItemLength` option and enforce per-item truncation when rendering hints. If a hint exceeds the configured length, it is truncated and an ellipsis is added. [[1]](diffhunk://#diff-a4f99ac14f20ad02a66050bc0a94bf425490341ae28bbdd690d1a7da73d1b35bR547) [[2]](diffhunk://#diff-a4f99ac14f20ad02a66050bc0a94bf425490341ae28bbdd690d1a7da73d1b35bR616-R630)

### DIFF
--- a/src/vs/editor/common/config/editorOptions.ts
+++ b/src/vs/editor/common/config/editorOptions.ts
@@ -3140,6 +3140,13 @@ export interface IEditorInlayHintsOptions {
 	 * Set to 0 to have an unlimited length.
 	 */
 	maximumLength?: number;
+
+	/**
+	 * Maximum length for a single inlay hint item.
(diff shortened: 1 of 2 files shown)


jackrobertscott:

PR #262754, opened 2025-08-21
User getters for events on a few common types

mjbvz:
This defers creating the event function until the property is accessed. Useful small optimization for very commonly created objects


bpasero (approved):

bpasero:
Is there an easy way to detect the top ones? How did you come up with the list?

mjbvz:
Thanks @bpasero. I created this issue to track this: #262772. It has a simple patch I used to find some of the top hitters

I don't think it's worth trying to convert over all callers, just the top hitting ones that we are confident are safe. I'm going to put together another set of fixes for those

benibenj (approved):

PR #262766, opened 2025-08-21
Add warning for when a disposable is registered multiple times on the same store

mjbvz:
Not super common but typically indicates unclear ownership patterns


bpasero (changes requested):
Red CI

mjbvz:
Turns out to be more common that I thought in our test code. Don't think it's worth the effort since it can rarely cause issues

PR #262782, opened 2025-08-21
add bg terminal /task auto reply

meganrogge:
fixes bg terminal part of https://github.com/microsoft/vscode-copilot/issues/16430


src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/runInTerminalToolTelemetry.ts
```diff
@@ -117,6 +119,7 @@ export class RunInTerminalToolTelemetry {
 			timingExecuteMs: { classification: 'SystemMetaData'; purpose: 'FeatureInsight'; isMeasurement: true; comment: 'How long the terminal took to execute the command' };
 			pollDurationMs: { classification: 'SystemMetaData'; purpose: 'FeatureInsight'; isMeasurement: true; comment: 'How long the tool polled for output, this is undefined when isBackground is true or if there\'s an error' };
 			terminalExecutionIdleBeforeTimeout: { classification: 'SystemMetaData'; purpose: 'FeatureInsight'; isMeasurement: true; comment: 'Indicates whether a terminal became idle before the run-in-terminal tool timed out or was cancelled by the user. This occurs when no data events are received twice consecutively and the model determines, based on terminal output, that the command has completed.' };
+			autoReplyCount: { classification: 'SystemMetaData'; purpose: 'FeatureInsight'; isMeasurement: true; comment: 'The number of times the tool automatically replied to the terminal requesting user input.' };
```
Tyriar, line 142: See https://github.com/microsoft/vscode/pull/257691/files#diff-ea55cda1a7b8acc564d0bc025f854fd670a7410a82a1d5a2ce709b9be96f1dc7 for the types of things I was hoping to capture here:

```
inputToolManualAcceptCount // how many times the user clicked a detected suggestion
inputToolManualRejectCount // how many times the user rejected a detected suggestion
inputToolManualChars // how many chars were input by inputToolManualAcceptCount 
inputToolAutoChars // how many chars were input by copilot automatically responding
```

This would also be useful:

```
inputToolAutoAcceptCount // how many times copilot automatically responded
```

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/monitoring/outputMonitor.ts
```diff
@@ -220,6 +225,14 @@ export class OutputMonitor extends Disposable implements IOutputMonitor {
 			return customPollingResult;
 		}
 		const modelOutputEvalResponse = await this._assessOutputForErrors(buffer, token);
+		const confirmationPrompt = await this._determineUserInputOptions(execution, token);
+		const executedOption = await this._selectAndRunOptionInTerminal(confirmationPrompt, execution, token);
+		if (executedOption) {
+			if (recursionDepth >= PollingConsts.MaxRecursionCount) {
```
Tyriar, line 231: We want to move away from recursion here so we don't need this escape, I don't think it should have a limit ideally but the polling needs to be restructured for that. Can be done later

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/monitoring/outputMonitor.ts
```diff
@@ -240,4 +253,91 @@ export class OutputMonitor extends Disposable implements IOutputMonitor {
...
+			{ role: ChatMessageRole.User, content: [{ type: 'text', value: promptText }] }
+		], {}, token);
+
+		const responseText = await getTextResponseFromStream(response);
+		try {
+			const match = responseText.match(/\{[\s\S]*\}/);
+			if (match) {
+				const obj = JSON.parse(match[0]);
+				if (obj && typeof obj.prompt === 'string' && Array.isArray(obj.options)) {
+					return obj;
```
Tyriar: Does using `satisfies IConfirmationPrompt` give a compile error here? `obj.options` isn't totally safe and I think we should also iterate over the options and ensure each type is string or it could ptoentially lead to a runtime error.

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/monitoring/outputMonitor.ts
```diff
@@ -240,4 +253,91 @@ export class OutputMonitor extends Disposable implements IOutputMonitor {
...
+		const models = await this._languageModelsService.selectLanguageModels({ vendor: 'copilot', id: model });
+		if (!models.length) {
+			return Promise.resolve(undefined);
+		}
+		const sanitizedPrompt = confirmationPrompt.prompt;
+		const sanitizedOptions = confirmationPrompt.options.map(opt => opt);
+		const promptText = `Given the following confirmation prompt and options from a terminal output, which option should be selected to proceed safely and correctly?\nPrompt: "${sanitizedPrompt}"\nOptions: ${JSON.stringify(sanitizedOptions)}\nRespond with only the option string.`;
+		const response = await this._languageModelsService.sendChatRequest(models[0], new ExtensionIdentifier('github.copilot-chat'), [
+			{ role: ChatMessageRole.User, content: [{ type: 'text', value: promptText }] }
+		], {}, token);
```
Tyriar: Knowing about safety may only be possible by asking the agent with its full context. I don't know how to do a "full agent request" personally.

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/monitoring/outputMonitor.ts
```diff
@@ -240,4 +253,91 @@ export class OutputMonitor extends Disposable implements IOutputMonitor {
...
+		if (selectedOption) {
+			// Validate that the selectedOption matches one of the original options
+			const validOption = confirmationPrompt.options.find(opt => selectedOption.replace(/['"`]/g, '').trim() === opt.replace(/['"`]/g, '').trim());
+			if (selectedOption && validOption && validOption !== this._lastAutoReply) {
+				await execution.instance.sendText(validOption, true);
+				this._lastAutoReply = validOption;
+				return Promise.resolve(validOption);
+			}
+		}
+		return Promise.resolve(undefined);
```
Tyriar: For security and especially until this is more tried and tested, I think we want to present the options in an elicitation prompt first. Auto replying should probably work exactly the same as other auto approvals to be on the safe side, where a dropdown is available to enable auto replying. We can still evaluate what we think is the right option and present that to the user.

Tyriar (dismissed):

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/runInTerminalToolTelemetry.ts
meganrogge: I think this can be deferred?

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/monitoring/outputMonitor.ts
meganrogge: no it does not

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/monitoring/outputMonitor.ts
meganrogge: I don't know either - @roblourens any suggestions?

meganrogge: can defer since the user has to confirm for now

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/runInTerminalToolTelemetry.ts
meganrogge: seems like it can be deferred as that's going to require a lot of code and we want to keep this simple?

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/monitoring/outputMonitor.ts
Tyriar: We should check the entries still. Maybe cast it as unknown originally and then narrow that type

Tyriar (approved):

PR #262785, opened 2025-08-21
Fix terminal sticky scroll breaking 'tab moves focus'

### LINKED ISSUE #213925: Terminal sticky scroll breaks 'Tab moves focus' behavior
Type: <b>Bug</b>

1. Enable terminal sticky scroll:
"terminal.integrated.shellIntegration.suggestEnabled": true

2. Focus on the terminal
3. Press Ctrl+M to enable 'Tab moves focus' behavior
4. Press 'Tab' several times

Focus moves to the sticky line and doesn't move after.
If the sticky scroll is disabled, then the focus moves just fine.

VS Code version: Code 1.89.1 (dc96b837cf6bb4af9cd736aa3af08cf8279f7685, 2024-05-07T05:16:23.416Z)
OS version: Linux x64 6.8.0-76060800daily20240311-generic
Modes:

anthonykim1:
Resolves: https://github.com/microsoft/vscode/issues/213925

The bug is that focus can get forever "stuck" inside terminal sticky scroll when tabbing through with setting `tab moves focus`. 

Setting the index for the sticky scroll overlay to -1 would prevent this from happening.
We'd also want to make sure clicking on terminal sticky scroll still works though hence the `hoverOverlay.style.pointerEvents = 'auto';`

### DIFF
--- a/src/vs/workbench/contrib/terminalContrib/stickyScroll/browser/terminalStickyScrollOverlay.ts
+++ b/src/vs/workbench/contrib/terminalContrib/stickyScroll/browser/terminalStickyScrollOverlay.ts
@@ -7,7 +7,7 @@ import type { SerializeAddon as SerializeAddonType } from '@xterm/addon-serializ
 import type { WebglAddon as WebglAddonType } from '@xterm/addon-webgl';
 import type { LigaturesAddon as LigaturesAddonType } from '@xterm/addon-ligatures';
 import type { IBufferLine, IMarker, ITerminalOptions, ITheme, Terminal as RawXtermTerminal, Terminal as XTermTerminal } from '@xterm/xterm';
-import { $, addDisposableListener, addStandardDisposableListener, getWindow, isHTMLElement } from '../../../../../base/browser/dom.js';
+import { $, addDisposableListener, addStandardDisposableListener, getWindow } from '../../../../../base/browser/dom.js';
 import { throttle } from '../../../../../base/common/decorators.js';
 import { Event } from '../../../../../base/common/event.js';
 import { Disposable, MutableDisposable, combinedDisposable, toDisposable } from '../../../../../base/common/lifecycle.js';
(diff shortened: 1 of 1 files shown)


src/vs/workbench/contrib/terminalContrib/stickyScroll/browser/terminalStickyScrollOverlay.ts
```diff
@@ -396,6 +396,15 @@ export class TerminalStickyScrollOverlay extends Disposable {
 
 		this._stickyScrollOverlay.open(this._element);
 
+		// Prevent tab navigation from getting stuck in sticky scroll
+		this._element.querySelectorAll('*').forEach(el => {
+			if (isHTMLElement(el)) {
+				el.tabIndex = -1;
+			}
+		});
```
Tyriar: This fix prevents keyboard focus, we want it to work naturally. The right fix I think is to prevent the tab from being handled by the xterm overlay. Look into `ITerminalOptions.disableStdin` (ideally, if that works) and `Terminal.attachCustomKeyEventHandler` to do this.

Tyriar (changes requested):

src/vs/workbench/contrib/terminalContrib/stickyScroll/browser/terminalStickyScrollOverlay.ts
anthonykim1: > This fix prevents keyboard focus, we want it to work naturally. 

Meaning the above fix doesnt allow tabbing into the sticky scroll at all right? Instead we still want to tab into sticky scroll, and get out of sticky scroll naturally ?

> Look into ITerminalOptions.disableStdin (ideally, if that works

Tried adding disableStdin to ITerminalOptions and didn't seem to do anything

> Terminal.attachCustomKeyEventHandler to do this.

I think this is the way to go. Using this allows more natural tabbing in/out of sticky scroll too [6a3cf4d](https://github.com/microsoft/vscode/pull/262785/commits/6a3cf4d46c9bf440b8f635da3b240c1625b7ff9a):

https://github.com/user-attachments/assets/c4b4a261-2b04-4118-a73c-66d83f079078

src/vs/workbench/contrib/terminalContrib/stickyScroll/browser/terminalStickyScrollOverlay.ts
```diff
@@ -396,6 +396,14 @@ export class TerminalStickyScrollOverlay extends Disposable {
 
 		this._stickyScrollOverlay.open(this._element);
 
+		// Prevent tab key from being handled by the xterm overlay to allow natural tab navigation
+		this._stickyScrollOverlay.attachCustomKeyEventHandler((event: KeyboardEvent) => {
```
anthonykim1, line 400: we won't need to this._register this event handler since `this._stickyScrollOverlay` was initialized with:
```
this._stickyScrollOverlay = this._register(new ctor({
				rows: 1,
				cols: this._xterm.raw.cols,
				allowProposedApi: true,
				...this._getOptions()
			}));
```
right?

Tyriar: It's good practice to just register everything you can. I think `attachCustomKeyEventHandler` returns void though so you're good.

Tyriar (approved):

PR #262787, opened 2025-08-21
extHostMcp: accept Content-Type parameters for SSE/JSON handling

### LINKED ISSUE #262786: charsets not supported in SSE events
There's a problem when looking at the header for MCP SSE events.

Namely here:

https://github.com/microsoft/vscode/blob/ab2e3cc8e2b29e94960df8d3676cad925ac485c8/src/vs/workbench/api/common/extHostMcp.ts#L471

It tries to match on "text/event-stream" or "application/json". However, this header may also include addtional information, like "text/event-stream; charset=utf-8".

Tritlo:
Treat Content-Type as case-insensitive and allow parameters (e.g. "text/event-stream; charset=utf-8", "application/json; charset=utf-8")

Fixes #262786

### DIFF
--- a/src/vs/workbench/api/common/extHostMcp.ts
+++ b/src/vs/workbench/api/common/extHostMcp.ts
@@ -468,36 +468,32 @@ class McpHTTPHandle extends Disposable {
 			return; // no body
 		}
 
-		switch (res.headers.get('Content-Type')?.toLowerCase()) {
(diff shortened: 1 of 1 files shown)


Tritlo:

connor4312 (approved):
thanks!

Tyriar (approved):

PR #262797, opened 2025-08-21
Revert "tool picker with toolsets"

meganrogge:
Reverts microsoft/vscode#262573

This broke core tools cc @aeschli


houmark:
I think I hit this bug. In Agent mode (on latest Insiders and latest pre-release of CoPilot and CoPilot Chat, my agent is no longer able to execute any commands.

Will you be pushing out a new release to fix this soonish or should I take the rest of the day off? :) (or try to load up older versions of all three?

TylerLeonhardt (approved):

Issue #262795, opened 2025-08-21
Leaking ChatInputParts and DOM elements


roblourens:
In two scenarios- using the experimental welcome view, and also when clicking to edit a request, we call `createInput` repeatedly and reinstantiate new ChatInputParts without disposing the previous one. This leaks memory and the process of creating a ChatInputPart also adds DOM elements that are never removed, and these accumulate

<img width="393" height="779" alt="Image" src="https://github.com/user-attachments/assets/269e49df-a4f6-4939-aee0-78687c07b90d" />

Please create a ChatInputPart and reuse it, or clean up the previous things before creating new ones

roblourens:
<img width="367" height="320" alt="Image" src="https://github.com/user-attachments/assets/2002ef0e-14e8-485f-bfa2-b5a6c7a1ee61" />

If you want to create new ChatInputParts, then a `MutableDisposable` is a good pattern instead of registering them

PR #262803, opened 2025-08-21
make sure to dispose of any leftover overlays

justschen:
we are already doing `this.input.dispose()`, which gets rid of the current input container. 

ref https://github.com/microsoft/vscode/issues/262795

### DIFF
--- a/src/vs/workbench/contrib/chat/browser/chatDragAndDrop.ts
+++ b/src/vs/workbench/contrib/chat/browser/chatDragAndDrop.ts
@@ -64,6 +64,15 @@ export class ChatDragAndDrop extends Themable {
 		this.updateStyles();
 	}
 
+	override dispose(): void {
+		for (const { overlay, disposable } of this.overlays.values()) {
+			overlay.remove();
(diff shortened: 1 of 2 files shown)


src/vs/workbench/contrib/chat/browser/chatDragAndDrop.ts
```diff
@@ -64,6 +64,15 @@ export class ChatDragAndDrop extends Themable {
 		this.updateStyles();
 	}
 
+	override dispose(): void {
```
Tyriar, line 67: It's better imo to avoid overriding dispose and instead register a toDisposable that do the cleaning up. That way you don't accidentally not call super.dispose and you can put the clean up code where the set up is.

Tyriar (changes requested):

src/vs/workbench/contrib/chat/browser/chatDragAndDrop.ts
Tyriar: MutableDisposable is even better if applicable

PR #262814, opened 2025-08-21
Try to avoid rerendering experimental welcome view too often

### LINKED ISSUE #262472: Perf: showing Chat view on startup is slow
Opening the secondary sidebar by default for workspaces adds `200-300ms` on our perf bots.

Image

It is not unexpected that showing more means startup is slower, but the bump is quite significant.

Running out of sources, this can easily be seen in the perf graphs.

With Secondary Sidebar (arrows pointing to significant time spend in the chat view):

Image

Without Secondary Sidebar:

Image

**Perf traces:**
* [Slow-Trace-20250820T121140.json.zip](https://github.com/user-attachments/files/21890511/Slow-Trace-20250820T121140.json.zip)
* [Fast-Trace-20250820T121220.json.zip](https://github.com/user-attachments/files/21890536/Fast-Trace-20250820T121220.json.zip)

Perf Telemetry Data

Image

Image

Image

roblourens:
Try to fix #262472

### DIFF
--- a/src/vs/workbench/contrib/chat/browser/chatWidget.ts
+++ b/src/vs/workbench/contrib/chat/browser/chatWidget.ts
@@ -487,8 +487,8 @@ export class ChatWidget extends Disposable implements IChatWidget {
 			]))) {
 				// reset the input in welcome view if it was rendered in experimental mode
 				if (this.container.classList.contains('experimental-welcome-view')) {
-					this.welcomePart.value = undefined;
 					this.container.classList.remove('experimental-welcome-view');
+					this.welcomePart.value = undefined;
 					const renderFollowups = this.viewOptions.renderFollowups ?? false;
 					const renderStyle = this.viewOptions.renderStyle;
 					this.createInput(this.container, { renderFollowups, renderStyle });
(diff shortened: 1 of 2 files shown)


src/vs/workbench/contrib/chat/browser/chatWidget.ts
```diff
@@ -488,6 +488,7 @@ export class ChatWidget extends Disposable implements IChatWidget {
 				// reset the input in welcome view if it was rendered in experimental mode
```
roblourens, line 488: Here we're checking `affectsSome` but not checking the value that these changed to, seems like we should do that too

bhavyaus: I fixed this up a few weeks ago and this is no longer be the case.
Sorry for not taking a look sooner.

src/vs/workbench/contrib/chat/browser/chatWidget.ts
```diff
@@ -718,8 +719,9 @@ export class ChatWidget extends Disposable implements IChatWidget {
 
 
 			// reset the input in welcome view if it was rendered in experimental mode
-			if (this.container.classList.contains('experimental-welcome-view')) {
+			if (treeItems.length > 0 && this.container.classList.contains('experimental-welcome-view')) {
```
bhavyaus, line 722: This is also fixed up.

PR #262885, opened 2025-08-22
Git - accept working set changes when a resource is staged/committed

lszomoru:
* Add a new workbench command - `_chat.editSessions.accept`
* Use the workbench command when resources are staged, committed


justschen (approved):
looks good, thanks!

PR #262892, opened 2025-08-22
Adds uniqueEventId to editTelemetry.reportEditArc

hediet:


lszomoru (approved):

bryanaknight:
👋 @hediet unfortunately we are not seeing a value in `unique_edit_id` coming through on our end. We do see the field, it's just empty. Could you investigate this? 🙏

Issue #262904, opened 2025-08-22
Introduce Node MCP Server definition for extensions to wrap a node MCP package


sandy081:
There are MCP servers with npm packages. In order to easily wrap such mcp servers in extensions introduce following API. Extensions can add a dependency to this npm package and contribute the MCP server using the mcpServerProvider contribuion point and provide the node mcp server config

```ts

/**
 * The MCP package type
 */
export enum McpPackageType {
	/**
	 * Node Package
	 */
	Node = 'node',
}


/**
 * McpPackageServerDefinition represents an MCP server available by running
 * a local process with the provided package and operating on its stdin and stdout streams.
 * The process will be spawned as a child process of the extension host and by default
 * will not run in a shell environment.
 */
export class McpPackageServerDefinition {
	/**
	 * The human-readable name of the server.
	 */
	readonly label: string;

	/**
	 * Name of the package that needs to be started.
	 */
	readonly packageName: string;

	/**
	 * The type of the mcp package
	 */
	readonly packageType: McpPackageType;

	/**
	 * The working directory used to start the server.
	 */
	cwd?: Uri;

	/**
	 * Additional command-line arguments passed to the server.
	 */
	args: string[];

	/**
	 * Optional additional environment information for the server. Variables
	 * in this environment will overwrite or remove (if null) the default
	 * environment variables of the editor's extension host.
	 */
	env: Record<string, string | number | null>;

	/**
	 * Optional version identification for the server. If this changes, the
	 * editor will indicate that tools have changed and prompt to refresh them.
	 */
	version?: string;

	/**
	 * @param label The human-readable name of the server.
	 * @param packageName The name of the package that needs to be started.
	 * @param packageType The type of the mcp package
	 * @param args Additional command-line arguments passed to the server.
	 * @param env Optional additional environment information for the server.
	 * @param version Optional version identification for the server.
	 */
	constructor(label: string, packageName: string, packageType: McpPackageType, args?: string[], env?: Record<string, string | number | null>, version?: string);
}
```

Eg:

```ts
new vscode.McpNodeServerDefinition(
		'context7',
		'@upstash/context7-mcp',
)
```

cc @connor4312

connor4312:
[...]

PR #262908, opened 2025-08-22
Implement #262904

sandy081:
Implement #262904


src/vscode-dts/vscode.d.ts
```diff
@@ -20293,6 +20303,62 @@ declare module 'vscode' {
...
+		 * in this environment will overwrite or remove (if null) the default
+		 * environment variables of the editor's extension host.
+		 */
+		env: Record<string, string | number | null>;
+
+		/**
+		 * Optional version identification for the server. If this changes, the
+		 * editor will indicate that tools have changed and prompt to refresh them.
+		 */
+		version?: string;
```
connor4312: Should this be required in the package definition? We generally don't want to just point at `@latest` since that can break unexpectedly.

sandy081: Actually, I do not want to have the version field because the package should be included in the extension and the version of the extension varies if the extension wants to include a new package version.

src/vs/workbench/api/common/extHostMcp.ts
```diff
@@ -125,12 +132,16 @@ export class ExtHostMcpService extends Disposable implements IExtHostMpcService
...
 				}
 
-				servers.push({
-					id,
-					label: item.label,
-					cacheNonce: item.version || '$$NONE',
-					launch: Convert.McpServerDefinition.from(item),
-				});
+				const launch = await this._resolveMcpServerLaunch(item, extension);
+				const version = Convert.McpServerDefinition.isMcpNodeServerDefinition(item) ? undefined : item.version;
```
connor4312: We should actually have the extension version as the 'version' here. The version is just used for `cacheNonce` for core to know if we need to re-run the server to discover possibly-new tools, and that should happen when the extension updates.

src/vs/workbench/api/common/extHostMcp.ts
```diff
@@ -163,6 +174,55 @@ export class ExtHostMcpService extends Disposable implements IExtHostMpcService
...
+				args: [
+					entrypoint.with({ scheme: Schemas.file }).fsPath,
+					...item.args
+				]
+			});
+		}
+		return Convert.McpServerDefinition.from(item);
+	}
+
+	private async findNodePackageEntryPoint(packageName: string, packageLocation: URI): Promise<URI> {
```
connor4312, line 211: We probably want to share this logic with core at some point to not duplicate as we continue to support more types of packages.

src/vs/workbench/api/common/extHostMcp.ts
sandy081: ✅ Used extension version

src/vs/workbench/api/common/extHostMcp.ts
sandy081: Agreed. Will extract it appropriately then when needed

sandy081 (commented):
@connor4312 Any other feedback?

connor4312 (dismissed):

sandy081:
Since it is too late in the milestone, I will bring this to the API call next milestone.

src/vscode-dts/vscode.d.ts
```diff
@@ -20293,6 +20303,56 @@ declare module 'vscode' {
...
+	export class McpPackageServerDefinition {
+		/**
+		 * The human-readable name of the server.
+		 */
+		readonly label: string;
+
+		/**
+		 * Name of the package that needs to be started.
+		 */
+		readonly packageName: string;
```
alexr00: It looks like when we have a "name" property we just call it `name`.

src/vscode-dts/vscode.d.ts
```diff
@@ -20293,6 +20303,56 @@ declare module 'vscode' {
...
+
+		/**
+		 * The type of the mcp package
+		 */
+		readonly packageType: McpPackageType;
```
alexr00: I think we usually prefer `kind` over `type`. I defer to @mjbvz though. 
```suggestion
		readonly kind: McpPackageKind;
```

sandy081: We also have types - Eg: FileType

alexr00: You are correct, we do also have `type`, but we usually prefer `kind`.

sandy081:
@mjbvz Can you please take a look and give your feedback?

alexr00 (commented):
Looks good to me, thank you!

src/vscode-dts/vscode.d.ts
```diff
@@ -20332,6 +20342,56 @@ declare module 'vscode' {
...
+
+		/**
+		 * @param label The human-readable name of the server.
+		 * @param packageName The name of the package that needs to be started.
+		 * @param packageType The type of the mcp package
+		 * @param args Additional command-line arguments passed to the server.
+		 * @param env Optional additional environment information for the server.
+		 * @param version Optional version identification for the server.
+		 */
+		constructor(label: string, packageName: string, packageType: McpPackageKind, args?: string[], env?: Record<string, string | number | null>, version?: string);
```
mjbvz, line 20394: Instead of talking about packages, could this be about node processes?

Instead of a server name, I'd expect that I could point VS Code to a `.js` file to spawn. This could be a package or it could be a random file in my extension

src/vscode-dts/vscode.d.ts
```diff
@@ -20332,6 +20342,56 @@ declare module 'vscode' {
 		constructor(label: string, command: string, args?: string[], env?: Record<string, string | number | null>, version?: string);
 	}
 
+	/**
+	 * McpPackageServerDefinition represents an MCP server available by running
+	 * a local process with the provided package and operating on its stdin and stdout streams.
+	 * The process will be spawned as a child process of the extension host and by default
```
mjbvz, line 20350: Why do we need this if we have `McpStdioServerDefinition`?

mjbvz (changes requested):
Let's hold off on finalization until this has gone through an api discussion and had some time to be tested as a proposal

src/vscode-dts/vscode.d.ts
sandy081: This is explained in the issue - https://github.com/microsoft/vscode/issues/262904 - This allows easy wrapping of an existing MCP server (npm package) into an extension.

src/vscode-dts/vscode.d.ts
sandy081: This requirement you are talking about is taken care by the `McpStdioServerDefinition`. As mentioned in my previous comment, if there is an npm package for mcp server this API allows wrapping that MCP server in VS Code extension easily. For example, taking `@upstash/context7-mcp` MCP server:
- Add npm dependency to '@upstash/context7-mcp' server in package.json
- Contribute the context7 mcp server in package.json and provide the definition programtically like below
```ts
new vscode.McpNodeServerDefinition(
		'context7',
		'@upstash/context7-mcp',
)
```

src/vscode-dts/vscode.d.ts
mjbvz: How difficult is this without the helper?

src/vscode-dts/vscode.d.ts
sandy081:
```ts
new vscode.McpStdioServerDefinition(
					'context7',
					'node',
					[
						path.join(__dirname, '..', 'node_modules', '@upstash', 'context7-mcp', 'dist', 'index.js'),
					]
				)
```

One has to manually compute the path

mjbvz: Maybe we could bundle up the `findNodePackageEntryPoint` function you added into a helper function that extensions can use? Either published as a new npm package or include it in our extension samples or docs for extensions to use?

I'd hope that for most packages you'd be able to use `import.meta.resolve` too:

```ts
new vscode.McpStdioServerDefinition(
	'context7',
	'node',
	[
		import.meta.resolve('@upstash/context7-mcp'),
	]
);
```

This doesn't work for `@upstash/context7-mcp` because it doesn't have a `main` entry in its `package.json`

sandy081: If every extension has to do the same then it is much better if it is wrapped inside core via API. It will make it easier for extensions to use. Going forward we would like to extend this also for other package types like python.

mjbvz: We can discuss  this at the next api sync but right now my feeling is that requiring extensions resolve paths isn't a big enough burden to justify a new API

I also am concerned that this api will commit us to having to support all sorts of quirky node package setups, and that supporting other languages opens the door for a long tail of language specific additions and complicated package resolver logic

PR #262931, opened 2025-08-22
make check safer

meganrogge:


src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/monitoring/outputMonitor.ts
```diff
@@ -299,9 +300,14 @@ export class OutputMonitor extends Disposable implements IOutputMonitor {
 		try {
 			const match = responseText.match(/\{[\s\S]*\}/);
```
roblourens, line 301: This check doesn't add anything IMO, it doesn't tell you that it's JSON, I would just throw it into `JSON.parse`.

If you have issues with it getting the format right, you could use a tool (an ad-hoc tool definition not a real registered tool) and forcing it to call it with `tool_choice` (the real right approach to this problem is using structured outputs, but we don't expose that unfortunately).

Consider passing `modelOptions: { max_tokens: something }` to prevent runaway incorrect responses from using tons of tokens, I just had to fix this in a similar 4o-mini request.

roblourens (approved):

PR #262959, opened 2025-08-22
fix(editor)(#261780): transform UPPER_CASE to PascalCase

Selva-Ganesh-M:
fixes #261780

The fix supports the new expectation:

> FOO_BAR → FooBar
> 
> FOO BAR A → FooBarA

while keeping the current behavior intact:

> parseHTMLString → ParseHTMLString,
> 
> audioConverter.convertM4AToMP3() → AudioConverter.ConvertM4AToMP3()

Added unit tests to cover both cases.

reviews and suggestions are welcome.


Selva-Ganesh-M:

hiiamkarati (commented):
Thanks for this contribution! The PascalCase refactor looks clean and consistent, and I can see that related test cases were updated as well.  

A couple of things to double-check before merging:  
1. Please confirm whether these identifiers are used in any public API/extensions. If yes, renaming them could introduce breaking changes.  
2. Ensure all references to the old UPPER_CASE values have been updated consistently across the codebase (not just in the files shown here).  
3. The updated tests cover common cases, but it would be good to verify edge cases (numbers, special characters, mixed casing) are still passing.  

Overall, this is a solid improvement in naming consistency 👍

hediet (dismissed):

dbaeumer (dismissed):

src/vs/editor/contrib/linesOperations/browser/linesOperations.ts
```diff
@@ -1267,6 +1267,7 @@ export class CamelCaseAction extends AbstractCaseAction {
 export class PascalCaseAction extends AbstractCaseAction {
 	public static wordBoundary = new BackwardsCompatibleRegExp('[_\\s-]', 'gm');
 	public static wordBoundaryToMaintain = new BackwardsCompatibleRegExp('(?<=\\.)', 'gm');
+	public static upperCaseWordMatcher = new BackwardsCompatibleRegExp('^[A-Z]+$', 'mu');
```
karuturi: regex doesnt handle unicode characters. Should this be `/^\p{Lu}+$/u` ?

EFanZh:
For acronyms, different programming languages have different styles:

- Keep upper case (like `HTTPServer`): [Go](https://go.dev/wiki/CodeReviewComments#initialisms), [Python](https://peps.python.org/pep-0008/#descriptive-naming-styles).
- Treat as word (like `HttpServer`): [C#](https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/capitalization-conventions#capitalization-rules-for-identifiers), [Kotlin](https://kotlinlang.org/docs/coding-conventions.html#choose-good-names), [Rust](https://rust-lang.github.io/api-guidelines/naming.html#casing-conforms-to-rfc-430-c-case).

Can we support both styles?

hediet (approved):

Tyriar (approved):

PR #262963, opened 2025-08-22
Create overlay to guard for given chat session type

joshspicer:
ref: https://github.com/microsoft/vscode/issues/260227


src/vs/workbench/contrib/chat/browser/chatContentParts/chatMultiDiffContentPart.ts
```diff
@@ -132,7 +133,14 @@ export class ChatMultiDiffContentPart extends Disposable implements IChatContent
 
 			const activeEditorUri = this.editorService.activeEditor?.resource;
 			let marshalledUri: any | undefined = undefined;
+			let contextKeyService: IContextKeyService = this.contextKeyService;
 			if (activeEditorUri) {
+				const { authority } = activeEditorUri;
+				const overlay: [string, any][] = [];
```
Tyriar: Can you use `unknown` instead of `any`?

Tyriar (changes requested):

src/vs/workbench/contrib/chat/browser/chatContentParts/chatMultiDiffContentPart.ts
joshspicer: done!

Tyriar (approved):

PR #255455, opened 2025-07-11
Add RTL support based on decorations

remcohaszing:
This adds RTL support to Monaco editor. A line can be marked as RTL using decorations. Decorations can now specify a text direction. For every line, the decorations on that line are counted which specify the text direction. If there are more RTL decorations than there are RTL decorations, the line is considered RTL.

If a line is RTL, the content is aligned to the right side of the screen, taking the scrollbar into account.

If text content matches the text direction, pressing a left or right arrow key, moves the cursor into the same direction. If the content is of the inverse direction, the cursor is moved into the opposite direction.

The <kbd>Home</kbd> key moves the cursor to the start of the line. In case of RTL, this is the right.

The <kbd>End</kbd> key moves the cursor to the end of the line. In case of RTL, this is the left.

These cursor movements also work for multi selections.

Both RTL and LTR work with or without text wrapping, even with the advanced text wrapping strategy.

PR #262964, opened 2025-08-22
Fix editor edge clicking

remcohaszing:
When clicking the 10 left-most pixels of the editor, the cursor was moved to the beginning of the line. When clicking the 10 right-most characters of the editor, the cursor was moved to the end of the line. This is not correct in RTL. Clicking the left pixels should move the cursor to the left. Clicking the right pixels should move the cursor to the right.

This was an oversight in #255455.

Also the `horizontalScrollPadding` is hardcoded to 10. I noticed in my own editor this means I can’t put the cursor between the first two characters in VSCode. It will always snap to the beginning of the line. Should this be made configurable (in a separate PR)?


aiday-mar (approved):

connor4312 (approved):

remcohaszing:
This was approved 2 weeks and marked to auto-merge, but it’s still not merged. Looks like a prerequisite is missing. I think CI needs to run?

PR #262980, opened 2025-08-22
Explorer: add `explorer.collapseResource` to collapse a specific folder (by URI) without changing focus (#261598)

### LINKED ISSUE #261598: Explorer command to collapse a resource by URI
**Problem:**

Extensions currently cannot collapse a specific folder in the built-in Explorer without first focusing it via ```revealInExplorer```, then calling ```list.collapse```/```list.collapseRecursively```. This causes visible flicker and steals focus from the editor.

**Currently:**

- list.collapse operates on the currently focused list/tree item and accepts no arguments; it’s generic by design.
- There is no public API/command to collapse a specific Explorer node by URI without moving focus.

**Solution:**

```explorer.collapseResource```
Args: A simple resource URI
Collapses the folder node matching resource (or the parent folder if resource is a file). No focus/selection change. No scrolling. No-op if not found.

**Backwards compatibility**

- New commands only; no behavior change to existing commands.
- No keybindings added by default.

Csabesz07:
## Summary
Adds two new commands `explorer.collapseResource` and `explorer.collapseResourceRecursive` that collapses a specific folder (and it's child folders with the recurscive command) in the Explorer given its resource URI, without changing focus/selection.

## Motivation
Extensions and automation need to collapse folders programmatically without UI flicker or stealing focus. Existing commands operate only on the focused element. This would allow the development of an extension which keeps the explorer tidy.

## Implementation
- Registers `explorer.collapseResource` (args: `URI`).
- Registers `explorer.collapseResourceRecursive` (args: `URI`).
- No behavior change to existing commands.

## Backwards compatibility
- New commands only; no changes to existing public commands.

## Related
Fixes #261598


Csabesz07:

PR #262987, opened 2025-08-22
Don't cancel an all-provider request when the single-provider request is a subset

hediet:
Fixes https://github.com/microsoft/vscode/issues/262986
FYI @benibenj

### DIFF
--- a/src/vs/editor/contrib/inlineCompletions/browser/structuredLogger.ts
+++ b/src/vs/editor/contrib/inlineCompletions/browser/structuredLogger.ts
@@ -53,7 +53,7 @@ interface IFetchResult {
  * The sourceLabel must not contain '@'!
 */
 export function formatRecordableLogEntry<T extends IRecordableLogEntry>(entry: T): string {
-	return entry.sourceId + ' @@ ' + JSON.stringify({ ...entry, sourceId: undefined });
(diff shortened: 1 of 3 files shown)


src/vs/editor/contrib/inlineCompletions/browser/structuredLogger.ts
```diff
@@ -53,7 +53,7 @@ interface IFetchResult {
  * The sourceLabel must not contain '@'!
 */
 export function formatRecordableLogEntry<T extends IRecordableLogEntry>(entry: T): string {
-	return entry.sourceId + ' @@ ' + JSON.stringify({ ...entry, sourceId: undefined });
+	return entry.sourceId + ' @@ ' + JSON.stringify({ ...entry, modelUri: (entry as any).modelUri?.toString(), sourceId: undefined });
```
Tyriar, line 56: Unsafe? Could this get mangled in product build?

hediet: Thanks for paying attention to code smells :)
But I think it is fine here, this is debug logging only. And no, it cannot get mangled, as this is a json object (which would break all kinds of interop, if json object properties would get mangled).

mjbvz (approved):

PR #263066, opened 2025-08-23
allow classes for syntax highlighting in collapsed cells

### LINKED ISSUE #16924: option to retain editor comment font for jupyter interactive window cell headings
### Applies To

- [ ] Notebooks (.ipynb files)
- [x] Interactive Window and\/or Cell Scripts (.py files with \#%% markers)

### What happened?

I'm using jupyter interactive windows with .py files in vscode using the # %% cell magic

previously, the first line of the cell was displayed above the collapsed code cell, in the editor comment font (same size and colour).  this was helpful to quickly and easily see cell breaks in the interactive window, and my code is organised so that the first line summarises the contents of the cell for easy navigation in the interactive window with the cells collapsed

since a recent change (maximum past 2 weeks) the cell headers now appear in a small white font, possibly markdown, although changing my markdown settings and other font settings in vscode don't affect it

this means that the cell headings are now a very similar size and colour to the notebook output font, which makes the interactive window much harder to read

I've tried changing various settings, including this one, which I think may be related

Jupyter › Interactive Window: Notebook Repl [Experimental.]
The interactive window will use a notebook as the editor model.

however, neither activating or deactivating that option resolved this

grateful if there could be an option to control the font of the cell headings directly, or to change back to the previous behaviour with the editor comment font (or at least a distinct font) used

### VS Code Version

Version: 1.103.1 (user setup) Commit: 360a4e4fd251bfce169a4ddf857c7d25d1ad40da Date: 2025-08-12T16:25:40.542Z Electron: 37.2.3 ElectronBuildId: 12035395 Chromium: 138.0.7204.100 Node.js: 22.17.0 V8: 13.8.500258-electron.0 OS: Windows_NT x64 10.0.26100

### Jupyter Extension Version

Installation Identifier ms-toolsai.jupyter Version 2025.7.0 Last Updated 2025-08-21, 11:44:17 Size 17.14MB Marketplace Published 2020-11-11, 19:14:18 Last Released 2025-08-21, 10:47:58

### Jupyter logs

```shell

```

_No response_

_No response_

_No response_

None

amunger:
fix https://github.com/microsoft/vscode-jupyter/issues/16924

This change restores just the classes for collapsed cell previews


amunger:
This looks like a pretty standard set of classnames that would be used for this type of thing, so perhaps there is a more common way to allow these tags (or should be)?

src/vs/workbench/contrib/notebook/browser/view/cellParts/markupCell.ts
```diff
@@ -262,8 +265,9 @@ export class MarkupCell extends Disposable {
 		const element = DOM.$('div');
 		element.classList.add('cell-collapse-preview');
 		const richEditorText = this.getRichText(this.viewCell.textBuffer, this.viewCell.language);
-		domSanitize.safeSetInnerHtml(element, richEditorText);
+		domSanitize.safeSetInnerHtml(element, richEditorText, getTokenizedPreviewSanitizerConfig());
```
mjbvz: Worth checking if you need the sanitize at all in this flow. If the html string just the output of `tokenizeLineToHTML`, I *think* it should be safe but Alex D or Martin should be able to confirm

amunger: @alexdima / @aeschli - should I just be setting `element.innerHtml` directly with the output of [_tokenizeToString](https://github.com/microsoft/vscode/blob/e2694277d97640102f601dd88561894f07423f96/src/vs/editor/common/languages/textToHtmlTokenizer.ts#L129)? the input text to that function is the first line of the cell input, which could be anything.

mjbvz: My thought is that so long as we are only doing syntax highlighting, this would be the same as rendering the content in a normal text editor, which we already do for untrusted content. However good to double check

alexdima: I reviewed `_tokenizeToString` and that looks to be escaping special characters (`<`, `>`, `&`) correctly, so it should be safe to set that as `.innerHtml`. However, it would be best to also go through trusted types like the rest of the places.

deepak1556 (approved):

deepak1556:
Hmm I meant to approve the changes in `src/vs/code/electron-browser/workbench/` not the entire PR.. @amunger please get a second verification from @alexdima for the innerHTML changes.

Issue #262756, opened 2025-08-21
Support CLI agents


roblourens:
We will explore supporting CLI agents (claude code, codex, etc) as part of Copilot Chat in vscode. 

Current code is in https://github.com/microsoft/vscode-copilot-chat/tree/roblou/classic-quail, `@claude` can be invoked as a participant, but that would not be the real UX.

### Required for a basically usable experience
- Register a chat session provider to show a separate section in the chat sessions list
	- And use the chat widget mode that locks it to the agent
- Store conversation history independently in extension? 
	- How to keep it in sync with real state?
	- Read their format from `~/.claude/projects/<dir>/<guid>`?
	- For now, will keep our own copy, but ideally the sdk would be able to list conversations and their data
- Show tool confirmations using vscode's tool confirmations
	- This is driven via an mcp tool plugged into the claude code sdk
	- Needs all the polish of different tool confirmation types, eg for terminals, with the same auto-approve logic
- Track edits
	- Allow user to undo edits like usual? Or will that confuse the agent?
	- Apply our logic for allowing edits inside the workspace
	- But requiring confirmation for certain files (how to do this when the agent is editing the file?)
	- The agent editing files that are unsaved in vscode?
- Hide tool picker or show the agent's tools in the tool picker

### Polish
- `/ide` context eg open files
- Support multi-root via `additionalDirectories`
- Show background bash shells (`/bashes`)?
- Support claude code modes? plan, accept edits, full permission
- `/compact`
- `/memory`
- Manage agents, show subtasks. The user's configured agents are used by the sdk
- Start and manage a claude code agent on a different directory (eg worktree)
- Support custom `/` commands
- Test all types of vscode attachments
- Enable mcp servers configured in vscode

PR #263114, opened 2025-08-24
Add ConfirmationTool

roblourens:
#262756

### DIFF
--- a/src/vs/workbench/contrib/chat/common/tools/confirmationTool.ts
+++ b/src/vs/workbench/contrib/chat/common/tools/confirmationTool.ts
@@ -1,0 +1,96 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { CancellationToken } from '../../../../../base/common/cancellation.js';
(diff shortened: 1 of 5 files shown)


src/vs/workbench/contrib/chat/common/tools/confirmationTool.ts
```diff
@@ -0,0 +1,96 @@
...
+			throw new Error('Missing required parameters for ConfirmationTool');
+		}
+
+		const confirmationType = parameters.confirmationType ?? 'basic';
+
+		// Create different tool-specific data based on confirmation type
+		let toolSpecificData: IChatTerminalToolInvocationData | undefined;
+
+		if (confirmationType === 'terminal') {
+			// For terminal confirmations, use the terminal tool data structure
```
Tyriar, line 64: This is so the agent can ask the user for confirmation about something? What's the use case around asking for terminal confirmations here? Isn't it a tool's job, not the agents, to handle permission of commands like this?

Is this going to use the terminal auto approve code unexpectedly?

TylerLeonhardt (approved):

src/vs/workbench/contrib/chat/common/tools/confirmationTool.ts
roblourens: This is for the cli agents. For now it's mainly based on claude code's concept of permissions. I'm still a bit split on leaning into it being an "external agent" that has its own native configuration vs trying to configure it the same way that our agent is configured.

PR #263136, opened 2025-08-24
fix: memory leak in codelens controller

SimonSiefke:
Before: 
code-lens-count-before

In the code lens controller, the old code lens models don't seem to be cleared in all cases.  It seems they are only cleared if `toResolve.length` is greater than zero. Due to the early return, the cleanup seems missing for the case `toResolve.length === 0 `.

```ts
// before
private _resolveCodeLensesInViewport(): void {

    /* ... */
    if (toResolve.length === 0) {
	    return; 
    }

    /* ... */ 
    this._oldCodeLensModels.clear();
}
```

---

After: 
code-lens-count-after

 The change is to clear oldCodeLensModels in both cases:

```ts
// after
private _resolveCodeLensesInViewport(): void {

    /* ... */
    if (toResolve.length === 0) {
		this._oldCodeLensModels.clear();
	    return; 
    }

    /* ... */ 
    this._oldCodeLensModels.clear();
}
```

### DIFF
--- a/src/vs/editor/contrib/codelens/browser/codelensController.ts
+++ b/src/vs/editor/contrib/codelens/browser/codelensController.ts
@@ -395,6 +395,7 @@ export class CodeLensContribution implements IEditorContribution {
 		});
 
 		if (toResolve.length === 0) {
+			this._oldCodeLensModels.clear();
 			return;
 		}
 


mjbvz (approved):
Thanks!

lszomoru (approved):

PR #263210, opened 2025-08-25
Preserve case in multicursor replace command

### LINKED ISSUE #106502: Multiple Cursors - Preserve Case
**Use Case**
Take the following:
`const [value, setValue] = useState();`

I want to change it to:
`const [array, setArray] = useState();`

**Feature Ask**
It would be great to:
1. Select `value`.
2. Press `Ctrl+D`.
3. Type `array`.
4. See that both symbols have changed, while preserving their proper casing.

Right now, I need to conduct two edit operations to do this.  Alternatively, I can use a single Find & Replace operation, but that would be disproportionately time-consuming for this task, as well as less precise.

alpalla:
# Problem
Fixes https://github.com/microsoft/vscode/issues/106502.

Reading through the issue it seems the main problem involves preserving the case of the first letter of a variable name when executing a replace command on multiple case-insensitive selections of that word.

# Solution
To fix this, when defining the replace command for a given selection, we can check the case of the first letter of the text to replace, and update the case of the character that will replace it accordingly.

So if we have selections `"value"` and `"Value"`, and type `"a"` (while typing `array`), when composing the replace command for `"Value"` we see that it starts with a capital and replace the selection with `"A"`, not with `"a"`.

For now this logic is only applied in the case of multiple selections.

# Testing
Given
```
const [value, setValue] = useState();
```

Select `"value"` and `"Value"`, and then type `"array"`.

Expected result
```
const [array, setArray] = useState();
```

https://github.com/user-attachments/assets/2a3f2716-1785-45f5-9ff7-ae37d0e482cd

Note: it might make sense to make this feature opt-in, whether through some setting or a dedicated button as was suggested in the issue. I'm open to suggestions. 🙏

### DIFF
--- a/src/vs/editor/common/cursor/cursorTypeEditOperations.ts
+++ b/src/vs/editor/common/cursor/cursorTypeEditOperations.ts
@@ -499,12 +499,14 @@ export class InterceptorElectricCharOperation {
 
 export class SimpleCharacterTypeOperation {
 
-	public static getEdits(config: CursorConfiguration, prevEditOperationType: EditOperationType, selections: Selection[], ch: string, isDoingComposition: boolean): EditOperationResult {
+	public static getEdits(config: CursorConfiguration, model: ITextModel, prevEditOperationType: EditOperationType, selections: Selection[], ch: string, isDoingComposition: boolean): EditOperationResult {
 		// A simple character type
 		const commands: ICommand[] = [];
 		for (let i = 0, len = selections.length; i < len; i++) {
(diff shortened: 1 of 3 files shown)


aiday-mar (commented):
Hi @alpalla thanks for the PR. I looked at it and it seems that the way it is currently coded, it checks that the previous full selected string contains an upper case letter at the beginning and if yes, it will replace the first character inserted by an upper case letter. This means that in the following case: 

const [value, setvAlue] = useState();

If you were to select the word `value` and replace with `array`, this would incorrectly replace the above to:

const [array, setarray] = useState();

instead of 

const [array, setaRray] = useState();

Additionally we want this to work with only one cursor, there is no reason why it should not work with one cursor too while replacing. Also this should be behind a setting.

Issue #240296, opened 2025-02-10
Resuming after OS suspend can cause sticky scroll to not work correctly until hovered


Tyriar:
I think we have a special handler to prevent this in the main terminal that probably needs to be done on the sticky scroll one:

![Image](https://github.com/user-attachments/assets/d54344e0-a11f-4a59-9f35-f24d5ea0a1b0)

Issue #224750, opened 2024-08-04
1.92 Regression: scrolling in integrated terminal is borked


maxpatiiuk:
Does this issue occur when all extensions are disabled?: **Yes**.

After updating to VS Code 1.92, I noticed several scroll-related issues with the integrated terminal:

1. After exiting a full screen CLI app (`git diff` or `git log`), sometimes scroll up or down doesn't work anymore until I press cmd+c - at which point I am teleported back to the start of the cli log.
  
   See the below video:

   https://github.com/user-attachments/assets/58e3c21e-c7fc-4e53-ba43-cc4708e0e8fd

   - start `git log`
   - scroll works (though way too sensitive - see issue 2)
   - press `q`
   - scroll up down no longer works
   - press `cmd+c`
   - as soon as I try to scroll again, I am teleported to the top of the cli log output (though, at least scroll works after that)

   - Note, while the above triggers the issue, sometimes the issue also occurs even without that (at one point I just run `git status` and this bug was triggered)

2. In full-screen CLI apps (`vim`, `git log`, `man`), the scroll sensitivity is way higher than when scrolling in non-full-screen apps. This scroll sensitivity is uncomfortably high, way higher than it was before, and remains high even if I set `terminal.integrated.mouseWheelScrollSensitivity` to a very low value like 0.1.

   https://github.com/user-attachments/assets/6315ab60-bf33-42ac-a5ac-b68981fa4790

3. When scrolling through full-screen CLI app (`git log` or `git diff`), sometimes the terminal will not let me scroll up. If I try to scroll up, I am "teleported" back to previous position. As soon as I scroll down, I can't scroll back up to the previous text.

    https://github.com/user-attachments/assets/5f8855ba-bbf4-4096-bcd0-ea17ada0aa58

Relevant information:
- VS Code Version: 1.92
- OS Version: macOS 14.5
- Issues are reproducible in ZSH. I was also able to trigger issue 1 in BASH
  - In ZSH, I am using oh-my-zsh and powerlevel19k, BUT I was able to reproduce the issues even after I temporary disabled oh-my-zsh and powerlevel19k in my `.zshrc`
- Issues are reproducible even if I start VS Code with empty user dir and empty extensions dir.

A few relevant settings:
`terminal.integrated.smoothScrolling` is off. Enabling it does not make things better
`terminal.integrated.fastScrollSensitivity` - N/A as I am not holding the `Alt` key - I am just scrolling using the touchpad
[...]

PR #263292, opened 2025-08-25
Enable terminal sticky scroll as default true in both stable and insiders

### LINKED ISSUE #209003: Consider enabling sticky scroll by default in stable
It's been on by default in Insiders for a few months now and there haven't been many complaints

fyi @meganrogge

anthonykim1:
Resolves: https://github.com/microsoft/vscode/issues/209003

With only https://github.com/microsoft/vscode/issues/198116 and https://github.com/microsoft/vscode/issues/240296 remaining under `terminal-sticky-scroll` label, shall we have default to true, and have more folks in stable try out.


Tyriar (changes requested):
Looks good, need to remove the `product` import to fix compile

Tyriar (approved):

mhsmith:
@anthonykim1: Sorry for posting on an unrelated page, but you sent me a message in #224750, which is a locked issue. Could you unlock it so I can reply?

Tyriar:
@mhsmith unlocked, thanks for letting us know

PR #263304, opened 2025-08-25
Update build TS versions

mjbvz:


build/lib/treeshaking.js
```diff
@@ -230,7 +230,7 @@ var NodeColor;
     NodeColor[NodeColor["Black"] = 2] = "Black";
 })(NodeColor || (NodeColor = {}));
 function getColor(node) {
-    return node.$$$color || NodeColor.White;
+    return node.$$$color || 0 /* NodeColor.White */;
```
Tyriar, line 233: Is this a change in TS output? I would have expected the `NodeColor` enum definition to disappear too?

Tyriar (approved):

build/lib/treeshaking.js
mjbvz: yes we're testing the native TS build here and they just merged initial const enum support https://github.com/microsoft/typescript-go/pull/1599

Opened https://github.com/microsoft/typescript-go/issues/1632 to track not emitting the enum itself in these cases too

PR #263350, opened 2025-08-25
Adding timing to chat sessions API and grouping

osortega:
Adding `timing` to chat sessions API and group them in the UI

### DIFF
--- a/src/vscode-dts/vscode.proposed.chatSessionsProvider.d.ts
+++ b/src/vscode-dts/vscode.proposed.chatSessionsProvider.d.ts
@@ -96,15 +96,15 @@ declare module 'vscode' {
 		tooltip?: string | MarkdownString;
 
 		/**
-		 * The times at which session started and ended,
+		 * The times at which session started and ended
 		 */
 		timing?: {
 			/**
-			 * Session start time.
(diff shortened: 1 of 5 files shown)


src/vscode-dts/vscode.proposed.chatSessionsProvider.d.ts
```diff
@@ -94,6 +94,20 @@ declare module 'vscode' {
...
 		tooltip?: string | MarkdownString;
+
+		/**
+		 * The times at which session started and ended,
+		 */
+		timing?: {
+			/**
+			 * Session start time.
+			 */
+			startTime: number;
```
mjbvz, line 105: These are timestamp, right? Maybe call that out in the jsdoc

mjbvz (commented):
API changes look good and match the timeline api proposal

src/vscode-dts/vscode.proposed.chatSessionsProvider.d.ts
osortega: Good idea, added it to the comment above following the same patterns as other types

rebornix (approved):

PR #263358, opened 2025-08-26
core: update thinking part shape for responses api

justschen:
fix https://github.com/microsoft/vscode/issues/262994

329-Vibrava

### DIFF
--- a/src/vscode-dts/vscode.proposed.languageModelThinkingPart.d.ts
+++ b/src/vscode-dts/vscode.proposed.languageModelThinkingPart.d.ts
@@ -16,7 +16,7 @@ declare module 'vscode' {
 		/**
 		 * The thinking/reasoning text content.
 		 */
-		value: string;
+		value: string | string[];
 
 		/**
 		 * Optional unique identifier for this thinking sequence.
@@ -28,7 +28,7 @@ declare module 'vscode' {
 		/**
 		 * Optional metadata associated with this thinking sequence.
 		 */
-		metadata?: string;
+		metadata?: { readonly [key: string]: any };
 
 		/**
 		 * Construct a thinking part with the given content.
@@ -36,7 +36,7 @@ declare module 'vscode' {
 		 * @param id Optional unique identifier for this thinking sequence.
 		 * @param metadata Optional metadata associated with this thinking sequence.
 		 */
-		constructor(value: string, id?: string, metadata?: string);
+		constructor(value: string | string[], id?: string, metadata?: { readonly [key: string]: any });
 	}
 
 	export interface LanguageModelChatResponse {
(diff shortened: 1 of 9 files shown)


karthiknadig (approved):

src/vscode-dts/vscode.proposed.languageModelThinkingPart.d.ts
```diff
@@ -16,7 +16,7 @@ declare module 'vscode' {
 		/**
 		 * The thinking/reasoning text content.
 		 */
-		value: string;
+		value: string | string[];
```
roblourens, line 19: Why `string[]`?

justschen: cc @karthiknadig 

apparently ResponsesAPI can have parallel summaries, so this was an ask from karthik to pass that thru (though atm i don't think we know what this will ultimately look like from a rendering perspective,)

karthiknadig: @roblourens Looking at the docs, it seems like, we can get delta for various thinking steps with `summary_index` indetifying which summary step the delta applies to.

essentially for a problem like create a python web app for me. The model can have 3 individual thoughts, like this:
step 0: understand what the user wants
step 1: pick a framework, like `flask`
step 2: make a single file app

something like this:
```
# Step 0 starts (partial)
{
  "type": "response.reasoning_summary_text.delta",
  "delta": {
    "reasoning": {
      "summary": "The user is asking me to make"
      ,
      "summary_index": 0
    }
  }
}

# Step 1 starts before step 0 is finished
{
  "type": "response.reasoning_summary_text.delta",
  "delta": {
    "reasoning": {
      "summary": "To implement this quickly, I should pick a framework"
      ,
      "summary_index": 1
    }
  }
}

# More text arrives for step 0
{
  "type": "response.reasoning_summary_text.delta",
  "delta": {
    "reasoning": {
      "summary": " a small web application."
      ,
      "summary_index": 0
    }
  }
}

# Step 2 even starts before step 1 is finished
{
  "type": "response.reasoning_summary_text.delta",
  "delta": {
    "reasoning": {
      "summary": "The simplest approach is to write everything into a single"
      ,
      "summary_index": 2
    }
  }
}

# Step 1 finishes
{
  "type": "response.reasoning_summary_text.delta",
  "delta": {
    "reasoning": {
      "summary": " — Flask is a good fit."
      ,
      "summary_index": 1
    }
  }
}

# Step 2 finishes
{
  "type": "response.reasoning_summary_text.delta",
  "delta": {
    "reasoning": {
      "summary": " file app for clarity."
      ,
      "summary_index": 2
    }
  }
}

```

My thought was if we make summary `string[]` then we can handle these deltas. Currently we are smushing them with `\n\n` which might not be correct for these.

I am open to other recommendations.

roblourens: Oh, interesting. And it would stream them in parallel like that? Seems very odd but ok.

PR #263369, opened 2025-08-26
Add configurable prompt file suggestions with when-clause support

digitarald:
## Summary

This PR introduces a new `promptFilesRecommendations` configuration setting that allows users to configure which prompt files should be suggested in the chat welcome view based on contextual when-clause expressions or simple boolean values.

image

cc @eli-w-king 

## Key Features

- **Contextual suggestions**: Configure prompt files to appear based on file type, language, or other VS Code context
- **Flexible configuration**: Support both boolean values (`true`/`false`) and when-clause expressions
- **Smart caching**: Async loading of prompt descriptions to avoid blocking UI rendering
- **Unified experience**: Suggestions now work in both experimental and regular welcome views

## Configuration Example

```json
{
  "promptFilesRecommendations": {
    "plan": true,                            // Always suggest
    "a11y-audit": "resourceExtname == .html", // Only for HTML files  
    "document": "resourceLangId == markdown", // Only for Markdown files
    "debug": false                           // Never suggest
  }
}
```

### Configuration Schema
- Added proper JSON schema with `oneOf` constraints supporting mixed boolean/string types
- Comprehensive validation and error handling for when-clause parsing
- Experimental tag with clear documentation and examples

### UI Integration
- Enhanced `ChatWidget` with intelligent prompt suggestion logic
- Added description caching to avoid async calls during rendering
- Graceful fallback when descriptions fail to load

### Welcome View Improvements
- Moved suggested prompts rendering outside experimental-only blocks
- Updated re-render logic to account for prompt suggestion changes
- Added helpful onboarding message when no agent-specific welcome exists

## Testing

- ✅ Configuration parsing handles various input types correctly
- ✅ When-clause evaluation integrates properly with VS Code context system
- ✅ Error handling gracefully degrades on invalid configurations
- ✅ UI updates correctly when suggestions change

## Related Issues

Enhances the reusable prompts feature by making suggestions context-aware and user-configurable.

## Breaking Changes

None - this is an additive feature with sensible defaults.

### DIFF
--- a/src/vs/workbench/contrib/chat/browser/chat.contribution.ts
+++ b/src/vs/workbench/contrib/chat/browser/chat.contribution.ts
@@ -577,21 +577,15 @@ configurationRegistry.registerConfiguration({
 		[PromptsConfig.PROMPT_FILES_SUGGEST_KEY]: {
 			type: 'object',
 			title: nls.localize(
-				'chat.promptFilesSuggest.title',
-				"Prompt File Suggestions",
+				'chat.promptFilesRecommendations.title',
+				"Prompt File Recommendations",
 			),
 			markdownDescription: nls.localize(
-				'chat.promptFilesSuggest.description',
-				"Configure which prompt files to suggest in different contexts. Each key is a prompt file name, and the value can be `true` to always suggest, `false` to never suggest, or a when clause expression like `resourceExtname == .js` or `resourceLangId == markdown`.",
+				'chat.promptFilesRecommendations.description',
+				"Configure which prompt files to recommend in the chat welcome view. Each key is a prompt file name, and the value can be `true` to always recommend, `false` to never recommend, or a [when clause](https://aka.ms/vscode-when-clause) expression like `resourceExtname == .js` or `resourceLangId == markdown`.",
 			),
 			default: {},
 			additionalProperties: {
-				oneOf: [
-					{ type: 'boolean' },
-					{ type: 'string' }
-				]
-			},
-			unevaluatedProperties: {
 				oneOf: [
 					{ type: 'boolean' },
 					{ type: 'string' }
--- a/src/vs/workbench/contrib/chat/browser/chat.contribution.ts
+++ b/src/vs/workbench/contrib/chat/browser/chat.contribution.ts
@@ -574,6 +574,38 @@ configurationRegistry.registerConfiguration({
 			disallowConfigurationDefault: true,
 			tags: ['experimental', 'prompts', 'reusable prompts', 'prompt snippets', 'instructions']
 		},
+		[PromptsConfig.PROMPT_FILES_SUGGEST_KEY]: {
+			type: 'object',
+			title: nls.localize(
+				'chat.promptFilesSuggest.title',
+				"Prompt File Suggestions",
+			),
+			markdownDescription: nls.localize(
+				'chat.promptFilesSuggest.description',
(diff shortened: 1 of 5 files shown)


src/vs/workbench/contrib/chat/browser/chat.contribution.ts
```diff
@@ -543,6 +543,38 @@ configurationRegistry.registerConfiguration({
...
+				"Configure which prompt files to suggest in different contexts. Each key is a prompt file name, and the value can be `true` to always suggest, `false` to never suggest, or a when clause expression like `resourceExtname == .js` or `resourceLangId == markdown`.",
+			),
+			default: {},
+			additionalProperties: {
+				oneOf: [
+					{ type: 'boolean' },
+					{ type: 'string' }
+				]
+			},
+			unevaluatedProperties: {
```
aeschli: you only need `additionalProperties`, that covers all properties

aeschli (commented):
Can we use a different name than `promptFilesSuggest`. We use the term suggestions in the context of code completions.

Maybe 'recommendations'? Or 'favorites'?

- The setting allows to use context keys in user settings. That might be a first.  Context keys are known to extension authors but typically not to regular users. We have a documentation if it, but it's in the extension section. There's no while editing in the  settings editor and users might struggle to debug. Do we need the full power?

- Is there listening for prompt description changes or context changes?

digitarald:
I added a link to the when clause in the settings. Non-extension devs users are also exposed to them for customizing keyboard shortcuts, which is a primary use case.

Given that this is meant to be a out of the box customization driven by power users for their other maintainers; it should be ok to not live-update descriptions, but biasing towards low performance impact and keep iterating on how this is being adopted.

justschen:
just a thought, if we have way to categorize suggestions, these icons could maybe be different to reflect that, but not sure how possible that is. 

Screenshot 2025-08-26 at 1 48 42 PM

digitarald:
> just a thought, if we have way to categorize suggestions, these icons could maybe be different to reflect that, but not sure how possible that is.

Havinc icons for prompts and modes would be neat, I agree. Locally I dropped in a play button and it makes it even more "clickable".

src/vs/workbench/contrib/chat/browser/chatWidget.ts
```diff
@@ -807,7 +810,15 @@ export class ChatWidget extends Disposable implements IChatWidget {
 
 			let welcomeContent: IChatViewWelcomeContent;
 			const defaultAgent = this.chatAgentService.getDefaultAgent(this.location, this.input.currentModeKind);
-			const additionalMessage = defaultAgent?.metadata.additionalWelcomeMessage;
+			let additionalMessage = defaultAgent?.metadata.additionalWelcomeMessage;
```
bhavyaus, line 813: should these be surfaced in `Ask` mode also? If not, we should move this further down into the `if(expEmtpyState)` block

bhavyaus: The suggested prompts always run the command in Agent mode fyi

digitarald: I am ok to show them only in agent mode as a first cut, as its the default. It is meant to be contextual to what users have open.

digitarald: Ok, always showing the prompts now (can help to switch more between modes as needed)

bhavyaus (approved):

PR #263373, opened 2025-08-26
Fix terminal link disposable leak 

### LINKED ISSUE #252302: LEAKED DISPOSABLE
```
[LEAKED DISPOSABLE] Error: CREATED via:
    at GCBasedDisposableTracker.trackDisposable (vscode-file://vscode-app/Users/bpasero/Development/Microsoft/vscode-electron-22.x/out/vs/base/common/lifecycle.js:27:23)
    at trackDisposable (vscode-file://vscode-app/Users/bpasero/Development/Microsoft/vscode-electron-22.x/out/vs/base/common/lifecycle.js:198:24)
    at new Disposable (vscode-file://vscode-app/Users/bpasero/Development/Microsoft/vscode-electron-22.x/out/vs/base/common/lifecycle.js:393:9)
    at new TerminalLink (vscode-file://vscode-app/Users/bpasero/Development/Microsoft/vscode-electron-22.x/out/vs/workbench/contrib/terminalContrib/links/browser/terminalLink.js:25:9)
    at InstantiationService._createInstance (vscode-file://vscode-app/Users/bpasero/Development/Microsoft/vscode-electron-22.x/out/vs/platform/instantiation/common/instantiationService.js:130:24)
    at InstantiationService.createInstance (vscode-file://vscode-app/Users/bpasero/Development/Microsoft/vscode-electron-22.x/out/vs/platform/instantiation/common/instantiationService.js:101:27)
    at TerminalLinkDetectorAdapter._createTerminalLink (vscode-file://vscode-app/Users/bpasero/Development/Microsoft/vscode-electron-22.x/out/vs/workbench/contrib/terminalContrib/links/browser/terminalLinkDetectorAdapter.js:86:43)
    at TerminalLinkDetectorAdapter._provideLinks (vscode-file://vscode-app/Users/bpasero/Development/Microsoft/vscode-electron-22.x/out/vs/workbench/contrib/terminalContrib/links/browser/terminalLinkDetectorAdapter.js:76:29)
    at async TerminalLinkDetectorAdapter.provideLinks (vscode-file://vscode-app/Users/bpasero/Development/Microsoft/vscode-electron-
```

anthonykim1:
Resolves: https://github.com/microsoft/vscode/issues/252302

### DIFF
--- a/src/vs/workbench/contrib/terminalContrib/links/browser/terminalLinkDetectorAdapter.ts
+++ b/src/vs/workbench/contrib/terminalContrib/links/browser/terminalLinkDetectorAdapter.ts
@@ -4,7 +4,7 @@
  *--------------------------------------------------------------------------------------------*/
 
 import { Emitter } from '../../../../../base/common/event.js';
-import { Disposable, DisposableStore } from '../../../../../base/common/lifecycle.js';
+import { Disposable } from '../../../../../base/common/lifecycle.js';
 import { localize } from '../../../../../nls.js';
 import { IInstantiationService } from '../../../../../platform/instantiation/common/instantiation.js';
 import { ITerminalLinkDetector, ITerminalSimpleLink, TerminalBuiltinLinkType, TerminalLinkType } from './links.js';
@@ -29,7 +29,6 @@ export interface IShowHoverEvent {
  */
 export class TerminalLinkDetectorAdapter extends Disposable implements ILinkProvider {
 	private _activeLinks: TerminalLink[] | undefined;
-	private readonly _activeLinksStore = this._register(new DisposableStore());
 
 	private readonly _onDidActivateLink = this._register(new Emitter<IActivateLinkEvent>());
 	readonly onDidActivateLink = this._onDidActivateLink.event;
@@ -51,12 +50,13 @@ export class TerminalLinkDetectorAdapter extends Disposable implements ILinkProv
 			callback(this._activeLinks);
 			return;
 		}
-		if (this._activeLinks) {
+
+		// Only dispose previous links if no other requests are pending
+		if (this._activeLinks && this._activeProvideLinkRequests.size === 0) {
 			for (const link of this._activeLinks) {
 				link.dispose();
 			}
 		}
-		this._activeLinksStore.clear();
 		activeRequest = this._provideLinks(bufferLineNumber);
 		this._activeProvideLinkRequests.set(bufferLineNumber, activeRequest);
 		this._activeLinks = await activeRequest;
@@ -95,9 +95,7 @@ export class TerminalLinkDetectorAdapter extends Disposable implements ILinkProv
 
 		const detectedLinks = await this._detector.detect(lines, startLine, endLine);
 		for (const link of detectedLinks) {
-			const terminalLink = this._createTerminalLink(link, async (event) => this._onDidActivateLink.fire({ link, event }));
-			links.push(terminalLink);
-			this._activeLinksStore.add(terminalLink);
+			links.push(this._createTerminalLink(link, async (event) => this._onDidActivateLink.fire({ link, event })));
 		}
 
 		return links;
--- a/src/vs/workbench/contrib/terminalContrib/links/browser/terminalLinkDetectorAdapter.ts
+++ b/src/vs/workbench/contrib/terminalContrib/links/browser/terminalLinkDetectorAdapter.ts
@@ -4,7 +4,7 @@
  *--------------------------------------------------------------------------------------------*/
 
 import { Emitter } from '../../../../../base/common/event.js';
-import { Disposable } from '../../../../../base/common/lifecycle.js';
+import { Disposable, DisposableStore } from '../../../../../base/common/lifecycle.js';
 import { localize } from '../../../../../nls.js';
 import { IInstantiationService } from '../../../../../platform/instantiation/common/instantiation.js';
 import { ITerminalLinkDetector, ITerminalSimpleLink, TerminalBuiltinLinkType, TerminalLinkType } from './links.js';
@@ -29,6 +29,7 @@ export interface IShowHoverEvent {
  */
 export class TerminalLinkDetectorAdapter extends Disposable implements ILinkProvider {
 	private _activeLinks: TerminalLink[] | undefined;
+	private readonly _activeLinksStore = this._register(new DisposableStore());
 
 	private readonly _onDidActivateLink = this._register(new Emitter<IActivateLinkEvent>());
 	readonly onDidActivateLink = this._onDidActivateLink.event;
@@ -50,11 +51,8 @@ export class TerminalLinkDetectorAdapter extends Disposable implements ILinkProv
 			callback(this._activeLinks);
 			return;
 		}
-		if (this._activeLinks) {
-			for (const link of this._activeLinks) {
-				link.dispose();
-			}
-		}
+		// Clear previous links from the store
+		this._activeLinksStore.clear();
 		activeRequest = this._provideLinks(bufferLineNumber);
 		this._activeProvideLinkRequests.set(bufferLineNumber, activeRequest);
 		this._activeLinks = await activeRequest;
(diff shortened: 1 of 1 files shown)


src/vs/workbench/contrib/terminalContrib/links/browser/terminalLinkDetectorAdapter.ts
```diff
@@ -93,9 +95,10 @@ export class TerminalLinkDetectorAdapter extends Disposable implements ILinkProv
 
 		const detectedLinks = await this._detector.detect(lines, startLine, endLine);
 		for (const link of detectedLinks) {
-			links.push(this._createTerminalLink(link, async (event) => this._onDidActivateLink.fire({ link, event })));
+			const terminalLink = this._createTerminalLink(link, async (event) => this._onDidActivateLink.fire({ link, event }));
+			links.push(terminalLink);
+			this._activeLinksStore.add(terminalLink);
```
anthonykim1, line 94: I couldn't repro the leak, but looking at the trace https://github.com/microsoft/vscode/issues/252302, here is my hypothesis:
 
--> The "active links" that gets added into `_activeLinks` may NOT get disposed when the `TerminalLinkDetectorAdapter` itself gets disposed. 

What I did:
* Created `DisposableStore` to add in all these links the moment they are added to `_activeLinks`, so when adapter gets disposed, all of these links would get disposed too.

Concern:
* I wanted to avoid having both `_activeLinks` and `_activeLinksStore`, but wasn't so easy just converting `_activeLinks` into DisposableStore itself since https://github.com/microsoft/vscode/blob/bb5e45ae593ba041df740974f24041b2ca6895f9/src/vs/workbench/contrib/terminalContrib/links/browser/terminalLinkDetectorAdapter.ts#L60-L64 already exists and felt wrong to change the type of `_activeLinks` into DisposableStore.

Tyriar: This is basically happening already here:

https://github.com/microsoft/vscode/blob/ee1ec076b8baa7c8f20163d00fca1959fecbbc42/src/vs/workbench/contrib/terminalContrib/links/browser/terminalLinkDetectorAdapter.ts#L53-L57

I think the issue is actually around the active request promise here, where new requests for different lines will abandon the old links:

https://github.com/microsoft/vscode/blob/ee1ec076b8baa7c8f20163d00fca1959fecbbc42/src/vs/workbench/contrib/terminalContrib/links/browser/terminalLinkDetectorAdapter.ts#L47-L57

Following that I actually got a really easy repro case, just add `await timeout(1000);` before the return in `_provideLinks`. That will simulate really slow link fetching and cause a lot of dispose warnings 🎉

Tyriar (changes requested):

Tyriar (dismissed):
lgtm once you verify

anthonykim1:
Managed to get the leak again, pretty easily actually by typing pretty fast :/

https://github.com/user-attachments/assets/e4e1f2fd-516f-4e8a-8d2c-331563a75e82


Will re-investigate..!

src/vs/workbench/contrib/terminalContrib/links/browser/terminalLinkDetectorAdapter.ts
```diff
@@ -28,7 +28,7 @@ export interface IShowHoverEvent {
  * Wrap a link detector object so it can be used in xterm.js
  */
 export class TerminalLinkDetectorAdapter extends Disposable implements ILinkProvider {
-	private _activeLinks: TerminalLink[] | undefined;
+	private readonly _activeLinksStore = this._register(new DisposableStore());
```
anthonykim1, line 31: Got rid of `_activeLinks` and added `_activeLinksStore` which is a disposable store. 
This is where I add all the links after creating them via `_createTerminalLink` below.

src/vs/workbench/contrib/terminalContrib/links/browser/terminalLinkDetectorAdapter.ts
```diff
@@ -46,20 +46,16 @@ export class TerminalLinkDetectorAdapter extends Disposable implements ILinkProv
 	async provideLinks(bufferLineNumber: number, callback: (links: ILink[] | undefined) => void) {
 		let activeRequest = this._activeProvideLinkRequests.get(bufferLineNumber);
 		if (activeRequest) {
-			await activeRequest;
-			callback(this._activeLinks);
+			const links = await activeRequest;
```
anthonykim1, line 49: I'm bit lost how before, we used `this._activeLinks` instead of result of `await activeRequest` (which is type `Promise<TerminalLinks[]>`, for callback. I might be missing some context though, 

By using `callback(links)`, where `await activeRequests` we won't need `_activeLinks` anymore and can just rely on `activeLinksStore`

anthonykim1:
@Tyriar  The current version is the only one that seem to "work" --> I couldn't repro disposable leak. 

What completely doesn't make sense is before, why we called `await activeRequest`, but then `callback(this._activeLinks)` instead of just `callback(await activeRequest)` 

Screenshot 2025-08-26 at 4 12 41 PM

src/vs/workbench/contrib/terminalContrib/links/browser/terminalLinkDetectorAdapter.ts
```diff
@@ -46,20 +46,16 @@ export class TerminalLinkDetectorAdapter extends Disposable implements ILinkProv
...
-			for (const link of this._activeLinks) {
-				link.dispose();
-			}
-		}
+		this._activeLinksStore.clear();
 		activeRequest = this._provideLinks(bufferLineNumber);
 		this._activeProvideLinkRequests.set(bufferLineNumber, activeRequest);
-		this._activeLinks = await activeRequest;
+		const links = await activeRequest;
 		this._activeProvideLinkRequests.delete(bufferLineNumber);
```
anthonykim1, line 57: I'm also wondering purpose of `_activeProvideLinkRequests`, maybeee it is somehow related to concurrency?

It seems like `_activeProvideLinkRequests` only gets referenced inside of this `provideLinks` function, and kind of confusing to `_activeProvideLinkRequests.set` and then `_activeProvideLinkRequests.delete` right afterwards 🧐🥲

Tyriar: It's possible this isn't needed and could have been related to caching that never ended up happening. I'd be careful about changing this stuff to clean up though without a full understanding of how links interact with xterm.js. For example, tracking different rows probably prevents duplicate requests when hovering multiple rows especially when there's a remote connection and it's slower to verify the links.

src/vs/workbench/contrib/terminalContrib/links/browser/terminalLinkDetectorAdapter.ts
Tyriar: Your proposal to use the return value does seem more correct, just make sure it's tested well

src/vs/workbench/contrib/terminalContrib/links/browser/terminalLinkDetectorAdapter.ts
Tyriar: Good change 👍

Tyriar (approved):

src/vs/workbench/contrib/terminalContrib/links/browser/terminalLinkDetectorAdapter.ts
anthonykim1: Just did more manual testing more via web link, `ls` and clicking on some different file links, spamming like above.

In terms of test files, it seems like `terminalManager` instantiates `terminalLinkDetectorAdapter` and [getLinks](https://github.com/microsoft/vscode/blob/db87c97754cef3a2ba89989caf281f8268dedc64/src/vs/workbench/contrib/terminalContrib/links/test/browser/terminalLinkManager.test.ts#L177) calls `provideLinks`, where provideLinks is what we want to test.

I don't see dedicated test file for https://github.com/microsoft/vscode/blob/db87c97754cef3a2ba89989caf281f8268dedc64/src/vs/workbench/contrib/terminalContrib/links/browser/terminalLinkDetectorAdapter.ts itself though, should I add one?

Or maybe that is intended since its an adapter.

Tyriar: You could try asking Copilot to try write some unit tests and see if they're useful? I never tried since it's the adapter that integrates with xterm.js.

PR #263403, opened 2025-08-26
Adding font info in colorTokenCustomization

aiday-mar:


src/vs/editor/common/model/tokens/tokenizationFontDecorationsProvider.ts
```diff
@@ -0,0 +1,167 @@
...
+		}));
+	}
+
+	public handleDidChangeContent(change: IModelContentChangedEvent) {
+		let edits: StringEdit = StringEdit.empty;
+		for (const c of change.changes) {
+			const offsetRange = new OffsetRange(c.rangeOffset, c.rangeOffset + c.rangeLength);
+			const stringEdit = StringEdit.replace(offsetRange, c.text);
+			edits = edits.compose(stringEdit);
+		}
```
hediet: There is a helper for this somewhere

src/vs/editor/common/model/tokens/annotations.ts
```diff
@@ -0,0 +1,260 @@
...
+			}
+		}
+		return endIndexExclusive;
+	}
+
+	public getAllAnnotations(): IAnnotation<T>[] {
+		return this._annotations.slice();
+	}
+
+	public applyEdit(edit: StringEdit): IAnnotation<T>[] {
```
hediet, line 127: Please add a comment that describes the result (and to what document state their ranges refer to).

src/vs/workbench/services/textMate/browser/backgroundTokenization/worker/textMateWorkerTokenizer.ts
```diff
@@ -157,10 +163,13 @@ export class TextMateWorkerTokenizer extends MirrorTextModel {
 				break;
 			}
 
+			const fontUpdate = AnnotationsUpdate.create<IFontTokenOption>(fontTokensUpdate);
+			const serializedFontUpdate = AnnotationsUpdate.serialize<IFontTokenOption>(fontUpdate, serializeFontTokenOptions());
```
hediet:
```
fontUpdate.serialize(...)
```

src/vs/workbench/services/textMate/browser/backgroundTokenization/textMateWorkerTokenizerController.ts
```diff
@@ -170,14 +174,21 @@ export class TextMateWorkerTokenizerController extends Disposable {
...
+					const text = innerChanges.text;
 					for (let j = 0; j < tokens.length; j++) {
-						tokens[j].applyEdit(innerChanges.range, innerChanges.text);
+						tokens[j].applyEdit(range, text);
 					}
+					const offsetEditStart = this._model.getOffsetAt(range.getStartPosition());
+					const offsetEditEnd = this._model.getOffsetAt(range.getEndPosition());
+					edits.push(StringEdit.replace(new OffsetRange(offsetEditStart, offsetEditEnd), text));
 				}
 			}
```
hediet: Please do this in a separate loop/function, maybe `const pendingStringEdit = stringEditFrom???(this._pendingChanges)`

hediet (changes requested):
Looks good! I left some comments.
Did you add tests for `annotations.ts` (we need lots of them)?

src/vs/editor/common/model/tokens/tokenizationFontDecorationsProvider.ts
aiday-mar: I guess you are referring to the function `offsetEditFromContentChanges`

src/vs/workbench/services/textMate/browser/backgroundTokenization/worker/textMateWorkerTokenizer.ts
aiday-mar: The reason I made it a static method, is because the AnnotationsUpdate contains two methods:

```
static serialize<T>(update: AnnotationsUpdate<T>, serializingFunc: (annotation: T | undefined) => ISerializedProperty): ISerializedAnnotation[];

static deserialize<T>(serializedAnnotations: ISerializedAnnotation[], deserializingFunc: (annotation: ISerializedProperty) => T): AnnotationsUpdate<T>;
```

The first of which could be a normal non-static method where we could access the current annotations field. The method would have only one argument: the serializing function. The second could be a normal non-static method, but it would have two arguments, the serialized annotations and the deserializing function. If I change both to non-static methods, they would no longer mirror each other. It also feels like it makes more sense to have the deserializing function be a static method, then a non-static method.

hediet: deserialize must be static, yes, but serialize not. It's a common pattern.
This is much more readable and discoverable:
```
const serializedFontUpdate = fontUpdate.serialize(serializeFontTokenOptions());
```

hediet (dismissed):

dbaeumer (dismissed):

bpasero (approved):

PR #263410, opened 2025-08-26
allow policy config to compute the policy value for account based policies

sandy081:


src/vs/base/common/policy.ts
```diff
@@ -3,14 +3,9 @@
...
 
-export type PolicyName = string;
+import { IDefaultAccount } from './defaultAccount.js';
 
-export enum PolicyTag {
-	Account = 'ACCOUNT',
-	MCP = 'MCP',
-	Preview = 'PREVIEW',
-	Agent = 'AGENT',
-}
```
Tyriar, line 13: FYI @joshspicer

Tyriar (approved):

sandy081 (commented):
I pinged Josh in the Slack already. Thanks for the ping though

PR #263438, opened 2025-08-26
add output monitoring for foreground terminals

### LINKED ISSUE #262933: add auto-reply for foreground terminals
We added support for auto-reply in the terminal for background terminals here https://github.com/microsoft/vscode/pull/262782

We also want to add support for this in foreground terminals

meganrogge:
fixes #262933 

Testing:
Rich ✅ 
Basic ✅ 
None ✅


scripts/confirm-multiple.ps1
```diff
@@ -0,0 +1,26 @@
+$title    = 'File Operation'
```
Tyriar: I wouldn't include these in the main repo, in some personal repo should be fine and you could add that to your PATH to make it convenient to use

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/executeStrategy/basicExecuteStrategy.ts
```diff
@@ -84,10 +89,11 @@ export class BasicExecuteStrategy implements ITerminalExecuteStrategy {
...
 			// the cursor is. This can happen in prompts where they clear the line and rerender it
 			// like powerlevel10k's transient prompt
-			let startMarker = store.add(xterm.raw.registerMarker());
-			store.add(startMarker.onDispose(() => {
+			this._startMarker = store.add(xterm.raw.registerMarker());
+			this._onDidCreateStartMarker.fire(this._startMarker);
+			store.add(this._startMarker.onDispose(() => {
 				this._log(`Start marker was disposed, recreating`);
-				startMarker = xterm.raw.registerMarker();
+				this._startMarker = xterm.raw.registerMarker();
```
Tyriar: Considering it can change, let's make the even `onDidChangeStartMarker` and refire here

Tyriar (dismissed):

Tyriar (approved):

PR #263458, opened 2025-08-26
fix: parse commands in chat.open commands

rwoll:
`workbench.action.chat.open` did not always trigger a parse of the input resulting in `/<prompt>` not being recognized as a prompt command when it's expected to.

If `isPartialQuery: false` (default), then the command used [`acceptInput` intead of `setInput`](https://github.com/microsoft/vscode/blob/709b8fabf55c505cbe5603ce654ecf3a8766a7a6/src/vs/workbench/contrib/chat/browser/actions/chatActions.ts#L244) only the latter of which will trigger a [re-parse of the query](https://github.com/microsoft/vscode/blob/709b8fabf55c505cbe5603ce654ecf3a8766a7a6/src/vs/workbench/contrib/chat/browser/chatWidget.ts#L1640) to critcially turn [`/<prompt>` into a `ChatRequestSlashPromptPart`](https://github.com/microsoft/vscode/blob/d2ff8d16065d4748b8a205bfebdc98d7d76f7864/src/vs/workbench/contrib/chat/common/chatRequestParser.ts#L231).

**Reviewers**: There's a few places `acceptInput` is called with a query in the codebase. It's not clear to me if the others have the same issue or if always triggering a re-parse would double parse.

### DIFF
--- a/src/vs/workbench/contrib/chat/browser/actions/chatActions.ts
+++ b/src/vs/workbench/contrib/chat/browser/actions/chatActions.ts
@@ -236,14 +236,15 @@ abstract class OpenChatGlobalAction extends Action2 {
 		let resp: Promise<IChatResponseModel | undefined> | undefined;
 
 		if (opts?.query) {
-			if (opts.isPartialQuery) {
-				chatWidget.setInput(opts.query);
-			} else {
+			chatWidget.setInput(opts.query);
+
+			if (!opts.isPartialQuery) {
 				await chatWidget.waitForReady();
 				await waitForDefaultAgent(chatAgentService, chatWidget.input.currentModeKind);
-				resp = chatWidget.acceptInput(opts.query);
+				resp = chatWidget.acceptInput();
 			}
 		}
+
 		if (opts?.toolIds && opts.toolIds.length > 0) {
 			for (const toolId of opts.toolIds) {
 				const tool = toolsService.getTool(toolId);


src/vs/workbench/contrib/chat/browser/actions/chatActions.ts
```diff
@@ -236,14 +236,15 @@ abstract class OpenChatGlobalAction extends Action2 {
 		let resp: Promise<IChatResponseModel | undefined> | undefined;
 
 		if (opts?.query) {
-			if (opts.isPartialQuery) {
-				chatWidget.setInput(opts.query);
-			} else {
+			chatWidget.setInput(opts.query);
```
rwoll, line 239: I need to double check I didn't just break `isPartialQuery` since `setInput` likely overrides the whole input.

rwoll: I in fact did regress that. Putting into Draft until I've addressed it properly. /cc @roblourens

rwoll: Okie - all is good. It appears to have the same `isPartialQuery` as before:

```
REQ 1 (isPartialQuery: true) "hello, "
REQ 2 (isPartialQuery: false) "world"
```

results in `world`.

Multiple `isPartialQuery: true` in a row did not and do not allow for additive query text message construction.

roblourens: Looks like you end up calling `setInput` the same as it was before in this case, LGTM

roblourens (approved):

benibenj (approved):

PR #263507, opened 2025-08-27
`workbench.action.chat.export` optionally accept path

rwoll:
Supports programmatically exporting without triggering a Save As…

### DIFF
--- a/src/vs/workbench/contrib/chat/browser/actions/chatImportExport.ts
+++ b/src/vs/workbench/contrib/chat/browser/actions/chatImportExport.ts
@@ -34,7 +34,7 @@ export function registerChatExportActions() {
 				f1: true,
 			});
 		}
-		async run(accessor: ServicesAccessor, outputPath?: URI | string) {
+		async run(accessor: ServicesAccessor, outputPath?: URI) {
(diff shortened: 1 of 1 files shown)


src/vs/workbench/contrib/chat/browser/actions/chatImportExport.ts
```diff
@@ -44,13 +45,26 @@ export function registerChatExportActions() {
...
-				filters
-			});
-			if (!result) {
-				return;
+			let finalOutputPath: URI;
+
+			if (outputPath) {
+				if (URI.isUri(outputPath)) {
+					finalOutputPath = outputPath;
+				} else if (typeof outputPath === 'string') {
```
roblourens: Unless you have a specific reason to support a string, I think we should just not support it, because then someone will use it incorrectly (eg in remote windows, when we transfer a URI from the EH to the renderer process, it gets mapped to the right format automatically, but a stringified URI will end up wrong)

rwoll: Updated!

amunger (approved):

Tyriar (approved):

PR #263583, opened 2025-08-27
skip terrapin check for PRs that don't touch package files

joaomoreno:
This allows us to set this pipeline as a required check.

cc @rebornix


Tyriar (approved):
👏

PR #263636, opened 2025-08-27
hide prompt on data input

### LINKED ISSUE #263474: When a user manually replies to input requested, hide confirmation prompt and continue
Request from @roblourens

meganrogge:
fixes https://github.com/microsoft/vscode/issues/263474

https://github.com/user-attachments/assets/03f46770-4d69-4dda-bad9-876fe70c8831


src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/monitoring/outputMonitor.ts
```diff
@@ -506,6 +507,13 @@ export class OutputMonitor extends Disposable implements IOutputMonitor {
 							run: async () => { }
 						}))
 					));
+					const inputDataDisposable = this._register(execution.instance.onDidInputData(() => {
```
Tyriar, line 510: Does this event ignored the focus in/out events? Click in and out of the terminal when the prompt is up to double check. If not we should filter those out like we're doing here:

https://github.com/microsoft/vscode/blob/a18d41e90a83fbe551fd7db4894a54cb7d436966/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/runInTerminalTool.ts#L417-L419

meganrogge: Yes, focus in /out is ignored

Tyriar (approved):

PR #263696, opened 2025-08-27
Disallow invalid states for `ChatSessionItem`

mjbvz:
This combines the state info into a single type so that you cannot create invalid states, such as `InProgress` with an endTime

In the future this can also be used to more neatly add other metadata, such as a failure reason string for failed sessions


src/vscode-dts/vscode.proposed.chatSessionsProvider.d.ts
```diff
@@ -85,29 +85,20 @@ declare module 'vscode' {
...
-			startTime: number;
-			/**
-			 * Session end timestamp in milliseconds elapsed since January 1, 1970 00:00:00 UTC.
-			 */
-			endTime?: number;
-		};
+		state?:
+		| { status: ChatSessionStatus.InProgress; timing?: { startTime: number } }
+		| { status: ChatSessionStatus.Completed; timing?: { startTime: number; endTime: number } }
+		| { status: ChatSessionStatus.Failed; timing?: { startTime: number; endTime: number } };
```
mjbvz: This union type is kind of ugly. Instead I wonder if we should introduce helper state classes like

```
state: ChatSessionCompletedState | ChatSessionFailedState | ChatSesssionInProgressState;

...

class ChatSesssionInProgressState {
 constructor(public timing? ...); 
}

...

```

src/vscode-dts/vscode.proposed.chatSessionsProvider.d.ts
```diff
@@ -5,23 +5,63 @@
 
 declare module 'vscode' {
```
mjbvz, line 8: @osortega Let me know if we should version this API to make it easier to avoid breaking copilot chat when making changes like this

src/vscode-dts/vscode.proposed.chatSessionsProvider.d.ts
```diff
@@ -3,25 +3,67 @@
...
-		 * The chat session failed to complete.
+		 * Optional  execution timing information about the session.
 		 */
-		Failed = 0,
+		timing?: {
+			/**
+			 * The time the session started as a unix timestamp.
+			 */
+			startTime: number;
+		};
```
osortega, line 26: Do we need to add endtime here? Or only for final states?

src/vscode-dts/vscode.proposed.chatSessionsProvider.d.ts
```diff
@@ -90,29 +132,12 @@ declare module 'vscode' {
...
-		 * An optional status indicating the current state of the session.
-		 */
-		status?: ChatSessionStatus;
-
 		/**
 		 * The tooltip text when you hover over this item.
 		 */
 		tooltip?: string | MarkdownString;
 
-		/**
```
osortega, line 103: Removing this creates a bit of ambiguity on the time to use in the chat session view.
Right now it's used as the time the pull request was created at, but with this new proposal do we select the in progress state to fetch the timing from? What happens if we come to a point where copilot takes a while to pick up a task and we don't change to "in progress" state for a while

src/vscode-dts/vscode.proposed.chatSessionsProvider.d.ts
mjbvz: Only for final states. My goal is to disallow invalid states, such as in-progress with an end time. I don't think that should ever happen

src/vscode-dts/vscode.proposed.chatSessionsProvider.d.ts
mjbvz: hm, this is a good example of why the current API is confusing 😆 

I think we could have a `createdTime` if we really want to track when the session is created. However `startTime` to me implies when the session actually started running

What info here do you think you need for the UI:

- Created time?
- Start and end times?
- Run time as a duration?

Once we know that we can work backwards to figure out the api shape

osortega: What we do today is we take the `createdAt` time from the PR and use that to render the time in the view: https://github.com/microsoft/vscode-pull-request-github/blob/eaa7c9fe0ebbe8d16479dcf962c52dbbc850980f/src/github/copilotRemoteAgent.ts#L917

But now that I look at it, we are also passing the status so we could just send the `createdAt` timestamp on the first status we report and use that in core to render the time in the view. Not sure if that is a strange way for core to implement the API though

mjbvz: So do you think we should have a top-level `createdAt` field on the `ChatSessionItem`? We can still use the `startedAt`/`endedAt` to compute duration

PR #263702, opened 2025-08-27
use stripped down version of PSReadline for screen reader support on windows to enable pwsh shell integration 

### LINKED ISSUE #233921: terminal shell integration doesn't work for screen reader users sometimes
From @jooyoungseo

waiting for the logs

Image

meganrogge:
fix #233921


src/vs/platform/terminal/node/terminalEnvironment.ts
```diff
@@ -63,7 +63,7 @@ export async function getShellIntegrationInjection(
 	env: ITerminalEnvironment | undefined,
 	logService: ILogService,
 	productService: IProductService,
-	skipStickyBit: boolean = false
+	skipStickyBit: boolean = false,
```
meganrogge:
```suggestion
	skipStickyBit: boolean = false
```

src/vs/workbench/contrib/terminal/common/scripts/shellIntegration.ps1
```diff
@@ -169,6 +169,34 @@ elseif ((Test-Path variable:global:GitPromptSettings) -and $Global:GitPromptSett
 	[Console]::Write("$([char]0x1b)]633;P;PromptType=posh-git`a")
 }
 
+function Global:__VSCode-TestScreenReader {
+	# Detect VS Code screen reader mode from common env vars (treat: 1/true/enabled as on)
```
meganrogge:
```suggestion
```

src/vs/workbench/contrib/terminal/common/psreadline/PSReadLine.psd1
Tyriar, line 1: Let's stick these inside the `scripts/` folder

src/vs/workbench/contrib/terminal/common/psreadline/License.txt
```diff
@@ -0,0 +1,23 @@
+Copyright (c) 2013, Jason Shirk
```
Tyriar: This would need to integrate into a cgmanifest. Example here:
 https://github.com/microsoft/vscode/blob/b70d0fc99f58928e12d07c78013916cdd5046485/src/vs/base/browser/dompurify/cgmanifest.json#L1-L18

src/vs/workbench/contrib/terminal/common/psreadline/Changes.txt
```diff
@@ -0,0 +1,861 @@
+### [2.4.3-beta3] - 2025-07-23
```
Tyriar: We should verify this one isn't getting published into our product build

src/vs/workbench/contrib/terminal/common/scripts/shellIntegration.ps1
```diff
@@ -169,6 +169,32 @@ elseif ((Test-Path variable:global:GitPromptSettings) -and $Global:GitPromptSett
 	[Console]::Write("$([char]0x1b)]633;P;PromptType=posh-git`a")
 }
 
+function Global:__VSCode-TestScreenReader {
+	return ($env:VSCODE_ACCESSIBILITY_SUPPORT -match 'true')
+}
```
Tyriar: Inline this? We want to avoid globals from SI as much as possible as it contaminates the shell session

src/vs/workbench/contrib/terminal/common/scripts/shellIntegration.ps1
```diff
@@ -169,6 +169,32 @@ elseif ((Test-Path variable:global:GitPromptSettings) -and $Global:GitPromptSett
...
+
+try {
```
Tyriar: Just above this block is a good place to explain what this version of PSReadLine is here for

Tyriar: I think the try could be more localized? I think the main thing it needs to wrap is `Import-Module`, including other things could hide unexpected bugs that break this support. We may also want to not have a try here at all, especially if we can attempt to import and continue then errors printing would be very useful to stabilize

src/vs/workbench/contrib/terminal/common/scripts/shellIntegration.ps1
```diff
@@ -169,6 +169,32 @@ elseif ((Test-Path variable:global:GitPromptSettings) -and $Global:GitPromptSett
...
+	if (-not (Get-Module -Name PSReadLine)) {
+		if ($PSVersionTable.PSVersion.Major -ge 7 -and (Global:__VSCode-TestScreenReader)) {
+			# Resolve the path relative to this script
+			$scriptRoot = $PSScriptRoot
+			if (-not $scriptRoot) { $scriptRoot = Split-Path -Parent $MyInvocation.MyCommand.Path }
```
Tyriar: Not sure when `-not $scriptRoot` could happen?

src/vs/workbench/contrib/terminal/common/scripts/shellIntegration.ps1
```diff
@@ -169,6 +169,32 @@ elseif ((Test-Path variable:global:GitPromptSettings) -and $Global:GitPromptSett
...
+
+			$specialPsrlPath = Join-Path $scriptRoot '..\psreadline\'
+
+			Import-Module $specialPsrlPath -ErrorAction Stop
```
Tyriar: Does `-ErrorAction Stop` means it'll brick the pwsh session? I don't think we want it to continue

src/vs/platform/terminal/node/terminalEnvironment.ts
```diff
@@ -112,6 +112,7 @@ export async function getShellIntegrationInjection(
 	}
 	// Windows
 	if (isWindows) {
+		envMixin['VSCODE_ACCESSIBILITY_SUPPORT'] = options.isScreenReaderOptimized.toString();
```
Tyriar: We should use `'1'` instead, see `VSCODE_STABLE`, `VSCODE_SUGGEST` (which needs to be removed and I missed https://github.com/microsoft/vscode/issues/263800)

build/filters.js
```diff
@@ -203,6 +206,7 @@ module.exports.tsFormattingFilter = [
 	'!extensions/html-language-features/server/lib/jquery.d.ts',
 	'!extensions/terminal-suggest/src/shell/zshBuiltinsCache.ts',
 	'!extensions/terminal-suggest/src/shell/fishBuiltinsCache.ts',
+	'!src/vs/workbench/contrib/terminal/common/psreadline/**'
```
Tyriar: Shouldn't be needed as there's no ts?

.eslint-ignore
```diff
@@ -38,4 +38,5 @@
 **/test/automation/out/**
 **/typings/**
 **/.build/**
+**/src/vs/workbench/contrib/terminal/common/psreadline/**
```
Tyriar: Shouldn't be needed as no js/ts?

extensions/package-lock.json
```diff
@@ -14,7 +14,7 @@
       },
       "devDependencies": {
         "@parcel/watcher": "2.5.1",
-        "esbuild": "0.25.0",
+        "esbuild": "^0.25.0",
```
Tyriar: Unrelated change?

extensions/package.json
```diff
@@ -11,7 +11,7 @@
   },
   "devDependencies": {
     "@parcel/watcher": "2.5.1",
-    "esbuild": "0.25.0",
+    "esbuild": "^0.25.0",
```
Tyriar: Unrelated change?

src/vs/platform/terminal/node/terminalEnvironment.ts
Tyriar: Usually env vars aren't this long, how about `VSCODE_A11Y_MODE`?

src/vs/workbench/contrib/terminal/common/psreadline/Microsoft.PowerShell.PSReadLine.dll
Tyriar, line 1: @andyleejordan any help here on what files we should be keep and which we should not?

@meganrogge some of these files will probably get removed without special rules in our `build/` product build packaging code

Tyriar (dismissed):

src/vs/workbench/contrib/terminal/common/scripts/shellIntegration.ps1
meganrogge: This means it will jump to the catch I believe, where we suppress errors.

meganrogge: Sounds like from the above comment, we want the errors to be propagated. So I'll remove this.

src/vs/workbench/contrib/terminal/common/scripts/shellIntegration.ps1
meganrogge: removed it

src/vs/workbench/contrib/terminal/common/scripts/shellIntegration.ps1
meganrogge: looks like only for lower versions of pwsh, removed

src/vs/workbench/contrib/terminal/common/psreadline/License.txt
meganrogge: done, just waiting on the commit to use from @andyleejordan

src/vs/workbench/contrib/terminal/common/psreadline/Changes.txt
meganrogge: I removed it

src/vs/workbench/contrib/terminal/common/psreadline/Microsoft.PowerShell.PSReadLine.dll
meganrogge: According to AI:

net6plus/…Polyfill.dll is used when PSReadLine runs on .NET 6+ (i.e., pwsh 7+ on macOS/Linux/Windows), **so we should keep it.**

netstd/…Polyfill.dll is for the legacy .NET Framework / Windows PowerShell 5.1 path. **(so I think we can remove it since we mandate pwsh 7?)**

We can remove `changes.txt` and `license.txt` as well.

src/vs/platform/terminal/node/terminalProcess.ts
```diff
@@ -147,7 +147,7 @@ export class TerminalProcess extends Disposable implements ITerminalChildProcess
 		private readonly _executableEnv: IProcessEnvironment,
 		private readonly _options: ITerminalProcessOptions,
 		@ILogService private readonly _logService: ILogService,
-		@IProductService private readonly _productService: IProductService
+		@IProductService private readonly _productService: IProductService,
```
meganrogge:
```suggestion
		@IProductService private readonly _productService: IProductService
```

src/vs/workbench/contrib/terminal/common/scripts/psreadline/cgmanifest.json
```diff
@@ -0,0 +1,38 @@
+{
+  "registrations": [
+    {
+      "component": {
+        "type": "git",
+        "git": {
+          "name": "PSReadLine",
+          "repositoryUrl": "https://github.com/PowerShell/PSReadLine",
+          "commitHash": "abc123"
```
meganrogge: TODO: real commit hash

src/vs/workbench/contrib/terminal/common/scripts/psreadline/PSReadLine.psd1
```diff
@@ -0,0 +1,19 @@
+@{
+RootModule = 'PSReadLine.psm1'
+NestedModules = @("Microsoft.PowerShell.PSReadLine.dll")
+ModuleVersion = '2.4.3'
+GUID = '5714753b-2afd-4492-a5fd-01d9e2cff8b5'
+Author = 'Microsoft Corporation'
+CompanyName = 'Microsoft Corporation'
+Copyright = '(c) Microsoft Corporation. All rights reserved.'
+Description = 'Great command line editing in the PowerShell console host'
+PowerShellVersion = '5.1'
```
meganrogge, line 10: I'm surprised this is 5.1 🤔

andyleejordan: Don't look too closely...I found net6.0 references still existing in the codebase...for building its unit tests on macOS/Linux...which I then found hadn't been done in _years_.

src/vs/workbench/contrib/terminal/common/scripts/psreadline/cgmanifest.json
```diff
@@ -0,0 +1,38 @@
...
+        }
+      },
+      "licenseDetail": [
+        "Copyright (c) 2013, Jason Shirk",
```
andyleejordan, line 15: It concerns me that we never updated this to be assigned to Microsoft Corporation 😬

src/vs/workbench/contrib/terminal/common/scripts/shellIntegration.ps1
```diff
@@ -169,6 +169,18 @@ elseif ((Test-Path variable:global:GitPromptSettings) -and $Global:GitPromptSett
 	[Console]::Write("$([char]0x1b)]633;P;PromptType=posh-git`a")
 }
 
+
+if (-not (Get-Module -Name PSReadLine)) {
+	# pwsh version 7+ is required for accessibility support in PSReadLine
```
andyleejordan: I just want to clarify where this requirement is coming from? I don't know of a technical reason beyond 5.1 being a lot less reliable than 7 and feature-frozen.

Tyriar:
Red CI

src/vs/platform/terminal/node/terminalEnvironment.ts
```diff
@@ -112,6 +112,7 @@ export async function getShellIntegrationInjection(
 	}
 	// Windows
 	if (isWindows) {
+		envMixin['VSCODE_A11Y_MODE'] = options.isScreenReaderOptimized ? '1' : '0';
```
Tyriar: We want to do this near the `VSCODE_STABLE` one and for pwsh only.

src/vs/workbench/contrib/terminal/common/scripts/shellIntegration.ps1
```diff
@@ -169,6 +169,14 @@ elseif ((Test-Path variable:global:GitPromptSettings) -and $Global:GitPromptSett
 	[Console]::Write("$([char]0x1b)]633;P;PromptType=posh-git`a")
 }
 
+
+if (-not (Get-Module -Name PSReadLine) -and $env:VSCODE_A11Y_MODE) {
+	$scriptRoot = Split-Path -Parent $MyInvocation.MyCommand.Path
+	$specialPsrlPath = Join-Path $scriptRoot '\psreadline\'
+	Import-Module $specialPsrlPath
+}
```
Tyriar: We should be unsetting `VSCODE_A11Y_MODE` like we do here:

https://github.com/microsoft/vscode/blob/e921a2ee51491b6b2b3825a7eded370760acc624/src/vs/workbench/contrib/terminal/common/scripts/shellIntegration.ps1#L29-L33

src/vs/workbench/contrib/terminal/common/scripts/shellIntegration.ps1
```diff
@@ -169,6 +169,14 @@ elseif ((Test-Path variable:global:GitPromptSettings) -and $Global:GitPromptSett
 	[Console]::Write("$([char]0x1b)]633;P;PromptType=posh-git`a")
 }
 
+
+if (-not (Get-Module -Name PSReadLine) -and $env:VSCODE_A11Y_MODE) {
+	$scriptRoot = Split-Path -Parent $MyInvocation.MyCommand.Path
+	$specialPsrlPath = Join-Path $scriptRoot '\psreadline\'
```
Tyriar: Guessing you don't need the `\`? Looks wrong seeing that since this is meant to work across platforms.

src/vs/workbench/contrib/terminal/common/scripts/shellIntegration.ps1
```diff
@@ -169,6 +169,14 @@ elseif ((Test-Path variable:global:GitPromptSettings) -and $Global:GitPromptSett
...
+	Import-Module $specialPsrlPath
```
Tyriar: We don't need to set that flag anymore?

Tyriar (dismissed):

Tyriar:
Windows pwsh: 
image

src/vs/platform/terminal/node/terminalEnvironment.ts
```diff
@@ -124,6 +124,7 @@ export async function getShellIntegrationInjection(
 			newArgs = [...newArgs]; // Shallow clone the array to avoid setting the default array
 			newArgs[newArgs.length - 1] = format(newArgs[newArgs.length - 1], appRoot, '');
 			envMixin['VSCODE_STABLE'] = productService.quality === 'stable' ? '1' : '0';
+			envMixin['VSCODE_A11Y_MODE'] = options.isScreenReaderOptimized ? '1' : '0';
```
Tyriar: Don't we need to do this when not Windows as well?

src/vs/workbench/contrib/terminal/common/scripts/shellIntegration.ps1
meganrogge: meaning remove the accessibility flag and just do this whenever the psreadline module is not installed? 

I think we still need the flag if we're going to try to load our special one for macOS because we'll first have to unimport the current version?

src/vs/workbench/contrib/chat/browser/chatContentParts/test.ps1
Tyriar, line 1: This shouldn't be included

src/vs/workbench/contrib/terminal/common/scripts/psreadline/Microsoft.PowerShell.PSReadLine.dll
Tyriar, line 1: You should do a product build to make sure all the relevant files are included in the packaged distributables and work when installed.

src/vs/workbench/contrib/terminal/common/scripts/psreadline/cgmanifest.json
```diff
@@ -0,0 +1,40 @@
...
+      "component": {
+        "type": "git",
+        "git": {
+          "name": "PSReadLine",
+          "repositoryUrl": "https://github.com/PowerShell/PSReadLine",
+          "commitHash": "b5aac30d26ce777497e85203dc59989296b73f8e"
+        }
+      },
+			"version": "2.4.4-beta4",
+      "licenseDetail": [
```
Tyriar:
```suggestion
      },
      "version": "2.4.4-beta4",
      "licenseDetail": [
```

src/vs/workbench/contrib/terminal/common/scripts/shellIntegration.ps1
```diff
@@ -169,6 +173,16 @@ elseif ((Test-Path variable:global:GitPromptSettings) -and $Global:GitPromptSett
 	[Console]::Write("$([char]0x1b)]633;P;PromptType=posh-git`a")
 }
 
+if ($Global:__VSCodeState.IsA11yMode -eq "1") {
+	if (Get-Module -Name PSReadLine) {
+		return
```
Tyriar: ⚠️ I think this will exit the entire script

src/vs/workbench/contrib/terminal/common/scripts/shellIntegration.ps1
```diff
@@ -169,6 +173,16 @@ elseif ((Test-Path variable:global:GitPromptSettings) -and $Global:GitPromptSett
...
+	}
+	$scriptRoot = Split-Path -Parent $MyInvocation.MyCommand.Path
+	$specialPsrlPath = Join-Path $scriptRoot 'psreadline'
+	Import-Module $specialPsrlPath
+	Set-PSReadLineOption -EnableScreenReaderMode
```
Tyriar: If `Import-Module` fails and the script continues, `Set-PSReadLineOption` would probably fail as well:

```suggestion
	Import-Module $specialPsrlPath
	if (Get-Module -Name PSReadLine) {
		Set-PSReadLineOption -EnableScreenReaderMode
	}
```

Tyriar (changes requested):

Tyriar (approved):
Test failures. LGTM provided the product build worked out

andyleejordan:
🥳

PR #263799, opened 2025-08-28
fix: ensure SVG images are loaded before clipboard copy

lukocode:
Fixes intermittent clipboard copy failures for SVG output where users
encountered "No blob data to write to clipboard" errors. The issue was
caused by a race condition where canvas operations proceeded before
the dynamically created Image element finished loading the SVG data URL.
- Add ensureImageLoaded helper to wait for image load completion
- Handle both already-loaded and loading states
- Include error handling and 5-second timeout protection
- Maintain backward compatibility with existing image copy functionality

Fixes: #265020


amunger:
is there an issue that can be linked to this?

lukocode:
I’ve registered issue #265020 and linked it to this pull request.

amunger (dismissed):

src/vs/workbench/contrib/notebook/browser/view/renderers/webviewPreloads.ts
```diff
@@ -1616,7 +1616,20 @@ async function webviewPreloads(ctx: PreloadContext) {
 			}
 
 			if (image) {
-				const imageToCopy = image;
+				//FIX: Ensure image is loaded before proceeding
```
Tyriar: `// FIX:` Normally means there's a todo here? Can we remove that?

src/vs/workbench/contrib/notebook/browser/view/renderers/webviewPreloads.ts
```diff
@@ -1616,7 +1616,20 @@ async function webviewPreloads(ctx: PreloadContext) {
...
+				const ensureImageLoaded = (img: HTMLImageElement): Promise<HTMLImageElement> => {
+					return new Promise((resolve, reject) => {
+						if (img.complete && img.naturalWidth > 0) {
+							resolve(img);
+						} else {
+							img.onload = () => resolve(img);
+							img.onerror = () => reject(new Error('Failed to load image'));
+							// Add timeout to prevent hanging
```
Tyriar: Seems redundant

Tyriar (changes requested):

src/vs/workbench/contrib/notebook/browser/view/renderers/webviewPreloads.ts
lukocode: Thanks for catching that. I’ve removed the `// FIX:` and clarified the comments.

src/vs/workbench/contrib/notebook/browser/view/renderers/webviewPreloads.ts
lukocode: Thank you for the feedback!
I understand why it might seem redundant, but in practice, the image is not always guaranteed to be fully loaded at this point - especially with dynamically created images. Adding this check prevents race conditions and intermittent clipboard errors, as seen in #265020.
I’ve clarified the comments in the code to explain this rationale.
Please let me know if you have further concerns or if I’ve missed something!

Tyriar (approved):

TylerLeonhardt (approved):

lukocode:
Thanks, guys! I really appreciate your support.

PR #263876, opened 2025-08-28
polish `OutputMonitor`

meganrogge:
Ensures:
- `onCommandFinished` always fires even if something throws to prevent hanging
- fixes a bug from a refactor - when a task was still active and there was no new data, we were returning idle


meganrogge:
one at a time bc something caused a leak

PR #263889, opened 2025-08-28
use last prompt and marker to ensure no duplicate question

### LINKED ISSUE #263888: can re-prompt with same question
The marker approach does not work since we include the last five lines of output and the prompt can occur within those. cc @Tyriar 

Image

meganrogge:
fix #263888

cc @tyriar

### DIFF
--- a/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/monitoring/outputMonitor.ts
+++ b/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/monitoring/outputMonitor.ts
@@ -45,6 +45,8 @@ export class OutputMonitor extends Disposable implements IOutputMonitor {
 
 	private _lastPromptMarker: XtermMarker | undefined;
 
+	private _lastPrompt: string | undefined;
+
 	private _pollingResult: IPollingResult & { pollDurationMs: number } | undefined;
 	get pollingResult(): IPollingResult & { pollDurationMs: number } | undefined { return this._pollingResult; }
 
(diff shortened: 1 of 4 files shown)


src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/monitoring/outputMonitor.ts
```diff
@@ -438,16 +437,10 @@ export class OutputMonitor extends Disposable implements IOutputMonitor {
 		const prompt = confirmationPrompt.prompt;
 		const options = confirmationPrompt.options;
 
-		const currentMarker = this._execution.instance.registerMarker();
-		if (!currentMarker) {
-			// Unable to register marker, so cannot track prompt location
-			return undefined;
-		}
-		if (this._lastPromptMarker?.line === currentMarker.line) {
+		if (this._lastPrompt?.includes(prompt) || this._lastPrompt && prompt?.includes(this._lastPrompt)) {
```
Tyriar: No understanding the problem mentioned in https://github.com/microsoft/vscode/issues/263888, in that scenario, the marker should be pointing at line 3 for the first one and line 9 for the second?

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/monitoring/outputMonitor.ts
```diff
@@ -438,16 +437,10 @@ export class OutputMonitor extends Disposable implements IOutputMonitor {
...
 			// Same prompt as last time, so avoid re-prompting
 			return;
 		}
-		this._lastPromptMarker = currentMarker;
```
meganrogge, line 458:
```suggestion
this._lastPrompt = prompt;
```

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/monitoring/outputMonitor.ts
```diff
@@ -438,17 +437,11 @@ export class OutputMonitor extends Disposable implements IOutputMonitor {
...
-
+this._lastPrompt = prompt;
```
meganrogge:
```suggestion
		this._lastPrompt = prompt;
```

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/monitoring/outputMonitor.ts
meganrogge: we don't set the marker where the question is asked. we set where the buffer ends atm.

mjbvz (approved):

PR #263894, opened 2025-08-28
fix: confirm response based on promise

rwoll:
When using the new `workbench.action.chat.open.blockOnResponse` feature, if the VSC instance was in auto-approve mode, it was observed that the command was resolving before the tool finished invocation.

This resulted in a race betwee sync and async code.

### DIFF
--- a/src/vs/workbench/contrib/chat/common/chatProgressTypes/chatToolInvocation.ts
+++ b/src/vs/workbench/contrib/chat/common/chatProgressTypes/chatToolInvocation.ts
@@ -29,7 +29,6 @@ export class ChatToolInvocation implements IChatToolInvocation {
 		return this._confirmDeferred;
 	}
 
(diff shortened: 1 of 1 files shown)


src/vs/workbench/contrib/chat/common/chatProgressTypes/chatToolInvocation.ts
```diff
@@ -31,7 +31,7 @@ export class ChatToolInvocation implements IChatToolInvocation {
 
 	private _isConfirmed: ConfirmedReason | undefined;
```
connor4312: We can get rid of _isConfirmed entirely

connor4312 (approved):

DonJayamanne (approved):

DonJayamanne (approved):

karthiknadig (approved):

zhichli (approved):

PR #263913, opened 2025-08-28
Add URL handler for chat message

ntrogh:
This URL handler allows external applications to insert messages into the VS Code chat interface via URLs. The chat message is inserted in the chat input field but is _not_ submitted. The user is prompted if they want to insert the specific prompt in chat.

## URL Format

```
code-oss:chat-message?prompt=<encoded-prompt-text>&mode=<chat-mode>
```

### Parameters

- `prompt` (required): The text to insert into the chat input field (URL encoded)
- `mode` (optional): The chat mode to set. Defaults to 'agent'. Supports both built-in modes ('ask', 'edit', 'agent') and custom chat modes defined in `.chatmode` files

## Examples

```
code-oss:chat-message?prompt=Hello%20World
```

```
code-oss:chat-message?prompt=Explain%20this%20code&mode=ask
```

```
code-oss:chat-message?prompt=Review%20my%20JavaScript%20code%20for%20potential%20bugs%20and%20security%20issues&mode=agent
```

## Test steps

1. Run **Developer: Open URL** command in Command Palette
1. Enter a URL like:
   - `code-oss:chat-message?prompt=Test%20prompt`
   - `code-oss:chat-message?prompt=Hello&mode=ask`

The handler should:
1. Show a confirmation dialog with the prompt preview
2. Upon confirmation, open the chat view
3. Set the specified chat mode
4. Insert the prompt text into the input field but not submit it
5. Focus the input field for user interaction


ntrogh:
@digitarald @aeschli Something I've been wanting for a while was to be able to use a URL to enter a prompt message in chat. Usage scenario is to use integrate this in our docs or other repos like `awesome-copilot`. You can provide a prompt text and chat mode (optional) as input.

I've based the code off of the `chat` CLI subcommand and the URL handlers we have for chat modes/prompt files/instructions.

Can you review this PR and check whether the functionality makes sense?

src/vs/workbench/contrib/chat/browser/chat.contribution.ts
```diff
@@ -850,6 +851,7 @@ registerWorkbenchContribution2(ChatTransferContribution.ID, ChatTransferContribu
 registerWorkbenchContribution2(ChatContextContributions.ID, ChatContextContributions, WorkbenchPhase.AfterRestored);
 registerWorkbenchContribution2(ChatResponseResourceFileSystemProvider.ID, ChatResponseResourceFileSystemProvider, WorkbenchPhase.AfterRestored);
 registerWorkbenchContribution2(PromptUrlHandler.ID, PromptUrlHandler, WorkbenchPhase.BlockRestore);
+registerWorkbenchContribution2(ChatMessageUrlHandler.ID, ChatMessageUrlHandler, WorkbenchPhase.BlockRestore);
```
bpasero: I think there is no good reason to block restore here, please pick another later phase.

bpasero (changes requested):

src/vs/workbench/contrib/chat/browser/chat.contribution.ts
ntrogh: @bpasero Good point. I'd suggest `AfterRestored` rather than `Eventually` because this is in immediate response to a user action. WDYT?

bpasero: Yeah after restored sounds good.

ntrogh: @aeschli Should we also update the `PromptUrlHandler` registration to happen after restored?

aeschli (dismissed):

src/vs/workbench/contrib/chat/browser/chatMessageUrlHandler.ts
```diff
@@ -0,0 +1,221 @@
...
+import { IWorkbenchContribution } from '../../../common/contributions.js';
+import { ICommandService } from '../../../../platform/commands/common/commands.js';
+import { Disposable } from '../../../../base/common/lifecycle.js';
+import { IViewsService } from '../../../services/views/common/viewsService.js';
+import { showChatView } from './chat.js';
+import { ChatModeKind } from '../common/constants.js';
+// import { IChatModeService } from '../common/chatModes.js';
+import { ILogService } from '../../../../platform/log/common/log.js';
+// import { IHostService } from '../../../services/host/browser/host.js';
+// import { mainWindow } from '../../../../base/browser/window.js';
```
Tyriar: Can these imports be cleaned up?

src/vs/workbench/contrib/chat/browser/chatMessageUrlHandler.ts
```diff
@@ -0,0 +1,221 @@
...
+	private static readonly MAX_PROMPT_LENGTH = 10000; // characters
+
+	constructor(
+		@IURLService urlService: IURLService,
+		@ICommandService private readonly commandService: ICommandService,
+		@IWorkspaceTrustRequestService private readonly workspaceTrustRequestService: IWorkspaceTrustRequestService,
+		@IViewsService private readonly viewsService: IViewsService,
+		// @IChatModeService private readonly chatModeService: IChatModeService,
+		@ILogService private readonly logService: ILogService,
+		// @IHostService private readonly hostService: IHostService,
```
Tyriar: Remove comments

bpasero (changes requested):
What is with the code that is commented out?

src/vs/workbench/contrib/chat/browser/chatMessageUrlHandler.ts
```diff
@@ -0,0 +1,221 @@
...
+import { MarkdownString } from '../../../../base/common/htmlContent.js';
+import { InvisibleCharacters, removeAnsiEscapeCodes, AmbiguousCharacters } from '../../../../base/common/strings.js';
+import { IWorkspaceTrustRequestService } from '../../../../platform/workspace/common/workspaceTrust.js';
+import { ACTION_ID_NEW_CHAT, CHAT_OPEN_ACTION_ID, IChatViewOpenOptions } from '../browser/actions/chatActions.js';
+
+// URL format: code-oss:chat-message?prompt=Hello%20World&mode=agent
+export class ChatMessageUrlHandler extends Disposable implements IWorkbenchContribution, IURLHandler {
+
+	static readonly ID = 'workbench.contrib.chatMessageUrlHandler';
+	private static readonly MAX_PROMPT_LENGTH = 10000; // characters
```
Tyriar: You could encode the fast it's characters into the variable to prevent bitrot: `MAX_PROMPT_CHAR_LENGTH`

src/vs/workbench/contrib/chat/browser/chatMessageUrlHandler.ts
```diff
@@ -0,0 +1,221 @@
...
+		if (uri.path !== 'chat-message') {
+			return false;
+		}
+
+		// Verify that workspace is trusted
+		const trusted = await this.workspaceTrustRequestService.requestWorkspaceTrust({
+			message: localize('copilotWorkspaceTrust', "Copilot is currently only supported in trusted workspaces.")
+		});
+
+		if (!trusted) {
```
Tyriar: Remove comment, it's obvious from context

src/vs/workbench/contrib/chat/browser/chatMessageUrlHandler.ts
```diff
@@ -0,0 +1,221 @@
...
+			const promptText = searchParams.get('prompt');
+			const modeParam = searchParams.get('mode');
+
+			if (!promptText) {
+				this.logService.error('[ChatMessageUrlHandler] No prompt parameter provided');
+				return true;
+			}
+
+			// Decode and clean prompt text
+			const decodedPrompt = this.sanitizePromptText(decodeURIComponent(promptText));
```
Tyriar: Comment obvious from context

src/vs/workbench/contrib/chat/browser/chatMessageUrlHandler.ts
```diff
@@ -0,0 +1,221 @@
...
+			const opts: IChatViewOpenOptions = {
+				query: decodedPrompt,
+				mode: modeParam ?? ChatModeKind.Agent,
+				isPartialQuery: true
+			};
+
+			// Show confirmation dialog
+			if (await this.shouldBlockChatMessage(decodedPrompt)) {
+				return true;
+			}
```
Tyriar: Don't need these comments

src/vs/workbench/contrib/chat/browser/chatMessageUrlHandler.ts
```diff
@@ -0,0 +1,221 @@
...
+
+			// Proceed with opening chat
+			const chatWidget = await showChatView(this.viewsService);
+
+			await chatWidget?.waitForReady();
```
Tyriar: Do we need to handle when `chatWidget` is undefined?

src/vs/workbench/contrib/chat/browser/chatMessageUrlHandler.ts
```diff
@@ -0,0 +1,221 @@
...
+			await this.commandService.executeCommand(ACTION_ID_NEW_CHAT);
+			await this.commandService.executeCommand(CHAT_OPEN_ACTION_ID, opts);
```
Tyriar, line 100: Are these doing the same thing? It's typically better to call into services than command like this where possible (I don't know this area though)

ntrogh:
@bpasero Refactored to use `IChatViewOpenOptions` upon advice from @aeschli. And I forgot to remove commented-out imports and ctor params that were no longer needed.

src/vs/workbench/contrib/chat/browser/chatMessageUrlHandler.ts
```diff
@@ -0,0 +1,221 @@
...
+
+			return true;
+		} catch (error) {
+			this.logService.error(`[ChatMessageUrlHandler] Error handling chat message URL ${uri.toString()}`, error);
+			return true;
```
Tyriar, line 105: Should this be false? I don't know what the return value does

src/vs/workbench/contrib/chat/browser/chatMessageUrlHandler.ts
```diff
@@ -0,0 +1,221 @@
...
+		}
+	}
+
+	private async shouldBlockChatMessage(messageText: string): Promise<boolean> {
+		// Truncate preview for display
```
Tyriar: Remove

src/vs/workbench/contrib/chat/browser/chatMessageUrlHandler.ts
```diff
@@ -0,0 +1,221 @@
...
+		const previewText = messageText.length > 200
+			? messageText.substring(0, 200) + '...'
```
Tyriar: Magic number referenced twice, my pref here is to put it in a const enum, but a const somewhere would be fine

src/vs/workbench/contrib/chat/browser/chatMessageUrlHandler.ts
```diff
@@ -0,0 +1,221 @@
...
+		const detail = new MarkdownString('', { supportHtml: true });
+		detail.appendMarkdown(localize('confirmChatMessageDetail', "This will insert the following text into chat:\n\n"));
+		detail.appendMarkdown(`\`\`\`\n${previewText}\n\`\`\``);
+		detail.appendMarkdown(localize('confirmChatMessageSecurity',
+			"\n\nIf you did not initiate this request, it may represent an attempted attack on your system. Unless you took an explicit action to initiate this request, you should press 'No'."));
+
+		const { confirmed } = await this.dialogService.confirm({
+			type: 'info',
+			primaryButton: localize({ key: 'yesButton', comment: ['&& denotes a mnemonic'] }, "&&Yes"),
+			cancelButton: localize('noButton', "No"),
```
Tyriar: It's possible the `No` could be localized differently in the button and the message

src/vs/workbench/contrib/chat/browser/chatMessageUrlHandler.ts
```diff
@@ -0,0 +1,221 @@
...
+		// Truncate preview for display
+		const previewText = messageText.length > 200
+			? messageText.substring(0, 200) + '...'
+			: messageText;
+
+		const detail = new MarkdownString('', { supportHtml: true });
+		detail.appendMarkdown(localize('confirmChatMessageDetail', "This will insert the following text into chat:\n\n"));
+		detail.appendMarkdown(`\`\`\`\n${previewText}\n\`\`\``);
+		detail.appendMarkdown(localize('confirmChatMessageSecurity',
+			"\n\nIf you did not initiate this request, it may represent an attempted attack on your system. Unless you took an explicit action to initiate this request, you should press 'No'."));
```
Tyriar: nit: appending like this is less efficient on memory/gc. How about this?

```ts
new MarkdownString([
  'first',
  'second',
  'third'
].join('\n\n')
```

src/vs/workbench/contrib/chat/browser/chatMessageUrlHandler.ts
```diff
@@ -0,0 +1,221 @@
...
+
+		const { confirmed } = await this.dialogService.confirm({
+			type: 'info',
+			primaryButton: localize({ key: 'yesButton', comment: ['&& denotes a mnemonic'] }, "&&Yes"),
+			cancelButton: localize('noButton', "No"),
+			message: localize('confirmChatMessage', "An external application wants to insert text into chat. Do you want to continue?"),
+			custom: {
+				markdownDetails: [{
+					markdown: detail
+				}]
```
Tyriar: If you rename `detail` to `markdown` you could make this line `markdownDetails: [{ markdown }]`

Tyriar:
I have a full review coming

src/vs/workbench/contrib/chat/browser/chatMessageUrlHandler.ts
```diff
@@ -0,0 +1,221 @@
...
+					// Skip the next character if this was a surrogate pair
+					if (codePoint > 0xFFFF) {
+						i++;
+					}
+				}
+			}
+		}
+
+		// Remove control characters except tab (9), newline (10), carriage return (13)
+		cleanText = cleanText.replace(/[\u0000-\u0008\u000B\u000C\u000E-\u001F\u007F-\u009F]/g, '');
```
Tyriar: Wouldn't this be handled by `InvisibleCharacters.isInvisibleCharacter`?

src/vs/workbench/contrib/chat/browser/chatMessageUrlHandler.ts
```diff
@@ -0,0 +1,221 @@
...
+
+		// Check for potentially deceptive URLs (homograph attacks)
+		cleanText = this.detectAndWarnAboutSuspiciousUrls(cleanText);
+
+		// Normalize multiple whitespace to single spaces and trim
+		return cleanText.replace(/\s+/g, ' ').trim();
```
Tyriar, line 163: Could whitespace be important to the query?

Tyriar (changes requested):

Tyriar:
@ntrogh we typically only label PRs with `candidate`

src/vs/workbench/contrib/chat/browser/chatMessageUrlHandler.ts
ntrogh: The first starts a new chat session and the second opens the Chat view with the requested options.

@bpasero I reused this implementation from your CLI `chat` subcommand. My understanding is that this ensures other checks happen like validating chat mode?

src/vs/workbench/contrib/chat/browser/chatMessageUrlHandler.ts
ntrogh: Should be `true`. The return value indicates if the URL should be processed by another URL handler. So, only if there's another command than`chat-command` should the URL be passed onwards.

src/vs/workbench/contrib/chat/browser/chatMessageUrlHandler.ts
ntrogh: I wanted to be on the safe side as a request could use blanks as a filler to obfuscate malicious content near the end and outside of the viewport.

ntrogh:
Reopening after discussing with @digitarald

src/vs/workbench/contrib/chat/browser/chatMessageUrlHandler.ts
digitarald: As a follow-up it would be nice to extend the command to add a "new thread" paramater, as the current default behavior appends to whatever thread was going.

src/vs/workbench/contrib/chat/browser/chatMessageUrlHandler.ts
digitarald: As one caveat, if a prompt is multi-line markdown with paragraphs or lists, this might break the structure.

digitarald (approved):
Small nits, but I think this is in good shape to merge for testing week

src/vs/workbench/contrib/chat/browser/chatMessageUrlHandler.ts
```diff
@@ -0,0 +1,209 @@
...
+		this._register(urlService.registerHandler(this));
+	}
+
+	async handleURL(uri: URI, options?: IOpenURLOptions): Promise<boolean> {
+		if (uri.path !== 'chat-message') {
+			return false;
+		}
+
+		const trusted = await this.workspaceTrustRequestService.requestWorkspaceTrust({
+			message: localize('copilotWorkspaceTrust', "Copilot is currently only supported in trusted workspaces.")
```
bpasero, line 50:
```suggestion
			message: localize('chatWorkspaceTrust', "AI features are currently only supported in trusted workspaces.")
```

src/vs/workbench/contrib/chat/browser/chatMessageUrlHandler.ts
```diff
@@ -0,0 +1,209 @@
...
+
+		// Normalize multiple whitespace to single spaces and trim
+		return cleanText.replace(/\s+/g, ' ').trim();
+	}
+
+	/**
+	 * Detects URLs with potentially deceptive characters (homograph attacks) and removes them.
+	 * Replaces suspicious URLs with a warning message for security.
+	 */
+	private detectAndWarnAboutSuspiciousUrls(text: string): string {
```
bpasero, line 170: Should this connect to the allow list of URLs we have when opening links to only trust those?

src/vs/workbench/contrib/chat/browser/chatMessageUrlHandler.ts
```diff
@@ -0,0 +1,209 @@
...
+
+		return !confirmed;
+	}
+
+	/**
+	 * Cleans prompt text by removing invisible characters, control sequences, and normalizing whitespace.
+	 * Uses VS Code's existing utilities for comprehensive and consistent text sanitization.
+	 * Also removes potentially deceptive URLs (homograph attacks) for security.
+	 */
+	private sanitizePromptText(text: string): string {
```
bpasero, line 139: I wonder if this should move to `base` and get some extensive test suite.

bpasero (changes requested):
Ok

PR #263925, opened 2025-08-28
Support alt+click to open diff editor to the side.

rebornix:


src/vs/workbench/contrib/chat/browser/media/chat.css
```diff
@@ -1708,10 +1708,16 @@ have to be updated for changes to the rules above, or to support more deeply nes
 	border: 1px solid var(--vscode-chat-requestBorder);
 
 	.chat-view-changes-icon {
+		padding: 3px;
```
rebornix, line 1711: @aiday-mar I added this style update to ensure that the buttons render the same as toolbar buttons (size and hover effects)

Image

joshspicer (approved):

PR #263927, opened 2025-08-28
Enable users to view and apply changes from worktree to main repo, per file

nguyenchristy:
'Compare with Workspace' is only available from a worktree's context menu under 'Changes' and opens the merge editor where users can apply changes from a worktree to its main repository. 
Screenshot 2025-08-28 164553

image

### DIFF
--- a/extensions/git/src/repository.ts
+++ b/extensions/git/src/repository.ts
@@ -182,6 +182,7 @@ export class Resource implements SourceControlResourceState {
 	get type(): Status { return this._type; }
 	get original(): Uri { return this._resourceUri; }
 	get renameResourceUri(): Uri | undefined { return this._renameResourceUri; }
+	get contextValue(): string | undefined { return this._repository?.kind; }
 
 	private static Icons: any = {
 		light: {
@@ -310,6 +311,7 @@ export class Resource implements SourceControlResourceState {
 		private _type: Status,
 		private _useIcons: boolean,
 		private _renameResourceUri?: Uri,
+		private _repository?: Repository,
 	) { }
 
 	async open(): Promise<void> {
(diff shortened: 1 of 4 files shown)


extensions/git/src/repository.ts
```diff
@@ -182,6 +182,7 @@ export class Resource implements SourceControlResourceState {
 	get type(): Status { return this._type; }
 	get original(): Uri { return this._resourceUri; }
 	get renameResourceUri(): Uri | undefined { return this._renameResourceUri; }
+	get contextValue(): string | undefined { return this._repository?.kind; }
```
lszomoru: Where is this being used?

lszomoru (changes requested):

extensions/git/src/repository.ts
nguyenchristy: In the `package.json` where I set `scmResourceState == worktree`
https://github.com/microsoft/vscode/blob/main/src/vscode-dts/vscode.d.ts#L16389

extensions/git/src/repository.ts
```diff
@@ -310,6 +311,7 @@ export class Resource implements SourceControlResourceState {
 		private _type: Status,
 		private _useIcons: boolean,
 		private _renameResourceUri?: Uri,
+		private _repository?: Repository,
```
lszomoru: Got it. Let's not pass in the complete repository object but just pass in the repository kind (similar to how the ResourceGroupType is being passed in.

lszomoru (changes requested):

extensions/git/src/repository.ts
nguyenchristy: Updated. Lmk if you'd eventually like to make RepositoryKind an enum. I just didn't do it here because I didn't want to make a bigger change than necessary to other files.

lszomoru (approved):

PR #263934, opened 2025-08-29
fix state issue with terminal polling

### LINKED ISSUE #263926: Window hangs when prompting to continue waiting for terminal
Reported by Simon originally but I can repro

Image

First I made this modification hoping to be able to trigger it faster

Image

Then "run top in a bg terminal"

Wait a bit, then it hangs.

First, I don't understand what we are waiting for the background terminal for. Isn't the point that you are expecting it to run in the background, so why would you ask me whether to "keep waiting" for it?

Then, we are stuck in an infinite while loop that has no way to exit when `state == 'Prompting'`. Please be very careful with while loops.

Image

meganrogge:
fix #263926

The issue was mostly `this._state = State.Prompting` which isn't handled in the switch.

I changed the polling and extended polling durations each to 5 seconds to test. Here's the behavior with this fix:

https://github.com/user-attachments/assets/ed4f256d-999b-493c-9afd-864a4b9f597d

cc @tyriar


roblourens (approved):
Thanks!

PR #263943, opened 2025-08-29
fix(chat): disable arrow key history navigation when input is filled

### LINKED ISSUE #282902: Remove "up/down arrow to navigate message history" feature from chat textarea
- Copilot Chat Extension Version: January 2025 (version 1.97)
- VS Code Version: Version: 1.97.2
- OS Version: MacOS 14.7.1

Steps to Reproduce:

1. Type `up arrow` at beginning of chat textarea, which can happen frequently when holding down up arrow to get back to the beginning of the message.
2. Observe that the message disappears, and caret is now somewhere in the first line of a previously sent message.
3. Type `down arrow` repeatedly until the message being edited re-appears.

My request is that the "up/down history navigation" feature is removed. I find it frustrating and unexpected every time it happens. If I wanted to re-send a message, I could just scroll up and copy and paste it in. This isn't a terminal window, and shouldn't behave like one. Claude, ChatGPT, and Cursor do not have this behavior.

See also: https://github.com/microsoft/vscode-copilot-release/issues/708#issuecomment-1875309496

>> I don't think that the arrow keys by themselves should move us from one chat to another. Perhaps alt-shift-downarrow/uparrow -- but not just the arrow keys themselves. On many occasions I use the arrow keys to go to the top or bottom of a chat with no desire to switch to a previous or later chat. If I do wish to cycle through chats, that's a unique case that should require some extra keys to make sure that it doesn't happen by accident.
>>
>> That is in addition to the annoying problem that it deletes what we are currently working on.
>
> I agree with this 100%. It would actually be better to remove this feature completely. If we want to find a previous chat we can use the clock (history) icon. Being able to flick through them with arrow keys in the input field doesn’t add anything (other than confusion).

scarf005:
prevents an issue where arrow key navigation erases very long prompt ([oftentimes current progress isn't retrievable](https://github.com/microsoft/vscode-copilot-release/issues/11340))

fixes https://github.com/microsoft/vscode/issues/282902

### DIFF
--- a/src/vs/workbench/contrib/chat/browser/chatInputPart.ts
+++ b/src/vs/workbench/contrib/chat/browser/chatInputPart.ts
@@ -1157,6 +1157,7 @@ export class ChatInputPart extends Disposable implements IHistoryNavigationWidge
 			const model = this._inputEditor.getModel();
 			const inputHasText = !!model && model.getValue().trim().length > 0;
 			this.inputEditorHasText.set(inputHasText);
+			onDidChangeCursorPosition();
 		}));
 		this._register(this._inputEditor.onDidContentSizeChange(e => {
 			if (e.contentHeightChanged) {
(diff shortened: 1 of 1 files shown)


ulugbekna:
I understand the problem but I think the solution should be that arrow-down brings back the in-progress prompts as well rather than disabling arrow-keys history navigation, which can be useful if you realise that you want to modify your past prompt instead creating a new one in the middle of writing a new prompt

scarf005:
> I understand the problem but I think the solution should be that arrow-down brings back the in-progress prompts as well rather than disabling arrow-keys history navigation, which can be useful if you realise that you want to modify your past prompt instead creating a new one in the middle of writing a new prompt

I see, but personally I never had the need to navigate history when I was in middle of writing prompts (other than accidentally pressing arrow keys and losing 10 minutes of work). I think this shouldn't be much of a problem since it's easy to select all then `ctrl+x`

roblourens:
> I understand the problem but I think the solution should be that arrow-down brings back the in-progress prompts as well rather than disabling arrow-keys history navigation, which can be useful if you realise that you want to modify your past prompt instead creating a new one in the middle of writing a new prompt

This is how it's supposed to work and it works that way for me. It has broken in the past and there could be another bug there due to all the churn in the chat input part/widget. So if you have an example of a case where it's not working #scarf005 we could get to the bottom of it. But I'm not going to delete the feature right now. Your linked issue has an example of how to set the keybindings to your preference.

PR #264012, opened 2025-08-29
Add support for extension version filter assignments

benibenj:
```Copilot Generated Description:``` Introduce a new filter provider for managing extension version assignments, allowing the system to track and utilize the versions of specific extensions. This includes updates to the assignment service to integrate the new filter provider and handle version changes effectively.

Debt item: https://github.com/microsoft/vscode/issues/264018


bpasero (dismissed):
I would suggest to turn the existing internal property into a string allowing it to have `microsoft` or `github` so that we do not have to introduce properties per organisation.

benibenj:
Unblocking this PR as we need this in by Monday. Created a debt item with @bpasero comment for @sbatten

chrmarti (approved):

bpasero:
@benibenj one thing I only see now is this line:

https://github.com/microsoft/vscode/blob/9702bf9cd3fbdf457225a9eeda78bec5db27adf6/src/vs/workbench/services/assignment/common/assignmentFilters.ts#L13-L14

That is a heavy layer breaker that sneaked in that must be fixed too. So that needs to be another debt item for September: 
* https://github.com/microsoft/vscode/issues/264280

sbatten:
@benibenj Sorry for making your simple change complicated
@bpasero Thanks for considering and creating a follow-up plan

PR #264020, opened 2025-08-29
fix terminal hang for `basic` and `none` SI strategies

### LINKED ISSUE #264013: Chat hangs if command is run in foreground terminal and then another is attempted
With `basic` or `none` shell integration strategy: 
1. Run `run ls in the terminal` in chat and accept the prompt
2. Manually run `top` in the Copilot terminal
3. Run `run cat README.md` in chat and accept the prompt
4. 🐛 hangs

cc @tyriar

https://github.com/user-attachments/assets/672b5e68-5c0a-4eea-9f4b-ab8deaf968da

meganrogge:
fix #264013

Before:

https://github.com/user-attachments/assets/a14fe82e-7463-4f5b-a435-b9dd89e94e74

After:

https://github.com/user-attachments/assets/7716de42-4ff5-4e86-8e17-09e03c949f4a


src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/executeStrategy/basicExecuteStrategy.ts
```diff
@@ -95,6 +96,15 @@ export class BasicExecuteStrategy implements ITerminalExecuteStrategy {
 				this._onDidCreateStartMarker.fire(this._startMarker = store.add(xterm.raw.registerMarker()));
 			}));
 
+			// Fixes #264013
+			const dataActive = await isDataActive(this._instance);
+			if (dataActive) {
+				this._log('Command timed out, sending SIGINT and retrying');
+				// Send SIGINT (Ctrl+C)
+				await this._instance.sendText('\x03', false);
+				await waitForIdle(this._instance.onData, 1000);
```
meganrogge: going to see if this `waitForIdle` is needed

meganrogge: if not, can move the rest of this into the helper function

meganrogge: yes, it's needed.

meganrogge: going to try reducing this time

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/executeStrategy/basicExecuteStrategy.ts
```diff
@@ -95,6 +96,15 @@ export class BasicExecuteStrategy implements ITerminalExecuteStrategy {
...
 			}));
 
+			// Fixes #264013
+			const dataActive = await isDataActive(this._instance);
+			if (dataActive) {
+				this._log('Command timed out, sending SIGINT and retrying');
+				// Send SIGINT (Ctrl+C)
+				await this._instance.sendText('\x03', false);
+				await waitForIdle(this._instance.onData, 100);
+			}
```
Tyriar: As discussed I don't think we should do this generally, even 100ms adds up. Instead only doing this workaround when we detect the user interacted with the terminal:

* Listen to data events on all copilot-owned terminals and flag them as receiving input if they get one of these events, this could go on `toolTerminal`
* Clear that state for the terminal at the end of an execution
* Check for basic/none SI whether it's been touched, if so, send ^C and wait briefly for something that resembles a prompt `waitForIdleWithPromptHeuristics` (this would avoid that p10k problem)

Tyriar (dismissed):

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/executeStrategy/basicExecuteStrategy.ts
```diff
@@ -43,11 +44,14 @@ export class BasicExecuteStrategy implements ITerminalExecuteStrategy {
 	private readonly _onDidCreateStartMarker = new Emitter<IXtermMarker | undefined>;
 	public onDidCreateStartMarker: Event<IXtermMarker | undefined> = this._onDidCreateStartMarker.event;
```
meganrogge, line 45: added this to keep the diff simple

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/runInTerminalTool.ts
```diff
@@ -521,7 +521,6 @@ export class RunInTerminalTool extends Disposable implements IToolImpl {
 					outputMonitor = store.add(this._instantiationService.createInstance(OutputMonitor, { instance: toolTerminal.instance, sessionId: invocation.context!.sessionId, getOutput: (marker?: IXtermMarker) => getOutput(toolTerminal.instance, marker ?? startMarker) }, undefined, invocation.context!, token, command));
 				}));
 				const executeResult = await strategy.execute(command, token);
```
meganrogge, line 523:
```suggestion
				const executeResult = await strategy.execute(command, token);

```

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/executeStrategy/basicExecuteStrategy.ts
```diff
@@ -107,6 +118,9 @@ export class BasicExecuteStrategy implements ITerminalExecuteStrategy {
 			// execution requested
 			const finishedCommand = await onDone;
 
+			// Reset user input state after command execution completes
+			this._toolTerminal.receivedUserInput = false;
```
Tyriar: This could live in `runInTerminalTool` after we `await execute()`, then the resetting is the same for all strategies

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/runInTerminalTool.ts
```diff
@@ -641,6 +643,14 @@ export class RunInTerminalTool extends Disposable implements IToolImpl {
 		return toolTerminal;
 	}
 
+	private _registerInputListener(toolTerminal: IToolTerminal): void {
+		const disposable = toolTerminal.instance.onData(data => {
+			toolTerminal.receivedUserInput = data.length > 0;
+		});
+		this._register(toolTerminal.instance.onDisposed(() => disposable.dispose()));
+	}
```
Tyriar: You should filter out `telemetryIgnoredSequences`

https://github.com/microsoft/vscode/blob/e2fb54b517a2180f3710a9853c43e5f9ad0848bc/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/runInTerminalTool.ts#L129-L132

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/executeStrategy/noneExecuteStrategy.ts
```diff
@@ -22,13 +23,16 @@ export class NoneExecuteStrategy implements ITerminalExecuteStrategy {
...
 	private _startMarker: IXtermMarker | undefined;
 
+	private readonly _instance: ITerminalInstance;
+
 	private readonly _onDidCreateStartMarker = new Emitter<IXtermMarker | undefined>;
 	public onDidCreateStartMarker: Event<IXtermMarker | undefined> = this._onDidCreateStartMarker.event;
 
 	constructor(
-		private readonly _instance: ITerminalInstance,
+		private readonly _toolTerminal: IToolTerminal,
```
Tyriar: It's probably nicer actually to keep it how it was and still pass in `_instance` but also `_hasReceivedUserInput: () => boolean`. That way the properties on the `IToolTerminal` object can't be messed with here at all (combined with my other comment).

Tyriar (dismissed):

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/executeStrategy/basicExecuteStrategy.ts
meganrogge: Thought about that but we don't do this for the rich strategy.

TylerLeonhardt (approved):

PR #264029, opened 2025-08-29
Allows monaco-editor createWorker to accept esmModuleLocation to avoid origin issues and forward localization strings correctly.

hediet:


src/vs/editor/standalone/browser/standaloneServices.ts
```diff
@@ -1076,7 +1076,7 @@ class StandaloneContextMenuService extends ContextMenuService {
 }
 
 const standaloneEditorWorkerDescriptor: IWebWorkerDescriptor = {
-	esmModuleLocation: undefined,
+	esmModuleLocation: new URL('../../common/services/editorWebWorkerMain.js', import.meta.url),
```
Tyriar, line 1079: Red CI

Tyriar (changes requested):

Issue #263896, opened 2025-08-28
Optimize rendering of long chat histories


mjbvz:
From #263496

For chats with long histories, we currently end up rendering the entire history on first load. This can cause big performance issues for users who never clear their chat history

I've debugged through this and have a few ideas we could try

- Paginate loading of the chat history

    Use pagination to only load the most recent 20 or so items. When you scroll back, we could then dynamically load in the previous 20 items as needed

    The main downside of this is that the scrollbars wouldn't be accurate. Without some thought it also would be difficult to jump back to the first item in the history

- Optimize tree/list rendering for cases where you start at the bottom of the list

    Our list rendering works from the top of the list on down. In chat though, we typically are most interested in showing content at the bottom of the list. This requires going through the entire list first just to find the handful of items we need to render

-  #263525 Optimize `NativeEditContext` for when we create many editors

- #263499: Break up long responses so that they can be rendered more efficiently

roblourens:
> It would nice if we could avoid the initial rendering if we know that we will showing content at the bottom of the list

The chat tree might be doing something wrong here in how we get the initial heights of elements or set the initial scroll position. Maybe if we remembered the previous element heights and started with that instead of some small default value, we'd be closer to correct on the first render.

isidorn:
I thought lazy rendering is built into the list widget, and @joaomoreno might be able to provide some guidance here.

joaomoreno:
If this was a list instead of a tree, you'd immediately get less overhead at runtime. I don't really know what tree features the chat widget relies on, but since I don't see any twisties or indentation, I assume there aren't many. If this were to move to a list, optimizing the "append only" case would be much simpler.

roblourens:
I think I moved it to a tree because only the tree supported searching. Then I never got searching to work and forgot that it was still a tree...

> I thought lazy rendering is built into the list widget

It is, the problem is that our rows can be very long, see https://github.com/microsoft/vscode/issues/263499#issuecomment-3226223589

isidorn:
> I think I moved it to a tree because only the tree supported searching. Then I never got searching to work and forgot that it was still a tree...

Sounds like we should first move it back to a list before we do any premature optimizations...

roblourens:
[...]

PR #264043, opened 2025-08-29
Avoid extra HTML re-parse when rendering markdown

mjbvz:
For #263896

Uses the dom fragment from dompurify instead of converting to a string and then re-parsing back to dom. This avoids this small cost every time markdown is rendered:

Screenshot 2025-08-29 at 9 40 46 AM


Tyriar (dismissed):

src/vs/workbench/contrib/chat/browser/chatMarkdownRenderer.ts
```diff
@@ -100,6 +100,7 @@ export class ChatMarkdownRenderer extends MarkdownRenderer {
 		// In some cases, the renderer can return text that is not inside a <p>,
 		// but our CSS expects text to be in a <p> for margin to be applied properly.
 		// So just normalize it.
+		result.element.normalize();
```
mjbvz, line 103: I'll create a separate PR for September that looks into this code. I think we actually need to do this replacement for all children, not just the last one. Otherwise you can still end up with top level text nodes

Tyriar (approved):

PR #264049, opened 2025-08-29
Make sure all text nodes in chat markdown are wrapped in elements

mjbvz:
Follow up on #264043

I'm pretty sure the tests were showing bugs where top level text nodes could still end up the result. With this change all text nodes should be wrapped in a `<p>`


src/vs/workbench/contrib/chat/browser/chatMarkdownRenderer.ts
```diff
@@ -97,12 +97,14 @@ export class ChatMarkdownRenderer extends MarkdownRenderer {
...
-		// but our CSS expects text to be in a <p> for margin to be applied properly.
+		// In some cases, the renderer can return top level text nodes  but our CSS expects
+		// all text to be in a <p> for margin to be applied properly.
 		// So just normalize it.
-		const lastChild = result.element.lastChild;
-		if (lastChild?.nodeType === Node.TEXT_NODE && lastChild.textContent?.trim()) {
-			lastChild.replaceWith($('p', undefined, lastChild.textContent));
+		result.element.normalize();
+		for (const child of result.element.childNodes) {
+			if (child.nodeType === Node.TEXT_NODE && child.textContent?.trim()) {
```
roblourens, line 105: Sure but is it weird to do this in chatMarkdownRenderer? 😅

roblourens (dismissed):

src/vs/workbench/contrib/chat/browser/chatMarkdownRenderer.ts
mjbvz: Yes definitely would love to remove this! I think this can only happen when `replaceWithPlaintext` is on so I'll see if I can fix that at some point too

roblourens (approved):

PR #264061, opened 2025-08-29
Allow rendering a theme icon with a chat model

lramos15:


src/vscode-dts/vscode.proposed.chatProvider.d.ts
```diff
@@ -3,7 +3,7 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-// version: 3
+// version: 4
```
roblourens, line 6: It's ok but since you're just adding an optional prop, you don't really need to update the api version.

roblourens (approved):

PR #264107, opened 2025-08-29
Chat sessions view experiment update

osortega:


joshspicer (dismissed):

src/vs/workbench/contrib/chat/browser/chat.contribution.ts
```diff
@@ -431,7 +431,10 @@ configurationRegistry.registerConfiguration({
 			enum: ['disabled', 'view'],
 			description: nls.localize('chat.sessionsViewLocation.description', "Controls where to show the agent sessions menu."),
 			default: 'disabled',
-			tags: ['experimental', 'onExp'],
+			tags: ['experimental', 'onExP'],
```
luabud: I think you can drop `onExP` in this case!

luabud (approved):

Issue #206480, opened 2024-02-28
Git Commit Message Wraps lines at 50 characters instead of 72


theory:
Type: <b>Bug</b>

*   Use Git for editing Git commit messages
*   Use it to edit a message (the name in the box in the bottom right corner is "Git Commit Message"
*   Note the wrap lines: one at 50 characters for the subject, another at 78 characters for the body
*   Place the cursor in some long body text
*   Use option-q (or alt-q?) to wrap the long text
*   See that it wraps at the 50 character line instead of 72

FWIW it wrapped at 72 characters until a few weeks ago.

VS Code version: Code 1.87.0 (Universal) (019f4d1419fbc8219a181fab7892ebccf7ee29a2, 2024-02-27T23:42:56.944Z)
OS version: Darwin arm64 23.3.0
Modes:

<details>
<summary>System Info</summary>

|Item|Value|
|---|---|
|CPUs|Apple M3 Max (14 x 24)|
|GPU Status|2d_canvas: enabled<br>canvas_oop_rasterization: enabled_on<br>direct_rendering_display_compositor: disabled_off_ok<br>gpu_compositing: enabled<br>multiple_raster_threads: enabled_on<br>opengl: enabled_on<br>rasterization: enabled<br>raw_draw: disabled_off_ok<br>skia_graphite: disabled_off<br>video_decode: enabled<br>video_encode: enabled<br>webgl: enabled<br>webgl2: enabled<br>webgpu: enabled|
|Load (avg)|2, 2, 2|
|Memory (System)|36.00GB (0.13GB free)|
|Process Argv|--crash-reporter-id 0f3c5b21-cecd-40d4-b17d-e913247e5700|
|Screen Reader|no|
|VM|0%|
</details><details><summary>Extensions (33)</summary>

Extension|Author (truncated)|Version
---|---|---
vscode-openapi|42C|4.21.1
terraform|4op|0.2.5
vscode-sql-formatter|adp|1.4.4
vscode-multimarkdown|and|0.1.1
vscode-abnf|arn|0.2.0
markdown-mermaid|bie|1.21.0
mermaid-markdown-syntax-highlighting|bpr|1.6.0
language-hugo-vscode|bud|1.3.1
perl-toolbox|d97|1.4.0
yash|dao|0.2.9
xml|Dot|2.5.1
EditorConfig|Edi|0.16.4
varnish-vcl|flo|0.1.3
go|gol|0.40.3
hcl|has|0.3.2
[...]

PR #264133, opened 2025-08-30
Git: Remove problematic ruler at 50 chars

walles:
The ruler at column 50, introduced in #201091, is wrong for all lines except for the first. Also, having it there breaks interaction with the (quite popular) rewrap extension:

https://marketplace.visualstudio.com/items?itemName=stkb.rewrap

The best solution would be if we could have different rulers for different lines, (50 for line 1, 0 for line 2 and 72 for the rest), but since that's not possible (or is it?), I believe dropping 50 is the best way forward.

Regarding the user experience, overflowing chars are highlighted anyway since #173195, so users who write too long subject lines will still be informed, even without that ruler in place.

Since I use rewrap all the time, and I commit things all the time, the current behavior is problematic for me.

"rewrap" uses the rulers to decide where to wrap text.

Before #201091, it wrapped at 72 chars, which was great!

With #201091 in place, it now wraps all lines at 50 chars, which is bad.

### DIFF
--- a/extensions/git/package.json
+++ b/extensions/git/package.json
@@ -3590,7 +3590,6 @@
     "configurationDefaults": {
       "[git-commit]": {
         "editor.rulers": [
-          50,
           72
         ],
         "editor.wordWrap": "off",


lszomoru:
To remove the ruler at 50 characters add the following to your user settings:

```json
  "[git-commit]": {
    "editor.rulers": [
      72
    ]
  }
```

walles:
@lszomoru, I'd like to understand the cost / benefit of having that ruler at column `50`.

Since I only see problems with it, I'd like to hear what the argument is for having it.

What value does the ruler at column 50 bring? Who would miss it if it was removed?

It was introduced here: https://github.com/microsoft/vscode/pull/201091

theory:
I worked around the issue with line wrapping by reversing the order of the rulers:

```json
  "[git-commit]": {
    "editor.rulers": [72, 50],
    "editor.wordWrap": "on"
  },
```

walles:
But these are just workarounds that will help just me.

Why is having `50` in there important to begin with? Whom does it benefit?

The purpose of having `50` in there can't be just to force people to work around it .

lszomoru:
@walles, you are the first user that complains about this issue, and instead of filing an issue to get the features owner and the community's take on this you went ahead and create a pull request instead. Given that fact that people that does not like the **default** rules can adjust them through a user setting, I do not want to incur the cost of adjusting the default behavior given the fact that nobody complained about this until now.

walles:
This is the first complaint that I could find: https://github.com/microsoft/vscode/issues/206480

But what *is* the value of having a ruler at column 50?

lszomoru:
The ruler at column 50 is for the commit message subject.

walles:
Right, but for whom does having this ruler bring value?

My point is that if it doesn't bring value to anybody, but is causing issues, then maybe it should be removed?

lszomoru:
> Right, but for whom does having this ruler bring value?

I cannot answer that question, as people only file issues for things that they do not like...

walles:
So [the community reported the ruler is causing problems](https://github.com/microsoft/vscode/issues/206480), and [there's no documented use case or user for it](https://github.com/microsoft/vscode/pull/201091).

What would be the requirements for having the ruler at column 50 removed?

Here's what typing a > 50 chars subject line looks like without it (Dark+ default theme).

Skärmavbild 2025-09-01 kl  20 22 07

PR #264219, opened 2025-08-31
win32(installer): Normalize PATH handling and fix uninstall filtering [needs testing]

yugook:
⚠️ This PR updates PATH handling logic in the Windows installer.
Changes are verified with string-based tests (PowerShell), but I would appreciate further testing on real Windows setups.

## Summary
Normalize PATH handling in the Windows installer (`build/win32/code.iss`) and fix uninstall filtering:
- Trim trailing separators when comparing (`\` and `/`)
- Case-insensitive comparison
- Consider 8.3 short names (when available)
- Handle empty / semicolon-terminated PATH without producing malformed values

## Why
- Avoid duplicate PATH entries when the existing entry has a trailing slash or is in 8.3 short form.
- Ensure uninstall removes VS Code’s PATH entry (both long and short forms).

## How
- Add `TrimTrailingSlash` helper.
- `NeedsAddToPath`: explode PATH, normalize each segment, compare against long and short names.
- `AddToPath`: treat empty/absent PATH specially; respect trailing `;`.
- Uninstall: normalize `{app}\bin`, compare against both long/short, rebuild PATH cleanly.

## Before / After (examples)
- **Before**: PATH contains `…;C:\Program Files\Microsoft VS Code\bin\` → installer adds `…;C:\Program Files\Microsoft VS Code\bin` (duplicate).
- **After**: no duplicate added.
- **Before**: PATH contains `…;C:\PROGRA~1\MICROS~1\bin` → uninstall sometimes leaves it behind.
- **After**: long/short both removed.

## Verification
- Windows Sandbox, PowerShell dry-run (strings only; no system writes):
  - trailing slash present → `NeedsAddToPath: False`; `RemoveFromPath`: entry removed
  - 8.3 short present → `NeedsAddToPath: False`; `RemoveFromPath`: entry removed
  - empty PATH → `AddToPath`: `C:\Program Files\Microsoft VS Code\bin`
  - PATH ends with `;` → no malformed value produced
(Full installer tests can be run if preferred.)

## Risk / Impact
Low: changes are isolated to installer script and guarded when 8.3 names are not available.

### DIFF
--- a/build/win32/code.iss
+++ b/build/win32/code.iss
@@ -1590,16 +1590,42 @@ begin
   until Length(Text)=0;
 end;
 
+// Normalize path string by trimming any trailing path separators
+function TrimTrailingSlash(p: string): string;
+begin
+  while (Length(p) > 0) and ((p[Length(p)] = '\\') or (p[Length(p)] = '/')) do
+    SetLength(p, Length(p) - 1);
+  Result := p;
+end;
+
 function NeedsAddToPath(VSCode: string): boolean;
 var
-  OrigPath: string;
+  OrigPath, seg, target, targetShort: string;
+  parts: TArrayOfString;
+  i: Integer;
 begin
-  if not RegQueryStringValue({#EnvironmentRootKey}, '{#EnvironmentKey}', 'Path', OrigPath)
-  then begin
+  if not RegQueryStringValue({#EnvironmentRootKey}, '{#EnvironmentKey}', 'Path', OrigPath) then begin
     Result := True;
     exit;
   end;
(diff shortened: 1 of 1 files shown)


yugook:

yugook:
Thanks again for taking a look.

I’m going to close this PR for now. I haven’t been able to validate it with enough confidence on a real Windows setup, and the behavior I checked on my work machine may have been affected by permission constraints rather than the PATH handling itself.

Since I don’t have reliable verification for the issue or the fix, I’d rather close this than leave an uncertain change open.

If I can retest it properly and confirm the behavior, I’ll open a new PR with clearer evidence. Thanks again.

Issue #252449, opened 2025-06-25
`terminal.findMatchBackground` not working


v-amorim:
Does this issue occur when all extensions are disabled?: Yes/No

- VS Code Version: 1.101.2
- OS Version: 23H2

Steps to Reproduce:

1. Edit the `workbench.colorCustomizations` `terminal.findMatchBackground` color.
2. See that it doesn't change the selected match color.

Print comparing the terminal and editor colors (should to be the same)
![Image](https://github.com/user-attachments/assets/6e947b1f-7b40-4c9c-a11f-d963ce61964f)

v-amorim:
Updated to latest version `1.101.2` and issue persists as the evidence.

Tyriar:
Looks like something fishy is going on. Probably some bug in xterm.js' color blending code

![Image](https://github.com/user-attachments/assets/47f4d1ca-40de-4a27-809f-8cd8c56f93f3)

Tyriar:
https://github.com/xtermjs/xterm.js/issues/5369

kevlarr:
Is this a blending issue or just VS Code ignoring `terminal.findMatchBackground`? Seems like no matter what I set the match background to for the terminal, the actual highlight color seems to be based off of `terminal.selectionBackground` (not exactly matching it but seems to be blended with the overall background color even when opaque).

Tyriar:
@kevlarr looks like you're right! So we're probably either not mapping the colors correctly from vscode theme -> xterm.js theme, or there's a bug applying that color in the xterm.js search addon.

jmg-duarte:
~`findMatchHighlightBorder` also seems broken~ after reloading it worked

<img width="649" height="237" alt="Image" src="https://github.com/user-attachments/assets/82eb8dd5-703e-45c3-bdc2-a24e5c0405a1" />

hyt658:
Version: 1.104.2 (Universal)
Commit: e3a5acfb517a443235981655413d566533107e92
Date: 2025-09-24T11:21:37.073Z
Electron: 37.3.1
ElectronBuildId: 12404162
Chromium: 138.0.7204.235
Node.js: 22.18.0
V8: 13.8.258.31-electron.0
OS: Darwin arm64 25.0.0

I met the same issue. However the color shows out when I keep typing in the terminal. Once I stopped, it turns back to gray. 
<img width="2886" height="550" alt="Image" src="https://github.com/user-attachments/assets/952a7190-394d-4134-b964-807f4544347c" />

[...]

PR #264228, opened 2025-08-31
Fix localization for terminal.hoverHighlightBackground

jmg-duarte:
Found while researching https://github.com/microsoft/vscode/issues/252449

Just replaces the wrong key with what should be the right one.

I took the description from the [theme color reference](https://code.visualstudio.com/api/references/theme-color)


Tyriar (approved):
Good find 👍

lramos15 (approved):

PR #264230, opened 2025-08-31
Fix localization for terminal.initialHintForeground

jmg-duarte:
Similar to #264228


Tyriar (commented):
This is actually harmless, the localization keys aren't used in the product, just to uniquely identify and to use as a hint for the translators. Changing it I believe triggers re-translation.

PR #264377, opened 2025-09-01
Adjust dropdown positioning in settings editor to prevent distance from controls

### LINKED ISSUE #264357: Settings editor workspace folder selector dropdown opens too far away from tab
Bug Description:

Setting tab dropdown opens too far away from control

Screenshot:

Image

Other context menus & dropdowns open at the bottom right of the control:

Image

Image

codekalinga:
Fixes #264357

## Problem
Dropdown menus in the settings editor were opening too far away from their trigger controls, causing visual disconnect and poor UX.

## Solution
Added a 10px offset specifically for dropdown positioning in the settings editor by detecting the `.settings-editor` class context.

## Changes
- Modified `getDomNodePagePosition()` in `src/vs/base/browser/dom.ts`
- Added settings page detection using `closest('.settings-editor')` 
- Applied 10px vertical offset only for settings editor dropdowns

## Testing
Verified dropdowns now open with proper proximity to controls in settings editor while maintaining normal behavior elsewhere.


rzhao271:
Not sure which dropdowns you're referring to. Could you add screenshots?

bpasero:
@rzhao271 I think this is for:

image

But the fix is not good.

PR #264611, opened 2025-09-02
chat history cleanup

### LINKED ISSUE #264603: Chat history hover, select, and focus states
Testing #264269

From  @justschen (thanks!)
> The selection persists when creating a new window after clicking on it to go to that history item. Also loses the blue focus border in this state.

https://github.com/user-attachments/assets/17123c35-f28f-4590-82ad-3a00137dc014

justschen:
fix https://github.com/microsoft/vscode/issues/264603


src/vs/workbench/contrib/chat/browser/chatWidget.ts
```diff
@@ -1022,13 +1023,9 @@ export class ChatWidget extends Disposable implements IChatWidget {
 					}, 0);
 				}
 			}));
-			const listEl = dom.append(container, $('.chat-welcome-history-list'));
+			const welcomeHistoryList = dom.append(container, $('.chat-welcome-history-list'));
```
roblourens, line 1026: Why isn't this stuff in `ChatViewWelcomePart`? It's essentially part of the welcome view right?

src/vs/workbench/contrib/chat/browser/chatWidget.ts
```diff
@@ -1050,17 +1047,15 @@ export class ChatWidget extends Disposable implements IChatWidget {
...
+			welcomeHistoryList.style.overflow = 'hidden';
 
-			// Set explicit height for the list to accommodate exactly the items we have
-			const listHeight = historyItems.length * 22; // 22px per item as defined in delegate
-			listEl.style.height = `${listHeight}px`;
-			listEl.style.minHeight = `${listHeight}px`;
-			listEl.style.overflow = 'hidden';
-
-			const historyList = this._register(this.instantiationService.createInstance(
+			this.historyList = this._register(this.instantiationService.createInstance(
```
roblourens, line 1055: Isn't all this stuff leaking? Because the welcome view is rerendered many times, that's why we set it up with the `MutableDisposable`, but this is registering all these things to the ChatWidget and not disposing them. And I don't think we should recreate a whole list from scratch repeatedly to render this. There is much more opportunity to clean this up, if we're going to build it let's do it right.

roblourens (approved):

src/vs/workbench/contrib/chat/browser/chatWidget.ts
justschen: welcomeHistoryList is a bit misleading, it's shown on every empty state. i'll rename this

src/vs/workbench/contrib/chat/browser/chatWidget.ts
justschen: taking another look at this. i'll make sure we're not leaking first and since we have a template, i think we can just update parts when new sessions come in 👍🏻

PR #264836, opened 2025-09-03
Instrument instructions collection

digitarald:


aeschli:
Note that the AgentInstructions also contain references inside `copilot-instructions.md`. Noy sure you wanted that.

The alternative implementation is to add a `kind` to the `toPromptFileVariableEntry` call and you can count them at the end.

aeschli (dismissed):

aeschli (approved):

dbaeumer (approved):

PR #264867, opened 2025-09-03
Don't fail if client does not implement "vscode/content" request

Maddimax:
Sending back an error to the client is unnecessary. It will only create noise.

### DIFF
--- a/extensions/json-language-features/server/src/jsonServer.ts
+++ b/extensions/json-language-features/server/src/jsonServer.ts
@@ -103,6 +103,9 @@ export function startServer(connection: Connection, runtime: RuntimeEnvironment)
 			return connection.sendRequest(VSCodeContentRequest.type, uri).then(responseText => {
 				return responseText;
 			}, error => {
+				if (error.code == -32601) {
+					return "{}";
+				}
 				return Promise.reject(error.message);
 			});
 		};


aeschli:
But it's wrong to return `{}` as schema.
We want an error to be throw so this can be shows as a validation error.

Maddimax:
But this is an optional, non-standard extension, that only works if the specific schema is implemented by the client. I don't want an error in that case.

Maddimax:
> If you want to solve this properly, add new option to LanguageClientOptions

I'd love to. Can you give me a bit more info on how you see that working? Do you mean a new option for [client.ts](https://github.com/microsoft/vscode-languageserver-node/blob/3412a17149850f445bf35b4ad71148cfe5f8411e/client/src/common/client.ts#L347)?

Should the client tell the server if its able to provide vscode/content?

aeschli:
As part the `initializationOptions`, the client can give information to the server.
e.g. `handledSchemaProtocols` tells the server to only load `file` schemas by itself and to otherwise forward back to the client using the  "vscode/content" request. 
https://github.com/microsoft/vscode/blob/c738ec6c40099671c763e1ee7023e321589cf815/extensions/json-language-features/client/src/jsonClient.ts#L262
On the server its handled here:
https://github.com/microsoft/vscode/blob/c738ec6c40099671c763e1ee7023e321589cf815/extensions/json-language-features/server/src/jsonServer.ts#L146

You could add a new property, or extend 'handledSchemaProtocols' with a new value '*' to signal that the server should handle all URIs itself and never call 'vscode/content'.

aeschli:
Pretending the schema is `{}` is not correct, at least it has to fail with an 'unable to load schema' error. I'm fine if it doesn't make a `vscode/content` request if you add an option for that.
Note that in `vscode` is just one of many virtual file systems. VS Code has an API for extensions to register new file systems. Only the client running in the extension host can handle these which is why the server forwards these calls to the client.

Why does your server wants to load JSON schemas from `vscode` URLs? All schema URI's are configured via ISchemaAssociation's. If you make sure you only schemas for `file` and `http`, you will not get any `vscode/content` requests.

Maddimax:
> Why does your server wants to load JSON schemas from `vscode` URLs? All schema URI's are configured via ISchemaAssociation's. If you make sure you only schemas for `file` and `http`, you will not get any `vscode/content` requests.

We would like to load the devcontainer schema, which in turn then tries to load vscode specific content (for the customization settings part)

aeschli:
ok, I see, that's not so cool/clean of the devcontainer schema and shows a weakness of their approach to merge editor/IDE specific schemas.
Can you only use the devcontainer base schema ? https://containers.dev/implementors/json_schema/#base-schema

Alternatively you can start the JSON schema server with a list of preconfigured schemas in the options such as `vscode://schemas/settings/machine` and `vscode://schemas/mcp`.

I know, not pretty, but just ignore calls to unknown schemas is also a hack.

Also, I want to mention that language servers typically need to be accompanied by a matching language client. You can't expect a server to be used 'as-is'. In the case of JSON, the  https://github.com/microsoft/vscode/blob/b48aefcad302d0d7000ed8b14096c2112dbc5fd7/extensions/json-language-features/server/README.md#L133 has all the requirements that are needed to run the server.

Maddimax:
I guess the only true solution would be to add "optional $ref"s to the json schema. Anything else is a hack I think.

Issue #263098, opened 2025-08-23
Caret jumps to column 0 when clicking between first two characters


alexchexes:
Does this issue occur when all extensions are disabled?: Yes

Clicking between the first and second characters in a line places the caret at column 0 instead of between the characters. Please see the screencast.

https://github.com/user-attachments/assets/ade140c6-8b40-4d1d-afda-8ff57dd3838d

Steps to Reproduce:

1. Open a new file and type: `let x`
2. Click between `l` and `e` in `let`
3. Observe the caret position

Expected:
Caret is placed between `l` and `e` (column 1).

Actual:
Caret is placed before `l` (column 0).

Notes:

- Only the first column is affected. Other positions (e.g., between `e` and `t`) work correctly.
- Reproduces in both **Stable** and the latest **Insiders** version, checked with a clean installation / logged-out.
- The screencast uses increased UI scale for visibility, but the issue also occurs with the default VSCode and Windows display scaling.

RedCMD:
can confirm

lszomoru:
@aiday-mar, could you please help triage this issue? Thank you!

RedCMD:
can we make this a candidate please?

aiday-mar:
Hi thanks for letting me know of this. I will look at this next week during debt week.

aiday-mar:
By using the vscode bisect tool and the git bisect tool I have found the PR https://github.com/microsoft/vscode/pull/235174 introduces this issue. I will further investigate how and why this happens.

aiday-mar:
To add more context about the issue above, the PR incorrectly computes that the cursor is outside of the editor, when it is inside of the editor in the following check:

```
		if (e.relativePos.x <= xLeftBoundary) {
			const outsideDistance = xLeftBoundary - e.relativePos.x;
			const position = new Position(possibleLineNumber, 1);			
            return MouseTarget.createOutsideEditor(mouseColumn, position, 'left', outsideDistance);
		}
```

ttttotem:
Thanks @aiday-mar, I've created https://github.com/microsoft/vscode/pull/264914 to try fix this

aiday-mar:
Thanks @ttttotem I will have a look at this PR.

aiday-mar:
I have looked at the PR and have asked some questions on it, which aim to help me understand this PR before approving.

alexdima:
Fixed by https://github.com/microsoft/vscode/pull/265131

PR #235174, opened 2024-12-03
Horizontal dragging auto-scroll

ttttotem:
Issue: #172652
Problem: horizontal dragging does not autoscroll
Solution:
Extension of vertical autoscroll introduced in #161378
Example:
https://github.com/user-attachments/assets/cee1f841-9a00-42bd-844d-9c4d4cecf1f6

PR #264914, opened 2025-09-03
Fix for issue 263098

ttttotem:
Fixes issue https://github.com/microsoft/vscode/issues/263098 that was caused by https://github.com/microsoft/vscode/pull/235174

Removing the padding except where it was needed for the minimap shadow should fix issue 263098 "Caret jumps to column 0 when clicking between first two characters"

### DIFF
--- a/src/vs/editor/browser/controller/mouseHandler.ts
+++ b/src/vs/editor/browser/controller/mouseHandler.ts
@@ -578,14 +578,12 @@ class MouseDownOperation extends Disposable {
 			return MouseTarget.createOutsideEditor(mouseColumn, new Position(possibleLineNumber, 1), 'left', outsideDistance);
 		}
 
-		const minimapShadowPadding = 10;
-
-		const contentRight = (
+		const minimapShadowPadding = 6; // See minimap-shadow-visible in minimap.css
+		const xRightBoundary = (
 			layoutInfo.minimap.minimapLeft === 0
 				? layoutInfo.width - layoutInfo.verticalScrollbarWidth // Happens when minimap is hidden
 				: layoutInfo.minimap.minimapLeft - minimapShadowPadding // Minimap needs padding because of its shadow
 		);
-		const xRightBoundary = contentRight;
 		if (e.relativePos.x >= xRightBoundary) {
 			const outsideDistance = e.relativePos.x - xRightBoundary;
 			return MouseTarget.createOutsideEditor(mouseColumn, new Position(possibleLineNumber, model.getLineMaxColumn(possibleLineNumber)), 'right', outsideDistance);
--- a/src/vs/editor/browser/controller/mouseHandler.ts
+++ b/src/vs/editor/browser/controller/mouseHandler.ts
@@ -573,7 +573,7 @@ class MouseDownOperation extends Disposable {
 		const horizontalScrollPadding = 10;
 		const layoutInfo = this._context.configuration.options.get(EditorOption.layoutInfo);
 
-		const xLeftBoundary = layoutInfo.contentLeft + horizontalScrollPadding;
+		const xLeftBoundary = layoutInfo.contentLeft;
 		if (e.relativePos.x <= xLeftBoundary) {
 			const outsideDistance = xLeftBoundary - e.relativePos.x;
 			return MouseTarget.createOutsideEditor(mouseColumn, new Position(possibleLineNumber, 1), 'left', outsideDistance);
--- a/src/vs/editor/browser/controller/mouseHandler.ts
+++ b/src/vs/editor/browser/controller/mouseHandler.ts
@@ -582,9 +582,9 @@ class MouseDownOperation extends Disposable {
 		const contentRight = (
 			layoutInfo.minimap.minimapLeft === 0
 				? layoutInfo.width - layoutInfo.verticalScrollbarWidth // Happens when minimap is hidden
-				: layoutInfo.minimap.minimapLeft
+				: layoutInfo.minimap.minimapLeft - horizontalScrollPadding // Minimap needs padding because of shadow
 		);
-		const xRightBoundary = contentRight - horizontalScrollPadding;
+		const xRightBoundary = contentRight;
 		if (e.relativePos.x >= xRightBoundary) {
 			const outsideDistance = e.relativePos.x - xRightBoundary;
 			return MouseTarget.createOutsideEditor(mouseColumn, new Position(possibleLineNumber, model.getLineMaxColumn(possibleLineNumber)), 'right', outsideDistance);
--- a/src/vs/editor/browser/controller/mouseHandler.ts
+++ b/src/vs/editor/browser/controller/mouseHandler.ts
@@ -570,7 +570,6 @@ class MouseDownOperation extends Disposable {
 
 		const possibleLineNumber = viewLayout.getLineNumberAtVerticalOffset(viewLayout.getCurrentScrollTop() + e.relativePos.y);
 
-		const horizontalScrollPadding = 10;
 		const layoutInfo = this._context.configuration.options.get(EditorOption.layoutInfo);
 
 		const xLeftBoundary = layoutInfo.contentLeft;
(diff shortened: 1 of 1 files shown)


src/vs/editor/browser/controller/mouseHandler.ts
```diff
@@ -570,21 +570,20 @@ class MouseDownOperation extends Disposable {
...
 
-		const xLeftBoundary = layoutInfo.contentLeft + horizontalScrollPadding;
+		const xLeftBoundary = layoutInfo.contentLeft;
 		if (e.relativePos.x <= xLeftBoundary) {
 			const outsideDistance = xLeftBoundary - e.relativePos.x;
 			return MouseTarget.createOutsideEditor(mouseColumn, new Position(possibleLineNumber, 1), 'left', outsideDistance);
 		}
 
-		const contentRight = (
+		const minimapShadowPadding = 6; // See minimap-shadow-visible in minimap.css
```
ttttotem, line 581: Not sure how to set this padding amount without hard coding it

src/vs/editor/browser/controller/mouseHandler.ts
```diff
@@ -570,21 +570,20 @@ class MouseDownOperation extends Disposable {
 
 		const possibleLineNumber = viewLayout.getLineNumberAtVerticalOffset(viewLayout.getCurrentScrollTop() + e.relativePos.y);
 
-		const horizontalScrollPadding = 10;
 		const layoutInfo = this._context.configuration.options.get(EditorOption.layoutInfo);
 
-		const xLeftBoundary = layoutInfo.contentLeft + horizontalScrollPadding;
+		const xLeftBoundary = layoutInfo.contentLeft;
```
aiday-mar, line 575: This variable `horizontalScrollPadding` was removed. I was wondering why this variable was initially added?

src/vs/editor/browser/controller/mouseHandler.ts
aiday-mar: Can you please explain what this variable represents? Why is it removed from the `xRightBoundary`? What do `xLeftBoundary` and `xRightBoundary` represent?

aiday-mar (commented):
Hi @ttttotem can you please help me understand this change by answering the questions? I would like to understand the following:
- What the initial issue was
- What the proposed solution is and how it fixes the issue

src/vs/editor/browser/controller/mouseHandler.ts
ttttotem: So please excuse the terible pictures, but when I was trying to add horizontal scrolling the shadow of the minimap was causing issues because it registers as outside the editor content
image

So to find where the editor content actually ends I need to add that extra bit of width covered by the shadow. This is what horizontalScrollPadding represented, it is now called minimapShadowPadding. If layoutInfo.minimap.minimapLeft accounted for the width of the shadow I would not need minimapShadowPadding:
image

This "Caret jumps to column 0" issue was caused by the padding on the left side of the editor content where there is no shadow, and therefore no padding should have been added
image

So this PR is to remove the padding/offset where there is no shadow, and to use the exact width of the shadow rather than a bigger value.

src/vs/editor/browser/controller/mouseHandler.ts
ttttotem: So the idea is if you are highlighting text and your mouse is left of xLeftBoundary then you should be scrolling left, likewise if your mouse is right of xRightBoundary then you should be scrolling right. We want to catch when the mouse leaves the editor content like what was already implmemented for vertical scrolling

xLeftBoundary is now just layoutInfo.contentLeft
xRightBoundary would be layoutInfo.contentRight if it existed

aiday-mar: Okay that makes sense thanks

src/vs/editor/browser/controller/mouseHandler.ts
aiday-mar: When you mentioned:

> So please excuse the terible pictures, but when I was trying to add horizontal scrolling the shadow of the minimap was causing issues because it registers as outside the editor content

What is the issue you are referring to exactly? Can you demonstrate the issue that happens otherwise? I would like to understand why you need to add this hard-coded value. We try to avoid hard-coded values where possible

ttttotem: Yeah so the issue was when you touch the shadow you jump to the end of the line (how it used to work before https://github.com/microsoft/vscode/pull/235174). This is because the shadow triggers a mouse event with MouseTargetType.SCROLLBAR when entered. So xRightBoundary needs to exactly at or before the shadow.

Putting xRightBoundary at layoutInfo.minimap.minimapLeft is 6 pixels too far right because it doesn't account for the minimap's 6 pixel shadow

https://github.com/user-attachments/assets/d23da71b-bbe9-41f9-810d-05da035c0fd4

aiday-mar: So to be clear are you saying this variable was added to fix an issue that existed already before? In that case this can be addressed in a separate PR, and does not need to be addressed here. 

If this issue already existed before, we could remove the padding you added and ship with a simpler fix. Please let me know if I understood correctly.

ttttotem: In this PR I'm just renaming the existing variable horizontalScrollPadding that was introduced by #235174 to the more fitting name minimapShadowPadding and setting it to a more accurate value, its not essential for fixing the raised issue.

If you would like a PR that only fixes issue #263098 and changes nothing else then this would be it https://github.com/microsoft/vscode/pull/265131

aiday-mar: The PR that you linked does not remove the variable horizontalScrollPadding variable. My suggestion was to not use it in the case of the right border either, because as you seemed to mention it fixes another issue which was already an issue prior to all this work.

aiday-mar:
I pushed the suggestion I mentioned in the review into your second PR : https://github.com/microsoft/vscode/pull/265131. 

The reason this is necessary is because with the initial change you made there remained still another issue. When I clicked on the second to last character column before the viewport end, the cursor would jump to the end of the line. This is essentially happening because of the padding you were removing on the right hand side, and is the same issue as the one where when you click on column 2 it takes us to column 1. I am discussing this issue with my team at the moment.

src/vs/editor/browser/controller/mouseHandler.ts
ttttotem: I meant that horizontalScrollPadding on the right was needed to implement horizontal drag scrolling https://github.com/microsoft/vscode/pull/235174. As in it fixed a problem I was having when trying to implement https://github.com/microsoft/vscode/pull/235174 not that it was fixing a seperate existing issue.

If you remove horizontalScrollPadding on the right then you cannot drag scroll to the right. In which case you should probably just roll back https://github.com/microsoft/vscode/pull/235174, otherwise you will only be able to drag scroll to the left which might be a little weird

aiday-mar: I am currently running the code on the other PR with the new commit , and I am able to drag scroll to the right. Please let me know if I am misunderstanding what you mean by horizontal scrolling.

https://github.com/user-attachments/assets/a0f91dd8-ac3e-4bf1-af1e-8a2fb363853d

ttttotem: Ah yeah, you're right sorry, drag scroll to the right will work fine as long as you have the minimap disabled

aiday-mar: I see, do you know if there is a way to make this work without adding a padding, or adding a padding in such a manner so that clicking on the right side does not place the cursor at the end of the line?

aiday-mar: So your fix was the following as I understand:

When the minimap is enabled, add a padding, so smooth horizontal scrolling is possible to the right
BUT this introduces the bug that clicking on the last character puts the cursor on the last column

When the minimap is disabled, it all works correctly.

ttttotem: Hmm, well disabling pointer events on the shadow would remove the need for padding. I've added a commit to the other PR to show what I mean which seems to work but I have no idea if this could affect some other completely different functionality

PR #264937, opened 2025-09-03
fix: memory leak in ReplAccessibilityAnnouncer

SimonSiefke:
Before (Running Debug 197 times):
Untitled

This changes the disposable to be a mutableDisposable. Since there seems to be only one global `ReplAccessibilityAnnouncer`, registering another disposable onto it when the debug session focus changes seems to make the number of disposables grow by one each time.


meganrogge (approved):
Thanks!

Tyriar (approved):

PR #265002, opened 2025-09-03
fix: memory leak in chat widget

SimonSiefke:
Before (Opening and Closing Chat Widget 197 times): 

Untitled

It seems that each time opening the chat widget, a new timeout disposable is registered, which seems to make the number of disposables grow by one each time.

The change uses a MutableDisposable to clear the previous timeout disposable when the setVisible function is called.


bpasero (approved):

lszomoru (approved):

justschen:
git blame says tyler, but this is a good change regardless! thanks @SimonSiefke

PR #265011, opened 2025-09-03
Fix to use arrow function in callback function

osortega:
Fixes to use arrow function in update callback to bind `this`


src/vs/workbench/contrib/chat/browser/chatStatus.ts
```diff
@@ -129,7 +129,7 @@ export class ChatStatusBarEntry extends Disposable implements IWorkbenchContribu
 		this.registerListeners();
 	}
 
-	private update(): void {
+	private update = () => {
```
mjbvz, line 132: I don't think we use this pattern very often. Could you use `() => this.update()` like other locations or use a real call to bind?

PR #265050, opened 2025-09-03
Add simulation context support across environment and context key services

bhavyaus:
I want to enable the manage_todo_list tool for s-tests. 
cc: @karthiknadig @rwoll


mjbvz (approved):

bpasero:
@bhavyaus it is not a good idea to use such a generic name like `SIMULATION`, at the very minimum you need to prefix with `VSCODE_` as we do with other environment variables.

PR #265066, opened 2025-09-04
Removing open chat in new window action

osortega:


rebornix (approved):

bpasero:
Care to explain? Is this dropping support for opening chats in new windows or something else?

PR #265082, opened 2025-09-04
Update Objective-C syntax hightlighting to support missing opaque types

SoCuul:
Adds syntax highlighting for the remaining opaque Objective-C types/structs: [Method](https://developer.apple.com/documentation/objectivec/method?language=objc), [Ivar](https://developer.apple.com/documentation/objectivec/ivar?language=objc), and [Protocol](https://developer.apple.com/documentation/objectivec/protocol?language=objc).

These are in addition to the currently highlighted type: [Class](https://developer.apple.com/documentation/objectivec/class?language=objc).

### DIFF
--- a/extensions/objective-c/syntaxes/objective-c.tmLanguage.json
+++ b/extensions/objective-c/syntaxes/objective-c.tmLanguage.json
@@ -154,7 +154,7 @@
 			"name": "storage.type.id.objc"
 		},
 		"anonymous_pattern_12": {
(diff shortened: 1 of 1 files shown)


RedCMD:
should report to upstream grammar repo https://github.com/jeff-hykin/better-objc-syntax

alexr00:
@RedCMD is correct, we pull our objective-c syntax highlighting grammar from https://github.com/jeff-hykin/better-objc-syntax. If your PR is merged there then we'll pull the changes into VS Code.

PR #265098, opened 2025-09-04
Fix stdin pipe file position issue

yuyalun-allen:
Fix #260309 that file position become inconsistent when using stdin redirection.

Fix this by replacing stdin redirection symbol directly with tmp pipe file, instead of filtering first, adding then. (I'm new to open source contribution, so I hope I could fix the "good first issue" for the community!)


yuyalun-allen:

PR #265131, opened 2025-09-04
Fixes "Caret jumps to column 0 when clicking between first two characters"

ttttotem:
Fixes issue https://github.com/microsoft/vscode/issues/263098 that was introduced in https://github.com/microsoft/vscode/pull/235174

Problem:
xLeftBoundary is not correctly at where the left most line of the editor content, it is blocking clicking on the first column

Solution:
Remove horizontalScrollPadding from the cursor leaving editor content on the left side check. horizontalScrollPadding should not have been added to the left side check because layoutInfo.contentLeft correctly matches the left most point of the editor content


alexdima (approved):

alexdima:
Thank you!

Tyriar (approved):

PR #265137, opened 2025-09-04
Window Border: There is no straight forward way to use Windows window border color (fix #264566)

### LINKED ISSUE #264566: Window Border: There is no straight forward way to use Windows window border color
Testing #263838

When I set `"window.border": "default"` and have windows "Show accent colour..." enabled, I get the window border color set by the theme. Personally, I would have expected to have the same border color as set by Windows by default as I think the default user would expect all Applications to use the color set in the Windows settings. There also does not seem to be an easy way to to use the same border color as set by Windows.

bpasero:

### DIFF
--- a/src/vs/workbench/electron-browser/desktop.contribution.ts
+++ b/src/vs/workbench/electron-browser/desktop.contribution.ts
@@ -316,10 +316,10 @@ import { registerWorkbenchContribution2, WorkbenchPhase } from '../common/contri
 				'markdownDescription': (() => {
 					let windowBorderDescription = localize('window.border.prefix', "Controls the border color of the window:");
 					windowBorderDescription += '\n- ' + [
-						localize('window.border.default', "`default`: respect color theme settings, fallback to Windows settings"),
-						localize('window.border.system', "`system`: respect Windows settings only"),
-						localize('window.border.off', "`off`: disable border colors"),
-						localize('window.border.color', "`<color>`: specific color in Hex, RGB, RGBA, HSL, HSLA format"),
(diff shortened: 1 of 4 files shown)


src/vs/workbench/electron-browser/desktop.contribution.ts
```diff
@@ -313,7 +313,18 @@ import { registerWorkbenchContribution2, WorkbenchPhase } from '../common/contri
 			'window.border': {
 				'type': 'string',
 				'default': 'default',
-				'markdownDescription': localize('window.border', "Controls the border color of the window. Set to `default` to respect Windows or color theme settings, `off` to disable, or to a specific color in Hex, RGB, RGBA, HSL, HSLA format. Use {0} to set different colors for active and inactive windows. This setting is ignored when {1} is set to {2}.", '`#workbench.colorCustomizations#`', '`#window.titleBarStyle#`', '`native`'),
+				'markdownDescription': (() => {
+					let windowBorderDescription = localize('window.border.prefix', "Controls the border color of the window:");
+					windowBorderDescription += '\n- ' + [
+						localize('window.border.default', "`default`: respect color theme settings, fallback to Windows settings"),
+						localize('window.border.system', "`system`: respect Windows settings only"),
+						localize('window.border.off', "`off`: disable border colors"),
```
Tyriar: Best not to localize the first part of the first 3 items. Also normally these live in enum descriptions?

bpasero: Good catch!

bpasero: The setting is open ended so it cannot be an enum...

Tyriar (approved):

PR #265147, opened 2025-09-04
Hide terminal inline chat if AI is disabled

### LINKED ISSUE #264612: cmd+i still works in terminal when `chat.disableAIFeatures: true`
Testing #263837

Image

Tyriar:
Fixes #264612


bpasero (changes requested):
Rather than each one depending on the setting, I would recommend to use `IChatEntitlementService` where you have a property `sentiment.hidden`.

bpasero (approved):

PR #265239, opened 2025-09-04
no progressive rendering for existing parts in incomplete chat response

rebornix:


src/vs/workbench/contrib/chat/browser/chatListRenderer.ts
```diff
@@ -627,6 +627,18 @@ export class ChatListItemRenderer extends Disposable implements ITreeRenderer<Ch
 		if (isResponseVM(element) && index === this.delegate.getListLength() - 1 && (!element.isComplete || element.renderData)) {
 			this.traceLayout('renderElement', `start progressive render, index=${index}`);
 
+			/**
+			 * There is no rendered parts yet, meaning this is the first time rendering this response
+			 * However, there is already parts in the response, we should render the existing parts without animation
+			 */
+			if (!templateData.renderedParts && !element.isComplete && element.response.value.length > 0) {
```
roblourens: Here's what I suggest

- You want to change getDataForProgressiveRender, that's where we decide how much to render
- Look at ChatViewModel.contentUpdateTimings.lastWordCount. I think that's what you're counting here
- Basically, if there is no content rendered yet, we want to render all the words (which is lastWordCount)
- OR if that doesn't work, `getDataForProgressiveRender` can return some special value which means "render everything" and then we render everything

rebornix: > OR if that doesn't work, getDataForProgressiveRender can return some special value which means "render everything" and then we render everything

great suggestion. `ChatViewModel.contentUpdateTimings.lastWordCount` is `0` initially, but that's a good indicator for checking if we should render all parts in the chat response.

roblourens (approved):
Yes LGTM, this is how it should have been written originally

PR #265265, opened 2025-09-05
Remove chat sessions view on no-ai

osortega:


mjbvz (approved):

bpasero:
@osortega please use `IChatEntitlementService` and the `sentiment` property to get service access to this information instead of having to go via context keys.

PR #265303, opened 2025-09-05
Add conversation id to mcp meta

leonard520:
Resolve #265302


src/vs/workbench/contrib/mcp/common/mcpServer.ts
```diff
@@ -895,12 +909,24 @@ export class McpTool implements IMcpTool {
 				}
 			});
 
-			return h.callTool({ name, arguments: params, _meta: { progressToken } }, token)
+			// Prepare meta data to include conversation id and progress token
+			const meta: Record<string, unknown> = {
+				'vscode.client': 'vscode',
+				'vscode.progressToken': progressToken
+			};
```
connor4312: The client and progress token are already sent in the protocol, we should not send them separately here

src/vs/workbench/contrib/mcp/common/mcpServer.ts
```diff
@@ -862,7 +862,21 @@ export class McpTool implements IMcpTool {
 		const name = this._definition.serverToolName ?? this._definition.name;
 		if (context) { this._server.runningToolCalls.add(context); }
 		try {
-			return await McpServer.callOn(this._server, h => h.callTool({ name, arguments: params }, token), token);
+			const meta: Record<string, unknown> = {
+				'vscode.client': 'vscode'
+			};
```
connor4312: Similarly, we no need to send the client here

connor4312 (changes requested):

src/vs/workbench/contrib/mcp/common/mcpServer.ts
leonard520: Sure. Updated.

src/vs/workbench/contrib/mcp/common/mcpServer.ts
leonard520: Updated

connor4312 (approved):

Tyriar (approved):

PR #265309, opened 2025-09-05
update exp configs on refetch

sandy081:


hediet (approved):

src/vs/workbench/services/configuration/browser/configurationService.ts
```diff
@@ -1351,6 +1351,7 @@ class ConfigurationDefaultOverridesContribution extends Disposable implements IW
 		super();
 
 		this.updateDefaults().then(() => {
+			this._register(workbenchAssignmentService.onDidRefetchAssignments(() => this.processExperimentalSettings(this.autoExperimentalSettings, true)));
```
hediet, line 1354: Just to note, this could leak in tests, if `updateDefaults` resolves after the service already got disposed!

PR #265319, opened 2025-09-05
Fire event on assignment context updates

benibenj:
```Copilot Generated Description:``` Introduce an event that fires when the assignment context updates, allowing for better handling of assignment refetching.

fixes https://github.com/microsoft/vscode/issues/265315

### DIFF
--- a/src/vs/workbench/services/assignment/common/assignmentService.ts
+++ b/src/vs/workbench/services/assignment/common/assignmentService.ts
@@ -51,7 +51,10 @@ class MementoKeyValueStorage implements IKeyValueStorage {
 	}
 }
 
-class WorkbenchAssignmentServiceTelemetry implements IExperimentationTelemetry {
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Review by hediet — APPROVED** [review]

File: src/vs/workbench/services/assignment/common/assignmentService.ts [hunk 1]
```diff
@@ -51,7 +51,10 @@ class MementoKeyValueStorage implements IKeyValueStorage {
 	}
 }
 
-class WorkbenchAssignmentServiceTelemetry implements IExperimentationTelemetry {
+class WorkbenchAssignmentServiceTelemetry extends Disposable implements IExperimentationTelemetry {
+
+	private readonly _onDidUpdateAssignmentContext = this._register(new Emitter<void>());
```
connor4312, line 56: Btw we also have a `Relay` class you can use in cases like this

benibenj: Thanks for the hint, I've forgot about that one.

PR #265326, opened 2025-09-05
fix: elicitation email validator

garciasdos:
This pull request fixes a bug in the email validation logic within the `McpElicitationService`. The previous implementation incorrectly marked emails containing an '@' as invalid; this change corrects the logic to properly validate email addresses.

- Bug fix: Corrected the email validation in the `_validateStringFormat` method of `McpElicitationService` to ensure that values containing an '@' are now considered valid emails.

fixes #265325


connor4312 (approved):
whoops 🤦

rzhao271 (approved):

PR #265390, opened 2025-09-05
fix button growing vertically in confirmation widget

### LINKED ISSUE #265382: Button should not grow vertically when there's horizontal space
From @digitarald 

cc @connor4312 and @justschen for the chat elicitation request part

Image

meganrogge:
fix #265382

Before:

image

After:

Screenshot 2025-09-05 at 2 18 31 PM

### DIFF
--- a/src/vs/workbench/contrib/chat/browser/chatContentParts/media/chatConfirmationWidget.css
+++ b/src/vs/workbench/contrib/chat/browser/chatContentParts/media/chatConfirmationWidget.css
@@ -12,6 +12,10 @@
 	align-items: center;
 }
 
+.chat-confirmation-widget .monaco-text-button {
+	height: 28px;
+}
+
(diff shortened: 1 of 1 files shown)


src/vs/workbench/contrib/chat/browser/chatContentParts/media/chatConfirmationWidget.css
```diff
@@ -12,6 +12,10 @@
 	align-items: center;
 }
 
+.chat-confirmation-widget .monaco-text-button {
+	height: 28px;
```
roblourens: I recommend not hard-coding heights like this, especially since chat font props are now customizable, what if I set a font-size larger than this?

Instead, figure out what is constraining the width and causing it to wrap

Tyriar (approved):

PR #265394, opened 2025-09-05
Handle registering and deregsitering chat sessions view

### LINKED ISSUE #264615: new "chat sessions" panel visible in ai disabled mode
Testing #263837

with chat.disableAIFeatures: true, I still have the "chat sessions" icon in my sidebar. It doesn't seem to load anything (which may also be because I don't have copilot installed)

Image

osortega:
Fixes: https://github.com/microsoft/vscode/issues/264615


src/vs/workbench/contrib/chat/browser/chatSessions.ts
```diff
@@ -304,13 +303,11 @@ export class ChatSessionsView extends Disposable implements IWorkbenchContributi
 
 	private updateViewContainerRegistration(): void {
 		const location = this.configurationService.getValue<string>(ChatConfiguration.AgentSessionsViewLocation);
-
-		if (location === 'view' && !this.isViewContainerRegistered) {
+		const isAIDisabled = this.configurationService.getValue<boolean>(ChatConfiguration.DisableAIFeatures);
```
Tyriar: I think you're meant to use sentiment for this? @bpasero?

lszomoru (dismissed):

lszomoru:
@osortega, PR build is 🔴

src/vs/workbench/contrib/chat/browser/chatSessions.ts
osortega: @Tyriar you are right, I just did the change to use sentiment instead

src/vs/workbench/contrib/chat/browser/chatSessions.ts
```diff
@@ -304,13 +308,11 @@ export class ChatSessionsView extends Disposable implements IWorkbenchContributi
 
 	private updateViewContainerRegistration(): void {
 		const location = this.configurationService.getValue<string>(ChatConfiguration.AgentSessionsViewLocation);
-
-		if (location === 'view' && !this.isViewContainerRegistered) {
+		const sentiment = this.chatEntitlementService.sentiment;
+		if (sentiment.disabled || (location !== 'view' && this.isViewContainerRegistered)) {
```
bpasero: This is maybe a bit misleading but you need to check on both `sentiment.hidden` and `sentiment.disabled` to cover all cases.

bpasero (changes requested):

src/vs/workbench/contrib/chat/browser/chatSessions.ts
osortega: @bpasero thanks for the tip, added it!

bpasero (approved):

PR #265408, opened 2025-09-05
Adding a way to identify chat models originated from contributions

osortega:
This PR adds a way to identify a chat model that originated from a contributed chat session.
It also filters out form history based on that new property


src/vs/workbench/contrib/chat/common/chatModel.ts
```diff
@@ -1422,9 +1422,14 @@ export class ChatModel extends Disposable implements IChatModel {
...
 	}
 
+	get inputType(): string {
+		return this._inputType;
+	}
+
 	constructor(
 		private readonly initialData: ISerializableChatData | IExportableChatData | undefined,
 		private readonly _initialLocation: ChatAgentLocation,
+		private readonly _inputType: string,
```
Tyriar, line 1432: Should we move to an options object instead? You can then have options and it'll be clearer what's being called

Tyriar (changes requested):

src/vs/workbench/contrib/chat/common/chatModel.ts
osortega: @Tyriar did that in a separate PR since it was easier to cherry pick the commit from the candidate: https://github.com/microsoft/vscode/pull/265923

PR #265507, opened 2025-09-06
feat: show binary size for non-binary files, too

### LINKED ISSUE #265506: Show File Size Infomation for Non-Binary files
Type: <b>Feature Request</b>

VSCode can show the file size of an image, for example, right in the editor, in the status bar, which is really convenient.

Users don't have to open the file in their OS's explorer and check the file size that way, saving them time.

It'd be nice if we could show file sizes for other file types, like JSON, too, so we can save users time and prevent them from context switching to another window to manually check file sizes themselves.

VS Code version: Code 1.103.2 (6f17636121051a53c88d3e605c491d22af2ba755, 2025-08-20T16:45:34.255Z)
OS version: Linux x64 6.14.0-29-generic
Modes:

hirehamir:
Closes #265506

### DIFF
--- a/extensions/media-preview/package.json
+++ b/extensions/media-preview/package.json
@@ -19,7 +19,9 @@
   "categories": [
     "Other"
   ],
-  "activationEvents": [],
+  "activationEvents": [
+    "onStartupFinished"
+  ],
   "capabilities": {
     "virtualWorkspaces": true,
     "untrustedWorkspaces": {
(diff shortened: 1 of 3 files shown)


eabase:
@hamirmahal 
Sorry, your video is not watchable...
Can you link somewhere else?

hamirmahal:
@lramos15 are there any changes I should make here?

eabase:
@bpasero 

This is really `s**t` behavior. The work was already done by @hamirmahal. 
So how can you possibly reach the conclusion that it is *`out-of-scoop`* ?

PR #265516, opened 2025-09-06
Don't create chat sessions as "editor" location

roblourens:
This means inline chat. I got into a state where a chat session was loaded in the panel instead of a chat editor and the location caused issues

### DIFF
--- a/src/vs/workbench/contrib/chat/browser/chatEditorInput.ts
+++ b/src/vs/workbench/contrib/chat/browser/chatEditorInput.ts
@@ -173,12 +173,12 @@ export class ChatEditorInput extends EditorInput implements IEditorCloseHandler 
 
 	override async resolve(): Promise<ChatEditorModel | null> {
 		if (this.resource.scheme === Schemas.vscodeChatSession) {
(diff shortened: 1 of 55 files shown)


Tyriar (approved):

src/vs/workbench/contrib/chat/browser/chatEditorInput.ts
```diff
@@ -173,12 +173,12 @@ export class ChatEditorInput extends EditorInput implements IEditorCloseHandler
 
 	override async resolve(): Promise<ChatEditorModel | null> {
 		if (this.resource.scheme === Schemas.vscodeChatSession) {
-			this.model = await this.chatService.loadSessionForResource(this.resource, ChatAgentLocation.Editor, CancellationToken.None);
+			this.model = await this.chatService.loadSessionForResource(this.resource, ChatAgentLocation.Chat, CancellationToken.None);
```
roblourens, line 176: fyi @mjbvz

mjbvz: Thanks for the enum rename! Never realized it was inline chat :)

PR #265623, opened 2025-09-08
chat - introduce support for `machineId`

bpasero:


bpasero:
Continues in https://github.com/microsoft/vscode/pull/265806

Issue #262778, opened 2025-08-21
Local chat history in Chat Sessions view appears to include coding agent sessions


bamurtaugh:
Version: 1.104.0-insider (user setup)
Commit: 6d631d487357dd5fcc772cb87e0bd155fc6febf9
Date: 2025-08-21T16:00:52.194Z
Electron: 37.2.3
ElectronBuildId: 12035395
Chromium: 138.0.7204.100
Node.js: 22.17.0
V8: 13.8.500258-electron.0
OS: Windows_NT arm64 10.0.26100

Steps to Reproduce:

1. Open Chat Sessions view
2. Open a variety of coding agent sessions, then close them all
3. `Show history...` in local chat sessions
4. ❔ I see some sessions titled `---`, which I know is from https://github.com/microsoft/vscode/issues/259207. When I click into them, they are indeed coding agent sessions, which I didn't think would appear in this section, as they're not local chat sessions?
<img width="494" height="460" alt="Image" src="https://github.com/user-attachments/assets/e2fb3871-2444-4048-a581-850a96d42dc5" />

osortega:
I believe this is fixed now, marking it for verification:
1. Create multiple chats
2. Check History items
3. Verify all have proper names

bamurtaugh:
I no longer see `---`, which is great!

I am still seeing remote coding agent sessions show up in local chat session history - that feels unexpected to me?

Example scenario:
- From the `GitHub Copilot Coding Agent` tab of the Chat Sessions view, open a bunch of editors with remote coding agent sessions
- Close the coding agent editors
- From the `Local Chat Sessions` tab of the Chat Sessions view, expand the `History` tab
- ❔ I see coding agent editors in my local chat history. Example of a coding agent session that appears in my local chat sessions below:

<img width="2484" height="1587" alt="Image" src="https://github.com/user-attachments/assets/412c82ea-8d8a-4519-ae09-e1e018c47655" />

mjbvz:
Confirmed that I still see this too. My repo:

1. Open a GitHub coding agent session from the view
2. 🐛 History now contains it (in this case I opened the session about `fix null reference`)

<img width="675" height="592" alt="Image" src="https://github.com/user-attachments/assets/1be09394-24ab-4828-9044-ff2282c595ac" />

bamurtaugh:
I still see previous coding agent sessions in my local history, though new ones aren't added (so if I open another coding agent session from the chat sessions view, it doesn't get added to local history).

[...]

PR #265702, opened 2025-09-08
Input type attribute to chat model

osortega:
Fixes: https://github.com/microsoft/vscode/issues/262778

### DIFF
--- a/src/vs/workbench/contrib/chat/browser/chatEditorInput.ts
+++ b/src/vs/workbench/contrib/chat/browser/chatEditorInput.ts
@@ -183,7 +183,7 @@ export class ChatEditorInput extends EditorInput implements IEditorCloseHandler 
 		} else if (!this.options.target) {
 			this.model = this.chatService.startSession(ChatAgentLocation.Panel, CancellationToken.None, undefined, inputType);
 		} else if ('data' in this.options.target) {
-			this.model = this.chatService.loadSessionFromContent(this.options.target.data, inputType);
+			this.model = this.chatService.loadSessionFromContent(this.options.target.data);
 		}
 
 		if (!this.model || this.isDisposed()) {
--- a/src/vs/workbench/contrib/chat/browser/chatEditorInput.ts
+++ b/src/vs/workbench/contrib/chat/browser/chatEditorInput.ts
@@ -172,16 +172,18 @@ export class ChatEditorInput extends EditorInput implements IEditorCloseHandler 
 	}
 
 	override async resolve(): Promise<ChatEditorModel | null> {
-		const isContributed = !!this.resource.authority || this.resource.query.startsWith('chatSessionType');
+		const searchParams = new URLSearchParams(this.resource.query);
+		const chatSessionType = searchParams.get('chatSessionType');
+		const inputType = chatSessionType ?? this.resource.authority;
 		if (this.resource.scheme === Schemas.vscodeChatSession) {
-			this.model = await this.chatService.loadSessionForResource(this.resource, ChatAgentLocation.Editor, CancellationToken.None, isContributed);
+			this.model = await this.chatService.loadSessionForResource(this.resource, ChatAgentLocation.Editor, CancellationToken.None, inputType);
 		} else if (typeof this.sessionId === 'string') {
-			this.model = await this.chatService.getOrRestoreSession(this.sessionId, isContributed)
-				?? this.chatService.startSession(ChatAgentLocation.Panel, CancellationToken.None, undefined, isContributed);
+			this.model = await this.chatService.getOrRestoreSession(this.sessionId, inputType)
+				?? this.chatService.startSession(ChatAgentLocation.Panel, CancellationToken.None, undefined, inputType);
 		} else if (!this.options.target) {
-			this.model = this.chatService.startSession(ChatAgentLocation.Panel, CancellationToken.None, undefined, isContributed);
+			this.model = this.chatService.startSession(ChatAgentLocation.Panel, CancellationToken.None, undefined, inputType);
 		} else if ('data' in this.options.target) {
-			this.model = this.chatService.loadSessionFromContent(this.options.target.data, isContributed);
+			this.model = this.chatService.loadSessionFromContent(this.options.target.data, inputType);
 		}
 
 		if (!this.model || this.isDisposed()) {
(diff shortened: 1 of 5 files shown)


src/vs/workbench/contrib/chat/common/chatModel.ts
```diff
@@ -1146,15 +1146,20 @@ export interface ISerializableChatData3 extends Omit<ISerializableChatData2, 've
 	customTitle: string | undefined;
 }
 
+export interface ISerializableChatData4 extends Omit<ISerializableChatData3, 'version'> {
+	version: 4;
+	inputType?: string;
```
roblourens, line 1148: You don't have to do this, since `inputType` is optional

src/vs/workbench/contrib/chat/common/chatModel.ts
```diff
@@ -1417,9 +1432,14 @@ export class ChatModel extends Disposable implements IChatModel {
...
 	}
 
+	get inputType(): string {
+		return this._inputType;
+	}
+
 	constructor(
 		private readonly initialData: ISerializableChatData | IExportableChatData | undefined,
 		private readonly _initialLocation: ChatAgentLocation,
+		private readonly _inputType: string,
```
roblourens: Can this go away if `ISerializableChatData` has inputType?

src/vs/workbench/contrib/chat/browser/chatEditorInput.ts
```diff
@@ -172,15 +172,18 @@ export class ChatEditorInput extends EditorInput implements IEditorCloseHandler
 	}
 
 	override async resolve(): Promise<ChatEditorModel | null> {
+		const searchParams = new URLSearchParams(this.resource.query);
+		const chatSessionType = searchParams.get('chatSessionType');
```
roblourens, line 176: Ideally we would have some code in one place that knows how to handle these special URIs rather than parsing a URI in an adhoc way

src/vs/workbench/contrib/chat/browser/chatEditorInput.ts
```diff
@@ -172,15 +172,18 @@ export class ChatEditorInput extends EditorInput implements IEditorCloseHandler
...
+		const inputType = chatSessionType ?? this.resource.authority;
```
roblourens, line 177: What are the values for inputType?

src/vs/workbench/contrib/chat/browser/chatEditorInput.ts
osortega: I believe that is something @rebornix and @joshspicer will be working on outside of a candidate

src/vs/workbench/contrib/chat/browser/chatEditorInput.ts
osortega: Right now we only have `claude-code` and `copilot-swe-agent`

src/vs/workbench/contrib/chat/common/chatModel.ts
osortega: Perfect, reverted to version 3

src/vs/workbench/contrib/chat/common/chatModel.ts
osortega: @roblourens I don't think so since the initialData can still be undefined in some cases. We could merge { ...initialData, inputType } but might end up being pretty ugly since `IExportableChatData` doesn't have `inputType`

roblourens: Seems like it should be `string | undefined`? or no? Should it be on `IExportableChatData`?

src/vs/workbench/contrib/chat/common/chatServiceImpl.ts
```diff
@@ -180,7 +180,7 @@ export class ChatService extends Disposable implements IChatService {
 	private saveState(): void {
 		const liveChats = Array.from(this._sessionModels.values())
 			.filter(session =>
-				this.shouldSaveToHistory(session.sessionId) && (session.initialLocation === ChatAgentLocation.Panel || session.initialLocation === ChatAgentLocation.Editor));
+				!session.inputType && (session.initialLocation === ChatAgentLocation.Panel || session.initialLocation === ChatAgentLocation.Editor));
```
roblourens, line 183: Is it right that `this.shouldSaveToHistory` isn't called here anymore? Is that function still needed?

src/vs/workbench/contrib/chat/test/common/chatModel.test.ts
```diff
@@ -52,8 +52,8 @@ suite('ChatModel', () => {
 	});
 
 	test('adoptRequest', async function () {
-		const model1 = testDisposables.add(instantiationService.createInstance(ChatModel, undefined, ChatAgentLocation.Editor));
-		const model2 = testDisposables.add(instantiationService.createInstance(ChatModel, undefined, ChatAgentLocation.Panel));
+		const model1 = testDisposables.add(instantiationService.createInstance(ChatModel, undefined, { initialLocation: ChatAgentLocation.Editor, inputType: '' }));
```
roblourens: inputType can be optional so I think we can clean up all these empty strings

src/vs/workbench/contrib/chat/common/chatServiceImpl.ts
```diff
@@ -609,11 +609,11 @@ export class ChatService extends Disposable implements IChatService {
 		return undefined;
 	}
 
-	loadSessionFromContent(data: IExportableChatData | ISerializableChatData): IChatModel | undefined {
-		return this._startSession(data, data.initialLocation ?? ChatAgentLocation.Panel, true, CancellationToken.None);
+	loadSessionFromContent(data: IExportableChatData | ISerializableChatData, inputType: string = ''): IChatModel | undefined {
```
roblourens: And this default we should be able to remove

src/vs/workbench/contrib/chat/test/common/chatModel.test.ts
osortega: Removed it

src/vs/workbench/contrib/chat/common/chatServiceImpl.ts
osortega: @roblourens good catch, there were a lot of more places where we needed to remove this

roblourens (approved):

src/vs/workbench/contrib/chat/browser/chatEditorInput.ts
joshspicer: Thats what `getChatSessionType` in `chatSessions/common.ts` was for 

https://github.com/microsoft/vscode/blob/5366847dcb9304bd5e695719192ea9cfc8764a75/src/vs/workbench/contrib/chat/browser/chatSessions/common.ts#L41-L62

PR #265707, opened 2025-09-08
fix: 'Create new branch...' label without ellipsis

gy-mate:
# Changes

Fixes #262054.

### Preparation

1. Open a folder tracked with Git
2. Click on the _Source Control Checkout_ button in the bottom Status Bar
3. Enter a new branch name

### Previously

There was an option named _Create new branch..._

### Now

That option is now called _Create new branch_. The ellipsis was removed.


lszomoru:
@gy-mate, thank you very much for your contribution. Most people hover create a new branch by pressing the action and inserting the name after since this flow takes into account a configured branch prefix as well as branch name generation. In light of that the action label containing `...` is correct.

PR #265712, opened 2025-09-08
provide progress messages to screen reader users, enable opting out

### LINKED ISSUE #265403: Progress messages are not indicated to screen reader users
From @rperez030: Codex provides thorough announcements for every step as the agent works.

Currently, we don’t announce progress in the same way. When there’s a spinner or checkmark indicating progress, we should add an opt-out setting to have these updates announced. It will be very verbose, but valuable so users can stay informed and interrupt the workflow if needed.

Image

I think we attempt to do this here:
https://github.com/microsoft/vscode/blob/208f629b439e05707c460a03f2f1d4256780a7cd/src/vs/workbench/contrib/chat/browser/chatContentParts/chatProgressContentPart.ts#L49-L53

cc @roblourens

meganrogge:
fix #265403

Visually, we indicate tool progress with messages. Before, screen reader users would only hear "Working" and the chat progress sound. 

Now, they'll hear the tool progress messages such as: `searched for x in ..test.ts, no results` so that they know if it's on the right track. Since this can be verbose, I've added an opt-out setting and included this in the accessibility help dialog.

### DIFF
--- a/src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolConfirmationSubPart.ts
+++ b/src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolConfirmationSubPart.ts
@@ -41,6 +41,7 @@ import { ChatCustomConfirmationWidget, IChatConfirmationButton } from '../chatCo
 import { IChatContentPartRenderContext } from '../chatContentParts.js';
 import { ChatMarkdownContentPart, EditorPool } from '../chatMarkdownContentPart.js';
 import { BaseChatToolInvocationSubPart } from './chatToolInvocationSubPart.js';
+import { alert } from '../../../../../../base/browser/ui/aria/aria.js';
 
 export const enum TerminalToolConfirmationStorageKeys {
 	TerminalAutoApproveWarningAccepted = 'chat.tools.terminal.autoApprove.warningAccepted'
@@ -134,6 +135,9 @@ export class ChatTerminalToolConfirmationSubPart extends BaseChatToolInvocationS
 				ariaLabel: typeof title === 'string' ? title : title.value
 			}
 		};
+		if (this.configurationService.getValue('accessibility.verboseChatProgressUpdates')) {
+			alert(codeBlockRenderOptions.editorOptions?.ariaLabel);
+		}
 		const languageId = this.languageService.getLanguageIdByLanguageName(terminalData.language ?? 'sh') ?? 'shellscript';
 		const model = this._register(this.modelService.createModel(
 			terminalData.commandLine.toolEdited ?? terminalData.commandLine.original,
--- a/src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolConfirmationSubPart.ts
+++ b/src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolConfirmationSubPart.ts
@@ -41,7 +41,6 @@ import { ChatCustomConfirmationWidget, IChatConfirmationButton } from '../chatCo
 import { IChatContentPartRenderContext } from '../chatContentParts.js';
 import { ChatMarkdownContentPart, EditorPool } from '../chatMarkdownContentPart.js';
 import { BaseChatToolInvocationSubPart } from './chatToolInvocationSubPart.js';
-import { alert } from '../../../../../../base/browser/ui/aria/aria.js';
 
 export const enum TerminalToolConfirmationStorageKeys {
 	TerminalAutoApproveWarningAccepted = 'chat.tools.terminal.autoApprove.warningAccepted'
@@ -135,9 +134,7 @@ export class ChatTerminalToolConfirmationSubPart extends BaseChatToolInvocationS
 				ariaLabel: typeof title === 'string' ? title : title.value
 			}
 		};
-		if (this.configurationService.getValue('accessibility.verboseChatProgressUpdates')) {
-			alert(codeBlockRenderOptions.editorOptions?.ariaLabel);
-		}
+
 		const languageId = this.languageService.getLanguageIdByLanguageName(terminalData.language ?? 'sh') ?? 'shellscript';
 		const model = this._register(this.modelService.createModel(
 			terminalData.commandLine.toolEdited ?? terminalData.commandLine.original,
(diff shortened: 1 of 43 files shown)


src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolConfirmationSubPart.ts
```diff
@@ -134,6 +135,9 @@ export class ChatTerminalToolConfirmationSubPart extends BaseChatToolInvocationS
 				ariaLabel: typeof title === 'string' ? title : title.value
 			}
 		};
+		if (this.configurationService.getValue('accessibility.verboseChatProgressUpdates')) {
+			alert(codeBlockRenderOptions.editorOptions?.ariaLabel);
```
Tyriar: `codeBlockRenderOptions.editorOptions` could be undefined?

src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatToolConfirmationSubPart.ts
```diff
@@ -129,7 +132,9 @@ export class ToolConfirmationSubPart extends BaseChatToolInvocationSubPart {
 					ariaLabel: typeof title === 'string' ? title : title.value
 				},
 			};
-
+			if (this.configurationService.getValue('accessibility.verboseChatProgressUpdates')) {
+				alert(codeBlockRenderOptions.editorOptions?.ariaLabel);
```
Tyriar: `codeBlockRenderOptions.editorOptions` could be undefined?

Tyriar (dismissed):
It's not clear how often this will happen. Also confirmations can get deserialized after reloading a window and that might alert when the window is starting up?

src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolConfirmationSubPart.ts
meganrogge: it's ok if it's undefined, i handle that in `alert`

src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatToolConfirmationSubPart.ts
meganrogge: same as above!

src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolConfirmationSubPart.ts
```diff
@@ -134,6 +134,7 @@ export class ChatTerminalToolConfirmationSubPart extends BaseChatToolInvocationS
 				ariaLabel: typeof title === 'string' ? title : title.value
 			}
 		};
+
```
meganrogge:
```suggestion
```

meganrogge:
```suggestion
```

src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatToolInvocationPart.ts
```diff
@@ -54,13 +57,19 @@ export class ChatToolInvocationPart extends Disposable implements IChatContentPa
...
 		super();
 
 		this.domNode = dom.$('.chat-tool-invocation-part');
 		if (toolInvocation.presentation === 'hidden') {
 			return;
 		}
+		this._register(this.lifecycleService.onBeforeShutdown(() => {
+			sessionStorage.setItem('vscode.chat.lastReload', String(Date.now()));
+		}));
+		this._alertForScreenReaders();
```
Tyriar: I think the right place to do this is where `IToolImpl.invoke` is called. @roblourens suggested tool service in standup

Tyriar (dismissed):

src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatToolInvocationPart.ts
roblourens: If you want to do this as part of rendering, it's a solvable problem, we can tell the difference between new content streaming in from a live response and old content just being rendered. But basically you just need to decide, do you want to alert when something happens in the system, or when there is certain content appearing on screen? Because these are different things.

meganrogge: I think we only want to do this when the tool is first invoked.

meganrogge: But we also want to indicate when it's finished. So, for the reading tool, it would say "reading file_name", "read file_name"

meganrogge: there are a bunch of implementations of `ItoolImpl.invoke`. 

Screenshot 2025-09-09 at 12 28 03 PM

TylerLeonhardt (approved):

src/vs/workbench/contrib/chat/browser/languageModelToolsService.ts
```diff
@@ -320,6 +321,9 @@ export class LanguageModelToolsService extends Disposable implements ILanguageMo
 				token = source.token;
 
 				const prepared = await this.prepareToolInvocation(tool, dto, token);
+
+				this.informScreenReader(prepared?.invocationMessage);
```
roblourens, line 325: Both of these are optional, `pastTenseMessage` is proposed and only used by some tools, maybe you want a fallback?

meganrogge: If we don't show it to users, we shouldn't be announcing it, so I think this is fine.

roblourens: But we would still show it to users

meganrogge: huh? what is shown to users when those are not set?

roblourens: Tool results get visualized in a few different ways. If you don't set `invocationMessage` we show a default message. If you don't use `pastTenseMessage` then it won't change the message when it's done but you'll see the progress finish

roblourens: Hey something else just ocurred to me, we should be thinking about this in terms of external chat sessions (padawan, claude code, etc) as well. In that case, they are showing the same tool call progress updates but they are running their own tools which are not passing through our internal tools service.

PR #265755, opened 2025-09-08
debug: fix incorrect thread state if "continued" arrives too soon after a "stopped"

### LINKED ISSUE #238591: Debugger shows incorrect thread state if "continued" arrives too soon after a "stopped" (or "started"?)
I had a bug raised against Dart where the debugger shows threads as "paused on entry" when they should not be. The DAP traffic looks good:

```
// Start
[5:17:21 PM] [DAP] [Info] <== {"seq":21,"type":"event","body":{"reason":"started","threadId":3},"event":"thread"}
// Pause
[5:17:21 PM] [DAP] [Info] <== {"seq":43,"type":"event","body":{"allThreadsStopped":false,"reason":"entry","threadId":3},"event":"stopped"}
// Resume
[5:17:21 PM] [DAP] [Info] <== {"seq":78,"type":"event","body":{"allThreadsContinued":false,"threadId":3},"event":"continued"}
```

Yet the UI shows the thread as paused. I was able to reproduce this in MockDebug. It seems to only occur when the events occur close together (if I add delays, it doesn't seem to happen):

```ts
// Start additional threads, pause on entry, continue.
for (var i = 2; i < 5; i++) {
	this._threads.push(new Thread(i, `Thread ${i}`));
	this.sendEvent(new ThreadEvent('started', i));
	this.sendEvent(new OutputEvent(`Thread ${i} started\n`));
	this.sendEvent(new StoppedEvent('entry', i));
	this.sendEvent(new OutputEvent(`Thread ${i} stopped on entry\n`));
	// await this.delay(1); // Uncomment this to fix it
	this.sendEvent(new ContinuedEvent(i, false));
	this.sendEvent(new OutputEvent(`Thread ${i} continued\n`));
}
```

Image

Full repro is here:

https://github.com/DanTup/vscode-mock-debug/tree/repro-dart-code-5397 / https://github.com/DanTup/vscode-mock-debug/commit/04f4764e853cbee2b09ca28c110c423fb4b99fd0

This felt a bit like https://github.com/microsoft/debug-adapter-protocol/issues/513 to me, but that actually turned out to be caused by `allThreadsStopped` defaulting to `true`. I don't believe that's the case here because in the real project where this happens (Dart debugger), we set both `allThreadsStopped` and `allThreadsContinued` to `false` (though I haven't set `allThreadsContinued` in the MockDebug example because it doesn't seem to be available in the constructor... I don't believe that's the issue though as `allThreadsContinued` seems to default to false and the delay shouldn't fix the issue in that case).

connor4312:
Fixes #238591

### DIFF
--- a/src/vs/workbench/contrib/debug/browser/debugSession.ts
+++ b/src/vs/workbench/contrib/debug/browser/debugSession.ts
@@ -7,7 +7,7 @@ import { getActiveWindow } from '../../../../base/browser/dom.js';
 import * as aria from '../../../../base/browser/ui/aria/aria.js';
 import { mainWindow } from '../../../../base/browser/window.js';
 import { distinct } from '../../../../base/common/arrays.js';
-import { Queue, RunOnceScheduler, raceTimeout } from '../../../../base/common/async.js';
+import { Queue, RunOnceScheduler, Sequencer, raceTimeout } from '../../../../base/common/async.js';
 import { CancellationToken, CancellationTokenSource } from '../../../../base/common/cancellation.js';
 import { canceled } from '../../../../base/common/errors.js';
 import { Emitter, Event } from '../../../../base/common/event.js';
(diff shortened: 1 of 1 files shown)


Tyriar (dismissed):

bpasero:
@connor4312 Red CI

Tyriar (approved):

PR #265769, opened 2025-09-09
fix oswe thinking because oswe model does not give us thinking ids

justschen:
note: we are already tracking the current thinking part. we can use this check again, but more often tha not, id check is good enough. 

360-Wynaut

### DIFF
--- a/src/vs/workbench/contrib/chat/browser/chatContentParts/chatThinkingContentPart.ts
+++ b/src/vs/workbench/contrib/chat/browser/chatContentParts/chatThinkingContentPart.ts
@@ -157,7 +157,18 @@ export class ChatThinkingContentPart extends ChatCollapsibleContentPart implemen
 			return false;
 		}
 
-		return other?.id !== this.id;
+		const otherId = other?.id;
+		const thisId = this.id;
+
+		// one off case where we have no ids, we compare text instead.
+		if (otherId === undefined && thisId === undefined) {
+			const rawValue = other.value;
+			const otherValueStr = typeof rawValue === 'string' ? rawValue : Array.isArray(rawValue) ? rawValue.join('') : '';
+			const otherValueNormalized = otherValueStr.trim();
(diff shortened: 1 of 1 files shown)


src/vs/workbench/contrib/chat/browser/chatContentParts/chatThinkingContentPart.ts
```diff
@@ -157,7 +157,18 @@ export class ChatThinkingContentPart extends ChatCollapsibleContentPart implemen
 			return false;
 		}
 
-		return other?.id !== this.id;
+		const otherId = other?.id;
```
roblourens, line 160: If it makes it easier, you could generate an id for them in the extension, right? But since the API apparently makes the IDs optional then this isn't an edge case, it's something that the API says is fair game to happen

roblourens (approved):

src/vs/workbench/contrib/chat/browser/chatContentParts/chatThinkingContentPart.ts
justschen: yeahhh anything is fair game, although OSWE is the only outlier since all the other models i've tested _do_ have ids. i'll keep an eye out on this to see if there is any weirdness (and also ask Poornima if OSWE should be adding cot_ids cc @karthiknadig )

karthiknadig: That sounds like a bug in the Endpoint. Can you get the delta logs to share with Poornima?

roblourens: The point is that you should be concerned with what the API says is ok, not what a particular model does on a tuesday. The API says that the id is optional so we should handle it well, and if that's too weird then we should change the API.

PR #212163, opened 2024-05-07
Fire onDidRegisterAllSupported executions if any execution type is re…

mahmoudsalah1993:
…gistered, this unblocks Tasks: Run Tasks and Configure Tasks from infinitely loading in web envrionments where this event never gets fired.

Fixes: https://github.com/microsoft/vscode/issues/212047

PR #265847, opened 2025-09-09
fix: race condition on supported task types

### LINKED ISSUE #265845: Race conditon on supported task types in web + server environment

- VS Code Version: 
- OS Version: 

Steps to Reproduce:

1. Open a project in VSCode web connected to a VSCode server (in a container for instance) that contains `shell` tasks that run on "folderOpen"
2. Sometimes, tasks are run before the server reports the supported tasks type,
3. The `task` output reports: `Warning: shell tasks are unavailable in the current environment.`

Here's the [suggested fix](https://github.com/microsoft/vscode/pull/265847), which may help understanding the issue

CGNonofr:
fix #265845

In https://github.com/microsoft/vscode/pull/212163, the `registerSupportedExecutions` was changed so that in web mode, the function doesn't wait for all default task extensions registration before running the task, because in web mode, only the `custom` mode is supported.

However, when there is a remote server, the server will register the support for all task execution types, but later.

A race condition can happen, in which the tasks are parsed before the server reports the supported task executions, leading to a `Warning: shell tasks are unavailable in the current environment.` error.

This PR makes the `registerSupportedExecutions` method wait for all task supported executions in web mode when there is a remote agent.

### DIFF
--- a/src/vs/workbench/contrib/tasks/browser/abstractTaskService.ts
+++ b/src/vs/workbench/contrib/tasks/browser/abstractTaskService.ts
@@ -432,7 +432,7 @@ export abstract class AbstractTaskService extends Disposable implements ITaskSer
 		// update tasks so an incomplete list isn't returned when getWorkspaceTasks is called
 		this._workspaceTasksPromise = undefined;
 		this._onDidRegisterSupportedExecutions.fire();
-		if (Platform.isWeb || (custom && shell && process)) {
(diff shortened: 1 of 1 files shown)


meganrogge (approved):

meganrogge:
Thanks!

connor4312 (approved):

PR #265920, opened 2025-09-09
add dictation actions to terminal overflow menu

### LINKED ISSUE #265805: Add Button for Voice Dictation in Terminal
Type: <b>Feature Request</b>

VS Code supports voice dictation in terminals via commands like "Terminal: Start Dictation in Terminal" and "Terminal: Stop Dictation in Terminal," but accessing them requires remembering specific commands, which can be inconvenient. Adding a dedicated button in the terminal interface for starting and stopping dictation would make it more intuitive and accessible, similar to other terminal controls. This would enhance usability for developers using voice input during terminal sessions.

VS Code version: Code - Insiders 1.104.0-insider (d1d3e550adbaa5c607f0b300f4bf16e2fcdc2f38, 2025-09-09T04:29:25.960Z)
Modes:

meganrogge:
fix #265805 (also added it to the editor's menu)

Screenshot 2025-09-09 at 2 53 14 PM

### DIFF
--- a/src/vs/workbench/contrib/terminal/browser/terminalMenus.ts
+++ b/src/vs/workbench/contrib/terminal/browser/terminalMenus.ts
@@ -33,8 +33,7 @@ export const enum TerminalMenuBarGroup {
 	Create = '1_create',
 	Run = '3_run',
 	Manage = '5_manage',
-	Configure = '7_configure',
-	Voice = '9_voice'
+	Configure = '7_configure'
 }
 
 export function setupTerminalMenus(): void {
@@ -108,7 +107,7 @@ export function setupTerminalMenus(): void {
 						id: TerminalCommandId.StartVoice,
 						title: localize('workbench.action.terminal.startVoice', "Start Dictation"),
 					},
-					group: TerminalMenuBarGroup.Voice,
+					group: 'navigation',
 					order: 5,
 					when: HasSpeechProvider,
 					isHiddenByDefault: true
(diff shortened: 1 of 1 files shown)


src/vs/workbench/contrib/terminal/browser/terminalMenus.ts
```diff
@@ -98,7 +100,33 @@ export function setupTerminalMenus(): void {
...
-			}
+			},
+			{
+				id: MenuId.ViewTitle,
+				item: {
+					command: {
+						id: TerminalCommandId.StartVoice,
+						title: localize('workbench.action.terminal.startVoice', "Start Dictation"),
+					},
+					group: TerminalMenuBarGroup.Voice,
```
Tyriar: I think these living in the voice group, not `'navigation'`, means it won't support right click to show it for terminal view specifically?

Tyriar (changes requested):

src/vs/workbench/contrib/terminal/browser/terminalMenus.ts
meganrogge: Oh, TIL!

meganrogge: Changed it to `navigation` and that didn't seem to work 🤔 

Screenshot 2025-09-09 at 4 56 22 PM

meganrogge: I also think it looks better in a separate group, even if that means no right clicking.

Tyriar: @meganrogge that's the tab menu, you want to right click on the view actions:

image

> I also think it looks better in a separate group

It does look better, but it defeats the purpose imo if you cannot toggle it to show always as it's not very convenient in the overflow.

meganrogge: ah, in that case we need Codicons for the actions too.

Tyriar (approved):

PR #265923, opened 2025-09-09
Adding chat input type to chat model

osortega:
Port of candidate https://github.com/microsoft/vscode/pull/265702

### DIFF
--- a/src/vs/workbench/contrib/chat/browser/chatEditorInput.ts
+++ b/src/vs/workbench/contrib/chat/browser/chatEditorInput.ts
@@ -172,13 +172,16 @@ export class ChatEditorInput extends EditorInput implements IEditorCloseHandler 
 	}
 
 	override async resolve(): Promise<ChatEditorModel | null> {
+		const searchParams = new URLSearchParams(this.resource.query);
+		const chatSessionType = searchParams.get('chatSessionType');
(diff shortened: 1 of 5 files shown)


src/vs/workbench/contrib/chat/browser/chatEditorInput.ts
```diff
@@ -172,13 +172,16 @@ export class ChatEditorInput extends EditorInput implements IEditorCloseHandler
 	}
 
 	override async resolve(): Promise<ChatEditorModel | null> {
+		const searchParams = new URLSearchParams(this.resource.query);
+		const chatSessionType = searchParams.get('chatSessionType');
+		const inputType = chatSessionType ?? this.resource.authority;
```
mjbvz, line 177: Why can't we always use the authority? Could the resource be persisted instead of the inputType?

osortega: @mjbvz we don't have an authority yet by the time we create a new chat editor on an external provider.
We are persisting the `inputType` in the model data but we still have cases where we need to pass it during the chat model creation, specifically when we are creating new empty chat editors

Yoyokrazy (approved):

Issue #224505, opened 2024-08-01
terminal completion provider extension API


Tyriar:
Now that the terminal suggest is getting more reliable and feature complete, we should look into what a completion provider could look like in the extension API. Some examples of where this would be useful:

- Providing completions for shells we don't support yet, like integrating fig completion specs.
- A provider for git that added smarts beyond what we can do in pwsh. For example it could complete `git clone https://` to `git clone https://github.com/` and provide the user's orgs/forks when completing `https://github.com/`, etc.

Tyriar:
@meganrogge you might be tracking this one elsewhere

meganrogge:
API sync feedback:

- look into why cwd is optional but somehow mandatory
- look into why env is on resource request config
- pass in id to registerCompletionProvider vs having id on the provider
- ensure it's clear that id needs to be unique, mention where it will show up (setting, hover w cmd +/ )
- add comments and examples jsdoc
- consider glob pattern vs file extensions?

Tyriar:
API sync feedback (mostly the same as above):

- Add docs for everything, explain commandLine, cursorPosition, replacement* etc.
- `TerminalCompletionProvider.id`
  - id should instead me an arg in in call to register the provider
  - Explain requirements: unique to extension, globally?
  - Do we need to prefix the ID with the extension ID internally so extensions won't step on each other?
  - Hint where id shows up for the user (providers setting, ctrl+/ debug info)
- `TerminalResourceRequestConfig.env`
  - Wasn't clear to me what this does?
- `TerminalResourceRequestConfig.cwd`
  - Is it mandatory or not? Comment contradicts somewhat
- Buffer access likely won't change `commandLine`
- `TerminalResourceRequestConfig.fileExtensions`
  - Should we use `RelativePattern`?

Tyriar:
We also need these so extension contributed providers integrate nicely and turn on automatically:

- https://github.com/microsoft/vscode/issues/254869
- https://github.com/microsoft/vscode/issues/254872

Tyriar:
Another one we should handle before stabilizing:

- https://github.com/microsoft/vscode/issues/257658

meganrogge:
Feedback:

- Remove static contribution in the API. 
- Add a selector for the completion provide so we know what type of shell it supports. We can allow * if it tries to support all shells

```
export function registerTerminalCompletionProvider<T extends TerminalCompletionItem>(
	selector: TerminalSelector,
[...]

PR #266073, opened 2025-09-10
address API feedback for terminal completion provider

meganrogge:
addresses https://github.com/microsoft/vscode/issues/224505#issuecomment-3049538549

- [x] look into why cwd is optional but somehow mandatory
    - I have made this mandatory
- [x] look into why env is on resource request config
    - I have removed this, wasn't necessary
- [x] pass in id to registerCompletionProvider vs having id on the provider
- [x] ensure it's clear that id needs to be unique, mention where it will show up (setting, hover w cmd +/ )
- [x] add comments and examples jsdoc
- [x] consider glob pattern vs file extensions?

### DIFF
--- a/extensions/terminal-suggest/src/terminalSuggestMain.ts
+++ b/extensions/terminal-suggest/src/terminalSuggestMain.ts
@@ -306,11 +306,12 @@ export async function activate(context: vscode.ExtensionContext) {
 
 			const cwd = result.cwd ?? terminal.shellIntegration?.cwd;
 			if (cwd && (result.filesRequested || result.foldersRequested)) {
+				const globPattern = createGlobPattern(result.fileExtensions);
 				return new vscode.TerminalCompletionList(result.items, {
 					filesRequested: result.filesRequested,
 					foldersRequested: result.foldersRequested,
-					fileExtensions: result.fileExtensions,
-					cwd
+					globPattern,
+					cwd,
 				});
 			}
 			return result.items;
@@ -563,3 +564,11 @@ export function sanitizeProcessEnvironment(env: Record<string, string>, ...prese
 		});
 }
 
+
+function createGlobPattern(fileExtensions?: string[]): string | undefined {
+	let globPattern: string | undefined;
+	if (fileExtensions && fileExtensions.length > 0) {
+		globPattern = `{${fileExtensions.map(ext => `*${ext.startsWith('.') ? ext : '.' + ext}`).join(',')}}`;
+	}
+	return globPattern;
+}
--- a/extensions/terminal-suggest/src/terminalSuggestMain.ts
+++ b/extensions/terminal-suggest/src/terminalSuggestMain.ts
@@ -566,9 +566,10 @@ export function sanitizeProcessEnvironment(env: Record<string, string>, ...prese
 
 
 function createGlobPattern(fileExtensions?: string[]): string | undefined {
-	let globPattern: string | undefined;
-	if (fileExtensions && fileExtensions.length > 0) {
-		globPattern = `{${fileExtensions.map(ext => `*${ext.startsWith('.') ? ext : '.' + ext}`).join(',')}}`;
+	if (!fileExtensions || fileExtensions.length === 0) {
+		return undefined;
 	}
-	return globPattern;
+	const exts = fileExtensions.map(ext => ext.startsWith('.') ? ext : '.' + ext);
+	// Create a regex that matches any string ending with one of the extensions
+	return `.*(${exts.map(ext => ext.replace('.', '\\.')).join('|')})$`;
 }
--- a/extensions/terminal-suggest/src/terminalSuggestMain.ts
+++ b/extensions/terminal-suggest/src/terminalSuggestMain.ts
@@ -250,8 +250,7 @@ export async function activate(context: vscode.ExtensionContext) {
 	const machineId = await vscode.env.machineId;
 	const remoteAuthority = vscode.env.remoteName;
 
-	context.subscriptions.push(vscode.window.registerTerminalCompletionProvider({
-		id: 'terminal-suggest',
+	context.subscriptions.push(vscode.window.registerTerminalCompletionProvider('terminal-suggest', {
 		async provideTerminalCompletions(terminal: vscode.Terminal, terminalContext: vscode.TerminalCompletionContext, token: vscode.CancellationToken): Promise<vscode.TerminalCompletionItem[] | vscode.TerminalCompletionList | undefined> {
 			currentTerminalEnv = terminal.shellIntegration?.env?.value ?? process.env;
 			if (token.isCancellationRequested) {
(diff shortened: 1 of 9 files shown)


src/vscode-dts/vscode.proposed.terminalCompletionProvider.d.ts
```diff
@@ -20,6 +19,16 @@ declare module 'vscode' {
 	}
 
 
+	/**
+	 * If the command line is `ls -|` where | is the cursor,
+	 * the following would be the values for the completion items:
```
meganrogge:
```suggestion
	 * the following would be the values for a completion item:
```

meganrogge:
@tyriar I wasn't sure what you meant when you said: Buffer access likely won't change commandLine 🤔 do you recall?

src/vscode-dts/vscode.proposed.terminalCompletionProvider.d.ts
```diff
@@ -20,6 +19,16 @@ declare module 'vscode' {
...
+	 * If the command line is `ls -|` where | is the cursor,
+	 * the following would be the values for a completion item:
+	 * label: -A
+	 * replacementIndex: 3
+	 * replacementLength: 1
+	 * detail: List all entries except for . and ...  Always set for the super-user
+	 * documentation: none
+	 * kind: TerminalCompletionItemKind.Flag
+	 */
 	export interface TerminalCompletionItem {
```
Tyriar, line 50: This isn't formatted how we'd want it in the eventual API. Something like this could live in an `@example` though.

src/vscode-dts/vscode.proposed.terminalCompletionProvider.d.ts
```diff
@@ -138,16 +146,12 @@ declare module 'vscode' {
...
 		/**
-		 * File extensions to filter by.
-		 */
-		fileExtensions?: string[];
-		/**
-		 * If no cwd is provided, no resources will be shown as completions.
+		 * A {@link GlobPattern glob pattern} that controls which files suggest should surface.
 		 */
-		cwd?: Uri;
+		globPattern?: GlobPattern;
```
Tyriar, line 201: Did you look into RelativePattern too? https://github.com/microsoft/vscode/issues/224505#issuecomment-3049547126

> TerminalResourceRequestConfig.fileExtensions
> Should we use RelativePattern?

src/vscode-dts/vscode.proposed.terminalCompletionProvider.d.ts
```diff
@@ -96,11 +104,11 @@ declare module 'vscode' {
 	export namespace window {
 		/**
 		 * Register a completion provider for a certain type of terminal.
-		 *
+		 * @param id The unique identifier of the terminal provider, used as a settings key and shown in the information hover of the suggest widget.
 		 * @param provider The completion provider.
 		 * @returns A {@link Disposable} that unregisters this provider when being disposed.
 		 */
-		export function registerTerminalCompletionProvider<T extends TerminalCompletionItem>(provider: TerminalCompletionProvider<T>, ...triggerCharacters: string[]): Disposable;
+		export function registerTerminalCompletionProvider<T extends TerminalCompletionItem>(id: string, provider: TerminalCompletionProvider<T>, ...triggerCharacters: string[]): Disposable;
```
Tyriar, line 149: I remember talking about `id` in the API sync, but the editor ones only use a `selector` and no `id`. Do we actually need ID and could we just use the extension ID as the ID instead?

extensions/terminal-suggest/src/terminalSuggestMain.ts
```diff
@@ -565,3 +564,17 @@ export function sanitizeProcessEnvironment(env: Record<string, string>, ...prese
...
+}
+
+function createGlobPattern(fileExtensions?: string[]): vscode.GlobPattern | undefined {
+	if (!fileExtensions || fileExtensions.length === 0) {
+		return undefined;
+	}
+	const exts = fileExtensions.map(ext => ext.startsWith('.') ? ext : '.' + ext);
+	// Create a regex that matches any string ending with one of the extensions
+	return `.*(${exts.map(ext => escapeRegExp(ext)).join('|')})$`;
+}
```
Tyriar: I don't think converting from a glob to a regex is this simple?

Tyriar (dismissed):

src/vscode-dts/vscode.proposed.terminalCompletionProvider.d.ts
meganrogge: yes, a `GlobPattern` just also allows strings

https://github.com/microsoft/vscode/blob/e069a91840775119de814c88eb4dbbfe624007a1/src/vscode-dts/vscode.d.ts#L2329

src/vscode-dts/vscode.proposed.terminalCompletionProvider.d.ts
meganrogge: An extension could contribute multiple providers though

extensions/terminal-suggest/src/terminalSuggestMain.ts
meganrogge: here, we're converting from a `string[]` to a `string`. `vscode.GlobPattern` can be a `string`.

src/vscode-dts/vscode.proposed.terminalCompletionProvider.d.ts
meganrogge: added support for `RelativePattern`

src/vscode-dts/vscode.proposed.terminalCompletionProvider.d.ts
Tyriar: The editor ones do too, they don't need to care about id's?

extensions/terminal-suggest/src/terminalSuggestMain.ts
Tyriar: The comment says create a regex?

src/vs/workbench/contrib/terminalContrib/suggest/browser/terminalCompletionService.ts
```diff
@@ -442,9 +443,12 @@ export class TerminalCompletionService extends Disposable implements ITerminalCo
 
 			label = escapeTerminalCompletionLabel(label, shellType, resourceRequestConfig.pathSeparator);
 
-			if (child.isFile && fileExtensions) {
-				const extension = child.name.split('.').length > 1 ? child.name.split('.').at(-1) : undefined;
-				if (extension && !fileExtensions.includes(extension)) {
+			if (child.isFile && globPattern) {
+				const filePath = child.resource.fsPath;
+				const matches = typeof globPattern === 'string'
+					? new RegExp(globPattern).test(filePath)
```
roblourens, line 449: A glob is not a valid regex, probably you want to use something from glob.ts

roblourens (approved):

Issue #266087, opened 2025-09-10
Don't require restart when switching to TS native preview


mjbvz:
Currently you have to restart the extension host when switching to the TS Native preview. This shouldn't be required

mjbvz:
To verify:

1. Install the `TypeScript (Native Preview)` extension
2. Run the `TypeScript Native Preview: enable` command
3. Confirm ext host isn't restarted (one easy way to check is to have an active debug session before running the command)
4. Check that IntelliSense works as expect
5. Disable the native preview and make sure no restart occurs and IntelliSense still works

alexr00:
I saw the extension host get restarted.

![Image](https://github.com/user-attachments/assets/5975f50e-6076-408f-9423-795b1068cd81)

rzhao271:
The extension still asks for a restart.

<img width="681" height="309" alt="Restart prompt shown after running the enable command" src="https://github.com/user-attachments/assets/86f2aae7-2c27-472a-b2e9-9b381e329993" />

mjbvz:
To verify, please wait until the next TS native build is released upstream

PR #266088, opened 2025-09-10
Don't require restart when switching to TS native

mjbvz:
For #266087

Dynamically registers / unregisters the language features

cc @andrewbranch


roblourens (approved):

andrewbranch (approved):
Thank you!

PR #266096, opened 2025-09-10
Avoid using disposables in domSanitize

mjbvz:
We can use `removeAllHooks` instead of unregistering each one individually to save a small amount of overhead

### DIFF
--- a/src/vs/base/browser/domSanitize.ts
+++ b/src/vs/base/browser/domSanitize.ts
@@ -3,11 +3,9 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-import { DisposableStore, IDisposable, toDisposable } from '../common/lifecycle.js';
 import { Schemas } from '../common/network.js';
 import { reset } from './dom.js';
 import dompurify from './dompurify/dompurify.js';
-
 
 /**
  * List of safe, non-input html tags.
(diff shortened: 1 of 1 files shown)


src/vs/base/browser/domSanitize.ts
```diff
@@ -304,21 +288,22 @@ function doSanitizeHtml(untrusted: string, config: DomSanitizerConfig | undefine
...
+			});
+
 		if (config?.replaceWithPlaintext) {
-			store.add(addDompurifyHook('uponSanitizeElement', replaceWithPlainTextHook));
+			dompurify.addHook('uponSanitizeElement', replaceWithPlainTextHook);
 		}
 
 		if (allowedAttrPredicates.size) {
-			store.add(addDompurifyHook('uponSanitizeAttribute', (node, e) => {
+			dompurify.addHook('uponSanitizeAttribute', (node, e) => {
```
roblourens, line 306: I think I remember doing this because otherwise it was too easy to forget to clean up hooks properly

roblourens: Maybe the difference was that we were removing them individually before and not using `removeAllHooks` and one was missed

roblourens (approved):

src/vs/base/browser/domSanitize.ts
mjbvz: Good point. Not a fan of global state like this and I agree this is potentially a little more confusing to use. However I'm confident in this change because I've gradually refactored our code so that `dompurify` is only used directly from the `domSanitize` file, and only called from one function: `doSanitizeHtml`

I will ban importing `dompurify` from anywhere else just to make sure no one tries using it directly and also add myself as a code owner for `domsanitize.ts`

PR #266099, opened 2025-09-10
focus terminal, await user enter if free form input is expected in the terminal

meganrogge:
fixes https://github.com/microsoft/vscode/issues/264584

Makes the `reject` action optional in the `ChatRequestElicitationPrompt`

https://github.com/user-attachments/assets/27755674-07f6-40f5-9688-74c542fc37c7

### DIFF
--- a/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/monitoring/outputMonitor.ts
+++ b/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/monitoring/outputMonitor.ts
@@ -439,10 +439,8 @@ export class OutputMonitor extends Disposable implements IOutputMonitor {
 			Now, analyze this output:
 			${lastFiveLines}
 			`;
-		const response = await this._languageModelsService.sendChatRequest(models[0], new ExtensionIdentifier('core'), [
-			{ role: ChatMessageRole.User, content: [{ type: 'text', value: promptText }] }
-		], {}, token);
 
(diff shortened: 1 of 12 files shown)


src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/monitoring/outputMonitor.ts
```diff
@@ -491,6 +523,54 @@ export class OutputMonitor extends Disposable implements IOutputMonitor {
...
+		const chatModel = this._chatService.getSession(execution.sessionId);
+		if (!(chatModel instanceof ChatModel)) {
+			return false;
+		}
+		const request = chatModel.getRequests().at(-1);
+		if (!request) {
+			return false;
+		}
+		const userPrompt = new Promise<boolean>(resolve => {
+			const thePart = this._register(new ChatElicitationRequestPart(
```
connor4312, line 536: I apparently missed when this was first added, but these parts should not be disposable imo. And if they were (in general) then it would be up to the chat model to dispose them since it's the 'owner' of them. As-is it's holding onto a disposed object which is a bit weird.

I would consider swapping `onDidRequestHide` into an `isHidden: IObservable<boolean>` which would avoid the need for any disposable data in the part.

connor4312 (approved):
lgtm but one comment, turned off auto merge in case you want to address in this PR

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/monitoring/outputMonitor.ts
meganrogge: I will address this separately, thanks!

PR #266101, opened 2025-09-10
Remove `copilot` from strings and localize them

### LINKED ISSUE #264155: Avoid the human readable string "Copilot"
https://github.com/microsoft/vscode/blob/3a40d7a6fc98f4c046f4678df7affaff3266e501/extensions/typescript-language-features/src/languageFeatures/quickFix.ts#L368-L384

We need to find wording that is not "Copilot", as we did in the rest of our UI. I also question the fact that these strings do not seem to be localizable?

mjbvz:
Fixes #264155

### DIFF
--- a/extensions/typescript-language-features/src/languageFeatures/quickFix.ts
+++ b/extensions/typescript-language-features/src/languageFeatures/quickFix.ts
@@ -366,22 +366,19 @@ class TypeScriptQuickFixProvider implements vscode.CodeActionProvider<VsCodeCode
 			let expand: Expand | undefined;
 			let title = action.description;
 			if (action.fixName === fixNames.classIncorrectlyImplementsInterface) {
-				title += ' with Copilot';
+				title = vscode.l10n.t('{0} with AI', action.description);
 				message = `Implement the stubbed-out class members for ${document.getText(diagnostic.range)} with a useful implementation.`;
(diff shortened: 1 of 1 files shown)


Tyriar (approved):

extensions/typescript-language-features/src/languageFeatures/quickFix.ts
```diff
@@ -366,30 +366,27 @@ class TypeScriptQuickFixProvider implements vscode.CodeActionProvider<VsCodeCode
 			let expand: Expand | undefined;
 			let title = action.description;
 			if (action.fixName === fixNames.classIncorrectlyImplementsInterface) {
-				title += ' with Copilot';
+				title = vscode.l10n.t('{0} with AI', action.description);
 				message = `Implement the stubbed-out class members for ${document.getText(diagnostic.range)} with a useful implementation.`;
```
DanTup, line 370: @mjbvz should this `message` on line 370 be wrapped with `vscode.l10n.t`?

(I'm not familiar with the localization, I was just being nosey and noticed that all the others were updated but not this one)

mjbvz: Yes thanks should be localized as well

PR #266123, opened 2025-09-11
feat: add resource-scoped prompt file recommendations

digitarald:
This PR enhances the `chat.promptFilesRecommendations` setting to support workspace folder-level configuration with proper inheritance hierarchy.

## Changes

- **Configuration scope**: Added `ConfigurationScope.RESOURCE` to enable folder-level settings
- **Context-aware recommendations**: Prompt suggestions now consider the active editor's workspace folder
- **Configuration merging**: Implemented proper inheritance (default → user → workspace → folder)

## Key Benefits

- Different projects can have tailored prompt recommendations
- Workspace folder settings override workspace/user settings appropriately  
- Maintains backward compatibility with existing global configurations
- Enables resource-aware when clauses (e.g., `resourceExtname == .js`) to work correctly

## Implementation

Follows established VS Code patterns:
- Uses `EditorResourceAccessor.getOriginalUri()` for resource context (consistent with 20+ other features)
- Leverages `configService.inspect()` with resource parameter for proper configuration inheritance
- Maintains type safety and robust error handling

The changes make the existing documentation for resource-aware when clauses fully functional.


src/vs/workbench/contrib/chat/common/promptSyntax/config/config.ts
```diff
@@ -165,19 +166,31 @@ export namespace PromptsConfig {
...
-		// note! this would be also true for `null` and `array`,
-		// 		 but those cases are already handled above
-		if (typeof configValue === 'object') {
-			const suggestions: Record<string, boolean | string> = {};
+		// Merge configurations in priority order: default → user → workspace → folder
+		// Later values override earlier ones for the same prompt name
+		const mergedSuggestions: Record<string, boolean | string> = {};
+
+		// Helper function to merge a configuration object
+		const mergeConfig = (configValue: Record<string, boolean | string> | undefined) => {
```
aeschli: From what I know (and @sandy081 confirms) the configuration service already does that merging of settings with object values.  `configService.getValue(key, {resource})` will do

src/vs/workbench/contrib/chat/browser/chatWidget.ts
```diff
@@ -1240,8 +1243,12 @@ export class ChatWidget extends Disposable implements IChatWidget {
 	}
 
 	private getPromptFileSuggestions(): IChatSuggestedPrompts[] {
+		// Get the current workspace folder context if available
+		const activeEditor = this.editorService.activeEditor;
```
aeschli, line 1247: Does this mean that the welcome view content will be re-rendered on every active editor change?

src/vs/workbench/contrib/chat/common/promptSyntax/config/config.ts
digitarald: Good find, fixed!

src/vs/workbench/contrib/chat/browser/chatWidget.ts
digitarald: No, the welcome view doesn't re-render on every active editor change. The `getPromptFileSuggestions()` method only runs when the welcome view itself needs to be rendered (when the chat is empty). The active editor access is just for reading configuration context, so the performance impact is minimal - just a configuration lookup when the welcome view is already being rendered for other reasons.

aeschli: Isn't it a bit random which editor is active at that time? Could be an external editor, or the settings editor. I'm not concerned about the performance, but about the predictability of this feature.

aeschli (approved):

PR #266239, opened 2025-09-11
Fix TOC in release notes

mjbvz:
Fixes #266235

Ports smaller fix to release. We always should create html strings before sanitization and then sanitize them

Also enables `nav` elements since the release notes uses these

### DIFF
--- a/src/vs/workbench/contrib/update/browser/releaseNotesEditor.ts
+++ b/src/vs/workbench/contrib/update/browser/releaseNotesEditor.ts
@@ -265,13 +265,21 @@ export class ReleaseNotesManager extends Disposable {
 	private async renderBody(fileContent: { text: string; base: URI }) {
 		const nonce = generateUuid();
 
-		const content: { toString: () => string } = await renderMarkdownDocument(fileContent.text, this._extensionService, this._languageService, {
+		// Remove HTML comment markers around table of contents navigation
+		const rawFileContent = fileContent.text
+			.replace(/<!--\s*TOC\s*/gi, '')
+			.replace(/\s*Navigation End\s*-->/gi, '');
+
+		const processedContent: { toString: () => string } = await renderMarkdownDocument(rawFileContent, this._extensionService, this._languageService, {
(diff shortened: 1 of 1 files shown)


src/vs/workbench/contrib/update/browser/releaseNotesEditor.ts
```diff
@@ -265,13 +265,21 @@ export class ReleaseNotesManager extends Disposable {
...
+		const processedContent: { toString: () => string } = await renderMarkdownDocument(rawFileContent, this._extensionService, this._languageService, {
 			sanitizerConfig: {
 				allowedMediaProtocols: {
 					override: '*' // TODO: remove once we can use <base> to find real resource locations
 				},
 				allowedLinkProtocols: {
 					override: [Schemas.http, Schemas.https, Schemas.command]
+				},
+				allowedTags: {
+					augment: ['nav']
```
mjbvz, line 282: 'nav' is not in the baseline set of tags as you can't normally create it with markdown

src/vs/workbench/contrib/update/browser/releaseNotesEditor.ts
```diff
@@ -265,13 +265,21 @@ export class ReleaseNotesManager extends Disposable {
 	private async renderBody(fileContent: { text: string; base: URI }) {
 		const nonce = generateUuid();
 
-		const content: { toString: () => string } = await renderMarkdownDocument(fileContent.text, this._extensionService, this._languageService, {
+		// Remove HTML comment markers around table of contents navigation
+		const rawFileContent = fileContent.text
+			.replace(/<!--\s*TOC\s*/gi, '')
+			.replace(/\s*Navigation End\s*-->/gi, '');
```
mjbvz, line 271: We were using html comments to wrap the TOC. However the sanitizer removes html comments

Moving the replace before makes sure we unwrap the TOC before trying to render it

JaclynCodes (commented):
Ok

osortega (approved):

PR #266241, opened 2025-09-11
Fix TOC in release notes

mjbvz:
For #266235

We should create the html content before sanitization, not after

Also adds tests so we can catch regressions in the future


alexr00 (approved):
This is great, thank you!

PR #248625, opened 2025-05-10
Fix PowerShell shell integration when strict mode is enabled.

mawosoft:
Check if variable exists before accessing it to prevent script failures in strict mode.

Issue #255699, opened 2025-07-14
VS Code Terminal Stack Overflow (-1073741571) Triggered by fnm PATH Injection


stevenried:
🐞 Bug Report: PowerShell 7 Crashes in Integrated Terminal with Exit Code -1073741571 When `fnm` Is Active

VS Code Version: 1.102.0
PowerShell Version: 7.5.2
Node Manager: fnm
OS: Windows 10/11
Shell: PowerShell 7 (pwsh.exe)

Error Message:
The terminal process "C:\Program Files\PowerShell\7\pwsh.exe" terminated with exit code: -1073741571

💥 Issue Summary
When launching PowerShell 7 in VS Code’s integrated terminal, the shell crashes with exit code -1073741571 (stack overflow). This only occurs when `fnm` is installed and active. PowerShell 7 works fine outside VS Code, and other shells work inside VS Code.

🔍 Steps to Reproduce
1. Install fnm and activate a Node version.
2. Launch VS Code.
3. Open integrated terminal with PowerShell 7 as default.
4. Observe crash with exit code -1073741571.

✅ What I’ve Tried
- Launching with -NoProfile
- Disabling all extensions (`code --disable-extensions`)
- Running VS Code as Admin
- Creating a clean wrapper script for PowerShell
- Comparing $env:PATH across environments

Does this issue occur when all extensions are disabled?: Yes

📎 Environment Clues
- fnm injects multiple fnm_multishells/... entries into $env:PATH
- Removing those entries resolves the crash
- Crash only occurs inside VS Code’s integrated terminal

📌 Suspected Cause
VS Code’s terminal renderer may be sensitive to deep or malformed environment chains introduced by `fnm`. This appears to trigger a stack overflow during shell startup.

🙏 Request
Can you investigate whether VS Code’s terminal subsystem can be made more resilient to injected PATH variables like those from fnm? Or suggest ways to isolate shell profiles more safely?

X9VoiD:
This issue happens even without `fnm`.

damadorPL:
<img width="539" height="192" alt="Image" src="https://github.com/user-attachments/assets/d70fe6a1-6a15-4d16-87cb-6228e8c4f823" />

got the same error bout only on one of my machines - both have the same config / os version etc ( win11 )  - error caused in python project

AlexJ-StL:
[...]

PR #266260, opened 2025-09-11
Fix PowerShell shell integration when strict mode is enabled.

mawosoft:
Check if variable exists before accessing it to prevent script failures in strict mode.

See #248625 for previous fix.


Tyriar (dismissed):
fyi @anthonykim1

Tyriar (changes requested):
@mawosoft can you comment on https://github.com/microsoft/vscode/issues/255699 which was fixed via https://github.com/microsoft/vscode/commit/d62248e1a1cfed85464b92ca11ebe33dc4e45db1? I'm not familiar with that syntax in pwsh and don't want this pretty bad issue to come back.

Tyriar (approved):
Makes sense, thanks!

alexr00 (approved):

PR #266277, opened 2025-09-11
tools: allow starting MCP servers from the tools picker

connor4312:
- Makes `onDidAccept` in the tree picker work (it previously no-op'd)
- Make `pickable: false` work correctly.
- Update the tree to let it recompute its entries within the tree
- Swapped `toolsEntries` to a getter since the map needs to be
  re-retrieved after the MCP server is available, or tools don't show.


TylerLeonhardt (approved):
Nice! Approving from the quick tree side but @aeschli  should maybe do a pass on the tool picker logic

PR #266537, opened 2025-09-13
Fix and update the definition of fullwidth characters

radobot:
This fixes and updates the list of wide and fullwidth Unicode characters. Previously, some wide characters were not included and some were incorrectly classified as fullwidth when they aren't.

To determine which characters to include, the Unicode East Asian Width property is used. Since this property is not commonly available in runtimes a hardcoded list is used instead.

### DIFF
--- a/src/vs/base/common/strings.ts
+++ b/src/vs/base/common/strings.ts
@@ -688,50 +688,34 @@ export function containsUnusualLineTerminators(str: string): boolean {
 	return UNUSUAL_LINE_TERMINATORS.test(str);
 }
 
-export function isFullWidthCharacter(charCode: number): boolean {
+export function isFullWidthCharacter(codePoint: number): boolean {
 	// Do a cheap trick to better support wrapping of wide characters, treat them as 2 columns
-	// http://jrgraphix.net/research/unicode_blocks.php
-	//          2E80 - 2EFF   CJK Radicals Supplement
-	//          2F00 - 2FDF   Kangxi Radicals
-	//          2FF0 - 2FFF   Ideographic Description Characters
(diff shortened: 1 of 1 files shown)


aiday-mar (commented):
Hi @radobot thank you for taking the time to do this lengthy work. I looked at the code and at the links you provided, and I see that as you mentioned, there are missing ranges for fullwidth characters. I think for the sake of completeness and in order to be rigorous, we will have to include all the appropriate fullwidth characters and not integrate certain characters marked as A (ambiguous) into the fullwidth category. Could you please mark them all as appropriate?

I am not sure also what the early termination condition in the code refers to. I was wondering if you could explain?

In addition to the above, could you please leave a comment like the one there is currently, which maps the ranges to specific character types as follows (example):

```
	//          3000 - 303F   CJK Symbols and Punctuation
	//          3040 - 309F   Hiragana
	//          30A0 - 30FF   Katakana
	//          3100 - 312F   Bopomofo
	//          3130 - 318F   Hangul Compatibility Jamo
	//          3190 - 319F   Kanbun
	//          31A0 - 31BF   Bopomofo Extended
	//          31F0 - 31FF   Katakana Phonetic Extensions
	//          3200 - 32FF   Enclosed CJK Letters and Months
	//          3300 - 33FF   CJK Compatibility
	//          3400 - 4DBF   CJK Unified Ideographs Extension A
	//          4DC0 - 4DFF   Yijing Hexagram Symbols
	//          4E00 - 9FFF   CJK Unified Ideographs
	//          A000 - A48F   Yi Syllables
	//          A490 - A4CF   Yi Radicals
	//          AC00 - D7AF   Hangul Syllables
```

Thank you.

PR #267272, opened 2025-09-17
thinking part OSWE fixes

justschen:
370-Luvdisc


lramos15 (approved):

src/vs/workbench/contrib/chat/browser/chatContentParts/chatThinkingContentPart.ts
```diff
@@ -18,9 +18,13 @@ import { ButtonWithIcon } from '../../../../../base/browser/ui/button/button.js'
 import { Codicon } from '../../../../../base/common/codicons.js';
 import { ThemeIcon } from '../../../../../base/common/themables.js';
 
+function stripImSepMarkers(text: string): string {
+	return text.replace(/<\|im_sep\|>(\*{4,})?/g, '');
```
roblourens, line 22: Ideally this would be done in the extension, not at the view level, IMO

roblourens: Well, ideally above us the API layer would not return this junk from the model I guess.

Issue #267312, opened 2025-09-18
Notebook overlay layout problem


harbin1053020115:
Does this issue occur when all extensions are disabled?: Yes/No

- VS Code Version: 1.104.0
- OS Version: Darwin arm64 24.3.0

Steps to Reproduce:

1. Add a Header Element which is 42px height on top of the vscode container element(className="monaco-workbench")
2. Open the ipynb editor
3. The notebookOverlay element position.top is wrong:

<img width="6016" height="1192" alt="Image" src="https://github.com/user-attachments/assets/b864468a-a94e-41d0-8926-60b3b527d40e" />

PR #267323, opened 2025-09-18
fix: fix notebook overlay layout problem

harbin1053020115:
#267312


dmitrivMS:
Current notebook overlay positioning has been reworked around anchor-based layout, making this PR's manual top/left offset adjustment obsolete.

PR #198463, opened 2023-11-16
Windows: Polish single prompt line and clear/cls behavior

Tyriar:
Fixes #198278
Fixes #198440

Issue #198278, opened 2023-11-15
Terminal's sticky scroll/shell integration doesn't handle cls well


abhijit-chikane:
- VS Code Version: 1.85.0-insider
- OS Version: Windows11

enable sticky scroll in terminal using "terminal.integrated.stickyScroll.enabled": true
there are couple of issues I observed
sometimes the output of prev. command get's stick along with the new command as shown in below image(cmd: git status)
![image](https://github.com/microsoft/vscode/assets/50770619/ae27aa03-0a6f-430a-9a1b-378174e7b09a)

If I ran the same command first time (git status) then it does stick to the top at but If I ran the same command multiple times it no longer stick to the top. And there are couple of issues if I ran multiline command.

Thanks
But overall love this new feature! ✨

Tyriar:
Glad you like it! I notice you don't have the shell integration decorations to the left of the terminal, did you hide them?

abhijit-chikane:
Nope, I didn't hide anything there

Tyriar:
Oh I was confused, I just noticed the shadow on sticky scroll.

There were some major changes that went in recently, does this still happen on the latest? https://github.com/microsoft/vscode/pull/197994

abhijit-chikane:
I observed it on today's insider release. I think it must be with those changes
Not sure though

Tyriar:
@abhijit-chikane another favor, could you set this `"terminal.integrated.developer.devMode": true`, create a new terminal and try again? I'm interested in where the little A, B, C and D letters are:

![image](https://github.com/microsoft/vscode/assets/2193314/0b8327de-5bf8-448c-8c69-f8a362d89f4e)

abhijit-chikane:
![image](https://github.com/microsoft/vscode/assets/50770619/bbc71d28-2bbf-400a-b7ff-f99e91ab4012)

I don't see any such setting there

Tyriar:
You did it right, it's a hidden setting that should work despite saying unknown

abhijit-chikane:
Oh, I see
For which the sticky scroll breaks I don't see those letters there
![image](https://github.com/microsoft/vscode/assets/50770619/46184349-7eb3-4c03-8d76-1d7f955ff333)

Tyriar:
> One more thing is whenever I open a new terminal I could see the box in the top right corner
is this intended behaviour?

Yep, this is an overlay we use to help improve shell integration.

---

[...]

PR #267439, opened 2025-09-19
Fix bad terminal sticky scroll after `clear` 

### LINKED ISSUE #267408: Bad terminal sticky scroll after running `clear` in terminal
@joaomoreno mentioned that we have bad terminal sticky scroll when running `clear` in terminal.

This seems ugly, doesn't make sense that we display `clear` sticky scroll after running `clear` in the terminal: 
(Tested on Mac)
https://github.com/user-attachments/assets/27f21f61-1ed8-4113-ab11-861737bd9153

We should probably just not show `clear` as terminal sticky scroll when `clear` is run.

/cc @Tyriar

anthonykim1:
Resolves: https://github.com/microsoft/vscode/issues/267408 

This would make sure `user ps1> clear` doesn't show as sticky scroll inappropriately after `clear` has been typed in the terminal.

### DIFF
--- a/src/vs/workbench/contrib/terminalContrib/stickyScroll/browser/terminalStickyScrollOverlay.ts
+++ b/src/vs/workbench/contrib/terminalContrib/stickyScroll/browser/terminalStickyScrollOverlay.ts
@@ -521,7 +521,6 @@ export class TerminalStickyScrollOverlay extends Disposable {
 			selectionInactiveBackground: undefined
 		};
 	}
-
 }
 
 function lineStartsWith(line: IBufferLine | undefined, text: string): boolean {
--- a/src/vs/workbench/contrib/terminalContrib/stickyScroll/browser/terminalStickyScrollOverlay.ts
+++ b/src/vs/workbench/contrib/terminalContrib/stickyScroll/browser/terminalStickyScrollOverlay.ts
@@ -244,6 +244,12 @@ export class TerminalStickyScrollOverlay extends Disposable {
 			return;
 		}
 
+		// Check if this is a clear command - don't show sticky scroll for clear
+		if (command.command && this._isClearCommand(command.command)) {
+			this._setVisible(false);
+			return;
+		}
+
 		// If the marker doesn't exist or it was trimmed from scrollback
 		const marker = command.marker;
 		if (!marker || marker.line === -1) {
@@ -520,6 +526,20 @@ export class TerminalStickyScrollOverlay extends Disposable {
 			selectionBackground: undefined,
 			selectionInactiveBackground: undefined
 		};
+	}
+
+	private _isClearCommand(command: string): boolean {
+		const trimmedCommand = command.trim().toLowerCase();
+
+		const clearCommands = [
+			'clear',       // Unix/Linux/macOS bash/zsh
+			'cls',         // Windows cmd.exe
+			'clear-host',  // PowerShell
+			'clh',         // PowerShell alias
+			'reset'        // Terminal reset (stronger than clear)
+		];
+
+		return clearCommands.includes(trimmedCommand);
 	}
 }
 
--- a/src/vs/workbench/contrib/terminalContrib/stickyScroll/browser/terminalStickyScrollOverlay.ts
+++ b/src/vs/workbench/contrib/terminalContrib/stickyScroll/browser/terminalStickyScrollOverlay.ts
@@ -244,12 +244,6 @@ export class TerminalStickyScrollOverlay extends Disposable {
 			return;
 		}
 
-		// Do not show sticky scroll after `clear` command
-		if (command.command && this._isClearCommand(command.command)) {
-			this._setVisible(false);
-			return;
-		}
-
 		// If the marker doesn't exist or it was trimmed from scrollback
 		const marker = command.marker;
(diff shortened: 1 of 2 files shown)


anthonykim1:
Additional problem -> I also found out whichever commands that comes right after `clear` displays wrong terminal sticky scroll.
For example, running `git status` after `clear`.
If I run `git status` twice in a row everything is fine.

https://github.com/user-attachments/assets/562cdb1b-a01d-4353-b7af-d9d1b882f889

I saw https://github.com/microsoft/vscode/blob/a8a4f224dc5ef4fc6c2eb38ab8ff140d43cf97af/src/vs/workbench/contrib/terminalContrib/stickyScroll/browser/terminalStickyScrollOverlay.ts#L315-L320 was getting set to content of `git status` command, instead of `git status` command itself:

Screenshot 2025-09-18 at 9 00 29 PM
Screenshot 2025-09-18 at 9 00 45 PM

@Tyriar Does this mean we are counting either `stickyScrollLineStart` or `stickyScrollLineCount` wrong?

meganrogge (dismissed):

src/vs/workbench/contrib/terminalContrib/stickyScroll/browser/terminalStickyScrollOverlay.ts
```diff
@@ -244,6 +244,12 @@ export class TerminalStickyScrollOverlay extends Disposable {
 			return;
 		}
 
+		// Do not show sticky scroll after `clear` command
+		if (command.command && this._isClearCommand(command.command)) {
+			this._setVisible(false);
```
Tyriar: I don't think we should do this, while it may end up working initially it's an edge case that will make it feel inconsistent. For example I suspect scrolling up and down will not be sticky as it should.

This is how it shows for me:

Screenshot 2025-09-26 at 8 06 50 am

The main issue is that this should not be possible where sticky scroll obscures the cursor line (or any part of the current command line).

Screenshot 2025-09-26 at 8 08 14 am

When I run another command, you can see something funny is happening with the `A` sequence which looks like the root cause as to why this is happening.

Screenshot 2025-09-26 at 8 09 23 am

The teal `A` should be at the same spot as the red `D`. Meaning we think that `echo a` starts on the following line and is therefore fine to obscure.

Tyriar (changes requested):

src/vs/workbench/contrib/terminalContrib/stickyScroll/browser/terminalStickyScrollOverlay.ts
Tyriar: Contrast that with how the following command looks:

Screenshot 2025-09-26 at 8 12 03 am

Tyriar: Here's a trace:

Screenshot 2025-09-26 at 8 14 32 am

`A` is in the right spot above, see just before the prompt (and before `B`). So I think when were adjusting the sequences we're messing something up. Look into how the `A` sequence is adjusted, we may need to disable that conditionally depending on if it's a clear command similar to what you do here.

anthonykim1: Yeah the A sequence should've been next to prompt not the output. I now prevent wrong adjustment of A sequence here: [f926735](https://github.com/microsoft/vscode/pull/267439/commits/f9267351c56fa7fcc4d00ec61b42e5d92dcc93fa) 

With the most recent approach,  we still won't show `clear` as sticky scroll but this is consistent with all the commands that preceded `clear`. 
It looks like we also didn't have sticky scroll for commands that preceded `clear` before this PR too.
https://github.com/user-attachments/assets/32cd31f8-3bb7-4eba-b614-044cf5f8dbb5

Is this fine?

Tyriar: That fix looks a bit like a hack as well. I'm investigating this now. What I've found is that https://github.com/microsoft/vscode/commit/6b8bbf8b6f41 (https://github.com/microsoft/vscode/pull/198463 / https://github.com/microsoft/vscode/issues/198278) was done which forces the D onto the next line if it's on the same as a C which isn't necessarily always what we want.

Tyriar: @anthonykim1 I spent a bunch of time trying to write a test but ended up giving up as it's a tricky edge case to cover unfortunately. Anyway, I pushed a change that I think fixes the root cause without more state/comms between hooks and the command detection capability. Could you very on your end and let's try get this in if it works for you.

anthonykim1: Seems working identical to PR!: 
https://github.com/user-attachments/assets/ce14a853-ac88-44f8-be7e-f3db15942798

We still wont show clear and previous commands preceding `clear` tho

Tyriar (approved):

karthiknadig (approved):

PR #267526, opened 2025-09-19
Implements code-no-observable-get-in-reactive-context

hediet:


.eslint-plugin-local/code-no-observable-get-in-reactive-context.ts
```diff
@@ -0,0 +1,212 @@
...
+				message: `Observable '.get()' should not be used in reactive context. Use '.read(${readerName})' instead to properly track dependencies or '.read(undefined)' to be explicit about a untracked read.`,
+				fix: (fixer) => {
+					const memberExpression = node.callee as TSESTree.MemberExpression;
+					return fixer.replaceText(node as any, `${context.getSourceCode().getText(memberExpression.object as any)}.read(undefined)`);
+				}
+			});
+		}
+
+		// Traverse child nodes
+		switch (node.type) {
```
connor4312: You can use eslint-visitor-keys for this, we should already have it as a transitive dependency at minimum https://www.npmjs.com/package/eslint-visitor-keys

connor4312 (dismissed):
overall lgtm but I do suggest swapping out to the library so we don't need to update the logic when we get new language constructs

lszomoru (approved):

alihasnain1920002-bit:
333333L;KLMMMM,

PR #267540, opened 2025-09-19
Add explain action to `DiffHunkWidget`

zhichli:
[Screencast from 09-19-2025 03:22:43 PM.webm](https://github.com/user-attachments/assets/291efac9-b9b9-4743-bd57-92659d02d7cc)


src/vs/workbench/contrib/chat/browser/chatEditing/chatEditingEditorActions.ts
```diff
@@ -322,6 +324,54 @@ class ToggleAccessibleDiffViewAction extends ChatEditingEditorAction {
...
+		const hunkWidget = args[0];
+		let diffRange: Range | undefined;
+
+		// Extract diff range from hunk widget if available
+		if (hunkWidget && typeof hunkWidget === 'object' && hunkWidget !== null) {
+			const widget = hunkWidget as { getStartLineNumber?(): number | undefined; _change?: { modified?: { toInclusiveRange?(): Range | null } } };
+
+			if (typeof widget.getStartLineNumber === 'function') {
+				const startLineNumber = widget.getStartLineNumber();
+				if (startLineNumber && widget._change?.modified?.toInclusiveRange) {
```
hediet, line 354: Why does this check for the presence of `toInclusiveRange`? I think that method is always defined.

src/vs/workbench/contrib/chat/browser/chatEditing/chatEditingEditorActions.ts
```diff
@@ -322,6 +324,54 @@ class ToggleAccessibleDiffViewAction extends ChatEditingEditorAction {
...
+					diffRange = widget._change.modified.toInclusiveRange() ?? undefined;
+				}
+			}
+		}
+
+		const attachFiles: { uri: any; range?: Range }[] = [];
```
hediet, line 360: Please avoid any.

hediet (changes requested):
Nice feature! Left some comments though. Handing over to @jrieken who owns this code area.

PR #267546, opened 2025-09-19
Do not allow localhost redirect in favor of loopback IP redirect

### LINKED ISSUE #267545: Default loopback redirect for enhanced security and compatibility

- VS Code Version: 1.104.1
- OS Version: Win 11 Enterprise 10.0.22631

**Observation/problem:**

Under OAuth 2.0 spec, an MCP (VSCode) client is a public client, as it cannot securely store a client secret. As a non-browser-based application, it aligns with the definition of a native client. Per RFC 8252, Section 8.3 - https://datatracker.ietf.org/doc/html/rfc8252#section-8.3, native clients should use loopback redirects (e.g., http://127.0.0.1:{port}) instead of localhost for enhanced security and compatibility:

`While redirect URIs using localhost (i.e., "http://localhost:{port}/{path}") function similarly to loopback IP redirects described in Section 7.3, the use of localhost is NOT RECOMMENDED.  Specifying a redirect URI with the loopback IP literal rather than localhost avoids inadvertently listening on network interfaces other than the loopback interface.  It is also less susceptible to client-side firewalls and misconfigured host name resolution on the user's device.`

Moreover, Microsoft Identity Platform makes a similar recommendation - https://learn.microsoft.com/en-us/entra/identity-platform/reply-url#prefer-127001-over-localhost - stating that:

`To prevent your app from breaking due to misconfigured firewalls or renamed network interfaces, use the IP literal loopback address 127.0.0.1 in your redirect URI instead of localhost.`

Additionally, some OAuth2 providers may reject "localhost" due to security policies while DNS misconfigurations or overrides in "/etc/hosts" can cause issues.

**Suggestion/enhancement:**

Proposed change is to eliminate default localhost redirects in favor of using loopback IP redirects.

narbit:
Fixes https://github.com/microsoft/vscode/issues/267545

### DIFF
--- a/src/vs/base/common/oauth.ts
+++ b/src/vs/base/common/oauth.ts
@@ -764,13 +764,11 @@ export async function fetchDynamicRegistration(serverMetadata: IAuthorizationSer
 			redirect_uris: [
 				'https://insiders.vscode.dev/redirect',
 				'https://vscode.dev/redirect',
-				'http://localhost/',
 				'http://127.0.0.1/',
 				// Added these for any server that might do
 				// only exact match on the redirect URI even
 				// though the spec says it should not care
 				// about the port.
-				`http://localhost:${DEFAULT_AUTH_FLOW_PORT}/`,
 				`http://127.0.0.1:${DEFAULT_AUTH_FLOW_PORT}/`
 			],
 			scope: scopes?.join(AUTH_SCOPE_SEPARATOR),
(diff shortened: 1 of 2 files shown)


narbit:
Addresses https://github.com/microsoft/vscode/issues/267545

TylerLeonhardt (approved):
LGTM. We only use 127.0.0.1 when we spin up the local server so this makes sense.

roblourens (approved):

Issue #267555, opened 2025-09-20
Use "workbench.editor.splitInGroupLayout" configuration to split editor group when using walkthrough "toSide" commands


harbin1053020115:
There is a configuration which ID="workbench.editor.splitInGroupLayout", but I find that the when clicking the  'toSide' commands in walkthrough start page did not follow the configuration.
So I will push a PR to fixed this.
Final result will be like this, when set "workbench.editor.splitInGroupLayout" to vertical, the editor group will added to the vertical direction:
<img width="3024" height="1572" alt="Image" src="https://github.com/user-attachments/assets/427e0cf4-bd1b-4df6-a286-ae524a83e501" />

PR #267557, opened 2025-09-20
feat: split editor group direction according to workbench.editor.splitInGroupLayout configuration when clicking walkthrough ':toSide' commands 

harbin1053020115:
…plitInGroupLayout configuration

#267555


src/vs/workbench/contrib/welcomeGettingStarted/browser/gettingStarted.ts
```diff
@@ -1243,13 +1244,15 @@ export class GettingStartedPage extends EditorPane {
 		const fullSize = this.groupsService.getPart(this.group).contentDimension;
 		if (!fullSize || fullSize.width <= 700 || this.container.classList.contains('width-constrained') || this.container.classList.contains('width-semi-constrained')) { return; }
 		if (this.groupsService.count === 1) {
-			const sideGroup = this.groupsService.addGroup(this.groupsService.groups[0], GroupDirection.RIGHT);
+			const splitOrientation = this.configurationService.getValue<'vertical' | 'horizontal'>(SideBySideEditor.SIDE_BY_SIDE_LAYOUT_SETTING);
```
bpasero: Thats not the right setting, I suggest to use `preferredSideBySideGroupDirection` utililty method.

bpasero (changes requested):

src/vs/workbench/contrib/welcomeGettingStarted/browser/gettingStarted.ts
harbin1053020115: Okay, already adjusted to preferredSideBySideGroupDirection now.

bpasero (approved):

benibenj (approved):

PR #267583, opened 2025-09-20
test(base): add unit tests for Event.fromObservable and fromObservabl…

SaaiAravindhRaja:
This pull request adds a new test suite for the `Event.fromObservable` and `Event.fromObservableLight` methods, ensuring their correct behavior when converting observables to events. The tests cover initial value emission, subsequent updates, disposal behavior, batch updates, and handling of undefined initial values.

**New tests for observable-to-event conversion:**

* Added `eventObservable.test.ts` with tests verifying that `Event.fromObservable` emits the initial value, subsequent changes, and stops after disposal, and that `Event.fromObservableLight` signals only on changes without payload.
* Included tests to ensure batch updates are coalesced into a single notification and that undefined initial values are handled correctly.

### DIFF
--- a/src/vs/base/test/common/eventObservable.test.ts
+++ b/src/vs/base/test/common/eventObservable.test.ts
@@ -1,0 +1,94 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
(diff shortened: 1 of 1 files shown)


SaaiAravindhRaja:
> Thanks for the PR, though we already have tests for these in event.test.ts/obsevable.test.ts

Sorry about the oversight, thanks for letting me know!

PR #267663, opened 2025-09-21
Fix build script for build VSCode in Nix

xiaoxiangmoe:
* Allow build vscode from tarball without .git directory
* Allow build vscode in macOS's clang without gcc alias
* Add node-gyp as dependency for node-pty's [postinstall script](https://github.com/microsoft/node-pty/blob/24b53bbf573ed7ab35f8897aa8b690dbd9064fc9/scripts/post-install.js#L81)
* Add integrity field for @emmetio/css-parser in extensions/emmet/package-lock.json by change git+ssh to https tarball
* Add buildTime env for deterministic build
* Allows customize the command that replaces `npm install`

### DIFF
--- a/build/azure-pipelines/common/codesign.ts
+++ b/build/azure-pipelines/common/codesign.ts
@@ -4,9 +4,10 @@
  *--------------------------------------------------------------------------------------------*/
 
 import { $, ProcessPromise } from 'zx';
+import { getDate } from '../../lib/util.ts';
 
 export function printBanner(title: string) {
-	title = `${title} (${new Date().toISOString()})`;
+	title = `${title} (${getDate().toISOString()})`;
 
 	console.log('\n');
 	console.log('#'.repeat(75));
--- a/build/azure-pipelines/common/createBuild.ts
+++ b/build/azure-pipelines/common/createBuild.ts
@@ -6,6 +6,7 @@
 import { ClientAssertionCredential } from '@azure/identity';
 import { CosmosClient } from '@azure/cosmos';
 import { retry } from './retry.ts';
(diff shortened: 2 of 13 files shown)


xiaoxiangmoe:
friendly ping @joaomoreno @lszomoru @deepak1556

xiaoxiangmoe:
friendly ping @joaomoreno @lszomoru @deepak1556

xiaoxiangmoe:
friendly ping @joaomoreno @lszomoru @deepak1556

xiaoxiangmoe:
NixOS has high requirements for deterministic builds.
Therefore, the following limitations apply: 

- Network access is generally unavailable. If network access is required, it must be downloaded in advance and a integrity hash must be provided. —— To this end, I added an integrity field to the @emmetio/css-parser dependency.
- NixOS build it without .git directory, just download tallbar, unpack then build it. So I change the postinstall script for check this —— if this dir is not git project, ignore git config command.
- The clang package manager in Unix does not have an alias named gcc for clang, so the build script should follow the requirements of the CC environment variable.
- node-pty called `npx node-gyp` in postinstall script. Because network access is generally unavailable. So I added it in devDependencies to make node-gyp is version deterministic and postinstall script can run without network.
- The new Date() in the build script is also one of the factors that cause unstable builds, so I passed it in through environment variables to eliminate this factor.
- In the current project, there are only two ways to install dependencies, `npm install` and `npm ci`. But in nixos linux, it should be ` npm ci --ignore-scripts && patchShebangs node_modules && npm rebuild && patchShebangs node_modules`. So I added custom npm_install_command env.

PR #267785, opened 2025-09-22
reduce memory by ~1.2 MB

SimonSiefke:
### Overview 

This changes the `toDisposable` function to use a class so that there is one dispose function on the class prototype instead of a dispose closure created for each `toDisposable` function call.

The memory savings seem to be ~1.2 MB.

### Function overview chart

This function overview chart of vscode shows how many function instances exist of each function: 

base

Notably at the top is `createSingleCallFunction`, which is used by `toDisposable`. 

### Heapsnapshot before
before2

### Heapsnapshot after

after

### Notable differences:
- The number of `Functions` is reduced from 117165 to 96305
- The number of `closures (System Context)` is reduced from 55728 to 35040
- The number of `{ dispose }` objects is reduced from 11342 to 1192
- The number of `FunctionDisposables` increases from 0 to 10087
- The memory savings seem to be ~1.2 MB.


src/vs/base/common/lifecycle.ts
```diff
@@ -367,19 +367,33 @@ export function combinedDisposable(...disposables: IDisposable[]): IDisposable {
 	return parent;
 }
 
+class FunctionDisposable implements IDisposable {
+	isDisposed: boolean;
```
mjbvz: [nit] Make these private

mjbvz (dismissed):
Thanks!

TylerLeonhardt (dismissed):

meganrogge (dismissed):

SimonSiefke:
Hm. Looks like it's always failing due to chat errors. I will try to see if I can reproduce the errors locally and maybe solve them.

mjbvz:
Ugh sorry this ended up being annoying and thanks for all the great follow up

I found one very obvious case that could have caused the failure here: https://github.com/microsoft/vscode/pull/268202. Will see if I spot any others. Hopefully we can avoid having to update all 480 locations as many are false positives

SimonSiefke:
Thanks very much! The tests are passing now. It might just have been that one occurrence. I'd say this is ready to merge now - but with a small chance of potentially breaking something.  :)

mjbvz (approved):
Thanks!

lszomoru (approved):

PR #267853, opened 2025-09-22
Don't eagerly attach pasted images in `provideDocumentPasteEdits`

mjbvz:
Fixes #267850

`provideDocumentPasteEdits` should be side effect free


justschen (approved):
thanks!

Issue #267105, opened 2025-09-16
Non existent prompt files crash in recommendations tile settings


eli-w-king:
Does this issue occur when all extensions are disabled?: No

- VS Code Version: 1.105
- OS Version: Mac

Steps to Reproduce:

1. Open Settings (JSON) and add a non existent or misspelled file to chat.promptFilesRecommendations ("no-file": true).
2. Open Chat.
3. 🐛 VS Code freezes and eventually crashes or requires reload/close-and-reopen.

bamurtaugh:
I'm also experiencing this, though my prompt file exists. Anytime I try to `Developer: Reload Window`, my Insiders crashes.

My setting:
```
  "chat.promptFilesRecommendations": {
    "accessibility": true,
  }  
```

File in my repo:

<img width="522" height="87" alt="Image" src="https://github.com/user-attachments/assets/7c8295e0-f9c4-4b49-b505-3219be720a11" />

deepak1556:
[profile.json.gz](https://github.com/user-attachments/files/22395134/profile.json.gz)

symbols https://github.com/electron/electron/releases/download/v37.3.1/electron-v37.3.1-darwin-arm64-symbols.zip

Steps to view the profile:

1) Download the latest release of https://github.com/mstange/samply/releases
2) Extract the symbols folder
3) `samply.exe load profile.json.gz --breakpad-symbol-dir <symbol-path>/breakpad_symbols`

PR #267856, opened 2025-09-22
Refactor welcome view rendering to use a scheduler for debouncing and prevent reentrant calls

bhavyaus:
Fixes:
https://github.com/microsoft/vscode/issues/267105


digitarald (approved):
LGTM, thanks for the fix!

aeschli (approved):

PR #267979, opened 2025-09-23
add header to chat list items

meganrogge:
Fixes https://github.com/microsoft/vscode/issues/261862

An issue with this still is when I use VoiceOver's quick nav (enable with `UpArrow` and `DownArrow` simultaneously), then `H` or `Shift H` to go between headers is that the chat list does not scroll to reveal new headers. 

I have commented on https://github.com/microsoft/vscode/issues/267152 to ask why header navigation is preferrable to our `upArrow/downArrow` list navigation mechanism that does scroll properly.

https://github.com/user-attachments/assets/688abcb6-2603-4d5f-b594-b8f5d698c2ec


meganrogge:
Based on user feedback, this won't solve things

PR #267991, opened 2025-09-23
Copy selection for deleted chunk in inline diff

### LINKED ISSUE #8226: Cannot select text from deleted lines diff view
- VSCode Version: 1.3.0-insider
- OS Version: OS X 10.11

Steps to Reproduce:
1. Open a file with changes in inline diff mode
2. Try to select the red (deleted) lines (You can't)

You can select them normally in the side-by-side. Being able to do so is useful to restore/revert small pieces of code manually by copying and pasting.

Mingpan:
An attempt to fix #8226 

Hi @hediet , can I interest you with (yet another) PR for #8226, a.k.a. inline diff deleted chunk copy selection? :)

The main idea is to calculate the exact line and column for the start & end of the selection in a view zone, and extract the corresponding code chunk using the text model API.

Demo: 
copySelectionInlineDelete

This works with both the menu item, and the browser 'copy' event (Ctrl/Cmd+C).

I have put the setup in a reusable method in a separate file, following the example of [renderLines.ts](https://github.com/microsoft/vscode/blob/main/src/vs/editor/browser/widget/diffEditor/components/diffEditorViewZones/renderLines.ts)

Due to the view zone, there are some hackery involved to get those line and column numbers right (the line number extraction is the same as that from the [inlineDiffDeletedCodeMargin.ts](https://github.com/microsoft/vscode/blob/main/src/vs/editor/browser/widget/diffEditor/components/diffEditorViewZones/inlineDiffDeletedCodeMargin.ts)). Let me know if you see any issues, or if you prefer to solve it in a different way.

Thanks!

### DIFF
--- a/src/vs/editor/browser/widget/diffEditor/style.css
+++ b/src/vs/editor/browser/widget/diffEditor/style.css
@@ -426,3 +426,14 @@
 		margin: 0 4px;
 	}
 }
+
+.monaco-editor .line-delete-selectable {
+	user-select: text !important;
+	-webkit-user-select: text !important;
+	z-index: 1 !important;
+}
+
+.line-delete-selectable .view-line {
+	user-select: text !important;
+	-webkit-user-select: text !important;
+}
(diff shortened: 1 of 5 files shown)


hediet (approved):
Thanks for the PR!
I'm not sure that we should show the context menu automatically, but its still better than before, so I'm happy to get this change in.
Also I think it would be great if if could copy the text of multiple deleted chunks at once, ideally including the unchanged lines in between.

src/vs/editor/browser/widget/diffEditor/style.css
```diff
@@ -426,3 +426,14 @@
...
+.monaco-editor .line-delete-selectable {
+	user-select: text !important;
+	-webkit-user-select: text !important;
+	z-index: 1 !important;
+}
+
+.line-delete-selectable .view-line {
+	user-select: text !important;
+	-webkit-user-select: text !important;
+}
```
Tyriar, line 439: @hediet might want to follow up with consistent selector prefix and/or nesting?

Tyriar (approved):

PR #268017, opened 2025-09-23
add opt-in setting for OS notification when long running (> 1 min) task finishes

meganrogge:
fixes #267459

Screenshot 2025-09-23 at 1 22 23 PM


src/vs/workbench/contrib/tasks/common/tasks.ts
```diff
@@ -1185,6 +1185,13 @@ export interface ITaskProcessEndedEvent extends ITaskCommon {
 	kind: TaskEventKind.ProcessEnded;
 	terminalId: number | undefined;
 	exitCode?: number;
+	durationMs?: number;
+}
+
+export interface ITaskInactiveEvent extends ITaskCommon {
+	kind: TaskEventKind.Inactive;
+	terminalId: number | undefined;
```
dmitrivMS, line 1193: nitpick: why this is typed as "number | undefined" but the line below is "?: number" - could be more consistent...

src/vs/workbench/contrib/tasks/browser/abstractTaskService.ts
```diff
@@ -456,6 +490,57 @@ export abstract class AbstractTaskService extends Disposable implements ITaskSer
...
+
+			disposables.add(this._hostService.onDidChangeFocus(focus => {
+				if (focus) {
+					disposables.dispose();
+				}
+			}));
+		}
+	}
+
+	private _formatTaskDuration(durationMs: number): string {
```
dmitrivMS, line 533: nitpick: this could be static (will make it easier to unit-test, if we added those)

src/vs/workbench/contrib/tasks/browser/abstractTaskService.ts
```diff
@@ -456,6 +490,57 @@ export abstract class AbstractTaskService extends Disposable implements ITaskSer
...
+		if (!terminalForTask) {
+			return;
+		}
+		const taskLabel = terminalForTask.title;
+		const durationText = this._formatTaskDuration(durationMs);
+		const message = taskLabel
+			? nls.localize('task.longRunningTaskCompletedWithLabel', 'Task "{0}" finished in {1}.', taskLabel, durationText)
+			: nls.localize('task.longRunningTaskCompleted', 'Task finished in {0}.', durationText);
+		const targetWindow = dom.getWindow(terminalForTask.domElement);
+		if (targetWindow.document.hasFocus()) {
```
dmitrivMS, line 503: nitpick: this check can probably move higher up to avoid extra calculations if we are going to return anyway

src/vs/workbench/contrib/tasks/browser/abstractTaskService.ts
```diff
@@ -456,6 +490,57 @@ export abstract class AbstractTaskService extends Disposable implements ITaskSer
...
+			return;
+		}
+		this._taskRunStartTimes.delete(event.taskId);
+		this._hostService.focus(targetWindow, { mode: FocusMode.Notify });
+		const notification = await dom.triggerNotification(message);
+		if (notification) {
```
dmitrivMS, line 517: Should notification be registered with the parent store?

src/vs/workbench/contrib/tasks/browser/abstractTaskService.ts
```diff
@@ -384,8 +391,35 @@ export abstract class AbstractTaskService extends Disposable implements ITaskSer
 		this._register(this._lifecycleService.onBeforeShutdown(e => {
 			this._willRestart = e.reason !== ShutdownReason.RELOAD;
 		}));
-		this._register(this.onDidStateChange(e => {
+		this._register(this.onDidStateChange(async e => {
 			this._log(nls.localize('taskEvent', 'Task Event kind: {0}', e.kind), true);
+			switch (e.kind) {
```
dmitrivMS, line 396: I'm not sure about the common practice in the codebase, but this change is adding non-trivial eventing mechanism - is it would adding a couple of unit-tests?

dmitrivMS (changes requested):

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/task/runTaskTool.ts
```diff
@@ -96,6 +98,13 @@ export class RunTaskTool implements IToolImpl {
 		const uniqueDetails = Array.from(new Set(details)).join('\n\n');
 		const toolResultDetails = toolResultDetailsFromResponse(terminalResults);
 		const toolResultMessage = toolResultMessageFromResponse(result, taskLabel, toolResultDetails, terminalResults);
+		const taskFinished = terminalResults.length > 0 && terminalResults.every(r => r.state === OutputMonitorState.Idle);
+		if (taskFinished && !this._configurationService.getValue<boolean>(TaskSettingId.ShowLongRunningTaskCompletionNotification)) {
+			const settingsCommandUri = `command:workbench.action.openSettings?%5B%22@id:${TaskSettingId.ShowLongRunningTaskCompletionNotification}%22%5D`;
```
dmitrivMS: BTW, I believe a change just went in that adds a helper for building command URIs - perhaps use it here?

dmitrivMS: https://github.com/microsoft/vscode/pull/268087

dmitrivMS (dismissed):

src/vs/workbench/contrib/tasks/browser/abstractTaskService.ts
meganrogge: Notification is already registered with the parent store https://github.com/microsoft/vscode/blob/41e48e9f19a1b96e3bf8657e7c7fb8c501476f7f/src/vs/workbench/contrib/tasks/browser/abstractTaskService.ts#L258

dmitrivMS: Ah, thanks, I missed that.

src/vs/workbench/contrib/tasks/browser/abstractTaskService.ts
meganrogge: Good idea, i'll defer to a new pr.

dmitrivMS (approved):

PR #268058, opened 2025-09-23
Show full date/time for history items

### LINKED ISSUE #264665: Chat sessions view should let me see actual date session was created/interacted with
Testing #264256

In addition to showing `X wks` for time, I'd like to see the actual date+time a session was created or interacted with, similar to what the timeline view shows on hover

vijayupadya:
Fixes #264665

For items in history in Local Chat Sessions window, show full datetime on hover over the relative time.

This is the new UI
image

### DIFF
--- a/src/vs/workbench/contrib/chat/browser/chatSessions/view/sessionsTreeRenderer.ts
+++ b/src/vs/workbench/contrib/chat/browser/chatSessions/view/sessionsTreeRenderer.ts
@@ -47,7 +47,7 @@ import { LocalChatSessionsProvider } from '../localChatSessionsProvider.js';
 import { IListRenderer, IListVirtualDelegate } from '../../../../../../base/browser/ui/list/list.js';
 import { ChatSessionTracker } from '../chatSessionTracker.js';
 import { CancellationToken } from '../../../../../../base/common/cancellation.js';
-import { language } from '../../../../../../base/common/platform.js';
+import { getLocalHistoryDateFormatter } from '../../../../localHistory/browser/localHistory.js';
 
 interface ISessionTemplateData {
 	readonly container: HTMLElement;
@@ -340,10 +340,10 @@ export class SessionsRenderer extends Disposable implements ITreeRenderer<IChatS
 
 			// Add tooltip showing full date/time when hovering over the timestamp
 			if (session.timing?.startTime) {
-				const fullDateTime = new Date(session.timing.startTime).toLocaleString(language);
+				const fullDateTime = getLocalHistoryDateFormatter().format(session.timing.startTime);
 				templateData.elementDisposable.add(
 					this.hoverService.setupDelayedHover(templateData.timestamp, {
-						content: fullDateTime
+						content: `Start Time: ${fullDateTime}`
 					})
 				);
 			}
--- a/src/vs/workbench/contrib/chat/browser/chatSessions/view/sessionsTreeRenderer.ts
+++ b/src/vs/workbench/contrib/chat/browser/chatSessions/view/sessionsTreeRenderer.ts
@@ -47,6 +47,7 @@ import { LocalChatSessionsProvider } from '../localChatSessionsProvider.js';
 import { IListRenderer, IListVirtualDelegate } from '../../../../../../base/browser/ui/list/list.js';
 import { ChatSessionTracker } from '../chatSessionTracker.js';
 import { CancellationToken } from '../../../../../../base/common/cancellation.js';
+import { language } from '../../../../../../base/common/platform.js';
 
 interface ISessionTemplateData {
 	readonly container: HTMLElement;
@@ -336,6 +337,16 @@ export class SessionsRenderer extends Disposable implements ITreeRenderer<IChatS
 			templateData.timestamp.ariaLabel = session.relativeTimeFullWord ?? '';
 			templateData.timestamp.parentElement!.classList.toggle('timestamp-duplicate', session.hideRelativeTime === true);
 			templateData.timestamp.parentElement!.style.display = '';
+
+			// Add tooltip showing full date/time when hovering over the timestamp
+			if (session.timing?.startTime) {
+				const fullDateTime = new Date(session.timing.startTime).toLocaleString(language);
+				templateData.elementDisposable.add(
+					this.hoverService.setupDelayedHover(templateData.timestamp, {
+						content: fullDateTime
+					})
+				);
+			}
 		} else {
 			// Hide timestamp container if no timestamp available
 			templateData.timestamp.parentElement!.style.display = 'none';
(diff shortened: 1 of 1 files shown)


src/vs/workbench/contrib/chat/browser/chatSessions/view/sessionsTreeRenderer.ts
```diff
@@ -336,6 +337,16 @@ export class SessionsRenderer extends Disposable implements ITreeRenderer<IChatS
 			templateData.timestamp.ariaLabel = session.relativeTimeFullWord ?? '';
 			templateData.timestamp.parentElement!.classList.toggle('timestamp-duplicate', session.hideRelativeTime === true);
 			templateData.timestamp.parentElement!.style.display = '';
+
+			// Add tooltip showing full date/time when hovering over the timestamp
+			if (session.timing?.startTime) {
+				const fullDateTime = new Date(session.timing.startTime).toLocaleString(language);
```
mjbvz: There's a `getLocalHistoryDateFormatter` helper you can use this to match what the timeline view does

src/vs/workbench/contrib/chat/browser/chatSessions/view/sessionsTreeRenderer.ts
```diff
@@ -336,6 +337,16 @@ export class SessionsRenderer extends Disposable implements ITreeRenderer<IChatS
...
+				templateData.elementDisposable.add(
+					this.hoverService.setupDelayedHover(templateData.timestamp, {
```
mjbvz, line 345: I think this label should also say what exactly this time represents. Is it created time? Updated time?

vijayupadya: Updated

src/vs/workbench/contrib/chat/browser/chatSessions/view/sessionsTreeRenderer.ts
vijayupadya: updated

src/vs/workbench/contrib/chat/browser/chatSessions/view/sessionsTreeRenderer.ts
```diff
@@ -336,6 +337,16 @@ export class SessionsRenderer extends Disposable implements ITreeRenderer<IChatS
 			templateData.timestamp.ariaLabel = session.relativeTimeFullWord ?? '';
 			templateData.timestamp.parentElement!.classList.toggle('timestamp-duplicate', session.hideRelativeTime === true);
 			templateData.timestamp.parentElement!.style.display = '';
+
+			// Add tooltip showing full date/time when hovering over the timestamp
+			if (session.timing?.startTime) {
+				const fullDateTime = getLocalHistoryDateFormatter().format(session.timing.startTime);
+				templateData.elementDisposable.add(
+					this.hoverService.setupDelayedHover(templateData.timestamp, {
+						content: `Start Time: ${fullDateTime}`
```
mjbvz: Make sure this is localized too

mjbvz: Also is this really the start time? I just tried adding to an old conversation and it seems like it got shifted towards the top of the list

vijayupadya: updated

vijayupadya: The startTime value itself seem to show the start time of the chat, for the test sessions I tested locally.
Are you seeing items in history getting resorted due to this?

mjbvz: So yes right now you are always showing the start time. But in the chat view, the timestamp seems to be updated whenever I add a new message to the chat. Before:

Screenshot 2025-09-23 at 4 02 36 PM

After adding a new message to these:

Screenshot 2025-09-23 at 4 03 25 PM

So it's confusing if I hover over `now` and only then  see the start time since that's not actually what the UI is showing

vijayupadya: lastMessageDate is currently being used as 'startTime' which is used to show in the current UI ('now', etc.). Its not absolute startTime when chat was first created.  Updated the tooltip label to "Last Activity" which is more appropriate.

src/vs/workbench/contrib/chat/browser/chatSessions/view/sessionsTreeRenderer.ts
```diff
@@ -336,6 +337,16 @@ export class SessionsRenderer extends Disposable implements ITreeRenderer<IChatS
 			templateData.timestamp.ariaLabel = session.relativeTimeFullWord ?? '';
 			templateData.timestamp.parentElement!.classList.toggle('timestamp-duplicate', session.hideRelativeTime === true);
 			templateData.timestamp.parentElement!.style.display = '';
+
+			// Add tooltip showing full date/time when hovering over the timestamp
+			if (session.timing?.startTime) {
+				const fullDateTime = getLocalHistoryDateFormatter().format(session.timing.startTime);
+				templateData.elementDisposable.add(
+					this.hoverService.setupDelayedHover(templateData.timestamp, {
+						content: `${nls.localize('chat.sessions.lastActivity', 'Last Activity')}: ${fullDateTime}`
```
mjbvz: You can use placeholders in the `nls.localize` call to make this a bit more clear (and potentially be more correctly localizable across languages). These are written using `{0}` with the arguments provided after the string. For example:


```suggestion
						content: nls.localize('chat.sessions.lastActivity', 'Last Activity: {0}', fullDateTime)
```

vijayupadya: updated

src/vs/workbench/contrib/chat/browser/chatSessions/view/sessionsTreeRenderer.ts
```diff
@@ -336,6 +337,16 @@ export class SessionsRenderer extends Disposable implements ITreeRenderer<IChatS
 			templateData.timestamp.ariaLabel = session.relativeTimeFullWord ?? '';
 			templateData.timestamp.parentElement!.classList.toggle('timestamp-duplicate', session.hideRelativeTime === true);
 			templateData.timestamp.parentElement!.style.display = '';
+
+			// Add tooltip showing full date/time when hovering over the timestamp
+			if (session.timing?.startTime) {
+				const fullDateTime = getLocalHistoryDateFormatter().format(session.timing.startTime);
+				templateData.elementDisposable.add(
+					this.hoverService.setupDelayedHover(templateData.timestamp, {
+						content: `${nls.localize('chat.sessions.lastActivity', 'Last Activity: {0}', fullDateTime)}`
```
mjbvz: I don't think this needs to be wrapped in a string. We can use the call to `localize` directly

mjbvz (changes requested):

src/vs/workbench/contrib/chat/browser/chatSessions/view/sessionsTreeRenderer.ts
vijayupadya: Updated.

mjbvz (approved):

Issue #267860, opened 2025-09-22
Chat empty state history preview: Remove secondary toolbar with history title and icon


eli-w-king:
1. Remove secondary toolbar with history title and icon from `chat.emptyState.history.enabled`.
2. Add 'Chat history..." link at end of list to open workspace history in command palette (replacing history icon functionality). 
<img width="1054" height="374" alt="Image" src="https://github.com/user-attachments/assets/ae93eb60-aea5-4b70-8d18-f625cf76554e" />

Issue #267522, opened 2025-09-19
History is not hidden when not required


DonJayamanne:
There doesn't seem to be a way to turn off this history feature
I'm referring to the 3 chat history items displayed on the top of the chat UI.

<img width="852" height="479" alt="Image" src="https://github.com/user-attachments/assets/0a15f3c6-8eb1-48cb-8a72-7f9cb16c6a49" />

DonJayamanne:
Still repros in latest insiders
From what I can tell I need to re-load VS Code for the changes to take into effect.

Would be nice if I didn't have to reload vs code, as we generally try to avoid requiring re-loads to VS Code for changes to settings or the like.

bpasero:
It is now `chat.recentSessions.enabled`

Issue #267430, opened 2025-09-19
no padding in the chat disclaimer welcome message when it has no chat history


DioLps:
I notice that in recent versions of copilot/ vscode, the chat padding for the disclaimer welcome message seems to have no padding.

<img width="401" height="576" alt="Image" src="https://github.com/user-attachments/assets/6468e492-2b05-4f04-ae61-7eea13831ba6" />

Created the pr here: #267054,  can I get someone to review it please?

Issue #266447, opened 2025-09-12
Add right click menu to chat history


eli-w-king:
To address the noise feedback, we can include a right click menu to relocate or hide chat history (considering the same for tiles). We could get ahead similar feedback once these setting are enabled by default for all users.

<img width="720" height="431" alt="Image" src="https://github.com/user-attachments/assets/77b00835-3979-41c0-8002-3b402b7af62b" />

eli-w-king:
<img width="243" height="150" alt="Image" src="https://github.com/user-attachments/assets/9b357ba9-6676-4a5a-83b5-21628d8d44f8" />

Right click menu added.

PR #268080, opened 2025-09-24
Empty state chat history 2.0

eli-w-king:
Redesign of history in chat empty state:
1. Adds right click menu to configure empty state settings and hide chat history (plan is to eventually set setting as default and remove, future proofing hide/show) https://github.com/microsoft/vscode/issues/267522, https://github.com/microsoft/vscode/issues/266447
2. Fixes overlapping disclaimer text and inconsistent padding https://github.com/microsoft/vscode/issues/267430, https://github.com/microsoft/vscode/issues/265071
3. Fixes hover state overlap with history entries https://github.com/microsoft/vscode/issues/266373
4. Removes secondary toolbar with history title and icon, replaces with link to open workspace chat history https://github.com/microsoft/vscode/issues/267860

https://github.com/user-attachments/assets/52a59735-130e-4208-bfb9-b2b946ca97c5

_Closing individual PRs in favor of this one. CC: @hawkticehurst @justschen @mrleemurray_


eli-w-king:
Removing hover for now, doesn't provide value when chat titles are displayed in full. For next phase, consider using rich hover:
> Here is an idea: what if we made the hover really show more info, like how many files changed, which model, the duration, the token use? That might also make it easier to relate to the chat session?

src/vs/workbench/contrib/chat/browser/actions/chatActions.ts
```diff
@@ -1989,3 +1989,69 @@ registerAction2(class EditToolApproval extends Action2 {
...
+				precondition: ChatContextKeys.enabled
+			});
+		}
+
+		async run(accessor: ServicesAccessor): Promise<void> {
+			const chatWidgetService = accessor.get(IChatWidgetService);
+			const widgets = chatWidgetService.getWidgetsByLocations(ChatAgentLocation.Chat);
+			const widget = widgets?.[0];
+			if (widget && 'toggleHistoryVisibility' in widget) {
+				(widget as any).toggleHistoryVisibility();
```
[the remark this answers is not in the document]
justschen: agree with copilot here. you can add it here: https://github.com/microsoft/vscode/blob/d47a6cec13028f3bdad1ac67f62e87df802fb3f5/src/vs/workbench/contrib/chat/browser/chat.ts#L257

src/vs/workbench/contrib/chat/browser/media/chatViewWelcome.css
```diff
@@ -30,6 +30,8 @@
 		align-items: center;
 		justify-content: center;
 		overflow: hidden;
+		container-type: inline-size;
+		container-name: chat-welcome;
```
justschen: ik container-name and `@container` is relatively new in css, so just for personal prefernce, let's maybe keep it simpler for now and not use it. 

`@media` is probably enough for what you want to do here, or maybe just giving it enough padding left/right might also be enough

src/vs/workbench/contrib/chat/browser/media/chatViewWelcome.css
```diff
@@ -324,4 +332,36 @@ div.chat-welcome-view {
 			margin-left: 8px;
 		}
 	}
+
+	.chat-welcome-history-more {
+		color: var(--vscode-textLink-foreground);
+		text-decoration: none;
+		cursor: pointer;
+		padding: 2px 12px 4px 12px;
+		font-size: 13px;
```
justschen: nit: this font size is already inherited from `.interactive-session`

src/vs/workbench/contrib/chat/browser/chatWidget.ts
```diff
@@ -852,6 +898,21 @@ export class ChatWidget extends Disposable implements IChatWidget {
...
+		this._historyVisible = !this._historyVisible;
+		// Find and hide/show the existing history section
+		const historyRoot = this.welcomeMessageContainer.querySelector('.chat-welcome-history-root') as HTMLElement;
+		if (historyRoot) {
+			if (this._historyVisible) {
+				historyRoot.style.removeProperty('display');
+				this.welcomeMessageContainer.classList.add('has-chat-history');
+			} else {
+				historyRoot.style.display = 'none';
+				this.welcomeMessageContainer.classList.remove('has-chat-history');
```
justschen: can we please use css + `classList.toggle` here. set the class list to be hidden or something, then in css set the style to be `display:none`

src/vs/workbench/contrib/chat/browser/chatWidget.ts
```diff
@@ -996,12 +1059,27 @@ export class ChatWidget extends Disposable implements IChatWidget {
...
+						getActionsContext: () => ({})
+					});
+				}));
 			}
 		}
 
 		this.updateChatViewVisibility();
 
-		if (numItems > 0) {
+		if (numItems === 0) {
```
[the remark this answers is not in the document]
justschen: i think something changed in the chat widget refresh due to the debouncing changes from @bhavyaus 

this and the other related copilot comment are probably okay

src/vs/workbench/contrib/chat/browser/chatWidget.ts
```diff
@@ -1040,12 +1110,22 @@ export class ChatWidget extends Disposable implements IChatWidget {
 			todayMidnight.setHours(0, 0, 0, 0);
 			const todayMidnightMs = todayMidnight.getTime();
 
+			// Create hover delegate for proper tooltip positioning
+			const getViewContainerLocation = () => {
+				const panelLocation = this.contextKeyService.getContextKeyValue<ViewContainerLocation>('chatPanelLocation');
```
justschen, line 1107: this is kinda a code smell iirc to get the context key like this.

but also, is there a reason we want to great a custom hover delegate? were there issues with the `defaultHoverDelegate` (you can get using `getDefaultHoverDelegate('element')`)

justschen: also for now, if you want to just get this merged so we can have some bug fixes to verify, let's just remove the hover like we discussed in ux sync.

src/vs/workbench/contrib/chat/browser/chatWidget.ts
```diff
@@ -162,13 +168,44 @@ interface IChatHistoryTemplate {
 	disposables: DisposableStore;
 }
 
+class ChatHistoryHoverDelegate extends WorkbenchHoverDelegate {
```
justschen, line 170: going back to previous comment below, let's remove hover work for this PR.

justschen (changes requested):

eli-w-king:
Moved prompt tile styling to separate branch.

justschen (approved):
i think this is fine for now, thanks 👍🏻 it is overall looking a lot better on the UI side.

there are a few things i really want to clean up here, such as the delegate (i think it is maybe little over cooked having whole separate delegate _just_ so that it renders nicely for one single hover, but maybe this will change with richer hovers here)

please be sure to test thoroughly during verification!!!

src/vs/workbench/contrib/chat/browser/actions/chatActions.ts
```diff
@@ -2027,3 +2027,69 @@ registerAction2(class EditToolApproval extends Action2 {
 		}
 	}
 });
+
+// Register actions for chat welcome history context menu
+MenuRegistry.appendMenuItem(MenuId.ChatWelcomeHistoryContext, {
+	command: {
+		id: 'workbench.action.chat.toggleChatHistoryVisibility',
+		title: localize('chat.showChatHistory.label', "✓ Chat History")
```
bpasero, line 2035: @eli-w-king this is not how we indicate a menu item that is set! We have a `toggled` condition for that on the menu. I will fix it.

src/vs/workbench/contrib/chat/browser/actions/chatActions.ts
```diff
@@ -2027,3 +2027,69 @@ registerAction2(class EditToolApproval extends Action2 {
...
+			precondition: ChatContextKeys.enabled
+		});
+	}
+
+	async run(accessor: ServicesAccessor): Promise<void> {
+		const chatWidgetService = accessor.get(IChatWidgetService);
+		const widgets = chatWidgetService.getWidgetsByLocations(ChatAgentLocation.Chat);
+		const widget = widgets?.[0];
+		if (widget) {
+			widget.toggleHistoryVisibility();
```
bpasero, line 2067: @eli-w-king this seems to toggle history only for the session but does not store in settings, whats the motivation behind that?

PR #268081, opened 2025-09-24
Claude with worktree

osortega:


src/vs/workbench/contrib/chat/browser/chatSessions.contribution.ts
```diff
@@ -265,7 +278,7 @@ export class ChatSessionsService extends Disposable implements IChatSessionsServ
 				});
 			}
 
-			async run(accessor: ServicesAccessor) {
+			async run(accessor: ServicesAccessor, args?: { workingDirectory?: string }) {
```
rebornix, line 281: should this be `args?: Record<string, any>;`?

PR #268087, opened 2025-09-24
Add `createCommandUri` helper

mjbvz:
This provides a consistent way of creating command uris

### DIFF
--- a/src/vs/base/common/htmlContent.ts
+++ b/src/vs/base/common/htmlContent.ts
@@ -200,11 +200,14 @@ export function parseHrefAndDimensions(href: string): { href: string; dimensions
 }
 
 export function markdownCommandLink(command: { title: string; id: string; arguments?: unknown[] }, escapeTokens = true): string {
-	const uri = URI.from({
-		scheme: Schemas.command,
-		path: command.id,
-		query: command.arguments?.length ? encodeURIComponent(JSON.stringify(command.arguments)) : undefined,
-	}).toString();
-
+	const uri = createCommandUri(command.id, ...(command.arguments || [])).toString();
 	return `[${escapeTokens ? escapeMarkdownSyntaxTokens(command.title) : command.title}](${uri})`;
+}
+
+export function createCommandUri(commandId: string, ...args: unknown[]): URI {
+	return URI.from({
+		scheme: Schemas.command,
+		path: commandId,
+		query: args.length ? encodeURIComponent(JSON.stringify(args)) : undefined,
+	});
 }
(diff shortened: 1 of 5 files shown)


src/vs/base/common/htmlContent.ts
```diff
@@ -200,11 +200,14 @@ export function parseHrefAndDimensions(href: string): { href: string; dimensions
...
-		scheme: Schemas.command,
-		path: command.id,
-		query: command.arguments?.length ? encodeURIComponent(JSON.stringify(command.arguments)) : undefined,
-	}).toString();
-
+	const uri = createCommandUri(command.id, ...(command.arguments || [])).toString();
 	return `[${escapeTokens ? escapeMarkdownSyntaxTokens(command.title) : command.title}](${uri})`;
 }
+
+export function createCommandUri(commandId: string, ...args: unknown[]): URI {
```
dmitrivMS, line 207: Is it possible to add some unit-tests (and perhaps a JSDoc comment) for this?

dmitrivMS, line 207: I am still new to the codebase, but "htmlContent" feels like a strange name/place for something that is command-related helper to be used elsewhere?

dmitrivMS (changes requested):
There appears to be a few more places where the new helper can be used:
https://github.com/microsoft/vscode/blob/bf7d40db74dba2445c259f3b4859db41eaa4dc50/extensions/git/src/timelineProvider.ts#L97

https://github.com/microsoft/vscode/blob/bf7d40db74dba2445c259f3b4859db41eaa4dc50/src/vs/workbench/contrib/files/browser/explorerViewlet.ts#L279

https://github.com/microsoft/vscode/blob/bf7d40db74dba2445c259f3b4859db41eaa4dc50/extensions/npm/src/scriptHover.ts#L106

Should any of these be updated as well?

mjbvz:
I'm intentionally not doing all of them at once as it will make the PRs difficult to review. Some instances need careful review as they involve a bunch of highly nested parens/strings

src/vs/base/common/htmlContent.ts
mjbvz: Our codebase is generally pretty jsdoc light (probably too light tbh) but in this case I don't think one would add much

Test maybe, but in this case it feels like adding it  just to say I have one. Not in favor of doing this unless there's something worth testing

mjbvz: I'd like to keep it next to the other helper for now. This function will be used mostly html/markdown

dmitrivMS: I don't mind not adding jsdoc, but not adding test for something that's reused across codebase feels like a mistake. It's your call though.

dmitrivMS: Again, your call.

dmitrivMS:
> I'm intentionally not doing all of them at once as it will make the PRs difficult to review. Some instances need careful review as they involve a bunch of highly nested parens/strings

Would adding tests help get more confidence in the changes?

dmitrivMS (approved):

PR #268090, opened 2025-09-24
add opt-in setting to show OS notification when a chat response is received

### LINKED ISSUE #259683: Support a notification when chat response has finished
## Description

VS Code has `chat.notifyWindowOnConfirmation` for when Copilot Chat needs user confirmation, but no equivalent for when agent mode finishes executing multi-step tasks. Agent mode is most valuable when users are able to be productive in other ways while it is executing. However, value is lost if users must manually keep checking for execution completion during long-running agent sessions.

## Proposed Solution

Add a setting called `chat.notifyWindowOnCompletion` (boolean, default `false`) (similar to `chat.notifyWindowOnConfirmation`) to show OS notifications when agent/edit mode execution finishes.

## Expected Behavior

-  Desktop notification when agent/edit mode completes (successfully or with errors)
-  The notification should include a basic status ("Copilot completed" or "Copilot finished with errors")
-  Clicking the notification brings VS Code to the foreground with focus in the chat window

## Use Case

Long-running agent sessions, e.g. with tool usage and multi-file changes.

## Related Issues

-  Similar to existing `chat.notifyWindowOnConfirmation` setting
-  #141506
-  #251621

meganrogge:
fixes https://github.com/microsoft/vscode/issues/259683

This is particularly useful because our chat response received sound doesn't always play due to chromium autoplay media restrictions. We should consider enabling this by default. 

https://github.com/user-attachments/assets/50fa25e5-8790-4aaa-b957-40eed1607d49


src/vs/workbench/contrib/chat/browser/chat.contribution.ts
```diff
@@ -311,6 +311,11 @@ configurationRegistry.registerConfiguration({
 			description: nls.localize('chat.emptyState.history.enabled', "Show recent chat history on the empty chat state."),
 			tags: ['experimental']
 		},
+		[ChatConfiguration.ShowResponseCompletionOSNotification]: {
+			type: 'boolean',
+			default: false,
+			description: nls.localize('chat.showResponseCompletionNotification', "Show an OS notification when a chat response is received and the window is not focused."),
```
bpasero: We already have a setting `chat.notifyWindowOnConfirmation`, why is this setting so much different? Ideally we align the setting names, wouldn't you agree?

src/vs/workbench/contrib/chat/browser/chatAccessibilityService.ts
```diff
@@ -58,4 +67,30 @@ export class ChatAccessibilityService extends Disposable implements IChatAccessi
...
+	private async _showOSNotification(container: HTMLElement): Promise<void> {
+		if (!this._configurationService.getValue(ChatConfiguration.ShowResponseCompletionOSNotification)) {
+			return;
+		}
+		const targetWindow = dom.getWindow(container);
+		if (targetWindow.document.hasFocus()) {
+			return;
+		}
+		this._hostService.focus(targetWindow, { mode: FocusMode.Notify });
+		const notification = await dom.triggerNotification('Chat response received');
```
bpasero: You are missing `nls.localize`?

bpasero (dismissed):

bpasero:
Enabling this by default sounds OK to me, but would like to get some selfhosting in insiders before making a final decision.

src/vs/workbench/contrib/chat/browser/chatAccessibilityService.ts
```diff
@@ -25,10 +32,14 @@ export class ChatAccessibilityService extends Disposable implements IChatAccessi
 
 	private _requestId: number = 0;
 
+	private readonly notifications: Set<DisposableStore> = new Set();
```
roblourens, line 35: I thought maybe we had a `DisposableSet`, but we only have `DisposableMap`. `DisposableSet` would be a very helpful utility for this kind of scenario though to avoid having to handle disposing and removing items manually!

roblourens (approved):

bpasero (approved):

src/vs/workbench/contrib/chat/browser/chatAccessibilityService.ts
roblourens: And the items in `this.notifications` should be disposed when the service is disposed, for everything to be tracked properly

Issue #107183, opened 2020-09-21
Add a new right-aligned "description" property to TreeItems


eamodio:
Similar to the timestamp in the Timeline view, it would be nice to have the ability to have a string that would be rendered right-aligned for each tree item.

Maybe something like:

```ts
export class TreeItem {
    label?: string;
    description?: string | boolean;
    ...
    endDescription?: string; // or endDetails? or even just details, but that feels wrong given the other usages of details
}
```

eamodio:
@misolori This could also be used for a checkmark for pr reviews

alexr00:
We should first discuss this in the UI sync.

The file decoration API is also being finalized soon. We can you that for a checkmark for PR reviews too.

alexr00:
Current:
![image](https://user-images.githubusercontent.com/38270282/95466055-3255a380-097c-11eb-912c-057d2cc2ca8a.png)

Example where we hide the right-aligned description on hover when the action is shown:
![recording (23)](https://user-images.githubusercontent.com/38270282/95467460-c6743a80-097d-11eb-9701-7b6bfb532bfd.gif)

Example in the panel:
![image](https://user-images.githubusercontent.com/38270282/95467941-50240800-097e-11eb-89a3-9c1770e0f512.png)

bpasero:
Seems like a fair request to me, but happy to discuss in a UX call.

misolori:
@bpasero we actually did discuss this in last week's ux sync, seemed like everyone was in agreement of this but the specific detail of whether we hide the description label was under discussion. I personally like hiding the description on hover as that is secondary information and actions have a higher priority.

alexr00:
Since we are trying not to add new proposed API in October, moving to November.

tonka3000:
@alexr00 Any  updates on this proposal? Would also love to have this in my vscode extension. I would also hide it by default, but you can add a second property to control the behavior (auto-hide, show, show-fully).

DaelonSuzuka:
This would be helpful for me. I'm the developer of [Godot Tools](https://github.com/godotengine/godot-vscode-plugin), that provides support for the [Godot Game Engine](https://godotengine.org/). 

One of my features is a TreeView that shows a preview of an open scene file:

![image](https://github.com/microsoft/vscode/assets/18042232/79351cf6-d2c9-4335-b482-ade3edf3de06)

[...]

PR #268093, opened 2025-09-24
Feature/107183 add ability to right align tree items description

trapgar:
Issue #107183: Add a new right-aligned "description" property to TreeItems

Added the ability for VS Code extensions utilizing the [`TreeView`](https://code.visualstudio.com/api/extension-guides/tree-view) to horizontally align the `description` property of `TreeItems` to the right. Default value remains left-aligned.

## How to Test
1. Clone the [vscode-extension-samples](https://github.com/microsoft/vscode-extension-samples) project
2. Open the [`tree-view-sample`](https://github.com/microsoft/vscode-extension-samples/tree/main/tree-view-sample) in Code
3. Open [`src/nodeDependencies.ts#`](https://github.com/microsoft/vscode-extension-samples/blob/main/tree-view-sample/src/nodeDependencies.ts#L98)
4. Add a new line right below where the `TreeItem.description` is being set
    > this.alignment = 'right';`

    *(since the repo references a released version of the vscode api you will want to also add an `// @ts-ignore` right about it just for testing)*
5. Run the extension as usual
6. Open any repo utilizing npm (I just created a new folder and ran `> npm i three`
7. Click on the new list clover in the left-hand sidebar that the extension added
8. Expand the tree and all the package versions will now be aligned to the right

Since `TreeItem` actions appear to the right, the description will hide any-time they show as to not clash.


alexr00 (commented):
Thank you for the PR! We're about to start our endgame. I'll take a close look at this PR once we branch for release at the end of next week.

src/vscode-dts/vscode.d.ts
```diff
@@ -12257,6 +12273,13 @@ declare module 'vscode' {
 		 */
 		description?: string | boolean;
 
+		/**
+		 * The horizontal alignment of the {@link TreeItem.description description} text.
+		 * When `right`, the text with be right-aligned and when `left`, it is left-aligned.
+		 * Defaults to `Left`
+		 */
+		alignment?: TreeItemAlignment;
```
alexr00, line 12281: I think we want to allow extension authors to set a description (left aligned) and a _second description_ (right aligned) instead of just the one description which we allow extension authors to set the alignment of. Not sure about a good name for this "second description". Maybe `detail`?

trapgar: Ah, understood. I'm assuming the styling (apart from the alignment) should mimic that of the description? e.g.: font/size/color/etc.

trapgar: @alexr00 Hi, can I get clarification on how we want this to function?

image

In this image, you can see that long text has the potential to hide or clash with the left-aligned text (all the TreeItems in the picture have both `description` and `detail` set, the `label` and `description` is just hidden when the right-aligned text overlaps).

In terms of styling, I also assume we want a different colouring/etc?

PR #268100, opened 2025-09-24
More `createCommandUri` adoption

mjbvz:
Follow up on #268087

Converting over a few more callers. This makes sure we're creating command Uris in a consistent way and also makes it easier to track where they are used

### DIFF
--- a/src/vs/platform/mcp/common/allowedMcpServersService.ts
+++ b/src/vs/platform/mcp/common/allowedMcpServersService.ts
@@ -4,9 +4,8 @@
  *--------------------------------------------------------------------------------------------*/
 
 import { Disposable } from '../../../base/common/lifecycle.js';
-import { URI } from '../../../base/common/uri.js';
 import * as nls from '../../../nls.js';
-import { IMarkdownString, MarkdownString } from '../../../base/common/htmlContent.js';
+import { createCommandUri, IMarkdownString, MarkdownString } from '../../../base/common/htmlContent.js';
 import { IConfigurationService } from '../../configuration/common/configuration.js';
 import { Emitter } from '../../../base/common/event.js';
 import { IAllowedMcpServersService, IGalleryMcpServer, IInstallableMcpServer, ILocalMcpServer, mcpAccessConfig, McpAccessValue } from './mcpManagement.js';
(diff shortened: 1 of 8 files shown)


src/vs/workbench/contrib/chat/browser/chatParticipant.contribution.ts
```diff
@@ -354,7 +354,7 @@ export class ChatCompatibilityNotifier extends Disposable implements IWorkbenchC
 		this.registeredWelcomeView = true;
 		const showExtensionLabel = localize('showExtension', "Show Extension");
 		const mainMessage = localize('chatFailErrorMessage', "Chat failed to load because the installed version of the Copilot Chat extension is not compatible with this version of {0}. Please ensure that the Copilot Chat extension is up to date.", this.productService.nameLong);
-		const commandButton = `[${showExtensionLabel}](command:${showExtensionsWithIdsCommandId}?${encodeURIComponent(JSON.stringify([[this.productService.defaultChatAgent?.chatExtensionId]]))})`;
+		const commandButton = `[${showExtensionLabel}](${createCommandUri(showExtensionsWithIdsCommandId, [this.productService.defaultChatAgent?.chatExtensionId])})`;
```
mjbvz, line 357: This arg is intentionally an array since the original values passed to stringify was `T[][]`.

src/vs/platform/mcp/common/allowedMcpServersService.ts
```diff
@@ -34,7 +33,7 @@ export class AllowedMcpServersService extends Disposable implements IAllowedMcpS
 			return true;
 		}
 
-		const settingsCommandLink = URI.parse(`command:workbench.action.openSettings?${encodeURIComponent(JSON.stringify({ query: `@id:${mcpAccessConfig}` }))}`).toString();
+		const settingsCommandLink = createCommandUri('workbench.action.openSettings', { query: `@id:${mcpAccessConfig}` }).toString();
```
dmitrivMS, line 36: In cases where we pass and object { ... } into commandArgs, doesn't it change the resulting JSON?
I.e. before it would only format the object itself, but now it becomes an array [{ ... }] - if so, would that be a problem?

dmitrivMS (approved):

src/vs/platform/mcp/common/allowedMcpServersService.ts
mjbvz: yes it does, but it will behave the same when run: https://github.com/microsoft/vscode/blob/61ceea4bb2e7afdc763bfcbcd0936bba49ba2dda/src/vs/editor/browser/services/openerService.ts#L60

The most correct way is to always use an array for the arguments. That's what `createCommandUri ` will do. But we do allow passing the first argument directly (so long as the first argument is not an array)

dmitrivMS: Got it, thanks for the explanation!
I still think adding some of this info as a comment for the new function would be helpful :-)

PR #268195, opened 2025-09-24
add PromptDocumentSemanticTokensProvider

aeschli:
For https://github.com/microsoft/vscode/issues/262159

### DIFF
--- a/src/vs/workbench/contrib/chat/common/promptSyntax/languageProviders/promptDocumentSemanticTokensProvider.ts
+++ b/src/vs/workbench/contrib/chat/common/promptSyntax/languageProviders/promptDocumentSemanticTokensProvider.ts
@@ -1,0 +1,82 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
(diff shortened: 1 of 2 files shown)


src/vs/workbench/contrib/chat/common/promptSyntax/languageProviders/promptDocumentSemanticTokensProvider.ts
```diff
@@ -0,0 +1,82 @@
...
+			if (token.isCancellationRequested) {
+				break; // Return what we have so far if cancelled.
+			}
+		}
+
+		return { data: new Uint32Array(data) };
+	}
+
+	getLegend(): SemanticTokensLegend {
+		return { tokenTypes: ['variable'], tokenModifiers: [] };
```
dmitrivMS, line 76: Should 'variable' be an extern const somewhere so consumers can use it?

dmitrivMS (approved):

src/vs/workbench/contrib/chat/common/promptSyntax/languageProviders/promptDocumentSemanticTokensProvider.ts
aeschli: We currently have no enum of the token types in our core. 
This is the only place in core where we implement a PromptDocumentSemanticTokensProvider. We normally do that in extensions.
At some point I'll move that code in an extension and then we can use the enum provided by the languageserver-types library

PR #268215, opened 2025-09-24
improve notification for chat - make it location agnostic

meganrogge:
cc @bpasero

### DIFF
--- a/src/vs/workbench/contrib/chat/browser/chatAccessibilityService.ts
+++ b/src/vs/workbench/contrib/chat/browser/chatAccessibilityService.ts
@@ -8,7 +8,7 @@ import { Disposable, DisposableMap, DisposableStore } from '../../../../base/com
 import { AccessibilitySignal, IAccessibilitySignalService } from '../../../../platform/accessibilitySignal/browser/accessibilitySignalService.js';
 import { IInstantiationService } from '../../../../platform/instantiation/common/instantiation.js';
 import { AccessibilityProgressSignalScheduler } from '../../../../platform/accessibilitySignal/browser/progressAccessibilitySignalScheduler.js';
-import { IChatAccessibilityService } from './chat.js';
+import { IChatAccessibilityService, IChatWidgetService } from './chat.js';
 import { IChatResponseViewModel } from '../common/chatViewModel.js';
 import { renderAsPlaintext } from '../../../../base/browser/markdownRenderer.js';
 import { MarkdownString } from '../../../../base/common/htmlContent.js';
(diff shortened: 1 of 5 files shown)


src/vs/workbench/contrib/chat/browser/chatAccessibilityService.ts
```diff
@@ -109,7 +108,8 @@ export class ChatAccessibilityService extends Disposable implements IChatAccessi
 
 		disposables.add(Event.once(notification.onClick)(async () => {
 			await this._hostService.focus(targetWindow, { mode: FocusMode.Force });
-			await this._commandService.executeCommand('workbench.panel.chat.view.copilot.focus');
+			const lastWidget = this._chatWidgetService.lastFocusedWidget;
```
roblourens: Seems like lastFocusedWidget isn't necessarily the same one that called acceptResponse

meganrogge: In most cases it will be though? I don't think it's worth tracking this more specifically

meganrogge: actually can maybe just pass the widget it. trying

roblourens: It would usually be fine but tracking the actual widget seems easy too. But if it turn out to be not (you have a circular dependency here essentially) then you can do it the other way.

meganrogge: it is easy

roblourens (dismissed):

src/vs/workbench/contrib/chat/browser/chatAccessibilityService.ts
```diff
@@ -49,15 +48,15 @@ export class ChatAccessibilityService extends Disposable implements IChatAccessi
 		this._pendingSignalMap.set(this._requestId, this._instantiationService.createInstance(AccessibilityProgressSignalScheduler, CHAT_RESPONSE_PENDING_ALLOWANCE_MS, undefined));
 		return this._requestId;
 	}
-	acceptResponse(container: HTMLElement, response: IChatResponseViewModel | string | undefined, requestId: number, isVoiceInput?: boolean): void {
+	acceptResponse(widget: ChatWidget, container: HTMLElement, response: IChatResponseViewModel | string | undefined, requestId: number, isVoiceInput?: boolean): void {
```
roblourens, line 51: shouldn't chatWidget.ts be in here somewhere?

meganrogge: https://github.com/microsoft/vscode/blob/4c6fffb8b7e24c173596812e392558234638cc78/src/vs/workbench/contrib/chat/browser/chatAccessibilityService.ts#L24

meganrogge: seems i didn't push 🫠

dmitrivMS (approved):

PR #268224, opened 2025-09-24
Use custom icon for chatSessions to continue showing when FileIcon is turned off

vijayupadya:
Fixes #265150

Chat Sessions is using VS Code's standard ResourceLabel component which respects global file icon theme settings. When file icons setting is turned off, global CSS rule .predefined-file-icon { display: none !important; } hides all icons rendered through this system.
This change implements a custom icon for Chat Sessions similar to Timeline and Outline views that bypasses the file icon theme.


dmitrivMS (dismissed):

mjbvz (commented):
I wonder if it's possible to do this entirely with css 

Take a look at how the file icons are rendered when enabled. Maybe we can avoid hiding them when the icons are disabled?

vijayupadya:
abandoning due to merge conflict related corruption

Issue #268219, opened 2025-09-24
Support for worktree in chat sessions


osortega:
This is to track the E2E support of creating chat sessions in a worktree

- [x] Support for contributed `create` actions to chat sessions view
- [ ] Support for create with metadata command
- [ ] Support for create worktree command with defaults
- [ ] Contribute new create with metadata command

lszomoru:
Closing this issue as this work has been covered by the local background agent session work.

PR #268236, opened 2025-09-24
Support for creating new chat with metadata

osortega:
Related to: https://github.com/microsoft/vscode/issues/268219

This adds a new command to support creating new chats with metadata


src/vs/workbench/contrib/chat/browser/chatSessions.contribution.ts
```diff
@@ -633,6 +637,7 @@ class CodingAgentChatImplementation extends Disposable implements IChatAgentImpl
 						request,
 						prompt: request.message,
 						history,
+						metadata
```
joshspicer, line 640: I am deleting this dynamic agent with https://github.com/microsoft/vscode/pull/268089

PR #268248, opened 2025-09-25
CSS cleanup

osortega:
Related: https://github.com/microsoft/vscode/issues/268329


src/vs/workbench/contrib/chat/browser/media/chatSessions.css
```diff
@@ -42,7 +42,7 @@
 	font-style: italic;
 }
 
-/* Simple approach - directly style the edit container for chat sessions */
+/* Style for editing chat sessions */
```
osortega: Starts the styling used for the style when renaming chat sessions. Similar to the styling of renaming files

src/vs/workbench/contrib/chat/browser/media/chatSessions.css
```diff
@@ -73,11 +72,15 @@
 	border: none !important; /* Remove any default border */
 }
 
+.getting-started-list-container .monaco-list-row {
```
osortega, line 21: Style for the welcome view when user is signed out or empty list

src/vs/workbench/contrib/chat/browser/media/chatSessions.css
```diff
@@ -73,11 +72,15 @@
...
+}
+
+/* Style for two line descriptions */
+
 .chat-sessions-tree-container .chat-session-item.multiline {
 	padding: 2px 0;
 }
-
-/* Position session content and actions inline */
 .chat-sessions-tree-container .chat-session-item .session-content {
```
osortega, line 29: Styling starts for multiline tree when user selected rich descriptions

src/vs/workbench/contrib/chat/browser/media/chatSessions.css
```diff
@@ -133,47 +134,12 @@
 	padding-left: 4px;
 }
 
-.chat-sessions-tree-container .chat-session-item .actions {
```
osortega, line 136: All of these not needed since the defautl behavior of the actions UI/UX is this already

src/vs/workbench/contrib/chat/browser/media/chatSessions.css
```diff
@@ -133,47 +134,12 @@
...
-	min-width: 0; /* Allow text to truncate */
-	text-overflow: ellipsis;
-	overflow: hidden;
-}
-
-.chat-sessions-tree-container .chat-session-item .monaco-icon-label::before {
-	text-align: center;
-}
 
 .chat-sessions-tree-container .chat-session-item .monaco-icon-label.codicon-loading::before {
```
osortega: Animating the progress icon, looks like there is not a native way of doing it with `ThemeIcons` from core

src/vs/workbench/contrib/chat/browser/media/chatSessions.css
```diff
@@ -133,47 +134,12 @@
...
 	animation: codicon-spin 1.5s steps(30) infinite;
 }
 
-/* Timestamp styling - similar to timeline pane */
+/* Needed for showing the timestamp on hover, from timeline pane */
```
osortega, line 91: Styling starts for shoing the timestamp on each entry when the user hovers on it

src/vs/workbench/contrib/chat/browser/media/chatSessions.css
```diff
@@ -219,6 +185,7 @@
 	width: initial;
 }
 
+/* Needed for not pushing the timestamp to the side of the action */
```
osortega, line 137: Styling starts for showing the timestamp to the right of the entries. Exactly the same as how timeline does it

src/vs/workbench/contrib/chat/browser/media/chatSessions.css
```diff
@@ -235,20 +202,10 @@
 	margin: 0;
 }
 
-/* Show twistie only for collapsible items (like "Show history...") */
 .chat-sessions-tree-container .monaco-list-row[aria-expanded] .monaco-tl-twistie {
```
osortega: This was really strange to me, seems like the treeview already supports hiding the twistie but it only removes the icon but doesn't change the padding.
Without this rule we get a really weird UI with lots of padding on the left

src/vs/workbench/contrib/chat/browser/chatSessions/view/sessionsViewPane.ts
```diff
@@ -284,6 +284,7 @@ export class SessionsViewPane extends ViewPane {
 				overrideStyles: {
 					listBackground: undefined
 				},
+				// hideTwistiesOfChildlessElements: true,
```
osortega: This doesn't really work the way I expected, it only removes the twistie but keeps the space for it

benibenj: Only parent nodes can have twisties. The option `hideTwistiesOfChildlessElements` specifies whether a twisty should be rendered for parent nodes which do not have any visible children. 

For example in the file explorer this option is set to false as we want folders which do not have any files to still render a twisty to make it easier for the user to understand that this is an empty folder.

In the outline view we hide twisties when parent nodes do not have any children as it is very normal for a funtion or method to not have any child nodes.

If `hideTwistiesOfChildlessElements` is set to true, then the twisty is not rendered for empty parent folders, however, the spacing/indentation stays the same. This has to be this way, otherwise the tree does not look like a nice tree anymore and the items do not align.

src/vs/workbench/contrib/chat/browser/media/chatSessions.css
benibenj: This is not needed. ThemeIcon supports adding modifiers in the following way `ThemeIcon.modify`. You can use this theme icon for your spinner: https://github.com/microsoft/vscode/blob/3bf92e795ffd1ef7c3ff38a774d2bdb1a747e0a6/src/vs/platform/theme/common/iconRegistry.ts#L345

src/vs/workbench/contrib/chat/browser/chatSessions/view/sessionsViewPane.ts
osortega: I ended up doing what the explorer does, basically adding the `file-icon-themable-tree` and toggling the class `align-icons-and-twisties` depending on the theme

src/vs/workbench/contrib/chat/browser/media/chatSessions.css
osortega: I tried doing that before but it doesn't really work not sure why. It doesn't render at all
Screenshot 2025-09-25 at 1 25 56 PM

src/vs/workbench/contrib/chat/browser/media/chatSessions.css
benibenj: This seems fine in general. But it should not use the `.explorer-item` but rather create a `.chat-session-item`. Also, as both explorer and chat sessions view need this CSS, it might make sense to have shared classes which are defined somewhere instead of having a copy. Maybe move all `.explorer-item-edited` css to views.css and call the class `.tree-item-edited`

src/vs/workbench/contrib/chat/browser/chatSessions/view/sessionsTreeRenderer.ts
```diff
@@ -361,6 +363,10 @@ export class SessionsRenderer extends Disposable implements ITreeRenderer<IChatS
 		}
 
 		// Create a simple container that mimics the file explorer's structure
+
```
benibenj: This is not the explorer, so we shouldn't use `explorer-item` naming

src/vs/workbench/contrib/chat/browser/chatSessions/view/sessionsViewPane.ts
```diff
@@ -381,6 +380,15 @@ export class SessionsViewPane extends ViewPane {
 			}
 		}));
 
+		// Handle icons and twisties correctly
+		const onDidChangeFileIconTheme = (theme: IFileIconTheme) => {
+			if (this.treeContainer) {
+				this.treeContainer.classList.toggle('hide-arrows', theme.hidesExplorerArrows === true);
```
benibenj, line 386: Not sure if this is needed, have you tested this?

src/vs/workbench/contrib/chat/browser/media/chatSessions.css
```diff
@@ -4,16 +4,20 @@
  *--------------------------------------------------------------------------------------------*/
 
 /* Ensure consistent title background regardless of number of views */
+
+/* Is this needed? This seems like something that the view management does or should do */
```
benibenj: This seems weird, why is this needed, I don't think this should be css fix

src/vs/workbench/contrib/chat/browser/media/chatSessions.css
```diff
@@ -4,16 +4,20 @@
...
 .composite.viewlet[id="workbench.view.chat.sessions"] .pane-header.expanded.not-collapsible {
 	background-color: var(--vscode-sideBarSectionHeader-background) !important;
 }
 
+/* Is this needed? It does not seem to do something and it seems a bit weird to me. This should be handled by the view management */
```
benibenj: some of these do not seem to be doing anything

benibenj (changes requested):

src/vs/workbench/contrib/chat/browser/chatSessions/view/sessionsViewPane.ts
benibenj: I think you might be right that this is needed

src/vs/workbench/contrib/chat/browser/media/chatSessions.css
osortega: I moved it to views.css as you suggested but only did the change to using that class within the sessions view for now so that this PR is self contained on the chat sessions UI.
In a separate PR we can migrate the explorer to use this as well

src/vs/workbench/contrib/chat/browser/chatSessions/view/sessionsTreeRenderer.ts
osortega: Removed it

src/vs/workbench/contrib/chat/browser/media/chatSessions.css
osortega: Removed it

src/vs/workbench/contrib/chat/browser/media/chatSessions.css
osortega: Removed all of this

src/vs/workbench/browser/parts/views/media/views.css
```diff
@@ -323,3 +323,32 @@
 	color: var(--vscode-inputOption-activeForeground);
 	background-color: var(--vscode-inputOption-activeBackground);
 }
+
+.tree-item-edited {
```
bpasero, line 327: Can we expand on why this is here? The only use seems to be in chat sessions, why is it made so generic?

bpasero (changes requested):

src/vs/workbench/browser/parts/views/media/views.css
benibenj: These styles were all copied from `explorerviewlet.css` into chat sessions view. Instead of having a copy, both explorer and chat sessions view should use the same styling, so moving in here seems reasonable. But it looks like these classes have not yet been adopted for the explorer. Maybe this work should be done in a separate PR (/debt item)

bpasero: Since we are doing the step of moving to `views.css` I would also suggest to do the explorer adoption unless that is a lot of work?

benibenj: This PR has reached a rather large size and to not delay merging this much more I would suggest doing this work in a separate PR. It will also make it easier to revert/understand the change in case something breaks. @osortega do you mind moving these styles back to `chatSessionss.css`. We can then merge this PR and in a follow up ensure we use the same styles across all trees that support renaming (explorer and chat sessions currently). Also there is a merge conflict to resolve.

PR #268263, opened 2025-09-25
Add explain action to DiffHunkWidget

zhichli:
[491826714-291efac9-b9b9-4743-bd57-92659d02d7cc.webm](https://github.com/user-attachments/assets/08759823-4ff9-4482-a6fb-7f13a494399d)


connor4312:
This is a cool idea, but I'm not sure how much I would use it because I generally don't want to pollute the context of my main agent session. I would prefer if this opened in inline chat, which is a good fit because I'm generally looking for an one-off Q and A there. Bonus if we actually spliced in the conversation history but that is more work.

PR #268362, opened 2025-09-25
add command to focus last focused chat item, include important info in a11y help

### LINKED ISSUE #268238: Track the last focused chat message so that moving between chat input and response list is smoother for kb users
Idea from @rperez030 : Provide a mechanism to move between the input field and the chat thread without losing focus. Right now, Ctrl+Up Arrow always moves focus to the latest message in the thread. We would need a mechanism to return to the chat that moves focus not to the latest message, but to the last active message, knowing that not all Windows screen readers are good at respecting focus events in a web view.

### LINKED ISSUE #268345: a11y chat help is missing important info
Such as how to navigate to the chat list

meganrogge:
fixes https://github.com/microsoft/vscode/issues/268238
fixes https://github.com/microsoft/vscode/issues/268345

https://github.com/user-attachments/assets/2887a192-9e32-4747-ba69-d0ad1787ceab


TylerLeonhardt (approved):

src/vs/workbench/contrib/chat/browser/chatWidget.ts
```diff
@@ -1555,6 +1557,18 @@ export class ChatWidget extends Disposable implements IChatWidget {
...
 			}));
+
+		this._register(this.tree.onDidChangeFocus(() => {
+			const focused = this.tree.getFocus();
+			if (focused && focused.length > 0) {
+				const focusedItem = focused[0];
+				const items = this.tree.getNode(null).children;
+				const idx = items.findIndex(i => i.element === focusedItem);
+				if (idx !== -1) {
+					this._mostRecentlyFocusedItemIndex = idx;
```
roblourens, line 1568: I actually don't think this part is necessary, because I believe the tree remembers its focus. You can see if I'm wrong, but I think this would work the same if you just don't do a `setFocus` in this scenario, but just focus the tree, I would expect it to have the same item focused.

For example when you move focus around the explorer or other trees, it always moves back to the same element automatically.

PR #268364, opened 2025-09-25
File icon display handling in chat sessions 

vijayupadya:
Fixes #265150

Chat session icons use `ThemeIcon` objects (like `Codicon.loading`, `Codicon.pass`, etc.) which get the `predefined-file-icon` CSS class. When file icons are globally disabled, a CSS rule in style.css hides all elements with this class, affecting chat session icons unintentionally. 
This change switches to IconLabel from the current resourceLabel in chatSessions to address this issue.

### DIFF
--- a/src/vs/workbench/contrib/chat/browser/chatSessions/view/sessionsTreeRenderer.ts
+++ b/src/vs/workbench/contrib/chat/browser/chatSessions/view/sessionsTreeRenderer.ts
@@ -58,6 +58,7 @@ interface ISessionTemplateData {
 	readonly descriptionRow: HTMLElement;
 	readonly descriptionLabel: HTMLElement;
 	readonly statisticsLabel: HTMLElement;
+	readonly customIcon: HTMLElement;
 }
 
 export interface IGettingStartedItem {
@@ -189,6 +190,8 @@ export class SessionsRenderer extends Disposable implements ITreeRenderer<IChatS
 
 		// Create a container that holds the label, timestamp, and actions
 		const contentContainer = append(element, $('.session-content'));
+		// Create custom icon element that won't be affected by file icon theme
+		const customIcon = append(contentContainer, $('.chat-session-custom-icon'));
 		const resourceLabel = this.labels.create(contentContainer, { supportHighlights: true });
 		const descriptionRow = append(element, $('.description-row'));
 		const descriptionLabel = append(descriptionRow, $('span.description'));
@@ -205,6 +208,7 @@ export class SessionsRenderer extends Disposable implements ITreeRenderer<IChatS
 		return {
 			container: element,
 			resourceLabel,
+			customIcon,
 			actionBar,
 			elementDisposable,
 			timestamp,
@@ -305,6 +309,20 @@ export class SessionsRenderer extends Disposable implements ITreeRenderer<IChatS
 				} : undefined) :
 			undefined;
 
+		// Handle custom icon rendering (independent of file icon theme)
+		if (iconTheme) {
+			templateData.customIcon.className = `chat-session-custom-icon ${ThemeIcon.asClassName(iconTheme)}`;
+			if (iconTheme.color) {
+				const theme = this.themeService.getColorTheme();
+				templateData.customIcon.style.color = theme.getColor(iconTheme.color.id)?.toString() ?? '';
+			} else {
+				templateData.customIcon.style.color = '';
+			}
+			templateData.customIcon.style.display = 'flex';
+		} else {
+			templateData.customIcon.style.display = 'none';
+		}
+
 		// Set the resource label
 		templateData.resourceLabel.setResource({
 			name: session.label,
@@ -312,7 +330,7 @@ export class SessionsRenderer extends Disposable implements ITreeRenderer<IChatS
 			resource: iconResource
 		}, {
 			fileKind: undefined,
-			icon: iconTheme,
+			hideIcon: true, // Hide the ResourceLabel's icon since we're using custom icon
 			// Set tooltip on resourceLabel only for single-row items
 			title: !renderDescriptionOnSecondRow || !session.description ? tooltipContent : undefined
 		});
--- a/src/vs/workbench/contrib/chat/browser/chatSessions/view/sessionsTreeRenderer.ts
+++ b/src/vs/workbench/contrib/chat/browser/chatSessions/view/sessionsTreeRenderer.ts
@@ -211,7 +211,7 @@ export class SessionsRenderer extends Disposable implements ITreeRenderer<IChatS
 			iconTheme = session.iconPath;
 		}
 
-		const renderDescriptionOnSecondRow = this.configurationService.getValue(ChatConfiguration.ShowAgentSessionsViewDescription) && session.provider.chatSessionType !== 'local';
+		const renderDescriptionOnSecondRow = this.configurationService.getValue<boolean>(ChatConfiguration.ShowAgentSessionsViewDescription) && session.provider.chatSessionType !== 'local';
 
 		if (renderDescriptionOnSecondRow && session.description) {
 			templateData.container.classList.toggle('multiline', true);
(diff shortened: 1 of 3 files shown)


src/vs/workbench/contrib/chat/browser/media/chatSessions.css
```diff
@@ -173,6 +173,32 @@
...
+}
+
+.monaco-workbench:not(.file-icons-enabled) .chat-sessions-tree-container .predefined-file-icon.codicon-circle-outline::before {
+	content: var(--vscode-icon-circle-outline-content) !important;
+	font-family: var(--vscode-icon-circle-outline-font-family, codicon) !important;
+}
+
+.monaco-workbench:not(.file-icons-enabled) .chat-sessions-tree-container .predefined-file-icon.codicon-chat-sparkle::before {
+	content: var(--vscode-icon-chat-sparkle-content) !important;
+	font-family: var(--vscode-icon-chat-sparkle-font-family, codicon) !important;
```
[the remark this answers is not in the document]
vijayupadya: updated

lramos15 (dismissed):

src/vs/workbench/contrib/chat/browser/media/chatSessions.css
```diff
@@ -173,6 +173,35 @@
 	animation: codicon-spin 1.5s steps(30) infinite;
 }
 
+/* Override global file icon hiding for chat session icons */
```
osortega: @benibenj is there a way to do this directly through the provider? Basically ignoring the setting?

osortega (commented):
We are trying to reduce this CSS class as much as possible to avoid having too many customizations in favor of what the tree view provider exposes in the API.
Let's confirm first if this can't be done through the API

src/vs/workbench/contrib/chat/browser/media/chatSessions.css
```diff
@@ -173,6 +173,35 @@
...
+.monaco-workbench:not(.file-icons-enabled) .chat-sessions-tree-container .predefined-file-icon.codicon-chat-sparkle::before {
+	font-family: codicon;
+}
+
+.monaco-workbench:not(.file-icons-enabled) .chat-sessions-tree-container .predefined-file-icon.codicon-loading::before {
+	content: var(--vscode-icon-loading-content) !important;
+}
+
+.monaco-workbench:not(.file-icons-enabled) .chat-sessions-tree-container .predefined-file-icon.codicon-pass::before {
+	content: var(--vscode-icon-pass-content) !important;
```
mjbvz: Hardcoding the list of supported icons seems fragile. I'd actually prefer your original approach over this, unless there is some way to re-enable the icons using css in a more generic way

vijayupadya: Switched back to earlier custom icon implementation.

bpasero (dismissed):
Actually this change seems to be within a tree item, not on the tree level.

src/vs/workbench/contrib/chat/browser/chatSessions/view/sessionsTreeRenderer.ts
```diff
@@ -305,14 +309,28 @@ export class SessionsRenderer extends Disposable implements ITreeRenderer<IChatS
...
+			} else {
+				templateData.customIcon.style.color = '';
+			}
+			templateData.customIcon.style.display = 'flex';
+		} else {
+			templateData.customIcon.style.display = 'none';
+		}
+
 		// Set the resource label
 		templateData.resourceLabel.setResource({
```
benibenj: I know this was not introduced in this PR, but I don't think this should be a `ResourceLabel` as there is not actually a resource backing this label (meaning a file or a folder for example). Below is `resource: iconResource` which is very unusual. It look like you are fighting against the resource label by setting `fileKind: undefined` and `hideIcon: true` and making CSS changes.

We typically use `ResourceLabels` when we want to render a label for a file or a folder (or anything which is backed by a resource / `URI`). In other cases we tend to use `IconLabel` which allows rendering an icon and a label. This is probably what you want in this case.

If you use `IconLabel`, you will probably not need the `customIcon` DOM element as well as the code in the `if (iconTheme)` above. Regarding the spinner, you can do `modify(iconTheme, 'spin')`, no CSS needed.

benibenj (changes requested):
I think it would make more sense to just move to use an `IconLabel` instead of doing these changes

src/vs/workbench/contrib/chat/browser/chatSessions/view/sessionsTreeRenderer.ts
vijayupadya: Updated to use IconLabel, @benibenj , @osortega

src/vs/workbench/contrib/chat/browser/chatSessions/view/sessionsTreeRenderer.ts
```diff
@@ -305,17 +253,29 @@ export class SessionsRenderer extends Disposable implements ITreeRenderer<IChatS
...
+				templateData.customIcon.style.color = theme.getColor(iconTheme.color.id)?.toString() ?? '';
+			} else {
+				templateData.customIcon.style.color = '';
+			}
+			templateData.customIcon.style.display = 'flex';
+		} else {
+			templateData.customIcon.style.display = 'none';
+		}
+
+		// Set the icon label
```
osortega, line 256: Q: what does this do?

src/vs/workbench/contrib/chat/browser/chatSessions/view/sessionsTreeRenderer.ts
```diff
@@ -305,17 +253,29 @@ export class SessionsRenderer extends Disposable implements ITreeRenderer<IChatS
...
-		}, {
-			fileKind: undefined,
-			icon: iconTheme,
-			// Set tooltip on resourceLabel only for single-row items
-			title: !renderDescriptionOnSecondRow || !session.description ? tooltipContent : undefined
-		});
+		// Handle custom icon rendering
+		if (iconTheme) {
+			templateData.customIcon.className = `chat-session-custom-icon ${ThemeIcon.asClassName(iconTheme)}`;
+			if (iconTheme.color) {
```
osortega: We don't support color anymore, we could remove this

src/vs/workbench/contrib/chat/browser/media/chatSessions.css
```diff
@@ -169,7 +169,17 @@
...
+.chat-sessions-tree-container .chat-session-item .chat-session-custom-icon {
+	flex-shrink: 0;
+	width: 16px;
+	height: 16px;
+	margin-right: 6px;
+	font-size: 16px;
+	color: var(--vscode-icon-foreground);
+}
+
+.chat-sessions-tree-container .chat-session-item .chat-session-custom-icon.codicon-loading::before {
```
osortega: Could you try changing [this loading icon](https://github.com/microsoft/vscode/blob/fc4d4d8ee53e384002f341411c71326a4fddfc51/src/vs/workbench/contrib/chat/browser/chatSessions/view/sessionsTreeRenderer.ts#L220-L221) to be `ThemeIcon.modify(Codicon.loading, 'spin')` and see if that works now that we are using IconLabel? 
If it does then we can remove this CSS rule

vijayupadya: Yes, that works. Removed the CSS rule

src/vs/workbench/contrib/chat/browser/chatSessions/view/sessionsTreeRenderer.ts
vijayupadya: Removed

src/vs/workbench/contrib/chat/browser/chatSessions/view/sessionsTreeRenderer.ts
vijayupadya: Hides the custom icon element when the current session has no icon. Added a comment.

src/vs/workbench/contrib/chat/browser/chatSessions/view/sessionsTreeRenderer.ts
```diff
@@ -51,13 +49,14 @@ import { getLocalHistoryDateFormatter } from '../../../../localHistory/browser/l
...
 	readonly container: HTMLElement;
-	readonly resourceLabel: IResourceLabel;
+	readonly iconLabel: IconLabel;
 	readonly actionBar: ActionBar;
 	readonly elementDisposable: DisposableStore;
 	readonly timestamp: HTMLElement;
 	readonly descriptionRow: HTMLElement;
 	readonly descriptionLabel: HTMLElement;
 	readonly statisticsLabel: HTMLElement;
+	readonly customIcon: HTMLElement;
```
osortega, line 59: Should we remove this?

osortega (dismissed):

mjbvz (dismissed):

benibenj (approved):

aeschli (approved):

PR #268377, opened 2025-09-25
Unblock terminal tool in evals

roblourens:
@meganrogge this is separate from auto-approve issue. In our current msbench simulator environment, tools run outside of a chat session context, so I skip the OutputMonitor in that scenario since we can't ask anyone for a confirmation anyway. Lmk if you think I should do something differently.

### DIFF
--- a/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/runInTerminalTool.ts
+++ b/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/runInTerminalTool.ts
@@ -524,8 +524,8 @@ export class RunInTerminalTool extends Disposable implements IToolImpl {
 				}
 				this._logService.debug(`RunInTerminalTool: Using \`${strategy.type}\` execute strategy for command \`${command}\``);
 				store.add(strategy.onDidCreateStartMarker(startMarker => {
-					if (!outputMonitor && invocation.context?.sessionId) {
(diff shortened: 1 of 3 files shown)


src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/runInTerminalTool.ts
```diff
@@ -524,8 +524,8 @@ export class RunInTerminalTool extends Disposable implements IToolImpl {
 				}
 				this._logService.debug(`RunInTerminalTool: Using \`${strategy.type}\` execute strategy for command \`${command}\``);
 				store.add(strategy.onDidCreateStartMarker(startMarker => {
-					if (!outputMonitor) {
-						outputMonitor = store.add(this._instantiationService.createInstance(OutputMonitor, { instance: toolTerminal.instance, sessionId: invocation.context!.sessionId, getOutput: (marker?: IXtermMarker) => getOutput(toolTerminal.instance, marker ?? startMarker) }, undefined, invocation.context!, token, command));
+					if (!outputMonitor && invocation.context?.sessionId) {
```
meganrogge: should this `invocation.context?.sessionId` check be moved outside of the `store.add`?

roblourens:
Good idea, I set it up to still use the OutputMonitor, took a lot of `| undefined` plumbing, lmk if you're ok with that approach

rzhao271 (approved):

roblourens:

PR #268488, opened 2025-09-26
Add support for MCP servers over Unix domain sockets and named pipes

bwateratmsft:
Fixes #268487.

This PR adds support for doing StreamableHTTP MCP servers over Unix domain sockets and named pipes. For local MCP servers, this is generally more secure than standard localhost HTTP. Standard IO servers are still preferable for local MCP servers, but some scenarios are better served by StreambleHTTP--for example, an MCP server can be hosted in the extension host process by an extension using this approach.

/cc @connor4312

Below is some sample extension code. In practice, a random pipe/socket name should be used, and permissions enforced as well, but this shows the gist.

```typescript
export function activate() {

    const isWindows = os.platform() === 'win32';
    const socketPath = isWindows ? '\\\\.\\pipe\\hello-world' : path.join(os.tmpdir(), 'hello-world.sock');

    const socketUri = vscode.Uri.from({
        scheme: isWindows ? 'pipe' : 'unix',
        path: socketPath,
        fragment: '/mcp' // The express app below is configured to serve MCP over the `/mcp` route
    });

    vscode.lm.registerMcpServerDefinitionProvider('hello-world', {
        provideMcpServerDefinitions(token: vscode.CancellationToken): vscode.McpServerDefinition[] {
            return [
                new vscode.McpHttpServerDefinition(
                    'hello-world',
                    socketUri,
                ),
            ];
        },
        resolveMcpServerDefinition(server: vscode.McpServerDefinition, token: vscode.CancellationToken): vscode.McpServerDefinition {
            startMcpServer(socketPath); // Additionally, the server should also be closed on extension deactivation
            return server;
        }
    });

}

function startMcpServer(socketPath: string) {
    // Uses express, same as the MCP TypeScript SDK examples
    const app = express();

    // MCP StreamableHTTP server setup, refer to examples at https://www.npmjs.com/package/@modelcontextprotocol/sdk#streamable-http
    // ...

    // Listen on the socket/named pipe
    app.listen(socketPath);
}
```

### DIFF
--- a/src/vs/workbench/api/common/extHostMcp.ts
+++ b/src/vs/workbench/api/common/extHostMcp.ts
@@ -33,9 +33,8 @@ export interface IExtHostMpcService extends ExtHostMcpShape {
 
 export class ExtHostMcpService extends Disposable implements IExtHostMpcService {
 	protected _proxy: MainThreadMcpShape;
-	protected _mcpHttpHandleType = McpHTTPHandle;
 	private readonly _initialProviderPromises = new Set<Promise<void>>();
-	private readonly _sseEventSources = this._register(new DisposableMap<number, McpHTTPHandle>());
+	protected readonly _sseEventSources = this._register(new DisposableMap<number, McpHTTPHandle>());
 	private readonly _unresolvedMcpServers = new Map</* collectionId */ string, {
 		provider: vscode.McpServerDefinitionProvider;
 		servers: vscode.McpServerDefinition[];
@@ -43,7 +42,7 @@ export class ExtHostMcpService extends Disposable implements IExtHostMpcService 
 
 	constructor(
 		@IExtHostRpcService extHostRpc: IExtHostRpcService,
-		@ILogService private readonly _logService: ILogService,
+		@ILogService protected readonly _logService: ILogService,
 		@IExtHostInitDataService private readonly _extHostInitData: IExtHostInitDataService,
 		@IExtHostWorkspace protected readonly _workspaceService: IExtHostWorkspace,
 		@IExtHostVariableResolverProvider private readonly _variableResolver: IExtHostVariableResolverProvider,
@@ -58,7 +57,7 @@ export class ExtHostMcpService extends Disposable implements IExtHostMpcService 
 
 	protected _startMcp(id: number, launch: McpServerLaunch, _defaultCwd?: URI, errorOnUserInteraction?: boolean): void {
 		if (launch.type === McpServerTransportType.HTTP) {
-			this._sseEventSources.set(id, new this._mcpHttpHandleType(id, launch, this._proxy, this._logService, errorOnUserInteraction));
+			this._sseEventSources.set(id, new McpHTTPHandle(id, launch, this._proxy, this._logService, errorOnUserInteraction));
 			return;
 		}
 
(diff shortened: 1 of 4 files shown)


src/vs/workbench/api/node/extHostMcpNode.ts
```diff
@@ -134,6 +136,42 @@ export class NodeExtHostMpcService extends ExtHostMcpService {
 	}
 }
 
+class McpHTTPHandleNode extends McpHTTPHandle {
+	protected override async _fetchInternal(url: string, init?: CommonRequestInit): Promise<Response> {
+		const { fetch, Agent } = await import('undici');
+
+		const undiciInit = { ...init, dispatcher: undefined as any };
```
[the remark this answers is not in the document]
bwateratmsft: Well, I can't really do that without doing an `import type` from `undici`, and that makes the linter upset, so...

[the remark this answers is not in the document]
bwateratmsft: Eh, I'll just suppress that linter warning...

bwateratmsft:
Not sure what I'm supposed to do about the failing check for `package-lock.json` edits...

src/vs/workbench/api/node/extHostMcpNode.ts
```diff
@@ -134,6 +136,42 @@ export class NodeExtHostMpcService extends ExtHostMcpService {
...
+
+		let httpUrl = url;
+		const uri = URI.parse(url);
+
+		if (uri.scheme === 'unix' || uri.scheme === 'pipe') {
+			// By convention, we put the *socket path* as the URI path, and the *request path* in the fragment
+			// So, set the dispatcher with the socket path
+			undiciInit.dispatcher = new Agent({
+				socketPath: uri.path,
+			});
```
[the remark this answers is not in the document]
bwateratmsft: You are hallucinating that `pipe` property...also, this *does* work as authored, I've tested it...so...

src/vs/workbench/api/common/extHostMcp.ts
```diff
@@ -33,6 +33,7 @@ export interface IExtHostMpcService extends ExtHostMcpShape {
 
 export class ExtHostMcpService extends Disposable implements IExtHostMpcService {
 	protected _proxy: MainThreadMcpShape;
+	protected _mcpHttpHandleType = McpHTTPHandle;
```
connor4312: Let's keep this class unchanged and instead modify our override of `_startMcp` in the Node implementation of this service

src/vs/workbench/api/node/extHostMcpNode.ts
```diff
@@ -134,6 +138,42 @@ export class NodeExtHostMpcService extends ExtHostMcpService {
 	}
 }
 
+class McpHTTPHandleNode extends McpHTTPHandle {
+	protected override async _fetchInternal(url: string, init?: CommonRequestInit): Promise<Response> {
+		const { fetch, Agent } = await import('undici');
```
connor4312: I would pop this to

```
private readonly _undici = new Lazy(() => import('undici'))
```

and then use it as

```
// Note: imported async so that we can ensure we load undici after proxy patches have been applied
const { fetch, Agent } = this._undici.value;
```

to avoid going through module resolution on each http call.

connor4312 (commented):
A few nits but overall good

src/vs/workbench/api/common/extHostMcp.ts
bwateratmsft: Done!

src/vs/workbench/api/node/extHostMcpNode.ts
bwateratmsft: Done! Still needed an `await` but I think it will keep the existing resolved promise for the rest of the life of the `Lazy`, right?

PR #268499, opened 2025-09-26
Update MCP schema to be more conscise

lramos15:
Some models seem to 400 due to this schema cc @TylerLeonhardt


rebornix (approved):

TylerLeonhardt (approved):
Thanks!

PR #268506, opened 2025-09-26
api: allow byok model providers to hint at good edit tools

connor4312:
For our own models, we set up the edit tools they get. For generic open-
ended providers like OpenRouter, we will have a 'learning' functionality.
But for BYOK providers that can provide a curated set of models (e.g.
Cerebras) we should let them tell us what edit tool is appropriate for
models they give us so users have a good experience from request 0.

The API is intentionally loose because edit tools continue to evolve and
our set of available tools will continue to change. So I do not surface
any of these in actual API types, but give a string array of hints the
provider can give us. If in the future none of the tools they hint are
ones we recognize, as the space evolves, then the chat extension will
fall back to its default learning behavior.

### DIFF
--- a/src/vscode-dts/vscode.proposed.chatProvider.d.ts
+++ b/src/vscode-dts/vscode.proposed.chatProvider.d.ts
@@ -56,6 +56,25 @@ declare module 'vscode' {
 		readonly statusIcon?: ThemeIcon;
 	}
 
+	export interface LanguageModelChatCapabilities {
+		/**
+		 * The tools the model prefers for making file edits. If not provided or if none of the tools,
+		 * are recognized, the editor will try multiple edit tools and pick the best one. The available
+		 * edit tools WILL change over time and this capability only serves as a hint to the editor.
+		 *
+		 * Edit tools currently recognized include:
+		 * - 'find-replace': Find and replace text in a document.
+		 * - 'multi-find-replace': Find and replace text in a document.
+		 * - 'apply-patch': A file-oriented diff format used by some OpenAI models
+		 * - 'code-rewrite': A general but slower editing tool that allows the model
+		 *   to rewrite and code snippet and provide only the replacement to the editor.
+		 *
+		 * The order of edit tools in this array has no significance; all of the recognized edit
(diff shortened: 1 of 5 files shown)


src/vscode-dts/vscode.proposed.chatProvider.d.ts
```diff
@@ -56,6 +56,25 @@ declare module 'vscode' {
...
+		 * - 'find-replace': Find and replace text in a document.
+		 * - 'multi-find-replace': Find and replace multiple text snippets across documents.
+		 * - 'apply-patch': A file-oriented diff format used by some OpenAI models
+		 * - 'code-rewrite': A general but slower editing tool that allows the model
+		 *   to rewrite and code snippet and provide only the replacement to the editor.
+		 *
+		 * The order of edit tools in this array has no significance; all of the recognized edit
+		 * tools will be made available to the model.
+		 */
+		readonly editTools?: string[];
```
roblourens, line 75: Referring to these specific tools in extension API feels wrong but I agree that it's a good idea for a model provider to decide this and tell us. Hoping there isn't much churn, like more custom formats like apply-patch in the future

`Set<string>` maybe?

connor4312: We so far don't use Sets or Maps in vscode.d.ts (avoiding required interaction with es6 constructs) but maybe it's time to change that in 2025 🤔

connor4312: > Hoping there isn't much churn, like more custom formats like apply-patch in the future

My bet is GPT keeps apply_patch and everyone else normalizes to replace_string, or goes to apply_patch since it's more expressive, but I'm no fortune teller 🔮

roblourens: Ah good point. There's also `{ string: boolean }` but that seems like overkill, array seems fine.

mjbvz (approved):
vscode.d.ts changes look good to me

PR #268526, opened 2025-09-26
always preserve task name, remove presentation option

### LINKED ISSUE #268389: Consider by default preventing task name override
A `presentation` option was added to tasks here https://github.com/microsoft/vscode/pull/268350. 

After testing, we should consider making this happen by default and removing that option.

meganrogge:
fixes #268389

### DIFF
--- a/src/vs/workbench/contrib/terminal/browser/terminalInstance.ts
+++ b/src/vs/workbench/contrib/terminal/browser/terminalInstance.ts
@@ -2256,10 +2256,7 @@ export class TerminalInstance extends Disposable implements ITerminalInstance {
 	}
 
 	private _setTitle(title: string | undefined, eventSource: TitleEventSource): void {
-		if (this._shellLaunchConfig?.type === 'Task' &&
-			(this._shellLaunchConfig?.preserveTaskName || this._titleSource === TitleEventSource.Api) &&
-			eventSource === TitleEventSource.Process) {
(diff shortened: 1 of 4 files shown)


src/vs/workbench/contrib/terminal/browser/terminalInstance.ts
```diff
@@ -2256,10 +2256,7 @@ export class TerminalInstance extends Disposable implements ITerminalInstance {
 	}
 
 	private _setTitle(title: string | undefined, eventSource: TitleEventSource): void {
-		if (this._shellLaunchConfig?.type === 'Task' &&
-			(this._shellLaunchConfig?.preserveTaskName || this._titleSource === TitleEventSource.Api) &&
-			eventSource === TitleEventSource.Process) {
-			// For task terminals with preserveTaskName enabled or with API-set titles, preserve the title even when the process title changes
+		if ((this._shellLaunchConfig?.type === 'Task' || this._titleSource === TitleEventSource.Api) && eventSource === TitleEventSource.Process) {
```
rzhao271, line 2259: Should the || be an &&

roblourens (approved):

src/vs/workbench/contrib/terminal/browser/terminalInstance.ts
meganrogge: nope!

PR #268548, opened 2025-09-26
Fix the type incompatibility issue in the MCP HTTP server handler

bwateratmsft:
Fixes a type incompatibility bug introduced by #268488 / #268524.

### DIFF
--- a/src/vs/workbench/api/common/extHostMcp.ts
+++ b/src/vs/workbench/api/common/extHostMcp.ts
@@ -812,7 +812,6 @@ export interface CommonResponse {
 	statusText: string;
 	headers: Headers;
 	body?: ReadableStream | null;
-	bodyUsed: boolean;
 	url: string;
 	json(): Promise<any>;
 	text(): Promise<string>;
--- a/src/vs/workbench/api/common/extHostMcp.ts
+++ b/src/vs/workbench/api/common/extHostMcp.ts
@@ -330,7 +330,7 @@ export class McpHTTPHandle extends Disposable {
 		}
 	}
 
-	private async _populateAuthMetadata(mcpUrl: string, originalResponse: Response): Promise<void> {
+	private async _populateAuthMetadata(mcpUrl: string, originalResponse: CommonResponse): Promise<void> {
 		// If there is a resource_metadata challenge, use that to get the oauth server. This is done in 2 steps.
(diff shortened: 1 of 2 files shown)


src/vs/workbench/api/common/extHostMcp.ts
```diff
@@ -807,6 +807,17 @@ export interface CommonRequestInit extends MinimalRequestInit {
...
 }
 
+export interface CommonResponse {
+	status: number;
+	statusText: string;
+	headers: Headers;
+	body?: ReadableStream | null;
+	bodyUsed: boolean;
+	url: string;
+	json(): Promise<any>;
```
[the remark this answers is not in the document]
bwateratmsft: The standard `json()` from `lib.dom.d.ts` also has `Promise<any>`, and I think it's better to match it exactly.

image

src/vs/workbench/api/node/extHostMcpNode.ts
```diff
@@ -165,13 +165,9 @@ class McpHTTPHandleNode extends McpHTTPHandle {
...
-		const undiciResponse = await fetch(httpUrl, undiciInit);
-
-		return new Response(undiciResponse.body as ReadableStream, {
-			status: undiciResponse.status,
-			statusText: undiciResponse.statusText,
-			headers: undiciResponse.headers
-		});
+		// An `as CommonResponse` is required because of a very slight type mismatch between undici's `Response` and the standard `Response` type
+		// (Way down in `ReadableStreamReadDoneResult<T>`, `value` is optional in the undici type but required [yet can be `undefined`] in the standard type)
+		return await fetch(httpUrl, undiciInit) as CommonResponse;
```
connor4312: It's a little more verbose but can we construct the CommonResponse properties manually and `as`-cast the one problematic property? `as` is a big hammer and would hide any drift that might happen between CommonResponse and undici's Response type.

bwateratmsft: Yes, I also thought about doing a very deep `as` cast on _just_ the `ReadableStreamReadDoneResult<T>`, for more or less the same reason

bwateratmsft: Fixed!

connor4312 (approved):

Tyriar (approved):

PR #268549, opened 2025-09-26
fix: update import paths

rzhao271:
Fixes the build.

The build broke after merging in https://github.com/microsoft/vscode/pull/265431. That PR passed PR checks, but the checks were run three weeks ago.


bhavyaus (approved):

bhavyaus:
How did this get checked in with the compilation broken? Would our PR build not catch this?

PR #268555, opened 2025-09-26
fix terminal tool disposable leak

meganrogge:
fix #268542


connor4312 (approved):

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/executeStrategy/richExecuteStrategy.ts
```diff
@@ -61,10 +61,27 @@ export class RichExecuteStrategy implements ITerminalExecuteStrategy {
...
+				}
+				const marker = xterm.raw.registerMarker();
+				if (!marker) {
+					this._startMarker = undefined;
+					this._onDidCreateStartMarker.fire(undefined);
+					return;
+				}
+				this._startMarker = marker;
+				this._onDidCreateStartMarker.fire(marker);
+				store.add(marker);
```
roblourens, line 76: If marker is disposed but still in the store then it's still being leaked

PR #268921, opened 2025-09-29
Prevent `clear` from flashing in terminal sticky scroll

anthonykim1:
Resolves: https://github.com/microsoft/vscode/issues/268917 

Edit: Just smooth out by hardcoded `clear` similar to iterm2? https://github.com/microsoft/vscode/issues/268917#issuecomment-3349428978


src/vs/workbench/contrib/terminalContrib/stickyScroll/browser/terminalStickyScrollOverlay.ts
```diff
@@ -191,10 +192,17 @@ export class TerminalStickyScrollOverlay extends Disposable {
 			this._pendingShowOperation = true;
 			this._show();
 		} else {
+			this._cancelPendingShow();
 			this._hide();
 		}
 	}
 
+	private _cancelPendingShow(): void {
+		const showFn = this._show as typeof this._show & { clear: () => void };
```
anthonykim1: I mean to cancel debounced `_show` when hide is called.
Maybe I can just put this function as part of `hide()` entirely.

anthonykim1: I thought debounce had `clear` or `.cancel` inside, but doesn't seem like it does https://github.com/microsoft/vscode/blob/2226318b9c664caa63f7d17bed487c50b9e38d7f/src/vs/base/common/decorators.ts#L65 

This is weird that it works though.. 


NVM it seemed working because it was crashing: Screenshot 2025-09-29 at 11 51 43 AM

src/vs/workbench/contrib/terminalContrib/stickyScroll/browser/terminalStickyScrollOverlay.ts
```diff
@@ -227,8 +227,8 @@ export class TerminalStickyScrollOverlay extends Disposable {
 		// scroll.
 		this._currentStickyCommand = undefined;
 
-		// No command
-		if (!command) {
+		// No command or clear command
+		if (!command || (command.command && this._isClearCommand(command.command))) {
```
Tyriar: Bit nicer imo to take the command object in `_isClearCommand`:

```
if (!command || this._isClearCommand(command)) {
```

Tyriar (changes requested):

src/vs/workbench/contrib/terminalContrib/stickyScroll/browser/terminalStickyScrollOverlay.ts
anthonykim1: Facts!! 
Here is the change: [565da26](https://github.com/microsoft/vscode/pull/268921/commits/565da26a75fa5ec76a1e34d34b338486a5d0b6ab)

Tyriar (approved):

PR #268937, opened 2025-09-29
Bump the version of tar-fs dependency to 2.1.4 to address CVE-2025-59343

dmitrivMS:


build/package.json
```diff
@@ -75,6 +75,7 @@
   "overrides": {
     "path-scurry": {
       "lru-cache": "11.2.1"
-    }
+    },
+    "tar-fs": "2.1.4"
```
joaomoreno, line 79: @dmitrivMS This is usually a bad idea. I see dependabot got there before you, which is good. In 99% of the cases, dependabot will just fix this for us, so always be on the lookout for those PRs. If dependabot somehow fails it's almost always because (1) there isn't a fix out yet or (2) the bump would be incompatible with our dependency declarations.

dmitrivMS: @joaomoreno Got it. I like dependabot as well, it just wasn't able to do its job in my previous team due to restrictions, so it's nice to have it back :-)

Issue #269006, opened 2025-09-30
Tasks leaks terminals


Tyriar:
Tasks ends up holding on to terminal instances indefinitely, leaking a bunch of memory.

Examples:

This listener is disposed if the terminal is disposed, but a reference to it is kept indefinitely as it's not deleted via `this._store.delete`:

https://github.com/microsoft/vscode/blob/50e37a458d877100ed90fe48f0319b88a8effd5a/src/vs/workbench/contrib/tasks/browser/terminalTaskSystem.ts#L1566-L1569

This type of thing happens in several places where the terminal is registered against `TerminalTaskSystem` but never released:

https://github.com/microsoft/vscode/blob/50e37a458d877100ed90fe48f0319b88a8effd5a/src/vs/workbench/contrib/tasks/browser/terminalTaskSystem.ts#L1410

Needs a proper audit by capturing heap snapshots to chase down all the leaks.

PR #269007, opened 2025-09-30
Refactor terminal reconnection to use instances filter

### LINKED ISSUE #268995: Move reconnected terminal tracking into tasks
This part of ITerminalService is only used by tasks: https://github.com/microsoft/vscode/blob/761fbb0ee2c430963e335a2c04113bf276f898d2/src/vs/workbench/contrib/terminal/browser/terminalService.ts#L105-L108

It should move into tasks via an event instead of being a persistent API. There may also be a leak here since referenced to the terminal instances are held onto.

Tyriar:
- Remove deprecated getReconnectedTerminals method usage
- Use instances.filter with reconnectionProperties check
- Clean up _addToReconnected private method
- Simplify terminal reconnection logic in task system
- Add deprecation notice to getReconnectedTerminals

Fixes #268995

---

This removed the possibility of a leak in terminal land and simplifies `ITerminalService`, there are remaining memory leak issues in task land though https://github.com/microsoft/vscode/issues/269006


meganrogge (approved):

meganrogge:
Thanks!

PR #269027, opened 2025-09-30
fix: memory leak in chat session tracker

SimonSiefke:
Fixes a memory leak in the chat session tracker. 

### Overview

[
Untitled
](https://simonsiefke.github.io/vscode-memory-leak-finder/named-function-count-3)

The chat session tracker registers a listener to a group each time a group is added. But when a group is removed, the listener doesn't seem to be disposed. 

```ts
// Listen for new groups
this._register(this.editorGroupsService.onDidAddGroup(group => {
  this.registerGroupListeners(group);
}));
```

### Change

The change is to also listen for group removal events and dispose the group change listener for the group when the group is removed:

```ts
// Listen to all editor groups
this.editorGroupsService.groups.forEach(group => {
	this.groupDisposables.set(group.id, this.registerGroupListeners(group));
});

// Listen for new groups
this._register(this.editorGroupsService.onDidAddGroup(group => {
	this.groupDisposables.set(group.id, this.registerGroupListeners(group));
}));

// listen for groups to be removed
this.editorGroupsService.onDidRemoveGroup(group => {
	this.groupDisposables.deleteAndDispose(group.id); // remove listener
});
```

### Results after change

After the change there seem to be no more chat session tracker leaks.  

editor split-down


src/vs/workbench/contrib/chat/browser/chatSessions/chatSessionTracker.ts
```diff
@@ -31,17 +33,20 @@ export class ChatSessionTracker extends Disposable {
...
-			this.registerGroupListeners(group);
+			this.groupDisposables.set(group.id, this.registerGroupListeners(group));
 		});
 
 		// Listen for new groups
 		this._register(this.editorGroupsService.onDidAddGroup(group => {
-			this.registerGroupListeners(group);
+			this.groupDisposables.set(group.id, this.registerGroupListeners(group));
 		}));
+		this.editorGroupsService.onDidRemoveGroup(group => {
```
osortega: By not using `this._register` we are losing a lot of disposing mechanisms inherited from the `Disposable` class.
In other words, now we will only dispose when `onDidRemoveGroup` fires and not in any other cases of the lifecycle of a disposable object

osortega (changes requested):

src/vs/workbench/contrib/chat/browser/chatSessions/chatSessionTracker.ts
alexr00: It looks like this new approach is better. Here's what it looks like was happening before:
- Editor group added
- we listen on model changes
- repeat for every added editor group.

At no point did we dispose the listener when the editor group was removed.

With this change, the model change listener is disposed when the group is removed ✅

With this line, we also ensure that all the model change listners are disposed when `this` disposable object is disposed:
https://github.com/SimonSiefke/vscode/blob/a386d75033ea79c42ef55ed74d771df985a57645/src/vs/workbench/contrib/chat/browser/chatSessions/chatSessionTracker.ts#L21-L21

alexr00 (dismissed):

Tyriar (dismissed):

src/vs/workbench/contrib/chat/browser/chatSessions/chatSessionTracker.ts
```diff
@@ -31,17 +33,20 @@ export class ChatSessionTracker extends Disposable {
...
+			this.groupDisposables.deleteAndDispose(group.id);
+		});
 	}
 
-	private registerGroupListeners(group: IEditorGroup): void {
-		this._register(group.onDidModelChange(e => {
+	private registerGroupListeners(group: IEditorGroup): IDisposable {
```
Tyriar: nit: this is a bad name as it doesn't register

src/vs/workbench/contrib/chat/browser/chatSessions/chatSessionTracker.ts
osortega: I missed that line in the chatSessionTracker, you are right it will be disposed with the object

osortega (dismissed):

alexr00 (approved):

src/vs/workbench/contrib/chat/browser/chatSessions/chatSessionTracker.ts
Tyriar: I'm not sure I understand this. We still need to register the `onDidRemoveGroup` listener in the case where `ChatSessionTracker` gets disposed? I expect a unit test on this class would fail due to the `onDidRemoveGroup` disposable leaking?

Tyriar (changes requested):

src/vs/workbench/contrib/chat/browser/chatSessions/chatSessionTracker.ts
alexr00: Nice catch.

alexr00 (changes requested):

src/vs/workbench/contrib/chat/browser/chatSessions/chatSessionTracker.ts
SimonSiefke: Registered the `onDidRemoveGroup` listener!

Tyriar (dismissed):

alexr00 (dismissed):

bpasero:
Merge conflict.

alexr00 (dismissed):

alexr00 (approved):

osortega (approved):

PR #269126, opened 2025-09-30
fix: properly update cloned stylesheets on mutation in firefox

### LINKED ISSUE #259835: Firefox: changes to global stylesheets are not applied live

CGNonofr:
fix #259835

It looks like the solution was quite simple at the end @bpasero 

Tested with firefox 143

### DIFF
--- a/src/vs/base/browser/domStylesheets.ts
+++ b/src/vs/base/browser/domStylesheets.ts
@@ -5,6 +5,7 @@
 
 import { DisposableStore, toDisposable, IDisposable } from '../common/lifecycle.js';
 import { autorun, IObservable } from '../common/observable.js';
+import { isFirefox } from './browser.js';
 import { getWindows, sharedMutationObserver } from './dom.js';
 import { mainWindow } from './window.js';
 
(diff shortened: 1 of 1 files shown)


bpasero:
@CGNonofr very nice (have to confirm it works), can we make this change with `isFirefox` only?

CGNonofr:
> @CGNonofr very nice (have to confirm it works), can we make this change with `isFirefox` only?

I have the feeling that characterData is the mutation we are looking for, and that childList is the one webkit-specific :thinking: What do you think?

bpasero:
@CGNonofr I actually tested this with a full insiders build and it turns out I only get it to work with both of them. Can you change it to this:

```ts
subtree: isFirefox, characterData: isFirefox
```

And update the PR?

bpasero (approved):
Thanks a lot!

benibenj (approved):

PR #269385, opened 2025-10-01
use `MutableDisposable` to manage `startMarker` in terminal execute strategies

### LINKED ISSUE #269240: If marker is disposed but still in the store then it's still being leaked
If marker is disposed but still in the store then it's still being leaked

_Originally posted by @roblourens in https://github.com/microsoft/vscode/pull/268555#discussion_r2392940599_

meganrogge:
fixes #269240

cc @tyriar

### DIFF
--- a/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/executeStrategy/noneExecuteStrategy.ts
+++ b/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/executeStrategy/noneExecuteStrategy.ts
@@ -6,7 +6,7 @@
 import type { CancellationToken } from '../../../../../../base/common/cancellation.js';
 import { CancellationError } from '../../../../../../base/common/errors.js';
 import { Emitter, Event } from '../../../../../../base/common/event.js';
-import { DisposableStore } from '../../../../../../base/common/lifecycle.js';
+import { DisposableStore, MutableDisposable, toDisposable, type IDisposable } from '../../../../../../base/common/lifecycle.js';
 import { ITerminalLogService } from '../../../../../../platform/terminal/common/terminal.js';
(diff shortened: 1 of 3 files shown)


lramos15 (approved):

roblourens (commented):
LGTM

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/executeStrategy/noneExecuteStrategy.ts
```diff
@@ -54,13 +54,31 @@ export class NoneExecuteStrategy implements ITerminalExecuteStrategy {
...
 			// like powerlevel10k's transient prompt
-			this._onDidCreateStartMarker.fire(this._startMarker = store.add(xterm.raw.registerMarker()));
-			store.add(this._startMarker.onDispose(() => {
-				this._log(`Start marker was disposed, recreating`);
-				this._onDidCreateStartMarker.fire(this._startMarker = store.add(xterm.raw.registerMarker()));
+			const markerListener = new MutableDisposable<IDisposable>();
+			const recreateStartMarker = () => {
+				if (store.isDisposed) {
+					return;
+				}
```
Tyriar, line 64: Lots of repetition here across files here, should this live in an executeStrategy/strategyHelpers.ts so they don't get out of sync?

meganrogge: I thought about that. But since we fire a class event from within `recreateStartMarker`, wasn't sure how that would work 🤔  and worried it would just complicate things. 

I also worried about breaking it right before the release, but can investigate de-duplicating for October

PR #269513, opened 2025-10-02
Enhance action label styling for high-contrast themes

mrleemurray:
Improve the styling of action labels in high-contrast themes by adding outlines for non-uri-icon elements and adjusting the wrapper for uri-icons to prevent masking issues.


mrleemurray:
Sounds goo. The summary is that currently the 3rd party extension icons are svg, masks are used to control the background-color property. This masking then hides the outline styles in high-contrast themes. After some experimentation with updating the parent container, ::before & ::after psuedo-elements, the most effective approach to was to add an additional wrapper container around `.uri-icon` elements. 

It is complicated & I would love some guidance on this

alexr00 (approved):

bpasero:
This was PENDING a review from selected people.

PR #269527, opened 2025-10-02
Clarify TypeScript version picker UI when workspace version is manually configured

magliocchetti:
## Summary

Fixes #263226 

Improves the "TypeScript: Select TypeScript Version..." UI by adding clarifying text when there's a mismatch between the user's explicit selection and the currently active TypeScript version.

## Problem

When users manually set `"typescript.tsdk": "node_modules/typescript/lib"` in their workspace settings (without using the version picker), the UI was confusing:

- The "Use VS Code's Version" item showed a bullet (•) indicating it was selected
- However, the workspace TypeScript version was actually active and running
- There was no indication that a workspace version was currently in use

This caused confusion about which TypeScript version was being used.

## Root Cause

The version picker UI was designed to show which version type the user has **explicitly selected** through the picker (tracked by workspace state), not which version is **currently active**. When users manually configure `typescript.tsdk` in settings.json:

- The workspace state flag remains in "bundled selected" state
- But the workspace TypeScript version becomes active
- The UI showed no indication of this mismatch

## Solution

Following maintainer feedback, this PR clarifies the UI rather than changing the selection indicator logic:

- The bullet (•) continues to indicate the user's explicit selection (maintains existing behavior)
- **NEW:** Adds clarifying text when there's a mismatch between selection and active version

### Specific Changes

1. **Workspace version active but not selected** (the reported issue):
   - Description now shows: `"5.9.1 (Currently Active)"`
   - Makes it clear the workspace version is running even though not explicitly selected

2. **Bundled version selected but different version active** (rare edge case):
   - Description now shows: `"5.8.3 (Currently Active: 5.9.1)"`
   - Clarifies when there's an unexpected mismatch

### Before (confusing)
```
• Use VS Code's Version          5.8.3          /path/to/bundled
  Use Workspace Version          5.9.1          node_modules/typescript/lib
```
No indication that workspace version is actually active!

### After (clear)
```
• Use VS Code's Version          5.8.3          /path/to/bundled
  Use Workspace Version          5.9.1 (Currently Active)          node_modules/typescript/lib
```
Now clearly shows workspace version is active.

## Impact

- ✅ Clarifies which TypeScript version is currently active
- ✅ Maintains existing UI convention (bullet = explicit selection)
- ✅ Backward compatible - no breaking changes
- ✅ Addresses user confusion reported in #263226
- ✅ Clearer distinction between "selected" vs "active"

## Testing

Manually tested scenarios:
1. User manually sets `typescript.tsdk` → workspace version shows "(Currently Active)"
2. User selects workspace through picker → workspace version shows bullet only
3. User selects bundled through picker → bundled version shows bullet only
4. Edge case with version mismatch → shows both versions in description

## Files Changed

- `extensions/typescript-language-features/src/tsServer/versionManager.ts`

### DIFF
--- a/extensions/typescript-language-features/src/tsServer/versionManager.ts
+++ b/extensions/typescript-language-features/src/tsServer/versionManager.ts
@@ -96,7 +96,7 @@ export class TypeScriptVersionManager extends Disposable {
 	private getBundledPickItem(): QuickPickItem {
 		const bundledVersion = this.versionProvider.defaultVersion;
 		return {
-			label: (!this.useWorkspaceTsdkSetting || !vscode.workspace.isTrusted
+			label: (this.currentVersion.eq(bundledVersion)
 				? '• '
 				: '') + vscode.l10n.t("Use VS Code's Version"),
 			description: bundledVersion.displayName,
@@ -111,7 +111,7 @@ export class TypeScriptVersionManager extends Disposable {
 	private getLocalPickItems(): QuickPickItem[] {
 		return this.versionProvider.localVersions.map(version => {
 			return {
-				label: (this.useWorkspaceTsdkSetting && vscode.workspace.isTrusted && this.currentVersion.eq(version)
+				label: (this.currentVersion.eq(version)
 					? '• '
 					: '') + vscode.l10n.t("Use Workspace Version"),
 				description: version.displayName,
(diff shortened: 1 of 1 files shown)


mjbvz (commented):
The current UI is confusing but it was actually designed to work like this. The most important info in this UI if the workspace or bundled TS version is being used

Instead of strictly checking the version, I think we should clarify the UI:

- Still show which type of TS the user has selected (workspace vs bundled)
- For cases where there is a mismatch between the active version and provided version, show a more detailed message. Something like `Workspace version (Currently Active: 5.9  Found: 5.7)`

PR #269528, opened 2025-10-02
Add copy context menu for chat math equations

### LINKED ISSUE #258286: Add copy button for chat math equations
> I would suggest to add interactive copy-latex-source feature from outputted equations, like in MathJax:
> 
> Image 

 _Originally posted by @hamad12a in [#255243](https://github.com/microsoft/vscode/issues/255243#issuecomment-3122099815)_

magliocchetti:
Fixes #258286

### Summary
Adds a right-click context menu to math equations in chat responses, allowing users to copy the LaTeX source. This implements a simplified version of the MathJax "Copy to Clipboard" feature requested in the issue.

#### Modified Files
1. **`src/vs/workbench/contrib/chat/browser/chatContentParts/chatMarkdownContentPart.ts`**
   - Added `IClipboardService` and `IContextMenuService` to constructor
   - Implemented `setupKatexContextMenus()` method to register context menu on KaTeX elements
   - Extracts LaTeX source from KaTeX's `<annotation encoding="application/x-tex">` element
   - Single context menu option: "Copy Math Source"

2. **`src/vs/workbench/contrib/chat/browser/chatContentParts/media/chatMarkdownPart.css`**
   - Added `.interactive-katex` styles with hover effect
   - Added `cursor: context-menu` for visual feedback
   - Subtle background highlight on hover

### Features

✅ Copy LaTeX source (raw LaTeX without delimiters)  
✅ Visual feedback on hover (background highlight + cursor change)  
✅ Works with both inline (`$...$`) and block (`$$...$$`) equations  
✅ Proper lifecycle management with disposables  

### User Experience

1. User hovers over a math equation in chat → subtle highlight appears with context-menu cursor
2. User right-clicks → "Copy Math Source" option appears in context menu
3. User clicks → raw LaTeX copied to clipboard (e.g., `\frac{1}{2}`)

### Example
- **Rendered equation**: $\frac{1}{2}$
- **Right-click → Copy Math Source**
- **Clipboard**: `\frac{1}{2}`

#### Simplified Approach
Started with a minimal single-option implementation as suggested by @mjbvz, since the original issue hasn't received much attention. This can be expanded based on user feedback to add:
- Additional copy formats (MathML, TeX with delimiters)
- Keyboard shortcuts
- Visible copy button on hover

#### Proper Separation of Concerns
- Styling via CSS classes (`.interactive-katex`) instead of inline styles
- Uses existing VS Code services for consistency
- Localized strings with `nls.localize()`

### Testing

To test:
1. Build VS Code with these changes
2. Open a chat session
3. Request math equations (e.g., "Show me the quadratic formula")
4. Hover over any rendered equation - verify highlight and cursor change
5. Right-click on equation - verify "Copy Math Source" menu item appears
6. Click the option - verify raw LaTeX (without delimiters) is copied to clipboard
7. Test with both inline and block equations

### DIFF
--- a/src/vs/workbench/contrib/chat/browser/chatContentParts/chatMarkdownContentPart.ts
+++ b/src/vs/workbench/contrib/chat/browser/chatContentParts/chatMarkdownContentPart.ts
@@ -340,81 +340,36 @@ export class ChatMarkdownContentPart extends Disposable implements IChatContentP
 				continue;
 			}
 
-			// Add hover style to indicate it's interactive
-			katexElement.style.cursor = 'context-menu';
+			// Mark element as having math content for CSS styling
+			katexElement.classList.add('interactive-katex');
 
 			disposables.add(dom.addDisposableListener(katexElement, dom.EventType.CONTEXT_MENU, (domEvent) => {
 				const event = new StandardMouseEvent(dom.getWindow(domEvent), domEvent);
-				dom.EventHelper.stop(domEvent, true);
+				
+				// Extract LaTeX source from the annotation element
+				const annotation = katexElement.querySelector('annotation[encoding="application/x-tex"]');
+				const latexSource = annotation?.textContent;
 
-				this.contextMenuService.showContextMenu({
-					getAnchor: () => event,
-					getActions: () => {
-						const actions = [];
-
-						// Extract LaTeX source from the annotation element
-						const latexSource = this.extractLatexSource(katexElement);
-						if (latexSource) {
-							actions.push({
-								id: 'chat.copyLatexSource',
-								label: localize('chat.copyLatexSource', "Copy LaTeX Source"),
+				if (latexSource) {
+					// Show context menu with copy math option
+					this.contextMenuService.showContextMenu({
+						getAnchor: () => event,
+						getActions: () => {
+							return [{
+								id: 'chat.copyMathSource',
+								label: localize('chat.copyMathSource', "Copy Math Source"),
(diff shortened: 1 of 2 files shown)


src/vs/workbench/contrib/chat/browser/chatContentParts/chatMarkdownContentPart.ts
```diff
@@ -324,6 +330,93 @@ export class ChatMarkdownContentPart extends Disposable implements IChatContentP
...
+		const katexElements = this.domNode.querySelectorAll('.katex');
+		for (const katexElement of katexElements) {
+			if (!dom.isHTMLElement(katexElement)) {
+				continue;
+			}
+
+			// Add hover style to indicate it's interactive
+			katexElement.style.cursor = 'context-menu';
+
+			disposables.add(dom.addDisposableListener(katexElement, dom.EventType.CONTEXT_MENU, (domEvent) => {
```
mjbvz, line 346: Let's add this as an extra option in the existing context menu instead of overriding the context menu completely

mjbvz (commented):
The original issue hasn't gotten much attention so let's go with a simple implementation until people ask for more: just a single `copy math source` context menu that copies the original markup inside of the dollar signs

dmitrivMS:
The underlying custom-editor stale-cache bug is already fixed on main by merged PR [#287966](https://github.com/microsoft/vscode/pull/287966) using a different implementation that disposes models on delete.

PR #269533, opened 2025-10-02
Format Selection now honors configured tab size

### LINKED ISSUE #250520: Tab Size is not honored by Control K + F

- VS Code Version: 
- OS Version: 

Steps to Reproduce:

1. Control K + F
When you press control K + F, it does not honor your tab size at all. My tab size (default) is set to 4, which is great and works like normal. But when I press control K + F, all my tab sizes move down to 2 in the code, but its still four if I were to tab manually in my code.

magliocchetti:
Fixes #250520

## Problem
When users press **Ctrl+K Ctrl+F** (Format Selection command), the formatter does not honor the configured `editor.tabSize` setting. For example, if a user has `tabSize` set to 4, the formatter might use 2 spaces instead, causing unexpected and inconsistent indentation.

## Root Cause
The bug was in the `getFormattingOptions()` method in `src/vs/editor/common/model/textModel.ts`. This method returns the formatting options that are passed to all document formatters.

The method was incorrectly returning `indentSize` for the `tabSize` property:
```typescript
// BEFORE (incorrect)
public getFormattingOptions(): FormattingOptions {
    return {
        tabSize: this._options.indentSize,  // ❌ Wrong!
        insertSpaces: this._options.insertSpaces
    };
}
```

## Solution
Updated `getFormattingOptions()` to return the correct `tabSize`:
```typescript
// AFTER (correct)
public getFormattingOptions(): FormattingOptions {
    return {
        tabSize: this._options.tabSize,  // ✅ Correct!
        insertSpaces: this._options.insertSpaces
    };
}
```

## Impact
This fix ensures that:
- ✅ Format Selection (Ctrl+K Ctrl+F) respects the configured `editor.tabSize` setting
- ✅ Format Document also uses the correct tab size
- ✅ All document formatters receive the correct tab size value
- ✅ Behavior is consistent whether `indentSize` is set to `'tabSize'` (default) or a custom value

## Testing
- No compilation errors after the change
- The fix aligns with the TypeScript interface definition for `FormattingOptions`
- All existing usages of `getFormattingOptions()` will now receive the correct tab size value

## Related
- Issue: #250520
- Labels: `bug`, `editor-autoindent`, `help wanted`

### DIFF
--- a/src/vs/editor/common/model/textModel.ts
+++ b/src/vs/editor/common/model/textModel.ts
@@ -655,7 +655,7 @@ export class TextModel extends Disposable implements model.ITextModel, IDecorati
 
 	public getFormattingOptions(): FormattingOptions {
 		return {
-			tabSize: this._options.indentSize,
+			tabSize: this._options.tabSize,
 			insertSpaces: this._options.insertSpaces
 		};
 	}


aiday-mar:
The indent size is defined as follows in the settings:

The number of spaces used for indentation or "tabSize" to use the value from `Editor: Tab Size`. This setting is overridden based on the file contents when `Editor: Detect Indentation` is on.

The formatting options will default to the detected indentation instead of the tab size. I think it is expected that we use the detected indentation. This PR will be closed. I commented on the associated issue, and asked the user to provide a sample code where they see this issue.

Issue #268782, opened 2025-09-28
run_task didn't wait for vscode build tasks to complete


roblourens:
- Agent ran for a bit, then did a `run_task` on `VS Code - Build`
- The task hadn't previously been running
- It looked like it only waited for a few seconds, then `run_task` resolved before it actually finished compiling everything
- There weren't any real errors but if there had been I guess it would have missed them

Main agent request
[16ba08f2.copilotmd](https://github.com/user-attachments/files/22585838/16ba08f2.copilotmd)

requests from the tool call

[96d757d3.copilotmd](https://github.com/user-attachments/files/22585842/96d757d3.copilotmd)
[dbeafdb9.copilotmd](https://github.com/user-attachments/files/22585841/dbeafdb9.copilotmd)
[abc9dd8b.copilotmd](https://github.com/user-attachments/files/22585844/abc9dd8b.copilotmd)
[2dc7aa36.copilotmd](https://github.com/user-attachments/files/22585843/2dc7aa36.copilotmd)

I guess it's confused by the fact that the task includes multiple smaller tasks so there are statements like `Finished compilation api-proposal-names with 0 errors` while the main task is still running?

And side note, I don't know why it felt the need to kill the task, but maybe we can tell to not do that
<img width="474" height="350" alt="Image" src="https://github.com/user-attachments/assets/b18b934c-b85d-4b29-b211-d6eb669d8ef7" />

meganrogge:
This check is meant to prevent such a thing from happening https://github.com/microsoft/vscode/blob/86369e5f21d6537b9fee337ef7a4a90527ada86f/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/monitoring/outputMonitor.ts#L320

I wonder why that wasn't hit here...

meganrogge:
I can reproduce though with that model.

meganrogge:
@roblourens can you repro? It seems to be working for me - as long as extensions are loaded, so problems are actually detected. Pls reopen if so.

roblourens:
The same thing still reproes. Exact steps

- vscode repo
- kill the build tasks
- (in chatInputPart.ts) `Rename _currentLanguageModel to _currLanguageModel, then run the build task to confirm`
- It runs the VS Code - Build task, seems to wait like 10 seconds then continues before the task has finished

[afaf334c.copilotmd](https://github.com/user-attachments/files/23373937/afaf334c.copilotmd)

[...]

PR #269568, opened 2025-10-02
for tasks with problem matchers, don't pass back output, only succinct reply

meganrogge:
fixes #269563

Discovered while investigating https://github.com/microsoft/vscode/issues/268782.

For tasks with problem matchers, we were still providing the task output, which could confuse the agent and lead it to mention earlier compilation problems when we know the exact number of problems. Now, for tasks with problem matchers, we only return the exact problems or that no problems were detected.

I thought by including the problems in `resources`, those were provided to the agent, but learned that's only for the UI 🤦🏼‍♀️ .

Before:

image

image

After:

image

Screenshot 2025-10-02 at 12 59 20 PM

Screenshot 2025-10-02 at 12 59 49 PM


dmitrivMS (approved):

meganrogge:
cc @Tyriar

PR #269596, opened 2025-10-02
Put terminal voice actions in separate group

### LINKED ISSUE #269587: Terminal menu would benefit from separators
Why are entries like "Start dictation" directly below "Run Selected Text":

Image

Menus have separators to provide meaningful grouping of entries that belong together. Why is that not applied here?

meganrogge:
fix #269587

Screenshot 2025-10-02 at 3 03 01 PM

### DIFF
--- a/src/vs/workbench/contrib/terminalContrib/voice/browser/terminalVoiceActions.ts
+++ b/src/vs/workbench/contrib/terminalContrib/voice/browser/terminalVoiceActions.ts
@@ -79,7 +79,6 @@ export function registerTerminalVoiceActions() {
 			{
 				id: MenuId.ViewTitle,
 				group: 'voice',
-				order: 9,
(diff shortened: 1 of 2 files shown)


src/vs/workbench/contrib/terminalContrib/voice/browser/terminalVoiceActions.ts
```diff
@@ -73,7 +74,16 @@ export function registerTerminalVoiceActions() {
 				}, learnMoreAction);
 			}
 			notificationService.notify({ severity: Severity.Info, message, actions: { primary: actions } });
-		}
+		},
+		menu: [
+			{
+				id: MenuId.ViewTitle,
+				group: 'voice',
+				order: 9,
```
meganrogge:
```suggestion
```

joshspicer (approved):

PR #269647, opened 2025-10-03
Remove .monaco-workbench from hover used in monaco

### LINKED ISSUE #4993: [Bug] [0.5.3.0] Rare "hints" in the search widget textarea inputs
### Reproducible in vscode.dev or in VS Code Desktop?

- [x] Not reproducible in [vscode.dev](https://vscode.dev) or VS Code Desktop

### Reproducible in the monaco editor playground?

- [ ] Not reproducible in [the monaco editor playground](https://microsoft.github.io/monaco-editor/playground.html)

Hello there!

First of all, thanks very much for this awesome project! 

As you can see in the below GIF, there is some "rare hints" when the mouse is over the textarea inputs of the search widget. This only happen in v0.53.0, I mean, not in previous versions.

Image

### Reproduction Steps

Just any sample at the playground with v0.53.0, like:

https://microsoft.github.io/monaco-editor/playground.html?source=v0.53.0#example-creating-the-editor-hello-world

### Actual (Problematic) Behavior

The rare hints which appears in the textarea inputs of the search widgets while the mouse hover it.

### Expected Behavior

No hints at all, like in previous versions?

Tyriar:
Fixes microsoft/monaco-editor#4993

@benibenj do you know how to verify this in the monaco playground?


bpasero (changes requested):
There is more usages:

https://github.com/microsoft/vscode/blob/782bd965b4ba09eb50ed1ac3e26f9b15b552926a/src/vs/editor/browser/services/hoverService/hoverService.ts#L606-L607

src/vs/editor/browser/services/hoverService/hover.css
```diff
@@ -3,7 +3,7 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-.monaco-workbench .workbench-hover {
+.monaco-hover.workbench-hover {
```
bpasero, line 8: Why do you sometimes not prefix with `.monaco-hover` in this file, isn't it all contained within?

bpasero (changes requested):

src/vs/editor/browser/services/hoverService/hover.css
Tyriar: Looks like I missed one unintentionally. For `.workbench-hover-container` that is the container of `.monaco-hover` so it cannot be the prefix.

bpasero (dismissed):
I would not expect `workbench` to be used anywhere in the `editor` layer, so please change the container to something different like `monaco-hover-container`.

Tyriar:
> I would not expect workbench to be used anywhere in the editor layer, so please change the container to something different like monaco-hover-container.

@bpasero the feature is called workbench hover, changing this is too big and risky of a change for the time I have available for this. This PR was meant to fix a bug in monaco, not clean up debt - you're being too strict in blocking this.

bpasero (approved):
Fair, the main task was to get rid of `.monaco-workbench` which this PR seems to address.

To clarify: I was just asking to find a better name for `workbench-hover` and then do a string+replace operation. Its not used a lot, basically just in 2 places.

PR #269732, opened 2025-10-03
prevent right prompt from being considered ghost text

meganrogge:
fix #269730

Before:

Screenshot 2025-10-03 at 12 24 07 PM
730

After:

Screenshot 2025-10-03 at 12 30 37 PM


src/vs/platform/terminal/common/capabilities/commandDetection/promptInputModel.ts
```diff
@@ -546,14 +546,20 @@ export class PromptInputModel extends Disposable implements IPromptInputModel {
...
+			}
 			// Ensure these positions are contiguous
 			for (let i = 1; i < positionsWithGhostStyle.length; i++) {
 				if (positionsWithGhostStyle[i] !== positionsWithGhostStyle[i - 1] + 1) {
 					// Discontinuous styles, so may be syntax highlighting vs ghost text
 					return -1;
 				}
 			}
-			// Calculate the ghost text start index
+			// Calculate the ghost text start index relative to command start.
```
meganrogge:
```suggestion
			// Calculate the ghost text start index
```

meganrogge:
Nice, test caught an issue! When a user navigates with their cursor, with this change, the suggest widget hides.

eleanorjboyd (approved):

src/vs/platform/terminal/test/common/capabilities/commandDetection/promptInputModel.test.ts
```diff
@@ -215,9 +215,6 @@ suite('PromptInputModel', () => {
 
 			await writePromise('foo\x1b[38;2;255;0;0m bar\x1b[0m\x1b[4D');
 			await assertPromptInput('foo|[ bar]');
-
-			await writePromise('\x1b[2D');
-			await assertPromptInput('f|oo[ bar]');
```
Tyriar, line 220: Isn't this an important test?

src/vs/platform/terminal/common/capabilities/commandDetection/promptInputModel.ts
```diff
@@ -546,6 +546,11 @@ export class PromptInputModel extends Disposable implements IPromptInputModel {
 		// Retrieve the positions of all cells with the same style as `lastNonWhitespaceCell`
 		const positionsWithGhostStyle = styleMap.get(this._getCellStyleAsString(lastNonWhitespaceCell));
 		if (positionsWithGhostStyle) {
+			// Ghost text must start at the cursor or one char after (e.g. a space),
+			// preventing right-prompt styles from being misdetected as ghost text.
+			if (positionsWithGhostStyle[0] > buffer.cursorX + 1) {
+				return -1;
+			}
```
Tyriar, line 553: You should detect right problem by looking for null characters between the prompt and the right side. Or by looking for several whitespace characters between the prompt and right prompt, depends how they are printed exactly.

src/vs/workbench/contrib/terminal/test/browser/xterm/recordings/basic/macos_zsh_p10k_ls_one_time.ts
```diff
@@ -35,7 +35,7 @@ export const events = [
 	},
 	{
 		"type": "promptInputChange",
-		"data": "|                                                                                                     [─╯]"
```
Tyriar, line 38: Doesn't this mean the right prompt issue wasn't fixed?

meganrogge: Before, we were detecting that right prompt as ghost text. Now we do not, thus updating this test

Image

src/vs/platform/terminal/common/capabilities/commandDetection/promptInputModel.ts
meganrogge: This seems to work ... can you think of a case where it wouldn't?

PR #269746, opened 2025-10-03
Don't try to save when reverting chat edits in untitled documents

### LINKED ISSUE #253116: undo chat edit opens a save dialog for untitled notebooks

- VS Code Version: 
- OS Version: 

Steps to Reproduce:

1. create new notebook
2. ask copilot edit/agent to change something in a cell
3. select `undo` to reject the changes

:bug: save dialog pops up

kycutler:
Fixes #253116 for both notebooks and text files.

The fix is slightly different for each:
- For text documents, fix the `isTextFileEditorModel` helper so it does not (incorrectly) match untitled models.
- For notebooks, add a check for untitled schemes -- this aligns with the behavior for accepting notebook edits.


src/vs/workbench/services/textfile/common/textfiles.ts
```diff
@@ -566,7 +566,16 @@ export interface ITextFileEditorModel extends ITextEditorModel, IEncodingSupport
...
 
-	return areFunctions(candidate.setEncoding, candidate.getEncoding, candidate.save, candidate.revert, candidate.isDirty, candidate.getLanguageId);
+	return areFunctions(
+		candidate.setEncoding,
+		candidate.getEncoding,
+		candidate.save,
+		candidate.revert,
+		candidate.isDirty,
+		candidate.getLanguageId,
+		candidate.hasState,
```
bpasero, line 576: Can you elaborate why this change is needed?

src/vs/workbench/services/textfile/common/textfiles.ts
```diff
@@ -566,7 +566,16 @@ export interface ITextFileEditorModel extends ITextEditorModel, IEncodingSupport
 export function isTextFileEditorModel(model: ITextEditorModel): model is ITextFileEditorModel {
```
amunger, line 566: it looks like this will actually cause an untitled file that was created to be left alone: https://insiders.vscode.dev/github/microsoft/vscode/blob/main/src/vs/workbench/contrib/chat/browser/chatEditing/chatEditingModifiedDocumentEntry.ts#L266

src/vs/workbench/services/textfile/common/textfiles.ts
kycutler: `IUntitledTextEditorModel` also has the other methods here, so this would return true for untitled models.

Edit rejection checks for this before saving, so would previously match untitled editors and trigger a save dialog:

https://github.com/microsoft/vscode/blob/cb76cf01f2b7163fc4ab5b2095fd6c8c1a4f5db2/src/vs/workbench/contrib/chat/browser/chatEditing/chatEditingModifiedDocumentEntry.ts#L269-L273

src/vs/workbench/services/textfile/common/textfiles.ts
amunger: looking at this again, `UntitledTextEditorModel` inherits from `BaseTextEditorModel`, so being untitled probably shouldn't make it return false.
We might need to just directly prevent saving for untitled undo's in the reject method in the [chat entry](https://insiders.vscode.dev/github/microsoft/vscode/blob/main/src/vs/workbench/contrib/chat/browser/chatEditing/chatEditingModifiedDocumentEntry.ts#L264) . (Which is Joh's area)

src/vs/workbench/services/textfile/common/textfiles.ts
bpasero: Yeah true, thanks for clarifying.

src/vs/workbench/services/textfile/common/textfiles.ts
bpasero: But this method is for testing for `ITextFileEditorModel` and that is exclusive for file based text models, not untitled.

kycutler: Updated to ensure created untitled files are also reverted.

Issue #269213, opened 2025-09-30
Clean up `as any` type assertions in the codebase


mjbvz:
We've seen a few build breaks caused by `as any` type assertions in our codebase. This can hide errors and break after mangling. Some language models seem to be specially keen on adding `as any` assertions to work around problems. This can break our mangling and results in more fragile code

We'd like to remove these `as any` cast in as many 

# Plan
The first priority is preventing new code from using `as any`. We'll then start addressing existing usages. Once complete we can also consider adding even stricter rules for type assertions 

- [x] @mjbvz add a new eslint rule for a `as any` and `<any>`.
- [x] @mjbvz Will add suppression comments for all existing breakers. There are around 900 or so
- [ ] 🏃   @mjbvz Will update this issue with a list of all files that have breaks
- [ ]  🏃 Assigned team member review  and fix these errors during debt week

Once complete:

- Evaluate if we also want to enabled the `@typescript-eslint/no-unsafe-type-assertion` option for even stricter rules

# Fixing `as any`
To find places in your code that need fixing, just search for: `// eslint-disable-next-line local/code-no-any-casts`

Here are some general patterns you can use to replace `as any` in code:

### Use real type guards 
This test code was using `as any` to simplify writing the test:

```ts
const tokens = store.root as any;
assert.strictEqual(tokens.children[0].token, 1);
```

Ww can use a real type check with `instanceof ` to get correct error reporting and Intellisense:

```ts
const tokens = store.root;
assert.ok(tokens instanceof ListNode);

assert.strictEqual(tokens.children[0].token, 1);
```

### Use more correct types
This `as any` was added because we need to cast an `string` to a `AXPropertyName` (a string union):

```ts
function createAXProperty(name: string, value: any, type: AXValueType = 'string'): AXProperty {
	return {
		name: name as any,
		value: createAXValue(type, value)
	};
}
```

If we just use proper types, we don't need the assertion:

```ts
function createAXProperty(name: AXPropertyName, value: any, type: AXValueType = 'string'): AXProperty {
	return {
		name: name,
		value: createAXValue(type, value)
	};
}
```

### Add better types
This code uses `as any` to access some internal types:

```ts
// HACK: xterm.js doesn't expose this by design as it's an internal core
// function an embedder could easily do damage with. Additionally, this
// can't really be upstreamed since the event relies on shell integration to
// verify the shifting is necessary.
(this._terminal as any)._core._bufferService.buffer.lines.onDeleteEmitter.fire({
[...]

PR #269779, opened 2025-10-03
Fix any cases in md extension

mjbvz:
For #269213


dmitrivMS (dismissed):

bpasero (dismissed):
Red CI

mjbvz:
Seriously considering if we can switch off of webpack because I cannot seem to track down why it doesn't check the TS the same way as tsc despite me giving it the same tsconfig

rzhao271 (approved):

PR #269782, opened 2025-10-03
fix: set dns result order to ipv4first for cli connections

### LINKED ISSUE #264136: CLI hangs indefinitely when updating extensions on IPv6 Network

- VS Code Version: 1.103.2
- OS Version: Windows 10 22H2

Steps to Reproduce:

1. have a working IPv6 connection
2. run `code --update-extensions` with at least one extension out of date

From what I can tell, on the CDN side this is the same issue as in winget (https://github.com/microsoft/winget-cli/issues/5693). Both `cdn.winget.microsoft.com` and `main.vscode-cdn.net` use the same CDN (`star-azurefd-prod.trafficmanager.net`).

In the VS Code UI, there seems to be a timeout and automatic fallback to IPv4, but the same is not happening with the CLI. It just hangs when trying to get `https://main.vscode-cdn.net/extensions/marketplace.json`.

On the CDN side, an incident should already be opened. But it would probably be good to have the same fallback logic in the CLI as exists in the UI.

deepak1556:
Fixes https://github.com/microsoft/vscode/issues/264136

### DIFF
--- a/src/vs/code/node/cliProcessMain.ts
+++ b/src/vs/code/node/cliProcessMain.ts
@@ -3,6 +3,7 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
+import { setDefaultResultOrder } from 'dns';
 import * as fs from 'fs';
 import { hostname, release } from 'os';
 import { raceTimeout } from '../../base/common/async.js';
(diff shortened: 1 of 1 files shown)


src/vs/code/node/cliProcessMain.ts
```diff
@@ -109,6 +110,10 @@ class CliMain extends Disposable {
 			// Error handler
 			this.registerErrorHandler(logService);
 
+			// DNS result order
+			// Refs https://github.com/microsoft/vscode/issues/264136
+			setDefaultResultOrder('ipv4first');
```
bpasero, line 115: Why are we setting this only in the CLI and not for all startups of VS Code?

bpasero (changes requested):

bpasero (approved):

PR #269791, opened 2025-10-03
Responsive chat panel (dynamic compression)

### LINKED ISSUE #271528: Chat placeholder not responsive to smaller height windows
Content should respond to smaller height viewports.

Screen Recording:

Image

---

eli-w-king:
CSS solution for responsive chat panel

https://github.com/user-attachments/assets/0dec2072-b2b7-4e1a-afe6-f1970b2b2a4b


src/vs/workbench/contrib/chat/browser/media/chatViewWelcome.css
```diff
@@ -372,3 +374,38 @@ div.chat-welcome-view {
...
+/* Responsive fallbacks to avoid overlapping welcome content */
+
+@media screen and (max-height: 760px), screen and (max-width: 620px) {
+	.interactive-session:not(.new-welcome-view):not(:has(.chat-welcome-view-icon .codicon-send-to-remote-agent)) .chat-welcome-view-container:has(.chat-welcome-view-suggested-prompts > .chat-welcome-view-suggested-prompt:nth-of-type(2)) div.chat-welcome-view > .chat-welcome-view-title {
+		display: none;
+	}
+}
+
+@media screen and (max-height: 720px), screen and (max-width: 580px) {
+	.interactive-session:not(.new-welcome-view):not(:has(.chat-welcome-view-icon .codicon-send-to-remote-agent)) .chat-welcome-view-container:has(.chat-welcome-view-suggested-prompts > .chat-welcome-view-suggested-prompt:nth-of-type(2)) div.chat-welcome-view > .chat-welcome-view-icon {
```
bhavyaus: - These selectors are very complex. We coul simplify them to target just the specific elements (.chat-welcome-view-icon, .chat-welcome-view-disclaimer, etc.).
- The screen qualifier isn’t needed here since it’s the default I think.
- Also, how are we choosing the max-height and max-width breakpoints?

src/vs/workbench/contrib/chat/browser/media/chatViewWelcome.css
```diff
@@ -242,13 +241,15 @@ div.chat-welcome-view {
...
-		margin: 0 4px;
+		margin: 0;
+		box-sizing: border-box;
+		flex: 0 0 auto;
 
 		& > .chat-welcome-view-suggested-prompt-title {
-			font-size: 14px;
-			font-weight: 600;
+			font-size: 13px;
+			font-weight: 500;
```
bpasero: Can we remove use of `font-weight` altogether? I think it still works without it well given the color.

src/vs/workbench/contrib/chat/browser/media/chatViewWelcome.css
```diff
@@ -396,3 +397,34 @@ div.chat-welcome-view {
 	max-width: 300px;
 	word-wrap: break-word;
 }
+
+/* Responsive fallbacks to avoid overlapping welcome content */
+
+.interactive-session:not(.new-welcome-view):not(:has(.chat-welcome-view-icon .codicon-send-to-remote-agent)) {
```
bpasero: I am a bit worried about the perf of this selector, is there a way we can avoid `:has` that seems to check on every child element recursively?

src/vs/workbench/contrib/chat/browser/media/chatViewWelcome.css
```diff
@@ -396,3 +397,34 @@ div.chat-welcome-view {
...
+	.chat-welcome-view-container:has(.chat-welcome-view-suggested-prompts > .chat-welcome-view-suggested-prompt:nth-of-type(2)) {
```
bpasero: Same here.

bpasero (changes requested):
Overall I wonder if we couldn't achieve this easily without CSS (which comes at a cost) by simply updating the visibility of elements from within our layout call, where we know exactly what kind of height and width the view gets:

https://github.com/microsoft/vscode/blob/79ff5507aaf5ee366ed5f16608f698e88dea8880/src/vs/workbench/contrib/chat/browser/chatWidget.ts#L2598

src/vs/workbench/contrib/chat/browser/media/chatViewWelcome.css
bhavyaus: +1. We've already broken down the selector for new-welcome-view etc above. Most of these can be nested under there.

bpasero (changes requested):
I would suggest to revisit this in November 🙏

PR #269818, opened 2025-10-04
 Refactor: Harden parseLsFiles against ReDoS

aka76bm:
The previous implementation of parseLsFiles used a regular expression
  to parse the output of git ls-files. While the regex was already
  improved to be more specific, this commit refactors the function to
  remove the regex entirely.

  The new implementation uses indexOf and split to parse the lines,
  which is safer and more performant, completely eliminating the risk of
  Regular Expression Denial of Service (ReDoS) vulnerabilities.


dmitrivMS:
Significant and risky churn in our build infrastructure without clear value.

Issue #269810, opened 2025-10-04
`loggerIpc#deregisterLogger` fails with Scheme contains illegal characters


bpasero:
Refs: https://errors.code.visualstudio.com/card?ch=f220831ea2d946c0dcb0f3eaa480eb435a2c1260&bH=76f31ebe-9cf2-d623-b3dc-613208f6d6ba

There is a very frequent bug around this call from millions of machines:

<img width="320" height="98" alt="Image" src="https://github.com/user-attachments/assets/8116e94d-f508-4d47-b2b8-2aa1a2d28e8c" />

https://github.com/microsoft/vscode/blob/9defb53a3bf4267858fb2a57a47cdd8c244c261f/src/vs/platform/log/electron-main/logIpc.ts#L36

The stack is hard to read but I found the location by scanning the minified code.

PR #269829, opened 2025-10-04
Fix LoggerChannel deregisterLogger crash by sanitizing URI scheme pro…

Riyazul555:
Closes issue #269810 

I made 2 changes 

Firstly I added a safeUri helper that
- Revives the URI
- Checks the scheme for illegal characters
- Falls back to 'file' if invalid

Secondly I made another change that 
- I replaced all URI.revive(arg[0]) calls with this.safeUri(arg[0])
The reason behind this is deregisterLogger always get a valid URI for ResourceMap.

Hope this helps 
Do check it out 
Thanks


sandy081:
Thanks for the fix. But this seems to be just a bandaid. We need to find the root cause why the URI contains illegal scheme.

PR #269836, opened 2025-10-04
Fix markdown preview not refreshing when file is overwritten externally

### LINKED ISSUE #265277: Markdown preview doesn't refresh/reload when the file is overwritten

- VS Code Version: 1.102.3

Steps to Reproduce:

Image
 

Image

magliocchetti:
**Closes #265277**

### Problem
When a markdown file is overwritten externally (e.g., by downloading or copying a file with the same name), the preview displays stale cached content instead of refreshing to show the updated file.

### Root Cause
The file watcher used `RelativePattern(resource, '*')` which only watches sibling files in the directory, not the markdown file itself. Additionally, the refresh logic only triggered when VS Code didn't know about the file, but VS Code always knows about open files.

### Solution
- **Added dedicated file watcher** for the markdown file itself using `RelativePattern(fileDir, fileName)` to watch the specific file
- The watcher **always triggers a full refresh** when the file changes externally, regardless of whether VS Code knows about it
- **Added refresh icon** (`$(refresh)`) to the manual refresh command for better visibility in the toolbar

### Changes
- `extensions/markdown-language-features/src/preview/preview.ts`: Added file system watcher for external file changes
- `extensions/markdown-language-features/package.json`: Added refresh icon to the refresh command

### Testing
**Manual test:**
1. Create a markdown file with some content
2. Open it in VS Code and open the preview (Ctrl+K V)
3. From outside VS Code, edit and save the file with different content
4. ✅ Preview automatically refreshes showing the new content

**Coverage:** Solution follows the same pattern used in the `media-preview` extension for handling external file changes.

### DIFF
--- a/extensions/markdown-language-features/src/preview/preview.ts
+++ b/extensions/markdown-language-features/src/preview/preview.ts
@@ -110,6 +110,7 @@ class MarkdownPreview extends Disposable implements WebviewResourceProvider {
 			}
 		}));
 
+		// Watch for changes to files in the same directory (for references, etc.)
 		const watcher = this._register(vscode.workspace.createFileSystemWatcher(new vscode.RelativePattern(resource, '*')));
 		this._register(watcher.onDidChange(uri => {
 			if (this.isPreviewOf(uri)) {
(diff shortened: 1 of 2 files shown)


extensions/markdown-language-features/src/preview/preview.ts
```diff
@@ -110,6 +110,7 @@ class MarkdownPreview extends Disposable implements WebviewResourceProvider {
 			}
 		}));
 
+		// Watch for changes to files in the same directory (for references, etc.)
 		const watcher = this._register(vscode.workspace.createFileSystemWatcher(new vscode.RelativePattern(resource, '*')));
```
mjbvz, line 114: We should only need one watcher. Also have you debugged why this doesn't work? It seems to work fine in my testing

mjbvz (changes requested):

Issue #269812, opened 2025-10-04
Respect submodule names defined in superrepo's `.gitmodules` file


Gravifer:
In the superrepo, with the following `.git/config`
```
[core]
	repositoryformatversion = 0
	filemode = false
	bare = false
	logallrefupdates = true
	symlinks = false
	ignorecase = true
[remote "origin"]
	url = https://github.com/Gravifer/lean-mathlib-1stproj.git
	fetch = +refs/heads/*:refs/remotes/origin/*
[branch "main"]
	remote = origin
	merge = refs/heads/main
	vscode-merge-base = origin/main
[submodule "mil"]
	url = https://github.com/Gravifer/lean-comm.mathematics_in_lean.git
	active = true
[submodule "mpil"]
	url = https://github.com/leanprover-community/lean4-metaprogramming-book.git
	active = true
```
and `.gitmodules`
```
[submodule "mil"]
	path = LeanMathlib1stproj/mil
	url = https://github.com/Gravifer/lean-comm.mathematics_in_lean.git
[submodule "mpil"]
	path = LeanMathlib1stproj/mpil/repo
	url = https://github.com/leanprover-community/lean4-metaprogramming-book.git
```
The SCM panel shows the second submodule with the name `repo`, as it determines the name using only the submodule root dirname (impressing skimming the source code). Upon the situation where the tags have been customized, it can be more informant to use the configured tag; the behavior can be left unaltered should the tag is left auto-generated, as it would just be the relative path to the submodule root.

Gravifer:
Current expectation to behavior: when the submodule directory is opened by itself, it can be argued that the current behavior should be prefered; only when the subrepo is detected from a superrepo, should the change take place.

Gravifer:
PR #269902 updated.

PR #269902, opened 2025-10-05
Git - respect configured submodule names in SCM

Gravifer:
## Summary

- respect customized submodule names from the superproject's `.gitmodules` throughout SCM UI
- add a creation-time repository name to the proposed `scmProviderOptions` API
- preserve existing workspace-folder and basename behavior for ordinary repositories, worktrees, generated submodule names, and independently opened submodules
- use the same resolved name in the built-in Git repository picker

Addresses #269812

## Behavior

A submodule discovered through an open superproject uses its configured name only when that name differs from its path. The explicit name flows into `ISCMProvider.name`, so the repositories tree, active repository status, SCM and history pickers, accessibility text, and Git repository picker remain consistent. Opening the submodule independently retains the current behavior.

## Validation

- `npm run typecheck-client`
- `npm run gulp compile-extension:git`
- `git diff --check upstream/main...HEAD`

The full `compile-extensions` command was also attempted; it stopped in the unrelated CSS language extension because that extension's separate Node typings were not installed in this clone. Focused core and Git test launchers could not start because the local `.build/electron` test runtime is absent.

### DIFF
--- a/extensions/git/src/model.ts
+++ b/extensions/git/src/model.ts
@@ -24,7 +24,7 @@ import { RepositoryCache } from './repositoryCache';
 
 class RepositoryPick implements QuickPickItem {
 	@memoize get label(): string {
-		return path.basename(this.repository.root);
+		return this.repository.name;
 	}
 
 	@memoize get description(): string {
--- a/src/vs/workbench/contrib/scm/browser/scmRepositoryRenderer.ts
+++ b/src/vs/workbench/contrib/scm/browser/scmRepositoryRenderer.ts
@@ -151,6 +151,7 @@ export class RepositoryRenderer implements ICompressibleTreeRenderer<ISCMReposit
 		if (this.scmViewService.explorerEnabledConfig.get() === false) {
 			label = repository.provider.name;
 		} else {
+			// Include the parent name when repositories are shown as a tree.
 			const parentRepository = this.scmViewService.repositories
 				.find(r => r.provider.id === repository.provider.parentId);
 
--- a/src/vs/workbench/contrib/scm/browser/scmViewService.ts
+++ b/src/vs/workbench/contrib/scm/browser/scmViewService.ts
@@ -11,9 +11,7 @@ import { IInstantiationService } from '../../../../platform/instantiation/common
 import { SCMMenus } from './menus.js';
 import { IStorageService, StorageScope, StorageTarget } from '../../../../platform/storage/common/storage.js';
 import { debounce } from '../../../../base/common/decorators.js';
-import { IWorkspaceContextService } from '../../../../platform/workspace/common/workspace.js';
 import { compareFileNames, comparePaths } from '../../../../base/common/comparers.js';
-import { basename } from '../../../../base/common/resources.js';
 import { binarySearch } from '../../../../base/common/arrays.js';
 import { IConfigurationService } from '../../../../platform/configuration/common/configuration.js';
 import { IContextKey, IContextKeyService, RawContextKey } from '../../../../platform/contextkey/common/contextkey.js';
@@ -30,15 +28,6 @@ import { getSCMRepositoryIcon } from './util.js';
 
 function getProviderStorageKey(provider: ISCMProvider): string {
 	return `${provider.providerId}:${provider.label}${provider.rootUri ? `:${provider.rootUri.toString()}` : ''}`;
-}
-
-function getRepositoryName(workspaceContextService: IWorkspaceContextService, repository: ISCMRepository): string {
-	if (!repository.provider.rootUri) {
-		return repository.provider.label;
-	}
-
-	const folder = workspaceContextService.getWorkspaceFolder(repository.provider.rootUri);
-	return folder?.uri.toString() === repository.provider.rootUri.toString() ? folder.name : basename(repository.provider.rootUri);
 }
 
 export const RepositoryContextKeys = {
@@ -239,8 +228,7 @@ export class SCMViewService implements ISCMViewService {
 		@IExtensionService extensionService: IExtensionService,
 		@IInstantiationService instantiationService: IInstantiationService,
 		@IConfigurationService private readonly configurationService: IConfigurationService,
-		@IStorageService private readonly storageService: IStorageService,
-		@IWorkspaceContextService private readonly workspaceContextService: IWorkspaceContextService
+		@IStorageService private readonly storageService: IStorageService
 	) {
 		this.menus = instantiationService.createInstance(SCMMenus);
 
@@ -530,8 +518,8 @@ export class SCMViewService implements ISCMViewService {
 		}
 
 		// Sort by repository name, using the path as a tie-breaker.
-		const name1 = getRepositoryName(this.workspaceContextService, op1.repository);
-		const name2 = getRepositoryName(this.workspaceContextService, op2.repository);
+		const name1 = op1.repository.provider.name;
+		const name2 = op2.repository.provider.name;
 
 		const nameComparison = compareFileNames(name1, name2);
 		if (nameComparison === 0 && op1.repository.provider.rootUri && op2.repository.provider.rootUri) {
--- a/src/vs/workbench/contrib/scm/browser/scmViewService.ts
+++ b/src/vs/workbench/contrib/scm/browser/scmViewService.ts
@@ -529,7 +529,7 @@ export class SCMViewService implements ISCMViewService {
 			return comparePaths(op1.repository.provider.rootUri.fsPath, op2.repository.provider.rootUri.fsPath);
 		}
 
-		// Sort by name, path
+		// Sort by repository name, using the path as a tie-breaker.
 		const name1 = getRepositoryName(this.workspaceContextService, op1.repository);
 		const name2 = getRepositoryName(this.workspaceContextService, op2.repository);
 
--- a/extensions/git/src/git.ts
+++ b/extensions/git/src/git.ts
@@ -890,6 +890,10 @@ export interface Submodule {
 	url: string;
 }
 
+export function getSubmoduleDisplayName(submodule: Submodule): string | undefined {
+	return submodule.name !== submodule.path ? submodule.name : undefined;
+}
+
 export function parseGitmodules(raw: string): Submodule[] {
 	const result: Submodule[] = [];
 
--- a/extensions/git/src/repository.ts
+++ b/extensions/git/src/repository.ts
@@ -18,7 +18,7 @@ import { ForcePushMode, GitErrorCodes, RefType, Status } from './api/git.constan
 import { AutoFetcher } from './autofetch';
 import { GitBranchProtectionProvider, IBranchProtectionProviderRegistry } from './branchProtection';
 import { debounce, memoize, sequentialize, throttle } from './decorators';
-import { Repository as BaseRepository, BlameInformation, Commit, CommitShortStat, GitError, IDotGit, LogFileOptions, LsTreeElement, PullOptions, RefQuery, Stash, Submodule, Worktree } from './git';
+import { Repository as BaseRepository, BlameInformation, Commit, CommitShortStat, getSubmoduleDisplayName, GitError, IDotGit, LogFileOptions, LsTreeElement, PullOptions, RefQuery, Stash, Submodule, Worktree } from './git';
 import { GitHistoryProvider } from './historyProvider';
 import { Operation, OperationKind, OperationManager, OperationResult } from './operation';
 import { CommitCommandsCenter, IPostCommitCommandsProviderRegistry } from './postCommitCommands';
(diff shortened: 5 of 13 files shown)


Gravifer:

Gravifer:
Current attempts at PR head heavily change method signatures; I hope to reduce the footprint eventually.

Gravifer:
OK, I think I can at least have Copilot go through my jank now. Sorry for LLM generated commit messages btw, I may squash everything later.

extensions/git/src/model.ts
```diff
@@ -199,7 +204,7 @@ export class Model implements IRepositoryResolver, IBranchProtectionProviderRegi
 	private openRepositories: OpenRepository[] = [];
 	get repositories(): Repository[] { return this.openRepositories.map(r => r.repository); }
 
-	private possibleGitRepositoryPaths = new Set<string>();
+	private possibleGitRepositoryPaths = new Map<string, string | undefined>(); // a set with optional labels
```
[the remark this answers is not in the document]
Gravifer: I don't even think this should be left as a Map... I hope there to be input from someone (that at least has ever written TS, as opposed to myself) to suggest a best practice instead of this.

extensions/git/src/model.ts
```diff
@@ -23,7 +23,12 @@ import { ISourceControlHistoryItemDetailsProviderRegistry } from './historyItemD
 
 class RepositoryPick implements QuickPickItem {
 	@memoize get label(): string {
-		return path.basename(this.repository.root);
+		const defaultLabel = path.basename(this.repository.root);
+		const suppliedLabel = this.repository.repoLabel;
+		if (!suppliedLabel || path.basename(suppliedLabel) === defaultLabel) {
```
Gravifer: This is janky! 

- [ ] Is there an `or_else` in TS?
- [ ] Is the [logic](https://github.com/microsoft/vscode/issues/269812#issuecomment-3369016426) I outlined the way to go?

Gravifer:
I have substantially revised the approach taken.

My original attempt carried an optional label through the repository scan queue, `Model.openRepository`, `Git.open`, and the low-level Git repository object. That label was then used only by the Git extension’s repository quick picker. Besides changing several unrelated method signatures, this produced an incomplete result: the picker could show the configured submodule name while the persistent SCM tree, status bar, history views, accessibility text, and other SCM surfaces continued to show the directory name. The result could also depend on repository discovery order.

The branch tip removes that plumbing. Instead, it adds an optional, creation-time repository name to the existing proposed `scmProviderOptions` API. The name is transported through the extension host and assigned to the core `ISCMProvider.name`. This is the value the workbench already uses throughout its SCM user interface, so the change applies consistently without modifying each renderer or view individually.

The Git extension now determines the name at the point where it creates the child repository’s source control. If the child is a submodule, its superproject is open, and the matching `.gitmodules` entry has a logical name different from its path, that configured name is supplied to SCM core. Git’s own repository picker uses the same resolved value.

This yields deliberately contextual behavior:

- A submodule discovered through its superproject uses the customized `.gitmodules` name.
- A submodule created while its matching superproject is not open retains the existing workspace-folder or directory name.
- A normal, automatically generated submodule name—where `name === path`—does not override existing behavior.
- Ordinary repositories and worktrees are unaffected.

I think this distinction is useful. When a repository is opened independently, its folder or workspace name is the most immediate identity available to the user. When it is discovered as part of a superproject, however, the superproject’s `.gitmodules` file supplies an intentional logical identity. That identity can be much more informative than the final path segment, particularly for deeply nested submodules or layouts where the directory name is generic. The logic is also consistent across the repository tree, status and history surfaces, SCM pickers, accessibility output, and the Git picker.

A customized submodule name takes precedence over a workspace-folder alias when the child is created while its superproject is open and resolvable. This avoids returning to the earlier split behavior, where different SCM surfaces could disagree about the same repository. Existing SCM rendering already handles duplicate names by showing parent and path context where necessary, so no new disambiguation mechanism is required.

The implementation is intentionally narrow from a maintenance perspective. It does not alter the stable extension API, add a setting, create persistent state, introduce a migration, or require dynamic rename events. The proposed API gains one optional final argument; callers that omit it retain the current workspace-folder, basename, and provider-label fallbacks. No existing extension needs to change.

Likewise, the workbench renderers are untouched. They already consume `ISCMProvider.name`, so the new value flows naturally to the existing UI. The Git-side lookup happens once during repository creation, uses submodule information that has already been parsed, and does not start additional Git processes or watchers. The earlier changes to the scan queue and low-level Git abstraction have been removed.

I hope this would not introduce much burden for the maintainers or become obstructive for future contributors. Thanks to you all for the excellent project that benefits all developers.

Gravifer:
Hi @lszomoru — when you have a chance, would you mind taking a look at the revised approach? The PR now routes repository display names through SCM core and removes the original Git-specific discovery plumbing. I’d particularly appreciate feedback on whether this direction fits the SCM API. No rush, and thank you for your time.

@coderabbitai @copilot can you initiate reviews as well? Perhaps focus on whether there are potential drawbacks to this.

------
I'm force pushing to make rebases from time to time. The material commits are:
- feat(scm): declare repository display names [wip] [no-ci]
- feat(scm): transport repository display names [wip] [no-ci]
- feat(scm): apply repository display names
- feat(git): select customized submodule names
- feat(git): name parent-detected submodules
- feat(git): align repository picker names
With moving SHAs.

Gravifer:
> Quoting the @copilot updated [overview](https://github.com/microsoft/vscode/tasks/92d32eca-7f35-4695-bac5-6803a4bb3dc6?session_id=4642bb99-b2f2-493c-944e-4c18337ec9e5) here

Extends SCM naming so Git submodules display customized .gitmodules names consistently while preserving existing fallback behavior.

**Changes**:
- Adds an optional creation-time repository name to the proposed SCM API and RPC flow.
- Resolves customized submodule names for SCM providers and Git repository picks.
- Adds coverage for name resolution and forwarding.

### Reviewed Changes

File | Description
-- | --
`src/vscode-dts/vscode.proposed.scmProviderOptions.d.ts` | Adds the optional SCM repository name API.
`src/vs/workbench/api/test/common/extHostSCM.test.ts` | Tests name forwarding.
`src/vs/workbench/api/common/extHostSCM.ts` | Passes names through the extension host.
`src/vs/workbench/api/common/extHost.protocol.ts` | Extends the SCM registration protocol.
`src/vs/workbench/api/common/extHost.api.impl.ts` | Exposes and proposal-gates the new argument.
`src/vs/workbench/api/browser/mainThreadSCM.ts` | Applies explicit names with existing fallbacks.
`extensions/git/src/test/git.test.ts` | Tests customized-name resolution.
`extensions/git/src/repository.ts` | Resolves and registers submodule names.
`extensions/git/src/model.ts` | Uses resolved names in repository picks.
`extensions/git/src/git.ts` | Adds the submodule display-name helper.

src/vs/workbench/contrib/scm/browser/scmRepositoryRenderer.ts
Gravifer, line 161: The repository renderer displays `ISCMProvider.name`

src/vs/workbench/contrib/scm/browser/scmViewService.ts
Gravifer, line 527: These values are normally identical, so the distinction has little effect today. With this PR, though, an explicit provider name can intentionally differ from the repository path.

Gravifer: "Sort by Name," however, still derives its sort key independently from the workspace-folder name or root basename

Gravifer (commented):
### Minor follow-up from local validation

For example:

| Configured name | Root basename | Displayed label |
|---|---|---|
| `zebra` | `alpha` | `parent / zebra` |
| `apple` | `zulu` | `parent / apple` |

The current implementation displays `zebra` and `apple`, but sorts using `alpha` and `zulu`. Selecting “Sort by Name” could therefore produce:

```text
parent / zebra
parent / apple
```

The PR already makes the explicit name authoritative while preserving the existing workspace-folder, basename, and provider-label fallbacks.



I am going to try a small follow-up correction that sorts by the displayed provider name:

```ts
const nameComparison = compareFileNames(
	op1.repository.provider.name,
	op2.repository.provider.name
);
```

The existing path tie-breaker and locale-aware comparator would remain unchanged.

I think this is justified within the PR because customized names would otherwise be displayed correctly but behave unexpectedly under the existing “Sort by Name” option. For callers that do not supply an explicit name, `provider.name` already resolves through the current fallbacks, so ordinary repositories should retain their existing ordering.

I do not plan to modify the pre-existing comparer tests. Their initial local failures came from the direct Electron test host inheriting `zh-CN`; all 51 relevant tests passed when the host was run with `en-US`. That appears separate from this small SCM sort-key adjustment.

Gravifer:
I have pushed the follow-up correction discussed above.

- [`2b68636`](https://github.com/Gravifer/microsoft.vscode/commit/2b68636) synchronizes the SCM transport test with the asynchronous RPC protocol before asserting registration and after disposal.
- [`e4e1568`](https://github.com/Gravifer/microsoft.vscode/commit/e4e1568) makes “Sort by Name” use the same canonical `ISCMProvider.name` displayed by the workbench. The existing locale-aware comparator and root-path tie-breaker remain unchanged.

https://github.com/Gravifer/microsoft.vscode/blob/e4e1568/src/vs/workbench/contrib/scm/browser/scmViewService.ts#L520-L529

The new regression test deliberately opposes provider names and root basenames, while giving two repositories the same provider name in reverse path order. It therefore verifies both that the displayed name controls sorting and that the existing path tie-breaker is preserved.

https://github.com/Gravifer/microsoft.vscode/blob/e4e1568/src/vs/workbench/contrib/scm/test/browser/scmViewService.test.ts#L36-L77

The transport test now explicitly waits for the test RPC queue:

https://github.com/Gravifer/microsoft.vscode/blob/2b68636/src/vs/workbench/api/test/common/extHostSCM.test.ts#L56-L100

Local validation included:

- `typecheck-client`: passed
- Targeted ESLint: passed
- Client esbuild transpilation: passed
- Focused `SCMViewService` suite: 2 passing
- Focused `ExtHostSCM` suite: 3 passing
- Full Git integration suite: 59 passing
- Broad Electron suite: 27,898 passing, 453 pending, 11 failing

None of the 11 Electron failures involved Git or SCM, and all occurred in source/test files byte-identical to upstream `main`. Their symptoms were environment-sensitive: Windows environment-variable assumptions, loopback tests interacting with the local proxy, a localized Kerberos error string, display-scale-dependent measurement, and timeouts. I stopped rather than modifying or repeatedly retrying unrelated tests. The four comparer tests that had previously exposed the host `zh-CN` locale all passed when the Electron host was run with VS Code’s `en-US` language.

The targeted coverage command also passed its tests, but the existing Electron coverage path produced an empty `lcov.info`, so I am not claiming numerical coverage.

These results were obtained on the patch-equivalent validation worktree immediately before the branch was refreshed onto newer `main`; range-diff confirmed that the PR patches and the two follow-up commits were unchanged.

Gravifer:
Hi @lszomoru — I'm still interested in contributing this improvement and would be grateful for your thoughts on the revised approach, particularly the optional SCM provider name API. I'm happy to adjust the implementation based on your guidance. If someone else would be better placed to offer input, I would also appreciate a pointer. Thank you for your time.

I've only stopped rebasing the branch so that we have a stable review surface; if rebasing to trunk tip is still desired, do please instruct me to do so :D

Gravifer:
I have pushed one final branch refresh at [`53f0c69`](https://github.com/Gravifer/microsoft.vscode/commit/53f0c69), based directly on current `main` `1f625ad`. I plan to keep the PR head stable from here.

During the private hardening pass I found a discovery-order race in the revised implementation: the child could be created before the open parent's first status operation published its parsed submodule list, causing the basename to be frozen as the creation-time SCM name. The [current resolution](https://github.com/Gravifer/microsoft.vscode/blob/53f0c69/extensions/git/src/model.ts#L708-L729) reads the parent's current `.gitmodules` before child/SCM creation, while retaining the last parsed parent state only as a fallback for genuine I/O errors.

A [real-repository integration test](https://github.com/Gravifer/microsoft.vscode/blob/53f0c69/extensions/git/src/test/submodule.test.ts#L17-L116) now drains the parent's initial status, deliberately empties its published submodule state, and opens the configured nested child through the Git model. It passes at the PR head; on the same stack without the race fix it fails with `metaprogramming` instead of `mpil`.

Current local validation includes full extension compilation, core and proposed-API typechecks, targeted ESLint, the focused SCM suites, and the full Git integration suite (59 passing, 2 pre-existing pending, 0 failing). The direct-file lookup starts no Git process or watcher; a warm-cache probe measured about 0.33 ms per child, or roughly 3.3 ms at the default 10-submodule detection limit.

I have updated the PR description with the exact behavior contract and validation details. Thank you for considering the revised and now race-hardened approach.

PR #270025, opened 2025-10-06
Fix proxy auth attempt default

yugook:
### Summary

  - Treat a missing req.firstAuthAttempt as the first attempt so cached or stored proxy
    credentials keep working.
  - Fixes the regression where VS Code started prompting for proxy credentials on every request
    after the latest Electron bump.

### Testing

  - Testing could not be performed behind an authenticating proxy, as development was conducted in an environment without proxy access. The issue has been confirmed on the official release build under proxy conditions.


bpasero:
Is there an issue for this with more details?

deepak1556:
> Fixes the regression where VS Code started prompting for proxy credentials on every request
after the latest Electron bump.

Do you know the commit in VSCode regressed this behavior ? Thanks!

deepak1556 (changes requested):
`firstAuthAttempt` is not an optional parameter https://github.com/electron/electron/blob/d6dfd4ed7a8f3664e13da7e60eb39529820b72bd/shell/browser/login_handler.cc#L75 unsure what you mean by it is missing ?

yugook:
Thanks for the quick reviews and questions.
This regression reproduces only on my corporate machine that sits behind an authenticating proxy; I don’t have a comparable setup at home,
so I’m unable to capture the login event values or bisect Electron from here.

My intent was simply to flag the behavior change I’m observing in the released build after the latest Electron update.
If that change isn’t expected, it might need investigation on the Electron side — or please feel free to close this PR if it isn’t helpful.

Let me know if there’s any additional detail I can provide from the limited environment I have.

yugook:
As a note: in VSCode 1.104.2 I did not observe frequent proxy auth prompts.
In 1.104.3, they appear more often, although the frequency varies from day to day.
I haven’t confirmed whether the bundled Electron version differs between the two releases.
I apologize if this issue turns out to be specific to my corporate environment.

deepak1556:
Between 1.104.2 and 1.104.3 we didn't have a change in runtime version, however there was a bump in `@vscode/proxy-agent` https://github.com/microsoft/vscode/commit/0a015e0ef041bf8077062604c2b87402a9996a0e with https://github.com/microsoft/vscode-proxy-agent/commit/df436a72430e322b6977f9a894538f5cc7d52ab6 which should affect only Node.js network requests.

@chrmarti do we use auth prompt when performing requests via Node.js ?

yugook:
Thanks for checking, @deepak1556. That makes sense — I hadn’t noticed the proxy-agent bump.
I mistakenly attributed the issue to the Electron update. I’m sorry for causing any confusion, and appreciate you looking into this further.

yugook:
Sorry for the delayed follow-up, and thanks again for taking a look.

On my side, I can no longer reproduce this in the same corporate environment where I previously saw the repeated proxy auth prompts. I first confirmed that after upgrading to VS Code 1.110.0, and I haven’t been able to reproduce it since.

Given that I don’t have a reliable repro or a confirmed root cause, I’m going to close this PR for now to avoid creating noise.

If I can reproduce it again with clearer evidence, I’ll open a new issue with more details. Thanks again.

Issue #269758, opened 2025-10-03
`accessibility.verboseChatProgressUpdates` does not read CoT


anthonykim1:
When `accessibility.verboseChatProgressUpdates` is enabled, it would not read CoT. 

/ cc @jooyoungseo and @rperez030

meganrogge:
@rperez030 I noticed that when I log what we're alerting, all the messages are present. But when I enable the screen reader, only some are read. I believe this is a timing issue. Using `status` vs `alert` did not fix it. Any suggestions? Not sure if there's anything we can do about it.

rperez030:
> [@rperez030](https://github.com/rperez030) I noticed that when I log what we're alerting, all the messages are present. But when I enable the screen reader, only some are read. I believe this is a timing issue. Using `status` vs `alert` did not fix it. Any suggestions? Not sure if there's anything we can do about it.

The problem with alerts is that they override each other if they are too close in time, or even if they are being announced when a new one arrives in the case of VoiceOver. 

**Status** should be the role for progress messages unless they are time-sensitive, such as confirmations or errors. 

I believe the logic uses a 4-second interval, so I don't think timing is the issue, but maybe increasing it to 5 seconds would help. 

Is this using the same existing infrastructure that is used for chat responses?
It is also interesting that I still hear the "working" message, which I thought had been replaced by the new progress messages.

meganrogge:
We still use the `working` message because extension contributed tool messages aren't read out currently. https://github.com/microsoft/vscode/issues/266256

I tried `status`, which led to nothing being announced 🫠 

Maybe if we only say `working` when these verbose ones are disabled, and set the chat response progress announcement to off by default, then the `status` will come through?

meganrogge:
Using this to track the issue when using Codex as the model and not hearing announcements about the thinking progress

justschen:
i believe this was fixed by @meganrogge ?

PR #270078, opened 2025-10-06
Move tool progress announcements from `LanguageModelToolsService` -> `ChatListRenderer` 

### LINKED ISSUE #266256: Tool invocation for extension tools are not announced to screen reader users
We only know about them in the rendering code, so would need a way to set / check for the ID to ensure it's only announced on initial render

cc @roblourens

meganrogge:
- Enables us to announce extension contributed tool progress messages
- We prevent re-announcement by storing the ID and by checking that the type is not a serialized tool invocation
- We now use `status` instead of `alert` to ensure overlapping announcements are not lost, but instead, queued
- Disables generic `Working` message when our verbose setting is enabled, already covered by our progress signal in most cases
- Sets the default for `progress.announcement` to `false` 

Investigated also adding announcements for `thinking` parts, but found that couldn't happen because we stream in the progress message, so it was trunchated. We also don't have a place where we know that streaming has finished. 
image

fixes #266256
fixes part of #269758

### DIFF
--- a/src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatToolProgressPart.ts
+++ b/src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatToolProgressPart.ts
@@ -4,11 +4,14 @@
  *--------------------------------------------------------------------------------------------*/
 
 import * as dom from '../../../../../../base/browser/dom.js';
+import { status } from '../../../../../../base/browser/ui/aria/aria.js';
 import { IMarkdownString, MarkdownString } from '../../../../../../base/common/htmlContent.js';
 import { autorun } from '../../../../../../base/common/observable.js';
 import { MarkdownRenderer } from '../../../../../../editor/browser/widget/markdownRenderer/browser/markdownRenderer.js';
+import { IConfigurationService } from '../../../../../../platform/configuration/common/configuration.js';
 import { IInstantiationService } from '../../../../../../platform/instantiation/common/instantiation.js';
 import { IChatProgressMessage, IChatToolInvocation, IChatToolInvocationSerialized, ToolConfirmKind } from '../../../common/chatService.js';
+import { AccessibilityWorkbenchSettingId } from '../../../../accessibility/browser/accessibilityConfiguration.js';
 import { IChatCodeBlockInfo } from '../../chat.js';
 import { IChatContentPartRenderContext } from '../chatContentParts.js';
 import { ChatProgressContentPart } from '../chatProgressContentPart.js';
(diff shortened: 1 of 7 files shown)


src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatToolProgressPart.ts
```diff
@@ -32,15 +37,21 @@ export class ChatToolProgressSubPart extends BaseChatToolInvocationSubPart {
...
 			return part.domNode;
 		} else {
 			const container = document.createElement('div');
 			const progressObservable = this.toolInvocation.kind === 'toolInvocation' ? this.toolInvocation.progress : undefined;
 			this._register(autorun(reader => {
 				const progress = progressObservable?.read(reader);
-				const part = reader.store.add(this.renderProgressContent(progress?.message || this.toolInvocation.invocationMessage));
+				const key = this.getAnnouncementKey('progress');
+				const progressContent = progress?.message ?? this.toolInvocation.invocationMessage;
+				const shouldAnnounce = this.toolInvocation.kind === 'toolInvocation' && this.hasMeaningfulContent(progress?.message) ? this.computeShouldAnnounce(key) : false;
```
roblourens, line 53: Just from looking at the code, it looks like if the chat is not being rendered while the response loads (sidebar collapsed, or chat is scrolled), then all of these would be read simultaneously as soon as they are rendered.

If that's an issue then you'd want to find another way to decide whether to read a particular progress item. `kind === 'toolInvocation'` probably works fine today to decide not read an item from a reloaded session, but I don't think it's safe, there's no real guarantee that this type is only used for new tool calls.

meganrogge: I tried what you suggested and not sure if I'm not hitting the case you're concerned about bc i'm forced to sign in again or if that's not an issue. Can you try pls?

roblourens: Sign in? I'm just talking about scrolling and closing the sidebar

meganrogge: Yeah I tried those and did not have an issue

src/vs/workbench/contrib/chat/browser/chatListRenderer.ts
```diff
@@ -1272,6 +1279,12 @@ export class ChatListItemRenderer extends Disposable implements ITreeRenderer<Ch
 		}
 	}
 
+	override dispose(): void {
+		this._announcedToolProgressKeys.clear();
```
roblourens, line 1283: This shouldn't be necessary. It's fine to leave it, there is some code in the repo that does this kind of thing when people feel paranoid about GC but I don't generally do this.

roblourens (approved):

PR #270100, opened 2025-10-06
Reduce `any` usage in ts extension

mjbvz:
Follow up on #269213

Doesn't fix all cases but handles most of the easier ones


dmitrivMS (approved):

mjbvz:

Issue #259730, opened 2025-08-05
Pressing "Enter" in Copilot agent chat when no text entered should accept "Continue" button


drewnoakes:
Type: <b>Feature Request</b>

Copilot agent chat is amazing, but it regularly stops processing to ask for confirmation of various actions.

When this happens, there's a "Continue" button on screen.

I'm primarily a keyboard user, and would like this button to be easily pressed via the "Enter" key.

However, pressing "Enter" when this big blue button appears on screen submits an empty prompt, which ends up cancelling whatever activity was previously underway -- the complete opposite action to what was intended.

Please allow pressing Enter when no prompt has been entered to accept "Continue". Thanks!

VS Code version: Code 1.102.3 (488a1f239235055e34e673291fb8d8c810886f81, 2025-07-29T03:00:23.339Z)
OS version: Windows_NT x64 10.0.26100
Modes:

drewnoakes:
Hmm, I guess this should be filed against the Copilot extension. Will close and resubmit.

drewnoakes:
https://github.com/microsoft/vscode-copilot-issues/issues/154

roblourens:
No, these issues go in this repo. But you can use the cmd+enter keybinding

drewnoakes:
Pressing <kbd>Enter</kbd> when no text has been entered in the chat window currently cancels the pending operation, which isn't a great experience.

roblourens:
That's not right

roblourens:
@bryanchen-d if this was fixed by a PR, it should have a bug label (or feature-request) so that it will get verified

PR #270117, opened 2025-10-07
feat: disable send button and prevent Enter key from creating new lines when input is empty

### LINKED ISSUE #264462: Disable chat buttons when input field is empty
1. Open Chat view
2. Leave chat input field empty   -> **Send** and **Delegate to Coding Agent** buttons are enabled and clickable.

   Image

**Expected result**: 
* Buttons should only be enabled when there's text in the chat input field.

**System info**:
Version: 1.104.0-insider (user setup)
Commit: 966ec1415440da5f1ef53acbd97ba2113add8853
Date: 2025-09-02T05:05:45.743Z
Electron: 37.3.1
ElectronBuildId: 12259562
Chromium: 138.0.7204.235
Node.js: 22.18.0
V8: 13.8.258.31-electron.0
OS: Windows_NT x64 10.0.26100

bryanchen-d:
When input is empty in chat, disable the send button and also prevent Enter key creating new lines.

Fixes #264462 264462

This also fixes an issue from https://github.com/microsoft/vscode/issues/259730 that when there is a tool pending confirmation, user hits enter when input is empty will cancel the tool.
Root cause of that is the Enter key is bound to the chat.submit action which eventually calls into cancel the pending tool invocation.
With this fix that disabled the action when intput is empty, that codeflow is not activated anymore thus that issue is addressed.

### DIFF
--- a/src/vs/base/browser/keyboardEvent.ts
+++ b/src/vs/base/browser/keyboardEvent.ts
@@ -114,6 +114,15 @@ export function printStandardKeyboardEvent(e: StandardKeyboardEvent): string {
 	return `modifiers: [${modifiers.join(',')}], code: ${e.code}, keyCode: ${e.keyCode} ('${KeyCodeUtils.toString(e.keyCode)}')`;
 }
 
+export function hasModifierKeys(keyStatus: {
+	readonly ctrlKey: boolean;
+	readonly shiftKey: boolean;
+	readonly altKey: boolean;
+	readonly metaKey: boolean;
+}): boolean {
+	return keyStatus.ctrlKey || keyStatus.shiftKey || keyStatus.altKey || keyStatus.metaKey;
+}
+
 export class StandardKeyboardEvent implements IKeyboardEvent {
 
 	readonly _standardKeyboardEventBrand = true;
--- a/src/vs/base/browser/keyboardEvent.ts
+++ b/src/vs/base/browser/keyboardEvent.ts
@@ -115,10 +115,10 @@ export function printStandardKeyboardEvent(e: StandardKeyboardEvent): string {
 }
 
 /**
- * Checks if a keyboard event has no modifier keys pressed
+ * Checks if a keyboard event has any modifier keys pressed
  */
-export function hasNoModifierKeys(e: IKeyboardEvent): boolean {
-	return !e.altKey && !e.ctrlKey && !e.metaKey && !e.shiftKey;
+export function hasAnyModifierKeys(e: IKeyboardEvent): boolean {
+	return e.altKey || e.ctrlKey || e.metaKey || e.shiftKey;
 }
 
 export class StandardKeyboardEvent implements IKeyboardEvent {
--- a/src/vs/base/browser/keyboardEvent.ts
+++ b/src/vs/base/browser/keyboardEvent.ts
@@ -114,13 +114,6 @@ export function printStandardKeyboardEvent(e: StandardKeyboardEvent): string {
 	return `modifiers: [${modifiers.join(',')}], code: ${e.code}, keyCode: ${e.keyCode} ('${KeyCodeUtils.toString(e.keyCode)}')`;
 }
 
-/**
- * Checks if a keyboard event has any modifier keys pressed
- */
-export function hasAnyModifierKeys(e: IKeyboardEvent): boolean {
-	return e.altKey || e.ctrlKey || e.metaKey || e.shiftKey;
-}
-
 export class StandardKeyboardEvent implements IKeyboardEvent {
 
 	readonly _standardKeyboardEventBrand = true;
--- a/src/vs/base/browser/keyboardEvent.ts
+++ b/src/vs/base/browser/keyboardEvent.ts
@@ -114,6 +114,13 @@ export function printStandardKeyboardEvent(e: StandardKeyboardEvent): string {
 	return `modifiers: [${modifiers.join(',')}], code: ${e.code}, keyCode: ${e.keyCode} ('${KeyCodeUtils.toString(e.keyCode)}')`;
 }
 
+/**
+ * Checks if a keyboard event has no modifier keys pressed
+ */
+export function hasNoModifierKeys(e: IKeyboardEvent): boolean {
+	return !e.altKey && !e.ctrlKey && !e.metaKey && !e.shiftKey;
+}
+
 export class StandardKeyboardEvent implements IKeyboardEvent {
 
 	readonly _standardKeyboardEventBrand = true;
--- a/src/vs/workbench/contrib/chat/browser/chatInputPart.ts
+++ b/src/vs/workbench/contrib/chat/browser/chatInputPart.ts
@@ -1162,8 +1162,6 @@ export class ChatInputPart extends Disposable implements IHistoryNavigationWidge
 		const inputScopedContextKeyService = this._register(this.contextKeyService.createScoped(inputContainer));
 		ChatContextKeys.inChatInput.bindTo(inputScopedContextKeyService).set(true);
 		this.currentlyEditingInputKey = ChatContextKeys.currentlyEditingInput.bindTo(inputScopedContextKeyService);
-		// Track whether a request is currently being processed so we can suppress plain Enter behavior
-		const requestInProgressKey = ChatContextKeys.requestInProgress.bindTo(inputScopedContextKeyService);
 		const scopedInstantiationService = this._register(this.instantiationService.createChild(new ServiceCollection([IContextKeyService, inputScopedContextKeyService])));
 
 		const { historyNavigationBackwardsEnablement, historyNavigationForwardsEnablement } = this._register(registerAndCreateHistoryNavigationContext(inputScopedContextKeyService, this));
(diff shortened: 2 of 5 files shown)


src/vs/workbench/contrib/chat/browser/chatInputPart.ts
```diff
@@ -1199,6 +1199,18 @@ export class ChatInputPart extends Disposable implements IHistoryNavigationWidge
 		options.overflowWidgetsDomNode?.classList.add('hideSuggestTextIcons');
 		this._inputEditorElement.classList.add('hideSuggestTextIcons');
 
+		// Prevent Enter key from creating new lines when input is empty
+		this._register(this._inputEditor.onKeyDown((e) => {
+			if (e.keyCode === KeyCode.Enter && !e.altKey && !e.ctrlKey && !e.metaKey && !e.shiftKey) {
+				const model = this._inputEditor.getModel();
+				const inputHasText = !!model && model.getValue().trim().length > 0;
```
roblourens: Hm, just ocurred to me, maybe this `inputHasText` isn't needed because maybe the action will handle the enter key when active and this won't run anyway? I'm not sure

roblourens: But if not, we would also need to check the case when a request is currently in progress

roblourens:

src/vs/workbench/contrib/chat/browser/chatInputPart.ts
bryanchen-d: Not sure if I am following. When the `precondition` disabled the action, the action won't handle the enter key so it (the key event) goes into the `chatInputPart` and creates new lines in the editor, right?

roblourens: Right, but that only happens when the action is disabled, meaning that `inputHasText` is redundant?

src/vs/workbench/contrib/chat/browser/chatInputPart.ts
```diff
@@ -1162,6 +1162,8 @@ export class ChatInputPart extends Disposable implements IHistoryNavigationWidge
 		const inputScopedContextKeyService = this._register(this.contextKeyService.createScoped(inputContainer));
 		ChatContextKeys.inChatInput.bindTo(inputScopedContextKeyService).set(true);
 		this.currentlyEditingInputKey = ChatContextKeys.currentlyEditingInput.bindTo(inputScopedContextKeyService);
+		// Track whether a request is currently being processed so we can suppress plain Enter behavior
+		const requestInProgressKey = ChatContextKeys.requestInProgress.bindTo(inputScopedContextKeyService);
```
roblourens: We shouldn't use context keys to share information across components like this. Context keys should really only be used for setting context that keyboard shortcuts and actions rely on.

src/vs/workbench/contrib/chat/browser/chatInputPart.ts
```diff
@@ -1199,6 +1199,18 @@ export class ChatInputPart extends Disposable implements IHistoryNavigationWidge
...
+		this._register(this._inputEditor.onKeyDown((e) => {
+			if (e.keyCode === KeyCode.Enter && !e.altKey && !e.ctrlKey && !e.metaKey && !e.shiftKey) {
+				const model = this._inputEditor.getModel();
+				const inputHasText = !!model && model.getValue().trim().length > 0;
+				if (!inputHasText) {
+					e.preventDefault();
+					e.stopPropagation();
+				}
+			}
+		}));
```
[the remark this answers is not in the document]
bryanchen-d: Extracted the modifierKey check to a utility module.

src/vs/workbench/contrib/chat/browser/chatInputPart.ts
bryanchen-d: As I just validated, simply swallowing the 'enter' key breaks the normal scenario: user cannot hit enter to send the prompt even when the action is active.

It seems the chatInputPart handle the keyboard event earlier than the action..

bryanchen-d: The keyboard event processing order is:
1. chatInputPart._inputEditor.onKeyDown
2. keybinding service (AbstractKeybindingService._dispatch)
3. monaco editor

so we cannot simply swallow the event at 1.

bryanchen-d: Nevermind, I figured out a work around. Updated.

src/vs/workbench/contrib/chat/browser/chatInputPart.ts
bryanchen-d: Updated removing this context key.

src/vs/workbench/contrib/chat/browser/chatInputPart.ts
```diff
@@ -1199,6 +1199,17 @@ export class ChatInputPart extends Disposable implements IHistoryNavigationWidge
 		options.overflowWidgetsDomNode?.classList.add('hideSuggestTextIcons');
 		this._inputEditorElement.classList.add('hideSuggestTextIcons');
 
+		// Prevent Enter key from creating new lines - but allow keybinding service to still handle the event
+		// We need to prevent Monaco's default Enter behavior while still allowing the VS Code keybinding service
+		// to receive and process the Enter key for ChatSubmitAction
+		this._register(this._inputEditor.onKeyDown((e) => {
+			if (e.keyCode === KeyCode.Enter && hasNoModifierKeys(e)) {
+				// Only prevent the default Monaco behavior (newline insertion)
+				// Do NOT call stopPropagation() so the keybinding service can still process this event
```
roblourens: Oh I see, makes sense

roblourens (dismissed):

src/vs/base/browser/keyboardEvent.ts
```diff
@@ -114,6 +114,13 @@ export function printStandardKeyboardEvent(e: StandardKeyboardEvent): string {
 	return `modifiers: [${modifiers.join(',')}], code: ${e.code}, keyCode: ${e.keyCode} ('${KeyCodeUtils.toString(e.keyCode)}')`;
 }
 
+/**
+ * Checks if a keyboard event has no modifier keys pressed
+ */
+export function hasNoModifierKeys(e: IKeyboardEvent): boolean {
```
bpasero: I would suggest to make this `hasModifierKeys` for easier understanding. Negations tend to make it hard to understand the function meaning.

src/vs/base/browser/keyboardEvent.ts
```diff
@@ -114,6 +114,13 @@ export function printStandardKeyboardEvent(e: StandardKeyboardEvent): string {
 	return `modifiers: [${modifiers.join(',')}], code: ${e.code}, keyCode: ${e.keyCode} ('${KeyCodeUtils.toString(e.keyCode)}')`;
 }
 
+/**
```
bpasero: Note that we have similar code already that we should revisit to use this method or align:

https://github.com/microsoft/vscode/blob/1332311cbaf9f21fb0a11ce5cfd9216208c52018/src/vs/base/browser/dom.ts#L1815-L1817

bpasero (changes requested):

src/vs/base/browser/keyboardEvent.ts
bryanchen-d: Good point. Updated it to `hasAnyModifierKeys`

src/vs/base/browser/keyboardEvent.ts
bryanchen-d: Agree, but any ideas how should we use this method to check a `IKeyboardEvent`?

bryanchen-d: Looks like we can do it by `!ModifierKeyEmitter.getInstance().isModifierPressed`, updated.

roblourens: Actually, I looked at this and I don't think we should use this as-is because it seems like the purpose of this is to check the current status of those keys. But we have an event object and we should be checking the flags on the event object itself. So there could be a race condition.

If there's a way to do it naturally, then maybe we can share a function just so we don't have to duplicate the list of modifiers?

Another idea for reusing our helpers, we could add a method to `StandardKeyboardEvent` for this

bryanchen-d: How about 
```
export function hasModifierKeys(keyStatus: {
	readonly ctrlKey: boolean;
	readonly shiftKey: boolean;
	readonly altKey: boolean;
	readonly metaKey: boolean;
}): boolean {
	return keyStatus.ctrlKey || keyStatus.shiftKey || keyStatus.altKey || keyStatus.metaKey;
}
```
The inline interface works for both`IKeyboardEvent` and `ModifierKeyStatus` in ModifierKeyEmitter.

roblourens: I'm not crazy about the parameter type because it doesn't really tell you how it's expected to be used

roblourens: Well, I think it's fine.

roblourens (dismissed):

src/vs/workbench/contrib/chat/browser/chatInputPart.ts
```diff
@@ -1199,6 +1199,27 @@ export class ChatInputPart extends Disposable implements IHistoryNavigationWidge
 		options.overflowWidgetsDomNode?.classList.add('hideSuggestTextIcons');
 		this._inputEditorElement.classList.add('hideSuggestTextIcons');
 
+		// Prevent Enter key from creating new lines - but respect user's custom keybindings
+		// Only prevent default behavior if ChatSubmitAction is bound to Enter AND its precondition is met
+		this._register(this._inputEditor.onKeyDown((e) => {
+			if (e.keyCode === KeyCode.Enter && !hasModifierKeys(e)) {
+				// Check if ChatSubmitAction has a keybinding for plain Enter in the current context
+				// This respects user's custom keybindings that disable the submit action
+				const chatSubmitKeybinding = this.keybindingService.lookupKeybinding(ChatSubmitAction.ID, this.contextKeyService);
```
roblourens: This says `Look up the preferred (last defined) keybinding for a command`. Should this be using `lookupKeybindings` which returns all of them?

bpasero (dismissed):

src/vs/workbench/contrib/chat/browser/chatInputPart.ts
bryanchen-d: Good catch. Updated to use `lookupKeybindings`.

roblourens (approved):

PR #270236, opened 2025-10-07
fix: Increase workbench border radius on macos tahoe

### LINKED ISSUE #270188: MacOS - Window corner border is cut off in high contrast mode
Screenshot:

Image

Image

Environment Information:
- VS Code Version: 1.105.0-insider
- OS: macOS (arm64)
- Workspace: 2 folder(s) open

---

avarayr:
fixes: https://github.com/microsoft/vscode/issues/270188
image

⚠️ important notes:

Electron will (sometime in the near future) bump their build workflow to build on MacOS 26 Tahoe, which uses XCode 26.
Apps built with XCode 26 will have *even bigger* corner radius, here's an example

image

electron preparing for xcode26 is tracked here: https://github.com/electron/electron/issues/47514

So this will be revisited in the future to increase *even more* (possibly to 18px?)

### DIFF
--- a/src/vs/workbench/electron-browser/desktop.main.ts
+++ b/src/vs/workbench/electron-browser/desktop.main.ts
@@ -45,7 +45,7 @@ import { WorkspaceTrustEnablementService, WorkspaceTrustManagementService } from
 import { IWorkspaceTrustEnablementService, IWorkspaceTrustManagementService } from '../../platform/workspace/common/workspaceTrust.js';
 import { safeStringify } from '../../base/common/objects.js';
 import { IUtilityProcessWorkerWorkbenchService, UtilityProcessWorkerWorkbenchService } from '../services/utilityProcess/electron-browser/utilityProcessWorkerWorkbenchService.js';
-import { isBigSurOrNewer, isCI, isMacintosh } from '../../base/common/platform.js';
+import { isBigSurOrNewer, isCI, isMacintosh, isTahoeOrNewer } from '../../base/common/platform.js';
 import { Schemas } from '../../base/common/network.js';
 import { DiskFileSystemProvider } from '../services/files/electron-browser/diskFileSystemProvider.js';
 import { FileUserDataProvider } from '../../platform/userData/common/fileUserDataProvider.js';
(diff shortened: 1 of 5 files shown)


src/vs/workbench/electron-browser/desktop.main.ts
```diff
@@ -153,8 +153,13 @@ export class DesktopMain extends Disposable {
 	}
 
 	private getExtraClasses(): string[] {
-		if (isMacintosh && isBigSurOrNewer(this.configuration.os.release)) {
-			return ['macos-bigsur-or-newer'];
+		if (isMacintosh) {
+			if (isTahoeOrNewer(this.configuration.os.release)) {
```
deepak1556, line 160:
```suggestion
             // TODO: Revisit the border radius values till Electron v40 adoption
             // Refs https://github.com/electron/electron/issues/47514 and
             // https://github.com/microsoft/vscode/pull/270236#issuecomment-3379301185
			if (isTahoeOrNewer(this.configuration.os.release)) {
```

deepak1556 (dismissed):
LGTM, Thank you!

src/vs/workbench/electron-browser/desktop.main.ts
```diff
@@ -153,8 +153,16 @@ export class DesktopMain extends Disposable {
...
-			return ['macos-bigsur-or-newer'];
+		if (isMacintosh) {
+			// TODO: Revisit the border radius values till Electron v40 adoption
+			// Refs https://github.com/electron/electron/issues/47514 and
+			// https://github.com/microsoft/vscode/pull/270236#issuecomment-3379301185
+			if (isTahoeOrNewer(this.configuration.os.release)) {
+				return ['macos-tahoe-or-newer'];
+			}
+			if (isBigSurOrNewer(this.configuration.os.release)) {
+				return ['macos-bigsur-or-newer'];
```
Tyriar, line 164: @deepak1556 this class name isn't the best anymore as it won't be present when the version is tahoe, despite tahoe being bigsur or newer.

Tyriar (approved):

src/vs/workbench/electron-browser/desktop.main.ts
avarayr: I propose removing the `-or-newer` suffix and opting for `macos-rounded-bigsur`, `macos-rounded-tahoe` that sort of resembles tailwind-esque classes. But I'm not sure how worth it is.

Plus having `macos-rounded-bigsur` sounds fine even when running on Sequoia, because sequoia inherits bigsur border radius so it checks out

deepak1556: Good catch! I think we can rewrite the logic to match the version table we currently support on macOS

```js
if (isMacintosh) {
  if (isTahoeOrNewer(this.configuration.os.release)) {
       return ['macos-rounded-tahoe'];
  } else {
       return ['macos-rounded-default'];
  }
```

bpasero (approved):

PR #270262, opened 2025-10-07
Add `TM_DIRECTORY_BASE` snippets variable

danielbayley:
Adds `TM_DIRECTORY_BASE` to [snippets variables](https://code.visualstudio.com/docs/editing/userdefinedsnippets#_variables).

### DIFF
--- a/src/vs/editor/contrib/snippet/browser/snippetVariables.ts
+++ b/src/vs/editor/contrib/snippet/browser/snippetVariables.ts
@@ -41,6 +41,7 @@ export const KnownSnippetVariableNames = Object.freeze<{ [key: string]: true }>(
 	'TM_FILENAME': true,
 	'TM_FILENAME_BASE': true,
 	'TM_DIRECTORY': true,
+	'TM_DIRECTORY_BASE': true,
 	'TM_FILEPATH': true,
 	'CURSOR_INDEX': true, // 0-offset
 	'CURSOR_NUMBER': true, // 1-offset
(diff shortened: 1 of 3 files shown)


danielbayley:

src/vs/editor/contrib/snippet/browser/snippetVariables.ts
```diff
@@ -185,6 +186,9 @@ export class ModelBasedVariableResolver implements VariableResolver {
 			}
 			return this._labelService.getUriLabel(dirname(this._model.uri));
 
+		} else if (name === 'TM_DIRECTORY_BASE') {
+			const parentDirname = path.basename(path.dirname(this._model.uri.fsPath));
```
jrieken: `dirname` might return `.` and that case should be handled. Similar to how it is done here: https://github.com/microsoft/vscode/blob/020dc65e0d3dd3072a7f5c4278461d30eb67f3c1/src/vs/editor/contrib/snippet/browser/snippetVariables.ts#L183

jrieken (changes requested):

src/vs/editor/contrib/snippet/browser/snippetVariables.ts
danielbayley: Feels like this whole thing should be a `switch` statement… Anyway, handled now.

jrieken (changes requested):
Looking good. Let's make sure you also update this file here: https://github.com/microsoft/vscode/blob/020dc65e0d3dd3072a7f5c4278461d30eb67f3c1/src/vs/editor/contrib/snippet/browser/snippet.md#L29

jrieken (approved):

alexr00 (approved):

PR #270313, opened 2025-10-08
Remove deprecated managedhover from findwidget + toggle

Tyriar:
Part of #243348
Part of #270314


src/vs/base/browser/ui/toggle/toggle.ts
```diff
@@ -24,7 +23,11 @@ export interface IToggleOpts extends IToggleStyles {
 	readonly title: string;
 	readonly isChecked: boolean;
 	readonly notFocusable?: boolean;
+	// TODO: Remove this, the previous default was mouse, so anything not mouse needs to be explicit
+	/** @deprecated Prefer hoverStyle and hoverLifecycleOptions instead */
 	readonly hoverDelegate?: IHoverDelegate;
+	readonly hoverStyle?: HoverStyle;
+	readonly hoverLifecycleOptions?: IHoverLifecycleOptions;
```
Tyriar: @benibenj what do you think about this as a way to get rid of hoverDelegate in base? I think base only needs to set a groupId and a style (pointer or mouse). We don't want to be passing around callbacks to `setupDelayedHover` or `setupDelayedHoverAtMouse` inside options so simple options like this are preferable.

Tyriar: I build upon this in https://github.com/microsoft/vscode/pull/270749 attempting to do the same for `IBaseActionViewItemOptions` but I might abandon it there as it's a lot of work to validate and a huge change that can't really be done in stages AFAICT and I wanted to get your view on the `HoverStyle` and lifecycle options replacing `hoverDelegate`.

benibenj (commented):
I like the idea and looks good overall, but wonder if we should reduce the options we provide (or pick recommended defaults for each widget) to ensure consistency across the different usages of the same base widgets

src/vs/base/browser/ui/toggle/toggle.ts
benibenj: How about removing `hoverStyle` and having it always use a `Pointer`. I do not see a use case where we would want to use `Mouse` style for a toggle. This would help with consistency

Tyriar: Example: 
image

There are a bunch of them, seems like a pretty radical and scary change to force pointer?

Tyriar: Note we also have https://github.com/microsoft/vscode/issues/243348 still to tweak defaults.

Tyriar: Switched all toggles to use pointer, keeping it on `IHoverService` and not changing the default though.

lszomoru (approved):

PR #270333, opened 2025-10-08
Do not prefix `Folders:` when explorer merges with container

benibenj:
```Copilot Generated Description:``` Remove the prefix `Folders:` from the title when the explorer merges with the container. Additionally, override the `singleViewPaneContainerTitle` method to return the name directly.

fixes https://github.com/microsoft/vscode/issues/269862

### DIFF
--- a/src/vs/workbench/contrib/files/browser/views/explorerView.ts
+++ b/src/vs/workbench/contrib/files/browser/views/explorerView.ts
@@ -185,7 +185,7 @@ export class ExplorerView extends ViewPane implements IExplorerView {
 	private decorationsProvider: ExplorerDecorationsProvider | undefined;
 	private readonly delegate: IExplorerViewContainerDelegate | undefined;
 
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/vs/workbench/contrib/files/browser/views/explorerView.ts [hunk 1]
```diff
@@ -185,6 +185,10 @@ export class ExplorerView extends ViewPane implements IExplorerView {
 	private decorationsProvider: ExplorerDecorationsProvider | undefined;
 	private readonly delegate: IExplorerViewContainerDelegate | undefined;
 
+	public override get singleViewPaneContainerTitle(): string {
```
bpasero:
```suggestion
	override get singleViewPaneContainerTitle(): string {
```

bpasero (dismissed):

bpasero (approved):

PR #270361, opened 2025-10-08
Fix disassembly view

DrSergei:
This patch fixes two problems in dissembly view.

The first issue is caused by the `disassemblyNotAvailable` entry. When a new debug session starts, this entry is pushed to the view, which triggers an `onDidScroll` event. This leads to a call to the `disassemble` request via `scrollUp_LoadDisassembledInstructions` function. However, an empty `memoryReference` from the `disassemblyNotAvailable` entry is used for that that leads to error on debug adapter side.

Log from debug adapter (lldb-dap)
```
1759923554.517830610 (stdio) --> {"command":"disassemble","arguments":{"memoryReference":"","offset":0,"instructionOffset":-50,"instructionCount":50,"resolveSymbols":true},"type":"request","seq":3}
1759923554.518007517 (stdio) queued (command=disassemble seq=3)
1759923554.518254757 (stdio) <-- {"body":{"error":{"format":"invalid arguments for request 'disassemble': malformed memory reference at arguments.memoryReference\n{\n  \"instructionCount\": 50,\n  \"instructionOffset\": -50,\n  \"memoryReference\": /* error: malformed memory reference */ \"\",\n  \"offset\": 0,\n  \"resolveSymbols\": true\n}","id":3,"showUser":true}},"command":"disassemble","request_seq":3,"seq":0,"success":false,"type":"response"}
```

The second issue is a successor to the first. If the `disassemble` request fails, the ` _loadingLock` remains permanently true, which blocks all further scrolling.

Before:

before

After:

after


dmitrivMS (approved):
LGTM

connor4312 (approved):
lgtm, thanks for the contribution!

roblourens (approved):

PR #270368, opened 2025-10-08
extract `setupRecreatingStartMarker` from terminal execution strategies

### LINKED ISSUE #269551: Lots of repetition here across files here, should this live in an executeStrategy/strategyHelpers.ts so they don't get out of sync?
Lots of repetition here across files here, should this live in an executeStrategy/strategyHelpers.ts so they don't get out of sync?

_Originally posted by @Tyriar in https://github.com/microsoft/vscode/pull/269385#discussion_r2396332572_

meganrogge:
fixes #269551

cc @tyriar


src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/executeStrategy/noneExecuteStrategy.ts
```diff
@@ -35,6 +36,7 @@ export class NoneExecuteStrategy implements ITerminalExecuteStrategy {
 
 	async execute(commandLine: string, token: CancellationToken): Promise<ITerminalExecuteStrategyResult> {
 		const store = new DisposableStore();
+
```
meganrogge:
```suggestion
```

ulugbekna (approved):

Issue #112595, opened 2020-12-15
Debug console sends evalution to parent session which fails


isidorn:
1. Create two js debug sessions
2. Open debug console, manually change the dropdown to the other session
3. try to evaluate something in the debug console -> vs code sends it against the parent session gets an error and no evaluation is shown

isidorn:
If the session is merged with the parent session we  should always send to the child session.
It seems like the drop down incorrectly chooses the parent session

weinand:
@isidorn you said
"If the session is merged with the parent session..."

But in the initial comment I do not see how two unrelated sessions could have a parent/child relation?
The initial problem does not occur if two sessions run in two debug consoles, right?

isidorn:
The two unrelated session do not have a parent/child relation between them.

However each of those two sessions has a parent child relationship with it's wrapper session (since this is using js-debug).

So yes the initial problem was ocurring when those two sessions are running in two debug consoles.

weinand:
Thanks, now I understand. And since we hide the parent now in one-child cases, I was not seeing the parent/child relation anymore...

isidorn:
Correct, sorry for not including the details in the original comment.

PR #270461, opened 2025-10-09
fix: debugger is unresponsive to repl evaluation after switching

### LINKED ISSUE #152407: Using debug console picker when not paused leaves console unresponsive
- Start two node launch configs
- Both sessions are running and not paused
- Use the debug console switcher to switch consoles
- Evaluate `1+1`, there is no response

In the console picker, we show parent sessions. The child needs to be focused to get the evaluate requests. When a child is paused, we will focus that one with this code

https://github.com/microsoft/vscode/blob/16c2a3ab3be4cf89c71fcce1ff1c11a06fcb923a/src/vs/workbench/contrib/debug/browser/repl.ts#L890-L896

I think this is the same as https://github.com/microsoft/vscode/issues/112595 except we still need to pick a child when not paused, and it also doesn't work when a deeper child is stopped instead.

https://github.com/microsoft/vscode/issues/138963 is relevant too

Basically when something tries to focus a new wrapper parent session, we should always focus one of its children instead. cc @isidorn

bryanchen-d:
Fixes https://github.com/microsoft/vscode/issues/152407:

## Problem Analysis
The issue occurred when using the debug console picker to switch between consoles. If the selected session was a parent session (wrapper) with child sessions, the console would become unresponsive because it focused the parent session instead of a child that can actually handle evaluation requests.

## Root Cause
The existing logic in both repl.ts and debugCommands.ts only looked for stopped child sessions, but ignored running child sessions. When both parent and children were running (not paused), the system would focus the parent session, which couldn't handle REPL evaluations properly.

## Solution

* Always check if the target session has child sessions
* Prefer stopped child sessions when available (preserves fix for #112595)
* Fallback to the first available child session if no stopped children exist
* Only focus parent if there are no child sessions at all


src/vs/workbench/contrib/debug/browser/debugCommands.ts
```diff
@@ -711,10 +711,7 @@ CommandsRegistry.registerCommand({
 	handler: async (accessor: ServicesAccessor, session: IDebugSession) => {
 		const debugService = accessor.get(IDebugService);
 		const editorService = accessor.get(IEditorService);
-		const stoppedChildSession = debugService.getModel().getSessions().find(s => s.parentSession === session && s.state === State.Stopped);
-		if (stoppedChildSession && session.state !== State.Stopped) {
-			session = stoppedChildSession;
-		}
+		session = resolveChildSession(session, debugService.getModel().getSessions());
```
roblourens, line 714: What's the difference between this command and the one in repl.ts?

roblourens (commented):
This looks good but I will test it a bit later

src/vs/workbench/contrib/debug/browser/debugCommands.ts
bryanchen-d: From what I understand, the repl.ts one is responsible for when user executes an expression in the repl pannel, and the command one is for general debugging context - like switching debugging session using the debug toolbar, etc. 
However I couldn't find similar issue for the debugCommand one, it seems better to have it consistent though.

roblourens (approved):

PR #270580, opened 2025-10-09
Support textDirection in languageConfiguration for RTL languages

abdelfetah18:
This PR fixes #268888 by adding a new `textDirection` field to [languageConfiguration](https://code.visualstudio.com/api/language-extensions/language-configuration-guide). It supports `"ltr"` (default) and `"rtl"`. See the linked issue for details.


alexdima (changes requested):
🤔 Why is this configurable in the language-configuration file? I would expect this to be configurable through a setting, like e.g. file encoding or end-of-line sequence, etc. Would you please mind changing it to a setting?

PR #270625, opened 2025-10-09
fix: search in open editors for files with square brackets in names

### LINKED ISSUE #233049: Search Only in Open Editors does not work if file name has square bracket

- VS Code Version: 1.94.2
- OS Version: Windows 11

Steps to Reproduce:

1. Name a file with closed or open square brackets []
2. Open in editor and search only in open editors
Image

bryanchen-d:
Files with square brackets in their names (e.g.,`foo[.js`,  `component[id].tsx`) leads to unexpected error when using "Search Only in Open Editors" because square brackets were interpreted as glob character classes instead of literal characters.

Fixes #233049

## Root Cause
When "Search Only in Open Editors" builds search queries, it creates include patterns from open file paths to restrict search scope. However, these file paths were used directly as glob patterns without escaping special characters. Square brackets [...] have special meaning in glob syntax - they define character classes that match any single character from the bracketed set. So a file named component[id].tsx created the pattern component[id].tsx, which the glob matcher interpreted as "component" + (any character from {i,d}) + ".tsx", failing to match the literal filename component[id].tsx.

## Solution
Applied the existing escapeGlobPattern function to file paths before adding them to include patterns in the commonQueryFromFileList method. This function escapes special glob characters by wrapping them in brackets - converting component[id].tsx to component[[]id[]].tsx, where [[] represents a literal opening bracket and []] represents a literal closing bracket. The escaped pattern now matches the literal filename correctly. Added comprehensive tests to verify the fix handles square brackets properly without breaking existing functionality.

### DIFF
--- a/src/vs/workbench/services/search/test/browser/queryBuilder.test.ts
+++ b/src/vs/workbench/services/search/test/browser/queryBuilder.test.ts
@@ -1201,6 +1201,8 @@ suite('QueryBuilder', () => {
 			});
 			assert.strictEqual(query.folderQueries.length, 1);
 		});
+
+
 	});
 });
 function makeExcludePatternFromPatterns(...patterns: string[]): {
--- a/src/vs/workbench/services/search/test/browser/queryBuilder.test.ts
+++ b/src/vs/workbench/services/search/test/browser/queryBuilder.test.ts
@@ -1201,8 +1201,6 @@ suite('QueryBuilder', () => {
 			});
 			assert.strictEqual(query.folderQueries.length, 1);
 		});
-
-
 	});
 });
 function makeExcludePatternFromPatterns(...patterns: string[]): {
--- a/src/vs/workbench/services/search/test/common/queryBuilder.test.ts
+++ b/src/vs/workbench/services/search/test/common/queryBuilder.test.ts
@@ -3,6 +3,7 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 import assert from 'assert';
+import * as glob from '../../../../../base/common/glob.js';
 import { isWindows } from '../../../../../base/common/platform.js';
 import { URI } from '../../../../../base/common/uri.js';
 import { ensureNoDisposablesAreLeakedInTestSuite } from '../../../../../base/test/common/utils.js';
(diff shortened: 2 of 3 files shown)


src/vs/workbench/services/search/test/browser/queryBuilder.test.ts
```diff
@@ -1201,6 +1201,8 @@ suite('QueryBuilder', () => {
 			});
 			assert.strictEqual(query.folderQueries.length, 1);
 		});
+
+
```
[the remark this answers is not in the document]
bryanchen-d: removed

src/vs/workbench/services/search/test/common/queryBuilder.test.ts
```diff
@@ -29,4 +30,27 @@ suite('QueryBuilderCommon', () => {
...
+		// This test verifies the fix for issue #233049 where files with square brackets in names
+		// were not found when using "Search Only in Open Editors"
+
+		// Test file name with square brackets
+		const fileName = 'file[test].txt';
+
+		// Function matching the one used in queryBuilder.ts
+		function escapeGlobPattern(path: string): string {
+			return path.replace(/([?*[\]])/g, '[$1]');
+		}
```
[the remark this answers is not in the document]
bryanchen-d: This one is in the test so it can be a second validation to the logic from queryBuild.ts.

src/vs/workbench/services/search/test/common/queryBuilder.test.ts
```diff
@@ -29,4 +30,27 @@ suite('QueryBuilderCommon', () => {
...
+
+	test('escapeGlobPattern properly escapes square brackets for literal matching', () => {
+		// This test verifies the fix for issue #233049 where files with square brackets in names
+		// were not found when using "Search Only in Open Editors"
+
+		// Test file name with square brackets
+		const fileName = 'file[test].txt';
+
+		// Function matching the one used in queryBuilder.ts
+		function escapeGlobPattern(path: string): string {
```
roblourens: But then this isn't testing the real function right?

bryanchen-d: You are right, exported the escapeGlobPattern from queryBuild for testing.

roblourens: I meant something more along the lines of calling into the query builder, building a query in the way that they get built in this scenario in the product, but this is ok

roblourens (approved):

Issue #259868, opened 2025-08-05
make notebook inline agent independent of inline chat v2 setting


amunger:
We should be able to use the new notebook inline agent tools without needing to enable the mostly unrelated setting for inline chat v2

56789KD:
Issue a change view comment

PR #270658, opened 2025-10-09
give inline notebook chat its own setting for using v2 controller

amunger:
https://github.com/microsoft/vscode/issues/259868

corresponding chat PR https://github.com/microsoft/vscode-copilot-chat/pull/1320

### DIFF
--- a/src/vs/workbench/contrib/inlineChat/browser/inlineChatActions.ts
+++ b/src/vs/workbench/contrib/inlineChat/browser/inlineChatActions.ts
@@ -11,7 +11,7 @@ import { EmbeddedDiffEditorWidget } from '../../../../editor/browser/widget/diff
 import { EmbeddedCodeEditorWidget } from '../../../../editor/browser/widget/codeEditor/embeddedCodeEditorWidget.js';
 import { EditorContextKeys } from '../../../../editor/common/editorContextKeys.js';
 import { InlineChatController, InlineChatController1, InlineChatController2, InlineChatRunOptions } from './inlineChatController.js';
-import { ACTION_ACCEPT_CHANGES, CTX_INLINE_CHAT_HAS_AGENT, CTX_INLINE_CHAT_HAS_STASHED_SESSION, CTX_INLINE_CHAT_FOCUSED, CTX_INLINE_CHAT_INNER_CURSOR_FIRST, CTX_INLINE_CHAT_INNER_CURSOR_LAST, CTX_INLINE_CHAT_VISIBLE, CTX_INLINE_CHAT_OUTER_CURSOR_POSITION, MENU_INLINE_CHAT_WIDGET_STATUS, CTX_INLINE_CHAT_REQUEST_IN_PROGRESS, CTX_INLINE_CHAT_RESPONSE_TYPE, InlineChatResponseType, ACTION_REGENERATE_RESPONSE, ACTION_VIEW_IN_CHAT, ACTION_TOGGLE_DIFF, CTX_INLINE_CHAT_CHANGE_HAS_DIFF, CTX_INLINE_CHAT_CHANGE_SHOWS_DIFF, MENU_INLINE_CHAT_ZONE, ACTION_DISCARD_CHANGES, CTX_INLINE_CHAT_POSSIBLE, ACTION_START, CTX_INLINE_CHAT_HAS_AGENT2, MENU_INLINE_CHAT_SIDE } from '../common/inlineChat.js';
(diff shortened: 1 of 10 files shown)


src/vs/workbench/contrib/inlineChat/browser/inlineChatActions.ts
```diff
@@ -637,7 +637,7 @@ class KeepOrUndoSessionAction extends AbstractInline2ChatAction {
 				id: MENU_INLINE_CHAT_WIDGET_STATUS,
 				group: '0_main',
 				order: 1,
-				when: ContextKeyExpr.and(CTX_INLINE_CHAT_HAS_AGENT2, ContextKeyExpr.greater(ctxRequestCount.key, 0), ctxHasEditorModification),
+				when: ContextKeyExpr.and(ContextKeyExpr.greater(ctxRequestCount.key, 0), ctxHasEditorModification),
```
jrieken, line 640: I think we should keep the V2 condition, otherwise these actions might show up in the v1 UI

src/vs/workbench/contrib/inlineChat/browser/inlineChatActions.ts
```diff
@@ -696,12 +696,12 @@ export class CloseSessionAction2 extends AbstractInline2ChatAction {
 			menu: [{
 				id: MENU_INLINE_CHAT_SIDE,
 				group: 'navigation',
-				when: ContextKeyExpr.and(CTX_INLINE_CHAT_HAS_AGENT2, ctxRequestCount.isEqualTo(0)),
+				when: ContextKeyExpr.and(ctxRequestCount.isEqualTo(0)),
```
jrieken, line 699: same

src/vs/workbench/contrib/inlineChat/browser/inlineChatActions.ts
```diff
@@ -696,12 +696,12 @@ export class CloseSessionAction2 extends AbstractInline2ChatAction {
...
 			}, {
 				id: MENU_INLINE_CHAT_WIDGET_STATUS,
 				group: '0_main',
 				order: 1,
-				when: ContextKeyExpr.and(CTX_INLINE_CHAT_HAS_AGENT2, ctxHasEditorModification.negate()),
+				when: ContextKeyExpr.and(ctxHasEditorModification.negate()),
```
jrieken, line 704: same

src/vs/workbench/contrib/inlineChat/browser/inlineChatController.ts
```diff
@@ -1315,16 +1317,14 @@ export class InlineChatController2 implements IEditorContribution {
 						location.location = ChatAgentLocation.Notebook;
```
jrieken: could use the new `isCellEditor` method

src/vs/workbench/contrib/inlineChat/browser/inlineChatSessionServiceImpl.ts
```diff
@@ -449,6 +455,11 @@ export class InlineChatEnabler {
 			}
 		}));
 
+		this._store.add(autorun(r => {
+			this._ctxHasNotebookInline.set(!notebookAgentConfigObs.read(r) && !!agentObs.read(r));
```
jrieken, line 459: The docs for this setting say it depends on the v2-setting, but I don't think that true, right?

amunger: no, need to update that

src/vs/workbench/contrib/inlineChat/browser/inlineChatActions.ts
amunger: the parent action `AbstractInline2ChatAction` adds `CTX_INLINE_CHAT_V2_ENABLED` to all menu when clauses with `massageMenu`

src/vs/workbench/contrib/inlineChat/browser/inlineChatController.ts
amunger: `isCellEditor` is not behaving as hoped, perhaps because the model doesn't necessarily exist when the check is hit, so it wasn't always creating the correct controller type.(Also, somehow the delegate is retrieved for three different editors, all with distinct _id's, when the open inline chat action is run)

amunger: never mind, I had a bug in that method that just checked the first cell editor. It works now

rebornix (approved):

jrieken (approved):

PR #270660, opened 2025-10-09
Filter unsupported items from 'Add Context' quick pick

benvillalobos:
Fixes https://github.com/microsoft/vscode/issues/255443

## Context
If you add an unsupported view (webview, settings pane, welcome page) to the context via quick pick, it gets included in context but shows up as an error.
image

## Changes Made
Added a filter to the quick pick that excludes views with specific schemes.


eleanorjboyd (approved):

src/vs/workbench/contrib/chat/browser/actions/chatContextActions.ts
```diff
@@ -51,6 +51,9 @@ import { resizeImage } from '../imageUtils.js';
 import { registerPromptActions } from '../promptSyntax/promptFileActions.js';
 import { CHAT_CATEGORY } from './chatActions.js';
 
+// Schemes that should not be available for chat context attach
+const UNSUPPORTED_CONTEXT_SCHEMES = new Set(['webview-panel', 'walkThrough', 'vscode-settings']);
```
roblourens, line 55: We have this list here: https://github.com/microsoft/vscode/blob/a3fc87930654143c8c775d2a1a791ac985cf3db2/src/vs/workbench/contrib/chat/common/constants.ts#L88 because there are multiple pathways for files to be attached

PR #270777, opened 2025-10-10
Prevent flicker when opening a webview in a panel

### LINKED ISSUE #270776: Flicker when opening a panel

- VS Code Version: 1.105.0 web
- OS web Version: Ubuntu 24.04

Steps to Reproduce:

1. Switch active panel
2. The panel is first rendered outside (at the bottom) of the rendering area, they relayout in the right place

It can be easily reproduce by adding a breakpoint in AbstractPaneCompositePart.showComposite

CGNonofr:
Because the empty message div is hidden AFTER the actual webview is layout, the first webview layout occurs while it's placed outside of the workbench (at the bottom on the empty message div).

So another layout is required to place it correctly, leading to a flicker

fix #270776

### DIFF
--- a/src/vs/workbench/browser/parts/paneCompositePart.ts
+++ b/src/vs/workbench/browser/parts/paneCompositePart.ts
@@ -231,15 +231,15 @@ export abstract class AbstractPaneCompositePart extends CompositePart<PaneCompos
 	}
 
 	protected override showComposite(composite: Composite): void {
+		this.layoutEmptyMessage(false);
 		super.showComposite(composite);
 		this.layoutCompositeBar();
-		this.layoutEmptyMessage();
 	}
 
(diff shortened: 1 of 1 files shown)


src/vs/workbench/browser/parts/paneCompositePart.ts
```diff
@@ -231,15 +231,15 @@ export abstract class AbstractPaneCompositePart extends CompositePart<PaneCompos
 	}
 
 	protected override showComposite(composite: Composite): void {
+		this.layoutEmptyMessage(false);
```
CGNonofr, line 234: The parameter is mandatory, because at that point, `this.getActiveComposite()` still returns null (for overlay webviews at least)

mjbvz (approved):
Thanks!

PR #270821, opened 2025-10-10
wip terminal inline chat

meganrogge:
Branched from https://github.com/microsoft/vscode/pull/261503 because there were many merge conflicts and I was nervous to break something

To do:
- how will this be accessible?
- fix width, update on chat width changes
- restore on window reload
- don't render instance in panel


src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
```diff
@@ -4,19 +4,31 @@
...
 import { IMarkdownRenderer } from '../../../../../../platform/markdown/browser/markdownRenderer.js';
 import { IInstantiationService } from '../../../../../../platform/instantiation/common/instantiation.js';
+import { IKeybindingService } from '../../../../../../platform/keybinding/common/keybinding.js';
+import { TerminalCapabilityStore } from '../../../../../../platform/terminal/common/capabilities/terminalCapabilityStore.js';
+import { TerminalLocation } from '../../../../../../platform/terminal/common/terminal.js';
 import { IPreferencesService, type IOpenSettingsOptions } from '../../../../../services/preferences/common/preferences.js';
+import type { ITerminalInstance } from '../../../../terminal/browser/terminal.js';
+import { TerminalInstance, TerminalInstanceColorProvider } from '../../../../terminal/browser/terminalInstance.js';
+import { XtermTerminal } from '../../../../terminal/browser/xterm/xtermTerminal.js';
+// TODO@meganrogge fix
```
meganrogge: To do: fix somehow, this was also in initial PR

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/executeStrategy/strategyHelpers.ts
```diff
@@ -38,8 +38,11 @@ export function setupRecreatingStartMarker(
...
-		startMarker.clear();
-		fire(undefined);
+		// TODO@meganrogge !!! does this cause disposables to be leaked?
+		// Don't clear the start marker on disposal - we want it to persist
+		// so terminal UI parts can still access it
+		console.log('strategy disposed but keeping start marker');
+		// Don't fire undefined marker event - keep the last valid marker
 	}));
-	store.add(startMarker);
+	// We don't register startMarker with the store to prevent it from being auto-disposed
```
meganrogge, line 47: todo: fix somehow, with this clearing code, it doesn't work.

But the peristentMarker thing should make this ok 🤔 . maybe timing issue

src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
```diff
@@ -27,12 +39,187 @@ import { CommandsRegistry } from '../../../../../../platform/commands/common/com
...
+	private static readonly trackingInstances = new Map<string, {
+		instance: ITerminalInstance;
+		executeStrategy: ITerminalExecuteStrategy;
+	}>();
+
+	// Map from real terminal instance IDs to chat terminal parts
+	private static readonly instanceToPartMap = new Map<string, Set<ChatTerminalToolProgressPart>>();
+
+	// Track the latest part for each terminal instance type (command, output display, etc)
+	private static readonly latestPartPerInstance = new Map<string, string>();
```
Tyriar: This thing was really hacky just to get something working, we need to come up with a proper way of transferring the terminal over to the tool. Since we don't necessarily know the instance to be used before this object is constructed it makes things more complicated.

Perhaps a new `ITerminalChatService` who's job it is initially purely to track these things and make sure they're serialized correctly (so things don't break on reload). Might be worth chatting about this in a sync together.

Tyriar (changes requested):

src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
meganrogge: Agreed, and this file is getting way too big, needs separation.

Here are some notes I made:

```
// Chat Session
//  ├── Chat Request → Xterm → Instance (start marker → end marker)
//  ├── Chat Request → Xterm → Instance (start marker → end marker)
//  └── Chat Request → Xterm → Instance (start marker → end marker)
// ...

// Notes:
// - Each Instance should be headless, hidden, and per Chat Request?
// - Should allow opening terminals from the chat-embedded one?

interface ChatEmbeddedTerminalService {
 map: Map<ChatSessionId, Xterm[]>;

 createTerminal(): Xterm;
 serialize(): string;
 deserialize(data: string): void;
```

PR #270827, opened 2025-10-10
Terminal suggest addon should always multiply line height config

### LINKED ISSUE #255851: Icon are clipped in terminal suggest when using lineHeight
Repro

```
"terminal.integrated.lineHeight": 1.1,
```

Image

Without that it's fine:

Image

1.1 is meant to multiply the default line height by 1.1, and it looks like it's making it smaller. The line height settings in terminal and editor work a little differently from each other.

pwang347:
| Before | After |
| -- | -- |
| image | image |

Based on the configuration description the treatment here should differ from regular editor suggest line heights:
image

Compare with:
image

Fixes https://github.com/microsoft/vscode/issues/255851


meganrogge (approved):
Thank you!

PR #270839, opened 2025-10-10
Fix issues with monitoring dependency tasks

### LINKED ISSUE #269451: Agent fails to give input to compound tasks consistently
Found while verifying https://github.com/microsoft/vscode/issues/264656

Notice how the `tunnel web` task doesn't require input but `serve web does`. The agent _says_ it looks for `serve web` task output but doesn't notice the "ok to proceed?". This behavior consistently repros in the scenario where `tunnel web` worked but `serve web` didn't.

https://github.com/user-attachments/assets/e2ed8c5a-3069-4a7a-a1cf-944c777a8383

pwang347:
## Problem Overview
- We are setting up the output monitor on the wrong task (should use the dependent tasks since those map to the terminal instances). This made it so that `isActive` always returned false since the parent task is never in the active list.

## Changes
- Updated to point to the correct dependency task
- Noticed sometimes the title was not showing up properly, so added a fallback
- Small optimization was also to check for (y) scenarios, which is used for npm install from npx.

Fixes #269451


Tyriar:
Just @meganrogge is good to handle task PRs

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/monitoring/outputMonitor.ts
```diff
@@ -318,14 +313,10 @@ export class OutputMonitor extends Disposable implements IOutputMonitor {
 				const recentlyIdle = consecutiveIdleEvents >= PollingConsts.MinIdleEvents;
 				const isActive = execution.isActive ? await execution.isActive() : undefined;
 				this._logService.trace(`OutputMonitor: waitForIdle check: waited=${waited}ms, recentlyIdle=${recentlyIdle}, isActive=${isActive}`);
-				// Keep polling if still active with no recent data
```
meganrogge, line 321: This code was added to prevent the agent from continuing on when the task terminal was idle for several seconds, but the task itself was still active. Have you tested that case?

meganrogge: Was this necessary for the fix? It would be best to bring this back.

pwang347: Good catch - I think I misunderstood the intent here and this check is needed. Thanks!

meganrogge (dismissed):
Thank you, good finds!

meganrogge (approved):

PR #270844, opened 2025-10-11
Add support for navigating to absolute character offset in the editor with "::" shortcut

### LINKED ISSUE #196067: Add way to "Go to offset", ideally accessible from the "Go to" command
It sometimes useful to be able to go a certain offset of a file (e.g. when a tool reports "error xxx at position 599").

This has been implemented for [vscode-hexeditor](https://github.com/microsoft/vscode-hexeditor/pull/203) and it would be nice to have this mainstream and built-in (rather than having to depend on some extensions ... that might be buggy).

A nice way to access it would be from Go-to (<kbd>Ctrl</kbd> + <kbd>G</kbd>) as it is the case in Notepad++ for example, maybe like:
* `:n` would go to line n
* `:n:c` would go to character  c of line n
* `::c` would go to the character c of the whole file (i.e. offset)

The first two being already implemented, that would be the logical continuation

dmitrivMS:
Fixes #196067
"::offset" will navigate to corresponding character.
Supports "::-offset" to navigate to a position from the end of the buffer.


src/vs/editor/contrib/quickAccess/browser/gotoLineQuickAccess.ts
```diff
@@ -3,29 +3,42 @@
...
+		return this._useZeroBasedOffset;
+	}
+
+	set useZeroBasedOffset(value: boolean) {
+		this._useZeroBasedOffset = value;
+	}
+
 	protected provideWithoutTextEditor(picker: IQuickPick<IGotoLineQuickPickItem, { useSeparators: true }>): IDisposable {
-		const label = localize('cannotRunGotoLine', "Open a text editor first to go to a line.");
+		const label = localize('cannotRunGotoLine', "Open a text editor first to go to a line or offset.");
```
TylerLeonhardt:
```suggestion
		const label = localize('cannotRunGotoLine', "Open a text editor first to go to a line.");
```

At the end of the day, the previous text was correct. It also didn't say "go to a line or column". Column being supported.

Let's keep this simple to understand.

src/vs/editor/contrib/quickAccess/browser/gotoLineQuickAccess.ts
```diff
@@ -110,6 +147,24 @@ export abstract class AbstractGotoLineQuickAccessProvider extends AbstractEditor
...
+			let offset = parseInt(value.substring(1), 10);
+			if (!isNaN(offset)) {
+				const model = this.getModel(editor);
+				if (model) {
+					if (offset >= 0) {
+						// If offset is 1-based, we need to convert to model's 0-based.
+						offset -= this.useZeroBasedOffset ? 0 : 1;
+					} else {
+						// Offset from the end of the buffer
+						offset += model.getValueLength();
```
TylerLeonhardt, line 154: Does zero-based offset apply in the negative case as well?

src/vs/workbench/browser/workbench.contribution.ts
```diff
@@ -512,6 +512,12 @@ const registry = Registry.as<IConfigurationRegistry>(ConfigurationExtensions.Con
 				'description': localize('workbench.quickOpen.preserveInput', "Controls whether the last typed input to Quick Open should be restored when opening it the next time."),
 				'default': false
 			},
+			'workbench.quickOpen.useZeroBasedOffset': {
```
TylerLeonhardt, line 515: My feeling is, I don't know if I love the name of this setting... both the:
* `quickOpen` ... which, the settings are a bit of a mess, Quick Open I guess applies to `Go To Line` but ... idk I usually associate Quick Open with the AnythingQuickAccess.
* `useZeroBasedOffset` ... keep thinking, do I wish this was an enum with `one` and `zero` as options... but I don't love that either.

Ultimately, I don't have a better idea yet. This is fine for now.

src/vs/editor/contrib/quickAccess/browser/gotoLineQuickAccess.ts
dmitrivMS: I cannot come up with a way/reason to do so, but if you can tell me what that means, I can add it :-)

src/vs/workbench/browser/workbench.contribution.ts
dmitrivMS: Yeah, there is nothing with "goto" in the name AFAIK, so I ended up adding to the closest group.

src/vs/editor/contrib/quickAccess/browser/gotoLineQuickAccess.ts
dmitrivMS: Reverted

src/vs/editor/contrib/quickAccess/browser/gotoLineQuickAccess.ts
TylerLeonhardt: If I have a file with:
```
12345
```
and I do `::1` it places it here:
```
|12345
```
At the beginning. But if I do `::-1` it places it here:
```
1234|5
```

Now, with 0-based indexing, and I do `::1` it places it here:
```
1|2345
```
At the beginning. But if I do `::-1` it still places it here:
```
1234|5
```

To me, it feels like one of these are wrong. My expectation is that there would be away to get to:
```
12345|
```

My leaning is that this would be the behavior for 1-indexing so that it's a mirror. This actually makes 1-indexing superior because you cannot do `-0` in zero indexing.

src/vs/editor/contrib/quickAccess/browser/gotoLineQuickAccess.ts
```diff
@@ -133,6 +188,12 @@ export abstract class AbstractGotoLineQuickAccessProvider extends AbstractEditor
...
 		const position = editor.getPosition() || { lineNumber: 1, column: 1 };
+
+		// When in offset mode, prompt for an offset.
+		if (inOffsetMode) {
+			return localize('gotoLineOffsetLabel', "Current Line: {0}, Character: {1}. Type a character offset to navigate to.", position.lineNumber, position.column);
+		}
+
 		const lineCount = this.lineCount(editor);
 		if (lineCount > 1) {
 			return localize('gotoLineLabelEmptyWithLimit', "Current Line: {0}, Character: {1}. Type a line number between 1 and {2} to navigate to.", position.lineNumber, position.column, lineCount);
```
TylerLeonhardt, line 192: Image

I wish this included text around navigating to the offset. I wonder if a prompt makes sense for this. This can be done later though

src/vs/workbench/contrib/codeEditor/browser/quickaccess/gotoLineQuickAccess.ts
```diff
@@ -34,6 +35,14 @@ export class GotoLineQuickAccessProvider extends AbstractGotoLineQuickAccessProv
 		this.onDidActiveTextEditorControlChange = this.editorService.onDidActiveEditorChange;
 	}
 
+	override get useZeroBasedOffset(): boolean {
+		return this.configurationService.getValue<boolean>(GotoLineQuickAccessProvider.zeroBasedOffsetSetting);
+	}
+
+	override set useZeroBasedOffset(value: boolean) {
+		this.configurationService.updateValue(GotoLineQuickAccessProvider.zeroBasedOffsetSetting, value);
+	}
```
TylerLeonhardt: There's one trap in this design where if I use `this._useZeroBasedOffset` in the base class, I won't get the correct value.

dmitrivMS: Updated to pass it in constructor.

src/vs/editor/contrib/quickAccess/browser/gotoLineQuickAccess.ts
dmitrivMS: We removed note about offset from another message, so I think keeping it consistent would be good.

src/vs/editor/contrib/quickAccess/browser/gotoLineQuickAccess.ts
dmitrivMS: Updated it to be consistent, so that 1-based "-1" will go to "12345|" which seems to be the right behavior.
To summarize:
- In either mode "0" goes to "|12345"
- 0-based "1" goes to "1|2345"
- 0-based "-1" goes to "1234|5"
- 1-based "1" goes to "|12345"
- 1-based "-1" goes to "12345|"

TylerLeonhardt (approved):

PR #270876, opened 2025-10-11
chat - fix height compute of tree

bpasero:
The `gap: 4px` directive for the chat input part needs to be accounted for.

In addition, the tree container itself should not require to get a height applied provided its parent list container has a proper height.

This fixes issues where the blue outline box would be cut off at the bottom.

### DIFF
--- a/src/vs/workbench/contrib/chat/browser/chatWidget.ts
+++ b/src/vs/workbench/contrib/chat/browser/chatWidget.ts
@@ -2606,8 +2606,7 @@ export class ChatWidget extends Disposable implements IChatWidget {
 
 		if (this.container.classList.contains('new-welcome-view')) {
 			this.inputPart.layout(layoutHeight, Math.min(width, 650));
-		}
-		else {
+		} else {
 			this.inputPart.layout(layoutHeight, width);
 		}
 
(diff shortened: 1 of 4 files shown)


src/vs/workbench/contrib/chat/browser/chatWidget.ts
```diff
@@ -2622,7 +2621,6 @@ export class ChatWidget extends Disposable implements IChatWidget {
 			this.listContainer.style.setProperty('--chat-current-response-min-height', contentHeight * .75 + 'px');
 		}
 		this.tree.layout(contentHeight, width);
-		this.tree.getHTMLElement().style.height = `${contentHeight}px`;
```
roblourens, line 2625: I remember this being needed for some reason? Maybe not. I haven't run it, I'll try it on Monday

roblourens (approved):

src/vs/workbench/contrib/chat/browser/chatWidget.ts
bpasero: @roblourens yeah lets see, you don't typically give the tree HTML element a height from what I can tell, it should be sufficient to call `layout` with the right height. Also, the `listContainer` is getting this height, so it should be fine. We will know next week.

This fixes a real issue where the blue focus outline would be cropped at the bottom when you select the entire tree (not when you select an individual response).

PR #271020, opened 2025-10-12
Ai app modification

Jamestran1304:

### DIFF
--- a/setup-ai-app.bat
+++ b/setup-ai-app.bat
@@ -1,0 +1,86 @@
+@echo off
+echo VSCode AI App Setup Script
+echo ==========================
+
+echo.
+echo This script will help you set up the VSCode AI App.
+echo.
+
+echo Checking prerequisites...
+echo.
+
+REM Check if Node.js is installed
+node --version >nul 2>&1
+if %errorlevel% neq 0 (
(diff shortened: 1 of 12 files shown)


setup-ai-app.bat
```diff
@@ -0,0 +1,86 @@
+@echo off
```
salimuddin07, line 1: Script assumes it’s run from the correct directory (npm install etc.), but doesn’t check.
Script will stop if Visual Studio Build Tools are missing, but doesn’t attempt to auto-install or give direct next steps.

src/vs/workbench/contrib/aiApp/browser/browserComponent.ts
```diff
@@ -0,0 +1,184 @@
...
+
+	private createWebview(): HTMLElement {
+		const webview = append(this._container, $('.webview-container'));
+		webview.style.cssText = `
+			flex: 1;
+			display: flex;
+			flex-direction: column;
+			background: var(--vscode-editor-background);
+		`;
+
```
salimuddin07, line 85: The iframe may fail to load many websites due to CORS and X-Frame-Options restrictions. Your code doesn’t handle fallback gracefully.
The “Demo” button is referenced in docs but not implemented in code.

PR #271089, opened 2025-10-13
per-model rendering for thinking

justschen:
380-Latias


src/vs/workbench/contrib/chat/browser/chatContentParts/chatThinkingContentPart.ts
```diff
@@ -33,6 +34,34 @@ function extractTitleFromThinkingContent(content: string): string | undefined {
 	return headerMatch ? headerMatch[1].trim() : undefined;
 }
 
+function isResponseViewModel(item: ChatTreeItem): item is IChatResponseViewModel {
+	return typeof (item as IChatResponseViewModel).model !== 'undefined';
```
roblourens: See the common helper `isResponseVM`. But honestly I don't know why that isn't just an `instanceof`

src/vs/workbench/contrib/chat/browser/chatContentParts/chatThinkingContentPart.ts
```diff
@@ -33,6 +34,34 @@ function extractTitleFromThinkingContent(content: string): string | undefined {
...
+ * all others     => collapsed
+ */
+function resolvePerModelDefaultThinkingStyle(modelId: string | undefined): ThinkingDisplayMode {
+	if (!modelId) {
+		return ThinkingDisplayMode.Collapsed;
+	}
+	const id = modelId.toLowerCase();
+
+	// TODO @justschen: could have additional styles specific to gemini/codex besides collapased.
+	if (id.includes('gpt-5-codex') || id.includes('gemini')) {
```
roblourens: Instead of checking model IDs, I wonder would it make sense to do this based on the behavior/actual format of the response? So when you have any model returning thinking text in a certain format, then pick the right display style for that format.

The advantage would be that you don't have to hardcode a bunch of model IDs and maintain that list as new models come out.

And maybe sometimes some models act one way, sometimes they act another way, this would react to what they are doing for the prompt.

src/vs/workbench/contrib/chat/browser/chatContentParts/chatThinkingContentPart.ts
justschen: hmm copilot added it's own function LOL 👍🏻

src/vs/workbench/contrib/chat/browser/chatContentParts/chatThinkingContentPart.ts
justschen: when looking at the response, everything looks about the same tho i think.

ie, the different between gpt-5-codex and gpt-5 is that in gpt-5-codex, sometimes we only get a header summary, but sometimes we will get text too, but gpt-5 will give both.

with gemini, i'm only doing this because they aren't sending `done` tokens like we expected them too (google doesn't do this anyways), so the fixed scrolling type looks broken, while the lighter-weight collapse works fine

i feel like the differences in what we see in _just_ thinking tokens is pretty hard to differentiate ATP?

roblourens: > sometimes we only get a header summary, but sometimes we will get text too, but gpt-5 will give both.

Then do you just want to be distinguishing between having headers and not?

justschen: as of late, everything seems to have a header (formatted like `**header**\n\n` or `**header**`) 🥲

bpasero (dismissed):

src/vs/workbench/contrib/chat/browser/chatContentParts/chatThinkingContentPart.ts
roblourens: Sorry I meant having text after the header or not. Like is that the model-specific difference that you're trying to respond to?

justschen: that is one of the model specific things. the other thing is the issue where we aren't getting `done` from gemini.

cc @karthiknadig not sure when we think this might get fixed from CAPI and/or if we should just fix this ourselves from the extension side by adding in `done` when it goes from thinking -> text

karthiknadig: Adding the `done` ourselves when content goes from thinking -> text is an ok point in time solution.

bpasero:

src/vs/workbench/contrib/chat/browser/media/chat.css
```diff
@@ -30,6 +30,10 @@
 	display: none !important;
 }
 
+.interactive-list > .monaco-list > .monaco-scrollable-element > .monaco-list-rows > .monaco-list-row > .monaco-tl-row > .monaco-tl-indent {
+	display: none !important;
```
roblourens: Why do you need this now?

src/vs/workbench/contrib/chat/browser/chatListRenderer.ts
```diff
@@ -1183,27 +1206,40 @@ export class ChatListItemRenderer extends Disposable implements ITreeRenderer<Ch
...
-	// 	}
-
-	// 	return false;
-	// }
+	private shouldPinPart(part: IChatRendererContent, element?: IChatResponseViewModel): boolean {
+
+		if ((part.kind === 'toolInvocation' || part.kind === 'toolInvocationSerialized') && element) {
+			// Explicit set of tools that should be pinned when there has been thinking
+			const specialToolIds = new Set<string>([
+				'copilot_searchCodebase',
```
roblourens, line 1214: If we keep this kind of approach, we would want another way to identify the tools to group. A tool tag might be one. Think about MCP/extension tools too.

src/vs/workbench/contrib/chat/browser/chatListRenderer.ts
```diff
@@ -1018,6 +1028,19 @@ export class ChatListItemRenderer extends Disposable implements ITreeRenderer<Ch
 				contentIndex: contentIndex,
 				container: templateData.rowContainer,
 			};
+
+			// combine tool invocations into thinking part if needed. render the tool, but do not replace the working spinner with the new part's dom node since it is already inside the thinking part.
+			if (this._currentThinkingPart && (partToRender.kind === 'toolInvocation' || partToRender.kind === 'toolInvocationSerialized') && this.shouldCombineThinking(element) && this.shouldPinPart(partToRender, element)) {
+				const newPart = this.renderChatContentPart(partToRender, templateData, context);
```
roblourens, line 1034: I don't understand this, isn't it the same line `const newPart = this.renderChatContentPart(partToRender, templateData, context);` as below? How does it get rendered differently?

roblourens: I guess it's not removing/replacing the `alreadyRenderedPart.domNode` that I don't understand

src/vs/workbench/contrib/chat/browser/chatListRenderer.ts
sbatten: not for this PR, but I'm leaning more and more that if the tool doesn't require me to interact, pin it

sbatten (approved):

PR #271145, opened 2025-10-13
Switch to standard `findLastIndex`

mjbvz:
Doing this before also removing our impl of `findLast`


mjbvz:
Will follow up in December with a PR bumping the Monaco target too

PR #271219, opened 2025-10-13
Fix output for microsoft-authentication

mjbvz:
For #271167

All out of extensions normally target `out`. `dist` is for webpack


TylerLeonhardt (approved):
Looks fine, but I know the MSAL typings are weird... so hopefully this is fine with those

Issue #250648, opened 2025-06-04
Migration path for extensions relying on navigator for web environment detection


deepak1556:
Refs https://github.com/microsoft/vscode-docs/pull/8446/files

- [ ] Enable navigator object in global scope of desktop and remote extension host behind setting
- [ ] :runner: Capture telemetry for extensions usage and surface visible error https://github.com/microsoft/vscode/pull/250619
- [ ] :runner: Document changes in release notes https://github.com/microsoft/vscode-docs/pull/8446

deepak1556:
**Steps for Verification**

* Add `console.log(navigator)` to https://github.com/microsoft/vscode-extension-samples/blob/cf30922221d1c2dcb819da857b7f64583991fed0/helloworld-minimal-sample/extension.js#L15
* Debug and Run the extension
* Confirm that the output of the log is `undefined`
* Verify that extension host log contains `PendingMigrationError is deprecated. navigator is now a global in nodejs, please see https://aka.ms/vscode-extensions/navigator for additional info on this error.` followed by error stack that makes sense
* Change the setting `extensions.supportNodeGlobalNavigator` to `true` and reload the window
* Debug and Run the extension
* Confirm that the output of the log contains a valid navigator object and there are no error messages from this extension

PR #271252, opened 2025-10-14
chore: avoid reporting PendingMigrationError to error telemetry

deepak1556:
For https://github.com/microsoft/vscode/issues/250648#issuecomment-3394040431


jrieken (approved):
Thanks. This looks good and better than extending from `ErrorNoTelemetry`. In the long run we should think about some well-known property/symbol that marks an error as not reported. This would allow for easy composition. @lramos15 what do think?

PR #271322, opened 2025-10-14
chat - properly apply height to last element in list

bpasero:
We update the `min-height` of the last response during `layout` call, but we fail to tell the list widget that the height changed.

With the fix, the last chat response will stay in view when resizing the window:

Recording 2025-10-14 at 16 19 06

//cc @benibenj as this concerns tree/list usage

### DIFF
--- a/src/vs/workbench/contrib/chat/browser/chatWidget.ts
+++ b/src/vs/workbench/contrib/chat/browser/chatWidget.ts
@@ -2600,6 +2600,7 @@ export class ChatWidget extends Disposable implements IChatWidget {
 
 	layout(height: number, width: number): void {
 		width = Math.min(width, 950);
+		const heightUpdated = this.bodyDimension && this.bodyDimension.height !== height;
 		this.bodyDimension = new dom.Dimension(width, height);
 
 		const layoutHeight = this._dynamicMessageLayoutData?.enabled ? this._dynamicMessageLayoutData.maxHeight : height;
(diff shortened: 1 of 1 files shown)


src/vs/workbench/contrib/chat/browser/chatWidget.ts
```diff
@@ -2616,12 +2617,16 @@ export class ChatWidget extends Disposable implements IChatWidget {
...
 		const lastElementVisible = this.tree.scrollTop + this.tree.renderHeight >= this.tree.scrollHeight - 2;
+		const lastItem = this.viewModel?.getItems().at(-1);
 
 		const contentHeight = Math.max(0, height - inputHeight - chatTodoListWidgetHeight);
 		if (this.viewOptions.renderStyle === 'compact' || this.viewOptions.renderStyle === 'minimal') {
 			this.listContainer.style.removeProperty('--chat-current-response-min-height');
 		} else {
 			this.listContainer.style.setProperty('--chat-current-response-min-height', contentHeight * .75 + 'px');
+			if (heightUpdated && lastItem) {
+				this.tree.updateElementHeight(lastItem, undefined);
```
bpasero, line 2628: @benibenj my understanding is that passing in `undefined` leaves the height computation to the list widget.

I am trying to be defensive here and only do this when the overall height changed, which typically should only happen if the window size changes.

roblourens (approved):
This looks right to me, thanks!

Issue #240226, opened 2025-02-10
Terminal suggest: Allow disabling fallback suggestions from quick suggest


Tyriar:
Feedback from @ulugbekna:

> I definitely want a terminal suggest but only when it's useful, eg I'm typing cd ./ or cd subdir/ and want to find the file/directory I'm looking for. If I want a suggest explicitly

Tyriar:
With https://github.com/microsoft/vscode/pull/240889 this is the default behavior for now. The fallback suggestions are available via an explicit ctrl+space though.

ulugbekna:
I think since the change of default I'm quite enjoying the terminal suggest now :-) 👍🏽

Tyriar:
@ulugbekna this one wasn't merged just yet, I don't think it'll make it into Insiders until tomorrow. You're likely enjoying the expanded fig specs which has a similar effect but only for the CLIs we include now which includes vim, brew, etc. Let us know if you use a CLI often but it's not at https://github.com/microsoft/vscode/tree/main/extensions/terminal-suggest/src/completions/upstream

PR #240889, opened 2025-02-15
Expand terminal suggest quickSuggestions setting

Tyriar:
This moves from a simple boolean to an object that allows customizing whether completions are provided for commands and arguments, and whether fallback suggestions are allowed.

The default is on for commands and arguments and off for unknown (fallback).

Fixes #240226 (`unknown` defaults to `"off"`)
Fixes #239978 (can disable `commands` by setting to "off"`)

Default values:

```json
  "terminal.integrated.suggest.quickSuggestions": {
    "commands": "on",
    "arguments": "on",
    "unknown": "off"
  },
```

Issue #207504, opened 2024-03-12
Explore terminal buffer API


Tyriar:
We may want to implement access to the terminal buffer before finalizing https://github.com/microsoft/vscode/issues/145234. Copilot also has use for such an API.

Tyriar:
Rough draft of what this could look like:

```ts
declare module 'vscode' {
	export interface Terminal {
		readonly buffers: TerminalBuffers;
	}

	export interface TerminalBuffers {
		readonly active: TerminalBuffer;
		readonly normal: TerminalBuffer;
		readonly alternate: TerminalBuffer;
	}

	export interface TerminalBuffer {
		readonly type: TerminalBufferType;

		/**
		 * The number of lines that have been trimmed from the scrollback.
		 */
		readonly trimmedLineCount: number;

		/**
		 * The length of the buffer. This does not include trimmed lines.
		 */
		readonly length: number;

		// TODO: Can we omit cursor x, y, base y, viewport y?

		// NOTE: Throws when line isn't valid or is trimmed
		lineAt(line: number): TerminalBufferLine;
	}

	export enum TerminalBufferType {
		Normal,
		Alternate
	}

	export interface TerminalBufferRange {
		// TODO: Could we just share Position here?
		start: Position;
		end: Position;
	}

	/**
	 * TerminalBufferLine objects are __immutable__. When a {@link TerminalBuffer buffer}'s content
	 * changes, previously retrieved lines will not represent the latest state.
	 */
	export interface TerminalBufferLine {
		/**
		 * The zero-based line number.
		 */
		readonly lineNumber: number;

		/**
		 * The text of this line without the line separator characters.
		 */
		readonly text: string;

		/**
		 * The range this line covers. This includes "whitespace" at the end of
		 * the line if the terminal cells were written to.
		 */
		readonly range: TerminalBufferRange;

		/**
		 * The offset of the first character which is not a whitespace character as defined
		 * by `/\s/`. **Note** that if a line is all whitespace the length of the line is returned.
		 */
		readonly firstNonWhitespaceCharacterIndex: number;

		/**
		 * Whether this line is whitespace only, shorthand
		 * for {@link TerminalBufferLine.firstNonWhitespaceCharacterIndex} === {@link TerminalBufferLine.text TerminalBufferLine.text.length}.
		 */
		readonly isEmptyOrWhitespace: boolean;
	}

	export namespace window {
		export const onDidChangeTerminalBuffer: Event<Terminal>;
	}

[...]

PR #271338, opened 2025-10-14
finalize terminal completion provider API

### LINKED ISSUE #224505: terminal completion provider extension API

meganrogge:
fixes https://github.com/microsoft/vscode/issues/224505


src/vscode-dts/vscode.d.ts
```diff
@@ -8377,10 +8377,9 @@ declare module 'vscode' {
 		 */
 		readonly cursorIndex: number;
 		/**
-		 * Whether completions should be provided when none are explicitly suggested. This will display
-		 * fallback suggestions like files and folders.
+		 * Whether completions were explicitly invoked.
 		 */
-		readonly allowFallbackCompletions: boolean;
+		readonly wasExplicitlyInvoked: boolean;
```
mjbvz: Should we use `CompletionTriggerKind`, or a new type like it instead?

src/vscode-dts/vscode.d.ts
```diff
@@ -8192,6 +8192,252 @@ declare module 'vscode' {
...
+		kind?: TerminalCompletionItemKind;
+	}
+
+
+	/**
+	 * The kind of an individual terminal completion item.
+	 *
+	 * The kind is used to render an appropriate icon in the suggest list and to convey the semantic
+	 * meaning of the suggestion (file, folder, flag, commit, branch, etc.).
+	 *
```
mjbvz:
```suggestion
```

src/vscode-dts/vscode.d.ts
```diff
@@ -8192,6 +8192,252 @@ declare module 'vscode' {
...
+		 * Example: `#41 Fix bug (closed)`
+		 */
+		PullRequestDone = 16,
+	}
+
+
+	/**
+	 * Context information passed to {@link TerminalCompletionProvider.provideTerminalCompletions}.
+	 *
+	 * It contains the full command line, the current cursor position, and a flag indicating whether
```
mjbvz, line 8375: I think the part about the flag is out of date now

src/vscode-dts/vscode.d.ts
meganrogge: Unlike `CompletionTriggerKind`, we only have two kinds, explicitly invoked or triggered. Thus, the boolean seems fine here to me. WDYT?

mjbvz: I think we should probably introduce a new `TerminalCompletionTriggerKind` enum to align this with the completions API and to support expansion in the future if needed

src/vscode-dts/vscode.d.ts
```diff
@@ -8192,6 +8192,251 @@ declare module 'vscode' {
...
+	 */
+	export interface TerminalCompletionResourceOptions {
+		/**
+		 * Show files as completion items.
+		 */
+		showFiles?: boolean;
+		/**
+		 * Show folders as completion items.
+		 */
+		showDirectories?: boolean;
```
alexr00: @mjbvz should `showFiles` and `showDirectories` be required instead of optional? The extension author has to set at least one for the options to be meaninful.

src/vscode-dts/vscode.d.ts
jrieken: > we only have two kinds, explicitly invoked or triggered. 

Is there a strong reason for having this? We explicitly have not exposed this in text completions because we don't want extensions to differentiate their results based on implicit or explicit triggering. The worry is that it can confuse user why sometimes completions show and sometimes not

src/vscode-dts/vscode.d.ts
```diff
@@ -8192,6 +8192,251 @@ declare module 'vscode' {
...
+	 * 	replacementIndex: 3,
+	 * 	replacementLength: 1,
+	 * 	detail: 'List all entries except for . and .. (always set for the super-user)',
+	 * 	kind: TerminalCompletionItemKind.Flag
+	 * };
+	 *
+	 * The fields on a completion item describe what text should be shown to the user
+	 * and which portion of the command line should be replaced when the item is accepted.
+	 */
+	export interface TerminalCompletionItem {
```
jrieken: Must be a class with a ctor like it all other providers. The ctor should require all non optional fields

src/vscode-dts/vscode.d.ts
```diff
@@ -8192,6 +8192,251 @@ declare module 'vscode' {
...
+
+		/**
+		 * The index of the start of the range to replace.
+		 */
+		replacementIndex: number;
+
+		/**
+		 * The length of the range to replace.
+		 */
+		replacementLength: number;
```
jrieken: Should these `replacement` properties by one type or tuple? Similar to `InputBoxOptions#valueSelection` or some TerminalRange?

jrieken (dismissed):

src/vscode-dts/vscode.d.ts
Tyriar: `valueSelection` looks like exactly what we should be copying 👍. The reason it's replacement index/length is because an early prototype was based on how pwsh's completions worked which used these properties.

---

RE `TerminalRange`: even if we had the terminal buffer API we've talked about in the past (https://github.com/microsoft/vscode/issues/207504), we would probably still want an API similar to this because the prompt is not necessarily a contiguous range in the terminal buffer.

Simple example:

image

Should be `echo "a\nb`, but a single range would include the `>> ` line continuation. It gets more complicated with right prompts.

What we're actually exposing in `TerminalCompletionContext` is the smarts of an internal `PromptInputModel` class that does a lot of work to figure out the prompt given all the other stuff that could show up, and reprinting content, etc.

We could however do a `TerminalRange[]`, but it feels like it's overcomplicating things. This is on top of the buffer API being more complicated too if it ever happens. Here's an example of what getting the actual string would look like if we had the proposal in https://github.com/microsoft/vscode/issues/207504:

```ts
const input = ranges.map(e => {
  // Hope it's the right activeBufferType since we tried to hide that in the most recent proposal
  return terminal.buffer.getText(e)
}).join('\n')
// Map cursor Index to that string somehow
```

src/vscode-dts/vscode.d.ts
Tyriar: Here's the PR it was added in https://github.com/microsoft/vscode/pull/240889

@meganrogge after our chat yesterday, seeing @jrieken's feedback and seeing that PR, I've leaning towards align with the editor and always have a single set of completions instead of differing explicit triggering. We can still keep `unknown: on` when when there's no argument found by fig which I believe will keep the main problem some people on the team were having fixed https://github.com/microsoft/vscode/issues/240226#issuecomment-2663193011

meganrogge: I have removed it from the API

extensions/terminal-suggest/src/terminalSuggestMain.ts
```diff
@@ -493,13 +493,9 @@ export async function getCompletionItemsFromSpecs(
...
 		showFiles = true;
 		showFolders = true;
-	}
-	// For arguments when no fig suggestions are found these are fallback suggestions
-	else if (!items.length && !showFiles && !showFolders && !hasCurrentArg) {
-		if (terminalContext.allowFallbackCompletions) {
-			showFiles = true;
-			showFolders = true;
-		}
+	} else if (!items.length && !showFiles && !showFolders && !hasCurrentArg) {
```
meganrogge, line 496: @Tyriar I think we still want this, correct?

src/vscode-dts/vscode.d.ts
```diff
@@ -8192,6 +8192,258 @@ declare module 'vscode' {
...
+
+		/**
+		 * Selection range in the command line to replace when the completion is accepted. Defined
+		 * as a tuple where the first entry is the inclusive start index and the second entry is the
+		 * exclusive end index. When `undefined` the completion will be inserted at the cursor
+		 * position. When the two numbers are equal only the cursor position changes (insertion).
+		 *
+		 * This mirrors {@link QuickPick.valueSelection} semantics for consistency.
+		 */
+		valueSelection: readonly [number, number] | undefined;
```
mjbvz: Not a fan of the name `valueSelection`. I think we should align more with the completion API instead  and use the name `replacement`, maybe `replacementRange`?

src/vscode-dts/vscode.d.ts
```diff
@@ -8192,6 +8192,258 @@ declare module 'vscode' {
...
+		/**
+		 * Creates a new terminal completion item.
+		 *
+		 * @param label The label of the completion.
+		 * @param valueSelection The inclusive start and exclusive end index of the text to replace.
+		 * @param kind The completion's kind.
+		 */
+		constructor(
+			label: string | CompletionItemLabel,
+			valueSelection: readonly [number, number] | undefined,
```
mjbvz: Should this be optional?

src/vscode-dts/vscode.d.ts
mjbvz: Yes we can encourage more correct usage that way. Let's make these just `boolean` instead

extensions/terminal-suggest/src/terminalSuggestMain.ts
Tyriar: As in the `else if` case? It should depend upon `unknown = 'on'` now I think since that's no longer coming through as `wasExplicitlyInvoked`

src/vs/workbench/contrib/terminal/browser/terminalInstance.ts
```diff
@@ -2651,7 +2651,7 @@ export class TerminalLabelComputer extends Disposable {
 			case 0: return '';
 			case 1: return `${Math.round(progressState.value)}%`;
 			case 2: return '$(error)';
-			case 3: return '$(loading~spin)';
```
Tyriar, line 2654: What's this change for? We don't want the sync icon for terminal command in progress as it's not syncing?

Tyriar (changes requested):

src/vs/workbench/contrib/terminal/browser/terminalInstance.ts
meganrogge: I had changed this on a different branch, accidentally pushed it here, so reverted that. You'll find there's no change in the overall diff to that file.

I closed the PR associated with the other branch because it did not do what i want.

extensions/terminal-suggest/src/terminalSuggestMain.ts
meganrogge: If we remove this, in `terminalCompletionsService`, then we won't have `cwd` or `pathSeparator`, as those come only from the provider and the extension setting `resourceRequestOptions`. I investigated getting those a different way and it's very involved/ doesn't feel like the right approach.

Since our suggest extension is the thing we want to trigger this anyway, especially when knowing `!hasCurrentArg`, I believe we should leave this.

Tyriar: Oh actually we do want to keep this as is. The quickSuggestions/unknown value is purely about whether it should show when you type or not. So we just need to make sure that still works, then all suggestions should be the same regardless of whether they were explicitly triggered or via quick suggest.

src/vs/workbench/api/browser/mainThreadTerminalService.ts
```diff
@@ -277,8 +277,8 @@ export class MainThreadTerminalService implements MainThreadTerminalServiceShape
 	public $registerCompletionProvider(id: string, extensionIdentifier: string, ...triggerCharacters: string[]): void {
 		this._completionProviders.set(id, this._terminalCompletionService.registerTerminalCompletionProvider(extensionIdentifier, id, {
 			id,
-			provideCompletions: async (commandLine, cursorPosition, allowFallbackCompletions, token) => {
-				const completions = await this._proxy.$provideTerminalCompletions(id, { commandLine, cursorIndex: cursorPosition, allowFallbackCompletions }, token);
+			provideCompletions: async (commandLine, cursorPosition, token) => {
+				const completions = await this._proxy.$provideTerminalCompletions(id, { commandLine, cursorIndex: cursorPosition }, token);
```
Tyriar:
```suggestion
			provideCompletions: async (commandLine, cursorIndex, token) => {
				const completions = await this._proxy.$provideTerminalCompletions(id, { commandLine, cursorIndex }, token);
```

src/vscode-dts/vscode.d.ts
```diff
@@ -8192,6 +8192,257 @@ declare module 'vscode' {
...
+		PullRequest = 15,
+		/**
+		 * A closed pull request completion item.
+		 * Example: `#41 Fix bug (closed)`
+		 */
+		PullRequestDone = 16,
+	}
+
+
+	/**
```
Tyriar, line 8372:
```suggestion
		PullRequestDone = 16,
	}

	/**
```

src/vscode-dts/vscode.d.ts
```diff
@@ -8192,6 +8192,257 @@ declare module 'vscode' {
...
+		/**
+		 * Creates a new terminal completion item.
+		 *
+		 * @param label The label of the completion.
+		 * @param valueSelection The inclusive start and exclusive end index of the text to replace.
+		 * @param kind The completion's kind.
+		 */
+		constructor(
+			label: string | CompletionItemLabel,
+			valueSelection: readonly [number, number],
```
Tyriar: All `valueSelection` in the PR should be `replacementRange` I think?

src/vs/workbench/services/suggest/browser/simpleCompletionModel.ts
```diff
@@ -113,7 +113,7 @@ export class SimpleCompletionModel<T extends SimpleCompletionItem> {
 			// filter and score against. In theory each suggestion uses a
 			// different word, but in practice not - that's why we cache
 			// TODO: Fix
-			const overwriteBefore = item.completion.replacementLength; // item.position.column - item.editStart.column;
+			const overwriteBefore = item.completion.replacementRange ? (item.completion.replacementRange[1] - item.completion.replacementRange[0]) : 0; // derived from valueSelection
 			const wordLen = overwriteBefore + characterCountDelta; // - (item.position.column - this._column);
```
Tyriar: Do these comments add value or confusion?

Tyriar (dismissed):

src/vs/workbench/services/suggest/browser/simpleCompletionModel.ts
meganrogge: rm comments

Tyriar (approved):
Let's make sure we do a full TPI for this again

mjbvz (approved):
API looks good. Thanks for making all these changes!

Issue #268856, opened 2025-09-29
Chat: Model picker value should be cached


joaomoreno:
The model picker currently doesn't show up a few good seconds after startup. We should really cache these models as best we can. Then:

- If the model list didn't change since the last exec, all good;
- Otherwise:
  - If the currently selected model is still there, all good;
  - Otherwise, just switch the user to Auto (or next best thing).

https://github.com/user-attachments/assets/8d69ec5e-6d12-41e3-a3ff-ffac9d33f3d3

PR #271376, opened 2025-10-14
First round of caching of the model picker

lramos15:
Part of https://github.com/microsoft/vscode/issues/268856

### DIFF
--- a/src/vs/workbench/contrib/chat/browser/chatInputPart.ts
+++ b/src/vs/workbench/contrib/chat/browser/chatInputPart.ts
@@ -510,8 +510,8 @@ export class ChatInputPart extends Disposable implements IHistoryNavigationWidge
 
 		this._register(this.languageModelsService.onDidChangeLanguageModels(() => {
 			// We've changed models and the current one is no longer available. Select a new one
-			const selectedModel = this._currentLanguageModel ? this.languageModelsService.lookupLanguageModel(this._currentLanguageModel.identifier) : undefined;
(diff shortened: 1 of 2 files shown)


dmitrivMS (approved):

src/vs/workbench/contrib/chat/browser/chatInputPart.ts
```diff
@@ -698,22 +693,20 @@ export class ChatInputPart extends Disposable implements IHistoryNavigationWidge
 	}
 
 	private getModels(): ILanguageModelChatMetadataAndIdentifier[] {
-		const models = this.languageModelsService.getLanguageModelIds()
-			.map(modelId => ({ identifier: modelId, metadata: this.languageModelsService.lookupLanguageModel(modelId)! }))
-			.filter(entry => entry.metadata?.isUserSelectable && this.modelSupportedForDefaultAgent(entry));
+		const cachedModels = this.storageService.getObject<ILanguageModelChatMetadataAndIdentifier[]>('chat.cachedLanguageModels', StorageScope.APPLICATION, []);
```
roblourens, line 696: Why cache these in the view, can the ILanguageModelsService handle the caching?

lramos15: Caching in the view state makes more sense as a lot of things rely on the language model service as the actual data source for when things are ready. Such as the chat setup / proxy. Effectively what caching here allows us to do is have up to date UI state while having all the proper event handlers + gets on the language model service be reserved for "real" models

PR #271379, opened 2025-10-14
feat: add PascalCase helper with edge-case tests

youssefnabil2030:
- Handles multiple underscores, leading/trailing underscores
- Handles acronyms (HTML_API -> HtmlApi)
- Includes unit tests

### DIFF
--- a/src/vs/editor/contrib/customTransform/pascalCase.test.ts
+++ b/src/vs/editor/contrib/customTransform/pascalCase.test.ts
@@ -1,0 +1,19 @@
+import { toPascalCase } from './pascalCase';
+import * as assert from 'assert';
+
+// Basic test
+assert.strictEqual(toPascalCase('FOO_BAR'), 'FooBar');
+
+// Double underscores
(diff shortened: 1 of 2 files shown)


aeschli:
How is VS Code using this?

youssefnabil2030:
> How is VS Code using this?

Hi @aeschli 👋

Thanks for the review!

The purpose of this helper is to normalize identifiers and keys that need to follow PascalCase naming across the codebase — for example, when generating command IDs, variable names, or labels used by extensions.

Technically, the helper takes any string (including those with underscores or acronyms) and converts it into a clean PascalCase form:

"__user_name__" → "UserName"

"HTML_API" → "HtmlApi"

It’s mainly useful for code transformation utilities where string naming consistency is important.

I can adjust or relocate the helper if there’s a more suitable place for it within the VS Code source. 😊

aeschli:
We only add new utilities when we have multiple usages for it. Until then we place you can code along with the component that requires it.

aeschli:
We don't want to add code that is not used. Or di I miss something?

PR #271397, opened 2025-10-14
Support advanced settings.

### LINKED ISSUE #270142: Settings: Support advanced setting registration
We've had some cases in which extensions like Copilot Chat has the need for having advanced settings that aren't exposed to the user in the Settings UI or JSON editor. But over time we had to compensate for this and augment the JSON schema of the JSON editor to somewhat support this scenario.

Let's simply come up with an **advanced settings** concept. An advanced setting should be still registered with an `advanced: true` property. But:

- By default, users should not be able to discover advanced settings via the Settings UI or JSON editor.
- Unless they toggle `View advanced settings` (this can either be a setting or a global state value).
- (Optional) Insiders have `View advanced settings` set to `true`, by default.

sandy081:
Fixes #270142


src/vs/workbench/api/common/configurationExtensionPoint.ts
```diff
@@ -119,9 +119,24 @@ const configurationEntrySchema: IJSONSchema = {
...
+										'experimental',
+										'telemetry',
+										'usesOnlineServices',
+									],
+									enumDescriptions: [
+										nls.localize('accessibility', 'Accessibility settings'),
+										nls.localize('advanced', 'Advanced settings are hidden by default in the Settings editor unless the user chooses to show advanced settings.'),
+										nls.localize('experimental', 'Experimental settings'),
+										nls.localize('telemetry', 'Telemetry settings'),
+										nls.localize('usesOnlineServices', 'Settings that use online services')
```
rzhao271, line 136: We have some more tags accessible through the filter icon on the right of the Settings editor input. Would it make sense to add `preview` considering there's also `experimental`?

Settings editor tags

src/vs/workbench/contrib/preferences/browser/settingsEditor2.ts
```diff
@@ -1727,6 +1737,10 @@ export class SettingsEditor2 extends EditorPane {
 			this.viewState.languageFilter = parsedQuery.languageFilter;
 		}
 
+		if (showAdvanced !== this.viewState.tagFilters?.has(ADVANCED_SETTING_TAG)) {
```
rzhao271, line 1740: I'm wondering why this tag requires waiting for a refresh?

rzhao271 (changes requested):
The feature request seems to be for a new "show advanced settings" toggle within the UI and JSON Settings editors. Would it make more sense for the toggle to be a setting? Tags, such as the preview tag, currently act more as filters rather than as on/off toggles.

sandy081:
> The feature request seems to be for a new "show advanced settings" toggle within the UI and JSON Settings editors. Would it make more sense for the toggle to be a setting? Tags, such as the preview tag, currently act more as filters rather than as on/off toggles.

Thats right. After discussing locally, we decided to have this similar to tag filtering except that advanced settings are not visible by default. IMO when user want to see advanced settings, showing them only would be helpful. We can later introduce the setting/state to show the advanced settings by default or not

src/vs/workbench/api/common/configurationExtensionPoint.ts
sandy081: Done

src/vs/workbench/contrib/preferences/browser/settingsEditor2.ts
sandy081: To update the TOC and Settings Entries:

https://github.com/microsoft/vscode/blob/2eaa41bb253b0ecbf4a31a1f59d1904b96d50c9c/src/vs/workbench/contrib/preferences/browser/settingsEditor2.ts#L1409

rzhao271: Don't the other tags update those automatically? I'm not sure why the advanced tag requires additional processing compared to other tags such as experimental.

src/vs/workbench/contrib/preferences/browser/settingsEditor2.ts
```diff
@@ -1397,9 +1398,17 @@ export class SettingsEditor2 extends EditorPane {
...
 		const groups = this.defaultSettingsEditorModel.settingsGroups.slice(1); // Without commonlyUsed
+		const coreSettingsGroups = [], extensionSettingsGroups = [];
+		for (const group of groups) {
+			if (group.extensionInfo) {
+				extensionSettingsGroups.push(group);
+			} else {
+				coreSettingsGroups.push(group);
+			}
+		}
+		const filter = this.viewState.tagFilters?.has(ADVANCED_SETTING_TAG) ? undefined : { exclude: { tags: [ADVANCED_SETTING_TAG] } };
```
rzhao271, line 1409: Note to self: confirmed that this line explains why there's exclude logic. If no advanced tag then exclude those settings.

rzhao271 (approved):

PR #271400, opened 2025-10-14
hide copilot terminals by default in insider's, make configurable

### LINKED ISSUE #271384: Hide chat terminal sessions by default
All chat session terminals should be `hideFromUser: true` and have a popout button to show the terminal in panel

- New setting, insiders on by default to get early self hosting
    ```jsonc
    // Begin with in MVP
    "chat.tools.terminal.outputLocation": 'terminal' | 'none',
    // After we get final output showing this could be expanded
    "chat.tools.terminal.outputLocation": 'chat' | 'terminal' | 'both' | 'none',
    ```
- Pop out button to show the terminal
    - Pop out button's initial location can be on very right of tool call
    - Once popped out, should behave as it does today

meganrogge:
fixes #271384

**Tricky parts:**  
- **Persistence:** `hideFromUser` terminals aren’t restored on window reload, but our tool terminals must persist. To handle this, we reveal them and update their state in `onBeforeShutdown`.  
- **Debounced state saving:** The terminal service’s `_saveState` method is debounced, which delays persistence. We’ve added a `_saveStateNow` method that’s called when an argument is passed to `showBackgroundTerminal`, ensuring immediate state saving.  
- **Minimizing noise:** To avoid unnecessary UI noise on window reload, we only reveal and persist terminals that have active child processes or a shell integration state indicating an executing command.  
https://github.com/microsoft/vscode/blob/f9c9dbe971d419e4d229bdf8ea3d25454c3edf87/src/vs/workbench/contrib/terminalContrib/chat/browser/terminalChatService.ts#L52


src/vs/workbench/contrib/terminalContrib/chatAgentTools/common/terminalChatAgentToolsConfiguration.ts
```diff
@@ -386,6 +388,17 @@ export const terminalChatAgentToolsConfiguration: IStringDictionary<IConfigurati
 		tags: ['experimental'],
 		markdownDescription: localize('autoReplyToPrompts.key', "Whether to automatically respond to prompts in the terminal such as `Confirm? y/n`. This is an experimental feature and may not work in all scenarios."),
 	}
+	,
+	[TerminalChatAgentToolsSettingId.OutputLocation]: {
+		markdownDescription: localize('outputLocation.description', "Where to show the run in terminal tool session when running a foreground command."),
```
meganrogge:
```suggestion
		markdownDescription: localize('outputLocation.description', "Where to show the output from the run in terminal tool session."),
```

src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
```diff
@@ -63,6 +72,40 @@ export class ChatTerminalToolProgressPart extends BaseChatToolInvocationSubPart
...
 
+		// Only show focus action when output location is set to 'none'
+		if (configurationService.getValue<string>('chat.tools.terminal.outputLocation') === 'none') {
+			let terminalInstance = getTerminalInstanceByToolSessionId(terminalData.terminalToolSessionId);
+			if (elements.actionBar) {
+				const actionBar = new ActionBar(elements.actionBar, {});
+				this._register(actionBar);
+				const focusAction = new Action(
+					TerminalCommandId.FocusInstance,
+					localize('runInTerminalTool.showTerminalButtonLabel', "Focus Terminal"),
```
Tyriar: Can this conditionally say Show and Focus Terminal or Focus Terminal depending on whether it's hidden from the user or not?

src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
```diff
@@ -63,6 +72,40 @@ export class ChatTerminalToolProgressPart extends BaseChatToolInvocationSubPart
...
+					ThemeIcon.asClassName(Codicon.linkExternal),
```
Tyriar: Created https://github.com/microsoft/vscode-codicons/issues/350 for a new icon

src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
```diff
@@ -63,6 +72,40 @@ export class ChatTerminalToolProgressPart extends BaseChatToolInvocationSubPart
...
+				);
+				focusAction.tooltip = localize('runInTerminalTool.focusTerminal', "Focus Terminal");
+				actionBar.push([focusAction], { icon: true, label: false });
+
+				if (!terminalInstance && terminalData.terminalToolSessionId) {
+					const listener = onDidRegisterTerminalInstanceForToolSession(e => {
+						if (e.terminalToolSessionId === terminalData.terminalToolSessionId) {
+							terminalInstance = getTerminalInstanceByToolSessionId(e.terminalToolSessionId);
+							if (terminalInstance) {
+								focusAction.enabled = true;
```
Tyriar: Can we `this._store.delete(listener)` here to avoid holding onto that listener when we don't need it anymore?

Tyriar: Here we need to listen to when this instance is disposed and hide it in that case.

src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
```diff
@@ -63,6 +72,40 @@ export class ChatTerminalToolProgressPart extends BaseChatToolInvocationSubPart
...
+				const actionBar = new ActionBar(elements.actionBar, {});
+				this._register(actionBar);
+				const focusAction = new Action(
+					TerminalCommandId.FocusInstance,
+					localize('runInTerminalTool.showTerminalButtonLabel', "Focus Terminal"),
+					ThemeIcon.asClassName(Codicon.linkExternal),
+					!!terminalInstance,
+					async () => {
+						if (terminalInstance) {
+							await commandService.executeCommand(TerminalCommandId.FocusInstance, terminalInstance);
```
Tyriar: Instead of registering a command and giving it an implementation in terminalActions.ts, we should just implement it here instead. It doesn't make sense to call it any other way I think?

We could reuse `TerminalCommandId.Focus` unless we care about tracking command telemetry. In that case the command ID should be more specific to chat.

src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
```diff
@@ -63,6 +72,40 @@ export class ChatTerminalToolProgressPart extends BaseChatToolInvocationSubPart
...
+						}
+					}
+				);
+				focusAction.tooltip = localize('runInTerminalTool.focusTerminal', "Focus Terminal");
```
Tyriar: Don't think we need a tooltip since the content is identical to label?

src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
```diff
@@ -63,6 +72,40 @@ export class ChatTerminalToolProgressPart extends BaseChatToolInvocationSubPart
...
+				if (!terminalInstance && terminalData.terminalToolSessionId) {
+					const listener = onDidRegisterTerminalInstanceForToolSession(e => {
+						if (e.terminalToolSessionId === terminalData.terminalToolSessionId) {
+							terminalInstance = getTerminalInstanceByToolSessionId(e.terminalToolSessionId);
+							if (terminalInstance) {
+								focusAction.enabled = true;
+							}
+						}
+					});
+					this._register(listener);
```
Tyriar: Could move this to: `const listener = this._register(...)`

src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
Tyriar: Little nicer to use a class for this, for example:

https://github.com/microsoft/vscode/blob/0a70ffec1cb7d85eb0a4bcbd95cc34fe325512d2/src/vs/workbench/contrib/mcp/browser/mcpServerEditor.ts#L242-L250

src/vs/workbench/contrib/terminalContrib/browser/terminalChatAgentToolsExports.ts
Tyriar, line 1: We should never have a file in `terminalContrib/browser/`, the only exports from contribs must live within `terminalContribExports.ts`. Additionally this breaks some implicit rules around imports as contrib should never import from terminalContrib, only the other way around.

src/vs/workbench/contrib/terminalContrib/browser/terminalChatAgentToolsExports.ts
```diff
@@ -0,0 +1,28 @@
...
+ *--------------------------------------------------------------------------------------------*/
+
+import type { ITerminalInstance } from '../../terminal/browser/terminal.js';
+import { Emitter, Event } from '../../../../base/common/event.js';
+
+// Local registry used as a bridge between the run-in-terminal tool implementation
+// and chat UI components
+const terminalInstancesByToolSessionId = new Map<string, ITerminalInstance>();
+const onDidRegisterTerminalInstanceForToolSessionEmitter = new Emitter<{ terminalToolSessionId: string }>();
+export const onDidRegisterTerminalInstanceForToolSession: Event<{ terminalToolSessionId: string }> = onDidRegisterTerminalInstanceForToolSessionEmitter.event;
```
Tyriar: Ok so we have to solve that hacky code issue from my WIP PR for this. Let's solve it like this:

- Create an `ITerminalChatService` _interface_ in `vs\workbench\contrib\terminal\browser\terminal.ts` that maintains this relationship
- Implement the service in `vs\workbench\contrib\terminalContrib\chat\browser` and register it in `terminal.chat.contribution.ts`

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/runInTerminalTool.ts
```diff
@@ -396,16 +397,19 @@ export class RunInTerminalTool extends Disposable implements IToolImpl {
 
 		const timingStart = Date.now();
 		const termId = generateUuid();
+		// The prepared data should always include a terminalToolSessionId; assert and narrow the type.
+		const terminalToolSessionId = (toolSpecificData as IChatTerminalToolInvocationData).terminalToolSessionId;
+		if (!terminalToolSessionId) {
+			throw new Error('terminalToolSessionId missing from toolSpecificData');
+		}
```
Tyriar: Little concerned about this, but I'm not 100% sure if it will break something or not. Could we support running the tool without this?

src/vs/workbench/contrib/terminalContrib/browser/terminalChatAgentToolsExports.ts
```diff
@@ -0,0 +1,28 @@
...
+
+export function registerTerminalInstanceForToolSession(terminalToolSessionId: string, instance: ITerminalInstance): void {
+	terminalInstancesByToolSessionId.set(terminalToolSessionId, instance);
+	onDidRegisterTerminalInstanceForToolSessionEmitter.fire({ terminalToolSessionId });
+	instance.onDisposed(() => {
```
Tyriar: This needs to be registered somewhere or it may leak a terminal instance

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/runInTerminalTool.ts
```diff
@@ -584,6 +588,8 @@ export class RunInTerminalTool extends Disposable implements IToolImpl {
 				});
 			}
 
+			this._handleTerminalVisibility(toolTerminal);
```
Tyriar: Why was this moved? Doesn't that mean the terminal should only show when the command is finished, not when it's started? We want to keep the old behavior when the setting is `'terminal'`.

Tyriar (changes requested):

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/runInTerminalTool.ts
meganrogge: This was moved to occur after the shell integration element is added.

Tyriar: But it's now after `await strategy.execute` isn't it?

src/vs/workbench/contrib/terminal/browser/terminalService.ts
```diff
@@ -1172,6 +1172,8 @@ export class TerminalService extends Disposable implements ITerminalService {
 		}
 		this._backgroundedTerminalDisposables.delete(instance.instanceId);
 		this._terminalGroupService.createGroup(instance);
+		// This is no longer a background terminal
+		instance.shellLaunchConfig.hideFromUser = false;
```
Tyriar: It's a launch config only, not meant to be changed later

Tyriar (changes requested):

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/runInTerminalTool.ts
meganrogge: yeah now that we don't render anything, can move it back

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/runInTerminalTool.ts
meganrogge: it looks like that's only undefined when a legacy config is used. Can we remove that legacy handling instead?

meganrogge: i'll support it for now

src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
```diff
@@ -89,19 +89,26 @@ export class ChatTerminalToolProgressPart extends BaseChatToolInvocationSubPart
...
+					}));
+				}
+				const instanceListener = terminalInstance?.onDisposed(() => {
+					focusAction.enabled = false;
+					actionBar.dispose();
+					instanceListener?.dispose();
+				});
+				if (instanceListener) {
+					this._register(instanceListener);
 				}
```
Tyriar: We need to register the `onDisposed` listener if `terminalInstance` is set or if it's not set and we get it via `getTerminalInstanceByToolSessionId`.

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/runInTerminalTool.ts
Tyriar: We don't want to break things when opening chat with data restored from old versions. I don't think we need to worry about that inside `runInTerminalTool` but we do for the UI chat parts so we don't break rendering when updating.

src/vs/workbench/contrib/chat/browser/chatContentParts/media/chatTerminalToolProgressPart.css
```diff
@@ -30,6 +30,14 @@
 	}
 }
 
+.chat-terminal-content-part .chat-terminal-action-bar {
+	position: absolute;
```
Tyriar: Using absolute here means that it would potentially overlap the terminal command. Instead you should use flexbox to put the action bar on the right.

src/vs/workbench/contrib/terminal/browser/terminal.ts
```diff
@@ -100,6 +101,35 @@ export interface ITerminalInstanceService {
 	didRegisterBackend(backend: ITerminalBackend): void;
 }
 
+/**
+ * Service for bridging terminal chat related tool invocations with terminal instances.
+ * Maintains an in-memory association between a tool session id (generated when a chat tool
+ * prepares/invokes a terminal run) and the actual {@link ITerminalInstance}. Consumers (eg. chat
+ * UI parts) can look up the instance asynchronously or listen for registrations should the
+ * terminal not yet exist.
+ */
```
Tyriar, line 107: You could simplify this by saying it's mainly a communication mechanism to enable communication via the chat tool implementation in terminalContrib and workbench contribs.

> Maintains an in-memory association between a tool session id

This will probably quickly get out of date as we add functionality.

src/vs/workbench/contrib/terminal/browser/terminal.ts
```diff
@@ -100,6 +101,35 @@ export interface ITerminalInstanceService {
...
+export interface ITerminalChatService {
+	readonly _serviceBrand: undefined;
+
+	/**
+	 * Fired when a terminal instance is registered for a tool session id. This can happen after
+	 * the chat UI first renders, enabling late binding of the focus action.
+	 */
+	readonly onDidRegisterTerminalInstanceForToolSession: Event<{ terminalToolSessionId: string }>;
```
Tyriar: Adding instance to the event would simplify the usage of this. Also make it a `IRegisterTerminalInstanceForToolSessionEvent`?

src/vs/workbench/contrib/terminal/browser/terminal.ts
```diff
@@ -100,6 +101,35 @@ export interface ITerminalInstanceService {
...
+	 * Associate a tool session id with a terminal instance. The association is automatically
+	 * cleared when the instance is disposed.
+	 */
+	registerTerminalInstanceForToolSession(terminalToolSessionId: string | undefined, instance: ITerminalInstance): void;
+
+	/**
+	 * Resolve a terminal instance by its tool session id.
+	 * @param terminalToolSessionId The tool session id provided in toolSpecificData.
+	 */
+	getTerminalInstanceByToolSessionId(terminalToolSessionId: string | undefined): ITerminalInstance | undefined;
```
Tyriar: Explaining what an undefined session id does would be good, how it's just there for convenience.

src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
Tyriar: Don't think this one was resolved?

src/vs/workbench/contrib/terminalContrib/chat/browser/terminalChatService.ts
```diff
@@ -0,0 +1,43 @@
...
+
+/**
+ * Used to manage chat tool invocations and the underlying terminal instances they create/use.
+ */
+export class TerminalChatService extends Disposable implements ITerminalChatService {
+	declare _serviceBrand: undefined;
+
+	private readonly _terminalInstancesByToolSessionId = new Map<string, ITerminalInstance>();
+	private readonly _onDidRegisterTerminalInstanceForToolSession = new Emitter<{ terminalToolSessionId: string }>();
+	readonly onDidRegisterTerminalInstanceForToolSession: Event<{ terminalToolSessionId: string }> = this._onDidRegisterTerminalInstanceForToolSession.event;
```
Tyriar: I think getters are the preferred way of doing this since then the event doesn't get created until it's actually used:

```suggestion
	get onDidRegisterTerminalInstanceForToolSession() { return this._onDidRegisterTerminalInstanceForToolSession.event; }
```

src/vs/workbench/contrib/terminalContrib/chat/browser/terminalChatService.ts
```diff
@@ -0,0 +1,43 @@
...
+	}
+
+	registerTerminalInstanceForToolSession(terminalToolSessionId: string | undefined, instance: ITerminalInstance): void {
+		if (!terminalToolSessionId) {
+			this._logService.warn('Attempted to register a terminal instance with an undefined tool session ID');
+			return;
+		}
+		this._terminalInstancesByToolSessionId.set(terminalToolSessionId, instance);
+		this._onDidRegisterTerminalInstanceForToolSession.fire({ terminalToolSessionId });
+		instance.onDisposed(() => {
```
Tyriar: Needs to get registered appropriately and disposed as appropriate. Could use a `DisposableMap` for example.

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/runInTerminalTool.ts
```diff
@@ -396,16 +397,18 @@ export class RunInTerminalTool extends Disposable implements IToolImpl {
 
 		const timingStart = Date.now();
 		const termId = generateUuid();
+		// The prepared data should always include a terminalToolSessionId; assert and narrow the type.
```
Tyriar: Comment no longer relevant.

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/runInTerminalTool.ts
```diff
@@ -640,7 +650,7 @@ export class RunInTerminalTool extends Disposable implements IToolImpl {
 			};
 		}
 
-		return defaultProfile;
+		return { ...defaultProfile, icon: undefined };
```
Tyriar, line 653: Other fix leaking into this PR?

src/vs/workbench/contrib/terminal/browser/terminal.ts
```diff
@@ -100,6 +101,35 @@ export interface ITerminalInstanceService {
...
+	 * Fired when a terminal instance is registered for a tool session id. This can happen after
+	 * the chat UI first renders, enabling late binding of the focus action.
+	 */
+	readonly onDidRegisterTerminalInstanceForToolSession: Event<{ terminalToolSessionId: string }>;
+
+	/**
+	 * Associate a tool session id with a terminal instance. The association is automatically
+	 * cleared when the instance is disposed.
+	 */
+	registerTerminalInstanceForToolSession(terminalToolSessionId: string | undefined, instance: ITerminalInstance): void;
```
Tyriar: `registerTerminalInstanceWithToolSession` instead of `register...` would be preferable as register has a very specific meaning in our codebase.

Tyriar (changes requested):

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/runInTerminalTool.ts
meganrogge: Just need to merge main.

src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
meganrogge: it now implemented here https://github.com/microsoft/vscode/blob/fee466e866bf2fa805ce299bd31346b4b5b73be2/src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts#L169

and I'll add telemetry

src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
```diff
@@ -212,7 +214,11 @@ export class FocusChatInstanceAction extends Action implements IAction {
 		}
 		this.label = localize('focusTerminal', 'Focus Terminal');
 		this._terminalService.setActiveInstance(this._instance);
-		this._terminalGroupService.showPanel(true);
+		if (this._instance.target === TerminalLocation.Panel) {
+			this._terminalGroupService.showPanel(true);
+		} else {
+			this._terminalEditorService.openEditor(this._instance);
+		}
```
Tyriar: I guess this was a bug before hand, could you create an issue to get it verified?

src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
```diff
@@ -129,3 +165,60 @@ CommandsRegistry.registerCommand(openTerminalSettingsLinkCommandId, async (acces
...
+			const listener = this._register(this._terminalChatService.onDidRegisterTerminalInstanceForToolSession(e => {
+				if (e.terminalToolSessionId === this._terminalToolSessionId) {
+					const terminalInstance = this._terminalChatService.getTerminalInstanceByToolSessionId(e.terminalToolSessionId);
+					if (terminalInstance) {
+						this._registerInstanceListener(terminalInstance);
+						this._store.delete(listener);
+					}
+				}
+			}));
+		}
```
Tyriar: Having an action have listeners and state like this seems dangerous. Regardless, we will definitely need to instance later so these listeners should be owned by the progress part and passed into the action as an argument. Here's an example of how to pass an argument:

https://github.com/microsoft/vscode/blob/88c209229b6d89d8b9e362f9f57ed59234f4d1c3/src/vs/workbench/contrib/terminal/browser/terminalTabsList.ts#L302-L303

Tyriar (changes requested):

src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
```diff
@@ -35,34 +42,41 @@ export class ChatTerminalToolProgressPart extends BaseChatToolInvocationSubPart
...
 			elements.title,
-			new MarkdownString(`$(${Codicon.terminal.id})\n\n\`\`\`${terminalData.language}\n${command}\n\`\`\``, { supportThemeIcons: true }),
+			new MarkdownString(`$(${Codicon.terminal.id})\n\n\`\`\`${_terminalData.language}\n${command}\n\`\`\``, { supportThemeIcons: true }),
 			undefined,
 		));
 		this._register(titlePart.onDidChangeHeight(() => this._onDidChangeHeight.fire()));
 
+		if (configurationService.getValue<string>('chat.tools.terminal.outputLocation') === 'none') {
+			this._createActionBar(elements);
+		}
```
Tyriar: I think we actually want to show the action bar regardless?

src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
```diff
@@ -84,13 +98,37 @@ export class ChatTerminalToolProgressPart extends BaseChatToolInvocationSubPart
...
+		const isTerminalHidden = this._terminalChatService.terminalIsHidden(terminalToolSessionId);
+
+		const actionBar = this._register(new ActionBar(elements.actionBar, {}));
+		const focusAction = this._register(this._instantiationService.createInstance(FocusChatInstanceAction,
+			'chat.focusTerminalInstance',
+			isTerminalHidden ? localize('showTerminal', 'Show Terminal') : localize('focusTerminal', 'Focus Terminal'),
+			ThemeIcon.asClassName(Codicon.linkExternal),
+			true,
+			terminalToolSessionId,
+			isTerminalHidden,
```
Tyriar: Let's inline some of this stuff into the `FocusChatInstanceAction.ctor`

Tyriar (dismissed):

src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
meganrogge: i'm not sure if it's a bug - maybe somewhere we depend on `setActiveInstance` for background terminals not revealing the panel?

src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
```diff
@@ -24,10 +25,17 @@ import { ConfigurationTarget } from '../../../../../../platform/configuration/co
...
+import { ITerminalChatService, ITerminalEditorService, ITerminalGroupService, ITerminalInstance, ITerminalService } from '../../../../terminal/browser/terminal.js';
+import { Action, IAction } from '../../../../../../base/common/actions.js';
+import { ThemeIcon } from '../../../../../../base/common/themables.js';
+import { localize } from '../../../../../../nls.js';
+import { TerminalLocation } from '../../../../../../platform/terminal/common/terminal.js';
 
 export class ChatTerminalToolProgressPart extends BaseChatToolInvocationSubPart {
 	public readonly domNode: HTMLElement;
 
+	private _actionBar: ActionBar | undefined;
```
Tyriar, line 37: Let's make it a `MutableDisposable` to make it harder to accidentally leak and save setting it to undefined (which isn't done below, perfect example 🙂)

src/vs/workbench/contrib/terminal/browser/terminal.ts
```diff
@@ -100,6 +101,35 @@ export interface ITerminalInstanceService {
...
+	registerTerminalInstanceWithToolSession(terminalToolSessionId: string | undefined, instance: ITerminalInstance): void;
+
+	/**
+	 * Resolve a terminal instance by its tool session id.
+	 * @param terminalToolSessionId The tool session id provided in toolSpecificData.
+	 * If no tool session ID is provided, we do nothing.
+	 */
+	getTerminalInstanceByToolSessionId(terminalToolSessionId: string): ITerminalInstance | undefined;
+
+	terminalIsHidden(terminalToolSessionId: string): boolean;
```
Tyriar, line 130: Could be `ITerminalService.isBackgroundTerminal` as it's generally useful? That'll save exposing `foregroundInstances` too.

src/vs/workbench/contrib/terminal/browser/terminalService.ts
```diff
@@ -1174,10 +1178,15 @@ export class TerminalService extends Disposable implements ITerminalService {
...
-		if (this.instances.length === 1) {
+		if (this.instances.length === 1 && !suppressSetActive) {
 			this._terminalGroupService.setActiveInstanceByIndex(0);
 		}
 
+		if (forceSaveState) {
+			// Skips the debounce of _saveState as we are shutting down
+			// and need to save the revealed hideFromUser terminals
+			await this._saveStateNow();
+		}
```
Tyriar, line 1189: `showBackgroundTerminal` doesn't imply it's shutting down. It's fine to do this, just should tweak to comment to "in case it's shutting down" or similar.

src/vs/workbench/contrib/terminalContrib/chat/browser/terminalChatService.ts
```diff
@@ -0,0 +1,159 @@
...
+		super();
+
+		this._restoreFromStorage();
+		this._register(this._lifecycleService.onBeforeShutdown(async e => {
+			let veto = false;
+			// Show all hidden terminals before shutdown so they are restored
+			for (const [toolSessionId, instance] of this._terminalInstancesByToolSessionId) {
+				if (this.terminalIsHidden(toolSessionId) && (instance.capabilities.get(TerminalCapability.CommandDetection)?.promptInputModel.state === PromptInputState.Execute || instance.hasChildProcesses)) {
+					await this._terminalService.showBackgroundTerminal(instance, true, true);
+					veto = true;
```
Tyriar, line 50: A veto will prevent shutdown, we don't want that right?

src/vs/workbench/contrib/terminalContrib/chat/browser/terminalChatService.ts
```diff
@@ -0,0 +1,159 @@
...
+	) {
+		super();
+
+		this._restoreFromStorage();
+		this._register(this._lifecycleService.onBeforeShutdown(async e => {
+			let veto = false;
+			// Show all hidden terminals before shutdown so they are restored
+			for (const [toolSessionId, instance] of this._terminalInstancesByToolSessionId) {
+				if (this.terminalIsHidden(toolSessionId) && (instance.capabilities.get(TerminalCapability.CommandDetection)?.promptInputModel.state === PromptInputState.Execute || instance.hasChildProcesses)) {
+					await this._terminalService.showBackgroundTerminal(instance, true, true);
```
Tyriar, line 49: The await here is a bit of a problem actually. We'll likely get pinged by Ben about slowing shutdown is this lands, especially in its current state.

Maybe a better solution is to add the flag to `IShellLaunchConfig` where we force persistence despite `hideFromUser`? That also comes with the benefit of keeping the terminal hidden.

Tyriar (changes requested):

Tyriar (approved):
Approving so we can start self hosting, but let's get that async in lifecycle shutdown events gone tomorrow.

src/vs/workbench/contrib/terminalContrib/chat/browser/terminalChatService.ts
meganrogge: fixed https://github.com/microsoft/vscode/pull/271806

src/vs/workbench/contrib/terminal/browser/terminal.ts
meganrogge: we still need `foregroundInstances` exposed on `ITerminalService`

meganrogge: but I like the rename

src/vs/workbench/contrib/terminalContrib/chat/browser/terminalChatService.ts
meganrogge: `hideFromUser` isn't explicitly ignored. We save state using the group service in the form of the panel layout. We need to reveal the terminal there and save state for it to work.

PR #271454, opened 2025-10-15
fixup todo styling

bhavyaus:
- Removed todo widget rendering on top of the chat panel
- Added todo widget rendering on top of the chat input as the only supported UX
- Added a setting to turn off widget display with `chat.tools.todos.showWidget`
- Matched styling of the todos with the edit-session widget


bpasero:
There also seems to be an issue when the todo list is empty but the container still applies its styling:

image

amunger (dismissed):

roblourens (approved):

PR #271503, opened 2025-10-15
fix compile errors, fyi @roblourens

jrieken:


Tyriar (approved):
@jrieken I recommend avoiding @ in commit messages as that sends people unnecessary notifications (forks can uplevel in interesting ways).

PR #271556, opened 2025-10-15
Resolve issue where first input gets ignored on new conpty with a ding sound

### LINKED ISSUE #243584: First input gets ignored on pwsh/conpty
1. Open Insiders on Windows
2. Open the terminal, immediately type in `git status`, and press enter
3. `it status` appears to be typed into the terminal, and the terminal gives an error.

I'm not sure if it's from the history, shell integration, or suggest feature.

Terminal error after issuing it status

anthonykim1:
Resolves: https://github.com/microsoft/vscode/issues/243584

Updated summary:
There were two request I think in `\e[6n\e[c`.
We were sending `\e[?61;4c\e[2;1R`, which is wrong.
The one ending with 4c, was supposed to come later.

Because the order was wrong, ConPTY "ate" the first input from user thinking that must've been the response to their cursor position request, and made a ding sound because it was invalid.

### DIFF
--- a/src/vs/platform/terminal/node/ptyService.ts
+++ b/src/vs/platform/terminal/node/ptyService.ts
@@ -238,14 +238,18 @@ export class PtyService extends Disposable implements IPtyService {
 	private async _reviveTerminalProcess(workspaceId: string, terminal: ISerializedTerminalState): Promise<void> {
 		const restoreMessage = localize('terminal-history-restored', "History restored");
 
-		// For Windows ConPTY, we must NOT pass the replay data as initialText because that enables
-		// conptyInheritCursor which tries to parse cursor positions from the serialized xterm data.
-		// This causes cursor misalignment because the escape sequences were from a different session.
-		// Instead, we let ConPTY start with a completely clean slate and handle everything via replay.
-		let initialText: string | undefined;
-		if (!isWindows) {
-			// Non-Windows: Preserve legacy behavior of passing replay as initialText (safe on pty implementations)
-			initialText = terminal.replayEvent.events[0].data + formatMessageForTerminal(restoreMessage, { loudFormatting: true });
+		// Conpty v1.22+ uses passthrough and doesn't reprint the buffer often, this means that when
+		// the terminal is revived, the cursor would be at the bottom of the buffer then when
+		// PSReadLine requests `GetConsoleCursorInfo` it will be handled by conpty itself by design.
+		// This causes the cursor to move to the top into the replayed terminal contents. To avoid
+		// this, the post restore message will print new lines to get a clear viewport and put the
+		// cursor back at to top left.
+		let postRestoreMessage = '';
+		if (isWindows) {
+			const lastReplayEvent = terminal.replayEvent.events.length > 0 ? terminal.replayEvent.events.at(-1) : undefined;
+			if (lastReplayEvent) {
+				postRestoreMessage += '\r\n'.repeat(lastReplayEvent.rows - 1) + `\x1b[H`;
+			}
 		}
 
 		// TODO: We may at some point want to show date information in a hover via a custom sequence:
@@ -258,7 +262,7 @@ export class PtyService extends Disposable implements IPtyService {
 				color: terminal.processDetails.color,
 				icon: terminal.processDetails.icon,
 				name: terminal.processDetails.titleSource === TitleEventSource.Api ? terminal.processDetails.title : undefined,
-				initialText
+				initialText: terminal.replayEvent.events[0].data + formatMessageForTerminal(restoreMessage, { loudFormatting: true }) + postRestoreMessage
 			},
 			terminal.processDetails.cwd,
 			terminal.replayEvent.events[0].cols,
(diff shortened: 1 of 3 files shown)


src/vs/platform/terminal/node/ptyService.ts
```diff
@@ -912,11 +932,54 @@ class PersistentTerminalProcess extends Disposable {
...
+			this._logService.info(`Persistent process "${this._persistentProcessId}": Skipping in-start Windows replay (already handled pre-start)`);
+			return; // Avoid double replay on Windows revival
+		}
+
 		this._onProcessReplay.fire(ev);
 		this._terminalProcess.clearUnacknowledgedChars();
 		this._onPersistentProcessReady.fire();
 	}
 
+	private async _preStartReplayWindows(): Promise<void> {
```
anthonykim1: this _preStartReplayWindows() is copilot generated.
I wanted it to correctly generate the history and the history restored message. Seems good from testing, but I want to be careful.

src/vs/platform/terminal/node/ptyService.ts
```diff
@@ -823,19 +839,23 @@ class PersistentTerminalProcess extends Disposable {
 
 	async start(): Promise<ITerminalLaunchError | ITerminalLaunchResult | undefined> {
 		if (!this._isStarted) {
+			// On Windows when reviving, perform replay BEFORE starting the actual pty so that
+			// the viewport and scrollback are prepared (blank viewport, history in scrollback)
+			// before the shell prompt is emitted. This avoids the race where the shell prompt
+			// appears in the middle of replay manipulation causing cursor/beep issues.
+			if (this._wasRevived && isWindows) {
+				await this._preStartReplayWindows();
+			}
```
Tyriar: This issue is purely about a race condition on startup? If so, can we await `triggerReplay` on all platforms instead? This PR adds a bunch of Windows only stuff which makes testing harder and me self-hosting on Windows basically no longer is as effective self-hosting of this feature on other platforms.

src/vs/platform/terminal/node/ptyService.ts
Tyriar: Main concern here is this is mostly already done elsewhere and duplicates code doesn't it?

Tyriar (changes requested):

src/vs/platform/terminal/node/ptyService.ts
anthonykim1: Hmm.. I thought there was more to this:

I thought having `initialText` makes ConPTY and shells to think we performed bunch of operations -> ding sound from position request.

But rather with the approach I have right now, we are not informing ConPTY+shell, but rather just rendering things on xterm so user can see. The `initialText` becomes part of `shellLaunchConfig` which then goes into constructor of `TerminalProcess` and not `PersistentTerminalProcess` 
https://github.com/microsoft/vscode/blob/d9210f67df6ad2feb2e9c37bab99eb4ce6a9942c/src/vs/platform/terminal/node/ptyService.ts#L254-L261
https://github.com/microsoft/vscode/blob/d9210f67df6ad2feb2e9c37bab99eb4ce6a9942c/src/vs/platform/terminal/node/ptyService.ts#L307
When I tried with only `await this.triggerReplay()`, I get the ding + first character ignored again.

src/vs/workbench/contrib/terminal/browser/terminalInstance.ts
```diff
@@ -844,7 +840,6 @@ export class TerminalInstance extends Disposable implements ITerminalInstance {
 
 		this._register(this._processManager.onProcessData(e => this._onProcessData(e)));
 		this._register(xterm.raw.onData(async data => {
-			await this._pauseInputEventBarrier?.wait();
```
anthonykim1, line 847: Now the approach is we remove the wait from `onData` and explicitly put `this._processManager.write('\x1b[?61;4c');`. I'm thinking this would allow `this._processManager.write('\x1b[?61;4c');` to be sent later.

src/vs/workbench/contrib/terminal/browser/terminalInstance.ts
```diff
@@ -858,7 +853,9 @@ export class TerminalInstance extends Disposable implements ITerminalInstance {
 			if (processTraits?.windowsPty?.backend === 'conpty') {
 				this._register(xterm.raw.parser.registerCsiHandler({ final: 'c' }, params => {
 					if (params.length === 0 || params.length === 1 && params[0] === 0) {
-						this._processManager.write('\x1b[?61;4c');
+						queueMicrotask(() => {
+							this._processManager.write('\x1b[?61;4c');
+						});
```
Tyriar: Instead of this, a method for the onData listener and have this call that would be better. That will also mean that we get logs in the Terminal channel for this (one of the reasons it was hard to debug), and any future changes to that listener will also behave properly.

Tyriar (changes requested):
Forgot to submit the review 🤦

src/vs/workbench/contrib/terminal/browser/terminalInstance.ts
anthonykim1: sounds good!
Do you like the name `_sendDataToProcess` or should it be `_writeDataToProcess`? [c53bef2](https://github.com/microsoft/vscode/pull/271556/commits/c53bef2a1cf7922de75d3a0f146db3cf7049d7ca)

src/vs/workbench/contrib/terminal/browser/terminalInstance.ts
```diff
@@ -856,9 +855,9 @@ export class TerminalInstance extends Disposable implements ITerminalInstance {
...
 			// Reference: https://github.com/microsoft/terminal/blob/3760caed97fa9140a40777a8fbc1c95785e6d2ab/src/terminal/adapter/adaptDispatch.cpp#L1471-L1495
 			if (processTraits?.windowsPty?.backend === 'conpty') {
-				this._register(xterm.raw.parser.registerCsiHandler({ final: 'c' }, params => {
+				this._register(xterm.raw.parser.registerCsiHandler({ final: 'c' }, async params => {
 					if (params.length === 0 || params.length === 1 && params[0] === 0) {
-						this._processManager.write('\x1b[?61;4c');
+						await this._sendDataToProcess('\x1b[?61;4c');
 						return true;
 					}
 					return false;
```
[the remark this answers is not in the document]
anthonykim1: I thought we should await both since we are awaiting for generic onData above:
```
this._register(xterm.raw.onData(async data => {
			await this._sendDataToProcess(data);
		}));
```

src/vs/workbench/contrib/terminal/browser/terminalInstance.ts
Tyriar: `_handleOnData` is what I'd probably have gone with, but these are fine names too.

src/vs/workbench/contrib/terminal/browser/terminalInstance.ts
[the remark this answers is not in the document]
Tyriar: We can use async, but we don't need to and shouldn't. There are some bad performance problems with using async as it must block the parser: https://github.com/xtermjs/xterm.js/blob/1bf90d9f478f993362467a92f7d6197a22cba473/typings/xterm.d.ts#L1818-L1829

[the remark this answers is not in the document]
Tyriar: Copilot's description is correct I think before we added the parser API, after it works but is just slow.

Tyriar (changes requested):
Just the async thing needs to change

Tyriar (approved):

PR #271590, opened 2025-10-15
Add support for copying math source from katex blocks

### LINKED ISSUE #258286: Add copy button for chat math equations

pwang347:
## Screenshot
image

Fixes #258286


mjbvz (dismissed):
Seems reasonable to me. Thanks for looking into this!

Tyriar (approved):
Looks a little fragile to me, but @mjbvz had a look

pwang347:
> Looks a little fragile to me, but @mjbvz had a look

Happy to discuss any changes you think would improve the stability here! Probably would be nicer to not have to do the `closest` call twice, but this fit cleaner into the existing action framework imo without needing to forward args.

Tyriar:
Dealing with classes like this is basically touching internals of katex that could change in the future and therefore brick this feature on our end without any indication. Unless they have committed to the DOM structure that is.

pwang347:
> Dealing with classes like this is basically touching internals of katex that could change in the future and therefore brick this feature on our end without any indication. Unless they have committed to the DOM structure that is.

That makes sense, to my knowledge the class name has been around basically since the beginning so it should be relatively stable (but not guaranteed). This is good feedback though, I can check if we can inject a custom attribute at the same element and use that instead in a follow-up PR. Thanks!

PR #271187, opened 2025-10-13
feat: display description of prompt as placeholder text when user types `/{prompt}` in chat input

bryanchen-d:
display description of prompt as placeholder text when user types `/{prompt}` in chat input

Fixes: https://github.com/microsoft/vscode/issues/269237

PR #271596, opened 2025-10-15
Add context menus to suggested prompts

### LINKED ISSUE #266374: Right click welcome prompts should allow me to configure them
Right click does nothing, it could be a way to open the prompt file directly.

Image

benvillalobos:
Fixes https://github.com/microsoft/vscode/issues/266374

To get a welcome prompt to show, add this setting to your user settings json (or the project's settings.json in .vscode)

You can also add your own user or project prompt (Settings Gear -> Prompt Files -> Add new) and include the name (excluding `.prompt.md`) here.
```
  "chat.promptFilesRecommendations": {
    "setup-environment": true,
    "deepwiki": true
  },
```

Note: the video shows the label as `Open Prompt File`, but I've since changed it to `Edit Prompt File`

https://github.com/user-attachments/assets/966dad70-6b59-44eb-9bb3-63c15bf3ea8c


OrenMe:
That's nice
Maybe also add option to go to their configuration to add:remove

eleanorjboyd (approved):

src/vs/workbench/contrib/chat/browser/chatWidget.ts
```diff
@@ -1551,6 +1554,7 @@ export class ChatWidget extends Disposable implements IChatWidget {
 				if (promptNames.includes(promptCommand.command)) {
 					try {
 						if (promptCommand.promptPath) {
+							this.promptUriCache.set(promptCommand.command, promptCommand.promptPath.uri);
```
roblourens, line 1665: In https://github.com/microsoft/vscode/pull/271187/files Bryan will be adding this cache to the prompt file service. It needs to be there IMO so we can properly invalidate the cache and use it in multiple places

src/vs/workbench/contrib/chat/browser/viewsWelcome/chatViewWelcomeController.ts
```diff
@@ -271,6 +298,27 @@ export class ChatViewWelcomePart extends Disposable {
 		}
 	}
 
+	private getPromptContextMenuActions(prompt: IChatSuggestedPrompts): IAction[] {
+		const actions: IAction[] = [];
+		if (prompt.uri) {
+			const uri = prompt.uri;
+			actions.push(new Action(
```
roblourens, line 305: Does it make sense for this to be a registered Action contributed to a Menu? An advantage would be things like having a keyboard shortcut for it or showing it in the command palette. If not and if you don't see this getting much more complex, I think the one-off is ok too.

benvillalobos: > Does it make sense for this to be a registered Action contributed to a Menu?

This particular action should be fine as a one off, since it will always open specific file. A general "Open Prompt File" command sounds like a good idea though, will open an issue.

benvillalobos:

src/vs/workbench/contrib/chat/browser/chatWidget.ts
benvillalobos: Filed https://github.com/microsoft/vscode/issues/272302

Issue #271101, opened 2025-10-13
Git: Support deep link to continue working on PR in VS Code


joaomoreno:
We should support a deep link URL that allows us to continue working on a PR in VS Code:

- IMO we should augment the git clone URL support to allow remembering which local workspace already had a particular repository open such that the second time a user would open such a deep git clone URL, they would get prompted: would you like to open folder X, or clone anyways?
- Once that functionality is implemented, we could extend it to a deep link GHPRI URL that leverages it to stitch together the flow of continue working/reviewing a PR from the Web to the Desktop.

PR #271652, opened 2025-10-16
Maintain a cache of repo remotes that we know about

alexr00:
Part of #271101


extensions/git/src/knownFolders.ts
```diff
@@ -0,0 +1,143 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { Memento, workspace } from 'vscode';
+import * as path from 'path';
+import { LRUCache } from './cache';
+
+export class KnownFolders {
```
lszomoru: nit: should this be called RepositoryCache?

extensions/git/src/knownFolders.ts
```diff
@@ -0,0 +1,143 @@
...
+
+	private static readonly STORAGE_KEY = 'git.knownFolders';
+	private static readonly MAX_REPO_ENTRIES = 30; // Max repositories tracked
+	private static readonly MAX_FOLDER_ENTRIES = 10; // Max folders per repository
+	private static readonly EXPIRY_MS = 90 * 24 * 60 * 60 * 1000; // 90 days in ms
```
lszomoru: Given that we already cap both the inner/outer entries I am not sure we need to prune based on time. Would that also simplify the inner cache as well as some of the cache management code?

extensions/git/src/knownFolders.ts
```diff
@@ -0,0 +1,143 @@
...
+		}
+
+
+		foldersLru.set(folderPathOrWorkspaceFile, Date.now()); // touch/update timestamp
+		this.lru.set(repoUrl, foldersLru);
+		this.save();
+	}
+
+	/**
+	 * We should possibly support converting between ssh remotes and http remotes.
```
lszomoru: I would probably "normalize" the remote when storing it but we can tackle this later.

extensions/git/src/knownFolders.ts
```diff
@@ -0,0 +1,143 @@
...
+
+export class KnownFolders {
+
+	private static readonly STORAGE_KEY = 'git.knownFolders';
+	private static readonly MAX_REPO_ENTRIES = 30; // Max repositories tracked
+	private static readonly MAX_FOLDER_ENTRIES = 10; // Max folders per repository
+	private static readonly EXPIRY_MS = 90 * 24 * 60 * 60 * 1000; // 90 days in ms
+
+	// Outer LRU: repoUrl -> inner LRU (folderPathOrWorkspaceFile -> timestamp ms). Only keys matter externally, timestamp tracks last seen.
+	private readonly lru = new LRUCache<string, LRUCache<string, number>>(KnownFolders.MAX_REPO_ENTRIES);
```
lszomoru: Should the "inner" cache use URI instead of a string?

extensions/git/src/knownFolders.ts
```diff
@@ -0,0 +1,143 @@
...
+						}
+						inner.set(folderPath, ts);
+					}
+					if (inner.size) {
+						this.lru.set(repo, inner);
+					}
+				}
+			}
+		} catch {
+			// Ignore corrupt state; start fresh.
```
lszomoru: I think that we should log a "warning" here.

extensions/git/src/model.ts
```diff
@@ -849,6 +855,16 @@ export class Model implements IRepositoryResolver, IBranchProtectionProviderRegi
 		this._onDidOpenRepository.fire(repository);
 	}
 
+	private _updateKnownFolders(repository: Repository): void {
```
lszomoru: nit: I would move this method/logic into the cache.

extensions/git/src/repository.ts
```diff
@@ -1848,10 +1850,16 @@ export class Repository implements Disposable {
 
 	async addRemote(name: string, url: string): Promise<void> {
 		await this.run(Operation.Remote, () => this.repository.addRemote(name, url));
+		this.knownFolders.set(url, this.root);
```
lszomoru: This should be moved inside the callback above so that it is part of the operation.

extensions/git/src/repository.ts
```diff
@@ -1848,10 +1850,16 @@ export class Repository implements Disposable {
...
 	}
 
 	async removeRemote(name: string): Promise<void> {
 		await this.run(Operation.Remote, () => this.repository.removeRemote(name));
+		const remote = this.remotes.find(remote => remote.name === name);
```
lszomoru: This should be moved inside the callback above so that it is part of the operation.

extensions/git/src/repository.ts
```diff
@@ -1848,10 +1850,16 @@ export class Repository implements Disposable {
...
+		const remoteUrl = remote?.fetchUrl ?? remote?.pushUrl;
```
lszomoru: For both "add" and "remove" I would just pass in the repository into let's say `updateCache()` that is a method in the Cache which would just update the cache for all the remotes of the repository.

lszomoru (changes requested):

extensions/git/src/knownFolders.ts
alexr00: I don't think that will work, as the the the URIs may not have the same reference.

extensions/git/src/knownFolders.ts
alexr00: It seemed nice to have, but no, we don't need it.

extensions/git/src/repository.ts
alexr00: I ended up needing to pass in the removed remotes. We can have this case:
- User clones `microsoft/vscode` into `./vscode-1`
- User clones `microsoft/vscode` into `./vscode-2`
- User deletes `microsoft/vscode` remote from `./vscode-1`
- We need to pass in the repo URL so that we know which repo to remove from the cache
- We need to pass in the repo root path so that we don't just remove all folders associated with `microsoft/vscode` (we only want to remove `./vscode-1`, not `./vscode-2`)

extensions/git/src/test/knownFolders.test.ts
```diff
@@ -0,0 +1,135 @@
...
+	append(_value: string): void { }
+	appendLine(_value: string): void { }
+	replace(_value: string): void { }
+	clear(): void { }
+	show(_column?: unknown, _preserveFocus?: unknown): void { }
+	hide(): void { }
+	dispose(): void { }
+}
+
+class RepoCache extends RepositoryCache {
```
lszomoru: Let's call this `TestRepositoryCache`

extensions/git/src/test/knownFolders.test.ts
lszomoru, line 1: Can we rename the file to `reposiotryCache.test.ts`? Thanks!

extensions/git/src/repositoryCache.ts
```diff
@@ -0,0 +1,173 @@
...
+				}
+			}
+			return folderPathOrWorkspaceFile;
+		} catch {
+			return;
+		}
+
+	}
+
+	updateRepository(addedRemotes: Remote[], removedRemotes: Remote[], rootPath: string): void {
```
lszomoru: Can we call this method `update` so that it is consistent with the rest?

lszomoru: Could we just pass in an array of remote: Remote[] instead of added/removed and work with that?

alexr00: We cannot. We need to know which remotes were specifically removed. If we simply pass in the "current" remotes", then we may remove a remote that is used by both this folder and some other folder.

lszomoru (approved):

Issue #270855, opened 2025-10-11
Support Agent/Mode/Model picker for new session creation


rebornix:
Currently, chat editors for contributed agents don’t support mode, or model pickers. Users also can’t select an agent when starting a new or untitled chat session.

We’ll explore how to enable these pickers for agents that support such capabilities, ensuring a consistent creation experience across chat editors.

PR #271728, opened 2025-10-16
ChatSessions: Select Model

joshspicer:
ref https://github.com/microsoft/vscode/issues/270855

First pass at method for chat session providers to forward a list of 'models' and those models be selected in a chat session. 

https://github.com/user-attachments/assets/69ceb82d-61fa-4960-abbd-b206524000e6

### DIFF
--- a/src/vs/workbench/contrib/chat/browser/chatInputPart.ts
+++ b/src/vs/workbench/contrib/chat/browser/chatInputPart.ts
@@ -80,7 +80,8 @@ import { ChatContextKeys } from '../common/chatContextKeys.js';
 import { IChatEditingSession, ModifiedFileEntryState } from '../common/chatEditingService.js';
 import { IChatRequestModeInfo } from '../common/chatModel.js';
 import { ChatMode, IChatMode, IChatModeService } from '../common/chatModes.js';
-import { IChatFollowup } from '../common/chatService.js';
+import { IChatFollowup, IChatService } from '../common/chatService.js';
+import { IChatSessionsService } from '../common/chatSessionsService.js';
 import { ChatRequestVariableSet, IChatRequestVariableEntry, isElementVariableEntry, isImageVariableEntry, isNotebookOutputVariableEntry, isPasteVariableEntry, isPromptFileVariableEntry, isPromptTextVariableEntry, isSCMHistoryItemChangeRangeVariableEntry, isSCMHistoryItemChangeVariableEntry, isSCMHistoryItemVariableEntry } from '../common/chatVariableEntries.js';
 import { IChatResponseViewModel } from '../common/chatViewModel.js';
 import { ChatInputHistoryMaxEntries, IChatHistoryEntry, IChatInputState, IChatWidgetHistoryService } from '../common/chatWidgetHistoryService.js';
@@ -292,6 +293,7 @@ export class ChatInputPart extends Disposable implements IHistoryNavigationWidge
 	private chatModeKindKey: IContextKey<ChatModeKind>;
 	private withinEditSessionKey: IContextKey<boolean>;
 	private filePartOfEditSessionKey: IContextKey<boolean>;
+	private inContributedSessionWithModelsKey: IContextKey<boolean>;
 
 	private modelWidget: ModelPickerActionItem | undefined;
 	private modeWidget: ModePickerActionItem | undefined;
@@ -382,6 +384,11 @@ export class ChatInputPart extends Disposable implements IHistoryNavigationWidge
 	 */
 	private _generating?: { rc: number; defer: DeferredPromise<void> };
 
+	/**
+	 * Reference to the chat widget this input part belongs to
+	 */
+	private _widget?: IChatWidget;
+
 	constructor(
 		// private readonly editorOptions: ChatEditorOptions, // TODO this should be used
 		private readonly location: ChatAgentLocation,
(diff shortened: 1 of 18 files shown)


src/vscode-dts/vscode.proposed.chatSessionsProvider.d.ts
```diff
@@ -166,6 +171,32 @@ declare module 'vscode' {
...
 		 */
 		provideChatSessionContent(sessionId: string, token: CancellationToken): Thenable<ChatSession> | ChatSession;
+
+		/**
+		 *
+		 * @param sessionId Identifier of the chat session being updated.
+		 * @param updates Collection of option identifiers and their new values. Only the options that changed are included.
+		 * @param token A cancellation token that can be used to cancel the notification if the session is disposed.
+		 */
+		provideHandleOptionsChange?(sessionId: string, updates: ReadonlyArray<ChatSessionOptionUpdate>, token: CancellationToken): void;
```
mjbvz, line 190: Can this live on the `ChatSession` as an optional `acceptOptionsChange` method?

Also I think we should type out what the options are instead of having a generic `ChatSessionOptionUpdate`

src/vscode-dts/vscode.proposed.chatSessionsProvider.d.ts
```diff
@@ -166,6 +171,32 @@ declare module 'vscode' {
...
+
+		/**
+		 * Called as soon as you register (call me once)
+		 * @param token
+		 */
+		provideChatSessionOptions?(token: CancellationToken): Thenable<ChatSessionOptions> | ChatSessionOptions;
```
mjbvz: We may want to move this to the `registerChatSessionProvider` call to align with our other APIs. It's tied to the provider itself but having it here makes it feel like it's going to be invoked per session

Maybe at least call it `provideChatSessionProviderOptions` to make it clear it's not per-session

src/vscode-dts/vscode.proposed.chatSessionsProvider.d.ts
```diff
@@ -212,6 +243,13 @@ declare module 'vscode' {
 		supportsInterruptions?: boolean;
 	}
 
+	export interface ChatSessionOptions {
```
mjbvz: Maybe `ChatSessionProviderOptions` to make it very clear this is per chat session type, not per session

mjbvz (commented):
Adding some initial thoughts on the API part of things

src/vs/workbench/contrib/chat/browser/chatWidget.ts
```diff
@@ -2188,6 +2192,16 @@ export class ChatWidget extends Disposable implements IChatWidget {
 		this.container.setAttribute('data-session-id', model.sessionId);
 		this.viewModel = this.instantiationService.createInstance(ChatViewModel, model, this._codeBlockModelCollection);
 
+		// Contributed session model handling: register contributed models with unified language model service
+		// TODO: This needs to be 'reset' or if will affect new chat editors. Understand this more.
+		const sessionCtx = this.chatService.getChatSessionFromInternalId(model.sessionId);
+		if (sessionCtx && sessionCtx.chatSessionType !== 'local') {
+			const contributedModels = this.chatSessionsService.getModelsForSessionType(sessionCtx.chatSessionType) ?? [];
+			this.languageModelsService.setContributedSessionModels(sessionCtx.chatSessionType, contributedModels);
```
roblourens: Ideally the view wouldn't have to pass these models around, can the ChatService or something else be responsible for syncing these into the LM service?

src/vs/workbench/contrib/chat/browser/chatInputPart.ts
```diff
@@ -702,19 +712,45 @@ export class ChatInputPart extends Disposable implements IHistoryNavigationWidge
...
 			this.storageService.store('chat.cachedLanguageModels', models, StorageScope.APPLICATION, StorageTarget.MACHINE);
 		}
-		models.sort((a, b) => a.metadata.name.localeCompare(b.metadata.name));
 		return models.filter(entry => entry.metadata?.isUserSelectable && this.modelSupportedForDefaultAgent(entry));
 	}
 
+	/**
+	 * Called by the owning ChatWidget after a model is bound, if the session is contributed.
+	 */
+	public setContributedSession(sessionType: string, sessionId: string, cachedModelId?: string): void {
```
roblourens: Can this happen in the constructor or in `initForNewChatModel` so that it's less of a special case?

src/vs/workbench/contrib/chat/browser/chatWidget.ts
roblourens: Or the LM Service just depends on the ChatSessionService and gets the models from there? Or we just always read them from ChatSessionService? Because you're basically just copying the data from one service to another right?

lramos15: I assume what we want here is similar to what mode does which is one place where the picker gets filtered down to the capable models. Unless you're proposing that the language model service starts accepting mode + session and filters based on those params?

src/vs/workbench/contrib/chat/browser/chatInputPart.ts
joshspicer: removed

src/vs/workbench/contrib/chat/browser/chatWidget.ts
joshspicer: relying on chatsessionservice to populate directly

src/vscode-dts/vscode.proposed.chatSessionsProvider.d.ts
joshspicer: in https://github.com/microsoft/vscode/pull/271728/commits/dde0567bf23a9bea8bdf8438449e471f168dff8a

src/vscode-dts/vscode.proposed.chatSessionsProvider.d.ts
joshspicer: I like this. It wasn't straightforward to me how to initially do this so lets talk about it more perhaps?  I left out the typing and curious what you think that should look like?

rebornix (approved):

PR #271758, opened 2025-10-16
add View chat terminals action

### LINKED ISSUE #271385: Provide command to manage of active chat terminal sessions
After https://github.com/microsoft/vscode/issues/271384, terminal sessions will not show in the terminal view by default. So we should have some way of managing them since they are active processes behind the scenes.

- Initially this could be a command that shows in the chat overflow menu.
- The most important thing here is to show all backgrounded (hideFromUser) terminals, but we may want to show all linked terminal sessions even the visible ones?

meganrogge:
fixes #271385

Screenshot 2025-10-17 at 4 21 27 PM

Screenshot 2025-10-17 at 11 57 50 AM


src/vs/workbench/contrib/terminal/common/terminalContextKey.ts
```diff
@@ -140,6 +140,9 @@ export namespace TerminalContextKeys {
 	/** Whether shell integration is enabled in the active terminal. This only considers full VS Code shell integration. */
 	export const terminalShellIntegrationEnabled = new RawContextKey<boolean>(TerminalContextKeyStrings.TerminalShellIntegrationEnabled, false, localize('terminalShellIntegrationEnabled', "Whether shell integration is enabled in the active terminal"));
 
+	/** Whether there is at least one tool terminal (created via run_in_terminal). */
+	export const hasToolTerminal = new RawContextKey<boolean>('hasToolTerminal', false, localize('hasToolTerminalContextKey', "Whether there is at least one tool terminal (created via a tool session)."));
```
meganrogge:
```suggestion
	export const hasToolTerminal = new RawContextKey<boolean>('hasToolTerminal', false, localize('hasToolTerminalContextKey', "Whether there is at least one tool terminal (created via run_in_terminal)."));
```

meganrogge:
Should we say `View` vs `Show` chat terminals?

connor4312 (approved):

PR #271918, opened 2025-10-17
Toolbars handle rerenders

benibenj:
```Copilot Generated Description:``` Implement event handling for toolbar and composite bar rerenders to ensure proper layout updates. Add mechanisms to trigger updates when items are rerendered in various components.

https://github.com/microsoft/vscode/issues/248571


bpasero (commented):
Wouldn't a nicer approach be to let `IActionViewItem` have an optional event that downstream clients could implement if they want to notify about a label change? And then the widget registers to this event if the view item provides it?

PR #271944, opened 2025-10-17
align terminal task spinner with status bar one

### LINKED ISSUE #271688: Spinners rotate at different speeds
It also looks like they are using different easing curves:

Image

---

meganrogge:
fix #271688


bpasero (changes requested):
🤔 this changes the icon from the one codicon to another codicon, is that what the issue talks about? I am confused.

meganrogge:
Yep! We want them to match.

bpasero:
@meganrogge my understanding is that the icons match but spin at different intervals:

image

We have 2 rotating icons: one with arrows and one without, but that does not seem to be the issue that @mrleemurray reported?

PR #271967, opened 2025-10-17
Ensure restored terminal sessions are accounted for in `runInTerminal` tool

### LINKED ISSUE #271964: Restored hideFromUser sessions are not disposed of when the session is cleared
1. Have the agent run something in the terminal 
2. Click show or have it be a long running process so that it gets restored
3. Reload the window
4. Click the new chat button, which should clear the existing terminals
5. 🐛 terminal sticks around

meganrogge:
fix #271964

Now that we persist terminals in `terminalChatService`, we need to be sure that the `RunInTerminal` tool's associations are updated.

It would be good to eventually just have the `RunInTerminalTool` pull from `terminalChatService`.

### DIFF
--- a/src/vs/workbench/contrib/terminalContrib/chat/browser/terminalChatService.ts
+++ b/src/vs/workbench/contrib/terminalContrib/chat/browser/terminalChatService.ts
@@ -22,8 +22,8 @@ export class TerminalChatService extends Disposable implements ITerminalChatServ
 
 	private readonly _terminalInstancesByToolSessionId = new Map<string, ITerminalInstance>();
 	private readonly _terminalInstanceListenersByToolSessionId = this._register(new DisposableMap<string, IDisposable>());
-	private readonly _onDidRegisterTerminalInstanceForToolSession = new Emitter<ITerminalInstance>();
-	readonly onDidRegisterTerminalInstanceWithToolSession: Event<ITerminalInstance> = this._onDidRegisterTerminalInstanceForToolSession.event;
+	private readonly _onDidRegisterTerminalInstanceForToolSession = new Emitter<{ toolSessionId: string; instance: ITerminalInstance }>();
+	readonly onDidRegisterTerminalInstanceWithToolSession: Event<{ toolSessionId: string; instance: ITerminalInstance }> = this._onDidRegisterTerminalInstanceForToolSession.event;
 
 	/**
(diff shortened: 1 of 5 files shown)


src/vs/workbench/contrib/terminalContrib/chat/browser/terminalChatService.ts
```diff
@@ -69,6 +69,10 @@ export class TerminalChatService extends Disposable implements ITerminalChatServ
 		}
 	}
 
+	getToolSessionTerminalInstances(): readonly ITerminalInstance[] {
+		return Array.from(this._terminalInstancesByToolSessionId.values());
+	}
```
Tyriar, line 74: Returning the iterator is better here, we can then use that instead of constructing the array on the outside.

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/runInTerminalTool.ts
```diff
@@ -196,6 +196,18 @@ export class RunInTerminalTool extends Disposable implements IToolImpl {
...
+		for (const instance of this._terminalChatService.getToolSessionTerminalInstances()) {
+			this._restoreAssociationForInstance(instance);
+		}
+
+		// Listen for future registrations (including restored mappings) so we can rebuild
+		// session associations if they weren't present when _restoreTerminalAssociations ran.
+		// This ensures when a session is cleared, the terminal is disposed of
+		this._register(this._terminalChatService.onDidRegisterTerminalInstanceWithToolSession(e => {
+			this._restoreAssociationForInstance(e.instance);
+		}));
```
Tyriar, line 210: I think we need to handle the case where the association isn't made and kill the terminal? Don't know the repro not, but keeping hidden terminals around needs a bunch of care or we'll start "leaking" procs.

Tyriar (changes requested):

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/runInTerminalTool.ts
Tyriar: Note that this is taking into account the direction I'm thinking wrt restoring hidden terminals instead of forcing them on.

meganrogge:
I'm hoping that by properly saving and restoring hide from user terminals, this gets fixed

PR #271982, opened 2025-10-17
Initial support for opening custom resource from sessions list

mjbvz:


src/vs/workbench/contrib/chat/browser/chatSessions/view/sessionsViewPane.ts
```diff
@@ -204,9 +204,9 @@ export class SessionsViewPane extends ViewPane {
 	}
 
 	/**
-	 * Updates the empty state message based on current tree data.
-	 * Uses the tree's existing data to avoid redundant provider calls.
-	 */
+		* Updates the empty state message based on current tree data.
```
mjbvz: Oh man  ts native formatter  went a bit crazy  here

Tyriar (approved):

PR #271983, opened 2025-10-17
persist `forcePersist` background terminals

### LINKED ISSUE #271964: Restored hideFromUser sessions are not disposed of when the session is cleared

meganrogge:
fixes #271964

### DIFF
--- a/src/vs/workbench/contrib/terminal/browser/terminalService.ts
+++ b/src/vs/workbench/contrib/terminal/browser/terminalService.ts
@@ -508,7 +508,7 @@ export class TerminalService extends Disposable implements ITerminalService {
 
 	private async _reviveBackgroundTerminalInstances(backgroundLayouts: (IPtyHostAttachTarget | null)[]): Promise<ITerminalInstance[]> {
 		const instances: ITerminalInstance[] = [];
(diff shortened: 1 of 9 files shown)


src/vs/workbench/contrib/terminal/browser/terminalService.ts
```diff
@@ -505,6 +506,19 @@ export class TerminalService extends Disposable implements ITerminalService {
 		return Promise.all(groupPromises).then(result => result.filter(e => !!e) as ITerminalGroup[]);
 	}
 
+	private async _reviveBackgroundTerminalInstances(backgroundLayouts: (IPtyHostAttachTarget | null)[]): Promise<ITerminalInstance[]> {
+		const instances: ITerminalInstance[] = [];
+		for (const i in backgroundLayouts) {
```
Tyriar: Should use `for of` instead of `for in`

Tyriar (dismissed):

roblourens (approved):

PR #272013, opened 2025-10-17
Use URIs to identify sessions

mjbvz:
Switches us from id+types to use uris to identify chat sessions. I already added very basic support for this but in this next step I think we need a breaking change so that we can avoid adding more and more hacks and complexity to support this


amunger (dismissed):

bpasero (dismissed):

joshspicer (changes requested):
The mechanism we use to modify the UI based on whether a chat session is opened needs to be updated

This is returning 'local': 

https://github.com/microsoft/vscode/blob/040f4a99fe6640ce7fe52cce26043da6a2a2b6d2/src/vs/workbench/contrib/chat/browser/chatSessions/common.ts#L51-L72

When checked here:

https://github.com/microsoft/vscode/blob/040f4a99fe6640ce7fe52cce26043da6a2a2b6d2/src/vs/workbench/contrib/chat/browser/chatEditor.ts#L182-L217

causing

https://github.com/microsoft/vscode/blob/040f4a99fe6640ce7fe52cce26043da6a2a2b6d2/src/vs/workbench/contrib/chat/browser/chatEditor.ts#L207-L208

to not run

joshspicer (approved):

Yoyokrazy (approved):

PR #272018, opened 2025-10-17
Generate policy data as JSON

pwang347:
## Follow-ups
1. Localization: introduce utility to make creation less verbose
2. Terminal configuration changes: remove hacks
3. Refactor script into smaller pieces in the folder and add tests
4. Linux support

## Behaviour Changes
- All categories and policies are now sorted (policies sorted by category first)
- ChatMCP and TelemetryLevel now use enums instead of string


src/vs/workbench/contrib/policyExport/test/node/policyExport.integrationTest.ts
```diff
@@ -0,0 +1,84 @@
...
+					stderr += data.toString();
+				});
+
+				proc.on('close', (code: number | null) => {
+					if (code === 0) {
+						resolve();
+					} else {
+						reject(new Error(`VS Code exited with code ${code}\nStdout: ${stdout}\nStderr: ${stderr}`));
+					}
+				});
```
joaomoreno: Could we use `util.promisify(require('node:child_process').exec)` instead?

src/vs/workbench/contrib/policyExport/electron-browser/policyExport.contribution.ts
```diff
@@ -0,0 +1,117 @@
...
+
+		// Skip for non-development flows
+		if (this.nativeEnvironmentService.isBuilt) {
+			return;
+		}
+
+		const policyDataPath = this.nativeEnvironmentService.exportPolicyData;
+		if (policyDataPath !== undefined) {
+			const defaultPath = join(this.nativeEnvironmentService.appRoot, PolicyExportContribution.DEFAULT_POLICY_EXPORT_PATH);
+			void this.exportPolicyDataAndQuit(policyDataPath ? policyDataPath : defaultPath);
```
joaomoreno, line 46: `void `? I'm guessing it's because of that function being called is `async`? Does the compile barf without this? Or was it just AI being overly cautious?

build/lib/policies/copyPolicyDto.js
joaomoreno, line 1: I see there's a `@ts-check` here... but let's just use TypeScript.

build/package.json
```diff
@@ -64,9 +64,12 @@
   },
   "type": "commonjs",
   "scripts": {
-    "compile": "cd .. && npx tsgo --project build/tsconfig.build.json",
-    "watch": "cd .. && npx tsgo --project build/tsconfig.build.json --watch",
-    "npmCheckJs": "cd .. && npx tsgo --project build/tsconfig.build.json --noEmit"
+    "copy-policy-dto": "node lib/policies/copyPolicyDto.js",
+    "prebuild-ts": "npm run copy-policy-dto",
```
joaomoreno, line 68: Why is this needed? 😕

src/vs/workbench/contrib/policyExport/test/node/policyExport.integrationTest.ts
pwang347: yes! I can look into cleaning this part up

src/vs/workbench/contrib/policyExport/electron-browser/policyExport.contribution.ts
pwang347: this was mostly just me doing it out of habit actually, since my previous project had a lint rule for this 😆  I do think it helps visually see that we're calling async without waiting for it though

joaomoreno: I mean... I don't **_not_** like it. 😄

build/lib/policies/copyPolicyDto.js
pwang347: discussed offline, we can just add in both the .ts and .js file! will make the change

build/package.json
pwang347: discussed offline, this is needed to make sure we have shared types between build/core

src/vs/workbench/contrib/chat/browser/chat.contribution.ts
```diff
@@ -240,18 +241,21 @@ configurationRegistry.registerConfiguration({
 		},
 		[ChatConfiguration.GlobalAutoApprove]: {
 			default: false,
-			// HACK: Description duplicated for policy parser. See https://github.com/microsoft/vscode/issues/254526
-			description: nls.localize('autoApprove2.description',
-				'Global auto approve also known as "YOLO mode" disables manual approval completely for all tools in all workspaces, allowing the agent to act fully autonomously. This is extremely dangerous and is *never* recommended, even containerized environments like Codespaces and Dev Containers have user keys forwarded into the container that could be compromised.\n\nThis feature disables critical security protections and makes it much easier for an attacker to compromise the machine.'
-			),
 			markdownDescription: globalAutoApproveDescription.value,
```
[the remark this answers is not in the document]
pwang347: this isn't relevant because they were never using the same string

src/vs/workbench/contrib/policyExport/common/policyDto.ts
```diff
@@ -0,0 +1,33 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+export type LocalizedValueDto = {
```
bpasero, line 6: Any reason we cannot just have these interfaces and types in `policy.ts`? I would keep the contributions to `vs/base/common` isolated in 1 file since its all going to be top level together with a lot of other things.

Besides, if the use is only in upstream components, why not move it into `vs/platform` or even `workbench`?

pwang347: The reason it's split into separate files is because this types file needs to be shared between VS Code source and a script in the `build` folder - maybe I can move this new file to workbench instead where it's being used?

pwang347: updated - let me know what you think!

bpasero: Yes thanks!

joaomoreno (approved):

PR #272055, opened 2025-10-18
270403: Handle windows UNC paths in markdown doc

Parasaran-Python:
fix #270403

### DIFF
--- a/src/vs/editor/contrib/links/browser/links.ts
+++ b/src/vs/editor/contrib/links/browser/links.ts
@@ -6,6 +6,7 @@
 import { createCancelablePromise, CancelablePromise, RunOnceScheduler } from '../../../../base/common/async.js';
 import { CancellationToken } from '../../../../base/common/cancellation.js';
 import { onUnexpectedError } from '../../../../base/common/errors.js';
+import { isUNC } from '../../../../base/common/extpath.js';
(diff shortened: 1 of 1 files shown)


Parasaran-Python:
The fix is inspired from how terminal link opener works in VSCode

https://github.com/microsoft/vscode/blob/cdbfba6dbf8be50184553ed8e6c8fd4e25c74051/src/vs/workbench/contrib/terminalContrib/links/browser/terminalLinkOpeners.ts#L101-L108

https://github.com/microsoft/vscode/blob/cdbfba6dbf8be50184553ed8e6c8fd4e25c74051/src/vs/workbench/contrib/terminalContrib/links/browser/terminalLinkOpeners.ts#L185-L195

Parasaran-Python:
@mjbvz could you please review the change 🙂

bpasero (commented):
We have a `isUNC` utility method in `extpath`.

Parasaran-Python:
> Having a uri like this `file:///\\\\` seems already fishy. How/where do you create these? Are you using `Uri.file`?

In markdown files when a file path is used and a ctrl click is done on it vscode tries to open it, during the process the URI class of vscode is used, relative and absolute paths work fine, but when an UNC path is used, the URI class interprets it this way

PR #272060, opened 2025-10-18
Move auto approval messages into hover

### LINKED ISSUE #271965: Move auto approval into tool status indicator
Feedback I've seen a bunch is that: Auto approve info is important, but not important enough to constantly spam chat with. We should just add it to the tool status hover or something like that. For terminal specifically, this will need some mechanism of communicating back the auto approval status with details since it has rich rendering and links to action auto approval.

Tyriar:
Fixes #271965
Related #264370 - we want terminal and other tools in the future to have special handling as it does, it's co-located near the general rendering now though

image
image

@connor4312 this loses it for MCP, I'll need your advice on how to connect those pieces. Not sure why the tool progress is on the right on MCP but this seems like a good time to move it to the left and outside the box to make things consistent unless I'm missing something?


src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatToolInvocationPart.ts
```diff
@@ -94,8 +95,8 @@ export class ChatToolInvocationPart extends Disposable implements IChatContentPa
 			partStore.add(this.subPart.onDidChangeHeight(() => this._onDidChangeHeight.fire()));
 			partStore.add(this.subPart.onNeedsRerender(render));
 
-			// todo@connor4312/tyriar: standardize how these are displayed
-			if (!(this.subPart instanceof ChatTerminalToolProgressPart)) {
+			// todo@connor4312: Move MCP spinner to left to get consistent auto approval presentation
+			if (this.subPart instanceof ChatInputOutputMarkdownProgressPart) {
```
Tyriar, line 98: @connor4312 I ended up giving up on moving progress out of MCP and just showing the old experience for now to unblock this PR:

Image

I think this is the right condition. The problems I was hitting were:

- I wasn't sure the impacts of adding a progress to the left of `ChatInputOutputMarkdownProgressPart`, not sure if that's just MCP that uses that or not
- Not sure how to get `toolInvocation` inside `ChatMcpServersInteractionContentPart`
- General concern about breaking stuff I don't know about

I'll keep https://github.com/microsoft/vscode/issues/264370 open to follow that up

src/vs/workbench/contrib/chat/browser/chatContentParts/chatTaskContentPart.ts
```diff
@@ -41,7 +41,7 @@ export class ChatTaskContentPart extends Disposable implements IChatContentPart
 				true;
 			this.isSettled = isSettled;
 			const showSpinner = !isSettled && !context.element.isComplete;
-			const progressPart = this._register(instantiationService.createInstance(ChatProgressContentPart, task, chatContentMarkdownRenderer, context, showSpinner, true, undefined));
+			const progressPart = this._register(instantiationService.createInstance(ChatProgressContentPart, task, chatContentMarkdownRenderer, context, showSpinner, true, undefined, undefined));
```
Tyriar, line 44: @connor4312 I was expecting undefined here to fail for tasks, but it works 🤷 

Image

lramos15 (approved):

PR #272170, opened 2025-10-19
create fork of only the default branch

JoeyShapiro:
This is a feature for the "create fork" option. This happens when you try to publish a branch to a repo you don't have permission to. A popup will be supplied asking if you instead want to fork the repo and publish your new branch there.

I am not sure how most people work, but when I fork a repo, I only fork the default branch. I don't need all of the other branches. So this PR adds that as an additional option. It can easily plug into the existing `octokit` package by adding a single option. Below is a video with my requested change.

https://github.com/user-attachments/assets/29ad8be8-f2a7-4172-b3c6-003d1f8c2980

I am not sure if this makes the menu too cluttered. Maybe it's fine, maybe this should be a setting, or even a second menu after the first. Although I think the second menu idea would be too much complexity. Also, maybe the "no" option should be removed, as I don't think its any different from the "cancel" option already there.


JoeyShapiro:
has anyone had a chance to look this over yet

Issue #270852, opened 2025-10-11
Copilot CLI integration


rebornix:
This month we will work on shipping the Copilot CLI out of the box, deeply integrates it with the Chat Editor and Sessions view, and enables smooth transition between CLI sessions, UI sessions, and cloud (Padawan) sessions. Users should be able to start, manage, and reopen sessions in either interface, see them unified in the Background Agents view, and fluidly delegate or continue work across environments.

### Basic user experience

- [x] Browse and Resume
    - [x] Users can browse all their CLI sessions in the Sessions view
    - [x] Users can open/resume CLI sessions in the Chat Editor
- [ ] Terminal Integration
    - [x] Users can launch the Copilot CLI in an integrated terminal from the Chat Sessions View. The terminal should open in REPL mode. VS Code help setup `copilot` if users don't have it installed.
    - [x] Users can launch the Copilot CLI in an integrated terminal by typing `copilot`.
    - [x] 🏃Users can resume CLI session in the integrated terminal from the Chat Sessions View.
    - [ ] Manually initiated copilot sessions in terminal are tracked/managed by the chat sessions view
- [x] Users can interact with the agent
    - [x] 🏃Show basic tool messages in the view
    - [x] 🏃Support tool confirmation messages (edit, mcp, terminal commands)
- [x] Support picking language models
- [ ] Track edits

### Advanced

- [x] Support picking models
- [x] Push to Coding Agent (Cloud)
- [ ] 🏃🏾 Context attachments @DonJayamanne 
- [ ] MCP server support
  - [ ] 🏃 GitHub MCP first
- [ ] Image support (w/ proper shim for sharp)
- [ ] Support multi-root via `additionalDirectories`
- [ ] Custom instructions support (double check might just work)
- [ ] 💪 Slash commands
- [ ] 💪 Git worktree support
- [ ] 💪 Support custom `/` commands

bamurtaugh:
We'll have new issues for November!

PR #272195, opened 2025-10-20
Support slash commands for silent agent selection (such as chat editors created for a specific agent)

DonJayamanne:
This will allow users to see the list of commands in the chat input.
On the extension side we can add this back to the prompt if required (e.g. the cli tools.)

For #270852

### DIFF
--- a/src/vs/workbench/contrib/chat/common/chatRequestParser.ts
+++ b/src/vs/workbench/contrib/chat/common/chatRequestParser.ts
@@ -206,8 +206,9 @@ export class ChatRequestParser {
 		const slashRange = new OffsetRange(offset, offset + full.length);
 		const slashEditorRange = new Range(position.lineNumber, position.column, position.lineNumber, position.column + full.length);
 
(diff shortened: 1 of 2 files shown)


DonJayamanne:
@roblourens 

For Copilot CLI we want to add support for slash commands.
However I noticed that in you don't add slash commands into package.json in other clis.
This means users won't see the available slash commands when typing `/`.

* My plan is to support this (get completions for slash commands)
    On extension side we can add this back into the prompt if we want (before we send the prompt onto LLM)
* This also gives us the ability to handle slash commands our selves


I can make the changes in copilot chat extension to adopt these changes .

roblourens:
The claude code slash commands are in package.json, they are here https://github.com/microsoft/vscode-copilot-chat/blob/c077cda2e4097fad15dca36546a7500dc3d67e69/package.json#L4103-L4139

Sorry, I don't understand your question

DonJayamanne:
> The claude code slash commands are in package.json, they are he

Thats weird, it didn't work at all for me.
Apologies, i thought you added it but didn't bother due to some other reason.

PR #272228, opened 2025-10-20
Support GitHub-style alert syntax in MarkdownStrings

kycutler:
Related: #209652


src/vs/base/browser/markdownRenderer.ts
```diff
@@ -115,6 +115,32 @@ const defaultMarkedRenderers = Object.freeze({
...
+	const alertIcons: Record<string, string> = {
+		'NOTE': 'info',
+		'TIP': 'light-bulb',
+		'IMPORTANT': 'comment',
+		'WARNING': 'alert',
+		'CAUTION': 'stop'
+	};
+
+	// Match blockquotes that start with [!TYPE]
+	const alertRegex = /<blockquote>\s*<p>\[!(NOTE|TIP|IMPORTANT|WARNING|CAUTION)\](\s*(?:<br\/?>|[\r\n]))/gi;
```
alexr00: I just learned about named capturing groups a year or so ago: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Regular_expressions/Named_capturing_group

src/vs/workbench/contrib/markdown/browser/markdownAlertStyles.ts
```diff
@@ -0,0 +1,60 @@
...
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { registerThemingParticipant } from '../../../../platform/theme/common/themeService.js';
+
+/**
+ * Registers global alert styles for markdown rendering across the workbench.
+ * These styles apply to all markdown content rendered via IMarkdownRendererService.
+ */
+registerThemingParticipant((theme, collector) => {
```
alexr00: Can this just be done with CSS? IIRC, the theming participant should only be used when something needs to be calculated. With CSS variables we mostly don't need it anymore. I think style sheet is included everywhere: https://github.com/microsoft/vscode/blob/15026b09c7355e2ca9f0ab6572c326edc58a4117/src/vs/workbench/browser/media/style.css

src/vs/base/browser/markdownRenderer.ts
```diff
@@ -115,6 +115,32 @@ const defaultMarkedRenderers = Object.freeze({
...
+
+	return html.replace(alertRegex, (_, alertType: string, separator: string) => {
+		// Create the alert structure with a heading, keeping the rest of the content intact
+		const alertTypeCapitalized = alertType.charAt(0).toUpperCase() + alertType.slice(1).toLowerCase();
+		const iconHtml = renderIcon({ id: alertIcons[alertType] }).outerHTML;
+		const severity = alertType.toLowerCase();
+		return `<blockquote data-severity="${severity}"><p><span title="">${iconHtml}${alertTypeCapitalized}</span>${separator}`;
```
alexr00: Why an empty `title`?

src/vs/workbench/contrib/comments/browser/commentThreadWidget.ts
```diff
@@ -409,11 +409,6 @@ export class CommentThreadWidget<T extends IRange | ICellRange = IRange> extends
 			content.push(`.review-widget .body .review-comment blockquote { background: ${blockQuoteBackground}; }`);
 		}
 
-		const blockQuoteBOrder = theme.getColor(textBlockQuoteBorder);
```
alexr00, line 412: I forgot all this css was here. I should move it into a css file. Opened https://github.com/microsoft/vscode/issues/272240 to track moving it.

alexr00 (changes requested):
Nice! My only real comment is on the CSS.

src/vs/workbench/contrib/markdown/browser/markdownAlertStyles.ts
kycutler: It surely can -- just not sure where a good place to put such CSS would be. Adding it directly to the main stylesheet feels a bit wrong so I opted to inject it dynamically.

src/vs/base/browser/markdownRenderer.ts
kycutler: To help identify it for styling. Definitely a bit underhanded but I'm trying not to add additional cases to the sanitizer. The other option would be to just always grab `blockquote[data-severity] > p > :first-child`, but I figure the `title` helps make it clearer.

src/vs/workbench/contrib/markdown/browser/markdownAlertStyles.ts
alexr00: That's fair. Worst case, you can put it in https://github.com/microsoft/vscode/blob/15026b09c7355e2ca9f0ab6572c326edc58a4117/src/vs/workbench/contrib/comments/browser/media/review.css, but using straight css is preferable to using the theming participant.

alexr00: If you can, a markdown.css file next to review.css, which review.css imports, would be better.

src/vs/base/browser/markdownRenderer.ts
alexr00: The selector you have already uses `data-severity`, which seems unique enough. The empty title feels like something we'll accidentally break in the future.

kycutler: Fair point -- removed

src/vs/workbench/contrib/markdown/browser/markdownAlertStyles.ts
kycutler: Actually it looks like I could just create a new one and import it. I want to keep it central and not specific to reviews. Updated, thanks!

src/vs/base/browser/markdownRenderer.ts
kycutler: Good suggestion -- though it turns out that `String.replace()` takes the groups as its last argument, so we would just have even more ignored parameters in the implementation below. So I think it's still cleaner this way.

src/vscode-dts/vscode.proposed.markdownAlerts.d.ts
```diff
@@ -0,0 +1,29 @@
...
+		 * > [!NOTE]
+		 * > This is a note alert
+		 *
+		 * > [!WARNING]
+		 * > This is a warning alert
+		 * ```
+		 *
+		 * Supported alert types: `NOTE`, `TIP`, `IMPORTANT`, `WARNING`, `CAUTION`.
+		 */
+		supportAlerts?: boolean;
```
kycutler: supportAlertSyntax

bpasero:
fyi @mjbvz

src/vs/base/common/htmlContent.ts
```diff
@@ -19,6 +19,8 @@ export interface IMarkdownString {
 	readonly isTrusted?: boolean | MarkdownStringTrustedOptions;
 	readonly supportThemeIcons?: boolean;
 	readonly supportHtml?: boolean;
+	/** @internal */
+	readonly supportAlertSyntax?: boolean;
```
alexr00, line 23: I don't think we really use `@internal`, but @mjbvz can decide if he wants this or not.

alexr00: Actually, what's the `@internal` for?

alexr00 (dismissed):
LGTM, but we also need @mjbvz's review for markdown renderer changes.

src/vs/base/common/htmlContent.ts
kycutler: It prevents this from being exposed in the Monaco API. Since the styles are for workbench and I figure it's better for this to be supported explicitly within hover etc. before being enabled there.

alexr00 (dismissed):

src/vs/base/browser/markdownRenderer.ts
mjbvz: I think we should try using a `marked` extension for rendering this instead. Here's an example one from npm: https://github.com/bent10/marked-extensions/blob/main/packages/alert/src/index.ts

Should be a little more reliable for handling this

src/vs/workbench/contrib/markdown/browser/media/markdown.css
```diff
@@ -0,0 +1,38 @@
...
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+/* Alert color mappings using theme variables */
+blockquote[data-severity="note"] {
+	--vscode-textBlockQuote-border: var(--vscode-editorInfo-foreground);
+}
+
+blockquote[data-severity="tip"] {
+	--vscode-textBlockQuote-border: var(--vscode-charts-green);
```
mjbvz: A few of these hardcode specific colors while others use more general colors item, such as `--vscode-editorWarning-foreground`. Could all of them use these more broad color names? Maybe we need to introduce some new colors?

src/vs/base/browser/markdownRenderer.ts
kycutler: Done!

src/vs/workbench/contrib/markdown/browser/media/markdown.css
```diff
@@ -0,0 +1,38 @@
...
+/* Alert title styling */
+blockquote[data-severity] > p > :first-child {
+	display: inline-flex;
+	align-items: center;
+	color: var(--vscode-textBlockQuote-border);
+	font-weight: bolder;
+}
+
+blockquote[data-severity] > p > :first-child .codicon {
+	color: var(--vscode-textBlockQuote-border);
```
mjbvz, line 36: Based on how these colors are used, I think we should not use border in the new color names. Maybe: `--vscode-markdownAlert-foreground`?

mjbvz: Oh actually I see that these do change the border color too which makes this more complicated. I still think that using the word `border` is confusing and think `markdownAlert.foreground` is clearer, but it's also not perfect

kycutler: Makes sense! Updated

mjbvz (approved):
Looks great! Thanks for making the changes

PR #272265, opened 2025-10-20
feat(monaco): Add option to remove Paste option from editor context menu

jimthedev:
Hello. 

This PR mitigates [this monaco issue](https://github.com/microsoft/monaco-editor/issues/5068). In the linked issue Paste does not work in the context menu in modern browsers (except Firefox which is already excluded). By allowing someone to disable the Paste context menu item we will allow them to ship a custom Paste option that works instead of relying on the built in one that uses the deprecated `execCommand('paste')`. By exposing this first we allow them to remove Paste for now without breaking any existing apis or workflows. There can be a future PR to update the existing functionality to not use the deprecated commands (which would be a higher risk change).

All you need to do to test this is to go into your User Settings as JSON and set it to:

```
{
    "editor.removePasteFromEditorContextMenu": true
}
```

Then go back to the main editor window, right click and notice that the Paste context menu item is no longer there.

https://github.com/user-attachments/assets/46ddf3a4-7868-458a-91be-d54131bf6ae7


jimthedev:

jimthedev:

qqaimh:
Please quick

aiday-mar:
Thank you for this PR. I don't think we should be removing the paste command from the context menu, rather we should focus on fixing the paste command not working from the context menu. I will close this PR and add the other issue to my priority list of issues to fix.

PR #272294, opened 2025-10-20
Enhance RunInTerminalTool to conditionally reveal terminal based on v…

Mateus-Reis:
# Fix terminal auto-opening when AI suggests code via external CLI tools

  ## Problem

  When using AI assistants through external CLI tools, the VS Code integrated terminal would unexpectedly reveal itself whenever the AI suggested code. This happened because the
  `_handleTerminalVisibility()` method in `RunInTerminalTool` would always call `revealTerminal()` when `chat.tools.terminal.outputLocation` was set to `'terminal'`, regardless of whether the terminal was
  already visible or not.

  This created a poor user experience for users who:
  - Use external terminals (Warp, iTerm, etc.) with AI tools
  - Keep VS Code's integrated terminal closed
  - Don't want the terminal panel to auto-open unexpectedly

  ## Solution

  Added a visibility check before revealing the terminal. The terminal panel is now only revealed if it's already visible when `chat.tools.terminal.outputLocation` is set to `'terminal'`.

  ### Changes
  - Added `IViewsService` dependency to check terminal panel visibility
  - Modified `_handleTerminalVisibility()` to call `isViewVisible(TERMINAL_VIEW_ID)` before revealing
  - Added unit tests to validate the new behavior

  ## Behavior

  - **Before**: Terminal always opens when AI suggests code (if `outputLocation` = `'terminal'`)
  - **After**: Terminal only reveals if already open (if `outputLocation` = `'terminal'`)
  - **Unchanged**: When `outputLocation` = `'none'`, terminal never opens (existing behavior preserved)

  ## Testing

  1. Set `chat.tools.terminal.outputLocation` to `'terminal'`
  2. Close the integrated terminal panel
  3. Use an AI tool via external CLI that triggers the run-in-terminal tool
  4. Verify terminal does NOT auto-open
  5. Open the terminal manually
  6. Trigger AI code suggestion again
  7. Verify terminal is properly revealed (existing behavior)

  ## Notes

  This is a conservative fix that:
  - Maintains 100% backward compatibility
  - Preserves all existing behavior when terminal is already visible
  - Only changes behavior when terminal is explicitly closed by the user
  - Does not require any configuration changes
  
image

### DIFF
--- a/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/runInTerminalTool.ts
+++ b/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/runInTerminalTool.ts
@@ -31,7 +31,8 @@ import { ChatConfiguration } from '../../../../chat/common/constants.js';
 import { CountTokensCallback, ILanguageModelToolsService, IPreparedToolInvocation, IToolData, IToolImpl, IToolInvocation, IToolInvocationPreparationContext, IToolResult, ToolDataSource, ToolInvocationPresentation, ToolProgress, type IToolConfirmationMessages, type ToolConfirmationAction } from '../../../../chat/common/languageModelToolsService.js';
 import { ITerminalService, type ITerminalInstance, ITerminalChatService } from '../../../../terminal/browser/terminal.js';
 import type { XtermTerminal } from '../../../../terminal/browser/xterm/xtermTerminal.js';
-import { ITerminalProfileResolverService } from '../../../../terminal/common/terminal.js';
+import { ITerminalProfileResolverService, TERMINAL_VIEW_ID } from '../../../../terminal/common/terminal.js';
+import { IViewsService } from '../../../../../services/views/common/viewsService.js';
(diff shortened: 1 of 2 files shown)


Mateus-Reis:
> Hey, thanks for the PR. I don't see the point, thus closing this. When the terminal ouput location is `terminal`, we want to reveal it.

Hey! I think there might be a misunderstanding about the use case.

  The issue happens when using AI tools via **external CLI** (like Claude Code in Warp terminal). In this scenario:

  1. User is working with an external terminal (Warp, iTerm, etc.)
  2. VS Code's integrated terminal is closed (user doesn't use it)
  3. AI tool suggests code via external CLI
  4. VS Code terminal opens automatically and empty (even though user isn't using it)

  This creates a poor (and really annoying) UX because the terminal opens unexpectedly when the user has explicitly closed it and is using a different terminal.

  **Current behavior**: Terminal always opens when AI suggests code (even if closed)
  **Expected behavior**: Terminal should only reveal if already open (respecting user's choice to keep it closed)

  The fix preserves the existing behavior when terminal IS open, while preventing the unexpected auto-open when it's closed.

  Would you reconsider? This affects users who use external terminals with AI CLI tools.

PR #272324, opened 2025-10-20
Update terminal config policies

pwang347:
Reverts the hack added in https://github.com/microsoft/vscode/pull/262226


pwang347:
@Tyriar let me know if this change is what you had in mind! I tried removing the `TerminalContribSettingId` enum, but noticed that there's a layering issue that happens here if I do: https://github.com/microsoft/vscode/blob/88a26c105cd5ec0c86322af7ebae9515cbb5c751/src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolConfirmationSubPart.ts#L108

src/vs/workbench/contrib/terminalContrib/chatAgentTools/common/terminalChatAgentToolsConfiguration.ts
```diff
@@ -61,6 +62,22 @@ const terminalChatAgentProfileSchema: IJSONSchema = {
...
+	[TerminalChatAgentToolsSettingId.EnableAutoApprove]: {
+		description: localize('autoApproveMode.description', "Controls whether to allow auto approval in the run in terminal tool."),
+		type: 'boolean',
+		default: true,
+		policy: {
+			name: 'ChatToolsTerminalEnableAutoApprove',
+			category: PolicyCategory.IntegratedTerminal,
+			minimumVersion: '1.104',
+			localization: {
+				description: {
```
pwang347, line 75: Note: I discussed with @TylerLeonhardt about whether we can introduce a way to reduce the localization work here (e.g., with a `localize3` that avoids writing the loc key twice). Preference for now is to require some duplication, but enforcing correctness via linting: https://github.com/microsoft/vscode/pull/272334

src/vs/workbench/contrib/terminal/common/terminalConfiguration.ts
```diff
@@ -638,24 +637,6 @@ export async function registerTerminalConfiguration(getFontSnippets: () => Promi
...
-					minimumVersion: '1.104',
-					localization: {
-						description: {
-							key: 'autoApproveMode.description',
-							value: localize('autoApproveMode.description', "Controls whether to allow auto approval in the run in terminal tool."),
-						}
-					}
-				}
-			}
 		}
```
Tyriar: No need to recreate the object via `...` spread now

src/vs/workbench/contrib/terminalContrib/chatAgentTools/common/terminalChatAgentToolsConfiguration.ts
Tyriar: Not understanding this, can't you just pull it from the outer `description`? Also localization doesn't seem that important for policy stuff to me?

src/vs/workbench/contrib/terminalContrib/chatAgentTools/common/terminalChatAgentToolsConfiguration.ts
```diff
@@ -61,6 +62,22 @@ const terminalChatAgentProfileSchema: IJSONSchema = {
 };
 
 export const terminalChatAgentToolsConfiguration: IStringDictionary<IConfigurationPropertySchema> = {
+	[TerminalChatAgentToolsSettingId.EnableAutoApprove]: {
```
Tyriar, line 66: It also needs to be removed from here now:

https://github.com/microsoft/vscode/blob/main/src/vs/workbench/contrib/terminal/terminalContribExports.ts#L37

Terminal is it's own little land and to help reduce code complexity has `terminalContrib`'s that enforce import rules, mainly `contrib/terminal` or any other workbench contrib cannot import from `terminalContrib`, only the other way around. That way there's a lint time error if a bad import appears, effectively forcing terminalContribs to strictly build upon terminal for simple isolated features.

`TerminalContribSettingId` and `terminalContribExports.ts` in general is the way for `contrib/` to import from `terminalContrib`s in the off chance it's needed. See the HACK comments:

https://github.com/microsoft/vscode/blob/0734942d32769ea1e77e42e056d6f16579e78fd6/src/vs/workbench/contrib/terminal/terminalContribExports.ts#L22-L39

Tyriar (changes requested):

src/vs/workbench/contrib/terminalContrib/chatAgentTools/common/terminalChatAgentToolsConfiguration.ts
pwang347: Hey, yes let me summarize a bit the thought process here.

> Not understanding this, can't you just pull it from the outer `description`?

We can't do this directly because the policy generation script needs localization key to access strings in a JSON downloaded from CDN. This was one of the drawbacks I mentioned for moving from static to runtime flow; localization keys are not available at runtime. So the minimal change here even if we pull from outer description is something like:

```typescript
...
localization: {
  description: 'autoApproveMode.description'
}
...
```
And then the corresponding linting rule we add would probably be something like "make sure this same string appears in a `localize` call somewhere in this file".

I think we can definitely consider the above (I had this initially) if we want the localization block to be smaller at the loss of ability to customize a policy's description. I modified it to be the current more verbose approach hoping that we could get it to be a single `localize3` call. I'm not sure if customizing policy descriptions are super useful though, since the current usage is mostly just to avoid markdown descriptions but we already have the regular description property there. It's also a bit more self-contained here, so the linting is a bit more straightforward. Happy to make the change here if you think this would be better though.

> Also localization doesn't seem that important for policy stuff to me?

I will defer to @joaomoreno to comment on this one, I'm just preserving the original policy generation behaviour.

src/vs/workbench/contrib/terminalContrib/chatAgentTools/common/terminalChatAgentToolsConfiguration.ts
pwang347: so I did try that, but ended up with a lint error when trying to import here: https://github.com/microsoft/vscode/pull/272324#issuecomment-3430056143 unless you meant that I shouldn't be importing `TerminalChatAgentToolsSettingId.EnableAutoApprove`?

pwang347: image

Tyriar: Ah yes, nevermind it needs to stay 😅

src/vs/workbench/contrib/terminalContrib/chatAgentTools/common/terminalChatAgentToolsConfiguration.ts
joaomoreno: We actually create localized policy files:

image

pwang347: Thanks @joaomoreno. @Tyriar just to confirm, are we fine with the current state of the localization as is? Or would you prefer that we move to just a dictionary of keys to be less verbose and reuse what we can?

Tyriar: It's fine, not my area just an observation and I'm not the expert here; the duplication is another code smell like the `markdownDescription` not working problem.

Tyriar (dismissed):

mjbvz (approved):

PR #272355, opened 2025-10-20
Use `toAction` instead of `new Action` to prevent some leaks

mjbvz:


amunger (approved):

bpasero:

PR #272366, opened 2025-10-20
add `id` for `TerminalCommand`, revamp `terminalLastCommand`

### LINKED ISSUE #259261: Allow addressing particular terminals and terminal commands in chat attachments
Currently the terminal contexts `#terminalSelection` and `#terminalLastCommand` work purely off the active (most recent) terminal. Since everything used to live in the extension, this was much more difficult to accomplish as there's no unique addressable API such as `Terminal.resource: URI` that we could have used, and referencing a particular command was basically impossible due to not having access to CommandDetectionCapability in core.

We're moving all this stuff to core (core https://github.com/microsoft/vscode/issues/252650, other tools https://github.com/microsoft/vscode/issues/259260), so this is something we could do.

meganrogge:
fixes #259261

https://github.com/user-attachments/assets/3117748a-4a57-4c23-b8e4-55557b84ea32


src/vs/workbench/contrib/chat/browser/chatContentParts/chatAttachmentsContentPart.ts
```diff
@@ -141,6 +141,8 @@ export class ChatAttachmentsContentPart extends Disposable {
 			widget = this.instantiationService.createInstance(PromptTextAttachmentWidget, attachment, undefined, { shouldFocusClearButton: false, supportsDeletion: false }, container, this._contextResourceLabels);
 		} else if (resource && (attachment.kind === 'file' || attachment.kind === 'directory')) {
 			widget = this.instantiationService.createInstance(FileAttachmentWidget, resource, range, attachment, correspondingContentReference, undefined, { shouldFocusClearButton: false, supportsDeletion: false }, container, this._contextResourceLabels);
+		} else if (isTerminalCommandVariableEntry(attachment)) {
+			widget = this.instantiationService.createInstance(TerminalCommandAttachmentWidget, attachment, undefined, { shouldFocusClearButton: false, supportsDeletion: false }, container, this._contextResourceLabels);
```
roblourens, line 145: So this is rendered differently when you attach the terminal tool? Or what is a `ITerminalCommandVariableEntry`?

src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatResultListSubPart.ts
```diff
@@ -27,17 +32,44 @@ export class ChatResultListSubPart extends BaseChatToolInvocationSubPart {
...
+			if (URI.isUri(detail) && detail.scheme === Schemas.vscodeTerminal) {
+				terminalReferences.push(detail);
+			} else {
+				otherReferences.push(detail);
+			}
+		}
+
+		if (terminalReferences.length) {
+			const attachmentEntries = terminalReferences.map(toTerminalCommandVariableEntry);
+			const attachmentsPart = this._register(instantiationService.createInstance(ChatAttachmentsContentPart, {
```
roblourens, line 62: When is there more than one? To me the goal of `ChatResultListSubPart` is to show a collapsible list of tool results but it looks like this isn't a collapsible list? Is there a case where there are terminalReferences and otherReferences?

src/vs/workbench/contrib/chat/common/chatVariableEntries.ts
```diff
@@ -112,6 +112,13 @@ export interface INotebookOutputVariableEntry extends IBaseChatRequestVariableEn
 	readonly mimeType?: string;
 }
 
+export interface ITerminalCommandVariableEntry extends IBaseChatRequestVariableEntry {
+	readonly kind: 'terminalCommand';
+	readonly value: URI;
+	readonly commandId?: string;
+	readonly commandLine?: string;
```
Tyriar, line 119: We should store output here too? Otherwise if the terminal is killed it won't be possible to access?

src/vs/workbench/contrib/chat/common/chatVariableEntries.ts
```diff
@@ -249,6 +257,29 @@ export namespace IChatRequestVariableEntry {
...
 
+export function isTerminalCommandVariableEntry(entry: IChatRequestVariableEntry): entry is ITerminalCommandVariableEntry {
+	return entry.kind === 'terminalCommand';
+}
+
+export function toTerminalCommandVariableEntry(uri: URI): ITerminalCommandVariableEntry {
+	const query = uri.query ?? '';
+	const params = new URLSearchParams(query);
+	const commandId = params.get('command') ?? undefined;
+	const commandLine = params.get('commandLine') ?? undefined;
```
Tyriar: Can we avoid encoding this in the URI? I see `toPromptTextVariableEntry` allows passing content in via an arg

Tyriar (changes requested):

src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatResultListSubPart.ts
meganrogge: There isn't more than one, at least yet. You can imagine a user might eventually want to attach multiple terminal commands though. 

Open to ideas for where to put this instead of here 🤷🏼‍♀️

Tyriar:
More focused feature requests for what I was describing:

https://github.com/microsoft/vscode/issues/272902
https://github.com/microsoft/vscode/issues/272905

PR #272369, opened 2025-10-20
Use unified timeout setting under shell integration in runCommand 

### LINKED ISSUE #270813: Use timeout setting under `terminal.integrated.shellIntegration.` in runCommand
@rebornix saw the behavior, when calling `terminal.shellIntegration`'s `executeCommand` duplicate commands can show.

Video of repro:
https://github.com/user-attachments/assets/a5c5abe8-6328-4344-9697-bbb3801cde4b

However, this was avoidable when there was a explicit timeout of 5 seconds before the call to `executeCommand`.

The duplicate command showing up behavior is the same behavior we see when user first opens the terminal, and tries to type things before the prompt even shows up.
   - In this case, the expected behavior is that we actually want these commands to show up before the prompt does, so the user has feedback of what they have just typed in real-time.

This is the 2s timeout we currently have: https://github.com/microsoft/vscode/blob/6a660317a1cf9f202b0f48a7a9189930b9bae39b/src/vs/workbench/contrib/terminal/browser/terminalInstance.ts#L924-L934

And here is what we do in chat terminal: https://github.com/microsoft/vscode/blob/6a660317a1cf9f202b0f48a7a9189930b9bae39b/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/toolTerminalCreator.ts#L75-L82

We could move this setting under `terminal.integrated.shellIntegration` and wait for however long time `defined` via setting, when `executeCommand` is called right after startup. 

From @rebornix:
"""
Maybe we just completely differentiate between user typing to run command vs. `executeCommand` called from extension, and wait however long to prevent the dup command to show when `executeCommand` is called from extension.
"""

My concern for this one here is that things may feel slower to user in a sense the prompt and command will only show when (shell integration&its executeCommand and the second one from in the dup command) is ready.
Without upper bound defined, I'm worried something will go wrong in between shell integration being available and executeCommand being called, and user having to wait forever.

/cc @Tyriar

anthonykim1:
Resolves: https://github.com/microsoft/vscode/issues/270813 

Main points from: https://github.com/microsoft/vscode/pull/272369#issue-3534265187

* Move the chat shell integration timeout setting under move it under terminal.integrated.shellIntegration.
* Use that setting in runCommand
* Record the time the terminal process was ready and only apply that timeout however long after that time occurs. If the timeout is 5 seconds and the terminal has been ready for 6 seconds, there's no need to wait at all.


src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/toolTerminalCreator.ts
```diff
@@ -72,7 +72,12 @@ export class ToolTerminalCreator {
 		const siInjectionEnabled = this._configurationService.getValue(TerminalSettingId.ShellIntegrationEnabled);
 
 		// Get the configurable timeout to wait for shell integration
-		const configuredTimeout = this._configurationService.getValue(TerminalChatAgentToolsSettingId.ShellIntegrationTimeout) as number | undefined;
+		// Try the new unified setting first, fall back to the deprecated chat-specific setting?
+		// TODO: Maybe we should just entirely get rid of the chat setting.
+		let configuredTimeout = this._configurationService.getValue(TerminalSettingId.ShellIntegrationTimeout) as number | undefined;
+		if (configuredTimeout === undefined || typeof configuredTimeout !== 'number') {
+			configuredTimeout = this._configurationService.getValue(TerminalChatAgentToolsSettingId.ShellIntegrationTimeout) as number | undefined;
```
anthonykim1: Am I safe to just completely get rid of the chat timeout setting?

src/vs/workbench/contrib/terminalContrib/chatAgentTools/common/terminalChatAgentToolsConfiguration.ts
```diff
@@ -325,11 +325,13 @@ export const terminalChatAgentToolsConfiguration: IStringDictionary<IConfigurati
 		} satisfies Record<string, boolean | { approve: boolean; matchCommandLine?: boolean }>,
 	},
 	[TerminalChatAgentToolsSettingId.ShellIntegrationTimeout]: {
-		markdownDescription: localize('shellIntegrationTimeout.description', "Configures the duration in milliseconds to wait for shell integration to be detected when the run in terminal tool launches a new terminal. Set to `0` to wait the minimum time, the default value `-1` means the wait time is variable based on the value of {0} and whether it's a remote window. A large value can be useful if your shell starts very slowly and a low value if you're intentionally not using shell integration.", `\`#${TerminalSettingId.ShellIntegrationEnabled}#\``),
+		// TODO: We should probably just remove
```
anthonykim1: Probably just remove, but need to get confirmation

src/vs/workbench/contrib/terminal/browser/terminalCommands.ts
```diff
@@ -4,10 +4,37 @@
...
+import { ITerminalGroupService, ITerminalService } from './terminal.js';
+import { registerAction2, Action2 } from '../../../../platform/actions/common/actions.js';
+import { ServicesAccessor } from '../../../../platform/instantiation/common/instantiation.js';
 
 export function setupTerminalCommands(): void {
 	registerOpenTerminalAtIndexCommands();
+	registerTestRunCommandAction();
+}
+
+// TODO: Remove
```
anthonykim1: This is my dummy command for testing. Please ignore for now!

src/vs/workbench/contrib/terminal/browser/terminalInstance.ts
```diff
@@ -917,19 +917,36 @@ export class TerminalInstance extends Disposable implements ITerminalInstance {
...
+
+			const processReadyTime = this._processManager.processReadyTime;
+			if (processReadyTime) {
+				const elapsed = Date.now() - processReadyTime;
+				timeoutMs = Math.max(0, timeoutMs - elapsed);
+			}
+
+			if (timeoutMs > 0) {
+				await timeout(timeoutMs);
+			}
```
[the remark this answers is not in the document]
anthonykim1: Command detection being available alone was not enough to avoid the dup command UI.
This is intentional, we would just want to wait for timeout, but still need listener/handler for commandDetectionCapability to avoid bunch of unnecessary `^C`.

src/vs/workbench/contrib/terminal/browser/terminalInstance.ts
```diff
@@ -917,19 +917,36 @@ export class TerminalInstance extends Disposable implements ITerminalInstance {
 
 	async runCommand(commandLine: string, shouldExecute: boolean): Promise<void> {
 		let commandDetection = this.capabilities.get(TerminalCapability.CommandDetection);
+		const timeOutValue = this._configurationService.getValue(TerminalSettingId.ShellIntegrationTimeout) as number | undefined;
```
Tyriar: timeOut -> timeout

Tyriar: Preferred way of doing the cast (though still unsafe like more of `getValue` uses):

```suggestion
		const timeOutValue = this._configurationService.getValue<number | undefined>(TerminalSettingId.ShellIntegrationTimeout);
```

src/vs/workbench/contrib/terminal/browser/terminalInstance.ts
```diff
@@ -917,19 +917,36 @@ export class TerminalInstance extends Disposable implements ITerminalInstance {
...
+		let timeoutMs: number;
+
+		if (timeOutValue === undefined || typeof timeOutValue !== 'number' || timeOutValue < 0) {
```
Tyriar: You're basically doing full checking of the type here so you may as well use `getValue<unknown>`. Also prefer `isNumber(...)` to checking `typeof` (I know you're copying my code over but it can improve 😉)

src/vs/workbench/contrib/terminal/browser/terminalInstance.ts
```diff
@@ -917,19 +917,36 @@ export class TerminalInstance extends Disposable implements ITerminalInstance {
...
+			const siEnabled = this._configurationService.getValue(TerminalSettingId.ShellIntegrationEnabled) === true;
+			// Discuss: This default from copilot chat setting feels pretty slow..
```
Tyriar: It's slow, but only if things don't work and only for the first command. We need to check the time against when `processReady` happens

src/vs/workbench/contrib/terminal/browser/terminalInstance.ts
```diff
@@ -917,19 +917,36 @@ export class TerminalInstance extends Disposable implements ITerminalInstance {
...
+			timeoutMs = siEnabled ? 5000 : (this.isRemote ? 3000 : 2000);
+		} else if (timeOutValue === 0) {
+			// Discuss: This is supposed to be fastest value.
+			// Past: We used to have Promise race between onDidAddCommandDetectionCapability and timeout of 2000ms.
+			timeoutMs = 500;
+		} else {
+			timeoutMs = Math.max(timeOutValue, 500);
+		}
```
Tyriar: This is equivalent:

```suggestion
		} else {
			// Discuss: This is supposed to be fastest value.
			// Past: We used to have Promise race between onDidAddCommandDetectionCapability and timeout of 2000ms.
			timeoutMs = Math.max(timeOutValue, 500);
		}
```

src/vs/workbench/contrib/terminal/browser/terminalInstance.ts
```diff
@@ -917,19 +917,36 @@ export class TerminalInstance extends Disposable implements ITerminalInstance {
...
+		const timeOutValue = this._configurationService.getValue(TerminalSettingId.ShellIntegrationTimeout) as number | undefined;
+		let timeoutMs: number;
+
+		if (timeOutValue === undefined || typeof timeOutValue !== 'number' || timeOutValue < 0) {
+			const siEnabled = this._configurationService.getValue(TerminalSettingId.ShellIntegrationEnabled) === true;
+			// Discuss: This default from copilot chat setting feels pretty slow..
+			timeoutMs = siEnabled ? 5000 : (this.isRemote ? 3000 : 2000);
+		} else if (timeOutValue === 0) {
+			// Discuss: This is supposed to be fastest value.
+			// Past: We used to have Promise race between onDidAddCommandDetectionCapability and timeout of 2000ms.
```
Tyriar: This is a user configured value so it's fine it's now smaller than 2000.

src/vs/workbench/contrib/terminal/browser/terminalInstance.ts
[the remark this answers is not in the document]
Tyriar: We should use Promise.race here, but the command detection condition is that the capability is ready and also that `PromptInputModel.state == PromptInputState.Input` (event: `onDidStartInput`)

src/vs/workbench/contrib/terminalContrib/chatAgentTools/common/terminalChatAgentToolsConfiguration.ts
Tyriar: There are 2 approaches:

1. Deprecate and remove after x months
2. Set up migration, then remove that after x months

The second option is probably best as the user doesn't have to do anything. For migration on startup we will check the old setting and if it exists set the new setting to that value if it doesn't exist already.

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/toolTerminalCreator.ts
Tyriar: See https://github.com/microsoft/vscode/pull/272369/files#r2446823086

src/vs/workbench/contrib/terminalContrib/chatAgentTools/common/terminalChatAgentToolsConfiguration.ts
```diff
@@ -325,11 +325,13 @@ export const terminalChatAgentToolsConfiguration: IStringDictionary<IConfigurati
...
+		markdownDescription: localize('shellIntegrationTimeout.description.deprecated', "**Deprecated**: Use {0} instead. Configures the duration in milliseconds to wait for shell integration to be detected when the run in terminal tool launches a new terminal.", `\`#${TerminalSettingId.ShellIntegrationTimeout}#\``),
```
Tyriar: For future, no need to mention deprecated in the description since it gets special treatment when `deprecationMessage` is there.

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/toolTerminalCreator.ts
Tyriar: This should share the same code as the use in runCommand.

Tyriar (changes requested):

src/vs/workbench/contrib/terminal/browser/terminalInstance.ts
anthonykim1: This is really useful, thank you :)

Is the using of unknown because its safer or we get to catch the error at this point (rather than in getValue itself) since we are doing the type checking here? 

Also, would it also be better to just remove `timeoutValue === undefined ` case? Since `!isNumber(timeoutValue)` could cover it.

src/vs/workbench/contrib/terminal/browser/terminalInstance.ts
anthonykim1: Is what I have now below good?
I do math to subtract the processReadyTime from now. 

https://github.com/microsoft/vscode/blob/e46d70bb1d56aa48ee95a984ebd758b7cb17da10/src/vs/workbench/contrib/terminal/browser/terminalInstance.ts#L935-L955

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/toolTerminalCreator.ts
```diff
@@ -56,52 +55,74 @@ export class ToolTerminalCreator {
...
+				store.dispose();
+			}
+
 			if (token.isCancellationRequested) {
+				processReadyListener.dispose();
 				instance.dispose();
 				throw new CancellationError();
 			}
 
-			// If SI is rich, wait for the prompt state to change. This prevents an issue with pwsh
```
anthonykim1, line 95: We won't need to wait extra, since we wait above, right?

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/toolTerminalCreator.ts
```diff
@@ -162,65 +183,4 @@ export class ToolTerminalCreator {
...
-	private _waitForShellIntegration(
-		instance: ITerminalInstance,
-		timeoutMs: number
-	): Promise<ShellIntegrationQuality> {
-		const store = new DisposableStore();
-		const result = new DeferredPromise<ShellIntegrationQuality>();
-
-		const siNoneTimer = store.add(new MutableDisposable());
-		siNoneTimer.value = disposableTimeout(() => {
-			this._logService.info(`ToolTerminalCreator#_waitForShellIntegration: Timed out ${timeoutMs}ms, using no SI`);
```
anthonykim1, line 175: There were bunch of logs here, did you want me to try to leave similar logs in the new version?

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/toolTerminalCreator.ts
anthonykim1: Tried to equate what I wrote in `runCommand` into `createTerminal` inside this` toolTerminalCreator.ts.`
Would appreciate another look 🙏

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/toolTerminalCreator.ts
Tyriar, line 1: So many changes in this file is really scary, the flow for creating a terminal is really important to chat and is relatively fragile. I was only expecting the timeoutMs to come from some shared function but otherwise everything else to remain the same.

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/terminal.chatAgentTools.contribution.ts
```diff
@@ -27,7 +31,22 @@ import { CreateAndRunTaskTool, CreateAndRunTaskToolData } from './tools/task/cre
 import { GetTaskOutputTool, GetTaskOutputToolData } from './tools/task/getTaskOutputTool.js';
 import { RunTaskTool, RunTaskToolData } from './tools/task/runTaskTool.js';
 
-// #region Workbench contributions
+class ShellIntegrationTimeoutMigrationContribution extends Disposable implements IWorkbenchContribution {
+	static readonly ID = 'terminal.shellIntegrationTimeoutMigration';
+
+	constructor(
+		@IConfigurationService private readonly _configurationService: IConfigurationService,
```
Tyriar: Should not be `private readonly` or have `_` prefix as it's only used in the ctor.

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/terminal.chatAgentTools.contribution.ts
```diff
@@ -27,7 +31,22 @@ import { CreateAndRunTaskTool, CreateAndRunTaskToolData } from './tools/task/cre
...
+	) {
+		super();
+		const newSettingValue = this._configurationService.getValue<unknown>(TerminalSettingId.ShellIntegrationTimeout);
+		const deprecatedSettingValue = this._configurationService.getValue<unknown>(TerminalChatAgentToolsSettingId.ShellIntegrationTimeout);
+
+		if (!isNumber(newSettingValue) && isNumber(deprecatedSettingValue)) {
+			this._configurationService.updateValue(TerminalSettingId.ShellIntegrationTimeout, deprecatedSettingValue);
+		}
```
Tyriar: Do the minimum work to gate this:

```ts
const deprecatedSettingValue = this._configurationService.getValue<unknown>(TerminalChatAgentToolsSettingId.ShellIntegrationTimeout);
if (isNumber(deprecatedSettingValue)) {
  //...
}
```

That makes it more clear and in the common case saves a couple of function calls.

src/vs/workbench/contrib/terminal/common/terminalConfiguration.ts
```diff
@@ -593,6 +593,14 @@ const terminalConfiguration: IStringDictionary<IConfigurationPropertySchema> = {
 		],
 		default: 'both'
 	},
+	[TerminalSettingId.ShellIntegrationTimeout]: {
+		restricted: true,
+		markdownDescription: localize('terminal.integrated.shellIntegration.timeout', "Configures the duration in milliseconds to wait for shell integration to be detected when the terminal launches or when extension APIs like {0} are called. Set to {1} to wait the minimum time (500ms), the default value {2} means the wait time is variable based on whether shell integration injection is enabled and whether it's a remote window. A large value can be useful if your shell starts very slowly.", '`executeCommand`', '`0`', '`-1`'),
```
Tyriar:
```suggestion
		markdownDescription: localize('terminal.integrated.shellIntegration.timeout', "Configures the duration in milliseconds to wait for shell integration after launch before declaring it's not there. Set to {0} to wait the minimum time (500ms), the default value {1} means the wait time is variable based on whether shell integration injection is enabled and whether it's a remote window. Consider setting this to a small value if you intentionally disabled shell integration, or a large value if your shell starts very slowly.", '`0`', '`-1`'),
```

src/vs/workbench/contrib/terminal/common/terminal.ts
```diff
@@ -277,6 +277,7 @@ export const isTerminalProcessManager = (t: ITerminalProcessInfo | ITerminalProc
 
 export interface ITerminalProcessManager extends IDisposable, ITerminalProcessInfo {
 	readonly processTraits: IProcessReadyEvent | undefined;
+	readonly processReadyTime: number | undefined;
```
Tyriar: Time -> Timestamp

Tyriar (changes requested):

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/terminal.chatAgentTools.contribution.ts
anthonykim1: Nice, yeah.
Something like:
```
if (isNumber(deprecatedSettingValue)) {
			const newSettingValue = configurationService.getValue<unknown>(TerminalSettingId.ShellIntegrationTimeout);
			if (!isNumber(newSettingValue)) {
				configurationService.updateValue(TerminalSettingId.ShellIntegrationTimeout, deprecatedSettingValue);
			}
		}
``` 

Tried to go one more step, and inverted the first if and return ear, what do you think? https://github.com/microsoft/vscode/pull/272369/commits/262be843ebbb9aa4d2f449d6ab20968cddf1df1d

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/toolTerminalCreator.ts
anthonykim1: Made bunch of more changes to take this feedback! lmk what you think

src/vs/workbench/contrib/terminal/browser/terminalInstance.ts
```diff
@@ -2798,3 +2813,27 @@ function guessShellTypeFromExecutable(os: OperatingSystem, executable: string):
...
+	}
+
+	// Adjust timeout based on how long the process has already been running
+	if (processReadyTimestamp !== undefined) {
+		const elapsed = Date.now() - processReadyTimestamp;
+		timeoutMs = Math.max(0, timeoutMs - elapsed);
+	}
+
+	return timeoutMs;
+}
```
Tyriar: terminalEnvironment.ts is a better place to put this as a general helper file.

Tyriar (changes requested):
Red CI

Tyriar (approved):

DonJayamanne (approved):

Issue #272367, opened 2025-10-20
Handoffs show in chat editors for Copilot coding agent tasks in Agent sessions view


pierceboggan:
<img width="1034" height="842" alt="Image" src="https://github.com/user-attachments/assets/e76e3584-7051-44d4-bcfa-63ef5c0f5235" />

PR #272377, opened 2025-10-21
Skip rendering the "Suggest Next" widget when locked to a coding agent

bhavyaus:
Fixes:
https://github.com/microsoft/vscode/issues/272367

### DIFF
--- a/src/vs/workbench/contrib/chat/browser/chatWidget.ts
+++ b/src/vs/workbench/contrib/chat/browser/chatWidget.ts
@@ -1715,6 +1715,12 @@ export class ChatWidget extends Disposable implements IChatWidget {
 	}
 
 	private renderChatSuggestNextWidget(): void {
+		// Skip rendering in coding agent sessions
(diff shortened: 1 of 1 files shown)


src/vs/workbench/contrib/chat/browser/chatWidget.ts
```diff
@@ -1715,6 +1715,12 @@ export class ChatWidget extends Disposable implements IChatWidget {
 	}
 
 	private renderChatSuggestNextWidget(): void {
+		// Skip rendering in coding agent sessions
+		if (ChatContextKeys.lockedToCodingAgent.getValue(this.contextKeyService) === true) {
```
joshspicer: In the chat widget you can check
https://github.com/microsoft/vscode/blob/2c8835790a9ac1819e4ce2e7d71fe40e254e6bd4/src/vs/workbench/contrib/chat/browser/chatWidget.ts#L2512-L2515

Tyriar (approved):

PR #272378, opened 2025-10-21
Allow custom editors to set custom icons 

### LINKED ISSUE #105028: Add ability to apply custom icon to Custom Editors
With webviews, we can add icons we like to the webview with webviewPanel.iconPath but when this is applied to the Custom Text Editor's webview panel, nothing is displayed.

Create a method to apply a custom icon to Custom Text Editors without having to provide an entire Icon Theme. Ideally this would override ONLY the Custom Editor associated to that viewType, which would allow us to differentiate open text editors vs Custom Editors viewing the same file by icon.

mjbvz:
Follow up on #272375 which i'll merge first

Fixes https://github.com/microsoft/vscode/issues/105028

Enables icons for custom editors if set by extensions. Also removes the `WebviewIconManager` we were using to support icons previously in favor of this more standard approach


bpasero (dismissed):

bpasero (dismissed):

Tyriar (approved):

mjbvz:

Issue #251527, opened 2025-06-15
Cannot open files that contains zero width chars


Mingyueyixi:
Type: <b>Bug</b>

vscode can't open file when file's name contains zero width chars.
env:
windows 10
python3.12.1

Setp1
create a fold which name testDemo.
mkdir testDemo

Setp2
run this python3 script to create a c.txt file：

```python
# encoding: utf-8

from pathlib import Path

if __name__ == "__main__":
    Path("\ufeffc.txt").write_text("this is test file", encoding="utf-8")

```

Setp3
open testDemo fold by vscode

Setp4
click c.txt file on vscde , then vscode will show a error dialog like :
![img](https://upload.cc/i1/2025/06/15/E0iOI2.gif)

VS Code version: Code 1.101.0 (dfaf44141ea9deb3b4096f7cd6d24e00c147a4b1, 2025-06-11T15:00:50.123Z)
OS version: Windows_NT x64 10.0.19045
Modes:

<details>
<summary>System Info</summary>

|Item|Value|
|---|---|
|CPUs|Intel(R) Core(TM) i7-10750H CPU @ 2.60GHz (12 x 2592)|
|GPU Status|2d_canvas: enabled<br>canvas_oop_rasterization: enabled_on<br>direct_rendering_display_compositor: disabled_off_ok<br>gpu_compositing: enabled<br>multiple_raster_threads: enabled_on<br>opengl: enabled_on<br>rasterization: enabled<br>raw_draw: disabled_off_ok<br>skia_graphite: disabled_off<br>video_decode: enabled<br>video_encode: enabled<br>vulkan: disabled_off<br>webgl: enabled<br>webgl2: enabled<br>webgpu: enabled<br>webnn: disabled_off|
|Load (avg)|undefined|
|Memory (System)|31.78GB (16.82GB free)|
|Process Argv|testDemo --crash-reporter-id 692a477b-bf45-428a-b0f6-405ec06a6b86|
|Screen Reader|no|
|VM|29%|
</details><details><summary>Extensions (48)</summary>

Extension|Author (truncated)|Version
---|---|---
tongyi-lingma|Ali|2.5.12
vscode-django|bat|1.15.0
better-json5|Blu|1.4.0
dart-code|Dar|3.112.0
flutter|Dar|3.112.0
vscode-eslint|dba|3.0.10
githistory|don|0.6.20
python-extension-pack|don|1.7.0
vscode-html-css|ecm|2.0.13
[...]

PR #272389, opened 2025-10-21
Set TextDecoder.ignoreBOM to true in VSBuffer

busorgin:
Fix for https://github.com/microsoft/vscode/issues/251527

Issue can be replicated with the steps shown in the issue.

DiskFileSystemProvider.readdir() will create a promise which will contain the list of files in our folder, this message is serialized correctly with the BOM character preserved. The issue occurs when ChannelClient.onBuffer() tries to deserialize the message. In VSBuffer.toString() a default instance of TextDecoder is used which has ignoreBOM set to false meaning the BOM character will be stripped.


bpasero (approved):

bpasero:
Great finding! Our original intent with `buffer.ts` was to stay as close as possible to node.js, so this makes sense.

One change I push is to leave the encoding not spelled out explicitly and rely on the default: https://github.com/microsoft/vscode/pull/272389/commits/bef678a0faa2894d7d5d9b2ce3faff9fddc825b9

Tyriar (approved):

PR #272425, opened 2025-10-21
Fix terminal tab prompt input breaking when backticks are included

### LINKED ISSUE #272416: Command text on the Shell Integration details pop-up can escape the code block
I haven't checked the code who renders this part, but I'd believe this is only a visual bug, aside from maybe creating rickroll/phishing links.

For what I could skim from the visual representation, the terminal tab popup's "Shell Integration" section renders the command text by wrapping it with backticks:

```markdown
`echo like this;`
```
However, it is possible to use backticks in a valid command, which breaks this wrapping

```bash
apt install --only-upgrade `grep 'dev' ./apt-dependencies.lock`
```
(the code doesn't matter, using backticks matters)

Since the command has a backtick, this breaks the wrapping:
```markdown
`apt install --only-upgrade `grep 'dev' ./apt-dependencies.lock``
```

Does this issue occur when all extensions are disabled?: Will test later as it's a remote machine right now.

- VS Code Version: 1.105.1 (user setup)
- OS Version: Windows_NT x64 10.0.19045

Steps to Reproduce:

1. Run a long running command with backticks (`\``) from terminal
2. Hover over the terminal icon in the Terminal View.

The popup after hovering over the terminal tab, which has a broken input of the code

3. The text is not escaped, which breaks the backtick.

P.S: I might be able to make a PR in a few days if the issue really happens to be simple.
  Changing the wrapping to a `<pre>` could fix it... 
  and then escaping the &lt; and &gt; with &amp;lt; and &amp;gt; to make sure the command can't break by having html in the command text...
I'm open to comments!

Abrifq:
Closes #272416.
I have tested the changes manually via codespaces.

### DIFF
--- a/src/vs/workbench/contrib/terminal/browser/terminalTooltip.ts
+++ b/src/vs/workbench/contrib/terminal/browser/terminalTooltip.ts
@@ -6,7 +6,7 @@
 import { localize } from '../../../../nls.js';
 import { ITerminalInstance } from './terminal.js';
 import { asArray } from '../../../../base/common/arrays.js';
-import { escapeMarkdownSyntaxTokens, MarkdownString } from '../../../../base/common/htmlContent.js';
+import { MarkdownString } from '../../../../base/common/htmlContent.js';
 import type { IHoverAction } from '../../../../base/browser/ui/hover/hover.js';
 import { TerminalCapability } from '../../../../platform/terminal/common/capabilities/capabilities.js';
 import { TerminalStatus } from './terminalStatusList.js';
@@ -44,7 +44,7 @@ export function getInstanceHoverInfo(instance: ITerminalInstance, storageService
 	});
 
 	const shellProcessString = getShellProcessTooltip(instance, !!showDetailed);
-	const content = new MarkdownString(instance.title + shellProcessString + statusString, { supportThemeIcons: true, supportHtml: true });
+	const content = new MarkdownString(instance.title + shellProcessString + statusString, { supportThemeIcons: true });
 
 	return { content, actions };
 }
@@ -109,11 +109,15 @@ export function refreshShellIntegrationInfoStatus(instance: ITerminalInstance) {
 	}
 	const combinedString = instance.capabilities.get(TerminalCapability.CommandDetection)?.promptInputModel.getCombinedString();
 	if (combinedString !== undefined) {
-		const escapedPromptInput = escapeMarkdownSyntaxTokens(combinedString
-			.replaceAll('<', '&lt;').replaceAll('>', '&gt;') 		 //Prevent escaping from wrapping
-			.replaceAll(/\((.+?)(\|?(?: (?:.+?)?)?)\)/g, '(<$1>$2)') //Escape links as clickable links
-		);
-		detailedAdditions.push(`Prompt input: <code>\n${escapedPromptInput}\n</code>`);
+		if (combinedString.includes('`')) {
+			detailedAdditions.push('Prompt input:' + [
+				'```',
+				combinedString, // No new lines so no need to escape ```
+				'```',
+			].map(e => `\n    ${e}`).join(''));
+		} else {
+			detailedAdditions.push(`Prompt input: \`${combinedString.replaceAll('`', '&#96;')}\``);
+		}
 	}
 	const detailedAdditionsString = detailedAdditions.length > 0
 		? '\n\n' + detailedAdditions.map(e => `- ${e}`).join('\n')
--- a/src/vs/workbench/contrib/terminal/browser/terminalTooltip.ts
+++ b/src/vs/workbench/contrib/terminal/browser/terminalTooltip.ts
@@ -6,7 +6,7 @@
 import { localize } from '../../../../nls.js';
 import { ITerminalInstance } from './terminal.js';
 import { asArray } from '../../../../base/common/arrays.js';
-import { MarkdownString } from '../../../../base/common/htmlContent.js';
+import { escapeMarkdownSyntaxTokens, MarkdownString } from '../../../../base/common/htmlContent.js';
 import type { IHoverAction } from '../../../../base/browser/ui/hover/hover.js';
 import { TerminalCapability } from '../../../../platform/terminal/common/capabilities/capabilities.js';
 import { TerminalStatus } from './terminalStatusList.js';
@@ -109,10 +109,10 @@ export function refreshShellIntegrationInfoStatus(instance: ITerminalInstance) {
 	}
 	const combinedString = instance.capabilities.get(TerminalCapability.CommandDetection)?.promptInputModel.getCombinedString();
 	if (combinedString !== undefined) {
-		const escapedPromptInput = combinedString
+		const escapedPromptInput = escapeMarkdownSyntaxTokens(combinedString
 			.replaceAll('<', '&lt;').replaceAll('>', '&gt;') 		 //Prevent escaping from wrapping
 			.replaceAll(/\((.+?)(\|?(?: (?:.+?)?)?)\)/g, '(<$1>$2)') //Escape links as clickable links
-			.replaceAll(/([\[\]\(\)\-\*\!\#\`])/g, '\\$1'); 		 //Comment most of the markdown elements to not render them inside
+		);
 		detailedAdditions.push(`Prompt input: <code>\n${escapedPromptInput}\n</code>`);
 	}
 	const detailedAdditionsString = detailedAdditions.length > 0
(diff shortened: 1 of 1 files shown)


Abrifq (commented):
OoT/pet peeve:
Prompt input sometimes gets an extra space appended.

Need to check `_doSync()` for what could cause it:

https://github.com/microsoft/vscode/blob/55505bf0b5b7abbe820137bf8e01bed9b72675cf/src/vs/platform/terminal/common/capabilities/commandDetection/promptInputModel.ts#L273-L449

Abrifq:
> I will send the patch in after also testing with links with titles

I have tested (<https://regexr.com/8hu9f>) and it should be good for now.

Testing with this input:
```bash
echo `date`; echo </code>[escaped link?](https://youtu.be/dQw4w9WgXcQ "asdasdasd")
```

---

## Main Branch (https://github.com/microsoft/vscode/commit/124710c5ffc06cedba2e4975d89f3fabcc0dada9)

image



---

## My patch (https://github.com/microsoft/vscode/commit/3ccfea3edcd9495030c11641f6c2b9170fb8fe2c)

image

Abrifq:
> I'll also check if just applying https://github.com/microsoft/vscode/commit/db999f4ba0508dcc82b86081addf1904bcf4a98e onto main works.

It does not. Until someone confirms that markdown rendering inside `<code>` is a bug and gets that fixed, this seems like the best solution.

If I didn't misread the code when I first opened this issue/PR, the changes I made should run once per keystroke, so it should -hopefully- have negligible performance impacts.

src/vs/workbench/contrib/terminal/browser/terminalTooltip.ts
```diff
@@ -44,7 +44,7 @@ export function getInstanceHoverInfo(instance: ITerminalInstance, storageService
 	});
 
 	const shellProcessString = getShellProcessTooltip(instance, !!showDetailed);
-	const content = new MarkdownString(instance.title + shellProcessString + statusString, { supportThemeIcons: true });
+	const content = new MarkdownString(instance.title + shellProcessString + statusString, { supportThemeIcons: true, supportHtml: true });
```
meganrogge: Why do we need to supportHtml here to fix this?

src/vs/workbench/contrib/terminal/browser/terminalTooltip.ts
```diff
@@ -109,7 +109,11 @@ export function refreshShellIntegrationInfoStatus(instance: ITerminalInstance) {
 	}
 	const combinedString = instance.capabilities.get(TerminalCapability.CommandDetection)?.promptInputModel.getCombinedString();
 	if (combinedString !== undefined) {
-		detailedAdditions.push(`Prompt input: \`${combinedString}\``);
+		const escapedPromptInput = combinedString
+			.replaceAll('<', '&lt;').replaceAll('>', '&gt;') 		 //Prevent escaping from wrapping
+			.replaceAll(/\((.+?)(\|?(?: (?:.+?)?)?)\)/g, '(<$1>$2)') //Escape links as clickable links
+			.replaceAll(/([\[\]\(\)\-\*\!\#\`])/g, '\\$1'); 		 //Comment most of the markdown elements to not render them inside
+		detailedAdditions.push(`Prompt input: <code>\n${escapedPromptInput}\n</code>`);
```
meganrogge: this seems pretty complicated. shouldn't we be checking if the string contains a backtick before doing all of this?

src/vs/workbench/contrib/terminal/browser/terminalTooltip.ts
Abrifq: Current markdown parsing doesn't allow for escaping backticks via a backslash, and to workaround that I have changed the backtick wrapping with \<code\> wrapping.

Code wrapping does not work unless supportHtml is enabled, though. Which is why I enabled it.

src/vs/workbench/contrib/terminal/browser/terminalTooltip.ts
Abrifq: I'll make an update tonight but yeah, good call.

Could you also provide test cases where something should or should not be escaped?

Line 114 checks for the href part of a markdown link for a pipe character, removes it if exists and lets the url render as a clickable link again.

It is not vital to this PR, so, it can be removed.

P.S: I got no idea WHY a link inside an **inline** code element gets rendered as a link.

Tyriar: isn't this just as simple as escaping inner backticks like this?

```ts
combinedString.replaceAll('`', '\\`')
```

Tyriar (changes requested):

src/vs/workbench/contrib/terminal/browser/terminalTooltip.ts
Abrifq: I have tried it and it didn't work that time
 ( https://github.com/microsoft/vscode/pull/272425#issuecomment-3455257831 )

https://github.com/Abrifq/vscode/tree/Abrifq/alt-wrap-fix

But I will update the branch & try again and report here.

Tyriar: The current proposal seems far too complex for what this actually does imo, all it's doing is making debug information show a little nicer. Does this helper do the job?

https://github.com/microsoft/vscode/blob/1d4fe40748197f0478cdf6247d43ec133ccfc67b/src/vs/base/common/htmlContent.ts#L153-L156

Abrifq: > Does this helper do the job?
> 
> https://github.com/microsoft/vscode/blob/1d4fe40748197f0478cdf6247d43ec133ccfc67b/src/vs/base/common/htmlContent.ts#L153-L156

Yeah, it should replace line 115. Thanks!

> The current proposal seems far too complex for what this actually does imo, all it's doing is making debug information show a little nicer.

I need to check if lines 112, 113 & 116 themselves are good enough to ONLY fix the wrapping, but I faintly recall they were not.

As for line 114, I have stated in meganrogge's review that it's basically optional and can be removed.

I'll see what I can do within my lunch break, and thanks for the helper.

Tyriar (approved):
@meganrogge simplified it, basically handling the edge case where backtick is included specially instead of adding HTML support/complicated regexes. Could you check it out?

meganrogge (approved):
Works for me

src/vs/workbench/contrib/terminal/browser/terminalTooltip.ts
```diff
@@ -109,11 +109,15 @@ export function refreshShellIntegrationInfoStatus(instance: ITerminalInstance) {
...
-		);
-		detailedAdditions.push(`Prompt input: <code>\n${escapedPromptInput}\n</code>`);
+		if (combinedString.includes('`')) {
+			detailedAdditions.push('Prompt input:' + [
+				'```',
+				combinedString, // No new lines so no need to escape ```
+				'```',
+			].map(e => `\n    ${e}`).join(''));
+		} else {
+			detailedAdditions.push(`Prompt input: \`${combinedString.replaceAll('`', '&#96;')}\``);
```
Abrifq, line 119: You already checked if it has any backticks on line 112, the ```replaceAll('`' ...)``` is unnecessary.

Have you checked if it works as is? The solution might just be this.

Abrifq: > Have you checked if it works as is? The solution might just be this.

Nope. Didn't work as expected. Hmm.

Abrifq: The simplest solution will be only wrapping with triple backticks and assume the user is not writing markdown in command line.

Using two backticks in a row is pretty much useless as it just makes an empty string.

Tyriar: Ah that was an alternate solution I was trying, it's redundant and should be removed.

> The simplest solution will be only wrapping with triple backticks and assume the user is not writing markdown in command line.

It actually works when writing markdown too since we put it on a single line.

Abrifq: > It actually works when writing markdown too since we put it on a single line.

Except for the case they try to write markdown _code blocks_. (I meant to write that before, sorry. Bad case of ADHD, I guess.)

Abrifq: I also forgot to _create the PR_... https://github.com/microsoft/vscode/pull/277406

Thanks for peer-reviewing, by the way. Have a nice weekend.

PR #272435, opened 2025-10-21
Support basic MarkdownStrings in tree view labels

### LINKED ISSUE #271523: Support codicons in tree item label
Context: https://github.com/microsoft/vscode-pull-request-github/pull/8001

In the past, we may have said we don't want to support this because it will encourage lots of icon use in tree item labels, but really extensions can already do this if they want by using emojis. Better to encourage a consistent look with codicons than to allow the wild west of rendering the difference we get with emojis.

kycutler:
Closes #271523, #115365


src/vscode-dts/vscode.proposed.treeViewRenderLabelIcons.d.ts
```diff
@@ -0,0 +1,18 @@
...
+
+	// @kycutler https://github.com/microsoft/vscode/issues/271523
+
+	export interface TreeViewOptions<T> {
+		/**
+		 * Whether to render icons in tree item labels. When `true`, icons embedded in the label
+		 * (e.g., codicons like `$(icon-name)`) will be rendered as icons. When `false`, they will
+		 * be rendered as plain text. Defaults to `false`.
+		 */
+		renderLabelIcons?: boolean;
```
alexr00: Instead of having an option here, I think we should allow `TreeItemLabel`'s `label`, to be `string | MarkdownString`.

alexr00: Using `MarkdownString` will also let us support italics, underline, and bold if we ever want to.

alexr00 (changes requested):

src/vscode-dts/vscode.proposed.treeViewRenderLabelIcons.d.ts
kycutler: `TreeView` uses `IResourceLabel`s for the item labels, which end up being far more complex than just simple strings. But they do support icon rendering. Rather than trying to add Markdown support to those, or updating `TreeView` to not use resource labels, I figure the easiest way is to just use that flag.

The reason it's a flag on the tree view and not in the label itself is due to the way tree views are rendered -- from my understanding the tree items are rendered using templates, and icon support is configured in the template. So it's not really feasible to configure on each individual item.

alexr00: In our API, the way we support codicons is with `MarkdownString`, so to be consistent with the rest of the API we should also use `MarkdownString` here. Especially given that we might want to support italic, underline, and bold someday ™️([issue](https://github.com/microsoft/vscode/issues/115365)), we need to make sure that we design the API in such a way that we won't end up with conflicting or duplicate ways to express the same thing in the future (i.e., we don't want the API user to be able to say `renderLabelIcons` here, but then provide a `MarkdownString` with `supportThemeIcons: false`). 

We don't need to use a markdown string in the tree view itself, we can keep it as a string there. At a glance, I don't see a reason why `IIconLabelValueOptions` couldn't be updated to have a `supportIcons` at set time instead of requiring at label creating time.

kycutler: Done! Also supports bold, italics, and strikethrough around the entire markdown string.

src/vs/base/browser/ui/iconLabel/iconLabel.ts
```diff
@@ -103,10 +105,12 @@ export class IconLabel extends Disposable {
 
 	private readonly hoverDelegate: IHoverDelegate;
 	private readonly customHovers: Map<HTMLElement, IDisposable> = new Map();
+	private currentSupportIcons: boolean = false;
```
bpasero: I cannot wrap my head around this field, can we do this differently? It seems complex, and used for the description only? Could we get this differently?

bpasero (dismissed):

src/vs/base/browser/ui/iconLabel/iconLabel.ts
kycutler: Good callout! I meant to revisit this. In theory it was because `HighlightedLabel` only expected the flag in the constructor. But it looks like this is the only place it was ever used anyway. So refactored to pass it dynamically and remove this field. Thanks!

src/vs/base/browser/ui/iconLabel/iconLabel.ts
```diff
@@ -115,7 +117,7 @@ export class IconLabel extends Disposable {
 		this.nameContainer = dom.append(this.labelContainer, dom.$('span.monaco-icon-name-container'));
 
 		if (options?.supportHighlights || options?.supportIcons) {
```
alexr00, line 119: Can this just be this now, or does only the `LabelWithHightlights`, and not just `Label` support icons?


```suggestion
		if (options?.supportHighlights) {
```

src/vs/workbench/browser/parts/views/treeView.ts
```diff
@@ -1318,24 +1326,84 @@ class TreeRenderer extends Disposable implements ITreeRenderer<ITreeItem, FuzzyS
 						node.resolve(token).then(() => resolve(node.tooltip));
 					});
 				},
-			markdownNotSupportedFallback: resource ? undefined : (label ?? '') // Passing undefined as the fallback for a resource falls back to the old native hover
+			markdownNotSupportedFallback: resource ? undefined : (label ? (isMarkdownString(label) ? label.value : label) : '') // Passing undefined as the fallback for a resource falls back to the old native hover
 		};
 	}
 
+	private processLabel(label: string | IMarkdownString | undefined, matches: { start: number; end: number }[] | undefined): { label: string | undefined; bold: boolean; italic: boolean; strikethrough: boolean; supportIcons: boolean | undefined } {
+		if (isMarkdownString(label)) {
```
alexr00: nit: early return if `!isMarkdownString(label)` to reduce indentation depth.

alexr00 (dismissed):
Looks good to me. @benibenj, do you own iconlabel/do you want to review it?

src/vs/workbench/browser/parts/views/media/views.css
```diff
@@ -184,6 +184,11 @@
 	overflow: hidden;
 }
 
+.customview-tree .monaco-list .monaco-list-row .custom-view-tree-node-item .custom-view-tree-node-item-resourceLabel .monaco-highlighted-label .codicon {
```
benibenj, line 187: This CSS only applies to a handful of trees we have. You've added support to IconLabel which can be rendered anywhere in our UI. Have you tried adding an icon to an iconlabel which is rendered somewhere else in the workbench? Is CSS really only needed for the custom tree?

kycutler: `IconLabel` already had support for inline icons -- this is already used elsewhere like in terminal tab labels. So there are already styles in other places that I don't want to regress. This PR really just adds support for `MarkdownString` (including the icon flag which is just used to trigger the existing logic) in custom tree items, so I figure it's best to keep the styles here specific to that scenario.

src/vs/workbench/browser/parts/views/treeView.ts
kycutler: Fair point, updated

src/vs/base/browser/ui/iconLabel/iconLabel.ts
kycutler: Only `LabelWithHighlights` supports icon rendering, so this check is still needed.

src/vs/base/browser/ui/iconLabel/iconLabel.ts
```diff
@@ -338,14 +344,17 @@ class LabelWithHighlights extends Disposable {
 		this.label = label;
 		this.options = options;
 
+		// Determine supportIcons: use option if provided, otherwise use constructor value
+		const supportIcons = options?.supportIcons ?? this.supportIcons;
```
alexr00, line 348: Looks like the idea here is that the `options` here overrides `this.supportIcons`. Should probably check if `options?.supportIcons` is `undefined` instead of truthy.

alexr00 (changes requested):
One small comment! Please correct me if I'm wrong.

src/vs/base/browser/ui/iconLabel/iconLabel.ts
kycutler: `??` only checks for nullish values, so this will still be `false` if `options.supportIcons` is `false`.

alexr00: Nice.

alexr00 (commented):
🚀

alexr00 (approved):

PR #272482, opened 2025-10-21
Ignore default auto approve rules

### LINKED ISSUE #270641: chat.tools.terminal.autoApprove object format with "approve": true doesn't work
I'm attempting to enable auto approve for curl commands that go to localhost, but no other locations. In settings.json I replaced "curl": false with:
```
        "/^curl\\b.*localhost.*$/i": {
            "approve": true,
            "matchCommandLine": true
        },
```
However, curl to localhost was not approved.
The documentation says that both true and false are valid values for "approve", but all of the examples used "approve": false.

- Copilot Chat Extension Version: v0.31.5 (Build 1)
- VS Code Version: Version: 1.104.3
- OS Version: macOS 15.6.1
- Selected model (e.g. GPT 4.1, Claude 3.7 Sonnet): Cluade Sonnet 4
- Logs: None

Steps to Reproduce:

1. Edit settings.json
2. Replace "curl": false with
```
        "/^curl\\b.*localhost.*$/i": {
            "approve": true,
            "matchCommandLine": true
        },
```
3. Save the file
4. Set the copilot chat to agent mode
5. Ask copilot to execute the command "**curl -X GET http://localhost:3001/api/state**". This should be auto approved, but is not. Note that if executed, that command would fail, but that is ok, we only care if the command autoApproves or not.

I have a work-around, by allowing curl commands, but then setting "approve": false for any curl command that doesn't go to localhost:
```
        "curl": true,
        "/^curl\\b(?!.*localhost).*$/i": {
            "approve": false,
            "matchCommandLine": true
        },
```
The fact that the "approve": false version works but the "approve": true version does not makes me conclude that the issue is "approve": true setting is not being respected

### LINKED ISSUE #272250: Consider option for disabling default terminal auto approve rules
https://github.com/microsoft/vscode/issues/267206 gives an example I see over and over where people are confused about the `false` auto approve rules. Part of this is it's a little complex, the other part is it's hard to know where to list this rules out (open settings.json and auto complete `chat.tools.terminal.autoApprove`) since there's no settings UI (https://github.com/microsoft/vscode/issues/270653).

We should consider a setting to disable either the default rules or maybe just the false ones.

Tyriar:
Fixes #272250
FIxes #270641

### DIFF
--- a/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/commandLineAutoApprover.ts
+++ b/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/commandLineAutoApprover.ts
@@ -45,6 +45,7 @@ export class CommandLineAutoApprover extends Disposable {
 		this._register(this._configurationService.onDidChangeConfiguration(e => {
 			if (
 				e.affectsConfiguration(TerminalChatAgentToolsSettingId.AutoApprove) ||
+				e.affectsConfiguration(TerminalChatAgentToolsSettingId.IgnoreDefaultAutoApproveRules) ||
 				e.affectsConfiguration(TerminalChatAgentToolsSettingId.DeprecatedAutoApproveCompatible)
 			) {
 				this.updateConfiguration();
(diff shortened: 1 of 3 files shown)


chrmarti (approved):

Tyriar:

PR #272501, opened 2025-10-21
Fix #272423

### LINKED ISSUE #272423: Handoffs not setting the correct chat mode
When using handoffs in VS Code Copilot Chat, the chat mode is not set correctly to the agent specified in the handoff.

- Copilot Chat Extension Version: 0.33.2025102001
- VS Code Version: latest Insiders
- OS Version: Windows
- Feature (e.g. agent/edit/ask mode): custom chat modes
- Selected model: Claude Sonnet 4

**Steps to Reproduce:**

1. Create a custom chat mode A
2. Create custom chat mode B that specifies `handoffs.agent` = <chat-mode-A>
3. Run prompt using custom mode B and select handoff to mode A
4. The handoff happens but the selected mode is **Agent** and not mode A

digitarald:
## Problem

The `handoff.agent` property contains human-readable mode names (e.g., "Default"), but `setChatMode` expects mode IDs. This caused a mismatch where:
- Builtin modes have `id === ChatModeKind` (e.g., "agent")  
- Custom modes have `id === URI.toString()` (e.g., "file:///...")
- Handoffs specify modes by their display name (e.g., "Default")

When `setChatMode(handoff.agent)` was called with a name like "Default", it couldn't find the matching mode because it only looked up by ID.

## Solution

Added targeted mode resolution in the `handleNextPromptSelection` method in `chatWidget.ts` where handoffs are actually processed. The fix:

1. First tries `findModeById(handoff.agent)` for builtin modes
2. Falls back to `findModeByName(handoff.agent)` for custom modes specified by name
3. Passes the resolved mode's ID to `setChatMode`

```typescript
if (handoff.agent) {
    // Handoffs specify modes by name, so try ID first (for builtin modes), then name (for custom modes)
    const mode = this.chatModeService.findModeById(handoff.agent) ?? this.chatModeService.findModeByName(handoff.agent);
    if (mode) {
        this.input.setChatMode(mode.id);
    }
}
```

## Why This Approach

- **Targeted**: Only affects handoff processing, not all `setChatMode` calls
- **Clear intent**: Comment explains that handoffs use mode names
- **Explicit resolution**: Shows the ID→name fallback logic where it's needed
- **Safer**: Doesn't change `setChatMode` behavior for other callers
- **Maintainable**: Future developers see the special handling at the handoff point

## Testing

- ✅ No compilation errors
- ✅ Handoffs with custom agent names now correctly resolve modes
- ✅ Falls back gracefully if mode is not found
- ✅ Builtin modes still work by ID

### DIFF
--- a/src/vs/workbench/contrib/chat/browser/chatInputPart.ts
+++ b/src/vs/workbench/contrib/chat/browser/chatInputPart.ts
@@ -727,7 +727,6 @@ export class ChatInputPart extends Disposable implements IHistoryNavigationWidge
 		}
 
 		const mode2 = this.chatModeService.findModeById(mode) ??
-			this.chatModeService.findModeByName(mode) ??
 			this.chatModeService.findModeById(ChatModeKind.Agent) ??
 			ChatMode.Ask;
 		this.setChatMode2(mode2, storeSelection);
--- a/src/vs/workbench/contrib/chat/browser/chatInputPart.ts
+++ b/src/vs/workbench/contrib/chat/browser/chatInputPart.ts
@@ -727,6 +727,7 @@ export class ChatInputPart extends Disposable implements IHistoryNavigationWidge
 		}
 
 		const mode2 = this.chatModeService.findModeById(mode) ??
+			this.chatModeService.findModeByName(mode) ??
 			this.chatModeService.findModeById(ChatModeKind.Agent) ??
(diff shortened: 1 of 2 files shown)


aeschli:
Replaced by [272544](https://github.com/microsoft/vscode/pull/272544)

PR #272523, opened 2025-10-21
Custom tool for confirming terminal executions in other agents

DonJayamanne:
For microsoft/vscode#270852

### DIFF
--- a/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/terminal.chatAgentTools.contribution.ts
+++ b/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/terminal.chatAgentTools.contribution.ts
@@ -42,7 +42,7 @@ class ChatAgentToolsContribution extends Disposable implements IWorkbenchContrib
 
 		// #region Terminal
 
-		const terminalConfirmationHelper = instantiationService.createInstance(RunInTerminalToolConfirmationHelper);
+		const terminalConfirmationHelper = this._register(instantiationService.createInstance(RunInTerminalToolConfirmationHelper));
 		const confirmTerminalCommandTool = instantiationService.createInstance(ConfirmTerminalCommandTool, terminalConfirmationHelper);
 		this._register(toolsService.registerTool(ConfirmTerminalCommandToolData, confirmTerminalCommandTool));
 		const getTerminalOutputTool = instantiationService.createInstance(GetTerminalOutputTool);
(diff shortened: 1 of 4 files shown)


src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/terminal.chatAgentTools.contribution.ts
```diff
@@ -41,6 +42,9 @@ class ChatAgentToolsContribution extends Disposable implements IWorkbenchContrib
 
 		// #region Terminal
 
+		const terminalConfirmationHelper = instantiationService.createInstance(RunInTerminalToolConfirmationHelper);
```
rebornix: let's track it in disposable

rebornix (commented):
nit

rebornix (approved):
let's get the changes in since it's exposing the functionality from existing tool. @Tyriar let us know if you have comments we can resolve in followup PRs.

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/terminal.chatAgentTools.contribution.ts
```diff
@@ -41,6 +42,9 @@ class ChatAgentToolsContribution extends Disposable implements IWorkbenchContrib
 
 		// #region Terminal
 
+		const terminalConfirmationHelper = this._register(instantiationService.createInstance(RunInTerminalToolConfirmationHelper));
+		const confirmTerminalCommandTool = instantiationService.createInstance(ConfirmTerminalCommandTool, terminalConfirmationHelper);
+		this._register(toolsService.registerTool(ConfirmTerminalCommandToolData, confirmTerminalCommandTool));
```
Tyriar, line 47: We need to register this conditionally don't we? Keeping tools to a minimum is important and as I understand it this would add a permanent useless tool to anyone who's model isn't instructed to use it.

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/runInTerminalTool.ts
```diff
@@ -277,9 +276,9 @@ export class RunInTerminalTool extends Disposable implements IToolImpl {
 	}
 
 	constructor(
+		private readonly _confirmationHelper: RunInTerminalToolConfirmationHelper,
```
Tyriar, line 279: Could the tool ctor include whether or not to invoke or confirm only instead? Wouldn't that be a lot simpler of a change?

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/terminal.chatAgentTools.contribution.ts
DonJayamanne: I think this_ will need to be permanent, as its not based on the model

Tyriar: But then regular agent mode in chat will have this useless tool available that will cause unexpected results if it happens to get called?

PR #272532, opened 2025-10-21
Support auto approval of tools in custom confirmation tool

DonJayamanne:
For microsoft/vscode#270852

### DIFF
--- a/src/vs/workbench/contrib/chat/common/tools/confirmationTool.ts
+++ b/src/vs/workbench/contrib/chat/common/tools/confirmationTool.ts
@@ -78,7 +78,7 @@ export class ConfirmationTool implements IToolImpl {
 			confirmationMessages: {
 				title: parameters.title,
 				message: new MarkdownString(parameters.message),
-				allowAutoConfirm: false
+				allowAutoConfirm: true
 			},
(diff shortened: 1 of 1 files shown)


src/vs/workbench/contrib/chat/common/tools/confirmationTool.ts
```diff
@@ -78,7 +78,7 @@ export class ConfirmationTool implements IToolImpl {
 			confirmationMessages: {
 				title: parameters.title,
 				message: new MarkdownString(parameters.message),
-				allowAutoConfirm: false
+				allowAutoConfirm: true
```
DonJayamanne, line 81: @roblourens 
I'm not sure why this was `false` 
Let me know if you have any concerns

rebornix (approved):

src/vs/workbench/contrib/chat/common/tools/confirmationTool.ts
roblourens: Yeah, this is a generic tool for different types of confirmations. It should not allow auto confirm because that would auto-confirm different types of tools. We would need a different solution to enable auto-confirm with the right logic for external tools

PR #272554, opened 2025-10-21
Specify hot key for 'Select folder' button on the Open Folder dialog

### LINKED ISSUE #165161: [Accessibility] Open Folder dialog controls do not have accelerator keys
CC @isidorn

## System Info

Version: 1.73.0-insider (system setup)
Commit: 3544aabd5a2b0da2cd3a6f3fa3f00759aa9d9709
Date: 2022-10-27T21:34:51.585Z
Electron: 19.0.17
Chromium: 102.0.5005.167
Node.js: 16.14.2
V8: 10.2.154.15-electron.0
OS: Windows_NT x64 10.0.22621
Sandboxed: Yes

## Reproducible Steps

1. In VSCode, press Ctrl+K, Ctrl+O to trigger "Open Folder ..." dialog box.

1. See if the following fields have any acceleator keyboard shortcuts:

  - [ ] Folder name text field

  - [ ] Folder location treeview

  - [ ] Select Folder button

None of them above has accelerator keys whereas "Open File ... (Ctrl+O)" dialog has them, such as File Name (Alt+N), File Type (Alt+T), OK (Alt+O), etc.

Please add keyboard shorcuts to each control for better accessibility.

dmitrivMS:
Partial fix for #165161


src/vs/platform/dialogs/electron-main/dialogMainService.ts
```diff
@@ -62,7 +62,8 @@ export class DialogMainService implements IDialogMainService {
 	}
 
 	pickFolder(options: INativeOpenDialogOptions, window?: electron.BrowserWindow): Promise<string[] | undefined> {
-		return this.doPick({ ...options, pickFolders: true, title: localize('openFolder', "Open Folder") }, window);
+		const buttonLabel = mnemonicButtonLabel(localize({ key: 'selectFolder', comment: ['&& denotes a mnemonic'] }, "&&Select folder")).withMnemonic;
```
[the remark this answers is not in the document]
dmitrivMS: This is an OS label, so keeping it as is

dmitrivMS:
Hot key in action:
Recording 2025-10-21 at 14 26 42

src/vs/platform/dialogs/electron-main/dialogMainService.ts
TylerLeonhardt: I wonder what this looks like on macOS/Linux

dmitrivMS: Yeah, me too :-) I don't have a Mac, but I'll see if I can get a Linux VM running to check this out.
I suspect it may be using a different "file or folder" dialog there (the function above).

bpasero (changes requested):
I would like for us to understand why the folder dialog is not keyboard accessible but the others are. In a simple Electron Fiddle I see the folder picker being keyboard accessible, so we do something wrong and need to understand what that is.

bpasero:
Looks like a real Electron bug to me, I would wait for their official fix: https://github.com/electron/electron/issues/48631

dmitrivMS:
> I would like for us to understand why the folder dialog is not keyboard accessible but the others are. In a simple Electron Fiddle I see the folder picker being keyboard accessible, so we do something wrong and need to understand what that is.

Did you try this on Windows?

bpasero:
@dmitrivMS yes

src/vs/platform/dialogs/electron-main/dialogMainService.ts
bpasero: But this issue is specifically for Windows, so please put this change behind a `isWindows` check. I think we should also document as to why we are doing this with a reference to the Electron issue.

bpasero (changes requested):
As this seems to actually be a Windows bug, we can go ahead with this change, with the feedback I posted as change request. This will at least make the button accessible.

src/vs/platform/dialogs/electron-main/dialogMainService.ts
dmitrivMS: Done

bpasero (approved):
Thanks!

PR #272623, opened 2025-10-22
Adding directory support for MCP resources

dileepyavan:
fixes #262368


src/vs/workbench/contrib/mcp/browser/mcpResourceQuickAccess.ts
```diff
@@ -391,21 +395,45 @@ export abstract class AbstractMcpResourceAccessPick {
...
-				runOptions.handleAccept?.(picker.activeItems[0], event.inBackground);
-			} else {
-				const [item] = picker.selectedItems;
-				const uri = await helper.toURI((item as ResourceQuickPickItem).resource);
-				if (uri) {
-					this._editorService.openEditor({ resource: uri, options: { preserveFocus: event.inBackground } });
+			try {
+				picker.busy = true;
+				if (runOptions?.handleAccept) {
+					runOptions.handleAccept?.(picker.activeItems[0], event.inBackground);
```
connor4312: We should handle this case too -- `handleAccept` is used in the quick access flow and we should only accept files, not directories

src/vs/workbench/contrib/mcp/browser/mcpResourceQuickAccess.ts
```diff
@@ -391,21 +395,45 @@ export abstract class AbstractMcpResourceAccessPick {
...
+									description: child.isDirectory ? localize('mcp.directory', 'Directory') : '',
+								}));
+								picker.items = directoryItems;
+								picker.show();
+							}
+						} catch (error) {
+							this._notificationService.error(localize('mcp.readDirError', "Unable to read directory '{0}'", uri.toString()));
+						}
+					} else if (uri) {
+						picker.hide();
```
connor4312, line 547: We seem to have lost the `event.inBackground` option here. This can be tested by alt-clicking a resource in the Add Context flow.

src/vs/workbench/contrib/mcp/browser/mcpResourceQuickAccess.ts
```diff
@@ -391,21 +395,45 @@ export abstract class AbstractMcpResourceAccessPick {
...
+						try {
+							const stat = await this._fileService.resolve(uri, { resolveMetadata: false });
+							if (stat.children) {
+								isDirectoryMode = true;
+								const directoryItems: IQuickPickItem[] = stat.children.map((child) => ({
+									id: child.resource.toString(),
+									label: child.name,
+									description: child.isDirectory ? localize('mcp.directory', 'Directory') : '',
+								}));
+								picker.items = directoryItems;
```
connor4312: We should provide a "back" navigation too. This can be done with just a `..` option like we do for the normal File dialog, or an explicit "Back" item

src/vs/workbench/contrib/mcp/browser/mcpResourceQuickAccess.ts
```diff
@@ -391,21 +395,45 @@ export abstract class AbstractMcpResourceAccessPick {
...
+					const [item] = picker.selectedItems;
+					let uri: URI | undefined;
+					if (isDirectoryMode) {
+						uri = URI.parse((item as IQuickPickItem).id!);
+					} else {
+						uri = await helper.toURI((item as ResourceQuickPickItem).resource);
+					}
+					// If the URI is a directory, we need to read its contents and show them in the picker.
+					// directory URIs end with a `/`
+					if (uri?.path.endsWith('/')) {
```
connor4312: I think it's fine to do this unconditionally. The resource is going to get resolved as soon as you pick it anyway so we may as well `resolve` here all the time and then we don't need to demand any specific URI pattern from the server.

connor4312:
I think we actually want to add much of this logic/handling in `getPicks`

https://github.com/microsoft/vscode/blob/fd24d77753b568e7fae6334cbab5b94fd7ef29ee/src/vs/workbench/contrib/mcp/browser/mcpResourceQuickAccess.ts#L258

`getPicks` is used directly by the the "📎 Add Context" button in chat, where those are pulled in and used in a slightly more generic way in the chat. Probably how I would implement is:

- Make the `McpResourcePickHelper` stateful. We only ever use one `McpResourcePickHelper` per quick pick, so this should be fine.
- Introduce a new `helper.navigate(resource: IMcpResource | IMcpResourceTemplate): Promise<boolean>` which resolves `true` if the item is nested, _and_ it then statefully updates the items it may be emitting in `helper.getPicks` with the new children. (Internally maybe have an `inDirectory = observableValue<undefined | { parent: URI }>` that drive this logic.) If it resolves false then the consumer should know that item is picked and do the existing logic.
- Adopt this flow in the `AbstractMcpResourceAccessPick` where you have your current changes
- Update the [`asAttachment` flow](https://github.com/microsoft/vscode/blob/fe497361bc1705f5be28785df5cbc3258aa2b651/src/vs/workbench/contrib/mcp/browser/mcpAddContextContribution.ts#L66) for the "Add Context" pick. Currently `asAttachment` just gets called when the item is picked, the picker is hidden, and the item is returned (it's defined as `asAttachment(): IChatRequestVariableEntry | Promise<IChatRequestVariableEntry>`). Instead we should have a new `type ChatContextPickerPickItemAttechment = IChatRequestVariableEntry | { type: 'continuePicking' }` and make asAttachment return that.
- Handle this [in chatContextActions](https://github.com/microsoft/vscode/blob/fe497361bc1705f5be28785df5cbc3258aa2b651/src/vs/workbench/contrib/chat/browser/actions/chatContextActions.ts#L639-L659). If `asAttachment` is a promise, mark the picker as busy and wait for it to resolve before doing anything. If what it returns is the `continuePicking` type, then do nothing -- we can expect the new pick items to be updated.

src/vs/workbench/contrib/chat/browser/chatContextPickService.ts
```diff
@@ -21,6 +21,7 @@ export interface IChatContextPickerPickItem extends Partial<IQuickItem> {
 	description?: string;
 	detail?: string;
 	disabled?: boolean;
+	validateForAttachment?(): Promise<boolean>;
 	asAttachment(): IChatRequestVariableEntry | Promise<IChatRequestVariableEntry>;
```
connor4312: I suggest we have a new type like

```
type ChatContextPickAttachment = IChatRequestVariableEntry |'noop';
```

and return that from asAttachment

```suggestion
	asAttachment(): ChatContextPickAttachment | Promise<ChatContextPickAttachment>;
```

That way we don't ever have the weird state of calling `asAttachment` on something that's not 'valid' for an attachment. Trying to make invalid states irrepresentable is good :)

src/vs/workbench/contrib/chat/browser/actions/chatContextActions.ts
```diff
@@ -633,17 +633,27 @@ export class AttachContextAction extends Action2 {
...
 
-		store.add(qp.onDidAccept(e => {
+		store.add(qp.onDidAccept(async e => {
 			const [selected] = qp.selectedItems;
 			if (isChatContextPickerPickItem(selected)) {
-				const attachment = selected.asAttachment();
-				if (isThenable(attachment)) {
-					addPromises.push(attachment.then(v => widget.attachmentModel.addContext(v)));
+				const isValidForAttachment = await selected.validateForAttachment?.();
+				if (isValidForAttachment) {
```
connor4312: We should mark the picker as busy while calling asAttachment/validateForAttachment

src/vs/workbench/contrib/mcp/browser/mcpAddContextContribution.ts
```diff
@@ -39,17 +54,25 @@ export class McpAddContextContribution extends Disposable implements IWorkbenchC
...
+			},
 		});
 	}
 
-	private _getResourcePicks(token: CancellationToken) {
+	private _getResourcePicks(token: CancellationToken, helper: McpResourcePickHelper) {
 		const observable = observableValue<{ busy: boolean; picks: ChatContextPick[] }>(this, { busy: true, picks: [] });
-
-		this._helper.getPicks(servers => {
+		// TODO: still need to dispose helper
```
connor4312: I think this is still relevant!

src/vs/workbench/contrib/mcp/browser/mcpAddContextContribution.ts
```diff
@@ -39,17 +54,25 @@ export class McpAddContextContribution extends Disposable implements IWorkbenchC
...
+		const picksObservable = helper.getPicks(token);
+		this._register(autorun(reader => {
```
connor4312: You can use `derive` instead of `autorun` with a separate observable

src/vs/workbench/contrib/mcp/browser/mcpResourceQuickAccess.ts
```diff
@@ -255,19 +329,28 @@ export class McpResourcePickHelper {
...
 			publish();
 		}, 5_000));
 
+		this._register(autorun(reader => {
+			const directoryResource = this._inDirectory.read(reader);
+			if (directoryResource) {
+				// When navigating into a directory, filter to show only that server's resources
+				const filtered = new Map<IMcpServer, (IMcpResourceTemplate | IMcpResource)[]>();
+				filtered.set(directoryResource.server, directoryResource.resources);
+				this._resources.set(filtered, undefined);
```
connor4312: This usage of _resources is racey because you could navigate to a subdirectory before all resources have been populated.

Setting another observable from inside an autorun is a code smell, in general you should `derive` the necessary output instead

src/vs/workbench/contrib/mcp/browser/mcpResourceQuickAccess.ts
```diff
@@ -36,6 +41,20 @@ export class McpResourcePickHelper {
 		};
 	}
 
+	public AddCurrentMCPQuickPickItemLevel(server: IMcpServer, resources: (IMcpResource | IMcpResourceTemplate)[]): void {
```
connor4312: Seems unused? But should be `camelCased` if kept

src/vs/workbench/contrib/mcp/browser/mcpResourceQuickAccess.ts
```diff
@@ -390,23 +477,35 @@ export abstract class AbstractMcpResourceAccessPick {
...
+						}
+					}
 				}
+			} finally {
+				picker.busy = false;
 			}
 		}));
-
-		return store;
+		return Disposable.None;
```
connor4312: should we not return anything here now?

src/vs/workbench/contrib/mcp/browser/mcpResourceQuickAccess.ts
```diff
@@ -332,55 +418,56 @@ export class McpResourcePickHelper {
...
-			store.dispose();
-		});
+		}));
+		return this._resources;
 	}
 }
 
-
-export abstract class AbstractMcpResourceAccessPick {
+export abstract class AbstractMcpResourceAccessPick extends Disposable {
```
connor4312: I think the `McpResourceQuickAccess` may be a singleton. In this case we will need to rework it such that it create an instance of an implementation of `AbstractMcpResourceAccessPick` rather than extending it, otherwise everything that was changed to `this._register` will be held onto indefinitely.

If you want to make `AbstractMcpResourceAccessPick` disposable you will also need to make sure everywhere we create an instance of an implementation of it get disposed properly (e.g. in mcpCommands.ts for the McpResourceQuickPick)

src/vs/workbench/contrib/mcp/browser/mcpResourceQuickAccess.ts
```diff
@@ -426,7 +527,7 @@ export class McpResourceQuickPick extends AbstractMcpResourceAccessPick {
 		const store = new DisposableStore();
```
connor4312, line 572: In the vein of the above comment on `AbstractMcpResourceAccessPick`, this is a bit funky now -- we create local disposables that are cleaned up when the picker closes, but also have disposables held onto until the entire `McpResourceQuickPick` is disposed. We should go with one or the other.

src/vs/workbench/contrib/mcp/browser/mcpResourceQuickAccess.ts
```diff
@@ -332,55 +418,56 @@ export class McpResourcePickHelper {
...
+		}));
 
-		const store = new DisposableStore();
-		store.add(picker.onDidTriggerItemButton(event => {
+		this._register(picker.onDidTriggerItemButton(event => {
 			if (event.button.tooltip === attachButton) {
 				picker.busy = true;
-				helper.toAttachment((event.item as ResourceQuickPickItem).resource).then(async a => {
+				const resourceItem = event.item as ResourceQuickPickItem;
+				helper.toAttachment(resourceItem.resource, resourceItem.server).then(async a => {
```
connor4312: What happens if it's not `valid[ate]ForAttachment`?

src/vs/workbench/contrib/mcp/common/mcpResourceFilesystem.ts
```diff
@@ -31,6 +31,16 @@ interface IReadData {
...
 
+/**
+ * Helper function to check if a resource path represents a directory.
+ * Works cross-platform by checking if the path ends with a forward slash.
+ * @param path The resource path to check
+ * @returns true if the path represents a directory, false otherwise
+ */
+function isDirectoryPath(path: string): boolean {
+	return path.endsWith('/');
+}
```
connor4312: Is this in the MCP spec somewhere? From the info shared in the original issue it looks like the MCP server does not include a trailing slash on its 'directory' https://github.com/microsoft/vscode/issues/262368

connor4312 (changes requested):

dileepyavan (commented):
> Is this in the MCP spec somewhere? From the info shared in the original issue it looks like the MCP server does not include a trailing slash on its 'directory' https://github.com/microsoft/vscode/issues/262368 

I dont find this as a rule specified in official MCP documentation, i referenced this earlier from chatgpt and i think default agent mode. The only rule specified is https://modelcontextprotocol.io/specification/2025-06-18/server/resources#file%3A%2F%2F . This also contains "MAY" so unsure if it will be always the case

src/vs/workbench/contrib/mcp/browser/mcpAddContextContribution.ts
dileepyavan: Currently the helper disposes when the parent(McpAddContextContribution) disposes. Currently this might create an issue with multiple instances of the MCPResourcehelper. I think we can dispose the helper when the quick pick is closed using "onDidHide" event. For dispoing the helper, need to listen to an observable on onDidHide event

dileepyavan (commented):
> Currently the helper disposes when the parent(McpAddContextContribution) disposes. Currently this might create an issue with multiple instances of the MCPResourcehelper. I think we can dispose the helper when the quick pick is closed using "onDidHide" event.

src/vs/workbench/contrib/mcp/browser/mcpResourceQuickAccess.ts
dileepyavan: As discussed, this is no longer needed for MCP resources workflow.

src/vs/workbench/contrib/chat/browser/actions/chatContextActions.ts
dileepyavan: updated.

src/vs/workbench/contrib/mcp/browser/mcpResourceQuickAccess.ts
dileepyavan: We do not have anything to dispose here. We need to dispose helper only when picker is hidden. Added a new handler for that and removed this._register for onDidAccept method.

src/vs/workbench/contrib/mcp/browser/mcpResourceQuickAccess.ts
dileepyavan: Updated the code to allow only for valid attcachments

src/vs/workbench/contrib/mcp/browser/mcpResourceQuickAccess.ts
dileepyavan: We do have event.inBackground present in this method.

src/vs/workbench/contrib/mcp/browser/mcpResourceQuickAccess.ts
dileepyavan: Removed disposable for AbstractMcpResourceAccessPick

dileepyavan (commented):
> The directory detection logic is incorrect. A resource with exactly 1 content item will be treated as a File, but the check at line 164 (`contents.length <= 1`) treats single-item resources as errors when calling `readdir()`. This creates inconsistent behavior where `stat()` says it's a file but `readdir()` rejects it as not being a directory. 

These changes are no longer needed.

src/vs/workbench/contrib/chat/browser/chatContextPickService.ts
```diff
@@ -21,9 +21,12 @@ export interface IChatContextPickerPickItem extends Partial<IQuickItem> {
 	description?: string;
 	detail?: string;
 	disabled?: boolean;
-	asAttachment(): IChatRequestVariableEntry | Promise<IChatRequestVariableEntry>;
+	validateForAttachment?(): Promise<boolean>;
```
connor4312: I think we no longer need `validateForAttachment` now that we have the updated return type for `asAttachment`

src/vs/workbench/contrib/chat/browser/actions/chatContextActions.ts
```diff
@@ -635,17 +635,32 @@ export class AttachContextAction extends Action2 {
...
 
-		store.add(qp.onDidAccept(e => {
+		store.add(qp.onDidAccept(async e => {
 			const [selected] = qp.selectedItems;
 			if (isChatContextPickerPickItem(selected)) {
-				const attachment = selected.asAttachment();
-				if (isThenable(attachment)) {
-					addPromises.push(attachment.then(v => widget.attachmentModel.addContext(v)));
+				qp.busy = true;
+				const isValidForAttachment = await selected.validateForAttachment?.().finally(() => {
```
connor4312: If we get rid of `validateForAttachment`, we would essentially selected.asAttachment here instead (keeping qp.busy=true while that's resolving).

We still want to keep `addPromises` in case the user closes the quickpick after selecting something and before `asAttachment` resolves, but if the picker is closed by that point we would do nothing if it returns `noop`

src/vs/workbench/contrib/chat/browser/actions/chatContextActions.ts
```diff
@@ -659,6 +674,7 @@ export class AttachContextAction extends Action2 {
 
 		store.add(qp.onDidHide(() => {
 			defer.complete(true);
+			pickerConfig.dispose?.();
```
connor4312, line 681: We also want to call this up on line 632 where we `return true`

src/vs/workbench/contrib/mcp/browser/mcpResourceQuickAccess.ts
```diff
@@ -36,6 +42,20 @@ export class McpResourcePickHelper {
 		};
 	}
 
+	public addCurrentMCPQuickPickItemLevel(server: IMcpServer, resources: (IMcpResource | IMcpResourceTemplate)[]): void {
+		this._pickItemsStack.push({ server, resources });
+	}
+
+	public navigateBack(): boolean {
+		const items = this._pickItemsStack.pop();
+		if (items) {
```
connor4312, line 64: We don't set inDirectory if the stack is empty -- does this work for going back to the root?

src/vs/workbench/contrib/mcp/browser/mcpResourceQuickAccess.ts
```diff
@@ -284,14 +373,17 @@ export class McpResourcePickHelper {
 					break;
 				}
 			}
-			onChange(output);
+
+			if (this._inDirectory.get() === undefined) {
```
connor4312: The check for `this._inDirectory.get()` is unnecessary, and could lead to missed items if the user picks a directory while other server resources are still loading in

src/vs/workbench/contrib/mcp/browser/mcpResourceQuickAccess.ts
```diff
@@ -284,14 +373,17 @@ export class McpResourcePickHelper {
...
+				this._resources.set(output, undefined);
```
connor4312: I suggest we create a new value on each publish, otherwise the observables won't detect a change because they do an equality check on the old and new values. Because we're just updating the same map each time, this won't trigger updates.

We could have the output here be a flat `(IMcpServer | IMcpResourceTemplate | IMcpResource)[]` and let the consumers either filter the results or turn them into separators as necessary

src/vs/workbench/contrib/mcp/browser/mcpResourceQuickAccess.ts
connor4312: We should remove this return now then.

src/vs/workbench/contrib/chat/browser/chatContextPickService.ts
dileepyavan: I misunderstood the earlier comment :). Intention of making it as noop is to include this scenario in asattachment. got it. will update

src/vs/workbench/contrib/mcp/browser/mcpResourceQuickAccess.ts
dileepyavan: If the stack is empty navigateBack function returns false causing the default goback handler to kick in

src/vs/workbench/contrib/mcp/browser/mcpResourceQuickAccess.ts
dileepyavan: correct, i think this got left over from previous changes before derived observable is implemented.

src/vs/workbench/contrib/chat/browser/actions/chatContextActions.ts
dileepyavan: cancellation token is requested on DidHide event. We are already disposing on this event. Do we still need to dispose on line 632?

src/vs/workbench/contrib/mcp/browser/mcpResourceQuickAccess.ts
dileepyavan: In Publish method, we are creating new Map for every update.

src/vs/workbench/contrib/mcp/browser/mcpResourceQuickAccess.ts
dileepyavan: This method is called by `provide(picker: IQuickPick<IQuickPickItem, { useSeparators: true }>, token: CancellationToken, runOptions?: IQuickAccessProviderRunOptions): IDisposable` which is implementing IQuickAccessProvider

src/vs/workbench/contrib/chat/browser/actions/chatContextActions.ts
```diff
@@ -640,17 +640,25 @@ export class AttachContextAction extends Action2 {
...
+		store.add(qp.onDidAccept(async e => {
 			const [selected] = qp.selectedItems;
 			if (isChatContextPickerPickItem(selected)) {
 				const attachment = selected.asAttachment();
+				if (!attachment || attachment === 'noop') {
+					return;
+				}
 				if (isThenable(attachment)) {
-					addPromises.push(attachment.then(v => widget.attachmentModel.addContext(v)));
+					addPromises.push(attachment.then(v => widget.attachmentModel.addContext(v as IChatRequestVariableEntry)));
```
connor4312: In the interface we say that you can return `Promise<IChatRequestVariableEntry | 'noop'>`. We should either change the interface to not allow that (making the `as` cast unnecessary) or support it here

src/vs/workbench/contrib/chat/browser/actions/chatContextActions.ts
connor4312: Yea, if we bail out early (e.g. if the promise was cancelled before awaitable `picks` resolve) then we should still dispose the pickerConfig

src/vs/workbench/contrib/mcp/browser/mcpAddContextContribution.ts
```diff
@@ -63,21 +88,11 @@ export class McpAddContextContribution extends Disposable implements IWorkbenchC
...
-							}
-						}),
+						asAttachment: () => helper.toAttachment(resource, server)
 					});
 				}
 			}
-			observable.set({ picks, busy: true }, undefined);
-		}, token).finally(() => {
-			observable.set({ busy: false, picks: observable.get().picks }, undefined);
+			return { picks, busy: false };
```
connor4312: We should propagate the loading state from the `helper.getPicks` function too I think -- in this change we lost the 'busy' indicator that is shown while the picks are loading in from the servers

src/vs/workbench/contrib/mcp/browser/mcpResourceQuickAccess.ts
```diff
@@ -80,13 +113,84 @@ export class McpResourcePickHelper {
...
+			if (this._isDirectoryResource(resource) && (stat.children?.length ?? 0) > 0) {
+				// Save current state to stack before navigating
+				const currentResources = this._resources.get().get(server);
+				if (currentResources) {
+					this.addCurrentMCPQuickPickItemLevel(server, currentResources);
+				}
+
+				// Convert all the children to IMcpResource objects
+				const childResources: IMcpResource[] = stat.children!.map(child => {
+					// const mcpUri = McpResourceURI.fromServer(server.definition, child.resource.toString());
```
connor4312: Doing this would be good!

src/vs/workbench/contrib/mcp/browser/mcpResourceQuickAccess.ts
```diff
@@ -80,13 +113,84 @@ export class McpResourcePickHelper {
...
+						description: resource.description,
+						mimeType: undefined,
+						sizeInBytes: child.size,
+						icons: McpIcons.fromParsed(undefined)
+					};
+				});
+				this._inDirectory.set({ server, resources: childResources }, undefined);
+				return true;
+			}
+		} catch (error) {
```
connor4312: nit: would prefer to keep this try/catch more specific around the `stat` call to avoid hiding actual programming errors that might crop up

src/vs/workbench/contrib/mcp/browser/mcpResourceQuickAccess.ts
```diff
@@ -80,13 +113,84 @@ export class McpResourcePickHelper {
...
+	}
+
+	public toAttachment(resource: IMcpResource | IMcpResourceTemplate, server: IMcpServer): Promise<ChatContextPickAttachment> | 'noop' {
+		const noop = 'noop';
+		let attachmentResult: ChatContextPickAttachment = noop;
+
+		const deferred = new DeferredPromise<ChatContextPickAttachment>();
+		if (this._isDirectoryResource(resource)) {
+			//Check if directory
+			this.checkIfDirectoryAndPopulate(resource, server);
```
connor4312, line 169: do we want to await this?

src/vs/workbench/contrib/mcp/browser/mcpResourceQuickAccess.ts
```diff
@@ -80,13 +113,84 @@ export class McpResourcePickHelper {
...
+				deferred.complete(attachmentResult);
+			});
 		} else {
-			return this._resourceToAttachment(resource);
+			this._resourceToAttachment(resource).then(val => {
+				attachmentResult = (val as IChatRequestVariableEntry) || noop;
+				deferred.complete(attachmentResult);
+			});
+		}
+		return deferred.p;
```
connor4312: I don't think you need a deferred here. Also the `as` casts are not correct -- the methods can return undefined if the user cancels template URI creation, and that should be bubbled up to the caller.

```suggestion
		if (this._isDirectoryResource(resource)) {
			//Check if directory
			this.checkIfDirectoryAndPopulate(resource, server);
			return noop;
		}

		if (isMcpResourceTemplate(resource)) {
			return this._resourceTemplateToAttachment(resource)
		} else {
			return this._resourceToAttachment(resource);
		}
```

src/vs/workbench/contrib/mcp/browser/mcpResourceQuickAccess.ts
```diff
@@ -332,81 +449,124 @@ export class McpResourcePickHelper {
...
 		}));
 
-		store.add(picker.onDidAccept(async event => {
-			if (!event.inBackground) {
-				picker.hide(); // hide picker unless we accept in background
-			}
+		store.add(picker.onDidHide(() => {
+			helper.dispose();
+			store.dispose();
+		}));
```
connor4312, line 526: We actually do not want to do this, and instead we want to keep returning the `store`. In the case of quick access, the `McpResourceQuickAccess` does not own the quickpick, it only gets the opportunity to populate it. If I open MCP quick access (type `mcpr` + space in the command palette), that triggers this MCP quick access logic. But I can then backspace and type something else. That will cause the value returned by `McpResourceQuickAccess.provide` to be disposed and the quickpick can later be used by some other logic.

src/vs/workbench/contrib/mcp/browser/mcpResourceQuickAccess.ts
```diff
@@ -417,6 +577,8 @@ export class McpResourceQuickPick extends AbstractMcpResourceAccessPick {
 		@IEditorService editorService: IEditorService,
 		@IChatWidgetService chatWidgetService: IChatWidgetService,
 		@IViewsService viewsService: IViewsService,
+		@IFileService fileService: IFileService,
+		@INotificationService notificationService: INotificationService,
```
connor4312: Unused

```suggestion
```

connor4312 (commented):
Almost there, some final more localized comments

src/vs/workbench/contrib/mcp/browser/mcpResourceQuickAccess.ts
dileepyavan: This is kind of fire and forget.

src/vs/workbench/contrib/mcp/browser/mcpResourceQuickAccess.ts
dileepyavan: Agree, deferred was introduced earlier when checkIfDirectoryAndPopulate was returning a value.No longer needed.

src/vs/workbench/contrib/mcp/browser/mcpResourceQuickAccess.ts
dileepyavan: Updated

src/vs/workbench/contrib/mcp/browser/mcpResourceQuickAccess.ts
dileepyavan: updated

src/vs/workbench/contrib/mcp/browser/mcpAddContextContribution.ts
dileepyavan: updated

src/vs/workbench/contrib/mcp/browser/mcpAddContextContribution.ts
```diff
@@ -42,42 +56,46 @@ export class McpAddContextContribution extends Disposable implements IWorkbenchC
...
+					picks.push(McpResourcePickHelper.sep(server));
+					for (const resource of resources) {
+						picks.push({
+							...McpResourcePickHelper.item(resource),
+							asAttachment: () => helper.toAttachment(resource, server)
+						});
+					}
 				}
+				return { picks, busy: pickItems.isBusy };
+			} catch {
```
connor4312: I suggest removing the try/catch. Errors are not expected here and it will just hide anything that comes up.

src/vs/workbench/contrib/mcp/browser/mcpResourceQuickAccess.ts
```diff
@@ -426,7 +577,7 @@ export class McpResourceQuickPick extends AbstractMcpResourceAccessPick {
 		const store = new DisposableStore();
 		const qp = store.add(this._quickInputService.createQuickPick({ useSeparators: true }));
 		qp.placeholder = localize('mcp.quickaccess.placeholder', "Search for resources");
-		store.add(this.applyToPick(qp, token));
+		this.applyToPick(qp, token);
```
connor4312: This should be disposed again now

src/vs/workbench/contrib/mcp/browser/mcpResourceQuickAccess.ts
```diff
@@ -331,81 +437,126 @@ export class McpResourcePickHelper {
...
+
+				const resourceItem = item as ResourceQuickPickItem;
+				const resource = resourceItem.resource;
+				let uri: URI | undefined;
+
+				// Try to navigate into the resource if it's a directory
+				const isNested = await helper.navigate(resource, resourceItem.server);
+				if (isNested) {
+					// Navigation succeeded, picker will be updated by getPicks() observing _inDirectory changes
+					picker.show();
```
connor4312: Technically don't need the picker.show() calls here and above. It is not hidden by default when an item is accepted, unless you explicitly do so

dileepyavan: Updated

src/vs/workbench/contrib/mcp/browser/mcpAddContextContribution.ts
dileepyavan: updated

connor4312 (approved):

PR #272628, opened 2025-10-22
tools: add postapproval step for tool calls

### LINKED ISSUE #270618: Make approval for read tools happen after rather than before the tool call
For chat tools that read data, what the user _actually_ wants to approve is not the read request but the response. For example a Github issue may have content that's invisible when seeing the rendered issue and would only be detectable if approval happens after retrieving it and before giving it to the model. Same deal with the `fetch` tool.

For MCP land, we currently prompt pre-approval for tools that have `readOnlyHint` _or_ `openWorldHint`. Once this happens I will not request pre-approval for `readOnlyHint` and only post-approval for `openWorldHint`.

Downside is that I think for fetch we kind of want to approve both cases, at least for untrusted domains, which is a little annoying. Likewise tools that are open world but not readonly may 'properly' need two approvals.

cc @Tyriar @TylerLeonhardt for discussion

connor4312:
Closes https://github.com/microsoft/vscode/issues/270618

### DIFF
--- a/src/vs/workbench/contrib/chat/browser/chatContentParts/chatToolOutputContentSubPart.ts
+++ b/src/vs/workbench/contrib/chat/browser/chatContentParts/chatToolOutputContentSubPart.ts
@@ -1,0 +1,267 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import * as dom from '../../../../../base/browser/dom.js';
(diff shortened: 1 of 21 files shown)


src/vs/workbench/contrib/chat/browser/chatContentParts/chatToolOutputContentSubPart.ts
```diff
@@ -0,0 +1,267 @@
+/*---------------------------------------------------------------------------------------------
```
connor4312, line 1: This file is a mostly literal move from chatToolInputOutputContentPart.ts, doesn't need granular review

dbaeumer (approved):

PR #272730, opened 2025-10-22
Fix builtin extension detection

### LINKED ISSUE #272729: Support custom built in extensions

- VS Code Version: 1.105.1
- OS Version: Ubuntu 24.04

Steps to Reproduce:

1. add another extension in the VSCode default built-in extension folder
2. start VSCode
3. edit `.vscode/extensions.json`
4. in the `recommendations`, the extension is suggested, while all the other built-in extensions are not

CGNonofr:
fix #272729

Even if it's very minor, is touches the public VSCode extension api, so is there any special process for it?


sandy081:
Closed the original issue - https://github.com/microsoft/vscode/issues/272729 - As we do not support adding folders to system built in extensions

PR #272793, opened 2025-10-22
ensure we carry over title when 'swapping' chat sessions

joshspicer:
ref https://github.com/microsoft/vscode-internalbacklog/issues/6026

quick patch for title when swapping chat sessions.  Our title setting logic is now quite complicated, more thought needed on how to simplify it.


src/vs/workbench/api/browser/mainThreadChatSessions.ts
```diff
@@ -409,6 +411,13 @@ export class MainThreadChatSessions extends Disposable implements MainThreadChat
 			originalGroup = this.editorGroupService.activeGroup;
 		}
 
+		const options: IChatEditorOptions = {
+			title: {
+				preferred: originalEditor?.getName() || undefined,
```
pwang347, line 416: would it be better if we just fired off label change on the initial `resolve` call in the ChatEditorInput for this scenario?

pwang347 (approved):

PR #272805, opened 2025-10-23
Support for multiple same-type chat session contrubtions

osortega:


rebornix (changes requested):
let's do https://github.com/microsoft/vscode/pull/272817 instead

bpasero (commented):
Not doing a review here but an observation: I would prefer if `.contribution.ts` files are purely about registering/contributing things but then moving the meaty stuff into their own files. For example, why is there no separate file that is the chat sessions service implementation?

PR #272860, opened 2025-10-23
Store global tools selection per profile (not window)

### LINKED ISSUE #266058: Tool enablement is per window, how do I set it globally?
I tried to cut down the number of tools I'm using (eg. disabling new, fetch, runTasks, etc.) but it turns out if I want to do that I need to do it for every single window I want to work in.

aeschli:
Fixes https://github.com/microsoft/vscode/issues/266058


bpasero (approved):

src/vs/workbench/contrib/chat/browser/chatSelectedTools.ts
```diff
@@ -114,7 +114,7 @@ export class ChatSelectedTools extends Disposable {
 			toStorage: ToolEnablementStates.toStorage
 		});
 
-		this._globalState = this._store.add(globalStateMemento(StorageScope.WORKSPACE, StorageTarget.MACHINE, _storageService));
+		this._globalState = this._store.add(globalStateMemento(StorageScope.PROFILE, StorageTarget.MACHINE, _storageService));
```
Tyriar, line 117: Should this be `StorageTarget.USER` so it gets synced?

PR #272943, opened 2025-10-23
add `id` for `TerminalCommand`, add `Add to Chat` action for commands

### LINKED ISSUE #259261: Allow addressing particular terminals and terminal commands in chat attachments

meganrogge:
fixes #259261

https://github.com/user-attachments/assets/c2d223ad-7d24-4cef-b1da-fc2ce2068d50


src/vs/workbench/contrib/terminal/browser/terminalInstance.ts
```diff
@@ -788,7 +788,7 @@ export class TerminalInstance extends Disposable implements ITerminalInstance {
 		}
 
 		const disableShellIntegrationReporting = (this.shellLaunchConfig.executable === undefined || this.shellType === undefined) || !shellIntegrationSupportedShellTypes.includes(this.shellType);
-		const xterm = this._scopedInstantiationService.createInstance(XtermTerminal, Terminal, {
+		const xterm = this._scopedInstantiationService.createInstance(XtermTerminal, this.instanceId, Terminal, {
```
Tyriar: `XtermTerminal` isn't meant to know about `instanceId` (or even what a terminal instance is). It would be better to pass through the URI of the terminal and then add the query parameter in DecorationAddon.

src/vs/workbench/contrib/terminal/browser/terminalActions.ts
```diff
@@ -510,6 +510,27 @@ export function registerTerminalActions() {
...
+			const instance = c.service.getInstanceFromResource(resource);
+			if (!instance) {
+				return;
+			}
+			c.service.setActiveInstance(instance);
+			focusActiveTerminal(instance, c);
+			const command = instance.capabilities.get(TerminalCapability.CommandDetection)?.commands.find(c => c.id === resource.query);
+			if (command) {
+				instance.xterm?.markTracker.revealCommand(command);
+			}
```
Tyriar: Could this change to `markTracker.revealCommand(resource)` so it's convenient to do this in the future?

src/vs/workbench/contrib/chat/browser/chatAttachmentWidgets.ts
```diff
@@ -160,6 +161,14 @@ abstract class AbstractChatAttachmentWidget extends Disposable {
 			return;
 		}
 
+		if (resource.scheme === Schemas.vscodeTerminal) {
+			this.element.style.cursor = 'pointer';
```
Tyriar: Setting style when open is called?

src/vs/workbench/contrib/chat/browser/chatAttachmentWidgets.ts
```diff
@@ -160,6 +161,14 @@ abstract class AbstractChatAttachmentWidget extends Disposable {
...
+			this._register(dom.addDisposableListener(this.element, dom.EventType.CLICK, () => {
```
Tyriar: Doesn't click already happen at this point as `openResource` is called?

src/vs/workbench/contrib/chat/browser/chatAttachmentWidgets.ts
```diff
@@ -160,6 +161,14 @@ abstract class AbstractChatAttachmentWidget extends Disposable {
...
+				this.commandService.executeCommand('workbench.action.terminal.revealCommand', resource);
```
Tyriar: Something like this would be preferable imo:

```
this.terminalService.openResource(resource)
```

We generally don't want commands that aren't designed to be used in the command palette or keybindings imo. They are less safe ways of calling into services.

src/vs/workbench/contrib/chat/browser/actions/chatContext.ts
```diff
@@ -258,6 +261,60 @@ class ClipboardImageContextValuePick implements IChatContextValueItem {
...
 }
 
+export class TerminalContext implements IChatContextValueItem {
+
+	readonly type = 'valuePick';
+	readonly icon = Codicon.terminal;
+	readonly label = localize('terminal', 'Terminal');
+	constructor(private readonly _resource: URI, @ITerminalService private readonly _terminalService: ITerminalService) {
+
+	}
```
Tyriar, line 271: I haven't tested but is there a way to control the tooltip? Truncating the output could be interesting here?

Tyriar (dismissed):

src/vs/workbench/contrib/terminal/browser/terminalActions.ts
meganrogge: the `revealCommand` function is already a little hard to read. Tried adding `| URI` and don't think we should do that as it complicates the function.

src/vs/workbench/contrib/chat/browser/actions/chatContext.ts
meganrogge: I don't see that here https://github.com/microsoft/vscode/blob/fe1e01eb479399b8332ac535629b6b19f233b3ff/src/vs/workbench/contrib/chat/common/chatVariableEntries.ts#L20-L37  but files show a tooltip, so there must be a way

meganrogge: looks like the file one uses a custom hover. trying to figure out how

meganrogge: need to add our own widget to control this, doing so

src/vs/workbench/contrib/chat/browser/actions/chatContext.ts
```diff
@@ -258,6 +261,73 @@ class ClipboardImageContextValuePick implements IChatContextValueItem {
...
+	async isEnabled(widget: IChatWidget) {
```
Tyriar, line 272: Doesn't need async

src/vs/workbench/contrib/chat/browser/actions/chatContext.ts
```diff
@@ -258,6 +261,73 @@ class ClipboardImageContextValuePick implements IChatContextValueItem {
...
+		const terminal = this._terminalService.getInstanceFromResource(this._resource);
+		return !!widget.attachmentCapabilities.supportsTerminalAttachments && terminal?.isDisposed === false;
+	}
+	async asAttachment(widget: IChatWidget): Promise<IChatRequestVariableEntry | undefined> {
```
Tyriar, line 276: Doesn't need async

src/vs/workbench/contrib/chat/browser/media/chat.css
```diff
@@ -2277,6 +2277,19 @@ have to be updated for changes to the rules above, or to support more deeply nes
...
+	font-family: var(--monaco-monospace-font);
+	white-space: pre-wrap;
+	word-break: break-word;
+	margin: 4px 0;
+}
+
+.chat-attached-context-hover .chat-terminal-command-output {
+	max-height: 200px;
+	overflow: auto;
+}
```
Tyriar, line 2291: Other change leaking into this one?

Yoyokrazy (approved):

src/vs/workbench/contrib/terminal/browser/terminalUri.ts
```diff
@@ -15,14 +21,17 @@ export function parseTerminalUri(resource: URI): ITerminalIdentifier {
...
 }
 
-export function getTerminalUri(workspaceId: string, instanceId: number, title?: string): URI {
+export function getTerminalUri(workspaceId: string, instanceId: number, title?: string, commandId?: string): URI {
+
 	return URI.from({
 		scheme: Schemas.vscodeTerminal,
 		path: `/${workspaceId}/${instanceId}`,
 		fragment: title || undefined,
+		query: commandId
```
Tyriar, line 30: The query string should be `command=${commandId}`, just a guid or number after `?` isn't a valid URI and will cause problems later

src/vs/workbench/contrib/chat/browser/chatAttachmentWidgets.ts
```diff
@@ -249,6 +258,118 @@ export class FileAttachmentWidget extends AbstractChatAttachmentWidget {
...
+	attachment: ITerminalVariableEntry,
+	ariaLabel: string,
+	hoverService: IHoverService,
+	clickHandler: () => Promise<void>
+): IDisposable {
+	const disposable = new DisposableStore();
+	element.ariaLabel = ariaLabel;
+	element.style.cursor = 'pointer';
+
+	const pillIcon = dom.$('div.chat-attached-context-pill', {}, dom.$('span.codicon.codicon-terminal'));
```
Tyriar, line 310: Should use icon helpers to get the class instead of hard coding it. Should also be a theme icon so users can style it.

src/vs/workbench/contrib/chat/browser/chatAttachmentWidgets.ts
```diff
@@ -249,6 +258,118 @@ export class FileAttachmentWidget extends AbstractChatAttachmentWidget {
...
+	const textLabel = dom.$('span.chat-attached-context-custom-text', {}, attachment.command);
+	element.appendChild(pillIcon);
+	element.appendChild(textLabel);
+
+	disposable.add(dom.addDisposableListener(element, dom.EventType.CLICK, e => {
+		if ((e.target as HTMLElement | null)?.closest('.monaco-button')) {
+			return;
+		}
```
Tyriar, line 318: This looks like a hack, aren't you attaching it to the right `element` so you don't need to check this?

Tyriar (commented):
Wasn't working when I tested it:

Image

src/vs/workbench/contrib/chat/browser/actions/chatContext.ts
meganrogge: I was just following suit here

meganrogge: needs to match this https://github.com/microsoft/vscode/blob/3ccc52f9fcf0c77f589b0b3bc5b0efc7710437a8/src/vs/workbench/contrib/chat/browser/chatContextPickService.ts#L42

src/vs/workbench/contrib/chat/browser/chatAttachmentWidgets.ts
```diff
@@ -91,6 +93,7 @@ abstract class AbstractChatAttachmentWidget extends Disposable {
 		protected readonly currentLanguageModel: ILanguageModelChatMetadataAndIdentifier | undefined,
 		@ICommandService protected readonly commandService: ICommandService,
 		@IOpenerService protected readonly openerService: IOpenerService,
+		@ITerminalService protected readonly terminalService?: ITerminalService,
```
Tyriar, line 96: Accidental `?` ?

src/vs/workbench/contrib/chat/browser/actions/chatContext.ts
Tyriar: Oh it must be a promise ok, thanks for fixing the enabled one

PR #272994, opened 2025-10-23
expose history to chat session agents

joshspicer:


joshspicer:
@copilot this doesn't expose the history entire chat history of a conversation to the chat participant -why?  Fix that (and any other build errors from the latest merge)

Issue #232258, opened 2024-10-25
Powershell 7 Ctrl + G from Terminal fails for paths with space


albertosantini:
Does this issue occur when all extensions are disabled?: Yes/No

- VS Code Version: 1.94.2
- OS Version: Windows 11

Steps to Reproduce:

1. Open Terminal with Powershell `7.4.6`
2. mkdir "TEST TEST"
3. cd "TEST TEST"
4. cd -
5. Ctrl + G
6. Select a directory to go: `___ path ___\TEST TEST`
7. In the Terminal you see `cd & '___ path ___\TEST TEST'`

Changing directory fails due to the space in path: the command contains a weird `&`.
Other paths, withous spaces, correctly work.

Any hint?

meganrogge:
does this work on an external terminal?

albertosantini:
No, it doesn't work on external Terminal.

I don't see how it can work: : `cd & any_path` is executed, but it means another thing.

The point is the following: 
- when I select a path, via `Ctrl + G`, without any space, in the terminal I see the command: `cd 'C:\my\path\test'`
- when I select a path, via `Ctrl + G`, with a space, in the terminal I see the command: `cd & 'C:\my\path\test test'`

That character `&` is not correct.

meganrogge:
Given it does not work on an external terminal, closing this as it's not an issue with VS Code.

albertosantini:
No. :)

The problem is the shortcut Ctrl + G, key binding used in the terminal inside VS Code to change directory.

It "creates" a wrong command for the terminal.
It is a VS Code affair, not the fail of the wrong command in the external terminal.

albertosantini:
`cd & 'C:\my\path\test test'` is a wrong command for VS Code terminal and external Terminal.
`cd'C:\my\path\test test'` is correctly executed on VS Code terminal and external Terminal.

`cd & 'C:\my\path\test test'` is created only in VS Code when the terminal has focus, typing Ctrl + G displays a combo box with paths (`select a directory to go`).

Better explanation?

albertosantini:
I dig the code.

Here the code of the combo box, displaying the list of the paths.
`src/vs/workbench/contrib/terminalContrib/history/browser/terminalRunRecentQuickPick.ts`

Basically here it is generated the command to be executed on terminal. The path is prepared for changing folder.
```
// line 300
if (type === 'cwd') {
	text = `cd ${await instance.preparePathForShell(result.rawLabel)}`;
```

[...]

PR #273008, opened 2025-10-23
Deduplicate and format entries in terminal's Go To quick pick

### LINKED ISSUE #157932: Terminal go to recent directory should standardize path formats
Currently various paths formats are added depending on the shell. For example pwsh, wsl(bash) and git bash are shown below for the same folder:

image

We should convert the paths into a URI and store them in a consistent format, then when go to recent directory is triggered the path can be changed back to the right format for the shell in question.

This is technically an all platform problem but in practice it would probably only be seen on Windows.

dmitrivMS:
Fix for #157932: This will address duplicate entries and formatting in the quick pick list.

This does not address #232258 which was marked as a duplicate of 157932 because display strings formatting is different from strings sent to the shell (e.g. no quoting or escaping is necessary for display).


src/vs/workbench/contrib/terminal/browser/terminal.ts
```diff
@@ -1033,6 +1033,12 @@ export interface ITerminalInstance extends IBaseTerminalInstance {
 	 */
 	preparePathForShell(originalPath: string): Promise<string>;
 
+	/**
+	 * Formats a file system URI for display in UI so that it appears in the terminal shell's format.
+	 * @param uri The URI to format.
+	 */
+	formatUriForShellDisplay(uri: URI): Promise<string>;
```
Tyriar: Do these do the same thing? Could `preparePathForShell` take a `string | URI`? The name "for shell display" in particular is unclear to me.

src/vs/workbench/contrib/terminal/common/terminalEnvironment.ts
```diff
@@ -390,3 +389,26 @@ export function getWorkspaceForTerminal(cwd: URI | string | undefined, workspace
...
+				return backend.getWslPath(path.replace(/\//g, '\\'), 'win-to-unix');
+			} else {
+				return path.replace(/\\/g, '/');
+			}
+		} else if (shellType === WindowsShellType.GitBash) {
+			// Convert \ to / and replace 'c:\' with '/c/'.
+			return path.replace(/\\/g, '/').replace(/^([a-zA-Z]):\//, '/$1/');
+		} else {
+			// If the frontend is not Windows but the terminal is, convert / to \.
+			path = typeof uri === 'string' ? path : uriToFsPath(uri, true);
```
Tyriar, line 404: Is it necessary to do this?

src/vs/workbench/contrib/terminal/common/terminalEnvironment.ts
```diff
@@ -390,3 +389,26 @@ export function getWorkspaceForTerminal(cwd: URI | string | undefined, workspace
...
+
+export async function formatUriForShellDisplay(uri: URI | string, shellType?: TerminalShellType, os?: OperatingSystem, backend?: Pick<ITerminalBackend, 'getWslPath'>, isWindowsFrontend: boolean = isWindows): Promise<string> {
+	let path = typeof uri === 'string' ? uri : uri.fsPath;
+	if (os === OperatingSystem.Windows) {
+		if (shellType === WindowsShellType.Wsl) {
+			if (backend) {
+				return backend.getWslPath(path.replace(/\//g, '\\'), 'win-to-unix');
+			} else {
+				return path.replace(/\\/g, '/');
+			}
```
Tyriar: I think this will fail, but anyway backend will always exist after `onProcessReady` has fired. An easy fix that makes this a little better is to use `this._processManager.backend!` in terminal instance's `preparePathForShell` and `formatUriForShellDisplay`, then get rid of the backend condition.

src/vs/workbench/contrib/terminal/common/terminalEnvironment.ts
```diff
@@ -390,3 +389,26 @@ export function getWorkspaceForTerminal(cwd: URI | string | undefined, workspace
...
+			return path.replace(/\\/g, '/').replace(/^([a-zA-Z]):\//, '/$1/');
+		} else {
+			// If the frontend is not Windows but the terminal is, convert / to \.
+			path = typeof uri === 'string' ? path : uriToFsPath(uri, true);
+			return !isWindowsFrontend ? path.replace(/\//g, '\\') : path;
+		}
+	} else {
+		// If the frontend is Windows but the terminal is not, convert \ to /.
+		return isWindowsFrontend ? path.replace(/\\/g, '/') : path;
+	}
```
Tyriar, line 410: The newer replaceAll API is a little nicer:

```
path.replaceAll('\\', '/')
```

src/vs/workbench/contrib/terminal/test/common/terminalEnvironment.test.ts
```diff
@@ -319,4 +320,24 @@ suite('Workbench - TerminalEnvironment', () => {
...
+			strictEqual(await formatUriForShellDisplay('C:\\Foo\\Bar', WindowsShellType.CommandPrompt, OperatingSystem.Windows, wslPathBackend, true), 'C:\\Foo\\Bar');
+			strictEqual(await formatUriForShellDisplay('C:/Foo/Bar', WindowsShellType.CommandPrompt, OperatingSystem.Windows, wslPathBackend, false), 'C:\\Foo\\Bar');
+		});
+		test('Non-Windows backend', async () => {
+			strictEqual(await formatUriForShellDisplay('\\foo\\bar', PosixShellType.Bash, OperatingSystem.Linux, wslPathBackend, true), '/foo/bar');
+			strictEqual(await formatUriForShellDisplay('/foo/bar', PosixShellType.Bash, OperatingSystem.Linux, wslPathBackend, true), '/foo/bar');
+			strictEqual(await formatUriForShellDisplay('\\foo\\bar', PosixShellType.Bash, OperatingSystem.Linux, wslPathBackend, false), '\\foo\\bar');
+			strictEqual(await formatUriForShellDisplay('/foo/bar', PosixShellType.Bash, OperatingSystem.Linux, wslPathBackend, false), '/foo/bar');
+		});
+	});
```
Tyriar, line 352: Could these tests be organized into shell tests instead then test both windows and non-windows backends inside that?

src/vs/workbench/contrib/terminal/test/common/terminalEnvironment.test.ts
```diff
@@ -319,4 +320,24 @@ suite('Workbench - TerminalEnvironment', () => {
...
+			strictEqual(await formatUriForShellDisplay('c:/foo/bar', WindowsShellType.Wsl, OperatingSystem.Windows, undefined, false), 'c:/foo/bar');
+			strictEqual(await formatUriForShellDisplay('c:\\foo\\bar', WindowsShellType.GitBash, OperatingSystem.Windows, wslPathBackend, true), '/c/foo/bar');
+			strictEqual(await formatUriForShellDisplay('c:/foo/bar', WindowsShellType.GitBash, OperatingSystem.Windows, wslPathBackend, false), '/c/foo/bar');
+			strictEqual(await formatUriForShellDisplay('c:\\foo\\bar', WindowsShellType.CommandPrompt, OperatingSystem.Windows, wslPathBackend, true), 'c:\\foo\\bar');
+			strictEqual(await formatUriForShellDisplay('c:/foo/bar', WindowsShellType.CommandPrompt, OperatingSystem.Windows, wslPathBackend, false), 'c:\\foo\\bar');
+			strictEqual(await formatUriForShellDisplay('C:\\Foo\\Bar', WindowsShellType.CommandPrompt, OperatingSystem.Windows, wslPathBackend, true), 'C:\\Foo\\Bar');
+			strictEqual(await formatUriForShellDisplay('C:/Foo/Bar', WindowsShellType.CommandPrompt, OperatingSystem.Windows, wslPathBackend, false), 'C:\\Foo\\Bar');
```
Tyriar: Prefer testing pwsh to command prompt since it's the default.

src/vs/workbench/contrib/terminalContrib/history/browser/terminalRunRecentQuickPick.ts
```diff
@@ -212,9 +226,11 @@ export async function showRunRecentQuickPick(
 		const previousSessionItems: (IQuickPickItem & { rawLabel: string })[] = [];
 		// Only add previous session item if it's not in this session and it matches the remote authority
 		for (const [label, info] of history.entries) {
-			if ((info === null || info.remoteAuthority === instance.remoteAuthority) && !cwds.includes(label)) {
+			const itemUri = URI.file(label);
+			if ((info === null || info.remoteAuthority === instance.remoteAuthority) && !uniqueUris.has(itemUri)) {
```
Tyriar: See `TerminalInstance.getCwdResource` for the right way to get the URI when the terminal is remote:

https://github.com/microsoft/vscode/blob/281f94f487b53bb2274215538e36fb4062d38c33/src/vs/workbench/contrib/terminal/browser/terminalInstance.ts#L2230-L2245

src/vs/workbench/contrib/terminal/common/terminalEnvironment.ts
```diff
@@ -390,3 +389,26 @@ export function getWorkspaceForTerminal(cwd: URI | string | undefined, workspace
 	}
 	return workspaceFolder;
 }
+
+export async function formatUriForShellDisplay(uri: URI | string, shellType?: TerminalShellType, os?: OperatingSystem, backend?: Pick<ITerminalBackend, 'getWslPath'>, isWindowsFrontend: boolean = isWindows): Promise<string> {
+	let path = typeof uri === 'string' ? uri : uri.fsPath;
+	if (os === OperatingSystem.Windows) {
```
Tyriar: If we have an actual URI with scheme set properly (see my comment about `remoteAuthority` below), `LabelService` will give a good formatting of the URI instead and you won't need to check windows frontend.

Tyriar (changes requested):
Thanks for the PR! FYI some background here, URIs in the terminal historically haven't been used so much since we're not actually 100% sure what they are. The current best practice is to use `TerminalInstance.getCwdResource` to pull the cwd now which handles it properly for remotes.

This PR could be split into a few smaller chunks, for example supporting actual URIs in `CwdDetectionCapability` with a separate event if needed.

src/vs/workbench/contrib/terminal/browser/terminal.ts
dmitrivMS: No, they are different - formatting for UI display does not include escaping of quoting since that doesn't look nice in the UI. What would be a better name?

src/vs/workbench/contrib/terminal/common/terminalEnvironment.ts
dmitrivMS: Yeah, in order to preserve drive letter casing - i.e. we want C:\ instead of c:\ and .fsPath will lowercase it whereas uriToFsPath will preserve.

src/vs/workbench/contrib/terminal/common/terminalEnvironment.ts
dmitrivMS: Good to know, will fix.

src/vs/workbench/contrib/terminal/common/terminalEnvironment.ts
dmitrivMS: I like it! Will fix

src/vs/workbench/contrib/terminal/test/common/terminalEnvironment.test.ts
dmitrivMS: Will do

src/vs/workbench/contrib/terminal/test/common/terminalEnvironment.test.ts
dmitrivMS: Will do

src/vs/workbench/contrib/terminal/common/terminalEnvironment.ts
dmitrivMS: I may need more context on this - LabelService does not respect shell type as I do here. What would be an example of a remoteAuthority URI?

src/vs/workbench/contrib/terminal/browser/terminal.ts
dmitrivMS: Would `getUriLabelForShell` be more clear?

dmitrivMS: renamed

src/vs/workbench/contrib/terminalContrib/history/browser/terminalRunRecentQuickPick.ts
dmitrivMS: fixed

src/vs/workbench/contrib/terminal/common/terminalEnvironment.ts
Tyriar: This is for remote connected windows. It might not actually be a problem, I think directories might be cached per workspace as I only see WSL paths in my [WSL](https://marketplace.visualstudio.com/items?itemName=ms-vscode-remote.remote-wsl) window

image

Tyriar (approved):
Thanks!

PR #273073, opened 2025-10-24
Increase codicon font-size and adjust padding in SCM history view

mrleemurray:
Enhance the visibility of codicons in the SCM history view by increasing their font size to 16px and reducing padding to 1px. This improves the overall user experience and consistency across various components.

Before:
image

After:
image

### DIFF
--- a/src/vs/workbench/contrib/scm/browser/media/scm.css
+++ b/src/vs/workbench/contrib/scm/browser/media/scm.css
@@ -205,7 +205,7 @@
 }
 
 .scm-view .monaco-list-row .history-item > .label-container > .label > .codicon {
-	font-size: 14px;
+	font-size: 16px;
 	color: inherit !important;
 	padding: 2px;
 }
--- a/src/vs/workbench/contrib/scm/browser/media/scm.css
+++ b/src/vs/workbench/contrib/scm/browser/media/scm.css
@@ -205,7 +205,6 @@
 }
 
 .scm-view .monaco-list-row .history-item > .label-container > .label > .codicon {
-	font-size: 16px;
 	color: inherit !important;
 	padding: 1px;
 }
@@ -572,7 +571,6 @@
 
 .monaco-hover.history-item-hover .hover-row.status-bar .action .codicon {
 	color: inherit;
-	font-size: 16px;
 }
 
 .monaco-hover.history-item-hover .markdown-hover .hover-contents:not(.code-hover-contents):not(.html-hover-contents) span.codicon {
@@ -619,7 +617,6 @@
 
 .monaco-toolbar .action-label.scm-graph-repository-picker .codicon,
 .monaco-toolbar .action-label.scm-graph-history-item-picker .codicon {
-	font-size: 16px;
 	color: inherit;
 }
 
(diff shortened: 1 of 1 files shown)


src/vs/workbench/contrib/scm/browser/media/scm.css
```diff
@@ -205,9 +205,9 @@
 }
 
 .scm-view .monaco-list-row .history-item > .label-container > .label > .codicon {
-	font-size: 14px;
+	font-size: 16px;
```
lszomoru: Instead of specifying `16px` which is the default font size, I think that we should just remove the `font-size: 14px` altogether.

src/vs/workbench/contrib/scm/browser/media/scm.css
```diff
@@ -572,7 +572,11 @@
 
 .monaco-hover.history-item-hover .hover-row.status-bar .action .codicon {
 	color: inherit;
-	font-size: 12px;
+	font-size: 16px;
```
lszomoru: Same as above

src/vs/workbench/contrib/scm/browser/media/scm.css
```diff
@@ -615,7 +619,7 @@
 
 .monaco-toolbar .action-label.scm-graph-repository-picker .codicon,
 .monaco-toolbar .action-label.scm-graph-history-item-picker .codicon {
-	font-size: 14px;
+	font-size: 16px;
```
lszomoru: Same as above

src/vs/workbench/contrib/scm/browser/media/scm.css
```diff
@@ -572,7 +572,11 @@
 
 .monaco-hover.history-item-hover .hover-row.status-bar .action .codicon {
 	color: inherit;
-	font-size: 12px;
+	font-size: 16px;
+}
+
+.monaco-hover.history-item-hover .markdown-hover .hover-contents:not(.code-hover-contents):not(.html-hover-contents) span.codicon {
+	font-size: 16px;
```
lszomoru: Why do we need to set this?

lszomoru (changes requested):

src/vs/workbench/contrib/scm/browser/media/scm.css
mrleemurray: `.monaco-hover.workbench-hover.compact` in `hover.css` specifies the font-size as `12px`, which overrides the codicon sizes. Should we add the `16px` icon size there instead?

lszomoru: The git graph hover should honor the compact hover settings. 
In order to do that I think that the graph hover class should not be setting any font sizes for codicons.

mrleemurray: Okay - so if we were to update the icon sizes it would happen in the compact hover settings. I'm not sure where else in the product it is used, so should we hold off updating that widget until we have a better understanding?

lszomoru: Yes, I would get confirmation from Daniel that it is expected the the compact hover sets **all** fonts to `12px`.

lszomoru (approved):

PR #273074, opened 2025-10-24
Add dotenv support

### LINKED ISSUE #267333: Add support for dotenv
The dotenv format uses subset of the bourne shell syntax to define key-value pairs.

```dotenv
# This is a comment
SOME_VARIABLE=some value  # Another comment
another_variable="A value in a quoted string"
YET_another_variable='A value with a # sign in it'
```

Unlike other languages, files of this type are not recognizable by a name **suffix** (file extension). Instead, these files are typically recognizable by a name **prefix**. Some popular file names include:

- `.env`
- `.env.development`
- `.env.production`
- `.env.test`
- `.env.local`
- `.env.development.local`
- `.env.production.local`
- `.env.test.local`
- `.flaskenv`
- `user-dirs.dirs`

This format is supported by many popular environments, including the Node.js, Next.js, Flask, and the XDG user dir specification.

There are two popular existing Dotenv extensions:

- [DotENV](https://marketplace.visualstudio.com/items?itemName=mikestead.dotenv)
- [Dotenv Official +Vault](https://marketplace.visualstudio.com/items?itemName=dotenv.dotenv-vscode)

The most popular is _DotENV_. It doesn’t specify the correct file patterns. Instead, it defines [extensions](https://github.com/dotenv-org/dotenv-vscode/blob/v0.28.1/package.json#L109-L162). This extension was last updated 7 years ago.

The extension _Dotenv Official +Vault_ looks more official. It has the same problem as _DotENV_, but additionally it writes to user settings as a workaround. This is solved by https://github.com/dotenv-org/dotenv-vscode/pull/105, but this has gone unnoticed for over a year.

Both projects appear to be unmaintained. Given how commonly dotenv files are used, I think it would be good to include basic dotenv support in VSCode as a builtin extension.

### LINKED ISSUE #267355: Add Native Syntax Highlighting Support for .env Files in VS Code
Type: <b>Feature Request</b>

VS Code currently lacks built-in syntax highlighting for .env files, which are simple, widely used configuration files for environment variables. Users must rely on third-party extensions to get basic highlighting, which is inconvenient for such a common format. Adding native support would improve the editing experience by automatically recognizing key-value pairs, comments, and basic structure without requiring additional plugins. This would align with VS Code's goal of providing out-of-the-box support for popular file types and reduce dependency on external tools.

VS Code version: Code - Insiders 1.105.0-insider (b41ed0096e1ed219d4740ed3a4dc40531eef3a8b, 2025-09-17T23:32:07.124Z)
Modes:

remcohaszing:
This adds basic language support for the dotenv language. This includes syntax highlighting and a language configuration file.

The dotenv language is applied to files with the `.env` extension, files named `.env`, `.flaskenv`, or `user-dirs.dirs`, and file names starting with `.env.`.

The grammar is downloaded from https://github.com/dotenv-org/dotenv-vscode.

Closes #267333
Closes #267355


extensions/dotenv/cgmanifest.json
```diff
@@ -0,0 +1,40 @@
...
+				"THE SOFTWARE IS PROVIDED \"AS IS\", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR",
+				"IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,",
+				"FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE",
+				"AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER",
+				"LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,",
+				"OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE",
+				"SOFTWARE."
+			],
+			"license": "MIT License",
+			"version": "0.0.0"
```
remcohaszing: I don’t know the meaning of this number. It exists in other `cgmanifest.json` files I just set it to `0.0.0`

alexr00: It's grabbed from the package.json of the upstream repo when the update-grammar script is run.

remcohaszing:
I noticed this too. It’s a bug in the JSON schema. This affects all `language-configuration.json` files. I tried tweaking this line, but it doesn’t seem to make a difference.

https://github.com/microsoft/vscode/blob/c7e7ef5444582260f5b0ac1387a68649b0fa97c0/src/vs/workbench/contrib/codeEditor/common/languageConfigurationExtensionPoint.ts#L550

extensions/dotenv/package.json
```diff
@@ -0,0 +1,48 @@
...
+        "extensions": [
+          ".env"
+        ],
+        "filenames": [
+          ".env",
+          ".flaskenv",
+          "user-dirs.dirs"
+        ],
+        "filenamePatterns": [
+          ".env.*"
```
alexr00, line 28: This conflicts with shellscript: 

https://github.com/remcohaszing/vscode/blob/3cb630867018c9dec57d748c8b75b05966e33bf5/extensions/shellscript/package.json#L72-L74

extensions/dotenv/package.json
```diff
@@ -0,0 +1,48 @@
...
+  "categories": ["Programming Languages"],
+  "contributes": {
+    "languages": [
+      {
+        "id": "dotenv",
+        "extensions": [
+          ".env"
+        ],
+        "filenames": [
+          ".env",
```
alexr00, line 23: This conflicts with ini:

https://github.com/remcohaszing/vscode/blob/3cb630867018c9dec57d748c8b75b05966e33bf5/extensions/ini/package.json#L42-L45

alexr00 (commented):
Thanks for the PR. I have a few comments. Not sure if we want to conflict with shellscript and ini.

extensions/dotenv/package.json
remcohaszing: It looks like 3 ([also here](https://github.com/remcohaszing/vscode/blob/3cb630867018c9dec57d748c8b75b05966e33bf5/extensions/ini/package.json#L49)) individuals marked dotenv files as `shellscript` / `properties` in 3 separate commits due to the lack of a proper builtin dotenv language. Should I remove them?

alexr00: Oh, that's funny. Thanks for looking at the history!

alexr00 (commented):
Alright, one last question: why is this change better than using the existing shellscript grammar to highlight dotenv files?

alexr00 (approved):

isidorn (approved):

Issue #272922, opened 2025-10-23
[Markdown] Do not auto close angle brackets inside latex blocks


dragon-archer:
Does this issue occur when all extensions are disabled?: No

This issue is part of the bundled `markdown-basics` extension.

- VS Code Version: 1.105.0
- OS Version: Windows 11 24H2 26100.6899

Steps to Reproduce:

1. Open an arbitary markdown file
2. Type a `<` inside `$ ... $` or `$$ ... $$`
3. A `>` will be added as part of the auto-closing
4. However, HTML tags are not allowed inside latex blocks, and when people use latex blocks, mostly they will use `<` as **less than** meaning instead of **left angle bracket**, so it should not be auto-closed

dragon-archer:
After some invesgeting, I think perhaps we can simply move every textmate scope inside `markup.math.*.markdown` to `markup.math.string.*.markdown`, thus the `Standard Token Type` of these content will change from `Other` to `String`, which will disable the annoying autoclosing behaviour as declared in [language-configuration.json](https://github.com/microsoft/vscode/blob/578fc30603d5bfb73815ea11bdfbbbdb8db4050d/extensions/markdown-basics/language-configuration.json#L42)

Or, we may pick another standard token type from [the guide](https://code.visualstudio.com/api/language-extensions/semantic-highlight-guide#semantic-token-classification), and add it to the `notIn` list of `language-configuration.json`. Sadly it seemes that none of the token types in the guide fit the math block well.

RedCMD:
actually is dup:
* https://github.com/microsoft/vscode/issues/133397

PR #273094, opened 2025-10-24
Fix: Prevent auto-closing of angle brackets in LaTeX blocks

Rishi1878:
This PR addresses [issue #272922](https://github.com/microsoft/vscode/issues/272922) by preventing auto-closing of `< >` inside LaTeX blocks in Markdown.

### Changes:
- Modified `language-configuration.json` to include `"latex"` in the `notIn` array for `< >` auto-closing.
- Scoped LaTeX blocks in `markdown.tmLanguage.json` using a new `markup.math.latex.markdown` rule.
- Verified that `<` is no longer auto-closed inside `$<a + b>$` and `$$<x = y>$$`.

### Testing:
Tested in GitHub Codespaces using a Markdown file with LaTeX blocks. Confirmed correct token scopes and behavior.
Let me know if further adjustments are needed!

### DIFF
--- a/extensions/markdown-basics/language-configuration.json
+++ b/extensions/markdown-basics/language-configuration.json
@@ -39,7 +39,7 @@
 			"open": "<",
 			"close": ">",
 			"notIn": [
-				"string"
+				"string", "latex"
 			]
 		},
 	],
--- a/extensions/markdown-basics/syntaxes/markdown.tmLanguage.json
+++ b/extensions/markdown-basics/syntaxes/markdown.tmLanguage.json
@@ -25,6 +25,9 @@
 					"include": "#heading"
 				},
 				{
+        			"include": "#latex"
+      			},
+				{
 					"include": "#blockquote"
 				},
 				{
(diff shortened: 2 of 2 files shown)


RedCMD:
but this doesn't work?

only `string` and `comment` are supported
image

and now the language assignment is broken
image
should be `latex` not `markdown`

extensions/markdown-basics/language-configuration.json
```diff
@@ -39,7 +39,7 @@
 			"open": "<",
 			"close": ">",
 			"notIn": [
-				"string"
+				"string", "latex"
```
mjbvz, line 42: `notIn` only supports a limited set of scopes such as `string`. `latex` won't work

extensions/markdown-basics/syntaxes/markdown.tmLanguage.json
```diff
@@ -50,6 +53,12 @@
 				}
 			]
 		},
+		"latex": {
```
mjbvz, line 56: We should already parse latex using this injection grammar: https://github.com/microsoft/vscode/blob/3a52543b01e83598eb60089946e0d3bcebb5b082/extensions/markdown-math/syntaxes/md-math-block.tmLanguage.json#L3

mjbvz (changes requested):

Issue #273052, opened 2025-10-24
Quick open: files picker shows description aligned to the right


bpasero:
OLD:

<img width="725" height="296" alt="Image" src="https://github.com/user-attachments/assets/8b24a6cc-0b8c-4e01-b224-9681d1082cfc" />

NEW:

<img width="711" height="590" alt="Image" src="https://github.com/user-attachments/assets/2e25142e-2329-40b7-b8a8-0febf21b5799" />

PR #273098, opened 2025-10-24
fix: file-found Badge vertical

### LINKED ISSUE #235159: Number-of-files-found Badge is not at the center of the list item vertically
Testing #234889

## Actual

Image

## Expected

Image

Version: 1.96.0-insider
Commit: 5d6531aabba3abfbc263c1e17e965db97df9210f
Date: 2024-12-03T05:13:15.471Z
Electron: 32.2.6
ElectronBuildId: 10629634
Chromium: 128.0.6613.186
Node.js: 20.18.1
V8: 12.8.374.38-electron.0
OS: Darwin arm64 24.1.0

ohah:
Fixed https://github.com/microsoft/vscode/issues/235159

This caused a regression in https://github.com/microsoft/vscode/issues/273052, so I've fixed the specific line to avoid side effects.

Fixes #235159
Addresses #273052

FIXED ScreenShot

스크린샷 2025-10-24 오후 9 10 19

스크린샷 2025-10-24 오후 9 13 10


src/vs/workbench/contrib/files/browser/media/explorerviewlet.css
```diff
@@ -64,6 +64,7 @@
 
 .explorer-folders-view .monaco-list-row[aria-expanded="false"] .explorer-item.highlight-badge .monaco-count-badge {
 	display: inline-block;
+	transform: translateY(-1.5px);
```
benibenj: can you explain the math behind this? How do you get to 1.5px?

ohah: The original value was -2px because the .monaco-tl-contents element has a height of 22px and a line height of 22px.

However, since .monaco-count-badge has a line-height of 11px, it was slightly misaligned (0.5px too high), so I adjusted the value to -1.5px instead.

I think using flex would be the cleanest approach, but it might affect properties like text-overflow: ellipsis or other layout behavior, so I only moved this specific part.

스크린샷 2025-10-25 오전 1 07 46
스크린샷 2025-10-25 오전 1 08 01

스크린샷 2025-10-25 오전 1 17 08

benibenj (dismissed):

src/vs/workbench/contrib/files/browser/media/explorerviewlet.css
Tyriar: @benibenj using the `.5` here is likely to cause blurriness on low DPI displays. I'm unsure whether the use of transform here will create a new render layer and the overhead associated with that.

Tyriar (dismissed):

dmitrivMS (approved):

ohah:
@dmitrivMS thx

alexr00 (approved):

PR #273111, opened 2025-10-24
Improve performance of UriIdentityService (#273108)

### LINKED ISSUE #273108: Performance: UriIdentityService uses non-optimal caching strategy

- VS Code Version:  1.105.1
- OS Version: MacOS 15.7.1

[UriIdentityService](https://github.com/microsoft/vscode/blob/main/src/vs/platform/uriIdentity/common/uriIdentityService.ts) is heavily used to normalize URIs. While this code is not very heavy it sits on a critical path in many situation and already caused significant performance problems before. See [this issue](https://github.com/microsoft/vscode/issues/194853) as an example when a user had more then 1 million items in cache.

Currently [UriIdentityService] has some room for improvement:

1. For caching it uses `SkipList` with `extUri.compare` as a comparator. Compare method currently [creates comparison key](https://github.com/microsoft/vscode/blob/f958c6f0e6e9e862c7b6ba438e256a1b0488d4be/src/vs/base/common/resources.ts#L147) for each URI before comparison. It leads to creation of [2 new URIs](https://github.com/microsoft/vscode/blob/f958c6f0e6e9e862c7b6ba438e256a1b0488d4be/src/vs/base/common/resources.ts#L160-L164) and stringifying them which is also [non-trivial](https://github.com/microsoft/vscode/blob/f958c6f0e6e9e862c7b6ba438e256a1b0488d4be/src/vs/base/common/uri.ts#L649-L720). While each individual operation is not very heavy they can build up due to huge amount of them. With (almost) full cache get/set operation will require around 35 comparisons (see SkipList [source](https://github.com/microsoft/vscode/blob/f958c6f0e6e9e862c7b6ba438e256a1b0488d4be/src/vs/base/common/skipList.ts) and [paper](https://www.epaperpress.com/sortsearch/download/skiplist.pdf)). Which means around 70 URI creations and toString calls in `getComparisonKey`. All together this leads to high costs of reading/writing many items from the list.

2. It is important to mention that huge amount of `getComparisonKey` is needed not only for insertion but also for reads, making reading from the cache almost as expensive operation as insertion.

3. While one can notice problems (1) and (2) only when reading/writing huge amounts of items in cache (thousands of items) which should not happen very often, such a situation actually happens when cache overflows. The current implementation sorts all existing items in cache in chronological order, clears the cache and [reinsert items back to cache](https://github.com/microsoft/vscode/blob/edff076b3ae7e6147430f32440818068fe3bf145/src/vs/platform/uriIdentity/common/uriIdentityService.ts#L109-L113) one by one. With the current limit it is 32768 insertions. It can take up to 200-300 ms

4. Last tiny problem is that current implementation of `UriIdentityService` reverses cache age on cleanup. First it sorts items in descending order (max time === newest first), then resets cache timer, and then reinserts first half of the resulting array into new cache starting from the head of the list. So the most recent item in the old cache gets 0 time in a new cache, so it becomes the oldest. But in fact this is not important because we are not specifically interested in relative timing for items left in cache. Since this is exactly half of the cache, and after cleanup this is by definition the oldest half all of them. If none of them are touched they all together will be removed anyway. And if some of them are touched they will get updated time.

yavanosta:
This CL improves perfrormance of UriIdentityService:

1. Migrates to string keys and native Map for cache to reduce amount of URI objects creation/conversion to string. 
2. Improves cache cleanup algorithm by using `quickSelect` to pick median value instead of sorting the entire cache.

Changes measured on my local machine (MacBook Air M4):

1.  Case 1: inserting 1000 URIs on almost full cache (has 64000 items): `153ms` -> `18ms`
2. Case 2: reading 50000 URIs from cache on almost full cache  (has 65000 items): `297ms` -> `81ms`
3. Case 3: cache trim `297ms` -> `17ms`

Fixes #273108


src/vs/platform/uriIdentity/test/common/uriIdentityService.test.ts
```diff
@@ -102,6 +102,30 @@ suite('URI Identity', function () {
 		assertCanonical(URI.parse('foo://bar/BAZZ#DDD'), b.with({ fragment: 'DDD' })); // lower-case path, but fragment is kept
 	});
 
+	test('cleares cache when overflown with respect to access time', () => {
```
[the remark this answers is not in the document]
yavanosta: Fixed

src/vs/platform/uriIdentity/common/uriIdentityService.ts
```diff
@@ -92,24 +139,18 @@ export class UriIdentityService implements IUriIdentityService {
...
+				entry.time = 0;
 			}
-		});
-
-		Entry._clock = 0;
-		this._canonicalUris.clear();
-		const newSize = this._limit * 0.5;
-		for (let i = 0; i < newSize; i++) {
-			this._canonicalUris.set(entries[i][0], entries[i][1].touch());
 		}
```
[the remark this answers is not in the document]
yavanosta: Not sure if I can correctly understand this AI generated commit. The logic behind the change if the following:

1. We calculate median based on current items
2. Then we remove items with time <= median and reset it to `0` for those which left.
3. This logic utilizes the fact that we clean up exactly the half of the cache.
4. All remain items will have time `0`.
5. On next cleanup all new items will have time >= 1. Items left from previous cleanup will have time === 0 if they have not been touched and >= 1 if they have been touched. If they have not been touched we do not care in their order, since all of them will be trimmed. 
6. So the algorithm guarantees that amount of items with time === 0 is always strictly less then a half, and this is enough for correct trim.

src/vs/platform/uriIdentity/common/uriIdentityService.ts
```diff
@@ -22,49 +22,95 @@ class Entry {
 	}
 }
 
-export class UriIdentityService implements IUriIdentityService {
+interface IFileSystemCasingChangedEvent {
+	scheme: string;
+}
+
+class PathCasingCache extends Disposable {
```
jrieken: Extracting `PathCasingCache` is just cosmetic and to your personal preference or am I missing some "real" change here?

src/vs/platform/uriIdentity/common/uriIdentityService.ts
```diff
@@ -22,49 +22,95 @@ class Entry {
...
 
-	dispose(): void {
-		this._dispooables.dispose();
-		this._canonicalUris.clear();
+	private _handleFileSystemCasingChanged(e: IFileSystemCasingChangedEvent): void {
+		for (const [key, entry] of this._canonicalUris.entries()) {
+			if (entry.uri.scheme !== e.scheme) {
+				continue;
+			}
+			this._canonicalUris.delete(key);
```
jrieken: This is new behaviour, right? So when the FS casing has changes old canonical uris are dropped

src/vs/platform/uriIdentity/common/uriIdentityService.ts
```diff
@@ -92,24 +139,21 @@ export class UriIdentityService implements IUriIdentityService {
...
+			(a, b) => a - b);
+		for (const [key, entry] of this._canonicalUris.entries()) {
+			// Its important to remove the median value here (<= not <).
+			// If we have not touched any items since the last trim, the
+			// median will be 0 and no items will be removed otherwise.
+			if (entry.time <= median) {
+				this._canonicalUris.delete(key);
 			} else {
-				return 0;
+				entry.time = 0;
```
jrieken, line 122: Wouldn't we before `touch` all surviving items (after resetting the clock) and hence they would have times `1`, `2`, `3`, etc. But now all surviving items have time `0`

jrieken (changes requested):
Can we leave cosmetic changes (formatting, extracting code) out of this PR. It makes reviewer harder and only serves personal preferences. Thanks

src/vs/platform/uriIdentity/common/uriIdentityService.ts
yavanosta: Hi Johannes. Thanks for the thoughtful review!

This is mixture of cosmetic and real change. Let me try to clarify this and [the next comment](https://github.com/microsoft/vscode/pull/273111#discussion_r2510833090).

The current behavior works in the following way:
1. We create a `new ExtUri(ignorePathCasing)` passing there `ignorePathCasing` callback as a parameter
2. On the `UriIdentityService` side we maintain a map that keeps track of case sensitivity for each file system
3. Since SkipList uses `this.extUri` for each comparison it automatically uses the latest version of case-sensitivity state

However with the new pre-generated string keys situation is different:
1. We generate keys on insertion using the current state of case-sensitivity
2. On read we do NOT read the state of case-sensitivity, so we can read wrong URI if case-sensitivity for a certain file system has changed
3. To fix this, there are two options: regenerate keys or simply clean them up from cache and let them to be regenerated on the next read
4. Since in comparison to just updating boolean (what happens today), both of that options is relatively expensive we don't want to do that on each `onDidChangeFileSystemProviderCapabilities` event, since it may be related to some other capability. So we only do that when case sensitivity changes.
5. Overall this logic became slightly more complicated. Today: update boolean on every event. Now: maintain case-sensitivity state and clean up the main cache if it changes. So I decided that it would make sense to extract all case-sensitivity tracking logic into a separate class to keep each class focused on they primary purpose. Case sensitivity cache - track case sensitivity state and inform the main class when it changes. The main class maintain URI cache and (partially) clear it when needed.

Does this split make sense from your perspective?

src/vs/platform/uriIdentity/common/uriIdentityService.ts
yavanosta: See response here. I tried to cover both comments in [a single response](https://github.com/microsoft/vscode/pull/273111#discussion_r2513406499).

src/vs/platform/uriIdentity/common/uriIdentityService.ts
yavanosta: That's correct but this is not needed.

First of all, as I described in [the issue](https://github.com/microsoft/vscode/issues/273108) currently we do that in reverse order, so the newest item becomes oldens in cleaned up cache.

See [this code snippet](https://github.com/yavanosta/vscode/blob/ec7952e66e36baa31f04facc9bfd680dfd4f226a/src/vs/platform/uriIdentity/common/uriIdentityService.ts#L95-L114). We sort items in descending order (newest first) and then re-insert them from item 0 (the newest item) to the middle of the array into new cache. So the newest item becomes the oldest in the new cache.

However while this probably is a mistake it does NOT cause any issues. And the reason is - we clean up exactly half of the cache. Let me explain.

1. When we first fill the cache `time` for each item matters since when we will have to trim the cache we will need to find the "oldest" half of items in cache.
2. Situations gets more interesting when we already have trimmed cache once. Lets call all items that left in cache "generation 1". All items that are inserted after that lets call "generation 2".
3. If we touch an item from "generation 1" it will move it automatically to "generation 2".
4. When it's time to trim cache again, we can guarantee that: (a) all "generation 1" items are older then "generation 2" and (b) "generation 1" is not more then half of the cache.
5. Then we can state the following: we should remove all items from "generation 1" and as many items from "generation 2" as needed to trim half of the cache.
6. You can notice that we are NOT interested in relative position of "generation 1" items. We don't care which one of them is olden then another one. If it is still in "generation 1" till next trim - it is removed.
7. It means we can safely omit information about their relative timing as long as we can distinguish "generation 1" and "generation 2".
8. This is exactly what happens here. All items left after first trim will have `time = 0` and this is a marker for "generation 1". All new items will have `time > 1` this is a marker for "generation 2". If any item with `time = 0` is read, it's time will be increased, so it will be moved to "generation 2".
9. Essentially it is exactly **the same as how the code works today**, the only difference is that "generation boundary" will be at `newSize` and this is exactly the reason that the code works currently despite reinserting items in reverse order, relative position of items left in cache does not matter.
10. Why we are changing that to zeros than? Should we keep the current behavior or fixing it to the correct one? If we want to re-insert items that left in cache in the correct order we will have to **sort** them (the current implementation). However if we only want to drop the oldest half we only need to find median value (the new implementation). Quickselect has [lesser complexity](https://en.wikipedia.org/wiki/Quickselect) then Quicksort. I know `O(n*log(n))` and `O(n)` does not make a big difference when we only have 2^16 items, but this optimization is free and Quickselect implementation already exists in VSCode's codebase, so I thought it would be nice to use it.

src/vs/platform/uriIdentity/common/uriIdentityService.ts
yavanosta: Reverted split.

src/vs/platform/uriIdentity/common/uriIdentityService.ts
jrieken: Thanks for thorough explaination and detailed analysis

jrieken (approved):

bpasero (approved):

PR #273118, opened 2025-10-24
Improve history item label text legibility in SCM view

mrleemurray:
Enhance the contrast of history item labels against colored badges by using the panel background color, improving overall readability. Remove unused imports for cleaner code.

Before:
image

After:
Recording 2025-10-24 at 13 28 41


src/vs/workbench/contrib/scm/browser/scmHistoryViewPane.ts
```diff
@@ -529,7 +530,7 @@ class HistoryItemRenderer implements ICompressibleTreeRenderer<SCMHistoryItemVie
 
 		const elements = h('div.label', {
 			style: {
-				color: historyItemRefs[0].color ? asCssVariable(historyItemHoverLabelForeground) : asCssVariable(foreground),
+				color: historyItemRefs[0].color ? asCssVariable(PANEL_BACKGROUND) : asCssVariable(foreground),
```
lszomoru: This problem with this approach is that you are removing the users capability to customize the color using the `scmGraph.historyItemHoverLabelForeground` token since it is not being used any more. The right approach would be to update the color registration for `historyItemHoverLabelForeground` so that instead of `buttonForeground` it defaults to `PANEL_BACKGROUND`. 

```ts
export const historyItemHoverLabelForeground = registerColor('scmGraph.historyItemHoverLabelForeground', PANEL_BACKGROUND, localize('scmGraphHistoryItemHoverLabelForeground', "History item hover label foreground color."));
```

lszomoru (changes requested):

src/vs/workbench/contrib/scm/browser/scmHistoryViewPane.ts
mrleemurray: Good catch - I have made the update, thank you

lszomoru (approved):

PR #273131, opened 2025-10-24
Re-write && to ; in pwsh

### LINKED ISSUE #272704: Re-write && to ; in pwsh commands
https://github.com/microsoft/vscode/pull/272281 adds parsing, but the pwsh grammar doesn't seem to like `&&`. We should see if we can re-wrote those to `;` reliably using the AST.

Tyriar:
Fixes #272704

### DIFF
--- a/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/treeSitterCommandParser.ts
+++ b/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/treeSitterCommandParser.ts
@@ -6,7 +6,7 @@
 import { BugIndicatingError } from '../../../../../base/common/errors.js';
 import { derived, waitForState } from '../../../../../base/common/observable.js';
 import { ITreeSitterLibraryService } from '../../../../../editor/common/services/treeSitter/treeSitterLibraryService.js';
-import type { Language, Parser, Query } from '@vscode/tree-sitter-wasm';
+import type { Language, Parser, Query, QueryCapture } from '@vscode/tree-sitter-wasm';
 
 export const enum TreeSitterCommandParserLanguage {
 	Bash = 'bash',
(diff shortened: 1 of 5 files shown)


src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/treeSitterCommandParser.ts
```diff
@@ -42,9 +42,32 @@ export class TreeSitterCommandParser {
 
 		const captures = query.captures(tree.rootNode);
 		const subCommands = captures.map(e => e.node.text);
+
 		return subCommands;
 	}
```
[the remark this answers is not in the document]
Tyriar: Not necessary

meganrogge (dismissed):
This looks like a good candidate for tests so we don't break anything

src/vs/workbench/contrib/terminalContrib/chatAgentTools/test/electron-browser/commandSimplifier.test.ts
```diff
@@ -344,4 +365,22 @@ suite('command re-writing', () => {
...
+		}
+		setup(() => {
+			commandSimplifier = instantiationService.createInstance(CommandSimplifier, Promise.resolve(OperatingSystem.Windows), parser);
+		});
+
+		test('should rewrite && to ; in PowerShell commands', () => t('echo hello && echo world', 'echo hello ; echo world'));
+		test('should rewrite multiple && to ; in PowerShell commands', () => t('echo first && echo second && echo third', 'echo first ; echo second ; echo third'));
+		test('should handle complex commands with && operators', () => t('npm install && npm test && echo "build complete"', 'npm install ; npm test ; echo "build complete"'));
+		test('should work with Windows PowerShell shell identifier', () => t('Get-Process && Stop-Process', 'Get-Process ; Stop-Process'));
+		test('should preserve existing semicolons', () => t('echo hello; echo world && echo final', 'echo hello; echo world ; echo final'));
```
meganrogge, line 383: does this also work with other string variants like single quote?

meganrogge (dismissed):

src/vs/workbench/contrib/terminalContrib/chatAgentTools/test/electron-browser/commandSimplifier.test.ts
Tyriar: Yep, that's already covered in treeSitterCommandParser.test.ts which does a more general check of the tree capabilities

lramos15 (approved):

PR #273152, opened 2025-10-24
Fix: Restore clickable terminal link hover

Rishi1878:
Fixes #273097

This PR restores mouse-only clickable behavior for terminal links in the integrated terminal. Previously, the `_getLinkHoverString` function escaped the URI and label using `appendText()`, which stripped markdown formatting and rendered the link as plain text.

🔧 Fix:
- Removed unnecessary escaping of `uri` and `label`
- Used `appendLink()` directly to preserve clickable behavior

✅ Tested:
- Hover displays clickable link
- Modifier key hint appears
- Link opens on mouse click without keyboard
- Verified across reloads and multiple terminal sessions


Tyriar (commented):
This doesn't seem to fix it

PR #273175, opened 2025-10-24
add terminal output dropdown, reveal command on focus 

### LINKED ISSUE #271387: Pop out button should show the terminal command via addressable URI
Should flash when navigating to that URI, similar to command navigation (ctrl/cmd+up/down)

### LINKED ISSUE #271388: Showing completed terminal tool output in chat view
This is the first step to getting inline terminal content into the chat view. This part is relatively easy that gets a lot of the way there.

- UX thinking needed around where pop out button shows. How toggle works, etc.

meganrogge:
fixes https://github.com/microsoft/vscode/issues/271387
fixes https://github.com/microsoft/vscode/issues/271388

Also updates xterm

https://github.com/user-attachments/assets/18355fcd-cbab-44c3-ad68-05d62cbc4280


src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
```diff
@@ -106,42 +121,153 @@ export class ChatTerminalToolProgressPart extends BaseChatToolInvocationSubPart
...
 		}));
 	}
+
+	private _collectOutput(terminalInstance: ITerminalInstance): { text: string; truncated: boolean } {
+		this._attachedCommand ??= this._resolveCommand(terminalInstance);
+		let text = this._attachedCommand?.getOutput() ?? '';
+
+		if (!text && terminalInstance.xterm) {
+			text = terminalInstance.xterm.getContentsAsText();
+		}
```
Tyriar: You'll keep colors if you use `getContentsAsHtml`

Tyriar (dismissed):

src/vs/workbench/contrib/chat/browser/chatContentParts/media/chatTerminalToolProgressPart.css
```diff
@@ -53,3 +53,30 @@
...
+
+.chat-terminal-output-container.collapsed { display: none; }
+.chat-terminal-output-container {
+	margin-top: 6px;
+	border: 1px solid var(--vscode-chat-requestBorder);
+	border-radius: 4px;
+	padding: 4px 6px;
+	background: var(--vscode-panel-background);
+	max-height: 300px;
+	overflow: auto;
```
Tyriar: This should be a scrollable element, this is what it looks like on Windows:

Image

src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
```diff
@@ -106,42 +121,164 @@ export class ChatTerminalToolProgressPart extends BaseChatToolInvocationSubPart
...
+			'chat.showTerminalOutput',
+			localize('showTerminalOutput', 'Show Output'),
+			ThemeIcon.asClassName(Codicon.chevronRight),
+			true,
+			async () => {
+				if (!this._showOutputAction) {
+					return;
+				}
+				const expanded = !this._outputContainer.classList.contains('expanded');
+				this._outputContainer.classList.toggle('expanded', expanded);
```
Tyriar: This was a previous command I had before checking out the branch and expanding:

Image

See the height is not correct and the output isn't either (it should just be `2`). Note the actual terminal no longer exists.

src/vs/workbench/contrib/chat/browser/chatContentParts/media/chatTerminalToolProgressPart.css
Tyriar: I think the experience will be best if the text is unwrapped with a horizontal scroll bar so it matches what it looks like in the terminal.

Image

It would look more like this if we unwrapped which is much more familiar:

Image

src/vs/workbench/contrib/chat/browser/chatContentParts/media/chatTerminalToolProgressPart.css
```diff
@@ -53,3 +53,30 @@
 		}
 	}
 }
+
+.chat-terminal-output-container.collapsed { display: none; }
+.chat-terminal-output-container {
+	margin-top: 6px;
+	border: 1px solid var(--vscode-chat-requestBorder);
+	border-radius: 4px;
```
Tyriar: We want to connect the 2 boxes and remove the radius between them:

Image

From design:

Image

I think we want the border in between like this though:

Image

src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
```diff
@@ -106,42 +121,164 @@ export class ChatTerminalToolProgressPart extends BaseChatToolInvocationSubPart
...
+
+		const pre = document.createElement('pre');
+		pre.classList.add('chat-terminal-output');
+		pre.textContent = result.text || localize('chat.terminalNoOutput', 'No output captured for this command.');
+		container.appendChild(pre);
+
+		if (result.truncated) {
+			const note = document.createElement('div');
+			note.classList.add('chat-terminal-output-info');
+			note.textContent = localize('chat.terminalOutputTruncated', 'Output truncated to first {0} characters.', MAX_TERMINAL_OUTPUT_PREVIEW_LENGTH);
```
Tyriar: Current:

Image

Don't need to do this in this PR, but one of the benefits of using xterm is better rendering of the text:

Image

We also won't end up hitting problems when too much content is added like this and performance problems when scrolling won't be a problem since scrolling is virtualized.

src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
```diff
@@ -106,42 +121,164 @@ export class ChatTerminalToolProgressPart extends BaseChatToolInvocationSubPart
...
+
+		if (!text) {
+			return { text: '', truncated: false };
+		}
+
+		let truncated = false;
+		if (text.length > MAX_TERMINAL_OUTPUT_PREVIEW_LENGTH) {
+			const lines = text.split('\n');
+			const maxLines = Math.floor(MAX_TERMINAL_OUTPUT_PREVIEW_LENGTH / 80); // Estimate ~80 chars per line
+			if (lines.length > maxLines) {
```
Tyriar: These are estimated lines using an 80 character width, but the UI says it's restricted to 20k characters.

src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
```diff
@@ -106,42 +121,164 @@ export class ChatTerminalToolProgressPart extends BaseChatToolInvocationSubPart
...
+			this._actionBar.value?.clear();
+			instanceListener.dispose();
 		}));
 	}
+
+	private async _collectOutput(terminalInstance: ITerminalInstance): Promise<{ text: string; truncated: boolean }> {
+		this._attachedCommand ??= this._resolveCommand(terminalInstance);
+		let text = this._attachedCommand?.getOutput() ?? '';
+
+		if (!text && terminalInstance.xterm) {
```
Tyriar: You want to `await terminalInstance.xtermReadyPromise` to avoid this condition

src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
```diff
@@ -106,42 +121,164 @@ export class ChatTerminalToolProgressPart extends BaseChatToolInvocationSubPart
...
+			text = await terminalInstance.xterm.getContentsAsText();
+		}
+
+		if (!text) {
+			return { text: '', truncated: false };
+		}
```
Tyriar, line 418: I think using `terminalInstance.xtermReadyPromise` lets you avoid this error condition.

src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
```diff
@@ -106,42 +121,164 @@ export class ChatTerminalToolProgressPart extends BaseChatToolInvocationSubPart
...
+				this._attachedCommand = undefined;
+			}
+			this._actionBar.value?.clear();
+			instanceListener.dispose();
 		}));
 	}
+
+	private async _collectOutput(terminalInstance: ITerminalInstance): Promise<{ text: string; truncated: boolean }> {
+		this._attachedCommand ??= this._resolveCommand(terminalInstance);
+		let text = this._attachedCommand?.getOutput() ?? '';
```
Tyriar: `TerminalCommand` knows about xterm and the markers, this should use `getContentsAsHtml` internally using that information

src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
```diff
@@ -106,42 +121,164 @@ export class ChatTerminalToolProgressPart extends BaseChatToolInvocationSubPart
...
+
+		if (!text && terminalInstance.xterm) {
+			text = await terminalInstance.xterm.getContentsAsText();
+		}
```
Tyriar: I'm not sure this is edge case is useful? Seems best to just not present the option to expand if we don't know where the command is. It would be confusing if you press this on an older command and see 1000 lines of mostly unrelated content.

src/vs/platform/terminal/common/capabilities/commandDetection/terminalCommand.ts
```diff
@@ -28,6 +28,7 @@ export interface ITerminalCommandProperties {
 	executedMarker?: IMarker | undefined;
 	aliases?: string[][] | undefined;
 	wasReplayed?: boolean | undefined;
+	chatSessionId?: string | undefined;
```
Tyriar: Terminal commands should not know about `chatSessionId` or chat in general. This linking should be handled entirely using `URI`s and `ITerminalChatService`.

Tyriar (dismissed):

src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
meganrogge: Repro steps exactly? When I kill the associated terminal, both of those actions are removed for me

src/vs/workbench/contrib/chat/browser/chatContentParts/media/chatTerminalToolProgressPart.css
```diff
@@ -53,3 +53,41 @@
...
+	font-family: var(--monaco-monospace-font);
+	box-sizing: border-box;
+	overflow: hidden;
+	position: relative;
+}
+.chat-terminal-output-container.expanded { display: block; }
+.chat-terminal-output-container > .monaco-scrollable-element {
+	width: 100%;
+	height: 100%;
+}
```
Tyriar, line 79: This looked a little wrong to me, but I see we do set the width elsewhere so I guess it's fine.

Do we need height though?

src/vs/workbench/contrib/chat/browser/chatContentParts/media/chatTerminalToolProgressPart.css
```diff
@@ -53,3 +53,41 @@
 		}
 	}
 }
+
+.chat-terminal-output-container.collapsed { display: none; }
+.chat-terminal-output-container {
+	margin-top: 0;
+	border: 1px solid var(--vscode-chat-requestBorder);
+	border-radius: 0;
```
Tyriar: This should be polished:

Image

It should look like:

Image

src/vs/workbench/contrib/terminal/browser/xterm/xtermTerminal.ts
```diff
@@ -368,20 +368,49 @@ export class XtermTerminal extends Disposable implements IXtermTerminal, IDetach
...
 
-	async getSelectionAsHtml(command?: ITerminalCommand): Promise<string> {
+	async getSelectionAsHtml(command?: ITerminalCommand, maxLines?: number, explicitRange?: boolean): Promise<string> {
 		if (!this._serializeAddon) {
 			const Addon = await this._xtermAddonLoader.importAddon('serialize');
 			this._serializeAddon = new Addon();
 			this.raw.loadAddon(this._serializeAddon);
 		}
-		if (command) {
+		if (explicitRange && command) {
```
Tyriar: `command` should never have been an argument here. I see it was added in https://github.com/microsoft/vscode/commit/5cc4bc671b8889e6b06d9ba9ab9b12bc820c3cd1

We're calling `getSelectionAsHtml` which means it needs to use the selection exactly as is. This should be pulled into either a generic `getRangeAsHtml` or `getCommandOutputAsHtml`.

src/vs/workbench/contrib/chat/common/chatService.ts
```diff
@@ -305,6 +305,8 @@ export interface IChatTerminalToolInvocationData {
 	alternativeRecommendation?: string;
 	language: string;
 	terminalToolSessionId?: string;
+	terminalCommandUri?: UriComponents;
```
Tyriar, line 316: Can't we just use `URI` here? `UriComponents` normally is only needed when dealing with remote/IPC.

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/runInTerminalTool.ts
```diff
@@ -734,6 +740,48 @@ export class RunInTerminalTool extends Disposable implements IToolImpl {
 		}
 	}
 
+	/**
+	 * Enables persistence of terminal command across sessions.
+	 *
+	 * - `terminalCommandUri` stores the command id. When command detection replays with the same id we can look it up directly.
+	 * - `terminalCommandIndex` captures the position within the command history. On reload, the command ID is regenerated, but the
+	 *   command reappears at the same index, so this acts as a fallback identifier.
```
Tyriar: This doesn't seem necessary? If the Uri is restored properly it should be able to uniquely link them?

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/runInTerminalTool.ts
```diff
@@ -635,7 +639,7 @@ export class RunInTerminalTool extends Disposable implements IToolImpl {
 			let exitCode: number | undefined;
 			try {
 				let strategy: ITerminalExecuteStrategy;
-				const commandDetection = toolTerminal.instance.capabilities.get(TerminalCapability.CommandDetection);
+				this._setTerminalCommandIdentity(toolSpecificData, toolTerminal.instance, commandDetection, commandDetection?.currentCommand?.id);
```
Tyriar: The command isn't executed yet, so `currentCommand` may not be correct

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/runInTerminalTool.ts
```diff
@@ -734,6 +740,48 @@ export class RunInTerminalTool extends Disposable implements IToolImpl {
...
+		if (!toolSpecificData.terminalCommandUri && resolvedCommand.id) {
+			const params = new URLSearchParams(instance.resource.query);
+			params.set('command', resolvedCommand.id);
+			const commandUri = instance.resource.with({ query: params.toString() || undefined });
+			toolSpecificData.terminalCommandUri = commandUri.toJSON();
+		}
+
+		if (toolSpecificData.terminalCommandIndex === undefined) {
+			toolSpecificData.terminalCommandIndex = resolvedIndex;
+		}
```
Tyriar: `toolSpecificData` is meant to be immutable at this point, it's purely for passing from prepare/part to invoke. The `TerminalChatService` should be used to create this link.

Tyriar: https://github.com/microsoft/vscode/issues/273931

src/vs/workbench/contrib/terminal/browser/xterm/xtermTerminal.ts
```diff
@@ -368,20 +368,49 @@ export class XtermTerminal extends Disposable implements IXtermTerminal, IDetach
...
+			let startCol: number;
+			if (command.executedMarker && command.executedMarker.line >= 0) {
+				startLine = command.executedMarker.line;
+				startCol = Math.max(command.executedX ?? 0, 0);
+			} else {
+				startLine = command.marker?.line !== undefined ? command.marker.line + 1 : 1;
+				startCol = Math.max(command.startX ?? 0, 0);
+			}
+
+			const endLine = command.endMarker?.line !== undefined ? command.endMarker.line - 1 : this.raw.buffer.active.length - 1;
```
Tyriar: The end marker may be on the start of the next command line, so we want to take the end X into account as well. Alternatively you could check the following command's prompt start line/X.

src/vs/workbench/contrib/terminal/browser/xterm/xtermTerminal.ts
```diff
@@ -368,20 +368,49 @@ export class XtermTerminal extends Disposable implements IXtermTerminal, IDetach
...
+			}
+
+			const bufferLine = this.raw.buffer.active.getLine(startLine);
+			if (bufferLine) {
+				startCol = Math.min(startCol, bufferLine.length);
+			}
+
+			const range = { startLine, endLine, startCol };
+			const result = this._serializeAddon.serializeAsHTML({ range });
+			return result;
```
Tyriar: We may want to massage the start/end lines to trim empty lines or maybe just 1 empty line on either side?

src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
```diff
@@ -106,42 +144,298 @@ export class ChatTerminalToolProgressPart extends BaseChatToolInvocationSubPart
...
+
+	private _ensureShowOutputAction(): void {
+		if (this._showOutputActionAdded || !this._attachedCommand?.endMarker) {
+			return;
+		}
+		if (!this._focusAction) {
+			return;
+		}
+		if (!this._showOutputAction) {
+			this._showOutputAction = this._register(new Action(
```
Tyriar: Can this be split out like `FocusChatInstanceAction` and then refactored a little for smaller functions? It's quite complex now with the amount of indentation/conditions

src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
```diff
@@ -106,42 +144,298 @@ export class ChatTerminalToolProgressPart extends BaseChatToolInvocationSubPart
...
+	}
+
+	private _layoutOutput(): void {
+		if (!this._outputScrollbar || !this._outputContainer.classList.contains('expanded')) {
+			return;
+		}
+		const scrollableDomNode = this._outputScrollbar.getDomNode();
+		scrollableDomNode.style.width = '100%';
+		const viewportHeight = Math.min(this._outputBody.scrollHeight, MAX_TERMINAL_OUTPUT_PREVIEW_HEIGHT);
+		scrollableDomNode.style.height = `${viewportHeight}px`;
```
Tyriar, line 382: Is this the same element in CSS that gets width and height set?

src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
```diff
@@ -106,42 +144,298 @@ export class ChatTerminalToolProgressPart extends BaseChatToolInvocationSubPart
...
+			const note = document.createElement('div');
+			note.classList.add('chat-terminal-output-info');
+			note.textContent = localize('chat.terminalOutputTruncated', 'Output truncated to first {0} characters.', MAX_TERMINAL_OUTPUT_PREVIEW_LINES);
+			container.appendChild(note);
+		}
+
+		return container;
+	}
+
+	private _matchesInvocationCommand(command: ITerminalCommand, instance: ITerminalInstance): boolean {
```
Tyriar: Might not be understanding all the cases, but this seems overly complex

Tyriar (changes requested):

src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
meganrogge: Yes, removed the static width part

src/vs/workbench/contrib/terminal/browser/xterm/xtermTerminal.ts
meganrogge: can you think of an example when this would happen?

meganrogge: is this overkill? it has seemed to work well for me in testing

src/vs/platform/terminal/common/capabilities/commandDetection/terminalCommand.ts
```diff
@@ -283,8 +286,10 @@ export class PartialTerminalCommand implements ICurrentPartialCommand {
 
 	constructor(
 		private readonly _xterm: Terminal,
+		id?: string
 	) {
-		this.id = generateUuid();
+		this.id = id ?? generateUuid();
+		console.log('terminal partial command id ', this.id);
```
meganrogge:
```suggestion
```

src/vs/workbench/contrib/terminal/browser/xterm/xtermTerminal.ts
```diff
@@ -368,20 +368,80 @@ export class XtermTerminal extends Disposable implements IXtermTerminal, IDetach
...
+			this._serializeAddon = new Addon();
+			this.raw.loadAddon(this._serializeAddon);
+		} else if (command) {
 			const length = command.getOutput()?.length;
 			const row = command.marker?.line;
 			if (!length || !row) {
 				throw new Error(`No row ${row} or output length ${length} for command ${command}`);
 			}
-			this.raw.select(0, row + 1, length - Math.floor(length / this.raw.cols));
+			this.raw.select(command.startX ?? 0, row + 1, length - Math.floor(length / this.raw.cols));
```
meganrogge:
```suggestion
			this.raw.select(0, row + 1, length - Math.floor(length / this.raw.cols));
```

src/vs/platform/terminal/common/capabilities/commandDetection/terminalCommand.ts
```diff
@@ -283,8 +286,9 @@ export class PartialTerminalCommand implements ICurrentPartialCommand {
 
 	constructor(
 		private readonly _xterm: Terminal,
+		id?: string
 	) {
-		this.id = generateUuid();
+		this.id = id ?? generateUuid();
```
Tyriar, line 289: Add a `TODO` here and create a follow up issue about how it doesn't restore properly due to ptyhost?

src/vs/workbench/contrib/terminal/browser/terminalProcessManager.ts
```diff
@@ -277,7 +277,7 @@ export class TerminalProcessManager extends Disposable implements ITerminalProce
 
 				// this is a copy of what the merged environment collection is on the remote side
 				const env = await this._resolveEnvironment(backend, variableResolver, shellLaunchConfig);
-				const shouldPersist = ((this._configurationService.getValue(TaskSettingId.Reconnection) && shellLaunchConfig.reconnectionProperties) || !shellLaunchConfig.isFeatureTerminal) && this._terminalConfigurationService.config.enablePersistentSessions && !shellLaunchConfig.isTransient;
+				const shouldPersist = shellLaunchConfig.forcePersist || ((this._configurationService.getValue(TaskSettingId.Reconnection) && shellLaunchConfig.reconnectionProperties) || !shellLaunchConfig.isFeatureTerminal) && this._terminalConfigurationService.config.enablePersistentSessions && !shellLaunchConfig.isTransient;
```
meganrogge:
```suggestion
				const shouldPersist =  ((this._configurationService.getValue(TaskSettingId.Reconnection) && shellLaunchConfig.forcePersist || shellLaunchConfig.reconnectionProperties) || !shellLaunchConfig.isFeatureTerminal) && this._terminalConfigurationService.config.enablePersistentSessions && !shellLaunchConfig.isTransient;
```

src/vs/workbench/contrib/terminal/browser/xterm/xtermTerminal.ts
```diff
@@ -368,20 +368,80 @@ export class XtermTerminal extends Disposable implements IXtermTerminal, IDetach
...
+		const result = this._serializeAddon.serializeAsHTML({ range });
+		return result;
+	}
+
+	async getSelectionAsHtml(command?: ITerminalCommand): Promise<string> {
+		if (!this._serializeAddon) {
+			const Addon = await this._xtermAddonLoader.importAddon('serialize');
+			this._serializeAddon = new Addon();
+			this.raw.loadAddon(this._serializeAddon);
+		} else if (command) {
```
Tyriar: This else if breaks the `command` arg depending on whether serialize addon is loaded

src/vs/workbench/contrib/terminal/browser/xterm/xtermTerminal.ts
```diff
@@ -368,20 +368,80 @@ export class XtermTerminal extends Disposable implements IXtermTerminal, IDetach
...
+		if (bufferLine) {
+			startCol = Math.min(startCol, bufferLine.length);
+		}
+
+		const range = { startLine, endLine, startCol };
+		const result = this._serializeAddon.serializeAsHTML({ range });
+		return result;
+	}
+
+	async getSelectionAsHtml(command?: ITerminalCommand): Promise<string> {
```
Tyriar, line 430: Should make an issue to remove command completely if it's not trivial to move other uses onto the other function.

src/vs/workbench/contrib/terminal/browser/xterm/xtermTerminal.ts
```diff
@@ -368,20 +368,80 @@ export class XtermTerminal extends Disposable implements IXtermTerminal, IDetach
 		return this._serializeAddon.serializeAsHTML();
 	}
 
-	async getSelectionAsHtml(command?: ITerminalCommand): Promise<string> {
+	async getHtmlForCommand(command: ITerminalCommand, maxLines: number): Promise<string> {
```
Tyriar: `Command` is ambiguous, `CommandOutput` is more precise

src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
```diff
@@ -106,42 +143,281 @@ export class ChatTerminalToolProgressPart extends BaseChatToolInvocationSubPart
...
+		await terminalInstance.xtermReadyPromise;
+		let command = this._attachedCommand;
+		if (!command) {
+			command = this._resolveCommand(terminalInstance);
+		}
+		if (!command) {
+			return { text: '', truncated: false };
+		}
+		this._attachedCommand = command;
+		const text = await terminalInstance.xterm?.getHtmlForCommand(command, MAX_TERMINAL_OUTPUT_PREVIEW_LINES);
```
```suggestion
		const xterm = await terminalInstance.xtermReadyPromise;
		let command = this._attachedCommand;
		if (!command) {
			command = this._resolveCommand(terminalInstance);
		}
		if (!command) {
			return { text: '', truncated: false };
		}
		this._attachedCommand = command;
		const text = await xterm.getHtmlForCommand(command, MAX_TERMINAL_OUTPUT_PREVIEW_LINES);
```

src/vs/workbench/contrib/terminal/browser/xterm/xtermTerminal.ts
```diff
@@ -368,20 +368,80 @@ export class XtermTerminal extends Disposable implements IXtermTerminal, IDetach
...
+
+		if (maxLines && endLine - startLine > maxLines) {
+			startLine = endLine - maxLines;
+			startCol = 0;
+		}
+
+		const bufferLine = this.raw.buffer.active.getLine(startLine);
+		if (bufferLine) {
+			startCol = Math.min(startCol, bufferLine.length);
+		}
```
Tyriar: The C sequence is typically at the end of the prompt input, but it may be on the following line. So it's important to check the actual startX and not assume it's on the prompt line.

src/vs/workbench/contrib/terminal/browser/xterm/xtermTerminal.ts
```diff
@@ -368,20 +368,80 @@ export class XtermTerminal extends Disposable implements IXtermTerminal, IDetach
...
-		if (command) {
+		let startLine: number;
+		let startCol: number;
+		if (command.executedMarker && command.executedMarker.line >= 0) {
+			startLine = command.executedMarker.line;
+			startCol = Math.max(command.executedX ?? 0, 0);
+		} else {
+			startLine = command.marker?.line !== undefined ? command.marker.line + 1 : 1;
+			startCol = Math.max(command.startX ?? 0, 0);
+		}
```
Tyriar, line 385: This `startCol` seems to be ignored?

src/vs/workbench/contrib/terminal/browser/terminalProcessManager.ts
```diff
@@ -277,7 +277,7 @@ export class TerminalProcessManager extends Disposable implements ITerminalProce
 
 				// this is a copy of what the merged environment collection is on the remote side
 				const env = await this._resolveEnvironment(backend, variableResolver, shellLaunchConfig);
-				const shouldPersist = ((this._configurationService.getValue(TaskSettingId.Reconnection) && shellLaunchConfig.reconnectionProperties) || !shellLaunchConfig.isFeatureTerminal) && this._terminalConfigurationService.config.enablePersistentSessions && !shellLaunchConfig.isTransient;
+				const shouldPersist =  ((this._configurationService.getValue(TaskSettingId.Reconnection) && shellLaunchConfig.forcePersist || shellLaunchConfig.reconnectionProperties) || !shellLaunchConfig.isFeatureTerminal) && this._terminalConfigurationService.config.enablePersistentSessions && !shellLaunchConfig.isTransient;
```
Tyriar: Good candidate to fix in different PR

src/vs/workbench/contrib/terminal/browser/terminalService.ts
```diff
@@ -1034,6 +1034,7 @@ export class TerminalService extends Disposable implements ITerminalService {
 					this._onDidDisposeInstance.fire(instance);
 				})
 			]);
+			this._onDidChangeInstances.fire();
```
Tyriar: Why was this needed?

src/vs/workbench/contrib/terminalContrib/chat/browser/terminalChatService.ts
```diff
@@ -10,16 +10,21 @@ import { ITerminalChatService, ITerminalInstance, ITerminalService } from '../..
...
 /**
  * Used to manage chat tool invocations and the underlying terminal instances they create/use.
  */
 export class TerminalChatService extends Disposable implements ITerminalChatService {
 	declare _serviceBrand: undefined;
 
-	private static readonly _storageKey = 'terminalChat.toolSessionMappings';
+
+	private static readonly _toolSessionInstancesKey = 'terminalChat.toolSessionMappings';
+	private static readonly _commandIdByToolSessionIdKey = 'terminalChat.commandIdMappings';
```
Tyriar: Const enum would be better for these

src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
```diff
@@ -106,42 +143,281 @@ export class ChatTerminalToolProgressPart extends BaseChatToolInvocationSubPart
...
+
 		const instanceListener = this._register(terminalInstance.onDisposed(() => {
-			this._actionBar.clear();
-			instanceListener?.dispose();
+			if (this._terminalForOutput === terminalInstance) {
+				this._terminalForOutput = undefined;
+				this._attachedCommand = undefined;
+			}
+			commandDetectionListener.clear();
+			this._actionBar.value?.clear();
```
Tyriar: Can go back to 
```suggestion
			this._actionBar.clear();
```

src/vs/workbench/contrib/terminalContrib/chat/browser/terminalChatService.ts
```diff
@@ -108,6 +139,15 @@ export class TerminalChatService extends Disposable implements ITerminalChatServ
...
 			}
+			const rawCommandIds = this._storageService.get(TerminalChatService._commandIdByToolSessionIdKey, StorageScope.WORKSPACE);
+			if (rawCommandIds) {
+				const parsedCommandIds: [string, string][] = JSON.parse(rawCommandIds);
+				for (const [toolSessionId, commandId] of parsedCommandIds) {
+					if (typeof toolSessionId === 'string' && typeof commandId === 'string') {
+						this._commandIdByToolSessionId.set(toolSessionId, commandId);
+					}
+				}
+			}
```
Tyriar: Haven't thought it through fully, but here's another idea for associating the tool call with the command which may end up being simpler:

- Only support linking commands for rich command detection
- In prepare, create the command ID manually via `tool-${generateUuid()}` and set on `toolSpecificData`, this would make serialize of the part just work at long as we restore the custom command ID
- In invoke, pass that command ID into `runCommand`
- Tell the pty host and renderer shell integration addon that the next command with the matching command line is the ID we created

src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
```diff
@@ -27,16 +26,51 @@ import { ChatConfiguration } from '../../../common/constants.js';
...
 	public readonly domNode: HTMLElement;
 
 	private readonly _actionBar = this._register(new MutableDisposable<ActionBar>());
 
+	private readonly _outputContainer: HTMLElement;
+	private readonly _outputBody: HTMLElement;
+	private readonly _titlePart: HTMLElement;
+	private _outputScrollbar: DomScrollableElement | undefined;
+	private _outputContent: HTMLElement | undefined;
+	private _outputResizeObserver: ResizeObserver | undefined;
```
Tyriar, line 64: We should serialize the output when it's ready, not just the command ID. Doing the output means it will work even after the terminal is killed which is much more important than being able to focus the command after a reload.

Tyriar (dismissed):

src/vs/workbench/contrib/terminal/browser/xterm/xtermTerminal.ts
meganrogge: it's used below in the returned `range`?

meganrogge: oh, never mind, I see I was overriding it. fixed

src/vs/workbench/contrib/terminalContrib/chat/browser/terminalChatService.ts
meganrogge: Will defer to new PR

rebornix (approved):

src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
Tyriar, line 1: https://github.com/microsoft/vscode/issues/274322

src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
```diff
@@ -106,42 +142,325 @@ export class ChatTerminalToolProgressPart extends BaseChatToolInvocationSubPart
...
+	}
+
+	private async _collectOutput(terminalInstance: ITerminalInstance | undefined): Promise<{ text: string; truncated: boolean }> {
+		const storedOutput = this._terminalData.output;
+		if (storedOutput?.html) {
+			return { text: storedOutput.html, truncated: storedOutput.truncated ?? false };
+		}
+		if (!terminalInstance) {
+			return { text: '', truncated: false };
+		}
```
Tyriar, line 406: `_collectOutput` is only called when `terminalInstance` is truthy I believe

src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
```diff
@@ -106,42 +142,325 @@ export class ChatTerminalToolProgressPart extends BaseChatToolInvocationSubPart
...
+
+		if (!this._terminalInstance) {
+			const resource = this._getTerminalResource();
+			if (resource) {
+				this._terminalInstance = this._terminalService.getInstanceFromResource(resource);
+			}
+		}
+		const output = await this._collectOutput(this._terminalInstance);
+		const content = this._renderOutput(output);
+		const theme = this._terminalInstance?.xterm?.getXtermTheme();
```
Tyriar, line 331: `xterm` is used in collect output and in this function. There should be a single `xtermReadyPromise` call and a single check whether it's truthy.

src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
```diff
@@ -106,42 +142,325 @@ export class ChatTerminalToolProgressPart extends BaseChatToolInvocationSubPart
...
+		if (theme) {
+			const inlineTerminal = content.querySelector('div');
```
Tyriar, line 333: Should use the `h` helper, and keep track of elements we care about. We should very rarely query sub-elements like this. Doing this should also make the `if (inlineTerminal)` on the following line unnecessary.

src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
```diff
@@ -106,42 +142,325 @@ export class ChatTerminalToolProgressPart extends BaseChatToolInvocationSubPart
...
+		const command = this._resolveCommand(terminalInstance);
+		if (!command?.endMarker) {
+			return { text: '', truncated: false };
+		}
+		const text = await xterm.getCommandOutputAsHtml(command, CHAT_TERMINAL_OUTPUT_MAX_PREVIEW_LINES);
+		if (!text) {
+			return { text: '', truncated: false };
+		}
+
+		return { text, truncated: false };
```
Tyriar, line 420: This `if (!text)` seems redundant as `getCommandOutputAsHtml` returns a string and when it's `''` the behavior would be the same if it fell through.

src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
```diff
@@ -199,12 +552,16 @@ export class FocusChatInstanceAction extends Action implements IAction {
...
 	public override async run() {
 		this.label = localize('focusTerminal', 'Focus Terminal');
+		this._terminalService.setActiveInstance(this._instance);
 		if (this._instance.target === TerminalLocation.Editor) {
 			this._terminalEditorService.openEditor(this._instance);
 		} else {
-			this._terminalGroupService.showPanel(true);
+			await this._terminalGroupService.showPanel(true);
 		}
 		this._terminalService.setActiveInstance(this._instance);
```
Tyriar, line 561: `setActiveInstance` is now run twice? This shouldn't be needed. What case were you trying to fix here, as this code is relatively fragile and caused a new insiders to go out when changed before.

src/vs/workbench/contrib/terminal/browser/xterm/xtermTerminal.ts
```diff
@@ -368,6 +368,65 @@ export class XtermTerminal extends Disposable implements IXtermTerminal, IDetach
...
+		let emptyLinesFromStart = 0;
+		for (let i = startLine; i <= endLine; i++) {
+			const line = this.raw.buffer.active.getLine(i);
+			if (line && line.translateToString(true).trim() === '') {
+				emptyLinesFromStart++;
+			} else {
+				break;
+			}
+		}
+		startLine = startLine + emptyLinesFromStart;
```
Tyriar, line 413: Lots of code repetition

src/vs/workbench/contrib/terminal/browser/xterm/xtermTerminal.ts
```diff
@@ -368,6 +368,65 @@ export class XtermTerminal extends Disposable implements IXtermTerminal, IDetach
...
+
+		if (maxLines && endLine - startLine > maxLines) {
+			startLine = endLine - maxLines;
+			startCol = 0;
+		}
+
+		const bufferLine = this.raw.buffer.active.getLine(startLine);
+		if (bufferLine) {
+			startCol = Math.min(startCol, bufferLine.length);
+		}
```
Tyriar, line 423: It's not clear to me if this does anything

src/vs/workbench/contrib/terminal/browser/xterm/xtermTerminal.ts
```diff
@@ -368,6 +368,65 @@ export class XtermTerminal extends Disposable implements IXtermTerminal, IDetach
...
+
+		const range = { startLine, endLine, startCol };
+		const result = this._serializeAddon.serializeAsHTML({ range });
+		return result;
```
Tyriar, line 427: nit: Easier to read when it's all on the same line imo

```
return this._serializeAddon.serializeAsHTML({ startLine, endLine, startCol });
```

PR #273180, opened 2025-10-24
Add a happy path for URI path normalization when we can just return t…

irengrig:
…he incoming value

String search should work very efficiently - for the path without ./ and ../ we typically expect one dot symbol for the file extension, so it will mostly only compare each symbol with dot.

This should save us a lot or operations (and memory allocation) for most cases when relative paths are not used.


bpasero:
That file is a copy of node.js `path` lib. You need to make this PR against node.js: https://github.com/nodejs/node

PR #273187, opened 2025-10-24
terminal command attachment follow-ups

meganrogge:
follow up to #272943

### DIFF
--- a/src/vs/workbench/contrib/chat/browser/actions/chatContext.ts
+++ b/src/vs/workbench/contrib/chat/browser/actions/chatContext.ts
@@ -269,7 +269,7 @@ export class TerminalContext implements IChatContextValueItem {
 	constructor(private readonly _resource: URI, @ITerminalService private readonly _terminalService: ITerminalService) {
 
 	}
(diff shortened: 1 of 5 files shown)


rzhao271 (approved):

src/vs/workbench/contrib/chat/browser/actions/chatContext.ts
```diff
@@ -279,7 +279,7 @@ export class TerminalContext implements IChatContextValueItem {
 			return;
 		}
 
-		const command = terminal.capabilities.get(TerminalCapability.CommandDetection)?.commands.find(cmd => cmd.id === this._resource.query);
+		const command = terminal.capabilities.get(TerminalCapability.CommandDetection)?.commands.find(cmd => cmd.id === this._resource.query.replace('command=', ''));
```
Tyriar, line 282: This isn't great either and will also fail if there is more than one query param. I think this is the right way to do it:

```ts
new URLSearchParams('commandId=abc&something=else').get('commandId')
```

Issue #273029, opened 2025-10-24
Chat response: render thinking parts more like progress steps


bpasero:
<img width="557" height="298" alt="Image" src="https://github.com/user-attachments/assets/c59de264-ba3d-46ae-835e-ef2bd8604ac4" />

Here you can see 2 thinking parts around a progress step, almost the same representation (with checkmark etc.) but vastly different fonts and spacings and borders. I would suggest to render them all the same, esp. considering that its more for the AI to help the result than for the user to look at.

Another example that is calling out for a lot of attention:

<img width="551" height="562" alt="Image" src="https://github.com/user-attachments/assets/3830d14a-f002-47e2-b001-b0c6c5bf286f" />

//cc @isidorn @lszomoru @sbatten @mrleemurray

justschen:
with the latest thinking changes, i think this aligns way better than what the screenshot shows!

PR #273207, opened 2025-10-24
PROTOTYPE: Lightweight rendering of fixed scroll height

sbatten:
This prototype updates the fixed height scrolling thinking part to be consistent with the general tool invocation (e.g. Search tool invocation)

image

image

image

cc @bpasero per https://github.com/microsoft/vscode/issues/273029

### DIFF
--- a/src/vs/workbench/contrib/chat/browser/chatContentParts/chatThinkingContentPart.ts
+++ b/src/vs/workbench/contrib/chat/browser/chatContentParts/chatThinkingContentPart.ts
@@ -4,7 +4,6 @@
  *--------------------------------------------------------------------------------------------*/
 
 import { $, clearNode } from '../../../../../base/browser/dom.js';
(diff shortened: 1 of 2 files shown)


bpasero:
@sbatten yes, thats some nice progress for the issue. I wonder if the expanded body of the thinking part needs to be so complex in rendering: whats the idea behind the bullets and the line going from top to bottom on the left? It looks so different from anything else in Chat that I wonder if it isn't too much. What if we would simply render the thinking text after you expand without much borders and bullets?

sbatten:
> expand without much borders

Other collapsible chat widgets use the rounded corner border as well, like search results. I think this helps with the visual hierarchy personally.

As for the bullets. This is supposed to convey "chain of thought" It also helps with the visual hierarchy and helps distinguish the thinking section from the rest, but I'll let @justschen comment.

PR #273312, opened 2025-10-25
Enable Inner Scopes Option for Sticky Scroll

### LINKED ISSUE #208406: Option for sticky scroll to prioritize inner blocks
Currently sticky scroll always stick the first lines, which may make it less useful in some situations.

Consider the following case:

```html
<!DOCTYPE html>
<html>
    <body>
        <script>
            class SomeClass {
                constructor() {
                    // Long & complicated code
                    // Long & complicated code
                    // Long & complicated code
                }
            }
        </script>
    </body>
</html>
```

If we scroll to the comments with a max line count set to 3 for sticky scroll, `<html>`, `<body>` and `<script>` will stick around. However, we might not want to always stare at `<html>` whenever we write HTML, so it may be better to let `<html>` make room for inner code blocks.

| Current | Suggested alternative |
| ------- | --------------------- |
| "<html>", "<body>" and "<script>" stuck around, as they normally would | It might be better if "<script>", "class SomeClass {", and "constructor() {" stuck around instead. |

This problem also applies to other languages with a complicated program structure. Adding such an option can sometimes make it more efficient to work with these languages.

benvillalobos:
Fixes https://github.com/microsoft/vscode/issues/208406

## Changes Made
- Added a `scopePreference` setting. Values are `innerScope` and `outerScope`.
  - Note on the naming: VS has the same feature, they use the same naming
- Modified `stickyScrollController` to support inner scopes

## To Test
1. Settings (UI) > Search "Sticky scroll" > `Scope Preference: innerScopes`
2. Consider: `Default Model: foldingProviderModel` to help test.
3. Open a file that contains lots of nesting (functions, classes, html nodes, etc.) Optionally, create an html file and use the contents I've provided below.
4. Scroll around and observe the sticky behavior

## Video (Updated 10/31 🎃)

https://github.com/user-attachments/assets/478ea1a4-944e-4b17-875e-4769cb0b3c5e

I tested the logic on the following files:

```html
<html>




    <head>



        <h1>



            <h2>



                <h3>
                    
                    <h1>

                        <h2>

                            <h3>

                            </h3>
                        </h2>   
                    </h1>

                </h3>           

                
            </h2>
        </h1>
    </head>

</html>
```

```
export function level1() {



    function level2() {


    
        function level3() {
            
            function level4() {


                function level5() {
                    function level6() {
                        function level7() {
                            function level8() {
                                function level9() {
                                    function level10() {
                                        // Deepest level reached
                                    }
                                    level10();
                                }
                                level9();
                            }
                            level8();
                        }
                        level7();
                    }
                    level6();
                }
                level5();
            }
            level4();
        }
        level3();
    }
    level2();
}



```

### DIFF
--- a/src/vs/editor/contrib/stickyScroll/browser/stickyScrollController.ts
+++ b/src/vs/editor/contrib/stickyScroll/browser/stickyScrollController.ts
@@ -592,29 +592,26 @@ export class StickyScrollController extends Disposable implements IEditorContrib
 	}
 
 	private _shouldAppendStickyLine(
-		line: StickyLineCandidate,
-		startLineNumbers: number[],
-		innerScopes: boolean,
-		maxNumberStickyLines: number,
+		candidate: StickyLineCandidate,
+		useInnerScopes: boolean,
+		isAtMaxLines: boolean,
 		scrollTop: number,
-		stickyWidgetHeight: number
+		currentWidgetHeight: number
 	): boolean {
-		const topOfBeginningLine = this._editor.getTopForLineNumber(line.startLineNumber) - scrollTop;
-		const bottomOfEndLine = this._editor.getBottomForLineNumber(line.endLineNumber + (innerScopes ? 1 : 0)) - scrollTop;
+		const topOfBeginningLine = this._editor.getTopForLineNumber(candidate.startLineNumber) - scrollTop;
+		const bottomOfEndLine = this._editor.getBottomForLineNumber(candidate.endLineNumber + (useInnerScopes ? 1 : 0)) - scrollTop;
 
-		if (!innerScopes) {
-			return topOfBeginningLine < stickyWidgetHeight && bottomOfEndLine >= stickyWidgetHeight;
+		if (!useInnerScopes) {
+			return topOfBeginningLine < currentWidgetHeight && bottomOfEndLine >= currentWidgetHeight;
 		}
-
-		const isAtMaxLines = startLineNumbers.length === maxNumberStickyLines;
 
 		if (isAtMaxLines) {
-			const bottomOfBeginningLine = this._editor.getBottomForLineNumber(line.startLineNumber) - scrollTop;
-			return (bottomOfBeginningLine) < stickyWidgetHeight
-				&& (bottomOfEndLine) >= stickyWidgetHeight;
+			const bottomOfBeginningLine = this._editor.getBottomForLineNumber(candidate.startLineNumber) - scrollTop;
+			return (bottomOfBeginningLine) < currentWidgetHeight
+				&& (bottomOfEndLine) >= currentWidgetHeight;
 		}
 
-		return topOfBeginningLine < stickyWidgetHeight && bottomOfEndLine >= stickyWidgetHeight;
+		return topOfBeginningLine < currentWidgetHeight && bottomOfEndLine >= currentWidgetHeight;
 	}
 
 	findScrollWidgetState(): StickyScrollWidgetState {
@@ -631,7 +628,7 @@ export class StickyScrollController extends Disposable implements IEditorContrib
 			let stickyWidgetHeight = 0;
 			for (const range of candidateRanges) {
 				const bottomOfEndLine = this._editor.getBottomForLineNumber(range.endLineNumber + (innerScopes ? 1 : 0)) - scrollTop;
-				const shouldAppendStickyLine = this._shouldAppendStickyLine(range, startLineNumbers, innerScopes, maxNumberStickyLines, scrollTop, stickyWidgetHeight);
+				const shouldAppendStickyLine = this._shouldAppendStickyLine(range, innerScopes, startLineNumbers.length === maxNumberStickyLines, scrollTop, stickyWidgetHeight);
 
 				if (shouldAppendStickyLine) {
 					startLineNumbers.push(range.startLineNumber);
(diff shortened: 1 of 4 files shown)


src/vs/editor/contrib/stickyScroll/browser/stickyScrollWidget.ts
```diff
@@ -152,7 +152,9 @@ export class StickyScrollWidget extends Disposable implements IOverlayWidget {
 		const previousLineNumbers = this._lineNumbers;
 		this._lineNumbers = data.lineNumbers;
 		this._lastLineRelativePosition = data.lastLineRelativePosition;
-		const rebuildFromIndex = this._findIndexToRebuildFrom(previousLineNumbers, this._lineNumbers, rebuildFromIndexCandidate);
+		const innerScopes = this._editor.getOption(EditorOption.stickyScroll).scopePreference === 'innerScopes';
+		const indexCandidate = innerScopes ? 0 : rebuildFromIndexCandidate;
+		const rebuildFromIndex = this._findIndexToRebuildFrom(previousLineNumbers, this._lineNumbers, indexCandidate);
```
benvillalobos, line 157: This approach prioritizes a minimal diff over performance.
By forcing `stickyScrollWidget` to re-render every line on each update, there's a perf cost (scoped only to this `innerScopes` setting). The upside is this avoids a more extensive refactor of the widget’s rendering logic to properly cache rendered lines.

src/vs/monaco.d.ts
```diff
@@ -4450,6 +4450,10 @@ declare namespace monaco.editor {
 		 * Define whether to scroll sticky scroll with editor horizontal scrollbae
 		 */
 		scrollWithEditor?: boolean;
+		/**
+		 * Define whether to show outer scopes or inner scopes when the max line count is reached
+		 */
+		scopePreference?: 'outerScopes' | 'innerScopes';
```
benvillalobos, line 4456: Also, does this need to go through API review?

src/vs/editor/contrib/stickyScroll/browser/stickyScrollController.ts
```diff
@@ -601,21 +601,29 @@ export class StickyScrollController extends Disposable implements IEditorContrib
...
 				const topOfBeginningLine = this._editor.getTopForLineNumber(start) - scrollTop;
 				const bottomOfEndLine = this._editor.getBottomForLineNumber(end) - scrollTop;
-				if (topOfElement > topOfBeginningLine && topOfElement <= bottomOfEndLine) {
+
+				if (topOfBeginningLine < stickyWidgetHeight && bottomOfEndLine >= stickyWidgetHeight) {
 					startLineNumbers.push(start);
 					endLineNumbers.push(end + 1);
-					if (bottomOfElement > bottomOfEndLine) {
-						lastLineRelativePosition = bottomOfEndLine - bottomOfElement;
+					stickyWidgetHeight += range.height;
```
aiday-mar, line 641: Hi Ben, thank for the PR. I am not sure I quite understand this logic. I was wondering if you could help me understand it using a diagram?

aiday-mar (commented):
Thanks for the PR. I was tesing this PR and saw the following behavior on a nested function in the screen recording below. Notice how the last sticky scroll line appears even before we hit the line. I think there is a bug in the PR. 

https://github.com/user-attachments/assets/481cc64e-4289-489b-9776-022abc59ec4e

src/vs/editor/contrib/stickyScroll/browser/stickyScrollController.ts
benvillalobos: `bottomOfElement` used to mean "height of the sticky scroll widget", but it's not reliable WRT inner scopes.

- See that `bottomOfElement` was constructed with `range.top` + `range.height`.
- `range.top` comes from here: https://github.com/benvillalobos/vscode/blob/7ae5be88cb262b513ee08f05b4edfc38b60a28eb/src/vs/editor/contrib/stickyScroll/browser/stickyScrollProvider.ts#L169
- Notice that `top` starts at 0, and every recursive call of [getCandidateStickyLinesIntersectingFromStickyModel](https://github.com/benvillalobos/vscode/blob/7ae5be88cb262b513ee08f05b4edfc38b60a28eb/src/vs/editor/contrib/stickyScroll/browser/stickyScrollProvider.ts#L219) passes `top + height` for the value of `top`, where `height` is the lineHeight. 

Say we have a consistent lineHeight of `19`, `4` functions scoped inside of each other, and a max sticky count of `3`

- function1
  - top: `0`
  - height: `19`
- function2
  - top: `19` (top + height)
  - height: `19`
- function3
  - top: `38`
  - height: `19`
- function4 (once this stickies, function1 gets pushed out)
  - top: `57`
  - height: `19`

So the `bottomOfElement` for `function4` will be `76`, but the height of the sticky scroll widget is actually `57` because we don't append a 4th line. So `bottomOfElement` no longer represents "height of the sticky scroll widget" like it used to.

So the solution is to dynamically figure out how large the sticky scroll widget is by adding / subtracting to `stickyWidgetHeight` as we append / remove sticky lines.

`getCandidateStickyLinesIntersectingFromStickyModel` should probably be refactored to remove the `top` param. IIRC that param's use case was specifically meant for sticky scroll, so its probably a safe removal

aiday-mar: Hi thanks for the comment. I don't think `bottomOfElement` represents the height of the sticky scroll widget. The variable `bottomOfElement` and `topOfElement` are used to find the theoretical top and bottom of the sticky scroll lines relative to the top of the viewport if the sticky scroll widget did not have a cropped height. These values are compared to the actual position of the top and bottom of the corresponding editor lines, to determine if the candidate sticky lines should be stuck to the top or not. In effect, the comparison is a hypothetical comparison which is used to determine whether the line should be stuck or not. 

The reason we don't use the widget height directly is because we can not know the widget height until all the lines that are stuck are computed. It's a recursive problem. To better understand, consider lines 1, 2 and 3 which are nested within each other as follows:

line 1: function f1 () {
line 2: function f2 () {
line 3: function f3 () {
line 4: }}}

If we add line 2 to sticky scroll then according to your calculation the widget would have height 38 (line1 + line2). But actually as soon as you add line 2, the widget becomes larger and it actually now touches line 3. So the truth is that the widget height at this point would be 57 (line1 + line2 + line3) not height 38 as was initially hypothesized. In essence adding lines modulates recursively the sticky scroll widget height depending on where the next lines are situated. We can only know the height at the very end.

Therefore I think we need to continue using the variables `bottomOfElement` and `topOfElement` to do the required computation you need.

PR #273450, opened 2025-10-26
Search results: configurable truncation and size threshold

### LINKED ISSUE #273410: Search Results - Improve truncation - Show full line, Truncate on the Left or the Right
```
Version: 1.105.1 (user setup)
Electron: 37.6.0
ElectronBuildId: 12502201
Chromium: 138.0.7204.251
Node.js: 22.19.0
V8: 13.8.258.32-electron.0
OS: Windows_NT x64 10.0.26100
```

This has been mentioned a few times in the past and i think it should be revisited as I am struggling with how truncation works

### See this example

- A line is truncated when it is over a certain length.
    - this should be a configurable option.
- The truncation happens on the beginning of the line
    - I read, left to right, and find this tricky to use. I would prefer the truncation to happen on the end of the line.

Image

### Proposed options

- `search.resultsTruncation`
   - Off (no truncation)
   - Beginning of Line
   - End of line
   - Both (the search term would be in the middle of 2 `...` = good for context)
- `search.allowedSizeBeforeTruncation`
   - 123  ( a number of characters before truncation happens) 

### Links
- https://github.com/Microsoft/vscode/issues/57895
- https://github.com/microsoft/vscode/issues/59310

Yonas650:
**Summary**

*   Improves readability of search result previews by making truncation behavior configurable.
    
*   Adds options to show full lines, truncate from the start, end, or both sides (keeping the match centered).
    
*   Keeps default behavior unchanged for existing users.
    

**New Settings**

*   search.resultsTruncation: 'start' | 'end' | 'both' | 'off' (default: 'start')
    
*   search.allowedSizeBeforeTruncation: number of characters before truncating (default: 1000; regex searches enforce a minimum of 10000 internally for replace correctness)
    

Example:

*   "search.resultsTruncation": "both"
    
*   "search.allowedSizeBeforeTruncation": 80
    

**Behavior**

*   start: elides the beginning (“⟪ N characters skipped ⟫”), match remains in view.
    
*   end: shows the start, elides the tail with the same elision token.
    
*   both: elides both sides to center the match for context.
    
*   off: no elision (full line preview; UI may still visually clip).
    

**Scope**

*   Applies to Search view, Quick Text Search, and Search Editor.
    
*   Works with both ripgrep and non-rg (in-memory) result paths.
    

**Implementation Notes**

*   Uses existing elision token (unchanged): “⟪ N characters skipped ⟫” to preserve Search Editor offset parsing.
    
*   Honors regex replace constraints by using >= 10000 chars per line for regex queries.
    
*   Default behavior remains identical to current (truncation at the start).
    

**Test Plan**

*   Unit tests: src/vs/workbench/services/search/test/common/search.test.ts
    
    *   Covers off, end, and both modes in addition to existing behavior.
        
*   Manual:
    
    *   Create a file with a very long single line containing the search term.
        
    *   Toggle search.resultsTruncation through all options and validate:
        
        *   start → left elision only
            
        *   end → trailing elision only
            
        *   both → both sides elide, match centered
            
        *   off → full line shown
            
    *   Adjust search.allowedSizeBeforeTruncation to a small value (e.g., 40) to see clear effects.
        
    *   Verify in Search view, Quick Text Search (%), and Search Editor.
        

**Release Notes**

*   New: Control how search results are truncated with search.resultsTruncation and search.allowedSizeBeforeTruncation for improved readability and context.
    

Fixes #273410


Yonas650:

osortega (changes requested):
I'm not sure I understand what problem this PR is trying to fix, when we do search we can truncate both beginning or end depending on where the term that was searched for.
Example: searching for `interface` you can see truncation in all places.
Image

With this change the search results are pretty much impossible to parse
Image

shoulders:
@osortega it is not my PR and my level of programming is not at that level.

For me, the most important thing is to disable truncation.

The ability to also control the truncation amounts, left and right, is important to have a flexible workflow.

shoulders:
@osortega Why was this closed?

PR #174829, opened 2023-02-20
composition enter should not accept rename

lawvs:
Fix the issue where pressing enter while using CJK input method triggers unexpected confirmation of rename operation, which may also result in malfunctioning of rename operation.

https://user-images.githubusercontent.com/18554747/220090240-c9baa156-4082-44dc-9fd3-48006492f3dd.mov

PR #273452, opened 2025-10-27
Ignore Enter for replace while composing. Fix #257776

### LINKED ISSUE #257776: Pressing Enter to confirm IME composition in Search/Replace widget also triggers the replace action

- VS Code Version: 
Version: 1.102.2 (Universal)
Commit: c306e94f98122556ca081f527b466015e1bc37b0
Date: 2025-07-22T12:15:48.520Z (2 days ago)
Electron: 35.6.0
ElectronBuildId: 11847422
Chromium: 134.0.6998.205
Node.js: 22.15.1
V8: 13.4.114.21-electron.0
OS: Darwin arm64 24.5.0

- OS Version: 
macOS 15.5 (24F74)
Steps to Reproduce:

As a user of CJK (Chinese/Japanese) Input Method Editors (IMEs), I've encountered a workflow-breaking issue with the Search/Replace widget.

When my IME is active , pressing Enter should serve two distinct purposes depending on the context:

    Finalize an IME composition: When I type English letters like myNewFunction, the IME holds them in a pre-edit or composition state (often underlined). Pressing Enter confirms this composition, turning myNewFunction into finalized text in the input field.

    Trigger an action: When there is no active IME composition, pressing Enter should trigger the default action, such as "Replace Next".

The problem is that in the VS Code "Replace" input box, both of these actions are triggered by a single Enter press.

Steps to Reproduce:

    Switch your operating system's input method to a Chinese or Japanese IME.

    Open any file in VS Code.

    Press Ctrl+F to open the Search/Replace widget.

    Enter some text in the "Find" input.

    Click to focus the "Replace" input box.

    With the IME active, type some English letters (e.g., testReplace). You will see the IME's composition underline, indicating the text is not yet finalized.

    Press the Enter key.

Expected Behavior:

The Enter key press should only confirm the IME composition. The text testReplace should appear in the "Replace" input box as finalized text, the focus should remain in the input box, and no replacement action should be performed in the editor. A second Enter press would then be required to trigger the "Replace Next" action.

Actual Behavior:

The single Enter key press both confirms the IME composition and immediately triggers the "Replace Next" action in the editor. This makes it impossible to simply type English text into the replace box with an active IME without performing an unwanted replacement.

Thank you for looking into this

tamuratak:
Ignore Enter for replace while composing (add isComposing precondition). Fix #257776. 
 
 Similar to #174829

### DIFF
--- a/src/vs/editor/contrib/find/browser/findController.ts
+++ b/src/vs/editor/contrib/find/browser/findController.ts
@@ -1107,7 +1107,7 @@ registerEditorCommand(new FindCommand({
 	handler: x => x.replace(),
 	kbOpts: {
 		weight: KeybindingWeight.EditorContrib + 5,
-		kbExpr: ContextKeyExpr.and(EditorContextKeys.focus, CONTEXT_REPLACE_INPUT_FOCUSED),
+		kbExpr: ContextKeyExpr.and(EditorContextKeys.focus, CONTEXT_REPLACE_INPUT_FOCUSED, ContextKeyExpr.not('isComposing')),
 		primary: KeyCode.Enter
 	}
 }));
--- a/src/vs/editor/contrib/find/browser/findController.ts
+++ b/src/vs/editor/contrib/find/browser/findController.ts
@@ -1107,7 +1107,7 @@ registerEditorCommand(new FindCommand({
 	handler: x => x.replace(),
 	kbOpts: {
 		weight: KeybindingWeight.EditorContrib + 5,
-		kbExpr: ContextKeyExpr.and(EditorContextKeys.focus, CONTEXT_REPLACE_INPUT_FOCUSED, ContextKeyExpr.not('isComposing')),
+		kbExpr: ContextKeyExpr.and(EditorContextKeys.focus, CONTEXT_REPLACE_INPUT_FOCUSED, EditorContextKeys.isComposing.negate()),
 		primary: KeyCode.Enter
 	}
 }));
--- a/src/vs/editor/common/editorContextKeys.ts
+++ b/src/vs/editor/common/editorContextKeys.ts
@@ -60,6 +60,9 @@ export namespace EditorContextKeys {
 
 	export const standaloneColorPickerVisible = new RawContextKey<boolean>('standaloneColorPickerVisible', false, nls.localize('standaloneColorPickerVisible', "Whether the standalone color picker is visible"));
 	export const standaloneColorPickerFocused = new RawContextKey<boolean>('standaloneColorPickerFocused', false, nls.localize('standaloneColorPickerFocused', "Whether the standalone color picker is focused"));
+
+	export const isComposing = new RawContextKey<boolean>('isComposing', false, nls.localize('isComposing', "Whether the editor is in the composition mode"));
+
 	/**
 	 * A context key that is set when an editor is part of a larger editor, like notebooks or
 	 * (future) a diff editor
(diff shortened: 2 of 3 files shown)


osortega (dismissed):
Thank you ✨

src/vs/editor/contrib/find/browser/findController.ts
```diff
@@ -1107,7 +1107,7 @@ registerEditorCommand(new FindCommand({
 	handler: x => x.replace(),
 	kbOpts: {
 		weight: KeybindingWeight.EditorContrib + 5,
-		kbExpr: ContextKeyExpr.and(EditorContextKeys.focus, CONTEXT_REPLACE_INPUT_FOCUSED),
+		kbExpr: ContextKeyExpr.and(EditorContextKeys.focus, CONTEXT_REPLACE_INPUT_FOCUSED, EditorContextKeys.isComposing.negate()),
```
Tyriar, line 1110: This context key is never used so it doesn't do anything? Am I missing someting? @osortega

Tyriar (dismissed):

src/vs/editor/contrib/find/browser/findController.ts
osortega: Looks like we are creating it here https://github.com/microsoft/vscode/blob/541da45a06f489730064d4b300f238ff52d94657/src/vs/workbench/services/keybinding/browser/keybindingService.ts#L202-L203
But it was never added to EditorContextKeys
@tamuratak could you make sure we use is there?

tamuratak: Do you mean to apply the following changes?

```diff
diff --git a/src/vs/workbench/services/keybinding/browser/keybindingService.ts b/src/vs/workbench/services/keybinding/browser/keybindingService.ts
index aa62f865df0..bb360fb9b81 100644
--- a/src/vs/workbench/services/keybinding/browser/keybindingService.ts
+++ b/src/vs/workbench/services/keybinding/browser/keybindingService.ts
@@ -46,6 +46,8 @@ import { Registry } from '../../../../platform/registry/common/platform.js';
 import { ITelemetryService } from '../../../../platform/telemetry/common/telemetry.js';
 import { IUriIdentityService } from '../../../../platform/uriIdentity/common/uriIdentity.js';
 
+import { EditorContextKeys } from '../../../../editor/common/editorContextKeys.js';
+
 // workbench
 import { remove } from '../../../../base/common/arrays.js';
 import { commandsExtensionPoint } from '../../actions/common/menusExtensionPoint.js';
@@ -199,7 +201,7 @@ export class WorkbenchKeybindingService extends AbstractKeybindingService {
 	) {
 		super(contextKeyService, commandService, telemetryService, notificationService, logService);
 
-		this.isComposingGlobalContextKey = contextKeyService.createKey('isComposing', false);
+		this.isComposingGlobalContextKey = EditorContextKeys.isComposing.bindTo(contextKeyService);
 
 		this.kbsJsonSchema = new KeybindingsJsonSchema();
 		this.updateKeybindingsJsonSchema();
```

osortega: Could you just do
```
this.isComposingGlobalContextKey = contextKeyService.createKey(EditorContextKeys.isComposing.key, false);
```

tamuratak: done

osortega (approved):

roblourens (approved):

otbotb371-lgtm:
> ## Pull Request Overview
> 
> This PR fixes an issue where pressing Enter during IME composition in the find/replace widget would trigger the replace action prematurely. The fix adds an `isComposing` precondition to prevent Enter from executing the replace command while text is being composed.
> 
> **Key Changes:**
> - Added `isComposing` precondition to the replace command's keyboard binding expression to prevent premature activation during IME composition.
> 
> 
> 
>

PR #273550, opened 2025-10-27
Fix memory leak in InlineEditsGutterIndicator (#273549)

yavanosta:
Clean up previous observable every time we create a new one.

Fixes #273549


src/vs/editor/contrib/inlineCompletions/browser/view/inlineEdits/components/gutterIndicatorView.ts
```diff
@@ -92,11 +93,14 @@ export class InlineEditsGutterIndicator extends Disposable {
 
 		this._originalRangeObs = mapOutFalsy(this._originalRange);
 		this._state = derived(this, reader => {
+			this._stateDisposables?.dispose();
+			this._stateDisposables = this._register(new DisposableStore());
```
hediet: Now `this._register` would leak.

Instead, use `reader.store`

hediet:
Thanks for the PR!

Tyriar (approved):

hediet (approved):

PR #273605, opened 2025-10-27
Debug log terminal resize in terminal output 

anthonykim1:


src/vs/workbench/contrib/terminal/browser/terminalInstance.ts
```diff
@@ -802,15 +796,15 @@ export class TerminalInstance extends Disposable implements ITerminalInstance {
 			() => this._isVisible,
 			() => xterm,
 			async (cols, rows) => {
-				xterm.raw.resize(cols, rows);
+				xterm.resize(cols, rows);
```
Tyriar, line 799: This one too https://github.com/microsoft/vscode/blob/d3f0dce897975675ca356d699047b9285257c16b/src/vs/workbench/contrib/terminal/browser/terminalInstance.ts#L1534

src/vs/workbench/contrib/terminal/browser/xterm/xtermTerminal.ts
```diff
@@ -442,6 +440,7 @@ export class XtermTerminal extends Disposable implements IXtermTerminal, IDetach
 	}
 
 	resize(columns: number, rows: number): void {
+		this._logService.debug(`Resizing terminal to ${columns} col, ${rows} row`);
```
Tyriar: We can save making a bunch of unique strings in the JS engine that will need to be garbage collected by logging like this:

```suggestion
		this._logService.debug('resizing', columns, rows);
```

Tyriar (changes requested):

Tyriar (approved):

Issue #217987, opened 2024-06-25
SCM commit graph context menu


benibenj:
It would be really nice if the SCM commit graph would have a context menu with actions such as
- Checkout Commit
- Revert Commit
- Maybe contributable ones (open commit on github, ...)
- Share
- . . .

PR #273616, opened 2025-10-27
feat(extensions/git): add revert and reset commit graph commands

Rajaniraiyn:
adds `Revert Commit` and `Reset Commit` to SCM graph view - part of #217987

- Introduced `git.graph.revertCommit` and `git.graph.resetCommit` commands
- Implemented functionality for reverting commits.

https://github.com/user-attachments/assets/123d2ae0-c1c9-4f48-9141-5df753d3ccac

### DIFF
--- a/extensions/git/package.nls.json
+++ b/extensions/git/package.nls.json
@@ -138,6 +138,8 @@
 	"command.graphCheckout": "Checkout",
 	"command.graphCheckoutDetached": "Checkout (Detached)",
 	"command.graphCherryPick": "Cherry Pick",
+	"command.graphRevertCommit": "Revert Commit",
+	"command.graphResetCommit": "Reset Commit",
 	"command.graphDeleteBranch": "Delete Branch",
 	"command.graphDeleteTag": "Delete Tag",
 	"command.graphCompareRef": "Compare with...",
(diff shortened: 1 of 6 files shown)


extensions/git/package.nls.json
```diff
@@ -135,6 +135,10 @@
 	"command.graphCheckout": "Checkout",
 	"command.graphCheckoutDetached": "Checkout (Detached)",
 	"command.graphCherryPick": "Cherry Pick",
+	"command.graphRevertCommit": "Revert Commit",
+	"command.graphResetCommit": "Reset Commit",
+	"command.graphCompareBranches": "Compare Branches...",
```
lszomoru: Are the "Compare Branch..." and "Compare Branches..." strings being used?

extensions/git/src/git.ts
```diff
@@ -2141,6 +2141,14 @@ export class Repository {
 		}
 	}
 
+	async revertCommit(commitHash: string, noEdit: boolean = false): Promise<void> {
```
lszomoru: Let's remove the `noEdit` for now as it does not seem to be used.

lszomoru (changes requested):
Could you please take a look at the comments? Thanks!

extensions/git/package.nls.json
Rajaniraiyn: opps. my bad. messed up the upstream merge

PR #273639, opened 2025-10-27
Ability to close a contributed chat session

DonJayamanne:
Ability to dispose the contributed copilot chat sessions.

For https://github.com/microsoft/vscode/issues/270852
For https://github.com/microsoft/vscode-internalbacklog/issues/5982

### DIFF
--- a/src/vscode-dts/vscode.proposed.chatSessionsProvider.d.ts
+++ b/src/vscode-dts/vscode.proposed.chatSessionsProvider.d.ts
@@ -170,6 +170,11 @@ declare module 'vscode' {
 		// TODO: Should we introduce our own type for `ChatRequestHandler` since not all field apply to chat sessions?
 		// TODO: Revisit this to align with code.
 		readonly requestHandler: ChatRequestHandler | undefined;
+
+		/**
+		 * Implement to handle when the Chat Session is closed.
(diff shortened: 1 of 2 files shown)


src/vscode-dts/vscode.proposed.chatSessionsProvider.d.ts
```diff
@@ -170,6 +170,11 @@ declare module 'vscode' {
 		// TODO: Should we introduce our own type for `ChatRequestHandler` since not all field apply to chat sessions?
 		// TODO: Revisit this to align with code.
 		readonly requestHandler: ChatRequestHandler | undefined;
+
+		/**
+		 * Implement to handle when the Chat Session is closed.
+		 */
+		close?(): void;
```
DonJayamanne, line 177: Going with something similar to `CustomDocument` & `PseudoTerminal`
Didn't know whether to go with `close` or `dispose`
Went with `close` as that feels the right terminology for chat sessions.

mjbvz (commented):
Can you add more info about the use case? A close method isn't a great fit for the current api design but maybe we can come up with something better

mjbvz:
What does the chat object do? Is it something like an event listen that gets updates from a server?

PR #273679, opened 2025-10-28
feat(git): add `no-verify` option for git commit

### LINKED ISSUE #244550: SCM: Add a options to commit without verify when run git's hooks fail
**Background:** 

When commit fails, sometimes it is because of running hooks. I hope there is a button to ignore the hooks script and continue to commit.

The Git command has a flag to ignore hooks

```
--no-verify, -n
```

**Why**:
Because sometimes, we just make some simple modifications. If we want to verify the hooks, we may need to install/reinstall the dependencies, which is a waste of time.

This prompt pops up, for Git noob, he will panic, adding an option is more friendly

Image

Image

louisunlimited:
- Fixes: #244550

## Overview
This pr adds the option for commit with `no-verify` flag on the menu next to green commit button.

Screenshot 2025-10-27 at 20 11 37

Also added an option to show the same `no-verify` option when dialog popup if hooks failed. 

Screenshot 2025-10-27 at 20 13 04

By default, users will get this warning if they want to proceed, and can be toggled on / off in settings.
Screenshot 2025-10-27 at 20 13 19

## Question
I did find it difficult to get a general test for git pre-commit hook errors in `extensions/git/src/git.ts`, as copilot has pointed out, would love to see some better suggestions. If you feel copilot's suggestions are better, feel free to accept that change and run CI.

## Testing Instructions

https://github.com/user-attachments/assets/d26d106c-5c7e-4264-81ea-18f5b808191e


louisunlimited:

alphaofficial:
Can't wait to have this 👀

PR #273736, opened 2025-10-28
Contributable chat context

alexr00:
Implicit context:

image

Explicit context:

image


alexr00 (commented):
@mjbvz could you take a look at the webview changes and see if they're ok?
@roblourens I'd love to get your feedback on all the chat related changes.

src/vs/workbench/contrib/webviewPanel/browser/webviewEditorInput.ts
```diff
@@ -50,7 +50,8 @@ export class WebviewInput extends EditorInput {
 	get resource() {
 		return URI.from({
 			scheme: Schemas.webviewPanel,
-			path: `webview-panel/webview-${this._resourceId}`
+			path: `webview-panel/webview-${this.providedId}`,
+			query: JSON.stringify({ id: this._resourceId })
```
mjbvz: I'm pretty sure this is a misspelling in the original code. `provided` should have been `provider`. It is meant to track the view type

Moving the actual id to the query makes me a bit worried since not all of our code does uri equality checks properly. Could we keep it in the path?

src/vs/workbench/contrib/webviewPanel/browser/webviewEditor.ts
```diff
@@ -153,6 +156,7 @@ export class WebviewEditor extends EditorPane {
 		await input.resolve();
 
 		if (token.isCancellationRequested || this._isDisposed) {
+			this._onDidChangeInput.fire(input instanceof WebviewInput ? input : undefined);
```
mjbvz: I don't think we should fire in these cases, especially the disposed case

src/vs/workbench/contrib/chat/browser/contrib/chatImplicitContext.ts
```diff
@@ -89,6 +94,19 @@ export class ChatImplicitContextContribution extends Disposable implements IWork
 						() => undefined,
 						500)(() => this.updateImplicitContext()));
 				}
+				const webviewEditor = this.findActiveWebviewEditor();
+				if (webviewEditor) {
+					const inputDisposable: MutableDisposable<IDisposable> = activeEditorDisposables.add(new MutableDisposable<IDisposable>());
+					activeEditorDisposables.add(webviewEditor.onDidChangeInput(input => {
```
mjbvz: Instead of adding a new event for the input, can you use `IWebviewService.onDidChangeActiveWebview`? That may get fired for all webviews though so you'll have to check 

Or maybe use the editor service? Would be better to avoid subscribing on the individual webvieweditor if we can avoid it

src/vscode-dts/vscode.proposed.chatContextProvider.d.ts
```diff
@@ -0,0 +1,35 @@
...
+	export interface ChatContextItem {
+		icon: ThemeIcon;
+		label: string;
+		modelDescription?: string;
+		value?: string;
+	}
+
+	export interface ChatContextProvider<T extends ChatContextItem = ChatContextItem> {
+
+		provideChatContext?(options: {}, token: CancellationToken): ProviderResult<T[] | undefined>;
```
mjbvz: [nit] I think ProviderResult already makes its value `| undefined`

mjbvz: When is this called? Could the name make this more clear?

src/vscode-dts/vscode.proposed.chatContextProvider.d.ts
```diff
@@ -0,0 +1,35 @@
...
+		// id is from `ChatPromptReference`
+		export function registerChatContextProvider(selector: DocumentSelector, id: string, provider: ChatContextProvider): Disposable;
+
+	}
+
+	export interface ChatContextItem {
+		icon: ThemeIcon;
+		label: string;
+		modelDescription?: string;
+		value?: string;
```
mjbvz, line 23: What's this used for?

src/vscode-dts/vscode.proposed.chatContextProvider.d.ts
```diff
@@ -0,0 +1,35 @@
...
+	}
+
+	export interface ChatContextProvider<T extends ChatContextItem = ChatContextItem> {
+
+		provideChatContext?(options: {}, token: CancellationToken): ProviderResult<T[] | undefined>;
+
+		provideChatContextForResource?(resource: Uri, options: {}, token: CancellationToken): ProviderResult<T | undefined>;
+		resolveChatContext(context: T, token: CancellationToken): ProviderResult<ChatContextItem>;
```
mjbvz, line 55: Need some doc comments to help understand these and when they are called

src/vs/workbench/workbench.common.main.ts
```diff
@@ -379,6 +379,9 @@ import './contrib/remoteCodingAgents/browser/remoteCodingAgents.contribution.js'
 // Chat Sessions
 import './contrib/chat/browser/chatSessions.contribution.js';
 
+// Chat Context
+import './contrib/chat/browser/chatContext.contribution.js';
+
 // Code Actions
 import './contrib/codeActions/browser/codeActions.contribution.js';
```
bpasero: can we move `chatSessions.contribution` and  `chatContext.contribution ` under the already existing `// Chat` area above?

src/vscode-dts/vscode.proposed.chatContextProvider.d.ts
alexr00: This is the value that is given to the llm as context.

src/vs/workbench/contrib/chat/browser/contrib/chatImplicitContext.ts
alexr00: Actually manage to make it even more simple

src/vscode-dts/vscode.proposed.chatContextProvider.d.ts
```diff
@@ -0,0 +1,58 @@
...
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+
+declare module 'vscode' {
+
+	// https://github.com/microsoft/vscode/issues/271104 @alexr00
+
+	export namespace workspace {
```
roblourens, line 11: `chat` namespace?

src/vs/workbench/contrib/chat/browser/contrib/chatImplicitContext.ts
```diff
@@ -207,7 +226,21 @@ export class ChatImplicitContextContribution extends Disposable implements IWork
 			}
 		}
 
-		const uri = newValue instanceof URI ? newValue : newValue?.uri;
+		const webviewEditor = this.findActiveWebviewEditor();
+		if (webviewEditor?.input?.resource) {
+			const webviewContext = await this.chatContextService.contextForResource(webviewEditor.input.resource);
+			if (webviewContext?.value) {
```
roblourens, line 232: Does this mean implicit context can't use the 'resolve' flow too?

roblourens (approved):

src/vscode-dts/vscode.proposed.chatContextProvider.d.ts
alexr00: Will move in an immediate follow up PR.

src/vs/workbench/contrib/chat/browser/contrib/chatImplicitContext.ts
alexr00: For now, yes, but I will enable it.

PR #273742, opened 2025-10-28
Testing - Uncompleted result should be marked as skipped

### LINKED ISSUE #272888: [Testing] Enqued test that was never completed should be shown as skipped, not as the last result

- VS Code Version: 8fd821a0c7c0a04724bc51f0566b438e6f35052e
- OS Version: macOS 26.0.1

Steps to Reproduce:

1. Have a Test Provider that provides two tests A and B.
2. Start a test run and mark both tests as passed. 
3. Start another test run and do: `testRun.enqueued(A); testRun.enqueued(B); testRun.passed(B); testRun.end();`

Expected Behaviour: The user can tell from the UI that test A was never completed.

Actual Behaviour:
You'll see how both A and B switch to the "enqueued" icon, and then both switch to the "passed" icon. To the user, this looks like both tests were executed and both tests are green.

The reason for this is that:

- [tests that were never finished are switched back to `Unset`](https://github.com/microsoft/vscode/blob/8fd821a0c7c0a04724bc51f0566b438e6f35052e/src/vs/workbench/contrib/testing/common/testResult.ts#L467-L471)
- [`Unset` results are ignored in the Icon selection](https://github.com/microsoft/vscode/blob/8fd821a0c7c0a04724bc51f0566b438e6f35052e/src/vs/workbench/contrib/testing/common/testResultService.ts#L129-L131)

This is a footgun, because a buggy test extension that forgets executing or recording a test result will produce false negatives. Extensions can work around this by tracking unfinished tests themselves, and marking them as skipped before ending the test run.

Skn0tt:
Closes https://github.com/microsoft/vscode/issues/272888

### DIFF
--- a/src/vs/workbench/contrib/testing/common/testResult.ts
+++ b/src/vs/workbench/contrib/testing/common/testResult.ts
@@ -465,7 +465,7 @@ export class LiveTestResult extends Disposable implements ITestResult {
 		task.output.end();
 
 		this.setAllToState(
-			TestResultState.Unset,
+			TestResultState.Skipped,
 			taskId,
 			t => t.state === TestResultState.Queued || t.state === TestResultState.Running,
 		);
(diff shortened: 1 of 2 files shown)


connor4312 (approved):

connor4312:
thanks!

mjbvz (approved):

PR #273805, opened 2025-10-28
chat: add command for switching to custom chat agent

### LINKED ISSUE #270662: Enable keyboard shortcuts for custom chat modes
I should be able to set a keyboard shortcut for custom chat modes. I can do it for the builtin modes with these commands

Image

Image

But we don't register these for custom modes

We register these actions for builtin modes here: https://github.com/microsoft/vscode/blob/46b60ac2ce4fce4eab4f45d488b1b0942d5a0427/src/vs/workbench/contrib/chat/browser/actions/chatActions.ts#L444-L447

I think it needs to work differently for custom modes because they are dynamic. I think that the ChatModeService should handle dynamically registering actions for custom modes when they are discovered.

https://github.com/microsoft/vscode/blob/46b60ac2ce4fce4eab4f45d488b1b0942d5a0427/src/vs/workbench/contrib/chat/common/chatModes.ts#L33

As a starting point, we can just register the commands so people can customize the shortcuts in the keyboard shortcuts editor. As a next step, we could figure out the UX to make it easy to discover from the "Manage Mode" command

bryanchen-d:
Register custom agent/mode into the command list so user can assign keybinding for the custom agents.

Fixes https://github.com/microsoft/vscode/issues/270662


src/vs/workbench/contrib/chat/common/chatModes.ts
```diff
@@ -196,6 +218,23 @@ export class ChatModeService extends Disposable implements IChatModeService {
 	private getCustomModes(): IChatMode[] {
 		return this.chatAgentService.hasToolsAgent ? Array.from(this._customModeInstances.values()) : [];
 	}
+
+	setActionsManager(manager: IChatCustomAgentActionsManager | undefined): void {
```
roblourens: I don't see anywhere that calls `setActionsManager`?

src/vs/workbench/contrib/chat/browser/actions/chatActions.ts
```diff
@@ -439,6 +439,69 @@ abstract class ModeOpenChatGlobalAction extends OpenChatGlobalAction {
...
+
+		let selectedModeName: string | undefined;
+		const agentName = opts?.mode;
+		if (agentName) {
+			// If agent name is provided as argument, use it directly
+			const selectedMode = modes.find(m => m.name === agentName);
+			selectedModeName = selectedMode?.name;
+		} else {
+			// Show quick pick to select - use a setTimeout to defer execution after accessor becomes invalid
+			const selected = await quickInputService.pick(
```
roblourens: I don't think we need an action with a separate quickpick for this case, users can already open the mode picker with the keybinding and type to pick a mode

src/vs/workbench/contrib/chat/browser/actions/chatActions.ts
```diff
@@ -439,6 +439,69 @@ abstract class ModeOpenChatGlobalAction extends OpenChatGlobalAction {
...
+		const modes = chatModeService.getModes().custom;
+		if (modes.length === 0) {
+			return undefined;
+		}
+
+		let selectedModeName: string | undefined;
+		const agentName = opts?.mode;
+		if (agentName) {
+			// If agent name is provided as argument, use it directly
+			const selectedMode = modes.find(m => m.name === agentName);
```
roblourens: The ToggleChatModeAction handles this case of switching to a mode by id, so users can already set up their keybindings that way. The reason I wanted separate actions vs one action with args is that this way is not discoverable at all, and we have no UX for it in the keyboard shortcuts editor.

src/vs/workbench/contrib/chat/common/chatModes.ts
bryanchen-d: That was for creating separate action per each agent, will use it if we decide to do that.

src/vs/workbench/contrib/chat/browser/actions/chatActions.ts
bryanchen-d: reverted back to use separate actions so we don't need the quickpick in the case of using single action.

src/vs/workbench/contrib/chat/browser/actions/chatActions.ts
bryanchen-d: reverted back to using separate actions.

src/vs/workbench/contrib/chat/common/chatModes.ts
roblourens: Instead of setting a delegate type and calling back to it, can the contribution listen to an event on the service and keep itself in sync? That's how we typically do this and I think it's nicer that the service exposes an interface that others can build on, it should not have to know about "actions" or what this feature is.

roblourens (commented):
Based on our standup conversation I would also suggest mentioning this in Slack so that others are not too surprised if they see this in the product and don't understand why we did it this way.

src/vs/workbench/contrib/chat/common/chatModes.ts
bryanchen-d: That make sense. 

Updated adding a `onDidAddCustomMode` and `onDidRemoveCustomMode` event to handle registration and unregistration.
One question though: I think the add/remove events kind of overlap with the `onDidChangeChatModes` event, it will be cleaner if we can merge them, but that event is also connected with `chatAgentService.onDidChangeAgents`, which seems a broader change if we want to make it consistent. Any comments?

roblourens: I would suggest a 'onDidChange' rather than having two separate events. Or, an observable

src/vs/workbench/contrib/chat/browser/chat.contribution.ts
```diff
@@ -904,6 +904,63 @@ class ChatAgentSettingContribution extends Disposable implements IWorkbenchContr
...
+
+		// Listen for custom mode changes by tracking snapshots
+		this._register(this.chatModeService.onDidChangeChatModes(() => {
+			const { custom } = this.chatModeService.getModes();
+			const currentModeIds = new Set<string>();
+
+			// Register new modes
+			for (const mode of custom) {
+				currentModeIds.add(mode.id);
+				if (!this._previousCustomModeIds.has(mode.id)) {
```
roblourens: Does this have the same values as the keys of _modeActionDisposables?

bryanchen-d: You are right. This is not needed.

src/vs/workbench/contrib/chat/browser/chat.contribution.ts
```diff
@@ -904,6 +904,59 @@ class ChatAgentSettingContribution extends Disposable implements IWorkbenchContr
...
 
+
+/**
+ * Workbench contribution to register actions for custom chat modes via events
+ */
+class ChatAgentActionsContribution extends Disposable implements IWorkbenchContribution {
+
+	static readonly ID = 'workbench.contrib.chatAgentActions';
+
+	private readonly _modeActionDisposables = new DisposableMap<string>();
```
jrieken, line 948: this will leak, needs to be added `this._store`

jrieken (dismissed):
left some nit for dispoables

src/vs/workbench/contrib/chat/browser/chat.contribution.ts
bryanchen-d: updated, thanks!

justschen (approved):

roblourens (approved):

PR #273829, opened 2025-10-28
feat(server): minimal websocket collab server with rooms

Koiiichi:


mjbvz:
Looks like a good candidate to ship as an extension but not something we'd include as a built-in extension

PR #273831, opened 2025-10-28
chore: non-interactive `workbench.action.exportLogs` variant

rwoll:
During automation scenarios, we want to be able to export *all* logs without interacting with the QuickInput or system Save As dialog.

### DIFF
--- a/src/vs/workbench/contrib/output/browser/output.contribution.ts
+++ b/src/vs/workbench/contrib/output/browser/output.contribution.ts
@@ -720,7 +720,7 @@ class OutputContribution extends Disposable implements IWorkbenchContribution {
 					}],
 				});
 			}
-			async run(accessor: ServicesAccessor, outputPath?: URI): Promise<void> {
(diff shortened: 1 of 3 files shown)


src/vs/workbench/contrib/output/browser/output.contribution.ts
```diff
@@ -718,7 +720,7 @@ class OutputContribution extends Disposable implements IWorkbenchContribution {
 					}],
 				});
 			}
-			async run(accessor: ServicesAccessor): Promise<void> {
+			async run(accessor: ServicesAccessor, outputPath?: URI): Promise<void> {
```
amunger: It's kind of strange that providing an output path is the flag for exporting all logs. It's probably better to just add an additional command

rwoll: Thanks - will update!

amunger (approved):

PR #273837, opened 2025-10-28
Support case-insensitive glob

### LINKED ISSUE #10633: Glob matching should be case insensitive
- VSCode Version: 1.4.0
- OS Version: Win10 1511

Steps to Reproduce:
1. using VSCode, open a directory stored in an NTFS volume, which has files named FooContainer.cs
2. press CTRL+SHIFT+F to invoke Find in Files
3. press "..." to show 'files to include'
4. enter "*_/_container.cs"
5. search for something that is known to exist in the file.
6. results: nothing is found
7. replace 'files to include' with "*_/_Container.cs" (only change is the upper case C instead of lower case)
8. search again, this time results are found as expected.

Since Windows NTFS is not a case-sensitive file system, the expected results is that I can search in files using any case in the 'files to include'/'files to exclude'.

dmitrivMS:
First step towards fixing #10633 and many related duplicates. 

This PR adds an option to enable case-insensitive glob expressions and matching. 
The new option is not used here and default behavior is the same as before. 
The option will be used by follow up PRs to enable case-insensitive behavior in various features.


dmitrivMS:
NOTE: I have file search working locally with this and other changes, but prefer to split the changes into multiple PRs for ease of review and/or discussions/feedback.

src/vs/base/common/strings.ts
```diff
@@ -438,13 +438,19 @@ export function equalsIgnoreCase(a: string, b: string): boolean {
 	return a.length === b.length && compareSubstringIgnoreCase(a, b) === 0;
 }
 
+export function equals(a: string | undefined, b: string | undefined, ignoreCase?: boolean): boolean {
```
dmitrivMS, line 441: This is not used in this PR, but is used in other parts - particularly where === needs to be replaced with a comparison that can take ignoreCase parameter.

src/vs/base/common/strings.ts
```diff
@@ -438,13 +438,19 @@ export function equalsIgnoreCase(a: string, b: string): boolean {
...
+	return a === b || !!ignoreCase && a !== undefined && b !== undefined && equalsIgnoreCase(a, b);
```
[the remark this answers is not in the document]
dmitrivMS: ok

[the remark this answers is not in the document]
dmitrivMS: wrong

dmitrivMS:
FYI - second PR dependent on this (in draft):
https://github.com/microsoft/vscode/pull/273860

src/vs/base/common/glob.ts
```diff
@@ -522,14 +546,19 @@ function toRegExp(pattern: string): ParsedStringPattern {
  * * `[]` to declare a range of characters to match in a path segment (e.g., `example.[0-9]` to match on `example.0`, `example.1`, …)
  * * `[!...]` to negate a range of characters to match in a path segment (e.g., `example.[!0-9]` to match on `example.a`, `example.b`, but not `example.0`)
  */
-export function match(pattern: string | IRelativePattern, path: string): boolean;
+export function match(pattern: string | IRelativePattern, path: string, options?: IGlobOptions): boolean;
+export function match(expression: IExpression, path: string, options?: IGlobOptions): boolean;
 export function match(expression: IExpression, path: string, hasSibling?: (name: string) => boolean): string /* the matching pattern */;
-export function match(arg1: string | IExpression | IRelativePattern, path: string, hasSibling?: (name: string) => boolean): boolean | string | null | Promise<string | null> {
+export function match(arg1: string | IExpression | IRelativePattern, path: string, arg3?: IGlobOptions | ((name: string) => boolean)): boolean | string | null | Promise<string | null> {
```
bpasero: I wonder: can we move `hasSibling?: (name: string) => boolean` into `IGlobOptions` as optional thing and then the signature becomes simpler and we support everything in all cases?

bpasero (changes requested):
This is overall a well engineered PR as far as I can tell. Only 1 comment about how we could simplify the `match` method. See below.

src/vs/base/common/glob.ts
dmitrivMS: I would suggest we remove support for hasSibling parameter in this function:
- It is not used anywhere in the codebase but in tests
- It's only sugar for calling parse(...) and then calling expression with hasSibling, so nothing is lost
- Combining it with options will change return type to string | boolean which is confusing for the caller

Agreed?

dmitrivMS: Also, the return type of hasSibling overload is wrong right now - it should be "string | null" or even "string | null | boolean" since hasSibling is optional.

bpasero: Hm, I thought `hasSibling` would still be used, maybe by the explorer. If you can ensure its not used, then yes lets remove it, but please double check, it might be just a weird typing issue. Please check all callers.

dmitrivMS: It is used, but as a separate call to the parsed expression, not through match(...).
I'll update the PR.

dmitrivMS: Updated. And I absolutely went through each and every caller of match and parse for this and other PRs.

bpasero (approved):
@dmitrivMS can we do a follow up PR then to adopt a few places where we seem to be doing some `toLowercase` magic to get case insensitive matching today:

- https://github.com/microsoft/vscode/blob/1f645aa8e556c67bdf481f60013d3ac4a3d8cfda/src/vs/workbench/contrib/notebook/common/notebookCommon.ts#L957

- https://github.com/microsoft/vscode/blob/1f645aa8e556c67bdf481f60013d3ac4a3d8cfda/src/vs/workbench/contrib/notebook/common/notebookCommon.ts#L959

- https://github.com/microsoft/vscode/blob/1f645aa8e556c67bdf481f60013d3ac4a3d8cfda/src/vs/workbench/contrib/notebook/common/notebookProvider.ts#L82

- https://github.com/microsoft/vscode/blob/1f645aa8e556c67bdf481f60013d3ac4a3d8cfda/src/vs/workbench/contrib/notebook/common/notebookProvider.ts#L88

- https://github.com/microsoft/vscode/blob/1f645aa8e556c67bdf481f60013d3ac4a3d8cfda/src/vs/workbench/contrib/notebook/common/notebookProvider.ts#L100

- https://github.com/microsoft/vscode/blob/1f645aa8e556c67bdf481f60013d3ac4a3d8cfda/src/vs/workbench/contrib/notebook/common/notebookProvider.ts#L102

- https://github.com/microsoft/vscode/blob/1f645aa8e556c67bdf481f60013d3ac4a3d8cfda/src/vs/workbench/services/editor/common/editorResolverService.ts#L223

- https://github.com/microsoft/vscode/blob/1f645aa8e556c67bdf481f60013d3ac4a3d8cfda/src/vs/workbench/services/label/common/labelService.ts#L196

dmitrivMS:
@bpasero Yep, absolutely, I had the same idea and started on it, but wanted to have it as a separate PR.
My second PR builds on this one without touching UI or changing behavior, so it would be a good place to make these changes as well since they don't change behavior either.

dmitrivMS:
Follow up PR: https://github.com/microsoft/vscode/pull/273837

Issue #273727, opened 2025-10-28
Activity badges: update SVG mask for in-progress


mrleemurray:
Currently the `update` badge for downloading new updates looks different from other badges (the white stroke), is a little vague & doesn't represent the various states of checking, downloading, & updating. I also couldn't find the icon used in the current implementation in the codicon library.

Screenshot:
<img width="93" height="44" alt="Image" src="https://github.com/user-attachments/assets/a60ca765-d2d2-4805-9a33-021d705a8d8b" />

Proposal to indicate state more clearly:

<img width="1012" height="304" alt="Image" src="https://github.com/user-attachments/assets/11db5bfa-bfc7-4315-9873-08b257569426" />

UX/Design Issue:
This issue is related to user interface, design, or user experience.

Environment Information:
- VS Code Version: 1.106.0-insider
- OS: macOS (arm64)
- Workspace: 2 folder(s) open

---
This issue was filed using the Issue Filer extension

bpasero:
This is done?

eleanorjboyd:
verified that it no longer has the white stroke around the blue badge but was unable to get it in all the states to check each icon. What is the desired repro here?

PR #273897, opened 2025-10-29
activity - change to codicon styled progress (#273727)

bpasero:
@mrleemurray turns out badges already support codicons. I went ahead and changed the progress badge to use our animated loading icon:

Recording 2025-10-29 at 10 17 50

Overall the icon appears quite small, the previous one appeared larger:

image


bpasero:
Which is hard because the badge itself is small:

Recording 2025-10-29 at 10 37 02

PR #274001, opened 2025-10-29
Expand hover setting to allow for key modifier mode

### LINKED ISSUE #72025: Disable hover but keep Ctrl+Hover
Hover can be very frustrating when selecting text etc. Longer delay (eg 1500ms) solves this but when you actually need the hover information it is frustrating to wait. 

Having the full hover information by pressing control and hover the mouse on specific code is very nice. Please leave this functionality while let us disable the simple hover (the one without pressing Ctrl). In other words separate "hover" and "ctrl+hover" functionalities, with two seaparate options for turning them on/off.

benvillalobos:
Fixes https://github.com/microsoft/vscode/issues/72025

## Changes Made
- Modifies existing `hover.enabled` setting from a boolean to an enum with values [`on`, `off`, `onKeyboardModifier`]
  - Mimics how [`inlayHints.enabled`](https://github.com/microsoft/vscode/commit/5a0df1cec15847cf7f234d3fe77397fc5f666581#diff-0dfa7db579eface8250affb76bc88717725a121401d4d8598bc36b92b0b6ef62) went from boolean to enum
- When set to `onKeyboardModifier`, hover only appears when `ctrl`, `cmd`, or `alt` are held down.
  - The key is determined by which key is currently set at `Multi Cursor Modifier`. If MCM is set to `ctrlCmd`, hover triggers on `alt` and vice versa.

https://github.com/user-attachments/assets/8ee8f4c7-0b2c-4cbc-b6d9-43a1b143145f

## To Test
- Set `Hover: Enabled` to `onKeyboardModifier`
- Observe your current setting for `Multi Cursor Modifier`
- Observe that hovers do not show unless the opposite value of `Multi Cursor Modifier` is held down
  - If set to `ctrlCmd`, hover should only show when `alt` is held down.
  - If set to `alt`, hover should show when either `ctrl` or `cmd` (macOS) are held down.
- Set `Multi Cursor Modifier` to whatever it is not currently
- Observe that hovers do not show unless the opposite button is pressed
- Also worth ensuring `Hover: Enabled` `on` and `off` work as expected

### DIFF
--- a/src/vs/editor/browser/config/migrateOptions.ts
+++ b/src/vs/editor/browser/config/migrateOptions.ts
@@ -251,3 +251,10 @@ registerEditorSettingMigration('inlineSuggest.edits.codeShifting', (value, read,
 		write('inlineSuggest.edits.allowCodeShifting', value ? 'always' : 'never');
 	}
 });
+
+// Migrate Hover
+registerEditorSettingMigration('hover.enabled', (value, read, write) => {
+	if (typeof value === 'boolean') {
+		write('hover.enabled', value ? undefined : 'off');
+	}
+});
--- a/src/vs/editor/browser/config/migrateOptions.ts
+++ b/src/vs/editor/browser/config/migrateOptions.ts
@@ -255,6 +255,6 @@ registerEditorSettingMigration('inlineSuggest.edits.codeShifting', (value, read,
 // Migrate Hover
 registerEditorSettingMigration('hover.enabled', (value, read, write) => {
 	if (typeof value === 'boolean') {
-		write('hover.enabled', value ? undefined : 'off');
+		write('hover.enabled', value ? 'on' : 'off');
 	}
 });
--- a/src/vs/editor/common/config/editorOptions.ts
+++ b/src/vs/editor/common/config/editorOptions.ts
@@ -2344,7 +2344,7 @@ class EditorHover extends BaseEditorOption<EditorOption.hover, IEditorHoverOptio
 					markdownEnumDescriptions: [
 						nls.localize('hover.enabled.on', "Hover is enabled."),
 						nls.localize('hover.enabled.off', "Hover is disabled."),
-						nls.localize('hover.enabled.onKeyboardModifier', "Hover is shown when holding {0} or Alt (the opposite modifier of `#editor.multiCursorModifier#`)", platform.isMacintosh ? `Cmd` : `Ctrl`)
+						nls.localize('hover.enabled.onKeyboardModifier', "Hover is shown when holding `{0}` or `Alt` (the opposite modifier of `#editor.multiCursorModifier#`)", platform.isMacintosh ? `Command` : `Control`)
 					],
 					description: nls.localize('hover.enabled', "Controls whether the hover is shown.")
 				},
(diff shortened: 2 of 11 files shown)


src/vs/editor/common/config/editorOptions.ts
```diff
@@ -2338,8 +2338,14 @@ class EditorHover extends BaseEditorOption<EditorOption.hover, IEditorHoverOptio
...
 			{
 				'editor.hover.enabled': {
-					type: 'boolean',
+					type: 'string',
+					enum: ['on', 'off', 'onKeyboardModifier'],
 					default: defaults.enabled,
+					markdownEnumDescriptions: [
+						nls.localize('hover.enabled.on', "Hover is enabled."),
+						nls.localize('hover.enabled.off', "Hover is disabled."),
+						nls.localize('hover.enabled.onKeyboardModifier', "Hover is shown when the opposite modifier key of the Multi Cursor Modifier is pressed. (ctrlCmd or alt)")
```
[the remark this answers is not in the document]
benvillalobos: How verbose are we okay being with these descriptions?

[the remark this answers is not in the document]
alexdima: How about two distinct strings for mac and everyone else. In the mac string we use "cmd" and everywhere else we use "ctrl". We can also use the syntax to link to the other setting e.g.:

```
"Hover is shown when holding cmd or alt (the opposite modifier of `#editor.multiCursorModifier#`)"
```

src/vs/editor/common/config/editorOptions.ts
```diff
@@ -2374,8 +2380,12 @@ class EditorHover extends BaseEditorOption<EditorOption.hover, IEditorHoverOptio
 			return this.defaultValue;
 		}
 		const input = _input as Unknown<IEditorHoverOptions>;
+		// Backward compatibility
```
alexdima: The better way to do backward compatibility is via `migrateOptions.ts` which would also update the settings.json in the user's settings or the workspace settings. See e.g.
```
registerEditorSettingMigration('lightbulb.enabled', (value, read, write) => {
	if (typeof value === 'boolean') {
		write('lightbulb.enabled', value ? undefined : 'off');
	}
});
```

alexdima (commented):
LGTM. I left a couple of minor comments about how to better migrate and about the description.

@aiday-mar please also review

src/vs/editor/browser/config/migrateOptions.ts
```diff
@@ -251,3 +251,10 @@ registerEditorSettingMigration('inlineSuggest.edits.codeShifting', (value, read,
 		write('inlineSuggest.edits.allowCodeShifting', value ? 'always' : 'never');
 	}
 });
+
+// Migrate Hover
+registerEditorSettingMigration('hover.enabled', (value, read, write) => {
+	if (typeof value === 'boolean') {
+		write('hover.enabled', value ? undefined : 'off');
```
benvillalobos: Mimics `lightbulb.enabled`.
Is using `undefined` for true (the default value) correct here? I assume this means it does not write a value or clears the value if there is one.

src/vs/editor/common/config/editorOptions.ts
benvillalobos: Removed this code in favor of the migrateOptions route. Thanks for the tip!

src/vs/editor/browser/config/migrateOptions.ts
aiday-mar: Hi Ben that is a good question. Just to be on the safe side I changed it to 'on' for the case when the boolean is true.

alexdima (approved):
LGTM

PR #274016, opened 2025-10-29
Adopt uuids for generating webview nonces

mjbvz:

### DIFF
--- a/extensions/markdown-language-features/src/util/dom.ts
+++ b/extensions/markdown-language-features/src/util/dom.ts
@@ -11,11 +11,3 @@ export function escapeAttribute(value: string | vscode.Uri): string {
 		.replace(/'/g, '&#39;');
 }
 
-export function getNonce() {
-	let text = '';
-	const possible = 'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789';
-	for (let i = 0; i < 64; i++) {
-		text += possible.charAt(Math.floor(Math.random() * possible.length));
-	}
-	return text;
-}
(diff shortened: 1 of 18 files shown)


rzhao271 (dismissed):

extensions/markdown-language-features/src/util/dom.ts
```diff
@@ -11,11 +11,3 @@ export function escapeAttribute(value: string | vscode.Uri): string {
 		.replace(/'/g, '&#39;');
 }
```
rwoll, line 13: Instead of deleting `getNonce(…)`, I recommend using it to alias `generateUuid()`:

```
getNonce() { return generateUuid(); }
```

Your `nonce`s now happen to be UUIDs but they conceptually are there own kind of thing.

If you change from UUID to something else in the future, it'll be much easier to find all references of `getNonce` rather than `generateUuid` which could be used in a variety of non-nonce contexts.

extensions/markdown-language-features/src/util/uuid.ts
```diff
@@ -0,0 +1,58 @@
...
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+/**
+ * Copied from src/vs/base/common/uuid.ts
+ */
+export function generateUuid(): string {
+	// use `randomUUID` if possible
+	if (typeof crypto.randomUUID === 'function') {
```
rwoll, line 11: What are some examples in VSCode context, that `randomUUID` would not be available?

Tyriar (approved):

PR #274134, opened 2025-10-30
debt - set `es2024` as target for `esbuild`

bpasero:
Our `tsconfig.json` is using ES2024

fyi @mjbvz @jrieken


mjbvz:
Thanks for finding this! I've actually got another PR that should pick this automatically so we hopefully don't get out of sync again: https://github.com/microsoft/vscode/pull/274139

PR #274139, opened 2025-10-30
Pick up esbuild target from tsconfig files

mjbvz:


build/lib/tsconfig.ts
```diff
@@ -0,0 +1,24 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+import { dirname } from 'path/posix';
```
mjbvz:
```suggestion
import { dirname } from 'path';
```

bpasero (approved):

PR #274173, opened 2025-10-30
SCM - plumbing to support artifacts in the repositories view

lszomoru:


extensions/git/src/artifactProvider.ts
```diff
@@ -0,0 +1,62 @@
...
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { LogOutputChannel, SourceControlArtifactProvider, SourceControlArtifactGroup, SourceControlArtifact, Event, EventEmitter, ThemeIcon, l10n, workspace, Uri } from 'vscode';
+import { IDisposable } from './util';
+import { Repository } from './repository';
+
+export class GitArtifactProvider implements SourceControlArtifactProvider, IDisposable {
+	private readonly _onDidChangeArtifacts = new EventEmitter<string>();
```
Tyriar, line 11: Register?

extensions/git/src/commands.ts
```diff
@@ -5182,6 +5182,16 @@ export class CommandCenter {
 		config.update(setting, !enabled, true);
 	}
 
+	@command('git.repositories.checkout', { repository: true })
+	async artifactCheckout(repository: Repository, artifact: SourceControlArtifact): Promise<void> {
+		console.log(repository, artifact);
```
Tyriar, line 5187: Accident?

src/vs/workbench/api/browser/mainThreadSCM.ts
```diff
@@ -534,6 +562,14 @@ class MainThreadSCMProvider implements ISCMProvider {
 		this._historyProvider.get()?.$onDidChangeHistoryItemRefs(historyItemRefs);
 	}
 
+	$onDidChangeArtifacts(group: string): void {
+		if (!this.artifactProvider.get()) {
+			return;
+		}
+
+		this._artifactProvider.get()?.$onDidChangeArtifacts(group);
```
Tyriar, line 570: First part is redundant? get is a function as well and could theoretically return different things, which I'm assuming is why the `?` was still there

```suggestion
		this._artifactProvider.get()?.$onDidChangeArtifacts(group);
```

src/vs/workbench/api/common/extHostSCM.ts
```diff
@@ -887,6 +910,8 @@ class ExtHostSourceControl implements vscode.SourceControl {
 		this._acceptInputDisposables.dispose();
 		this._actionButtonDisposables.dispose();
 		this._statusBarDisposables.dispose();
+		this._historyProviderDisposable.dispose();
+		this._artifactProviderDisposable.dispose();
```
Tyriar, line 914: nit: `MutableDisposable`s registered on class init are a nicer way of handling this. That way the variable declaration and registering are in a single spot.

Tyriar (approved):

PR #274186, opened 2025-10-30
More plumbing through of session resources instead of ids

mjbvz:


roblourens (approved):

src/vs/workbench/contrib/chat/common/chatModel.ts
```diff
@@ -1360,11 +1362,16 @@ export class ChatModel extends Disposable implements IChatModel {
 
 	// TODO to be clear, this is not the same as the id from the session object, which belongs to the provider.
```
roblourens, line 1363: ugh I think this TODO is very obsolete, no async init today, adding to debt list...

PR #274194, opened 2025-10-30
Improve Katex copy robustness

pwang347:
Follow up for https://github.com/microsoft/vscode/pull/271590#issuecomment-3418382156


src/vs/workbench/contrib/markdown/common/markedKatexExtension.ts
```diff
@@ -45,6 +49,19 @@ export namespace MarkedKatexExtension {
 		};
 	}
 
+	function escapeHtmlAttribute(text: string): string {
```
mjbvz: We have `htmlAttributeEncodeValue` that looks like it does the same thing

src/vs/workbench/contrib/markdown/common/markedKatexExtension.ts
```diff
@@ -32,11 +32,15 @@ export namespace MarkedKatexExtension {
...
-				out = katex.renderToString(token.text, {
+				const html = katex.renderToString(token.text, {
 					...options,
 					throwOnError: true,
 					displayMode: token.displayMode,
 				});
+
+				// Wrap in a container with data-latex attribute as a fallback for extracting the original LaTeX source
+				// This ensures we can always retrieve the source even if the annotation element is not present
+				out = `<span class="vscode-katex-container" data-latex="${escapeHtmlAttribute(token.text)}">${html}</span>`;
```
mjbvz: Could we extract the class name to a constant string that we can reference in the copy logic too?

mjbvz (changes requested):

src/vs/workbench/contrib/markdown/common/markedKatexExtension.ts
pwang347: thanks! I updated this

src/vs/workbench/contrib/markdown/common/markedKatexExtension.ts
pwang347: done! also added the name of the new data-latex attribute to a constant

mjbvz (approved):

PR #274228, opened 2025-10-31
Add screenreader warning on terminal tool notifications

### LINKED ISSUE #258590: Warning not announced to screenreader
Testing https://github.com/microsoft/vscode/issues/258230

With NVDA, I was able to tab to the shell integration link, but I was not able to see any announcement for the "Enable shell integration" message.

benvillalobos:
Fixes https://github.com/microsoft/vscode/issues/258590

## Context
When a terminal tool runs and leaves a [`toolResultMessage` behind](https://github.com/benvillalobos/vscode/blob/f93afd7150d850608c83765df9aaac267ef5ee9e/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/runInTerminalTool.ts#L558), that message does not get announced by the screenreader.

## Changes Made
- Makes `chatTerminalToolProgressPart`s more accessible
  - Assigns tabIndex = 0
  - Assigns role = region
  - Assigns an aria label
  - If a message was left behind by a terminal tool, includes it in the aria label
  - Ensures screenreaders announce when a message is left behind by a terminal tool


src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
```diff
@@ -148,6 +152,18 @@ export class ChatTerminalToolProgressPart extends BaseChatToolInvocationSubPart
 
 		const progressPart = this._register(_instantiationService.createInstance(ChatProgressSubPart, elements.container, this.getIcon(), terminalData.autoApproveInfo));
 		this.domNode = progressPart.domNode;
+		this.domNode.tabIndex = 0;
```
Tyriar: Isn't it bad practice/UX to make something focusable when it can't actually be interacted with?

src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
```diff
@@ -148,6 +152,18 @@ export class ChatTerminalToolProgressPart extends BaseChatToolInvocationSubPart
...
+		this.domNode.role = 'region';
+
+		if (pastTenseMessage) {
+			const sanitizedMessage = stripIcons(pastTenseMessage);
+			this.domNode.ariaLabel = localize('terminalToolCommand', "{0} command: {1} - {2}", toolInvocation.toolId, command, sanitizedMessage);
```
Tyriar: `pastTenseMessage` is rendered via `ChatMarkdownContentPart`, shouldn't any a11y accommodations around it live there?

src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
```diff
@@ -148,6 +152,18 @@ export class ChatTerminalToolProgressPart extends BaseChatToolInvocationSubPart
...
+			if (this._configurationService.getValue(AccessibilityWorkbenchSettingId.VerboseChatProgressUpdates)) {
+				status(sanitizedMessage);
+			}
```
Tyriar: Same as above, shouldn't this live in `ChatMarkdownContentPart`?

Tyriar (changes requested):

src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
meganrogge: It can be interactive though - does this link not exist anymore?

Screenshot 2025-10-31 at 12 40 21 PM

meganrogge: If not, we should remove `tabIndex`. I was suggesting a `tabIndex` on the link.

src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
benvillalobos: I've refactored the logic:
- `CharMarkdownContentPart` now receives an optional `accessibilityOptions` which contains a `statusMessage`. It sets `statusMessage` as its ariaLabel and announces it when `VerboseChatProgressUpdates` is set.
- `ChatTerminalToolProgressPart`, when given a `pastTenseMessage`, passes the localized-and-icon-stripped message to the markdown part to announce and set as aria label

src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
benvillalobos: the link is already tab-navigable, so I've removed the tabIndex on the terminal tool part

meganrogge (dismissed):
Thanks!

Tyriar (dismissed):

Tyriar (approved):

PR #274254, opened 2025-10-31
fix one off error for chat edits and checkpoints

### LINKED ISSUE #274084: Restoring checkpoint does not delete last message, shows broken button
Repro:

1. Fairly long chat session
2. Click Restore Checkpoint, 🐛 below it isn't deleted, checkpoint restored shows underneath, redo button doesn't work

I've had this sort of experience before, makes me not want to engage with checkpoints.

Image

justschen:
fix https://github.com/microsoft/vscode/issues/274084

### DIFF
--- a/src/vs/workbench/contrib/chat/browser/chatEditing/chatEditingCheckpointTimelineImpl.ts
+++ b/src/vs/workbench/contrib/chat/browser/chatEditing/chatEditingCheckpointTimelineImpl.ts
@@ -203,7 +203,7 @@ export class ChatEditingCheckpointTimelineImpl implements IChatEditingCheckpoint
 			throw new Error(`Checkpoint ${checkpointId} not found`);
 		}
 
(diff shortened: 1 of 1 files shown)


src/vs/workbench/contrib/chat/browser/chatEditing/chatEditingCheckpointTimelineImpl.ts
```diff
@@ -203,7 +203,7 @@ export class ChatEditingCheckpointTimelineImpl implements IChatEditingCheckpoint
 			throw new Error(`Checkpoint ${checkpointId} not found`);
 		}
 
-		return this._navigateToEpoch(targetCheckpoint.epoch + 1);
```
connor4312, line 206: This line should have been good as-is. Having `_currentEpoch === thing.epoch` means 'thing' hasn't happened yet.

PR #274276, opened 2025-10-31
Chat todos: revisit icons used for tasks and in progress (fix #273083)

### LINKED ISSUE #273083: Chat todos: revisit icons used for tasks and in progress
I find the codicon we use for a task that has not yet started very large:

Image

Can we find a different smaller one?

I also think for an in-progress task, we should show our usual spinner icon.

cc @mrleemurray

bpasero:


bpasero:
@bhavyaus 

> This removes the icons from the header -> was this intentional?

this was only done so that todos - when collapsed - look similar to the edit working set below since that one also does not have an icon. We could play with having the icon to the right though.

> Also, the spinner continues to animate after the response is completed or stopped, when there are in-progress todos.

That is a good point, but that was already the case before, with the previous in-progress icon, even though not animated. Do we know when the response stopped and could track that as another state in the todo widget?

bpasero:
Also not very pretty:
image


This is not ready, closing for now.

PR #274284, opened 2025-10-31
Make `applyTo` glob pattern in prompt instructions case-insensitive

dmitrivMS:
Pass `{ ignoreCase: true }` to `glob.match()` to make glob matching on the `applyTo` pattern of prompt instructions case-insensitive.

This simplifies experience for the users of this setting and I believe a case-sensitive functionality in this case is not expected/required.

Contributes towards fixing #10633.


aeschli:
@dmitrivMS How do we decide which globs are case sensitive and which ones not. Does it depend on the file system they are matched against?

aeschli (approved):

PR #274291, opened 2025-10-31
Add the missing 'StructuredBuffer' keyword in the HLSL syntax highlighting

### LINKED ISSUE #274290: HLSL's "StructuredBuffer" keyword is not getting syntax highlighting

- VS Code Version: 1.105.1
- OS Version: Any

Steps to Reproduce:

1. Create a file
2. Select HLSL as language mode
3. Write `StructuredBuffer` in the editor

`StructuredBuffer` won't be highlighted unlike `RWStructuredBuffer` for example. It seems `StructuredBuffer` is missing from `hlsl.tmLanguage.json`

godlikepanos:
`StructuredBuffer` is missing from `hlsl.tmLanguage.json` and this PR just adds that

Fixes #274290

### DIFF
--- a/extensions/hlsl/syntaxes/hlsl.tmLanguage.json
+++ b/extensions/hlsl/syntaxes/hlsl.tmLanguage.json
@@ -104,7 +104,7 @@
 		},
 		{
 			"name": "support.type.object.hlsl",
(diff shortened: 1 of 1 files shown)


alexr00 (commented):
Thank you for the PR, but we pull this grammar from https://github.com/tgjones/shaders-tmLanguage/blob/master/grammars/hlsl.json.

PR #274363, opened 2025-10-31
Fix hover display for nested targets

### LINKED ISSUE #273888: [a11y] Tooltip of context button is overlapped by tree item

- VS Code Version: Version: 1.105.1 (user setup)
Electron: 37.6.0
ElectronBuildId: 12502201
Chromium: 138.0.7204.251
Node.js: 22.19.0
V8: 13.8.258.32-electron.0
OS: Windows_NT x64 10.0.26200
- OS Version: 

Steps to Reproduce:

1. Tab navigate through the context buttons on tree item
2. When context button is focused, first the tooltip of the context button is displayed, then it disappear and the tooltip of the tree item is displayed
3. This can be reproduced with Agents Toolkit extension or the Source control panel in VSC

https://github.com/user-attachments/assets/bc3147c6-d3d7-455d-bd47-c682dd8e6eb5

benibenj:
```Copilot Generated Description:``` Update hover service logic to correctly handle mouse and focus events for nested hover targets, ensuring that hover does not display when interacting with other hover targets.

fixes https://github.com/microsoft/vscode/issues/273888

### DIFF
--- a/src/vs/platform/hover/browser/hoverService.ts
+++ b/src/vs/platform/hover/browser/hoverService.ts
@@ -602,7 +602,7 @@ class HoverContextViewDelegate implements IDelegate {
 }
 
 function eventIsRelatedToTarget(event: UIEvent, target: HTMLElement): boolean {
-	return isHTMLElement(event.target) && getHoverTargetElement(event.target as HTMLElement, target) !== target;
+	return isHTMLElement(event.target) && getHoverTargetElement(event.target as HTMLElement, target) === target;
 }
 
 function getHoverTargetElement(element: HTMLElement, stopElement?: HTMLElement): HTMLElement {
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/vs/platform/hover/browser/hoverService.ts [hunk 1]
```diff
@@ -597,6 +601,10 @@ class HoverContextViewDelegate implements IDelegate {
 	}
 }
 
+function eventIsRelatedToTarget(event: UIEvent, target: HTMLElement): boolean {
+	return isHTMLElement(event.target) && getHoverTargetElement(event.target as HTMLElement, target) !== target;
```
benibenj: this should be `===`

Tyriar (approved):
Do you know if this a regression or was it always like this?

PR #274376, opened 2025-10-31
extension unification ignore settings validation

benibenj:
This is the only way I could come up with which would support both extensions contributing the same configuration and being able to write to them when either one or both are enabled.

https://github.com/microsoft/vscode-internalbacklog/issues/6073


DonJayamanne (approved):

src/vs/workbench/services/configuration/common/configurationEditing.ts
```diff
@@ -522,7 +522,7 @@ export class ConfigurationEditing {
 		 * 	- the key is an override identifier
 		 * 	- the operation is to delete the key
 		 */
-		if (!operation.workspaceStandAloneConfigurationKey) {
+		if (!operation.workspaceStandAloneConfigurationKey && !IGNORE_VALIDATION_SETTINGS.has(operation.key)) {
```
benibenj, line 525: Remove this one in https://github.com/microsoft/vscode/pull/274797

PR #274387, opened 2025-10-31
Update node-pty version for conpty 1.23.251008001

anthonykim1:
This is so we get https://github.com/microsoft/node-pty/pull/811 
Also updating conpty dll version number defined in setting description.

### DIFF
--- a/package.json
+++ b/package.json
@@ -107,7 +107,7 @@
     "native-is-elevated": "0.7.0",
     "native-keymap": "^3.3.5",
     "native-watchdog": "^1.4.1",
-    "node-pty": "1.1.0-beta35",
+    "node-pty": "1.1.0-beta38",
     "open": "^10.1.2",
     "tas-client-umd": "0.2.0",
     "undici": "^7.9.0",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -108,7 +107,7 @@
     "native-is-elevated": "0.7.0",
     "native-keymap": "^3.3.5",
     "native-watchdog": "^1.4.1",
-    "node-pty": "1.1.0-beta35",
+    "node-pty": "1.1.0-beta38",
```
Tyriar, line 110: Need to also update remote/package.json for node-pty as it's shipped with the server.

Tyriar (changes requested):

package.json
anthonykim1: Silly question.. I just realized for remote/ package-lock.json, I had to manually edit the version number.

Usually `npm install` would update package-lock.json . But for remote is it manual like this?

Tyriar: I always run `npm install node-pty@...` in both repos to update. Note that we need the update to go in also for https://github.com/microsoft/node-pty/pull/812 now

PR #274399, opened 2025-10-31
only show chat terminal hint if there are hidden terminals

### LINKED ISSUE #274313: Chat terminal UI should not show up when all chat terminals are visible
Following up on https://github.com/microsoft/vscode/pull/274149

This is meant to be a hint for hidden chat terminals:

Image

If there's only a single one and it's visible the indicator shouldn't show. Might also be best to make it explicit in the text `1 Hidden Chat Terminal`?

meganrogge:
fix https://github.com/microsoft/vscode/issues/274313

cc @tyriar

Also removes a bad `!`

### DIFF
--- a/src/vs/workbench/contrib/terminal/browser/terminalTabbedView.ts
+++ b/src/vs/workbench/contrib/terminal/browser/terminalTabbedView.ts
@@ -140,6 +140,12 @@ export class TerminalTabbedView extends Disposable {
 			this._updateChatTerminalsEntry();
 		}));
 
+		this._register(contextKeyService.onDidChangeContext(e => {
+			if (e.affectsSome(new Set(['hasHiddenChatTerminals']))) {
+				this._refreshShowTabs();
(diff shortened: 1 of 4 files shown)


src/vs/workbench/contrib/terminal/browser/terminalTabbedView.ts
```diff
@@ -140,6 +140,12 @@ export class TerminalTabbedView extends Disposable {
 			this._updateChatTerminalsEntry();
 		}));
 
+		this._register(contextKeyService.onDidChangeContext(e => {
+			if (e.affectsSome(new Set(['hasHiddenChatTerminals']))) {
```
dmitrivMS, line 144: Should this be a constant somewhere?

benvillalobos (dismissed):
unblock approval (not area owner, but confirmed it unblocks my scenario)

dmitrivMS (approved):

PR #274414, opened 2025-10-31
Fix watch view context menu when Variables pane is hidden

### LINKED ISSUE #273721: Context menu in watch during debugging not working

- VS Code Version: 1.105.1
- OS Version: Kubuntu 24.04

Hello,
since a few weeks, I’ve been unable to open the context menu in the Watch window during debugging (for example, to copy values).

Debugger: gdb
Language: C++

bryanchen-d:
## Problem
The context menu in the Watch view fails to appear during debugging when the Variables pane is hidden/collapsed. This prevents users from accessing debugging actions like "Break on Value Change", "Copy Value", etc.

Fixes: #273721

## Root Cause
When the Variables pane is hidden, we never fetches scope variables from the debug adapter. When the Watch view tries to get data breakpoint info for expressions, the debug adapter doesn't have the necessary variable context and throws "Unable to find child property" errors.

## Solution
Before requesting data breakpoint info, ensure the debug adapter has the necessary context:

For nested variable `s` (like x within expanded s):

* Fetch the parent's children to populate adapter state
* Pass the parent's `variablesReference`

For simple expressions (like `s` or `a`):
* Get the local scope from the focused stack frame
* Fetch the scope's variables to populate adapter state
* Pass the scope's `variablesReference`

For complex expressions (like `s.x`):

* Pass the expression with `frameId` and let the adapter parse it
* Fall back gracefully if not supported

Error handling:

* Wrap `dataBreakpointInfo()` calls in try-catch
* On failure, silently continue without data breakpoint menu items

### DIFF
--- a/src/vs/workbench/contrib/debug/browser/watchExpressionsView.ts
+++ b/src/vs/workbench/contrib/debug/browser/watchExpressionsView.ts
@@ -422,24 +422,35 @@ async function getContextForWatchExpressionMenuWithDataAccess(parentContext: ICo
 
 	try {
 		// Per DAP spec:
-		// - For Variables (children in a tree): use variable.name + parent's variablesReference
-		// - For Expressions (top-level items): use expression string (evaluateName preferred) + frameId
-		if (expression instanceof Variable) {
+		// - If evaluateName is available: use it as an expression (top-level evaluation)
+		// - Otherwise, check if it's a Variable: use name + parent reference (container-relative)
+		// - Otherwise: use name as an expression
+		if ('evaluateName' in expression && expression.evaluateName) {
+			// Use evaluateName if available (more precise for evaluation context)
+			dataBreakpointInfoResponse = await session.dataBreakpointInfo(
+				expression.evaluateName as string,
+				undefined,
+				stackFrame?.frameId
+			);
+		} else if (expression instanceof Variable) {
+			// Variable without evaluateName: use name relative to parent container
 			dataBreakpointInfoResponse = await session.dataBreakpointInfo(
 				expression.name,
 				expression.parent.reference,
 				stackFrame?.frameId
 			);
 		} else {
+			// Expression without evaluateName: use name as the expression to evaluate
 			dataBreakpointInfoResponse = await session.dataBreakpointInfo(
-				'evaluateName' in expression ? expression.evaluateName as string : expression.name,
-				undefined,  // No parent reference for top-level expressions
+				expression.name,
+				undefined,
 				stackFrame?.frameId
 			);
 		}
 	} catch (error) {
+		// If dataBreakpointInfo fails (e.g., expression not found, invalid expression),
 		// silently continue without data breakpoint support for this item
-		logService.error('Failed to get data breakpoint info for watch expression:', error);
+		logService.trace('Failed to get data breakpoint info for watch expression:', error);
 	}
 
 	const dataBreakpointId = dataBreakpointInfoResponse?.dataId;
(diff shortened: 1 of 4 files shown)


src/vs/workbench/contrib/debug/browser/watchExpressionsView.ts
```diff
@@ -409,14 +408,65 @@ function getContextForWatchExpressionMenu(parentContext: IContextKeyService, exp
...
+						scopeReference = localScope.reference || 0;
+					}
+				}
+
+				dataBreakpointInfoResponse = await session.dataBreakpointInfo(
+					expressionName,
+					scopeReference
+				);
+			}
+		}
```
connor4312, line 449: I think passing the frameId is good but I'm confused why we need to look at scopes or check if the expression is nested. evaluateName is always expected to be based on the stackframe, so a scope reference is not needed. If there isn't an evaluate name, we should evaluate `expression.name` in the parent context which is `expression.parent`.

connor4312 (changes requested):
comment

src/vs/workbench/contrib/debug/browser/watchExpressionsView.ts
bryanchen-d: When the Variables window is off, the scopes are not populated so here we have to populate the scope and pass it in. 
From my test, if we don't pass a scope reference here, the `dataBreakpointInfo()` throws "Unable to find child property".

Regarding the expression.parent, from my debug, it is undefined for a top-level expression like `s.x` (accessing a member variable in a struct object) or `a` (a simple variable)

connor4312: I'm still not sure why we need scopes. There's nothing in DAP that ties scopes to the `dataBreakpointInfo`. It needs only a name, and optionally variablesReference/frameId. We didn't pass the frameId before, and you fixed that here, which I think is good.

> Regarding the expression.parent, from my debug, it is undefined for a top-level expression like s.x (accessing a member variable in a struct object) or a (a simple variable)

Yea, that is intended. It's only defined for children of DAP Variables.

src/vs/workbench/contrib/debug/browser/watchExpressionsView.ts
```diff
@@ -408,63 +410,22 @@ function getContextForWatchExpressionMenu(parentContext: IContextKeyService, exp
...
-				}
-
-				dataBreakpointInfoResponse = await session.dataBreakpointInfo(
-					expressionName,
-					scopeReference
-				);
-			}
-		}
+		const expressionName = 'evaluateName' in expression ? expression.evaluateName as string : expression.name;
+		dataBreakpointInfoResponse = await session.dataBreakpointInfo(expressionName, undefined, stackFrame?.frameId);
```
connor4312: you had another change we should keep here:

- If there's an `expression.evaluateName`, then calling it as `evaluateName, undefined, frameId` is good
- If there's no `expression.evaluateName`, we should call it as `name, expression.parent.reference, frameId` when available

evaluateName is absolute, but `variable.name` is relative to its parent, so the parent reference should be preserved when possible

bryanchen-d: From the strong typing and also from my debugging, `expression.parent` is only avaiable when `expression instanceof Variable` is true. Can we assume when there is no `evaluateName` , the expression is a Variable?

bryanchen-d: From what I see in debugging, when there is a top level expression, for example, if user type `s` or `s.x` in the expression,  there is no evaluateName, nor expression.parent.

connor4312: For the top level expression, there will not be an expression.parent. For properties of the object/expression result, there will be a parent.

> expression.parent is only avaiable when expression instanceof Variable

Yes, if this is the case we should pass in the parent refrence when there is no `evaluateName`

connor4312 (approved):

PR #274417, opened 2025-10-31
persist command in pty service, refactor how commands are restored on window reload for inline chat terminal

### LINKED ISSUE #274179: Link run in terminal session with command ID
We have:

https://github.com/microsoft/vscode/blob/21fa01dcece707d8a252b06431970407951a26ec/src/vs/workbench/contrib/terminalContrib/chat/browser/terminalChatService.ts#L91-L100

But it fails right now after reload because the `command.id` differs since it's created once for the renderer and once for the ptyHost here:

https://github.com/microsoft/vscode/blob/21fa01dcece707d8a252b06431970407951a26ec/src/vs/platform/terminal/common/capabilities/commandDetection/terminalCommand.ts#L289

Haven't thought it through fully, but here's another idea for associating the tool call with the command which may end up being simpler:

- Only support linking commands for rich command detection
- In prepare, create the command ID manually via `tool-${generateUuid()}` and set on `toolSpecificData`, this would make serialize of the part just work at long as we restore the custom command ID
- In invoke, pass that command ID into `runCommand`
- Tell the pty host and renderer shell integration addon that the next command with the matching command line is the ID we created

_Originally posted by @Tyriar in https://github.com/microsoft/vscode/pull/273175#discussion_r2478161085_

### LINKED ISSUE #274311: Focus terminal button doesn't work after reloading window
Repro:

1. `run ls` in chat
2. Expand and click focus terminal, the command is revealed
3. Reload the window
4. Click focus terminal, 🐛 nothing happens despite the terminal being visible

Image

meganrogge:
fixes #274311
fixes https://github.com/microsoft/vscode/issues/274179

This PR implements a mechanism to pre-assign command IDs for terminal commands executed by chat agent tools, enabling better tracking and linking of commands between the renderer and ptyHost. The system generates a custom command ID early in the process flow and propagates it through the execution pipeline.

Key changes:
- Introduced `terminalCommandId` field to `IChatTerminalToolInvocationData` for pre-assigned command IDs
- Added `setNextCommandId` method throughout the terminal process chain (from `ITerminalInstance` to `ITerminalChildProcess`)
- Updated execution strategies to pass the command ID when executing terminal commands
- Modified command detection to use pre-assigned IDs when available
- Refactored output collection to use the pre-assigned command ID for lookup
- Removes stored command ids from chat service


rebornix (approved):

src/vs/platform/terminal/common/capabilities/commandDetectionCapability.ts
```diff
@@ -355,7 +364,20 @@ export class CommandDetectionCapability extends Disposable implements ICommandDe
...
-
+		// If a custom command ID is provided, use it for the current command
+		// Otherwise, check if there's a pending next command ID
+		if (options?.commandId) {
+			this._currentCommand.id = options.commandId;
+			this._nextCommandId = undefined; // Clear the pending ID
+		} else if (
+			this._nextCommandId &&
+			typeof this.currentCommand.command === 'string' &&
+			typeof this._nextCommandId.command === 'string' &&
```
Tyriar, line 375: Prefer `isString`

src/vs/workbench/contrib/chat/common/chatService.ts
```diff
@@ -313,12 +313,9 @@ export interface IChatTerminalToolInvocationData {
...
 	language: string;
 	terminalToolSessionId?: string;
-	terminalCommandUri?: UriComponents;
+	/** The predefined command ID that will be used for this terminal command */
+	terminalCommandId?: string;
 	autoApproveInfo?: IMarkdownString;
-	output?: {
-		html: string;
-		truncated?: boolean;
-	};
```
Tyriar, line 321: Does deserializations still restore the output after reload if the terminal's gone? We want it so this works:

1. Run ls
2. Expand in chat, works
3. Collapse in chat
4. Kill the terminal
5. Expand in chat, works
6. Collapse in chat
7. Reload window
8. Expand in chat, works (this is what I'm worried about)

src/vs/workbench/contrib/terminal/browser/terminalProcessExtHostProxy.ts
```diff
@@ -140,6 +140,10 @@ export class TerminalProcessExtHostProxy extends Disposable implements ITerminal
 		// No-op
 	}
 
+	async setNextCommandId(commandLine: string, commandId: string): Promise<void> {
+		// No-op
+	}
```
Tyriar, line 145: This isn't great that there are a bunch of no-op ones. That's due to being on `ITerminalChildProcess`?

Could we put it on `IPtyService` instead?

src/vs/workbench/contrib/terminal/browser/terminalProcessManager.ts
```diff
@@ -591,6 +592,35 @@ export class TerminalProcessManager extends Disposable implements ITerminalProce
...
+		}
+		if (!error.message.includes('Method not found: setNextCommandId') && !error.message.includes('Method not found: $setNextCommandId')) {
+			return false;
+		}
+		if (!this._hasLoggedSetNextCommandIdFallback) {
+			this._hasLoggedSetNextCommandIdFallback = true;
+			this._logService.trace('setNextCommandId not supported by current terminal backend, falling back.');
+		}
+		return true;
+	}
```
Tyriar, line 622: Instead of string matching on error, can we return something more useful from `setNextCommandId`? An error would be an unexpected ideally but it looks like this is expected.

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/runInTerminalTool.ts
```diff
@@ -479,13 +481,13 @@ export class RunInTerminalTool extends Disposable implements IToolImpl {
 			let pollingResult: IPollingResult & { pollDurationMs: number } | undefined;
 			try {
 				this._logService.debug(`RunInTerminalTool: Starting background execution \`${command}\``);
-				const execution = new BackgroundTerminalExecution(toolTerminal.instance, xterm, command, chatSessionId);
+				const commandId = (toolSpecificData as IChatTerminalToolInvocationData).terminalCommandId;
```
Tyriar, line 484: Can this live next to where `terminalToolSessionId` is set so the dangerous looking casts live next to each other?

src/vs/workbench/contrib/chat/common/chatService.ts
meganrogge: No, if the terminal is killed, it won't work. I can add that back in.

Tyriar: Yes please, when output is ready we should serialize it in addition to the command URI so that we can toggle even when the terminal's long gone. Otherwise looking back at history will omit all the output

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/runInTerminalTool.ts
meganrogge: that's set in `prepare`, so they can't be next to eachother. it should be safe though

PR #274542, opened 2025-11-01
edits: fix 'restore checkpoint' leaving the request

connor4312:
- For restoring to a request, we actually want to restore the request's
  initial checkpoint but put our pointer before it, so that further
  requests overwrite that request checkpoint.
- We previously didn't discard undone checkpoints when new ones were applied
- Refined the behavior of disablement to more explicitly provide the
  right UX (show up to and including the last checkpoint that contains
  any applied operation)

### DIFF
--- a/src/vs/workbench/contrib/chat/browser/chatEditing/chatEditingCheckpointTimelineImpl.ts
+++ b/src/vs/workbench/contrib/chat/browser/chatEditing/chatEditingCheckpointTimelineImpl.ts
@@ -123,16 +123,23 @@ export class ChatEditingCheckpointTimelineImpl implements IChatEditingCheckpoint
 		{ equalsFn: (a, b) => arraysEqual(a, b, objectsEqual) },
 		reader => {
 			const currentEpoch = this._currentEpoch.read(reader);
-			const checkpoints = this._checkpoints.read(reader);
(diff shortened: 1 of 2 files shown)


src/vs/workbench/contrib/chat/browser/chatEditing/chatEditingCheckpointTimelineImpl.ts
```diff
@@ -123,16 +123,23 @@ export class ChatEditingCheckpointTimelineImpl implements IChatEditingCheckpoint
...
 		reader => {
 			const currentEpoch = this._currentEpoch.read(reader);
-			const checkpoints = this._checkpoints.read(reader);
+			const operations = this._operations.read(reader);
+			const firstUnappliedOperationIdx = operations.findIndex(op => op.epoch >= currentEpoch);
+			if (firstUnappliedOperationIdx === -1) {
+				return [];
+			}
 
+			const lastAppliedOperation = firstUnappliedOperationIdx > 0 ? operations[firstUnappliedOperationIdx - 1].epoch : 0;
```
Tyriar, line 132: nit: `find` is a bit nicer unless you'd using the index?

Tyriar (approved):

Issue #233555, opened 2024-11-11
Increase hit target of +/- type zoom buttons


Tyriar:
The hit target is very small:

![Image](https://github.com/user-attachments/assets/c1cc40c2-70a4-489b-94e3-1b7a7e5a8690)

This is actually much smaller than the breakpoint hit target:

![Image](https://github.com/user-attachments/assets/4a7177e2-b3a7-4f1f-a341-1fb091812635)

PR #274632, opened 2025-11-02
Fix: hit target, padding and appearance #233555

juliabush:
Closes issue #233555 

Description of changes:
- Added padding-bottom: 2px to .monaco-resizable-hover to increase space at the bottom of the hover container, because minus sign was overflowing. This is why overflow-y was also removed.
- Adjusted .codicon with padding: 2px, margin: -2px, and margin-bottom: -4px to enlarge the clickable area without affecting visual size. 
- Increased gap to 4px in .verbosity-actions-inner because the click area of the plus and minus signs began to overlap. 

Hopefully these changes are helpful or on the mark. Can change anything or make improvements if necessary.

### DIFF
--- a/src/vs/editor/contrib/hover/browser/hover.css
+++ b/src/vs/editor/contrib/hover/browser/hover.css
@@ -11,6 +11,7 @@
 	border: 1px solid var(--vscode-editorHoverWidget-border);
 	border-radius: 3px;
 	box-sizing: content-box;
+	padding-bottom: 2px;
 }
 
 .monaco-editor .monaco-resizable-hover > .monaco-hover {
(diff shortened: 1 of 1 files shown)


aiday-mar:
Hi thanks for the PR. The padding bottom adds a padding bottom which is visible when the inner hover is focused. It should not be added. I believe we need to clip the overflow-y, so that should be removed too. While testing I noticed that the codicons are not centered inside of their containers. Could you please do screen-recordings for how the codicons looks like before the fix and after the fix, so we see the fix visually, and show the enlarged hit target area? Thank you.

https://github.com/user-attachments/assets/0d03d00a-016b-4602-b318-ce81925875f0

aiday-mar (commented):
Hi thanks for the PR. The padding bottom adds a padding bottom which is visible when the inner hover is focused. It should not be added. I believe we need to clip the overflow-y, so that should be removed too. While testing I noticed that the codicons are not centered inside of their containers. Could you please do screen-recordings for how the codicons looks like before the fix and after the fix, so we see the fix visually, and show the enlarged hit target area? Thank you.

https://github.com/user-attachments/assets/0d03d00a-016b-4602-b318-ce81925875f0

PR #274645, opened 2025-11-02
JSDoc

m-housni:
add JSDoc to equals()


mjbvz:
Thanks but we generally only add docs when they add clear value over the code itself

PR #274650, opened 2025-11-02
Enhance removeFastWithoutKeepingOrder with error checks

m-housni:
Add error handling for out-of-bounds index and empty array in removeFastWithoutKeepingOrder function.


jrieken:
@m-housni Thanks for this contribution. Tho, I do agree with @ulugbekna - we typically don't throw from utils but keep on trucking (the JS way). This also doesn't solve an issue because the only user of this function is correct

PR #274656, opened 2025-11-03
thinking settings cleanup and style fix

justschen:
we now only have 3 thinking styles exposed:
1. collapsed
2. collapsed preview
3. fixedScrolling (default, fixed height)

a new setting will be checked which will choose if we add tools into the container or not.
1. none
2. readonly (default)
3. all tools

https://github.com/user-attachments/assets/ec33b6fe-2e52-4eac-a4ac-0e4e270f26d3


Tyriar (dismissed):

roblourens (approved):
Image

you love to see it!

Issue #162995, opened 2022-10-07
Status bar content appears misaligned


daviddossett:
<img width="1023" alt="image" src="https://user-images.githubusercontent.com/25163139/194638274-2e0a1650-37db-4a9a-9f39-fc9c5281922f.png">

bpasero:
@daviddossett seems OK to me at least in my instance (Parallels):

<img width="1028" alt="image" src="https://user-images.githubusercontent.com/900690/206096656-1056ed52-7bc7-4fd1-b9af-1d77033e32d0.png">

Any zooming or display scaling involved?

VSCodeTriageBot:
This issue has been closed automatically because it needs more information and has not had recent activity. See also our [issue reporting](https://aka.ms/vscodeissuereporting) guidelines.

Happy Coding!

bpasero:
Though there is something to it where I often feel icons and text are not perfectly aligned, on macOS:

![image](https://user-images.githubusercontent.com/900690/208091158-33a9a304-cf60-434d-8ca3-bcd8a9f3b0aa.png)

Maybe a CSS champion wants to look into this, the CSS is at `src/vs/workbench/browser/parts/statusbar/media/statusbarpart.css`

bpasero:
To clarify, I am on macOS where I often feel there is misalignment, though I also always have a zoom level of 1.

bpasero:
I see no evidence of a `1px` border, unless maybe you configured the `window.activeBorder`:

<img width="346" alt="image" src="https://user-images.githubusercontent.com/900690/208240441-6baa596c-1f1a-4471-90a6-6685c8da4fbd.png">

daviddossett:
I think it's just Windows adding a native border then? I'll look again next time I'm on my windows machine.

VSCodeTriageBot:
We closed this issue because we [don't plan to address it](https://aka.ms/vscode-out-of-scope) in the foreseeable future. If you disagree and feel that this issue is crucial: we are happy to listen and to reconsider.

If you wonder what we are up to, please see our [roadmap](https://aka.ms/vscoderoadmap) and [issue reporting guidelines](https://aka.ms/vscodeissuereporting).

Thanks for your understanding, and happy coding!

bpasero:
Reopening for @aeschli https://github.com/microsoft/vscode/issues/174434#issue-1585347915. Opening for help if someone is able to figure this out.

bpasero:
Cool findings 👍

[...]

PR #274733, opened 2025-11-03
#162995

24f2008896:


24f2008896:

bpasero (changes requested):
thanks, but why is this removing existing comments? I would prefer to review only your actual changes

PR #274754, opened 2025-11-03
enable inline chat v2 for team selfhost

jrieken:


Tyriar (approved):

.vscode/settings.json
```diff
@@ -176,7 +167,6 @@
 			"${workspaceFolder}/extensions/*/out/**/*.js",
 		]
 	},
-
 	// --- Coverage ---
```
Tyriar, line 170: You may not have meant to delete these empty lines?

PR #274776, opened 2025-11-03
Git - improve compare reference notification

lszomoru:

### DIFF
--- a/extensions/git/src/commands.ts
+++ b/extensions/git/src/commands.ts
@@ -3116,13 +3116,12 @@ export class CommandCenter {
 			return;
 		}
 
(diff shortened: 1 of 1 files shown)


extensions/git/src/commands.ts
```diff
@@ -3195,11 +3192,11 @@ export class CommandCenter {
 
 			await commands.executeCommand('_workbench.openMultiDiffEditor', {
 				multiDiffSourceUri,
-				title,
+				title: `${ref1DisplayId} ↔ ${ref2DisplayId}`,
```
Tyriar, line 3195: Best to use \u for unicode in general:

```suggestion
				title: `${ref1DisplayId} \u2194 ${ref2DisplayId}`,
```

Some chars can impact file encoding which caused perf problems in the past.

Tyriar (approved):

Issue #274723, opened 2025-11-03
Reduce use of explicit `any` type


bpasero:
Related issue: https://github.com/microsoft/vscode/issues/269213

We want to enable `@typescript-eslint/no-explicit-any` for all our TypeScript code. Today we exclude a lot of files that violate this rule:

https://github.com/microsoft/vscode/blob/c24799b06932465a06b21031bd7f5512f32ca7a4/eslint.config.js#L427-L1078

If you are assigned to this issue then please review the `ignores` list of files you own and try to remove the use of explicit `any`. After that, remove the file from the `ignores` list.

- [ ] @jrieken 
- [ ] @benibenj 
- [ ] @bpasero 
- [x] @Tyriar 
- [ ] @meganrogge 
- [ ] @justschen 
- [x] @lszomoru 
- [ ] @mjbvz 
- [ ] @roblourens 
- [ ] @sandy081 
- [ ] @DonJayamanne

PR #274804, opened 2025-11-03
Make `executeCommand` return `unknown` instead of `any` by default

mjbvz:
For #274723


src/vs/workbench/contrib/welcomeGettingStarted/browser/gettingStarted.ts
```diff
@@ -1337,7 +1326,7 @@ export class GettingStartedPage extends EditorPane {
 				}
 			}
 
-			this.commandService.executeCommand(commandURI.path, ...args).then(result => {
+			this.commandService.executeCommand<any>(commandURI.path, ...args).then(result => {
```
mjbvz, line 1329: Needs proper fix by owner as part of `any` pass

Tyriar (approved):

Issue #274198, opened 2025-10-30
Agent sessions close and reopen editor when they complete


connor4312:
1. Have multiple editors open for an claude code/copilot cli agent session
2. Have the agent make changes (which are now tracked in our edit sessions)
3. After the response complete you get this dialog 

<img width="616" height="155" alt="Image" src="https://github.com/user-attachments/assets/398c7ae7-3fa6-4bf3-9ee2-9e052af42544" />

This is because we try to close and recreate the session here

https://github.com/microsoft/vscode/blob/1393bde719465678fdebe71a7721bfa29ee85d05/src/vs/workbench/api/browser/mainThreadChatSessions.ts#L408

@joshspicer looks like you added this, not sure of the context, let me know if you need help

connor4312:
@DonJayamanne for context too

connor4312:
(1) now already happens, but the problem is the chat history from core which includes the edit annotations is not transferred to the new session, so the edit pills disappear from the first turn.

(2) is probably suffering from the same thing. I'm guessing the edit session is actually saved and restored but you don't restore the edit history parts so there are no pills.

joshspicer:
@connor4312 is this still relevant with all your recent changes?

DonJayamanne:
I think we can close this, 
> ed wanting to do some further cleanups with this in general

I was referring to getting rid of this whole `swapping` as there are too many hacks/work arounds in Core and Chat extension for background agents (i've been submitting PRs regularly to get around this)
Lets close this for now.

PR #274840, opened 2025-11-03
edits: transfer edit session to committed agent chat session

connor4312:
Working set and edit history transfers, however chat pills do not yet
show up since they don't survive the chat history transfer.

Refs #274198 cc @joshspicer


src/vs/workbench/contrib/chat/common/chatSessionsService.ts
```diff
@@ -111,6 +112,11 @@ export interface IChatSession extends IDisposable {
 	readonly isCompleteObs?: IObservable<boolean>;
 	readonly interruptActiveResponseCallback?: () => Promise<boolean>;
 
+	/**
+	 * Editing session transferred from a previously-untitled chat session in `onDidCommitChatSessionItem`.
+	 */
+	initialEditingSession?: IChatEditingSession;
```
connor4312, line 118: This is not the nicest thing, but it does the job, let me know if you have a preference for something better

mjbvz (approved):

PR #274862, opened 2025-11-03
Fix flicker in focus terminal

### LINKED ISSUE #274314: Chat terminal button will flicker after selecting one to focus
Following up on https://github.com/microsoft/vscode/pull/274149

Repro:

1. Open 2 terminals
2. `run ls`
3. Click 1 chat terminal in terminal tabs
    Image
4. Select the single entry, 🐛 there's a focus border flicker for ~1 frame around the 1 chat terminal button

Quick pick has some feature where it will focus what was focused before, we may be able to suppress that feature since we want to control focus unless the quick pick is hidden without action?

Tyriar:
Fixes #274314


meganrogge (approved):
Thank you 😄 !

PR #274897, opened 2025-11-03
edits: fix bad undo/redo around messages with no edits

connor4312:
Handles cases around edits with no contained undo stops

### DIFF
--- a/src/vs/base/common/arraysFind.ts
+++ b/src/vs/base/common/arraysFind.ts
@@ -5,9 +5,9 @@
 
 import { Comparator } from './arrays.js';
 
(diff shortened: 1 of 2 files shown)


src/vs/base/common/arraysFind.ts
```diff
@@ -5,21 +5,43 @@
...
-		if (predicate(element)) {
+		if (predicate(element, i)) {
+			return i;
+		}
+	}
+
+	return -1;
+}
+
+export function findFirst<T, R extends T>(array: readonly T[], predicate: (item: T, index: number) => item is R, fromIndex?: number): R | undefined;
```
mjbvz, line 30: Can we use the built-in `[].find` and `[].findIndex` for this?

mjbvz (approved):

PR #275085, opened 2025-11-04
Improve file permissions detection

### LINKED ISSUE #275084: Detect and indicate readonly files and file permissions

- VS Code Version: 1.105.1
- OS Version: Ubuntu 24.04

Steps to Reproduce:

1. Create a file in a project
2. Apply `chattr +i` on that file, or make it readonly by any other mean
3. Open the containing folder in vscode

- The file is not marked as readonly in the explorer (while the feature exists in VSCode)
- A weird error is reported if we try to delete the file (`/tmp/demo/test`):
Image

CGNonofr:
fix #275084

explorer before:
image
after:
image

deleting the file before:
image
after:
image
(sorry for the french)

### DIFF
--- a/src/vs/platform/files/node/diskFileSystemProvider.ts
+++ b/src/vs/platform/files/node/diskFileSystemProvider.ts
@@ -3,7 +3,8 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-import { Stats, promises } from 'fs';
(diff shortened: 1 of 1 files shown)


CGNonofr:
> The way it works today is intentional, we allow to edit protected files with the option to remove the protection.
> 
> Closing.

What's wrong is that is the system if not able to remove the protection, the displayed error is weird and confusing

You can't say that displaying:
image
when trying to delete a file without permissions is `intentional`

PR #275465, opened 2025-11-05
Revert "Git - adopt new QuickInputButton location (#275071)"

dmitrivMS:
This reverts commit e092054c98f425d9a8e55e22dd0f49eb206c7de3.

Unfortunetely new API only supports toggles, not regular buttons so using it as a button will not provide the right appearance. There is additional work to enable regular buttons with `Inline` location.

This is how it would look like right now (that pressed state is sticky, since it's a toggle):
image


lszomoru (approved):

lszomoru:
@dmitrivMS, sounds good. Looking forward to have this as a button.

PR #274994, opened 2025-11-04
Optimize animation frame queue

erkinalp:
Solves #272155 
To test it, use https://github.com/erkinalp/vscode-animation-frame-harness (there's two variants of the testing harness, one for deterministically timing fresh installations, one for profiling existing installations)

Issue #272155, opened 2025-10-19
Improve main rendering loop


erkinalp:
# Main Bottleneck in the Rendering Loop

The primary bottleneck is at /vscode/src/vs/base/browser/dom.ts:365 - the animation frame queue sorts on every iteration inside the while loop:
```typescript

while (currentQueue.length > 0) {
    currentQueue.sort(AnimationFrameQueueItem.sort);  // ⚠️ BOTTLENECK: O(n² log n)
    const top = currentQueue.shift()!;
    top.execute();
}
```

## Why This is the Bottleneck:

- Repeated sorting: For n callbacks, this sorts n times with decreasing sizes (n, n-1, n-2, ..., 1)
- Complexity: O(n² log n) instead of O(n log n) if sorted once
- Real-world impact: With 50+ view parts (text, cursors, minimap, scrollbar, widgets, decorations, etc.), this wastes 1-2ms per frame
- Critical path: Happens during the 16ms frame budget (60fps)

# Proposed Solution: Binary Heap Priority Queue

Replace the array-based queue with a binary min-heap that maintains priority order automatically:
Benefits:

- Insertion: O(log n)
- Extraction: O(log n)
- Overall: O(n log n) instead of O(n² log n)
- Performance gain: 85-90% reduction in queue overhead (~1.5ms → ~0.2ms)

### Implementation:

1. Create a BinaryHeap class:
```typescript

class BinaryHeap<T extends { priority: number }> {
    private _items: T[] = [];

    get length(): number {
        return this._items.length;
    }

    push(item: T): void {
        this._items.push(item);
        this._bubbleUp(this._items.length - 1);
    }

    shift(): T | undefined {
        if (this._items.length === 0) return undefined;
        if (this._items.length === 1) return this._items.pop()!;

        const result = this._items[0];
        this._items[0] = this._items.pop()!;
        this._bubbleDown(0);
        return result;
    }

    private _bubbleUp(index: number): void {
        const item = this._items[index];
        const priority = -item.priority; // Negative for max-heap behavior

        while (index > 0) {
            const parentIndex = Math.floor((index - 1) / 2);
            const parent = this._items[parentIndex];
            if (priority >= -parent.priority) break;

            this._items[index] = parent;
            index = parentIndex;
        }
        this._items[index] = item;
    }

    private _bubbleDown(index: number): void {
        const length = this._items.length;
        const item = this._items[index];
        const priority = -item.priority;

        while (true) {
            let minIndex = index;
[...]

PR #275565, opened 2025-11-05
Optimize animation frame queue ­— attempt 2: tri-buffer RAF queue with up to once per frame sorting

erkinalp:
Solves #272155, replaces #274994
To test it, use https://github.com/erkinalp/vscode-animation-frame-harness (there's two variants of the testing harness, one for deterministically timing fresh installations, one for profiling existing installations)


erkinalp:
@jrieken

PR #275617, opened 2025-11-05
fix disposable leak

### LINKED ISSUE #275388: `chatTerminalToolProgressPart.ts` —  Trying to add a disposable to a DisposableStore
Seeing this after running some terminal commands in chat 

```
lifecycle.ts:477 Error: Trying to add a disposable to a DisposableStore that has already been disposed of. The added object will be leaked!
    at r2i.add (lifecycle.ts:477:18)
    at _rt.D (lifecycle.ts:555:22)
    at _rt.P (chatTerminalToolProgressPart.ts:260:41)
    at s (chatTerminalToolProgressPart.ts:180:9)
    at _rt.L (chatTerminalToolProgressPart.ts:183:9)
    at async chatTerminalToolProgressPart.ts:124:4
```

Maybe the instance gets disposed of around these async calls: https://github.com/microsoft/vscode/blob/ebb5f678cb79d86b18de20cb8c4a2442a160a98f/src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts#L183

meganrogge:
fixes #275388

### DIFF
--- a/src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
+++ b/src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
@@ -174,7 +174,7 @@ export class ChatTerminalToolProgressPart extends BaseChatToolInvocationSubPart 
 		}
 
 		const attachInstance = async (instance: ITerminalInstance | undefined) => {
-			if (!instance || this._terminalInstance === instance) {
+			if (!instance || this._terminalInstance === instance || this._store.isDisposed) {
 				return;
 			}
 			this._terminalInstance = instance;


src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
```diff
@@ -63,6 +63,7 @@ export class ChatTerminalToolProgressPart extends BaseChatToolInvocationSubPart
 	private _outputContent: HTMLElement | undefined;
 	private _outputResizeObserver: ResizeObserver | undefined;
 	private _renderedOutputHeight: number | undefined;
+	private _isDisposed = false;
```
Tyriar: `this._store.isDisposed` already tracks

src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
```diff
@@ -165,6 +169,17 @@ export class ChatTerminalToolProgressPart extends BaseChatToolInvocationSubPart
...
+	public override dispose(): void {
+		if (this._isDisposed) {
+			return;
+		}
+
+		this._isDisposed = true;
+		this._outputResizeObserver?.disconnect();
+		this._outputResizeObserver = undefined;
+		super.dispose();
+	}
```
Tyriar: Let's not override `dispose`.  `_outputResizeObserver` is already being disconnected in the `toDisposable` in `_ensureOutputResizeObserver`

Tyriar (dismissed):
This is a lot, do we need everything here? Would rather something like this went in during debt week but I'm not sure we need this check in so many places.

Tyriar (approved):

PR #275619, opened 2025-11-05
don't use relative path as absolute

### LINKED ISSUE #275110: Path to save at is guessed incorrectly when using "Create: New File..." command and picking jupyter notebook
repro (jupyter notebook specific) : 

(have setting `"files.simpleDialog.enable": true`)

1. invoke "Create: new file..." command
2. pick jupyter notebook in the opened quick pick 
3. press `cmd+s` or `ctrl+s` on windows to save the file 
4. the proposed path is `/Untitled-1.ipynb` which's incorrect

Image

Version: 1.106.0-insider (Universal)
Commit: 18d828e5d596e56b516fac7baa53e4b32cf087be
Date: 2025-11-04T05:03:04.572Z
Electron: 37.7.0
ElectronBuildId: 12597478
Chromium: 138.0.7204.251
Node.js: 22.20.0
V8: 13.8.258.32-electron.0
OS: Darwin arm64 24.6.0

amunger:
fix https://github.com/microsoft/vscode/issues/275110

### DIFF
--- a/src/vs/workbench/contrib/notebook/common/notebookEditorInput.ts
+++ b/src/vs/workbench/contrib/notebook/common/notebookEditorInput.ts
@@ -33,6 +33,7 @@ import { ITextResourceConfigurationService } from '../../../../editor/common/ser
 import { ICustomEditorLabelService } from '../../../services/editor/common/customEditorLabelService.js';
 import { IWorkbenchEnvironmentService } from '../../../services/environment/common/environmentService.js';
 import { IPathService } from '../../../services/path/common/pathService.js';
+import { isAbsolute } from '../../../../base/common/path.js';
 
 export interface NotebookEditorInputOptions {
 	startDirty?: boolean;
@@ -255,7 +256,7 @@ export class NotebookEditorInput extends AbstractResourceEditorInput {
 	}
 
 	private async ensureAbsolutePath(resource: URI): Promise<URI> {
-		if (resource.scheme !== Schemas.untitled || resource.path.startsWith('/')) {
+		if (resource.scheme !== Schemas.untitled || isAbsolute(resource.path)) {
 			return resource;
 		}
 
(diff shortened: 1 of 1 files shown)


Tyriar (dismissed):

rzhao271 (dismissed):

src/vs/workbench/contrib/notebook/common/notebookEditorInput.ts
```diff
@@ -249,11 +249,20 @@ export class NotebookEditorInput extends AbstractResourceEditorInput {
...
 
 	private async _suggestName(provider: NotebookProviderInfo) {
-		const resource = this.ensureProviderExtension(provider);
+		const resource = await this.ensureAbsolutePath(this.ensureProviderExtension(provider));
 		const remoteAuthority = this.environmentService.remoteAuthority;
 		return toLocalResource(resource, remoteAuthority, this.pathService.defaultUriScheme);
 	}
 
+	private async ensureAbsolutePath(resource: URI): Promise<URI> {
+		if (resource.scheme !== Schemas.untitled || resource.path.startsWith('/')) {
```
Tyriar: FYI there's an isAbsolute helper in `path` which might be useful here?

src/vs/workbench/contrib/notebook/common/notebookEditorInput.ts
```diff
@@ -255,7 +256,7 @@ export class NotebookEditorInput extends AbstractResourceEditorInput {
 	}
 
 	private async ensureAbsolutePath(resource: URI): Promise<URI> {
-		if (resource.scheme !== Schemas.untitled || resource.path.startsWith('/')) {
+		if (resource.scheme !== Schemas.untitled || isAbsolute(resource.path)) {
```
Tyriar, line 259: This won't actually work for Windows -> WSL now, since isAbsolute would use win32 parsing in that case.

Tyriar (changes requested):

src/vs/workbench/contrib/notebook/common/notebookEditorInput.ts
amunger: so just revert that change to use the helper method? Seems like a pretty big shortfall for that method

amunger: the win32 isAbsolute will check for a path separator at char 0, so I think it will still work

Tyriar (approved):

PR #275688, opened 2025-11-05
Avoid saving inline chats to history

### LINKED ISSUE #252849: Everytime when inline chat is opened it adds entry to Chat history which cannot be removed (if `inlineChat.enableV2` is enabled)
Does this issue occur when all extensions are disabled?: It's probably related to Copilot extension

- VS Code Version:
```
Version: 1.102.0-insider (user setup)
Commit: b68fa590ef54bc8f07dd79cd3175bf78e0113d4c
Date: 2025-06-27T05:03:40.827Z
Electron: 35.5.1
ElectronBuildId: 11727614
Chromium: 134.0.6998.205
Node.js: 22.15.1
V8: 13.4.114.21-electron.0
OS: Windows_NT x64 10.0.26100
```
- OS Version: Windows 11

Steps to Reproduce:

1. Enable `inlineChat.enableV2` option
2. Open any file
3. Open inline chat <kbd>Ctrl</kbd>+<kbd>I</kbd> few times
4. In chat panel click on the `Show Chats...` button
5. There are many entries which cannot be removed (these entries disappear after reloading window)

Also if you click on one of these entries and click on the `Show Chats...` button, then error occurs:
```
Cannot access a disposed editing session
```

https://github.com/user-attachments/assets/ed88904d-ba7d-4331-8de9-19e91a2b4ed3

osortega:
Fixes: https://github.com/microsoft/vscode/issues/252849

@jrieken could you please review this? Looks like saving inline chats was purposely added at some point, so I want to make sure it doesn't cause any regressions

### DIFF
--- a/src/vs/workbench/contrib/chat/common/chatServiceImpl.ts
+++ b/src/vs/workbench/contrib/chat/common/chatServiceImpl.ts
@@ -239,7 +239,7 @@ export class ChatService extends Disposable implements IChatService {
 				if (!LocalChatSessionUri.parseLocalSessionId(session.sessionResource)) {
 					return false;
 				}
-				return session.initialLocation === ChatAgentLocation.Chat || session.initialLocation === ChatAgentLocation.EditorInline;
+				return session.initialLocation === ChatAgentLocation.Chat;
 			});
 
 		this._chatSessionStore.storeSessions(liveChats);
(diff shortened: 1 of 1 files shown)


jrieken:
> Looks like saving inline chats was purposely added at some point, so I want to make sure it doesn't cause any regressions

Looks like this happened through: https://github.com/microsoft/vscode/commit/fc55871185baaba8adeb8c69b394c062d65c40b8#diff-3f1e8dab034f6103af41784444685e373d9059b1579c8b4cab2fca06b4f96ea4. I don't know why that is and maybe it's enough to undo that change. I would not add even more "if location is XYZ" checks because it clutters the code.

dmitrivMS (approved):

PR #275782, opened 2025-11-06
Throw when executeCommand called on task

### LINKED ISSUE #275667: Attempts to activate Python environment in task terminal
- Set up the task described in https://github.com/microsoft/vscode/issues/268595#issuecomment-3353831971, in a workspace that has a .venv
- Just run the task. The task runs, then we try to activate the virtual environment in the task terminal. But the task script is already running at that point, so the command is handled as input, and it continues. That's not good.
- I'm trying to remember if I messed something up when we were working on env activation for Codex, but my settings are the defaults for things related to this

Image

Tyriar:
Fixes #275667

### DIFF
--- a/src/vscode-dts/vscode.d.ts
+++ b/src/vscode-dts/vscode.d.ts
@@ -7791,6 +7791,8 @@ declare module 'vscode' {
 		 * @param commandLine The command line to execute, this is the exact text that will be sent
 		 * to the terminal.
 		 *
+		 * @throws When run on a terminal doesn't support this API, such as task terminals.
+		 *
 		 * @example
 		 * // Execute a command in a terminal immediately after being created
 		 * const myTerm = window.createTerminal();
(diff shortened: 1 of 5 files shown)


anthonykim1 (approved):

src/vscode-dts/vscode.d.ts
```diff
@@ -7791,6 +7791,8 @@ declare module 'vscode' {
 		 * @param commandLine The command line to execute, this is the exact text that will be sent
 		 * to the terminal.
 		 *
+		 * @throws When run on a terminal doesn't support this API, such as task terminals.
```
alexr00, line 7794: I guess it's better that we break extensions that do this rather than allow this behavior.

alexr00 (approved):

src/vscode-dts/vscode.d.ts
Tyriar: Yeah it's somewhat breaking but we must throw as we can't return undefined or some fake/broken object here.

PR #275878, opened 2025-11-06
fix disposable leak

### LINKED ISSUE #275388: `chatTerminalToolProgressPart.ts` —  Trying to add a disposable to a DisposableStore

meganrogge:
fix #275388

### DIFF
--- a/src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
+++ b/src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
@@ -181,17 +181,27 @@ export class ChatTerminalToolProgressPart extends BaseChatToolInvocationSubPart 
 			this._registerInstanceListener(instance);
 		};
 
-		await attachInstance(await this._terminalChatService.getTerminalInstanceByToolSessionId(terminalToolSessionId));
+		const initialInstance = await this._terminalChatService.getTerminalInstanceByToolSessionId(terminalToolSessionId);
+		await attachInstance(initialInstance);
+
+		if (this._store.isDisposed) {
+			return;
+		}
 
 		const listener = this._terminalChatService.onDidRegisterTerminalInstanceWithToolSession(async instance => {
-			if (instance !== await this._terminalChatService.getTerminalInstanceByToolSessionId(terminalToolSessionId)) {
+			const registeredInstance = await this._terminalChatService.getTerminalInstanceByToolSessionId(terminalToolSessionId);
+			if (instance !== registeredInstance) {
 				return;
 			}
-			attachInstance(instance);
-			listener.dispose();
+			await attachInstance(instance);
+			this._store.delete(listener);
 		});
-		if (!(this._store.isDisposed)) {
-			this._register(listener);
+
+		if (this._store.isDisposed) {
+			listener.dispose();
+			return;
+		} else {
+			this._store.add(listener);
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
```diff
@@ -190,7 +190,9 @@ export class ChatTerminalToolProgressPart extends BaseChatToolInvocationSubPart
 			attachInstance(instance);
 			listener.dispose();
 		});
-		this._register(listener);
+		if (!(this._store.isDisposed)) {
+			this._register(listener);
+		}
```
Tyriar: The handle to `listener` would leak here when `this._store.isDisposed`? Also we shouldn't register against `this` as listener is short lived. I think this is what we need to do:

- Check `isDisposed` before `const listener`, that way it won't be created if it's disposed already
- Change `listener.dispose()` to `this._store.delete(listener)` to free the handle on `this._store` in addition to disposing `listener`
- nit: Pull await call out as it's harder to think about when it's in a condition

Tyriar (dismissed):

src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
```diff
@@ -181,16 +181,28 @@ export class ChatTerminalToolProgressPart extends BaseChatToolInvocationSubPart
...
-			attachInstance(instance);
-			listener.dispose();
+			await attachInstance(instance);
+			this._store.delete(listener);
 		});
-		this._register(listener);
+
+		if (this._store.isDisposed) {
+			listener.dispose();
+			return;
```
Tyriar: This case shouldn't be possible as you checked just before setting up the listener, and listening to an event is sync and couldn't cause it to be disposed

src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
```diff
@@ -181,16 +181,28 @@ export class ChatTerminalToolProgressPart extends BaseChatToolInvocationSubPart
...
 		const listener = this._terminalChatService.onDidRegisterTerminalInstanceWithToolSession(async instance => {
-			if (instance !== await this._terminalChatService.getTerminalInstanceByToolSessionId(terminalToolSessionId)) {
+			const registeredInstance = await this._terminalChatService.getTerminalInstanceByToolSessionId(terminalToolSessionId);
+			if (instance !== registeredInstance) {
 				return;
 			}
-			attachInstance(instance);
-			listener.dispose();
+			await attachInstance(instance);
+			this._store.delete(listener);
```
Tyriar: Safer:

```suggestion
			this._store.delete(listener);
			await attachInstance(instance);
```

Tyriar (dismissed):

Tyriar (approved):

PR #276108, opened 2025-11-07
rm unneeded event

meganrogge:
cc @alexr00


alexr00 (approved):
🙏 Since the API is being finalized, we should also have this in the release branch.

PR #276211, opened 2025-11-07
make terminal output feature accessible to screen reader users

### LINKED ISSUE #274220: Make terminal inline output accessible
May render the content as text vs HTML in screen reader mode and/or add a specific accessible view for this

cc @Tyriar

meganrogge:
fix #274220

If/when we end up using xterm here instead, we can reuse all of this and just change the `getCommandAndOutputAsText` function.

Screenshot 2025-11-07 at 4 31 37 PM

Screenshot 2025-11-07 at 4 33 39 PM


src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
```diff
@@ -446,6 +464,68 @@ export class ChatTerminalToolProgressPart extends BaseChatToolInvocationSubPart
...
+
+	private _updateOutputAriaLabel(): void {
+		if (!this._outputScrollbar) {
+			return;
+		}
+		const scrollableDomNode = this._outputScrollbar.getDomNode();
+		scrollableDomNode.setAttribute('role', 'region');
+		const accessibleViewHint = this._accessibleViewService.getOpenAriaHint(AccessibilityVerbositySettingId.TerminalChatOutput);
+		const label = accessibleViewHint
+			? localize('chatTerminalOutputAriaLabelWithHint', '{0}. {1}', this._outputAriaLabelBase, accessibleViewHint)
```
Tyriar: This looks hard to localize without some hint. Could we skip localizing if it was a `\n` instead?

src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
```diff
@@ -53,6 +57,7 @@ const sanitizerConfig = Object.freeze<DomSanitizerConfig>({
 
 export class ChatTerminalToolProgressPart extends BaseChatToolInvocationSubPart {
 	public readonly domNode: HTMLElement;
+	private static _lastFocusedPart: WeakRef<ChatTerminalToolProgressPart> | undefined;
```
Tyriar: Is this how this is normally done? Best to not deal with statics or weak refs unless absolutely necessary

Tyriar (dismissed):

Tyriar (approved):

PR #276409, opened 2025-11-10
Terminal suggest - include persistent options in suggestions and improve suggestion grouping

### LINKED ISSUE #272352: Terminal suggest: ability to group arg suggestions together
Does this issue occur when all extensions are disabled?: N/A

- VS Code Version: `main`
- OS Version: Windows 11

I was testing out a new Fig spec (https://github.com/microsoft/vscode/pull/272348) and noticed that argument suggestions aren't grouped together by default:

Image

I think it would be useful to have the ability to group them together instead of having them mixed with options/flags.

Spec snippet:

```ts
{
	name: ['extension', 'ext'],
	description: 'Manage azd extensions.',
	subcommands: [
		{
			name: ['install'],
			description: 'Installs specified extensions.',
			options: [
				{
					name: ['--force', '-f'],
					description: 'Force installation even if it would downgrade the current version',
					isDangerous: true,
				},
				...
			],
			args: {
				name: 'extension-id',
				generators: azdGenerators.listExtensions,
			},
		},
	],
}
```

### LINKED ISSUE #276406: Terminal suggest: symlinks are not grouped together with regular files/folders
Notice how the symlinks are grouped with options/flags instead of the regular files/folders:
Image

### LINKED ISSUE #276407: Terminal suggest: options/flags marked as `isPersistent` are not shown at all
For example, the GitHub CLI spec has a root level `--help` option that's marked as persistent, but `--help` does not appear in the root command suggestions or any of its subcommands:

https://github.com/microsoft/vscode/blob/6a3b90260e379120bdca24d1053cdc5361a76c4a/extensions/terminal-suggest/src/completions/gh.ts#L3393-L3400

Image

Image

JeffreyCA:
Fixes #272352 
Arguments are now grouped together instead of being mixed with flags/options:
arg-grouped

Fixes #276406
Symlink files and folders are now grouped together with regular files and folders
symlinks

Fixes #276407
Options marked with `isPersistent: true` are now shown in suggestions:
persistent

### DIFF
--- a/extensions/terminal-suggest/src/fig/figInterface.ts
+++ b/extensions/terminal-suggest/src/fig/figInterface.ts
@@ -341,6 +341,7 @@ export async function collectCompletionItemResult(
 	}
 	if (parsedArguments.suggestionFlags & SuggestionFlag.Options) {
 		await addSuggestions(parsedArguments.completionObj.options, vscode.TerminalCompletionItemKind.Flag, parsedArguments);
+		await addSuggestions(parsedArguments.completionObj.persistentOptions, vscode.TerminalCompletionItemKind.Flag, parsedArguments);
 	}
 
 	return { showFiles, showFolders, fileExtensions };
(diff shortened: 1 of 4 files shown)


meganrogge (approved):
Thank you!

Tyriar (approved):

PR #276420, opened 2025-11-10
Clear attachPersistantProcess for relaunch

### LINKED ISSUE #228359: relaunching the terminal will often just close the terminal
Steps to Reproduce:

1. With the python extension installed, open a terminal
2. use `python: select interpreter` so that the python extension will want to relaunch the terminal with the new environment
3. select relaunch terminal from the terminal hover UI

:bug: about half the time, the terminal just closes

anthonykim1:
Resolves: https://github.com/microsoft/vscode/issues/228359 
Was able to repro on Windows, specifically when `relaunch terminal` right after reloading window or restarting VS Code.
 - Bug doesnt occur if I open new terminal before relaunching the terminal with warning sign.
 

I thought the `attachPersistentProcess` is only meant for initial revival/restoration after reload, not for user-initiated relaunches.

I think the problem is that we try to shutdown process and attach to that same process because `attachPersistentProcess` gets passed on relaunch.

### DIFF
--- a/src/vs/workbench/contrib/terminal/browser/terminalInstance.ts
+++ b/src/vs/workbench/contrib/terminal/browser/terminalInstance.ts
@@ -1826,7 +1826,12 @@ export class TerminalInstance extends Disposable implements ITerminalInstance {
 
 	@debounce(1000)
 	relaunch(): void {
-		this.reuseTerminal(this._shellLaunchConfig, true);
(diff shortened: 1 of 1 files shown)


Tyriar (approved):
This makes sense and seems like it should be fairly safe. Good find 👍

PR #276477, opened 2025-11-10
terminal suggest - adopt `vscode.fs` watcher

bpasero:
@meganrogge do you recall why you were using `fs` for file watching and not our `vscode` API?

//cc @hediet


meganrogge (dismissed):
I think I was not aware of this API. Thank you!

sandy081 (approved):

PR #276478, opened 2025-11-10
Update editor info color for better contrast

### LINKED ISSUE #276475: Low contrast issue
Testing #253045

Image

I was too late to take the picture with the blue font, but I think its contrast is too low.

alexr00:
Fixes https://github.com/microsoft/vscode/issues/276475


mrleemurray (approved):
Great, thanks for fixing this

PR #276492, opened 2025-11-10
Remove special case margin from hover span

### LINKED ISSUE #275115: Model hover has bottom padding
Not sure why the bottom has a bunch of padding?

Image

Ideal (edited in image editor):

Image

Tyriar:
Fixes #275115

cc @alexr00


bpasero (approved):

alexr00 (commented):
Checked the case that this was initially added for, and it looks fine 👍

PR #276518, opened 2025-11-10
Mark Cursor mdc files as markdown

remcohaszing:
The Cursor editor uses `.mdc` files to configure its LLM. It uses the markdown language ID for these files. People tend to commit these files. For users whose coworkers use Cursor, it’s nice if VSCode also uses the markdown language ID for these files.

### DIFF
--- a/extensions/markdown-basics/package.json
+++ b/extensions/markdown-basics/package.json
@@ -28,6 +28,9 @@
           ".mdtext",
           ".workbook"
         ],
+        "filenamePatterns": [
+          "**/.cursor/rules/*.mdc"
+        ],
         "configuration": "./language-configuration.json"
       }
     ],
--- a/extensions/markdown-basics/package.json
+++ b/extensions/markdown-basics/package.json
@@ -29,7 +29,7 @@
           ".workbook"
         ],
         "filenamePatterns": [
-          "**/.cursor/rules/*.mdc"
+          "**/.cursor/**/*.mdc"
         ],
         "configuration": "./language-configuration.json"
       }


extensions/markdown-basics/package.json
```diff
@@ -28,6 +28,9 @@
           ".mdtext",
           ".workbook"
         ],
+        "filenamePatterns": [
+          "**/.cursor/rules/*.mdc"
```
mjbvz: Any reason not to apply this for all `.mdc` files?

mjbvz (commented):
Thanks. Should also be added [here](https://github.com/microsoft/vscode/blob/6293b5452cb221b1da20f22c7d0e03b60a0c0e4e/extensions/markdown-language-features/src/util/file.ts#L19) if you want most markdown features to work in these files

extensions/markdown-basics/package.json
remcohaszing: Nuxt uses `.mdc` files for their own custom markdown variation. They register this in their [own VSCode extension](https://github.com/nuxt-content/vscode-mdc) (>250k installs). So in that sense there’s a bit of a conflict between Nuxt and Cursor.

Because Cursor uses files in a specific directory, we can target only those specific files with a filename pattern.

With that information I’m unsure if you still think if `.mdc` should be added to https://github.com/microsoft/vscode/blob/6293b5452cb221b1da20f22c7d0e03b60a0c0e4e/extensions/markdown-language-features/src/util/file.ts#L19

mjbvz: Ok thanks, let's make the language mapping more generic then but keep it out from the main md language intellisense:

```suggestion
          "**/.cursor/**/*.mdc"
```

mjbvz (approved):

rzhao271 (approved):

PR #276524, opened 2025-11-10
account for height of horizontal scrollbar so terminal output is not cutoff

### LINKED ISSUE #275884: bottom of terminal output slightly cutoff
> Yes. Confirming that I also see this on a mac when I scroll to the bottom
> 
> Image 

 _Originally posted by @bhavyaus in [#275690](https://github.com/microsoft/vscode/issues/275690#issuecomment-3498618602)_

meganrogge:
fix #275884

Screenshot 2025-11-10 at 10 46 01 AM


Tyriar (changes requested):
This isn't what I had in mind for this. The scroll element's contents should show regardless of whether the scroll bar is underneath, I think the bug here is that the calculated size is not correct.

Because of this, adding the scroll bar height would probably also fall over theoretically when the font size becomes large enough.

PR #276554, opened 2025-11-10
detect `press any/a key` and ask if user wants to send `a` to terminal

### LINKED ISSUE #275605: Copilot Agent blocked by "Press any key to continue ..." - Loss of autonomous capabilities
Does this issue occur when all extensions are disabled?: N/A

- VS Code Version: 1.106.0 Insider (05/11/2025)
- OS Version: Windows 11 25H2 (26200.7019)

The issue is about VS Code not allowing Copilot Chat extension (and possibly other AI extensions) to proceed in Terminal when some commands/scripts are finished. The point of **"Press any key to continue ..."** is being taken as not finished yet, and requires the human user to press to proceed.

This is a blocker for the autonomic functioning of the affected AI agent, and if this condition is hit multiple times per session, will significantly slow down the user's workflow. Imagine going AFK for a while and returning to discover it has only just began the task (and became blocked, - paused - on this), to then have to sit next to it to manually continue on further occasions throughout the task. Not everyone has their view settings to always show the Terminal console.

Image

Image

**Crucially, as you can see, the terminal session's in-progress spinner is still running the circle, so the agent doesn't consider it as done.** If the user doesn't bother to click it, open it, and spot that it was just waiting for user keystroke like Enter, he'd have never found out why the Agent was stuck indefinitely and would never be completing their task.

**Steps to Reproduce:**

1. Use Copilot Agent Chat and wait for it to need to run an affected PowerShell, .bat or .cmd command/script
2. Observe the whole Agent flow being stuck due to "Press any key to continue ..." being prompted in Terminal Window, hidden or retrievable by clicking the Agent window's active terminal session.

It can occur with many types of queries, commands and scripts, which are called by the AI agent extension's available tools. The degree of an user being affected depends on if they work with solutions that require the use of the impacted commands. I am affected because i let the AI agent rebuild my program, as you can see in the screenshot.

To fix this issue, VS Code must let extensions like GitHub Copilot automatically send a keystroke (or internal proceed-now) when the text "Press any key to continue ..." is detected in one of the agent's active terminal sessions.

meganrogge:
fixes #275605

### DIFF
--- a/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/executeStrategy/executeStrategy.ts
+++ b/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/executeStrategy/executeStrategy.ts
@@ -116,8 +116,10 @@ const LINE_ENDS_WITH_COLON_RE = /:\s*$/;
 
 const END = /\(END\)$/;
 
+const ANY_KEY = /press any key|press a key/i;
+
 export function detectsInputRequiredPattern(cursorLine: string): boolean {
-	return PS_CONFIRM_RE.test(cursorLine) || YN_PAIRED_RE.test(cursorLine) || YN_AFTER_PUNCT_RE.test(cursorLine) || CONFIRM_Y_RE.test(cursorLine) || LINE_ENDS_WITH_COLON_RE.test(cursorLine.trim()) || END.test(cursorLine);
+	return PS_CONFIRM_RE.test(cursorLine) || YN_PAIRED_RE.test(cursorLine) || YN_AFTER_PUNCT_RE.test(cursorLine) || CONFIRM_Y_RE.test(cursorLine) || LINE_ENDS_WITH_COLON_RE.test(cursorLine.trim()) || END.test(cursorLine) || ANY_KEY.test(cursorLine);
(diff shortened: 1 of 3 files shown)


src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/executeStrategy/executeStrategy.ts
```diff
@@ -116,8 +116,10 @@ const LINE_ENDS_WITH_COLON_RE = /:\s*$/;
 
 const END = /\(END\)$/;
 
+const ANY_KEY = /press any key|press a key/i;
```
Tyriar, line 119:
```suggestion
const ANY_KEY = /press a(?:ny)? key/i;
```

Tyriar (approved):

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/executeStrategy/executeStrategy.ts
meganrogge: I will adopt this in my related PR

PR #276580, opened 2025-11-10
Enable support for Http based MCP resources

dileepyavan:
fixes #259253

https://github.com/user-attachments/assets/5580b9cb-24ab-4571-ac79-1434eeab59ed

### DIFF
--- a/src/vs/workbench/contrib/mcp/common/mcpResourceFilesystem.ts
+++ b/src/vs/workbench/contrib/mcp/common/mcpResourceFilesystem.ts
@@ -22,7 +22,7 @@ import { IWorkbenchContribution } from '../../../common/contributions.js';
 import { McpServer } from './mcpServer.js';
 import { McpServerRequestHandler } from './mcpServerRequestHandler.js';
 import { IMcpService, McpCapability, McpResourceURI } from './mcpTypes.js';
-import { ValidateHttpResources } from './mcpTypesUtils.js';
+import { canLoadMcpNetworkResourceDirectly } from './mcpTypesUtils.js';
 import { MCP } from './modelContextProtocol.js';
 
 const MOMENTARY_CACHE_DURATION = 3000;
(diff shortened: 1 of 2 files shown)


Tyriar (dismissed):

src/vs/workbench/contrib/mcp/common/mcpTypesUtils.ts
```diff
@@ -91,3 +93,24 @@ export function mcpServerToSourceData(server: IMcpServer): ToolDataSource {
...
+
+
+/**
+ * Validates whether the given HTTP or HTTPS resource is allowed for the specified MCP server.
+ *
+ * @param resource The URI of the resource to validate.
+ * @param server The MCP server instance to validate against, or undefined.
+ * @returns True if the resource request is valid for the server, false otherwise.
+ */
+export function ValidateHttpResources(resource: URI, server: IMcpServer | undefined) {
```
connor4312: I'd suggest a camelCase name like `canLoadMcpNetworkResourceDirectly`

src/vs/workbench/contrib/mcp/common/mcpResourceFilesystem.ts
```diff
@@ -273,12 +275,28 @@ export class McpResourceFilesystem extends Disposable implements IWorkbenchContr
...
 		const { resourceURI, server } = this._decodeURI(uri);
+		const matchedServer = this._mcpService.servers.get().find(s => s.definition.id === server.definition.id);
+		//check for http/https resources and use web content extractor service to fetch the contents.
+		if (ValidateHttpResources(uri, matchedServer)) {
+			const extractURI = URI.parse(resourceURI.toString());
+			const result = await this._webContentExtractorService.extract([extractURI], { followRedirects: false });
+			return {
+				contents: result.map(r => {
+					if (r.status === 'ok') {
+						return { uri: resourceURI.toString(), text: r.result };
```
connor4312: If the status is not okay, we should fall back to requesting the resource from the MCP server

src/vs/workbench/contrib/mcp/common/mcpTypesUtils.ts
```diff
@@ -91,3 +93,24 @@ export function mcpServerToSourceData(server: IMcpServer): ToolDataSource {
...
+	let isResourceRequestValid = false;
+	if (resource.path.startsWith('/http/')) {
```
connor4312: I suggest matching the `protocol` unpacked `resourceURL` instead just so you know you're validating against the real URL that's being requested--and avoid the implicit dependency on how the namespaces resource URIs get constructed

connor4312 (commented):
Couple small comments!

src/vs/workbench/contrib/mcp/common/mcpTypesUtils.ts
dileepyavan: Updated

justschen (approved):

PR #276582, opened 2025-11-10
Hover on keyboard modifier should trigger instantly

### LINKED ISSUE #276558: Hover on keyboard modifier should trigger instantly
https://github.com/microsoft/vscode/pull/274001 added support for a new setting to only allow hovers when a keyboard modifier is held.

The issue is:
- Hover over something (no hover comes up)
- Press ctrl or alt
- 🐛 See that no hover comes up UNLESS you move the mouse

## Original Report
----

This looks good! One thing I noticed though is that when the setting is `onModifierKeyPressed` and I press `option` the hover does not appear immediately, it only appears after I move the mouse again so the calculation retriggers. I think if we implement this it would make sense to add code to immediately trigger the hover (after a delay potentially) without having to move the mouse again.

_Originally posted by @aiday-mar in https://github.com/microsoft/vscode/issues/274001#issuecomment-3481296095_

ChaseKnowlden:
Fixes #276558


src/vs/editor/contrib/hover/browser/contentHoverController.ts
```diff
@@ -269,6 +271,33 @@ export class ContentHoverController extends Disposable implements IEditorContrib
...
 		}
+		// New behavior: if hover is configured for keyboard modifier and the user presses the triggering modifier
+		// we should show the hover immediately for the current (last known) mouse location even without moving the mouse again.
+		if (this._hoverSettings.enabled === 'onKeyboardModifier' && this._mouseMoveEvent) {
+			const multiCursorModifier = this._editor.getOption(EditorOption.multiCursorModifier); // 'altKey' | 'ctrlKey' | 'metaKey'
+			// Inverse logic: if multiCursor uses Alt (multiCursorModifier === 'altKey'), hover triggers on Ctrl/Cmd.
+			// Otherwise (multiCursorModifier is ctrlKey/metaKey), hover triggers on Alt.
+			const triggerPressed = multiCursorModifier === 'altKey'
+				? (e.ctrlKey || e.metaKey)
+				: e.altKey;
```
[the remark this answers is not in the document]
ChaseKnowlden: Fixed.

src/vs/editor/contrib/hover/browser/contentHoverController.ts
```diff
@@ -34,6 +34,7 @@ interface IHoverSettings {
 	readonly enabled: 'on' | 'off' | 'onKeyboardModifier';
 	readonly sticky: boolean;
 	readonly hidingDelay: number;
+	readonly delay: number; // show delay from editor hover options
```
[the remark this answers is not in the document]
ChaseKnowlden: Fixed.

ChaseKnowlden:
Ready to check.

src/vs/editor/contrib/hover/browser/contentHoverController.ts
```diff
@@ -269,6 +269,29 @@ export class ContentHoverController extends Disposable implements IEditorContrib
 		if (this._ignoreMouseEvents) {
 			return;
 		}
+		// New behavior: if hover is configured for keyboard modifier and the user presses the triggering modifier
```
benvillalobos: We can remove this comment. The code explains itself here.

src/vs/editor/contrib/hover/browser/contentHoverController.ts
```diff
@@ -269,6 +269,29 @@ export class ContentHoverController extends Disposable implements IEditorContrib
...
+		// we should show the hover immediately for the current (last known) mouse location even without moving the mouse again.
+		if (this._hoverSettings.enabled === 'onKeyboardModifier' && this._mouseMoveEvent) {
+			const multiCursorModifier = this._editor.getOption(EditorOption.multiCursorModifier); // 'altKey' | 'ctrlKey' | 'metaKey'
+			const triggerPressed = isTriggerModifierPressed(multiCursorModifier, e);
+			if (triggerPressed) {
```
benvillalobos: Nit, this is simpler:
```suggestion
			const multiCursorModifier = this._editor.getOption(EditorOption.multiCursorModifier);
			if (isTriggerModifierPressed(multiCursorModifier, e)) {
```

src/vs/editor/contrib/hover/browser/hoverUtils.ts
```diff
@@ -37,9 +51,6 @@ export function shouldShowHover(
 	if (hoverEnabled === 'off') {
 		return false;
 	}
-	if (multiCursorModifier === 'altKey') {
-		return mouseEvent.event.ctrlKey || mouseEvent.event.metaKey;
-	} else {
-		return mouseEvent.event.altKey;
-	}
+	// onKeyboardModifier
```
benvillalobos: nit: This comment is unnecessary

benvillalobos (commented):
Looking good! I noticed a bug where the hover wouldn't go away when moving the mouse to an area that wasn't the squiggly or the hover itself.

When the mouse moves away from the hover and the squiggly, the hover should go away. It should essentially mimic the behavior of how hover works in the "On" state.

https://github.com/user-attachments/assets/164c7de7-013a-478e-9e56-f685833f6cfe

src/vs/editor/contrib/hover/browser/contentHoverController.ts
ChaseKnowlden: Removed.

src/vs/editor/contrib/hover/browser/contentHoverController.ts
ChaseKnowlden: Fixed.

src/vs/editor/contrib/hover/browser/hoverUtils.ts
ChaseKnowlden: Removed.

benvillalobos:
Thanks for the updates @ChaseKnowlden! Before we merge, we should fix the issue mentioned here: https://github.com/microsoft/vscode/pull/276582#pullrequestreview-3484325169

Any idea what might be causing it? Happy to take a look as well if you'd like.

benvillalobos (dismissed):

src/vs/editor/contrib/hover/browser/hoverUtils.ts
```diff
@@ -37,9 +37,19 @@ export function shouldShowHover(
...
 		return false;
 	}
+	return isTriggerModifierPressed(multiCursorModifier, mouseEvent.event);
+}
+
+/**
+ * Returns true if the trigger modifier (inverse of multi-cursor modifier) is pressed.
+ * This works with both mouse and keyboard events by relying only on the modifier flags.
+ */
+export function isTriggerModifierPressed(
```
TylerLeonhardt, line 47: some tests for this would be good.

src/vs/editor/contrib/hover/browser/contentHoverController.ts
```diff
@@ -266,12 +266,18 @@ export class ContentHoverController extends Disposable implements IEditorContrib
...
 			return;
 		}
-		if (!this._contentWidget) {
-			return;
+
+		if (this._hoverSettings.enabled === 'onKeyboardModifier'
+			&& isTriggerModifierPressed(this._editor.getOption(EditorOption.multiCursorModifier), e)
+			&& this._mouseMoveEvent) {
+			if (!this._contentWidget.isVisible) {
+				this._contentWidget.showsOrWillShow(this._mouseMoveEvent);
```
TylerLeonhardt, line 277: should this return after? the rest of the logic in this function talks about hiding the content widget... so is there a scenario where this is shown and then hidden?

benvillalobos: This scenario would have been caught by the `isModifierKeyPressed` a few lines down, but it's cleaner to explicitly return here. Added

benvillalobos (dismissed):

benvillalobos (approved):

mjbvz (approved):

Yoyokrazy (approved):

PR #276601, opened 2025-11-10
store command and output so it can persist when the terminal is killed

### LINKED ISSUE #274870: store command output so it persists after the terminal is killed
When terminal command is finished, we should serialize the command output in addition to the command URI so that we can toggle the show output action even when the terminal's long gone.

We'll put those on `toolSpecificData` in `runInTerminalTool`.

_Originally posted by @Tyriar in https://github.com/microsoft/vscode/pull/274417#discussion_r2483717854_

meganrogge:
fixes #274870

Also makes the terminal focus action navigate to the command after reload of window

https://github.com/user-attachments/assets/f930a7d4-d68a-4acd-9e7a-5c177d520109

### DIFF
--- a/src/vs/workbench/contrib/chat/common/chatService.ts
+++ b/src/vs/workbench/contrib/chat/common/chatService.ts
@@ -316,6 +316,19 @@ export interface IChatTerminalToolInvocationData {
 	terminalToolSessionId?: string;
 	/** The predefined command ID that will be used for this terminal command */
 	terminalCommandId?: string;
+	/** Serialized URI for the command that was executed in the terminal */
+	terminalCommandUri?: UriComponents;
+	/** Serialized output of the executed command */
+	terminalCommandOutput?: {
+		text: string;
+		truncated?: boolean;
+		isHtml?: boolean;
+	};
+	/** Stored theme colors at execution time to style detached output */
+	terminalTheme?: {
+		background?: string;
+		foreground?: string;
+	};
 	autoApproveInfo?: IMarkdownString;
 }
 
--- a/src/vs/workbench/contrib/chat/common/chatService.ts
+++ b/src/vs/workbench/contrib/chat/common/chatService.ts
@@ -322,7 +322,6 @@ export interface IChatTerminalToolInvocationData {
 	terminalCommandOutput?: {
 		text: string;
 		truncated?: boolean;
-		isHtml?: boolean;
 	};
 	/** Stored theme colors at execution time to style detached output */
 	terminalTheme?: {
--- a/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/runInTerminalTool.ts
+++ b/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/runInTerminalTool.ts
@@ -791,7 +791,7 @@ export class RunInTerminalTool extends Disposable implements IToolImpl {
 			}
 			const serialized = await this._tryGetSerializedCommandOutput(instance, commandId);
 			if (serialized) {
-				toolSpecificData.terminalCommandOutput = { text: serialized.text, truncated: serialized.truncated, isHtml: true };
+				toolSpecificData.terminalCommandOutput = { text: serialized.text, truncated: serialized.truncated };
 				const theme = instance.xterm?.getXtermTheme();
 				if (theme) {
 					toolSpecificData.terminalTheme = { background: theme.background, foreground: theme.foreground };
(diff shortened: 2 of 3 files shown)


src/vs/workbench/contrib/chat/common/chatService.ts
```diff
@@ -316,6 +316,19 @@ export interface IChatTerminalToolInvocationData {
...
+	terminalCommandOutput?: {
+		text: string;
+		truncated?: boolean;
+		isHtml?: boolean;
+	};
+	/** Stored theme colors at execution time to style detached output */
+	terminalTheme?: {
+		background?: string;
+		foreground?: string;
+	};
```
Tyriar: Isn't the theme handled by `isHtml`?

src/vs/workbench/contrib/chat/common/chatService.ts
```diff
@@ -316,6 +316,19 @@ export interface IChatTerminalToolInvocationData {
...
 	/** The predefined command ID that will be used for this terminal command */
 	terminalCommandId?: string;
+	/** Serialized URI for the command that was executed in the terminal */
+	terminalCommandUri?: UriComponents;
+	/** Serialized output of the executed command */
+	terminalCommandOutput?: {
+		text: string;
+		truncated?: boolean;
+		isHtml?: boolean;
+	};
```
Tyriar, line 325: I always envisioned this being done in xterm.js instead of HTML when we do streaming, we need to do something like that to avoid the problems we're having now in chat with slow markdown rendering.

So we may end up changing this

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/runInTerminalTool.ts
```diff
@@ -627,14 +631,15 @@ export class RunInTerminalTool extends Disposable implements IToolImpl {
...
 				}));
-				const commandId = (toolSpecificData as IChatTerminalToolInvocationData).terminalCommandId;
 				const executeResult = await strategy.execute(command, token, commandId);
 				// Reset user input state after command execution completes
 				toolTerminal.receivedUserInput = false;
 				if (token.isCancellationRequested) {
 					throw new CancellationError();
 				}
 
+				await this._captureCommandArtifacts(toolSpecificData, toolTerminal.instance, commandId, executeResult.output);
```
Tyriar, line 641: Could kick this off here and not await? Any delays in returning a tool response is bad

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/runInTerminalTool.ts
```diff
@@ -762,6 +767,79 @@ export class RunInTerminalTool extends Disposable implements IToolImpl {
...
+
+	private static _escapeAsHtml(value: string): string {
+		return value
+			.replace(/\r\n/g, '\n')
+			.replace(/&/g, '&amp;')
+			.replace(/</g, '&lt;')
+			.replace(/>/g, '&gt;')
+			.replace(/"/g, '&quot;')
+			.replace(/'/g, '&#39;');
+	}
```
Tyriar: Do we have a helper for this? It's certainly not complete, not sure exactly what it's protecting against.

Tyriar (dismissed):

src/vs/workbench/contrib/chat/common/chatService.ts
meganrogge: We need to save the theme so once the terminal is disposed of and we recreate the output, it can be styled correctly. I think `isHtml` can actually be removed.

src/vs/workbench/contrib/chat/common/chatService.ts
meganrogge: Yes, makes sense that we will change this when it's xterm vs html

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/runInTerminalTool.ts
meganrogge: We needed that before we called `domSanitize.safeSetInnerHtml`. I believe that can be safely removed.

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/runInTerminalTool.ts
```diff
@@ -762,6 +767,69 @@ export class RunInTerminalTool extends Disposable implements IToolImpl {
...
+
+	// #region Command serialization
+
+	private _createTerminalCommandUri(instance: ITerminalInstance, commandId: string): URI {
+		const params = new URLSearchParams(instance.resource.query);
+		params.set('command', commandId);
+		return instance.resource.with({ query: params.toString() });
+	}
+
+	private async _captureCommandArtifacts(
```
Tyriar, line 780: Since this is a region containing related functionality, think about how we can pull this out. `runInTerminalTool` is the entry point to a lot of functionality, so we want it to be slim and straightforward

Tyriar (approved):

Issue #276599, opened 2025-11-10
selections not added as implicit context


justschen:
caused by https://github.com/microsoft/vscode/pull/275432/files which attempted to fix the issue where `implicit` was showing up

stable, implicit selection attached:
<img width="864" height="739" alt="Image" src="https://github.com/user-attachments/assets/3cc04519-40df-4d57-8016-626e7912c390" />

insiders, implicit selection not attached:
<img width="1408" height="686" alt="Image" src="https://github.com/user-attachments/assets/38550aa9-0cf4-4b88-8d6f-34dd1aad0cb1" />

cc. @alexr00

PR #276604, opened 2025-11-10
fix selections not included in implicit

justschen:
release branch fix for https://github.com/microsoft/vscode/issues/276599

(which just reverts https://github.com/microsoft/vscode/pull/275432)

### DIFF
--- a/src/vs/workbench/contrib/chat/browser/chatInputPart.ts
+++ b/src/vs/workbench/contrib/chat/browser/chatInputPart.ts
@@ -189,7 +189,7 @@ export class ChatInputPart extends Disposable implements IHistoryNavigationWidge
 
 		const contextArr = this.getAttachedContext(sessionResource);
 
-		if (this.implicitContext?.enabled && this.implicitContext?.value && this.configurationService.getValue<boolean>('chat.implicitContext.suggestedContext')) {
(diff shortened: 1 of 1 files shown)


mjbvz (approved):

src/vs/workbench/contrib/chat/browser/chatInputPart.ts
```diff
@@ -189,7 +189,7 @@ export class ChatInputPart extends Disposable implements IHistoryNavigationWidge
 
 		const contextArr = this.getAttachedContext(sessionResource);
 
-		if (this.implicitContext?.enabled && this.implicitContext?.value && this.configurationService.getValue<boolean>('chat.implicitContext.suggestedContext')) {
+		if ((this.implicitContext?.enabled && this.implicitContext?.value) || (this.implicitContext && !URI.isUri(this.implicitContext.value) && this.configurationService.getValue<boolean>('chat.implicitContext.suggestedContext'))) {
```
roblourens, line 192: Is `!isUri` correct here, because it will then always include `StringChatContextValue`?

roblourens: Side note, do we still need this setting?

justschen: i tried it out and i think it's fine, but yeah it's a revert from what it was before we added the new `value` type, so i'm not 100% sure

and yes, we probably do not need this setting anymore.

roblourens: You tried it with a custom context provider?

PR #276631, opened 2025-11-10
Fix various terminal suggest widget layout issues

### LINKED ISSUE #269178: Terminal suggest location is odd when in editor area in the side
https://github.com/user-attachments/assets/269a9621-7e73-45f5-8a29-8951f8ec3326

I'm getting completions, but they are appearing in non-ideal way.

pwang347:
## Overview
The goal here is to align the simple suggest widget's layout behaviour to the content suggest widget.

## Fixes

Addresses the following bugs:
1. Vertical out of bounds

| Before | After |
| -- | -- |
| image | image |

2. Only showing 1-2 items in "compat mode" (when the vertical space is small), it now expands to fill the remaining space. This only became a problem after the vertical out of bounds issue was addressed, so no "before/after" comparison.
3. Horizontal out of bounds

| Before | After |
| -- | -- |
| image | image |

4. Suggest widget is covered by other terminal window editors

| Before | After |
| -- | -- |
| image | image |

Fixes #269178


src/vs/workbench/services/suggest/browser/simpleSuggestWidget.ts
```diff
@@ -738,12 +738,19 @@ export class SimpleSuggestWidget<TModel extends SimpleCompletionModel<TItem>, TI
...
-		const fullHeight = info.statusBarHeight + this._list.contentHeight + this._messageElement.clientHeight + info.borderHeight;
+		// Cap list content height to a reasonable maximum (12 items worth), matching suggestWidget behavior
+		const cappedListContentHeight = Math.min(this._list.contentHeight, info.itemHeight * 12);
+		const fullHeight = info.statusBarHeight + cappedListContentHeight + this._messageElement.clientHeight + info.borderHeight;
 		const minHeight = info.itemHeight + info.statusBarHeight;
 		// const editorBox = dom.getDomNodePagePosition(this.editor.getDomNode());
 		// const cursorBox = this.editor.getScrolledVisiblePosition(this.editor.getPosition());
 		const editorBox = dom.getDomNodePagePosition(this._container);
-		const cursorBox = this._cursorPosition; //this.editor.getScrolledVisiblePosition(this.editor.getPosition());
+		// Convert absolute cursor position to relative position (relative to container)
```
pwang347, line 748: This was the main bug causing the original behaviour (should be using relative coords). The other changes  are to make the height of the suggest widget fill as much as it can when the height is small, instead of only showing around 1-2 items in the list

Image

This aligns with the suggest widget's behaviour

Image

meganrogge (approved):
Thanks, seems to work very well!

PR #276635, opened 2025-11-11
Use electron to trigger copy/paste in webviews

### LINKED ISSUE #239228: Migrate from unsupported paste execCommand in desktop
With https://github.com/electron/electron/pull/45277 Electron has dropped support for `document.execCommand("paste")` to align with browsers. We might be able to get some transition time via https://github.com/electron/electron/pull/45377

This issue captures the existing usages and the work needed to migrate to a web friendly equivalent api.

- [ ] Webview https://github.com/microsoft/vscode/blob/main/src/vs/workbench/contrib/webview/browser/webviewElement.ts
- [x] Editor clipboard https://github.com/microsoft/vscode/blob/main/src/vs/editor/contrib/clipboard/browser/clipboard.ts
- [x] Text input actions https://github.com/microsoft/vscode/blob/main/src/vs/workbench/browser/actions/textInputActions.ts **(This has a web fallback we can switch to)**
- [x] Input clipboard actions https://github.com/microsoft/vscode/blob/main/src/vs/workbench/contrib/codeEditor/electron-sandbox/inputClipboardActions.ts

mjbvz:
Fixes #239228

### DIFF
--- a/src/vs/platform/webview/common/webviewManagerService.ts
+++ b/src/vs/platform/webview/common/webviewManagerService.ts
@@ -36,7 +36,12 @@ export interface IWebviewManagerService {
 
 	setIgnoreMenuShortcuts(id: WebviewWebContentsId | WebviewWindowId, enabled: boolean): Promise<void>;
 
+	// Find in frame
 	findInFrame(windowId: WebviewWindowId, frameName: string, text: string, options: FindInFrameOptions): Promise<void>;
-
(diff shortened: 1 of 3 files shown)


deepak1556 (commented):
WebContents.copy/paste/cut action will only operate on a [focused frame](https://source.chromium.org/chromium/chromium/src/+/main:content/browser/web_contents/web_contents_impl.cc;l=4651-4677;) in the current webcontents frame tree, @mjbvz does this expectation match with the current usages in webview, do you perform these actions when the webview is not focused in anyway ?

PR #276642, opened 2025-11-11
Removing history custom node from agents view

osortega:

### DIFF
--- a/src/vs/workbench/contrib/chat/browser/chatSessions/common.ts
+++ b/src/vs/workbench/contrib/chat/browser/chatSessions/common.ts
@@ -117,12 +117,14 @@ function applyTimeGrouping(sessions: ChatSessionItemWithProvider[]): void {
 }
 
 // Helper function to process session items with timestamps, sorting, and grouping
-export function processSessionsWithTimeGrouping(sessions: ChatSessionItemWithProvider[]): void {
+export function processSessionsWithTimeGrouping(sessions: ChatSessionItemWithProvider[]): ChatSessionItemWithProvider[] {
+	const sessionsTemp = [...sessions];
 	// Only process if we have sessions with timestamps
 	if (sessions.some(session => session.timing?.startTime !== undefined)) {
-		sortSessionsByTimestamp(sessions);
-		applyTimeGrouping(sessions);
+		sortSessionsByTimestamp(sessionsTemp);
+		applyTimeGrouping(sessionsTemp);
 	}
+	return sessionsTemp;
 }
 
 // Helper function to create context overlay for session items
--- a/src/vs/workbench/contrib/chat/browser/chatSessions/common.ts
+++ b/src/vs/workbench/contrib/chat/browser/chatSessions/common.ts
@@ -168,3 +168,12 @@ export function getSessionItemContextOverlay(
 
 	return overlay;
 }
+
+export interface IChatSessionGroupItem extends ChatSessionItemWithProvider {
+	readonly isGroupNode: true;
+	readonly groupChildren: ChatSessionItemWithProvider[];
+}
+
+export function isGroupNode(item: ChatSessionItemWithProvider): item is IChatSessionGroupItem {
+	return 'isGroupNode' in item && (item as IChatSessionGroupItem).isGroupNode === true;
+}
--- a/src/vs/workbench/contrib/chat/browser/chatSessions/common.ts
+++ b/src/vs/workbench/contrib/chat/browser/chatSessions/common.ts
@@ -20,7 +20,6 @@ export const NEW_CHAT_SESSION_ACTION_ID = 'workbench.action.chat.openNewSessionE
 
 export type ChatSessionItemWithProvider = IChatSessionItem & {
 	readonly provider: IChatSessionItemProvider;
-	isHistory?: boolean;
 	relativeTime?: string;
 	relativeTimeFullWord?: string;
 	hideRelativeTime?: boolean;
@@ -145,15 +144,15 @@ export function getSessionItemContextOverlay(
 	}
 
 	// Mark history items
-	overlay.push([ChatContextKeys.isHistoryItem.key, session.isHistory]);
+	overlay.push([ChatContextKeys.isArchivedItem.key, session.archived]);
 
 	// Mark active sessions - check if session is currently open in editor or widget
 	let isActiveSession = false;
 
-	if (!session.isHistory && provider?.chatSessionType === localChatSessionType) {
+	if (!session.archived && provider?.chatSessionType === localChatSessionType) {
 		// Local non-history sessions are always active
 		isActiveSession = true;
-	} else if (session.isHistory && chatWidgetService && chatService && editorGroupsService) {
+	} else if (session.archived && chatWidgetService && chatService && editorGroupsService) {
 		// Check if session is open in a chat widget
 		const widget = chatWidgetService.getWidgetBySessionResource(session.resource);
 		if (widget) {
@@ -167,13 +166,4 @@ export function getSessionItemContextOverlay(
 	overlay.push([ChatContextKeys.isActiveSession.key, isActiveSession]);
 
 	return overlay;
-}
-
-export interface IChatSessionGroupItem extends ChatSessionItemWithProvider {
-	readonly isGroupNode: true;
-	readonly groupChildren: ChatSessionItemWithProvider[];
-}
-
-export function isGroupNode(item: ChatSessionItemWithProvider): item is IChatSessionGroupItem {
-	return 'isGroupNode' in item && (item as IChatSessionGroupItem).isGroupNode === true;
 }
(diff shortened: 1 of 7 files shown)


src/vs/workbench/contrib/chat/browser/chatSessions/view/sessionsTreeRenderer.ts
```diff
@@ -191,12 +190,27 @@ export class SessionsRenderer extends Disposable implements ITreeRenderer<IChatS
...
+	private renderGroupNode(groupNode: IChatSessionGroupItem, templateData: ISessionTemplateData): void {
+		templateData.customIcon.className = '';
+		templateData.descriptionRow.style.display = 'none';
+		templateData.timestamp.parentElement!.style.display = 'none';
 
+		const childCount = groupNode.groupChildren.length;
+		templateData.iconLabel.setLabel(groupNode.label, undefined, {
+			title: `${childCount} session${childCount === 1 ? '' : 's'}`
+		});
 	}
```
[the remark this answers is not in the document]
osortega: No icon for groups, it should only be the twistie

src/vs/workbench/contrib/chat/browser/chatSessions/view/sessionsTreeRenderer.ts
```diff
@@ -538,72 +551,67 @@ export class SessionsDataSource implements IAsyncDataSource<IChatSessionItemProv
...
+						label: groupName,
+						resource: URI.from({ scheme: 'chat-group', path: `/${groupName}` }),
+						provider: this.provider,
+						isGroupNode: true,
+						groupChildren: children,
+						timing: {
+							startTime: 0
+						}
+					};
+					result.push(groupNode);
```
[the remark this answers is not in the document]
osortega: That is done in the provider for now

src/vs/workbench/contrib/chat/browser/chatSessions/common.ts
```diff
@@ -168,3 +168,12 @@ export function getSessionItemContextOverlay(
 
 	return overlay;
 }
+
+export interface IChatSessionGroupItem extends ChatSessionItemWithProvider {
```
mjbvz: Can this class live in `sessionViewPane` / `sessionsTreeRenderer` instead of being exposed publicly?

src/vs/workbench/contrib/chat/browser/chatSessions/view/sessionsTreeRenderer.ts
```diff
@@ -191,12 +191,27 @@ export class SessionsRenderer extends Disposable implements ITreeRenderer<IChatS
...
+	}
+
+	private renderGroupNode(groupNode: IChatSessionGroupItem, templateData: ISessionTemplateData): void {
+		templateData.customIcon.className = '';
+		templateData.descriptionRow.style.display = 'none';
+		templateData.timestamp.parentElement!.style.display = 'none';
 
+		const childCount = groupNode.groupChildren.length;
+		templateData.iconLabel.setLabel(groupNode.label, undefined, {
+			title: `${childCount} session${childCount === 1 ? '' : 's'}`
```
dmitrivMS: Can it be localized?

src/vs/workbench/contrib/chat/common/chatSessionsService.ts
```diff
@@ -60,7 +60,7 @@ export interface IChatSessionsExtensionPoint {
 }
 export interface IChatSessionItem {
 	/** @deprecated Use {@link resource} instead */
-	id: string;
+	id?: string;
```
mjbvz: This seems unrelated to the rest of this change. Could it be done in a separate PR?

dmitrivMS (dismissed):

src/vs/workbench/contrib/chat/browser/chatSessions/view/sessionsTreeRenderer.ts
```diff
@@ -538,72 +551,66 @@ export class SessionsDataSource implements IAsyncDataSource<IChatSessionItemProv
...
+					processSessionsWithTimeGrouping(ungroupedItems);
 				}
 
-				return itemsWithProvider;
+				const result: ChatSessionItemWithProvider[] = [];
+				result.push(...ungroupedItems);
+				for (const [groupName, children] of groupedItems) {
+					const groupNode: IChatSessionGroupItem = {
+						label: groupName,
+						resource: URI.parse(`${Schemas.vscodeLocalChatSession}://history`),
```
mjbvz: I find these fake resources confusing. Do group nodes actually need to be extending a `ChatSessionItemWithProvider` or could you introduce a new type just for groups. You tree would then be of `ChatSessionItemWithProvider | NewGroupType`

src/vs/workbench/contrib/chat/browser/chatSessions/common.ts
bpasero: I agree, this should be an implementation detail of the classic view.

src/vs/workbench/contrib/chat/browser/chatSessions/common.ts
```diff
@@ -168,3 +168,12 @@ export function getSessionItemContextOverlay(
...
+	readonly isGroupNode: true;
+	readonly groupChildren: ChatSessionItemWithProvider[];
+}
+
+export function isGroupNode(item: ChatSessionItemWithProvider): item is IChatSessionGroupItem {
+	return 'isGroupNode' in item && (item as IChatSessionGroupItem).isGroupNode === true;
```
bpasero: Using `'xy' in` is discouraged, please do NOT use it. You should have seen the related discussion in #dev-discuss

src/vs/workbench/contrib/chat/common/chatSessionsService.ts
```diff
@@ -76,7 +76,7 @@ export interface IChatSessionItem {
 		insertions: number;
 		deletions: number;
 	};
-
+	groupName?: string;
```
bpasero: I am not sure what `groupName` is supposed to mean in the agent sessions API? Think about the single-view, how would it apply there?

I would suggest we rather think about introducing a new property to agentic sessions that signals if a session item has been archived. We likely need this concept in the future. And then I would argue our local chat history is archived chats.

src/vs/workbench/contrib/chat/browser/chatSessions/localChatSessionsProvider.ts
```diff
@@ -241,16 +241,34 @@ export class LocalChatSessionsProvider extends Disposable implements IChatSessio
 				}
 			}
 		});
+		const history = await this.getHistoryItems();
```
bpasero, line 244: As indicated in my previous comment, I would put in the history items as agent sessions with `archived: true` and then we need to adopt this properly in the classic view and the single view.

bpasero (changes requested):

src/vs/workbench/contrib/chat/browser/chatSessions/localChatSessionsProvider.ts
roblourens: I suggested something like this in our standup

src/vs/workbench/contrib/chat/browser/chatSessions/view/sessionsTreeRenderer.ts
osortega: Good call! Changed it

src/vs/workbench/contrib/chat/browser/chatSessions/common.ts
osortega: Removed this in favor of the ArchivedSessionItems class

src/vs/workbench/contrib/chat/common/chatSessionsService.ts
osortega: Removed

src/vs/workbench/contrib/chat/browser/chatSessions/common.ts
osortega: Removed it in favor of using a class

src/vs/workbench/contrib/chat/common/chatSessionsService.ts
osortega: Removed it in favor of `archived` property

src/vs/workbench/contrib/chat/browser/chatSessions/localChatSessionsProvider.ts
osortega: Changed it so that all history items are marked as archived, each renderer can now decide how to show them in the tree

lszomoru (approved):

src/vs/workbench/contrib/chat/browser/chatSessions/view/sessionsTreeRenderer.ts
osortega: Thanks for the tip, I did that and used `ArchivedSessionItems`

bpasero (approved):

PR #271950, opened 2025-10-17
chat - simplify send action by removing dropdown options

bpasero:

PR #276661, opened 2025-11-11
Support contribution to chat execute menu (for contributed agents)

rebornix:

### DIFF
--- a/extensions/git/src/commands.ts
+++ b/extensions/git/src/commands.ts
@@ -3515,7 +3515,8 @@ export class CommandCenter {
 	@command('git.createWorktreeWithDefaults', { repository: true, repositoryFilter: ['repository'] })
 	async createWorktreeWithDefaults(
 		repository: Repository,
-		commitish: string = 'HEAD'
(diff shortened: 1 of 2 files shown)


bpasero (commented):
Can we first have a discussion and plan for how we envision the delegation to work? In my mind it should go into the send-dropdown but offer a way to promote actions back to the top if the user wants to.

For reference, the dropdown was removed in https://github.com/microsoft/vscode/pull/271950

rebornix (commented):
@bpasero yep for sure. I'm not keen on having the delegate button (but mostly want to self host if it ever makes sense for CLI). One thing worth mentioning that CLI has a different gesture for "delegate". The last time when I tried it, it sends request to the cloud, and clears the whole local CLI session. It's a big surprise to me at the time.

bpasero:
I would like delegation to work like this, no matter where you delegate to:
* your local chat remains functional and visible
* you see a response in that chat that something was delegated (and a link to open the task)
* a new agentic session appears in the view reporting progress
* NO editor opens automatically unless I do a user action

PR #276751, opened 2025-11-11
add tsgo watcher

meganrogge:


extensions/typescript-language-features/package.json
```diff
@@ -1783,6 +1783,27 @@
           "${cwd}"
         ],
         "pattern": "$tsc"
+      },
+       {
+        "name": "tsgo",
```
mjbvz: Should this be `tsgo-watch`?

extensions/typescript-language-features/package.json
```diff
@@ -1783,6 +1783,27 @@
...
+        "label": "%typescript.problemMatchers.tsgo.label%",
```
mjbvz: Looks like this string is missing from the localizations

extensions/typescript-language-features/package.json
meganrogge: yes, updated

mjbvz (approved):

extensions/typescript-language-features/package.json
meganrogge: added it

PR #276754, opened 2025-11-11
Allow setting a side for `SimpleSuggestWidget` details, set to `right` for terminal to prevent input from being blocked

meganrogge:
Before:

Detail is rendered on the left, sometimes blocking the terminal input

image

After:

Detail is always rendered on the right to avoid ever blocking terminal input, so is sometimes off screen

Screenshot 2025-11-11 at 12 26 43 PM


Tyriar (changes requested):
We shouldn't need a setting for this, instead the detail in the screenshot should be anchored at the bottom and never obscure the cursor line:

Screenshot 2025-11-11 at 9 32 18 am

PR #276763, opened 2025-11-11
allow disabling certain detail orientation, add `north` rendering 

### LINKED ISSUE #276741: terminal hints are blocking input

- VS Code Version: 
- OS Version: 

Steps to Reproduce:

I can't really see what I'm typing here

Image

meganrogge:
fix #276741

We never want the details to render `west` of the list even if that fits best. The next best fitting in certain cases, `east`, looks very crammed. Adds `north` rendering to address this.

Before:

image

After:

Screenshot 2025-11-11 at 1 04 49 PM

### DIFF
--- a/src/vs/workbench/services/suggest/browser/simpleSuggestWidget.ts
+++ b/src/vs/workbench/services/suggest/browser/simpleSuggestWidget.ts
@@ -83,7 +83,6 @@ export interface IWorkbenchSuggestWidgetOptions {
 
 	/**
 	 * Disables specific detail placements when positioning the details overlay.
-	 * Valid values: `'east' | 'west' | 'south' | 'north'`.
 	 */
 	preventDetailsPlacements?: readonly SimpleSuggestDetailsPlacement[];
 }
--- a/src/vs/workbench/services/suggest/browser/simpleSuggestWidget.ts
+++ b/src/vs/workbench/services/suggest/browser/simpleSuggestWidget.ts
@@ -21,7 +21,7 @@ import { SuggestWidgetStatus } from '../../../../editor/contrib/suggest/browser/
 import { MenuId } from '../../../../platform/actions/common/actions.js';
 import { IConfigurationService } from '../../../../platform/configuration/common/configuration.js';
 import { IStorageService, StorageScope, StorageTarget } from '../../../../platform/storage/common/storage.js';
-import { canExpandCompletionItem, SimpleSuggestDetailsOverlay, SimpleSuggestDetailsWidget } from './simpleSuggestWidgetDetails.js';
+import { canExpandCompletionItem, SimpleSuggestDetailsOverlay, SimpleSuggestDetailsWidget, type SimpleSuggestDetailsPlacement } from './simpleSuggestWidgetDetails.js';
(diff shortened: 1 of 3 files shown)


src/vs/workbench/services/suggest/browser/simpleSuggestWidget.ts
```diff
@@ -80,6 +80,12 @@ export interface IWorkbenchSuggestWidgetOptions {
 	 * The setting for selection mode.
 	 */
 	selectionModeSettingId?: string;
+
+	/**
+	 * Disables specific detail placements when positioning the details overlay.
+	 * Valid values: `'east' | 'west' | 'south' | 'north'`.
```
Tyriar: Redundant, possible future bitrot


```suggestion
```

src/vs/workbench/services/suggest/browser/simpleSuggestWidgetDetails.ts
```diff
@@ -23,6 +23,8 @@ export function canExpandCompletionItem(item: SimpleCompletionItem | undefined):
 
 export const SuggestDetailsClassName = 'suggest-details';
 
+export type SimpleSuggestDetailsPlacement = 'east' | 'west' | 'south' | 'north';
```
Tyriar: const enum without values would be better as they end up being numbers in compiled code.

src/vs/workbench/services/suggest/browser/simpleSuggestWidgetDetails.ts
```diff
@@ -415,8 +420,28 @@ export class SimpleSuggestDetailsOverlay {
...
+		const placementEntries: [SimpleSuggestDetailsPlacement, Placement][] = [
+			['east', eastPlacement],
+			['south', southPacement],
+			['north', northPlacement],
+			['west', westPlacement]
+		];
+		const filteredEntries = this._preventPlacements
+			? placementEntries.filter(([direction]) => !this._preventPlacements!.has(direction))
+			: placementEntries;
+		const placements = (filteredEntries.length ? filteredEntries : placementEntries).map(([, entry]) => entry);
```
Tyriar: nit: This is a bit messy and could be simplified

Tyriar (dismissed):

src/vs/workbench/services/suggest/browser/simpleSuggestWidgetDetails.ts
```diff
@@ -23,6 +23,13 @@ export function canExpandCompletionItem(item: SimpleCompletionItem | undefined):
 
 export const SuggestDetailsClassName = 'suggest-details';
 
+export const enum SimpleSuggestDetailsPlacement {
+	East = 'east',
+	West = 'west',
+	South = 'south',
+	North = 'north'
+}
```
Tyriar, line 31: nit: It should always be referenced using this enum and not a string, so setting it as a string isn't necessary here and will just end with a large bundle size.

Tyriar (dismissed):

meganrogge:
Looks like CI is failing so i'll make that change when I fix that

alexdima (approved):

PR #276816, opened 2025-11-11
store expanded state for terminal output items

### LINKED ISSUE #275929: Chat terminal widget hides itself after scrolling up and back down
On Windows 11 with pwsh 7.

1. Ask Copilot to run ls in panel chat. Repeat four more times.
2. Expand the last ls output.
3. Scroll up. The last ls output should be off-screen, now.
4. Scroll back down.
5. :bug: The last ls output is collapsed again.

meganrogge:
fixes #275929

When you scroll the chat list far enough, it throws away off screen rows. Later, when you scroll back up, it rebuilds the row for that tool invocation. The rebuilt part starts in its default collapsed state unless we remember what the user did before, which is why we now store the expanded flag so the new instance opens back up automatically.

Not sure if we care about storing this on reload of the window. I lean toward no here, but lmk what you think @tyriar

https://github.com/user-attachments/assets/639773a4-629a-4ccc-9959-aa0b0dde6445

### DIFF
--- a/src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
+++ b/src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
@@ -58,6 +58,11 @@ const sanitizerConfig = Object.freeze<DomSanitizerConfig>({
 
 let lastFocusedProgressPart: ChatTerminalToolProgressPart | undefined;
 
+/**
+ * Remembers whether a tool invocation was last expanded so state survives virtualization re-renders.
+ */
+const expandedStateByInvocation = new WeakMap<IChatToolInvocation | IChatToolInvocationSerialized, boolean>();
+
 export class ChatTerminalToolProgressPart extends BaseChatToolInvocationSubPart {
 	public readonly domNode: HTMLElement;
 
@@ -182,6 +187,10 @@ export class ChatTerminalToolProgressPart extends BaseChatToolInvocationSubPart 
 
 		const progressPart = this._register(_instantiationService.createInstance(ChatProgressSubPart, elements.container, this.getIcon(), terminalData.autoApproveInfo));
 		this.domNode = progressPart.domNode;
+
+		if (expandedStateByInvocation.get(toolInvocation)) {
+			void this._toggleOutput(true);
+		}
 	}
 
 	private async _createActionBar(elements: { actionBar: HTMLElement }): Promise<void> {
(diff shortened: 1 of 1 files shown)


src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
```diff
@@ -58,6 +58,11 @@ const sanitizerConfig = Object.freeze<DomSanitizerConfig>({
 
 let lastFocusedProgressPart: ChatTerminalToolProgressPart | undefined;
 
+/**
+ * Remembers whether a tool invocation was last expanded so state survives virtualization re-renders.
+ */
+const expandedStateByInvocation = new WeakMap<IChatToolInvocation | IChatToolInvocationSerialized, boolean>();
```
Tyriar, line 64: If that's the reason it's collapsing on scroll, we should serialize it properly, then it'll work on scroll and reload I think? @roblourens @justschen any tips?

Tyriar (changes requested):

src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
meganrogge: You mean adding it to `IChatTerminalToolInvocationData`? I thought we only wanted to touch that in the tool

Tyriar: Not sure what the right way to do it is, we expand other items like search that need to retain that state when scrolling up and down though. Could check there?

roblourens: The state for list tool parts is not implemented correctly, IIRC. They all pile onto `usedReferencesExpanded`. We don't really have an established pattern for storing and persisting view state that is per-content-part. I'm not exactly sure what it should look like but someone could invent it! The WeakMap isn't optimal but seems ok in the meantime.

roblourens: Ideally I think you would want the parts being rendered to have a way to declare view state, which would get picked up in `IChatViewState` and persisted with other chat widget view state.

And what's the key for the view state? Tool calls have a unique ID but other parts don't, so I guess ideally keyed by the index of the part.

Tyriar (approved):
Let's go with this then until proper view state serialization happens 👍

PR #276823, opened 2025-11-11
Fix task rerun

### LINKED ISSUE #248866: rerun task button for task terminals always runs the last task

- VS Code Version: v1.101 insiders
- OS Version: macOS Sequoia 15.4.1

Steps to Reproduce:

1. Add 2 tasks to tasks.json or run tasks from contributed extensions.
2. Run task 1, then run task 2.
3. Click the rerun task button on the first task.

### Expected

Task 1 is rerun.

### Actual

Task 2 is rerun

The button seems to always rerun the last task similar to `rerun last task` command regardless of which terminal the button was clicked within.

https://github.com/user-attachments/assets/72e3bc04-a323-4801-bb02-881d29aaa1df

pwang347:
## Overview
- Since task is no longer in active map, we aren't able to retrieve it (add fallback mechanism that does a lookup using the lastTask key)
- Update task restart logic to handle case where it's already terminated

Fixes https://github.com/microsoft/vscode/issues/248866

### DIFF
--- a/src/vs/workbench/contrib/tasks/browser/abstractTaskService.ts
+++ b/src/vs/workbench/contrib/tasks/browser/abstractTaskService.ts
@@ -2172,22 +2172,31 @@ export abstract class AbstractTaskService extends Disposable implements ITaskSer
 		if (!this._taskSystem) {
 			return;
 		}
-		const response = await this._taskSystem.terminate(task);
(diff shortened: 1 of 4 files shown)


meganrogge (approved):
Thank you!

PR #276917, opened 2025-11-12
Fixed maximize/minimize toggle icon

Rahul-069:
Fix : Updated bottom panel maximize toggle icon
Description : This PR fixes the issue where the bottom panel’s maximize toggle button displayed the maximize icon even after expanding. The icon now correctly switches between maximize and minimize depending on the panel’s current state.

Steps to Test :
1. Open VS Code (or your local build)
2. Open bottom panel (e.g. Terminal)
3. Look on the maximize toggle icon:
     When the panel is maximized, it shows the minimize icon
     When the panel is minimized, it shows the maximize icon
4. The functionality and animations remain unchanged

### DIFF
--- a/src/vs/workbench/browser/parts/panel/panelActions.ts
+++ b/src/vs/workbench/browser/parts/panel/panelActions.ts
@@ -23,6 +23,7 @@ import { KeybindingWeight } from '../../../../platform/keybinding/common/keybind
 import { SwitchCompositeViewAction } from '../compositeBarActions.js';
 
 const maximizeIcon = registerIcon('panel-maximize', Codicon.screenFull, localize('maximizeIcon', 'Icon to maximize a panel.'));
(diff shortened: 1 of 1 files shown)


bpasero:
I actually think this is as intended, you have to pick either the state it is going to be in when you click or the current state, there is no right or wrong, but it should be consistent. We use this icon in other places too.

//cc @mrleemurray

Rahul-069:
@bpasero Thanks for the clarification!
I was trying to show the next state(usual convention). Even if we go with the current state approach, the minimized panel still shows the maximize icon, which feels inconsistent.
The same inconsistency is also present with the maximize icon on the secondary side bar.
Apologies if I missed something. I’m looking forward to your feedback on this.

Rahul-069:
@bpasero Yes!

PR #276959, opened 2025-11-12
fix: resolve tsconfig extends paths when using package exports remapping

### LINKED ISSUE #276592: typescript language features do not resolve tsconfig extends specifier with package exports subpath remapping

- VS Code Version: any
- OS Version: any

Steps to Reproduce:

1. Extend a tsconfig from your (or another) package and/or project. The exact specifier does not matter, as long as it is correct (according to what is exported/exposed in that package):
    ```json
    {
      "//": "my tsconfig",
      "extends": "@scope/example/base/tsconfig.json"
    }
    ```
    and the dependency, in case you don't extend from within the same package:
    ```json
    {
      "//": "my package.json",
      "name": "@scope/my-package",
      "type": "module",
      "devDependencies": {
        "@scope/example": "*"
      }
    }
    ```
2. Remap export paths in the package/project to extend from. In theory this could be the package housing the project from step (1) itself, or another package in your mono repo, or a third party package. In practice this has been tested in a mono repo with one package/project/workspace depending on another package/project/workspace in the same mono repo:
    ```json
    {
      "//": "other package.json",
      "name": "@scope/example",
      "type": "module",
      "exports": {
        "./*/tsconfig.json": "./lib/*/tsconfig.json"
      }
    }
    ```
3. Observe that TypeScript correctly resolves the specifier from step (1). You can test this by running the type checker, or by trusting the VSCode problems panel not to show any problems about that config.
4. Observe that the VSCode TypeScript and JavaScript language features fail to resolve the config specifier. For example `cmd + click` to follow the link does *not* work.
5. Observe that using an (incorrect) specifier, which ignores the exports field of the dependency makes `cmd + click` work, but the type checker (correctly) is not able to resolve this specifier. In this example this could be something like the following, assuming there is a `tsconfig.json` in `./lib/base/` of the `@scope/example` package:
    ```json
    {
      "//": "my tsconfig",
      "extends": "@scope/example/lib/base/tsconfig.json" 
    }
    ```

Edit: Removed unnecessary steps. Added a repro repo instead: https://github.com/valler/repro-repo-for-ms-vscode-issue-276592

souhil25:
This PR updates the TypeScript language features to correctly resolve tsconfig.json extends fields that rely on package exports subpath remapping.

Problem:
Currently, when a tsconfig.json file extends another config using an exported subpath (e.g. "extends": "@scope/example/base/tsconfig.json"), VS Code fails to resolve the path, showing “Failed to resolve ... as module”.

Fix:
This change enhances getTsconfigPath in tsconfig.ts to handle exports-based resolution, aligning VS Code’s behavior with TypeScript’s module resolution logic.

Testing:
Verified with a repro project using a package with exports like:

"exports": {
  "./*/tsconfig.json": "./lib/*/tsconfig.json"
}

The extends link in the dependent project is now correctly resolved and opened in VS Code.

Fixes: #276592
Impact: Low (only affects tsconfig link resolution)


mjbvz:
Closing as this looks like a work in progress. Please resubmit once it's in a more complete state

PR #277047, opened 2025-11-13
Align tool names and add support for legacy reference identification

### LINKED ISSUE #272734: agent files: align tool names to names from custom agents specification
The [custom agents specification](https://github.com/github/sweagentd/blob/main/internal/customagent/SPEC.md) defines a set of tool names:

| Primary alias  | Purpose / Notes                                                                                          |
| ------------- | -------------------------------------------------------------------------------------------------------- |
| `shell`       | Execute a command in the appropriate shell for the operating system.                                     |
| `read`         | Read file contents.                                                                                      |
| `edit`        | Allow LLM to edit . Exact arguments can vary.                                                            |
| `search`      | Search for files or text in files.                                                                       |
| `custom-agent` | Allows a different custom agent to be invoked to accomplish a task. Claude Code uses "Task" for triggering subagents which is close to the same idea. |
| `web`          | Allows fetching content from URLs and performing a web search                                            |
| `todo`        | Creates and manages structured task lists |

#### Tool names for "out-of-box" MCP servers

The following MCP servers are available out-of-box for Copilot coding agent and can be referenced using namespacing.

| MCP server name              | Available tools                                                                  |
| ------------------------ | ------------------------------------------------------------------------------------------------ |
| `github`              | All [read-only tools](https://github.com/github/github-mcp-server?tab=readme-ov-file#read-only-mode) are available by default, but the token the server receives is scoped to the source repository. `github/*` includes all of them, or you can reference `github/<tool name>` where `<tool name>` is a value from the [MCP server documentation](https://github.com/github/github-mcp-server?tab=readme-ov-file#tools).                                                                      |
| `playwright` | All playwright tools are available by default, but the server runs is configured to only access localhost. `playwright/*` includes all of them, or you can reference `playwright/<tool name>` where `<tool name>` is a value from the [MCP server documentation](https://github.com/microsoft/playwright-mcp?tab=readme-ov-file#tools). By default the token it has access to is scoped to the source code repository. |

We want to align our tool names with these tool names. 
- show the new names in our tool picker
- existing agent and prompt file should continue to work, but we want to to guide users to update with squiggles and code actions  
- we want to group all tools that require vscode as environment under a tool set `vscode`

pwang347:
## Overview

The goal of this PR is to unify the tool names. As part of this change, we want to introduce a system that allows for deprecating old names but retaining backwards compatibility. Added as a contribution which will also allow extension authors similar capabilities.

### Out of scope
* Handling for nested toolsets
* Manual ordering of tools

### Updated picker
image

### GIF
tools

See related change: https://github.com/microsoft/vscode-copilot-chat/pull/1995

Fixes https://github.com/microsoft/vscode/issues/272734


src/vs/workbench/contrib/chat/browser/actions/chatToolPicker.ts
```diff
@@ -412,6 +419,10 @@ export async function showToolsPicker(
 			if (!tool.canBeReferencedInPrompt || !toolsEntries.has(tool)) {
```
pwang347, line 412: this flag `canBeReferencedInPrompt` is causing tools with this flag to be shown at top-level and in the toolset, not sure if we still need this block at all? @connor4312

src/vs/workbench/contrib/chat/common/languageModelToolsService.ts
```diff
@@ -301,9 +303,9 @@ export class ToolSet {
 	constructor(
 		readonly id: string,
 		readonly referenceName: string,
-		readonly icon: ThemeIcon,
+		public icon: ThemeIcon,
```
aeschli: keep it `readonly`

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/terminal.chatAgentTools.contribution.ts
```diff
@@ -69,16 +69,16 @@ class ChatAgentToolsContribution extends Disposable implements IWorkbenchContrib
 		const getTerminalOutputTool = instantiationService.createInstance(GetTerminalOutputTool);
 		this._register(toolsService.registerTool(GetTerminalOutputToolData, getTerminalOutputTool));
 
-		const runCommandsToolSet = this._register(toolsService.createToolSet(ToolDataSource.Internal, 'runCommands', VSCodeToolReference.runCommands, {
+		const shellToolSet = this._register(toolsService.createToolSet(ToolDataSource.Internal, 'shell', VSCodeToolReference.shell, {
```
aeschli, line 70: I think this is where we also need a legacy name for tool sets.
Old agent files will have a `runCommands` that now is `shell`

src/vs/workbench/contrib/chat/browser/actions/chatToolPicker.ts
aeschli: Maybe for now we keep things as they are. It's the duty of the tool definition to only set canBeReferencedInPrompt when top level.
So when moving it to a tool set, this property need to be removed.
Maybe in a follow up PR we should look who depends on this flag. Maybe we should deprecated it. There should not be any hidden tools as the user wants to be in control which tools are available.

src/vs/workbench/contrib/chat/browser/languageModelToolsService.ts
```diff
@@ -675,7 +675,16 @@ export class LanguageModelToolsService extends Disposable implements ILanguageMo
 				}
 			} else {
 				if (!result.has(tool)) { // already set via an enabled toolset
-					const enabled = toolOrToolSetNames.has(toolReferenceName) || toolOrToolSetNames.has(tool.toolReferenceName ?? tool.displayName);
+					let enabled = toolOrToolSetNames.has(toolReferenceName) || toolOrToolSetNames.has(tool.toolReferenceName ?? tool.displayName);
+					// Check legacy names
+					if (!enabled && tool.legacyToolReferenceFullNames) {
```
aeschli: `|| tool.legacyToolReferenceFullNames?.some(name =>  toolOrToolSetNames.has(name))`

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/terminal.chatAgentTools.contribution.ts
```diff
@@ -103,12 +103,15 @@ class ChatAgentToolsContribution extends Disposable implements IWorkbenchContrib
...
 		this._register(toolsService.registerTool(CreateAndRunTaskToolData, createAndRunTaskTool));
 
-		const runTasksToolSet = this._register(toolsService.createToolSet(ToolDataSource.Internal, 'runTasks', 'runTasks', {
-			description: localize('toolset.runTasks', 'Runs tasks and gets their output for your workspace'),
-		}));
-		runTasksToolSet.addTool(RunTaskToolData);
-		runTasksToolSet.addTool(GetTaskOutputToolData);
-		runTasksToolSet.addTool(CreateAndRunTaskToolData);
+		// Add to vscode toolset
+		const vscodeToolSet = this._register(toolsService.getOrCreateToolSet(
```
aeschli: Can we find a single place where we define that tool set?

src/vs/workbench/contrib/chat/common/tools/languageModelToolsContribution.ts
```diff
@@ -335,16 +344,32 @@ export class LanguageModelToolsExtensionPointHandler implements IWorkbenchContri
...
-
-					const obj = languageModelToolsService.createToolSet(
-						source,
-						toToolSetKey(extension.description.identifier, toolSet.name),
-						toolSet.referenceName ?? toolSet.name,
-						{ icon: toolSet.icon ? ThemeIcon.fromString(toolSet.icon) : undefined, description: toolSet.description }
-					);
+					const referenceName = toolSet.referenceName ?? toolSet.name;
+					const existingToolSet = languageModelToolsService.getToolSetByName(referenceName);
+					const mergeExisting = isBuiltinTool && existingToolSet?.source === ToolDataSource.Internal;
```
aeschli, line 369: Is merging only a thing for built-in tools?

aeschli:
Looks good! Should also have some tests. Check out https://github.com/microsoft/vscode/blob/37eac26c04f356f06474223d35e19706439bdbe4/src/vs/workbench/contrib/chat/test/browser/languageModelToolsService.test.ts#L766

src/vs/workbench/contrib/chat/common/languageModelToolsService.ts
pwang347: what are your thoughts on who should take precedence for the icon/description setting if both core/copilot-chat are registering to the same toolset? maybe we can just have core define the values and just pass in nothing from copilot-chat?

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/terminal.chatAgentTools.contribution.ts
pwang347: good point, yes I'll make that fix

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/terminal.chatAgentTools.contribution.ts
pwang347: what do you think about just defining the `vscode` toolset directly within the `toolsService` itself as a built-in? it will likely span various modules, while toolsets like `shell` can be expected to be more contained

edit: discussed offline, this is fine

src/vs/workbench/contrib/chat/common/tools/languageModelToolsContribution.ts
pwang347: yes, generally this is just to handle scenario where we contribute some tools from both core and copilot-chat extension to the same toolset (it's not ideal, but with the rename we would have such a scenario)

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/terminal.chatAgentTools.contribution.ts
pwang347: added `legacyNames` field

pwang347: Some specific scenarios to consider as part of this refactor:

### Simple cases
- `runCommands` -> `shell`

### Interesting cases (should we even do this?)
- `new` -> `vscode`
- `runNotebooks` -> `vscode`
- `runTasks` -> `vscode`

I think the following example is something that might arise from this sort of support (not sure if it's a pattern we intend to support):

```
Toolset A
- Tool 1
- Tool 2
```

Suppose we now deprecate Toolset A:

```
Toolset B
- Tool 1
- ...other tools

Toolset C
- Tool 2
- ...other tools
```

Should referencing `Toolset A` enable both `Toolset B` and `Toolset C`?

This is a practical example
image

src/vs/workbench/contrib/chat/common/tools/languageModelToolsContribution.ts
```diff
@@ -78,6 +79,14 @@ const languageModelToolsExtensionPoint = extensionsRegistry.ExtensionsRegistry.r
 					type: 'string',
 					pattern: '^[\\w-]+$'
 				},
+				legacyToolReferenceFullNames: {
```
roblourens, line 82: If this is only for our use, and not other extensions, then it should be gated by an API proposal, and the description should either mention that or we exclude it from the JSON schema

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/terminal.chatAgentTools.contribution.ts
roblourens: At one point we wanted to do a specific `notebooks` toolset for all notebook tools, because many users are not working with toolsets and may want to disable them, and I reverted it because we didn't have a way to maintain back-compat like we do now. Can we revisit that?

src/vs/workbench/contrib/chat/common/tools/languageModelToolsContribution.ts
roblourens: Which toolset do we do that for? IMO ideally we just create the toolset in one place like in the extension, and can list tools even if they are contributed by core, that seems valid to me.

roblourens (commented):
So with the legacy names, do we respect them but still show a warning?

src/vs/workbench/contrib/chat/common/tools/languageModelToolsContribution.ts
pwang347: @roblourens the current thinking is to include a new `vscode` toolset which would encompass any tools that are relevant only to the VS Code application (since we now share a single YAML format that should be mostly compatible with remote agents like GH cloud agents). I expect this one to have a large amount of contributions from both core and the chat extension.

pwang347: > IMO ideally we just create the toolset in one place like in the extension

Got it, so maybe the idea is we would declare these tools from core but de-dupe them (at least visually) if they are registered under a toolset from the chat extension (and I suppose we would disallow this for any other extension)?

src/vs/workbench/contrib/chat/common/tools/languageModelToolsContribution.ts
pwang347: @aeschli thoughts on allowing other extensions to use this? It seems like tool authors will likely want this capability as well for stability. Or, we can defer that part until later if we think this proposal needs more iterations.

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/terminal.chatAgentTools.contribution.ts
pwang347: I think a notebook toolset makes sense, though GitHub distributes the notebook functionality into the other toolsets:
image

What do you think?

src/vs/workbench/contrib/chat/common/tools/languageModelToolsContribution.ts
roblourens: I think any extension should be allowed to register toolsets that include tools that it doesn't own

src/vs/workbench/contrib/chat/common/tools/languageModelToolsContribution.ts
roblourens: Eh, ideally other extensions wouldn't be refactoring all their tool names all the time I guess

aeschli: I think we should allow extensions to use it too.

src/vs/workbench/contrib/chat/test/browser/promptSytntax/promptValidator.test.ts
```diff
@@ -67,6 +67,32 @@ suite('PromptValidator', () => {
...
+		const toolWithLegacy = { id: 'newTool', toolReferenceName: 'newToolRef', displayName: 'New Tool', canBeReferencedInPrompt: true, modelDescription: 'New Tool', source: ToolDataSource.External, inputSchema: {}, legacyToolReferenceFullNames: ['oldToolName', 'deprecatedToolName'] } satisfies IToolData;
+		disposables.add(toolService.registerToolData(toolWithLegacy));
+
+		const toolSetWithLegacy = disposables.add(toolService.createToolSet(
+			ToolDataSource.External,
+			'newToolSet',
+			'newToolSetRef',
+			{ description: 'New Tool Set', legacyNames: ['oldToolSet', 'deprecatedToolSet'] }
+		));
+		const toolInSet = { id: 'toolInSet', toolReferenceName: 'toolInSetRef', displayName: 'Tool In Set', canBeReferencedInPrompt: true, modelDescription: 'Tool In Set', source: ToolDataSource.External, inputSchema: {} } satisfies IToolData;
```
aeschli: Should that be `canBeReferencedInPrompt: false`?

src/vs/workbench/contrib/chat/test/browser/promptSytntax/promptValidator.test.ts
```diff
@@ -390,11 +507,9 @@ suite('PromptValidator', () => {
 			const messages = markers.map(m => m.message);
 			assert.deepStrictEqual(messages, [
 				'Attribute \'mcp-servers\' is ignored when running locally in VS Code.',
-				'Unknown tool \'shell\'.',
+				'Unknown tool \'edit\'.',
 			]);
-		});
-
-		test('undefined target with mcp-servers and github-tools', async () => {
+		}); test('undefined target with mcp-servers and github-tools', async () => {
```
aeschli: needs a new line
```suggestion
		});
		
		test('undefined target with mcp-servers and github-tools', async () => {
```

src/vs/workbench/contrib/extensions/browser/extensions.contribution.ts
```diff
@@ -2024,6 +2024,7 @@ class ExtensionToolsContribution extends Disposable implements IWorkbenchContrib
 		super();
 		const searchExtensionsTool = instantiationService.createInstance(SearchExtensionsTool);
 		this._register(toolsService.registerTool(SearchExtensionsToolData, searchExtensionsTool));
+		toolsService.vscodeToolSet.addTool(SearchExtensionsToolData);
```
aeschli: Returns a disposable. Needs a `this._register(` ?

src/vs/workbench/contrib/chat/common/languageModelToolsService.ts
pwang347: discussed offline, let's keep it as is (core contributes icons/descriptions and copilot-chat inherits this)

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/terminal.chatAgentTools.contribution.ts
pwang347: Discussed this with @aeschli. Will share in dev-discuss for more feedback from the team so we can chat more there - tentatively, because we have notebook reading in `read` and notebook editing in `edit` for alignment I don't think we'll add a new group for notebooks, and instead merge the remaining run notebook action into `vscode`.

src/vs/workbench/contrib/chat/common/tools/languageModelToolsContribution.ts
pwang347: discussed offline, we'll add an API proposal for now but follow-up to enable support for it for other extensions (out of caution we gate it now so we can think through a bit some of the extension namespacing pieces and to see if we have any issues with this change)

src/vs/workbench/contrib/chat/common/tools/languageModelToolsContribution.ts
pwang347: I think I'll leave this as is for now, as it is also an anti-pattern for the chat extension to know about specific core tool IDs

roblourens: I don't feel that way, I think all built-in tools are well-known names that can be referenced from an extension. It's not that different than referencing them from a prompt file or something IMO

roblourens (dismissed):
I don't know if you were in standup today, we were talking about the `chat.tools.eligibleForAutoApproval` setting, will the legacy reference names get applied to that?

pwang347:
> I don't know if you were in standup today, we were talking about the `chat.tools.eligibleForAutoApproval` setting, will the legacy reference names get applied to that?

Thanks @roblourens, I synced with @joshspicer about this and there will be some small work involved for that setting, but should be doable.

DonJayamanne (approved):

PR #277160, opened 2025-11-13
update distro

sandy081:


Tyriar (changes requested):
@sandy081 this doesn't update distro?

sandy081:
@Tyriar Thats right. I ran the `npm run update-distro` command and created PR w/o checking changes. Updated now. Thanks for checking

Tyriar (approved):

PR #277172, opened 2025-11-13
add dialog role and aria-labelledby to non-modal dialog widget

### LINKED ISSUE #277170: [Accessibility]: Screen reader do not announce when user is in the "Search workspace" dialog.
PROBLEM:
Screen readers are not giving any indication that the "Search workspace" dialog has been opened.

OBSERVED RESULTS:
When a user triggers the "Search workspace" dialog, there is no indication from screen readers that a dialog has been opened.

EXPECTED RESULTS:
Screen Reader should announce that the user is in a dialog as well as the accessible name of the dialog.

USER IMPACT:
Keyboard users will not be able to understand whether they are in a dialog for "Search workspace" or part of the entire web page.

STEPS TO REPRODUCE:
1. Open VSCode
2. Turn screen reader on
3. With the keyboard, press ENTER or SPACE while focused on "Search workspace" field at top of UI
4. Notice how there is no indication that the user is in a dialog

GIF of Issue: 

Image

ENVIRONMENT DETAILS:
Machine: Chromebook
Browser: Version 136.0.7103.150 (Official Build) (64-bit)
Accessibility Tool: ChromeVox.
Bug Filling Time: 20 mins.

Machine: MacBook Pro
Browser: Version 137.0.7151.104 (Official Build) (arm64)
Accessibility Tool: VoiceOver.
Bug Filling Time: 20 mins."

DBowen33:
Add role="dialog" and aria-labelledby attributes to quick-input-widget to let screen reader users know that they are in a dialog as well as giving the dialog an accessible name

Fixes #277170

Steps to Test:
1. Open VSCode
2. Turn screen reader on
3. With the keyboard, press ENTER or SPACE while focused on "Search workspace" field at top of UI
4. Notice when dialog opens, screen readers now announce that user is in a dialog along with the accessible name of the dialog.

### DIFF
--- a/src/vs/platform/quickinput/browser/quickInputController.ts
+++ b/src/vs/platform/quickinput/browser/quickInputController.ts
@@ -140,6 +140,8 @@ export class QuickInputController extends Disposable {
 		const container = dom.append(this._container, $('.quick-input-widget.show-file-icons'));
 		container.tabIndex = -1;
 		container.style.display = 'none';
+		container.setAttribute('role', 'dialog');
+		container.setAttribute('aria-labelledby', 'quick-input-title');
 
 		const styleSheet = domStylesheetsJs.createStyleSheet(container);
 
@@ -149,6 +151,7 @@ export class QuickInputController extends Disposable {
 		leftActionBar.domNode.classList.add('quick-input-left-action-bar');
 
 		const title = dom.append(titleBar, $('.quick-input-title'));
+		title.setAttribute('id', 'quick-input-title');
 
 		const rightActionBar = this._register(new ActionBar(titleBar, { hoverDelegate: this.options.hoverDelegate }));
 		rightActionBar.domNode.classList.add('quick-input-right-action-bar');


DBowen33:
Adding @a-stewart for review

src/vs/platform/quickinput/browser/quickInputController.ts
```diff
@@ -140,6 +140,8 @@ export class QuickInputController extends Disposable {
 		const container = dom.append(this._container, $('.quick-input-widget.show-file-icons'));
 		container.tabIndex = -1;
 		container.style.display = 'none';
+		container.setAttribute('role', 'dialog');
+		container.setAttribute('aria-labelledby', 'quick-input-title');
```
TylerLeonhardt, line 144: this is not so straightforward. The title is "Manage Trusted Extensions" in this case:
Image

but there is no title bar in this case:
Image

Can you tell me why this container needs `the aria-labelledby`? The true aria label depends on a lot of different factors.

DBowen33: The `aria-labelledby` attribute is to meaningfully name the dialog if there is a visible name/header so that it can relay it to the screen reader. It's best practice to have a visible label, but if there is not then `aria-label` should be used instead.

PR #277181, opened 2025-11-13
Don't automatically include contributed context value in implicit context

alexr00:
Part of #271104


src/vs/workbench/contrib/chat/common/chatContext.ts
```diff
@@ -22,6 +22,6 @@ export interface IChatContextSupport {
 
 export interface IChatContextProvider {
 	provideChatContext(options: {}, token: CancellationToken): Promise<IChatContextItem[]>;
-	provideChatContextForResource?(resource: URI, options: {}, token: CancellationToken): Promise<IChatContextItem | undefined>;
+	provideChatContextForResource?(resource: URI, withValue: boolean, token: CancellationToken): Promise<IChatContextItem | undefined>;
```
bpasero, line 25: Sorry for side tracking, but why is `chatContext.ts` in this folder? It seems only be used in chat context thus should be in `contrib/chat`?

The only reason chat entitlement service is here is because its used in non-chat environments.

bpasero (dismissed):

src/vs/workbench/contrib/chat/common/chatContext.ts
bpasero: Thanks!

alexr00: It was just an oversight, happy to move it!

justschen (approved):
awesome, thanks!

PR #277210, opened 2025-11-13
fix when terminal tabs show 

### LINKED ISSUE #277174: Terminal tabs view shows when there's only a single terminal
Image

Before, it only showed when there was more than one 

cc @Tyriar

meganrogge:
fixes #277174


Tyriar (commented):
lgtm, just need standup sign off

rzhao271 (approved):

PR #277225, opened 2025-11-13
fix: memory leak in terminal tabs list

SimonSiefke:
Fixes a memory leak in terminal tabs list. 

The main change is changing the TabsRenderer from a Disposable to a not-Disposable. Looking at other workbench lists, it seems list renderer is always a not-Disposable, and instead uses templateDisposables and elementDisposables to dispose elements.

### Before

When changing something like a task icon, the number of fillActionBar functions seems to grow each time

terminal-action-bar

### After

No more memory leak is detected:

```json
{
  "namedFunctionCount3": [],
  "isLeak": false
}
```


meganrogge (approved):
Thank you!

Tyriar (approved):

PR #277234, opened 2025-11-13
extract `ChatTerminalToolOutputSection`, fix a bug

### LINKED ISSUE #274322: Refactor ChatTerminalToolProgressPart to make simpler
https://github.com/microsoft/vscode/pull/273175 really ramped up how complex `ChatTerminalToolProgressPart` is, consider what we can pull out and simplify. For example the part that does the actual rendering of the output could be its own class such that it's not intermingled with the rest of the part.

### LINKED ISSUE #277131: LEAKED DISPOSABLE ChatTerminalToolProgressPart._createActionBar
Image

Image

### LINKED ISSUE #277453: Focus terminal/toggle output does not go to the correct element if a user has scrolled
The focus terminal/toggle output action should always apply to the last response element, even if you scroll up in the list.

Run several terminal requests

Scroll up in the list

Use the focus terminal or toggle output command via keybinding

🐛 it would not apply to the last one

meganrogge:
fixes #274322
fixes #277131
fixes #277453

Also addresses a bug that could happen where the the focus would move to an unexpected element's terminal or output - caused by the list re-rendering. We now keep track of where the part is in the list to fix this.


Tyriar (approved):
It's hard to review such a large change effectively, make sure it's tested well before merging as it looks risky

PR #277406, opened 2025-11-14
Cleanup microsoft#272425

Abrifq:
https://github.com/microsoft/vscode/pull/272425#discussion_r2527318019

- Removes redundant backtick replacement, as the check makes sure there is no backticks.
- Simplifies the wrapping on the assumption the user does not use triple backticks on their command.


Tyriar (approved):
I thought this would have stuck it on a new line, this is nicer 👍

TylerLeonhardt (approved):

PR #277498, opened 2025-11-14
tools: fix invocation token not round tripping with new sessions URIs

connor4312:
Caused the fetch tool (and possibly more tools) to fail to be called.


benvillalobos (approved):
Unblock approval

PR #277522, opened 2025-11-14
Allow approving terminal tool for session

### LINKED ISSUE #260819: Allow commands in session
This difference just occured to me. In other tools I can allow them in the session or the workspace. For terminal commands, it only offers to change my settings. Allowing a command just for a session would be nice. I sometimes would like to allow something riskier for a certain task but I know that I don't want it in my settings

Image

Image

I actually don't know whether "session" is "chat session" or "window session", I think I want "chat session" but it might not work that way

Tyriar:
Fixes #260819

@connor4312 what do you think? I considered pulling in the standard actions instead, but:

- I feel the standard wording "Allow in this Session" is too ambiguous
- We can't do workspace, adding workspace support would mean duplicating the rule suggestions and indicating user vs workspace (so out of scope imo)
- I also don't want to add "Always Allow" for similar reasons.

Something that's missing here is the ability to undo it. I could add that 

image

image

New command:

image

Clicking Disable is a bit ugly as it gives no feedback, but it does do what's advertised.

image


Tyriar:
There's an issue tracking that :+1:

roblourens (approved):

PR #277592, opened 2025-11-15
Disclaimer when `eligibleForAutoApproval` halts tool call

joshspicer:
ref https://github.com/microsoft/vscode-internalbacklog/issues/6255

image


joshspicer:
@connor4312 i think it's ok to add my `!disclaimer` here, but flagging in case you think otherwise

bpasero (approved):

PR #277601, opened 2025-11-15
Fix stdin position handling in diff mode

Amogh-2404:
## Summary

Fixes #260309

This PR preserves the position of the `-` argument when used with `--diff`, allowing users to control which side stdin appears on in the diff view.

## Problem

Previously, when using stdin with `--diff`, the position of `-` was ignored:
- `cat foo | code --diff - bar` → bar on left, stdin on right (wrong)
- `cat foo | code --diff bar -` → bar on left, stdin on right (same, wrong)

Both commands produced identical output because the `-` argument was filtered out and the stdin temp file was always appended to the end.

## Solution

Instead of filtering out `-` from `argv` and appending the temp file at the end, this PR replaces `-` in-place with the stdin temp file path. This preserves the original position:
- `cat foo | code --diff - bar` → stdin on left, bar on right ✓
- `cat foo | code --diff bar -` → bar on left, stdin on right ✓

## Changes

**File modified:** `src/vs/code/node/cli.ts`

1. Removed the line that filtered `-` from `argv` (line 261)
2. Added logic to replace `-` in-place with the stdin temp file path (lines 296-314)
3. Maintained backward compatibility by falling back to `addArg` if `-` is not found

## Testing

The fix has been verified to work correctly for:
- `--diff` with stdin on the left: `echo "test" | code --diff - file.txt`
- `--diff` with stdin on the right: `echo "test" | code --diff file.txt -`
- Backward compatibility: piping without explicit `-` still works

## Impact

- **Scope:** Single file, ~10 lines changed
- **Risk:** Minimal - only affects stdin handling with diff mode
- **Compatibility:** Backward compatible via fallback to existing behavior


bpasero:
Closing this in favour of the first PR: https://github.com/microsoft/vscode/pull/265098

PR #277759, opened 2025-11-16
remove headers and replace with checkmark for now

### LINKED ISSUE #271883: Chat response: certain actions make my avatar appear as request rendering
Notice how suddenly my name and photo appears when I retry a chat request:

Image

This is quite inconsistent with the rest of the UI where neither my handle nor photo appear.

//cc @isidorn @mrleemurray

justschen:
fix https://github.com/microsoft/vscode/issues/271883

Screenshot 2025-11-16 at 3 53 03 PM

I don't want to mess too much with the margins since those are set for the entire monaco list row,  but if it bothers folks, we can reduce them for this scenario specifically perhaps?

### DIFF
--- a/src/vs/workbench/contrib/chat/browser/chatListRenderer.ts
+++ b/src/vs/workbench/contrib/chat/browser/chatListRenderer.ts
@@ -693,8 +693,10 @@ export class ChatListItemRenderer extends Disposable implements ITreeRenderer<Ch
 	private renderConfirmationAction(element: IChatRequestViewModel, templateData: IChatListItemTemplate) {
 		dom.clearNode(templateData.detail);
 		if (element.confirmation) {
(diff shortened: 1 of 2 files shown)


bpasero (approved):

roblourens (commented):
Nice, @lramos15 was looking at this too

lramos15:
> Nice, @lramos15 was looking at this too

I notice this PR doesn't get rid of the backing data objects that powered this view before. See https://github.com/microsoft/vscode/blob/56328c6a54e6a6e6b6f459e677faf92b57a78a65/src/vs/workbench/contrib/chat/common/chatViewModel.ts#L400-L406

Should we clean these up too if nobody will use the info. cc @justschen @roblourens

PR #277869, opened 2025-11-17
add command decoration to chat terminal 

### LINKED ISSUE #272575: Surface exit code, start time and duration in chat terminal tool UI
tl;dr: Indicator for how long a terminal command has been waiting (only show after a bit of time has passed).

With more agent sessions, I am now coming back to agent sessions that are showing running terminal commands:

Image

(In this case, the tests were reliably stuck)

Is it stuck, has it just started, no idea.

cc @Tyriar

meganrogge:
fixes #272575

https://github.com/user-attachments/assets/08d94990-a1c1-4d55-9932-4d83505eca69

### DIFF
--- a/src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
+++ b/src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
@@ -6,6 +6,7 @@
 import { h } from '../../../../../../base/browser/dom.js';
 import { ActionBar } from '../../../../../../base/browser/ui/actionbar/actionbar.js';
 import { Codicon } from '../../../../../../base/common/codicons.js';
+import { getDurationString } from '../../../../../../base/common/date.js';
 import { KeyCode, KeyMod } from '../../../../../../base/common/keyCodes.js';
 import { isMarkdownString, MarkdownString } from '../../../../../../base/common/htmlContent.js';
 import { IInstantiationService, ServicesAccessor } from '../../../../../../platform/instantiation/common/instantiation.js';
@@ -31,6 +32,8 @@ import { Action, IAction } from '../../../../../../base/common/actions.js';
 import { Disposable, MutableDisposable, toDisposable, type IDisposable } from '../../../../../../base/common/lifecycle.js';
 import { Emitter, Event } from '../../../../../../base/common/event.js';
 import { ThemeIcon } from '../../../../../../base/common/themables.js';
+import { terminalDecorationError, terminalDecorationIncomplete, terminalDecorationSuccess } from '../../../../terminal/browser/terminalIcons.js';
+import { DecorationSelector } from '../../../../terminal/browser/xterm/decorationStyles.js';
 import * as dom from '../../../../../../base/browser/dom.js';
 import { DomScrollableElement } from '../../../../../../base/browser/ui/scrollbar/scrollableElement.js';
 import { ScrollbarVisibility } from '../../../../../../base/common/scrollable.js';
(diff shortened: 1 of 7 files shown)


src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
```diff
@@ -67,6 +70,200 @@ const sanitizerConfig = Object.freeze<DomSanitizerConfig>({
...
+		super();
+		this._register(toDisposable(() => this._stopTimer()));
+	}
+
+	public ensureElement(context?: HTMLElement): HTMLElement | undefined {
+		const title = context ?? this._options.getTitleElement();
+		if (!title) {
+			return undefined;
+		}
+		// eslint-disable-next-line no-restricted-syntax
```
meganrogge: Is there a prefeerred way to do this?

src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
```diff
@@ -67,6 +70,200 @@ const sanitizerConfig = Object.freeze<DomSanitizerConfig>({
...
+		const decoration = this.ensureElement();
+		if (!decoration) {
+			this._stopTimer();
+			return;
+		}
+		const resolvedCommand = command ?? this._options.getResolvedCommand();
+		this._apply(decoration, resolvedCommand);
+		const storedState = this._options.terminalData.terminalCommandState;
+		if (resolvedCommand?.exitCode === undefined || (!resolvedCommand && storedState && storedState.exitCode === undefined)) {
+			this._startTimer();
```
meganrogge: Considered keeping track of duration on the `ITerminalCommand` instead. LMK if that's preferable.

src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
```diff
@@ -67,6 +70,200 @@ const sanitizerConfig = Object.freeze<DomSanitizerConfig>({
...
 const expandedStateByInvocation = new WeakMap<IChatToolInvocation | IChatToolInvocationSerialized, boolean>();
 
+interface ITerminalCommandDecorationOptions {
+	readonly terminalData: IChatTerminalToolInvocationData;
+	getTitleElement(): HTMLElement | undefined;
+	getHost(): HTMLElement | undefined;
+	getResolvedCommand(): ITerminalCommand | undefined;
+}
+
+class TerminalCommandDecoration extends Disposable {
```
Tyriar, line 102: Can we move it to the right like the mock ups?

Image

src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
```diff
@@ -67,6 +70,200 @@ const sanitizerConfig = Object.freeze<DomSanitizerConfig>({
...
+		} else if (status === 'success') {
+			icon = terminalDecorationSuccess;
+			decoration.classList.add('success');
+		} else {
+			icon = terminalDecorationIncomplete;
+			decoration.classList.add(DecorationSelector.DefaultColor, DecorationSelector.Default);
+		}
+
+		decoration.classList.add(...ThemeIcon.asClassNameArray(icon));
+		const hoverMessage = localize('chatTerminalCommandDecoration.hover', 'Exit Code: {0}\nStart: {1}\nDuration: {2}', exitCodeText, startText, durationText);
```
Tyriar: Can we share the code/message of the one in the terminal view? 

Image

It looks like this change would be a lot simplifier if we shared this code.

src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
Tyriar: We shouldn't be using `querySelector` anymore, the idea is you should construct elements you need using the `h` helper. If this element comes from the markdown renderer and is not accessible, it means we shouldn't be touching it as it's internal to that component.

Tyriar: For this specifically, it looks like the pattern with chat parts is to expose a `domNode` and allow the embedder of the part to control where it goes.

src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
```diff
@@ -67,6 +70,200 @@ const sanitizerConfig = Object.freeze<DomSanitizerConfig>({
...
+			container.insertBefore(wrapper, markdown);
+			wrapper.appendChild(markdown);
+			container = wrapper;
+		}
+		// eslint-disable-next-line no-restricted-syntax
+		let decoration = container.querySelector<HTMLElement>(':scope > .chat-terminal-command-decoration');
+		if (!decoration) {
+			decoration = document.createElement('span');
+			decoration.classList.add('chat-terminal-command-decoration');
+			decoration.setAttribute('role', 'img');
```
Tyriar: Shouldn't this have an aria-label and be a button instead?

src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
```diff
@@ -67,6 +70,200 @@ const sanitizerConfig = Object.freeze<DomSanitizerConfig>({
...
+		this._element = decoration;
+		return decoration;
+	}
+
+	public update(command?: ITerminalCommand): void {
+		const decoration = this.ensureElement();
+		if (!decoration) {
+			this._stopTimer();
+			return;
+		}
```
Tyriar: If the element is created in the ctor as `domNode` we don't need the `ensureElement` check as it will always be there.

src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
```diff
@@ -67,6 +70,200 @@ const sanitizerConfig = Object.freeze<DomSanitizerConfig>({
...
+				this._stopTimer();
+				return;
+			}
+			const command = this._options.getResolvedCommand();
+			this._apply(element, command);
+			const storedState = this._options.terminalData.terminalCommandState;
+			if ((!command && (!storedState || storedState.exitCode !== undefined)) || (command && command.exitCode !== undefined)) {
+				this._stopTimer();
+			}
+		}, 1000);
```
Tyriar: This timer is to check for whether the command is ready? This should be event based instead

Tyriar (dismissed):

src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
meganrogge: It's also to update the duration

Tyriar: We only need it on hover, so it could live in the hover content callback?

src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
meganrogge: it doesn't have actions, so shouldn't be a button. it has content, and is mostly a visual feature, so treating it as an image seems right.

benibenj (approved):

PR #277920, opened 2025-11-17
Fix issue where local terminal failed to open in remote scenarios

### LINKED ISSUE #148572: Local terminal can't be opened: \bin\bash not found (though VS Code running on Windows)
- VSCode Version: 1.66.1
- Local OS Version: Windows 10
- Remote OS Version: Linux
- Remote Extension/Connection Type: Docker/WSL2 running on host
- Logs: error message: ```The terminal process failed to launch: Path to shell executable "\bin\bash" does not exist.```

Steps to Reproduce:

1. VS Code command: ```> Terminal: Create New Integrated Terminal (Local)```
2. The error popup appears (see above "Logs")

Does this issue occur when you try this locally?: N/A
Does this issue occur when you try this locally and all extensions are disabled?: Yes

Seems like microsoft/vscode-remote-release#730 is broken.
I checked with ```@modified terminal``` if there are any changes. Only inside the container the following settings are set:
```terminal.integrated.profiles.linux``` and ```terminal.integrated.defaultProfile.linux```.
Shouldn't it run cmd or powershell locally?

### PR DESCRIPTION
**Pull request by anthonykim1**
Resolves: https://github.com/microsoft/vscode/issues/148572 
This would allow default shell to launch on non-wsl remote/local scenario as well.

### CODE REVIEW

**Comment by anthonykim1** [comment]
This would prevent the aggressive error message and immediate close of local terminal, but it won't respect local terminal profile setting. It would just open powershell for windows, regardless of if I had `  "terminal.integrated.defaultProfile.windows": "Git bash",` for local windows settings in wsl scenario.

I think we need something like https://github.com/microsoft/vscode/pull/277968 to make sure profiles are picked up properly as well.

**Review by Tyriar — DISMISSED** [review]

File: src/vs/workbench/contrib/terminal/browser/terminalService.ts [hunk 1]
```diff
@@ -968,9 +968,12 @@ export class TerminalService extends Disposable implements ITerminalService {
 		// local terminal in a remote workspace as profile won't be used in those cases and these
 		// terminals need to be launched before remote connections are established.
 		const isLocalInRemoteTerminal = this._remoteAgentService.getConnection() && URI.isUri(options?.cwd) && options?.cwd.scheme === Schemas.vscodeFileResource;
+		this._logService.debug(`#terminalService#createTerminal: Value of isLocalInRemoteTerminal is: ${isLocalInRemoteTerminal}`);
```
anthonykim1: This is coming out to be false when we are trying to launch local terminal from remote hmm..

src/vs/workbench/contrib/terminal/browser/terminalService.ts
```diff
@@ -982,20 +985,25 @@ export class TerminalService extends Disposable implements ITerminalService {
 		}
 
 		let config = options?.config;
+		this._logService.debug(`#terminalService#createTerminal: config ${config}`);
 		if (!config && isLocalInRemoteTerminal) {
+			this._logService.debug(`#terminalService#createTerminal: !config and isLocalInRemoteTerminal`);
 			const backend = await this._terminalInstanceService.getBackend(undefined);
```
anthonykim1, line 994: Because of https://github.com/microsoft/vscode/pull/277920/commits/1b762fc6ca5743b863267aa24fdbf9d2cdcff0af#r2540901561 Code never reaches here actually.

Tyriar (approved):

Issue #251844, opened 2025-06-18
Add "Admin Contact Info" policy


joshspicer:
Provide a way to encode contact information (url/email) where the admin configuring this policy can be reached. 

There is precedent for this in other MDM solutions, for example [tailscale](https://tailscale.com/kb/1315/mdm-keys#set-an-info-message).

This information is helpful for https://github.com/microsoft/vscode/issues/239798#issuecomment-2960782214, where you'd ideally want such an alert to provide actionable information.

cc/ @sandy081

digitarald:
For GH-controlled settings, we could point them to the org page that the policy comes from.

joshspicer:
We don't currently internally track _who_ is applying policy, but that is possible. There are really only two options - your IT department that owns the hardware you're on (Microsoft IT) and the organization admin that controls your GitHub copilot seat.

This issue was for the former - I imagined another 'policy' (in addition to [these](https://code.visualstudio.com/docs/setup/enterprise#_centrally-manage-vs-code-settings)) that is `AdminContactEmail`, `AdminContactUrl`, etc...

I suppose it would also be helpful for GitHub to send us a org contact email that we could surface?

Other products like tailscale do this https://tailscale.com/kb/1315/mdm-keys#set-your-organization-name

PR #277943, opened 2025-11-17
release: eligibleForAutoApproval

joshspicer:
closes https://github.com/microsoft/vscode-internalbacklog/issues/6255

### DIFF
--- a/src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/abstractToolConfirmationSubPart.ts
+++ b/src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/abstractToolConfirmationSubPart.ts
@@ -60,6 +60,7 @@ export abstract class AbstractToolConfirmationSubPart extends BaseChatToolInvoca
 		const skipTooltip = skipKeybinding ? `${config.skipLabel} (${skipKeybinding})` : config.skipLabel;
 
 
+		const additionalActions = this.additionalPrimaryActions();
 		const buttons: IChatConfirmationButton<(() => void)>[] = [
 			{
 				label: config.allowLabel,
@@ -67,7 +68,7 @@ export abstract class AbstractToolConfirmationSubPart extends BaseChatToolInvoca
 				data: () => {
 					this.confirmWith(toolInvocation, { type: ToolConfirmKind.UserAction });
 				},
-				moreActions: this.additionalPrimaryActions(),
+				moreActions: additionalActions.length > 0 ? additionalActions : undefined,
 			},
 			{
 				label: localize('skip', "Skip"),
--- a/src/vs/workbench/contrib/chat/browser/languageModelToolsService.ts
+++ b/src/vs/workbench/contrib/chat/browser/languageModelToolsService.ts
@@ -445,9 +445,26 @@ export class LanguageModelToolsService extends Disposable implements ILanguageMo
 			prepared = await preparePromise;
 		}
 
+		// TODO: If the user has _previously_ auto-approved this tool, I don't think we make it to this check.
+		const isEligibleForAutoApproval = this.isToolEligibleForAutoApproval(tool.data);
+
+		// Default confirmation messages if tool is not eligible for auto-approval
+		if (!isEligibleForAutoApproval && !prepared?.confirmationMessages?.title) {
+			if (!prepared) {
+				prepared = {};
+			}
+			const toolReferenceName = getToolReferenceName(tool.data);
+			// TODO: This should be more detailed per tool.
+			prepared.confirmationMessages = {
+				title: localize('defaultToolConfirmation.title', 'Allow tool to execute?'),
+				message: localize('defaultToolConfirmation.message', 'Run the "{0}" tool.', toolReferenceName),
+				allowAutoConfirm: false,
+			};
+		}
+
 		if (prepared?.confirmationMessages?.title) {
 			if (prepared.toolSpecificData?.kind !== 'terminal' && typeof prepared.confirmationMessages.allowAutoConfirm !== 'boolean') {
-				prepared.confirmationMessages.allowAutoConfirm = true;
+				prepared.confirmationMessages.allowAutoConfirm = isEligibleForAutoApproval;
 			}
 
 			if (!prepared.toolSpecificData && tool.data.alwaysDisplayInputOutput) {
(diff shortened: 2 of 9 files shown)


src/vs/workbench/contrib/chat/browser/languageModelToolsService.ts
```diff
@@ -445,9 +445,30 @@ export class LanguageModelToolsService extends Disposable implements ILanguageMo
...
+			if (!prepared) {
+				prepared = {};
+			}
+			const toolReferenceName = getToolReferenceName(tool.data);
+			// TODO: This should be more detailed per tool.
+			prepared.confirmationMessages = {
+				title: localize('defaultToolConfirmation.title', 'Allow tool to execute?'),
+				message: localize('defaultToolConfirmation.message', 'Run the \'{0}\' tool?', toolReferenceName),
+				disclaimer: localize('defaultToolConfirmation.disclaimer', 'Auto approval for \'{0}\' is restricted by \'{1}\'.', toolReferenceName, ChatConfiguration.EligibleForAutoApproval),
+				allowAutoConfirm: false,
```
connor4312, line 463: If a tool is marked ineligible for auto approval, but if it would otherwise only require post-approval (e.g. an MCP tool with readOnlyHint and openWorldHint) this overwrites the potential `confirmResults?: boolean` on the confirmation messages object and so post-approvel would not happen

src/vs/workbench/contrib/chat/browser/languageModelToolsService.ts
```diff
@@ -445,9 +445,30 @@ export class LanguageModelToolsService extends Disposable implements ILanguageMo
...
+			};
+		}
+
+		if (!isEligibleForAutoApproval && prepared?.confirmationMessages?.title && !prepared.confirmationMessages.disclaimer) {
+			prepared.confirmationMessages.disclaimer = localize('defaultToolConfirmation.disclaimer', 'Auto approval for \'{0}\' is restricted by \'{1}\'.', getToolReferenceName(tool.data), ChatConfiguration.EligibleForAutoApproval);
```
connor4312: I'd append this unconditionally if `!isEligibleForAutoApproval && prepared?.confirmationMessages?.title`. The disclaimer we show for MCP's is very generic

src/vs/workbench/contrib/chat/browser/languageModelToolsService.ts
joshspicer: thanks, makes sense. Fixed in https://github.com/microsoft/vscode/pull/277943/commits/023d12f30e76a2f89635ad96c19b6d7a4c25c66b

src/vs/workbench/contrib/chat/browser/languageModelToolsService.ts
joshspicer: That makes sense.  I cannot think of an example where this disclaimer is not the 'most important' one if the setting is enabled.

I am wary to change it for the candidate and may just take it in main?

src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/abstractToolConfirmationSubPart.ts
```diff
@@ -60,14 +60,15 @@ export abstract class AbstractToolConfirmationSubPart extends BaseChatToolInvoca
...
+		const additionalActions = this.additionalPrimaryActions();
 		const buttons: IChatConfirmationButton<(() => void)>[] = [
 			{
 				label: config.allowLabel,
 				tooltip: allowTooltip,
 				data: () => {
 					this.confirmWith(toolInvocation, { type: ToolConfirmKind.UserAction });
 				},
-				moreActions: this.additionalPrimaryActions(),
+				moreActions: additionalActions.length > 0 ? additionalActions : undefined,
```
Tyriar, line 71: Ideally this wouldn't be necessary, the dropdown code should know that an array of length 0 is the same as undefined.

src/vs/workbench/contrib/chat/browser/languageModelToolsService.ts
```diff
@@ -445,9 +445,31 @@ export class LanguageModelToolsService extends Disposable implements ILanguageMo
...
+				...prepared.confirmationMessages,
+				title: localize('defaultToolConfirmation.title', 'Allow tool to execute?'),
+				message: localize('defaultToolConfirmation.message', 'Run the \'{0}\' tool?', toolReferenceName),
+				disclaimer: localize('defaultToolConfirmation.disclaimer', 'Auto approval for \'{0}\' is restricted by \'{1}\'.', toolReferenceName, ChatConfiguration.EligibleForAutoApproval),
+				allowAutoConfirm: false,
+			};
+		}
+
+		if (!isEligibleForAutoApproval && prepared?.confirmationMessages?.title && !prepared.confirmationMessages.disclaimer) {
+			prepared.confirmationMessages.disclaimer = localize('defaultToolConfirmation.disclaimer', 'Auto approval for \'{0}\' is restricted by \'{1}\'.', getToolReferenceName(tool.data), ChatConfiguration.EligibleForAutoApproval);
```
Tyriar: Calling out the policy/administrator here would be good if we know that setting is enforced via policy, rather than just the setting name.

Tyriar (commented):
Took a quick look and have some small comments, don't need to handle these in the release branch though. Defer to @connor4312 for final approval

src/vs/workbench/contrib/chat/browser/languageModelToolsService.ts
joshspicer: I agree. I didn't do that _just because_ you could set this yourself and we don't have a great way to determine if the setting was set via policy.

Even better, with https://github.com/microsoft/vscode/issues/251844 we could provide actionable info on who to contact for information

connor4312 (dismissed):
code lgtm

src/vs/workbench/contrib/chat/browser/languageModelToolsService.ts
joshspicer: i've changed this since fetch will show a misleading message. Good call out.

src/vs/workbench/contrib/chat/browser/languageModelToolsService.ts
joshspicer: Link to the setting itself, which will indicate 'managed by organization'. Moving forward I would love to make the admin info more visible.

connor4312 (approved):

Issue #277318, opened 2025-11-14
Improve ChatModel lifecycle


roblourens:
This issue captures some work in the area of managing ChatModels so that we're set for everything we want to do with managing agent sessions. The primary task is

- [x] ChatModel can live independently of a ChatWidget (https://github.com/microsoft/vscode/issues/246134)

and in order to do it right, here are some other improvements we can do
- [x] Support creating sessions fully in the background without first opening an editor
  - This should work but no scenario triggers this right now
- [x] ChatModel should have a clear observable `state` to let consumers watch for it to be waiting on a confirmation, without having to watch the `value` of a particular response. https://github.com/microsoft/vscode/pull/278171
  - [x] And a timestamp to determine how long the current response has been running https://github.com/microsoft/vscode/pull/278449
- [x] Try to clean up how the ChatModel/editing session are initialized https://github.com/microsoft/vscode/pull/277924 https://github.com/microsoft/vscode/pull/277944
  - This is too complicated: https://github.com/microsoft/vscode/blob/b59493ac03bf9646a6ec17b88c758a93b14bd5e3/src/vs/workbench/contrib/chat/common/chatModel.ts#L1493-L1503
  - It makes chat session init async when it doesn't need to be
  - And led to this hack, which is something that makes it hard to work with ChatWidgets: https://github.com/microsoft/vscode/blob/b59493ac03bf9646a6ec17b88c758a93b14bd5e3/src/vs/workbench/contrib/chat/browser/chatWidget.ts#L775
- [x] Have really solid interfaces for working with chat sessions and chat widgets
	- eg, I should be able to say "open this chat session URI" and it "just works" without having to check whether it's already open
	- After a lot of work to fix this, we apparently regressed that the agent sessions view can open multiple copies of the same chat session. Then when you close one, everything breaks.
	- This code https://github.com/microsoft/vscode/blob/bfe4e1aa7e22ccbf667d0e8ac5dad9224c89057b/src/vs/workbench/contrib/chat/browser/chatSessions/localChatSessionsProvider.ts#L25 should get chat sessions in a much simpler way
	- Register a custom opener so that `IOpenerService#open` can be used to reveal/open chat sessions easily
- [x] Fully adopt `IChatWidgetService#openSession` in actions that open chats @roblourens https://github.com/microsoft/vscode/pull/278973
- [ ] Support moving open chats via `IChatWidgetService#openSession` @roblourens 
- [ ] Fix bugs around having multiple chats active
[...]

PR #277944, opened 2025-11-17
chat: remove `waitForReady`

### LINKED ISSUE #247484: Debt- clean up async ChatModel/IChatEditingSession initialization
The ChatModel and editing session intialization got complicated after unifying the chat views and loading sessions from disk, async, instead of the storage service, sync.

Since the editing session setup blocks on the ChatModel being ready in the ChatService, the ChatModel creation can't block on the editing session being ready.

Workarounds led to https://github.com/microsoft/vscode-copilot/issues/16038 and https://github.com/microsoft/vscode-copilot/issues/16060, now we have a better workaround, but the whole thing needs to be reworked so that the model and editing session setup are more aligned.

connor4312:
- Now that edit sessions are sync, we don't need to wait for them (just
  guard edits) as in https://github.com/microsoft/vscode-copilot/issues/16060
- I noticed `widget.onDidClear()` listeners were actually async and we
  do need to wait for them before we can safely make new input. Revised
  this to a delegate pattern.
- Wait for a view model when sending chat. Also, improved Event.toPromise
  so it can safely be reused -- it now cleans up its own disposable from
  the disposable store or array after it settles.

Resolves #247484
Refs https://github.com/microsoft/vscode/issues/277318

### DIFF
--- a/src/vs/workbench/contrib/chat/browser/actions/chatActions.ts
+++ b/src/vs/workbench/contrib/chat/browser/actions/chatActions.ts
@@ -325,7 +325,9 @@ abstract class OpenChatGlobalAction extends Action2 {
 			chatWidget.setInput(opts.query);
 
 			if (!opts.isPartialQuery) {
-				await chatWidget.waitForReady();
+				if (!chatWidget.viewModel) {
+					await Event.toPromise(chatWidget.onDidChangeViewModel);
+				}
 				await waitForDefaultAgent(chatAgentService, chatWidget.input.currentModeKind);
 				resp = chatWidget.acceptInput();
 			}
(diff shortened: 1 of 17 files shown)


src/vs/workbench/contrib/chat/browser/actions/chatActions.ts
```diff
@@ -325,7 +325,9 @@ abstract class OpenChatGlobalAction extends Action2 {
 			chatWidget.setInput(opts.query);
 
 			if (!opts.isPartialQuery) {
-				await chatWidget.waitForReady();
+				if (!chatWidget.viewModel) {
```
connor4312, line 328: This is the case where we do need to wait still, because we might change agents I believe

roblourens (approved):

PR #277971, opened 2025-11-18
chat: implement and migrate to IChatWidgetService.reveal/revealWidget

connor4312:
Previously we did a couple of different ad-hoc things when showing chat,
which often didn't support editor chat, aux window chat, and never
supported quick chat at all.

This implement a method on the IChatWidgetService to reveal the chat
widget regardless of location, and also to reveal+open the last widget
or create a new one if there wasn't one, replacing the popular `showChatView`
which only supported sidebar chat.

Refs #277318

cc @roblourens

### DIFF
--- a/src/vs/workbench/contrib/chat/browser/chatWidget.ts
+++ b/src/vs/workbench/contrib/chat/browser/chatWidget.ts
@@ -10,7 +10,7 @@ import { IHoverOptions } from '../../../../base/browser/ui/hover/hover.js';
 import { HoverPosition } from '../../../../base/browser/ui/hover/hoverWidget.js';
 import { IListRenderer, IListVirtualDelegate } from '../../../../base/browser/ui/list/list.js';
 import { ITreeContextMenuEvent, ITreeElement } from '../../../../base/browser/ui/tree/tree.js';
-import { disposableTimeout, RunOnceScheduler, timeout } from '../../../../base/common/async.js';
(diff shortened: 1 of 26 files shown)


src/vs/workbench/contrib/chat/browser/chatWidget.ts
```diff
@@ -2882,6 +2893,58 @@ export class ChatWidgetService extends Disposable implements IChatWidgetService
...
+		if (widget.viewModel?.sessionResource) {
+			for (const group of this.editorGroupsService.getGroups(GroupsOrder.MOST_RECENTLY_ACTIVE)) {
+				const editor = group.findEditors(widget.viewModel?.sessionResource).at(0);
+				if (!editor) {
+					continue;
+				}
+
+				// focus transfer to other documents is async. If we depend on the focus
+				// being synchronously transferred in consuming code, this can fail, so
+				// wait for it to propagate
```
roblourens, line 2915: This is a little confusing, was there somewhere that we depend on this? And is this just async when switching to a new aux window, or always async?

roblourens (approved):

src/vs/workbench/contrib/chat/browser/chatWidget.ts
connor4312: It's only async when switching to an aux window. This was specially with the "open model picker" command where the context view service looks for the active document to determine where to open the picker. If I don't wait, there is actually no active document and so it falls back to the main window. There just seems to be some asyncness in chrome/electron around the focus transfer.

PR #277972, opened 2025-11-18
Update `@types/vscode` package.json too

jakebailey:
Per: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/74099#discussion_r2534961302

This updates the script to also update the version number in `package.json`, so that it does not need to be done manually.

cc @mjbvz


mjbvz (approved):
Thanks!

Tyriar (approved):

PR #277979, opened 2025-11-18
chat thinking content part cleanup

justschen:
lots of cleanup:
1. use the chat-used-context/collapsible part piece
2. move thinking content part out of `chat.css`


dmitrivMS (approved):
Less code is better code!

justschen:
@dmitrivMS indeed!

a few of the copilot commentsa are actually good though, will address those in a second PR later tonight 😴

PR #278056, opened 2025-11-18
fix: properly compute notebook element position

### LINKED ISSUE #278055: Wrong notebook layout position when the workbench is not positioned on (0, 0)

- VS Code Version: VSCode web 1.106.1
- OS Version: Ubuntu 24.04

Steps to Reproduce:

1. Go to vscode.dev
2. Open the browser devtools, edit the `body` element to add a `padding-top` of 50px to simulate a header
3. Run the `Create: New Jupyter Notebook` command from the command palette
4. The notebook is not correctly positioned:

Image

5. Resize the page or any element to force a layout
6. The notebook is now correctly positioned:

Image

Obviously the bug doesn't impact official builds of VSCode because I guess the container always takes the whole body, but it impacts forks that are adding a header

CGNonofr:
fix #278055

When a position is given to the updateShadowElement function, it's already an absolute position, so we shouldn't substract the parentElement top from it


dmitrivMS:
The notebook positioning issue was fixed by merged PR [#310070](https://github.com/microsoft/vscode/pull/310070).

PR #278100, opened 2025-11-18
build/npm: add explicit npm version check to preinstall.js

mithilP007:
Adds explicit npm version validation to build/npm/preinstall.js, mirroring the existing Node.js version check.

- Requires npm v10.5.0 or later for development
- Reads npm version from npm_config_user_agent environment variable
- Uses semver.coerce() and semver.lt() for version comparison
- Provides clear error message when older npm version is detected

Fixes #252372


dmitrivMS:
We already have npm version check in place.

PR #278165, opened 2025-11-18
prevent execute strategy from hanging

### LINKED ISSUE #278002: Terminal hangs when running `echo` command
- Copilot Chat Extension Version: 0.34.2025111702
- VS Code Version: af29d352965dd3b3ed8f3cf46e9f55b896533fcc
- OS Version: macOS
- Feature (e.g. agent/edit/ask mode): run_in_terminal
- Selected model (e.g. GPT 4.1, Claude 3.7 Sonnet): N/A
- Logs: N/A

https://github.com/user-attachments/assets/18ae9bf8-0c56-4d0c-a8ad-902300498287

Steps to Reproduce:

1. Open a new chat session and prompt the agent with:
   ```
   #runInTerminal run `echo "(hi!)"`
   ```
2. Click *Allow*.
3. Observe the terminal running command hangs indefinitely.

When you pop open the hidden chat it shows it's complete, but the Chat Window still thinks the command is running.

NB: The example I've given is a minimal repro, but this organically happened in a more complex session where the agent was running some python and bash.

meganrogge:
fixes #278002

@digitarald and others have said that sometimes chat will hang when the agent runs something in the terminal. Ross found a case that I could reproduce 🎉 . I found that when we do not receive the `commandExecuted` sequence, the agent will hang.

This prevents that by considering the command finished if the start sequence comes in and the command executed or finished does not occur within a second of that.

Before, hanging indefinitely:

https://github.com/user-attachments/assets/65b5d3bb-03ae-4454-bf23-62e8ffc26fbc

After:

Screenshot 2025-11-18 at 12 44 29 PM

### DIFF
--- a/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/executeStrategy/executeStrategy.ts
+++ b/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/executeStrategy/executeStrategy.ts
@@ -165,13 +165,14 @@ export async function trackIdleOnPrompt(
 	const scheduler = store.add(new RunOnceScheduler(() => {
 		idleOnPrompt.complete();
 	}, idleDurationMs));
+	let state: TerminalState = TerminalState.Initial;
+
 	// Fallback in case prompt sequences are not seen but the terminal goes idle.
 	const promptFallbackScheduler = store.add(new RunOnceScheduler(() => {
-		if (hasSeenExecuteSequenceSinceLastPrompt) {
+		if (state === TerminalState.Executing || state === TerminalState.PromptAfterExecuting) {
 			promptFallbackScheduler.cancel();
 			return;
 		}
-		hasSeenExecuteSequenceSinceLastPrompt = false;
 		state = TerminalState.PromptAfterExecuting;
 		scheduler.schedule();
 	}, 1000));
@@ -187,8 +188,6 @@ export async function trackIdleOnPrompt(
 		Executing,
 		PromptAfterExecuting,
 	}
-	let state: TerminalState = TerminalState.Initial;
-	let hasSeenExecuteSequenceSinceLastPrompt = false;
 	store.add(onData(e => {
 		// Update state
 		// p10k fires C as `133;C;`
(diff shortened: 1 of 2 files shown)


src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/executeStrategy/executeStrategy.ts
```diff
@@ -165,6 +165,16 @@ export async function trackIdleOnPrompt(
 	const scheduler = store.add(new RunOnceScheduler(() => {
 		idleOnPrompt.complete();
 	}, idleDurationMs));
+	// Fallback in case prompt sequences are not seen but the terminal goes idle.
+	const promptFallbackScheduler = store.add(new RunOnceScheduler(() => {
+		if (hasSeenExecuteSequenceSinceLastPrompt) {
+			promptFallbackScheduler.cancel();
+			return;
+		}
```
Tyriar: Instead of a new `hasSeenExecuteSequenceSinceLastPrompt`, you can check `state` instead. If it's `Executing` or `PromptAfterExecuting` we want to cancel it.

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/executeStrategy/executeStrategy.ts
```diff
@@ -178,26 +188,36 @@ export async function trackIdleOnPrompt(
 		PromptAfterExecuting,
 	}
 	let state: TerminalState = TerminalState.Initial;
+	let hasSeenExecuteSequenceSinceLastPrompt = false;
```
Tyriar: nit: These should move above promptFallbackScheduler since they're references earlier than they're defined. Even though this compiles/runs it helps with readability.

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/executeStrategy/executeStrategy.ts
```diff
@@ -178,26 +188,36 @@ export async function trackIdleOnPrompt(
...
 		if (state === TerminalState.PromptAfterExecuting) {
+			promptFallbackScheduler.cancel();
 			scheduler.schedule();
 		} else {
 			scheduler.cancel();
+			if (!hasSeenExecuteSequenceSinceLastPrompt) {
+				promptFallbackScheduler.schedule();
+			} else {
+				promptFallbackScheduler.cancel();
+			}
```
Tyriar: We only to need to do this when state is `Initial` or `Prompt`?

Tyriar (dismissed):

benibenj (approved):

Tyriar (approved):

PR #278171, opened 2025-11-18
chat: add requestNeedsInput on the chat model

connor4312:
With a side quest that makes elicitations correctly trigger this, which
involved refactoring their state into an observable.

Also swap requestInProgress/Obs to a single requestInProgress observable.
Synchronous callers can just `.get()` it as needed.

Refs #277318

### DIFF
--- a/src/vs/workbench/contrib/chat/common/chatModel.ts
+++ b/src/vs/workbench/contrib/chat/common/chatModel.ts
@@ -4,6 +4,7 @@
  *--------------------------------------------------------------------------------------------*/
 
 import { asArray } from '../../../../base/common/arrays.js';
+import { softAssertNever } from '../../../../base/common/assert.js';
 import { BugIndicatingError } from '../../../../base/common/errors.js';
 import { Emitter, Event } from '../../../../base/common/event.js';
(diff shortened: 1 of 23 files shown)


src/vs/workbench/contrib/chat/common/chatModel.ts
```diff
@@ -1481,9 +1484,13 @@ export class ChatModel extends Disposable implements IChatModel {
 
 		const lastResponse = observableFromEvent(this, this.onDidChange, () => this._requests.at(-1)?.response);
 
-		this.requestInProgressObs = lastResponse.map((response, r) => {
+		this.requestInProgress = lastResponse.map((response, r) => {
 			return response?.isInProgress.read(r) ?? false;
 		});
+
+		this.requestNeedsInput = lastResponse.map((response, r) => {
```
roblourens, line 1491: How do you feel about two observables vs one state enum like `Idle, Running, WaitingForConfirmation`? The number of states might increase

roblourens: I mean on second thought we should keep the basic requestInProgress regardless, and then could refactor if there are more things to track.

roblourens (approved):

PR #278183, opened 2025-11-18
chore: enable BinSkim on Linux

rzhao271:
This PR enables BinSkim scanning on our Linux build agents.
It also removes the unused BinSkim configuration within the macOS build template.

Verification build: https://dev.azure.com/monacotools/Monaco/_build/results?buildId=376247&view=results

### DIFF
--- a/build/azure-pipelines/linux/product-build-linux.yml
+++ b/build/azure-pipelines/linux/product-build-linux.yml
@@ -31,6 +31,7 @@ jobs:
     templateContext:
       sdl:
         binskim:
+          analyzeTargetGlob: '+:file|$(Agent.BuildDirectory)/VSCode-linux-*/**/*.node;+:file|$(Agent.BuildDirectory)/VSCode-linux-*/**/*.so;'
           preReleaseVersion: '4.3.1'
       outputParentDirectory: $(Build.ArtifactStagingDirectory)/out
       outputs:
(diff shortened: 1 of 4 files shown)


build/azure-pipelines/linux/product-build-linux.yml
```diff
@@ -29,6 +29,11 @@ jobs:
       NPM_ARCH: ${{ parameters.NPM_ARCH }}
       VSCODE_ARCH: ${{ parameters.VSCODE_ARCH }}
     templateContext:
+      sdl:
+        binskim:
+          scanOutputDirectoryOnly: true
+          analyzeTargetGlob: '**/.build/**/*.node'
```
deepak1556: are these two parameters `scanOutputDirectoryOnly` and `analyzeTargetGlob` exclusive ? Because `.build` directory will not be present in the out folder, it is usually under the src folder unless this is talking about the `.build` folder being published to the pipeline artifact. In either case it will not contain any `.node` files. Why not focus on the `.node` modules in the out folder instead specifically `$(Build.ArtifactStagingDirectory)/out/client` and `$(Build.ArtifactStagingDirectory)/out/server`, that is what we end up shipping ?

rzhao271: I forgot that I could check the templateContext to see which folders corresponded to the archives. Fixed!

rzhao271: It seems those output directories only contain archives. I'll run another build to confirm and then check which parts of `.build` contain the actual source files.

deepak1556: Sorry the unarchived builds are one level up the src checkout thats what you would want to use, https://github.com/microsoft/vscode/blob/9c4786defa45e4c37ca6d7b9f909a0d27bb7c33e/build/azure-pipelines/linux/steps/product-build-linux-compile.yml#L205-L210

deepak1556 (approved):

PR #278195, opened 2025-11-18
add event for chat view openend

sbatten:


mjbvz (approved):

sbatten:
It won't show on new chats in the same window. I am able to reliably see this trigger in a new window or reloading.

PR #278204, opened 2025-11-18
Adding file stats to local chat provider

osortega:
Note this is backed by the edit session, which means the stats will be lost once the user has accepted the changes or the edit session closes


Tyriar (dismissed):

src/vs/workbench/contrib/chat/browser/chatSessions/localChatSessionsProvider.ts
```diff
@@ -254,21 +257,48 @@ export class LocalChatSessionsProvider extends Disposable implements IChatSessio
...
 
 		} catch (error) {
 			return [];
 		}
 	}
+
+	private getSessionStatistics(chatModel: IChatModel) {
+		let linesAdded = 0;
+		let linesRemoved = 0;
+		const modifiedFiles = new Set<URI>();
```
mjbvz: This needs to use a `ResourceSet` to work correctly

mjbvz: @osortega Please make sure to fix this before merging. Thanks

Tyriar: ☝️

Tyriar (changes requested):

src/vs/workbench/contrib/chat/browser/chatSessions/localChatSessionsProvider.ts
osortega: Ahh nice, I didn't know that existed. Changed it

Tyriar (dismissed):

dmitrivMS (approved):

PR #278219, opened 2025-11-18
stream content to inline chat terminal

### LINKED ISSUE #276946: Embed xterm.js instead of text content

meganrogge:
fixes #276946
fixes #271390

To do/defer:
- occasionally the next prompt is included in the output, tried to fix that, but looks like the endMarker is just wrong in these cases
- add setting to determine if it's expanded when streaming or if only expanded for non-zero exit code
- calculate max width so horizontal scrollbar is hidden unless necessary

Summary:
- Owns the rolling chat-terminal buffer (terminalCommandOutput), resetting it from stored output or when a new stream begins so the preview knows the stream state (isStreaming, needsReplay, shouldRender).
- Adds incremental VT data via appendData, which keeps the stream active, concatenates chunks, and records them for later replay.
- Diffs each snapshot from xterm against the last one in applySnapshot, emitting one of four mutations: no-op, append (shared prefix), truncate+append (partial overlap), or full replace when the buffer diverges.
- Detects overlap for truncation with a bounded KMP search and overlap/ratio thresholds, protecting against quadratic scans and avoiding trims for small prompt churn.
- Provides helpers (applyEmptyOutput, countRenderableLines, markRenderableOutput, getBufferedText, getBuffer) so the view can clear, size, or replay the preview without re-reading the terminal.


src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
```diff
@@ -754,168 +927,321 @@ class ChatTerminalToolOutputSection extends Disposable {
...
-		const { content } = h('div.chat-terminal-output-content@content');
-		let inlineOutput: HTMLElement | undefined;
-		let preElement: HTMLElement | undefined;
+	private _hasRenderableOutput(): boolean {
+		return this._streamBuffer.some(chunk => {
+			const withoutCsi = chunk.replace(CSI_SEQUENCE_REGEX, '');
+			const withoutAnsi = removeAnsiEscapeCodes(withoutCsi);
+			return withoutAnsi.replace(/\r/g, '').trim().length > 0;
+		});
+	}
```
Tyriar: `removeAnsiEscapeCodes` should remove CSI too:

https://github.com/microsoft/vscode/blob/7e98d11d64b4d8d1c6aa60d72c5c6c27ecced23c/src/vs/base/common/strings.ts#L788-L796

We should use events for this instead of doing a bunch of string manipulations to try determine it. If `onWriteParsed`/`onCursorMove`/`onRender` from xterm don't give us what we need we could expose another event.

src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
```diff
@@ -669,19 +735,95 @@ export class ChatTerminalToolProgressPart extends BaseChatToolInvocationSubPart
...
+			await this._syncStreamingSnapshot(instance, command);
+		};
+		// Keep the queue alive even if a prior snapshot rejected (markers can disappear as terminals dispose).
+		// Without catching here, a single rejection would leave the chain permanently rejected and skip future runs.
+		const next = this._streamingQueue.catch(() => undefined).then(() => run());
+		this._streamingQueue = next.catch(() => undefined);
+		return next;
+	}
+
+	private async _syncStreamingSnapshot(instance: ITerminalInstance, command: ITerminalCommand): Promise<void> {
```
Tyriar: Everything related to mirroring the xterm content should be encapsulated in it's own class

src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
```diff
@@ -51,23 +48,24 @@ import { ChatContextKeys } from '../../../common/chatContextKeys.js';
...
-	}
-});
+const CSI_SEQUENCE_REGEX = /\x1b\[[0-9;?]*[ -/]*[@-~]/g;
 
 /**
  * Remembers whether a tool invocation was last expanded so state survives virtualization re-renders.
  */
 const expandedStateByInvocation = new WeakMap<IChatToolInvocation | IChatToolInvocationSerialized, boolean>();
 
+const DEFAULT_ROW_HEIGHT_PX = 16;
```
Tyriar: This should be derived from the terminal font size setting via `ITerminalConfigurationService.getFont`?

src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
```diff
@@ -153,12 +151,10 @@ class TerminalCommandDecoration extends Disposable {
...
 			storedState = terminalData.terminalCommandState;
-		} else if (!this._options.terminalData.terminalCommandOutput) {
-			if (!storedState) {
-				const now = Date.now();
-				terminalData.terminalCommandState = { exitCode: undefined, timestamp: now };
-				storedState = terminalData.terminalCommandState;
-			}
+		} else if (!storedState) {
+			const now = Date.now();
+			terminalData.terminalCommandState = { exitCode: undefined, timestamp: now };
```
```suggestion
			terminalData.terminalCommandState = { exitCode: undefined, timestamp: Date.now() };
```

src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
```diff
@@ -226,9 +222,14 @@ export class ChatTerminalToolProgressPart extends BaseChatToolInvocationSubPart
...
 	private _terminalCommandUri: URI | undefined;
 	private _storedCommandId: string | undefined;
+
 	private readonly _isSerializedInvocation: boolean;
 	private _terminalInstance: ITerminalInstance | undefined;
 	private readonly _decoration: TerminalCommandDecoration;
+	private readonly _commandStreamingListener: MutableDisposable<DisposableStore>;
+	private _streamingCommand: ITerminalCommand | undefined;
+	private _trackedCommandId: string | undefined;
+	private _streamingQueue: Promise<void>;
```
Tyriar: This ends up being a potentially enormous chain of promises. Instead of this if you need to track a queue of updates store the items in a `IDataType[]` or similar, then schedule the update somehow like in an animation frame.

src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
```diff
@@ -286,6 +287,10 @@ export class ChatTerminalToolProgressPart extends BaseChatToolInvocationSubPart
 			getIconElement: () => undefined,
 			getResolvedCommand: () => this._getResolvedCommand()
 		}));
+		this._commandStreamingListener = this._register(new MutableDisposable<DisposableStore>());
```
Tyriar: `<DisposableStore>` isn't necessary here as you don't need anything beyond `IDisposable` which is the default `T`

src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
```diff
@@ -754,168 +927,321 @@ class ChatTerminalToolOutputSection extends Disposable {
...
+			this._lastRawSnapshot = snapshot;
 			return;
 		}
-		const commandId = this._terminalData.terminalCommandId ?? this._getStoredCommandId();
-		if (!commandId) {
+		if (this._lastRawSnapshot.startsWith(snapshot)) {
+			this._replaceStreamingDataFromRaw(snapshot);
+			this._lastRawSnapshot = snapshot;
 			return;
 		}
```
Tyriar: This approach will be slow when the terminal commands are large (`largeString.startsWith(largeString2)` needs to iterate `largeString2.length` chars). Additionally I think when the command starts to get trimmed it'll fail entirely and reprint everything since the beginning will be slow.

src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
```diff
@@ -754,168 +927,321 @@ class ChatTerminalToolOutputSection extends Disposable {
...
+	}
 
-		if (result.truncated) {
-			const { info } = h('div.chat-terminal-output-info@info');
-			info.textContent = localize('chat.terminalOutputTruncated', 'Output truncated to first {0} lines.', CHAT_TERMINAL_OUTPUT_MAX_PREVIEW_LINES);
-			content.appendChild(info);
+	private _disposeDetachedTerminal(): void {
+		if (this._detachedTerminal.value) {
+			this._detachedTerminal.clear();
+		}
```
```suggestion
		this._detachedTerminal.clear();
```

src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
```diff
@@ -754,168 +927,321 @@ class ChatTerminalToolOutputSection extends Disposable {
...
+		if (this._outputResizeObserver) {
+			this._outputResizeObserver.disconnect();
+			this._outputResizeObserver = undefined;
 		}
```
```suggestion
		this._outputResizeObserver?.disconnect();
		this._outputResizeObserver = undefined;
```

src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
```diff
@@ -926,51 +1252,227 @@ class ChatTerminalToolOutputSection extends Disposable {
...
 			this._outputResizeObserver = undefined;
 		}));
 	}
+
+	private _calculateVisibleContentHeight(): number {
+		const nonEmptyLines = this._countNonEmptyStreamLines();
+		const effectiveLines = Math.max(nonEmptyLines, 1);
+		const infoHeight = this._infoElement?.offsetHeight ?? 0;
+		// TODO: handle this better, call some function that tells us if there's any output instead of checking text content
+		if (this._infoElement?.textContent.includes('No output was produced by the command.')) {
```
Tyriar: Think this is localized so would fail for non-English?

src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
```diff
@@ -926,51 +1252,227 @@ class ChatTerminalToolOutputSection extends Disposable {
...
+			this._captureXtermElements(instance);
+			this._scrollable.scanDomNode();
+			return instance;
+		} catch {
+			return undefined;
+		}
+	}
+
+	private _captureXtermElements(instance: IDetachedTerminalInstance): void {
+		const xterm = instance.xterm as unknown as XtermTerminal | undefined;
```
Tyriar: This is a dangerous cast, also it looks like `IDetachedTerminalInstance.xterm` is never `undefined`

src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
```diff
@@ -926,51 +1252,227 @@ class ChatTerminalToolOutputSection extends Disposable {
...
+				for (let i = cur.children.length - 1; i >= 0; i--) {
+					stack.push(cur.children[i]);
+				}
+			}
+			return;
+		};
+
+		const xtermElement = rawElement.classList.contains('xterm')
+			? rawElement
+			: findElementWithClass(rawElement, 'xterm');
```
Tyriar: No understanding this? `findElementWithClass` seems to just be reimplementing `querySelector`, but isn't `xterm?.raw.element` exactly what we want always?

src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
```diff
@@ -754,168 +927,321 @@ class ChatTerminalToolOutputSection extends Disposable {
...
+		// Streams raw chunks into the preview buffer, enforcing the configured line limit and
+		// mirroring any appended data into the detached xterm when it's live.
+		if (!data || this._lastOutputTruncated) {
+			return this._lastOutputTruncated;
+		}
+		this._isStreaming = true;
+		const storedOutput = this._terminalData.terminalCommandOutput ?? (this._terminalData.terminalCommandOutput = { text: '', truncated: false });
+		const maxLines = CHAT_TERMINAL_OUTPUT_MAX_PREVIEW_LINES;
+		const remainingLines = Math.max(0, maxLines - this._bufferedLineCount);
+		const trimmed = this._trimToRemainingLines(data, remainingLines);
```
Tyriar: I don't think we need to care about tracking a max line count if working with xterm.js when we fix the perf problems by reducing the rendered amount to only changed lines.

Tyriar (dismissed):

src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
```diff
@@ -698,38 +840,58 @@ class ChatTerminalToolOutputSection extends Disposable {
...
-		this._ensureTerminalInstance = options.ensureTerminalInstance;
-		this._resolveCommand = options.resolveCommand;
-		this._getTerminalTheme = options.getTerminalTheme;
-		this._getStoredCommandId = options.getStoredCommandId;
+		this._detachedTerminal = this._register(new MutableDisposable<IDetachedTerminalInstance>());
+		this._container = _options.container;
+		this._title = _options.title;
+		this._displayCommand = _options.displayCommand;
+		this._terminalData = _options.terminalData;
+		this._accessibleViewService = _options.accessibleViewService;
```
Tyriar: DI?

src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
meganrogge: this was used before when I was hacking things together, but pretty sure we can remove it.

src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
meganrogge: this is needed because `raw` doesn't exist on the element until `attachToElement` is called.

src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
meganrogge: I thought we needed this, but you're right, the raw element seems to work too

src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
meganrogge: that's already the case - we onoly render changed lines.

src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
Tyriar: If we need `raw.element`, let's expose a `get xtermElement(): HTMLElement | undefined` on `IXtermTerminal`. The reason for the interface here hiding `raw` is generally we want to hide the complex xterm interface from outside components (like `contrib/chat`).

DonJayamanne (dismissed):

src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
Tyriar, line 1: Testing out and it doesn't handle commands with lots of output well

Finished inline:

Image

Actual command:

Image

Tyriar, line 1: Content is written to the very top in what's meant to be scrollback, the diffing logic doesn't appear to work:

Image

File it's running:

```
// Print a random number every 500ms for 10 seconds
let count = 0;
const maxCount = 20; // 10 seconds / 0.5 seconds = 20 iterations

const interval = setInterval(() => {
    count++;
    const randomNumber = Math.floor(Math.random() * 1000);
    console.log(`${count}/20: ${randomNumber}`);
    
    if (count >= maxCount) {
        clearInterval(interval);
        console.log('Done!');
    }
}, 500);

console.log('Starting random number generator...');
```

src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
```diff
@@ -207,6 +200,338 @@ class TerminalCommandDecoration extends Disposable {
...
+	| { readonly kind: 'truncate'; readonly truncated: number; readonly appended: string };
+
+// Encapsulates the rolling buffer of serialized terminal output so the UI only needs to worry
+// about mirroring data into the preview. The heavy lifting happens here, including diffing the
+// newest VT snapshot to decide when we can append, truncate, or fully replace content.
+class ChatTerminalStreamingModel {
+	// Tuning notes:
+	//  - Prefix sample (256 chars) is big enough to catch prompt churn while staying cheaper than diffing full snapshots.
+	//  - Overlap window (32 KiB) keeps the KMP scan bounded to a few dozen lines of VT output so we avoid quadratic scans.
+	//  - When trimming, we require both 60% overlap and at least 2 KiB of shared content so we only treat large shifts as truncations.
```
Tyriar: This seems needlessly complex. I would not think we need to capture incoming data buffers but a set of dirty and/or up to date lines relative to the start marker. Some runtime characteristics:

- The chance existing lines outside the current line need to be changes is small
- The chance it needs to happen excessively is even smaller
- Rendering a single line is relatively cheap, so I wouldn't bother trying to diff at the character level unless it ends up being a problem

src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
Tyriar, line 1: I recommending taking a step back from this and first replacing the HTML with the xterm instance in its own PR, focusing on maintainability and small modules. At 2100 lines that makes this larger than the entirety of terminalTaskSystem.ts or terminalService.ts.

Tyriar, line 1469: Seeing this now, we should not be registering keybindings, menus or commands outside of contribution files

src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
```diff
@@ -230,6 +555,14 @@ export class ChatTerminalToolProgressPart extends BaseChatToolInvocationSubPart
 	private _terminalInstance: ITerminalInstance | undefined;
 	private readonly _decoration: TerminalCommandDecoration;
 
+	private readonly _commandStreamingListener: MutableDisposable<IDisposable>;
+	private _streamingCommand: ITerminalCommand | undefined;
+	private _trackedCommandId: string | undefined;
+	private _streamingQueue: IStreamingSnapshotRequest[];
+	private readonly _streamingSnapshotRetryCounts = new WeakMap<ITerminalCommand, number>();
+	private _isDrainingStreamingQueue = false;
+	private _streamingDrainScheduled = false;
```
Tyriar: I question the queue approach in general, but these are the sort of properties that should live in one of the other classes.

src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
```diff
@@ -230,6 +555,14 @@ export class ChatTerminalToolProgressPart extends BaseChatToolInvocationSubPart
 	private _terminalInstance: ITerminalInstance | undefined;
 	private readonly _decoration: TerminalCommandDecoration;
 
+	private readonly _commandStreamingListener: MutableDisposable<IDisposable>;
+	private _streamingCommand: ITerminalCommand | undefined;
+	private _trackedCommandId: string | undefined;
+	private _streamingQueue: IStreamingSnapshotRequest[];
+	private readonly _streamingSnapshotRetryCounts = new WeakMap<ITerminalCommand, number>();
```
Tyriar: It's very rare to actually need `WeakMap`, using it makes red flags go off in my mind that the lifecycle of this class is not managed properly.

src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
```diff
@@ -754,168 +1411,295 @@ class ChatTerminalToolOutputSection extends Disposable {
...
+	private _clearDetachedTerminal(): void {
+		// Clears the detached xterm prior to replaying content so the terminal reflects the latest
+		// snapshot exactly.
+		const instance = this._detachedTerminal.value;
+		if (!instance) {
 			return;
 		}
-		const command = commands.find(c => c.id === commandId);
-		if (!command?.endMarker) {
+		const xterm = instance.xterm as unknown as XtermTerminal | undefined;
```
Tyriar: As mentioned before, `as unknown as` is dangerous and I don't think this can ever be `undefined` looking at the interface.

Tyriar (changes requested):

src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
meganrogge: I can repro

meganrogge: moving `ChatTerminalStreamingModel` to its own file helps with this, but yes, can do

src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
meganrogge: I thought this is what you meant by not chaining promises together and keeping track of the requests in a Data[]. What do you think would be better?

src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
meganrogge: `raw` does not exist on `DetachedTerminalInstance`. Perhpas i'll fix that and make `raw` optional?

Screenshot 2025-11-20 at 12 37 17 PM

meganrogge: did so

src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
meganrogge: not sure what's happening there 🤔

meganrogge: oh I see what's wrong

meganrogge: fixed

meganrogge: separate pr

PR #278233, opened 2025-11-18
fix thinking flickering and scrolling

justschen:
fixes this issue where after we get a tool and start a new thinking part, we try to add that thinking part to `value`, but it is already added somewhere else, so it causes the scroll to get weird!

https://github.com/user-attachments/assets/f7c277c6-81c1-4a0a-a2cd-82394836f769

### DIFF
--- a/src/vs/workbench/contrib/chat/browser/chatContentParts/chatThinkingContentPart.ts
+++ b/src/vs/workbench/contrib/chat/browser/chatContentParts/chatThinkingContentPart.ts
@@ -105,7 +105,7 @@ export class ChatThinkingContentPart extends ChatCollapsibleContentPart implemen
 			}));
 		}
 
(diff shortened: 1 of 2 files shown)


src/vs/workbench/contrib/chat/browser/chatContentParts/chatThinkingContentPart.ts
```diff
@@ -105,6 +104,9 @@ export class ChatThinkingContentPart extends ChatCollapsibleContentPart implemen
 				}
 			}));
 		}
+
+		const label = localize('chat.thinking.progress.withHeader', '{0}{1}', this.lastExtractedTitle ?? this.currentTitle, this.hasMultipleItems ? '...' : '');
```
roblourens: You don't have to localize this, it's just concatenating two unlocalized strings?

roblourens (dismissed):

dmitrivMS (approved):

PR #278275, opened 2025-11-19
Fix not properly reading observable

roblourens:
Tried to simplify, dropped the reader

### DIFF
--- a/src/vs/workbench/contrib/chat/browser/chatViewPane.ts
+++ b/src/vs/workbench/contrib/chat/browser/chatViewPane.ts
@@ -7,7 +7,7 @@ import { $, getWindow } from '../../../../base/browser/dom.js';
 import { CancellationToken } from '../../../../base/common/cancellation.js';
 import { DisposableStore } from '../../../../base/common/lifecycle.js';
 import { MarshalledId } from '../../../../base/common/marshallingIds.js';
-import { autorun } from '../../../../base/common/observable.js';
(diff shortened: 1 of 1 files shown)


chrmarti (approved):

src/vs/workbench/contrib/chat/browser/chatViewPane.ts
```diff
@@ -229,14 +229,14 @@ export class ChatViewPane extends ViewPane implements IViewWelcomeDelegate {
 			}));
 		this._widget.render(parent);
 
-		const updateWidgetVisibility = () => {
```
hediet, line 232: My recommendation would be to inline `updateWidgetVisibility`. In that case, the lint rule would warn you about the `get` inside of an autorun/derived.
Also, it seems like `isBodyVisible` is not observable? Can it update?

roblourens: Ah, didn't know there was a lint rule, thanks. I had it inlined originally, then it felt like duplicating code. For isBodyVisible, I just have the event, it would be nice to have an observable. I guess I could create the observable at this level too though.

hediet: There is `observableFromEvent` when you have an actual value (it only fires when the value actually changes).
There is `signalFromEvent` when you don't have an value (it always fires when the event fires).
With that you can do it in one autorun!

PR #278277, opened 2025-11-19
make MCP titles look more like regular tool titles

justschen:
Screenshot 2025-11-19 at 10 53 32 AM


Tyriar (approved):
@justschen I was hoping this clean up would use the same component to wrap all tools so we get guaranteed consistent rendering for the title and border, basically everything except what's in the expanded part.

PR #278319, opened 2025-11-19
Adjust padding for select box dropdown and remove border radius styles

mrleemurray:
Modify the padding for the select box dropdown to improve layout and remove border radius styles for a cleaner appearance.

Aligns listbox content with select content & removes any radius mismatch on hovered list items:

image


bpasero (changes requested):
So how does it really look like with the border radius removed? I think your screenshot is still with radius.

bpasero (approved):
👍

PR #278331, opened 2025-11-19
fix: possible memory leak with decoration registration

SimonSiefke:
Fixes a few possible memory leaks with decoration registration by registering the returned disposable. 

### Details
Currently decorations are registered like this:

```ts
this.codeEditorService.registerDecorationType(decorationDescription, placeholderDecorationType, {});
```

But it can be easy to forget that most likely the decoration also needs to be removed on dispose:

```ts
this.codeEditorService.removeDecorationType(placeholderDecorationType);
```

### Change

The change is to switch to using `this._register`, which seems to make it easier to register the decoration as a disposable. Now for example in `commentsController` or `breakPointWidget` the decoration registration is registered as a disposable.

### DIFF
--- a/src/vs/editor/browser/services/codeEditorService.ts
+++ b/src/vs/editor/browser/services/codeEditorService.ts
@@ -43,7 +43,7 @@ export interface ICodeEditorService {
 	 */
 	getFocusedCodeEditor(): ICodeEditor | null;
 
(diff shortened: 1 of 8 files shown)


hediet (approved):

hediet:
Thank you!

Tyriar (approved):

mjbvz (approved):

PR #278359, opened 2025-11-19
fix hover scrollbar not appearing when content overflows

reagenzn:
Fixes the issue where hover tooltips with `maxHeightRatio` option don't show scrollbars when content exceeds the maximum height.

### Problem
When `IHoverService.showInstantHover()` is called with `appearance.maxHeightRatio`, the scrollbar doesn't appear even when content overflows. This affects:
- Terminal tab hovers
- Status bar hovers  
- Other components using `maxHeightRatio`

### Root Cause
`adjustHoverMaxHeight()` only set `maxHeight` on `containerDomNode`, but not on `contentsDomNode`. This caused `DomScrollableElement.scanDomNode()` to see `clientHeight === scrollHeight` (both equal to full content height), preventing scrollbar detection.

### Solution
Set `maxHeight` on both `containerDomNode` and `contentsDomNode`:
- `containerDomNode`: Controls the visible window size
- `contentsDomNode`: Enables scrollbar detection (`clientHeight < scrollHeight`)

This aligns with the existing implementation in `ContentHoverWidget._setHoverWidgetMaxDimensions()`.

### Testing

Manual Test:
Tested with terminal tab hovers containing long status information.

|Before|After|
|---|---|
|image|image|

### DIFF
--- a/src/vs/platform/hover/browser/hoverWidget.ts
+++ b/src/vs/platform/hover/browser/hoverWidget.ts
@@ -572,6 +572,7 @@ export class HoverWidget extends Widget implements IHoverWidget {
 		}
 
 		this._hover.containerDomNode.style.maxHeight = `${maxHeight}px`;
+		this._hover.contentsDomNode.style.maxHeight = `${maxHeight}px`;
 		if (this._hover.contentsDomNode.clientHeight < this._hover.contentsDomNode.scrollHeight) {
 			// Add padding for a vertical scrollbar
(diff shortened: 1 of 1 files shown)


aiday-mar (commented):
Hi thanks for this PR. Could you please link the issue you are trying to fix? I am not sure I am able to reproduce this issue. Could you include a numbered set of steps for how to reproduce this issue.

PR #278374, opened 2025-11-19
Add platform and releaseDate exp filter

vijayupadya:
Fixes https://github.com/microsoft/vscode-internalbacklog/issues/6295 and https://github.com/microsoft/vscode-internalbacklog/issues/6325

Add platform and release date filter for experimentation


sbatten:
Do we know what happens in remote scenarios? What should we do for user on Windows using SSH to connect to Linux?

I *think* this would always be the render process and that is where our ExP run so it is correct. Just want to verify.

sbatten (approved):

PR #278423, opened 2025-11-19
Implements TextEdit.compose

hediet:


src/vs/editor/common/core/edits/textEdit.ts
```diff
@@ -204,6 +204,380 @@ export class TextEdit {
 		return equals(this.replacements, other.replacements, (a, b) => a.equals(b));
 	}
 
+	/**
+	 * Combines two edits into one with the same effect.
+	 * WARNING: This is written by AI, but well tested. I do not understand the implementation myself.
```
dmitrivMS, line 209: !

dmitrivMS (approved):

src/vs/editor/common/core/edits/textEdit.ts
hediet: Yes, this is a first! The test-suite contains a fuzzer that I ran with millions of different inputs. Luckily, composition has a nice definition of correctness that can easily be checked automatically, which is why I fully trust the code.

I was looking for an implementation for this for quite some years already and was more than impressed when Gemini 3 finally solved it.
The implementation of [`combineTextEditInfos`](https://github.com/microsoft/vscode/blob/f0e31b98fb8be45b50481070aafffebdfd285c78/src/vs/editor/common/model/bracketPairsTextModelPart/bracketPairsTree/combineTextEditInfos.ts#L10) is close (and took me quite some time to get right), but just tracking the lengths is easier than also handling the text correctly.
Composing offset-based `StringEdit`s is also [much easier](https://github.com/microsoft/vscode/blob/f0e31b98fb8be45b50481070aafffebdfd285c78/src/vs/editor/common/core/edits/edit.ts#L101).

In the past, whenever I needed to compose (line/column based) TextEdits, I first converted them into offset-based StringEdits, composed them and converted them back. This is not only imperformant, but sometimes the conversion cannot be done, as the positions of the `\n` is not available anymore.

PR #278640, opened 2025-11-20
Support streaming of partial tool data

lramos15:
I'm not sure I love this flow, but I wanted each tool to be able to handle its own messaging while data streams in because it would likely know best. If we want to simplify the API we could just have one centralized streaming tool call processor in Copilot that handles everything before any tools have prepare called.

The current flow is 
1. Stream gets a `prepareToolInvocation` pushed to it with partial stream data
2. Core calls `handleToolStream` with that data
3. The returned invocation message is then rendered as a progress message which updates the progress message rather than stacking messages


src/vscode-dts/vscode.proposed.chatParticipantAdditions.d.ts
```diff
@@ -668,6 +684,36 @@ declare module 'vscode' {
 		model?: LanguageModelChat;
 	}
 
+	export interface LanguageModelToolInvocationStreamOptions<T> {
+		/**
+		 * Raw argument payload, such as the streamed JSON fragment from the language model.
+		 */
+		readonly rawInput?: unknown;
```
connor4312, line 695: I question whether this is useful/something we'd want to expose in the API like this. Tools are generic to any specific model. The tool doesn't know how the input is encoded. JSON is the default right now, but OAI's has `raw tool calls, and maybe in the future something like [Toon](https://github.com/toon-format/toon) gains traction.

We can also do some transforms on the call object after validation -- for example, Sonnet has a propensity to double JSON-encode nested objects and arrays and we can correct those.

That said I really like the idea of being able to show e.g. the files that are about to be edited while the call is streaming in... maybe we just

1. spec this is a fragment of a JSON-encoded `string` (and it's up to the implementation to convert that if needed)
2. explicitly say the input might be transformed before the actual tool call

(of course this is all proposed API right now, but for the general shape of things)

lramos15: I think it's safe to say `handleToolStream` is just a string that will be partial data from whatever the stream returns. Then when `prepareInvocation` is called we guarnatee it's in some format the tool expects. My guess is the tool will have to know how to handle various formats.

The benefit of exposing this is if you have a durable JSON parser that can parse incomplete data you could likely render that file `foo.js` is being created and it has x lines at the moment

connor4312: I agree, but I'm suggesting we have a way to identify that. Maybe there's a mimeType on the stream so the producer/consumer can handle the data correctly? OAI "raw" tools would emit `text/plain` as their mimetype. (We don't have any of these yet, but they have said that raw works a lot better for apply_patch and it's on our todo to adopt that)

src/vs/workbench/api/browser/mainThreadChatAgents2.ts
```diff
@@ -321,6 +323,37 @@ export class MainThreadChatAgents2 extends Disposable implements MainThreadChatA
...
+						let progressContent: IMarkdownString;
+						if (streamResult?.invocationMessage) {
+							progressContent = typeof streamResult.invocationMessage === 'string'
+								? new MarkdownString(streamResult.invocationMessage)
+								: streamResult.invocationMessage;
+						} else {
+							progressContent = new MarkdownString('Invoking tool: ' + toolData.displayName);
+						}
+						// Generate a progress message part from the handleToolStream result
+						const progressMessage: IChatProgressMessage = {
```
roblourens: Not sure I follow why there's a progress message involved here. I'd expect that the partial tool content just gets pushed into the ChatModel and then is rendered similarly to normal tool invocations

roblourens: It seems like you're making a decision about how content should be rendered at this level, when the renderer should decide that

src/vscode-dts/vscode.proposed.chatParticipantAdditions.d.ts
```diff
@@ -81,7 +81,19 @@ declare module 'vscode' {
...
 		toolName: string;
-		constructor(toolName: string);
+		/**
+		 * Partial arguments that have streamed in for the tool invocation.
+		 */
+		streamData?: ChatToolInvocationStreamData;
+		constructor(toolName: string, streamData?: ChatToolInvocationStreamData);
+	}
+
+	export interface ChatToolInvocationStreamData {
```
roblourens, line 83: Maybe you need a tool call ID here, so we know when the real tool invocation has started? Not sure how that works now

roblourens (commented):
Can I see the extension side of it too?

src/vscode-dts/vscode.proposed.chatParticipantAdditions.d.ts
lramos15: Honestly the only thing I use here is the partial stream data, I wasn't able to extract the rest of the information and it doesn't seem all that important for rendering progress.

src/vs/workbench/api/browser/mainThreadChatAgents2.ts
lramos15: I want this to behave the same as every other progress message pretty much so it felt a bit silly to implement the rendering the same and have code duplication. But I also agree that I don't love the interception here.

roblourens: I just think that decision about how a certain data type is rendered should be made at the view level

src/vscode-dts/vscode.proposed.chatParticipantAdditions.d.ts
roblourens: I'll have to look more closely because I'm not sure how it knows to stop rendering the partial call and start rendering the real tool call

roblourens: Oh, maybe that just happens based on the logic for progress messages which get hidden when more content shows up. If that's the case, then this might be wrong for parallel tool calls?

lramos15: Yeah it just gets hidden when more content shows up. I figured for parallel tool calls we can just render the last one and don't necessarily need parallel progress. This is less about showing the progress to the user and more about showing something that doesn't feel like it's stuck in a working state

lramos15: Added tool call id

src/vs/workbench/api/browser/mainThreadChatAgents2.ts
lramos15: Moved to view layer

src/vs/workbench/api/browser/mainThreadLanguageModelTools.ts
```diff
@@ -93,6 +93,7 @@ export class MainThreadLanguageModelTools extends Disposable implements MainThre
 					}
 				},
 				prepareToolInvocation: (context, token) => this._proxy.$prepareToolInvocation(id, context, token),
+				handleToolStream: (context, token) => this._proxy.$handleToolStream(id, context, token),
```
connor4312, line 96: Should this only be defined if the tool in the ext host side has this method defined?

src/vs/workbench/contrib/chat/browser/widget/chatContentParts/chatPrepareToolInvocationPart.ts
```diff
@@ -0,0 +1,117 @@
...
+		// Create initial progress message - will be updated async
+		const initialProgressMessage = ChatPrepareToolInvocationPart.createInitialProgressMessage(prepareToolInvocation, languageModelToolsService);
+		super(initialProgressMessage, chatContentMarkdownRenderer, context, undefined, undefined, undefined, undefined, instantiationService, chatMarkdownAnchorService, configurationService);
+
+		this.element = context.element;
+
+		// Hide when following content contains parts that are not prepareToolInvocation or progressMessage
+		// This is similar to how progressMessage parts hide when other content arrives
+		const followingContent = context.content.slice(context.contentIndex + 1);
+		this.isHiddenByFollowingContent = followingContent.some(part => part.kind !== 'prepareToolInvocation' && part.kind !== 'progressMessage');
```
connor4312: I'm not a huge fan of of this approach of rendering a new part for each piece of the stream. If we imagine a pedantic case where I have a tool call emitting a word at a time generating an enormous file this could get expensive even for relatively small files of a few hundred KB.

How I would imagine it should work is the stream gets passed to `invokeTool()` in the LanguageModelToolsService, which internally could create a content part that includes an `IObservable` with whatever information should get shown in the UI

src/vs/workbench/contrib/chat/browser/widget/chatContentParts/chatPrepareToolInvocationPart.ts
```diff
@@ -0,0 +1,117 @@
...
+
+		const streamContext: IToolInvocationStreamContext = {
+			toolCallId: prepareToolInvocation.toolCallId,
+			rawInput: prepareToolInvocation.streamData?.partialInput,
+			chatRequestId: isResponseVM(this.element) ? this.element.requestId : undefined,
+			chatSessionId: this.element.sessionId
+		};
+
+		try {
+			const streamResult = await this.languageModelToolsService.handleToolStream(toolData.id, streamContext, CancellationToken.None);
```
connor4312: We cannot do side-effects like this from the chat widgets, as this will not work for background sessions.

One could say that partial results don't matter if the user isn't actively looking at the window, but we then might get into weird states if the user opens or closes the session mid-stream. Better imo to keep the view and the data model separate.

connor4312 (changes requested):

src/vs/workbench/api/browser/mainThreadLanguageModelTools.ts
lramos15: How would we know if the tool in the ext host has this defined? Isn't this the same as prepareToolInvocation?

src/vs/workbench/contrib/chat/browser/widget/chatContentParts/chatPrepareToolInvocationPart.ts
lramos15: We already do this approach for progress which is why I followed it https://github.com/microsoft/vscode/blob/95bfedc00a413d4f6047feec079a4cf497fdf97a/src/vs/workbench/contrib/chat/browser/chatContentParts/chatProgressContentPart.ts#L51-L53

since effectively this is how we want this to behave as well. But I guess with progress that is a part that gets emitted a lot less

src/vs/workbench/contrib/chat/browser/widget/chatContentParts/chatPrepareToolInvocationPart.ts
```diff
@@ -0,0 +1,117 @@
...
+	}
+
+	override hasSameContent(other: IChatRendererContent, followingContent: IChatRendererContent[], element: ChatTreeItem): boolean {
+		if (other.kind !== 'prepareToolInvocation') {
+			return false;
+		}
+
+		// If following content contains parts that are not prepareToolInvocation or progressMessage,
+		// we need to re-render to hide this part (similar to progressMessage behavior)
+		const shouldBeHidden = followingContent.some(part => part.kind !== 'prepareToolInvocation' && part.kind !== 'progressMessage');
```
roblourens: I don't think we should follow this pattern of hiding things. Now we have a toolcallid so we can associate the incomplete call with the real call, it seems to me like the incoplete call should just sort of resolve into the real call widget.

connor4312: We could potentially reuse IChatToolInvocation, moving the parameters and message parts into the state so they're only present once prepare resolves

src/vs/workbench/contrib/chat/common/chatService/chatService.ts
connor4312, line 460: I suggest having the full `parameters`, confirmationMessages, etc. only available on the `state` objects for states after `StateKind.Streaming` part. Callers will need to check the state first to know they're valid, and so having them only on those state objects forces that to happen

src/vs/workbench/contrib/chat/browser/tools/languageModelToolsService.ts
```diff
@@ -549,6 +563,104 @@ export class LanguageModelToolsService extends Disposable implements ILanguageMo
 		return prepared;
 	}
 
+	beginToolCall(options: IBeginToolCallOptions): IChatToolInvocation | undefined {
+		// First try to look up by tool ID (the package.json "name" field),
+		// then fall back to looking up by toolReferenceName
+		const toolData = this.getTool(options.toolId) ?? this.getToolByName(options.toolId);
```
connor4312: I think this should identical logic to invokeTool which uses `this._tools.get(toolId)` not name.

src/vs/workbench/contrib/chat/browser/tools/languageModelToolsService.ts
```diff
@@ -549,6 +563,104 @@ export class LanguageModelToolsService extends Disposable implements ILanguageMo
...
+				}
+			} else {
+				this._logService.warn(`[LanguageModelToolsService#beginToolCall] Session not found for ${options.sessionResource}`);
+			}
+		} else {
+			this._logService.warn(`[LanguageModelToolsService#beginToolCall] No sessionResource provided`);
+		}
+
+		// Call handleToolStream to get initial streaming message
+		const toolEntry = this._tools.get(toolData.id);
```
connor4312: 2nd lookup can be removed if you swap to `this._tools.get` above

src/vs/workbench/contrib/chat/browser/tools/languageModelToolsService.ts
```diff
@@ -549,6 +563,104 @@ export class LanguageModelToolsService extends Disposable implements ILanguageMo
...
+					chatRequestId: invocation.chatRequestId,
+				}, token);
+
+				this._logService.info(`[LanguageModelToolsService#updateToolStream] handleToolStream result: ${JSON.stringify(result)}`);
+				if (result?.invocationMessage) {
+					invocation.updateStreamingMessage(result.invocationMessage);
+				}
+			} catch (error) {
+				this._logService.error(`[LanguageModelToolsService#updateToolStream] Error calling handleToolStream for tool ${invocation.toolId}:`, error);
+			}
```
connor4312: this block could be deduped

src/vs/workbench/contrib/chat/browser/tools/languageModelToolsService.ts
```diff
@@ -94,6 +94,9 @@ export class LanguageModelToolsService extends Disposable implements ILanguageMo
 
 	private readonly _callsByRequestId = new Map<string, ITrackedCall[]>();
 
+	/** Pending tool calls in the streaming phase, keyed by toolCallId */
+	private readonly _pendingToolCalls = new Map<string, ChatToolInvocation>();
```
connor4312, line 98: what happens to pending tool calls if the tool call is cancelled before they complete?

src/vscode-dts/vscode.proposed.chatParticipantAdditions.d.ts
```diff
@@ -349,7 +352,21 @@ declare module 'vscode' {
...
+		 * @param streamData Optional initial streaming data with partial arguments.
+		 */
+		beginToolInvocation(toolCallId: string, toolName: string, streamData?: ChatToolInvocationStreamData): void;
+
+		/**
+		 * Update the streaming data for a tool invocation that was started with `beginToolInvocation`.
+		 * @param toolCallId The tool call ID that was passed to `beginToolInvocation`.
+		 * @param streamData New streaming data with updated partial arguments.
+		 */
+		updateToolInvocation(toolCallId: string, streamData: ChatToolInvocationStreamData): void;
```
connor4312, line 369: We'd technically be okay with a single method here that acts as a create+update

connor4312 (approved):
lgtm overall

src/vscode-dts/vscode.proposed.chatParticipantAdditions.d.ts
lramos15: Yeah I thought of that long term, will follow up with a cleanup 👍

PR #278663, opened 2025-11-20
Fix terminal mouse wheel zoom interference with scrollbar

### LINKED ISSUE #251874: terminal.integrated.mouseWheelZoom": true This configuration is repeated with the scroll bar

- VS Code Version: 1.101.0
- OS Version: win11

Steps to Reproduce:

I set the config "terminal.integrated.mouseWheelZoom": true, but when the terminal has scrollbars, the scrollbar must be at the top or bottom in order to set the font size using ctrl and mouse

zhichli:
Fixes #251874

#### Problem
`Ctrl+Wheel` zoom only works when the terminal scrollbar is at top/bottom. In the middle, it scrolls instead of zooming.

#### Cause
xterm.js's internal `ScrollableElement` consumes wheel events during bubbling phase, preventing the zoom handler from receiving them.

#### Solution
Use DOM capture phase (`{ capture: true }`) to intercept wheel events before they reach the scrollbar. This ensures zoom shortcuts work regardless of scroll position.

#### Alternative considered
Could fix in xterm.js by making `attachCustomWheelEventHandler` use capture phase, but that's riskier and affects all library consumers. This VS Code-level fix is more targeted.

### DIFF
--- a/src/vs/workbench/contrib/terminalContrib/zoom/browser/terminal.zoom.contribution.ts
+++ b/src/vs/workbench/contrib/terminalContrib/zoom/browser/terminal.zoom.contribution.ts
@@ -7,7 +7,7 @@ import type { Terminal as RawXtermTerminal } from '@xterm/xterm';
 import { Event } from '../../../../../base/common/event.js';
 import { IMouseWheelEvent } from '../../../../../base/browser/mouseEvent.js';
 import { MouseWheelClassifier } from '../../../../../base/browser/ui/scrollbar/scrollableElement.js';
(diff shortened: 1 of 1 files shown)


Tyriar (dismissed):
~This fix should go into xterm.js instead since it looks like the bug is there and we want to fix it for more than just vscode.~

Tyriar (approved):
This looks good. `attachCustomWheelEventHandler` is specifically for handling the event after the scrollable element has does its thing, so a regular dom event is fine here.

PR #278677, opened 2025-11-20
Integrated Browser

### LINKED ISSUE #277298: Exploration: rich integrated web browser
Possibly using [WebContentsView](https://www.electronjs.org/docs/latest/api/web-contents-view), or [controlledframe](https://chromestatus.com/feature/5199572022853632) (pending electron/electron#38577).

The new implementation would need to overcome the limitations of the current iframe-based [Simple Browser](https://github.com/microsoft/vscode/tree/main/extensions/simple-browser), but won't necessarily replace it (for example, the new browser might only work on Desktop).

kycutler:
Resolves #277298, #285151


src/vs/base/common/keyCodes.ts
jruales, line 464: shouldn't  `SCAN_CODE_STR_TO_EVENT_KEY_CODE` be included in `ScanCodeUtils` like the other scan code maps? And if so, `SCAN_CODE_STR_TO_EVENT_KEY_CODE` itself would need to be renamed to `scanCodeToEventKeyCode` and not exported by itself anymore but only as part of ScanCodeUtils

src/vs/code/electron-main/app.ts
```diff
@@ -1106,6 +1116,8 @@ export class CodeApplication extends Disposable {
 		// MCP
 		services.set(INativeMcpDiscoveryHelperService, new SyncDescriptor(NativeMcpDiscoveryHelperService));
 
+		// Browser View
```
jruales: nit: maybe let's move this registration higher in this file, so it's closer to other web-related registrations like `// Browser Elements`, `// Web Contents Extractor` and `// Webview Manager` ?

src/vs/workbench/contrib/browserView/electron-browser/overlayManager.ts
```diff
@@ -0,0 +1,172 @@
...
+			let didRemove = false;
+			for (const mutation of mutations) {
+				if (mutation.type === 'childList') {
+					for (const node of mutation.removedNodes) {
+						if (this.elementObservers.delete(node as HTMLElement)) {
+							didRemove = true;
+						}
+						if (this.overlayRectangles.delete(node as HTMLElement)) {
+							didRemove = true;
+						}
```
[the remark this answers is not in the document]
kycutler: fixed

src/vs/code/electron-main/app.ts
kycutler: Fair point! Moved

src/vs/workbench/contrib/browserView/electron-browser/browserEditor.ts
```diff
@@ -0,0 +1,543 @@
...
+		super({
+			id: GoForwardAction.ID,
+			title: localize2('browserGoForward', 'Go Forward'),
+			icon: Codicon.arrowRight,
+			f1: true,
+			menu: {
+				id: MenuId.BrowserNavigationToolbar,
+				group: 'navigation',
+				order: 2,
+				when: ContextKeyExpr.and(BROWSER_EDITOR_ACTIVE, CONTEXT_BROWSER_CAN_GO_FORWARD)
```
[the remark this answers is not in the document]
kycutler: This is incorrect -- the `when` in the menu also controls visibility.

src/vs/workbench/contrib/browserView/electron-browser/media/browser.css
```diff
@@ -0,0 +1,113 @@
+/*---------------------------------------------------------------------------------------------
```
bpasero, line 1: I would suggest to use CSS nesting here to ensure the selectors are encapsulated properly. Currently they seem very flat and might potentially match something else?

src/vs/workbench/contrib/browserView/electron-browser/browserEditorInput.ts
```diff
@@ -0,0 +1,232 @@
...
+import { IThemeService } from '../../../../platform/theme/common/themeService.js';
+import { TAB_ACTIVE_FOREGROUND } from '../../../common/theme.js';
+import { localize } from '../../../../nls.js';
+import { IInstantiationService } from '../../../../platform/instantiation/common/instantiation.js';
+import { IBrowserViewWorkbenchService, IBrowserViewModel } from '../../../services/browserView/common/browserView.js';
+import { hasKey } from '../../../../base/common/types.js';
+import { ILifecycleService, ShutdownReason } from '../../../services/lifecycle/common/lifecycle.js';
+import { BrowserEditor } from './browserEditor.js';
+
+const LOADING_SPINNER_SVG = (color: string | undefined) => `
```
bpasero, line 21: I am not sure I have seen this pattern used in other places, where did you get the idea from? Are we missing to support animated codicons in a editor tab?

src/vs/workbench/services/browserView/electron-browser/browserViewWorkbenchService.ts
```diff
@@ -0,0 +1,64 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import {
```
bpasero: This seems not well formatted.

src/vs/workbench/contrib/browserView/common/browserView.ts
```diff
@@ -0,0 +1,271 @@
+/*---------------------------------------------------------------------------------------------
```
bpasero, line 1: We typically do not add something to `services` unless it is considered to be a core workbench service that is used across many components. So I wonder if `services/browserView` should move into `contrib/browserView`?

src/vs/workbench/contrib/browserView/electron-browser/browserEditor.ts
```diff
@@ -0,0 +1,619 @@
...
+		const editorService = accessor.get(IEditorService);
+		const activeEditorPane = editorService.activeEditorPane;
+		if (activeEditorPane instanceof BrowserEditor) {
+			await activeEditorPane.reload();
+		}
+	}
+}
+
+// Register actions
+registerAction2(GoBackAction);
```
bpasero: Maybe to reduce file complexity, consider to move actions into a `browserViewActions.ts`?

src/vs/platform/browserView/electron-main/browserView.ts
```diff
@@ -0,0 +1,13 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { IBrowserViewService } from '../common/browserView.js';
+import { createDecorator } from '../../instantiation/common/instantiation.js';
+
+export const IBrowserViewMainService = createDecorator<IBrowserViewMainService>('browserViewMainService');
```
bpasero: This file seems not providing much value, should it be gone and merged into another one?

src/vs/platform/browserView/electron-main/plugins/themePlugin.ts
```diff
@@ -0,0 +1,68 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { Disposable } from '../../../../base/common/lifecycle.js';
```
bpasero: Is there a specific reason to have a `plugins` folder? It seems somewhat off from our typical folder structure where maybe we have a `contrib` folder for dedicated extensions to some component, but that does not seem to be the case here. I would move it one level up or inline it.

src/vs/platform/browserView/electron-main/browserViewMainService.ts
```diff
@@ -0,0 +1,593 @@
...
+import { ILogService } from '../../log/common/log.js';
+import { IInstantiationService } from '../../instantiation/common/instantiation.js';
+
+export class BrowserViewMainService extends Disposable implements IBrowserViewMainService {
+	declare readonly _serviceBrand: undefined;
+
+	/**
+	 * Check if a webContents belongs to an integrated browser view
+	*/
+	private static readonly knownSessions = new WeakSet<Electron.Session>();
```
bpasero, line 36: Using a `static` here seems off: this service is already a singleton, so why not provide these things as methods from the service?

src/vs/platform/browserView/electron-main/browserViewMainService.ts
```diff
@@ -0,0 +1,593 @@
...
+		@IInstantiationService private readonly instantiationService: IInstantiationService
+	) {
+		super();
+	}
+
+	/**
+	 * Get the session for a browser view based on data storage setting and workspace
+	 */
+	private getSession(viewId: string, workspaceId?: string): Electron.Session {
+		const dataStorageSetting = this.configurationService.getValue<string>('workbench.browser.dataStorage', { resource: undefined }) ?? 'workspace';
```
bpasero: Its a bit of a layer breaker here to depend on a workbench setting from `platform`, but not sure this can be done differently.

bpasero (changes requested):
I glanced over the changes high level and the code seems to follow our usual code patterns and contributions, which looks legit to me. There is still some feedback from me and I did not yet go into an in-depth review as the change is quite complex. 

I would also like @deepak1556 to chime in specifically on the use of `WebContentsView` here: https://github.com/microsoft/vscode/blob/d7736eb78fd13576be6fb8e2a771597717e75527/src/vs/platform/browserView/electron-main/browserViewMainService.ts#L235

And the possible consequences this could have.

src/vs/platform/browserView/electron-main/browserViewMainService.ts
kycutler: It's tricky to access the service instance from the `will-navigate` callback in app.ts, since the service is in `appInstantiationService` (and not `mainInstantiationService`) and this isn't accessible on the class. I didn't want to break any existing assumptions by making it accessible so just opted for a static. Please do let me know if you think there's a better way!

src/vs/platform/browserView/electron-main/browserView.ts
kycutler: It was more useful when putting `isBrowserViewWebContents` on the instance, to keep the Electron types isolated. I think it's nice to leave around in case we add more Electron-specific methods in the future

src/vs/workbench/contrib/browserView/electron-browser/browserEditorInput.ts
kycutler: Yes, currently label icons cannot be animated and it seems to be a non-trivial change -- e.g. just updating [this line](https://github.com/microsoft/vscode/blob/bc0c55891c7a8fe9f59d1b76e60771da99537e05/src/vs/editor/common/services/getIconClasses.ts#L19) to use `ThemeIcon.getClassArray(icon)` results in:
05c2c586-6935-4a71-9fd4-e220f1abe319

It seems the label would need some restructuring so the icon can be animated in its own element instead of just rendering in `::before`. So for now I opted to just use a custom SVG.

src/vs/platform/browserView/electron-main/browserViewMainService.ts
kycutler: Updated to take a scope option from the workbench. This is more flexible anyway so we can e.g. open an in-memory session using a command. Thanks!

src/vs/platform/browserView/electron-main/plugins/themePlugin.ts
kycutler: I had anticipated having more than just one and putting them under a folder to keep things tidy. But since it's just this one for now, and it's quite small anyway, moved it out into the same file as the view and removed the `plugins` folder.

src/vs/platform/browserView/electron-main/browserView.ts
kycutler: Actually I realized you might not be commenting on the interface itself, just the having a separate file for it. You're right 😄 moved these types into the same file as the service implementation, and moved the `BrowserView` class into this file instead of below the service.

src/vs/workbench/contrib/browserView/common/browserView.ts
kycutler: Good call! Previously the hygiene checks failed, and I thought it was due to the channel creation here, but it must have been something else. Moved all the workbench changes into `contrib`. Thanks!

src/vs/base/common/keyCodes.ts
kycutler: I don't think it should go in `ScanCodeUtils`, since that's just for converting between the numeric and string forms of scan codes. And similar with `KeyCodeUtils` for key codes. Since there are already other maps for converting across different types of codes (like `EVENT_KEY_CODE_MAP`), it seems reasonable to just follow that pattern for converting scan code strings -> event key codes

src/vs/platform/browserView/electron-main/browserView.ts
```diff
@@ -0,0 +1,478 @@
...
+		return this._view;
+	}
+
+	override dispose(): void {
+		// Remove from parent window
+		this._window?.win?.contentView.removeChildView(this._view);
+
+		// Clean up the view and all its event listeners
+		// Note: webContents.close() automatically removes all event listeners
+		this._view.webContents.close();
```
deepak1556: 1) This is not synchronous, you should wait till [destroyed](https://github.com/electron/electron/blob/main/docs/api/web-contents.md#event-destroyed) event is called ensure successful termination of the underlying content
2) The unload handler for pages inside `WebContentsView` will be called but the runtime won't respect to prevent the unload https://github.com/electron/electron/blob/main/docs/api/web-contents.md#event-will-prevent-unload, fyi
3) What about handling `render-process-gone`, `unresponsive` and `responsive` events to provide appropriate dialogs to reload, close or wait depending on the event ?

kycutler: * Updated this to pass `waitForBeforeUnload: false` so there's nothing to wait on here. Also updated so `will-prevent-unload` is not respected in general (i.e. unload is never prevented). We should revisit this in the future, but it requires some care for requiring interaction / blocking multiple dialogs / etc. to prevent abuse.
* Added handling for `render-process-gone` but will hold off on `unresponsive` for now. Though I did validate that the behavior there is reasonable -- e.g. the rest of the UI is still responsive, and closing the tab kills the process.

Regarding your other comment below:
* Updated to make `global` the default, and to always use `ephemeral` if the workspace is not trusted.
* Added commands to clear all data for global and workspace sessions
* Added explicit permission handling with the same allowed permissions as webviews (`pointerLock`, `notifications`, `clipboard-read`, `clipboard-sanitized-write`)

Thanks!

src/vs/platform/browserView/electron-main/browserView.ts
```diff
@@ -0,0 +1,530 @@
...
+
+					return `data:${type};base64,${Buffer.from(buffer).toString('base64')}`;
+				})();
+
+				this._faviconRequestCache.set(url, request);
+
+				try {
+					this._lastFavicon = await request;
+					this._onDidChangeFavicon.fire({ favicon: this._lastFavicon });
+					// On success, leave the promise in the cache and stop looping
```
deepak1556, line 150: why cache the promise instead of the response string ? Maybe I am missing something but `_faviconRequestCache` is populated and consumed in the same suspended function.

deepak1556 (commented):
LGTM for session, permission handling, browserview lifetime, webPreferences and storage cleanups

src/vs/platform/browserView/electron-main/browserView.ts
kycutler: It's so we don't make duplicate requests if the event fires again before the favicon finishes loading

src/vs/platform/browserView/electron-main/browserViewMainService.ts
bpasero: Ok lets leave it for now.

src/vs/workbench/contrib/browserView/electron-browser/browserEditorInput.ts
bpasero: Ok lets follow up on this separately.

src/vs/platform/browserView/electron-main/browserView.ts
deepak1556: As I see the cache is only used for a given BrowserView (not shared across BrowserViews), the event will only be emitted when the main page of the BrowserView stopped loading (it won't be emitted for child frame navigations). So in that case wouldn't it be enough to cache the response rather ? Why would the event re-enter before it has finished processing the requests it is waiting on ?

bpasero (dismissed):
@kycutler I went over the changes high level and think we can go ahead and merge this in, provided all feedback from @deepak1556 is resolved. I am mainly looking at how well it integrates into our contributions story and it seems to do that well, following how other components do this. This is a non trivial change considering that it also touches on `electron-main` but I do not see red flags from my basic review 🙏 

As you might have seen, `EditorInput` has some methods (some optional) to drive how it behaves, so maybe check over the API here to see if you want to adopt more things (close handler e.g. is used to bring up a dialog when you close, not sure this ever applies):

https://github.com/microsoft/vscode/blob/077294bf9c99b2092e798d92e0ca1fde88a67bef/src/vs/workbench/common/editor/editorInput.ts#L61

Here are things to consider next based on my testing:

- the loading animation for browser editor tab seem somewhat non-standard compared to our usual spinner. maybe you can get help from @mrleemurray on this and see if we could make it consistent: [1]

- I cannot drag the browser tab over itself to split the editor group into half, something I can do with normal editor tabs. I am not sure how you decide to show the overlay today, but I assume for DND it would need to also be shown: [2]

- you are duplicating the editor toolbar with specific actions and I wonder if these should move into the editor toolbar for consistency, maybe at least on the right hand side? see also how they are not well aligned: [3]

- I find the blurry overlay not matching our usual VS Code style, could we just simply not blur it? For a user it might come at a surprise because other editors are not blurry like that and it does not really seem to add much value?: [4]

[1]
https://github.com/user-attachments/assets/3c16e392-d5a1-4f46-8cb4-6682b3688870

[2]
https://github.com/user-attachments/assets/dcce5152-4c6c-404a-ad32-7610460827d4

[3]
image

[4]
image

src/vs/workbench/contrib/browserView/electron-browser/browserEditorInput.ts
```diff
@@ -0,0 +1,239 @@
...
+					void this._model.setVisible(false);
+				} else {
+					this._model.dispose();
+					this._model = undefined;
+				}
+			}
+		}));
+	}
+
+	public get id() {
```
bpasero: @kycutler I notice that you sometimes seem to be using `public` but most often leave it away, can we make it consistently left out?

src/vs/workbench/contrib/browserView/electron-browser/browserEditorInput.ts
```diff
@@ -0,0 +1,239 @@
...
+		const authority = URI.parse(url).authority;
+		return authority || BrowserEditorInput.DEFAULT_LABEL;
+	}
+
+	override getDescription(): string | undefined {
+		// Use model URL if available, otherwise fall back to initial data
+		return this._model ? this._model.url : this._initialData.url;
+	}
+
+	override canReopen(): boolean {
```
bpasero, line 171: @kycutler fyi I added this because this threw an error:
* open a browser editor
* close it
* select "Reopen closed editor"

src/vs/platform/browserView/electron-main/browserViewMainService.ts
```diff
@@ -0,0 +1,231 @@
...
+
+	/**
+	 * Check if a webContents belongs to an integrated browser view
+	*/
+	private static readonly knownSessions = new WeakSet<Electron.Session>();
+	static isBrowserViewWebContents(contents: Electron.WebContents): boolean {
+		return BrowserViewMainService.knownSessions.has(contents.session);
+	}
+
+	private readonly browserViews = new Map<string, BrowserView>();
```
bpasero: @kycutler fyi we have a `DisposableMap` that would come in handy here.

src/vs/workbench/contrib/browserView/electron-browser/browserEditorInput.ts
kycutler: This actually exposed some bugs -- updated to fix those, reopening works now 🙂

src/vs/workbench/contrib/browserView/electron-browser/browserEditorInput.ts
kycutler: Good callout, old habits 😄 updated

src/vs/base/common/keyCodes.ts
```diff
@@ -755,6 +756,9 @@ for (let i = 0; i <= KeyCode.MAX_VALUE; i++) {
 		if (eventKeyCode) {
 			EVENT_KEY_CODE_MAP[eventKeyCode] = keyCode;
 		}
+		if (scanCodeStr) {
```
TylerLeonhardt, line 759: for such a low level change, it'd be good to have a test.

src/vs/platform/browserView/common/browserViewUri.ts
```diff
@@ -0,0 +1,62 @@
...
+ *--------------------------------------------------------------------------------------------*/
+
+import { Schemas } from '../../../base/common/network.js';
+import { URI } from '../../../base/common/uri.js';
+import { generateUuid } from '../../../base/common/uuid.js';
+
+/**
+ * Helper for creating and parsing browser view URIs.
+ */
+export namespace BrowserViewUri {
```
TylerLeonhardt, line 13: This also seems like a good class for unit tests

roblourens (approved):

DonJayamanne (approved):

PR #278476, opened 2025-11-20
Manage ChatModels with ref-counting, allow running in bg

roblourens:
Towards #277318

PR #278683, opened 2025-11-21
chat: refactor input state to be central

connor4312:
This is a big PR that centralizes input state into an aptly named `inputState` on the model. This allows state to be passed around and transferred more completely and consistently.

Model, agent, input, and selection is transferred between panel chat, editor chat, and even quick chat, and also in session history and in sessions transferred between workspaces.

More stuff can be cleaned up after #278476 is in because there is still some explicit state passing to make it work with empty sessions.

I also want to do some work on history cleanup/unification. Additionally, this code basically syncs UI state back and forth with the input model, but I want to do a little more work to actually have the model as the point of truth for some of this data. For example, instead of updating your model in the chat picker which triggers and update to save state in the model, instead have the picker take an `IObservable<ChatModel>` and emit an event to update the chat model and observe that change.

But that should be a less wide-reaching change.

cc @roblourens

refs #277318


dmitrivMS (approved):

joaomoreno:
FYI this broke the chat input as the tools were suddenly missing:

image

@aeschli fixed it with https://github.com/microsoft/vscode/pull/278761

connor4312:
Ah, sorry 🙈  I think the change Martin made in that PR is an overall improvement

PR #278692, opened 2025-11-21
Optimize initial dev build with esbuild for transpiling

bryanchen-d:

### DIFF
--- a/build/gulpfile.mjs
+++ b/build/gulpfile.mjs
@@ -8,6 +8,8 @@ import glob from 'glob';
 import gulp from 'gulp';
 import { createRequire } from 'node:module';
 import { fileURLToPath } from 'url';
+import fancyLog from 'fancy-log';
+import ansiColors from 'ansi-colors';
 import { monacoTypecheckTask /* , monacoTypecheckWatchTask */ } from './gulpfile.editor.mjs';
 import { compileExtensionMediaTask, compileExtensionsTask, watchExtensionsTask } from './gulpfile.extensions.mjs';
 import compilation from './lib/compilation.js';
@@ -26,7 +28,7 @@ gulp.task(compileApiProposalNamesTask);
 gulp.task(watchApiProposalNamesTask);
 
 // SWC Client Transpile
-const transpileClientSWCTask = task.define('transpile-client-esbuild', task.series(util.rimraf('out'), transpileTask('src', 'out', true)));
+const transpileClientSWCTask = task.define('transpile-client-esbuild', task.series(util.rimraf('out'), transpileTask('src', 'out', { esbuild: true, logTopic: 'src-transpile' })));
 gulp.task(transpileClientSWCTask);
 
 // Transpile only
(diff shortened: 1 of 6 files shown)


build/lib/compilation.ts
```diff
@@ -47,24 +47,31 @@ interface ICompileTaskOptions {
...
+export function createCompile(src: string, { build, emitError, transpileOnly, preserveEnglish, noEmit, logTopic }: ICompileTaskOptions) {
 	const tsb = require('./tsb') as typeof import('./tsb');
 	const sourcemaps = require('gulp-sourcemaps') as typeof import('gulp-sourcemaps');
+	const currentReporter = logTopic ? createReporter(logTopic) : reporter;
 
 
 	const projectPath = path.join(__dirname, '../../', src, 'tsconfig.json');
 	const overrideOptions = { ...getTypeScriptCompilerOptions(src), inlineSources: Boolean(build) };
+	if (noEmit) {
+		overrideOptions.noEmit = true;
```
jrieken, line 63: I didn't test this yet but doesn't `noEmit` also mean no `.d.ts` files are created and therefore the impact of a code change cannot be assessed correctly anymore? E.g did its d.ts change and hence more checking is need? 

I think we need to check the `noEmit` flag where we send the JS files into the stream

build/gulpfile.mjs
```diff
@@ -37,7 +39,17 @@ gulp.task(transpileClientTask);
 const compileClientTask = task.define('compile-client', task.series(util.rimraf('out'), compileApiProposalNamesTask, compileTask('src', 'out', false)));
 gulp.task(compileClientTask);
 
-const watchClientTask = task.define('watch-client', task.series(util.rimraf('out'), task.parallel(watchTask('out', false), watchApiProposalNamesTask)));
+const watchClientEsbuildTask = task.define('watch-client-esbuild', watchTask('out', false, 'src', { useEsbuild: true, logTopic: 'src-transpile', skipInitial: true }));
```
jrieken, line 42: Don't remove the `watch-client` task yet, instead have both

build/gulpfile.mjs
```diff
@@ -37,7 +39,17 @@ gulp.task(transpileClientTask);
...
+const watchClientCheckTask = task.define('watch-client-check', watchTask('out', false, 'src', { useEsbuild: false, noEmit: true, logTopic: 'src-check' }));
+const watchClientTask = task.define('watch-client', task.series(
+	util.rimraf('out'),
+	task.parallel(transpileClientSWCTask, compileApiProposalNamesTask),
+	task.define('client-ready', () => {
+		fancyLog(ansiColors.bgGreen.black('*** Finished transpiling the client. Ready to launch for debugging! *** '));
+		return Promise.resolve();
+	}),
+	task.parallel(watchClientEsbuildTask, watchClientCheckTask, watchApiProposalNamesTask)
```
jrieken, line 51: Why do we need first `parallel(transpile, compile)` and why isn't it just `parallel(watch, watch, watch)`? Shouldn't the first watch-transpile do all files and be fast from then one?

jrieken (changes requested):
Left some feedback, I think the noEmit is severe. Also note that we don't have the JS-files anymore, so you can delete them

PR #278699, opened 2025-11-21
Support ISO 8601 in terminal link

### LINKED ISSUE #247568: Terminal Ctrl+Click on a file with colon in filename does not open the file, preceding zeroes are deleted

- VS Code Version: 1.99.3
- OS Version: Arch Linux

Steps to Reproduce:

1. `echo hello > test-$(date -Is).log`
2. `echo ./test-2025-04-28T11:03:09+02:00.log`
3. `echo test-2025-04-28T11:03:09+02:00.log`

ctrl+click in the terminal. it will show "./test-2025-04-28T11:3:9 no matching results" instead of opening the file

Image

Interestingly, if you echo WITHOUT `./` prefix, and WITHOUT numbers (echo `test-2025-04-28T11:19:24+02:00.log`) it works fine, but only if the other file does not exist (prefix matching?)

There is some semi-related issues ( https://github.com/microsoft/vscode/issues/193629 ) but I didn't see any where colons are part of the filename and thus misdetected.

anthonykim1:
Resolves: https://github.com/microsoft/vscode/issues/247568

### DIFF
--- a/src/vs/workbench/contrib/terminalContrib/links/browser/terminalLinkParsing.ts
+++ b/src/vs/workbench/contrib/terminalContrib/links/browser/terminalLinkParsing.ts
@@ -90,8 +90,7 @@ function generateLinkSuffixRegex(eolOnly: boolean) {
 		// "foo",339.12
 		// "foo",339.12-789
 		// "foo",339.12-341.789
-		// Negative lookbehind, support ISO 8601   [#247568]
-		`(?<!\\d{4}-\\d{2}-\\d{2}T\\d{1,2}(?::\\d{1,2}){0,2}(?:[+-]\\d{2})?)(?::|#| |['"],|, )${r()}([:.]${c()}(?:-(?:${re()}\\.)?${ce()})?)?` + eolSuffix,
+		`(?::|#| |['"],|, )${r()}([:.]${c()}(?:-(?:${re()}\\.)?${ce()})?)?` + eolSuffix,
 		// The quotes below are optional           [#171652]
 		// "foo", line 339                         [#40468]
 		// "foo", line 339, col 12
--- a/src/vs/workbench/contrib/terminalContrib/links/browser/terminalLinkParsing.ts
+++ b/src/vs/workbench/contrib/terminalContrib/links/browser/terminalLinkParsing.ts
@@ -90,7 +90,8 @@ function generateLinkSuffixRegex(eolOnly: boolean) {
 		// "foo",339.12
 		// "foo",339.12-789
 		// "foo",339.12-341.789
-		`(?::|#| |['"],|, )${r()}([:.]${c()}(?:-(?:${re()}\\.)?${ce()})?)?` + eolSuffix,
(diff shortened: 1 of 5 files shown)


src/vs/workbench/contrib/terminalContrib/links/browser/terminalLinkParsing.ts
Tyriar, line 1: We'd want unit tests for the link formats you're trying to cover

src/vs/workbench/contrib/terminalContrib/links/browser/terminalLinkParsing.ts
```diff
@@ -90,7 +90,8 @@ function generateLinkSuffixRegex(eolOnly: boolean) {
 		// "foo",339.12
 		// "foo",339.12-789
 		// "foo",339.12-341.789
-		`(?::|#| |['"],|, )${r()}([:.]${c()}(?:-(?:${re()}\\.)?${ce()})?)?` + eolSuffix,
+		// Negative lookbehind, support ISO 8601   [#247568]
+		`(?<!\\d{4}-\\d{2}-\\d{2}T\\d{1,2}(?::\\d{1,2}){0,2}(?:[+-]\\d{2})?)(?::|#| |['"],|, )${r()}([:.]${c()}(?:-(?:${re()}\\.)?${ce()})?)?` + eolSuffix,
```
Tyriar: This specific regex handles links with at least a line and at most a line start/end and column start/end. This isn't the right place to fix this since the example `test-2025-04-28T11:03:09+02:00.log` doesn't actually contain a line number.

Tyriar (dismissed):

Tyriar:
Looking a little closer so it's working find without the `./` as mentioned in the report. This is because the full link text is being included as a link candidate along side the ones with the badly detected line/col numbers. So first thing I would investigate is why the full file isn't being parsed out as a link candidate when the `./` prefix is included.

The fallback case where the quick pick shows up (search opener) isn't as important since if this is a file it should have been validated.

src/vs/workbench/contrib/terminalContrib/links/browser/terminalLocalLinkDetector.ts
```diff
@@ -101,11 +101,25 @@ export class TerminalLocalLinkDetector implements ITerminalLinkDetector {
 				continue;
 			}
 
+			// https://github.com/microsoft/vscode/issues/247568
+			// If the path ends with an ISO 8601 timestamp pattern and has a suffix,
+			// the suffix is likely part of the timestamp, not a line:col. Extend the path to include it.
+			let pathForCandidate = parsedLink.path.text;
+			let extendedLength = 0;
+			if (parsedLink.suffix && parsedLink.path.text.match(/\d{4}-\d{2}-\d{2}T\d{2}$/)) {
```
anthonykim1: @Tyriar Trying different approach.. I'm kind of worried regex match and substring would be expensive, since these detections happen pretty frequently when people hover around terminal.

lszomoru (approved):

PR #278758, opened 2025-11-21
Refactor watermark entry management and improve styling

mrleemurray:
Simplify the management of watermark entries and enhance the layout styling for better visual presentation. 

Adjustments include changes to the structure and CSS for improved alignment and spacing.


bpasero (changes requested):
See Slack

bpasero (changes requested):
When a user disables AI features or runs in vscode.dev/github.dev there is only 1 entry showing up which looks very sparse.

image


Also, when you are in an empty window, you get a lot more entries, is that still intentional?

image

bpasero (approved):

lszomoru (approved):

PR #278811, opened 2025-11-21
Add delegate button on untitled prompt files

joshspicer:
ref https://github.com/microsoft/vscode-internalbacklog/issues/6311


Tyriar (dismissed):

src/vs/workbench/contrib/chat/browser/actions/chatContinueInAction.ts
```diff
@@ -148,10 +176,25 @@ export class ChatContinueInSessionActionItem extends ActionWidgetDropdownActionV
...
-		return super.renderLabel(element);
+			const iconSpan = document.createElement('span');
+			iconSpan.classList.add(...ThemeIcon.asClassNameArray(Codicon.forward));
+			container.appendChild(iconSpan);
+
+			const textSpan = document.createElement('span');
+			textSpan.textContent = localize('delegate', "Delegate to...");
+			container.appendChild(textSpan);
+
+			element.appendChild(container);
```
Tyriar, line 191: Consider using the `h` helper, this can be made a lot simpler with that

Tyriar (approved):

PR #278831, opened 2025-11-21
Fix MCP registry servers being disabled on restart

### LINKED ISSUE #278830: MCP registry servers disabled on restart
# MCP Registry Servers Disabled on Restart

## Issue Description

MCP servers installed from a custom registry are incorrectly marked as "disabled" after restarting VSCode, even though they were installed from an authorized registry source.

## Steps to Reproduce

1. Configure VSCode to use a custom MCP registry:
   ```json
   {
     "chat.mcp.gallery.enabled": true,
     "chat.mcp.gallery.serviceUrl": "http://localhost:9000/mcp_registry",
     "chat.mcp.access": "registry"
   }
   ```

2. Install an MCP server from the registry
3. Verify the server works correctly
4. Restart VSCode
5. **Bug**: Server shows as "disabled" with message about access restrictions

## Expected Behavior

MCP servers installed from the registry should remain enabled after restart, since they came from an authorized source.

## Actual Behavior

Registry-installed servers are marked as disabled on every restart, requiring manual re-enabling.

## Root Cause

Race condition during VSCode startup:
1. Local MCP configurations load from settings
2. Enablement checks run (`getRuntimeStatus()` and `getEnablementStatus()`)
3. Gallery metadata fetches from registry (async)
4. Servers register in `mcpService.servers` collection

The enablement checks (step 2) run **before** gallery metadata loads (step 3), causing:
- `mcpServer.gallery` is `false` during the check window
- Servers incorrectly identified as non-registry servers
- Access control logic disables them

## Environment

- **VS Code Version**: 1.107.0 (or any version with MCP support)
- **OS**: macOS (reproducible on all platforms)
- **MCP Access Setting**: `"registry"`
- **Custom Registry**: Any custom MCP registry URL

## Impact

- **Severity**: High - Breaks user experience for custom registry users
- **Workaround**: Manually re-enable servers after each restart (poor UX)
- **Affected Users**: Anyone using custom MCP registries with `chat.mcp.access: "registry"`

## Additional Context

This issue only affects custom registries because the official Microsoft registry may have different timing characteristics. The bug is in the client-side enablement logic, not the registry API.

abhishekgupta-lab:
Fixes #278830

Registry-installed MCP servers are incorrectly marked as disabled after restarting VSCode due to a race condition during startup where enablement checks run before gallery metadata has loaded from the registry.

This fix addresses two issues:

1. Skip the mcpService.servers collection check for gallery servers during startup to avoid a circular dependency. Gallery servers are managed by the InstalledMcpServersDiscovery system which registers them as collections after they're enabled.

2. Use the persistent local.galleryUrl marker instead of installable to identify registry servers during the access control check. The galleryUrl property is set during gallery installation and persists across restarts, providing a reliable indicator even before gallery metadata loads.

Security: The fix maintains all security guarantees. Only servers with galleryUrl (set during official gallery installation) are allowed when chat.mcp.access is set to "registry". Manual/unauthorized servers without this marker are correctly disabled.

### DIFF
--- a/src/vs/workbench/contrib/mcp/browser/mcpWorkbenchService.ts
+++ b/src/vs/workbench/contrib/mcp/browser/mcpWorkbenchService.ts
@@ -742,6 +742,14 @@ export class McpWorkbenchService extends Disposable implements IMcpWorkbenchServ
 			return enablementStatus;
 		}
 
+		// Don't mark registry/gallery servers as disabled during startup when collections haven't been loaded yet.
+		// Gallery servers are managed by the InstalledMcpServersDiscovery system which registers them as
+		// collections after they're enabled. Checking mcpService.servers before collections are registered
+		// creates a circular dependency that prevents registry servers from being enabled on restart.
+		if (mcpServer.gallery) {
+			return undefined;
+		}
+
 		if (!this.mcpService.servers.get().find(s => s.definition.id === mcpServer.id)) {
 			return { state: McpServerEnablementState.Disabled };
 		}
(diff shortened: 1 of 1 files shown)


sandy081:
Discussed with @abhishekgupta-lab over teams call and we got to an understanding. Hence closing this.

PR #278858, opened 2025-11-21
Add better timing info to chat sessions

### LINKED ISSUE #278567: Agent sessions: provide `inProgressTime` and `finishedOrFailedTime` from provider
**Usecase:** show how long the current task is running (if in-progress) and when the last response finished.

Today we do tracking of these properties in-memory:

https://github.com/microsoft/vscode/blob/5942390add05dddd450fe83bc4da636a77deae0a/src/vs/workbench/contrib/chat/browser/agentSessions/agentSessionViewModel.ts#L217-L225

We need to get this data from the provider to show it properly. This needs support for that info from the provider API.

//cc @mjbvz @osortega @rebornix @DonJayamanne

mjbvz:
Fixes https://github.com/microsoft/vscode/issues/278567


src/vs/workbench/contrib/chat/browser/agentSessions/agentSessionsModel.ts
```diff
@@ -376,9 +370,10 @@ class AgentSessionsCache {
 				status: session.status,
 				archived: session.archived,
 
-				timing: {
-					startTime: session.timing.startTime,
-					endTime: session.timing.endTime,
+				timing: hasKey(session.timing, { created: true }) ? session.timing : {
```
bpasero: Are you doing this to preserve some old data the has the previous shape? I would prefer we would change the storage key simply, I do not think a migration is needed from the previous data.

bpasero: Btw I also think @osortega started to implement some kind of caching for the classic view.

src/vs/workbench/contrib/chat/browser/agentSessions/agentSessionsModel.ts
```diff
@@ -344,10 +341,7 @@ class AgentSessionsCache {
 				status: session.status,
 				archived: session.archived,
 
-				timing: {
-					startTime: session.timing.startTime,
-					endTime: session.timing.endTime,
-				},
+				timing: session.timing,
```
bpasero: We have to be careful here to not include the synthetic properties `inProgressTime` and `finishedOrFailedTime` which are only generated by knowing state transitions in memory. We cannot have them restored from storage as then they would get out of sync.

Once all providers adopted the new API we can actually remove these propertis as they are just workarounds for not having the very API you are now adding.

bpasero (changes requested):

src/vs/workbench/contrib/chat/common/chatSessionsService.ts
```diff
@@ -69,8 +69,9 @@ export interface IChatSessionItem {
 	status?: ChatSessionStatus;
 	tooltip?: string | IMarkdownString;
 	timing: {
-		startTime: number;
-		endTime?: number;
+		created: number | undefined; // Old sessions may not have this set
```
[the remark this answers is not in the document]
bpasero: This?

src/vs/workbench/api/common/extHostChatSessions.ts
```diff
@@ -166,15 +166,19 @@ export class ExtHostChatSessions extends Disposable implements ExtHostChatSessio
...
 			label: sessionContent.label,
 			description: sessionContent.description ? typeConvert.MarkdownString.from(sessionContent.description) : undefined,
 			status: this.convertChatSessionStatus(sessionContent.status),
 			tooltip: typeConvert.MarkdownString.fromStrict(sessionContent.tooltip),
 			timing: {
-				startTime: sessionContent.timing?.startTime ?? 0,
-				endTime: sessionContent.timing?.endTime
+				created: timing?.created ?? timing?.startTime ?? 0,
+				lastRequestStarted: timing?.lastRequestStarted ?? undefined,
+				lastRequestEnded: timing?.lastRequestEnded ?? timing.endTime ?? undefined,
```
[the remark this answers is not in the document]
bpasero: This?

src/vs/workbench/contrib/chat/browser/chatSessions/view/sessionsTreeRenderer.ts
```diff
@@ -333,16 +333,17 @@ export class SessionsRenderer extends Disposable implements ITreeRenderer<IChatS
 		}
 
 		// Handle timestamp display and grouping
-		const hasTimestamp = session.timing?.startTime !== undefined;
+		const startTime = session.timing.lastRequestStarted ?? session.timing.created;
```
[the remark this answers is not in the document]
bpasero: This?

src/vs/workbench/contrib/chat/browser/chatSessions/common.ts
```diff
@@ -20,6 +20,8 @@ export const NEW_CHAT_SESSION_ACTION_ID = 'workbench.action.chat.openNewSessionE
 
 export type ChatSessionItemWithProvider = IChatSessionItem & {
 	readonly provider: IChatSessionItemProvider;
+
+	// no :(
```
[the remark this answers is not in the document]
bpasero: This?

src/vs/workbench/api/common/extHostChatSessions.ts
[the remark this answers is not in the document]
mjbvz: Good catch and very confusing TS behavior here. Rewriting this to use a safer type assertion

src/vs/workbench/contrib/chat/browser/agentSessions/agentSessionsModel.ts
mjbvz: Ok will take a look at invaliding the old cache in this case

src/vs/workbench/contrib/chat/browser/chatSessions/view/sessionsTreeRenderer.ts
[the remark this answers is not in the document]
mjbvz: This seems incorrect since we want a `startTime`. I believe  this should either be the last request start time or the creation time

src/vs/workbench/contrib/chat/common/chatSessionsService.ts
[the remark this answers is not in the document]
mjbvz: Intentional for now. I want to make sure we don't assume that all sessions will have this property set

src/vs/workbench/contrib/chat/browser/agentSessions/agentSessionsModel.ts
```diff
@@ -404,14 +400,18 @@ class AgentSessionsCache {
...
+					lastRequestStarted: session.timing.lastRequestStarted,
+					lastRequestEnded: session.timing.lastRequestEnded,
 				},
 
 				statistics: session.statistics,
 			}));
 
-		this.storageService.store(AgentSessionsCache.SESSIONS_STORAGE_KEY, JSON.stringify(serialized), StorageScope.WORKSPACE, StorageTarget.MACHINE);
+		this.storageService.store(AgentSessionsCache.SESSIONS_STORAGE_KEY, JSON.stringify({
+			version: AgentSessionsCache.SESSIONS_STORAGE_VERSION,
```
mjbvz: @bpasero If you prefer I can instead update the key we use to store the sessions so that we don't need a version check here

src/vs/workbench/contrib/chat/browser/agentSessions/agentSessionsViewer.ts
```diff
@@ -243,7 +244,7 @@ export class AgentSessionRenderer implements ICompressibleTreeRenderer<IAgentSes
 			}
 
 			if (!timeLabel) {
-				timeLabel = fromNow(session.timing.endTime || session.timing.startTime, true);
+				timeLabel = fromNow(session.timing.lastRequestEnded || session.timing.lastRequestStarted || session.timing.created || 0, true);
```
bpasero: How can the time ever be `undefined` though? I would prefer we can ensure the API is sound around this and never allows `undefined`, otherwise we will play catch up game with bad behaving providers.

src/vs/workbench/api/common/extHostChatAgents2.ts
```diff
@@ -601,6 +601,12 @@ export class ExtHostChatAgents2 extends Disposable implements ExtHostChatAgentsS
 					chatSessionItem: {
 						resource: URI.revive(context.chatSessionContext.chatSessionResource),
 						label: context.chatSessionContext.isUntitled ? 'Untitled Session' : 'Session',
+						// TODO@mjbvz: what is the correct value for these?
```
bpasero, line 605: Can we resolve this in this PR, do we need more proposed API? can you explain?

src/vs/workbench/contrib/chat/browser/agentSessions/agentSessionsModel.ts
bpasero: Yes that sounds good to me, I do not think we need to migrate. Sorry for not having made that clear before. Use `agentSessions.sessions.cache` as key.

bpasero (changes requested):

PR #278874, opened 2025-11-21
Chat view: introduce session title and back button (fix #277537)

### LINKED ISSUE #277537: Chat view: introduce session title and back button
**Replace the “Chat” header in active sessions with the session title, matching how titles appear in chat history.** 

Showing the session title directly in the header helps users stay oriented across multiple types of chat sessions and reduces cognitive load by making each session identifiable at a glance. This mirrors how users already identify sessions in the chat sessions view, carrying that context into the active sessions.

| Current State      | Suggested State |
| ----------- | ----------- |
| Image      | Image       |

hawkticehurst:
Resolves #277537 

Changes include:

- When activity bar is in Default position, a new "secondary title bar" is appended below the primary title bar
  - When the Show Labels configuration is set, the secondary title will use the format: "CUSTOM TITLE"
  - When the Show Icons configuration is set, the secondary title will use the format: "CHAT: CUSTOM TITLE"
  - When multiple views are rendered in the chat view pane the secondary bar will not be rendered and instead follow the same conventions used for Top, Bottom, and Hidden positions (listed below)
- When activity bar is in Top, Bottom, or Hidden positions the primary title will be updated (no secondary bar rendered)
  - The primary title will use the format: "CHAT: CUSTOM TITLE"
- When waiting for a new chat title to be generated the secondary bar will not be rendered and/or the CHAT label will be the fallback default text
- An arrow icon button is included next to generated title that will bring you back to the empty chat view state

**_Default position with labels (secondary title bar IS rendered WITHOUT chat prefix):_**

Screenshot 2025-11-21 at 7 30 48 PM

**_Default position with icons (secondary title bar IS rendered WITH chat prefix):_**

Screenshot 2025-11-21 at 7 30 59 PM

**_Top position (secondary title bar IS NOT rendered):_**

Screenshot 2025-11-21 at 7 39 18 PM

**_Bottom/Hidden positions (secondary title bar IS NOT rendered):_**

Screenshot 2025-11-21 at 7 39 10 PM

**_Default, Top, Bottom/Hidden positions with multiple views (secondary title bar IS NOT rendered):_**

Screenshot 2025-11-21 at 7 32 26 PM

Screenshot 2025-11-21 at 7 33 11 PM

Screenshot 2025-11-21 at 7 33 29 PM


src/vs/workbench/contrib/chat/browser/chatViewPane.ts
```diff
@@ -192,6 +204,16 @@ export class ChatViewPane extends ViewPane implements IViewWelcomeDelegate {
 	protected override async renderBody(parent: HTMLElement): Promise<void> {
 		super.renderBody(parent);
 
+		this._secondaryTitleContainer = $('.chat-secondary-title', { style: 'display: none; padding: 4px 16px 8px 16px; border-bottom: 1px solid var(--vscode-sideBarSectionHeader-border); align-items: center;' });
```
hawkticehurst: @roblourens wasn't sure how the team feels about using inline styles versus creating a new css file. Opted to not create a new file but let me know if you think I should change that.

src/vs/workbench/contrib/chat/browser/chatViewPane.ts
```diff
@@ -45,6 +51,7 @@ interface IViewPaneState extends IChatViewState {
 }
 
 export const CHAT_SIDEBAR_PANEL_ID = 'workbench.panel.chat';
+export const ChatSecondaryTitleMenuId = new MenuId('ChatSecondaryTitle');
```
roblourens: All MenuIds are centralized here https://github.com/microsoft/vscode/blob/d302bf002e54cce6e2230eeb7313b65b9a783d92/src/vs/platform/actions/common/actions.ts#L287

src/vs/workbench/contrib/chat/browser/chatViewPane.ts
roblourens: I would prefer a real css file

src/vs/workbench/contrib/chat/browser/chatViewPane.ts
```diff
@@ -317,4 +351,56 @@ export class ChatViewPane extends ViewPane implements IViewWelcomeDelegate {
...
+			this.layoutBody(this._lastLayout.height, this._lastLayout.width);
+		}
+	}
 }
+
+registerAction2(class BackToChatAction extends Action2 {
+	constructor() {
+		super({
+			id: 'workbench.action.chat.backToChat',
+			title: localize('backToChat', "Back to Chat"),
```
roblourens: "Back to chat" feels off, aren't you currently on a chat?

src/vs/workbench/contrib/chat/browser/chatViewPane.ts
```diff
@@ -185,6 +204,16 @@ export class ChatViewPane extends ViewPane implements IViewWelcomeDelegate {
...
 		super.renderBody(parent);
 
+		this._secondaryTitleContainer = $('.chat-secondary-title', { style: 'display: none; padding: 4px 16px 8px 16px; border-bottom: 1px solid var(--vscode-sideBarSectionHeader-border); align-items: center;' });
+		this._register(this.instantiationService.createInstance(MenuWorkbenchToolBar, this._secondaryTitleContainer, ChatSecondaryTitleMenuId, {
+			menuOptions: {
+				shouldForwardArgs: true
+			}
+		}));
+
+		this._secondaryTitle = $('span', { style: 'margin-left: 4px;' });
```
roblourens: This should hide when a welcome view is shown (these are a bit weird and inconsistent because I think people forgot they exist)

roblourens: image

src/vs/workbench/contrib/chat/browser/chatViewPane.ts
```diff
@@ -317,4 +351,56 @@ export class ChatViewPane extends ViewPane implements IViewWelcomeDelegate {
...
+		} else {
+			this._secondaryTitleContainer.style.display = 'none';
+		}
+		if (this._lastLayout) {
+			this.layoutBody(this._lastLayout.height, this._lastLayout.width);
+		}
+	}
 }
+
+registerAction2(class BackToChatAction extends Action2 {
```
roblourens: This can go with other actions in one of the chatActions.ts files

src/vs/workbench/contrib/chat/browser/chatViewPane.ts
```diff
@@ -312,13 +344,63 @@ export class ChatViewPane extends ViewPane implements IViewWelcomeDelegate {
...
+			// Assign all props to the memento so they get saved
+			(this.viewState as Record<string, unknown>)[key] = value;
+		}
+	}
+
+	private updateViewTitle(model?: IChatModel): void {
+		model = model ?? this._widget.viewModel?.model;
+		const title = (model && model.hasCustomTitle) ? model.title : localize('chat', "Chat");
+
+		this.updateTitle(localize('chat', "Chat"));
```
roblourens: Why is this updateTitle call needed?

src/vs/workbench/contrib/chat/browser/chatViewPane.ts
```diff
@@ -317,4 +351,56 @@ export class ChatViewPane extends ViewPane implements IViewWelcomeDelegate {
...
+		if (model && model.hasCustomTitle) {
+			this.setSecondaryTitle(title);
+		} else if (model && model.getRequests().length > 0) {
+			this.setSecondaryTitle(localize('generatingTitle', "Generating title..."));
```
roblourens: I have some saved chats that didn't get a custom title generated. That can happen for different reasons, like the request might fail, or maybe I was just debugging, but then we end up showing "generating title" forever. In that case we still show the request text as the chat title, we should show that here too. This state could probably look nicer, maybe it can show the request text instead, or "new chat", what do you think?

src/vs/workbench/contrib/chat/browser/chatViewPane.ts
hawkticehurst: Same, this was just temporary filler text. I sent a message to Eli earlier asking for his thoughts on what we should put here and am waiting to hear back.

src/vs/workbench/contrib/chat/browser/chatViewPane.ts
hawkticehurst: It does nothing right now, but it's the initial start to the logic that will replace the primary title (i.e. CHAT) with the AI generated title when the activity bar is in Top, Bottom, or Hidden positions

src/vs/workbench/contrib/chat/browser/chatViewPane.ts
hawkticehurst: Talked with Eli and the "solution" we landed on is to actually not include the arrow back button for now. We'll revisit once the agent sessions view and new chat views are unified / rationalized a bit better.

src/vs/workbench/contrib/chat/browser/chatViewPane.ts
hawkticehurst: Decided to remove the "Generating title..." text and just have the fallback always be "CHAT" so I'm hoping this issue is now resolved. That said, can you tell me how to get into this welcome view state? I want to make sure the secondary title bar isn't being rendered here.

src/vs/workbench/contrib/chat/browser/chatViewPane.ts
hawkticehurst: As mentioned elsewhere, I removed the back to chat action, but will keep this in mind if/when we add it back :)

src/vs/workbench/contrib/chat/browser/chatViewPane.ts
hawkticehurst: How do things look now with the new changes?

src/vs/workbench/contrib/chat/browser/chatViewPane.ts
roblourens: You can disable wifi and reload the window (I think this used to always show up, now it needs a window reload...)

src/vs/workbench/contrib/chat/browser/chatViewPane.ts
```diff
@@ -317,4 +365,123 @@ export class ChatViewPane extends ViewPane implements IViewWelcomeDelegate {
...
+		} else {
+			this.setSecondaryTitle(undefined);
+		}
+	}
+
+	private withChatPrefix(title: string | undefined): string | undefined {
+		if (!title) {
+			return undefined;
+		}
+		return title.startsWith('Chat: ') ? title : `Chat: ${title}`;
```
roblourens: When does the chatmodel title start with 'Chat: '? Would that be a localized string, and should the 'Chat: ' in this method be localized?

src/vs/workbench/contrib/chat/browser/chatViewPane.ts
```diff
@@ -52,8 +56,16 @@ export class ChatViewPane extends ViewPane implements IViewWelcomeDelegate {
 	private readonly modelDisposables = this._register(new DisposableStore());
 	private memento: Memento<IViewPaneState>;
 	private readonly viewState: IViewPaneState;
+	private showSecondaryTitleBar: boolean;
+	private secondarySideBarShowsLabels: boolean;
```
roblourens: Maybe there's a way to reduce the number of different properties you have to store, for example checking IConfigurationService is pretty fast and you don't necessarily have to cache a value from it.

roblourens (commented):
I think the title shouldn't wrap
Image

roblourens:
I don't think people will like the shift down when the title gets generated, but I assume things are all still in flux

src/vs/workbench/contrib/chat/browser/chatViewPane.ts
roblourens: For those chats, I don't see the title rendered, which seems fine

hawkticehurst:
> I don't think people will like the shift down when the title gets generated, but I assume things are all still in flux

Yeah I'm guessing the same, but tis indeed in flux. Perhaps we can leave that one to the scream test and then adjust based on whatever feedback we get.

src/vs/workbench/contrib/chat/browser/chatViewPane.ts
hawkticehurst: Oop yes this should totally be localized. Thank you for the catch.

Also, the "Chat: " prefix will be rendered when:

- The activity bar is in the Top, Bottom, Hidden positions
- The activity bar is in the Default position AND the Show Icons config set
- The activity bar is in any position AND there are multiple views being rendered in the chat view pane

The "Chat: " prefix will NOT be rendered when:

- The activity bar is in the Default position AND the Show Labels config set

The idea (in theory) was to include the "Chat: " prefix as a sort of method of way finding –– it becomes easier to contextualize what the custom titles represent. Without it could become confusing, especially in configurations like when the activity bar in the hidden position:

Screenshot 2025-11-21 at 5 49 31 PM

And then the idea for not including the Chat prefix when the activity bar is in the Default position and the Show Labels config set is due to redundant "Chat" text right next to each other. Looks weird/off.

Screenshot 2025-11-21 at 8 48 43 PM

src/vs/workbench/contrib/chat/browser/chatViewPane.ts
```diff
@@ -249,7 +292,7 @@ export class ChatViewPane extends ViewPane implements IViewWelcomeDelegate {
 		this._widget.acceptInput(query);
 	}
 
-	private async clear(): Promise<void> {
+	public async clear(): Promise<void> {
```
bpasero: This seems like an unrelated change.

bpasero (changes requested):
I am not at all an owner of the `chatViewPane.ts` file, but when I look at the before and after, I find the change makes the rather simple `chatViewPane.ts` quite complex. Is there any chance we could encapsulate this logic into a separate file that then `chatViewPane.ts` simply depends on, just to separate concerns?
Having looked at `chatWidget.ts`, I think there we really missed the opportunity to keep the file readable (not just for humans but also AI!). The smaller the file, I think the easier to make changes in the future.

As for the UX: its a bit jarring that a chat at first does not have a title and then only later gets a title, which in turn pushes the chat down. I wonder if we can invent some generic title that then later is replaced? I am also not 100% sure about using all uppercase, at least when the title is above the chat and not part of the view title.

src/vs/workbench/contrib/chat/browser/chatViewPane.css
```diff
@@ -0,0 +1,16 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+.chat-secondary-title {
```
bpasero: I would really encourage to make this CSS rule more specific to the chat view pane, otherwise you risk matching on a rule that already exists somewhere else.

bpasero (changes requested):

src/vs/workbench/contrib/chat/browser/chatViewPane.ts
```diff
@@ -185,6 +224,10 @@ export class ChatViewPane extends ViewPane implements IViewWelcomeDelegate {
 	protected override async renderBody(parent: HTMLElement): Promise<void> {
 		super.renderBody(parent);
 
+		this._secondaryTitleContainer = $('.chat-secondary-title');
+		this._secondaryTitle = $('span');
+		append(this._secondaryTitleContainer, this._secondaryTitle);
+		append(parent, this._secondaryTitleContainer);
```
bpasero: We have a utility method `dom.ts#h()` to make building DOM elements easier, see for example https://github.com/microsoft/vscode/blob/2871c51409fda02cfa68bdf4ed7b4b7bc700827a/src/vs/workbench/contrib/chat/browser/agentSessions/agentSessionsViewer.ts#L78-L95

src/vs/workbench/contrib/chat/browser/chatViewPane.ts
hawkticehurst: Ooh thank you for the catch! This was needed when we still had the back icon button but forgot to revert this when I removed the back button from this PR.

src/vs/workbench/contrib/chat/browser/chatViewTitle.ts
```diff
@@ -0,0 +1,184 @@
...
+	}
+
+	render(parent: HTMLElement): void {
+		if (this._secondaryTitleContainer) {
+			return;
+		}
+		this._secondaryTitleContainer = h('div.chat-view-secondary-title-container').root;
+		this._secondaryTitle = h('span.chat-view-secondary-title-text').root;
+		append(this._secondaryTitleContainer, this._secondaryTitle);
+		append(parent, this._secondaryTitleContainer);
```
hawkticehurst: @bpasero @roblourens I know this isn't quite the intended use of an `h` style function for building UI but felt just a touch wasteful to build the UI then go back and query select for the secondary title when I could just create/get a reference to the title at the same time.

If you'd prefer that I don't do this (and just build the whole UI and then query select) I can change this. 

_*As I think about it now, I suppose query selecting a single nested element is probably a wash in terms of performance anyway?_

src/vs/workbench/contrib/chat/browser/chatViewPane.ts
hawkticehurst: Nice thank you, confirmed the secondary title is not showing up in this scenario.

src/vs/workbench/contrib/chat/browser/chatViewTitle.ts
roblourens: Yeah, we don't like querying for elements because you should have a direct reference to any element inside the widget itself. One improvement is that you can make one full element tree with a nested call to `h`

roblourens: I pushed a commit to show how this is done since it's hard to explain

src/vs/workbench/contrib/chat/browser/chatViewPane.ts
roblourens: But if I'm understanding the code correctly, you're operating on the chatmodel title, not the view title, so I don't think there will be a case where the title already is prefixed with "chat:"

src/vs/workbench/contrib/chat/browser/chatViewPane.ts
```diff
@@ -81,6 +85,20 @@ export class ChatViewPane extends ViewPane implements IViewWelcomeDelegate {
...
 		this.memento = new Memento('interactive-session-view-' + CHAT_PROVIDER_ID, this.storageService);
 		this.viewState = this.memento.getMemento(StorageScope.WORKSPACE, StorageTarget.MACHINE);
+		this.titleController = this._register(new ChatViewTitleController(
+			options.id,
+			{
+				getCurrentModel: () => this._widget?.viewModel?.model,
+				updatePrimaryTitle: title => this.updateTitle(title),
+				requestLayout: () => {
+					if (this._lastLayout) {
+						this.layoutBody(this._lastLayout.height, this._lastLayout.width);
```
roblourens: This delegate pattern is fine, but another way to think about this one would be to expose a `onDidChangeHeight` event on the title part, and listen to that and react to it in the view pane.

src/vs/workbench/contrib/chat/browser/chatViewTitle.ts
```diff
@@ -0,0 +1,178 @@
...
+		const elements = h('div.chat-view-secondary-title-container', [
+			h('span.chat-view-secondary-title-text@text')
+		]);
+		this._secondaryTitleContainer = elements.root;
+		this._secondaryTitle = elements.text;
+		append(parent, this._secondaryTitleContainer);
+	}
+
+	update(model?: IChatModel): void {
+		model = model ?? this.delegate.getCurrentModel();
```
roblourens: Either pass in a model or use the delegate to check the current model, having both is a little confusing

src/vs/workbench/contrib/chat/browser/chatViewTitle.ts
```diff
@@ -0,0 +1,178 @@
...
+		private readonly delegate: IChatViewTitleDelegate,
+		@IConfigurationService private readonly configurationService: IConfigurationService,
+		@IViewDescriptorService private readonly viewDescriptorService: IViewDescriptorService,
+	) {
+		super();
+		const viewContainer = this.viewDescriptorService.getViewContainerByViewId(this.viewId);
+		if (viewContainer) {
+			this.viewContainerModel = this.viewDescriptorService.getViewContainerModel(viewContainer);
+			this.hasMultipleVisibleViews = this.viewContainerModel.visibleViewDescriptors.length > 1;
+			this._register(this.viewContainerModel.onDidAddVisibleViewDescriptors(() => this.handleVisibleViewDescriptorsChanged()));
```
roblourens: I think it would be easier to follow the logic if instead of having different handlers for the different specific things that can change, you would just have one `update` method which always runs when anything changes, checks the state of everything, and renders the content as needed. Caching something like `hasMultipleVisibleViews` is premature optimization IMO

src/vs/workbench/contrib/chat/browser/chatViewTitle.ts
hawkticehurst: Ooh the `@text` thing is clever! If I'm understanding correctly, it's attaching a reference to a given node, accessible as a property of the final returned `element` object where the property name is whatever follows the `@`?

roblourens: That's right, and the real magic is that typescript knows about the name after the `@` so you get proper TS suggestions for the names. I couldn't even tell you how that works.

src/vs/workbench/contrib/chat/browser/chatViewPane.ts
hawkticehurst: Oh if the team is okay with the event bubbling pattern I'm happy to pivot! 

I've done a lot [frontend benchmarking](https://bsky.app/profile/hawkticehurst.com/post/3lguah5dvjs2e) in the past (for fun) and found that event bubbling is pretty meaningfully slower than directly delegating methods that can be called within a child class, but again if you're okay with it I do prefer event bubbling pattern for it's readability / simplified mental model.

roblourens: We do that kind of thing in a lot of places

hawkticehurst: Oh to clarify any potential confusion, the Bluesky thread I linked to doesn't explicitly talk about the event bubbling pattern being slower. It's more so just me generally talking about the benchmarking.

But it still is a discovery that came up while I was doing the benchmarking.

hawkticehurst: Oh hold up, realizing VS Code has custom `Event` / `Emitter` implementations. I'll skim through them, but I'm gonna assume these are more performant than the event bubbling pattern I was thinking of.

An example of what I was thinking of (with custom elements) that hopefully gets the idea across:

```ts
class ChatViewTitle extends HTMLElement {
	constructor() {
		super();
		this.textContent = 'Some chat title';
		this.addEventListener('view-layout-change', () => {
			const currentHeight = this.getCurrentHeight();
			this.dispatchEvent(new CustomEvent('height-change', {
				detail: { height: currentHeight },
				bubbles: true
			}));
		});
	}
	// ... other methods ...
}
customElements.define('chat-view-title', ChatViewTitle);

class ChatViewPane extends HTMLElement {
	constructor() {
		super();
		this.innerHTML = `<chat-view-title></chat-view-title>`;
		this.addEventListener('height-change', e => {
			const { height } = (e as CustomEvent<{ height: string }>).detail;
			// Do stuff with the height...
		});
	}
	// ... other methods ...
}
customElements.define('chat-view-pane', ChatViewPane);
```

src/vs/workbench/contrib/chat/browser/chatViewTitle.ts
hawkticehurst: Oooh very very cool. I'm definitely gonna check out how that works.

bpasero (dismissed):

lszomoru (approved):

Issue #275679, opened 2025-11-05
Support model-specific tool enablement in language model API


bhavyaus:
## Context

When working with different language models that have native/server-side tools (e.g., Anthropic's web search tool), we need the ability to control which tools are available on a per-model basis.

Currently, tools registered via the language model API are available to all models. However, some scenarios require model-specific tool visibility:

1. **Native/server-side tools**: Some models like Anthropic have native tools (e.g., web search) that are executed server-side. To provide a consistent experience, we'd want to offer wrapper tools that are only available when using those specific models.

2. **Model-specific capabilities**: Different models have varying levels of tool-calling support, and some tools may work better or only work with certain models.

3. **Provider-specific features**: Tool implementations may be optimized for specific model providers and shouldn't appear for models that can't use them effectively.

## Related

- This came up in the context of #275201 where Anthropic's server-side web search tool doesn't provide pre-call signals, and we'd want to offer a wrapper tool only for Anthropic models.
- Todo tool is currently disabled for grok and gpt-5-codex, but they still show up in the tool picker

Issue #277472, opened 2025-11-14
Support per-model/per-request tools


connor4312:
Currently tools are global on the `vscode.lm.tools` API. This makes it difficult/hacky to override tools per-model: we have the `alternativeDefinition?()` on `ICopilotTool` in the copilot extension, but this only works for tools that extension provides, and we're getting more of these cases with each model release.

1. Easy: `ChatRequest.tools` is instead `LanguageModelToolInformation[]` https://github.com/microsoft/vscode/blob/0147d461600b28e98fb555679b971203ae230050/src/vscode-dts/vscode.proposed.chatParticipantAdditions.d.ts#L402-L405
2. Would adding a `model` in the `when` clause in the package.json tool contribution be sufficient to signal enablement? I think there is some DX we'd want to do in copilot-chat if that's the case (i.e. programmatically generate that contribution based on per-model overrides in the code) but it would fit the best into our existing contribution model.
3. How do we invoke what could be a model-specific variant of an existing tool that is otherwise globally regsitered? Perhaps have [`invokeTool`](https://github.com/microsoft/vscode/blob/0147d461600b28e98fb555679b971203ae230050/src/vscode-dts/vscode.d.ts#L20713) take `name: string | LanguageModelToolInformation` and route accordingly if we get specific information passed in.

Per-request tools--the API for (1)/(3)--is also needed to support MCP's sampling with tools: https://github.com/modelcontextprotocol/modelcontextprotocol/pull/1796 where tools are included in a 'sampling' request and are available as tool calls only in that specific context.

pwang347:
Here's a related use case from GitHub Custom Agents:

<img width="719" height="298" alt="Image" src="https://github.com/user-attachments/assets/0cddf809-f035-47e2-9240-d768931a1a57" />

jrieken:
Today `modelDescription` of the `languageModelTools`-array is just a string. Should we extend this to also be a complex structure that allows to target certain model/model-families and allow them to (a) give a special definition and (b) allow to define exclusive use?

roblourens:
I would rather have one place in code where the agent is configured for a model, with all prompts and core tool customizations. That might mean breaking with how extension API works now, but I think that core tools are special, need to be closely aligned with specific model versions, and extension tools don't need this flexibility.

PR #278904, opened 2025-11-22
Support tool filtering by model

bhavyaus:
For https://github.com/microsoft/vscode/issues/275679

Ensures `ChatRequest.tools` correctly reflects tool availability and enablement state by filtering based on the active model's compatibility.

Key changes:
- Added `supportedModels` field to `IToolDataDto` and passed through main-to-exthost communication
- Updated `ExtHostChatAgents2.getToolsForRequest()` to accept `modelId` parameter and filter tools using `isToolAvailableForModel()`
- Modified tool map to include all model-compatible tools with correct enablement state (defaults to enabled if not explicitly set)
- Updated all `getToolsForRequest()` call sites to pass the active model ID


src/vs/workbench/contrib/chat/common/tools/languageModelToolsContribution.ts
```diff
@@ -138,6 +139,13 @@ const languageModelToolsExtensionPoint = extensionsRegistry.ExtensionsRegistry.r
 						type: 'string',
 						pattern: '^(?!copilot_|vscode_)'
 					}
+				},
```
roblourens: Is a static contribution flexible enough? It wouldn't let us control this with a setting/exp, or do more flexible checks like `family.contains('-codex')` that we do in code.

Also, going along with wanting to do other prompt tweaks in a more centralized place, it would be nice to have one place in code where I can determine the prompts and tools enabled for a certain model family.

How do you see this working for websearch, do you register a tool in api that doesn't have a invoke implementation but can show up in UI?

src/vscode-dts/vscode.proposed.chatParticipantAdditions.d.ts
```diff
@@ -402,7 +402,7 @@ declare module 'vscode' {
 		/**
 		 * A map of all tools that should (`true`) and should not (`false`) be used in this request.
 		 */
-		readonly tools: Map<string, boolean>;
+		readonly tools: Map<LanguageModelToolInformation, boolean>;
```
roblourens, line 477: Why is this needed?

connor4312: I think this is needed for the more general cases in https://github.com/microsoft/vscode/issues/277472 where `LanguageModelToolInformation` given to a request might not be globally available on `vscode.lm.tools`, or where there are multiple definitions of a tool that vary per model and the tool name alone is not enough to identify it.

I would pair this with making `invokeTool(tool: string | LanguageModelToolInformation` so that extensions are able to invoke tools that are only visible in the scope of a single request

src/vs/workbench/contrib/chat/common/tools/languageModelToolsContribution.ts
bhavyaus: I was thinking of piggy backing off `ICopilotToolExtension` that Bryan added to support alrernativeDefintions and have that dynamically compute the supported models for now.  

> How do you see this working for websearch, do you register a tool in api that doesn't have a invoke implementation but can show up in UI?


Yes. But I don't think you can't NOT have an invoke with the way things are setup currently though.

connor4312: I agree that we probably want more flexibility here. Otherwise every single release of every model will require changes. In the related issue #277472 I brought up the notion of using a `when` clause.

This does not necessarily mean you need to drive it off global context keys (though maybe using `when` would overload it, perhaps use a different name?) as there is a standalone `Parser` in contextkey.ts that returns an expression you can easily evaluate by passing an object with a `getValue` method. So having something specific there that is evaluated when a request is sent with information about the model, family, etc. is doable.

bhavyaus: How about updating `LanguageModelTool`  to allow tools to dynamically specify which models they support?

```
declare module 'vscode' {

	export interface LanguageModelTool<T> {
		/**
		 * Called to check if this tool supports a specific language model. If this method is not implemented,
		 * the tool is assumed to support all models.
		 *
		 * This method allows extensions to dynamically determine which models a tool can work with,
		 * enabling fine-grained control over tool availability based on model capabilities.
		 *
		 * @param modelId The identifier of the language model (e.g., 'gpt-4o', 'claude-3-5-sonnet')
		 * @returns `true` if the tool supports the given model, `false` otherwise
		 */
		supportsModel?(modelId: string): Thenable<boolean>;
	}
}

```

src/vs/workbench/api/common/extHostLanguageModelTools.ts
```diff
@@ -171,6 +171,28 @@ export class ExtHostLanguageModelTools implements ExtHostLanguageModelToolsShape
...
+	 * @returns `true` if supported, `false` if not, `undefined` if no supportsModel implementation (treat as supported)
+	 */
+	async supportsModel(toolId: string, modelId: string, token: CancellationToken): Promise<boolean | undefined> {
+		const item = this._registeredTools.get(toolId);
+		if (!item) {
+			// Tool is not registered in this extension host, assume it supports all models
+			return undefined;
+		}
+
+		// supportsModel is a proposed API, cast to access it
```
bhavyaus: Copilot what?

src/vscode-dts/vscode.proposed.chatParticipantAdditions.d.ts
bhavyaus: I needed this to be able to filter the tools for a request in the chat extension to just the supported ones. 
Will explore updating invokeTools per Connor's suggestion above.

src/vs/workbench/contrib/chat/common/tools/languageModelToolsContribution.ts
connor4312: While that would definitely be a bit easier to author, the downside with that approach is that we must activate every extension that provides language model tools when a chat message is sent in order to filter them appropriately. Previously we've only needed to activate tool-providing extensions when a tool they provide is actually invoked.

roblourens: Are you expecting this to be something that would be useful as real API? I was only thinking of this for our core tools. Is MCP considering something like this?

bhavyaus: Was only looking to support this for the builit-in tools we expose. @connor4312 do you anticipate needing this per model activation for MCP tools ?

connor4312: For MCP there are two scenarios:

1. Tools from MCP servers that are defined in `agents.md` and similar. We're seeing clients start to support inline MCP's in this way and so their tools should only be applied to requests that involve those agent instructions. 
2. Sampling with tools. In these cases tools are purely internal, we get them in core in the MCP sampling request and can register them however we want.

In both cases no particular extension API is needed for the MCP scenarios.

connor4312:
@aeschli in this PR I have suggested moving from storing the `tool.id` as the key for enablement to instead storing the `toolReferenceName` when possible. This allows two differing tool registrations (e.g. a web search tool from Anthropic and one from OAI) to both share the same name for the user and have their enablement persist together, while having different underlying `id`s, schemas, and implementation.

DonJayamanne (dismissed):

roblourens (dismissed):

DonJayamanne (dismissed):

aeschli:
I like the addition of `toolset` to the tool definition. 
On the [IToolData in core](https://github.com/microsoft/vscode/blob/72bacda9f3493262af0705a4314ffa5ea4ac5a56/src/vs/workbench/contrib/chat/common/tools/languageModelToolsService.ts#L46) I would now add a property `fullToolReferenceName` with derived value `${toolset}/${toolReferenceName ?? displayName`. That should eliminates a lot of complexity on my side and we can remember enabled tools by 'fullToolReferenceName'.

aeschli:
Is the main purpose to have specialized tool implementations for a tool? It might be cleaner to model that and have an additional contribution point or provider for tool implementations. A tool implementation has the list of models it supports and points to a tool definition id. It doesn't need to repeat all the properties.
Maybe just the extension needs to deal with the tool implementations

dbaeumer (dismissed):

roblourens (approved):

PR #278958, opened 2025-11-22
agent sessions - removal of history from chat view

bpasero:


joshspicer (approved):

MominRaza:
Bring back the history icon as the history list is removed https://github.com/microsoft/vscode/issues/267860

PR #279088, opened 2025-11-23
fix: memory leak in terminal editor

SimonSiefke:
Fixes a memory leak in terminal editor by removing the focus listener when `clearInput` is called.

### Before

When moving a terminal from panel to editor area 97 times, a function seems to be created each time at `terminalEditor.ts:87:64` which is never removed.

```json
{
  "namedFunctionCount3": [
    {
      "count": 99,
      "delta": 97,
      "name": "anonymous",
      "sourceLocation": "out/vs/workbench/workbench.desktop.main.js:2841:56564",
      "originalLocation": "src/vs/workbench/contrib/terminal/browser/terminalEditor.ts:87:64",
      "originalName": "TerminalEditor.setInput"
    }
  ],
  "isLeak": true
}
```

### After

No more leak is detected.

```json
{
  "namedFunctionCount3": [],
  "isLeak": false
}
```


meganrogge (approved):
Thank you!

roblourens (approved):

PR #279094, opened 2025-11-24
tool calls in dropdown and thinking headers

justschen:
1. all tools inside dropdown
2. updates header using chat title service
3. refactor chat list renderer parts stuff to not have a persistant state - state is handled per part.


src/vs/workbench/contrib/chat/browser/chat.contribution.ts
```diff
@@ -750,15 +750,15 @@ configurationRegistry.registerConfiguration({
...
-				nls.localize('chat.agent.thinking.collapsedTools.all', "All tool calls are added into the collapsible thinking section."),
-				nls.localize('chat.agent.thinking.collapsedTools.readOnly', "Only read-only tool calls are added into the collapsible thinking section."),
-			],
-			markdownDescription: nls.localize('chat.agent.thinking.collapsedTools', "When enabled, tool calls are added into the collapsible thinking section according to the selected mode."),
+			type: 'boolean',
+			default: true,
+			markdownDescription: nls.localize('chat.agent.thinking.collapsedTools', "When enabled, all tool calls are added into the collapsible thinking section."),
+			tags: ['experimental'],
+		},
+		'chat.agent.thinking.collapsedTools2': {
```
roblourens: Why two settings instead of the one enum?

src/vs/workbench/contrib/chat/browser/chatContentParts/chatCollapsibleContentPart.ts
```diff
@@ -95,6 +101,14 @@ export abstract class ChatCollapsibleContentPart extends Disposable implements I
 		this._register(disposable);
 	}
 
+	override dispose(): void {
+		if (this.renderedTitleWithWidgets) {
```
roblourens: Good candidate for MutableDisposable?

src/vs/workbench/contrib/chat/browser/chatContentParts/chatThinkingContentPart.ts
```diff
@@ -212,51 +250,124 @@ export class ChatThinkingContentPart extends ChatCollapsibleContentPart implemen
...
-			if (this._collapseButton) {
-				this._collapseButton.icon = Codicon.check;
-				this._collapseButton.label = finalLabel;
-			}
-		} else {
-			if (this.currentTitle === this.defaultTitle) {
-				const suffix = localize('chat.thinking.fixed.done.generic', 'Thought for a few seconds');
-				this.setTitle(suffix);
-				this.currentTitle = suffix;
+	public finalizeTitleIfDefault(agentService?: IChatAgentService, agentId?: string, sessionResource?: URI): void {
```
roblourens: Why do you have to pass a service to another class?

src/vs/workbench/contrib/chat/browser/chatContentParts/chatThinkingContentPart.ts
```diff
@@ -212,51 +250,124 @@ export class ChatThinkingContentPart extends ChatCollapsibleContentPart implemen
...
+		if (this.extractedTitles.length > 0 && agentService && agentId && sessionResource) {
+			this.generateTitleViaLLM(agentService, agentId, sessionResource);
+		}
+	}
+
+	private async generateTitleViaLLM(agentService: IChatAgentService, agentId: string, sessionResource: URI): Promise<void> {
+		try {
+			const titlesContext = this.extractedTitles.join(', ');
+			const message = `Generate a concise header for thinking that contains the following thoughts: ${titlesContext}`;
+			const generatedTitle = await agentService.getChatTitle(agentId, [{
```
roblourens: I think preferably we would make a request to a model direclty rather than creating a fake request and using chat title generation, that way you can have full control over the prompt, which model is used, etc.

roblourens: Does this happen every time we render this content? I don't see it being saved anywhere. Seems like we should only do this once then save it.

src/vs/workbench/contrib/chat/browser/chatListRenderer.ts
```diff
@@ -291,11 +290,14 @@ export class ChatListItemRenderer extends Disposable implements ITreeRenderer<Ch
 		this.viewModel = viewModel;
 		this._announcedToolProgressKeys.clear();
 
-		if (this._currentThinkingPart) {
-			this._currentThinkingPart.dispose();
-			this._currentThinkingPart = undefined;
+		for (const templateData of this.templateDataByRequestId.values()) {
```
roblourens, line 292: This statefulness seems like an issue, why do we keep these around from a previous viewmodel?

justschen: this was added in for edting previous requests (since sometimes the ids and positions would get moved around on different renders) and is just leveragin those. 

in all the other cases we can get `templateData` from the parameter

src/vs/workbench/contrib/chat/browser/chatContentParts/chatThinkingContentPart.ts
justschen: updated to request the model directly.

it does happen everytime on render (so mostly scrolling or reloads), but not sure the best way to save or seralize this?

src/vs/workbench/contrib/chat/browser/chatContentParts/chatThinkingContentPart.ts
```diff
@@ -287,37 +287,54 @@ export class ChatThinkingContentPart extends ChatCollapsibleContentPart implemen
...
+				models = await this.languageModelsService.selectLanguageModels({ vendor: 'copilot', family: 'gpt-4o-mini' });
+			}
+			if (!models.length) {
+				this.setFallbackTitle();
+				return;
+			}
+
+			let context: string;
+			if (this.extractedTitles.length > 0) {
+				context = this.extractedTitles.join(', ');
```
roblourens, line 328: Could you skip the call when there is exactly 1 title?

src/vs/workbench/contrib/chat/browser/chatContentParts/chatThinkingContentPart.ts
roblourens: Seems a bit tricky, but important to avoid seeing flickering when these update unnecessarily, getting different titles at different times, or just wasting gpu capacity.

It makes me think that this work of grouping and extracting titles then computing nice summaries should happen at the ChatModel level, not the view level. I could help figure out what that can look like, in the short term we could probably figure out how to hack it as persisted view state (a similar flow to Connor's new `IInputModel`).

src/vs/workbench/contrib/chat/browser/chatListRenderer.ts
roblourens: Ah I remember why we have to do this for editing

roblourens (approved):

PR #279107, opened 2025-11-24
ChatSessionService test fix

osortega:


deepak1556 (approved):

bpasero (changes requested):
This is NOT good and was the original reason this never caused a compile error:

https://github.com/microsoft/vscode/blob/03d31bd4db9eecac576a6018c571424b6e852e5f/src/vs/workbench/contrib/chat/test/browser/chatSessionsService.test.ts#L23-L27

PR #279148, opened 2025-11-24
fix: memory leak in terminal decoration addon

SimonSiefke:
Fixes a memory leak in terminal decoration addon by registering two listeners as disposables

```ts
decoration.onRender(element => { // need to register here
	/* ... */
	if (!this._decorations.get(decoration.marker.id)) {
		decoration.onDispose(() => this._decorations.delete(decoration.marker.id)); // need to register here
       /* ... */
	}
});
```

### Before

When running a task 97 times, the number of various functions related to DecorationAddon grows each time

Untitled

<br>

### After 

When running a task 97 times, while there still seem to be some memory leaks, the number of various functions related to DecorationAddon stays constant now. 

task quick-pick-run-task


meganrogge (dismissed):
Thanks!

dmitrivMS:
@meganrogge CI is not happy...

bpasero (dismissed):

meganrogge (approved):

alexdima (approved):

PR #279198, opened 2025-11-24
Update to gulp 5

mjbvz:
For https://github.com/microsoft/vscode/issues/277526

The 4 to 5 gulp update mostly is about internal cleanup. It switches to more maintained dependencies and reduces the total number of dependencies. It also fixes some annoying bugs, such as not being able to use top level awaits in scripts


Tyriar (approved):
Red CI

Issue #252605, opened 2025-06-27
[engineering] add build/checker for typescript `isolatedModules: true`


tmm1:
Previously enabled in https://github.com/microsoft/vscode/pull/212913 and some errors were fixed

Later reverted in https://github.com/microsoft/vscode/issues/227372 due to [open typescript bug](https://github.com/microsoft/TypeScript/issues/16671)

esbuild [does support](https://github.com/evanw/esbuild/issues/128#issuecomment-1002341244) the const enum inlining optimization, however [strongly recommends `isolatedModules`](https://github.com/evanw/esbuild/issues/128#issuecomment-1599085571) to ensure correctness at compile time.

Currently on main some failures have snuck in again:

```
src/vs/workbench/contrib/chat/common/tools/promptTsxTypes.ts:58:20 - error TS2748: Cannot access ambient const enums when 'isolatedModules' is enabled.

58  if (node.type === PromptNodeType.Text) {
                      ~~~~~~~~~~~~~~
src/vs/workbench/contrib/chat/common/tools/promptTsxTypes.ts:66:27 - error TS2748: Cannot access ambient const enums when 'isolatedModules' is enabled.

66  } else if (node.ctor === PieceCtorKind.ImageChatMessage) {
                             ~~~~~~~~~~~~~
src/vs/workbench/contrib/chat/common/tools/promptTsxTypes.ts:69:27 - error TS2748: Cannot access ambient const enums when 'isolatedModules' is enabled.

69  } else if (node.ctor === PieceCtorKind.BaseChatMessage || node.ctor === PieceCtorKind.Other) {
                             ~~~~~~~~~~~~~
src/vs/workbench/contrib/chat/common/tools/promptTsxTypes.ts:69:74 - error TS2748: Cannot access ambient const enums when 'isolatedModules' is enabled.

69  } else if (node.ctor === PieceCtorKind.BaseChatMessage || node.ctor === PieceCtorKind.Other) {
                                                                            ~~~~~~~~~~~~~
```

Is it worth adding this as another config to the tsconfig variants being used for layer checking?

cc @jrieken @bpasero

bryanchen-d:
From the errors, it seems it will only ban the usage of `declare const enum`. Normal `const enum` is good.

PR #279208, opened 2025-11-24
Turn on isolatedModule in tsconfig

bryanchen-d:
Turn on isolatedModule flag in tsconfig, and fix some `declare const enum`

Address https://github.com/microsoft/vscode/issues/252605

### DIFF
--- a/src/tsconfig.json
+++ b/src/tsconfig.json
@@ -7,7 +7,7 @@
 		"sourceMap": false,
 		"allowJs": true,
 		"resolveJsonModule": true,
-		"isolatedModules": false,
+		"isolatedModules": true,
 		"outDir": "../out/vs",
 		"types": [
 			"@webgpu/types",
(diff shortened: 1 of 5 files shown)


jrieken (changes requested):
My testing (compile our sources with this change) shows that const enums are not inlined anymore. This can also be seen in the [TS Playground](https://www.typescriptlang.org/play/?esModuleInterop=false&preserveConstEnums=true&declaration=false&target=11&isolatedModules=false#code/FAimA8AcHsCcBcAEBjaA7Azk0aCuBbRAMWmg0QG9hFjSBGRAXkToBprboAmJxL9miWgBmXsPYBfEKkzQANqAB0c6AHMAFEIyKhdAJTAZGeUpUatO0sINGTytZtLahXPUA) This seems to be in contrast with how esbuild does it and something we need to consider very carefully because of the performance implications.

PR #279230, opened 2025-11-24
fix: memory leak in folder configuration

SimonSiefke:
Fixes some memory leaks in folder configuration / configuration service. 

### Issue one:  Registering to global configuration service

Since there is only one global configuration service, registering a folder configuration to that class using `this._register` makes it so the class holds on to the folder configuration forever, since the global configuration service is never disposed. 

```ts
private loadFolderConfigurations(folders: IWorkspaceFolder[]): Promise<ConfigurationModel[]> {
	/* ...*/
	let folderConfiguration = this.cachedFolderConfigs.get(folder.uri);
	if (!folderConfiguration) {
		folderConfiguration = new FolderConfiguration(!this.initialized, folder, FOLDER_CONFIG_FOLDER_NAME, this.getWorkbenchState(), this.isWorkspaceTrusted, this.fileService, this.uriIdentityService, this.logService, this.configurationCache);
		this._register(folderConfiguration.onDidChange(() => this.onWorkspaceFolderConfigurationChanged(folder)));
		this.cachedFolderConfigs.set(folder.uri, this._register(folderConfiguration));
	}
	return folderConfiguration.loadConfiguration();
})]);
}
```

### Fix

Registering all folder related disposables only to the map `cachedFolderConfigs: Map<URI, FolderConfiguration>`

<br>

### Issue two: Resetting caches

This code resets the caches, but it doesn't dispose the values of the map.

```ts
// reset caches
this.cachedFolderConfigs = new ResourceMap<FolderConfiguration>();
```

### Fix

Switching to a `DisposableMap` and making use of it's `clearAndDisposeAll` function.

<br>

### Before

When executing a task 97 times,  the number of functions/objects related to `FolderConfiguration` grows each time

Untitled (3)

### After

When executing a task 97 times, while the number of various functions/objects still grows, it seems there are no more leaks related to `FolderConfiguration` 

task quick-pick-run-task


src/vs/workbench/services/configuration/browser/configurationService.ts
```diff
@@ -77,7 +77,7 @@ export class WorkspaceService extends Disposable implements IWorkbenchConfigurat
 	private readonly localUserConfiguration: UserConfiguration;
 	private readonly remoteUserConfiguration: RemoteUserConfiguration | null = null;
 	private readonly workspaceConfiguration: WorkspaceConfiguration;
-	private cachedFolderConfigs: ResourceMap<FolderConfiguration>;
+	private cachedFolderConfigs: DisposableMap<URI, FolderConfiguration> = this._register(new DisposableMap());
```
sandy081: I do not think ResourceMap<T> is same as Map<Uri, T>. Therefore we cannot use DisposableMap. Can you please dispose manually from the map instead of using DisposableMap which disposes automatically?

sandy081 (changes requested):

src/vs/workbench/services/configuration/browser/configurationService.ts
SimonSiefke: I changed it to `new DisposableMap(new ResourceMap())`.

sandy081 (approved):

dmitrivMS (approved):

SimonSiefke:
(updated branch to try to fix MacOS CI timeout)

dmitrivMS (approved):

DonJayamanne (approved):

PR #279242, opened 2025-11-24
fix: memory leak in accessibility signal

SimonSiefke:
Fixes a memory leak in accessibility signal by ensuring to dispose the request in all cases, even when there is no result.

### Before
Screenshot From 2025-10-13 14-57-29

### After

No more leaks are detected.


meganrogge (approved):
Thanks!

bpasero (approved):

PR #279249, opened 2025-11-24
Resolve regression where build task did not start in wsl

### LINKED ISSUE #275011: Getting strange terminal message when opening VS Code in WSL on a trusted workspace
Image

The workspace is actual trusted.

anthonykim1:
Resolves: https://github.com/microsoft/vscode/issues/275011

Long story short, after https://github.com/microsoft/vscode/commit/975a1efd611d8d5cc0a2bb2bb7463edb3c4cd058, we ensure capabilities registered before shell integration events fire.

~SOMEHOW
ShellIntegrationAddon._updatecwd  and `onDidChangeCwd` was getting called before `createProcess` in remote.
Whereas, in non-remote, create process happened first, then updateCwd happened.~

Edit: Maybe its more suspicious that activeWorkspaceUri is undefined.. for the wsl active workspace case:

```
2025-11-24 22:51:57.466 [debug] [9f844c9] MYLOG TerminalInstance#_createProcess: activeWorkspaceRootUri=undefined []
2025-11-24 22:51:57.466 [debug] [9f844c9] MYLOG TerminalInstance#_createProcess: Empty workspace cwd mismatch detected - cwd="\mnt\c\Users\antho\Desktop\All\vscode-copilot-chat", userHome="\home\anthonykim1", hasRemoteAuthority=true []
```

Regardless, we were going into https://github.com/microsoft/vscode/blob/51e7119ccb43f1e1dbb3d059bfc9bac09eebb8cb/src/vs/workbench/contrib/terminal/browser/terminalInstance.ts#L1539-L1543 which was what was causing the build tasks to fail.


anthonykim1:
I'm not sure if its expected that remote updateCwd happens before createProcess, which is the opposite of non-remote scenario

src/vs/workbench/contrib/terminal/browser/terminalInstance.ts
```diff
@@ -1530,24 +1539,35 @@ export class TerminalInstance extends Disposable implements ITerminalInstance {
 		if (this.isDisposed) {
 			return;
 		}
+		this._logService.debug(`MYLOG TerminalInstance#_createProcess: Starting process creation, _cwd="${this._cwd}", _userHome="${this._userHome}", hasRemoteAuthority=${this.hasRemoteAuthority}, remoteAuthority=${this.remoteAuthority}`);
 		const activeWorkspaceRootUri = this._historyService.getLastActiveWorkspaceRoot(Schemas.file);
+		this._logService.debug(`MYLOG TerminalInstance#_createProcess: activeWorkspaceRootUri=${activeWorkspaceRootUri?.toString()}`);
 		if (activeWorkspaceRootUri) {
```
Tyriar: Based on the screenshot in https://github.com/microsoft/vscode/issues/275011, I would have expected this to reproduce in a non-empty workspace in which case it would trigger `_onProcessExit` based purely on `trusted`?

anthonykim1: Ahhh The activeWorkspaceUri was undefined for wsl (for both task and regular terminal it seems like)

I might have to debug why `const activeWorkspaceRootUri = this._historyService.getLastActiveWorkspaceRoot(Schemas.file);` is undefined now

DonJayamanne (approved):

PR #279250, opened 2025-11-24
Fix code block rendering logic to prevent unnecessary re-renders. Fix #277574

### LINKED ISSUE #277574: Codeblocks aren't rendered when restoring chat sessions
Version: 1.107.0-insider (Universal)
Commit: 18279e23d7d90c429d329c1104d2108d388b8210
Date: 2025-11-14T20:20:20.505Z
Electron: 39.1.2
ElectronBuildId: 12766293
Chromium: 142.0.7444.134
Node.js: 22.21.1
V8: 14.2.231.17-electron.0
OS: Darwin arm64 24.6.0

Name: GitHub Copilot Chat
Id: GitHub.copilot-chat
Description: AI chat features powered by Copilot
Version: 0.34.2025111403
Publisher: GitHub
VS Marketplace Link: https://marketplace.visualstudio.com/items?itemName=GitHub.copilot-chat

1. Type the following prompt to Ask mode:

~~~
output the following three times:

aaa message

```text
a


a



a




a



a



aaaa

bbb


aa
```
~~~

2. Start a new chat
3. Restore the previous chat session from the history.
4. Codeblocks aren't rendered 

Image

tamuratak:
When restoring a chat from history, there’s still a chance that some code blocks may not render. Fix #277574

### DIFF
--- a/src/vs/workbench/contrib/chat/browser/chatContentParts/chatMarkdownContentPart.ts
+++ b/src/vs/workbench/contrib/chat/browser/chatContentParts/chatMarkdownContentPart.ts
@@ -416,13 +416,12 @@ export class ChatMarkdownContentPart extends Disposable implements IChatContentP
 		if (other.kind !== 'markdownContent') {
 			return false;
 		}
-
-		if (other.content.value === this.markdown.content.value) {
+		const lastCodeblock = this._codeblocks.at(-1);
+		if (other.content.value === this.markdown.content.value && !lastCodeblock) {
 			return true;
 		}
 
 		// If we are streaming in code shown in an edit pill, do not re-render the entire content as long as it's coming in
-		const lastCodeblock = this._codeblocks.at(-1);
 		if (lastCodeblock && lastCodeblock.codemapperUri !== undefined && lastCodeblock.isStreamingEdit) {
 			return other.content.value.lastIndexOf('```') === this.markdown.content.value.lastIndexOf('```');
 		}
--- a/src/vs/workbench/contrib/chat/browser/chatContentParts/chatMarkdownContentPart.ts
+++ b/src/vs/workbench/contrib/chat/browser/chatContentParts/chatMarkdownContentPart.ts
@@ -416,12 +416,13 @@ export class ChatMarkdownContentPart extends Disposable implements IChatContentP
 		if (other.kind !== 'markdownContent') {
 			return false;
 		}
-		const lastCodeblock = this._codeblocks.at(-1);
-		if (other.content.value === this.markdown.content.value && !lastCodeblock) {
+		const allCodeblocksComplete = this._codeblocks.every(cb => cb.codemapperUri !== undefined && !cb.isStreamingEdit);
+		if (other.content.value === this.markdown.content.value && allCodeblocksComplete) {
 			return true;
 		}
 
 		// If we are streaming in code shown in an edit pill, do not re-render the entire content as long as it's coming in
+		const lastCodeblock = this._codeblocks.at(-1);
 		if (lastCodeblock && lastCodeblock.codemapperUri !== undefined && lastCodeblock.isStreamingEdit) {
 			return other.content.value.lastIndexOf('```') === this.markdown.content.value.lastIndexOf('```');
 		}


src/vs/workbench/contrib/chat/browser/chatContentParts/chatMarkdownContentPart.ts
```diff
@@ -416,8 +416,8 @@ export class ChatMarkdownContentPart extends Disposable implements IChatContentP
 		if (other.kind !== 'markdownContent') {
 			return false;
 		}
-
-		if (other.content.value === this.markdown.content.value) {
+		const allCodeblocksComplete = this._codeblocks.every(cb => cb.codemapperUri !== undefined && !cb.isStreamingEdit);
```
connor4312, line 419: I'm kind of confused why this is necessary. If the markdown contents are identical between parts then there should be no need to re-render--you should get the same output each time. If a code block is not rendering correctly then I think that may be an underlying bug with how the block is rendered

tamuratak: When a session is loaded, it can happen that the Markdown is identical but the code blocks haven’t been rendered yet. The `hasSameContent` function is being used to check whether code block rendering is complete. That’s why I believe the correct fix is to update `hasSameContent`.

I haven’t traced the session-loading implementation in detail, though.

Below is the response I received when I asked gpt-5.1-codex to investigate.


---

- `ChatMarkdownContentPart.hasSameContent()` short-circuits purely on markdown string equality [chatMarkdownContentPart.ts#L415-L430](https://github.com/microsoft/vscode/blob/main/src/vs/workbench/contrib/chat/browser/chatContentParts/chatMarkdownContentPart.ts#L415-L430). When you reopen an older session, the list already holds stale `ChatMarkdownContentPart` instances for those response IDs, but their DOM was cleared when the template was recycled (`clearRenderedParts` disposes and empties nodes when the template is reassigned) [chatListRenderer.ts#L526-L538](https://github.com/microsoft/vscode/blob/main/src/vs/workbench/contrib/chat/browser/chatListRenderer.ts#L526-L538). Because `hasSameContent()` returns `true`, `ChatListItemRenderer.diff()` treats every markdown part as unchanged and never calls `renderMarkdown` again [chatListRenderer.ts#L1207-L1219](https://github.com/microsoft/vscode/blob/main/src/vs/workbench/contrib/chat/browser/chatListRenderer.ts#L1207-L1219), leaving those rows without any code-block DOM.  

- Skipping `renderMarkdown()` also means the code paths that actually instantiate editors or edit pills never run. The editor/pill creation happens inside `codeBlockRendererSync` (which pushes entries into `_codeblocks`) and in `renderCodeBlock`/`renderCodeBlockPill` [chatMarkdownContentPart.ts#L173-L312](https://github.com/microsoft/vscode/blob/main/src/vs/workbench/contrib/chat/browser/chatContentParts/chatMarkdownContentPart.ts#L173-L312) and [chatMarkdownContentPart.ts#L397-L411](https://github.com/microsoft/vscode/blob/main/src/vs/workbench/contrib/chat/browser/chatContentParts/chatMarkdownContentPart.ts#L397-L411). Because the markdown part is never recreated after a session reload, `_codeblocks` stays empty, `codeBlockModelCollection.update()` is never re-run to fetch text/codemapper URIs, and the subsequent `layout()` pass can’t call `CollapsedCodeBlock.render()` to paint anything [chatMarkdownContentPart.ts#L433-L445](https://github.com/microsoft/vscode/blob/main/src/vs/workbench/contrib/chat/browser/chatContentParts/chatMarkdownContentPart.ts#L433-L445). This is why historical responses show a blank area where the code block should be.  

- The streaming guard compounds the issue: the only non-string check in `hasSameContent()` is the final `isStreamingEdit` clause, but it only looks at the last code block and assumes that as soon as the textual content stabilizes, rendering can stop [chatMarkdownContentPart.ts#L424-L428](https://github.com/microsoft/vscode/blob/main/src/vs/workbench/contrib/chat/browser/chatContentParts/chatMarkdownContentPart.ts#L424-L428). When a response finished while a code-edit pill was still waiting for its `codemapperUri`, every block except the last is ignored, so we mark the whole part as “same content” even though none of the pills have transitioned from “streaming” to “interactive.” When that session is loaded later, we immediately reuse the stale part, never re-run `renderCodeBlockPill()`, and therefore never call `CollapsedCodeBlock.render()` with the persisted URI. The patch you shared (checking that every `_codeblock` has a URI and is no longer streaming before returning `true`) addresses precisely this hole by forcing one more render pass so the pills/editors are rebuilt for historical sessions.

connor4312: > When you reopen an older session, the list already holds stale ChatMarkdownContentPart instances for those response IDs

Not sure the the model is correct here. The list view is used to render whatever the current session is. Its contents get replaced when you switch models and there should be no 'stale' components around anywhere.

Are you still reproducing this with the same steps from https://github.com/microsoft/vscode/issues/277574#issue-3627720123? I don't seem to hit that on current Insiders.

tamuratak: I can reproduce the issue.

image

steps:

1. Start a new chat.
2. Type the prompt.
3. Move the session to the editor area.
4. Close the chat in the editor area.
5. Restore the session in the chat.

Version: 1.107.0-insider (Universal)
Commit: 6e0893efb3d4b50c43dcff8979c722e6019de61d
Date: 2025-12-02T05:02:35.190Z
Electron: 39.2.3
ElectronBuildId: 12869810
Chromium: 142.0.7444.175
Node.js: 22.21.1
V8: 14.2.231.21-electron.0
OS: Darwin arm64 24.6.0

Name: GitHub Copilot Chat
Id: GitHub.copilot-chat
Description: AI chat features powered by Copilot
Version: 0.34.2025120202
Publisher: GitHub
VS Marketplace Link: https://marketplace.visualstudio.com/items?itemName=GitHub.copilot-chat


settings:

```
    "github.copilot.chat.summarizeAgentConversationHistory.enabled": false,
    "github.copilot.chat.useResponsesApi": true,
    "github.copilot.chat.codesearch.enabled": true,
    "chat.detectParticipant.enabled": false,
    "chat.math.enabled": true,
    "chat.agent.maxRequests": 75,
    "chat.mcp.autostart": "never",
    "chat.recentSessions.enabled": false
```

tamuratak: I opened a new PR for an alternative fix. #281505. You might prefer it.

Issue #279266, opened 2025-11-25
showRange field regression: functionality lost during InlineCompletionsSource refactoring


yanmofeixi:
Does this issue occur when all extensions are disabled?: No

- VS Code Version: 
- OS Version: 

 Description

  The showRange field in InlineCompletionItem (part of the inlineCompletionsAdditions proposed API) is currently not working. While the field is
  still defined in the API and accepted from extensions, the implementation that checks cursor position against this range was completely removed during a refactoring.

  Steps to Reproduce

  1. Create an extension using the inlineCompletionsAdditions proposed API
  2. Return an InlineCompletionItem with a showRange field set to a specific range
  3. Move the cursor outside of the showRange
  4. Expected: The inline completion should not be shown
  5. Actual: The inline completion is always shown regardless of cursor position

  Root Cause Analysis

  Working implementation (Jan 9, 2025 - commit a016c0be118):
  - PR #237532 added support for showRange
  - InlineCompletionItem had a cursorShowRange field
  - inlineCompletionsModel.ts checked: cursorShowRange?.containsPosition(cursorPos) ?? true

  Regression (Apr 7, 2025 - commit b0abf06c7dc):
  - Major refactoring of InlineCompletionsSource
  - The cursorShowRange field was removed from InlineCompletionItem
  - All cursor position checking logic was deleted
  - However, the API definition remains in vscode.proposed.inlineCompletionsAdditions.d.ts

  Evidence

  API layer still accepts the field:
  // src/vs/workbench/api/common/extHostLanguageFeatures.ts:1431
  showRange: (this._isAdditionsProposedApiEnabled && item.showRange)
      ? typeConvert.Range.from(item.showRange) : undefined

  Internal implementation doesn't use it:
  - InlineSuggestData constructor doesn't receive showRange parameter
  - InlineCompletionItem.create() doesn't process showRange from inlineCompletion
  - No cursor position checks against showRange exist in current codebase

  Affected Files

  - src/vs/editor/contrib/inlineCompletions/browser/model/provideInlineCompletions.ts
  - src/vs/editor/contrib/inlineCompletions/browser/model/inlineSuggestionItem.ts
  - src/vs/editor/contrib/inlineCompletions/browser/model/inlineCompletionsModel.ts

  Expected Fix

  Restore the showRange functionality:
  1. Store showRange in InlineSuggestData or InlineCompletionItem
  2. Check cursor position against this range before showing inline completions
[...]

PR #279269, opened 2025-11-25
Fix showRange field regression in InlineCompletionItem

yanmofeixi:
The showRange field in InlineCompletionItem (part of the inlineCompletionsAdditions proposed API) stopped working after the InlineCompletionsSource refactoring in commit b0abf06c7dc. While the field is still defined in the API and accepted from extensions, the implementation that checks cursor position against this range was removed.

See more detail: https://github.com/microsoft/vscode/issues/279266

This commit restores the showRange functionality by:

1. Adding showRange field to InlineSuggestData constructor
2. Passing showRange from InlineCompletion to InlineSuggestData in toInlineSuggestData
3. Adding showRange getter to InlineSuggestionItemBase
4. Implementing showRange check in InlineCompletionItem.isVisible() method
5. Adding cursor position check against showRange in inlineCompletionsModel state computation

The fix ensures that inline completions are only shown when the cursor is within the specified showRange (if provided), matching the original behavior from PR #237532.

Fixes regression from commit b0abf06c7dc (Refactors InlineCompletionsSource) Restores functionality from commit a016c0be118 (Support show range for inline edits)

### DIFF
--- a/src/vs/editor/contrib/inlineCompletions/browser/model/inlineCompletionsModel.ts
+++ b/src/vs/editor/contrib/inlineCompletions/browser/model/inlineCompletionsModel.ts
@@ -626,7 +626,14 @@ export class InlineCompletionsModel extends Disposable {
 			let edit = inlineEditResult.getSingleTextEdit();
 			edit = singleTextRemoveCommonPrefix(edit, model);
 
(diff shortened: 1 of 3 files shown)


yanmofeixi:

PR #279351, opened 2025-11-25
fix: memory leak in local process extension host

SimonSiefke:
Fixes a memory leak in local process extension host by setting the port message listener to null on dispose

```ts
portPromise.then((port) => {
	this._register(toDisposable(() => {
		port.close();
		port.onmessage = null; // here
	}));
	clearTimeout(handle);

});
```

Also changes `this._toDispose.add` to using `this._register` which overall seems a bit simpler.

### Before & After Comparison

Barely noticable, but one can see at the bottom there is one less leak. 
 
Untitled (5)

The chart shows restarting extensions 97 times, before and after.


alexdima (approved):
Thank you!

dmitrivMS (approved):

PR #263998, opened 2025-08-29
feat: create versioned resources for windows setup

deepak1556:
Fixes https://github.com/microsoft/vscode/issues/249239

PR #279444, opened 2025-11-25
fix: insiders cli launch from wsl terminal

### LINKED ISSUE #279561: WSL: `code-insiders` 1.107.0-insider fails to start (`MODULE_NOT_FOUND` for `cli.js`)

**Title:** WSL: `code-insiders` fails with `MODULE_NOT_FOUND` for `resources/app/out/cli.js` in 1.107.0-insider

**VS Code Version:** 1.107.0-insider
**Commit:** 2d69e225bfc64f111fdb9b9294a76bfbff1e99c0
**OS Version:** Windows 11 Enterprise 24H2 (Build 26100.7171) + WSL Ubuntu 22.04.5 LTS (jammy)
**Node.js (from error):** v22.21.1
**Edition:** VS Code Insiders for Windows, launched from WSL (`code-insiders`)

---

### Steps to Reproduce

1. Install VS Code Insiders 1.107.0-insider on Windows.

2. Open a WSL Ubuntu terminal.

3. Run:

   ```sh
   code-insiders .
   ```

4. Observe the error.

---

### Actual Behavior

`code-insiders` fails to start and prints:

```text
node:internal/modules/cjs/loader:1393
  throw err;
  ^

Error: Cannot find module 'C:\Users\<user>\AppData\Local\Programs\Microsoft VS Code Insiders\resources\app\out\cli.js'
    at Module._resolveFilename (node:internal/modules/cjs/loader:1390:15)
    at defaultResolveImpl (node:internal/modules/cjs/loader:1032:19)
    at resolveForCJSWithHooks (node:internal/modules/cjs/loader:1037:22)
    at Module._load (node:internal/modules/cjs/loader:1199:37)
    at c._load (node:electron/js2c/node_init:2:17993)
    at TracingChannel.traceSync (node:diagnostics_channel:328:14)
    at wrapModuleLoad (node:internal/modules/cjs/loader:244:24)
    at Module.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:186:5)
    at node:internal/main/run_main_module:36:49 {
  code: 'MODULE_NOT_FOUND',
  requireStack: []
}

Node.js v22.21.1
```

---

### Expected Behavior

`code-insiders .` from WSL should launch VS Code Insiders on Windows and open the current folder.

---

### Additional Info

* In WSL, `which code-insiders` returns:

  ```text
  /mnt/c/Users/<user>/AppData/Local/Programs/Microsoft VS Code Insiders/bin/code-insiders
  ```

* The Windows install directory looks like this:

  ```text
  Microsoft VS Code Insiders\
    Code - Insiders.exe
    bin\
    2d69e225bf\
      resources\app\out\cli.js
    (no resources\ directory at the install root)
  ```

* So the WSL launcher appears to look for:

  ```text
  <install-root>\resources\app\out\cli.js
  ```

  but in this Insiders build the `resources` directory is under a versioned subfolder (`2d69e225bf\resources\app\out\cli.js`, where `2d69e225bf` matches the first 10 chars of the commit hash).

* This reproduces for me on two different machines and two different Windows versions and Ubuntu WSL distros.

deepak1556:
Followup to https://github.com/microsoft/vscode/pull/263998

Fixes https://github.com/microsoft/vscode/issues/279561

### DIFF
--- a/src/vs/server/node/server.cli.ts
+++ b/src/vs/server/node/server.cli.ts
@@ -279,7 +279,12 @@ export async function main(desc: ProductDescription, args: string[]): Promise<vo
 		} else {
 			const cliCwd = dirname(cliCommand);
 			const env = { ...process.env, ELECTRON_RUN_AS_NODE: '1' };
-			newCommandline.unshift('resources/app/out/cli.js');
+			if (process.platform === 'win32' && cliCommand.endsWith('Code - Insiders.exe')) {
+				const versionFolder = desc.commit.substring(0, 10);
+				newCommandline.unshift(`${versionFolder}/resources/app/out/cli.js`);
+			} else {
+				newCommandline.unshift('resources/app/out/cli.js');
+			}
 			if (verbose) {
 				console.log(`Invoking: cd "${cliCwd}" && ELECTRON_RUN_AS_NODE=1 "${cliCommand}" "${newCommandline.join('" "')}"`);
 			}
--- a/src/vs/server/node/server.cli.ts
+++ b/src/vs/server/node/server.cli.ts
@@ -279,8 +279,8 @@ export async function main(desc: ProductDescription, args: string[]): Promise<vo
 		} else {
 			const cliCwd = dirname(cliCommand);
 			const env = { ...process.env, ELECTRON_RUN_AS_NODE: '1' };
-			if (process.platform === 'win32' && cliCommand.endsWith('Code - Insiders.exe')) {
-				const versionFolder = desc.commit.substring(0, 10);
+			const versionFolder = desc.commit.substring(0, 10);
+			if (fs.existsSync(join(cliCwd, versionFolder))) {
 				newCommandline.unshift(`${versionFolder}/resources/app/out/cli.js`);
 			} else {
 				newCommandline.unshift('resources/app/out/cli.js');
--- a/resources/win32/insider/bin/code.sh
+++ b/resources/win32/insider/bin/code.sh
@@ -39,7 +39,7 @@ fi
 if [ $IN_WSL = true ]; then
 
 	export WSLENV="ELECTRON_RUN_AS_NODE/w:$WSLENV"
-	CLI=$(wslpath -m "$VSCODE_PATH/resources/app/out/cli.js")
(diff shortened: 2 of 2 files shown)


Tyriar (dismissed):

src/vs/server/node/server.cli.ts
```diff
@@ -279,7 +279,12 @@ export async function main(desc: ProductDescription, args: string[]): Promise<vo
 		} else {
 			const cliCwd = dirname(cliCommand);
 			const env = { ...process.env, ELECTRON_RUN_AS_NODE: '1' };
-			newCommandline.unshift('resources/app/out/cli.js');
+			if (process.platform === 'win32' && cliCommand.endsWith('Code - Insiders.exe')) {
```
deepak1556: @aeschli a couple of things with this code,

1) I want to guard this behind insiders quality since we are not enabling versioned layout for other flavors, due to lack of quality field went with exe name check
2) I see desc.commit flow as follows, initial value from https://github.com/microsoft/vscode/blob/cf7f57448b6385aa34f8f0af97f607cad62944ea/resources/win32/insider/bin/code.sh#L9 gets passed onto `wslCode.sh` which eventually downloads a remote server if not present via `wslDownload.sh` for that commit, so it seems like the commit from the initial client script acts as the source of truth. In which case, attempting to extract back the version folder name from the server commit should be fine ?

let me know if you prefer a different approach to carry this info. Maybe a new env to carry the script path but that will require an extension side change as well..

aeschli: The command is run from WSL. I would not expect process.platform === 'win32' to be true.

My suggestion would be to do this:
```
const versionFolder = desc.commit.substring(0, 10);
if (fs.existsSync(cliCwd, versionFolder) {
  newCommandline.unshift(`${versionFolder}/resources/app/out/cli.js`);
} else {
  newCommandline.unshift('resources/app/out/cli.js');
}
```
The extra check doesn't hurt and will be removed later again. If I remember right, that code is only called from WSL.

- yes, server commit and client commit always match

aeschli (approved):

PR #279476, opened 2025-11-25
Polish custom menu + dark modern list theme color

daviddossett:
This was probably my own doing a couple years ago, but I think there is significant room for polish on both custom menus and the now-not-so-new default dark theme. This PR:
- Updates the list item height and font size, aligning the height to quick pick list items
- Updates dark modern's list item background color to move away from the dated-looking blue 
- Tones down the jarring border color for widgets

CleanShot 2025-11-25 at 13 58 48@2x

CleanShot 2025-11-25 at 14 02 09@2x

CleanShot 2025-11-25 at 14 02 42@2x

CleanShot 2025-11-25 at 14 02 54@2x


mrleemurray:
My only comment regarding colors is the selected item & hovered item colors in the Quick Pick & in chat menus are very similar & difficult to discern. Should we make them the same color or differentiate them more?

image

image

bpasero:
I would merge next iteration.

PR #279481, opened 2025-11-25
Add performance markers to inline completions

benibenj:
```Copilot Generated Description:``` Introduce performance markers to track various stages in the inline completions lifecycle, including creation, resolution, and display. This change adds methods to log performance data and updates relevant data structures to include performance information.

### DIFF
--- a/src/vs/editor/contrib/inlineCompletions/browser/model/provideInlineCompletions.ts
+++ b/src/vs/editor/contrib/inlineCompletions/browser/model/provideInlineCompletions.ts
@@ -501,13 +501,13 @@ export class InlineSuggestData {
 		);
 	}
 
-	private performance = new Performance();
+	private performance = new InlineSuggestionsPerformance();
 	public addPerformanceMarker(marker: string): void {
 		this.performance.mark(marker);
 	}
 }
 
-class Performance {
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: src/vs/editor/contrib/inlineCompletions/browser/model/provideInlineCompletions.ts [hunk 1]
```diff
@@ -498,6 +500,31 @@ export class InlineSuggestData {
 			this._correlationId,
 		);
 	}
+
+	private performance = new Performance();
+	public addPerformanceMarker(marker: string): void {
+		this.performance.mark(marker);
+	}
```
Tyriar: This looks a lot like the `performance.mark` API so I was confused at first FYI:

https://github.com/microsoft/vscode/blob/643ae5a99809dd34ae1c69d0bbbab34b9a0a2bc3/src/vs/base/common/performance.ts#L43-L51

Consider changing the name to differentiate it a little?

Tyriar (dismissed):

Yoyokrazy (dismissed):

rwoll (dismissed):

DonJayamanne (dismissed):

rzhao271 (approved):

PR #279507, opened 2025-11-26
Osortega/multi-diff-core

osortega:


src/vs/workbench/contrib/chat/browser/chatInputPart.ts
```diff
@@ -2112,6 +2112,199 @@ export class ChatInputPart extends Disposable implements IHistoryNavigationWidge
...
 	}
 
+	/**
+	 * Renders file changes in the input widget from timeline diff data.
+	 * This is used for agent sessions that don't have a traditional editing session
+	 * but instead have diff data computed from the timeline.
+	 * NOTE: this is pretty much an addaptation of renderChatEditingSessionState above
+	 * but using IEditSessionEntryDiff observable instead.
+	 */
+	renderFileChangesFromDiffs(diffs: IObservable<readonly IEditSessionEntryDiff[]>, sessionResource: URI) {
```
osortega, line 2122: We could merge this with the copy above, and just consolidate it based on the entries type but would like to hear from input on this approach first

src/vs/workbench/contrib/chat/browser/chatWidget.ts
```diff
@@ -495,6 +495,29 @@ export class ChatWidget extends Disposable implements IChatWidget {
 			this.renderChatEditingSessionState();
 		}));
 
+		this._register(autorun(r => {
```
connor4312, line 498: I would have expected this state gets controlled internally in the chat input like it's already doing for the editing session, rather than driven externally in the way.

---

Also, driving an observable input from autoruns is rather questionable. Instead, you should prefer to derive the input to the method and give it directly, something like

```ts
		const diffs = derived(r => {
			const viewModel = viewModelObs.read(r);
			const model = viewModel?.model;

			// Should we calculate the diff for all chat sessions?
			if (!model?.sessionResource || !this.chatSessionsService.getContentProviderSchemes().includes(model.sessionResource.scheme)) {
				return;
			}

			return model.editingSession?.getDiffsForFilesInSession();
		}).map((diffs, reader) => diffs?.read(reader) || []);

		this.input.renderFileChangesFromDiffs(diffs, model.sessionResource);
```

I would expect `renderFileChangesFromDiffs` handles the cases of diffEntries.length == 0, and `this.input` is set unconditionally.

src/vs/workbench/contrib/chat/browser/chatInputPart.ts
```diff
@@ -2112,6 +2112,199 @@ export class ChatInputPart extends Disposable implements IHistoryNavigationWidge
...
+				);
+			} else {
+				dom.clearNode(this.chatEditingSessionWidgetContainer);
+				this._chatEditsDisposables.clear();
+				this._chatEditList = undefined;
+			}
+		});
+	}
+
+	private renderFileChangesFromDiffsWithEntries(
```
connor4312, line 2174: Yes, these should definitely be collapsed, but I also feel string that we should not try and use identical UX to show two different things, that is confusing for users.

PR #279513, opened 2025-11-26
Submit Contributed Sessions requetsts using ChatWidget

DonJayamanne:


joshspicer (changes requested):
i think this fixes what you're trying to accomplish, is that true?

https://github.com/microsoft/vscode/pull/279664

Issue #279520, opened 2025-11-26
Quick Onboarding Snippet Generator


franciszver:
New contributors often struggle with environment setup and reproducibility traps (e.g., symlinks, mounts, config separation) so this adds a CLI tool to generate curated command snippets for new contributors.
Covers environment setup, Docker, WSL, linting, testing, and build workflows.

Usage: npm run onboarding:snippet <topic>

PR #279521, opened 2025-11-26
Feature/snippet generator

franciszver:
Quick Snippet Generator to help new contributors with environment setup.
https://github.com/microsoft/vscode/issues/279520


dmitrivMS:
See link in the issue for existing solution

PR #279562, opened 2025-11-26
Return early from rename processor if no rename provider exists.

dbaeumer:


ulugbekna (approved):
just a heads up on copilot suggestion looking reasonable

PR #279579, opened 2025-11-26
fix: fix terminal webgl context memory leak

Beace:
My testing found that on my Mac, when 10 terminals were opened and left idle for a period of time, the GPU process memory increased by 167MB (319MB ->486MB). After deleting all terminals, the memory of GPU processes will not significantly decrease.

By taking a snapshot of JavaScript, it has been confirmed that the corresponding canvas has been destroyed, but the `WebGL2RenderingContext` has not been destroyed, resulting in the inability to release rendering memory.

image


Tyriar (approved):
Thanks for spotting this, I also did some other webgl addon lifecycle improvements recently in https://github.com/microsoft/vscode/pull/285579 but missed this

lszomoru (approved):

PR #279690, opened 2025-11-27
Remove chatSessionTracker

osortega:


dmitrivMS:
Copilot seems to have good points, and CI is red...

src/vs/workbench/contrib/chat/browser/chatSessions.contribution.ts
```diff
@@ -318,6 +322,12 @@ export class ChatSessionsService extends Disposable implements IChatSessionsServ
 				stripPathStartingSeparator: true,
 			}
 		}));
+
+		const chatService = this._chatServiceLazy.value;
+		this._register(autorun(reader => {
+			const models = chatService.chatModels.read(reader);
+			this.registerModelProgressListener(models);
```
roblourens: I don't have the big picture, but it seems like a step backwards to do this in this service. I think _chatServiceLazy is trying to avoid a recursive dependency? Actually I don't really see how it helps since you still have to get a IChatService instance in the constructor of this service. And then if the end result of this is just to inform the list that the local sessions have changed, why shouldn't the local sessions provider do that?

src/vs/workbench/contrib/chat/browser/chatSessions.contribution.ts
```diff
@@ -794,58 +804,63 @@ export class ChatSessionsService extends Disposable implements IChatSessionsServ
...
-			return;
-		}
-		this._registeredModels.add(model);
-
-		// Helper function to register listeners for a request
-		const registerRequestListeners = (request: IChatRequestModel) => {
-			if (!request.response || this._registeredRequestIds.has(request.id)) {
-				return;
+	public registerModelProgressListener(models: Iterable<IChatModel>): void {
+		const disposables = new DisposableStore();
```
roblourens: This is never disposed?

src/vs/workbench/contrib/chat/browser/chatSessions.contribution.ts
roblourens: I was debugging this because I wasn't sure how it worked, and I think it works because our services are created lazily, and it would fail if the `IChatSessionsService` was referenced in the `IChatService` constructor

roblourens (changes requested):

src/vs/workbench/contrib/chat/browser/chatSessions.contribution.ts
osortega: This is because we need to do the same for all providers and not just the local chat sessions one. We need to detect progress in the chat model response and indicate the UI update so that we can show the latest progress status.
Do you have other suggestions of where to put the progress listeners if we wanted to do it for across all providers?

src/vs/workbench/contrib/chat/browser/chatSessions.contribution.ts
```diff
@@ -794,58 +804,63 @@ export class ChatSessionsService extends Disposable implements IChatSessionsServ
...
+			disposables.add(model.onDidChange(() => {
+				const currentRequests = model.getRequests();
+				currentRequests.forEach(registerRequestListeners);
+			}));
+
+			// Clean up when model is disposed
+			disposables.add(model.onDidDispose(() => {
+				this._registeredModels.delete(model);
+			}));
+		}
```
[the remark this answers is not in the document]
osortega: Doing this._register

src/vs/workbench/contrib/chat/browser/chatSessions.contribution.ts
osortega: Fixed it

src/vs/workbench/contrib/chat/browser/chatSessions.contribution.ts
roblourens: Oh I see. Maybe each provider should just be responsible for reporting on its sessions? Even if the underlying data source is the chat service, it can filter the available models to the ones that it is in charge of. Isn't it odd now that the local sessions provider has a `onDidChangeChatSessionItems` and doesn't fire it?

src/vs/workbench/contrib/chat/browser/chatSessions.contribution.ts
```diff
@@ -794,58 +804,62 @@ export class ChatSessionsService extends Disposable implements IChatSessionsServ
...
+					return;
+				}
 
-			this._registeredRequestIds.add(request.id);
+				this._registeredRequestIds.add(request.id);
 
-			this._register(request.response.onDidChange(() => {
-				callback();
-			}));
+				this._register(request.response.onDidChange(() => {
```
roblourens: registerModelProgressListener runs every time the set of models changes, but then any disposables using `this._register` in here will be retained for the lifetime of the service (so forever), I believe that would cause us to run these listeners multiple times per change and leak memory when chatmodels are disposed.

src/vs/workbench/contrib/chat/browser/chatSessions.contribution.ts
roblourens: I guess I don't quite understand how this works for non-local providers, what is the extension's responsibility and what is core's?

osortega: Yeah I think the provider having `onDidChangeChatSessionItems` and not firing is strange, the other providers do fire them in other cases but the local chat session provider is very bare bones so it doesn't really need to update for anything beyond tracking progress.
The extensions used to have the full responsability for the rendering but we want to start migrating things off of that pattern to create some rendering standard, for now status, description and icons fall into that category.

Not sure what the best pattern is for cases like this where core owns part of the UI and the extension another part of it, I guess we need to restrict the API accordingly but how do we manage the updates based on the chat model progress in this case?

roblourens: I don't quite understand and don't have full context on the API, happy to dig into it with you when we're back. But in general, I think we need a well-defined extension API, and a well-defined internal API for the view to be built on.

DonJayamanne (dismissed):

DonJayamanne (dismissed):

src/vs/workbench/contrib/chat/common/chatServiceImpl.ts
```diff
@@ -187,7 +187,7 @@ export class ChatService extends Disposable implements IChatService {
 
 		this._register(storageService.onWillSaveState(() => this.saveState()));
 
-		this.chatModels = derived(this, reader => this._sessionModels.observable.read(reader).values());
+		this.chatModels = derived(this, reader => [...this._sessionModels.observable.read(reader).values()]);
```
roblourens, line 199: Sorry for the wasted time here - I might push a change to tweak this but you can check this in.

src/vs/workbench/api/test/browser/mainThreadChatSessions.test.ts
```diff
@@ -388,6 +389,11 @@ suite('MainThreadChatSessions', function () {
 				};
 			}
 		});
+		instantiationService.stub(IChatService, new class extends mock<IChatService>() {
```
roblourens: You can use the MockChatService class

src/vs/workbench/api/browser/mainThreadChatSessions.ts
```diff
@@ -370,8 +371,34 @@ export class MainThreadChatSessions extends Disposable implements MainThreadChat
 			provider,
 			onDidChangeItems: changeEmitter,
 		});
+
+		this._register(autorun(reader => {
```
roblourens: This should be on `disposables` (for this provider) not registered to the class.

roblourens: Not necessary, but, breaking all this code into a new class that implements `IChatSessionItemProvider` might be a good way to make the lifecycle easier to reason about. When I end up with standalone DisposableStores then I usually start to think about doing that.

src/vs/workbench/api/browser/mainThreadChatSessions.ts
```diff
@@ -370,8 +371,34 @@ export class MainThreadChatSessions extends Disposable implements MainThreadChat
...
+			const models = this._chatService.chatModels.read(reader);
+			// Note that local has it's own way of handling model progress
+			for (const model of models) {
```
roblourens: Does this need to call `registerModelProgressListener`?

roblourens (commented):
Should there also be a change in LocalAgentsSessionsProvider for it to just watch local models?

src/vs/workbench/api/browser/mainThreadChatSessions.ts
osortega: Refactored `registerModelProgressListener` now both localagent provider calls that one too

src/vs/workbench/contrib/chat/common/chatServiceImpl.ts
osortega: No worries at all! Really appreciate the help ❤

meganrogge (dismissed):

Yoyokrazy (dismissed):

src/vs/workbench/api/browser/mainThreadChatSessions.ts
```diff
@@ -370,8 +369,18 @@ export class MainThreadChatSessions extends Disposable implements MainThreadChat
 			provider,
 			onDidChangeItems: changeEmitter,
 		});
+
+		disposables.add(autorun(reader => {
+			const models = this._chatService.chatModels.read(reader);
+			const onProgress = this._chatSessionsService.registerModelProgressListener(Array.from(models), chatSessionType);
+			disposables.add(onProgress(() => {
```
roblourens: You are re-adding a new event listener every time the autorun runs - I don't think that what you want, the listeners will accumulate over time

roblourens (dismissed):

src/vs/workbench/contrib/chat/browser/chatSessions.contribution.ts
```diff
@@ -833,58 +832,35 @@ export class ChatSessionsService extends Disposable implements IChatSessionsServ
...
-					const toolInvocation = part as IChatToolInvocation;
-					// Use autorun to listen for state changes
-					this._register(autorunSelfDisposable(reader => {
-						const state = toolInvocation.state.read(reader);
-
-						// Also track progress changes when executing
-						if (state.type === IChatToolInvocation.StateKind.Executing) {
-							state.progress.read(reader);
+	public registerModelProgressListener(models: IChatModel[], type: string): Event<void> {
+		const changeEmitter = new Emitter<void>();
```
roblourens: When is this disposed?

src/vs/workbench/contrib/chat/browser/chatSessions.contribution.ts
```diff
@@ -833,58 +832,35 @@ export class ChatSessionsService extends Disposable implements IChatSessionsServ
...
+		for (const model of models) {
+			if (model.sessionResource.scheme === type && !this._registeredModels.has(model)) {
+				this._registeredModels.add(model);
+				const modelDisposables = new DisposableStore();
+
+				modelDisposables.add(autorun(reader => {
+					const lastResponse = model.lastRequestObs.read(reader);
+					if (lastResponse?.response) {
+						if (responseDisposable) {
+							responseDisposable.dispose();
```
roblourens: Could be a mutable disposable?

src/vs/workbench/contrib/chat/browser/chatSessions.contribution.ts
osortega: Good catch, now I'm storing this disposable as well

src/vs/workbench/api/browser/mainThreadChatSessions.ts
osortega: Good catch, fixed it

src/vs/workbench/api/browser/mainThreadChatSessions.ts
```diff
@@ -370,8 +369,29 @@ export class MainThreadChatSessions extends Disposable implements MainThreadChat
...
+			if (progressDisposable) {
+				progressDisposable.dispose();
+			}
+			const onProgress = this._chatSessionsService.registerModelProgressListener(Array.from(models), chatSessionType);
+			progressDisposable = onProgress(() => {
+				changeEmitter.fire();
+			});
+		}));
+
+		disposables.add({
```
roblourens: Isn't this equivalent to just adding `progressDisposable` to `disposables`?

src/vs/workbench/contrib/chat/browser/chatSessions.contribution.ts
```diff
@@ -833,58 +832,34 @@ export class ChatSessionsService extends Disposable implements IChatSessionsServ
...
 	}
 
-	public registerModelProgressListener(model: IChatModel, callback: () => void): void {
-		// Prevent duplicate registrations for the same model
-		if (this._registeredModels.has(model)) {
-			return;
-		}
-		this._registeredModels.add(model);
+	public registerModelProgressListener(models: IChatModel[], type: string): Event<void> {
+		const changeEmitter = this._register(new Emitter<void>());
```
roblourens: This is still very funky overall to me. Someone can call the service and create disposables that stick around forever and have no way to be disposed. They should be owned by the calling method. We can merge it as long as it's not leaking but we just still need to revisit this for next month.

src/vs/workbench/contrib/chat/browser/chatSessions.contribution.ts
```diff
@@ -833,58 +832,34 @@ export class ChatSessionsService extends Disposable implements IChatSessionsServ
...
+		const event = changeEmitter.event;
+		const responseDisposables = this._register(new DisposableStore());
+
+		for (const model of models) {
+			if (model.sessionResource.scheme === type && !this._registeredModels.has(model)) {
+				this._registeredModels.add(model);
+				const modelDisposables = this._register(new DisposableStore());
```
roblourens: We can't register `modelDisposables` on the class if it has an independent lifecycle (which it does because it tracks the model and you clear it below). Doing this creates a memory leak because the reference to the DisposableStore and everything it touches will be retained by the ChatSessionsService forever.

roblourens (commented):
Does this fix a bug or is it just code cleanup?

src/vs/workbench/contrib/chat/browser/chatSessions.contribution.ts
osortega: @roblourens we can collect all disposables and return it to the caller but does that mean that the caller needs to listen to `onDispose` for each model and know when to dispose the model specific disposables?

src/vs/workbench/api/browser/mainThreadChatSessions.ts
osortega: Good catch

src/vs/workbench/contrib/chat/browser/chatSessions.contribution.ts
osortega: I think I understand what you meant, pushed a fix so that the caller has access to the disposables.
The function also adds disposables that listens to the disposae event of each model

src/vs/workbench/contrib/chat/browser/chatSessions.contribution.ts
osortega: Changed it so that it returns the disposable store to the caller

src/vs/workbench/api/browser/mainThreadChatSessions.ts
roblourens: I think this is still an issue? It's adding a new disposable on every autorun iteration and they accumulate over time

src/vs/workbench/api/browser/mainThreadChatSessions.ts
```diff
@@ -370,8 +369,14 @@ export class MainThreadChatSessions extends Disposable implements MainThreadChat
 			provider,
 			onDidChangeItems: changeEmitter,
 		});
+
+		disposables.add(autorun(reader => {
+			const models = this._chatService.chatModels.read(reader);
+			disposables.add(this._chatSessionsService.registerModelProgressListener(Array.from(models), chatSessionType, changeEmitter));
```
connor4312: This will create and register a new listener each time the chat models change, which will leak memory. You may want to diff to get the added/removed listeners and just register those. Check out `autorunIterableDelta` to help with this, and then maybe have `DisposableMap` to add/remove the listener from each model individually

src/vs/workbench/contrib/chat/browser/agentSessions/localAgentSessionsProvider.ts
```diff
@@ -49,7 +49,7 @@ export class LocalAgentsSessionsProvider extends Disposable implements IChatSess
 		// Listen for models being added or removed
 		this._register(autorun(reader => {
 			const models = this.chatService.chatModels.read(reader);
-			this.registerModelListeners(models);
+			this._register(this._chatSessionsService.registerModelProgressListener(Array.from(models), Schemas.vscodeLocalChatSession, this._onDidChangeChatSessionItems));
```
connor4312: same issue here ^ this leaks memory every time the models change

src/vs/workbench/contrib/chat/browser/chatSessions.contribution.ts
```diff
@@ -833,58 +832,29 @@ export class ChatSessionsService extends Disposable implements IChatSessionsServ
...
+			if (model.sessionResource.scheme === type && !this._registeredModels.has(model)) {
+				this._registeredModels.add(model);
+				disposables.add(autorun(reader => {
+					const lastResponse = model.lastRequestObs.read(reader);
+					if (lastResponse?.response) {
+						disposables.add(lastResponse.response.onDidChange(() => {
+							emitter.fire();
+						}));
+					}
+				}));
```
connor4312: This too leaks a bit because you will retain listeners for all responses, I suggest something more like

```ts
				const lastResponse = model.lastRequestObs.map(last => {
					return last.response ? observableSignalFromEvent(last.response.onDidChange) : undefined;
				});
				disposables.add(autorun(reader => {
					lastResponse.read(reader)?.read(reader);
					emitter.fire();
				}));
```

src/vs/workbench/contrib/chat/browser/chatSessions.contribution.ts
```diff
@@ -833,58 +832,29 @@ export class ChatSessionsService extends Disposable implements IChatSessionsServ
...
-		requests.forEach(registerRequestListeners);
-
-		// Listen for new requests being added
-		this._register(model.onDidChange(() => {
-			const currentRequests = model.getRequests();
-			currentRequests.forEach(registerRequestListeners);
-		}));
+				disposables.add(model.onDidDispose(() => {
+					this._registeredModels.delete(model);
+					disposables.dispose();
```
connor4312: Seems incorrect, you have multiple sessions in the disposables but dispose all of them when any model gets disposed

connor4312 (changes requested):
some code-level things you probably want to fix

src/vs/workbench/contrib/chat/browser/chatSessions.contribution.ts
```diff
@@ -877,60 +875,43 @@ export class ChatSessionsService extends Disposable implements IChatSessionsServ
...
 
-			// Track tool invocation state changes
-			const responseParts = request.response.response.value;
-			responseParts.forEach((part: IChatProgressResponseContent) => {
-				if (part.kind === 'toolInvocation') {
-					const toolInvocation = part as IChatToolInvocation;
-					// Use autorun to listen for state changes
-					this._register(autorunSelfDisposable(reader => {
-						const state = toolInvocation.state.read(reader);
+		const listeners = new ResourceMap<IDisposable>();
```
connor4312, line 904: I think we actually want to return a DisposableStore containing both the `autorunIterableDelta` and something like

```ts
store.add(toDisposable(() => {
  for (const listener of listeners) { listener.dispose(); }
}));
```

otherwise they might get orphaned

src/vs/workbench/contrib/chat/browser/chatSessions.contribution.ts
```diff
@@ -877,60 +875,43 @@ export class ChatSessionsService extends Disposable implements IChatSessionsServ
...
+				});
+				addedValues.forEach((added) => {
+					const lastResponseListener = new MutableDisposable();
+					listeners.set(added.sessionResource, autorun(reader => {
+						const lastResponse = added.lastRequestObs.read(reader);
+						if (lastResponse?.response) {
+							lastResponseListener.value = lastResponse.response.onDidChange(() => {
+								onChange();
+							});
 						}
```
connor4312: We also want to make sure to dispose the `lastResponseListener` when the session gets removed. Alternatively we can avoid that by doing something like


```suggestion
					const lastResponseListener = new MutableDisposable();
					const changedSignal = added.lastRequestObs.map(last => last?.response && observableSignalFromEvent(last.response.onDidChange));

					listeners.set(added.sessionResource, autorun(reader => {
						changedSignal.read(reader)?.read(reader); 
						onChange();
					}));
}
```

(downside is this fires once the first time the new session gets added, but maybe that's actually desirable?)

connor4312 (dismissed):

rebornix (dismissed):

joshspicer (approved):

PR #279703, opened 2025-11-27
Agent session progress clean up

osortega:


src/vs/workbench/contrib/chat/browser/chatSessions.contribution.ts
```diff
@@ -879,36 +880,40 @@ export class ChatSessionsService extends Disposable implements IChatSessionsServ
...
-			const fileName = groups?.path?.split('/').pop() || groups?.path || '';
-			return (groups?.linkText?.trim() || fileName);
-		});
+	private cleanUpProgress(progress: IMarkdownString | string | undefined): string {
+		let cleaned = '';
+		if (typeof progress === 'string') {
+			cleaned = progress;
+		} else if (progress) {
+			// Remove file links but keep the link text or filename
+			cleaned = progress.value.replace(/\[(?<linkText>[^\]]*)\]\(file:\/\/\/(?<path>[^)]+)\)/g, (match: string, _p1: string, _p2: string, _offset: number, _string: string, groups?: { linkText?: string; path?: string }) => {
```
mjbvz: We have a `renderAsPlaintext` helper that can strip out markdown syntax. Maybe what you are looking for here too?

mjbvz: You can also use `marked.lexer` to parse the string and find links in a more reliable way

mjbvz (dismissed):
Let's see if it's possible to avoid having custom parsing logic

src/vs/workbench/contrib/chat/browser/chatSessions.contribution.ts
osortega: Screenshot 2025-11-26 at 6 51 46 PM
Screenshot 2025-11-26 at 6 54 03 PM
@mjbvz using the string doesn't strip out the markdown link syntax and using the markdown just remove the link altogether :/

mjbvz: I think `renderAsPlaintext` should work if you pass in a markdown string. You can create one if you need to:

```ts
console.log(renderAsPlaintext(new MarkdownString('[text](http://example.com)')));
// Logs 'text'
```

osortega: @mjbvz the issue here is that we want to display part of the link so that we indicate which file is being modified.
That helper seems to be stripping out the link completely if the hyperlink doesn't have a text like `[]`
Screenshot 2025-12-02 at 4 38 02 PM
Screenshot 2025-12-02 at 4 38 52 PM

mjbvz: The empty link seems very fishy to me but if you can't change the markdown itself, look at how `renderAsPlaintext` works. You need to override the link renderer. The `includeCodeBlocksFences` option shows one example of this

mjbvz: However I'd look at why you have an empty link. That really would be invisible to the user if rendered directly as html

osortega: This is the response coming from github.com tool invocation, we don't have much control over it but I just added the new renderer to handle this cases

src/vs/base/browser/markdownRenderer.ts
```diff
@@ -750,6 +759,18 @@ const plainTextWithCodeBlocksRenderer = new Lazy<marked.Renderer>(() => {
 	return renderer;
 });
 
+const plainTextWithFileLinksRenderer = new Lazy<marked.Renderer>(() => {
+	const renderer = createPlainTextRenderer();
+	renderer.link = ({ text, href }: marked.Tokens.Link): string => {
+		if (href && href.startsWith('file://')) {
+			const fileName = href.split('/').pop() || href;
+			return text.trim() || fileName;
```
mjbvz: This should use `basename` to get the file name

src/vs/base/browser/markdownRenderer.ts
```diff
@@ -662,7 +664,14 @@ export function renderAsPlaintext(str: IMarkdownString | string, options?: {
 		value = `${value.substr(0, 100_000)}…`;
 	}
 
-	const html = marked.parse(value, { async: false, renderer: options?.includeCodeBlocksFences ? plainTextWithCodeBlocksRenderer.value : plainTextRenderer.value });
+	let renderer = plainTextRenderer.value;
+	if (options?.includeCodeBlocksFences) {
+		renderer = plainTextWithCodeBlocksRenderer.value;
+	} else if (options?.preserveFileLinks) {
```
mjbvz: Using the API it would be surprising you can't set both `includeCodeBlocksFences` and `preserveFileLinks` can be set. Can we make it so both can be enabled?

src/vs/base/browser/markdownRenderer.ts
```diff
@@ -651,6 +651,8 @@ function getDomSanitizerConfig(mdStrConfig: MdStrConfig, options: MarkdownSaniti
 export function renderAsPlaintext(str: IMarkdownString | string, options?: {
 	/** Controls if the ``` of code blocks should be preserved in the output or not */
 	readonly includeCodeBlocksFences?: boolean;
+	/** Controls if file:// links should be preserved as filenames in the output */
+	readonly preserveFileLinks?: boolean;
```
mjbvz: Instead of something so specific, could this be  a `linkFormatter` that takes the text and href and returns the text to display? That way the caller can handle all the file uri / chat specific logic

src/vs/workbench/contrib/chat/browser/chatSessions.contribution.ts
```diff
@@ -944,36 +946,35 @@ export class ChatSessionsService extends Disposable implements IChatSessionsServ
...
-		});
+	private cleanUpProgress(progress: IMarkdownString | string | undefined): string {
+		let cleaned = '';
+		if (typeof progress === 'string') {
+			cleaned = progress;
+		} else if (progress) {
+			cleaned = renderAsPlaintext(progress, { preserveFileLinks: true });
+		}
+		// Remove ticks
+		return cleaned.replace(/`+/g, '');
```
mjbvz: Is this still needed after adopting `renderAsPlaintext`?

src/vs/workbench/contrib/chat/browser/chatSessions.contribution.ts
```diff
@@ -944,36 +946,35 @@ export class ChatSessionsService extends Disposable implements IChatSessionsServ
...
 	}
 
-	private extractFileNameFromLink(filePath: string): string {
-		return filePath.replace(/\[(?<linkText>[^\]]*)\]\(file:\/\/\/(?<path>[^)]+)\)/g, (match: string, _p1: string, _p2: string, _offset: number, _string: string, groups?: { linkText?: string; path?: string }) => {
-			const fileName = groups?.path?.split('/').pop() || groups?.path || '';
-			return (groups?.linkText?.trim() || fileName);
-		});
+	private cleanUpProgress(progress: IMarkdownString | string | undefined): string {
+		let cleaned = '';
+		if (typeof progress === 'string') {
```
mjbvz: You should be able to pass strings to `renderAsPlaintext` which will return them directly so you don't need this conditional

src/vs/workbench/contrib/chat/browser/chatSessions.contribution.ts
```diff
@@ -944,36 +946,35 @@ export class ChatSessionsService extends Disposable implements IChatSessionsServ
...
-		return description;
+		return this.cleanUpProgress(description);
 	}
 
-	private extractFileNameFromLink(filePath: string): string {
-		return filePath.replace(/\[(?<linkText>[^\]]*)\]\(file:\/\/\/(?<path>[^)]+)\)/g, (match: string, _p1: string, _p2: string, _offset: number, _string: string, groups?: { linkText?: string; path?: string }) => {
-			const fileName = groups?.path?.split('/').pop() || groups?.path || '';
-			return (groups?.linkText?.trim() || fileName);
-		});
+	private cleanUpProgress(progress: IMarkdownString | string | undefined): string {
```
mjbvz: Where is the progress markdown string coming from? Could we update whatever is generating it to have better text for the link instead of using the href?

src/vs/base/browser/markdownRenderer.ts
```diff
@@ -750,6 +759,18 @@ const plainTextWithCodeBlocksRenderer = new Lazy<marked.Renderer>(() => {
 	return renderer;
 });
 
+const plainTextWithFileLinksRenderer = new Lazy<marked.Renderer>(() => {
+	const renderer = createPlainTextRenderer();
+	renderer.link = ({ text, href }: marked.Tokens.Link): string => {
+		if (href && href.startsWith('file://')) {
```
mjbvz: I see bad formatting for other schemes too, such as `untitled`:

Image

src/vs/workbench/contrib/chat/browser/chatSessions.contribution.ts
osortega: Yes because `renderAsPlaintext` returns the same string if it's of type string. And it's usually a string and not a markdown the one that has the ```

mjbvz: So if `typeof progress  === 'string'`, it could still actually be markdown and needs special markdown characters stripped out? That seems odd to me

Could we update whoever generates `progress` to make sure it always a `MarkdownString`?

src/vs/base/browser/markdownRenderer.ts
osortega: This is the model title, will fix that separately

src/vs/base/browser/markdownRenderer.ts
```diff
@@ -740,15 +749,18 @@ function createPlainTextRenderer(): marked.Renderer {
...
-	return renderer;
-});
+const codeBlockFences = ({ text }: marked.Tokens.Code): string => {
+	return `\n\`\`\`\n${escape(text)}\n\`\`\`\n`;
+};
+
+const linkFormatter = ({ text, href }: marked.Tokens.Link): string => {
+	if (href) {
+		const uri = URI.parse(href);
+		return text.trim() || basename(uri);
```
mjbvz: I wonder if this should just be the default behavior for rendering links that don't have text?

src/vs/base/browser/markdownRenderer.ts
```diff
@@ -740,15 +749,18 @@ function createPlainTextRenderer(): marked.Renderer {
...
-	};
-	return renderer;
-});
+const codeBlockFences = ({ text }: marked.Tokens.Code): string => {
+	return `\n\`\`\`\n${escape(text)}\n\`\`\`\n`;
+};
+
+const linkFormatter = ({ text, href }: marked.Tokens.Link): string => {
+	if (href) {
+		const uri = URI.parse(href);
```
mjbvz: Make sure to wrap this in a try catch in case it's not a valid URI. In those cases you could still try using `basename` from `path.ts` instead of our uri based one

src/vs/base/browser/markdownRenderer.ts
```diff
@@ -651,6 +650,8 @@ function getDomSanitizerConfig(mdStrConfig: MdStrConfig, options: MarkdownSaniti
 export function renderAsPlaintext(str: IMarkdownString | string, options?: {
 	/** Controls if the ``` of code blocks should be preserved in the output or not */
 	readonly includeCodeBlocksFences?: boolean;
+	/** Controls if we want to format links from "Link [text](file)" to "Link text" */
```
mjbvz: This comment is a bit off since that's already the default behavior. Maybe something about handling of empty links?

src/vs/base/browser/markdownRenderer.ts
osortega: Leaving it as an option for now and we can always decide if more callers need this

src/vs/base/browser/markdownRenderer.ts
osortega: Good idea, changed it

src/vs/base/browser/markdownRenderer.ts
osortega: Changed it

rebornix (approved):

src/vs/workbench/contrib/chat/browser/chatSessions.contribution.ts
```diff
@@ -944,36 +946,24 @@ export class ChatSessionsService extends Disposable implements IChatSessionsServ
...
 
-					if (description) {
-						description = this.extractFileNameFromLink(description);
-					}
 					if (state.type === IChatToolInvocation.StateKind.WaitingForConfirmation) {
 						const message = toolInvocation.confirmationMessages?.title && (typeof toolInvocation.confirmationMessages.title === 'string'
 							? toolInvocation.confirmationMessages.title
 							: toolInvocation.confirmationMessages.title.value);
-						description = message ?? localize('chat.sessions.description.waitingForConfirmation', "Waiting for confirmation: {0}", description);
+						description = message ?? localize('chat.sessions.description.waitingForConfirmation', "Waiting for confirmation: {0}", typeof description === 'string' ? description : description.value);
```
mjbvz, line 955: Should this use `renderAsPlainText` too? Or even better, preserve it as a markdown string if it starts as one?

src/vs/base/browser/markdownRenderer.ts
```diff
@@ -662,7 +664,15 @@ export function renderAsPlaintext(str: IMarkdownString | string, options?: {
 		value = `${value.substr(0, 100_000)}…`;
 	}
 
-	const html = marked.parse(value, { async: false, renderer: options?.includeCodeBlocksFences ? plainTextWithCodeBlocksRenderer.value : plainTextRenderer.value });
+	const renderer = createPlainTextRenderer();
+	if (options?.includeCodeBlocksFences) {
+		renderer.code = codeBlockFences;
+	}
+	if (options?.useLinkFormatter) {
```
mjbvz, line 671: `useLinkFormatter` still feels like an odd name to me. Could we try doing one of these (best to worst):

1. Make whoever is generating the string generate a valid link in the first place
1. Make the new link logic the default behavior so we don't need an option. That seems reasonable if there's no `text`
1. Give the option a more clear name such as `useHrefForLinks`

PR #279740, opened 2025-11-27
chat - tweak behaviour of `openSession`

bpasero:
@roblourens making this PR to solve a few paper cuts around `openSession` when used from agents view. I am leaving comments inline to better explain the rationale. Addresses these issues:
* double-click from agent sessions view does not pin a chat editor
* opening from agent sessions view does not enforce opening chats in editor if opened in panel chat
* opening from panel chat does not enforce opening chat in panel chat if opened as chat editor


src/vs/workbench/contrib/chat/browser/agentSessions/agentSessionsControl.ts
```diff
@@ -153,7 +153,6 @@ export class AgentSessionsControl extends Disposable {
 		sessionOptions.ignoreInView = true;
 
 		const options: IChatEditorOptions = {
-			preserveFocus: false,
```
bpasero, line 196: This should not be defined explicitly but be derived from `e.editorOptions` where this is set properly.

src/vs/workbench/contrib/chat/browser/chatWidgetService.ts
```diff
@@ -94,46 +93,57 @@ export class ChatWidgetService extends Disposable implements IChatWidgetService
...
-			if (!preserveFocus) {
+			if (!options?.preserveFocus) {
 				view?.focus();
 			}
 			return chatView.widget;
 		}
 
 		// Already open in an editor?
-		const existingEditor = this.findExistingChatEditorByUri(sessionResource);
+		const existingEditor = this.editorService.findEditors({ resource: sessionResource, typeId: ChatEditorInput.TypeID, editorId: ChatEditorInput.EditorID }).at(0);
```
bpasero, line 136: We have `IEditorService.findEditors` to help you find an existing editor, no need to iterate over groups yourself.

src/vs/workbench/contrib/chat/browser/chatWidgetService.ts
```diff
@@ -94,46 +93,57 @@ export class ChatWidgetService extends Disposable implements IChatWidgetService
...
-			return chatViewPane?.widget;
+			return chatView?.widget;
 		}
 
 		// Open in chat editor
-		const pane = await this.editorService.openEditor({ resource: sessionResource, options }, target);
+		const pane = await this.editorService.openEditor({
+			resource: sessionResource, options: {
+				...options,
+				revealIfOpened: true // always try to reveal if already opened
```
bpasero, line 118: We maintain an editor option `revealIfOpened` to prefer revealing an opened editor if it exists, no need to manually search for it.

src/vs/workbench/contrib/chat/browser/chatWidgetService.ts
```diff
@@ -94,46 +93,57 @@ export class ChatWidgetService extends Disposable implements IChatWidgetService
 	openSession(sessionResource: URI, target?: typeof ChatViewPaneTarget): Promise<IChatWidget | undefined>;
 	openSession(sessionResource: URI, target?: PreferredGroup, options?: IChatEditorOptions): Promise<IChatWidget | undefined>;
 	async openSession(sessionResource: URI, target?: typeof ChatViewPaneTarget | PreferredGroup, options?: IChatEditorOptions): Promise<IChatWidget | undefined> {
-		const alreadyOpenWidget = await this.revealSessionIfAlreadyOpen(sessionResource, options?.preserveFocus);
-		if (alreadyOpenWidget) {
-			return alreadyOpenWidget;
+		// Reveal if already open unless a explicit target is specified
+		if (typeof target === 'undefined') {
```
bpasero: This is up for debate: if I am passing in an explicit target (chat view or editor), I want that to be respected because otherwise you end up with a weird interaction where sometimes it would open an editor and sometimes it would reveal in the sidebar. We want sessions from the agent sessions view to always open in the editor.

However: if you click on a session that is also live in a sidebar, you essentially end up having it opened twice, which seems to work fine, but nevertheless did not happen before. Should we detect this case and then move the session from the sidebar into the editor? I guess the same can happen if you have it open in the editor and want to open it in the sidebar.

roblourens: I don't see why anyone would want to have the same chat opened twice. It doesn't work well IMO now that we have one 'input model' attached to the ChatModel.

Forcibly moving chats around from the session view seems weird, if I started a chat in the view, why can't I just continue it there?

bpasero: @roblourens first of all: if you provide an API that allows me to pass in an editor group, then you should support that and not just decide to open somewhere else.

And then: the agent sessions view should never open or reveal a chat in the side bar, it would be very weird to click through sessions in the sidebar and sometimes you end up in a different view and sometimes in the editor. For a user that is very unpredictable, almost as if a file from the explorer sometimes open in the window and sometimes in a new window. Its crucial that the experience is consistent.

You may have used a sidebar chat without real intent, just because you ended there, thats up to the user. But then I would still expect to open the chat in the editor so that the experience is consistent.

At the same time, ideally all sessions opened from within the sidebar open in the sidebar. This is something we can currently not do for background/cloud chats, but should be able to do so in the future.

roblourens: The chat is in the view because the user put it there- I just don't really get why not allow them to continue it there? That's not unpredictable to me, it would be unpredictable to move my chat in some other new location.

From an API perspective, it would make sense to me to move the chat to the requested editor group unless `revealIfOpened` is set. But I don't think we should open a chat in a second view by default because I don't think I would ever need my window to look like this? 
image

roblourens: I guess I'm not thinking of the sessions view as like a file explorer, more like tabs that show my various running agents in different places. If I click one that is open in a view I expect it to reveal the view, not move them around.

roblourens: I think it's up to you how it works though- if the change to openSession that I described makes sense, I could make it work that way. Or do you think we should leave the chat in the original location and also open a second instance?

roblourens: And if we want to just open a second copy, then we should clean up the things that are weird about this right now (mainly the chat input kept in sync between the two copies) although it works fine afaik, it's just cosmetic.

bpasero: Ok lets wait for more real usage feedback on this one before making a decision.

bpasero: I found another case which does not work well and suggest we continue in https://github.com/microsoft/vscode/issues/280128

chrmarti (approved):

PR #279851, opened 2025-11-27
Fix terminal search overwriting clipboardFix terminal search overwriting clipboard

### LINKED ISSUE #272894: Terminal: highlighted (matched) search term CONSTANTLY overwrites clipboard

Version: 1.105.1 (Universal)
Electron: 37.6.0
ElectronBuildId: 12502201
Chromium: 138.0.7204.251
Node.js: 22.19.0
V8: 13.8.258.32-electron.0
OS: Darwin x64 23.6.0

- OS Version: 
MacOS Sonoma 14.7.7 (23H723)

This looks like https://github.com/microsoft/vscode/issues/151902 is still unfixed despite 4 years in making and being closed. 

Steps to Reproduce:

1. Search for a term in Terminal pane
2. Copy something into clipboard (not necessarily inside the Terminal pane, as in the video repro below)
3. Paste the copied term into the terminal _(or simply, interact with the terminal in any way),_ then paste again, and observe  that the clipboard item has been overwritten by your previous search term from the terminal.

To clarify: pasting action is not necessary to lose clipboard contents, it is simply for repro purposes. In real life scenarios, any interaction with the Terminal pane can replace clipboard if terminal search field is open and populated.

The way I see it, is that searching highlights found items (in terminal pane), and this highlighting acts as if user has selected something, hence copy-on-select is triggered and clipboard is replaced with the found / highlighted string.

https://github.com/user-attachments/assets/052fe079-3c7c-4c21-b36b-fd113ba0377d

Bitshifter-9:
Fixes #272894

## Summary
When searching in the terminal with 'copy on selection' enabled, the search term was being automatically copied to the clipboard, overwriting any previously copied content.

## Root Cause
The search creates a selection which triggers the `onSelectionChange` event handler in the clipboard contribution. This handler automatically copies the selection when the 'copy on selection' setting is enabled, without distinguishing between user-initiated and programmatic selections.

## Solution
Extended the existing copy-on-selection override mechanism to be active whenever the find widget is shown, not just when it has focus. The override is now:
- Set when `show()` or `reveal()` is called
- Cleared when `hide()` is called

This prevents search-related selections from triggering clipboard copies while preserving normal copy-on-selection behavior for user-initiated selections.

## Testing
Manually tested with the following scenarios:
1. Search in terminal with 'copy on selection' enabled
2. Copy other content to clipboard
3. Interact with terminal (paste, close find widget)
4. Verify clipboard contains the manually copied content, not the search termWhen searching in the terminal with 'copy on selection' enabled, the search term was being automatically copied to the clipboard, overwriting any previously copied content. This happened because the search creates a selection which triggers the onSelectionChange event handler.

This fix extends the existing copy-on-selection override mechanism to be active whenever the find widget is shown, not just when it has focus. The override is now set when show() or reveal() is called and cleared when hide() is called.

Fixes #272894

### DIFF
--- a/src/vs/workbench/contrib/terminalContrib/find/browser/terminalFindWidget.ts
+++ b/src/vs/workbench/contrib/terminalContrib/find/browser/terminalFindWidget.ts
@@ -115,6 +115,11 @@ export class TerminalFindWidget extends SimpleFindWidget {
 	}
 
 	override reveal(): void {
+		// Disable copy-on-selection for search-related selections to prevent overwriting clipboard
+		if (!this._overrideCopyOnSelectionDisposable && TerminalClipboardContribution.get(this._instance)?.overrideCopyOnSelection) {
+			this._overrideCopyOnSelectionDisposable = TerminalClipboardContribution.get(this._instance)?.overrideCopyOnSelection(false);
+		}
+
 		const initialInput = this._instance.hasSelection() && !this._instance.selection!.includes('\n') ? this._instance.selection : undefined;
 		const inputValue = initialInput ?? this.inputValue;
 		const xterm = this._instance.xterm;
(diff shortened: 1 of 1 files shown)


anthonykim1:
Thank you for the PR.
Let's close this against: https://github.com/microsoft/vscode/pull/285276

PR #279871, opened 2025-11-28
Fix: do not show implicit activation warning for engines <1.75 (#251898)

### LINKED ISSUE #251898: Do not suggest implicit activation message when engine does not support it
When targeting versions prior than 1.75 and `activationEvents` contains events that implicitly activate the extension, like `onCommand`
```jsonc
/* 
 * package.json: 
 */

{
	/* ... */
	"engines": {
		"vscode": "^1.74.0"
	},
	"activationEvents": [
		"onCommand:x.a",
		"onCommand:x.b",
		"onCommand:x.c",
		"onCommand:x.d",
		"onCommand:x.e"
	],
	/* ... */
}
```

the problems panel shows:
`This activation event can be removed for extensions targeting engine version ^1.75 as VS Code will generate these automatically from your package.json contribution declarations.`
for every such event.

Image

This is unnecessarily annoying. I hope this can be fixed by someone as a https://github.com/microsoft/vscode/labels/good%20first%20issue
This message is pushed by this code:
https://github.com/microsoft/vscode/blob/315b0949786b3807f05cb6acd13bf0029690a052/extensions/extension-editing/src/extensionLinter.ts#L36
https://github.com/microsoft/vscode/blob/315b0949786b3807f05cb6acd13bf0029690a052/extensions/extension-editing/src/extensionLinter.ts#L167-L172

Also it looks like the source code does not suppress this message for versions 2.0.0<=>2.74.x

I noticed the source code of this builtin extension `extension-editing` does not contain any *.test.js files.
Are these tests available somewhere else? Or is this extension currently not tested?

Dsp023:
This PR fixes an issue where the VS Code extension linter displays the “redundant implicit activation event” warning even when the extension’s "engines.vscode" version does not support implicit activation (versions earlier than 1.75).

Problem

Implicit activation events were introduced in VS Code 1.75, but the linter currently shows the warning regardless of the engine version.

Example of incorrect behavior:

"engines": { "vscode": "^1.74.0" }

Still triggers:

“This activation event can be removed for extensions targeting engine version ^1.75...”

Fix

Before emitting the diagnostic, the linter now checks:

if (!isImplicitActivationSupported) {
    continue;
}

This ensures:

Extensions targeting versions < 1.75 → no warning shown

Extensions targeting >= 1.75 → warning shown correctly

Why This Fix Is Safe

Uses the existing boolean: isImplicitActivationSupported

Does not change any behavior for supported versions

Does not modify any other diagnostics

Minimally scoped, easy to review

Related Issue

Fixes: #251898


extensions/extension-editing/src/extensionLinter.ts
```diff
@@ -2,6 +2,12 @@
  *  Copyright (c) Microsoft Corporation. All rights reserved.
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
+function engineSupportsImplicitActivation(engineRange: string | undefined): boolean {
+    if (!engineRange) return false;
+
+    const semver = require('semver');
+    return semver.intersects(engineRange, '>=1.75.0');
+}
```
alexdima, line 11: This function is not invoked anywhere

extensions/extension-editing/src/extensionLinter.ts
```diff
@@ -165,6 +171,9 @@ export class ExtensionLinter {
 						const isImplicitActivationSupported = info.engineVersion && info.engineVersion?.majorBase >= 1 && info.engineVersion?.minorBase >= 75;
 						// Redundant Implicit Activation
 						if (info.implicitActivationEvents?.has(activationEvent) && redundantImplicitActivationEventPrefixes.some((prefix) => activationEvent.startsWith(prefix))) {
+							if (!isImplicitActivationSupported) {
+								   continue;
+							   }
```
alexdima, line 176: This code is not formatted correctly

alexdima (changes requested):

PR #279992, opened 2025-11-28
chat: allow attaching debug resources to chat

connor4312:
Some initial work. Got tired of manually adding debug context :)

https://github.com/user-attachments/assets/9e2a34a7-3744-4284-bdab-8f6f7058ee83


jrieken (approved):

erezak:
@connor4312 Great! Looks like a time-saver.
Hey, could we add something like a "pin state" action that saves a snapshot of the current paused location plus selected debug items (variables, scopes, watches) as a named attachment?
This way, we can continue debugging without losing that contact, and we can continue referring to it in chat.

PR #269635, opened 2025-10-02
Update katex regex for jQuery expressions in KaTeX matching

nikdmello:
Updates katex regex coverage for jQuery expressions with multiple dollar signs.

Ensures that when jQuery expressions like `$.getJSON, $.ajax, $.get `and `$("#dialogDetalleZona").dialog(...) / $("#dialogDetallePDC").dialog(...)` appear on the same line, the dollar signs are not incorrectly paired as opening and closing KaTeX delimiters.

Fixes #268378

PR #280021, opened 2025-11-28
fix: update mathInlineRegExp to correctly handle $(a+b)^2$

tamuratak:
fix: update mathInlineRegExp to correctly handle `$(a+b)^2$`. Fix #279080

Allowing `$(` should be fine, since it’s very common in mathematical expressions. It should be enough to block only `$.`, `$("` and `$('`.

cc: @mjbvz 

related to #269635


mjbvz (approved):
Thanks!

rzhao271 (approved):

PR #280217, opened 2025-12-01
Fix illegal characters in Dynamic Auth Provider logger filename

Josbleuet:
The Dynamic Auth Provider was using the full authorization server URL (e.g., 'https://example.com/') as the logger ID, which caused 'Illegal characters in path' errors on Windows when mkdirp tried to create the log file directory.

This change sanitizes the provider ID before using it as a logger filename by replacing characters illegal in file paths (:, /, ?, *, etc.) with underscores.

Changes:
- Added sanitizeForFilename() private method to DynamicAuthProvider
- Updated logger creation to use sanitized ID
- Added unit tests for filename sanitization

Example transformation:
  Before: 'https://example.com/' After:  'https___example.com_'

Fixes dynamic auth provider initialization on Windows with MCP servers.

### DIFF
--- a/src/vs/platform/log/common/log.ts
+++ b/src/vs/platform/log/common/log.ts
@@ -649,7 +649,7 @@ export abstract class AbstractLoggerService extends Disposable implements ILogge
 	}
 
 	protected toResource(idOrResource: string | URI): URI {
-		return isString(idOrResource) ? joinPath(this.logsHome, `${idOrResource}.log`) : idOrResource;
+		return isString(idOrResource) ? joinPath(this.logsHome, `${idOrResource.replace(/[\\/:\*\?"<>\|]/g, '_')}.log`) : idOrResource;
 	}
 
 	setLogLevel(logLevel: LogLevel): void;


src/vs/workbench/api/common/extHostAuthentication.ts
```diff
@@ -404,7 +404,9 @@ export class DynamicAuthProvider implements vscode.AuthenticationProvider {
 		// Auth Provider label is just the resource name if provided, otherwise the authority of the authorization server.
 		this.label = _resourceMetadata?.resource_name ?? this.authorizationServer.authority;
 
-		this._logger = loggerService.createLogger(this.id, { name: `Auth: ${this.label}` });
+		// Sanitize the ID for use as a logger file name (remove characters illegal in file paths)
+		const loggerIdSafe = this.sanitizeForFilename(this.id);
```
TylerLeonhardt: Maybe instead just url encode it?

Josbleuet: # Response to Reviewer

Good question about URL encoding. I looked into the VS Code codebase to see how this kind of thing is typically handled.

## What I found

I searched through the existing logger creation patterns and filename sanitization in the codebase. A few things stood out:

**How logger IDs become filenames:**

In `src/vs/platform/log/common/log.ts` (lines 653-655), when you pass a string ID, it goes straight into `joinPath()`:

```typescript
protected toResource(idOrResource: string | URI): URI {
    return isString(idOrResource) ? joinPath(this.logsHome, `${idOrResource}.log`) : idOrResource;
}
```

There's no sanitization or encoding happening there. That's why we hit the Windows path error with characters like `:`, `/`, `?`.

**Existing patterns:**

Looking at 50+ logger creation calls across the codebase, they all use simple IDs like `'network'`, `'terminal'`, `'remoteagent'`. I couldn't find any examples of URLs or complex strings being used as logger IDs. Also couldn't find any standard utility function for making arbitrary strings filesystem-safe.

**Similar case I found:**

The user profile export code does something similar in `userDataProfileImportExportService.ts` (line 277):

```typescript
await profileContentHandler.saveProfile(profile.name.replace('/', '-'), ...)
```

They replace `/` with `-` rather than URL encoding it.

## Why I went with character replacement

The main reason is readability. When you're debugging MCP auth issues, you'll see log files in `.vscode/logs/`. Compare:

- `https___example.com_oauth_authorize.log`
- `https%3A%2F%2Fexample.com%2Foauth%2Fauthorize.log`

The first one is immediately recognizable as a URL. The second looks like gibberish until you mentally decode it.

Also, it's consistent with what I found in the codebase - simple transformations rather than encoding. And underscore is safe everywhere (Windows, macOS, Linux, all shells).

That said, if there's a preference for URL encoding or another approach, I'm happy to change it. Just wanted to explain the reasoning.

Josbleuet (commented):
Good catch! You're absolutely right that this should be handled at the logger service level. I've refactored the fix to add the sanitization directly in `AbstractLoggerService.toResource()` instead of in the auth provider.

Now any code that passes a string with invalid filename characters as a logger ID will be automatically protected. Much better solution than fixing it in one place.

Updated in the latest commit: e526f0937dc

Josbleuet:

src/vs/platform/log/common/log.ts
```diff
@@ -649,7 +649,23 @@ export abstract class AbstractLoggerService extends Disposable implements ILogge
...
+	/**
+	 * Sanitizes a string to be safe for use as a filename.
+	 * Replaces characters that are illegal in Windows filenames with underscores.
+	 * @param str The string to sanitize
+	 * @returns A filesystem-safe version of the string
+	 */
+	private sanitizeForFilename(str: string): string {
+		// Replace characters that are illegal in Windows filenames: < > : " / \ | ? *
+		// Also replace whitespace for consistency
+		return str.replace(/[<>:"/\\|?*\s]/g, '_');
```
sandy081:
```suggestion
	return isString(idOrResource) ? joinPath(this.logsHome, `${idOrResource.replace(/[\\/:\*\?"<>\|]/g, '')}.log`) : idOrResource;
```

sandy081 (changes requested):

Josbleuet:
I've updated the PR with your suggested approach - using an inline replace instead of a separate sanitization function.

sandy081 (approved):

aeschli (approved):

PR #280222, opened 2025-12-01
openSession should move an existing session to the requested target

### LINKED ISSUE #280128: Chat: revisit reveal logic of `openSession`
(continuation from https://github.com/microsoft/vscode/pull/279740)

Today `openSession` will force to reveal a chat as a first thing:

https://github.com/microsoft/vscode/blob/8ccd1dae54571a5e8e9ca33aa2ec7509dc39ba03/src/vs/workbench/contrib/chat/browser/chatWidgetService.ts#L97-L100

Even though a parameter maybe passed in to control where to open. This leads to subtle bugs and potential inconsistencies:

**Bug: "Open In New Window" can fail**

If a chat is opened in an editor or sidebar, if the user picks "Open In New Window", this will actually not work because we force to reveal the session.

Image

I think this is quite frustrating because the user expressed explicit intent.

**UX: clicking on session in sessions view can sometimes not open in editor**

For now we decided to let sessions opened from the sessions view open in editor. If a chat is opened in sidebar, it would though prefer that over opening in the editor.

roblourens:
Fix #280128

### DIFF
--- a/src/vs/workbench/contrib/chat/browser/chatWidgetService.ts
+++ b/src/vs/workbench/contrib/chat/browser/chatWidgetService.ts
@@ -10,8 +10,8 @@ import { combinedDisposable, Disposable, IDisposable, toDisposable } from '../..
 import { isEqual } from '../../../../base/common/resources.js';
 import { URI } from '../../../../base/common/uri.js';
 import { ILayoutService } from '../../../../platform/layout/browser/layoutService.js';
-import { IEditorService, PreferredGroup } from '../../../../workbench/services/editor/common/editorService.js';
-import { IEditorGroupsService } from '../../../services/editor/common/editorGroupsService.js';
+import { ACTIVE_GROUP, IEditorService, type PreferredGroup } from '../../../../workbench/services/editor/common/editorService.js';
+import { IEditorGroupsService, isEditorGroup } from '../../../services/editor/common/editorGroupsService.js';
 import { IViewsService } from '../../../services/views/common/viewsService.js';
 import { ChatAgentLocation } from '../common/constants.js';
 import { ChatViewId, ChatViewPaneTarget, IChatWidget, IChatWidgetService, IQuickChatService, isIChatViewViewContext } from './chat.js';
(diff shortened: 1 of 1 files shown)


connor4312 (approved):

src/vs/workbench/contrib/chat/browser/chatWidgetService.ts
```diff
@@ -112,12 +115,7 @@ export class ChatWidgetService extends Disposable implements IChatWidgetService
 		}
 
 		// Open in chat editor
-		const pane = await this.editorService.openEditor({
-			resource: sessionResource, options: {
-				...options,
-				revealIfOpened: true // always try to reveal if already opened
-			}
-		}, target);
+		const pane = await this.editorService.openEditor({ resource: sessionResource, options: options }, target);
```
bpasero, line 118: This is maybe up for debate, but I was adding the `revealIfOpened: true` here actually intentionally because I am assuming that a user (such as Kai) will have setup a careful editor layout where chats are all in one group and files are in other groups for review and then it would be frustrating if a chat editor moves around only because another group is active. But I can be convinced to drop this intent and see from real user feedback.

src/vs/workbench/contrib/chat/browser/chatWidgetService.ts
```diff
@@ -164,6 +162,36 @@ export class ChatWidgetService extends Disposable implements IChatWidgetService
...
+				return;
+			}
+
+			if (!isIChatViewViewContext(existingWidget.viewContext) && target !== ChatViewPaneTarget && existingEditor && this.isSameEditorTarget(existingEditor.groupId, target)) {
+				return;
+			}
+
+			if (existingEditor) {
+				// widget.clear() on an editor leaves behind an empty chat editor
+				await this.editorService.closeEditor(existingEditor);
```
bpasero, line 182: Do we need to check if the editor was closed after all? Isn't it possible that this still brings up a dialog asking for accepting or discarding edits from the session? I am not sure how Connors change is timed for this dialog to disappear?

src/vs/workbench/contrib/chat/browser/chatWidgetService.ts
```diff
@@ -93,10 +93,13 @@ export class ChatWidgetService extends Disposable implements IChatWidgetService
 	openSession(sessionResource: URI, target?: typeof ChatViewPaneTarget): Promise<IChatWidget | undefined>;
 	openSession(sessionResource: URI, target?: PreferredGroup, options?: IChatEditorOptions): Promise<IChatWidget | undefined>;
 	async openSession(sessionResource: URI, target?: typeof ChatViewPaneTarget | PreferredGroup, options?: IChatEditorOptions): Promise<IChatWidget | undefined> {
-		// Reveal if already open
-		const alreadyOpenWidget = await this.revealSessionIfAlreadyOpen(sessionResource, options);
-		if (alreadyOpenWidget) {
-			return alreadyOpenWidget;
+		if (options?.revealIfOpened) {
```
bpasero, line 96: `revealIfOpened` is actually an editor option, which is now here used in a non-editor context (sidebar). I would still prefer that we just respect the `target` that is passed in to determine if we need to reveal an existing chat in the sidebar or not. Because otherwise this will not work: pass in an editor group and `revealIfOpened: true`, will sometimes go to the sidebar and sometimes not, which is yet again undeterministic.

bpasero (changes requested):

src/vs/workbench/contrib/chat/browser/chatWidgetService.ts
roblourens: That dialog is removed (not in Insiders but in main)

PR #280236, opened 2025-12-01
Enabling sandboxing for terminal commands execution through copilot chat. 

### LINKED ISSUE #277286: Sandboxing terminal while running commands through agent.
This is part of broader feature to sandbox agent execution. 

This issue is scoped to sandboxing terminal while running commands through copilot chat extension.

OS: Mac OS and Linux

dileepyavan:
fixes #277286


bpasero (dismissed):
I would like to learn more what a `sandboxSettingsResource` is and what the new dependency is `@anthropic-ai/sandbox-runtime`

Are we really pushing this in on the last day of the iteration?

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/runInTerminalTool.ts
```diff
@@ -512,6 +530,11 @@ export class RunInTerminalTool extends Disposable implements IToolImpl {
 			toolSpecificData.commandLine.toolEdited !== toolSpecificData.commandLine.original
 		);
 
+		if (this._isSandboxedTerminal()) {
+			this._logService.info(`RunInTerminalTool: Sandboxing is enabled, wrapping command with srt.`);
+			command = `srt --settings "${this._sandboxConfigPath}" "${command}"`; // sandboxing
```
connor4312: This only works if `srt` happens to already be on the user's path, which is not guaranteed.

src/vs/platform/terminal/common/terminal.ts
```diff
@@ -667,8 +683,28 @@ export interface IShellLaunchConfig {
 	 * This allows extensions to control shell integration for terminals they create.
 	 */
 	shellIntegrationNonce?: string;
+
+	/** Whether to launch the terminal in a sandboxed environment. */
+	sandboxed?: boolean;
+
+	/** Sandbox settings to use when launching the terminal process in a sandboxed environment. */
+	sandboxSettings?: ISandboxRuntimeConfig;
```
connor4312: I'm not clear why these are needed -- it doesn't seem like we actually used them?

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/runInTerminalTool.ts
```diff
@@ -738,30 +765,121 @@ export class RunInTerminalTool extends Disposable implements IToolImpl {
...
+			// Workspace-level setting: use workspace .vscode folder
+			const workSpaceFolder = this._workspaceContextService.getWorkspace().folders[0]?.uri.fsPath;
+			if (workSpaceFolder) {
+				this._sandboxConfigPath = URI.file(workSpaceFolder).with({
+					path: `${workSpaceFolder}/.vscode/sandbox-settings.json`
+				}).fsPath;
+			}
+		} else {
+			// User-level setting: use user profile
+			this._sandboxConfigPath = this._userDataProfileService.currentProfile.sandboxSettingsResource.fsPath;
```
connor4312: This won't work for remotes where the user's profile is on a different machine than the workspace.

I suggest we PR in support for setting the sandbox config via an environment variable for the SRT library (if we want to commit to using that.) There are some other issues with the file handling here, but switching to an environment variable would make them moot.

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/terminal.chatAgentTools.contribution.ts
```diff
@@ -105,8 +117,55 @@ registerWorkbenchContribution2(ChatAgentToolsContribution.ID, ChatAgentToolsCont
 
 // #endregion Contributions
 
+
 // #region Actions
 
+// Retry action for sandboxed terminal commands (invoked via command link in toolResultMessage)
+registerAction2(class RetryWithoutSandboxingAction extends Action2 {
```
connor4312: This just resends the last request and turns sandboxing off in the terminal commands it makes. The terminal command could have been the 20th turn step in a long agentic request, and users would find restarting the whole last request to be frustrating.

We should instead re-run that individual command and proceed from there. I think this may need API changes to support unless we stall after the request fails and before returning to the model (which is also not good, some command failures are normal and expected). A cheap interim solution would be to generate a request saying something like `Rerun the terminal command 'X' and proceed from there`

connor4312 (dismissed):

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/runInTerminalTool.ts
dileepyavan: This is part of a npm package that is added as a dependency

src/vs/platform/terminal/common/terminal.ts
dileepyavan: These slipped from previous changes where the entire process is sandboxed.

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/runInTerminalTool.ts
connor4312: That is not available on the user path when VS Code is installed.

dileepyavan: Correct, this needs to be available. My testing was working as there was a global install on my machine.

.vscode/launch.json
Tyriar, line 304: What's all these launch.json changes about?

Tyriar (changes requested):
Please don't merge this without a proper review from me.

package.json
```diff
@@ -72,6 +72,7 @@
     "update-build-ts-version": "npm install -D typescript@next && npm install -D @typescript/native-preview && (cd build && npm run typecheck)"
   },
   "dependencies": {
+    "@anthropic-ai/sandbox-runtime": "^0.0.13",
```
Tyriar: This is quite old at this point:

https://www.npmjs.com/package/@anthropic-ai/sandbox-runtime?activeTab=versions

Tyriar: Same concerns about shipping beta quality software within in VS Code. This is the main package.json meaning every VS Code user will now get this, pushing binary size out 3+MB even for users that don't engage with AI. While installs have gone up, they're no where near the number that you would expect if Claude Code or some other CLI installed it via npm:

Image

src/vs/platform/terminal/common/terminal.ts
```diff
@@ -19,6 +19,22 @@ import type { IAction } from '../../../base/common/actions.js';
 import type { IDisposable } from '../../../base/common/lifecycle.js';
 import type { SingleOrMany } from '../../../base/common/types.js';
 
+/**
+ * Local type definition for sandbox runtime configuration to avoid importing external package
+ * in the common layer. The actual type should match @anthropic-ai/sandbox-runtime.
+ */
+export interface ISandboxRuntimeConfig {
```
Tyriar: Generally it's best to prefix with `ITerminal` since this is entering the global namespace.

src/vs/platform/terminal/common/terminal.ts
```diff
@@ -670,6 +686,20 @@ export interface IShellLaunchConfig {
 	shellIntegrationNonce?: string;
 }
 
+export interface ISandboxTerminalSettings {
+	enabled?: boolean;
```
Tyriar: `export interface ISandboxTerminalSettings extends ISandboxRuntimeConfig`?

src/vs/platform/terminal/common/terminal.ts
```diff
@@ -710,6 +740,8 @@ export interface IShellLaunchConfigDto {
 	isFeatureTerminal?: boolean;
 	tabActions?: ITerminalTabAction[];
 	shellIntegrationEnvironmentReporting?: boolean;
+	sandboxed?: boolean;
+	sandboxSettings?: ISandboxRuntimeConfig;
```
Tyriar: These should be combined? `sandboxed == !!sandboxSettings`

src/vs/platform/terminal/common/terminalPlatformConfiguration.ts
```diff
@@ -59,6 +59,51 @@ export const terminalProfileBaseProperties: IJSONSchemaMap = {
 	}
 };
 
+export const terminalSandboxPropertySchema: IJSONSchemaMap = {
```
Tyriar: This shouldn't be here, the profile one is here because it's a platform-level feature and needs to be shared across multiple settings. This is not shared and is workbench-level, so it should be inlined inside [terminalChatAgentToolsConfiguration.ts‎](https://github.com/microsoft/vscode/pull/280236/changes#diff-469c2f24bcff8e59e670c291064e46a4a44730fec19614f6360be30181de2487)

src/vs/platform/userDataProfile/common/userDataProfile.ts
```diff
@@ -56,6 +56,7 @@ export interface IUserDataProfile {
 	readonly useDefaultFlags?: UseDefaultProfileFlags;
 	readonly isTransient?: boolean;
 	readonly workspaces?: readonly URI[];
+	readonly sandboxSettingsResource: URI;
```
Tyriar: Why is sandbox stuff touching the user data feature?

src/vs/workbench/contrib/chat/browser/widget/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
```diff
@@ -49,6 +49,8 @@ import { removeAnsiEscapeCodes } from '../../../../../../../base/common/strings.
 import { PANEL_BACKGROUND } from '../../../../../../common/theme.js';
 import { editorBackground } from '../../../../../../../platform/theme/common/colorRegistry.js';
 import { IThemeService } from '../../../../../../../platform/theme/common/themeService.js';
+import { ISandboxUtility } from '../../../../common/sandboxUtility.js';
+
```
Tyriar:
```suggestion
```

src/vs/workbench/contrib/chat/browser/widget/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
```diff
@@ -242,6 +244,7 @@ export class ChatTerminalToolProgressPart extends BaseChatToolInvocationSubPart
 		@IContextKeyService private readonly _contextKeyService: IContextKeyService,
 		@IChatWidgetService private readonly _chatWidgetService: IChatWidgetService,
 		@IKeybindingService private readonly _keybindingService: IKeybindingService,
+		@ISandboxUtility private readonly _sandboxUtility: ISandboxUtility,
```
Tyriar: All services must end up `Service`

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/runInTerminalTool.ts
```diff
@@ -270,6 +272,7 @@ export class RunInTerminalTool extends Disposable implements IToolImpl {
 	private readonly _commandArtifactCollector: TerminalCommandArtifactCollector;
 	protected readonly _profileFetcher: TerminalProfileFetcher;
 
+
```
Tyriar:
```suggestion
```

src/vs/workbench/contrib/chat/common/sandboxUtility.ts
```diff
@@ -0,0 +1,124 @@
...
+export interface ISandboxUtility {
+	readonly _serviceBrand: undefined;
+	isEnabled(): boolean;
+	wrapCommand(command: string): string;
+	getSandboxConfigPath(forceRefresh?: boolean): Promise<string | undefined>;
+	getTempDir(): URI | undefined;
+	setNeedsForceUpdateConfigFile(): void;
+}
+
+export class SandboxUtility implements ISandboxUtility {
```
Tyriar: This should not live in chat/ imo unless we're going to be using it very soon in some non-terminal use case.

src/vs/workbench/contrib/chat/common/sandboxUtility.ts
```diff
@@ -0,0 +1,124 @@
...
+	readonly _serviceBrand: undefined;
+	private static _srtPath: string;
+	private static _sandboxConfigPath: string | undefined;
+	private _needsForceUpdateConfigFile = true;
+	private static _tempDir: URI | undefined;
+	private static _sandboxSettingsId: string | undefined;
+	private static readonly TERMINAL_SANDBOX_SETTING_ID = 'chat.tools.terminal.sandbox';
```
Tyriar: statics should generally be avoided

Tyriar: This setting is duplicating where it's defined? See `terminalContribExports.ts` for how to get setting ids out of `terminalContrib/` land - I'd argue this whole file should live inside `terminalContrib/` though.

src/vs/workbench/contrib/terminalContrib/chatAgentTools/common/terminalSandboxService.ts
```diff
@@ -0,0 +1,124 @@
...
+					deniedDomains: sandboxSetting.network?.deniedDomains || []
+				},
+				filesystem: {
+					denyRead: sandboxSetting.filesystem?.denyRead || [],
+					allowWrite: sandboxSetting.filesystem?.allowWrite || ['.'],
+					denyWrite: sandboxSetting.filesystem?.denyWrite || []
+				}
+			};
+			SandboxUtility._sandboxConfigPath = configFileUri.fsPath;
+			await this._fileService.createFile(configFileUri, VSBuffer.fromString(JSON.stringify(sandboxSettings, null, '\t')), { overwrite: true });
```
Tyriar, line 112: Guessing there's probably a security issue here where another user can disable the sandbox by editing this file.

src/vs/workbench/contrib/terminalContrib/chatAgentTools/common/terminalChatAgentToolsConfiguration.ts
```diff
@@ -460,6 +515,29 @@ export const terminalChatAgentToolsConfiguration: IStringDictionary<IConfigurati
...
 		}
 	},
+	// Experimental settings for object storage that control sandboxing with properties like allowed/denied system calls could go here in future.
+	[TerminalChatAgentToolsSettingId.TerminalSandbox]: {
+		markdownDescription: localize('terminalSandbox.description', "Controls whether to run commands in a sandboxed terminal. Support for sandboxed terminals is enabled for Linux and macOS."),
+		type: ['object', 'null'],
+		'anyOf': [
+			{ type: 'null' },
+			terminalSandboxPropertySchema
+		],
```
Tyriar: anyOf means the settings editor won't work properly. A sandbox config and enablement setting would mean the settings editor would natively support (at least if it was a simple non-nested object).

src/vs/workbench/contrib/terminalContrib/chatAgentTools/common/terminalChatAgentToolsConfiguration.ts
```diff
@@ -460,6 +515,29 @@ export const terminalChatAgentToolsConfiguration: IStringDictionary<IConfigurati
...
+				denyRead: [],
+				denyWrite: [],
+				allowWrite: []
+			}
+		},
+
+		tags: ['experimental'],
+		experiment: {
+			mode: 'auto'
+		}
```
Tyriar: I don't think it needs experiment?

src/vs/workbench/contrib/terminalContrib/chatAgentTools/common/terminalChatAgentToolsConfiguration.ts
```diff
@@ -460,6 +515,29 @@ export const terminalChatAgentToolsConfiguration: IStringDictionary<IConfigurati
...
+	},
+
```
Tyriar:
```suggestion
```

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/runInTerminalTool.ts
Tyriar, line 336:
```suggestion
		this._register(Event.runAndSubscribe(this._configurationService.onDidChangeConfiguration, e => {
			// Clear out warning accepted state if the setting is disabled
			if (!e || e.affectsConfiguration(TerminalChatAgentToolsSettingId.EnableAutoApprove)) {
```

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/runInTerminalTool.ts
```diff
@@ -551,13 +571,15 @@ export class RunInTerminalTool extends Disposable implements IToolImpl {
 
 		const store = new DisposableStore();
 
+
```
Tyriar:
```suggestion
```

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/runInTerminalTool.ts
```diff
@@ -551,13 +571,15 @@ export class RunInTerminalTool extends Disposable implements IToolImpl {
...
 		this._logService.debug(`RunInTerminalTool: Creating ${args.isBackground ? 'background' : 'foreground'} terminal. termId=${termId}, chatSessionId=${chatSessionId}`);
 		const toolTerminal = await (args.isBackground
 			? this._initBackgroundTerminal(chatSessionId, termId, terminalToolSessionId, token)
 			: this._initForegroundTerminal(chatSessionId, termId, terminalToolSessionId, token));
 
 		this._handleTerminalVisibility(toolTerminal, chatSessionId);
 
+
```
Tyriar, line 593:
```suggestion
```

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/runInTerminalTool.ts
```diff
@@ -411,6 +419,13 @@ export class RunInTerminalTool extends Disposable implements IToolImpl {
 			};
 		}
 
+		toolSpecificData.autoApproveInfo = new MarkdownString(localize('autoApprove.sandbox', 'In sandbox mode'));
+		if (this._sandboxUtility.isEnabled()) {
```
Tyriar: Should have a comment before this as it's very important what this is doing (bypassing regular security features).

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/runInTerminalTool.ts
```diff
@@ -538,6 +552,12 @@ export class RunInTerminalTool extends Disposable implements IToolImpl {
 			toolSpecificData.commandLine.toolEdited !== toolSpecificData.commandLine.original
 		);
 
+		if (this._sandboxUtility.isEnabled()) {
+			await this._sandboxUtility.getSandboxConfigPath();
+			this._logService.info(`RunInTerminalTool: Sandboxing is enabled, wrapping command with srt.`);
+			command = this._sandboxUtility.wrapCommand(command);
+		}
```
Tyriar: See `ICommandLineRewriter`, that's the generic mechanism for doing this.

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/runInTerminalTool.ts
```diff
@@ -739,6 +761,103 @@ export class RunInTerminalTool extends Disposable implements IToolImpl {
...
+							case ShellIntegrationQuality.None:
+								retryStrategy = this._instantiationService.createInstance(NoneExecuteStrategy, toolTerminal.instance, () => toolTerminal.receivedUserInput ?? false);
+								break;
+							case ShellIntegrationQuality.Basic:
+								retryStrategy = this._instantiationService.createInstance(BasicExecuteStrategy, toolTerminal.instance, () => toolTerminal.receivedUserInput ?? false, commandDetectionRetry!);
+								break;
+							case ShellIntegrationQuality.Rich:
+								retryStrategy = this._instantiationService.createInstance(RichExecuteStrategy, toolTerminal.instance, commandDetectionRetry!);
+								break;
+						}
```
Tyriar: This is duplicating a whole bunch of stuff which isn't maintainable.

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/runInTerminalTool.ts
```diff
@@ -739,6 +761,103 @@ export class RunInTerminalTool extends Disposable implements IToolImpl {
 					terminalResult = resultArr.join('\n\n');
 				}
 
+				//if sandboxed and there is error, retry without sandboxing.
+				if (exitCode !== 0 && this._sandboxUtility.isEnabled()) {
```
Tyriar: I don't think this is actually what we want? The point of sandboxing is less confirmations, so shouldn't it just keep going?

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/runInTerminalTool.ts
```diff
@@ -793,12 +912,13 @@ export class RunInTerminalTool extends Disposable implements IToolImpl {
...
 				content: [{
 					kind: 'text',
-					value: resultText.join(''),
+					value: resultText.join('')
 				}]
 			};
 		}
 	}
 
+
```
Tyriar:
```suggestion
```

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/runInTerminalTool.ts
```diff
@@ -826,6 +947,8 @@ export class RunInTerminalTool extends Disposable implements IToolImpl {
 		return toolTerminal;
 	}
 
+
+
```
Tyriar:
```suggestion
```

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/runInTerminalTool.ts
```diff
@@ -946,6 +1069,7 @@ export class RunInTerminalTool extends Disposable implements IToolImpl {
 		}
 	}
 
+
```
Tyriar:
```suggestion
```

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/runInTerminalTool.ts
```diff
@@ -967,7 +1091,6 @@ export class RunInTerminalTool extends Disposable implements IToolImpl {
 			}
 		}
 	}
```
Tyriar, line 1020:
```suggestion
	}

```

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/runInTerminalTool.ts
```diff
@@ -814,6 +934,7 @@ export class RunInTerminalTool extends Disposable implements IToolImpl {
 		const profile = await this._profileFetcher.getCopilotProfile();
 		const os = await this._osBackend;
 		const toolTerminal = await this._terminalToolCreator.createTerminal(profile, os, token);
+
```
Tyriar:
```suggestion
```

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/runInTerminalTool.ts
```diff
@@ -793,12 +912,13 @@ export class RunInTerminalTool extends Disposable implements IToolImpl {
 				},
 				content: [{
 					kind: 'text',
-					value: resultText.join(''),
+					value: resultText.join('')
```
Tyriar:
```suggestion
					value: resultText.join(''),
```

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/terminal.chatAgentTools.contribution.ts
```diff
@@ -105,8 +114,11 @@ registerWorkbenchContribution2(ChatAgentToolsContribution.ID, ChatAgentToolsCont
 
 // #endregion Contributions
 
+
 // #region Actions
 
+
+
```
Tyriar:
```suggestion
// #region Actions

```

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/terminal.chatAgentTools.contribution.ts
```diff
@@ -56,6 +64,7 @@ class ChatAgentToolsContribution extends Disposable implements IWorkbenchContrib
 	constructor(
 		@IInstantiationService instantiationService: IInstantiationService,
 		@ILanguageModelToolsService toolsService: ILanguageModelToolsService,
+		@IConfigurationService configurationService: IConfigurationService,
```
Tyriar: Not used?

src/vs/workbench/contrib/terminalContrib/chatAgentTools/common/terminalChatAgentToolsConfiguration.ts
```diff
@@ -473,6 +551,7 @@ export const terminalChatAgentToolsConfiguration: IStringDictionary<IConfigurati
 	}
 };
 
+
```
Tyriar:
```suggestion
```

src/vs/workbench/contrib/terminalContrib/chatAgentTools/test/electron-browser/runInTerminalTool.test.ts
```diff
@@ -1081,7 +1081,6 @@ suite('RunInTerminalTool', () => {
 				instance: mockTerminal2,
 				shellIntegrationQuality: ShellIntegrationQuality.None
 			});
-
 			ok(runInTerminalTool.sessionTerminalAssociations.has(sessionId1), 'Session 1 terminal association should exist');
```
Tyriar:
```suggestion

			ok(runInTerminalTool.sessionTerminalAssociations.has(sessionId1), 'Session 1 terminal association should exist');
```

Tyriar (changes requested):

src/vs/workbench/contrib/chat/common/sandboxUtility.ts
dileepyavan: This would not only be used by terminal but other tools for the chat in the future.

src/vs/workbench/contrib/chat/common/sandboxUtility.ts
dileepyavan: Would like to know more on this. Any reasons on why this should be avoided?

src/vs/workbench/contrib/chat/common/sandboxUtility.ts
Tyriar: Are you sure about that? If not this is just creating tech debt. We can always move it out in the future

src/vs/workbench/contrib/chat/common/sandboxUtility.ts
Tyriar: They're bad for tests, they're properties at runtime int he case of `TERMINAL_SANDBOX_SETTING_ID` as opposed to const enum which gets inlined. There's only one of these services anyway so having `static` here is just confusing and misleading to the developer reading it as it implies to me that there are multiple instances of this class.

package.json
dileepyavan: updated

src/vs/platform/terminal/common/terminal.ts
dileepyavan: I put it this to match with external, this might needs to be moved to a different place. Can address in coming PR's

src/vs/platform/terminal/common/terminal.ts
dileepyavan: updated

src/vs/platform/terminal/common/terminal.ts
dileepyavan: updated, this is no longer needed.

src/vs/platform/userDataProfile/common/userDataProfile.ts
dileepyavan: This is one of the earlier implementations I did for settings which I reverted.

src/vs/workbench/contrib/chat/browser/widget/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
dileepyavan:
```suggestion

```

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/runInTerminalTool.ts
dileepyavan: updated

src/vs/workbench/contrib/chat/common/sandboxUtility.ts
dileepyavan: As part of sandboxing, we need to implement the same for other tools/ processes. After terminal sandboxing, we need to work on other tools/process.

src/vs/workbench/contrib/chat/common/sandboxUtility.ts
dileepyavan: All the properties are expected to have a single global value for a vscode instance. Will update non-constant values to be instance based instead of static to reduce confusion.

dileepyavan: Sandboxing can be used by other processes other than terminal.

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/runInTerminalTool.ts
dileepyavan:
```suggestion
	private static readonly _backgroundExecutions = new Map<string, BackgroundTerminalExecution>();
```

.vscode/launch.json
dileepyavan: These changes are suggested connor for vscode remote debugging

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/terminal.chatAgentTools.contribution.ts
dileepyavan: Updated.

package.json
Tyriar: I see https://github.com/anthropic-experimental/sandbox-runtime doesn't have any native code actually, so I'm not so worried about this anymore.

src/vs/workbench/contrib/chat/common/sandboxUtility.ts
Tyriar: I wouldn't be so sure, running shell commands are inherently risky unlike pretty much any other tool. Fetch for example we have full control over as we know exactly what the URL is.

dileepyavan: Would like to discuss and take this up in follow-up PR's.

src/vs/workbench/contrib/chat/common/sandboxUtility.ts
dileepyavan: used terminalContribExports instead.

package.json
dileepyavan: updated to [0.0.23](https://www.npmjs.com/package/@anthropic-ai/sandbox-runtime/v/0.0.23) as there are issues reported in higher versions.

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/runInTerminalTool.ts
dileepyavan: Yeah I think I started with this but this will be an issue when retrying as there will be multiple tool edited commands and not easily mainatainable for retry.

src/vs/workbench/contrib/terminalContrib/chatAgentTools/common/terminalSandboxService.ts
dileepyavan: Temp file format is "vscode-sandbox-settings-6ce94558-99ac-4e5b-be0a-1a6e003fc806.json" and recreated on every instance of the vscode. The file is always overwritten when there is a change in the sandbox settings. This will be an issue if the user on the server access tmp folder and has permissions to modify, remove all the content of denied paths. We also show user if the command ran in sandboxed mode or not.  There are multiple recovery points here for the user. Similar to if an auth token is compromised unknowingly, it has x hrs of default expiration time.

package.json
```diff
@@ -73,6 +73,7 @@
     "update-build-ts-version": "npm install -D typescript@next && npm install -D @typescript/native-preview && (cd build && npm run typecheck)"
   },
   "dependencies": {
+    "@anthropic-ai/sandbox-runtime": "^0.0.23",
```
Tyriar: This should not have the `^`, v0 of a package means breaking changes can happen at any version, so running upgrading via npm may break the feature silently or in compile.

src/vs/workbench/contrib/chat/common/sandboxService.ts
```diff
@@ -0,0 +1,135 @@
...
+import { IFileService } from '../../../../platform/files/common/files.js';
+import { VSBuffer } from '../../../../base/common/buffer.js';
+import { joinPath } from '../../../../base/common/resources.js';
+import { generateUuid } from '../../../../base/common/uuid.js';
+import { IEnvironmentService } from '../../../../platform/environment/common/environment.js';
+import { TerminalContribSettingId } from '../../terminal/terminalContribExports.js';
+
+export const ISandboxService = createDecorator<ISandboxService>('sandboxService');
+
+export interface ISandboxService {
```
Tyriar: This should either move into terminal and be named `ITerminalSandboxService` or stay here and be named `IChatSandboxService`

src/vs/platform/terminal/common/terminal.ts
Tyriar: This should either move into chat and be `IChatSandboxRuntimeConfig` or here and be prefixed with `ITerminal`

src/vs/workbench/contrib/chat/common/sandboxService.ts
```diff
@@ -0,0 +1,135 @@
...
+			this._initTempDir();
+		}
+		if (this._tempDir) {
+			const networkSetting = this._configurationService.getValue<ITerminalSandboxSettings['network']>(TerminalContribSettingId.TerminalSandboxNetwork) ?? {};
+			const linuxFileSystemSetting = isLinux
+				? this._configurationService.getValue<ITerminalSandboxSettings['filesystem']>(TerminalContribSettingId.TerminalSandboxLinuxFileSystem) ?? {}
+				: {};
+			const macFileSystemSetting = isMacintosh
+				? this._configurationService.getValue<ITerminalSandboxSettings['filesystem']>(TerminalContribSettingId.TerminalSandboxMacFileSystem) ?? {}
+				: {};
```
Tyriar: This will fall over on remotes, you need to support this based on the OS where the pty is hosted, in the case of the terminal (`ITerminalInstance.os`)

src/vs/workbench/contrib/chat/browser/widget/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
```diff
@@ -455,7 +457,7 @@ export class ChatTerminalToolProgressPart extends BaseChatToolInvocationSubPart
 			showOutputAction = this._instantiationService.createInstance(ToggleChatTerminalOutputAction, () => this._toggleOutputFromAction());
 			this._showOutputAction.value = showOutputAction;
 			const exitCode = resolvedCommand?.exitCode ?? this._terminalData.terminalCommandState?.exitCode;
-			if (exitCode) {
```
Tyriar, line 458: What's this for?

src/vs/workbench/contrib/chat/common/chatSandboxService.ts
```diff
@@ -0,0 +1,135 @@
...
+		this._srtPath = join(appRoot, 'node_modules', '.bin', 'srt');
+		this._sandboxSettingsId = generateUuid();
+		this._initTempDir();
+	}
+
+	public isEnabled(): boolean {
+		if (isWindows) {
+			return false;
+		}
+		const enabledSetting = this._configurationService.getValue<boolean>(TerminalContribSettingId.TerminalSandboxEnabled);
```
Tyriar: If this is going to live in `chat/`, the setting should not be prefixed with `chat.tools.terminal.`. This is breaking architectural layers that aim to prevent spaghetti code/cyclic dependencies (terminalContrib depends on terminal depends on chat depends on terminalContrib).

I still think this should live inside terminal as it's only used there and we're sandboxing shell commands here. If we do happen to need it generally we can always move it and migrate at a later point.

src/vs/workbench/contrib/chat/common/chatSandboxService.ts
```diff
@@ -0,0 +1,135 @@
...
+		const isEnabled = enabledSetting === true;
+		if (!isEnabled) {
+			return false;
+		}
+		return true;
```
Tyriar:
```suggestion
		}
		return this._configurationService.getValue<boolean>(TerminalContribSettingId.TerminalSandboxEnabled);
```

src/vs/workbench/contrib/terminalContrib/chatAgentTools/common/terminalSandboxService.ts
```diff
@@ -0,0 +1,135 @@
...
+			return false;
+		}
+		return true;
+	}
+
+	public wrapCommand(command: string): string {
+		if (!this._sandboxConfigPath || !this._tempDir) {
+			throw new Error('Sandbox config path or temp dir not initialized');
+		}
+		return `"${this._srtPath}" TMPDIR=${this._tempDir.fsPath} --settings "${this._sandboxConfigPath}" "${command}"`;
```
Tyriar, line 67: It may be possible that `_sandboxConfigPath` can contain `"` and escape the argument. We should be defensive here as it's enforcing a security feature.

src/vs/workbench/contrib/terminalContrib/chatAgentTools/common/terminalSandboxService.ts
```diff
@@ -0,0 +1,135 @@
...
+	}
```
Tyriar, line 68: Command is likely to double quotes, this should also be sanitized.

src/vs/workbench/contrib/chat/common/chatSandboxService.ts
```diff
@@ -0,0 +1,135 @@
...
+			if (!tmpDir) {
+				this._logService.warn('SandboxService: Cannot create sandbox settings file because no tmpDir is available in this environment');
+				return;
+			}
+			this._tempDir = tmpDir;
+			this._fileService.exists(this._tempDir).then(exists => {
+				if (!exists) {
+					this._logService.warn(`SandboxService: tmp directory is not present at ${this._tempDir}`);
+				}
+			});
```
Tyriar: Switch to using await here so we don't have a dangling promise and we verify the temp dir does actually exist after the function is finished.

src/vs/workbench/contrib/chat/common/chatSandboxService.ts
```diff
@@ -0,0 +1,135 @@
...
+			await this._fileService.createFile(configFileUri, VSBuffer.fromString(JSON.stringify(sandboxSettings, null, '\t')), { overwrite: true });
+			return this._sandboxConfigPath;
+		}
+		return undefined;
+	}
+
+	private _initTempDir(): void {
+		if (this.isEnabled() && isNative) {
+			this._needsForceUpdateConfigFile = true;
+			const tmpDir = (this._environmentService as IEnvironmentService & { tmpDir: URI }).tmpDir;
```
Tyriar: Unsafe cast?

src/vs/workbench/contrib/terminalContrib/chatAgentTools/common/terminalSandboxService.ts
```diff
@@ -0,0 +1,135 @@
...
+				}
+			};
+			this._sandboxConfigPath = configFileUri.fsPath;
+			await this._fileService.createFile(configFileUri, VSBuffer.fromString(JSON.stringify(sandboxSettings, null, '\t')), { overwrite: true });
+			return this._sandboxConfigPath;
+		}
+		return undefined;
+	}
+
+	private _initTempDir(): void {
```
Tyriar, line 118: nit: You could return `URI | undefined` here to simplify its usage:

```ts
if (!this.isEnabled()) {
  return undefined
}

const tmpDir = this._initTempDir();
if (tmpDir === undefined) {
  return undefined;
}

// ...
```

src/vs/workbench/contrib/terminalContrib/chatAgentTools/common/terminalSandboxService.ts
```diff
@@ -0,0 +1,135 @@
...
+		if (this.isEnabled() && isNative) {
```
Tyriar, line 119: `isNative` shouldn't determine whether sandboxing is enabled, in fact, wouldn't this mean that the temp dir is never created since isNative is only true in a node environment, not the browser renderer thread?

src/vs/workbench/contrib/editSessions/test/browser/editSessions.test.ts
```diff
@@ -166,6 +166,7 @@ suite('Edit session sync', () => {
 				promptsHome: URI.file('promptsHome'),
 				extensionsResource: URI.file('extensionsResource'),
 				cacheHome: URI.file('cacheHome'),
+				sandboxSettingsResource: URI.file('sandbox-settings.json')
```
Tyriar: This feature should not touch edit sessions

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/terminal.chatAgentTools.contribution.ts
```diff
@@ -121,6 +129,7 @@ registerWorkbenchContribution2(ChatAgentToolsContribution.ID, ChatAgentToolsCont
 
 // #endregion Contributions
 
+
```
Tyriar:
```suggestion
```

src/vs/workbench/contrib/terminalContrib/chatAgentTools/common/terminalChatAgentToolsConfiguration.ts
```diff
@@ -503,6 +507,95 @@ export const terminalChatAgentToolsConfiguration: IStringDictionary<IConfigurati
 			mode: 'auto'
 		}
 	},
+	[TerminalChatAgentToolsSettingId.TerminalSandboxEnabled]: {
```
Tyriar, line 510: All these settings should be `restricted: true`

src/vs/workbench/contrib/terminalContrib/chatAgentTools/common/terminalChatAgentToolsConfiguration.ts
```diff
@@ -503,6 +507,95 @@ export const terminalChatAgentToolsConfiguration: IStringDictionary<IConfigurati
...
+			denyWrite: {
+				type: 'array',
+				description: localize('terminalSandbox.linuxFileSystemSetting.denyWrite', "Array of paths to deny write access within allowed paths (takes precedence over allowWrite)."),
+				items: { type: 'string' },
+				default: []
+			}
+		},
+		default: {
+			denyRead: [],
+			allowWrite: ['.'],
```
Tyriar, line 566: It's unclear if `.` mean only the workspace or everything? Only workspace seems like a good default

src/vs/workbench/contrib/terminalContrib/chatAgentTools/common/terminalChatAgentToolsConfiguration.ts
```diff
@@ -503,6 +507,95 @@ export const terminalChatAgentToolsConfiguration: IStringDictionary<IConfigurati
...
+			denyWrite: []
```
Tyriar, line 567: Not workspace seems like a good default for read/write?

src/vs/workbench/contrib/terminalContrib/chatAgentTools/common/terminalChatAgentToolsConfiguration.ts
```diff
@@ -503,6 +507,95 @@ export const terminalChatAgentToolsConfiguration: IStringDictionary<IConfigurati
...
+		default: false,
+		tags: ['experimental']
+	},
+	[TerminalChatAgentToolsSettingId.TerminalSandboxNetwork]: {
+		markdownDescription: localize('terminalSandbox.networkSetting', "Note: this setting is applicable only when {0} is enabled. Controls network access in the terminal sandbox.", `\`#${TerminalChatAgentToolsSettingId.TerminalSandboxEnabled}#\``),
+		type: 'object',
+		properties: {
+			allowedDomains: {
+				type: 'array',
+				description: localize('terminalSandbox.networkSetting.allowedDomains', "Array of allowed domains (supports wildcards like *.example.com). Empty array = no network access."),
```
Tyriar: Tweaking wording to reduce parens, not localize the example, use backticks for formatting, make markdown and avoid array as it's a technical term. The rest should be more in this style too:

```suggestion
				markdownDescription: localize('terminalSandbox.networkSetting.allowedDomains', "A list of allowed domains. Supports wildcards like {0} and an empty list means no network access.", '`*.example.com`'),
```

src/vs/workbench/contrib/terminalContrib/chatAgentTools/common/terminalChatAgentToolsConfiguration.ts
```diff
@@ -503,6 +507,95 @@ export const terminalChatAgentToolsConfiguration: IStringDictionary<IConfigurati
...
 		}
 	},
+	[TerminalChatAgentToolsSettingId.TerminalSandboxEnabled]: {
+		markdownDescription: localize('terminalSandbox.enabledSetting', "Controls whether to run commands in a sandboxed terminal for the run in terminal tool."),
+		type: 'boolean',
+		default: false,
+		tags: ['experimental']
+	},
+	[TerminalChatAgentToolsSettingId.TerminalSandboxNetwork]: {
+		markdownDescription: localize('terminalSandbox.networkSetting', "Note: this setting is applicable only when {0} is enabled. Controls network access in the terminal sandbox.", `\`#${TerminalChatAgentToolsSettingId.TerminalSandboxEnabled}#\``),
```
Tyriar, line 518: These settings should have a real description that says what it's doing. This is the main string users that go to configure it will see.

src/vs/workbench/contrib/terminalContrib/chatAgentTools/common/terminalSandboxService.ts
```diff
@@ -0,0 +1,135 @@
...
+		}
+		const enabledSetting = this._configurationService.getValue<boolean>(TerminalContribSettingId.TerminalSandboxEnabled);
+		const isEnabled = enabledSetting === true;
+		if (!isEnabled) {
+			return false;
+		}
+		return true;
+	}
+
+	public wrapCommand(command: string): string {
```
Tyriar, line 63: We should have some unit tests that test this function specifically, the quote escaping cases I mention below are good candidates to cover in the tests.

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/runInTerminalTool.ts
```diff
@@ -276,17 +278,14 @@ export class RunInTerminalTool extends Disposable implements IToolImpl {
 	private readonly _treeSitterCommandParser: TreeSitterCommandParser;
 	private readonly _telemetry: RunInTerminalToolTelemetry;
 	private readonly _commandArtifactCollector: TerminalCommandArtifactCollector;
-	protected readonly _profileFetcher: TerminalProfileFetcher;
-
 	private readonly _commandLineRewriters: ICommandLineRewriter[];
 	private readonly _commandLineAnalyzers: ICommandLineAnalyzer[];
-
+	private static readonly _backgroundExecutions = new Map<string, BackgroundTerminalExecution>();
+	protected readonly _profileFetcher: TerminalProfileFetcher;
```
Tyriar: Please don't rearrange things unnecessarily, can you move these back and keep the whitespace as it was?

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/runInTerminalTool.ts
```diff
@@ -417,6 +427,15 @@ export class RunInTerminalTool extends Disposable implements IToolImpl {
 			};
 		}
 
+		toolSpecificData.autoApproveInfo = new MarkdownString(localize('autoApprove.sandbox', 'In sandbox mode'));
+		// If in sandbox mode, skip confirmation logic. In sandbox mode, commands are run in a restricted environment and explicit
+		// user confirmation is not required.
```
Tyriar:
```suggestion
		// If in sandbox mode, skip confirmation logic. In sandbox mode, commands are run in a
		// restricted environment and explicit user confirmation is not required.
		toolSpecificData.autoApproveInfo = new MarkdownString(localize('autoApprove.sandbox', 'In sandbox mode'));
```

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/runInTerminalTool.ts
```diff
@@ -527,14 +546,13 @@ export class RunInTerminalTool extends Disposable implements IToolImpl {
 				}]
 			};
 		}
-
 		const args = invocation.parameters as IRunInTerminalInputParams;
```
Tyriar, line 551: Avoid messing with whitespace

src/vs/workbench/contrib/terminalContrib/chatAgentTools/common/terminalSandboxService.ts
Tyriar: There's a class of vulnerability that is about other users editing files that can influence security related features. I suspect on all platforms this is an issue currently where this file is created (not restricted to the user), some one logged into the same machine could see it and edit it, and it would be used after that with the wrong content.

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/runInTerminalTool.ts
Tyriar: That's because the retry mechanism is fragile and doesn't play nicely with how the terminal tool works. The command rewriter system is to avoid making this already complex file a mess.

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/runInTerminalTool.ts
```diff
@@ -763,6 +776,91 @@ export class RunInTerminalTool extends Disposable implements IToolImpl {
 					terminalResult = resultArr.join('\n\n');
 				}
 
+				//if sandboxed and there is error, retry without sandboxing.
+				if (exitCode !== 0 && this._sandboxService.isEnabled()) {
```
Tyriar: Please remove retry from this PR, it's not needed for the MVP. This feature is causing conflicts with the rewrite system, forks execute code which is currently being refactored and it's also optional and not necessarily what the users want. The PR is still not ready to be merged and this will push that out further, it's an easy feature to trim off to get this merged faster.

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/runInTerminalTool.ts
```diff
@@ -860,6 +960,8 @@ export class RunInTerminalTool extends Disposable implements IToolImpl {
 		return toolTerminal;
 	}
 
+
+
```
Tyriar: Please review and restore all whitespace as it was before. It was consistent before and the changes are just making the code messy.

Tyriar (changes requested):

package.json
dileepyavan: got it, normally assumed npm ci is used to deploy pinned versions in package-lock.json.

src/vs/workbench/contrib/chat/common/sandboxService.ts
```diff
@@ -0,0 +1,135 @@
...
+	private _needsForceUpdateConfigFile = true;
+	private _tempDir: URI | undefined;
+	private _sandboxSettingsId: string | undefined;
+
+
+	constructor(
+		@IConfigurationService private readonly _configurationService: IConfigurationService,
+		@IFileService private readonly _fileService: IFileService,
+		@IEnvironmentService private readonly _environmentService: IEnvironmentService,
+		@ITerminalLogService private readonly _logService: ITerminalLogService,
```
Tyriar: This should be `ILogService` if it's not in terminal

Tyriar (changes requested):

src/vs/workbench/contrib/terminalContrib/chatAgentTools/common/terminalSandboxService.ts
dileepyavan: Discussed and validated, temp files created -rw-r--r--   permissions

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/runInTerminalTool.ts
dileepyavan: removed retry

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/runInTerminalTool.ts
dileepyavan: These were part of merge conflicts

src/vs/workbench/contrib/terminalContrib/chatAgentTools/common/terminalChatAgentToolsConfiguration.ts
dileepyavan: Its the workspace

src/vs/workbench/contrib/terminalContrib/chatAgentTools/common/terminalChatAgentToolsConfiguration.ts
dileepyavan: I think default user should be available to edit and read the files as its loaded in IDE. Or else everyone should edit the settings file for basic operations

src/vs/workbench/contrib/editSessions/test/browser/editSessions.test.ts
dileepyavan: Agree, this leaked from my previous implementation which is updating sessions for sandbox config.

src/vs/workbench/contrib/terminalContrib/chatAgentTools/common/terminalSandboxService.ts
dileepyavan: Correct, renderer process should not create files in temp folders on user machines. Here, runinterminalTool is calling. This is tested. Native def in platform.ts .

if (typeof nodeProcess === 'object') {
	_isWindows = (nodeProcess.platform === 'win32');
	_isMacintosh = (nodeProcess.platform === 'darwin');
	_isLinux = (nodeProcess.platform === 'linux');
	_isLinuxSnap = _isLinux && !!nodeProcess.env['SNAP'] && !!nodeProcess.env['SNAP_REVISION'];
	_isElectron = isElectronProcess;
	_isCI = !!nodeProcess.env['CI'] || !!nodeProcess.env['BUILD_ARTIFACTSTAGINGDIRECTORY'] || !!nodeProcess.env['GITHUB_WORKSPACE'];
	_locale = LANGUAGE_DEFAULT;
	_language = LANGUAGE_DEFAULT;
	const rawNlsConfig = nodeProcess.env['VSCODE_NLS_CONFIG'];
	if (rawNlsConfig) {
		try {
			const nlsConfig: nls.INLSConfiguration = JSON.parse(rawNlsConfig);
			_locale = nlsConfig.userLocale;
			_platformLocale = nlsConfig.osLocale;
			_language = nlsConfig.resolvedLanguage || LANGUAGE_DEFAULT;
			_translationsConfigFile = nlsConfig.languagePack?.translationsConfigFile;
		} catch (e) {
		}
	}
	_isNative = true;
}

src/vs/workbench/contrib/chat/common/chatSandboxService.ts
dileepyavan: removed this as the tempDir is fed from environment service and there is already a null check

src/vs/workbench/contrib/terminalContrib/chatAgentTools/common/terminalSandboxService.ts
dileepyavan: anthropic sandbox runtime uses shell quotes which takes care of quotes.

src/vs/workbench/contrib/terminalContrib/chatAgentTools/common/terminalSandboxService.ts
dileepyavan: _sandboxConfigPath is code generated, not user entered.

src/vs/workbench/contrib/terminalContrib/chatAgentTools/common/terminalSandboxService.ts
dileepyavan: I have tests as an action item, planning to add them in a follow up PR.

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/runInTerminalTool.ts
```diff
@@ -545,6 +568,12 @@ export class RunInTerminalTool extends Disposable implements IToolImpl {
 			toolSpecificData.commandLine.toolEdited !== toolSpecificData.commandLine.original
 		);
 
+		if (this._sandboxService.isEnabled()) {
+			await this._sandboxService.getSandboxConfigPath();
+			this._logService.info(`RunInTerminalTool: Sandboxing is enabled, wrapping command with srt.`);
+			command = this._sandboxService.wrapCommand(command);
+		}
```
Tyriar, line 572: This should be converted into a command rewriter

src/vs/workbench/contrib/terminalContrib/chatAgentTools/common/terminalSandboxService.ts
```diff
@@ -0,0 +1,130 @@
...
+export interface ITerminalSandboxService {
+	readonly _serviceBrand: undefined;
+	isEnabled(): boolean;
+	wrapCommand(command: string): string;
+	getSandboxConfigPath(forceRefresh?: boolean): Promise<string | undefined>;
+	getTempDir(): URI | undefined;
+	setNeedsForceUpdateConfigFile(): void;
+}
+
+export class TerminalSandboxService implements ITerminalSandboxService {
```
Tyriar, line 33: This should move to `terminalContrib/chatAgentTools`, then you don't need the `terminalContribExports.ts` hack

src/vs/platform/terminal/common/terminal.ts
```diff
@@ -19,6 +19,22 @@ import type { IAction } from '../../../base/common/actions.js';
...
+	network?: {
+		allowedDomains?: string[];
+		deniedDomains?: string[];
+	};
+	filesystem?: {
+		denyRead?: string[];
+		allowWrite?: string[];
+		denyWrite?: string[];
+	};
+}
```
Tyriar: This can all move into terminalContrib too I think?

Tyriar (changes requested):

src/vs/workbench/contrib/terminalContrib/chatAgentTools/common/terminalSandboxService.ts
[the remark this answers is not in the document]
dileepyavan: anthropic sandbox runtime uses shell quote before this command is executed.

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/runInTerminalTool.ts
```diff
@@ -707,33 +772,141 @@ export class RunInTerminalTool extends Disposable implements IToolImpl {
...
+	private _createConfigFileForSandboxingAtWorkspaceLevel(settings: ISandboxTerminalSettings): void {
+		const workSpaceFolder = this._workspaceContextService.getWorkspace().folders[0]?.uri.fsPath;
+		if (!workSpaceFolder) {
+			return;
+		}
+		const sandboxSettingsUri = URI.file(workSpaceFolder).with({
+			path: `${workSpaceFolder}/.vscode/sandbox-settings.json`
+		});
+
+		this._createSandboxConfig(sandboxSettingsUri, settings);
```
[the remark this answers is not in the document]
dileepyavan: This is being called on the load before any action is taking place. This is intentional and using fire and forget approach. We do have a check again while the commands are executing.

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/runInTerminalTool.ts
dileepyavan: Will be doing it up in a follow up pr as I need to revisit the testing. Now as retry is removed, I should be able to implement this. Will send a follow up PR today after this merge.

src/vs/workbench/contrib/terminalContrib/chatAgentTools/common/terminalSandboxService.ts
```diff
@@ -0,0 +1,129 @@
...
+	readonly _serviceBrand: undefined;
+	private _srtPath: string;
+	private _sandboxConfigPath: string | undefined;
+	private _needsForceUpdateConfigFile = true;
+	private _tempDir: URI | undefined;
+	private _sandboxSettingsId: string | undefined;
+	private _os: OperatingSystem = OS;
```
Tyriar, line 40: You might want to make this a promise to avoid a possible race? You'd need isEnabled to be async then

Tyriar (approved):

src/vs/workbench/contrib/terminalContrib/chatAgentTools/common/terminalSandboxService.ts
dileepyavan: This is kind of fire and forget approach that is set in the constructor during singleton creation. Want to keep  the usage of isEnabled simple without async.

PR #280249, opened 2025-12-01
chat - further tweaks to `openSession`

bpasero:
**This does:**
- `revealSessionIfAlreadyOpen` if no `target` is passed in
- `revealIfOpened: true` on editors unless `options.revealIfOpened` is specified
- sets `preserveFocus: true` when closing the chat editor to not mess with focus


src/vs/workbench/contrib/chat/browser/chatWidgetService.ts
```diff
@@ -115,7 +116,13 @@ export class ChatWidgetService extends Disposable implements IChatWidgetService
 		}
 
 		// Open in chat editor
-		const pane = await this.editorService.openEditor({ resource: sessionResource, options: options }, target);
+		const pane = await this.editorService.openEditor({
+			resource: sessionResource,
+			options: {
+				...options,
+				revealIfOpened: options?.revealIfOpened ?? true // always try to reveal if already opened unless explicitly told not to
```
roblourens, line 123: I like this much better, I just thought you wanted the opposite, thanks

roblourens (approved):

PR #280263, opened 2025-12-01
Fix breakpoint range calculation

DrSergei:
This patch fixes wrong `bytes` calculation for `dataBreakpointInfo` request. Also, added a check that `end` address is greater than `start` address.


src/vs/workbench/contrib/debug/browser/breakpointsView.ts
```diff
@@ -1613,9 +1613,12 @@ abstract class MemoryBreakpointAction extends Action2 {
 
 		const start = BigInt(startStr);
 		const end = BigInt(endStr);
+		if (start > end) {
```
connor4312: This is allowed for the `+` operator

connor4312 (changes requested):

connor4312 (approved):
thanks!

DonJayamanne (approved):

PR #280273, opened 2025-12-01
fix(process-explorer): find name regexp error

yaxiaoliu:
before: 
find all xxxx.js will return xxxx.json and other startsWith xxxx.js
image

after:
find all xxxx.js only return `xxx.js`
image


bpasero (approved):
Thanks

dbaeumer (approved):

PR #280300, opened 2025-12-01
Add workaround for cases when execComand('copy') doesn't work with EditContext

alexdima:
Add workaround for #252371

### DIFF
--- a/src/vs/editor/browser/controller/editContext/clipboardUtils.ts
+++ b/src/vs/editor/browser/controller/editContext/clipboardUtils.ts
@@ -6,8 +6,57 @@ import { IViewModel } from '../../../common/viewModel.js';
 import { Range } from '../../../common/core/range.js';
 import { isWindows } from '../../../../base/common/platform.js';
 import { Mimes } from '../../../../base/common/mime.js';
+import { ViewContext } from '../../../common/viewModel/viewContext.js';
+import { ILogService, LogLevel } from '../../../../platform/log/common/log.js';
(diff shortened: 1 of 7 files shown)


dbaeumer (approved):

src/vs/editor/browser/controller/editContext/clipboardUtils.ts
```diff
@@ -6,8 +6,57 @@ import { IViewModel } from '../../../common/viewModel.js';
...
+import { EditorOption, IComputedEditorOptions } from '../../../common/config/editorOptions.js';
+import { generateUuid } from '../../../../base/common/uuid.js';
 
-export function getDataToCopy(viewModel: IViewModel, modelSelections: Range[], emptySelectionClipboard: boolean, copyWithSyntaxHighlighting: boolean): ClipboardDataToCopy {
+export function ensureClipboardGetsEditorSelection(e: ClipboardEvent, context: ViewContext, logService: ILogService, isFirefox: boolean): void {
+	const viewModel = context.viewModel;
+	const options = context.configuration.options;
+	let id: string | undefined = undefined;
+	if (logService.getLevel() === LogLevel.Trace) {
+		id = generateUuid();
```
deepak1556, line 19: @alexdima this is great, what do you think about adding a [performance marker](https://developer.mozilla.org/en-US/docs/Web/API/Performance/mark) with the uuid when the action fails so that when the user captures a cpu profile it would be easier to narrow in on the timeline.

PR #280362, opened 2025-12-01
Fix: Parallelize extension querying and target local server for insta…

Himanshu197200:
…ll action (#280336)

**Title:** Fix: Parallelize extension querying and target local server for install action

Investigated the issue where installing extensions on WSL2 would just spin forever. Turns out `ExtensionsWorkbenchService.queryLocal` was awaiting extension servers sequentially. If the remote server (WSL) didn't respond fast enough, it blocked the whole chain.

I've updated [queryLocal](cci:1://file:///Users/himanshumishramac198072gmail.com/.gemini/antigravity/scratch/vscode_fix/src/vs/workbench/contrib/extensions/common/extensions.ts:137:1-137:71) to fire off the queries in parallel so one slow server doesn't hold up the rest.

Also, for the "Install Local Extensions in Remote" action specifically, it was querying *all* servers for some reason. I switched it to only query the local server since that's all it actually needs. This should stop it from hanging if the remote side is acting up.

Tested locally and verified the queries are happening concurrently now.

### DIFF
--- a/src/vs/workbench/contrib/extensions/browser/extensionsActions.ts
+++ b/src/vs/workbench/contrib/extensions/browser/extensionsActions.ts
@@ -2960,6 +2960,7 @@ export abstract class AbstractInstallExtensionsInServerAction extends Action {
 
 	constructor(
 		id: string,
+		private readonly serverToQuery: IExtensionManagementServer | undefined,
 		@IExtensionsWorkbenchService protected readonly extensionsWorkbenchService: IExtensionsWorkbenchService,
 		@IQuickInputService private readonly quickInputService: IQuickInputService,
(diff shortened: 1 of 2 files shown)


sandy081:
> Tested locally and verified the queries are happening concurrently now.

This is good, but is the issue that https://github.com/microsoft/vscode/issues/280336 is seeing? If fetching remote extension hangs then it will block even if you do parallelly right?

sandy081:
> The root cause of https://github.com/microsoft/vscode/issues/280336 is queryLocal() awaiting servers sequentially. If WSL/remote hung, it blocked everything → frozen UI.

Have you tried this? Because UI asks extensions per server which means there will be only single request

PR #280426, opened 2025-12-01
Add proposed API for prompt files providers

pwang347:
## Overview
The goal of this change is to unify the internal prompt file provider APIs and to prepare the provider changes as a separate proposed API to allow for dynamic prompt file content registration.

The current PR provides minimal changes to include support for instructions as well as prompt files (.prompt.md). Internally, we refer to all of these categories as "prompt files".

## API follow-ups
- [ ] https://github.com/microsoft/vscode/pull/286457 - When I was testing the extension sample, it felt extremely wonky to have only a URI property and no way to just pass content directly. This update addresses it.
- [ ] https://github.com/microsoft/vscode/pull/286453 ([v2](https://github.com/microsoft/vscode/pull/286460)) - this is an internal addition to allow for custom source attribution, e.g. showing Organization label for resources related to GH Organizations.
- [ ] https://github.com/microsoft/vscode/pull/286445 ([v2](https://github.com/microsoft/vscode/pull/286466)) - I propose we add this to the API to allow for prompt files previously discovered to be available immediately (this needs to be optional in case stale data is not acceptable).

## API usage
- [ ] https://github.com/microsoft/vscode-extension-samples/pull/1259 (using V2) - samples update, including previous static files approach
- [ ] https://github.com/microsoft/vscode-copilot-chat/pull/2737 (TODO, use V2) - base PR adapting current GH org/enterprise agents to these changes
- [ ] https://github.com/microsoft/vscode-copilot-chat/pull/2310 (TODO, use V2) - adding GH custom instructions


src/vs/workbench/contrib/chat/common/promptSyntax/service/promptsServiceImpl.ts
```diff
@@ -254,6 +297,52 @@ export class PromptsService extends Disposable implements IPromptsService {
 		return result;
 	}
 
+	private async listInstructionsFromProvider(token: CancellationToken): Promise<IPromptPath[]> {
+		const result: IPromptPath[] = [];
+
+		if (this.instructionsProviders.length === 0) {
```
aeschli: This needs to be below the activateByEvent await

src/vs/workbench/contrib/chat/common/promptSyntax/computeAutomaticInstructions.ts
```diff
@@ -88,6 +88,9 @@ export class ComputeAutomaticInstructions {
 		const telemetryEvent: InstructionsCollectionEvent = newInstructionsCollectionEvent();
 		const context = this._getContext(variables);
 
+		// add instructions from providers (always included)
+		await this.addProviderInstructions(instructionFiles, context, variables, telemetryEvent, token);
```
aeschli: There should be no special handling in computeAutomaticInstructions. The provider instructions are not any special. If we want to to be always included they should define `applyTo: **`

aeschli (changes requested):
I think we should unify the code in core
rename `registerCustomAgentsProvider` to registerContributionsProvider(type

src/vs/workbench/api/common/extHostChatAgents2.ts
```diff
@@ -511,13 +526,22 @@ export class ExtHostChatAgents2 extends Disposable implements ExtHostChatAgentsS
 		return await provider.provider.provideRelatedFiles(extRequestDraft, token) ?? undefined;
 	}
 
-	async $provideCustomAgents(handle: number, options: ICustomAgentQueryOptions, token: CancellationToken): Promise<IExternalCustomAgent[] | undefined> {
-		const providerData = this._customAgentsProviders.get(handle);
+	async $providePromptFiles(handle: number, options: IPromptFileQueryOptions, token: CancellationToken): Promise<IPromptFileResource[] | undefined> {
+		const providerData = this._promptFileProviders.get(handle);
 		if (!providerData) {
 			return Promise.resolve(undefined);
```
aeschli:
```suggestion
			return undefined;
```

src/vs/workbench/contrib/chat/common/promptSyntax/service/promptsServiceImpl.ts
```diff
@@ -276,13 +305,31 @@ export class PromptsService extends Disposable implements IPromptsService {
...
+	private getProviderActivationEvent(type: PromptsType): string | undefined {
+		switch (type) {
+			case PromptsType.agent:
+				return CUSTOM_AGENT_PROVIDER_ACTIVATION_EVENT;
+			case PromptsType.instructions:
+				return INSTRUCTIONS_PROVIDER_ACTIVATION_EVENT;
+			case PromptsType.prompt:
+				return PROMPT_FILE_PROVIDER_ACTIVATION_EVENT;
+			default:
+				return undefined;
```
aeschli: I would just assert here. We can assume at this place that type is a valid PromptsType

src/vscode-dts/vscode.proposed.chatPromptFiles.d.ts
```diff
@@ -0,0 +1,192 @@
...
+	// #region CustomAgentProvider
+
+	/**
+	 * Represents a custom agent resource file (e.g., .agent.md) available for a repository.
+	 */
+	export interface CustomAgentResource {
+		/**
+		 * The unique identifier/name of the custom agent.
+		 */
+		readonly name: string;
```
aeschli: We expect the name and description to be in the frontmatter header. So I would remove it here to avoid confusion

src/vscode-dts/vscode.proposed.chatPromptFiles.d.ts
```diff
@@ -0,0 +1,192 @@
...
+		 */
+		readonly onDidChangeCustomAgents?: Event<void>;
+
+		/**
+		 * Provide the list of custom agents available.
+		 * @param options Optional query parameters.
+		 * @param token A cancellation token.
+		 * @returns An array of custom agent resources or a promise that resolves to such.
+		 */
+		provideCustomAgents(options: CustomAgentQueryOptions, token: CancellationToken): ProviderResult<CustomAgentResource[]>;
```
aeschli: I would call the `options` a `context`. See https://github.com/microsoft/vscode/blob/f4e9bc45294cc3618a1e7182595608891e65b45a/src/vscode-dts/vscode.d.ts#L5814

src/vscode-dts/vscode.proposed.chatPromptFiles.d.ts
```diff
@@ -0,0 +1,192 @@
...
+	/**
+	 * Options for querying custom agents.
+	 */
+	export type CustomAgentQueryOptions = object;
+
+	/**
+	 * A provider that supplies custom agent resources (from .agent.md files) for repositories.
+	 */
+	export interface CustomAgentProvider {
+		/**
```
aeschli, line 41: Could add an id and label for the provider that we could use to group and enable.

aeschli (changes requested):
The changes are good. Only some small suggestions

src/vscode-dts/vscode.proposed.chatPromptFiles.d.ts
aeschli: Maybe just a label is enough.

src/vs/workbench/api/common/extHostChatAgents2.ts
```diff
@@ -511,13 +526,22 @@ export class ExtHostChatAgents2 extends Disposable implements ExtHostChatAgentsS
 		return await provider.provider.provideRelatedFiles(extRequestDraft, token) ?? undefined;
 	}
 
-	async $provideCustomAgents(handle: number, options: ICustomAgentQueryOptions, token: CancellationToken): Promise<IExternalCustomAgent[] | undefined> {
-		const providerData = this._customAgentsProviders.get(handle);
+	async $providePromptFiles(handle: number, context: IPromptFileContext, token: CancellationToken): Promise<IPromptFileResource[] | undefined> {
```
aeschli: Thinking about this more, we should pass `type: PromptsType` to `$providePromptFiles` and use it to know whether to use 'provideCustomAgents', 'provideInstructions' or 'providePromptFiles'. A client could implement a provider that has all these methods that would be completly ok to do.

src/vs/workbench/api/common/extHostChatAgents2.ts
```diff
@@ -479,23 +480,37 @@ export class ExtHostChatAgents2 extends Disposable implements ExtHostChatAgentsS
...
 
 		const disposables = new DisposableStore();
 
 		// Listen to provider change events and notify main thread
-		if (provider.onDidChangeCustomAgents) {
-			disposables.add(provider.onDidChangeCustomAgents(() => {
-				this._proxy.$onDidChangeCustomAgents(handle);
+		// Check for the appropriate event based on the provider type
+		let changeEvent: vscode.Event<void> | undefined;
+		if ('onDidChangeCustomAgents' in provider) {
```
aeschli: We should use a switch over 'type' to know the name of the event. A instruction provider could also have a field 'onDidChangeCustomAgents'. That would be ok to do

aeschli (approved):

dmitrivMS (approved):

alexr00 (approved):

PR #280442, opened 2025-12-01
ensure that after terminal is killed, output can still be toggled

### LINKED ISSUE #280429: terminal output toggle no longer shows when the terminal is disposed of
1. run something in the terminal via agent
2. focus the terminal and kill it
3. 🐛 show terminal output dropdown action disappears

meganrogge:
fixes #280429

Moving over to xterm.js broke this

- Keeps command output visible across expand/collapse without reattaching the mirror.
- Falls back to the stored snapshot once the terminal is gone and keeps that DOM stable.
- Records terminal snapshots (text + line count) at execution time for reliable replay.
- Clears the detached terminal buffer before replaying snapshots so toggles don’t duplicate output.

cc @tyriar

### DIFF
--- a/src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
+++ b/src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
@@ -1039,7 +1039,6 @@ class DetachedTerminalSnapshotMirror extends Disposable {
 		const terminal = await this._getTerminal();
 		terminal.xterm.clearBuffer();
 		terminal.xterm.clearSearchDecorations?.();
-		await new Promise<void>(resolve => terminal.xterm.write('\x1bc', resolve));
 		if (this._container) {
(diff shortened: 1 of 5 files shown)


aiday-mar (dismissed):

src/vs/workbench/contrib/chat/browser/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
```diff
@@ -289,6 +296,9 @@ export class ChatTerminalToolProgressPart extends BaseChatToolInvocationSubPart
 			() => this._onDidChangeHeight.fire(),
 			() => this._ensureTerminalInstance(),
 			() => this._getResolvedCommand(),
+			() => this._terminalData.terminalCommandOutput,
```
roblourens, line 299: 6 unnamed callbacks 😅 I recommend creating a delegate with named props if you can't change the design to eliminate the coupling

roblourens (dismissed):

dmitrivMS (approved):

Issue #263554, opened 2025-08-27
Rendering broken on macOS


madrussa:
Does this issue occur when all extensions are disabled?: Unknown - issue is intermittent and difficult to replicate consistently.

Version: 1.103.2 (Universal)
Commit: 6f17636121051a53c88d3e605c491d22af2ba755
Date: 2025-08-20T16:45:34.255Z
Electron: 37.2.3
ElectronBuildId: 12035395
Chromium: 138.0.7204.100
Node.js: 22.17.0
V8: 13.8.500258-electron.0
OS: Darwin arm64 24.5.0

Steps to Reproduce:

1. Use VSCode with multiple window open. 
2. Allow MacOS to sleep
3. Wake up MacOS and navigate to VSCode
4. VSCode intermittently will completely fail to render - Every window

Rendering completely broken:
<img width="3829" height="1511" alt="Image" src="https://github.com/user-attachments/assets/79c5b4cd-a219-471f-8340-cf7e392a37a5" />

Sorry this bug report is pretty sparse but I'm not able to use the help tools built into vscode to make a report. This rendering issue happened once or twice in the previous version of VSCode, but since the most recent update it is failing more often and requires a restart everytime.

I do know that I haven't added any extensions in a while. Definitely in the last couple of updates.

If you have any suggestions on being able to get some diagnostic information via a terminal command or by setting a flag in VSCode I am happy to do so and to continue to update this issue. Unfortunately due to the nature of this bug I can't disable plugins as I am relying on them day to day.

MrSimmmons:
Version: 1.103.2 (Universal)
Commit: 6f17636121051a53c88d3e605c491d22af2ba755
Date: 2025-08-20T16:45:34.255Z
Electron: 37.2.3
ElectronBuildId: 12035395
Chromium: 138.0.7204.100
Node.js: 22.17.0
V8: 13.8.500258-electron.0
OS: Darwin arm64 24.6.0

MacOS 15.6, M1 Pro

I have been getting this as well. Same symptoms and reproduction steps as OP. Both from closing the lid the opening it after some time, and from letting it auto sleep from the screensaver screen. This does not happen 100% of the time however, my hit rate is maybe 1 in 40 ish?

<img width="1312" height="912" alt="Image" src="https://github.com/user-attachments/assets/bb623274-cc46-4028-9596-19d381e88714" />

[...]

PR #280591, opened 2025-12-02
feat: add setting to control throttling for chat sessions

deepak1556:
For https://github.com/microsoft/vscode/issues/263554

Based on some trace logs from @chrmarti who is occasionally affected by this issue, we are changing the background throttling of the affected windows but I am not sure why that would break compositing even after we restored the original throttling state when the chat requests ended. I will use this setting as a first step to either confirm or eliminate the chat customization as a source of the issue and then proceed to get additional diagnostics from the runtime.

### DIFF
--- a/src/vs/workbench/contrib/chat/browser/chat.contribution.ts
+++ b/src/vs/workbench/contrib/chat/browser/chat.contribution.ts
@@ -330,6 +330,11 @@ configurationRegistry.registerConfiguration({
 				},
 			}
 		},
+		[ChatConfiguration.SuspendThrottling]: {
+			type: 'boolean',
+			description: nls.localize('chat.suspendThrottling', "Controls whether background throttling is suspended when a chat request is in progress, allowing the chat session to continue even when the window is not in focus."),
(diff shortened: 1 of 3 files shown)


chrmarti (dismissed):

src/vs/workbench/contrib/chat/browser/chat.contribution.ts
```diff
@@ -330,6 +330,11 @@ configurationRegistry.registerConfiguration({
 				},
 			}
 		},
+		[ChatConfiguration.SuspendThrottling]: {
+			type: 'boolean',
+			description: nls.localize('chat.suspendThrottling', "Controls whether background throttling is suspended when a chat request is in progress, allowing the chat session to continue even when the window is not in focus."),
+			default: true,
```
bpasero, line 336: should this be `tags: ['preview']` and also have a TODO@deepak1556 to cleanup in the future?

bpasero (changes requested):

bpasero (approved):

PR #280600, opened 2025-12-02
Set implicit context enablement

### LINKED ISSUE #280202: Full file is attached automatically
I keep ending up with the full file attachment enabled in the implicit context widget. I just found one reliable set of repro steps but it happens in other cases too. I just reload the window with any text file open, and after reloading, that file is attached.

Image

alexr00:
Fixes #280202

### DIFF
--- a/src/vs/workbench/contrib/chat/browser/chatInputPart.ts
+++ b/src/vs/workbench/contrib/chat/browser/chatInputPart.ts
@@ -1386,7 +1386,7 @@ export class ChatInputPart extends Disposable implements IHistoryNavigationWidge
 				this._indexOfLastAttachedContextDeletedWithKeyboard = -1;
 				this._handleAttachedContextChange();
 			}));
-		} else {
+		} else if (!this.options.enableImplicitContext && this._implicitContext) {
 			this._implicitContext?.dispose();
 			this._implicitContext = undefined;
 		}
--- a/src/vs/workbench/contrib/chat/browser/chatInputPart.ts
+++ b/src/vs/workbench/contrib/chat/browser/chatInputPart.ts
@@ -522,10 +522,7 @@ export class ChatInputPart extends Disposable implements IHistoryNavigationWidge
 			if (this._inputEditor) {
 				this._inputEditor.updateOptions({ ariaLabel: this._getAriaLabel() });
 			}
-
-			if (this.implicitContext && this.configurationService.getValue<boolean>('chat.implicitContext.suggestedContext')) {
-				this.implicitContext.enabled = this._currentModeObservable.get() !== ChatMode.Agent;
-			}
+			this.setImplicitContextEnablement();
 		}));
 		this._register(this._onDidChangeCurrentLanguageModel.event(() => {
 			if (this._currentLanguageModel?.metadata.name) {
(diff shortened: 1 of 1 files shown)


src/vs/workbench/contrib/chat/browser/chatInputPart.ts
```diff
@@ -1373,15 +1376,19 @@ export class ChatInputPart extends Disposable implements IHistoryNavigationWidge
 		this.chatInputTodoListWidgetContainer = elements.chatInputTodoListWidgetContainer;
 		this.chatInputWidgetsContainer = elements.chatInputWidgetsContainer;
 
-		if (this.options.enableImplicitContext) {
+		if (this.options.enableImplicitContext && !this._implicitContext) {
 			this._implicitContext = this._register(
 				this.instantiationService.createInstance(ChatImplicitContext),
 			);
+			this.setImplicitContextEnablement();
```
alexr00, line 1383: This is the actual fix: set the enablement once we create the implicit context

src/vs/workbench/contrib/chat/browser/chatInputPart.ts
```diff
@@ -1373,15 +1376,19 @@ export class ChatInputPart extends Disposable implements IHistoryNavigationWidge
...
 
 			this._register(this._implicitContext.onDidChangeValue(() => {
 				this._indexOfLastAttachedContextDeletedWithKeyboard = -1;
 				this._handleAttachedContextChange();
 			}));
+		} else {
+			this._implicitContext?.dispose();
+			this._implicitContext = undefined;
```
alexr00: The `&& !this._implicitContext` check and this are not part of the fix, but it seems like we should be doing them.

src/vs/workbench/contrib/chat/browser/chatInputPart.ts
```diff
@@ -547,6 +544,12 @@ export class ChatInputPart extends Disposable implements IHistoryNavigationWidge
 		this.validateCurrentChatMode();
 	}
 
+	private setImplicitContextEnablement() {
+		if (this.implicitContext && this.configurationService.getValue<boolean>('chat.implicitContext.suggestedContext')) {
+			this.implicitContext.enabled = this._currentModeObservable.get().kind !== ChatMode.Agent.kind;
```
alexr00, line 549: I modified this check to compare the `.kind` as just comparing by reference was resulting in `false` for restored Agent mode chats. Do we actually want to include the whole file when not in agent mode?

alexr00 (commented):
@justschen here are some notes on this change.

src/vs/workbench/contrib/chat/browser/chatInputPart.ts
justschen: this was a big point when we changed the implicit context a few iterations ago. it's intended that in every mode except agent mode, the full fiile should be automatically added

alexr00: Ok, then I think my change is correct, as it seems like restored agent mode chats should count as agent mode. Though, maybe there's a separate problem that restored agent mode chats are not the same mode by reference as `ChatMode.Agent`?

src/vs/workbench/contrib/chat/browser/chatInputPart.ts
```diff
@@ -1373,15 +1376,19 @@ export class ChatInputPart extends Disposable implements IHistoryNavigationWidge
...
 				this.instantiationService.createInstance(ChatImplicitContext),
 			);
+			this.setImplicitContextEnablement();
 
 			this._register(this._implicitContext.onDidChangeValue(() => {
 				this._indexOfLastAttachedContextDeletedWithKeyboard = -1;
 				this._handleAttachedContextChange();
 			}));
+		} else {
+			this._implicitContext?.dispose();
```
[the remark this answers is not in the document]
alexr00: Oof. Good catch.

justschen (approved):
thanks! will test this out in the next build

PR #280738, opened 2025-12-02
show inline install X recommendations for alternative agents

joshspicer:
Reads product.json for the list of `chatSessionRecommendations`.  If the extension is not installed, show `Install X`.  If the recommendation has a `postInstallCommand`, fire that once it's available.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "code-oss-dev",
   "version": "1.107.0",
-  "distro": "55e124105c9ac2939053cdc9f50ff7d84fc86294",
+  "distro": "f7daaf68414ef6e47bec698f8babc297f0d90f0d",
   "author": {
     "name": "Microsoft Corporation"
   },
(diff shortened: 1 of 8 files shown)


bpasero (dismissed):
Since we already know that the agents view will merge into the chat view, I think we need to rethink how we approach this: we need to have a dynamic action contributed to the "+" action of the chat view itself because that is going to be the place where users will in the end pick which runtime to choose.

image

In other words: the blue button will be gone pretty soon!

bpasero:
@joshspicer the blue button will be removed as soon as we merge "Agents" into "Chat" view. I am just saying to not invest into the blue button but focus on the ➕ action of the Chat view, as that is the prime target going forward.

Sorry if this was not clear from the beginning 🙏 

Btw I can already predict that we may want to surface this dynamic action in other menus that we still need to implement, namely: some kind of picker in the chat input box. But I assume we can do that with a `MenuId` once we have it  introduced.

dmitrivMS (approved):

PR #280745, opened 2025-12-02
fix: memory leak in markdown hover

SimonSiefke:
Registering the `WorkbenchHoverDelegate` disposable. 

### Before

When calling the chat editor attach all problems and clearing chat context 197 times, the number of `WorkbenchHoverDelegate` functions grows by 1 each time.

before

### After

No more leak is detected


mjbvz:
Thanks! We are current testing and finalizing the upcoming release so I will merge this early next week for our January release

mjbvz (approved):

Tyriar (approved):

PR #280747, opened 2025-12-02
add setting chat.exitAfterDelegation

joshspicer:


meganrogge (approved):

joshspicer:
good point, 'preview' is more accurate https://github.com/microsoft/vscode/pull/280787

PR #280946, opened 2025-12-03
Adjust padding and font size in chat view title container

mrleemurray:
Modify the chat view title container's padding and font size for improved visual consistency. Test by checking the chat view to ensure the title appears correctly with the new styles applied.


bpasero (approved):
Thanks!

PR #280952, opened 2025-12-03
Clarify items in Add Model dropdown with headerFix/add model dropdown header 280906

### LINKED ISSUE #280906: It's not clear what are the items in the Add Model drop-down
Testing #270234

The drop-down lists some names - what are those? I read somewhere these are "model types" but how can I tell it from UI?

Image

Shubhz20:
Fixes microsoft/vscode#280906

## Problem
The "Add Model" dropdown in the Chat Models Widget lists items (providers) without any label or header, making it unclear to users what these items represent (e.g., "model types", "providers").

## Solution
Added a disabled header action "Select a provider to add models" and a separator at the top of the dropdown menu. This explicitly tells the user that the items listed below are providers they can select to add models from.

## Changes
- Modified `src/vs/workbench/contrib/chat/browser/chatManagement/chatModelsWidget.ts` to insert a header and separator into `dropdownActions`.

## Testing
- Verified that the code correctly constructs the action list with the header and separator.
- The header uses `enabled: false` so it acts as a label and is not clickable.


sandy081:
Thanks for the PR. The reported issue is still under discussion which means the fix is not sure. Hence I am closing this.

PR #280957, opened 2025-12-03
Fix #280927: Prevent false positives in rename inference

### LINKED ISSUE #280927: False positive in rename infer
Image

We shouldn't propose a rename in this case since skuPlan already exists as a property on the interface. Need to see if we detect this on the symbol or whether we need to inspect the AST / errors.

Raghav-Somasundaram:
Fixes #280927

Prevents false positive rename suggestions when a property already exists on the interface. The NES (Next Edit Suggestions) feature was incorrectly suggesting renames for properties that already existed on the interface.

This fix validates that the suggested rename doesn't conflict with existing properties before proposing the rename. Specific case: prevents suggesting rename for 'skuPlan' when it already exists as a property on the interface.

### DIFF
--- a/extensions/typescript-basics/renameInferFix.ts
+++ b/extensions/typescript-basics/renameInferFix.ts
@@ -1,0 +1,33 @@
+/**
+ * Fix for issue #280927: False positive in rename infer
+ * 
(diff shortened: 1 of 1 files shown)


dbaeumer:
I mean wrong project.

Issue #280859, opened 2025-12-03
Missing background for recent sessions


rebornix:
Does this issue occur when all extensions are disabled?: Yes/No

Not sure if it's intentionally, we show a dimmed background for the recent sessions view in empty chat in Dark Modern theme, which separates the recent sessions and the remaining of the empty view. 
 
<img width="330" height="446" alt="Image" src="https://github.com/user-attachments/assets/db6c84b0-9e60-4064-a9e9-7ee451f4cb61" />

However, we don't have a proper default bg color for most other default themes

Light Modern
<img width="385" height="455" alt="Image" src="https://github.com/user-attachments/assets/de037997-9258-447e-852b-2937cfec391d" />

Monokai 

<img width="382" height="403" alt="Image" src="https://github.com/user-attachments/assets/78d48997-f20a-45e1-a2b0-5848fd5273d8" />

Without a background, the padding around the sessions feels too large.

bpasero:
@eli-w-king @mrleemurray I commented, I find it way too strong. I think we need a light contrast over the background, not a strong one.

PR #281161, opened 2025-12-04
revised agent sessions bg color token

eli-w-king:
Replaced `editorWidget-background` with `input-background`. Addresses https://github.com/microsoft/vscode/issues/280859#issuecomment-3609646036

https://github.com/user-attachments/assets/549abc55-ed31-4108-99db-3540f4bffa68


bpasero (changes requested):
This is too strong of a color for me

Image

I liked the previous color in dark themes, I just think we need to make it proper for light themes.

PR #281163, opened 2025-12-04
Update input container width for better alignment

bhavyaus:
The chat input container and its associated widgets (editing session, todo list) were misaligned by approximately 0.5-1px on the right edge. max-width: 100% uses constraint-based sizing where the browser calculates intrinsic content width first, then constrains it potentially introducing subpixel rounding differences between elements. Changed to width: 100% so that all elements use explicit percentage-based width calculation

Also see linked issue for screen shot of issue


justschen (approved):
thanks!!!

PR #281268, opened 2025-12-04
Add background color for agent sessions in chat view

mrleemurray:
Introduce a background color for the agent session list and update styles for the agent sessions viewer and title container in the chat view. This enhances the visual distinction of agent sessions. Testing involves verifying the new background color in the chat view and ensuring it integrates well with existing styles.


bpasero (dismissed):
Thanks, I fixed the CI I think you removed one color that was in use.

src/vs/workbench/contrib/chat/browser/media/chatViewPane.css
```diff
@@ -26,7 +26,11 @@
 	.agent-sessions-container {
 		display: flex;
 		flex-direction: column;
-		background-color: var(--vscode-editorWidget-background);
+		background-color: var(--vscode-sessionList-background);
+
+		.agent-sessions-viewer .monaco-list-rows {
```
bpasero: Why is this needed?

bpasero (changes requested):

src/vs/workbench/contrib/chat/browser/media/chatViewPane.css
mrleemurray: I have now found where the original background color is set & have removed this override.

bpasero:
@mrleemurray thanks, I will take this change into my larger PR in https://github.com/microsoft/vscode/pull/281447 since we now have to merge into release branch too.

PR #281275, opened 2025-12-04
Fix terminal icon picker placement

### LINKED ISSUE #211083: Terminal Icon picker weird placement

milanchahar:
Fixes #211083. The terminal icon picker was using 'quickPickTop' offset which caused it to overlay the title bar when the Command Center is enabled. This change updates it to use 'activeContainerOffset.top' to ensure it appears below the title bar.


meganrogge (approved):
Works, thank you!

Image

rzhao271 (approved):

PR #281302, opened 2025-12-04
Fix: Do not suggest implicit activation message when engine does not support it

### LINKED ISSUE #251898: Do not suggest implicit activation message when engine does not support it

Ishiezz:
Fixes #251898
## Summary
This PR addresses the issue where the linter would incorrectly suggest bumping the engine version to `^1.75.0` even when the extension author intentionally targeted an older version. It also fixes a bug in the version check logic that would have caused failures for future major versions (e.g., `2.0.0`).
## Changes
- **Updated Version Check:** The logic `major >= 1 && minor >= 75` was flawed because it would fail for version `2.0.0` (since `0 < 75`). I updated it to correctly handle `major > 1`.
- **Removed Warning:** Removed the diagnostic that suggested bumping the engine version when implicit activation is not supported, as requested in the issue.
- **Cleanup:** Removed the unused `bumpEngineForImplicitActivationEvents` variable.
## Verification
- Verified the version check logic with a standalone script:
    - `1.74.0` -> Correctly identified as NOT supporting implicit activation.
    - `1.75.0`, `1.80.0` -> Correctly identified as supporting it.
    - `2.0.0` -> **Fix verified:** Now correctly identified as supporting it.
- Compiled the `extension-editing` extension successfully with `gulp compile-extension:extension-editing`.


alexdima (approved):
Thank you!

dmitrivMS (approved):

PR #281318, opened 2025-12-04
Fix node-gyp configuration for macOS

Magnus1X:
Removes Windows-specific build settings from .npmrc that were causing Visual Studio detection errors on macOS. This allows npm install to complete successfully on macOS systems.


deepak1556 (changes requested):
Please share the error you are facing

PR #281387, opened 2025-12-04
Add chat back button to activity bar

### LINKED ISSUE #281068: Agent sessions: Show back button when activity bar is in different positions
Testing #281061

*Talked about this morning, but formally filing an issue so we can track it.

Currently the back button will only be rendered when the activity bar is in the default position. We should still show back button when activity bar is in top, bottom, and hidden positions.

hawkticehurst:
Resolves #281068 

Implements support for a back button in the Chat view panel activity bar when in Top, Bottom, and Hidden positions. Specifically adds a `MenuWorkBenchToolbar` to the left of activity bar titles in the `AbstractPaneCompositePart` class.

Position Top:

Screenshot 2025-12-04 at 6 08 01 PM

Position Bottom/Hidden:

Screenshot 2025-12-04 at 6 08 23 PM


src/vs/workbench/browser/parts/paneCompositePart.ts
```diff
@@ -39,6 +40,33 @@ import { getActionBarActions } from '../../../platform/actions/browser/menuEntry
...
+interface IChatViewWithWidget {
+	widget?: {
+		viewModel?: {
+			model?: {
+				sessionResource?: URI;
+				getRequests?: () => unknown[];
+			};
+		};
+	};
+}
```
hawkticehurst, line 69: @benibenj @bpasero wasn't quite sure what to do here so for the moment I've defined local interfaces.

Was specifically unsure about making some of the private properties for model resources/requests public and/or pulling them out into a separate class/interface, so again just defined an interface locally to make TS happy. Would love input from @roblourens or @justschen on if this feels icky or not.

roblourens: I'm not an expert in this area, but I don't think there should be anything chat-specific in AbstractPaneCompositePart. Rather we should figure out how to turn it into a generic feature, which the chat view can enable/configure in some way, or, we figure out how to have a chat-specific `IPaneCompositePart` which has chat features. We might ask Steven for advice, or maybe Ben understands it well.

benibenj: I think there are 3 approaches
1. Have a contributable toolbar to the left of view pane title => I don't think this is a good idea
2. When creating a view pane, the title is passed as an option. We could also have a titleAction which is passed in which uses `when` condition for toggling visibility
3. Change the type of view pane `options.title` and `updateTitle()` to be `string | { action: Action; label: string }` instead of just `string`

Not sure yet which is better. I'm leaning towards 2.

src/vs/workbench/browser/parts/media/paneCompositePart.css
```diff
@@ -3,24 +3,51 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-.monaco-workbench .pane-composite-part > .title.has-composite-bar > .title-actions .monaco-action-bar .actions-container {
+.monaco-workbench
```
bpasero, line 6: Why is this file formatted? Its impossible for me to review this.

src/vs/workbench/browser/parts/paneCompositePart.ts
```diff
@@ -39,6 +40,33 @@ import { getActionBarActions } from '../../../platform/actions/browser/menuEntry
...
 import { HiddenItemStrategy, MenuWorkbenchToolBar } from '../../../platform/actions/browser/toolbar.js';
 import { DeferredPromise } from '../../../base/common/async.js';
+import { IConfigurationService } from '../../../platform/configuration/common/configuration.js';
+import { URI } from '../../../base/common/uri.js';
+
+/**
+ * Context for chat view title actions.
+ * Defined locally to avoid layering violations with contrib layer.
+ */
+interface IChatViewTitleActionContext {
```
bpasero, line 50: I think this is already being mentioned, but we cannot add Chat knowledge in to this generic pane that is the home for many other things like the Explorer.

What I would have expected is a menu that can be contributed to that appears to the left of the pane title, only if there is a contribution.

bpasero (changes requested):

src/vs/workbench/browser/parts/media/paneCompositePart.css
hawkticehurst: I have no idea, and was actually going to ask one of you about it. 

It was bugging the hell out of me, but assumed it was part of some formatting rules we have set up for VS Code –– I guess that's not the case though?

hawkticehurst: Any time I `cmd+s` the file / CSS selectors would be reformatted like that.

src/vs/workbench/browser/parts/paneCompositePart.ts
hawkticehurst: Yep, totally on board with this. Planning to update this PR next week / in January.

Just noticed I forgot to convert this PR to a draft though. I'll update that now so the state of this PR is better reflected.

PR #281536, opened 2025-12-05
fix(debug): support C# stack trace format in debug console links

odinGitGmail:
## Problem

When clicking on C# stack trace links in the Debug Console (e.g., `Program.cs:line 6`), VSCode only opens the file at line 1 instead of navigating to the specific line number.

## Root Cause

The link detector's regex pattern didn't support the `:line 123` format commonly used in C# stack traces.

## Solution

1. Updated `LINE_COLUMN` regex to support both `:line 123` and `:123:45` formats
2. Modified `createPathLink` to only set selection when line number > 0
3. Ensured column defaults to 1 when not specified

## Testing

- ✅ Code formatting passes (hygiene check)
- ⚠️ Local testing was blocked by build environment issues
- Recommend CI/CD testing by the VSCode team

## Related Issue

Fixes navigation in debug console for stack traces across all languages that use the `:line N` format.


connor4312 (dismissed):

dmitrivMS (dismissed):

connor4312 (approved):
thanks!

alexdima (approved):

PR #281621, opened 2025-12-05
Restoring chat model if opening changes

osortega:
This change makes sure the model is restored before trying to open the diff changes

### DIFF
--- a/src/vs/workbench/contrib/chat/browser/agentSessions/agentSessionsActions.ts
+++ b/src/vs/workbench/contrib/chat/browser/agentSessions/agentSessionsActions.ts
@@ -8,7 +8,7 @@ import { localize, localize2 } from '../../../../../nls.js';
 import { getAgentChangesSummary, IAgentSession } from './agentSessionsModel.js';
 import { Action, IAction } from '../../../../../base/common/actions.js';
 import { ActionViewItem, IActionViewItemOptions } from '../../../../../base/browser/ui/actionbar/actionViewItems.js';
-import { CommandsRegistry, ICommandService } from '../../../../../platform/commands/common/commands.js';
+import { ICommandService } from '../../../../../platform/commands/common/commands.js';
 import { EventHelper, h, hide, show } from '../../../../../base/browser/dom.js';
 import { assertReturnsDefined } from '../../../../../base/common/types.js';
 import { Action2, MenuId } from '../../../../../platform/actions/common/actions.js';
@@ -17,7 +17,6 @@ import { ServicesAccessor } from '../../../../../editor/browser/editorExtensions
 import { ViewAction } from '../../../../browser/parts/views/viewPane.js';
 import { AGENT_SESSIONS_VIEW_ID, AgentSessionProviders, IAgentSessionsControl } from './agentSessions.js';
 import { AgentSessionsView } from './agentSessionsView.js';
-import { URI } from '../../../../../base/common/uri.js';
 import { IChatService } from '../../common/chatService.js';
 import { ChatContextKeys } from '../../common/chatContextKeys.js';
 import { IMarshalledChatSessionContext } from '../actions/chatSessionActions.js';
@@ -103,7 +102,8 @@ export class AgentSessionDiffActionViewItem extends ActionViewItem {
 	constructor(
 		action: IAction,
 		options: IActionViewItemOptions,
-		@ICommandService private readonly commandService: ICommandService
+		@ICommandService private readonly commandService: ICommandService,
+		@IChatService private readonly chatService: IChatService
 	) {
 		super(null, action, options);
 	}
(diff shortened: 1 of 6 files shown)


src/vs/workbench/contrib/chat/browser/agentSessions/agentSessionsActions.ts
```diff
@@ -152,18 +155,27 @@ export class AgentSessionDiffActionViewItem extends ActionViewItem {
...
-	session?.editingSession?.show();
+	const commandService = accessor.get(ICommandService);
+	const type = getChatSessionType(resource);
+	let session = chatService.getSession(resource);
+	if (type === AgentSessionProviders.Local) {
+		if (!session) {
+			const chatModelRef = await chatService.getOrRestoreSession(resource);
+			session = chatModelRef?.object;
+		}
+		await session?.editingSession?.show();
```
rebornix: we want to dispose `chatModelRef` at the end

src/vs/workbench/contrib/chat/browser/agentSessions/agentSessionsActions.ts
```diff
@@ -152,18 +155,27 @@ export class AgentSessionDiffActionViewItem extends ActionViewItem {
...
+	} else {
+		if (!session) {
+			await chatService.loadSessionForResource(resource, ChatAgentLocation.Chat, CancellationToken.None);
+		}
+		commandService.executeCommand(`agentSession.${type}.openChanges`, resource);
```
rebornix: I'm not very sure about the command registration heuristics we had here. Is it fine that we don't register the command in core or extensions `package.json`?

src/vs/workbench/contrib/chat/browser/agentSessions/agentSessionsActions.ts
```diff
@@ -152,18 +155,27 @@ export class AgentSessionDiffActionViewItem extends ActionViewItem {
...
 	const chatService = accessor.get(IChatService);
-
-	const session = chatService.getSession(resource);
-	session?.editingSession?.show();
+	const commandService = accessor.get(ICommandService);
+	const type = getChatSessionType(resource);
+	let session = chatService.getSession(resource);
+	if (type === AgentSessionProviders.Local) {
+		if (!session) {
+			const chatModelRef = await chatService.getOrRestoreSession(resource);
```
roblourens: Who's going to dispose this?

src/vs/workbench/contrib/chat/browser/agentSessions/agentSessionsActions.ts
```diff
@@ -152,18 +155,27 @@ export class AgentSessionDiffActionViewItem extends ActionViewItem {
...
+			session = chatModelRef?.object;
+		}
+		await session?.editingSession?.show();
+	} else {
+		if (!session) {
+			await chatService.loadSessionForResource(resource, ChatAgentLocation.Chat, CancellationToken.None);
```
roblourens: This returned a reference too, who owns it?

src/vs/workbench/contrib/chat/browser/agentSessions/agentSessionsActions.ts
```diff
@@ -152,18 +155,27 @@ export class AgentSessionDiffActionViewItem extends ActionViewItem {
 
 	override onClick(event: MouseEvent): void {
 		EventHelper.stop(event, true);
-
-		const session = this.action.getSession();
-
-		this.commandService.executeCommand(`agentSession.${session.providerType}.openChanges`, this.action.getSession().resource);
+		this.commandService.executeCommand(`agentSession.core.openChanges`, this.action.getSession().resource);
```
roblourens: It's preferable to call a function than do `executeCommand`. That way it's type-safe

src/vs/workbench/contrib/chat/browser/agentSessions/agentSessionsActions.ts
roblourens: Where is the command to show the changes implemented? I assume the session needs to stay alive as long as the diff editor is open, how are those two things associated?

src/vs/workbench/contrib/chat/browser/agentSessions/agentSessionsActions.ts
osortega: Disposing now

src/vs/workbench/contrib/chat/browser/agentSessions/agentSessionsActions.ts
osortega: I removed the core command but still we depend on the extensions one, I think this should be part of the API in the future

src/vs/workbench/contrib/chat/browser/agentSessions/agentSessionsActions.ts
osortega: I tracked this and looks like it gets disposed right after loading it because we re-render the action bar and this view gets disposed.
It felt a bit weird that it's kind of a side effect that if the UI does something different then this will leak so I ended up registering to an event that fires when the editing session editor pane closes

src/vs/workbench/contrib/chat/browser/agentSessions/agentSessionsActions.ts
osortega: Changed it so that we keep a reference and dispose it accordingly.
This command is owned by the chat extension

src/vs/workbench/contrib/chat/browser/agentSessions/agentSessionsActions.ts
osortega: Changed it to be a function

src/vs/workbench/contrib/chat/browser/agentSessions/agentSessionsActions.ts
```diff
@@ -152,20 +155,34 @@ export class AgentSessionDiffActionViewItem extends ActionViewItem {
...
+		let session = this.chatService.getSession(resource);
+		if (!session) {
+			const chatModelRef = type === AgentSessionProviders.Local ? await this.chatService.getOrRestoreSession(resource) : await this.chatService.loadSessionForResource(resource, ChatAgentLocation.Chat, CancellationToken.None);
+			session = chatModelRef?.object;
+			const disposable = session?.editingSession?.onDidEditorPaneClose(() => {
+				chatModelRef?.dispose();
+				disposable?.dispose();
+			});
+			if (disposable) {
+				this._register(disposable);
```
roblourens: This is an action view item, on the toolbar, on the list template. You have a chance that the list will dispose the template at some point when it scrolls out of view. Once that happens, this will be disposed and your listener won't work. And if it scrolls out of view, the button would be reused for a different session (I don't think this would break anything as is, it would just be weird)

The button shouldn't own the reference to the ChatModel, they are independent. Now that I see how `editingSession.show()` works, I think the editing session should acquire a reference as long as the editor is visible. Or, something else external which can watch the editor should hold it.

src/vs/workbench/contrib/chat/browser/chatEditing/chatEditingSession.ts
```diff
@@ -435,6 +441,14 @@ export class ChatEditingSession extends Disposable implements IChatEditingSessio
 		}, this._instantiationService);
 
 		this._editorPane = await this._editorGroupsService.activeGroup.openEditor(input, { pinned: true, activation: EditorActivation.ACTIVATE }) as MultiDiffEditor | undefined;
```
roblourens, line 449: I don't know why this uses this way of opening the editor, vs `IEditorService.openEditor`. I see `IEditorService.onDidCloseEditor` and I wonder if that's a safer event to listen on? Not sure

src/vs/workbench/contrib/chat/browser/agentSessions/agentSessionsActions.ts
rebornix: > the editing session should acquire a reference as long as the editor is visible

This is how I expect it to work properly with chat model ref counting. `chatService.getSession(resource);` should not be used to get the chat model as it can be disposed by others (who acquire chat model via `loadSessionForResource`).

PR #281627, opened 2025-12-05
feat(scm): add setting to hide commit input box

### LINKED ISSUE #281562: Add setting to hide Source Control commit input box (scm.showInputBox)
Add a user-facing setting to hide the commit message input box in the Source Control panel.

**Use Case**

When using VS Code with multiple repositories in a workspace, the Source Control panel displays a commit input box for each repository. For users who:

- Prefer committing via terminal/CLI (e.g., using Claude Code, git CLI, or other tools)
- Want a cleaner, more compact Source Control view
- Only use the panel to view changed files, not to commit

The input boxes take up significant vertical space (34px per repository) and cannot currently be hidden.

**Current Workarounds**

- `scm.inputMinLineCount: 1` and `scm.inputMaxLineCount: 1` reduce height but don't hide the box
- CSS injection via extensions can hide the content but cannot collapse the gap due to VS Code's virtualized list using JavaScript-calculated inline styles for row positioning
- The `SourceControlInputBox.visible` property exists in the API but is only accessible to SCM provider extensions, not end users

**Proposed Solution**

Add a setting like:
```json
"scm.showInputBox": false
```

Or per-provider:
```json
"git.showInputBox": false
```

This would:
1. Hide the input box rows in the Source Control panel
2. Collapse the space they occupy (update the virtualized list to skip these rows)
3. Still allow commits via command palette, terminal, or other means

**Additional Context**

- Related: The `git.showActionButton` setting successfully hides commit/sync/publish buttons
- Related: `SourceControlInputBox.visible` API exists for extension authors (issue #62068, PR #60051)
- This would bring user-facing parity with what extension authors can already do programmatically

**System Info**

- VS Code Version: 1.95+
- OS: Linux (Arch)

kgn:
## Summary

Add `scm.showInputBox` setting that allows users to hide the commit message input box in the Source Control view.

- Adds new boolean setting `scm.showInputBox` (default: `true`)
- Filters input box elements from the SCM tree when setting is `false`
- Refreshes the tree when setting changes

## Use Case

This is useful for users who:
- Prefer committing via terminal/CLI (e.g., using git CLI or other tools)
- Want a cleaner, more compact Source Control view
- Only use the panel to view changed files, not to commit

Fixes #281562

### DIFF
--- a/src/vs/workbench/contrib/scm/browser/scm.contribution.ts
+++ b/src/vs/workbench/contrib/scm/browser/scm.contribution.ts
@@ -374,6 +374,11 @@ Registry.as<IConfigurationRegistry>(ConfigurationExtensions.Configuration).regis
 			markdownDescription: localize('showInputActionButton', "Controls whether an action button can be shown in the Source Control input."),
 			default: true
 		},
+		'scm.showInputBox': {
+			type: 'boolean',
+			markdownDescription: localize('showInputBox', "Controls whether the commit input box is shown in the Source Control view."),
+			default: true
+		},
 		'scm.workingSets.enabled': {
 			type: 'boolean',
 			description: localize('scm.workingSets.enabled', "Controls whether to store editor working sets when switching between source control history item groups."),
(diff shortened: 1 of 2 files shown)


kgn:

PR #281662, opened 2025-12-06
Use ref counted IChatModelReference to fix agentSession.local.openChanges

joshspicer:
Unless the chat was already open, clicking on the little file changes icon would not open the diff
image


src/vs/workbench/contrib/chat/browser/agentSessions/agentSessionsActions.ts
```diff
@@ -162,8 +162,8 @@ export class AgentSessionDiffActionViewItem extends ActionViewItem {
 CommandsRegistry.registerCommand(`agentSession.${AgentSessionProviders.Local}.openChanges`, async (accessor: ServicesAccessor, resource: URI) => {
 	const chatService = accessor.get(IChatService);
 
-	const session = chatService.getSession(resource);
-	session?.editingSession?.show();
+	const sessionRef = await chatService.getOrRestoreSession(resource);
```
connor4312: It looks like this reference is leaked

connor4312 (changes requested):

src/vs/workbench/contrib/chat/browser/agentSessions/agentSessionsActions.ts
joshspicer: of course, thanks. fixed.

connor4312 (approved):

PR #281767, opened 2025-12-07
fix: memory leak in debug session

SimonSiefke:
Fixes a memory leak with sources not being cleared when the debug session shuts down.

### Before

When debugging a javascript file 37 times, the number of sources grows from ~150 to 6085:

debug-javascript-step-original

### After

When debugging a javascript file 37 times, while there still seem to be some memory leaks, the number of sources stays constant:

debug-javascript-step


src/vs/workbench/contrib/debug/browser/debugSession.ts
```diff
@@ -50,7 +50,7 @@ import { RawDebugSession } from './rawDebugSession.js';
 
 const TRIGGERED_BREAKPOINT_MAX_DELAY = 1500;
 
-export class DebugSession implements IDebugSession {
+export class DebugSession extends Disposable implements IDebugSession {
```
connor4312: I'm unsure how safe `extends Disposable` is here. We have existing debt around the lifetime of DebugSessions, namely if a debug session gets restarted in a certain way we reuse the DebugSession instance. `Disposable` is backed by a `DisposableStore` which is only made to get disposed once and will warn/fail if reused.

Clearing sources and threads and such is good I would just not make DebugSession extend Disposable yet.

SimonSiefke: I remember doing that in order to prepare registering all the Emitters using this._register(new Emitter()),  

It seems to me these could be memory leaks:

```
private readonly _onDidChangeState = new Emitter<void>();
private readonly _onDidEndAdapter = new Emitter<AdapterEndEvent | undefined>();

private readonly _onDidLoadedSource = new Emitter<LoadedSourceEvent>();
private readonly _onDidCustomEvent = new Emitter<DebugProtocol.Event>();
private readonly _onDidProgressStart = new Emitter<DebugProtocol.ProgressStartEvent>();
private readonly _onDidProgressUpdate = new Emitter<DebugProtocol.ProgressUpdateEvent>();
private readonly _onDidProgressEnd = new Emitter<DebugProtocol.ProgressEndEvent>();
private readonly _onDidInvalidMemory = new Emitter<DebugProtocol.MemoryEvent>();

private readonly _onDidChangeREPLElements = new Emitter<IReplElement | undefined>();
```


Maybe something for later... :)

connor4312 (approved):

dmitrivMS (approved):

PR #209510, opened 2024-04-04
Fixed #114425 prevent mime pollution on install

telamon:
This software is not a plain text editor, and it's definitely not a file manager. Advertising compatibility with such overrides default OS handlers on many distributions, unnecessarily.

Issue #15741, opened 2016-11-18
Allow code to open directories in nautilus file manager


jamesbo13:
- VSCode Version: 1.7.1
- OS Version: Ubuntu 16.04

The nautilus file manager in Ubuntu and other GNOME-based Linux distros has a right click "Open with" option available for opening files and or directories. Currently, vscode supports open with for text files but not directories. Since I often open code repositories with vscode, it would be nice to be able to right click and select "open with ... " and then chose "code" to open a new instance of vscode using the directory.

Steps to Reproduce:

1. In nautilus, choose a folder. Right click and select "open with" submenu
2. Visual Studio Code is not listed in submenu.

If you repeat above with a plain text file, it works.

Nautilus uses the MimeType defined in the *.desktop file for associating applications with a particular file type. For directories, the desired mime-type is inode/directory. Simply changing the following line in code.desktop from

`MimeType=text/plain;`
to
`MimeType=text/plain;inode/directory;`

should suffice (though it may be necessary to log in/log out to have change take effect ... alternatively running update-desktop-database and restarting nautilus may work.

Tyriar:
Thanks @jamesbo13 this works perfectly! You shouldn't need to log in/out for it to apply as we call `desktop-file-install` in the postinst step. This is coming in 1.8.0.

aeschli:
verified on Elementary OS Freya

jonasnordlund:
Just adding a note that I have turn off this feature because it's extremely invasive throughout at least Debian & Xfce. It's not just VS Code getting added to a context menu. Most commands opening a folder in Xfce now opens VS Code. For example, opening the folder of a downloaded photo in Firefox -> Opens in VS Code.

Tyriar:
@jonasnordlund are you sure something else in your setup isn't affecting this? I believe the only change made here though was indicating to the OS that VS Code *can* open directories. You should be able to then configure which application is preferred for that mimetype (`inode/directory`).

What is the output of this command?

```
xdg-mime query default inode/directory
```

Mine points at `nautilus-folder-handler.desktop` which is why nautilus opens.

jamesbo13:
[This post](http://unix.stackexchange.com/questions/296418/audacious-run-when-i-press-open-with-or-show-in-folder-in-xfce-in-arch-for-any-k) seems like a similar issue ... I am suspecting that xfce has made VS Code the default rather than an option (as Tyriar also hints at).

Tyriar:
[...]

PR #281867, opened 2025-12-08
Add inode/directory mime type to code.desktop

qu1ck:
This allows shells to recognize that the app can be used to open directories:

image


qu1ck:
I found related issue and PR that disabled this behavior previously
#15741 and https://github.com/microsoft/vscode/pull/209510

For now I fixed this for myself by creating a custom .desktop in `~/.local` folder that overrides system one. Feel free to close this PR if you still think that having global mime type causes issues with some (in my opinion buggy) shells.

PR #281970, opened 2025-12-08
introduce maxRequestsLimit

isidorn:


Tyriar (approved):

src/vs/workbench/contrib/chat/browser/chat.contribution.ts
```diff
@@ -972,6 +975,7 @@ class ChatAgentSettingContribution extends Disposable implements IWorkbenchContr
 							type: 'number',
 							markdownDescription: nls.localize('chat.agent.maxRequests', "The maximum number of requests to allow per-turn when using an agent. When the limit is reached, will ask to confirm to continue."),
 							default: defaultValue,
+							maximum: maxLimit,
```
roblourens, line 978: This won't actually work as-is because this only drives a validation warning, it won't limit the real value. The setting is read in the extension and we would have to enforce it there.

lramos15 (approved):

PR #282067, opened 2025-12-08
Swap editing session for background agents

### LINKED ISSUE #281915: Agent sessions: `background` session does not indicate changes until opened
Steps to Reproduce:

1. start a new `background` session and pick `workspace` instead of `worktree`
2. let it add JSDoc to some file, leave it run in the background
3. wait for it to complete

=> 🐛 I see no changes indicator until I open the session.

Image

after opening:

Image

DonJayamanne:
Fix https://github.com/microsoft/vscode/issues/281915


src/vs/workbench/api/browser/mainThreadChatSessions.ts
```diff
@@ -451,7 +451,15 @@ export class MainThreadChatSessions extends Disposable implements MainThreadChat
 					inputState: originalModel.inputModel.toJSON()
 				};
 			}
+		}
```
Tyriar: This code flows a little strangely, `newSession` and `originalModel` are both scoped to their own {} block. The fact that it's using a `{}` block though makes me think there's some reason it's being used but I don't think there is a good reason?

Also can/should we use `newSession` instead of looking it up again below?

Tyriar (dismissed):

src/vs/workbench/api/browser/mainThreadChatSessions.ts
DonJayamanne: Agreed, fixed

lramos15 (approved):

PR #282106, opened 2025-12-09
Refresh sessions view on keep/undo

osortega:


bpasero (changes requested):
This makes no sense to me: we have a provider with events, that trigger refresh in the agent sessions. Why would we not be able to send an event through the provider API based on these conditions, why does this change have to go into the viewer? What if we decide to show the agent sessions somewhere else, would you duplicate the code there?

https://github.com/microsoft/vscode/blob/e605ba270d7f81bb29f3146926f8c12a78b74091/src/vs/workbench/contrib/chat/browser/agentSessions/agentSessionsModel.ts#L179-L181

cc @rebornix @joshspicer

PR #282152, opened 2025-12-09
Add command to announce editor line and column number 

### LINKED ISSUE #278855: Add command to announce the current editor line number via `aria.status`
Screen reader users use the go to line command just to hear what line number/column they're currently on. 

The fix for https://github.com/microsoft/vscode/issues/277538 only works when using NVDA. It does work if one arrows down.

@rperez030 and @jooyoungseo have said that a command that announces the current line and cursor position would be nice to have. 

We should also add this to our editor accessibility help dialog.

anthonykim1:
Resolves: https://github.com/microsoft/vscode/issues/278855

### DIFF
--- a/src/vs/editor/common/standaloneStrings.ts
+++ b/src/vs/editor/common/standaloneStrings.ts
@@ -34,6 +34,7 @@ export namespace AccessibilityHelpNLS {
 	export const showAccessibilityHelpAction = nls.localize("showAccessibilityHelpAction", "Show Accessibility Help");
 	export const listSignalSounds = nls.localize("listSignalSoundsCommand", "Run the command: List Signal Sounds for an overview of all sounds and their current status.");
 	export const listAlerts = nls.localize("listAnnouncementsCommand", "Run the command: List Signal Announcements for an overview of announcements and their current status.");
+	export const announceCursorPosition = nls.localize("announceCursorPosition", "Run the command: Announce Cursor Position{0} to hear the current line and column.", '<keybinding:editor.action.announceCursorPosition>');
 	export const quickChat = nls.localize("quickChatCommand", "Toggle quick chat{0} to open or close a chat session.", '<keybinding:workbench.action.quickchat.toggle>');
 	export const startInlineChat = nls.localize("startInlineChatCommand", "Start inline chat{0} to create an in editor chat session.", '<keybinding:inlineChat.start>');
 	export const startDebugging = nls.localize('debug.startDebugging', "The Debug: Start Debugging command{0} will start a debug session.", '<keybinding:workbench.action.debug.start>');
--- a/src/vs/workbench/contrib/codeEditor/browser/accessibility/accessibility.ts
+++ b/src/vs/workbench/contrib/codeEditor/browser/accessibility/accessibility.ts
@@ -12,9 +12,9 @@ import { Action2, registerAction2 } from '../../../../../platform/actions/common
 import { accessibilityHelpIsShown } from '../../../accessibility/browser/accessibilityConfiguration.js';
 import { KeybindingWeight } from '../../../../../platform/keybinding/common/keybindingsRegistry.js';
 import { KeyCode, KeyMod } from '../../../../../base/common/keyCodes.js';
-import { status } from '../../../../../base/browser/ui/aria/aria.js';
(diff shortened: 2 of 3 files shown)


src/vs/editor/common/standaloneStrings.ts
```diff
@@ -34,6 +34,7 @@ export namespace AccessibilityHelpNLS {
 	export const showAccessibilityHelpAction = nls.localize("showAccessibilityHelpAction", "Show Accessibility Help");
 	export const listSignalSounds = nls.localize("listSignalSoundsCommand", "Run the command: List Signal Sounds for an overview of all sounds and their current status.");
 	export const listAlerts = nls.localize("listAnnouncementsCommand", "Run the command: List Signal Announcements for an overview of announcements and their current status.");
+	export const announceCursorPosition = nls.localize("announceCursorPosition", "Run the command: Announce Cursor Position{0} to hear the current line and column.", '<keybinding:editor.action.announceCursorPosition>');
```
anthonykim1, line 37: @meganrogge Should we also add keybinding to this command? 
I'm not sure if we want something along the line with ctrl+g which was the old "hacky" way of folks accessing this info.

src/vs/workbench/contrib/codeEditor/browser/accessibility/accessibility.ts
```diff
@@ -48,3 +49,32 @@ class ToggleScreenReaderMode extends Action2 {
 }
 
 registerAction2(ToggleScreenReaderMode);
+
+class AnnounceCursorPosition extends Action2 {
```
anthonykim1, line 53: @meganrogge Do you think this would be the best/most appropriate place to register this command?
Other places I've been debating are:
1.  `src\vs\workbench\contrib\accessibilitySignals\browser\commands.ts`  
2. `src\vs\workbench\contrib\accessibility\browser\accessibleViewActions.ts`

src/vs/editor/common/standaloneStrings.ts
meganrogge: Wondering if `alt/option+shift+g` would be good ? @jooyoungseo and @rperez030 might know. We can ask on Friday!

src/vs/workbench/contrib/codeEditor/browser/accessibility/accessibility.ts
meganrogge: This looks fine to me!

meganrogge (commented):
Didn't test (we should) but LGTM! We should test using `status` and `alert` and see how each behaves.

src/vs/editor/common/standaloneStrings.ts
meganrogge: They also mentioned `ctrlCmd+shift+g`

meganrogge:
Let's go with `alert` per feedback!

src/vs/workbench/contrib/codeEditor/browser/accessibility/accessibility.ts
```diff
@@ -48,3 +49,32 @@ class ToggleScreenReaderMode extends Action2 {
...
+	constructor() {
+		super({
+			id: 'editor.action.announceCursorPosition',
+			title: nls.localize2('announceCursorPosition', "Announce Cursor Position"),
+			f1: true,
+			// TODO: Get suggestion on Keybinding!. Something with ctrl+g since people are used to it?
```
meganrogge: @anthonykim1 should we try the keybinding they suggested here?

anthonykim1:
Forgot to add keybinding🫢 ` ctrl/cmd + shift+ g ` , but I think we should be good to go after that

meganrogge (approved):
Thanks for working on this 😄

src/vs/workbench/contrib/codeEditor/browser/accessibility/accessibility.ts
anthonykim1: Yeap! @meganrogge  is 
[Add keybinding ctrlCmd+shift+g][0417961](https://github.com/microsoft/vscode/pull/282152/commits/0417961b347ca95c67795a8e6f95f5c02a3bf1bc) good place to put it?

meganrogge (approved):

PR #282223, opened 2025-12-09
Simplify `TestInstantiationService#stub` overloads

Andarist:
**Why**

Current types run into a problem with this proposed patch to TypeScript: https://github.com/microsoft/TypeScript/pull/62243#issuecomment-3629464729

The current types luck out on contextual typing for return expressions, a seamingly equivalent change from a method to an arrow function uncover this:

Screenshot 2025-12-09 at 15 27 19


mjbvz (approved):
Thanks!

Tyriar (approved):

PR #282246, opened 2025-12-09
fix: memory leak in status bar

SimonSiefke:
Fixes a memory leak in extension host status bar items. Generated by GitHub Copilot.

When a status bar item is added with the `$setEntry` method, a disposable is being added: 

```ts
// mainThreadStatus.ts

$setEntry(){
 // disposable added to store, but never removed from store
  this._store.add(toDisposable(() => this.statusbarService.unsetEntry(entryId)));
}
```

However it doesn't seem to be removed in `$disposeEntry`

```ts
$disposeEntry(entryId: string) {
	this.statusbarService.unsetEntry(entryId);
}
```

### Before

When opening and closing an image editor 97 times, which registers 3 status bar items, the number of `MainThreadStatusBar.$setEntry` functions grows by 291 (97*3). 

image-editor open-original

### After

When opening and closing an image editor 97 times, the number of `MainThreadStatusBar.$setEntry` functions stays constant. 

image-editor open


src/vs/workbench/api/browser/mainThreadStatusBar.ts
```diff
@@ -71,11 +72,13 @@ export class MainThreadStatusBar implements MainThreadStatusBarShape {
...
 		const kind = this.statusbarService.setOrUpdateEntry(entryId, id, extensionId, name, text, tooltipOrTooltipProvider, command, color, backgroundColor, alignLeft, priority, accessibilityInformation);
 		if (kind === StatusBarUpdateKind.DidDefine) {
-			this._store.add(toDisposable(() => this.statusbarService.unsetEntry(entryId)));
+			const disposable = toDisposable(() => this.statusbarService.unsetEntry(entryId));
+			this._entryDisposables.set(entryId, disposable);
 		}
 	}
 
 	$disposeEntry(entryId: string) {
+		this._entryDisposables.deleteAndDispose(entryId);
```
jrieken, line 77: I think `deleteAndDispose` also calls unsetEntry and we would now call it twice?

jrieken (changes requested):

src/vs/workbench/api/browser/mainThreadStatusBar.ts
SimonSiefke: Yeah. Good catch!

jrieken (approved):

aiday-mar (approved):

PR #282265, opened 2025-12-09
Cache windows executables 

### LINKED ISSUE #279262: terminal-suggest freezes extension host: isExecutable loop inefficiency on Windows
Yes

Version: 1.107.0-insider (system setup)
Commit: 29ea21caf8e29d76a96bcb651a05feb07891cfa7
Date: 2025-11-24T05:02:38.645Z
Electron: 39.2.0
ElectronBuildId: 12791201
Chromium: 142.0.7444.162
Node.js: 22.21.1
V8: 14.2.231.18-electron.0
OS: Windows_NT x64 10.0.22631

The integrated terminal suggest stopped working for me about 2-3 weeks ago, and found the issue as to why I think.

I profiled the process when the warning popped up, and 95% of the time is being spent in isExecutable inside terminalSuggestMain.js.

Looking at the source for `isExecutable` in the extension, it looks like there's a performance issue in the Windows path. It seems to be recalculating the list of allowed extensions (creating a new Set, new Array, and filtering) for every single file it checks.

Image

```
// It looks like this inner function runs for every file check?
const i = function(e) {
    const n = t.windowsDefaultExecutableExtensions,
          i = new Set;
    if (e)
        for (const [t, a] of Object.entries(e)) !0 === a ? n.push(t) : i.add(t);
    return Array.from(new Set(n)).filter(e => !i.has(e))
}(n);
```

Steps to Reproduce:

1. On Windows, open a workspace with a decent amount of files.
2. Trigger terminal suggestions (or just wait for the background scan).
3. Extension host becomes unresponsive.

Environment:
Windows 11
VS Code Insiders
terminal.integrated.suggest.enabled: true

I thought this might be been due to oh-my-posh, psreadline, malformed paths in path env var, but changing all of those don't seem to make a difference.

The weird thing is, if I run git-bash from the integrated terminal, it works fine:

Image

But any other profile (pwsh 7.5 / pwsh 5 / cmd prompt) dont work at all and just get stuck at "Loading..."

Image

my guess is that for some reason the path-var on my windows system causes it to scan everything, but that same var isnt preset in git-bash for whatever reason?

If you need anymore info let me know. Attaching 3 full cpuprofiles

[exthost-cd6f7d.cpuprofile](https://github.com/user-attachments/files/23736021/exthost-cd6f7d.cpuprofile)
[exthost-d91a91.cpuprofile](https://github.com/user-attachments/files/23736020/exthost-d91a91.cpuprofile)
[exthost-103659.cpuprofile](https://github.com/user-attachments/files/23736022/exthost-103659.cpuprofile)

meganrogge:
fixes #279262

Still can't build on windows, so a test would be appreciated @tyriar.


Tyriar (dismissed):
Yeah the original was really slow. We already have a caching mechanism here:

https://github.com/microsoft/vscode/blob/0a7c09faebf1ac0662d798aefc5414bd245f77b9/extensions/terminal-suggest/src/env/pathExecutableCache.ts#L35-L43

So we should move the transforming of the raw config value to the right data structure there. Perhaps pulled into a `WindowsExecutableExtensionsCache` class to encapsulate?

I think the right format we want is actually just a `Set<string>` with a list of executable extensions. We can then avoid the O(n) `Array.prototype.find` call and instead use an O(1) `Set.prototype.has`.

Tyriar:
Easy repro in pwsh, replace `d:\Playground\many_files` with an empty temp folder:

```
1..20000 | ForEach-Object { [System.IO.File]::Create("d:\Playground\many_files\file$_.txt").Dispose() }
```

Then type `./` and it's really laggy.

Tyriar (approved):

PR #282269, opened 2025-12-09
prevent any key from being detected and sent to terminal

### LINKED ISSUE #281569: "any key" input shouldn't be triggered by "terminal will be closed by tasks" message
It seems like we could handle this case specifically, since this is our message?

Image

This is from
```
🛠️ run_task (call_XfaBQhVuKOh7Vg6xCqOfzSMb) {
  "workspaceFolder": "/Users/roblou/code/vscode-copilot-chat",
  "id": "npm: compile"
}
```

meganrogge:
fixes #281569

Just finishes instead.

Screenshot 2025-12-09 at 12 37 46 PM


dmitrivMS (approved):

meganrogge:
We will release a new stable this week with the fix

PR #282325, opened 2025-12-09
Fix for completing response if needed

### LINKED ISSUE #281630: Duplicated file changes part for background session when `chat.checkpoints.showFileChanges` is turned on

When we turn on `chat.checkpoints.showFileChanges`, it shows up in background session as well. However background session already registers its own file changes part

Image

@osortega I think we should hide this for non-local sessions even if it's turned on. From what I see, `chat.checkpoints.showFileChanges` part is not working properly for background session.

osortega:
Fixes: https://github.com/microsoft/vscode/issues/281630


roblourens (dismissed):

src/vs/workbench/contrib/chat/common/chatServiceImpl.ts
```diff
@@ -615,6 +614,12 @@ export class ChatService extends Disposable implements IChatService {
 		});
 	}
 
+	private completeResponseIfNeeded(response: ChatResponseModel | undefined) {
```
roblourens: Actually can this logic live inside the ChatModel/ChatResponseModel?

roblourens (dismissed):
Seems fine for a candidate but in main, why keep both complete and completeResponseIfNeeded? I think `complete` can just have this behavior?

dmitrivMS (approved):

PR #282528, opened 2025-12-10
chore: delete node-pty prebuilds

rzhao271:
This PR deletes the prebuilt binaries that come with the latest node-pty considering that VS Code does not require them and that they break the VS Code build.


deepak1556 (changes requested):
This doesn't need a fix in node-pty, please follow the pattern similar to https://github.com/microsoft/vscode/blob/f1db08ab87ec1108d0e3c2a20c7ffe5b8a55a23b/build/npm/postinstall.ts#L112-L126 and remove the prebuilds of node-pty during postinstall. We build our own versions for each target platform and the prebuilds were meant for 3rd party clients of node-pty.

build/npm/postinstall.ts
```diff
@@ -152,6 +166,7 @@ for (const dir of dirs) {
 
 	if (/^(.build\/distro\/npm\/)?remote$/.test(dir)) {
 		// node modules used by vscode server
+		removeNodePtyPrebuild(dir);
```
deepak1556: Adding here would be a noop since we haven't installed deps for this directory, remove and add it in https://github.com/microsoft/vscode/blob/f1db08ab87ec1108d0e3c2a20c7ffe5b8a55a23b/build/npm/postinstall.ts#L71 for other directories

rzhao271:
New verification build: https://dev.azure.com/monacotools/Monaco/_build/results?buildId=383817&view=results

deepak1556 (dismissed):

deepak1556:
@rzhao271 update the cachesalt, x64 ran with cache that has prebuilds not removed https://dev.azure.com/monacotools/Monaco/_build/results?buildId=383817&view=logs&jobId=12844632-35fb-5b51-892a-9f6fe8a80e65&j=c3c6d686-fd72-58b2-fb22-32758a96abc0&t=43669b33-f7cc-5b15-d4df-074717949166 leading to the universal step failure.

rzhao271:
New verification build: https://dev.azure.com/monacotools/Monaco/_build/results?buildId=383839&view=results

rzhao271:
Closing now that we have https://github.com/microsoft/node-pty/pull/829

PR #282560, opened 2025-12-10
fix: memory leak in chat list renderer

SimonSiefke:
Fixes a memory leak in chatListRenderer. 

Using `this._register` adds the disposable to the `ChatListRenderer` instead of the  `elementDisposables` or `templateDisposables`. But list renderers (probably)? always need to use `elementDisposables` or `templateDisposables` for registering event listeners / disposables.


justschen (approved):
good catch, thanks!

connor4312 (approved):

PR #282589, opened 2025-12-10
fix: memory leak in pane composite bar

SimonSiefke:
Fixes a memory leak in `paneCompositeBar`. 

It seems the `paneCompositeBar` can be visible for a longer period of time, while it's actions can update many times. 

```ts
private getCompositeActions(){
  // composite actions are updated, but the already registered ones not disposed here
  this.compositeActions.set(compositeId, compositeActions)
}
```

By changing the code to use a `DisposableMap`, it ensures that the previous actions are disposed when the actions are updated.

### Before

When renaming a variable 97 times with refactor preview, it seems 3 actions are created each time:

editor rename-with-preview-before

### After

When renaming a variable 97 times with refactor preview, no more action leaks are detected:

editor rename-with-preview


dmitrivMS:
@SimonSiefke Thank you!

dmitrivMS (approved):

benibenj (approved):

PR #282609, opened 2025-12-11
Add button to stop page load

jruales:
* Adds "X" button to stop loading a website. This button shows instead of the reload button when the page is loading.
* Adds Escape key keybinding for the above action
    * Note: unlike other browser keybindings we've added, we don't prevent default, so that the page doesn't lose access to the Esc key
* When we stop loading a page, we don't show a fatal error. Instead, we ignore errors and just leave the page in a partially loaded state, as browsers normally do

https://github.com/user-attachments/assets/6ea324f3-99d7-4bb9-8da9-f82899b47267

Added after the above video was recorded:
* Refresh button cooldown period to match Edge and Chrome behavior:
    * Add cooldown period of 500ms after clicking the reload button before it turns into the cancel button, to prevent accidental load cancellation when double-clicking
    * Reset the cooldown if the user moves the mouse out of the toolbar


src/vs/platform/browserView/electron-main/browserView.ts
```diff
@@ -242,6 +249,18 @@ export class BrowserView extends Disposable {
 						metaKey: input.meta || false,
 						repeat: input.isAutoRepeat || false
 					});
+				} else if (isEscapeWhileLoading) {
```
kycutler: Let's just add Esc to the keycodes we bubble up. I think we'll want other shortcuts using it too, like cancelling element selection

src/vs/platform/browserView/electron-main/browserView.ts
```diff
@@ -181,6 +181,12 @@ export class BrowserView extends Disposable {
 		webContents.on('did-stop-loading', () => fireLoadingEvent(false));
 		webContents.on('did-fail-load', (e, errorCode, errorDescription, validatedURL, isMainFrame) => {
 			if (isMainFrame) {
+				// Ignore ERR_ABORTED (-3) which is the expected error when stopping a page load.
+				// This way, when the user presses X to stop loading, the page will just stop loading instead of showing an error page.
+				if (errorCode === -3) {
+					return;
```
kycutler, line 188: Do we still send an event with loading:false?

src/vs/workbench/contrib/browserView/electron-browser/browserViewActions.ts
```diff
@@ -137,6 +139,14 @@ class ReloadAction extends Action2 {
 	}
 
 	async run(accessor: ServicesAccessor): Promise<void> {
+		const contextKeyService = accessor.get(IContextKeyService);
+		const inCooldownPeriod = contextKeyService.getContextKeyValue(CONTEXT_BROWSER_IN_RELOAD_COOLDOWN_PERIOD.key);
+
+		// No-op if we're in the cooldown period (prevents accidental double-click double reload)
+		if (inCooldownPeriod) {
+			return;
```
kycutler: Can we just do this in the model? i.e. have it optimistically set itself to loading: true / false, and prevent sending the event from there. Then maybe we can remove the timeout too?

src/vs/platform/browserView/electron-main/browserView.ts
jruales: done

src/vs/platform/browserView/electron-main/browserView.ts
jruales: The `did-stop-loading` event is fired right after the `did-fail-load` event when the user stops the loading. So loading:false still happens. Do you prefer that I be more explicit by unconditionally setting `fireLoadingEvent(false);` here?

src/vs/workbench/contrib/browserView/electron-browser/browserViewActions.ts
jruales: The reason we need to keep the timeout is that I'm trying to achieve is an experience similar to Chrome and Edge, where there's a delay between clicking on the reload button and the time when the reload button turns into the Stop Loading button. We also don't want to mess with the loading state itself because we don't want this 500ms cooldown to apply to other aspects, such as showing the loading spinner.

I can however remove this no-op-enforcing code, since the most important part of the cooldown isn't to prevent double reloads, but rather to prevent accidental Stop Loading after a reload. I'll remove this now.

Please let me know if there's anything else you would suggest here.

src/vs/platform/browserView/electron-main/browserView.ts
jruales: added an explicit loading:false just for clarity, even though not strictly necessary

PR #282633, opened 2025-12-11
fix: Latest input is not stored to the history when sending message to a new chat

### LINKED ISSUE #282498: Chat input history stack position is not reset after starting a new chat
Image

Repro:

1. Press up in chat input to get last sent message
2. Press ctrl+shift+enter to run in new chat
3. Press up, 🐛 the input before the last message was shown, instead of the was that was just send

NriotHrreion:
fix #282498

## Cause

https://github.com/microsoft/vscode/issues/282498#issuecomment-3640269338

## Change

I added an option `storeToHistory` to the interface `IChatAcceptInputOptions`, and set it to `true` in `SendToNewChatAction.run()`. In this way, even if the message to be sent to new chat session is automatically inputted by the system, it will be stored into the history as expected.


src/vs/workbench/contrib/chat/browser/chat.ts
```diff
@@ -203,6 +203,7 @@ export interface IChatAcceptInputOptions {
 	noCommandDetection?: boolean;
 	isVoiceInput?: boolean;
 	enableImplicitContext?: boolean; // defaults to true
+	forceStoreToHistory?: boolean;
```
connor4312:
```suggestion
	// Whether to store the input to history. This defaults to 'true' if the input
	// box's current content is being accepted, or 'false' if a specific input
	// is being submitted to the widget.
	storeToHistory?: boolean;
```

src/vs/workbench/contrib/chat/browser/chatWidget.ts
```diff
@@ -2335,7 +2335,7 @@ export class ChatWidget extends Disposable implements IChatWidget {
 			return;
 		}
 
-		this.input.acceptInput(isUserQuery);
+		this.input.acceptInput(isUserQuery || options?.forceStoreToHistory);
```
connor4312:
```suggestion
		this.input.acceptInput(options?.storeToHistory ?? isUserQuery);
```

src/vs/workbench/contrib/chat/browser/actions/chatExecuteActions.ts
```diff
@@ -707,7 +707,7 @@ class SendToNewChatAction extends Action2 {
 		}
 
 		await widget.clear();
-		widget.acceptInput(inputBeforeClear);
+		widget.acceptInput(inputBeforeClear, { forceStoreToHistory: true });
```
connor4312:
```suggestion
		widget.acceptInput(inputBeforeClear, { storeToHistory: true });
```

connor4312 (commented):
Thanks for the diagnosis and the PR! Couple quick tweaks docs + suggested property naming

connor4312 (dismissed):
thanks!

mjbvz (dismissed):

connor4312 (approved):
thanks!

alexdima (approved):

PR #282693, opened 2025-12-11
Add ARGB color format support for color decorators

### LINKED ISSUE #282673: Add a setting for color picker to parse ARGB instead of RGBA
The Color Decorator feature in VS Code is highly useful.

However, when working with Qt, which utilizes the ARGB hexadecimal color format, the decorator incorrectly parses the color value, as it currently expects RGBA format with no way to change it.

This discrepancy leads to incorrect color previews.

Suggestion: Adding a configuration setting that allows users to choose between the RGBA and ARGB hex color parsing formats would be great.

pgivel:
Fixes #282673

This PR adds support for ARGB hexadecimal color format (`#AARRGGBB`) alongside the existing RGBA format (`#RRGGBBAA`).

This is my first time contributing, feel free to correct anything that may not be right.

## Changes
- **New setting**: `editor.colorDecoratorsHexFormat` with values `'rgba'` (default) and `'argb'`
- **Color parsing**: Modified `Color.Format.CSS.parseHex()` to accept both formats
- **Service chain**: Updated worker services to propagate format parameter (Not sure about this, maybe there could be a better way)
- **Tests**: Added tests for both 4-digit (#ARGB) and 8-digit (#AARRGGBB) formats

## Why ARGB?
Qt and some other frameworks use ARGB format where alpha comes first, while web/CSS uses RGBA where alpha comes last. This PR enables users working with Qt-based applications to have color decorators parse colors correctly.

## Default behavior
The default remains `'rgba'` to maintain backward compatibility. Users must explicitly set `editor.colorDecoratorsHexFormat` to `'argb'` to use the alternative format.


pgivel:
@microsoft-github-policy-service agree company="Genymobile SAS"

aiday-mar (commented):
Hi thank you for filing this issue. I have looked at this PR and have noticed two bugs:

- Notice the default color decorator in the screen recording below. It is initially blue when using the rgba format. When I change to argb, reload the window, and hover on the color decorator, the color picker shows a green color, while the color decorator remains dark blue. These two colors should both be green, with the new format. 

https://github.com/user-attachments/assets/10c71756-d9e4-4330-92f7-7962feda04f5

- When I change the setting from rgba to argb, the color decorators do not automatically update. You have to first reload the window for the changes to take effect.

pgivel:
> Hi thanks for the changes. Yes I got confused while testing. Indeed the color seems to be correct. One thing I noticed though which should also be changed is that when you use the argb color format and cycle through the color representations in the color picker, we still use the rgba format there. So when we arrive to the rgb representation, the color suddenly changes from argb to rgba. I think the color picker rgb representation should mirror the setting:
> 
>  Screen.Recording.2025-12-25.at.20.51.27.mov

Does that mean that I should add support for `argb(int, int, int)` when the setting is active? In my mind this PR was only for adding `#ARGB` support, but I can extend the scope if you wish.

pgivel:
Just resolved merge conflicts from upstream/main. Here's what I did:

`defaultDocumentColorProvider.ts`

- Upstream removed the alpha variable and added a compact parameter to hex formatters
- I kept my ARGB conditional logic but adapted it to work without alpha checks
- Applied compact: true to both formatAHex() and formatHexA() so fully opaque colors show as #RRGGBB instead of #RRGGBBAA/#AARRGGBB

`defaultDocumentColorsComputer.ts`

- Upstream added support for CSS Level 4 syntax (/ separator for alpha in rgb/hsl functions)
- I kept my argb color scheme support and added the \/ to the validation regex
- Decided not to add CSS Level 4 syntax to argb() since it's not a CSS standard (platform-specific format)

`defaultDocumentColorsComputer.test.ts`

- Upstream added tests for CSS Level 4 space-separated and slash-separated syntax
- I kept both my argb() tests and upstream's CSS Level 4 tests since they test different features

pgivel:
@aiday-mar I think it's ready for another review :)

PR #282839, opened 2025-12-11
Auto approve most rg calls by default

### LINKED ISSUE #282824: Terminal tool: Auto approve safe rg by default

Tyriar:
Fixes #282824


src/vs/workbench/contrib/terminalContrib/chatAgentTools/common/terminalChatAgentToolsConfiguration.ts
```diff
@@ -260,6 +260,12 @@ export const terminalChatAgentToolsConfiguration: IStringDictionary<IConfigurati
 			find: true,
 			'/^find\\b.*-(delete|exec|execdir|fprint|fprintf|fls|ok|okdir)\\b/': false,
 
+			// rg (ripgrep)
+			// - `--pre`: Executes arbitrary command as preprocessor for every file searched.
+			// - `--hostname-bin`: Executes arbitrary command to get hostname.
+			rg: true,
+			'/^rg\\b.*(--pre|--hostname-bin)\\b/': false,
```
roblourens, line 267: Are you worried about a scenario where the agent writes flags to a rg config file, then runs rg pointing at that config file, which can contain unsafe flags like --pre?

Tyriar: For this we don't want commands to write files (until we can detect them later) or be able to run any arbitrary command. `find --exec` is a good example of the main thing we're trying to prevent. Since Copilot can write files without approval, if `--pre` can point at it and then execute commands inside the config then it could be used as a method to execute anything which we want to prevent.

DonJayamanne (dismissed):

roblourens (approved):

PR #282857, opened 2025-12-11
Removing extra complete code

osortega:


Tyriar (changes requested):
I think adding this was a recent candidate https://github.com/microsoft/vscode/commit/07b88be629ad283d78b5a6b142c79d6330178354

cc @roblourens

Tyriar (approved):

PR #282883, opened 2025-12-11
Prevent alt buffer entrance from hanging chat agent

### LINKED ISSUE #280705: [inline terminal] less hangs the terminal
Prompt: "less ≤filepath≥". Results in the terminal indefinitely hanging with the loading spinner:

Image

Relates to TPI: https://github.com/microsoft/vscode/issues/280444

meganrogge:
fix #280705

Before: (hanging)

image

After:

Screenshot 2025-12-11 at 5 22 45 PM


Tyriar (dismissed):
Nice idea! Could we move the handling of this into the execute strategies instead so all these cases are consolidated here?

https://github.com/microsoft/vscode/blob/f83b2784ce8a4588efce81455b073b38592ed965/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/executeStrategy/richExecuteStrategy.ts#L49-L67

It'll be duplicated but it's way more obvious and keeps `RunInTerminalTool` simpler that way. If it makes sense to share code that can go into strategyHelpers.ts

dmitrivMS (approved):

Issue #276220, opened 2025-11-07
Agents sessions: indicate need for user input like approvals


rebornix:
Testing #274477

<img width="1024" height="325" alt="Image" src="https://github.com/user-attachments/assets/d771e54b-b26b-41bb-bcbf-f130fb69ab9e" />

When an agent session requests permission, we may want to render this info on the list view so users know they need to come back to the agent session to provide feedback.

bpasero:
fyi @jo-oikawa @mrleemurray our designs need to include this state: a session is in progress but also stuck on user input

connor4312:
related: https://github.com/microsoft/vscode/issues/280617

bpasero:
I really think we need something stronger. Lets work on it in December/January, this should surface to the user with a stronger indicator on a session card. 

//cc @jo-oikawa @eli-w-king

bpasero:
@roblourens , @osortega and me will drive this. Me from UX, others from core.

PR #282903, opened 2025-12-12
Support for NeedsInput response status

osortega:
Ref: https://github.com/microsoft/vscode/issues/276220

### DIFF
--- a/src/vs/workbench/contrib/chat/browser/agentSessions/agentSessionsFilter.ts
+++ b/src/vs/workbench/contrib/chat/browser/agentSessions/agentSessionsFilter.ts
@@ -166,6 +166,7 @@ export class AgentSessionsFilter extends Disposable implements Required<IAgentSe
 			{ id: ChatSessionStatus.Completed, label: localize('chatSessionStatus.completed', "Completed") },
 			{ id: ChatSessionStatus.InProgress, label: localize('chatSessionStatus.inProgress', "In Progress") },
 			{ id: ChatSessionStatus.Failed, label: localize('chatSessionStatus.failed', "Failed") },
+			{ id: ChatSessionStatus.NeedsInput, label: localize('chatSessionStatus.needsInput', "Needs Input") },
 		];
 
 		const that = this;
--- a/src/vs/workbench/contrib/chat/browser/chatSessions.contribution.ts
+++ b/src/vs/workbench/contrib/chat/browser/chatSessions.contribution.ts
@@ -348,7 +348,7 @@ export class ChatSessionsService extends Disposable implements IChatSessionsServ
 	private async updateInProgressStatus(chatSessionType: string): Promise<void> {
 		try {
 			const items = await this.getChatSessionItems(chatSessionType, CancellationToken.None);
-			const inProgress = items.filter(item => item.status === ChatSessionStatus.InProgress);
+			const inProgress = items.filter(item => item.status && this.isChatSessionInProgressStatus(item.status));
 			this.reportInProgress(chatSessionType, inProgress.length);
 		} catch (error) {
 			this._logService.warn(`Failed to update in-progress status for chat session type '${chatSessionType}':`, error);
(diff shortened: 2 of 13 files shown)


src/vs/workbench/contrib/chat/common/chatModel.ts
```diff
@@ -1011,9 +1012,12 @@ export class ChatResponseModel extends Disposable implements IChatResponseModel
...
-				lastStartedWaitingAt = pending.startedWaitingAt;
-			} else if (!pending && lastStartedWaitingAt) {
+			if (pending) {
+				this._modelState.set({ value: ResponseModelState.NeedsInput }, undefined);
+				if (!lastStartedWaitingAt) {
+					lastStartedWaitingAt = pending.startedWaitingAt;
+				}
+			} else if (lastStartedWaitingAt) {
 				this._timeSpentWaitingAccumulator += Date.now() - lastStartedWaitingAt;
 				lastStartedWaitingAt = undefined;
```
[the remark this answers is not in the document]
osortega: Added it

dmitrivMS (approved):

bpasero:
@osortega having a service method only for this purpose seems like a lot of overhead, https://github.com/microsoft/vscode/pull/282943 just adds a new method to use and thus avoids the service dependency.

src/vs/workbench/contrib/chat/common/chatModel.ts
```diff
@@ -1011,9 +1012,16 @@ export class ChatResponseModel extends Disposable implements IChatResponseModel
...
+				this._modelState.set({ value: ResponseModelState.NeedsInput }, undefined);
+				if (!lastStartedWaitingAt) {
+					lastStartedWaitingAt = pending.startedWaitingAt;
+				}
+			} else if (lastStartedWaitingAt) {
+				// Restore state to Pending if it was set to NeedsInput by this observable
+				if (this._modelState.read(reader).value === ResponseModelState.NeedsInput) {
+					this._modelState.set({ value: ResponseModelState.Pending }, undefined);
+				}
 				this._timeSpentWaitingAccumulator += Date.now() - lastStartedWaitingAt;
```
roblourens, line 1025: This removed the `&& lastStartedWaitingAt` check around this, is this still right?

osortega: I added it to the else if statement, I believe it's the same logical operation just split, or did I miss something?

roblourens: Sorry I think I read it wrong

src/vs/workbench/contrib/chat/browser/chatSessions.contribution.ts
```diff
@@ -1083,6 +1083,10 @@ export class ChatSessionsService extends Disposable implements IChatSessionsServ
 	public getContentProviderSchemes(): string[] {
 		return Array.from(this._contentProviders.keys());
 	}
+
+	public isChatSessionInProgressStatus(state: ChatSessionStatus): boolean {
```
connor4312, line 1087: personally I would tend to make this a free-floating function so that people don't need the whole service just to check the status. I like having a namespace so I get pseudo-methods on the enums like `ChatSessionStatus.isInProgress(state)`

https://github.com/microsoft/vscode/blob/17cf1ce11ac873561637f74dadb80dc84a4255d2/src/vs/workbench/contrib/mcp/common/mcpTypes.ts#L576-L588

src/vs/workbench/contrib/chat/browser/agentSessions/agentSessionsFilter.ts
```diff
@@ -166,6 +166,7 @@ export class AgentSessionsFilter extends Disposable implements Required<IAgentSe
 			{ id: ChatSessionStatus.Completed, label: localize('chatSessionStatus.completed', "Completed") },
 			{ id: ChatSessionStatus.InProgress, label: localize('chatSessionStatus.inProgress', "In Progress") },
 			{ id: ChatSessionStatus.Failed, label: localize('chatSessionStatus.failed', "Failed") },
+			{ id: ChatSessionStatus.NeedsInput, label: localize('chatSessionStatus.needsInput', "Needs Input") },
```
connor4312, line 169: I would perhaps special case this so that we also show "Needs Input" sessions when the user filters for "In Progress" sessions

PR #282969, opened 2025-12-12
Add tsconfig path resolution following Node's  semantics

### LINKED ISSUE #276592: typescript language features do not resolve tsconfig extends specifier with package exports subpath remapping

dmitrivMS:
Fixes #276592


extensions/typescript-language-features/src/utils/packageExports.ts
```diff
@@ -0,0 +1,291 @@
...
+/**
+ * Apply the captured wildcard segment to an exports target.
+ *
+ * @see https://nodejs.org/api/packages.html#subpath-patterns
+ */
+function replaceWildcard(target: string, star: string | undefined): string {
+	if (star === undefined) {
+		return target;
+	}
+	return target.replace('*', star);
```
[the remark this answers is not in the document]
dmitrivMS: It is actually meant to replace only one occurence.

extensions/typescript-language-features/src/languageFeatures/tsconfig.ts
```diff
@@ -156,6 +157,125 @@ async function resolveNodeModulesPath(baseDirUri: vscode.Uri, pathCandidates: st
...
+		subpath: parts.slice(1).join(posix.sep)
+	};
+}
+
+/**
+ * Walks up from `baseDirUri` and looks for `node_modules/<packageName>`.
+ *
+ * @see https://nodejs.org/api/esm.html#resolution-algorithm-specification (PACKAGE_RESOLVE)
+ */
+async function findNodeModulePackageRoot(baseDirUri: vscode.Uri, packageName: string): Promise<vscode.Uri | undefined> {
```
mjbvz, line 200: Most of our js language features are scoped to the current workspace. Maybe we should do the same here instead of walking outside of the workspace

extensions/typescript-language-features/src/test/unit/tsconfigExportsResolver.test.ts
```diff
@@ -0,0 +1,141 @@
...
+			// noop
+		}
+	});
+
+	test('resolves an exports subpath mapping from package.json in node_modules', async () => {
+		const baseDir = vscode.Uri.joinPath(root, 'a', 'b', 'c');
+		await vscode.workspace.fs.createDirectory(baseDir);
+
+		const packageRoot = vscode.Uri.joinPath(root, 'node_modules', 'pkg');
+		await vscode.workspace.fs.createDirectory(vscode.Uri.joinPath(packageRoot, 'configs'));
```
mjbvz, line 34: If the tests need to use the file system, can we mock it out instead? I've found using the real FS is a very common source of flaky tests and slowness

extensions/typescript-language-features/src/languageFeatures/tsconfig.ts
```diff
@@ -156,6 +157,125 @@ async function resolveNodeModulesPath(baseDirUri: vscode.Uri, pathCandidates: st
...
+ * Resolve a module specifier using `package.json#exports`.
+ *
+ * This is used for `tsconfig.json` `extends` links so that the link resolution
+ * matches Node/TypeScript behavior when packages use `exports` subpath remapping.
+ *
+ * @see https://nodejs.org/api/packages.html#package-entry-points
+ * @see https://nodejs.org/api/esm.html#resolution-algorithm-specification (PACKAGE_RESOLVE)
+ * @see https://nodejs.org/api/esm.html#resolution-algorithm-specification (PACKAGE_EXPORTS_RESOLVE)
+ */
+export async function resolveNodeModulePathUsingExports(baseDirUri: vscode.Uri, specifier: string): Promise<vscode.Uri | undefined> {
```
mjbvz, line 278: Could this use `require.resolve` or an external npm dependency that does this for us? It's quite a lot of code to add and maintain for a lower impact bug

mjbvz (commented):
Thanks for looking into this!

I agree it's a lot of code for a smaller bug. Let me try reaching out to TypeScript to see if they would like to take ownership of tsconfig intellisense with TS-go. That would likely provide the best user experience since it would exactly match how TS resolves these files

extensions/typescript-language-features/src/languageFeatures/tsconfig.ts
dmitrivMS: I considered it, but my understanding is that we need to use vscode's FS and `require.resolve` would not do that.

extensions/typescript-language-features/src/languageFeatures/tsconfig.ts
```diff
@@ -156,6 +157,125 @@ async function resolveNodeModulesPath(baseDirUri: vscode.Uri, pathCandidates: st
...
+	if (!parsed) {
+		return undefined;
+	}
+
+	const packageRoot = await findNodeModulePackageRoot(baseDirUri, parsed.packageName);
+	if (!packageRoot) {
+		return undefined;
+	}
+
+	const packageJson = await tryReadPackageJson(packageRoot);
```
dmitrivMS, line 289: If we go forward with this, we should probably add a cache for JSONC parse results to avoid reparsing on every call.

dmitrivMS: added cache

extensions/typescript-language-features/src/test/unit/tsconfigExportsResolver.test.ts
dmitrivMS: Mocking the FS here is quite some work since TestFS does not appear to be available in extension tests. I can do this, but would like to confirm this will be going in.

dmitrivMS:
Closing for now, waiting for TS team to follow up on their side.

Issue #272928, opened 2025-10-23
Less rounded corners for NES widget


ulugbekna:
code looks squished? but also inconsistent with the rest of vscode UI: tabs corner radius, block cursor radius, etc.

<img width="2294" height="518" alt="Image" src="https://github.com/user-attachments/assets/0d49ef46-df4a-44fb-83c2-814a3b488086" />

PR #283057, opened 2025-12-12
Refactor border radius for inline edit components

mrleemurray:
Reduce the border radius for various inline edit components from 4px to 3px to achieve a more consistent design.

Addresses #272928


benibenj (approved):
I extracted the border radius into a const so it's easier to tweak in the future. Also, I undid the padding change for the long distance preview editor. We can't just change the padding in the widget to be applied only vertically as these paddings are part of the layout computation. If you'd like to make that change you'll need to split `previewEditorMargin` into a horizontal and vertical constant.

rebornix (approved):

PR #283065, opened 2025-12-12
Remove throttle on pty input

### LINKED ISSUE #246204: Python host constantly loosing Terminal connection on Mac OS Silicon M1 since Version 1.93.1
Does this issue occur when all extensions are disabled?: Yes but the Python extensions which are required: Pylance, Python, Python Debugger

- VS Code Version: 1.93.1 to 1.99.1
- OS Version: Mac OS 15.1, Apple Silicon version, zsh terminal

Steps to Reproduce:

1. Select large chunk of python code (including "import" statements)
2. Use Shift + Enter to execute

=> the code starts to execute, but as soon as there are more time costly functions (after several seconds) the connection the the terminal is lost (Pty host lights up in yellow at the bottom of the screen).

The Issue started some time ago, and the only solution for me was to downgrade (to 1.92.2). It happened again now after the update to 1.99.1. I googled and chatGPTied and tried multiple fixes, but none works so far.

### LINKED ISSUE #283056: Remove ugly throttling mechanism when pasting to the terminal
@Jarred-Sumner pointed out in https://x.com/jarredsumner/status/1999317065237512224 this throttling mechanism on paste/`sendText` (my response at https://x.com/Tyriar/status/1999487637674147925):

https://github.com/microsoft/vscode/blob/f83b2784ce8a4588efce81455b073b38592ed965/src/vs/platform/terminal/node/terminalProcess.ts#L473-L481

https://github.com/microsoft/vscode/blob/f83b2784ce8a4588efce81455b073b38592ed965/src/vs/platform/terminal/node/terminalProcess.ts#L525-L544

https://github.com/microsoft/vscode/blob/f83b2784ce8a4588efce81455b073b38592ed965/src/vs/platform/terminal/common/terminalProcess.ts#L73-L109

This was added as one of the attempts to try fix an issue where sending large amounts of data to the terminal would drop chunks of it https://github.com/microsoft/vscode/issues/38137

This comment is the most relevant explanation as to why I thought this was happening at the time. I believe I was also never able to reproduce the problem on my machine:

https://github.com/microsoft/vscode/blob/f83b2784ce8a4588efce81455b073b38592ed965/src/vs/platform/terminal/common/terminalProcess.ts#L73-L81

The justification for the original fix is that correctness is much more important than performance here as pasting large amounts of data like this is an edge case, at least in a terminal within an editor. It sucks though so we should fix it.

Tyriar:
See https://github.com/microsoft/node-pty/pull/831 for a detailed write up on the cause and solution.

Fixes #283056
Fixes #246204

cc @deepak1556 

---

After test by pasting 50kb of data takes ~500ms - the reason that this is not instantaneous would likely require moving all writing to the fd into C++ (https://github.com/microsoft/node-pty/issues/833):

Recording 2025-12-15 at 04 29 24

After test ensure EAGAIN is correctly handled by pasting > 1kb during a shell sleep:

Recording 2025-12-15 at 04 32 49


dmitrivMS (dismissed):

deepak1556 (dismissed):

src/vs/platform/terminal/node/terminalProcess.ts
```diff
@@ -471,13 +459,15 @@ export class TerminalProcess extends Disposable implements ITerminalChildProcess
 	}
 
 	input(data: string, isBinary: boolean = false): void {
-		if (this._store.isDisposed || !this._ptyProcess) {
-			return;
+		this._logService.trace('node-pty.IPty#write', data, isBinary);
+		if (isBinary) {
+			// TODO: node-pty's write should accept a Buffer, needs https://github.com/microsoft/node-pty/pull/812
+			// eslint-disable-next-line local/code-no-any-casts, @typescript-eslint/no-explicit-any
```
deepak1556: is this still a TODO given the relevant node-pty change is merged and present in this version ?

Tyriar: Yes, I meant to remove that after the node-pty updated 👍

deepak1556 (approved):

PR #283138, opened 2025-12-12
debug: re-fetch threads when change without pause

### LINKED ISSUE #282777: Invalidated event clears thread list when all threads running

- VS Code Version: 1.106.3, 1.107.0
- OS Version: Linux x86_64 Ubuntu 20.04

## Problem

When a debug adapter sends an `invalidated` DAP event with `areas: ['threads']` while all threads are running, the threads erroneously disappear from the _Call Stack_ view, leaving an empty view.

### Steps to Reproduce

1. Clone https://github.com/microsoft/vscode-mock-debug.git and open it in Visual Studio Code.
2. Apply the following modifications:
   ```diff
   diff --git a/src/mockDebug.ts b/src/mockDebug.ts
   index b02f49e..ae7761c 100644
   --- a/src/mockDebug.ts
   +++ b/src/mockDebug.ts
   @@ -130,7 +130,7 @@ export class MockDebugSession extends LoggingDebugSession {
    			this.sendEvent(e);
    		});
    		this._runtime.on('end', () => {
   -			this.sendEvent(new TerminatedEvent());
   +			console.log(new TerminatedEvent());
    		});
    	}
    
   @@ -810,7 +810,7 @@ export class MockDebugSession extends LoggingDebugSession {
    		if (command === 'toggleFormatting') {
    			this._valuesInHex = ! this._valuesInHex;
    			if (this._useInvalidatedEvent) {
   -				this.sendEvent(new InvalidatedEvent( ['variables'] ));
   +				this.sendEvent(new InvalidatedEvent( ['threads'] ));
    			}
    			this.sendResponse(response);
    		} else {
   
   ```
3. Build and run the extension according to _readme.md_.
4. In the extension host, open the _Run_ view.
5. Uncheck both _Exceptions_ checkboxes under _Breakpoints_.
6. Start debugging the _Debug readme.md_ configuration. The debugger stops on the first line.
7. Click _Continue_. The _Call Stack_ view shows _thread 1_ and _thread 2_, both _RUNNING_. 
8. From the Command Palette, choose _Toggle between decimal and hex formatting_.

### Actual Result

The _Call Stack_ view becomes empty, both threads have disappeared.

### Expected Result

No visible change, both threads still present and running.

## Analysis

It looks like the code in

https://github.com/microsoft/vscode/blob/e6aeab60511647b9b00f0bf2f0f02b15f278abb5/src/vs/workbench/contrib/debug/browser/debugSession.ts#L1326-L1342

is missing a provision for the case that `this.stoppedDetails` is empty, in which case the old thread list is cleared but no `handleStop` calls are done to fill in a new one.

Also, the lines

https://github.com/microsoft/vscode/blob/e6aeab60511647b9b00f0bf2f0f02b15f278abb5/src/vs/workbench/contrib/debug/browser/debugSession.ts#L1333-L1335

look overcomplicated for what they do, which makes me suspicious whether they actually do what their author @connor4312 (in #200789) intended. Was the intention to call `handleStop` on all elements of `stoppedDetails`? That’s not what the code does (judging from inspection only, I have not debugged it). It only calls `handleStop` on the first element, and leaves behind `stoppedDetails` with the first element duplicated and all further elements thrown away. (The first line does not make a copy of the list, the second and third lines both operate on the same list.)

If desired, I can try to come up with a PR.

connor4312:
Closes #282777

### DIFF
--- a/src/vs/workbench/contrib/debug/browser/debugSession.ts
+++ b/src/vs/workbench/contrib/debug/browser/debugSession.ts
@@ -12,6 +12,7 @@ import { CancellationToken, CancellationTokenSource } from '../../../../base/com
 import { canceled } from '../../../../base/common/errors.js';
 import { Emitter, Event } from '../../../../base/common/event.js';
 import { normalizeDriveLetter } from '../../../../base/common/labels.js';
+import { Lazy } from '../../../../base/common/lazy.js';
 import { Disposable, DisposableMap, DisposableStore, MutableDisposable, dispose } from '../../../../base/common/lifecycle.js';
 import { mixin } from '../../../../base/common/objects.js';
 import * as platform from '../../../../base/common/platform.js';
@@ -65,7 +66,13 @@ export class DebugSession implements IDebugSession {
 	private cancellationMap = new Map<number, CancellationTokenSource[]>();
 	private readonly rawListeners = new DisposableStore();
 	private readonly globalDisposables = new DisposableStore();
-	private fetchThreadsScheduler: RunOnceScheduler | undefined;
+	private fetchThreadsScheduler = new Lazy(() => {
+		const inst = new RunOnceScheduler(() => {
+			this.fetchThreads();
+		}, 100);
+		this.rawListeners.add(inst);
+		return inst;
+	});
 	private passFocusScheduler: RunOnceScheduler;
 	private lastContinuedThreadId: number | undefined;
 	private repl: ReplModel;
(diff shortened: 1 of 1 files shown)


dmitrivMS (approved):

src/vs/workbench/contrib/debug/browser/debugSession.ts
```diff
@@ -1331,8 +1331,14 @@ export class DebugSession implements IDebugSession {
 				this.model.clearThreads(this.getId(), true);
 
 				const details = this.stoppedDetails;
-				this.stoppedDetails.length = 1;
-				await Promise.all(details.map(d => this.handleStop(d)));
+				this.stoppedDetails.length = 0;
+				if (details.length) {
```
cwalther, line 1335: Are you sure this works? I would say this is always false now, we always land in the `else`. `details` is the same object as `this.stoppedDetails` and you just set its `length` to zero. See my original concerns about these lines in #282777.

I would suggest replacing line 1333 by `const details = this.stoppedDetails.slice();` or `const details = [...this.stoppedDetails];`.

connor4312: Indeed 🤦

Issue #158548, opened 2022-08-19
ShellIntegration.ps1 fails when PowerShell running in `ConstrainedLanguage` mode


daxian-dbw:
Does this issue occur when all extensions are disabled?: Yes/No

- VS Code Version: 1.70.
- OS Version: Windows

@Tyriar, just want to bring this issue to your attention: https://github.com/PowerShell/PowerShell/issues/17900 -- the `ShellIntegration.ps1` failed for a PowerShell user because PowerShell was running in `ConstrainedLanguage` mode, where method invocation is not allowed.

I'm not sure if the integration script could work in the restricted language mode. If it's just the `[Console]::Write` method call causing the problem, maybe just replace them with `Write-Output`. If the script wouldn't work in that language mode, then it's better to not inject to the profile by checking `$ExecutionContext.SessionState.LanguageMode` (maybe only run the integration script when PowerShell is running in `FullLanguage`).

VSCodeTriageBot:
Thanks for creating this issue! It looks like you may be using an old version of VS Code, the latest stable release is 1.70.2. Please try upgrading to the latest version and checking whether this issue remains.

Happy Coding!

fflaten:
> If the script wouldn't work in that language mode, then it's better to not inject to the profile by checking `$ExecutionContext.SessionState.LanguageMode` (maybe only run the integration script when PowerShell is running in `FullLanguage`).

Hopefully `Write-Output` is enough to fix it.

If not, be aware that checking that variable will throw in `ConstrainedMode` and `RestrictedMode`. Restricted also blocks try/catch.

PR #283178, opened 2025-12-12
Terminal: allow PowerShell 7.4+ constrained language audit mode to work with shell integrations

### LINKED ISSUE #283151: Terminal Shell Integration requires update for newer PowerShell 7.4+ ConstrainedLanguage AUDIT mode

- VS Code Version:  1.107.0
- OS Version:  Windows 11 25H2

## Steps to Reproduce:

1. On Windows 11, enable a WDAC / App Control **audit-only** policy (On newer systems this is slowly being automatically enabled)
2. Install PowerShell **7.4 or later**. It will run in `ConstrainedLanguage AUDIT` Mode
3. Launch VS Code
4. Open the integrated terminal using `pwsh` profile
5. Observe that PowerShell terminal shell integration does not initialize (prompt tracking / command boundaries unavailable)

## Expected Result:

Shell integration should initialize when:
- `$ExecutionContext.SessionState.LanguageMode` is `FullLanguage`, or
- `$ExecutionContext.SessionState.LanguageMode` is `ConstrainedLanguage` **and** the system lockdown policy is `Audit`
(Refer to Additional Information below for more details as to why)

Shell integration should continue to be blocked when constrained language mode is **enforced**. (However it would be nice if the integration explicitly states that it is being blocked due to a restricted language mode - something that is not super evident to unfamiliar devs).

## Actual Result:

- PowerShell runs normally, in unrestricted fashion
- Scripts execute without restriction. For example, dynamic code (e.g. `Add-Type`) works fine
- However, Terminal shell integration does not initialize because it **assumes** restricted shell
- VS Code reports terminal integration failure, despite a fully functional shell
- This issue also affects downstream IDEs that reuse VS Code’s terminal integration logic (e.g. Windsurf, Cursor).

A minimal fix that allows shell integration in `ConstrainedLanguage + Audit mode`, while continuing to block other enforced restricted language modes, is being prepared as a follow-up PR and should be up soon.

## Additional Information:

PowerShell **7.4** (released Nov 2023) introduced **Constrained Language Audit Mode**.

In this mode:
- `LanguageMode` reports `ConstrainedLanguage`
- But runtime semantics are equivalent to `FullLanguage`
- No restrictions are actually enforced
- The mode exists solely to log what *would* be blocked under enforcement

Audit vs enforced CLM can be detected via:

```powershell
[System.Management.Automation.Security.SystemPolicy]::GetSystemLockdownPolicy()
```

The current language mode guard appears to originate from: #158548

That change predates the introduction of audit-only constrained language mode and was totally correct at the time. Since PowerShell semantics have evolved, the same guard now produces false negatives in audit-only scenarios by treating all `ConstrainedLanguage` sessions as enforced.

It is also worth noting that WDAC / App Control audit policies can become active implicitly on Windows 11 without explicit user configuration. Common cases include:
1) Activation of the `Verified and Reputable Desktop` baseline policy on consumer, developer-mode enabled machines not enrolled in any kind of MDM (think freelance/solo devs or small teams).
2) PowerShell Developer-oriented configurations where Windows prefers audit-first behavior instead of enforcement of PS1 script-breaking policies
3) OEM or security tooling that deploys audit policies as part of a hardening baseline, to meet Copilot+ PC certifications for example
4) User Toggling of advanced platform features (e.g. Copilot+ / Recall–class features), and connected VBS and related security protections

As a result, audit-only `ConstrainedLanguage` mode is becoming increasingly common on modern Windows 11 systems. In its defense, this is just `FullLanguage` mode with extra logs and accountability and should be treated as such.

skbtwiz:
# Fixes https://github.com/microsoft/vscode/issues/283151

## Description
PowerShell 7.4 introduced **Constrained Language Audit Mode**, where `$ExecutionContext.SessionState.LanguageMode` reports `ConstrainedLanguage` but runtime behavior is equivalent to `FullLanguage` Mode. This is intended only for logging what _would_ be blocked were ConstrainedLanguage mode actually being enforced seriously.

The PowerShell VSCode shell integration guard currently aborts unless `LanguageMode == FullLanguage`. This logic predates the existence of audit-only constrained language mode and was correct for its time. Presently however, it falsely blocks the shell integration from working in audit scenarios, even though PowerShell execution is fully functional.

This change updates the guard to allow shell integration when:

- `LanguageMode == FullLanguage`, or
- `LanguageMode == ConstrainedLanguage` **and** system lockdown policy is `Audit`

Enforced constrained language mode continues to be blocked.

## Why this change is safe
- Audit CLM is explicitly non-restrictive by design
- Enforced CLM behavior remains unchanged
- Backward compatible with all older PowerShell versions
- No elevation or system configuration changes required
- Uses PowerShell’s official `SystemPolicy.GetSystemLockdownPolicy()` API

## How to test
**Audit CLM scenario**
1. On Windows 11, enable a WDAC / App Control **audit-only** policy
2. Install PowerShell 7.4+. If step 1 went correctly, pwsh should be in ConstrainedLanguage Audit Mode
3. Open VS Code
4. Open the integrated terminal using `pwsh`
5. Verify shell integration initializes correctly (prompt tracking, command boundaries)

**Enforced CLM scenario**
1. Enable an enforced WDAC / App Control policy
2. Open the integrated terminal using `pwsh`
3. Verify shell integration does **not** initialize

**Baseline**
- Windows PowerShell 5.1 (FullLanguage) continues to work unchanged because it is too old to understand Audit Mode.

## Notes
The original language mode guard was introduced in #158548 and was correct at the time. This change adapts that logic to newer PowerShell semantics introduced in 7.4 without weakening security guarantees.

### DIFF
--- a/src/vs/workbench/contrib/terminal/common/scripts/shellIntegration.ps1
+++ b/src/vs/workbench/contrib/terminal/common/scripts/shellIntegration.ps1
@@ -10,7 +10,28 @@ if ((Test-Path variable:global:__VSCodeState) -and $null -ne $Global:__VSCodeSta
 
 # Disable shell integration when the language mode is restricted
 if ($ExecutionContext.SessionState.LanguageMode -ne "FullLanguage") {
-	return;
+	$LanguageMode = $ExecutionContext.SessionState.LanguageMode
+	if ($LanguageMode -ne "ConstrainedLanguage") {
+		return;
+	}
+	try {
+		# PowerShell 7.4+ supports audit-only constrained language mode, which is safe to run.
+		# Older PowerShell versions do not have SystemPolicy/GetSystemLockdownPolicy, so block.
+		$SystemPolicyType = [Type]::GetType("System.Management.Automation.Security.SystemPolicy, System.Management.Automation", $false)
+		if (-not $SystemPolicyType) {
+			return;
+		}
+		$SystemPolicyMethod = $SystemPolicyType.GetMethod("GetSystemLockdownPolicy")
+		if (-not $SystemPolicyMethod) {
+			return;
+		}
+		$Lockdown = $SystemPolicyMethod.Invoke($null, $null)
+		if ($Lockdown -and $Lockdown.ToString() -ne "Audit") {
+			return;
+		}
+	} catch {
+		return;
+	}
 }
 
 $Global:__VSCodeState = @{
--- a/src/vs/workbench/contrib/terminal/common/scripts/shellIntegration.ps1
+++ b/src/vs/workbench/contrib/terminal/common/scripts/shellIntegration.ps1
@@ -17,7 +17,7 @@ if ($ExecutionContext.SessionState.LanguageMode -ne "FullLanguage") {
 	try {
 		# Allow ConstrainedLanguage only when Windows lockdown policy is Audit (PS 7.4+).
 		$Lockdown = [System.Management.Automation.Security.SystemPolicy]::GetSystemLockdownPolicy()
-		if ($Lockdown -and $Lockdown -ne [System.Management.Automation.Security.SystemEnforcementMode]::Audit) {
+		if ($Lockdown -and $Lockdown -ne "Audit") {
 			return;
 		}
 	} catch {
--- a/src/vs/workbench/contrib/terminal/common/scripts/shellIntegration.ps1
+++ b/src/vs/workbench/contrib/terminal/common/scripts/shellIntegration.ps1
@@ -11,22 +11,13 @@ if ((Test-Path variable:global:__VSCodeState) -and $null -ne $Global:__VSCodeSta
 # Disable shell integration when the language mode is restricted
 if ($ExecutionContext.SessionState.LanguageMode -ne "FullLanguage") {
 	$LanguageMode = $ExecutionContext.SessionState.LanguageMode
-	if ($LanguageMode -ne "ConstrainedLanguage") {
+	if (-not $IsWindows -or $LanguageMode -ne "ConstrainedLanguage") {
 		return;
 	}
 	try {
-		# PowerShell 7.4+ supports audit-only constrained language mode, which is safe to run.
-		# Older PowerShell versions do not have SystemPolicy/GetSystemLockdownPolicy, so block.
-		$SystemPolicyType = [Type]::GetType("System.Management.Automation.Security.SystemPolicy, System.Management.Automation", $false)
-		if (-not $SystemPolicyType) {
-			return;
-		}
-		$SystemPolicyMethod = $SystemPolicyType.GetMethod("GetSystemLockdownPolicy")
-		if (-not $SystemPolicyMethod) {
-			return;
-		}
-		$Lockdown = $SystemPolicyMethod.Invoke($null, $null)
-		if ($Lockdown -and $Lockdown.ToString() -ne "Audit") {
+		# Allow ConstrainedLanguage only when Windows lockdown policy is Audit (PS 7.4+).
+		$Lockdown = [System.Management.Automation.Security.SystemPolicy]::GetSystemLockdownPolicy()
+		if ($Lockdown -and $Lockdown -ne [System.Management.Automation.Security.SystemEnforcementMode]::Audit) {
 			return;
 		}
 	} catch {


Tyriar (commented):
@andyleejordan could you review this when you get a chance? 🙏

src/vs/workbench/contrib/terminal/common/scripts/shellIntegration.ps1
```diff
@@ -10,7 +10,28 @@ if ((Test-Path variable:global:__VSCodeState) -and $null -ne $Global:__VSCodeSta
...
+		if (-not $SystemPolicyMethod) {
+			return;
+		}
+		$Lockdown = $SystemPolicyMethod.Invoke($null, $null)
+		if ($Lockdown -and $Lockdown.ToString() -ne "Audit") {
+			return;
+		}
+	} catch {
+		return;
+	}
```
daxian-dbw, line 25: System lockdown policy is a Windows only feature, so it'd be more efficient to skip the handling of `Audit` mode on non-Windows machines.

- Is this file only for PowerShell 7+? I noticed it uses `$IsWindows` below at line 68. If so, we can use `$IsWindows` here to skip processing for Linux and macOS.
- The `SystemPolicy.GetSystemLockdownPolicy` is a [public API](https://learn.microsoft.com/en-us/dotnet/api/system.management.automation.security.systempolicy.getsystemlockdownpolicy?view=powershellsdk-7.4.0) on Windows since v5.1. The `Audit` feature just added a new Enum member to `SystemEnforcementMode` (`SystemEnforcementMode.Audit`). So, you can directly call it on Windows instead of using reflection.

daxian-dbw (changes requested):

src/vs/workbench/contrib/terminal/common/scripts/shellIntegration.ps1
skbtwiz: Thanks for the feedback! Made the suggested changes and simplified the logic. Please review at your convenience.

src/vs/workbench/contrib/terminal/common/scripts/shellIntegration.ps1
```diff
@@ -10,7 +10,19 @@ if ((Test-Path variable:global:__VSCodeState) -and $null -ne $Global:__VSCodeSta
 
 # Disable shell integration when the language mode is restricted
 if ($ExecutionContext.SessionState.LanguageMode -ne "FullLanguage") {
-	return;
+	$LanguageMode = $ExecutionContext.SessionState.LanguageMode
```
daxian-dbw, line 13: nitpick: How aboue moving this line outside of the if block starting at line 12, and change line 12 to `if ($LanguageMode -ne "FullLanguage")`?

src/vs/workbench/contrib/terminal/common/scripts/shellIntegration.ps1
```diff
@@ -10,7 +10,19 @@ if ((Test-Path variable:global:__VSCodeState) -and $null -ne $Global:__VSCodeSta
...
+	if (-not $IsWindows -or $LanguageMode -ne "ConstrainedLanguage") {
+		return;
+	}
+	try {
+		# Allow ConstrainedLanguage only when Windows lockdown policy is Audit (PS 7.4+).
+		$Lockdown = [System.Management.Automation.Security.SystemPolicy]::GetSystemLockdownPolicy()
+		if ($Lockdown -and $Lockdown -ne [System.Management.Automation.Security.SystemEnforcementMode]::Audit) {
```
daxian-dbw:
```suggestion
		if ($Lockdown -and $Lockdown -ne "Audit") {
```
Let's use the string "Audit" here. The enum member "Audit" is not in versions prior to 7.4, so in case user is using a old version, `[System.Management.Automation.Security.SystemEnforcementMode]::Audit` would fail.

daxian-dbw (changes requested):

src/vs/workbench/contrib/terminal/common/scripts/shellIntegration.ps1
skbtwiz: I planned it initially because it clearly improves efficiency, but then I realized it would make `$LanguageMode` a global scope variable in the file. This was concerning because I did not want any of my changes to affect anything else in the file by accident, and wanted the patch to be entirely contained inside the scope of the `if` block. Open to hearing your thoughts, and will make this change if you believe that it is safe to do so.

src/vs/workbench/contrib/terminal/common/scripts/shellIntegration.ps1
skbtwiz: Thank you, this suggested change is applied.

src/vs/workbench/contrib/terminal/common/scripts/shellIntegration.ps1
daxian-dbw: That's fine. This is a minor comment that can be ignored. But be noted that the if statement doesn't create a new scope, so in case it's running in constrained audit mode, the variable will be visible in the script block.

daxian-dbw (approved):
LGTM!

skbtwiz:
Hello @anthonykim1 ,

Please review this PR if and when you get the opportunity. Since this is my first major contribution to a large open source project, I apologise in advance if this message is out of place. This PR would significantly improve QoL on VS Code as well as its fork IDEs as currently the IDE's Rich Terminal is broken on all of them if system is on ConstrainedLanguage Audit mode, which is true for majority of the new dev machines we have received at our company that are Copilot+/Recall certified.

Once again, no urgency implied, just that my team is eager to see these fixes on their machines. Currently we are manually applying these fixes after each update.

Issue #283215, opened 2025-12-12
create tool calls, markdown, and thinking out of sync


justschen:
<img width="450" height="623" alt="Image" src="https://github.com/user-attachments/assets/bf4b28d3-f40e-4613-ae49-f8b714a5ae83" />

as you can see from the image, the thinking keeps going despite a markdown part being hit, leading to the create tool calls to get added inot the thinking part

repro steps:
1. most common in claude 4.5 opus
2. most reliable repro prompt: create 5 files, each with a random word in each. after each tool call, explain what you did and the contents of the file.
3. it needs to have thinking -> markdown -> thinking -> markdown

PR #283219, opened 2025-12-12
fix tool call and markdown out of sync due to thinking

justschen:
release fix for https://github.com/microsoft/vscode/issues/283215


src/vs/workbench/contrib/chat/browser/chatListRenderer.ts
```diff
@@ -1650,6 +1650,7 @@ export class ChatListItemRenderer extends Disposable implements ITreeRenderer<Ch
 	}
 
 	private renderMarkdown(markdown: IChatMarkdownContent, templateData: IChatListItemTemplate, context: IChatContentPartRenderContext): IChatContentPart {
+		this.finalizeCurrentThinkingPart(context, templateData);
```
justschen, line 1653: here we are basically just ensuring that whenever _any_ markdown chat content part (text output from the model, non-tool calls, not thinking) is rendered, we know when to end the collapsible thinking part

lramos15 (approved):

PR #283223, opened 2025-12-12
Remove chat view pane border

### LINKED ISSUE #281928: Agent Sessions: Inconsistent scroll shadow/border in view header
If you scroll down in both the agents sessions view and the chat, you'll see that the chat has a shadow and a border and the agent sessions has only a shadow. I think the border should be removed from the chat view.

Image

dmitrivMS:
Fixes #281928


justschen (dismissed):

bpasero (changes requested):
Not sure I fully understand the suggested fix, but the issue is about hiding the border only when there is a scroll shadow, not always 🤔

bpasero (approved):
Thanks! Let's see how this flies next week in insiders in the team.

PR #283323, opened 2025-12-13
Update Copilot-related placeholders and picker items order

### LINKED ISSUE #253827: Select a location for custom mode prompt polish
Image

* Placeholder text should not have "...". It should be "Select a location to create the custom mode"
* User Data Dir should be first (this is the primary use case, users just create mode once and share across workspaces)

fyi @digitarald

dmitrivMS:
Fixes #253827

### DIFF
--- a/src/vs/workbench/contrib/chat/common/promptSyntax/service/promptsServiceImpl.ts
+++ b/src/vs/workbench/contrib/chat/common/promptSyntax/service/promptsServiceImpl.ts
@@ -282,6 +282,10 @@ export class PromptsService extends Disposable implements IPromptsService {
 	public getSourceFolders(type: PromptsType): readonly IPromptPath[] {
 		const result: IPromptPath[] = [];
 
+		// Add User Data Dir first (primary use case)
+		const userHome = this.userDataService.currentProfile.promptsHome;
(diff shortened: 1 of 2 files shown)


src/vs/workbench/contrib/chat/common/promptSyntax/service/promptsServiceImpl.ts
```diff
@@ -282,6 +282,10 @@ export class PromptsService extends Disposable implements IPromptsService {
 	public getSourceFolders(type: PromptsType): readonly IPromptPath[] {
 		const result: IPromptPath[] = [];
 
+		// Add User Data Dir first (primary use case)
```
aeschli: Actually I don't think user folder is the primary use case for instruction/prompt and agent files.
Let's keep the order as is now.

aeschli (changes requested):

src/vs/workbench/contrib/chat/common/promptSyntax/service/promptsServiceImpl.ts
dmitrivMS: reverted

aeschli (approved):

PR #283401, opened 2025-12-14
Fix input box jumping when sending a request

### LINKED ISSUE #281469: chat input jumps up after accepting input
https://github.com/user-attachments/assets/b4cc1002-bd37-476d-91c8-96544222b986

after submitting, the input jumps down a bit, while the empty state text jumps up, causing some flickering

roblourens:
Changing the display from flex to block forced a reflow/repaint. So keep it flex all the time.
Fix #281469


bpasero (changes requested):
Can we preserve the notion of `.has-sessions-control` and try to fix the CSS by moving the relevant pieces up into `.chat-viewpane {` rule? The semantic should be preserved otherwise future changes will be hard to understand. `.has-sessions-control` is present if the sessions are showing either stacked or side by side.

roblourens:
Can you finish this up @bpasero? I think this is closer to what you mean but I'm not sure which of these rules actually needs to change when `.has-sessions-control`.

bpasero (changes requested):
If we want to go down this path we also need to include the `.interactive-session` changes otherwise the contents over flow:

Image

I wonder: do we really want to do this as a recovery fix 🤔

bpasero (approved):

lszomoru (approved):

PR #283450, opened 2025-12-14
fix: memory leak in MainThreadWorkspace

SimonSiefke:
Fixes a memory leak in MainThreadWorkspace. 

### Details

When a provider is registered, it is both added to a Map called `registeredEditSessionProviders` as well as to `this._toDispose`.

```ts
this.registeredEditSessionProviders.set(handle, disposable);
this._toDispose.add(disposable);
```

However when a provider is unregistered, it is removed from the Map `registeredEditSessionProviders`, but not from `this._toDispose`

```ts
const disposable = this.registeredEditSessionProviders.get(handle);
disposable?.dispose();
this.registeredEditSessionProviders.delete(handle);
```

This causes the `this._toDispose` array to grow over time. 

### Change

The change is using a `DisposableMap`, ensuring `this._toDispose` doesn't grow when registering and unregistering providers. 

```ts
private registeredCanonicalUriProviders = this._register(new DisposableMap<number, IDisposable>());
```

### Before 

When restarting extensions 37 times, the number functions in `MainThreadWorkspace.$registerCanonicalUriProvider` seems to grow by 3 each time. Entries 10/11:

extensions restart

### After

When restarting extensions 37 times, the number functions in `MainThreadWorkspace.$registerCanonicalUriProvider` stays constant:

extensions restart


roblourens (approved):

roblourens (commented):
Thanks!

bhavyaus (approved):

PR #283466, opened 2025-12-14
fix: memory leak in terminal find widget

SimonSiefke:
Fixes a memory leak in terminal find widget. 

When calling find multiple times, it seems to register a selection change event listener each time:

```ts
this._register(Event.once(xterm.onDidChangeSelection)(() => xterm.clearActiveSearchDecoration()));
```

Using a `MutableDisposable` ensures there can be at most one selection change listener. 

---

Perhaps another solution might be to register the listener in the constructor once. Not entirely sure if that would be the same overall logic, but if possible it would be simpler since the `MutableDisposable` could be completly removed.

### Before

When searching something with terminal find 97 times, the number of functions in `TerminalFindWidget._focusPreviousWithEvent` seems to grow by 1 each time:

terminal search

### After

No more leaks are detected.


meganrogge (approved):
Thank you!

dmitrivMS (approved):

PR #283510, opened 2025-12-15
move attachment x and + to the left

justschen:
partially fix for https://github.com/microsoft/vscode-internalbacklog/issues/6482

Screenshot 2025-12-15 at 1 43 57 PM

### DIFF
--- a/src/vs/workbench/contrib/chat/browser/media/chat.css
+++ b/src/vs/workbench/contrib/chat/browser/media/chat.css
@@ -1582,10 +1582,9 @@ have to be updated for changes to the rules above, or to support more deeply nes
 .action-item.chat-attachment-button .action-label,
 .interactive-session .chat-attached-context .chat-attached-context-attachment {
 	display: flex;
-	gap: 2px;
 	overflow: hidden;
 	font-size: 11px;
-	padding: 0 4px;
+	padding: 0 4px 0 0;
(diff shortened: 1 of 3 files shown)


src/vs/workbench/contrib/chat/browser/media/chat.css
```diff
@@ -1657,16 +1656,16 @@ have to be updated for changes to the rules above, or to support more deeply nes
...
 	height: calc(100% + 4px);
 	outline-offset: -4px;
+	font-size: 12px;
 }
 
 .interactive-session .chat-attached-context .chat-attached-context-attachment .monaco-button.codicon.codicon-plus {
-	padding: 0 5px 0 3px;
-	font-size: 12px;
+	padding-left: 4px;
+	font-size: 11px;
```
justschen, line 1668: the `+` codicon for some reason looks a tad bit bigger than the `x` codicon even at the same fonts (originally set to 16, but i think 12 works better here) @mrleemurray

dmitrivMS (dismissed):

dmitrivMS (approved):

src/vs/workbench/contrib/chat/browser/media/chat.css
mrleemurray: @justschen we recently added an `add-small` codicon that is optimized for 11px. Here it is (via a quick hack):
image

PR #283571, opened 2025-12-15
WORKAROUND - paste fix - using readText from navigator API when triggerPaste fails

aiday-mar:
using readText from navigator API when triggerPaste fails
counterpart to copy issue


src/vs/editor/browser/controller/editContext/textArea/textAreaEditContextInput.ts
```diff
@@ -381,27 +376,14 @@ export class TextAreaInput extends Disposable {
...
 			}
-
-			// try the in-memory store
-			metadata = metadata || InMemoryClipboardMetadataManager.INSTANCE.get(text);
-
-			this._logService.trace(`TextAreaInput#onPaste (before onPaste)`);
-			this._onPaste.fire({
-				text: text,
-				metadata: metadata
-			});
```
deepak1556, line 404: should we keep "this._logService.trace(`TextAreaInput#onPaste (before onPaste)`);" ?

aiday-mar: Yep added back!

deepak1556 (approved):

PR #283575, opened 2025-12-15
Enable settings completion when the query is empty

### LINKED ISSUE #276462: Settings editor: suggest completions even when input is empty
When you open suggestions in the settings editor without text, you get nothing:

Image

But then when you type `@` you get a lot:

Image

Why don't we make this discoverable even when nothing is typed in the input?

dmitrivMS:
Fixes #276462


rzhao271 (approved):
Thanks!

PR #283584, opened 2025-12-15
fix: correct the separator for uvx packages #283572

### LINKED ISSUE #283572: UVX MCPs servers do not start when installed from extensions gallery
Does this issue occur when all extensions are disabled?: NA

- VS Code Version: Version: 1.108.0-insider (Universal) Commit: 5dab7809de7d46dffea7180dddcde7100f2ab4d6
- OS Version: Mac 14.7.6 

Issue:

There appears to be an issue with uvx MCP servers in VScode (stable and insiders), when they get installed the resulting config is wrong.

I installed the PagerDuty MCP from the GitHub/Microsoft registry and the server cannot start with the default config

It appears to be that when installed via MCP extensions, VScode writes the starting args as

```
"args": [
    "pagerduty-mcp==0.2.1",
    "--enable-write-tools"
],
```
with a Python style == where as the MCP stuff appears to need an @ instead.
i.e.
```
"args": [
    "pagerduty-mcp@0.2.1",
    "--enable-write-tools"
],
```
With the == config the server fails to start with:

```
2025-12-15 12:04:46.149 [warning] [server stderr] error: Not a valid package or extra name: "pagerduty-mcp==0.2.1". Names must start and end with a letter or digit and may only contain -, _, ., and alphanumeric characters.
2025-12-15 12:04:46.151 [info] Connection state: Error Process exited with code 2
```

Steps to Reproduce:

1.  Open extensions pannel and search `@mcp pagerduty`
2. Install the Pager Duty MCP server
3. Open mcp.json and see that the server does not start
4. amend the arg to `"pagerduty-mcp@0.2.1"` (replacing the `==` with `@`)
5. save and restart the server
6. it now loads 

With `==`
Image

With `@`
Image

robertsa1-els:
fixes #283572 

uvx uses an `@` for separating package and version.  This is different to pip which uses `==`
https://docs.astral.sh/uv/concepts/tools/#tool-versions

Currently the code that installs MCP servers into VSCode from a registry uses `==` as a separator which causes MCP servers to fail to start.

This PR updates that separator to the correct `@` as currently it looks like UVX is the only supported python runtime
https://github.com/microsoft/vscode/blob/main/src/vs/platform/mcp/common/mcpManagementService.ts#L171

### DIFF
--- a/src/vs/platform/mcp/common/mcpManagementService.ts
+++ b/src/vs/platform/mcp/common/mcpManagementService.ts
@@ -129,7 +129,7 @@ export abstract class AbstractCommonMcpManagementService extends Disposable impl
 				args.push(serverPackage.version ? `${serverPackage.identifier}@${serverPackage.version}` : serverPackage.identifier);
 				break;
 			case RegistryType.PYTHON:
-				args.push(serverPackage.version ? `${serverPackage.identifier}==${serverPackage.version}` : serverPackage.identifier);
+				args.push(serverPackage.version ? `${serverPackage.identifier}@${serverPackage.version}` : serverPackage.identifier);
 				break;
 			case RegistryType.DOCKER:
 				args.push(serverPackage.version ? `${serverPackage.identifier}:${serverPackage.version}` : serverPackage.identifier);


robertsa1-els:
Closing in favour of https://github.com/microsoft/vscode/pull/289056

PR #283635, opened 2025-12-15
debug: Fix UI freezing on "continue" with high number of threads

jcpetruzza:
This fixes a performance regression accidentally introduced in #265755.

Before that change, if the `continued` event had `allThreadsContinued` set, then we'd call `this.model.clearThreads()` passing `undefined` as `threadId`, which would update all threads in one pass. After that change, a call to `this.model.clearThreads()` would be done for each thread.

Because each call to `this.model.clearThreads()` ends up calling `this._onDidChangeCallStack.fire()`, we end up with quadratic overhead as some of the handlers traverse all threads in the call-stack box. This would lead to minute long freezes when using the Erlang debugger, to debug a system with ~20K Erlang processes.

The solution is to go back to the previous implementation, where `clearThreads()` is ever called with one thread-id or with `undefined`, but first wait on `this.fetchThreads()` if needed

### DIFF
--- a/src/vs/workbench/contrib/debug/common/debugModel.ts
+++ b/src/vs/workbench/contrib/debug/common/debugModel.ts
@@ -1458,6 +1458,9 @@ export class DebugModel extends Disposable implements IDebugModel {
 	private breakpointsActivated = true;
 	private readonly _onDidChangeBreakpoints = this._register(new Emitter<IBreakpointsChangeEvent | undefined>());
 	private readonly _onDidChangeCallStack = this._register(new Emitter<void>());
+	private _onDidChangeCallStackFire = new RunOnceScheduler(() => {
+		this._onDidChangeCallStack.fire(undefined);
+	}, 100);
 	private readonly _onDidChangeWatchExpressions = this._register(new Emitter<IExpression | undefined>());
 	private readonly _onDidChangeWatchExpressionValue = this._register(new Emitter<IExpression | undefined>());
 	private readonly _breakpointModes = new Map<string, IBreakpointModeInternal>();
@@ -1583,7 +1586,7 @@ export class DebugModel extends Disposable implements IDebugModel {
 
 		if (session) {
 			session.clearThreads(removeThreads, reference);
-			this._onDidChangeCallStack.fire(undefined);
+			this._onDidChangeCallStackFire.schedule();
 		}
 	}
 
--- a/src/vs/workbench/contrib/debug/common/debugModel.ts
+++ b/src/vs/workbench/contrib/debug/common/debugModel.ts
@@ -1578,13 +1578,24 @@ export class DebugModel extends Disposable implements IDebugModel {
 
 	clearThreads(id: string, removeThreads: boolean, reference: number | undefined = undefined): void {
 		const session = this.sessions.find(p => p.getId() === id);
-		this.schedulers.forEach(entry => {
-			entry.scheduler.dispose();
-			entry.completeDeferred.complete();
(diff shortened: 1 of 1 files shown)


src/vs/workbench/contrib/debug/browser/debugSession.ts
```diff
@@ -1132,35 +1132,34 @@ export class DebugSession implements IDebugSession {
...
-				affectedThreads = this.fetchThreads().then(() => this.threadIds);
+				needFreshFetchThreads = !this.threadIds.includes(event.body.threadId);
 			} else {
-				affectedThreads = this.threadIds;
+				needFreshFetchThreads = this.fetchThreadsScheduler.value.isScheduled();
+			}
+
+			if (needFreshFetchThreads) {
+				this.fetchThreadsScheduler.rawValue?.cancel();
+				await this.fetchThreads();
```
connor4312: We don't want to await outside the statusQueue since that opens up to race conditions (e.g. DA continues threads and pauses again before this fetchThreads finishes and we end in a torn state)

I think we probably just want to debounce the `_onDidChangeCallStack.fire` call in `DebugModel.clearThreads`. You can see how we do that with the `fetchThreadsScheduler` already today in the debug session, we'd do the same thing there.

Btw reviewing the code I notice a bug in `DebugModel.clearThreads` -- we don't check the `reference` (thread ID) when clearing the schedulers. If a thread ID is set, we should only clear its scheduler, not all schedulers. If you're up for fixing a low hanging bug there, feel free to do so 🙂

jcpetruzza:
@connor4312 thanks for the review! I've updated the branch with a new version that instead debounces the calls, and did a drive-by fix of the bug you spotted in `clearThreads()`

src/vs/workbench/contrib/debug/common/debugModel.ts
```diff
@@ -1575,15 +1578,26 @@ export class DebugModel extends Disposable implements IDebugModel {
...
+				if (entry !== undefined) {
+					entry.scheduler.dispose();
+					entry.completeDeferred.complete();
+					this.schedulers.delete(threadId);
+				}
+			}
+
 			session.clearThreads(removeThreads, reference);
-			this._onDidChangeCallStack.fire(undefined);
+			this._onDidChangeCallStackFire.schedule();
```
connor4312: Small thing: we should schedule only if not already scheduled. Otherwise a DA with a ton of threads and very frequent events could result in the call stack never updating.

src/vs/workbench/contrib/debug/common/debugModel.ts
```diff
@@ -1458,6 +1458,9 @@ export class DebugModel extends Disposable implements IDebugModel {
 	private breakpointsActivated = true;
 	private readonly _onDidChangeBreakpoints = this._register(new Emitter<IBreakpointsChangeEvent | undefined>());
 	private readonly _onDidChangeCallStack = this._register(new Emitter<void>());
+	private _onDidChangeCallStackFire = new RunOnceScheduler(() => {
+		this._onDidChangeCallStack.fire(undefined);
+	}, 100);
```
connor4312:
```suggestion
	private _onDidChangeCallStackFire = this._register(new RunOnceScheduler(() => {
		this._onDidChangeCallStack.fire(undefined);
	}, 100));
```

jcpetruzza:
Added a fixup for both issues

connor4312 (approved):

roblourens (approved):

PR #283663, opened 2025-12-15
Add 'Use Excludes and Ignore Files' toggle to quick access

### LINKED ISSUE #15604: Simple way to temporarily turn off exclusions for QuickOpen files
Using a transpiled language I sometimes find the need to look at the transpiled code which typically is excluded from QuickOpen. Having a simple way to turn exclusions off would simplify this. Some options come to mind:
- Have a 'Turn Off Exclusions' entry at the end of the results list.
- Have an additional QuickOpen prefix that runs QuickOpen with exclusions turned off.
- Have an option toggle next to the QuickOpen entry field to turn exclusions off.

dmitrivMS:
Fixes #15604

Added a toggle to file and text search quick access to disable exclusions/ignore files. The toggle uses same icon as in Search View and is bound to Ctrl+Dot shortcut.

### DIFF
--- a/src/vs/base/browser/ui/toggle/toggle.ts
+++ b/src/vs/base/browser/ui/toggle/toggle.ts
@@ -259,6 +259,9 @@ export class Toggle extends Widget {
 	}
 
 	toggle(): void {
+		if (!this.enabled) {
+			return;
+		}
 		this.checked = !this._checked;
 		this._onChange.fire(false);
 	}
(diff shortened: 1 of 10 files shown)


dmitrivMS:
To be fair, I think Copilot pulled most changes for me from https://github.com/microsoft/vscode/pull/281303 which is essentially the same. If we decide to take it, probably best to update that PR and let it go in.

bpasero:
I would like to review and test this change but will probably not get to it until next year.

dmitrivMS:
@bpasero Closing this one, will continue in https://github.com/microsoft/vscode/pull/281303 to which I've added you as a reviewer.

PR #283901, opened 2025-12-16
cancel chat progress sound when new chat is created, alert that session will continue in the background

### LINKED ISSUE #281595: Stop playing chat progress sound when new chat session is created
1. Start a chat session

meganrogge:
fixes https://github.com/microsoft/vscode/issues/281595

cc @roblourens

### DIFF
--- a/src/vs/workbench/contrib/chat/browser/actions/chatNewActions.ts
+++ b/src/vs/workbench/contrib/chat/browser/actions/chatNewActions.ts
@@ -18,7 +18,7 @@ import { ChatContextKeys } from '../../common/chatContextKeys.js';
 import { IChatEditingSession } from '../../common/chatEditingService.js';
 import { IChatService } from '../../common/chatService.js';
 import { ChatAgentLocation, ChatModeKind } from '../../common/constants.js';
-import { ChatViewId, IChatWidgetService } from '../chat.js';
+import { ChatViewId, IChatAccessibilityService, IChatWidgetService } from '../chat.js';
 import { EditingSessionAction, getEditingSessionContext } from '../chatEditing/chatEditingActions.js';
 import { ChatEditorInput } from '../chatEditorInput.js';
 import { ACTION_ID_NEW_CHAT, ACTION_ID_NEW_EDIT_SESSION, CHAT_CATEGORY, handleCurrentEditingSession } from './chatActions.js';
@@ -116,6 +116,8 @@ export function registerNewChatActions() {
 		}
 
 		async run(accessor: ServicesAccessor, ...args: unknown[]) {
+			const chatAccessibilityService = accessor.get(IChatAccessibilityService);
+			chatAccessibilityService.disposeRequest();
 			const executeCommandContext = args[0] as INewEditSessionActionContext | undefined;
 
 			// Context from toolbar or lastFocusedWidget
--- a/src/vs/workbench/contrib/chat/browser/actions/chatNewActions.ts
+++ b/src/vs/workbench/contrib/chat/browser/actions/chatNewActions.ts
@@ -23,6 +23,7 @@ import { EditingSessionAction, getEditingSessionContext } from '../chatEditing/c
 import { ChatEditorInput } from '../chatEditorInput.js';
 import { ACTION_ID_NEW_CHAT, ACTION_ID_NEW_EDIT_SESSION, CHAT_CATEGORY, handleCurrentEditingSession } from './chatActions.js';
 import { clearChatEditor } from './chatClear.js';
+import { alert } from '../../../../../base/browser/ui/aria/aria.js';
 
 export interface INewEditSessionActionContext {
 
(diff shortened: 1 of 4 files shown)


src/vs/workbench/contrib/chat/browser/actions/chatNewActions.ts
```diff
@@ -116,6 +117,9 @@ export function registerNewChatActions() {
 		}
 
 		async run(accessor: ServicesAccessor, ...args: unknown[]) {
+			const chatAccessibilityService = accessor.get(IChatAccessibilityService);
+			chatAccessibilityService.disposeRequest();
+			alert('Session will continue in the background');
```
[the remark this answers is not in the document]
dmitrivMS: Good point?

src/vs/workbench/contrib/chat/browser/actions/chatNewActions.ts
```diff
@@ -111,6 +117,9 @@ export function registerNewChatActions() {
...
```
roblourens: There are a lot of ways this can happen, so I think this isn't the right place. But also when you have multiple sessions, how does `chatAccessibilityService.disposeRequest` work, how do you know that it's operating on the right one?

roblourens: I'm not sure what the right place is because we haven't had to react to this specific case yet. Maybe I can add an event to `IChatWidgetService` for you?

roblourens: But on second thought, probably the call to the accessibility service could return a disposable, and we could find the right thing to attach that to so it would just get automatically disposed in this case? Where would that have to happen?

src/vs/workbench/contrib/chat/browser/chatWidget.ts
```diff
@@ -2249,7 +2249,7 @@ export class ChatWidget extends Disposable implements IChatWidget {
 		this.scrollLock = this.isLockedToCodingAgent || !!checkModeOption(this.input.currentModeKind, this.viewOptions.autoScroll);
 
 		const editorValue = this.getInput();
-		const requestId = this.chatAccessibilityService.acceptRequest();
+		const requestId = this.chatAccessibilityService.acceptRequest({ playProgressSignal: this.visible && !this.isLockedToCodingAgent });
```
roblourens, line 2252: Why the `this.isLockedToCodingAgent` check?

src/vs/workbench/contrib/chat/browser/actions/chatNewActions.ts
roblourens: > But on second thought, probably the call to the accessibility service could return a disposable, and we could find the right thing to attach that to so it would just get automatically disposed in this case

I think that's probably the right way to do it, that solves both problems

src/vs/workbench/contrib/chat/browser/chatWidget.ts
meganrogge: I was trying to understand what that is - no jsdoc... maybe it's wrong to have it here, I thought it meant is stored/backgrounded

src/vs/workbench/contrib/chat/browser/actions/chatNewActions.ts
meganrogge: An event on `IChatWidgetService` would be great. 

The reason I thought this would be fine (and seems to work based on my testing in the panel, which is the primary use case), is pressing the + button always means we want to stop whatever progress sound is occurring. I guess there is some weird edge case where you have a session in the editor, move to the panel, where you also have a session in progress, and click the +?

meganrogge: Since this is mostly for screen reader users, who based on my experience, do not use editor sessions, this is probably fine. Open to other ideas though.

src/vs/workbench/contrib/chat/browser/chatWidget.ts
roblourens: It means it can only be used with a particular agent (so background/cloud)

src/vs/workbench/contrib/chat/browser/actions/chatNewActions.ts
roblourens: Well there are several ways to open new chat sessions besides the + button. For example, using the agent sessions view, the history picker, DND, the other actions... I assume you want the same logic for all scenarios?

src/vs/workbench/contrib/chat/browser/actions/chatNewActions.ts
```diff
@@ -119,7 +119,7 @@ export function registerNewChatActions() {
 		async run(accessor: ServicesAccessor, ...args: unknown[]) {
 			const chatAccessibilityService = accessor.get(IChatAccessibilityService);
 			chatAccessibilityService.disposeRequest();
-			alert('Session will continue in the background');
+			alert(localize2('chat.session.continuesInBackground', "Session will continue in the background"));
```
roblourens, line 122: Shouldn't this only trigger if the session is actively running?

src/vs/workbench/contrib/chat/browser/actions/chatNewActions.ts
meganrogge: Oh, I see. Yes, we need something more generic.

src/vs/workbench/contrib/chat/browser/chatWidget.ts
meganrogge, line 2252:
```suggestion
		const requestId = this.chatAccessibilityService.acceptRequest({ playProgressSignal: this.visible });
```

PR #283942, opened 2025-12-17
cancel progress sound for backgrounded sessions

### LINKED ISSUE #281595: Stop playing chat progress sound when new chat session is created

meganrogge:
fixes https://github.com/microsoft/vscode/issues/281595

### DIFF
--- a/src/vs/workbench/contrib/chat/browser/chatAccessibilityService.ts
+++ b/src/vs/workbench/contrib/chat/browser/chatAccessibilityService.ts
@@ -9,6 +9,7 @@ import { alert, status } from '../../../../base/browser/ui/aria/aria.js';
 import { Event } from '../../../../base/common/event.js';
 import { MarkdownString } from '../../../../base/common/htmlContent.js';
 import { Disposable, DisposableMap, DisposableStore } from '../../../../base/common/lifecycle.js';
+import { URI } from '../../../../base/common/uri.js';
 import { localize } from '../../../../nls.js';
 import { AccessibilitySignal, IAccessibilitySignalService } from '../../../../platform/accessibilitySignal/browser/accessibilitySignalService.js';
 import { AccessibilityProgressSignalScheduler } from '../../../../platform/accessibilitySignal/browser/progressAccessibilitySignalScheduler.js';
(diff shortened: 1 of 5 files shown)


src/vs/workbench/contrib/chat/browser/chatAccessibilityService.ts
```diff
@@ -58,7 +71,15 @@ export class ChatAccessibilityService extends Disposable implements IChatAccessi
 		return this._requestId;
 	}
 
-	disposeRequest(requestId: number): void {
+	disposeRequest(requestId?: number): void {
+		if (!requestId) {
+			// Dispose last
+			const lastKey = Array.from(this._pendingSignalMap.keys()).pop();
```
meganrogge: there might be cases when this doesn't work? the issue is we don't have the `requestId` when we call `acceptRequest`... so we use this fake `requestId`. 

The alternative approach would be to move the map into `ChatWidget` and return a disposable from `acceptRequest`, disposing of that when the event is fired.

roblourens: I think the disconnect is that you start the signal by the chat widget calling into the service's `acceptRequest`, but you end it (in the bg case) by the service listening to an event and making an independent decision. That means that there isn't just one party is in charge of the signal. A couple ways to resolve that would be replacing `acceptRequest` with something that the service can do on its own, like listening to an event somewhere that a request was sent. Or, having ChatWidget be in charge of cancelling the signal.

A slightly hacky solution might be using the chat session uri instead of a separate request id

meganrogge: Yeah in my other PR, I tried having `ChatWidget` cancel the signal

meganrogge: https://github.com/microsoft/vscode/pull/283949/ 

The URI could be a good idea for the key of a map

meganrogge: Just thought we'd want this accessibility code isolated from the chat stuff

roblourens (approved):

PR #283949, opened 2025-12-17
alternative approach for canceling chat progress sound

### LINKED ISSUE #281595: Stop playing chat progress sound when new chat session is created

meganrogge:
fixes  #281595

### DIFF
--- a/src/vs/workbench/contrib/chat/browser/chatWidget.ts
+++ b/src/vs/workbench/contrib/chat/browser/chatWidget.ts
@@ -350,6 +350,8 @@ export class ChatWidget extends Disposable implements IChatWidget {
 		return this._location.resolveData?.();
 	}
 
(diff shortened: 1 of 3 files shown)


src/vs/workbench/contrib/chat/browser/chatWidget.ts
```diff
@@ -693,6 +697,18 @@ export class ChatWidget extends Disposable implements IChatWidget {
 		}).filter(isDefined);
 
 		this._register(this.chatWidgetService.register(this));
+		this._register(this.chatWidgetService.onDidBackgroundSession(e => {
```
roblourens, line 700: If ChatWIdget is going to drive this, then we probably don't want that event because it's just driven by the ChatWidget firing its own event anyway.

PR #284096, opened 2025-12-17
ci: check for valid compilation artifacts in alpine stage

deepak1556:
Fixes https://github.com/microsoft/vscode-internalbacklog/issues/6469

Verification:
bad build https://dev.azure.com/monacotools/Monaco/_build/results?buildId=386032&view=results
good build https://monacotools.visualstudio.com/Monaco/_build/results?buildId=386229&view=results


deepak1556:
Should I add this to other stages as well ?

lszomoru:
@deepak1556, I think that this validation should be added at the end of the Compile stage instead of each individual stage that consumes the peoduced artifact.

lszomoru:
The task currently prints the following: `All required compilation folders exist and are not empty`. 
Would it make sense to print all the folders that were checked?

lszomoru (approved):

PR #284134, opened 2025-12-17
make `focusTerminal` async for consistency

### LINKED ISSUE #275562: Revisit ITerminalService.revealTerminal and related methods
There has been some confusion in October about how you're meant to show a terminal, for example https://github.com/microsoft/vscode/pull/275369, https://github.com/microsoft/vscode/issues/273437

The idea behind the current ones if I remember right was to have each one have very specific task, but that means it's not ergonomic at all doing the common thing.

Related helper:

https://github.com/microsoft/vscode/blob/34096fd36a95893177bcee7a6f8fbc5fc4d7493d/src/vs/workbench/contrib/terminal/browser/terminalActions.ts#L1714-L1722

We should look at all related methods and see if we can make it harder to cause bugs like these.

cc @meganrogge

meganrogge:
fixes #275562

- Focus terminal is now async and sets the active instance before focusing
- Callers await focusing so reveal/focus completes before running follow-up work, preventing races where a task/chat action proceeds while the terminal isn’t actually focused or even available.
- Terminal editor/panel focus is more robust: editor focus now no-ops on unknown instances and waits for any in-flight openEditor before calling focusWhenReady, while group focus marks the instance active before showing the panel.
- `focusActiveTerminal` now falls back to any known active instance and relies on the service to focus the correct location, avoiding blank focus attempts when the provided instance is missing.

### DIFF
--- a/src/vs/workbench/contrib/terminal/browser/terminalService.ts
+++ b/src/vs/workbench/contrib/terminal/browser/terminalService.ts
@@ -396,10 +396,17 @@ export class TerminalService extends Disposable implements ITerminalService {
 	}
 
 	async focusInstance(instance: ITerminalInstance): Promise<void> {
-		if (instance.target === TerminalLocation.Editor) {
-			return this._terminalEditorService.focusInstance(instance);
+		if (!instance) {
+			return;
 		}
-		return this._terminalGroupService.focusInstance(instance);
+		if (this._activeInstance !== instance) {
+			this.setActiveInstance(instance);
+		}
+		if (instance.target === TerminalLocation.Editor) {
+			await this._terminalEditorService.focusInstance(instance);
+			return;
+		}
+		await this._terminalGroupService.focusInstance(instance);
 	}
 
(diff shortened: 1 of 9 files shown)


src/vs/workbench/contrib/terminal/browser/terminalService.ts
```diff
@@ -396,10 +396,17 @@ export class TerminalService extends Disposable implements ITerminalService {
 	}
 
 	async focusInstance(instance: ITerminalInstance): Promise<void> {
+		if (!instance) {
+			return;
+		}
```
Tyriar, line 401: Don't need this check as TS should prevent at compile time?

Tyriar (approved):
👏 let's be sure to keep a look out for any incoming issues around focus

src/vs/workbench/contrib/terminal/browser/terminalService.ts
meganrogge: https://github.com/microsoft/vscode/pull/284317

Issue #282015, opened 2025-12-08
Move off and remove `InlineChatController1`


jrieken:
It's time to sunset the `InlineChatController1` and everything around it. The controller is quite buggy with its internal state uses a parallel world applying edits. It also features/drives the old UI with responses et al. 

Before cleaning up we need to migrate notebooks off it. This can be done with relative ease and without major prompt/backend changes. The only thing needed is that the extension knows when no edits has been made and hence when to upgrade to the Chat view

PR #284160, opened 2025-12-17
use v2 controller with old notebook prompt

amunger:
https://github.com/microsoft/vscode/issues/282015


src/vs/workbench/contrib/chat/common/model/chatModel.ts
```diff
@@ -666,16 +666,20 @@ export class Response extends AbstractResponse implements IDisposable {
...
-			const useOldApproachForInlineNotebook = progress.uri.scheme === Schemas.vscodeNotebookCell && !this._responseParts.find(part => part.kind === 'notebookEditGroup');
 			// merge edits for the same file no matter when they come in
-			const notebookUri = useOldApproachForInlineNotebook ? undefined : CellUri.parse(progress.uri)?.notebook;
+			const notebookUri = CellUri.parse(progress.uri)?.notebook;
 			const uri = notebookUri ?? progress.uri;
 			let found = false;
 			const groupKind = progress.kind === 'textEdit' && !notebookUri ? 'textEditGroup' : 'notebookEditGroup';
 			// eslint-disable-next-line @typescript-eslint/no-explicit-any
-			const edits: any = groupKind === 'textEditGroup' ? progress.edits : progress.edits.map(edit => TextEdit.isTextEdit(edit) ? { uri: progress.uri, edit } : edit);
+			const edits: any = // TextEdit[] | (ICellTextEditOperation | ICellEditOperation)[] =
```
amunger, line 675: this is a typing puzzle that I wasn't able to solve. I don't think it likes the mixture of cell / celltext edit operations. Co-pilot's solution was to triplicate this section of code and still involved an `as` cast.

I think we're forcing too many operation types through the same method here.

src/vs/workbench/contrib/chat/common/model/chatModel.ts
```diff
@@ -689,7 +693,7 @@ export class Response extends AbstractResponse implements IDisposable {
 				this._responseParts.push({
 					kind: groupKind,
 					uri,
-					edits: groupKind === 'textEditGroup' ? [edits] : edits,
```
amunger, line 692: according to the types (that are still being ignored), all edits in this branch should be put into an array.
Nothing was broken before because notebook edits from the panel/agent always initially send a notebook entry through with 0 edits (to establish that edit group?), so only appended to the found edit group above

src/vs/workbench/contrib/chat/common/model/chatModel.ts
```diff
@@ -666,16 +666,20 @@ export class Response extends AbstractResponse implements IDisposable {
 			}
 			this._updateRepr(quiet);
 		} else if (progress.kind === 'textEdit' || progress.kind === 'notebookEdit') {
-			// If the progress.uri is a cell Uri, its possible its part of the inline chat.
-			// Old approach of notebook inline chat would not start and end with notebook Uri, so we need to check for old approach.
-			const useOldApproachForInlineNotebook = progress.uri.scheme === Schemas.vscodeNotebookCell && !this._responseParts.find(part => part.kind === 'notebookEditGroup');
```
amunger, line 671: the v1 backend still doesn't start with a notebook URI, but we don't want to create a raw text edit without an associated notebook URI. This is why this path was previously broken when mixing inline v1/v2 with notebooks

src/vs/workbench/contrib/inlineChat/common/inlineChat.ts
```diff
@@ -89,12 +87,11 @@ export const CTX_INLINE_CHAT_REQUEST_IN_PROGRESS = new RawContextKey<boolean>('i
 export const CTX_INLINE_CHAT_RESPONSE_TYPE = new RawContextKey<InlineChatResponseType>('inlineChatResponseType', InlineChatResponseType.None, localize('inlineChatResponseTypes', "What type was the responses have been receieved, nothing yet, just messages, or messaged and local edits"));
 
 export const CTX_INLINE_CHAT_V1_ENABLED = ContextKeyExpr.or(
```
amunger, line 89: this will no longer be true, but removing it will be a large change that i'd like to keep independent

src/vs/workbench/contrib/chat/common/model/chatModel.ts
connor4312: Yea I think we should just split this out, or maybe have a helper function with some generic parts that lets us express things correctly. THis specific `any` cast was already a source of a bug in the past

connor4312 (commented):
edit part looks fine, I'll jet @jrieken review the inline chat parts

src/vs/workbench/contrib/inlineChat/common/inlineChat.ts
jrieken: I can take care of this post PR

jrieken (dismissed):

jrieken:
Also looking good for inline chat

DonJayamanne (approved):

PR #284165, opened 2025-12-18
Add glob pattern filter settings to snippets

### LINKED ISSUE #75955: Snippets: Scope by specific file or pattern
Global snippets already support being scoped by language and language file snippets are naturally scoped.  I would like to expand that "scoping" logic to specific files by matching the filename, for example.

A good use case for this, let's say I want to create snippets for travis.yml file. Right now what I do is to create the snippet scoped to YAML language. 

The problem is that it will appear for every YAML file which will pollute the autocomplete with not very useful results if you have snippets for multiple kinds of files that have the same extension. (I don't want to see "travis"  related snippets in a "gitlab-ci.yml" file for example) as they are not applicable.

I think this is most critical for configuration files where they have a very specific scope and that you can easily restrict by looking at the filename.

Going further, we could even expand this concept for other kinds of snippets. For example, I can have some project level snippets for building queries and I only want to enable them, in Repository classes and not in Controller classes. You could scope the files it applies using a regex pattern, for example.

I guess this shouldn't be very hard to do and it would vastly improve the suggestions quality of the snippets.

I see adding a new field similar to "scope" called "context" or something where you can specify a regex that will match to the filename.

dmitrivMS:
Fixes #75955

Added `include` and `exclude` settings to snippet schema to allow specifying glob patterns for files which the snippet should run on, with semantics similar to `files.associations`.


src/vs/workbench/contrib/snippets/browser/snippets.ts
```diff
@@ -27,7 +28,7 @@ export interface ISnippetsService {
 
 	updateUsageTimestamp(snippet: Snippet): void;
 
-	getSnippets(languageId: string | undefined, opt?: ISnippetGetOptions): Promise<Snippet[]>;
+	getSnippets(languageId: string | undefined, opt?: ISnippetGetOptions, resourceUri?: URI): Promise<Snippet[]>;
 
-	getSnippetsSync(languageId: string, opt?: ISnippetGetOptions): Snippet[];
+	getSnippetsSync(languageId: string, opt?: ISnippetGetOptions, resourceUri?: URI): Snippet[];
```
jrieken: nit: put `resourceUri` after `languageId` and bofore `opt`

src/vs/workbench/contrib/snippets/browser/snippetsFile.ts
```diff
@@ -138,6 +143,36 @@ export class Snippet {
...
 		return this._bodyInsights.value.usesSelectionVariable;
 	}
+
+	isFileIncluded(resourceUri: URI): boolean {
+		const filePath = resourceUri.fsPath;
+
+		if (this.exclude) {
+			if (!this._parsedExclude) {
+				const expression: IExpression = {};
+				for (const pattern of this.exclude.filter(p => Boolean(p))) {
```
jrieken: nit: `.filter(Boolean)`

src/vs/workbench/contrib/snippets/browser/snippetsFile.ts
```diff
@@ -198,9 +235,13 @@ export class SnippetFile {
 		}
 	}
 
-	private _scopeSelect(selector: string, bucket: Snippet[]): void {
+	private _scopeSelect(selector: string, bucket: Snippet[], resourceUri?: URI): void {
 		// for `my.code-snippets` files we need to look at each snippet
 		for (const snippet of this.data) {
+			if (resourceUri && !snippet.isFileIncluded(resourceUri)) {
```
jrieken: We need to do the same in `_filePathSelect` or only define the include/exclude-properties for *.code-snippet files (`languageScopeSchema` vs `globalSchema`). Background is that in the past we only had `<languageName>.json` for snippets and later added `xyz.code-snippets` with support for the `scope` property.

jrieken (changes requested):

src/vs/workbench/contrib/snippets/browser/snippets.ts
dmitrivMS: Doing this, I found a bunch more places where resourceUri needs to be passed in.

jrieken (approved):

PR #284188, opened 2025-12-18
Add Import Profile command and use it to implement drag & drop of code-profile file into VS Code

### LINKED ISSUE #148860: Import Settings Profile when you drag & drop a `.code-profile` to VS Code window
When you drag & drop a `.code-profile` file to VS Code, it should Import the profile. 

Doing so, it will provide the same experience when you drag & drop a `.code-workspace` file.

Thanks in advance

dmitrivMS:
Fixes #148860


src/vs/workbench/browser/dnd.ts
```diff
@@ -106,6 +110,14 @@ export class ResourcesDropHandler {
 		// Make the window active to handle the drop properly within
 		await this.hostService.focus(targetWindow);
 
+		// Check for profile file being dropped
+		for (const { resource } of editors) {
+			if (resource && endsWithIgnoreCase(resource.path, `.${PROFILE_EXTENSION}`)) {
```
bpasero: This seems like a layer break to me where knowledge about profiles is encoded in `dnd.ts`. A cleaner solution would be to allow the profiles component to register a handler into `dnd.ts` for specific file extensions?

bpasero (changes requested):

src/vs/workbench/browser/dnd.ts
dmitrivMS: Added registry, updated code.

bpasero (changes requested):
I now realise that we already have a registry approach for drag and drop:

https://github.com/microsoft/vscode/blob/6c6ca76be9524a3812c1f99a373640d81ef8b33b/src/vs/platform/dnd/browser/dnd.ts#L333-L358

I would suggest to put your work in there too if possible so that we have 1 registry and not 1 registry and 1 service to control drag and drop. For this little usage, I think registry approach is totally fine and a service is a bit too much as its not going to be used by many components.

bpasero (commented):
Thanks, LGTM, but yielding for Sandeep for the final go.

bpasero (dismissed):

sandy081 (approved):

PR #284244, opened 2025-12-18
Terminal: show dimensions overlay while resizing

### LINKED ISSUE #284046: Add overlay that shows terminal dimensions during resize
Cool feature in ghostty:

Image

tharani-2006:
Fixes #284046

Adds a temporary overlay to the integrated terminal that displays
the current terminal dimensions (columns × rows) while resizing.

- Appears only during active resize
- Updates live with terminal size
- Automatically hides after resizing stops
- Theme-aware and accessible

### DIFF
--- a/src/vs/workbench/contrib/terminal/browser/media/terminal.css
+++ b/src/vs/workbench/contrib/terminal/browser/media/terminal.css
@@ -16,7 +16,7 @@
 	z-index: 0;
 }
 
-.monaco-workbench .pane-body.integrated-terminal .terminal-resize-overlay {
+.monaco-workbench .terminal-resize-overlay {
 	position: absolute;
 	left: 50%;
 	top: 50%;
@@ -31,17 +31,16 @@
 	opacity: 0;
 	transition: opacity 80ms ease-out;
 	z-index: 35;
-	font-family: var(--vscode-editor-font-family);
 	font-size: 11px;
 }
 
-.monaco-workbench.hc-black .pane-body.integrated-terminal .terminal-resize-overlay,
-.monaco-workbench.hc-light .pane-body.integrated-terminal .terminal-resize-overlay {
+.monaco-workbench.hc-black .terminal-resize-overlay,
+.monaco-workbench.hc-light .terminal-resize-overlay {
 	box-shadow: none;
 	border-color: var(--vscode-contrastBorder);
 }
 
-.monaco-workbench .pane-body.integrated-terminal .terminal-resize-overlay.visible {
+.monaco-workbench .terminal-resize-overlay.visible {
 	opacity: 1;
 }
 
@@ -173,8 +172,8 @@
 	position: relative;
 }
 
-.monaco-workbench .terminal-editor .terminal-wrapper > div,
-.monaco-workbench .pane-body.integrated-terminal .terminal-wrapper > div {
+.monaco-workbench .terminal-editor .terminal-wrapper > .terminal-xterm-host,
+.monaco-workbench .pane-body.integrated-terminal .terminal-wrapper > .terminal-xterm-host {
 	height: 100%;
 }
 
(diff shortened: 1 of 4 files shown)


src/vs/workbench/contrib/terminal/browser/terminalTabbedView.ts
```diff
@@ -111,6 +124,8 @@ export class TerminalTabbedView extends Disposable {
 
 		this._terminalService.setContainers(parentElement, this._terminalContainer);
 
+		this._register(this._terminalService.onDidChangeInstanceDimensions(instance => this._handleInstanceDimensionsChanged(instance)));
```
Tyriar: Something I noticed is this shows up when not explicitly dragging it, such as when it's resized on window launch. Could we restrict this to show only when manually resizing somehow?

src/vs/workbench/contrib/terminal/browser/terminalTabbedView.ts
```diff
@@ -214,6 +229,40 @@ export class TerminalTabbedView extends Disposable {
...
+		if (instance.target !== TerminalLocation.Panel) {
+			return;
+		}
+
+		if (instance !== this._terminalGroupService.activeInstance) {
+			return;
+		}
+
+		const overlay = this._ensureResizeOverlay();
+		overlay.textContent = `${instance.cols}c \u00D7 ${instance.rows}r`;
```
Tyriar: This looks pretty good and how I would have done it if I hadn't seen the feature elsewhere:

```suggestion
		overlay.textContent = `${instance.cols} x ${instance.rows}`;
```

src/vs/workbench/contrib/terminal/browser/media/terminal.css
```diff
@@ -16,6 +16,35 @@
...
+	position: absolute;
+	left: 50%;
+	top: 50%;
+	transform: translate(-50%, -50%);
+	padding: 4px 10px;
+	border-radius: 4px;
+	background-color: var(--vscode-editorWidget-background);
+	color: var(--vscode-editorWidget-foreground);
+	border: 1px solid var(--vscode-editorWidget-border);
+	box-shadow: 0 2px 8px rgba(0, 0, 0, 0.5);
```
Tyriar, line 29: Looks good overall, I'll tweak the style of it to fit a little better once the other parts are done.

Tyriar (changes requested):
Great work, looks fantastic 👏

src/vs/workbench/contrib/terminal/browser/terminalTabbedView.ts
```diff
@@ -214,6 +230,44 @@ export class TerminalTabbedView extends Disposable {
...
+	private _ensureResizeOverlay(): HTMLElement {
+		if (!this._resizeOverlay) {
+			this._resizeOverlay = $('.terminal-resize-overlay');
+			this._resizeOverlay.setAttribute('role', 'status');
+			this._resizeOverlay.setAttribute('aria-live', 'polite');
+			this._parentElement.append(this._resizeOverlay);
+			this._register(toDisposable(() => this._resizeOverlay?.remove()));
+		}
+		return this._resizeOverlay;
+	}
```
Tyriar: Playing around with this more, it should probably actually live in `TerminalInstance` since this falls over right now when there are multiple terminals or tabs are shown:

Image

It's also fine that it shows for all split terminals at once, not just the active one. But we do want to make sure that the window startup and terminal start up problem is handled. Moving it into `TerminalInstance` should make it a lot easier to accomplish this as it knows when it was just launched, it knows the first resize event, etc.

Tyriar (changes requested):

Tyriar (approved):
Thanks @tharani-2006 👏 

I made a few tweaks that were easier to do on my end than explain.

dmitrivMS (approved):

PR #284360, opened 2025-12-19
Finalize quickPickButtonLocation API proposal

### LINKED ISSUE #144956: Toggle support for QuickInput in API

dmitrivMS:
Fixes #144956


TylerLeonhardt:
@dmitrivMS PR is red

dmitrivMS:
> > @dmitrivMS PR is red
> 
> Yeah, sorry - bad merge, will fix.

Fixed

TylerLeonhardt (approved):

TylerLeonhardt:
One thing worth discussing in API Sync... where should we put a `secondary` property? I'm trying to light something like this up using `Toolbar`:
image

It could change the shape of this `location` property actually

alexr00:
Got it, thank you!

alexr00:
The way we do this in package.json is by allowing a button or menu. Not sure if we want to expose a similar concept here. Here, that would mean having a `QuickInputMenu` in addition to the current `QuickInputButton`. I think that would fit fine into the current API, but I don't know if it's the best approach (as opposed to having a `secondary` property). I agree that we should discuss in the API sync.

mjbvz (approved):

TylerLeonhardt:
I got the ball rolling: https://github.com/microsoft/vscode/pull/285984

Issue #65668, opened 2018-12-25
VS Code fails to detect the correct indentation for certain YAML files


segevfiner:
- VSCode Version: 1.30.1
- OS Version: macOS 10.14.2

Steps to Reproduce:

1. Save the following as `foo.yml`:
```yaml
version: 2

jobs:
  build:
    docker:
      - circleci/golang:1.11

  environment:
    TEST_RESULTS: /tmp/test-results

  steps:
    - checkout
    - run: mkdir -p $TEST_RESULTS # create the test results directory

    - restore_cache:
        keys:
          - v1-pkg-cache

    - run:
        name: dep ensure
        command: dep ensure -v

    - run:
        name: Run unit tests
        command: |
          trap "go-junit-report <${TEST_RESULTS}/go-test.out > ${TEST_RESULTS}/go-test-report.xml" EXIT
          go test -v ./... | tee ${TEST_RESULTS}/go-test.out

    - run:
        name: Build
        command: go build -v

    - save_cache:
        key: v1-pkg-cache
        paths:
          - "/go/pkg"

    - store_artifacts:
        path: /tmp/test-results
        destination: raw-test-output

    - store_test_results:
        path: /tmp/test-results
```
2. Check the detected indentation, it will be 4 spaces, despite the file being indented with 2 spaces.

I didn't change any related settings.

The code that does this is [src/vs/editor/common/model/indentationGuesser.ts](https://github.com/Microsoft/vscode/blob/master/src/vs/editor/common/model/indentationGuesser.ts) and using the developer tools I can see that it returns 4 spaces for this file.

Does this issue occur when all extensions are disabled?: Yes

PR #284371, opened 2025-12-19
Adjust heuristic for preferring 2 over 4 for indentation size

### LINKED ISSUE #249040: Unexpected 2-spaces indent detection despite of predominant 4-spaces usage
- VS Code Version: 1.95
- OS Version: Linux x64 5.15.0-139-generic

When _Editor: Detect Indentation_ is enabled and there are some areas indented using 2 spaces, VSCode detects a 2-space indentation even when the file is mostly indented with 4 spaces.

Image

This doesn't happen for other pairs of values such as {3, 6} or {4,8}.
Is there any reason for choosing 2 spaces over 4? Why does this happen only for the {2, 4} pair?

dmitrivMS:
Fixes #249040

The heuristic was added by @alexdima to fix issue #65668 where we were detecting indent in YAML incorrectly (because YAML indent skips `- ` portion). However, that change also broke detection in non-list based languages.

This change adjusts the heuristic to be less aggressive.
Added tests for both issues to ensure there is no regression.


src/vs/editor/common/model/indentationGuesser.ts
```diff
@@ -192,10 +192,7 @@ export function guessIndentation(source: ITextBuffer, defaultTabSize: number, de
 
 	// Guess tabSize only if inserting spaces...
 	if (insertSpaces) {
-		let tabSizeScore = (insertSpaces ? 0 : 0.1 * linesCount);
-
-		// console.log("score threshold: " + tabSizeScore);
-
+		let tabSizeScore = 0;
```
aiday-mar, line 195: Hi could you please describe why you changed this line?

dmitrivMS: At this point `insertSpaces` is `true` due to the `if` statement above, so the conditional always evaluates to `0`.

aiday-mar: I see yes

aiday-mar (approved):

Issue #283174, opened 2025-12-12
"Search with AI" option shows even if you have "Disable AI Features" enabled


Kampfkarren:
Does this issue occur when all extensions are disabled?: Yes/No

- VS Code Version: 1.107.0
- OS Version: Darwin arm64 24.6.0

Steps to Reproduce:

1. Enable chat.disableAIFeatures
2. Search something
3. Get "Search with AI" option, even when this usually disables all features.

<img width="307" height="45" alt="Image" src="https://github.com/user-attachments/assets/14ce6607-ec49-4f34-9568-b9d81bb285af" />

Clicking on the button does not do anything. I would like for this button to not be here, akin to the other AI functionality being hidden.

bpasero:
Good catch, this reproduces if copilot chat is installed and running: in that case the context key for AI search is never cleaning up.

@osortega please note my change in https://github.com/microsoft/vscode/pull/286307

PR #284463, opened 2025-12-19
"Search with AI" option shows even if you have "Disable AI Features" enabled (#283174)

bpasero:


osortega (approved):
Thanks for fixing it ✨ at first I thought we also needed to disable the command as well but looks like we disable the provider so the when clause handles that https://github.com/microsoft/vscode/blob/17cf1ce11ac873561637f74dadb80dc84a4255d2/src/vs/workbench/contrib/search/browser/searchActionsTopBar.ts#L218-L219

Issue #275056, opened 2025-11-04
Inline chat: no response returned in this case


bpasero:
Testing #274768

I am in https://github.com/Microsoft/vscode-bisect project, added a new file `helper.ts` into the `src` folder and opened inline chat with `add copyright header`.

No response was generated.

[Archive.zip](https://github.com/user-attachments/files/23329326/Archive.zip)

bpasero:
Even a similar request to "implement a method to print fibonacci numbers" yielded no result.

jrieken:
There is no loop. Once the tool-invocation is done the request is terminated and the tool result is never send back to the LM

jrieken:
I am aware of the `ToolCallingLoop` but I shied away from using a real loop (which is terminated by the LLM not calling another tool) for performance reasons. Today, I bet on my prompt (and very limited tool options) being strict enough so that the LM makes just a single tool call


>  correct any incorrect edit it may make (even the best current models fail 2-8% of the time).

I am obviously interested in correct edits but read this as in 92-98% of the time the optimistic and fast approach is good. Tho, I can improve on error handling and attempt a retry when tools fail

jrieken:
@connor4312 To help me staying fast but still recover from errors how about adding `LanguageModelToolResult#isError`. That's something I could use to re-start the loop and keep the immediate-done strategy of today

connor4312:
Edits can get partially applied in the apply_patch and multi_replace_string cases. So we don't throw in these cases and instead return a textual error for the model.

jrieken:
Closing this as inline chat now check for edit tool errors and then does new tries.

PR #284549, opened 2025-12-20
Add error handling properties to language model tool results

zhichli:
Introduce additional properties to handle error states in language model tool results. This enhancement allows for better communication of errors and messages from the tool, improving overall robustness and user experience.

### DIFF
--- a/src/vs/workbench/api/common/extHostTypeConverters.ts
+++ b/src/vs/workbench/api/common/extHostTypeConverters.ts
@@ -3594,7 +3594,7 @@ export namespace LanguageModelToolSource {
 
 export namespace LanguageModelToolResult {
 	export function to(result: IToolResult): vscode.LanguageModelToolResult {
-		return new types.LanguageModelToolResult(result.content.map(item => {
+		const toolResult = new types.LanguageModelToolResult(result.content.map(item => {
 			if (item.kind === 'text') {
 				return new types.LanguageModelTextPart(item.value, item.audience);
 			} else if (item.kind === 'data') {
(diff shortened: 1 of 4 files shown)


src/vs/workbench/api/common/extHostTypeConverters.ts
```diff
@@ -3603,6 +3603,21 @@ export namespace LanguageModelToolResult {
...
+		}
+		if (result.toolResultError) {
+			(toolResult as vscode.ExtendedLanguageModelToolResult).toolResultError = result.toolResultError;
+		}
+		if (result.hasError !== undefined) {
+			(toolResult as vscode.ExtendedLanguageModelToolResult).hasError = result.hasError;
+		}
+
+		return toolResult;
 	}
```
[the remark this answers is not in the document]
zhichli: not needed

src/vs/workbench/contrib/chat/common/languageModelToolsService.ts
```diff
@@ -208,6 +208,8 @@ export interface IToolResult {
 	toolMetadata?: unknown;
 	/** Whether to ask the user to confirm these tool results. Overrides {@link IToolConfirmationMessages.confirmResults}. */
 	confirmResults?: boolean;
+	/** Whether there was an error calling the tool. The tool may still have partially succeeded. */
+	hasError?: boolean;
```
roblourens, line 212: I'm a little confused about how hasError/toolResultError are used today, are those redundant? The tool can also throw.

zhichli: AFAIK, hasError might be only used for [this](https://github.com/microsoft/vscode/issues/275056#issuecomment-3534770436) but could be useful to cover other tool calls' failure cases later ? 

`toolResultError` is logged [here](https://github.com/microsoft/vscode/blob/d87d06f0e66419d71d9cfc1e46f2f8f18e4c1909/src/vs/workbench/contrib/chat/browser/languageModelToolsService.ts#L473) when tool call throws, does it mean the same as you mentioned by `The tool can also throw.`?

roblourens: I think this is ok, we can run it past @connor4312 that it is correct and useful to share these props with the extension for logging from eval runs

connor4312: Something that I think we may want to correct if we're expanding this is making this a percent rate (0-1) instead of just a boolean. For both apply_patch and multi_replace_string, we might apply some edits in the update but not all of the edits.

PR #284609, opened 2025-12-20
Show command descriptions as tooltips in Command Palette

### LINKED ISSUE #260868: Show command descriptions as tooltips in Command Palette
When using the Command Palette (`Ctrl+Shift+P`), only the command title is shown in the list.
If you don't already know what a command does, the only way to find out is to run it or go search for it.

I think we could improve this by showing the command’s description (if available) when you hover over it.

VS Code already stores descriptions for a lot of commands (both core and from extensions) in the `CommandsRegistry`. Right now that data isn’t surfaced in the UI, which feels like a missed opportunity.

### Why this would help
- Makes it easier to explore and understand less common commands.
- Good for new users who are still learning what’s available.
- Doesn’t require any new data — we’d just use what’s already there.

### Idea for implementation
- In the quick input list rendering (`quickInputList.ts`), grab the metadata for each command from `CommandsRegistry.getCommand(id)`.
- If it has a description, set it as the `title` attribute for that list item (browser/Electron will show this as a tooltip).
- Fallback to the command’s title if there’s no description.

**Example:**

Before:  
Hover over `Toggle Side Bar Visibility` → nothing happens.

After:  
Hover over `Toggle Side Bar Visibility` → shows “Shows or hides the sidebar in the workbench”.

If this sounds reasonable, I’m happy to try implementing it.

Riya-chandra:
## What this PR does

This PR surfaces existing command descriptions as native tooltips in the
Command Palette.

VS Code already stores command metadata (including descriptions) in
`CommandsRegistry`, but this information was not visible in the UI.
This change exposes that existing data via a simple hover tooltip,
making it easier to explore and understand less familiar commands.

## Why this is useful

- Improves discoverability of commands, especially for new users
- Reduces trial-and-error when exploring the Command Palette
- Uses existing metadata only — no new data or behavior introduced
- Minimal, UI-only change with low risk

## Implementation details

- Retrieves command metadata via `CommandsRegistry.getCommand`
- Uses `metadata.description` when available (with localized string support)
- Applies tooltip at the entry level (`title` attribute)

## Notes

Local build was not run due to environment constraints. This change is
isolated to the Quick Input UI layer.

Fixes #260868
@TylerLeonhardt...Please look into this


src/vs/platform/quickinput/browser/quickInputList.ts
```diff
@@ -543,6 +544,30 @@ class QuickPickItemElementRenderer extends BaseQuickInputListRenderer<QuickPickI
...
+					tooltip = command.metadata.description.value;
+				}
+			}
+
+			if (tooltip) {
+				data.entry.title = tooltip;
+			}
+		}
+
 	}
```
TylerLeonhardt, line 552: This change is changing the quick pick component for _all_ quick picks. I would expect a change only for the command palette, which is called the "Commands Quick Access". 

https://github.com/microsoft/vscode/blob/8e8e866dc462652b87189eacd85b13402e033bcb/src/vs/platform/quickinput/browser/commandsQuickAccess.ts#L49

TylerLeonhardt (changes requested):
please make the change for only the commands quick access

src/vs/platform/quickinput/browser/commandsQuickAccess.ts
```diff
@@ -274,7 +275,18 @@ export abstract class AbstractCommandsQuickAccessProvider extends PickerQuickAcc
...
+			if (command?.metadata?.description) {
+				commandPick = {
+					...commandPick,
+					tooltip:
+						typeof command.metadata.description === 'string'
+							? command.metadata.description
+							: command.metadata.description.value
+				};
+			}
+        }
```
TylerLeonhardt: the commandPick has the commandDescription already.

Riya-chandra: @TylerLeonhardt Thanks for the clarification!

You're absolutely right — `commandPick` already contains
`commandDescription`. I've removed the `CommandsRegistry.getCommand`
lookup and now directly use `commandPick.commandDescription` when
setting the tooltip.

This keeps the change minimal and aligns with the existing
Command Palette data flow.

Please let me know if this looks correct now.

src/vs/platform/quickinput/browser/commandsQuickAccess.ts
```diff
@@ -274,6 +275,12 @@ export abstract class AbstractCommandsQuickAccessProvider extends PickerQuickAcc
 		if (commandPick.type === 'separator') {
 			return commandPick;
 		}
+		if (!commandPick.tooltip && commandPick.commandDescription) {
```
TylerLeonhardt: I'd rather you compute the tooltip first, and then add it to the spread that's happening further down.

So
```
const tooltip = commandPick.tooltip || commandPick.commandDescription.value;
```

Similar to how aria label is handled

Riya-chandra: @TylerLeonhardt Thanks for the guidance!
I've updated the implementation.
Please let me know if this looks good now.

TylerLeonhardt (dismissed):

TylerLeonhardt:
Thanks for contributing! Nice work!

Tyriar (dismissed):

src/vs/platform/quickinput/browser/commandsQuickAccess.ts
```diff
@@ -27,6 +27,7 @@ import { IQuickAccessProviderRunOptions } from '../common/quickAccess.js';
 import { IKeyMods, IQuickPickSeparator } from '../common/quickInput.js';
 import { IStorageService, StorageScope, StorageTarget, WillSaveStateReason } from '../../storage/common/storage.js';
 import { ITelemetryService } from '../../telemetry/common/telemetry.js';
+import { CommandsRegistry } from '../../commands/common/commands.js';
```
TylerLeonhardt: This is unused, please delete it

TylerLeonhardt (approved):

dmitrivMS (approved):

gjsjohnmurray:
Only a few commands currently produce a tooltip. For verification purposes here's one that does:

image

PR #284645, opened 2025-12-21
include uppercase `-I` in sed in-place option detection

tamuratak:
include uppercase `-I` in sed in-place option detection

The `-I` option is available for `sed` on macOS and BSDs. 

- https://man.freebsd.org/cgi/man.cgi?sed
- https://ss64.com/mac/sed.html

Although There is no official macOS manual available online, you can check by executing `man sed` on macOS.

ProductName:		macOS
ProductVersion:		15.7.3
BuildVersion:		24G419

CC: @Tyriar


Tyriar (approved):
Thanks a bunch!

lszomoru (approved):

PR #284685, opened 2025-12-21
fix: close singleton editors when using 'Copy into New Window'

### LINKED ISSUE #284525: Integrated Browser: unexpected behavior with "Copy into New Window"
- VS Code Version: 1.109.0-insider
- OS Version: macOS Tahoe 26.2

Steps to Reproduce:

1. In command palette: "Browser: Open Integrated Browser" (Note that this is different from the Simple Browser)
2. Open a website with animations, such as https://doodles.google/
3. Notice the smooth animations on the page
4. Right click on the editor tab and "Copy into New Window"
    Image

5. Two issues happen:
    1. 🐛 The new window is synced to the old window, so every scroll or click or type is mirrored between these windows. I'm not sure if this is expected or not
    2. 🐛 At least one of the windows will now show some lag in the animation, or the popped out window might even glitch, very briefly disappearing every so often

**Expected behavior:**
either the two windows become independent of each other, not synced, or if they're synced there no perf hit, or as another option we just don't show the option to "Copy into New Window" if that's not meant to be supported.

The approach that Simple Browser takes is to keep the "Copy into New Window" button, but they force it to work like "Move into New Window": Once the copy is created, they force the closure of the original one. (This is not a perfect solution, because the original window ends up with focus, so the popped out window kind of disappears until you put it in focus again)

---

Related to feature: https://github.com/microsoft/vscode/issues/286579

Rani367:
## Summary
  Fixes singleton editors (webviews, terminals, etc.) being shared between
  windows when using "Copy into New Window".

  ## Problem
  When using "Copy into New Window" on a webview editor (like the integrated browser):
  - Both windows shared the same webview instance
  - Interactions were mirrored between windows
  - Performance lag/glitching occurred in animations

  ## Solution
  Treat singleton editors as moves even when the action is "copy" -
  the original editor is closed after opening in the new window.

  ## Test Plan
  - [x] Added unit tests for singleton editor copy behavior
  - [x] Added unit test for onWillMoveEditor event firing
  - [x] All 64 editor group tests pass

  Fixes #284525

### DIFF
--- a/src/vs/workbench/browser/parts/editor/editorGroupView.ts
+++ b/src/vs/workbench/browser/parts/editor/editorGroupView.ts
@@ -1447,17 +1447,21 @@ export class EditorGroupView extends Themable implements IEditorGroupView {
 			}
 		}
 
+		// Singleton editors cannot be copied, only moved (https://github.com/microsoft/vscode/issues/XXXXX)
+		const isSingletonEditor = editor.hasCapability(EditorInputCapabilities.Singleton);
+		const isActualMove = !keepCopy || isSingletonEditor;
+
 		// When moving/copying an editor, try to preserve as much view state as possible
 		// by checking for the editor to be a text editor and creating the options accordingly
 		// if so
 		const options = fillActiveEditorViewState(this, editor, {
 			...openOptions,
 			pinned: true, 																// always pin moved editor
-			sticky: openOptions?.sticky ?? (!keepCopy && this.model.isSticky(editor))	// preserve sticky state only if editor is moved or explicitly wanted (https://github.com/microsoft/vscode/issues/99035)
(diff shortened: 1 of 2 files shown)


bpasero (changes requested):
This change seems to break terminal editors badly:
* apply this change
* create a new terminal in editor
* type `ls`
* select "Copy into new window"
* close the window

=> I see a terminal tab opened that I cannot click on.

https://github.com/user-attachments/assets/e346f919-82e8-4e77-9f20-d8c86296c0f1

bpasero:
Besides, I cannot reproduce any issue with terminals or webview editors (such as markdown preview).

Rani367:
@bpasero  Looking back at #284525, I see it's specifically about the Integrated Browser on a feature branch - not relevant for main. My bad, I'll close this.

PR #284772, opened 2025-12-22
Add capitalized `.R` extension to R

### LINKED ISSUE #284740: Use .R (not .r) as default R file extension
- VS Code Version: 1.106
- OS Version: Fedora Linux 41 and Windows 11

Steps to Reproduce:

1. Create a New File (`Ctrl` + `N`)
2. Select the R language (`Ctrl` + `K` `M` and select `R`) 
3. Save the file (`Ctrl` + `S`)
4. The default file extension is `.r`

The default file extension should be `.R` (not `.r`), as used in [R source](https://svn.r-project.org/R/trunk/src/library/base/R/) and other R IDEs (e.g. [RStudio](https://posit.co/download/rstudio-desktop/) and [Positron](https://positron.posit.co/)), and as recommended in the [Tidyverse Style Guide](https://style.tidyverse.org/files.html#names).

Another user has already asked that on [StackOverflow](https://stackoverflow.com/q/76906572).

RedCMD:
fixes #284740
* #284740

adds `.R` as the default R file extension


extensions/r/package.json
```diff
@@ -17,6 +17,7 @@
       {
         "id": "r",
         "extensions": [
+          ".R",
```
aeschli, line 20: The better fix us to just rename `.r` to `.R`. Same for the `.Rhistory` and `.Rprofile`.
I tested on Windows and Linux that we still correctly associate `.r` files

RedCMD: I don't see any capitalized `.Rt` files so I just left it
https://github.com/search?q=path%3A*.Rt&type=code

fixed the other ones

aeschli (approved):

Tyriar (approved):

Issue #244862, opened 2025-03-27
Line number font size should be configurable 🥰


lopugit:
With the addition of hiding glyph, the hidden lineDecorationsWidth setting, and the ability to hide line numbers and folding icon altogether. 

I think the last step to seal the deal is to have a line number font size setting ! 

I definitely don't need the font-size to be THAT big hahaha ! 😂 

Oh wait is this open source.. ? 

HUH, cool !

PR #284836, opened 2025-12-23
feat(editor): make line number font size configurable (#244862)

JPSAUD501:
## Summary

- Add the `editor.lineNumbersFontSize` option (`0` uses the editor font size; otherwise a pixel value).
- Propagate the value into layout so line number width scales with the chosen font size.
- Render line numbers with the configured size in the view overlay.
- Add coverage in editor layout tests.
- Update Monaco typings/enums for the new option.

#244862

## Testing

- `./test.sh --grep "line numbers"`.
- Manual: set `"editor.lineNumbersFontSize": 10 | 13 | 16`, then verify line number width and rendering.

image

image


JPSAUD501:

PR #166240, opened 2022-11-13
Update mouseHandler.ts

jogibear9988:
Fix issue with popups not workin inside of a webcomponent, see: https://github.com/microsoft/monaco-editor/issues/3409

PR #284890, opened 2025-12-23
Fix shadow DOM mouseleave detection in editor

nick-romano:
### Bug:

When the editor is hosted inside a Shadow DOM, mouse moves could make the global mouseleave recovery think the pointer left the editor as soon as the event target falls outside the light-DOM container. This erroneously fires mouseLeave, which can break hover/selection behavior even though the pointer is still visually over the editor.

### Fix:

In the global mousemove recovery path, when the event target is outside the light-DOM container, we now query shadowRoot.elementsFromPoint and ensure none of the hit elements belong to the editor before emitting mouseLeave. This keeps the monitor active but only triggers leave when the pointer truly exits the editor, including Shadow DOM embeddings.

Fixes monaco issue: https://github.com/microsoft/monaco-editor/issues/3409

Based on the changes requested in https://github.com/microsoft/vscode/pull/166240


nick-romano:
@hediet could you please review, this has been a issue for well over a year.

PR #284922, opened 2025-12-24
update Lua grammar

sumneko:
I have updated the Lua syntax file to add support for the latest Lua 5.5


alexr00:
Thanks for the PR! I missed this, and we have just updated the grammars yesterday. Heads up @dmitrivMS.

PR #285020, opened 2025-12-25
Add inline reference handling for markdown conversion in RunSubagentTool

### LINKED ISSUE #282509: Subagent file path context is missing in response to orchestrator
- Copilot Chat Extension Version:
- VS Code Version:  1.107.0-insider (Universal)
- OS Version: Darwin arm64 24.6.0

Issue: When subagent returns file path in the response some formats get lost in the way.. 

Steps to Reproduce:

1. In agent mode, type the below prompt:
```Call a subagent with the following instructions:
Provide the path to a local file on the current repository. you must provide the path in at least 5 different formats (e.g. full path/ relative path / md file link / string / backtick etc.), only use ``` format in one output example
```
looks like the subagent response in the chat pane has all formats:
```
## File Path Formats for README.md

### 1. Full Absolute Path (Plain Text)
/<full path>/README.md

### 2. Relative Path from Workspace Root (Markdown Link)
[README.md](README.md)

### 3. Inline Backtick Code Format
`README.md`

### 4. Quoted String Format
"README.md"

### 5. Triple Backtick Code Block
\```
/<full path>/README.md
\```

### 6. URI/File Protocol Format
file:///<full path>/README.md

### 7. Tilde Home Directory Format
~/<path>/README.md

### 8. Markdown Link with Descriptive Text
[Repository README](README.md)

### 9. HTML-style Path Reference
<code>README.md</code>

### 10. Shell Variable Style
`$PWD/README.md` (when in workspace root)
```

but looking at the response the orchestrator gets at the start of the next chat looks like the common file formats are omitted (1, 2, 3, 6, 7)
```
## File Path Formats for 

### 1. Full Absolute Path (Plain Text)


### 2. Relative Path from Workspace Root (Markdown Link)


### 3. Inline Backtick Code Format


### 4. Quoted String Format
"README.md"

### 5. Triple Backtick Code Block
\```
/<full path>/README.md
\```

### 6. URI/File Protocol Format


### 7. Tilde Home Directory Format
~/<path>/README.md

### 8. Markdown Link with Descriptive Text
Repository README

### 9. HTML-style Path Reference
<code>README.md</code>

### 10. Shell Variable Style
`$PWD/README.md` (when in workspace root)
```

tamuratak:
Add inline reference handling for markdown conversion in RunSubagentTool. Fix #282509

CC: @roblourens

### DIFF
--- a/src/vs/workbench/contrib/chat/common/tools/runSubagentTool.ts
+++ b/src/vs/workbench/contrib/chat/common/tools/runSubagentTool.ts
@@ -36,12 +36,12 @@ import {
 	VSCodeToolReference,
 	IToolAndToolSetEnablementMap
 } from '../languageModelToolsService.js';
-import { basename } from '../../../../../base/common/resources.js';
 import { URI } from '../../../../../base/common/uri.js';
 import { isLocation } from '../../../../../editor/common/languages.js';
 import { ComputeAutomaticInstructions } from '../promptSyntax/computeAutomaticInstructions.js';
 import { ManageTodoListToolToolId } from './manageTodoListTool.js';
 import { createToolSimpleTextResult } from './toolHelpers.js';
+import { Schemas } from '../../../../../base/common/network.js';
 
 export const RunSubagentToolId = 'runSubagent';
 
@@ -213,7 +213,7 @@ export class RunSubagentTool extends Disposable implements IToolImpl {
 
 						if (part.kind === 'inlineReference') {
 							// Convert inline references into markdown links for the tool result
-							markdownParts.push(inlineReferenceToMarkdown(part));
+							markdownParts.push(inlineReferenceToUriString(part));
 						} else {
 							// Collect markdown content for the tool result
 							markdownParts.push(part.content.value);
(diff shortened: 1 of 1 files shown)


src/vs/workbench/contrib/chat/common/tools/runSubagentTool.ts
```diff
@@ -288,3 +296,28 @@ export class RunSubagentTool extends Disposable implements IToolImpl {
...
+		label = part.name;
+	} else if (URI.isUri(reference)) {
+		label = basename(reference);
+	} else if (isLocation(reference)) {
+		label = basename(reference.uri);
+	} else {
+		label = reference.name ?? basename(reference.location.uri);
+	}
+	const escapedLabel = label.replace(/[[\]()\\]/g, '\\$&');
+	return `[${escapedLabel}](${refUri.toString()})\n`;
```
roblourens: I think the formatting is a bit overkill, I think that just giving `uri.fsPath` would be best, rather than doing a md-link and stringifying a URI. We shouldn't stringify the URI here on the frontend because it will use the internal remote scheme `vscode-remote` when on a remote. I think the logic can be copied from here: https://github.com/microsoft/vscode-copilot-chat/blob/f809049bb00c86f54f7a2eb0b0be772b8da2ee79/src/platform/prompts/common/promptPathRepresentationService.ts#L47-L52

(fsPath for typical URIs but toString anything out of the ordinary)

roblourens (commented):
Thanks for the PR!

roblourens:
Thanks, but I'm actually handling this in a different way: https://github.com/microsoft/vscode-copilot-chat/pull/2959

PR #285099, opened 2025-12-26
Fix context handling in ChatWidget while re-editing

### LINKED ISSUE #281766: Editing submitted GitHub Copilot prompt removes file references
I wrote out a very long planning prompt for the Plan agent, including multiple `#file:` references. I submitted the prompt, and the LLM wanted to fetch https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/lang/CharSequence.html . I of course wanted it retrieve the latest Java version, so I canceled the request and edited the prompt, adding an explicit `#fetch` with https://docs.oracle.com/en/java/javase/25/docs/api/java.base/java/lang/CharSequence.html .

(I note that GitHub changed the agent to "Agent" instead of "Plan", but that's not the point of this ticket.)

Suddenly all my `#file:` references are no longer real references; they are just normal text! For example:

> Would this allow us to refactor `#file:Hash.java` to be …

was changed to:

> Would this allow us to refactor #file:Hash.java to be …

I know the references have been removed, because when I submitted it again, the only reference shown was `#fetch`.

Now I have to edit the prompt yet again (again it changed the agent to "Agent", and I had to change it back to "Plan") and go find all the file references and retype them so that they show up as real references. Not only is this tedious and error-prone, it chews up a lot of the time the LLM was supposed to save me!

The LLM is supposed to be reducing the amount of repetitive, tedious, replacement work I have to do—not adding to it!

```
Version: 1.106.3 (user setup)
Commit: bf9252a2fb45be6893dd8870c0bf37e2e1766d61
Date: 2025-11-25T22:28:18.024Z
Electron: 37.7.0
ElectronBuildId: 12781156
Chromium: 138.0.7204.251
Node.js: 22.20.0
V8: 13.8.258.32-electron.0
OS: Windows_NT x64 10.0.26200
```

tamuratak:
Fix context handling in ChatWidget while re-editing chat requests. `currentElement.variables` should be added to `currentContext`. Fix #281766.

### DIFF
--- a/src/vs/workbench/contrib/chat/browser/chatWidget.ts
+++ b/src/vs/workbench/contrib/chat/browser/chatWidget.ts
@@ -1608,14 +1608,23 @@ export class ChatWidget extends Disposable implements IChatWidget {
 
 			// set contexts and request to false
 			const currentContext: IChatRequestVariableEntry[] = [];
+			const addedContextIds = new Set<string>();
+			const addToContext = (entry: IChatRequestVariableEntry) => {
+				if (addedContextIds.has(entry.id) || isWorkspaceVariableEntry(entry)) {
+					return;
+				}
+				if ((isPromptFileVariableEntry(entry) || isPromptTextVariableEntry(entry)) && entry.automaticallyAdded) {
+					return;
+				}
+				addedContextIds.add(entry.id);
+				currentContext.push(entry);
+			};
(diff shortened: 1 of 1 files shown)


justschen (approved):
this looks like it's going in the right direction, but still has some issues with files attached via `#`

this is a good change regardless, thank you

lszomoru (approved):

Issue #284498, opened 2025-12-19
Copy All doesn't work for AI search


osortega:
Testing #9465

* use AI search
* right click a result
* select "Copy All"
* :bug: nothing happens

---
Migrated from microsoft/vscode-copilot#9557

PR #285112, opened 2025-12-26
Fix/ai search copy all #284498

eidriahn:
Fixes issue https://github.com/microsoft/vscode/issues/284498

The issue was caused by the fact that the copyAllCommand wasn't considering whether the click was done on normal searches or ai searches.

Test it:

Go to search -> AI Search
Search for anything
Right click on any result and select 'Copy All'
Paste the copied contents into a file and check if they correspond.


osortega (approved):
This is great! Thank you!

dmitrivMS (approved):

PR #285233, opened 2025-12-28
Avoid setting state-related ARIA attributes on separators

### LINKED ISSUE #282663: In Scan Mode, Screen Reader announces Separator as Checkbox Instead of Separator:A11y_Visual Studio Code Client_Screen reader
## GitHub Tags:
#A11yTCS; #A11ySev3; #Visual Studio Code Client; #BM_Visual Studio Code Client_Win32_JULY2024; #DesktopApp; #Win32; #Benchmark;#Name Role Value;#MAS4.1.2; #Closed; #A11yFixed;

## Environment and OS details:
Application Name: Visual Studio Code
Version: 1.107.0 (user setup)
Win11 25H2 OS build :26100.4770

## Reproduction Steps:
1. Launch VS Code 
2. Navigate to 'Explorer' tab from left navigation 
3. Access 'Open Editors' section and activate 'New File' button 
4. Open context menu (Ctrl + Shift + F10)  for newly added file
5. Press down arrow key to Reopen editor with  and press ENTER key.
6. Press down arrow key to Jupyter Notebook and press ENTER key.
7. TAB till More option button code cell and press ENTER key.
8. Now navigate in scan mode to separator with down arrow key and observe the issue

## Actual Result:
While navigating in Scan Mode using a screen reader, the focus moves to a separator element, but it is incorrectly announced as a checkbox. 

## Expected Result:
When the screen reader focus lands on a separator, it should be correctly announced as a "separator", not as a different element type like a checkbox or screen reader focus should not move on separator in scan mode

## User Impact:
Visually impaired users may get confused due to incorrect information about the UI element. This can lead to misunderstandings about the page structure and reduce overall accessibility. 

## Attachment:

Image

Code snippet

Image

dmitrivMS:
Fixes #282663


src/vs/base/browser/ui/actionbar/actionViewItems.ts
```diff
@@ -427,6 +431,10 @@ export class ActionViewItem extends BaseActionViewItem {
 	}
 
 	protected override updateChecked(): void {
+		if (this.action.id === Separator.ID) {
```
benibenj: What about making `checked` be undefined on the Separator. Then this would not be needed. It also seems to make more sense that it is set to `undefined` instead of `false`

dmitrivMS: Will take a look

dmitrivMS: FWIW, `updateChecked` is called unconditionally from `render`, so setting it to undefined may be futile.

dmitrivMS: Done: set `checked` to `undefined` on the `Separator`. Removed checks as requested.

benibenj (approved):
// cc @jrieken

Issue #284323, opened 2025-12-18
Checkbox quick pick hover doesn't support icons


Tyriar:
<img width="2944" height="514" alt="Image" src="https://github.com/user-attachments/assets/125b4091-4860-4ba5-b48b-5f2ab885b151" />

TylerLeonhardt:
Verification:
* Create an extension with this:
   ```ts
   await window.showQuickPick(['$(gear) one'], {
   	canPickMany: true,
   	placeHolder: 'hover over the checkbox',
   });
   ```
* show it and hover over the checkbox
* it should not have the gear in the hover (this is because the hover currently just passes in a string, I may change this to pass in a markdown string so that the gear can be rendered, but without modifying core at all this would verify this bug)

PR #285250, opened 2025-12-28
fix: fixes icons not showing when hovering quick pick checkboxes

eidriahn:
Fixes issue #284323

A simple fix that adds `supportThemeIcons: true` to the content of the setupDelayedHover function so that icons are correctly shown.

### DIFF
--- a/src/vs/base/browser/ui/toggle/toggle.ts
+++ b/src/vs/base/browser/ui/toggle/toggle.ts
@@ -6,8 +6,10 @@
 import { IAction } from '../../../common/actions.js';
 import { Codicon } from '../../../common/codicons.js';
 import { Emitter, Event } from '../../../common/event.js';
+import { IMarkdownString } from '../../../common/htmlContent.js';
 import { KeyCode } from '../../../common/keyCodes.js';
 import { ThemeIcon } from '../../../common/themables.js';
+import { hasKey } from '../../../common/types.js';
 import { $, addDisposableListener, EventType, isActiveElement } from '../../dom.js';
 import { IKeyboardEvent } from '../../keyboardEvent.js';
 import { BaseActionViewItem, IActionViewItemOptions } from '../actionbar/actionViewItems.js';
@@ -20,7 +22,7 @@ import './toggle.css';
 export interface IToggleOpts extends IToggleStyles {
 	readonly actionClassName?: string;
 	readonly icon?: ThemeIcon;
-	readonly title: string;
+	readonly title: string | IMarkdownString | HTMLElement;
 	readonly isChecked: boolean;
 	readonly notFocusable?: boolean;
 	readonly hoverLifecycleOptions?: IHoverLifecycleOptions;
@@ -126,7 +128,7 @@ export class Toggle extends Widget {
 	get onKeyDown(): Event<IKeyboardEvent> { return this._onKeyDown.event; }
 
 	private readonly _opts: IToggleOpts;
-	private _title: string;
+	private _title: string | IMarkdownString | HTMLElement;
 	private _icon: ThemeIcon | undefined;
 	readonly domNode: HTMLElement;
 
@@ -153,7 +155,7 @@ export class Toggle extends Widget {
 
 		this.domNode = document.createElement('div');
 		this._register(getBaseLayerHoverDelegate().setupDelayedHover(this.domNode, () => ({
-			content: { value: this._title, supportThemeIcons: true },
+			content: this._title,
 			style: HoverStyle.Pointer,
 		}), this._opts.hoverLifecycleOptions));
 		this.domNode.classList.add(...classes);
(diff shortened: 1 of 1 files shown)


src/vs/base/browser/ui/toggle/toggle.ts
```diff
@@ -153,7 +153,7 @@ export class Toggle extends Widget {
 
 		this.domNode = document.createElement('div');
 		this._register(getBaseLayerHoverDelegate().setupDelayedHover(this.domNode, () => ({
-			content: this._title,
+			content: { value: this._title, supportThemeIcons: true },
```
TylerLeonhardt: can you instead have the `title` options take in `string | IMarkdownString | HTMLElement`? since it's just used in the `content` property of the hover?

eidriahn: ah, this way the ones implementing the class are responsible for how the content is shown, much better.

eidriahn: @Tyriar I am not sure where the quick pick in your screenshot comes from, can't find it in vscode or vscode-copilot-chat. If that option is coming from one of those two repos let me know, if it's an option from an extension I guess they'll fix it.

src/vs/base/browser/ui/toggle/toggle.ts
```diff
@@ -246,9 +248,19 @@ export class Toggle extends Widget {
...
-	setTitle(newTitle: string): void {
+	setTitle(newTitle: string | IMarkdownString | HTMLElement): void {
 		this._title = newTitle;
-		this.domNode.setAttribute('aria-label', newTitle);
+
+		let ariaLabel = '';
+		if (typeof newTitle === 'string') {
+			ariaLabel = newTitle;
+		} else if (hasKey(newTitle, { value: true })) {
+			ariaLabel = newTitle.value;
```
TylerLeonhardt: Can you use this?

https://github.com/microsoft/vscode/blob/0551264cd7c263b02c18940126555b0acab6cfab/src/vs/base/common/iconLabels.ts#L39-L48

src/vs/base/browser/ui/toggle/toggle.ts
```diff
@@ -246,9 +248,19 @@ export class Toggle extends Widget {
...
 
-	setTitle(newTitle: string): void {
+	setTitle(newTitle: string | IMarkdownString | HTMLElement): void {
 		this._title = newTitle;
-		this.domNode.setAttribute('aria-label', newTitle);
+
+		let ariaLabel = '';
+		if (typeof newTitle === 'string') {
+			ariaLabel = newTitle;
+		} else if (hasKey(newTitle, { value: true })) {
```
TylerLeonhardt: Can you use this?

https://github.com/microsoft/vscode/blob/0551264cd7c263b02c18940126555b0acab6cfab/src/vs/base/common/htmlContent.ts#L126-L136

src/vs/base/browser/ui/toggle/toggle.ts
```diff
@@ -246,9 +248,19 @@ export class Toggle extends Widget {
...
+			ariaLabel = newTitle.value;
+		} else if (hasKey(newTitle, { textContent: true })) {
```
TylerLeonhardt: at this point we can then assume HTMLElement... or if you really wanna be clear, there's

https://github.com/microsoft/vscode/blob/0551264cd7c263b02c18940126555b0acab6cfab/src/vs/base/browser/dom.ts#L1012-L1015

src/vs/base/browser/ui/toggle/toggle.ts
TylerLeonhardt: Yeah I couldn't find that either. I am curious what @Tyriar thinks of this approach. How much massaging should we do in this hover? Should we strip/render the icons in the string case?

I was also thinking about the aria label as well... we for sure want to not have the icons in that... but right now it's inferred by the label.

src/vs/base/browser/ui/toggle/toggle.ts
eidriahn: opted for not using it

src/vs/base/browser/ui/toggle/toggle.ts
eidriahn: ✅

src/vs/base/browser/ui/toggle/toggle.ts
eidriahn: ✅

src/vs/base/browser/ui/toggle/toggle.ts
Tyriar: Either way is good with me

eidriahn: image

Opted for stripping the icons.

TylerLeonhardt (approved):
Awesome. Thanks for fixing this!

Tyriar (approved):

Issue #285320, opened 2025-12-29
decorationsEnabled takes space even when set to never


muhamedkarajic:
I would like to be able to get 10-20px on the left of my terminal below image demonstrates where and why

```json
"terminal.integrated.shellIntegration.decorationsEnabled": "gutter"
```

<img width="1403" height="458" alt="Image" src="https://github.com/user-attachments/assets/6c076301-2c49-40c9-bd8f-373edd805867" />

```json
"terminal.integrated.shellIntegration.decorationsEnabled": "never"
```

<img width="1512" height="501" alt="Image" src="https://github.com/user-attachments/assets/282f757f-e358-4d60-8f53-430d277daf70" />

___

P.S: Saw people talking here but no one raised a issue therefore I am raising it:
https://stackoverflow.com/questions/79848452/how-do-i-disable-the-left-and-right-side-empty-space-in-the-vs-code-integrated-t

PR #285325, opened 2025-12-29
decorationsEnabled takes space even when set to never (#285320)

muhamedkarajic:
I have not tested this, unfortunetly I got some issues on my machine. Could someone test it, how to reproduce stuff is written in https://github.com/microsoft/vscode/pull/285323.


dmitrivMS:
No issue is attached, the link goes to another PR.

PR #285365, opened 2025-12-29
Fix markdown content being split from thinking parts

### LINKED ISSUE #278557: Model response shows line break for no reason
Image

[255fdaa8.copilotmd.txt](https://github.com/user-attachments/files/23651958/255fdaa8.copilotmd.txt)

cc @roblourens

pwang347:
Fixes https://github.com/microsoft/vscode/issues/278557

### DIFF
--- a/src/vs/workbench/contrib/chat/common/chatModel.ts
+++ b/src/vs/workbench/contrib/chat/common/chatModel.ts
@@ -624,8 +624,9 @@ export class Response extends AbstractResponse implements IDisposable {
 
 			// last response which is NOT a text edit group because we do want to support heterogenous streaming but not have
 			// the MD be chopped up by text edit groups (and likely other non-renderable parts)
+			// Also skip thinking and progressMessage parts which can be interleaved between markdown fragments
 			const lastResponsePart = this._responseParts
-				.filter(p => p.kind !== 'textEditGroup')
+				.filter(p => p.kind !== 'textEditGroup' && p.kind !== 'thinking' && p.kind !== 'progressMessage')
 				.at(-1);
 
 			if (!lastResponsePart || lastResponsePart.kind !== 'markdownContent' || !canMergeMarkdownStrings(lastResponsePart.content, progress.content)) {
--- a/src/vs/workbench/contrib/chat/test/common/chatModel.test.ts
+++ b/src/vs/workbench/contrib/chat/test/common/chatModel.test.ts
@@ -244,6 +244,52 @@ suite('Response', () => {
 		assert.ok(!responseString.includes('Initial content'), 'Should not include content before clear');
 		assert.ok(responseString.endsWith('Made changes.'), 'Should end with "Made changes."');
 	});
+
+	test('markdown merging with thinking parts interleaved', async () => {
+		const response = store.add(new Response([]));
+		response.updateContent({ content: new MarkdownString('First markdown'), kind: 'markdownContent' });
+		response.updateContent({ kind: 'thinking', value: 'Thinking about something' });
+		response.updateContent({ content: new MarkdownString(' second markdown'), kind: 'markdownContent' });
+		response.updateContent({ kind: 'thinking', value: 'More thinking' });
+		response.updateContent({ content: new MarkdownString(' third markdown'), kind: 'markdownContent' });
+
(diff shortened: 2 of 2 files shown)


src/vs/workbench/contrib/chat/common/chatModel.ts
```diff
@@ -624,8 +624,9 @@ export class Response extends AbstractResponse implements IDisposable {
 
 			// last response which is NOT a text edit group because we do want to support heterogenous streaming but not have
 			// the MD be chopped up by text edit groups (and likely other non-renderable parts)
+			// Also skip thinking and progressMessage parts which can be interleaved between markdown fragments
 			const lastResponsePart = this._responseParts
-				.filter(p => p.kind !== 'textEditGroup')
+				.filter(p => p.kind !== 'textEditGroup' && p.kind !== 'thinking' && p.kind !== 'progressMessage')
 				.at(-1);
```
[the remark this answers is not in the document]
pwang347: @copilot open a new pull request to apply changes based on [this feedback](https://github.com/microsoft/vscode/pull/285365#discussion_r2651934011)

src/vs/workbench/contrib/chat/common/chatModel.ts
```diff
@@ -624,8 +624,9 @@ export class Response extends AbstractResponse implements IDisposable {
 
 			// last response which is NOT a text edit group because we do want to support heterogenous streaming but not have
 			// the MD be chopped up by text edit groups (and likely other non-renderable parts)
+			// Also skip thinking and progressMessage parts which can be interleaved between markdown fragments
```
roblourens, line 627: I'm not sure this is the right place for the fix. Assuming that the model didn't produce this thinking text right in the middle of the regular response text, wouldn't that mean that the response parts were re-ordered somewhere else upstream, and here we are trying to put them back in the right order? Or am I misunderstanding the problem?

pwang347: > Assuming that the model didn't produce this thinking text right in the middle of the regular response text

Isn't the problem that the thinking text is generated in the middle of the regular markdown by Gemini models which results in multiple markdown blocks (ie., a line break) to occur?
image
image

I think goal here is to make sure we never output two markdown blocks in these cases (note that we have merging logic already, this just makes it ignore thinking parts when it tries to merge with a previous markdown part)

roblourens: Does the model actually stop to produce more thinking in the middle of a sentence? The ChatModel parts are way downstream of a lot of async processing. If you look at the actual raw response coming from the model, is the content being produced in the same order?

pwang347: Hey thanks! Understood what you meant now.

For the following bug:
image

This is what it looks like in the raw request:
```
data: {"choices":[{"index":0,"delta":{"content":"And finally, here is the third large block of text. This concludes the attempt to reproduce the issue with larger blocks of interleaved content and thoughts. If the previous blocks rendered correctly, and this one also renders correctly, then the issue might be more subtle or dependent on specific content characters rather than just length. However, if","role":"assistant"}}],"created":1767120438,"id":"XXX","usage":{"completion_tokens":0,"prompt_tokens":0,"prompt_tokens_details":{"cached_tokens":0},"total_tokens":0,"reasoning_tokens":0},"model":"gemini-3-pro-preview"}
data: {"choices":[{"finish_reason":"stop","index":0,"delta":{"content":" **you see any visual glitches, missing text, or weird formatting, then we have successfully reproduced the issue. I will stop typing now.**","role":"assistant",**"reasoning_opaque":"XXX"**}}],"created":1767120438,"id":"XXX","usage":{"completion_tokens":368,"prompt_tokens":13368,"prompt_tokens_details":{"cached_tokens":10078},"total_tokens":14079,"reasoning_tokens":343},"model":"gemini-3-pro-preview"}
data: [DONE]
```

Specifically:
**you see any visual glitches, missing text, or weird formatting, then we have successfully reproduced the issue. I will stop typing now.**","role":"assistant",**"reasoning_opaque":"XXX"**

So seems like this happens when we have both reasoning and text in the same block. Exploring a solution in https://github.com/microsoft/vscode-copilot-chat/pull/2689

PR #285366, opened 2025-12-29
Add ThoughtSignature to chat content proposed API

vijayupadya:
Gemini 3 introduces ThoughtSignature which must be passed back during function calling, otherwise get a validation error (4xx status code). This change adds ThoughtSignature type to proposed API. This will enable consuming this type, needed for BYOK.

More info: https://ai.google.dev/gemini-api/docs/thought-signatures


vijayupadya (commented):
Submit

roblourens (commented):
How is this different from the opaque reasoning data which I believe we pack into LanguageModelThinkingPart today?

vijayupadya:
Ah, right. Didnt realize LanguageModelThinkingPart can contain arbitrary shaped objects. That will work. Changes in this PR not needed. Thanks!

PR #285407, opened 2025-12-30
Add `Collapse All` action to SCM view pane

### LINKED ISSUE #161929: SCM - Add action to collapse all resource groups for a repository
When all changed files are available under source control, if the user chooses to "View as Tree" a collapse all button should be available for the user to collapse all the tree directory structures, same as provided in explorer to collapse all the tree structures in the view.

Request to make available collapse all button here:
image

Example collapse all button available under explorer:
image

dmitrivMS:
Fixes #161929


lszomoru (dismissed):
Thanks for the pull request. Here is some feedback:
1. I think that the action should never collapse the resource group itself (that is what the twistie is for)
2. I think that the action should only be shown if the resources are being shown as a tree
3. Not sure about the position of the action, as it appears the leftmost action

joaomoreno (changes requested):
@copilot Found a bug when testing:
1. Have nested files in both staged and unstaged groups
2. Collapse All in the Staged group

🐛 Action will apply to all groups, even though it's located atop of each group. All folders in all groups will be collapsed.

joaomoreno (approved):
LGTM!

image

PR #285411, opened 2025-12-30
Add `Git: Delete` action to run  `git rm` command on the current document

### LINKED ISSUE #111767: Git - Add a command palette option for git rm
I'd like to be able to open the command palette and choose "Git: Delete Current File" which would do the equivalent of `git rm path` where `path` is the path to the current file.

A `git rm` will fail when the file is not currently committed. If it has uncommitted changes, or is not tracked by git, then it will fail. In this way, it's a "safer" form of delete, while avoiding the system trash folder.

dmitrivMS:
Fixes #111767

Added `Git: Delete` command which operates on the currently opened editor and runs `git rm` on the file, then closes the editor (if it is not dirty).

Added `git.confirmCommittedDelete` setting (on by default) to turn off the confirmation.

**NOTE** Since public API has no way to add a checkbox to the confirmation dialog, the setting can only be turned off from the settings file/UI.


extensions/git/src/commands.ts
```diff
@@ -1405,6 +1405,53 @@ export class CommandCenter {
...
+			}
+		}
+
+		await repository.rm([uri]);
+
+		// Close the active editor if it's not dirty
+		if (activeDocument?.uri.toString() === uriString && !activeDocument.isDirty) {
+			await commands.executeCommand('workbench.action.closeActiveEditor');
+		}
+	}
```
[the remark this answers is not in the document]
dmitrivMS: Most commands lack test coverage currently.

extensions/git/src/commands.ts
```diff
@@ -1405,6 +1405,53 @@ export class CommandCenter {
...
+		uri = uri ?? activeDocument?.uri;
+		if (!uri) {
+			return;
+		}
+
+		const repository = this.model.getRepository(uri);
+		if (!repository) {
+			return;
+		}
+
```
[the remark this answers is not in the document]
dmitrivMS: It's true if differs from `git.rename`, but rename's behavior in case of multiple repos is not ideal - it pops up repo picker every time even when focus is inside an editor for a file in the repo. `git.delete` behavior as implemented does not have this problem.

extensions/git/src/commands.ts
```diff
@@ -1405,6 +1405,53 @@ export class CommandCenter {
...
+		const allChangedResources = [
+			...repository.workingTreeGroup.resourceStates,
+			...repository.indexGroup.resourceStates,
+			...repository.mergeGroup.resourceStates,
+			...repository.untrackedGroup.resourceStates
+		];
+
+		// Check if file has uncommitted changes
+		const uriString = uri.toString();
+		if (allChangedResources.some(o => o.resourceUri.toString() === uriString)) {
```
lszomoru: You should use the pathEquals() utility method for this rather than doing a string compare.

extensions/git/src/commands.ts
```diff
@@ -1405,6 +1405,53 @@ export class CommandCenter {
...
+			window.showInformationMessage(l10n.t('Git: Delete can only be performed on committed files without uncommitted changes.'));
+			return;
+		}
+
+		// Show confirmation dialog if setting is enabled
+		const config = workspace.getConfiguration('git', Uri.file(repository.root));
```
lszomoru: Do we really need this setting? I think that we should always prompt, and only add it based on user feedback.

extensions/git/src/commands.ts
```diff
@@ -1405,6 +1405,53 @@ export class CommandCenter {
...
+			const choice = await window.showInformationMessage(message, { modal: true, detail }, deleteButton);
+			if (choice !== deleteButton) {
+				return;
+			}
+		}
+
+		await repository.rm([uri]);
+
+		// Close the active editor if it's not dirty
+		if (activeDocument?.uri.toString() === uriString && !activeDocument.isDirty) {
```
lszomoru: I think that we should check whether the active document is dirty before deleteting it and not after.

lszomoru (dismissed):

extensions/git/src/commands.ts
dmitrivMS: This is similar to Explorer delete functionality, but happy to remove if you feel we should wait.

extensions/git/src/commands.ts
dmitrivMS: This is not to prevent the user from deleting, but to be consistent with Explorer delete behavior where saved editors are closed upon deletion.

joaomoreno (approved):

PR #285453, opened 2025-12-30
Add --class command line parameter

### LINKED ISSUE #284420: Allow customizing WM_CLASS when launching vscode
Follow-up on #145455;

In Linux Window Managers such as Hyprland it is important to identify a window by its class (`WM_CLASS`) in order to apply rules (like positioning rules) to it.

Currently, the class is always `code` (or `code-oss` for the OSS version) which does not allow differenciating the windows from each other.
An example use would be `code --new-window --class mycustomclass`.

bagratte:
The parameter overrides the application name
(WM_CLASS).

Fixes #284420

### DIFF
--- a/src/vs/code/electron-main/main.ts
+++ b/src/vs/code/electron-main/main.ts
@@ -170,6 +170,16 @@ class CodeMain {
 
 		// Environment
 		const environmentMainService = new EnvironmentMainService(this.resolveArgs(), productService);
+
+		// Set custom application name for WM_CLASS/app_id when --class argument is provided
+		if (environmentMainService.args.class) {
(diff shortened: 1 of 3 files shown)


BlueskyFR:
Hey @bagratte! Thanks for this :P

I'm not sure how setting the window class for Electron works, but I'd just like to mention that it is important that the `initialClass` must be set, meaning the first class that is set should be the one passed to `--class`, and not **changed** later from `code` to the custom one, as some window managers such as Hyprland will match the first class that was set and not the subsequent ones.

Thanks again!

bagratte:
@BlueskyFR 
This should do but please test for your use case if you can.

PR #285467, opened 2025-12-31
Reorganize workbench/contrib/chat

### LINKED ISSUE #278766: Idea: More folder structure could improve overview
Personally, I believe more smaller files are usually better than fewer huge files.
However, if there are too many files, its also difficult to navigate the source code.
In contrib/chat/browser, there are currently almost 50 ts files:

Image

I suggest to group them by function or UI composition structure.

I can help here in debt week!

roblourens:
Fix #278766


bpasero (commented):
I find a `contrib` folder inside `contrib/chat` odd. Given its only few files, I would get rid of `contrib` and move the files into their respective folders in `contrib/chat`.

dmitrivMS (approved):

PR #285497, opened 2025-12-31
Feature/goto line preview toggle

MinuteReversal:
## Summary
- add the editor.gotoLine.autoReveal setting (window scope, default on) so users can disable live reveal while typing in Go to Line/Column
- persist the picker’s auto-reveal toggle in application storage whenever no explicit config is set
- add an eye-icon quick input button to flip auto reveal on demand while keeping the target-line decoration active

## Testing
- npm run compile
- ./scripts/code.sh


src/vs/editor/contrib/quickAccess/browser/gotoLineQuickAccess.ts
```diff
@@ -84,13 +187,23 @@ export abstract class AbstractGotoLineQuickAccessProvider extends AbstractEditor
 			toggle: { checked: this.useZeroBasedOffset }
 		};
 
+		const autoRevealButton: IQuickInputButton = {
+			iconClass: ThemeIcon.asClassName(Codicon.eye),
+			tooltip: localize('gotoLineAutoRevealToggleButton', "Toggle Auto Reveal"),
+			location: QuickInputButtonLocation.Input,
```
TylerLeonhardt: I'm not liking how prominent this toggle is for something that I will likely be setting once since it's a personal preference thing.

Like, I don't want this behavior personally (but understand that's _my_ preference) and will need to see this toggle all the time now... which in my theme stands out as the blue thing in the input box.

Instead, to not sacrifice discoverability, we should use "secondary" (which I just added as a concept during the holidays)... but unfortunately I didn't light up `secondary: true` for the Input location... but Inline will probably be fine for now. Can you show me what that looks like?

MinuteReversal: Thanks for the suggestion! I’ve moved the Auto Reveal toggle out of the input and into the inline toolbar, marking it as a secondary action so it tucks into the overflow unless needed. This keeps the preference easy to discover without the bright pill living inside the text box. Let me know if you’d still like to see further tweaks.

PR #285578, opened 2026-01-01
Use executable bit when completing commands in terminal suggest

### LINKED ISSUE #285577: Terminal suggest: Respect executable bit for command (non-arg) completions
https://github.com/microsoft/vscode/issues/282268#issuecomment-3703700802

Repro:

1. WSL remote, Linux or macOS
2. Open VS Code repo
3. Open terminal
4. `cd scripts`
5. Type `./` and hit ctrl+space to trigger suggest, 🐛 sh scripts should show up but not bat or js files.

Image

Verify executable bit by checking `x` on left when running `ls -la`

Tyriar:
Part of #285577

### DIFF
--- a/src/vs/platform/files/node/diskFileSystemProvider.ts
+++ b/src/vs/platform/files/node/diskFileSystemProvider.ts
@@ -90,6 +90,7 @@ export class DiskFileSystemProvider extends AbstractDiskFileSystemProvider imple
 				ctime: stat.birthtime.getTime(), // intentionally not using ctime here, we want the creation time
 				mtime: stat.mtime.getTime(),
 				size: stat.size,
+				permissions
 			};
 		} catch (error) {
 			throw this.toFileSystemProviderError(error);
--- a/src/vs/platform/files/node/diskFileSystemProvider.ts
+++ b/src/vs/platform/files/node/diskFileSystemProvider.ts
@@ -3,7 +3,7 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-import { Stats, promises } from 'fs';
+import { Stats, constants, promises } from 'fs';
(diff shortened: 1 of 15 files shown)


src/vs/workbench/test/common/workbenchTestServices.ts
```diff
@@ -245,7 +245,7 @@ export class TestWorkingCopy extends Disposable implements IWorkingCopy {
 	}
 }
```
mikeb42069: And stared at the wall and waited for you because you told me you would be here at 12 and I believed that  . 
Now I feel like a fool and I have your $$ come get it I guess People wanted me to come with them and I could've had a good time maybe and right now I really needed something to my mind off of other things but I sat her

src/vs/platform/files/node/diskFileSystemProvider.ts
```diff
@@ -73,12 +73,24 @@ export class DiskFileSystemProvider extends AbstractDiskFileSystemProvider imple
...
+				permissions = (permissions ?? 0) | FilePermission.Executable;
+			}
+
 			return {
 				type: this.toType(stat, symbolicLink),
 				ctime: stat.birthtime.getTime(), // intentionally not using ctime here, we want the creation time
 				mtime: stat.mtime.getTime(),
 				size: stat.size,
-				permissions: (stat.mode & 0o200) === 0 ? FilePermission.Locked : undefined
+				permissions: permissions ?? undefined
```
bpasero:
```suggestion
				permissions
```

bpasero (changes requested):
Looks good to me generally, but any chance we could add a test for this to the `src/vs/platform/files/test/node/diskFileService.integrationTest.ts` suite?. I could think of running it only on POSIX and tampering with the mode on a temporary file and then checking if the relevant executable flag is coming back properly.

bpasero (approved):

PR #285592, opened 2026-01-01
Update swipeToNavigate Commands (microsoft#285591)

kh0m:
This PR updates the commands that are called when using swipeToNavigate.

Steps to Reproduce:

1. click the back arrow, and the forward arrow suggests I can go forward
2. swipe forward, and nothing happens
3. swipe backwards, and it brings you forward in history, but to a previous editor
4. swipe forwards again, and the arrow doesn't reappear even though I am at the same place as before
5. now I don't know where I am in the history


dmitrivMS:
@kh0m Is there an issue tracking what you are fixing/implementing?

PR #285602, opened 2026-01-01
Fixes race condition where incomplete state of ViewModelLines could be observed by events fired during model access.

dmitrivMS:
## Problem description
During _constructLines(), the modelLineProjections array was emptied (this.modelLineProjections = [])
While reconstructing (40+ lines of code), synchronous callbacks could fire from model operations
These callbacks tried to access modelLineProjections[index] which was empty/incomplete
Result: Cannot read properties of undefined (reading 'getViewLineMinColumn'|'getViewLineContent'|'getViewLineMaxColumn')


bpasero:
@alexdima can you review this while Henning is out? The issues are among the top crashes in error telemetry as far as I can tell. 

image

hediet (approved):
Looks good! However, I'm not sure it fixes these referenced issues, as I didn't see `_constructLines` in the stack trace.
I also don't know which event could be fired within that method.

dmitrivMS:
Closing this since it does not address any of the existing problem reports. Will continue looking for the repro case(es).

PR #285631, opened 2026-01-02
fix: show restart extensions notification when disabling from Running Extensions view

### LINKED ISSUE #250201: Restart Extensions button not created when disabling extensions from Running Extensions view

- VS Code Version: 1.99.0
- OS Version: Windows 11

### The problem

When a running extension is disabled from the View->Extensions view, a Restart Extensions button is shown to notify the user that the Extensions must be restarted to disable running extensions.

But when an extension is disabled from the Running Extensions view, this Restart Extensions button is not shown.

Steps to Reproduce:

1.  Go to View->Extensions.
2.  In the Extensions view go to the Views and More Actions menu by clicking on the three dots (...)  in the top right corner of the Extension view.
3. Click on Show Running Extensions
4. Hide the Extensions view by going to the View->Explorer view
5. Go to the Running Extensions view
6. Right click on a running extension and click Disable
7. Notice that the extension is still running and there is no indication that the extensions need to be restarted.

### Proposed solution 1

Copy the behavior of the View->Extenions view and show a Reload Extensions button when an extension is disabled from the Running Extensions View.

### Proposed solution 2

Why is there a separate Running Extensions view in the first place? It only shows the running extensions, and their activation time. There is no separate view for installed, disabled, outdated or populair extensions. These are shown by an "@"filter in the extension search box. I propose to create an @<!-->running filter in the View->Extensions view and remove the Running Extensions view at all. It adds no value, creates this bug, and could create new bugs in the future when new behavior is added to the View->Extensions view and this behavior is not copied to the Running Extensions view, which is the cause of this bug.

When this solution is implemented, the 2 ways for viewing the Running Extensions need to be updated as well:
1. the Show Running Extensions option from step 3 of the reproduction steps above.
2. command "Developer: Show Running Extensions" from the command palette

murataslan1:
## Summary

When an extension is disabled from the Running Extensions view context menu, the user was not notified that extensions need to be restarted for the change to take effect. This was inconsistent with the Extensions view behavior.

## Changes

Added a notification with a 'Restart Extensions' button that appears after disabling an extension from the Running Extensions view. This matches the behavior of the main Extensions view.

### Modified File
- `src/vs/workbench/contrib/extensions/browser/abstractRuntimeExtensionsEditor.ts`

## Before
1. Go to Running Extensions (Developer: Show Running Extensions)
2. Right-click on an extension → Disable
3. No notification appears, user doesn't know a restart is needed

## After
1. Go to Running Extensions
2. Right-click on an extension → Disable
3. A notification appears: "The extension has been disabled. Please restart extensions to apply the change."
4. User can click "Restart Extensions" button to apply immediately

Fixes #250201

### DIFF
--- a/src/vs/workbench/contrib/extensions/browser/abstractRuntimeExtensionsEditor.ts
+++ b/src/vs/workbench/contrib/extensions/browser/abstractRuntimeExtensionsEditor.ts
@@ -487,8 +487,28 @@ export abstract class AbstractRuntimeExtensionsEditor extends EditorPane {
 			actions.push(new Separator());
 
 			if (e.element.marketplaceInfo) {
-				actions.push(new Action('runtimeExtensionsEditor.action.disableWorkspace', nls.localize('disable workspace', "Disable (Workspace)"), undefined, true, () => this._extensionsWorkbenchService.setEnablement(e.element!.marketplaceInfo!, EnablementState.DisabledWorkspace)));
-				actions.push(new Action('runtimeExtensionsEditor.action.disable', nls.localize('disable', "Disable"), undefined, true, () => this._extensionsWorkbenchService.setEnablement(e.element!.marketplaceInfo!, EnablementState.DisabledGlobally)));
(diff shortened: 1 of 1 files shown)


sandy081:
Closing this PR because the associated feature request is closed

PR #285636, opened 2026-01-02
fix: prevent duplicate disable actions in extension dropdown

murataslan1:
## Summary

When an extension is in the state:
- Disabled Globally
- Enabled Workspace

The "Disable" button dropdown would show both:
- Disable
- Disable (Workspace)

The "Disable" action (global) should not be shown because the extension is already disabled globally. The "Disable (Workspace)" action is the correct one to disable the workspace override.

## Change

Modified `DisableGloballyAction.update()` in `src/vs/workbench/contrib/extensions/browser/extensionsActions.ts` to explicitly check if the extension is already disabled globally.

If it is already disabled globally, the action becomes disabled.

Fixes #244138


huangxinyue8616 (commented):
lgtm

PR #285657, opened 2026-01-02
Fix: Replace terminal tabbed view lifecycle hack with Event.once

murataslan1:

### DIFF
--- a/src/vs/workbench/contrib/terminal/browser/terminalTabbedView.ts
+++ b/src/vs/workbench/contrib/terminal/browser/terminalTabbedView.ts
@@ -592,10 +592,8 @@ export class TerminalTabbedView extends Disposable {
 		// be focused. So wait for connection to finish, then focus.
 		const previousActiveElement = this._tabListElement.ownerDocument.activeElement;
 		if (previousActiveElement) {
-			// TODO: Improve lifecycle management this event should be disposed after first fire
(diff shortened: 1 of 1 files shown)


src/vs/workbench/contrib/terminal/browser/terminalTabbedView.ts
```diff
@@ -592,14 +592,13 @@ export class TerminalTabbedView extends Disposable {
 		// be focused. So wait for connection to finish, then focus.
 		const previousActiveElement = this._tabListElement.ownerDocument.activeElement;
 		if (previousActiveElement) {
-			// TODO: Improve lifecycle management this event should be disposed after first fire
-			this._register(this._terminalService.onDidChangeConnectionState(() => {
+			const listener = Event.once(this._terminalService.onDidChangeConnectionState)(() => {
```
Tyriar: Even better would be to register on the same line:

```ts
const listener = this._register(Event.once(...
```

Then dispose the listener inside the handler and also release the handle in `this._store`:

```ts
if (dom.isActiveElement(previousActiveElement)) {
	this._focus();
}
this._store.delete(listener)
```

Even though the once happens we should still explicitly dispose of it and release the handle or it would never be garbage collected.

Tyriar (changes requested):

Tyriar (approved):

lszomoru (approved):

PR #285661, opened 2026-01-02
Fix: Replace terminal view lifecycle hack with Event.once

murataslan1:
Identical to the fix in terminalTabbedView.ts: replaces a manual 'TODO' event listener in terminalView.ts with a robust Event.once pattern. Ensures proper disposal and prevents memory leaks during terminal connection state changes.

### DIFF
--- a/src/vs/workbench/contrib/terminal/browser/terminalView.ts
+++ b/src/vs/workbench/contrib/terminal/browser/terminalView.ts
@@ -337,10 +337,8 @@ export class TerminalViewPane extends ViewPane {
 		// be focused. So wait for connection to finish, then focus.
 		const previousActiveElement = this.element.ownerDocument.activeElement;
 		if (previousActiveElement) {
-			// TODO: Improve lifecycle management this event should be disposed after first fire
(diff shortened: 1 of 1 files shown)


src/vs/workbench/contrib/terminal/browser/terminalView.ts
```diff
@@ -337,10 +337,8 @@ export class TerminalViewPane extends ViewPane {
 		// be focused. So wait for connection to finish, then focus.
 		const previousActiveElement = this.element.ownerDocument.activeElement;
 		if (previousActiveElement) {
-			// TODO: Improve lifecycle management this event should be disposed after first fire
-			this._register(this._terminalService.onDidChangeConnectionState(() => {
+			this._register(Event.once(this._terminalService.onDidChangeConnectionState)(() => {
```
Tyriar: Same comment as https://github.com/microsoft/vscode/pull/285657

Tyriar (changes requested):

Tyriar (approved):

lszomoru (approved):

PR #285732, opened 2026-01-03
Add policies for more terminal chat settings

### LINKED ISSUE #284302: Allow MDM to prevent modifying list of allow/denied commands for chat auto-approval
Currently it is possible to globally allow or prevent auto-approval of chat actions. This kind of broad-strokes approach doesn't provide sufficient balance between security and usability. There is a setting, `chat.tools.terminal.autoApprove`, which accepts a list of commands that are allowed or not allowed. This comes with safe defaults, but a user can override those values (for example, allowing the `rm` command to be auto-approved) which undermines the safety this setting is intended to provide. It would be great if there was a way to rely on the default values (or custom values) and to guarantee that a user cannot override those values.

Tyriar:
Fixes #284302

@joshspicer thoughts on these policies? There is also `chat.tools.terminal.autoReplyToPrompts` (experimental) and `chat.tools.terminal.blockDetectedFileWrites` (experimental) which are potentially risky. It's not clear to me how defaults work when experimental features are disabled as the "safest" value for `blockDetectedFileWrites` is not the default.


build/lib/test/fixtures/policies/darwin/com.visualstudio.code.oss.mobileconfig
```diff
@@ -25,14 +25,20 @@
 				<false/>
 				<key>ChatToolsAutoApprove</key>
 				<false/>
+				<key>ChatToolsEligibleForAutoApproval</key>
+				<string></string>
```
Tyriar, line 29: @pwang347 this isn't present on main? 😕

build/lib/test/fixtures/policies/linux/policy.json
```diff
@@ -4,10 +4,13 @@
     "ChatMCP": "none",
     "ChatPromptFiles": false,
     "ChatToolsAutoApprove": false,
+    "ChatToolsEligibleForAutoApproval": "",
```
Tyriar, line 7: @pwang347 should this be the default JSON value that was generated when I ran `./scripts/code-cli.bat -- export-policy-data`?

build/lib/test/fixtures/policies/darwin/en-us/com.visualstudio.code.oss.plist
```diff
@@ -122,7 +122,7 @@
 <key>pfm_default</key>
```
pwang347, line 122: Just a heads up, the goal of the fixtures here is mostly to validate the generator behaviour on the data prior to refactor (and not necessarily the live version of the policy data; otherwise we could've just checked those in as well). Last time @joshspicer ran into this as well when asking Copilot to add a policy, we may need instructions to avoid having it modify it like this.

build/lib/test/fixtures/policies/linux/policy.json
pwang347: see comment above, the fixtures here are not meant to be updated when adding policies

build/lib/test/fixtures/policies/darwin/en-us/com.visualstudio.code.oss.plist
Tyriar: Not sure I understand, I was getting these failures: https://github.com/microsoft/vscode/actions/runs/20678051529/job/59368384249

I thought it would work like regenerating the files so they can be reviews carefully in PR?

pwang347: The only file that needs to be regenerated and reviewed is `policyData.jsonc`, which is then converted in product build pipeline into the OS-specific formats (the tests here are to make sure the conversions are as expected). I decided against checking in the OS-specific formats directly since it will bloat PR diffs and we need to regenerate those files anyways to account for product.json values and translations, so having it generated and checked-in could be misleading.

pwang347: that being said, if we feel like the policy being added is different from existing types (e.g., new format) then adding tests like this is the right thing to do

build/lib/test/fixtures/policies/linux/policy.json
pwang347: yes, I think this is due to https://github.com/microsoft/vscode/issues/277319 which was existing behaviour but something we should improve

build/lib/test/fixtures/policies/darwin/com.visualstudio.code.oss.mobileconfig
pwang347: (see other thread)

PR #285864, opened 2026-01-05
chat: support MCP apps

### LINKED ISSUE #260218: Built-in support for MCP-UI
MCP-UI is [gaining](https://shopify.engineering/mcp-ui-breaking-the-text-wall) adoption. We have proposed API for [extension-contributed tool rendering](https://github.com/microsoft/vscode/issues/257761), and we should consider including a built-in extension that would render MCP UI like we do for notebook rendering scenarios.

- Spec https://mcpui.dev/guide/embeddable-ui
- Repo https://github.com/idosal/mcp-ui

I still need to dive more into the spec to see feasibilty/challenges of this in our application.

cc @digitarald @mjbvz

connor4312:
Implements the [MCP apps](https://github.com/modelcontextprotocol/ext-apps) proposal. Unlike MCP UI, some deeper integration is needed so I opted to do it entirely within core. Controlled by the `chat.mcp.apps.enabled` setting which is disabled by default for the moment.

This supports all capabilities of MCP apps to the best I can tell, and all of their examples work (barring https://github.com/modelcontextprotocol/ext-apps/issues/199). Some notable changes outside of the new MCP stuff:

- There is a new `onDidRemount` method of chat parts. @mjbvz you may prefer to adopt this instead of what's [currently there](https://github.com/microsoft/vscode/blob/483e5be515177eba8e45cf167502b9d94a3a95d1/src/vs/workbench/contrib/chat/browser/widget/chatContentParts/toolInvocationParts/chatToolOutputPart.ts#L134-L140).
- I added a new optional `container` to `createWebviewOverlay`. The way I arrived on this was:
    1. webview elements alone are not the best in chat because they flicker when parts get re-rendered and they get re-created
    2. overlay webviews fix this, but they do not natively track the scroll position
    3. rather that adding a `scroll` listener and manually clipping to the container, we can make this work by just letting the overlay webview explicitly parent into the scroll container
    
  Though if we think this is not good I could also try and dive deeper into the rerenders in (i) to see what's happening there, but I'm unsure how readily solvable that will be 😛 

https://github.com/user-attachments/assets/0eb204ca-da62-472d-bd0d-aee96ff20a60

Closes https://github.com/microsoft/vscode/issues/260218


src/vs/workbench/contrib/chat/browser/widget/chatContentParts/toolInvocationParts/chatMcpAppSubPart.ts
```diff
@@ -0,0 +1,193 @@
...
+		// Acquire model from pool
+		this._model = this._context.mcpAppWebviewPool.acquire(toolInvocation, renderData);
+
+		// Create the DOM structure
+		this.domNode = dom.$('div.mcp-app-part');
+		this._webviewContainer = this._createWebviewContainer();
+		this.domNode.appendChild(this._webviewContainer);
+
+		// Claim the webview and layout over our container
+		this._model.claim(this, this.domNode);
```
mjbvz: Can you use `createWebviewElement` instead? The overlay should only be used if you need to avoid re-parenting the webview while keeping it alive. Using instead of an inline element makes layout and dom flow for stuff like accessibility much more difficult

connor4312: I mentioned this in my original comment https://github.com/microsoft/vscode/pull/285864#issue-3780235401

though after implementing I realized that the flickering might have been due to some other logic that fixed (in the render() call in the tool invocation part.) I'll revisit that and see if I can simplify

karthiknadig (approved):

Issue #280530, opened 2025-12-02
Ask for a whole metadata of MCP server tool in VS Code API (such as `vscode.lm.tools`)


Alive-Fish:
We are developing an extension (ATK) and we want to list all metadata of MCP server tools. I have found `vscode.lm.tools` which contains all the started tools, but it just contains some fields without all fields of a tool in MCP server.

<img width="888" height="805" alt="Image" src="https://github.com/user-attachments/assets/e7d3d261-1bf7-4ba9-a4ae-2bdd56c76f03" />

<img width="1428" height="703" alt="Image" src="https://github.com/user-attachments/assets/20d51d82-4968-403d-9f4e-1e76158378a3" />

Is there any way to get whole metadata of tool or is it possible to contain them in `LanguageModelToolInformation`? Very thanks!

connor4312:
We've no immediate plans to add this.

What's your use case for it?

Alive-Fish:
Hi @connor4312 , I have drafted a Pull Request #285882 for implementing it. Could you help to review it or give us some suggestions? We can make an internal discussion from Teams or Outlook, if needed. Thanks!

Cc. @SuryaMSFT , @pierceboggan , @wh-alice , @summzhan

connor4312:
This is available in the latest Insiders via [this](https://github.com/microsoft/vscode/blob/main/src/vscode-dts/vscode.proposed.mcpServerDefinitions.d.ts) API proposal. Let me know if you have an extension that you need allowlisted to use the proposal.

Info on using proposed API: https://code.visualstudio.com/api/advanced-topics/using-proposed-api

summzhan:
Hi @connor4312, thanks for rough ETA. Just kindly check whether you have more concrete ETA to share or design draft since it's been a week now and not sure the latest status

Alive-Fish:
Hi @connor4312 , could you help to add our extension `teamsdevapp.ms-teams-vscode-extension` to the allow list of the proposed API `mcpServerDefinitions`. Very appreciate!

Here is our usage: [feat: use startMcpGateway to get full definitions of tools by Alive-Fish · Pull Request #15347 · Of…](https://github.com/OfficeDev/microsoft-365-agents-toolkit/pull/15347/changes#diff-42f13443f1a9998d296d301460f278e1b9140e80be43195d50dffecdff8f01adR254)

PR #285882, opened 2026-01-05
feat: add fields to LanguageModelToolInformation for full definitions of MCP tools

Alive-Fish:
Issue: https://github.com/microsoft/vscode/issues/280530

Summary: this PR is to expose some fields of tools from MCP servers to extension users. [Our team's extension (Agents Toolkit)](https://github.com/OfficeDev/microsoft-365-agents-toolkit) wants to get the full definitions of tools from MCP servers, which can be directly used for declarative agents in M365.

image


muthu-rathinam:
To add more details on why this is an important requirement for MCP based DAs. Now we are currently utilizing the complete MCP tools details for DAs
* This helps the LLMs to make decisions clearly based on the tool's complete input and output schema.
* Annotations defines the tools nature, basically is it consequential (involves modifying data) or non-consequential (read only), this  helps in informing the user about the kind of operation that DA is going to make and potentially provide options like always allow (for read only) kind of operations to avoid unnecessary pauses in the DA's process.
* Meta property has the complete set of information for [Apps SDK ](https://developers.openai.com/apps-sdk/)related features that we are introducing for DAs.


cc: @Alive-Fish

Issue #263525, opened 2025-08-27
Speed up `NativeEditContext _updateSelectionAndControlBounds` for cases when a lot of editors are created


mjbvz:
As part of #263499, I'm trying to speed up rendering of chat responses. When a response contains a large number of code blocks, we end up creating a large number of editors for it. One slow part of this is the `NativeEditContext. _updateSelectionAndControlBounds` call after the editor is created

Here you can see a few of these calls:

<img width="686" height="308" alt="Image" src="https://github.com/user-attachments/assets/91cefd08-0b19-4dfd-8194-4433c2ce6891" />

This call is slow because it forces a recalculate styles + relayout

In my regular chat session log, we end up spending several hundred ms just on these calls when you first open the session log:

<img width="1071" height="838" alt="Image" src="https://github.com/user-attachments/assets/f8432634-7d01-4c1d-998b-a6bb6decc633" />

Would it be possible to avoid this call until there is a hint the NativeContext is needed, such as the editor becoming focused?

aiday-mar:
That is a good point, thanks for letting me know. I have to look into it.

PR #285906, opened 2026-01-05
Optimize rendering performance by scheduling DOM updates at the next animation frame in NativeEditContext and TextAreaEditContext

tamuratak:
Optimize rendering performance by scheduling DOM updates at the next animation frame in NativeEditContext and TextAreaEditContext.  Related to #263525 and #246523.

This is one of several PRs aimed at improving smooth scrolling in the chat view, even when it contains many code blocks.  `NativeEditContext._updateSelectionAndControlBoundsAfterRender` and `TextAreaEditContext._render` are one of the causes of reflow during chat view scrolling as mentioned in #263525. In this PR, reflow is avoided by deferring the actual method call to the next animation frame. There have been many complaints about scrolling in the chat view, so I think optimizations like this are well justified.

When this PR not merged:
editcontext_not_merged

When this PR not merged (`editor.editContext: false`):
editcontext_not_merged2

When all PRs merged:
allthree_merged


aiday-mar (approved):

joaomoreno (approved):

alexdima:
The render call for all ViewParts in the editor should already happen during animation frames, unless someone uses editor APIs that would cause forced synchronous rendering. I think I would like to be able to replay the repro to understand what is the root cause of this synchronous rendering.

aiday-mar:
Hi @tamuratak thank you for the PR. The issue you linked filed by my colleague and the performance profiles seems to suggest the slow performance is due to a recalculate styles + relayout calls triggered by the `_updateSelectionAndControlBoundsAfterRender`, not by the code being executed outside of an animation frame. It would seem that to improve the performance, what we could do instead is execute `_updateSelectionAndControlBoundsAfterRender` when the editor is focused for example, as suggested by my colleague. 

I am not sure we are forcing a synchronous rendering on the editor.

PR #285950, opened 2026-01-05
Allow writing prompt with voice when session is in progress (fix #285852)

### LINKED ISSUE #285852: Allow writing prompt with voice when session is in progress
When a chat session is running, the voice/mic button gets disabled but the text input box stays enabled. I can still type my next prompt while waiting for the response but I can't use Voice to archive that. The voice button should work the same way IMHO.

bpasero:


lramos15 (approved):

src/vs/workbench/contrib/chat/electron-browser/actions/voiceChatActions.ts
```diff
@@ -525,10 +520,7 @@ export class QuickVoiceChatAction extends VoiceChatWithHoldModeAction {
 			id: QuickVoiceChatAction.ID,
 			title: localize2('workbench.action.chat.quickVoiceChat.label', "Quick Voice Chat"),
 			category: CHAT_CATEGORY,
-			precondition: ContextKeyExpr.and(
```
asrar-mared: الملفات المتغيره

PR #285985, opened 2026-01-05
implement chat terminal streaming

### LINKED ISSUE #286593: Auto-expand chat terminal output on terminal command execution, collapse on success
We'll want a setting to configure this. Idea from @Tyriar

meganrogge:
fixes #271390
fixes https://github.com/microsoft/vscode/issues/286593

- Shows the output action `onCommandExecuted`
- `onData` we track cursor movement and dirty ranges so we can append new VT sequences instead of rewriting the full buffer.
- Allows providing vt sequence range even when start or end marker has been disposed to support cases where the output exceeds scrollback
- Expands on command execution and collapses on success (or stays expanded on failure)
- Enables writing input to the terminal within chat - esp nice for replying to prompts for input

https://github.com/user-attachments/assets/431460f9-706a-4bf3-80e1-f387f3573a0a


src/vs/workbench/contrib/terminal/browser/xterm/xtermTerminal.ts
```diff
@@ -912,22 +911,28 @@ export class XtermTerminal extends Disposable implements IXtermTerminal, IDetach
 		this._onDidRequestRefreshDimensions.fire();
 	}
```
meganrogge, line 923: start markers can be disposed due to scrollback - in that case, we want to start from 0. I ran into this when running `tree` in the terminal.

src/vs/workbench/contrib/terminal/browser/chatTerminalCommandMirror.ts
```diff
@@ -132,61 +112,242 @@ export async function getCommandOutputSnapshot(
...
-			colorProvider: {
-				getBackgroundColor: theme => getChatTerminalBackgroundColor(theme, this._contextKeyService),
-			},
-		}));
+	}
+
+	override dispose(): void {
+		this._stopStreaming();
+		super.dispose();
 	}
```
Tyriar: Register a `toDisposable` instead of overriding dispose

src/vs/workbench/contrib/terminal/browser/chatTerminalCommandMirror.ts
```diff
@@ -132,61 +112,242 @@ export async function getCommandOutputSnapshot(
...
+	private _lastUpToDateCursorY: number | undefined;
+	private _lowestDirtyCursorY: number | undefined;
+	private _flushPromise: Promise<void> | undefined;
+	private _dirtyScheduled = false;
+	private _isStreaming = false;
+	private _sourceRaw: RawXtermTerminal | undefined;
+
 	constructor(
-		private readonly _xtermTerminal: XtermTerminal,
+		private readonly _terminalInstance: ITerminalInstance,
```
Tyriar: Looking at the uses of this I think we could keep it to `XtermTerminal`? Bringing in `ITerminalInstance` increases the complexity quite a bit as that interface does so much.

src/vs/workbench/contrib/terminal/browser/chatTerminalCommandMirror.ts
```diff
@@ -132,61 +112,242 @@ export async function getCommandOutputSnapshot(
...
+
+	private _lastVT = '';
+	private _lineCount = 0;
+	private _lastUpToDateCursorY: number | undefined;
+	private _lowestDirtyCursorY: number | undefined;
+	private _flushPromise: Promise<void> | undefined;
+	private _dirtyScheduled = false;
+	private _isStreaming = false;
+	private _sourceRaw: RawXtermTerminal | undefined;
+
```
Tyriar, line 167:
```suggestion

	private _lastVT = '';
	private _lineCount = 0;
	private _lastUpToDateCursorY: number | undefined;
	private _lowestDirtyCursorY: number | undefined;
	private _flushPromise: Promise<void> | undefined;
	private _dirtyScheduled = false;
	private _isStreaming = false;
	private _sourceRaw: RawXtermTerminal | undefined;

	private readonly _onDidUpdateEmitter = this._register(new Emitter<number>());
	public readonly onDidUpdate = this._onDidUpdateEmitter.event;

```

src/vs/workbench/contrib/terminal/browser/chatTerminalCommandMirror.ts
```diff
@@ -132,61 +112,242 @@ export async function getCommandOutputSnapshot(
...
 		}
-		const detached = await this._getTerminal();
-		await new Promise<void>(resolve => {
-			detached.xterm.write(vt.text, () => resolve());
-		});
+
+		this._lastVT = vt.text;
+		this._lineCount = vt.lineCount;
+
+		const xterm = await this._terminalInstance.xtermReadyPromise;
```
Tyriar: If possible to pass in `_xtermTerminal` in the ctor we can avoid the async and undefined handling here.

src/vs/workbench/contrib/terminal/browser/chatTerminalCommandMirror.ts
```diff
@@ -132,61 +112,242 @@ export async function getCommandOutputSnapshot(
...
+
+	private async _getOrCreateTerminal(): Promise<IDetachedTerminalInstance> {
+		if (this._detachedTerminal) {
+			return this._detachedTerminal;
+		}
+		const targetRef = this._terminalInstance?.targetRef ?? new ImmortalReference<TerminalLocation | undefined>(undefined);
+		const colorProvider = this._instantationService.createInstance(TerminalInstanceColorProvider, targetRef);
+		const detached = await this._terminalService.createDetachedTerminal({
+			cols: this._terminalInstance?.cols ?? 80,
+			rows: 10,
```
Tyriar: `MirrorRowCount`, `MirrorColCountFallback` in const enum?

src/vs/workbench/contrib/terminal/browser/chatTerminalCommandMirror.ts
Tyriar: Can you add some tests here? Tests will be much easier if we pass in XtermTerminal

Tyriar (dismissed):
Looking great!

src/vs/workbench/contrib/terminal/browser/chatTerminalCommandMirror.ts
meganrogge: We need it for this 

https://github.com/microsoft/vscode/blob/ed10bcfe417ba6ecc73c74a43ac7105f47f47324/src/vs/workbench/contrib/terminal/browser/chatTerminalCommandMirror.ts#L223-L227

Tyriar: That just determines the terminal's background color, so using that is wrong. You want to make your own color provider instead of using `targetRef` that uses the right color theme for the chat view.

src/vs/workbench/contrib/terminal/browser/chatTerminalCommandMirror.ts
```diff
@@ -132,61 +115,357 @@ export async function getCommandOutputSnapshot(
...
+
+		if (vt.text === this._lastVT) {
+			this._lastUpToDateCursorY = currentCursor;
+			if (this._command.endMarker && !this._command.endMarker.isDisposed) {
+				this._stopStreaming();
+			}
+			return;
+		}
+
+		const canAppend = !!this._lastVT && startLine >= previousCursor;
```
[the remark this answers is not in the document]
meganrogge: @tyriar I think this would be a perf problem and unnecessary. WDYT?

src/vs/workbench/contrib/chat/browser/widget/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
```diff
@@ -858,9 +881,14 @@ class ChatTerminalToolOutputSection extends Disposable {
...
 			return false;
 		}
-		this._mirror = this._register(this._instantiationService.createInstance(DetachedTerminalCommandMirror, liveTerminalInstance.xterm!, command));
-		await this._mirror.attach(this._terminalContainer);
-		const result = await this._mirror.renderCommand();
+		const mirror = this._register(this._instantiationService.createInstance(DetachedTerminalCommandMirror, liveTerminalInstance.xterm, command));
+		this._mirror = mirror;
+		this._register(mirror.onDidUpdate(lineCount => {
+			this._layoutOutput(lineCount);
+			this._scrollOutputToBottom();
```
Tyriar: I think we want the standard behavior of locking scroll to bottom only when it's scrolled all the way down? We want to allow the user to scroll up an in progress command without it scrolling back down automatically.

src/vs/workbench/contrib/terminal/browser/xterm/xtermTerminal.ts
```diff
@@ -912,22 +912,26 @@ export class XtermTerminal extends Disposable implements IXtermTerminal, IDetach
...
 			const Addon = await this._xtermAddonLoader.importAddon('serialize');
 			this._serializeAddon = new Addon();
 			this.raw.loadAddon(this._serializeAddon);
 		}
 
-		assert(startMarker.line !== -1);
-		let end = endMarker?.line ?? this.raw.buffer.active.length - 1;
-		if (skipLastLine) {
+		const hasValidEndMarker = typeof endMarker?.line === 'number' && endMarker.line >= 0;
+		const start = startMarker?.line ?? 0;
```
Tyriar: `startMarker.line` could be `-1`

src/vs/workbench/contrib/terminal/browser/xterm/xtermTerminal.ts
```diff
@@ -912,22 +912,26 @@ export class XtermTerminal extends Disposable implements IXtermTerminal, IDetach
...
 		if (!this._serializeAddon) {
 			const Addon = await this._xtermAddonLoader.importAddon('serialize');
 			this._serializeAddon = new Addon();
 			this.raw.loadAddon(this._serializeAddon);
 		}
 
-		assert(startMarker.line !== -1);
-		let end = endMarker?.line ?? this.raw.buffer.active.length - 1;
-		if (skipLastLine) {
+		const hasValidEndMarker = typeof endMarker?.line === 'number' && endMarker.line >= 0;
```
Tyriar: `isNumber` helper

src/vs/workbench/contrib/terminal/browser/xterm/xtermTerminal.ts
```diff
@@ -912,22 +912,26 @@ export class XtermTerminal extends Disposable implements IXtermTerminal, IDetach
...
+		const start = startMarker?.line ?? 0;
+		let end = hasValidEndMarker ? endMarker!.line : this.raw.buffer.active.length - 1;
+		if (skipLastLine && hasValidEndMarker) {
 			end = end - 1;
 		}
+		if (end < start) {
+			end = start;
+		}
```
Tyriar: nit: I think this ends up being faster at runtime typically, plus it's nicer imo:

```suggestion
		end = Math.max(end, start)
```

src/vs/workbench/contrib/terminal/browser/terminal.ts
```diff
@@ -1476,6 +1476,19 @@ export interface IDetachedXtermTerminal extends IXtermTerminal {
 	 * Resizes the terminal.
 	 */
 	resize(columns: number, rows: number): void;
+
+	/**
+	 * Access to the terminal buffer for reading cursor position and content.
+	 */
+	readonly buffer: {
```
Tyriar: If we're keeping this you can make it `IBufferSet` from xterm's api

src/vs/workbench/contrib/terminal/browser/chatTerminalCommandMirror.ts
```diff
@@ -132,61 +115,367 @@ export async function getCommandOutputSnapshot(
...
-	attach(container: HTMLElement): Promise<void>;
-	renderCommand(): Promise<{ lineCount?: number } | undefined>;
-}
-
 /**
  * Mirrors a terminal command's output into a detached terminal instance.
  * Used in the chat terminal tool progress part to show command output for example.
+ *
+ * Streaming approach
+ * ------------------
```
Tyriar: Since these are internal implementation details they should live inside the class something, probably in a `//` command rather than jsdoc so it doesn't show up in autocomplete.

src/vs/workbench/contrib/terminal/browser/chatTerminalCommandMirror.ts
```diff
@@ -132,61 +115,367 @@ export async function getCommandOutputSnapshot(
...
+ *   changed), the mirror can *append* VT that corresponds only to the new rows.
+ *
+ * - If the cursor moves or data is written above the previously mirrored end (for example,
+ *   when the command rewrites lines, uses carriage returns, or modifies earlier rows),
+ *   `_lowestDirtyCursorY` will be before `_lastUpToDateCursorY`. In that case the mirror
+ *   cannot safely append and instead falls back to taking a fresh VT snapshot of the
+ *   entire command range and *rewrites* the detached terminal content.
  */
-export class DetachedTerminalCommandMirror extends DetachedTerminalMirror implements IDetachedTerminalCommandMirror {
+export class DetachedTerminalCommandMirror extends Disposable implements IDetachedTerminalCommandMirror {
```
Tyriar, line 123: We want this to also allow input into the mirror and get forward to the XtermTerminal. The user should be able to press ctrl+c there or type in awaited input, etc.

Tyriar, line 123: Can we expand this by default?

src/vs/workbench/contrib/terminal/test/browser/chatTerminalCommandMirror.test.ts
```diff
@@ -0,0 +1,743 @@
...
+					{} as IContextKeyService
+				));
+
+				const writes: string[] = [];
+				const detached = createDetachedTerminal(writes);
+
+				(mirror as unknown as Record<string, unknown>)['_getOrCreateTerminal'] = async () => {
+					(mirror as unknown as Record<string, unknown>)['_detachedTerminal'] = detached;
+					return detached;
+				};
```
Tyriar: 😱

src/vs/workbench/contrib/terminal/test/browser/chatTerminalCommandMirror.test.ts
```diff
@@ -0,0 +1,743 @@
...
+			const result = await getCommandOutputSnapshot(xterm, command);
+			strictEqual(result, undefined);
+		});
+
+		test('uses executed marker when available', async () => {
+			const xtermImpl = new TestXtermTerminal();
+			xtermImpl.nextResult = 'output';
+			const xterm = xtermImpl as unknown as XtermTerminal;
+
+			const executedMarker = new TestMarker(1, 10) as unknown as IXtermMarker;
```
Tyriar: Too much as unknown in the tests, it's pretty much always a hack doing this that we should avoid

src/vs/workbench/contrib/terminal/test/browser/chatTerminalCommandMirror.test.ts
```diff
@@ -0,0 +1,743 @@
...
+	suite('getCommandOutputSnapshot', () => {
+		test('returns undefined when end marker is missing', async () => {
+			const xterm = new TestXtermTerminal() as unknown as XtermTerminal;
+			const command = createCommand(undefined, undefined);
+
+			const result = await getCommandOutputSnapshot(xterm, command);
+			strictEqual(result, undefined);
+		});
+
+		test('uses executed marker when available', async () => {
```
Tyriar: The tests could do with some work, they're very verbose. I was also expecting some really simple ones that basically just create an XtermTerminal, a mirror write to the first one and make sure their contents are equal (setting the command as the entire buffer or whatever is needed for that). Can cover a bunch of edge cases in those tests (single char, multi-line, wrapped line, full scrollback/startMarker=-1, etc.)

Tyriar (changes requested):

src/vs/workbench/contrib/terminal/browser/chatTerminalCommandMirror.ts
```diff
@@ -132,61 +115,367 @@ export async function getCommandOutputSnapshot(
...
+
+	private _lastVT = '';
+	private _lineCount = 0;
+	private _lastUpToDateCursorY: number | undefined;
+	private _lowestDirtyCursorY: number | undefined;
+	private _flushPromise: Promise<void> | undefined;
+	private _dirtyScheduled = false;
+	private _isStreaming = false;
+	private _sourceRaw: RawXtermTerminal | undefined;
+	private _isDisposed = false;
```
Tyriar: `this._store.isDisposed`?

Tyriar (dismissed):

src/vs/workbench/contrib/terminal/browser/chatTerminalCommandMirror.ts
meganrogge: Yes

https://github.com/user-attachments/assets/c9018ebf-7152-4e25-bd00-58fecef4e679

meganrogge: Done

https://github.com/user-attachments/assets/e3b9bc73-4173-4491-8fe9-5a26b39fdb69

amunger (dismissed):

Tyriar (approved):

PR #286006, opened 2026-01-05
using relative values instead of absolute values for the font size and the line height

### LINKED ISSUE #285891: Use relative numbers instead of absolute ones in themes
As related to variable line heights

aiday-mar:
fixes https://github.com/microsoft/vscode/issues/285891


src/vs/workbench/services/textMate/browser/textMateTokenizationFeatureImpl.ts
```diff
@@ -348,7 +348,8 @@ export class TextMateTokenizationFeature extends Disposable implements ITextMate
 		this._grammarFactory?.setTheme(this._currentTheme, this._currentTokenColorMap);
 		const colorMap = toColorMap(this._currentTokenColorMap);
 		const colorCssRules = generateTokensCSSForColorMap(colorMap);
-		const fontCssRules = generateTokensCSSForFontMap(this._currentTokenFontMap);
+		const fontSize = this._configurationService.getValue<number>('editor.fontSize') ?? 0;
+		const fontCssRules = generateTokensCSSForFontMap(this._currentTokenFontMap, fontSize);
```
alexdima: 🤔 I think the CSS would get stale if someone changes the editor.fontSize, and only a theme change or a window reload would end up fixing it. Would it be possible to use CSS variables to express the relative font size, maybe each editor in the viewLines would define its base font size and the CSS could use calc or something else to create relative font sizes?

src/vs/editor/common/model/tokens/tokenizationFontDecorationsProvider.ts
```diff
@@ -135,8 +137,9 @@ export class TokenizationFontDecorationProvider extends Disposable implements De
 			const annotationEndPosition = this.textModel.getPositionAt(annotation.range.endExclusive);
 			const range = Range.fromPositions(annotationStartPosition, annotationEndPosition);
 			const anno = annotation.annotation;
-			const className = classNameForFontTokenDecorations(anno.fontToken.fontFamily ?? '', anno.fontToken.fontSize ?? '');
-			const affectsFont = !!(anno.fontToken.fontFamily || anno.fontToken.fontSize);
+			const defaultFontSize = this.configurationService.getValue<number>('editor.fontSize') ?? 0;
+			const className = classNameForFontTokenDecorations(anno.fontToken.fontFamily ?? '', anno.fontToken.fontSizeMultiplier ? anno.fontToken.fontSizeMultiplier * defaultFontSize : 0);
+			const affectsFont = !!(anno.fontToken.fontFamily || anno.fontToken.fontSizeMultiplier);
 			const id = anno.decorationId;
```
alexdima: Same here, could we avoid using editor.fontSize and thus the configurationService?

src/vs/editor/common/model/tokens/tokenizationFontDecorationsProvider.ts
```diff
@@ -25,8 +26,8 @@ export class TokenizationFontDecorationProvider extends Disposable implements De
 
 	private static DECORATION_COUNT = 0;
 
-	private readonly _onDidChangeLineHeight = new Emitter<Set<LineHeightChangingDecoration>>();
-	public readonly onDidChangeLineHeight = this._onDidChangeLineHeight.event;
+	private readonly _onDidChangeLineHeightMultiplier = new Emitter<Set<LineHeightMultiplierChangingDecoration>>();
+	public readonly onDidChangeLineHeightMultiplier = this._onDidChangeLineHeightMultiplier.event;
```
alexdima: The rename of this event to contain "Multiplier" seems a bit overkill, making things more complicated without too much value IMHO. Keeping the event name `onDidChangeLineHeight` and `LineHeightChangingDecoration` is IMHO good enough, the members can reflect that this is a multiplier.

src/vs/editor/common/textModelEvents.ts
```diff
@@ -518,21 +548,42 @@ export class ModelLineHeightChangedEvent {
...
-				}
+		return affectsRangeOrPosition(rangeOrPosition, this.changes);
+	}
+}
+
+/**
+ * An event describing a change of a line height multiplier.
+ * @internal
+ */
+export class ModelLineHeightMultiplierChangedEvent {
```
alexdima: Why do we now have two distinct events, ModelLineHeightChangedEvent and ModelLineHeightMultiplierChangedEvent ? Why not a single one? Are there still folks adding absolute values somewhere?

src/vs/editor/common/model/decorationProvider.ts
```diff
@@ -26,6 +26,20 @@ export interface DecorationProvider {
 
 }
 
+export class LineHeightMultiplierChangingDecoration {
```
alexdima: Same here, all line height changing decorations should now use relative values

src/vs/editor/common/model.ts
```diff
@@ -1310,6 +1310,14 @@ export interface ITextModel {
...
 	 */
 	readonly onDidChangeLineHeight: Event<ModelLineHeightChangedEvent>;
+	/**
+	 * An event emitted when line height multipliers from decorations changes.
+	 * This event is emitted only when adding, removing or changing a decoration
+	 * and not when doing edits in the model (i.e. when decoration ranges change)
+	 * @internal
+	 * @event
+	 */
+	readonly onDidChangeLineHeightMultiplier: Event<ModelLineHeightMultiplierChangedEvent>;
```
alexdima: Same here, merge into `onDidChangeLineHeight`, the text model would only support the multiplier version.

alexdima (commented):
I've left comments, mostly OK, except I would try to make it that everyone who adds line height decorations can only do relative line heights, no more absolutes to keep the code sane.

Also, when generating the CSS and the class name inside the text model it looks like the font size is being read, it would be better to avoid it and use the relative multiplier in the class name to remove that dependency on the config service (and avoid the bug of not reacting to the value changing)

alexdima (approved):
Nice!

I just found a problem and pushed a fix where configuring:
```
        "textMateRules": [{
            "scope": "entity.name.type.class.ts",
            "settings": {
                "foreground": "#FF0000",
                "fontStyle": "bold",
                "lineHeight": 1.5,
                "fontSize": 1.5
            }
        }]
    },
```
would generate a class name `.font-decoration-default-1.5` which wouldn't work due to the `.` in the font size.

bpasero (approved):

PR #286042, opened 2026-01-06
Fix Inline CSP style error

nigeldias27:
While enforcing nonce for CSP style directive in our web application, which uses Monaco Editor, multiple violations were found, originating from two areas:
1. While creating style tags at https://github.com/microsoft/vscode/blob/bf56edffb59c43fa0de636c3aa1d548770b168b8/src/vs/base/browser/domStylesheets.ts#L51 since it is missing the nonce attribute
2. Due to inline style usages such as https://github.com/microsoft/vscode/blob/bf56edffb59c43fa0de636c3aa1d548770b168b8/src/vs/editor/browser/view/viewOverlays.ts#L177

Hence, the proposed changes include:
1. Adding the nonce attribute to the generated style tag.
2. Converting inline styles to style tags, if nonce exists, in the `createHTML` method which is called in `_finishRenderingNewLines` and _finishRenderingInvalidLines before setting innerHTML

### DIFF
--- a/src/vs/editor/browser/view/viewLayer.ts
+++ b/src/vs/editor/browser/view/viewLayer.ts
@@ -384,24 +384,25 @@ class ViewLayerRenderer<T extends IVisibleLine> {
 
 	private static _ttPolicy = createTrustedTypesPolicy('editorViewLayer', {
 		createHTML: (value: string) => {
-			if (mainWindow?.cspNonce) {
-				const stylePattern: RegExp = new RegExp('style\\s*=\\s*([\"\'])(.*?)\\1', 'g');
+			if (mainWindow.cspNonce) {
+				const stylePattern = new RegExp('style\\s*=\\s*([\"\'])(.*?)\\1', 'g');
 				const styleMatches = Array.from(value.matchAll(stylePattern)).reverse();
-				const classNames = Array.from({ length: styleMatches.length }, () => 'csp-' + Math.random().toString(36).substring(7));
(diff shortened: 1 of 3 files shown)


Moharam10:
thanks guys for fixing this! We are waiting for this PR to be merged :D

src/vs/editor/browser/view/viewLayer.ts
```diff
@@ -381,7 +382,29 @@ interface IRendererContext<T extends IVisibleLine> {
 
 class ViewLayerRenderer<T extends IVisibleLine> {
 
-	private static _ttPolicy = createTrustedTypesPolicy('editorViewLayer', { createHTML: value => value });
+	private static _ttPolicy = createTrustedTypesPolicy('editorViewLayer', {
+		createHTML: (value: string) => {
```
hediet, line 386: This method looks very dangerous:
1. it could have performance impacts when cspNonce is set
2. analyzing html with regular expressions will have correctness problems

hediet (changes requested):
Thanks for the PR! Unfortunately I'm not sure if this problem can be properly fixed in `createHTML`.

src/vs/editor/browser/view/viewLayer.ts
nigeldias27: Thanks for reviewing this PR @hediet !  An alternative would be to update each inline style usage with a style tag. I was thinking that this would be a smaller, easier change. However, do you think that the alternative would be acceptable?

PR #286181, opened 2026-01-06
Disable tooltips for Quick Input to prevent obscuring dropdown result (fix #285927)

### LINKED ISSUE #285927: Fix hover label on search box
I will describe a common annoyance, and leave it to you to find a way to improve usability:

1. I use a window size that is not very wide, such as a non-maximized window. This is relevant because the UI elements on top are tight together, and the search box is not very wide.
2.  I click on the search box on the top (similar to pressing Ctrl-P, but I click it with the mouse).
3. I move my hand from the mouse to the keyboard, and the mouse pointer stays where I clicked.
4. I use the nice find-as-you-type search box to get to my file name quickly. The file I want seems to be the first result, I want to make sure it is and press ENTER to get to it and start editing...
5. BUT I get an annoying hover label on top of the file name. It says `Search files by name (append : to go to line or @ to go to symbol)`

I basically want that label out of my way. In general, it's not a good idea to cover a dropdown that you've just filled with new data, that you expect the user's eyes to be reading...

This happens to me VERY frequently and it is very disruptive of flow. My brain just sits there trying to decode a partial filename covered with a strange thing. And it makes the second line look the first result! So the file that I don't want just sits there while I wonder "why isn't it finding the obvious result?".

I've looked online, I know I can change the hover delay or remove hovers completely. But that means losing functionality, and I won't even know what I am missing. A better fix is surely possible from the fine UI designers that bring us the excellent VSCode. Thanks!

MohamedEmirHajji:
Disable tooltips for Quick Input to prevent obscuring dropdown result (fix #285927)

### DIFF
--- a/src/vs/base/browser/ui/inputbox/inputBox.ts
+++ b/src/vs/base/browser/ui/inputbox/inputBox.ts
@@ -40,6 +40,7 @@ export interface IInputOptions {
 	readonly actionViewItemProvider?: IActionViewItemProvider;
 	readonly inputBoxStyles: IInputBoxStyles;
 	readonly history?: IHistory<string>;
+	readonly hideHoverOnKeyDown?: boolean;
 }
 
 export interface IInputBoxStyles {
@@ -200,6 +201,12 @@ export class InputBox extends Widget {
 		this.oninput(this.input, () => this.onValueChange());
 		this.onblur(this.input, () => this.onBlur());
 		this.onfocus(this.input, () => this.onFocus());
+
+		if (this.options.hideHoverOnKeyDown) {
+			this._register(dom.addDisposableListener(this.input, dom.EventType.KEY_DOWN, () => {
+				getBaseLayerHoverDelegate().hideHover();
+			}));
+		}
 
 		this._register(this.ignoreGesture(this.input));
 
--- a/src/vs/base/browser/ui/inputbox/inputBox.ts
+++ b/src/vs/base/browser/ui/inputbox/inputBox.ts
@@ -202,12 +202,6 @@ export class InputBox extends Widget {
 		this.onblur(this.input, () => this.onBlur());
 		this.onfocus(this.input, () => this.onFocus());
 
-		if (this.options.hideHoverOnKeyDown) {
-			this._register(dom.addDisposableListener(this.input, dom.EventType.KEY_DOWN, () => {
-				getBaseLayerHoverDelegate().hideHover();
-			}));
-		}
-
 		this._register(this.ignoreGesture(this.input));
 
 		setTimeout(() => this.updateMirror(), 0);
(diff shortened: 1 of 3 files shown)


src/vs/platform/quickinput/browser/quickInputBox.ts
```diff
@@ -40,6 +41,11 @@ export class QuickInputBox extends Disposable {
 		input.role = 'textbox';
 		input.ariaHasPopup = 'menu';
 		input.ariaAutoComplete = 'list';
+
+		// Hide tooltip when user starts typing to prevent obscuring dropdown results
+		this._register(dom.addDisposableListener(input, dom.EventType.KEY_DOWN, () => {
```
TylerLeonhardt: For clean code, this should be handled by the input that Quick Pick takes advantage of and be a configuration option on that object upon creation.

MohamedEmirHajji: Thanks for your review. I moved the logic to InputBox and added a boolean to the configuration called hideHoverOnKeyDown

src/vs/base/browser/ui/inputbox/inputBox.ts
```diff
@@ -201,6 +202,12 @@ export class InputBox extends Widget {
 		this.onblur(this.input, () => this.onBlur());
 		this.onfocus(this.input, () => this.onFocus());
 
+		if (this.options.hideHoverOnKeyDown) {
+			this._register(dom.addDisposableListener(this.input, dom.EventType.KEY_DOWN, () => {
```
TylerLeonhardt: We already have an `onValueChange` function that runs when what is in the input changes. This feels like a good approximation without needing to add another event.

Additionally, rather than doing this, `getBaseLayerHoverDelegate().hideHover();` we already have a `this.hover.value` which we could probably just dispose of instead.

MohamedEmirHajji: Yep you are totally right about using `onValueChange`. About `getBaseLayerHoverDelegate().hideHover();`, while testing, I tried setting this.hover = undefined (using this.hover.clear()) but it didn't really made the hover disappear from the UI

src/vs/base/browser/ui/inputbox/inputBox.ts
```diff
@@ -40,6 +40,7 @@ export interface IInputOptions {
 	readonly actionViewItemProvider?: IActionViewItemProvider;
 	readonly inputBoxStyles: IInputBoxStyles;
 	readonly history?: IHistory<string>;
+	readonly hideHoverOnKeyDown?: boolean;
```
TylerLeonhardt: last bit of feedback (I think). Let's change this to `hideHoverOnValueChange` since that is the actual event it is reacting to.

TylerLeonhardt (approved):
Thanks! I think this is a good compromise

joshspicer (approved):

PR #286198, opened 2026-01-06
fix accessiblity signal bug

### LINKED ISSUE #285702: "Terminal is awaiting input" always triggers audio cue
Every time this dialog is rendered, it triggers the audio cue. I don't have screenreader mode or any audio cues enabled.

Image

meganrogge:
fixes #285702


src/vs/platform/accessibilitySignal/browser/accessibilitySignalService.ts
```diff
@@ -695,7 +695,6 @@ export class AccessibilitySignal {
 		name: localize('accessibilitySignals.chatUserActionRequired', 'Chat User Action Required'),
 		sound: Sound.chatUserActionRequired,
 		announcementMessage: localize('accessibility.signals.chatUserActionRequired', 'Chat User Action Required'),
-		settingsKey: 'accessibility.signals.chatUserActionRequired',
-		managesOwnEnablement: true
```
roblourens, line 699: I'm curious why this exists, it means I can't disable an audio cue?

roblourens (approved):
I'm very excited for this to be fixed, no more dinging 😅 thanks!

PR #286224, opened 2026-01-06
fix: memory leak editor group view

SimonSiefke:
Fixes a promise memory leak in editor group view.

### Before 

When opening two diff editors side by side and then closing all editors 37 times, the number of promises in editorGroup seems to grow by 1 each time:

diff-editor

### After

No more promise leak is detected.

### DIFF
--- a/src/vs/workbench/browser/parts/editor/editorGroupView.ts
+++ b/src/vs/workbench/browser/parts/editor/editorGroupView.ts
@@ -2209,9 +2209,6 @@ export class EditorGroupView extends Themable implements IEditorGroupView {
 		this._onWillDispose.fire();
 
 		this.whenRestored = undefined!;
-		this.whenRestoredPromise.cancel().catch(() => {
-			// ignore
-		});
 		this.whenRestoredPromise = undefined!;
 
 		super.dispose();
--- a/src/vs/workbench/browser/parts/editor/editorGroupView.ts
+++ b/src/vs/workbench/browser/parts/editor/editorGroupView.ts
@@ -136,8 +136,8 @@ export class EditorGroupView extends Themable implements IEditorGroupView {
 
 	private readonly containerToolBarMenuDisposable = this._register(new MutableDisposable());
 
-	private readonly whenRestoredPromise = new DeferredPromise<void>();
-	readonly whenRestored = this.whenRestoredPromise.p;
+	private whenRestoredPromise = new DeferredPromise<void>();
+	whenRestored = this.whenRestoredPromise.p;
 
 	constructor(
 		from: IEditorGroupView | ISerializedEditorGroupModel | null,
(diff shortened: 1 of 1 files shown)


SimonSiefke:
Hi @bpasero , 

Thanks for the feedback!



This is the full test result data of the diff-editor-open-side-by-side test, showing which functions grow between the start and end of the test:
diff-editor-side-by-side


The full test result data as json:
[folder.zip](https://github.com/user-attachments/files/24467871/folder.zip)



And the video of the test case:

[video.webm](https://github.com/user-attachments/assets/03ada0a4-f5b1-4588-b13c-d2f5aea2f2c9)

And the script to run it (ensure using node 24 or newer)

Measuring named function count:

```sh
git clone git@github.com:SimonSiefke/vscode-memory-leak-finder.git &&
cd vscode-memory-leak-finder &&
npm ci &&
node packages/cli/bin/test.js  --run-skipped-tests-anyway --check-leaks --measure named-function-count3 --measure-after --runs 37   --only diff-editor-side-by-side 
```


Measuring promises with stacktrace

```sh
node packages/cli/bin/test.js  --run-skipped-tests-anyway --check-leaks --measure promises-with-stack-trace --measure-after --runs 37   --only diff-editor-side-by-side 
```



It seems a lot of functions are leaking when opening multiple diff editors and then closing them, one of which is `EditorGroupView` and `EditorGroupModel`:

```json
{
  "count": 41,
  "delta": 37,
  "name": "EditorGroupView",
  "originalLocation": "vscode/src/vs/workbench/browser/parts/editor/editorGroupView.ts:143:2",
  "originalName": "",
  "sourceLocation": "vscode/out/vs/workbench/browser/parts/editor/editorGroupView.js:64:13"
},
{
  "count": 41,
  "delta": 37,
  "name": "EditorGroupModel",
  "originalLocation": "vscode/src/vs/workbench/common/editor/editorGroupModel.ts:220:2",
  "originalName": "",
  "sourceLocation": "vscode/out/vs/workbench/common/editor/editorGroupModel.js:47:13"
},
```


It seems that EditorGroupView isn't leaking itself, but rather something might be holding onto EditorGroupView instances. Which means also holding on to the promise that EditorGroupView creates. 


And setting the property to undefined in EditorGroupView is indeed a workaround to ensure the promise gets garbage collected, even when the EditorGroupView might be leaked.

What it doesn't do unfortunately is address the root cause (or causes).

SimonSiefke:
Thanks! Moving into https://github.com/microsoft/vscode/issues/286336


> At least from my simple test now I cannot see the leak in editor group view: I am opening an untitled editor, split it and close the split. I do this N times and then compare the heap snapshot.

I'm seeing that as well. The possible leaks do not occur when splitting and viewing two normal editors side by side. But when viewing two diff editors side by side like in the video.

bpasero:
image

So there is always 1 group and model alive because even when all editors are closed, there will be 1 active group.

PR #286236, opened 2026-01-06
fix: memory leak in main thread terminal service

SimonSiefke:
Fixes a memory leak in `mainThreadTerminalService`. 

### Details

Registering a terminal completion provider happens through this code:

```ts
private readonly _completionProviders = new Map<string, IDisposable>();

public $registerCompletionProvider(id: string, extensionIdentifier: string, ...triggerCharacters: string[]): void {
	this._completionProviders.set(id, this._terminalCompletionService.registerTerminalCompletionProvider(/*...*/)
```

When a completion provider with the same id gets registered again, instead of replacing the completion provider, another one gets created. And the previous `completionProviders[id]` `Disposable` kind of gets unreferenced.

### Fix

The fix uses a `DisposableMap` so that at most one completion provider gets registered with the same id. When another completion provider with the same id gets registered, the previous registration will be disposed.

### Before
When debugging a python file 37 times, the number of `TerminalCompletionService.registerTerminalCompletionProvider` functions seems to grow by one each time:

Untitled

### After

When debugging a python file 37 times, while the number of some functions still seems to grow, the number of `TerminalCompletionService.registerTerminalCompletionProvider` functions now stays constant:

debug-python


dmitrivMS:
@SimonSiefke Thank you!

dmitrivMS (dismissed):

dmitrivMS (approved):

alexr00 (approved):

PR #286246, opened 2026-01-06
fix: memory leak call stack widget

SimonSiefke:
Fixes a memory leak in call stack widget. 

### Details

In setFrames, a new disposable seems to be registered each time using `this._register`:

```ts
/** Replaces the call frames display in the view. */
public setFrames(frames: AnyStackFrame[]): void {
	// cancel any existing load
	this.currentFramesDs.clear();
	this.cts = new CancellationTokenSource();
	this._register(toDisposable(() => this.cts!.dispose(true)));

	this.list.splice(0, this.list.length, this.mapFrames(frames));
}
```

E.g. when setFrames is called 10 times, 10 disposables would be registered to the class. 

### Fix

By registering the disposable to `currentFramesDs` instead of using `this._register`, there will be at most one of this disposable registered. Similar to how mapFrames also uses `currentFramesDs` instead of `this._register`.

```ts
/** Replaces the call frames display in the view. */
public setFrames(frames: AnyStackFrame[]): void {
	// cancel any existing load
	this.currentFramesDs.clear();
	const cts = new CancellationTokenSource();
	this.currentFramesDs.add(toDisposable(() => cts.dispose(true)));
	this.cts = cts;

	this.list.splice(0, this.list.length, this.mapFrames(frames));


}
```

### Before

When opening the test view using the test-provider extension sample and running tests 97 times, the number of functions in `CallStackWidget.setFrames`  seems to grow by 1 each time:

Untitled (2)

### After

When opening the test view using the test-provider extension sample and running tests 97 times, while the number of some functions still grows, the number of functions in `CallStackWidget.setFrames`  stays constant:

test-view


connor4312 (approved):
thanks!

dmitrivMS (approved):

PR #286251, opened 2026-01-06
fix: memory leak terminal process manager

SimonSiefke:
Fixes a promise memory leak in terminal process manager. 

### Details

Sets the promise to `undefined` on dispose to ensure the promise is garbage collected.

### Before

When creating and removing a terminal 37 times, the number of promises seems to grow by 1 each time:

process-manager

### After

Running the same test from above also 37 times, no more promise memory in `terminalProcessManager` is detected.


meganrogge (dismissed):
Thank you!

Tyriar (approved):

alexr00 (approved):

PR #286297, opened 2026-01-07
vscode mcp: support custom workspace path; enable restart tool

rebornix:
- optional workspace paths through MCP setup/start so the server can open a specific folder or workspace when requested
- enable reload tool to support window reload

### DIFF
--- a/test/automation/src/application.ts
+++ b/test/automation/src/application.ts
@@ -54,17 +54,6 @@ export class Application {
 		return this._workspacePathOrFolder;
 	}
 
-	/**
-	 * Returns the workspace path, throwing if not set.
-	 * Use this in tests that require a workspace to be open.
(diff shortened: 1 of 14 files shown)


TylerLeonhardt (approved):
I'm fine with this from the MCP side

rebornix:
Merging it to allow us to continue experimenting with it. if there are feedbacks I can address in followup PRs.

bpasero:
@rebornix why do we have to repeat this code everywhwere?

```ts
const workspacePathOrFolder = app.workspacePathOrFolder;
if (!workspacePathOrFolder) {
	throw new Error('This test requires a workspace to be open');
}
```

Wouldn't it be better to just throw at the getter of `app.workspacePathOrFolder`

PR #286325, opened 2026-01-07
Fix a set of issues around transfer and save of custom editor documents

### LINKED ISSUE #125293: Custom editor reopened from untitled does not properly wire in dirty state?
Steps to Reproduce:

1. open an untitled text editor and type some content
2. reopen with hex editor

=> the editor indicates dirty, but e.g. the explorer does not

From my testing it seems like in this case the working copy that is created does not emit a `onDidChangeDirty` event?

recording (2)

dmitrivMS:
Fixes #125293

In addition to original scenario in the bug (error/duplicate opened documents) there were more issues down the line:
- Create untitled doc, enter some text
- Reopen in Hex Editor -> original bug
- Save from Hex Editor -> the untitled document was left around and a "save unsaved changes" dialog was popping up -> fixed
- Now create a new untitled document, enter some text
- Reopen the new doc in Hex Editor -> we were crashing

With these changes all the above issues are fixed and work as expected.

Moving to Jan milestone since it's a larger change.


src/vs/workbench/services/untitled/common/untitledTextEditorModel.ts
```diff
@@ -61,6 +61,12 @@ export interface IUntitledTextEditorModel extends ITextEditorModel, ILanguageSup
 	 */
 	setEncoding(encoding: string): Promise<void>;
```
bpasero, line 63: Can we explore a different route? Conceptually it seems wrong to me that the untitled model would have to become aware of custom editors or being transferred. Could we revert the model simply, which in turn would dispose it, considering that the data has been transfered into the custom editor?

bpasero (dismissed):

src/vs/workbench/services/untitled/common/untitledTextEditorModel.ts
dmitrivMS: Let me try

dmitrivMS: Done

lramos15 (approved):

PR #286341, opened 2026-01-07
Fix indentation rules for VB

### LINKED ISSUE #118932: Add language configuration file for visual basic
- VS Code Version: 1.51.1
- OS Version: Windows_NT x64 10.0.17763

Steps to Reproduce:

1. Open a Visual Basic file (.vb)
2. Type:
```
if True then
    ' Some code  
end if
```

**Expected:** When you type "End If", the indentation will go back to the same level as the opening block ("if"), as can be seen above.

**Actual:** The "end if" stays in the wrong indentation level:
```
if True then
    ' Some code  
    end if
```

**Possible cause and solution**:
I found out that this can happen in all programming languages where a closing tag includes the opening tag as a separate word (For example: _If ... End If_, or _While ... End While_).

**Solution:** This is related to the order of the pieces as defined in [these lines of code in the richEditBrackets.ts file](https://github.com/microsoft/vscode/blob/a699ffaee62010c4634d301da2bbdb7646b8d1da/src/vs/editor/common/modes/supports/richEditBrackets.ts#L228)

- If you change the order so that first the **closing** brackets would be pushed and only then the **opening** brackets, then it will work.
- Alternatively, you can modify the _createBracketOrRegExp_ function and just add a reverse() function there.

I'm not sure if it will mess things up or not (though my initial tests look promising) and I'm not too familiar with the code to change it myself and issue a PR, so please take a look and let me know what you think.

Thank you!

dmitrivMS:
Fixes #118932

Added more tests as well to cover onEnter scenarios.


extensions/vb/language-configuration.json
```diff
@@ -32,30 +32,22 @@
 			"flags": "i"
 		},
 		"increaseIndentPattern": {
-			"pattern": "^\\s*((If|ElseIf).*Then(?!\\s+(End\\s+If))\\s*(('|REM).*)?$)|\\b(Else|While|For|Do|Select\\s+Case|Case|Sub|Function|Class|Module|Enum|Structure|Interface|Namespace|With|Try|Catch|Finally|SyncLock|Using|Property|Get|Set|AddHandler|RaiseEvent|RemoveHandler|Event|Operator)\\b(?!.*\\bEnd\\s+(If|Sub|Function|Class|Module|Enum|Structure|Interface|Namespace|With|Select|Try|While|For|Property|Get|Set|SyncLock|Using|AddHandler|RaiseEvent|RemoveHandler|Event|Operator)\\b).*(('|REM).*)?$",
+			"pattern": "^\\s*((If|ElseIf).*Then(?!.*End\\s+If)\\s*(('|REM).*)?|(Else|While|For|Do|Select\\s+Case|Case|Sub|Function|Class|Module|Enum|Structure|Interface|Namespace|With|Try|Catch|Finally|SyncLock|Using|Property|Get|Set|AddHandler|RaiseEvent|RemoveHandler|Event|Operator)\\b(?!.*\\bEnd\\s+(If|Sub|Function|Class|Module|Enum|Structure|Interface|Namespace|With|Select|Try|While|For|Property|Get|Set|SyncLock|Using|AddHandler|RaiseEvent|RemoveHandler|Event|Operator)\\b).*(('|REM).*)?)$",
```
RedCMD: `(?!.*End\\s+If)` seems useless
as you're already enforcing only whitespace and comments till eol
and creates false negatives if `End If` appears inside the comment

should there be checks for word boundary between `If` and `Then`?

Image

aiday-mar (approved):

extensions/vb/language-configuration.json
RedCMD:
```suggestion
			"pattern": "^\\s*((If|ElseIf)\\b.*\\bThen\\s*(('|REM).*)?|(Else|While|For|Do|Select\\s+Case|Case|Sub|Function|Class|Module|Enum|Structure|Interface|Namespace|With|Try|Catch|Finally|SyncLock|Using|Property|Get|Set|AddHandler|RaiseEvent|RemoveHandler|Event|Operator)\\b(?!.*\\bEnd\\s+(If|Sub|Function|Class|Module|Enum|Structure|Interface|Namespace|With|Select|Try|While|For|Property|Get|Set|SyncLock|Using|AddHandler|RaiseEvent|RemoveHandler|Event|Operator)\\b).*(('|REM).*)?)$",
```

PR #286426, opened 2026-01-07
don't render empty leading space for terminal commands

### LINKED ISSUE #286011: Bad wrapping in terminal command
Image

meganrogge:
fixes #286011

### DIFF
--- a/src/vs/workbench/contrib/chat/common/chatService/chatService.ts
+++ b/src/vs/workbench/contrib/chat/common/chatService/chatService.ts
@@ -368,7 +368,10 @@ export interface IChatTerminalToolInvocationData {
 	commandLine: {
 		original: string;
 		userEdited?: string;
+		/** The tool-edited command for execution (may include leading whitespace for history prevention) */
 		toolEdited?: string;
+		/** The tool-edited command for display (trimmed). Falls back to toolEdited if not set. */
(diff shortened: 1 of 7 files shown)


src/vs/workbench/contrib/chat/common/chatService/chatService.ts
```diff
@@ -368,7 +368,10 @@ export interface IChatTerminalToolInvocationData {
 	commandLine: {
 		original: string;
 		userEdited?: string;
+		/** The tool-edited command for execution (may include leading whitespace for history prevention) */
 		toolEdited?: string;
+		/** The tool-edited command for display (trimmed). Falls back to toolEdited if not set. */
+		toolEditedDisplay?: string;
```
Tyriar, line 374: Instead of passing around this new property, how about we just special case the confirmation/progress part and `.trimStart()` there? It's already a little cumbersome with these props so I don't want to add to that when this is just trimming and not transforming it in a significant way.

Tyriar (changes requested):

src/vs/workbench/contrib/chat/common/chatService/chatService.ts
meganrogge: Ok - wasn't sure if that would mess with some command lines

Issue #285251, opened 2025-12-28
Persisting long chat sessions is slow and causes renderer freezes


garretwilson:
As I type a prompt in GitHub Copilot chat in VS Code in the side panel, every few seconds everything will freeze for several seconds. If I type, the information I typed will not appear until after the freezing stops. Sometimes the things I type during the freeze seem to appear in a different order after the freeze stops.

```
Version: 1.107.1 (user setup)
Commit: 994fd12f8d3a5aa16f17d42c041e5809167e845a
Date: 2025-12-17T14:15:14.850Z
Electron: 39.2.3
ElectronBuildId: 12895514
Chromium: 142.0.7444.175
Node.js: 22.21.1
V8: 14.2.231.21-electron.0
OS: Windows_NT x64 10.0.26200
```

This has happened for a while, but it seems to happen more often and the freezing lasts longer each time it happens. I have to just stop and wait for it to "come back to life". These "hiccups" are starting to get in the way of work, as I can't even get a full paragraph typed without needing to stop and wait for the world to start again. It's literally as if everything freezes for several seconds.

garretwilson:
The cursor literally stops flashing in the chat side panel during this freezing. If the cursor is flashing off at the time, it stays disappeared. If the cursor is flashing on at the time, it stays stuck as a vertical line until the freezing stops.

garretwilson:
When I first started using GitHub Copilot, even in new chats I would notice that during typing every few seconds the typing would "hiccup". GitHub Copilot was proving useful, so I gritted my teeth and tried to ignore it.

I don't know if that's the same thing, but on the same note as [my other comment about long conversation histories](https://github.com/microsoft/vscode/issues/285951#issuecomment-3715166055), could it be that this freezing is the same issue, that it simply grows into longer freezes correlated to the length of chat conversation? I'm just trying to provide ideas for you to look into; hope it's useful.

CleverCoder:
> > Here's an example of the nasty pause: https://drive.google.com/file/d/1iN_WF5q6vbNfgDQNO5UXCKB4eomyr1yW/view?usp=drive_link _(EDIT: Give me a 2 minute series of steps to capture CPU/perf stats for VSCode and I'll do it. I don't have time to parse that long page that was linked -- sorry... time is $$$)_
> 
> Here's a copy of the paragraph that I linked to. It's easier than a screen recording.
> 
> * Execute "F1 > Toggle Developer Tools." and select the "Performance"-tab
[...]

PR #286436, opened 2026-01-07
chat: store binary attachments as base64

connor4312:
Refs https://github.com/microsoft/vscode/issues/285251#issuecomment-3717207362

### DIFF
--- a/src/vs/workbench/contrib/chat/common/attachments/chatVariableEntries.ts
+++ b/src/vs/workbench/contrib/chat/common/attachments/chatVariableEntries.ts
@@ -16,6 +16,8 @@ import { ISCMHistoryItem } from '../../../scm/common/history.js';
 import { IChatContentReference } from '../chatService/chatService.js';
 import { IChatRequestVariableValue } from './chatVariables.js';
 import { IToolData, ToolSet } from '../tools/languageModelToolsService.js';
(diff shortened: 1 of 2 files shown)


src/vs/workbench/contrib/chat/common/attachments/chatVariableEntries.ts
```diff
@@ -288,6 +290,43 @@ export namespace IChatRequestVariableEntry {
...
+			return dup;
+		}
+
+		return v;
+	}
+
+	export function fromExport(v: IChatRequestVariableEntry): IChatRequestVariableEntry {
+		// Old variables format
+		// eslint-disable-next-line local/code-no-in-operator
+		if (v && 'values' in v && Array.isArray(v.values)) {
```
roblourens, line 308: Let's not add new instances of these?

connor4312: This is an existing instance moved from the chatModel file, which I think was excluded.

Maybe we want to update the lint rule because the recommended `hasKey` helper does not work on `unknown` types, so we must use `in` here

roblourens (approved):

src/vs/workbench/contrib/chat/common/attachments/chatVariableEntries.ts
roblourens: Ah nvm

PR #286445, opened 2026-01-07
Add extension prompt file caching

pwang347:
## Overview
By default, we should cache the prompt files that are provided. This can be opt-out using the `isCacheable` flag.


pwang347:
> * first the cached entries of all extensions that are currently installed are read
> * then the activation of extension is awaited
> * then the cached results are used of a provider returns undefined or on cancellation
> * if a provider returns a result, that result is merged onto to the cached results. If a the result contains less resources because one was removed that resource still shows up, but only until the next call of `listFromProviders`. That's looks like a bug.
> 
> This is quite tricky to get right. I would assume that most providers can return providers quickly and don't need any caching. So I would suggest to implement the caching on the extension side when required.

I think caching on extension side makes sense conceptually, but from the experience it means that during init we start in an empty state and we'll see custom agents, instructions, prompt files gradually appear in the UI which feels a bit bad (things will jump around and be inconsistent). Even if an extension provider itself returns instantly, we still need to wait for extension host to activate it which may or may not be slow. Could wait for feedback on this one potentially though.

PR #286453, opened 2026-01-07
Add support for custom source grouping

### LINKED ISSUE #280798: Distinguish where the enterprise agent mode comes from
Testing #280465

It would be nice if the quick pick showed more than just "Extensions" and instead showed that it was contributed by the enterprise/organization.

Image

pwang347:
image

Fixes https://github.com/microsoft/vscode/issues/280798


aeschli (commented):
I wonder if this should be rather be done as a label on the provider.

PR #286457, opened 2026-01-08
Prompt files provider API V2

pwang347:
## Overview
- Update provider API to take in content directly as well
- Add command `vscode.extensionPromptFileProvider` to query latest extension resource state


src/vscode-dts/vscode.proposed.chatPromptFiles.d.ts
```diff
@@ -7,12 +7,93 @@
```
aeschli, line 7: Having the headers specified in the vscode API is problematic and undesirable.
- vscode API needs to be backward compatible, so we can't easily remove or rename properties or their values (e.g. tool names)
- every addition needs to be discussed in a API call

So what providers provide should just be a string, no structure, that then will be parsed on the core side

src/vs/workbench/contrib/chat/browser/promptSyntax/chatPromptContentProvider.ts
```diff
@@ -0,0 +1,45 @@
...
+	}
+
+	async provideTextContent(resource: URI): Promise<ITextModel | null> {
+		const existing = this.modelService.getModel(resource);
+		if (existing) {
+			return existing;
+		}
+
+		// Decode the content from the query string
+		const content = decodeURIComponent(resource.query);
```
aeschli: Putting resource content in the query is problematic and a misuse of a URI.
- the query is part of the resource identifier. `foo:///bar?hello` and `foo:///bar?world` should not represent two different resources, but the same one (identified e.g. by 'foo:///bar')
- there's a limit on how long URIs can be typically

Maybe the better model s to create TextDocuments on the extension and share contents like that.

eg3359821-sys (commented):
SHIP THER NET WEB POINTS

src/vscode-dts/vscode.proposed.chatPromptFiles.d.ts
pwang347: What do you think about providing something like this?

```typescript
interface IPromptFileContent {
  header: {
    // for supported properties, see <link> (documentation will have a table showing what versions various properties were added)
    [key: string]: unknown;
  },
  body: string;
}

// we can take in either a string as value, or something structured like the above
```

Happy to discuss more; I think my main problem with the current contract is that anyone who is interested in adding dynamic content needs to get into the business of managing files as well as frontmatter formatting (and potentially parsing), which feels a bit wrong IMO for something quick and dirty.

src/vs/workbench/contrib/chat/browser/promptSyntax/chatPromptContentProvider.ts
pwang347: I'll need to explore this a bit more, from playing around with it when I tried opening untitled documents they would sometimes appear in the windows tab unexpectedly for me.

pwang347: Basically, I see this entrypoint as our internal recommendation for how users should be implementing ephemeral resources since I think it is a valid use case (not everyone will want to care about managing prompt files, the end goal is to inject instructions, agents, etc.)

src/vscode-dts/vscode.proposed.chatPromptFiles.d.ts
pwang347: pushed changes above, also updated the URI handling so we can take a look tomorrow - thanks!

sandy081 (dismissed):

sandy081 (dismissed):

src/vscode-dts/vscode.proposed.chatPromptFiles.d.ts
```diff
@@ -22,100 +53,167 @@ declare module 'vscode' {
 		 * Indicates whether the custom agent is editable. Defaults to false.
 		 */
 		readonly isEditable?: boolean;
```
aeschli: When created with a virtual id, the resource can not be edited, correct? We would not be able to store it back

aeschli: Should we maybe remove the 'isEditable' all together?
If created from a URI: Let the file system define the readonlyness
If created from a id: Not readonly

src/vs/workbench/api/common/extHostTypes.ts
```diff
@@ -3896,3 +3898,119 @@ export class McpHttpServerDefinition implements vscode.McpHttpServerDefinition {
...
+
+/**
+ * Creates a virtual URI for a prompt file.
+ * The content is NOT embedded in the URI - it will be registered separately
+ * with the ChatPromptContentStore on the main thread.
+ */
+function createVirtualPromptUri(id: string, extension: string): URI {
+	return URI.from({
+		scheme: Schemas.vscodeChatPrompt,
+		path: `/${id}${extension}`
```
aeschli: The ${id}${extension} might not be unique.
Better: $extension/$id.

aeschli (changes requested):
I suggest we just pass a string and remove the yaml stringify.
There are multiple ways to format an array or an object. Also there's stuff we don't support, such as new lines in values.
Easier to just leave this to the API user.

src/vscode-dts/vscode.proposed.chatPromptFiles.d.ts
pwang347: Should an extension be able to mark certain files it provides as protected though without necessarily changing file perms? e.g., for the scenario of GitHub organizations, we are simply caching the content locally but may not want a user to manually modify cache. Not sure if we should update to use the other API in this case to force it to be readonly.

aeschli: Wouldn't you use the string API for your case? Also its currently a bit of a hack to surface the workspace storage URIs.

aeschli (approved):

PR #286470, opened 2026-01-08
Rename fix for disposed models

### LINKED ISSUE #283514: Allow to rename local chat sessions
pre v0.35.0 it was possible to rename chat session. Now with the new chat session view it becomes more interesting to jump between chat sessions but makes it harder to track them. I would love to be able to rename my chat sessions again. 
In addition, either adding a tag or linking a chat session to a branch would be great

osortega:
Fixes: https://github.com/microsoft/vscode/issues/283514


TylerLeonhardt (dismissed):

src/vs/workbench/contrib/chat/common/chatService/chatServiceImpl.ts
```diff
@@ -243,6 +243,16 @@ export class ChatService extends Disposable implements IChatService {
 		const model = this._sessionModels.get(sessionResource);
 		if (model) {
 			model.setCustomTitle(title);
+		} else {
+			let modelRef: IChatModelReference | undefined;
+			try {
+				modelRef = await this.getOrRestoreSession(sessionResource);
```
osortega: TBH doing this seems a bit silly to me, just restore the chat model to update the title and dispose it, futher down then updating the store.
@roblourens when reviving a chat model, do we take the title from the store? If so maybe we just need to fire `onDidChange` event without the need of creating the modelRef

roblourens: I think the title is already in the index to avoid having to do this. I think the problem is just that when loading a model, we should take the title from the index and let it overwrite the one in the stored model.

roblourens: And I believe the block below this (`// Update the title in the file storage`) can be in the `else` from `if (model)`

rebornix (dismissed):

rebornix (approved):

PR #286560, opened 2026-01-08
Update TextMate grammars from upstream sources

dmitrivMS:
Includes fix for `scripts/test.bat` to use `electron.ts` instead of `electron.js`.
Includes a fix for `update-grammars` on Windows.
Includes baseline update for one of the tests based on grammar changes.


alexr00 (commented):
This is actually a surprising number of grammar updates. Great to see so much activity in the upstream repos!

alexr00 (approved):

PR #286603, opened 2026-01-08
Add instructions to ensure that new ai features are gated by the right context key

roblourens:

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -137,3 +137,9 @@ function f(x: number, y: string): void { }
 - When adding file watching, prefer correlated file watchers (via fileService.createWatcher) to shared ones.
 - When adding tooltips to UI elements, prefer the use of IHoverService service.
 - Do not duplicate code. Always look for existing utility functions, helpers, or patterns in the codebase before implementing new functionality. Reuse and extend existing code whenever possible.
+
+
+## Adding chat/AI-related features
+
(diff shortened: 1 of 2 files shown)


.github/copilot-instructions.md
```diff
@@ -137,3 +137,9 @@ function f(x: number, y: string): void { }
 - When adding file watching, prefer correlated file watchers (via fileService.createWatcher) to shared ones.
 - When adding tooltips to UI elements, prefer the use of IHoverService service.
 - Do not duplicate code. Always look for existing utility functions, helpers, or patterns in the codebase before implementing new functionality. Reuse and extend existing code whenever possible.
+
+
+## Adding chat/AI-related features
+
+- When adding a new chat/AI feature like a new surface where chat or agents appear, a new AI command, etc, these features must not show up for users when they've disabled AI features. The best way to do this is to gate the feature on the context key `ChatContextKeys.enabled` via a when clause.
+- When doing a code review for code that adds an AI feature, please ensure that the feature is properly gated.
```
Tyriar: Can we filter this in a new instructions file to something like `**chat**.ts`? The more feature-specific instructions we add to the top level instructions file the more context is used and the less focused

Tyriar (dismissed):

.github/copilot-instructions.md
roblourens: Good idea

dmitrivMS (approved):

PR #286607, opened 2026-01-08
Add enforceModelTimeout setting

### LINKED ISSUE #286598: Timeout parameter on terminal tool
Opus has done this 1000 times in my current session. Should the terminal tool have a timeout parameter so it doesn't feel the need to use this command?

Image

Maybe there is already an issue for this, not a new idea I think

Tyriar:
Fixes #286598

image


meganrogge (approved):
I'm slightly worried about giving this control over to the model, but 🤷🏼‍♀️ we shall see!

PR #286624, opened 2026-01-08
mcp: fix start all servers command

### LINKED ISSUE #283959: vscode.lm.tools does not include tools from extensions/MCP.json
vscode.lm.tools does not include MCP- or extension-contributed tools unless they are activated and started.

Chat auto-starts MCPs, so vscode.lm.tools or another programmatic API should mirror the Chat behavior.

rwoll:
Resolves #283959.

Previous implementation looked strictly for `undefined`, but the IPC layer
converts `undefined` to `null`. This meant that calling the command via IPC
like:

```
await vscode.commands.executeCommand('workbench.mcp.startServer', undefined);
```

was not working (while `vscode.commands.executeCommand('workbench.mcp.startServer')` was).

Instead of handling `null` as a value as well, use a specicial value of `*`
to indicate all servers should be started so the intent is clearer:

```
BEFORE: await vscode.commands.executeCommand('workbench.mcp.startServer');
AFTER : await vscode.commands.executeCommand('workbench.mcp.startServer', '*');
```

Additionally, add a `waitForLiveTools` parameter that blocks the command
from resolving until the server is downloaded, started, and tools are availble.

Without this, a static sleep or poll would be needed.


src/vs/workbench/contrib/mcp/browser/mcpCommands.ts
```diff
@@ -821,13 +822,18 @@ export class StartServer extends Action2 {
 		});
 	}
 
-	async run(accessor: ServicesAccessor, serverId: string | undefined, opts?: IMcpServerStartOpts) {
+	async run(accessor: ServicesAccessor, serverId: string, opts?: IMcpServerStartOpts & { waitForLiveTools?: boolean }) {
```
rwoll, line 825: @connor4312 - I added this as an option — instead of changing the default behavior — to avoid risk of breaking any other code that might depend on the old behavior.

However, should we change the default so we don't need this branching?

connor4312 (approved):

Issue #276243, opened 2025-11-08
Agents View: support deletion of sessions


rebornix:
**Updated from @bpasero:**

Today, deleting a session has gaps:
* it only really works for local sessions, not background or cloud who have their own actions and implementation
* it does not stop a running session
* it probably needs to call an API that the provider can do cleanup

**Original from @rebornix:**

Re #274477

In the grouped agents view, extensions contribute their own “delete” button to the session entry. This can easily fall apart, for example, when users delete a session that’s already opened in the Chat Editor, the editor still stays open even though the session is gone. Different extensions may also choose different icons for “delete”.

This makes me wonder if “Archive” / “Delete” should be part of the core API.

DonJayamanne:
> elete": shouldn't that be something that we tell via API to the provider so that the provider

Agreed, feels weird for provider to do this
Also this way, core can delete all other related state if required.

> Note that I still respect the archived state from the

Not sure how this gets used/set today. Its not used in chat extension
@osortega /cc

PR #286642, opened 2026-01-09
Explore a controller based chat session item API

mjbvz:
For #276243

Explores moving the chat session item api to use a controller instead of a provider. This uses managed objects and is based on the `TestController` we already have finalized

Main flow :

- Extensions call `createChatSessionItemController('my.sessionId')` to create the controller. The controller is a managed object with a `dispose` method

- VS Code then calls `refreshHandler` on the controller to build up the initial list

- During refresh, the extension updates the `ChatSessionItemCollection` managed object. This object is basically just a managed array. This lets the editor see changes that are made to the session list

A few other points:

- Chat session items would now also be managed objects. Extensions could set properties to update them in the UI

- If an extension needs to update the list of chat session items independently of VS Code — for example if a new session is created through GitHub.com —  it would simply update the `ChatSessionItemCollection `

- We would have event handlers for when the editor archives or disposes of a chat session item

Still some open questions but seems promising. Also although the general API shape looks pretty different , the logic shouldn't be that different for extensions

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -899,6 +899,7 @@ export default tseslint.config(
 					],
 					'verbs': [
 						'accept',
+						'archive',
 						'change',
 						'close',
 						'collapse',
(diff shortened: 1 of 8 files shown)


bpasero (commented):
I have no objections going with this API considering we have a similar one that seems finalised. But I am also not deep into providing sessions via API, so I would yield to @joshspicer and maybe @DonJayamanne and @osortega for final approval.

DonJayamanne:
I'll try this out later today.

DonJayamanne (approved):

Issue #286579, opened 2026-01-08
Integrated Browser V1


unthinkmedia:
### Summary
Replace VS Code's Simple Browser with an integrated browser that supports AI-assisted web development workflows. Users can visually select DOM elements, capture screenshots, and export context (HTML, CSS, error logs) directly to AI tools for debugging, modification, and implementation assistance.Context/Background

### Developers frequently need to:
- Inspect live web pages and communicate specific elements to AI assistants
- Debug rendering issues and share visual + code context efficiently
- Iterate on UI changes with AI by providing precise element selection
- Switch between internal preview and external browser testing

Without integrated selection and context export, developers resort to manual copying of HTML, taking separate screenshots, and describing elements verbally—a slow, error-prone process that breaks flow.

### Description
Build an integrated browser extension for VS Code that maintains all Simple Browser functionality:

Core Capabilities (Simple Browser Parity)
- [ ] Launch browser via command palette, editor context menu, and custom UI
- [x] Navigate to any external URL with address bar
- [x] Open current page in external browser (Chrome, Edge, default system browser)
- Element Selection
  - [x] Element Selection Mode: Click to select DOM elements, with visual highlighting
  - [ ] Multi-element Selection: Hold modifier key (Ctrl/Cmd) to select multiple elements or maybe it is always continuous with no need to use a key.
  - [x] Context Export: Selected elements export as structured context including may include code and screenshots
  - [x] On hover show element details (default Chrome popover)
- [x] Ability to update settings
- [x] Open browser content in standalone VS Code window (Move into New Window / Copy into New Window)

New:
- [x] Ability to open more than one browser instance at a time (on Simple Browser, it's limited to a single instance)
- [ ] When in its own window have a simple way to send a prompt to the the main prompt area
- [x] Toggle Browser Developer Tools
- [x] Create a setting to choose between the old and new browser experiences, this setting will be used to drive the rollout
- [ ] Allow the browser to remember the last URL that was loaded into the browser
- [ ] Ability to open links from chat output
- [x] Create a null state for when the integrated browser get launched

Out of scope:
- Select an area of the browser canvas to take a screenshot of
- Testing responsiveness based on predetermined browser sizes
- When an error, have one click ability to send logs as context
- Bookmark URLs to help with forgetting what port things are on

Related:
- https://github.com/microsoft/vscode/issues/277298
- https://github.com/microsoft/vscode/issues/289410
- https://github.com/microsoft/vscode/issues/289413
[...]

PR #286649, opened 2026-01-09
"Add Element to Chat" in integrated browser

kycutler:
#286579


lszomoru (dismissed):

TylerLeonhardt (dismissed):

src/vs/workbench/contrib/browserView/electron-browser/browserEditor.ts
```diff
@@ -447,6 +461,98 @@ export class BrowserEditor extends EditorPane {
 		return this._model?.toggleDevTools();
 	}
 
+	/**
+	 * Select an element and add it to chat
+	 */
+	public async selectElement(): Promise<void> {
```
jruales: nit: should we remove the `public` keyword here to be consistent with the other functions above like `async toggleDevTools()` or `async reload()`?

src/vs/workbench/contrib/browserView/electron-browser/browserEditor.ts
```diff
@@ -447,6 +461,98 @@ export class BrowserEditor extends EditorPane {
...
+			const locator: IBrowserTargetLocator = { browserViewId: BrowserViewUri.getId(this.input.resource) };
+
+			// Start debug session for integrated browser
+			await this.browserElementsService.startDebugSession(cts.token, locator);
+
+			// Get the browser container bounds
+			const { width, height } = this._browserContainer.getBoundingClientRect();
+
+			// Get element data from browser
+			const elementData = await this.browserElementsService.getElementData({ x: 0, y: 0, width, height }, cts.token, locator);
```
jruales, line 499: nit: it would be nice if getElementData() got all elements for the visible view by default (for example, when passed in undefined) instead of having to pass the rectangle in explicitly for this simple case

src/vs/workbench/contrib/browserView/electron-browser/browserViewActions.ts
```diff
@@ -139,6 +139,33 @@ class ReloadAction extends Action2 {
...
+			title: localize2('browser.selectElementAction', 'Add Element to Chat'),
+			icon: Codicon.inspect,
+			f1: true,
+			precondition: ChatContextKeys.enabled,
+			toggled: CONTEXT_BROWSER_ELEMENT_SELECTION_ACTIVE,
+			menu: {
+				id: MenuId.BrowserActionsToolbar,
+				group: 'actions',
+				order: 1,
+				when: ChatContextKeys.enabled
```
jruales, line 157: To match the logic of the other elements of this menu, we should add a check for `BROWSER_EDITOR_ACTIVE` here too:

```suggestion
				when: ContextKeyExpr.and(BROWSER_EDITOR_ACTIVE, ChatContextKeys.enabled)
```

src/vs/workbench/contrib/browserView/electron-browser/browserViewActions.ts
```diff
@@ -139,6 +139,33 @@ class ReloadAction extends Action2 {
 	}
 }
 
+class SelectElementAction extends Action2 {
```
jruales: nit: Thoughts on renaming this and related selectElement function names to be clearer that this sends the selection to chat or AI, and not only selects an element? Something like SelectElementForChatContextAction or something like that

src/vs/platform/browserElements/common/browserElements.ts
```diff
@@ -15,20 +15,52 @@ export interface IElementData {
...
+	getElementData(rect: IRectangle, token: CancellationToken, locator: IBrowserTargetLocator, cancellationId?: number): Promise<IElementData | undefined>;
+
+	startDebugSession(token: CancellationToken, locator: IBrowserTargetLocator, cancelAndDetachId?: number): Promise<void>;
+}
+
+/**
+ * Extract a display name from outer HTML (e.g., "div#myId.myClass")
+ */
+export function getDisplayNameFromOuterHTML(outerHTML: string): string {
+	const firstElementMatch = outerHTML.match(/^<(\w+)([^>]*?)>/);
```
jruales: I think that an HTML element name can technically have a `-`, which wouldn't be caught with `\w`. That would apply for HTML custom elements defined via `customElements.define()`. Updated regex:


```suggestion
	const firstElementMatch = outerHTML.match(/^<([\w\-]+)([^>]*?)>/);
```

src/vs/workbench/contrib/browserView/electron-browser/browserEditor.ts
```diff
@@ -447,6 +461,98 @@ export class BrowserEditor extends EditorPane {
...
+			// Browser view URIs have a special scheme we can match against
+			const locator: IBrowserTargetLocator = { browserViewId: BrowserViewUri.getId(this.input.resource) };
+
+			// Start debug session for integrated browser
+			await this.browserElementsService.startDebugSession(cts.token, locator);
+
+			// Get the browser container bounds
+			const { width, height } = this._browserContainer.getBoundingClientRect();
+
+			// Get element data from browser
```
jruales: Should we mention here that this includes waiting for the user to select an element? Currently the comment makes it sound as if just the browser is involved.

src/vs/workbench/contrib/browserView/electron-browser/browserEditor.ts
```diff
@@ -447,6 +461,98 @@ export class BrowserEditor extends EditorPane {
...
+
+			// Prepare HTML/CSS context
+			const displayName = getDisplayNameFromOuterHTML(elementData.outerHTML);
+			let value = 'Attached HTML and CSS Context\n\n' + elementData.outerHTML;
+			if (this.configurationService.getValue('chat.sendElementsToChat.attachCSS')) {
+				value += '\n\n' + elementData.computedStyle;
+			}
+
+			toAttach.push({
+				id: 'element-' + Date.now(),
```
jruales, line 516: Question just for context: Do these IDs need to be orderable, or is the `Date.now()` just used to make this unique? If the latter, would generating a Guid also work the same?

src/vs/workbench/contrib/browserView/electron-browser/browserEditor.ts
```diff
@@ -447,6 +461,98 @@ export class BrowserEditor extends EditorPane {
...
+			if (this.configurationService.getValue('chat.sendElementsToChat.attachImages') && this._model) {
+				const screenshotBuffer = await this._model.captureScreenshot({
+					quality: 90,
+					rect: bounds
+				});
+
+				toAttach.push({
+					id: 'element-screenshot-' + Date.now(),
+					name: 'Element Screenshot',
+					fullName: 'Element Screenshot',
```
jruales, line 534: In case the user attaches multiple elements, do you think that the model would benefit from somehow being able to know which `element-<blah>` item corresponds to which `element-screenshot-<blah>` item? I'm thinking there's the possibility that the model gets confused about this screenshot corresponds to which element. But maybe the model has access to the order in which the elements are attached, in which case it might be fine.

src/vs/workbench/contrib/browserView/electron-browser/browserEditor.ts
```diff
@@ -447,6 +461,98 @@ export class BrowserEditor extends EditorPane {
...
+			if (!elementData) {
+				throw new Error('Element data not found');
+			}
+
+			const bounds = elementData.bounds;
+			const toAttach: IChatRequestVariableEntry[] = [];
+
+			// Prepare HTML/CSS context
+			const displayName = getDisplayNameFromOuterHTML(elementData.outerHTML);
+			let value = 'Attached HTML and CSS Context\n\n' + elementData.outerHTML;
```
jruales: This should only say "and CSS" if `'chat.sendElementsToChat.attachCSS'` is enabled

src/vs/workbench/contrib/browserView/electron-browser/browserEditor.ts
```diff
@@ -447,6 +461,98 @@ export class BrowserEditor extends EditorPane {
...
+				});
+			}
+
+			// Attach to chat widget
+			const widget = await this.chatWidgetService.revealWidget() ?? this.chatWidgetService.lastFocusedWidget;
+			widget?.attachmentModel?.addContext(...toAttach);
+
+		} catch (error) {
+			if (!cts.token.isCancellationRequested) {
+				this.logService.error('BrowserEditor.selectElement: Failed to select element', error);
```
jruales: Do we know for sure that the issue was during element selection and not when adding the element to chat? If not, then can we make this error mention the adding to chat as well?

jruales:
I noticed that the following item was present in Integrated Browser but not in Simple Browser when doing Select Element to chat. Is this expected?

```
/* Inherited from ancestor level 6 (regular) */
:root {
  --ea-container-md: 720px;
  --ea-container-lg: 960px;
  --ea-container-xl: 1040px;
  --ea-image-width: 120px;
  --ea-image-width-xs: 44px;
  --ea-image-placement-width: 180px;
  --ea-image-placement-width-horizontal: 320px;
  --ea-image-logo-width: 80px;
  --ea-fixedheader-height: 50px;
  --ea-font-size: 14px;
  --ea-font-family: -apple-system, BlinkMacSystemFont, Segoe UI, Roboto, Helvetica Neue, Arial, Noto Sans, sans-serif, Apple Color Emoji, Segoe UI Emoji, Segoe UI Symbol, Noto Color Emoji;
  --ea-bgcolor: rgba(0, 0, 0, 0.03);
  --ea-bgcolor-dark: rgba(255, 255, 255, 0.05);
  --ea-font-size-text-only-header: 18px;
  --ea-stylefixed-bgcolor: rgb(220, 220, 220);
  --ea-stylefixed-bgcolor-dark: rgb(80, 80, 80);
  --ea-color-link: rgb(80, 80, 80);
  --ea-color-link-dark: rgb(220, 220, 220);
  --ea-color-link-active: rgb(54.5, 54.5, 54.5);
  --ea-color-link-dark-active: rgb(245.5, 245.5, 245.5);
  --ea-color-link-callout: rgb(105.5, 105.5, 105.5);
  --ea-color-link-callout-dark: rgb(194.5, 194.5, 194.5);
  --ea-color-link-callout-active: #505050;
  --ea-color-link-callout-dark-active: gainsboro;
  --ea-color-link-bold: #088cdb;
  --ea-color-link-bold-dark: rgb(80.3788546256, 185.6299559471, 248.6211453744);
  --ea-color-domain: rgb(156.5, 156.5, 156.5);
  --ea-color-domain-dark: rgb(143.5, 143.5, 143.5);
  --ea-image-logo-filter: grayscale(1);
  --ea-image-logo-filter-dark: grayscale(1) invert(1);
  --ea-container-md: 720px;
  --ea-container-lg: 960px;
  --ea-container-xl: 1040px;
  --ea-image-width: 120px;
  --ea-image-width-xs: 44px;
  --ea-image-placement-width: 180px;
  --ea-image-placement-width-horizontal: 320px;
  --ea-image-logo-width: 80px;
  --ea-fixedheader-height: 50px;
  --ea-font-size: 14px;
  --ea-font-family: -apple-system, BlinkMacSystemFont, Segoe UI, Roboto, Helvetica Neue, Arial, Noto Sans, sans-serif, Apple Color Emoji, Segoe UI Emoji, Segoe UI Symbol, Noto Color Emoji;
  --ea-bgcolor: rgba(0, 0, 0, 0.03);
  --ea-bgcolor-dark: rgba(255, 255, 255, 0.05);
  --ea-font-size-text-only-header: 18px;
  --ea-stylefixed-bgcolor: rgb(220, 220, 220);
  --ea-stylefixed-bgcolor-dark: rgb(80, 80, 80);
  --ea-color-link: rgb(80, 80, 80);
  --ea-color-link-dark: rgb(220, 220, 220);
  --ea-color-link-active: rgb(54.5, 54.5, 54.5);
  --ea-color-link-dark-active: rgb(245.5, 245.5, 245.5);
  --ea-color-link-callout: rgb(105.5, 105.5, 105.5);
  --ea-color-link-callout-dark: rgb(194.5, 194.5, 194.5);
  --ea-color-link-callout-active: #505050;
  --ea-color-link-callout-dark-active: gainsboro;
  --ea-color-link-bold: #088cdb;
  --ea-color-link-bold-dark: rgb(80.3788546256, 185.6299559471, 248.6211453744);
  --ea-color-domain: rgb(156.5, 156.5, 156.5);
  --ea-color-domain-dark: rgb(143.5, 143.5, 143.5);
  --ea-image-logo-filter: grayscale(1);
  --ea-image-logo-filter-dark: grayscale(1) invert(1);
}
```

src/vs/platform/browserElements/common/browserElements.ts
```diff
@@ -15,20 +15,52 @@ export interface IElementData {
...
-	LiveServer = 'liveServer',
+/**
+ * Locator for identifying a browser target/webview.
+ * Uses either the parent webview or browser view id to uniquely identify the target.
+ */
+export interface IBrowserTargetLocator {
+	/**
+	 * Identifier of the parent webview hosting the target.
+	 *
+	 * Exactly one of {@link webviewId} or {@link browserViewId} should be provided.
```
jruales, line 26: should this be enforced via type checks?

src/vs/platform/browserElements/common/browserElements.ts
```diff
@@ -15,20 +15,52 @@ export interface IElementData {
...
 	readonly windowId: number;
 
-	getElementData(rect: IRectangle, token: CancellationToken, browserType: BrowserType, cancellationId?: number): Promise<IElementData | undefined>;
+	getElementData(rect: IRectangle, token: CancellationToken, locator: IBrowserTargetLocator, cancellationId?: number): Promise<IElementData | undefined>;
+
+	startDebugSession(token: CancellationToken, locator: IBrowserTargetLocator, cancelAndDetachId?: number): Promise<void>;
+}
+
+/**
+ * Extract a display name from outer HTML (e.g., "div#myId.myClass")
```
jruales:
```suggestion
 * Extract a display name from outer HTML (e.g., "div#myId.myClass1.myClass2")
```

src/vs/workbench/contrib/chat/browser/attachments/simpleBrowserEditorOverlay.ts
```diff
@@ -249,7 +249,12 @@ class SimpleBrowserOverlayWidget {
 		const editorContainer = this._container.querySelector('.editor-container') as HTMLDivElement;
 		const editorContainerPosition = editorContainer ? editorContainer.getBoundingClientRect() : this._container.getBoundingClientRect();
 
-		const elementData = await this._browserElementsService.getElementData(editorContainerPosition, cts.token, this._activeBrowserType);
+		const elementData = await this._browserElementsService.getElementData({
+			x: editorContainerPosition.x,
+			y: editorContainerPosition.y + 32.4, // Height of the title bar
```
jruales, line 254: Speaking of the title bar height, I noticed that the title bar is now a bit taller than it was in Simple Browser. Is that intentional? Or should we keep the same, more compact height used in Simple Browser?

src/vs/workbench/contrib/chat/browser/attachments/simpleBrowserEditorOverlay.ts
jruales: Same here, the "and CSS" shouldn't appear when `chat.sendElementsToChat.attachCSS` is disabled

src/vs/workbench/contrib/browserView/electron-browser/browserEditor.ts
justschen: iirc attachments just have to be unique. for images, we do guid or hash instead of date, but not sure how necessary it is

jruales (dismissed):

src/vs/workbench/contrib/chat/browser/attachments/simpleBrowserEditorOverlay.ts
kycutler: Personally I don't like how short the title bar is in simple browser -- it feels cramped to me. Will leave this for feedback

src/vs/workbench/contrib/browserView/electron-browser/browserEditor.ts
kycutler: This rectangle is really just used for the returned `elementData.bounds` (which is just used for the screenshot). This essentially specifies where the viewport is relative to the view that the screenshot will come from so the bounds can be calculated relative to that. For webviews the screenshot is taken from the main window, so we provide the DOM rect of the iframe within the workbench. But here the screenshot is taken directly from the page, so we just provide the full viewport

src/vs/workbench/contrib/browserView/electron-browser/browserEditor.ts
kycutler: Will leave this as-is for now

src/vs/workbench/contrib/browserView/electron-browser/browserViewActions.ts
kycutler: That should actually be removed from the other actions (will do in a separate PR). The problem is that if you trigger the action (e.g. click back in the menu bar) when the editor isn't selected (such as in split view), the action should still trigger but with this when clause it doesn't.

src/vs/workbench/contrib/browserView/electron-browser/browserEditor.ts
kycutler: Aah I keep doing this 😄 removed, thanks!

src/vs/workbench/contrib/browserView/electron-browser/browserViewActions.ts
kycutler: Fair point, updated

src/vs/platform/browserElements/common/browserElements.ts
kycutler: Updated to just read until either a space or a closing `>`

src/vs/platform/browserElements/common/browserElements.ts
kycutler: I initially tried this but decided the code is cleaner this way

src/vs/workbench/contrib/browserView/electron-browser/browserEditor.ts
kycutler: Agree it could possibly get confused, although I think it can tell the order of the attachments. Ideally I think we could have both the text and image in one attachment, but AFAIK that isn't possible with the current APIs. Will leave it as-is for now

src/vs/workbench/contrib/browserView/electron-browser/browserEditor.ts
kycutler: I suppose so -- though we should get an idea based on the logged error. I'm not too worried about it honestly, will leave it for now

jruales (approved):

src/vs/workbench/contrib/browserView/electron-browser/browserEditor.ts
```diff
@@ -447,6 +461,99 @@ export class BrowserEditor extends EditorPane {
...
+				throw new Error('Element data not found');
+			}
+
+			const bounds = elementData.bounds;
+			const toAttach: IChatRequestVariableEntry[] = [];
+
+			// Prepare HTML/CSS context
+			const displayName = getDisplayNameFromOuterHTML(elementData.outerHTML);
+			const attachCss = this.configurationService.getValue<boolean>('chat.sendElementsToChat.attachCSS');
+			let value = (attachCss ? 'Attached HTML and CSS Context' : 'Attached HTML Context') + '\n\n' + elementData.outerHTML;
```
jruales, line 510: I'm assuming this is intentionally not localized since it's meant for the AI model?

kycutler: Correct

PR #286767, opened 2026-01-09
chat: support sendToNewChat option and SendToNewChatAction delegation…

### LINKED ISSUE #286724: Chat: support an action to send a chat request and clear widget
From our sessions sync, we raised the idea to allow to send a chat request and immediately return to an empty Chat so that you can essentially just fire of a chat request and then immediately another one without having to see the response.

We could wire this as alternative action into the execute one we have today:

https://github.com/user-attachments/assets/dad24cb3-90e0-4fa3-902c-71c54b93ac43

(the alternate action we have today allows to start a new chat)

@roblourens I tried to hack this in but failed so far, is this even supported from the API of the chat widget? I know I can call `widget.acceptInput` to start a chat, but it would need some kind of option to then proceed in the "background".

//cc @jo-oikawa

sandorsett:
fixes #286724
This change implements a new option so callers can submit a chat request and immediately return the UI to an empty/new chat view (submit in background). It addresses [https://github.com/microsoft/vscode/issues/286724](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html).

Adds [IChatAcceptInputOptions.sendToNewChat?: boolean](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html).
When set, [ChatWidget](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) captures the session, cancels any in-flight request for that session, clears the widget (returns UI to empty chat), and submits the request to the original session resource in the background.
Updates [SendToNewChatAction](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) to delegate to [widget.acceptInput(..., { sendToNewChat: true })](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) instead of manual clear+submit.
Adds unit tests validating delegation and the [_acceptInput](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) send+clear behavior.

### DIFF
--- a/src/vs/workbench/contrib/chat/browser/chat.ts
+++ b/src/vs/workbench/contrib/chat/browser/chat.ts
@@ -212,6 +212,11 @@ export interface IChatAcceptInputOptions {
 	// box's current content is being accepted, or 'false' if a specific input
 	// is being submitted to the widget.
 	storeToHistory?: boolean;
+	/**
+	 * If true, submit the input to the associated session and clear the widget (show empty/new chat) immediately.
+	 * This allows sending a request 'in the background' while returning the UI to an empty chat view.
+	 */
+	sendToNewChat?: boolean;
 }
 
 export interface IChatWidgetViewModelChangeEvent {
--- a/src/vs/workbench/contrib/chat/browser/actions/chatExecuteActions.ts
+++ b/src/vs/workbench/contrib/chat/browser/actions/chatExecuteActions.ts
@@ -634,19 +634,14 @@ class SendToNewChatAction extends Action2 {
 
 		const inputBeforeClear = widget.getInput();
 
-		// Cancel any in-progress request before clearing
-		if (widget.viewModel) {
-			chatService.cancelCurrentRequestForSession(widget.viewModel.sessionResource);
-		}
-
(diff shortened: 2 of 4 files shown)


src/vs/workbench/contrib/chat/browser/chat.ts
```diff
@@ -212,6 +212,11 @@ export interface IChatAcceptInputOptions {
 	// box's current content is being accepted, or 'false' if a specific input
 	// is being submitted to the widget.
 	storeToHistory?: boolean;
+	/**
+	 * If true, submit the input to the associated session and clear the widget (show empty/new chat) immediately.
+	 * This allows sending a request 'in the background' while returning the UI to an empty chat view.
+	 */
+	sendToNewChat?: boolean;
```
roblourens, line 219: Isn't this name backwards? It's not sending _to_ a new chat, it's sending to the current chat then opening a new chat... I will have to think of a catchy name

PR #286785, opened 2026-01-09
clean up edit processing, add new tests

amunger:
follow up to drop that any type, but also reduced some complexity within that branch. Written with opus, reviewed by me.


connor4312 (approved):
thanks! 🧹

PR #286812, opened 2026-01-09
feat-internal: Export Chat as Zip with Repository Context

### LINKED ISSUE #283943: Include workspace info in chat session export
We have logged [repo info in telemetry](https://github.com/microsoft/vscode-copilot-chat/blob/main/src/extension/prompt/node/repoInfoTelemetry.ts), but we also need to include similar workspace info (e.g. commit hash, diff) in exported artifacts during exporting chat session.

zhichli:
Adds "Export Chat as Zip" action to package chat sessions with repository context for benchmark creation.

## What it does
- Creates zip containing `chat.json` (session data) and `repos/*.json` (repo state per workspace folder)
- Captures git state: remote URL, branch, commits, working tree diffs
- Diffs stored as unified patches compatible with `git apply`

## Workspace Scenarios
| Scenario | workspaceType | syncStatus | Reproducibility |
|----------|---------------|------------|-----------------|
| Synced with remote | remote-git | synced | ✅ Clone + checkout |
| Unpushed commits | remote-git | unpushed | TBD: git bundle |
| Unpublished branch | remote-git | unpublished | TBD: git bundle |
| Local-only git | local-git | local-only | TBD: folder bundle |
| No git | plain-folder | no-git | TBD: folder bundle |

**This PR implements the synced scenario.** Other scenarios require git bundle support (future work).

## Sample repo.json
```json
{
  "workspaceType": "remote-git",
  "syncStatus": "synced",
  "remoteUrl": "https://github.com/example-org/my-project.git",
  "remoteVendor": "github",
  "remoteTrackingBranch": "origin/feature/chat-export",
  "remoteBaseBranch": "origin/main",
  "remoteHeadCommit": "abc123def456",
  "localBranch": "feature/chat-export",
  "localHeadCommit": "abc123def456",
  "diffs": [
    {
      "relativePath": "src/utils/helper.ts",
      "changeType": "modified",
      "status": "M",
      "unifiedDiff": "--- a/src/utils/helper.ts\n+++ b/src/utils/helper.ts\n@@ -10,6 +10,7 @@\n function helper() {\n+  console.log('debug');\n   return true;\n }"
    }
  ]
}
```

### DIFF
--- a/src/vs/platform/native/common/native.ts
+++ b/src/vs/platform/native/common/native.ts
@@ -235,10 +235,10 @@ export interface ICommonNativeHostService {
 	/**
 	 * Creates a zip file at the specified path containing the provided files.
 	 *
-	 * @param zipPath The absolute path where the zip file should be created.
+	 * @param zipPath The URI where the zip file should be created.
 	 * @param files An array of file entries to include in the zip, each with a relative path and string contents.
 	 */
-	createZipFile(zipPath: string, files: { path: string; contents: string }[]): Promise<void>;
+	createZipFile(zipPath: URI, files: { path: string; contents: string }[]): Promise<void>;
 }
 
 export const INativeHostService = createDecorator<INativeHostService>('nativeHostService');
--- a/src/vs/platform/native/common/native.ts
+++ b/src/vs/platform/native/common/native.ts
@@ -232,6 +232,12 @@ export interface ICommonNativeHostService {
 	windowsGetStringRegKey(hive: 'HKEY_CURRENT_USER' | 'HKEY_LOCAL_MACHINE' | 'HKEY_CLASSES_ROOT' | 'HKEY_USERS' | 'HKEY_CURRENT_CONFIG', path: string, name: string): Promise<string | undefined>;
 
 	// Zip
+	/**
+	 * Creates a zip file at the specified path containing the provided files.
+	 *
+	 * @param zipPath The absolute path where the zip file should be created.
+	 * @param files An array of file entries to include in the zip, each with a relative path and string contents.
+	 */
 	createZipFile(zipPath: string, files: { path: string; contents: string }[]): Promise<void>;
 }
 
--- a/src/vs/platform/native/common/native.ts
+++ b/src/vs/platform/native/common/native.ts
@@ -230,6 +230,9 @@ export interface ICommonNativeHostService {
 
 	// Registry (Windows only)
 	windowsGetStringRegKey(hive: 'HKEY_CURRENT_USER' | 'HKEY_LOCAL_MACHINE' | 'HKEY_CLASSES_ROOT' | 'HKEY_USERS' | 'HKEY_CURRENT_CONFIG', path: string, name: string): Promise<string | undefined>;
+
+	// Zip
+	createZipFile(zipPath: string, files: { path: string; contents: string }[]): Promise<void>;
 }
 
 export const INativeHostService = createDecorator<INativeHostService>('nativeHostService');
--- a/src/vs/platform/native/electron-main/nativeHostMainService.ts
+++ b/src/vs/platform/native/electron-main/nativeHostMainService.ts
@@ -1171,8 +1171,8 @@ export class NativeHostMainService extends Disposable implements INativeHostMain
 
 	//#region Zip
 
-	async createZipFile(windowId: number | undefined, zipPath: string, files: { path: string; contents: string }[]): Promise<void> {
-		await zip(zipPath, files);
+	async createZipFile(windowId: number | undefined, zipPath: URI, files: { path: string; contents: string }[]): Promise<void> {
+		await zip(zipPath.fsPath, files);
 	}
 
 	//#endregion
--- a/src/vs/platform/native/electron-main/nativeHostMainService.ts
+++ b/src/vs/platform/native/electron-main/nativeHostMainService.ts
@@ -43,6 +43,7 @@ import { IV8Profile } from '../../profiling/common/profiling.js';
 import { IAuxiliaryWindowsMainService } from '../../auxiliaryWindow/electron-main/auxiliaryWindows.js';
 import { IAuxiliaryWindow } from '../../auxiliaryWindow/electron-main/auxiliaryWindow.js';
 import { CancellationError } from '../../../base/common/errors.js';
+import { zip } from '../../../base/node/zip.js';
 import { IConfigurationService } from '../../configuration/common/configuration.js';
 import { IProxyAuthService } from './auth.js';
 import { AuthInfo, Credentials, IRequestService } from '../../request/common/request.js';
@@ -1171,7 +1172,6 @@ export class NativeHostMainService extends Disposable implements INativeHostMain
 	//#region Zip
 
 	async createZipFile(windowId: number | undefined, zipPath: string, files: { path: string; contents: string }[]): Promise<void> {
-		const { zip } = await import('../../../base/node/zip.js');
 		await zip(zipPath, files);
 	}
 
--- a/src/vs/platform/native/electron-main/nativeHostMainService.ts
+++ b/src/vs/platform/native/electron-main/nativeHostMainService.ts
@@ -1168,6 +1168,15 @@ export class NativeHostMainService extends Disposable implements INativeHostMain
 
 	//#endregion
 
+	//#region Zip
+
+	async createZipFile(windowId: number | undefined, zipPath: string, files: { path: string; contents: string }[]): Promise<void> {
+		const { zip } = await import('../../../base/node/zip.js');
+		await zip(zipPath, files);
+	}
+
+	//#endregion
+
 	private windowById(windowId: number | undefined, fallbackCodeWindowId?: number): ICodeWindow | IAuxiliaryWindow | undefined {
 		return this.codeWindowById(windowId) ?? this.auxiliaryWindowById(windowId) ?? this.codeWindowById(fallbackCodeWindowId);
 	}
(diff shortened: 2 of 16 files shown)


src/vs/workbench/contrib/chat/electron-browser/actions/chatExportZip.ts
```diff
@@ -0,0 +1,122 @@
...
+				} else {
+					notificationService.notify({
+						severity: Severity.Warning,
+						message: localize('chatExportZip.noRepoData', "Exported chat without repository context. No Git repository was detected.")
+					});
+				}
+			}
+
+			try {
+				await nativeHostService.createZipFile(result.fsPath, files);
```
roblourens: Let's pass around URIs, not string paths

src/vs/workbench/contrib/chat/browser/chatRepoInfo.ts
```diff
@@ -0,0 +1,387 @@
...
+	) {
+		super();
+
+		this._register(this._pendingSessions);
+
+		this._register(this.chatService.onDidCreateModel(async (model) => {
+			if (model.repoData) {
+				return;
+			}
+			await this.captureAndSetRepoData(model);
```
roblourens: It seems like this always captures the repo data for every chat model, every user? But this seems like it coudl be really expensive? Should it only happen when the user requests it?

src/vs/workbench/contrib/chat/browser/chatRepoInfo.ts
```diff
@@ -0,0 +1,387 @@
...
+		disposables.add(model.onDidDispose(() => {
+			disposables.dispose();
+		}));
+	}
+
+	private async captureAndSetRepoData(model: IChatModel): Promise<void> {
+		try {
+			const repoData = await captureRepoInfo(this.scmService, this.fileService);
+			if (repoData) {
+				model.setRepoData(repoData);
```
roblourens, line 544: It seems like we do this multiple times per ChatModel, and just keep replacing the previous repo data? I don't understand that part.

src/vs/workbench/contrib/chat/browser/chatRepoInfo.ts
```diff
@@ -0,0 +1,387 @@
...
+			}
+			await this.captureAndSetRepoData(model);
+			if (model.repoData) {
+				disposables.dispose();
+			}
+		}));
+
+		// Clean up when the model is disposed
+		disposables.add(model.onDidDispose(() => {
+			disposables.dispose();
```
roblourens: You dispose disposables but it's still in `_pendingSessions` so it would leak the memory

src/vs/workbench/contrib/chat/common/chatService/chatServiceImpl.ts
```diff
@@ -157,6 +160,9 @@ export class ChatService extends Disposable implements IChatService {
 		this._register(this._sessionModels.onDidDisposeModel(model => {
 			this._onDidDisposeSession.fire({ sessionResource: [model.sessionResource], reason: 'cleared' });
 		}));
+		this._register(this._sessionModels.onDidCreateModel(model => {
+			this._onDidCreateModel.fire(model);
```
roblourens: I think you can expose `this._sessionModels.onDidCreateModel` rather than create a new event emitter. eg `onDidCreateModel = this._sessionModels.onDidCreateModel`

roblourens: Maybe it has to be returned from a getter

src/vs/workbench/contrib/chat/browser/chat.contribution.ts
```diff
@@ -1199,6 +1200,7 @@ registerWorkbenchContribution2(ChatEditingNotebookFileSystemProviderContrib.ID,
 registerWorkbenchContribution2(UserToolSetsContributions.ID, UserToolSetsContributions, WorkbenchPhase.Eventually);
 registerWorkbenchContribution2(PromptLanguageFeaturesProvider.ID, PromptLanguageFeaturesProvider, WorkbenchPhase.Eventually);
 registerWorkbenchContribution2(ChatWindowNotifier.ID, ChatWindowNotifier, WorkbenchPhase.AfterRestored);
+registerWorkbenchContribution2(ChatRepoInfoContribution.ID, ChatRepoInfoContribution, WorkbenchPhase.AfterRestored);
```
roblourens: Can it be `.Eventually`?

src/vs/platform/native/common/native.ts
```diff
@@ -230,6 +230,15 @@ export interface ICommonNativeHostService {
...
 	windowsGetStringRegKey(hive: 'HKEY_CURRENT_USER' | 'HKEY_LOCAL_MACHINE' | 'HKEY_CLASSES_ROOT' | 'HKEY_USERS' | 'HKEY_CURRENT_CONFIG', path: string, name: string): Promise<string | undefined>;
+
+	// Zip
+	/**
+	 * Creates a zip file at the specified path containing the provided files.
+	 *
+	 * @param zipPath The absolute path where the zip file should be created.
+	 * @param files An array of file entries to include in the zip, each with a relative path and string contents.
+	 */
+	createZipFile(zipPath: string, files: { path: string; contents: string }[]): Promise<void>;
```
roblourens: @bpasero for the zip file stuff, maybe this would be better in `IFileService` or something

bpasero: No, this is easier.

src/vs/platform/native/electron-main/nativeHostMainService.ts
```diff
@@ -1168,6 +1168,15 @@ export class NativeHostMainService extends Disposable implements INativeHostMain
 
 	//#endregion
 
+	//#region Zip
+
+	async createZipFile(windowId: number | undefined, zipPath: string, files: { path: string; contents: string }[]): Promise<void> {
+		const { zip } = await import('../../../base/node/zip.js');
```
bpasero: No need to lazy import this.

bpasero (changes requested):
Left a minor comment, no objections having a zip method in native host service, thats easier than having it anywhere else.

src/vs/workbench/contrib/chat/browser/chatRepoInfo.ts
zhichli: changed to capture on first request only

src/vs/workbench/contrib/chat/browser/chatRepoInfo.ts
zhichli: changed to capture on first request only

src/vs/workbench/contrib/chat/browser/chatRepoInfo.ts
roblourens: But this is still running for every user by default? Does it need to work that way with no opt-in, or am I wrong for thinking it might be expensive?

zhichli: You’re right, it shouldn’t run for everyone due to the performance impact like git diff computation. Thanks for catching that—I should have included an internal user check on capture similar to what was done for the export action. I’ve now updated ChatRepoInfoContribution to add an IChatEntitlementService.isInternal check before capturing any data. Do you think we should add an opt-in experimental config key for internal users to enable this capture?

roblourens: What's the scenario you expect? I also don't want team members to set this setting, forget about it, then suffer from a slow process. Could it be a command that I run which then only affects the _next_ chat session?

zhichli: I verified that capturing logic is not blocking UI at least and is mostly doing the similar information capture as we do in https://github.com/microsoft/vscode-copilot-chat/blob/main/src/extension/prompt/node/repoInfoTelemetry.ts#L82 which also does similar things at the beginning of the chat:
- Only sent for internal Microsoft users (checks isInternal)
- Combines diffs from: upstream diff, working tree changes, and untracked changes
- Has guards against: >100 file changes, >900KB diff size, and files changing during collection
- Uses a file watcher to detect if files change during async diff operations (invalidates the data)

How about we also adding similar guard checks here ? 

> Could it be a command that I run which then only affects the next chat session?

Ideally users can decide to capture the session with repo info at the end of session so this won't help in that case. But it helps if users wants to capture a session before it happens though.

roblourens: Ah. One difference is that this would keep the content in memory forever, while the telemetry code would send an event and then let the memory be GC'd, I assume. Is that 900 KB total for all the diffs?

zhichli: 900 KB is the limit for all diffs combined (aligned with telemetry's 1MB payload limit).

Keeping data in memory is a valid concern. As far as I understand, chat sessions use reference counting and are disposed when not actively referenced—so data isn't held "forever" in memory. The actual persistence is to disk via `ChatSessionStore` (capped at 25 sessions).
- That said, I've added a limit to keep full diffs only for the 5 most recent sessions. Older sessions (6-25) have their diffs trimmed but retain metadata (commit hash, branch, etc.) with `diffsStatus: 'trimmedForStorage'` so exports can indicate what happened. This balances storage concerns with the most common export use case (recent sessions).

I've also added similar guards to what telemetry uses:

- `>100 file changes` -> `diffsStatus: 'tooManyChanges'`
- `>900 KB diff size` -> `diffsStatus: 'tooLarge'`

zhichli: added `chat.repoInfo.enabled` key to control this behavior with default as `true`

roblourens (approved):

bpasero (approved):

Issue #279302, opened 2025-11-25
[folder] Add empty line automatically


selonya:
When the last class/function is collapsed, it's impossible to add a line by selecting the last character and pressing enter. You have to expand it, scroll to be bottom, put the cursor at the end and press enter. It would be much easier if the editor added an empty line auotmatically upon collapse, if the last class in the editor has no empty line at the end.

selonya:
I tried files.insertFinalNewline. Yes, it creates the new line, but funny thing is, when the last class is collapsed you cannot see the new line. You have to expand it to see it. Another trick is to press Ctrl+End and then hit enter. But if the collapsed function is in the middle of the document, we need something else.

Wouldn't it be more reasonable if you put the cursor at the end of the collapsed function and hit enter, and it would create a new line at the end of that block/function? Right now it creates a new line inside the collapsed function. I don't think anybody would need it this way... IMO

https://ibb.co/Ng8KT7Kj

akyTheDev:
Hey @aeschli  @selonya ! I hope you are doing well 🙂 I've just submitted a PR to address this issue. I’d love to get your thoughts on the implementation whenever you have a moment. Looking forward to your feedback!

PR #286834, opened 2026-01-09
feat(editor): add option to insert newline when folding last line

akyTheDev:
Introduce a new editor option `foldingAddNewlineAtEnd` that controls whether a newline character is added when folding the last line of a file. This option defaults to true, ensuring that a newline is inserted to maintain a clean separation at the end of the file when the last line is folded.

The implementation includes updates to the folding model to check for this option and conditionally insert a newline. Additionally, tests have been added to verify the behavior of this new feature under various scenarios, including when the option is disabled or when the last line is already empty.

#279302


akyTheDev:
> ```
> @microsoft-github-policy-service agree
> ```

RedCMD:
should send screenshot/screen recording

PR #286850, opened 2026-01-10
fix remote terminal completions

### LINKED ISSUE #286155: File completions aren't working for WSL
Forking from https://github.com/microsoft/vscode/issues/285577#issuecomment-3713735898

Image

Image

I'm also seeing this on 1.106 so it's separate to the https://github.com/microsoft/vscode/issues/285577 work

Image

Version: 1.106.3 (user setup)
Commit: bf9252a2fb45be6893dd8870c0bf37e2e1766d61
Date: 2025-11-25T22:28:18.024Z
Electron: 37.7.0
ElectronBuildId: 12781156
Chromium: 138.0.7204.251
Node.js: 22.20.0
V8: 13.8.258.32-electron.0
OS: Windows_NT x64 10.0.26200

meganrogge:
Debugging showed the completions were getting created correctly, but then were filtered out. 

This was happening because we were using the local path sep instead of the one from the extension host. The fix is to pass `pathSep` into `TerminalCompletionItem`. 

fixes https://github.com/microsoft/vscode/issues/286155

### DIFF
--- a/src/vs/workbench/contrib/terminalContrib/suggest/browser/terminalCompletionItem.ts
+++ b/src/vs/workbench/contrib/terminalContrib/suggest/browser/terminalCompletionItem.ts
@@ -124,19 +124,19 @@ export class TerminalCompletionItem extends SimpleCompletionItem {
 	constructor(
 		override readonly completion: ITerminalCompletion,
 		/**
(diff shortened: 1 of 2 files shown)


Tyriar (approved):
Possible to write a test for this?

PR #286866, opened 2026-01-10
Fix missing stop button after elicitation is hidden

### LINKED ISSUE #283328: No stop button to interrupt agent
The AI keeps editing, but I cannot interrupt it without reloading the window:

Image

roblourens:
Fix #283328

@meganrogge When the elicitation part is automatically hidden, I guess we leave the state on Pending. ~~This fixes it but if I update the state, does it matter to you whether it's Rejected or Accepted or should it be another ElicitationState?~~

edit- I'm marking the elicitation as Rejected here, that seems fine, lmk if you think we should do something different though

cc @connor4312


src/vs/workbench/contrib/chat/common/model/chatProgressTypes/chatElicitationRequestPart.ts
```diff
@@ -54,7 +51,7 @@ export class ChatElicitationRequestPart extends Disposable implements IChatElici
 		}
 		this._isHiddenValue.set(true, undefined, undefined);
 		this.onHide?.();
-		this.dispose();
+		this.state.set(ElicitationState.Rejected, undefined);
```
connor4312: A hidden part is not necessarily rejected. An elicitaiton could be accepted and then hidden. This should only happen if the state was still pending

connor4312 (changes requested):

connor4312 (approved):

meganrogge (approved):

PR #286872, opened 2026-01-10
Add .mpp and .mxx extensions to C++

damster101:
These are common extensions for C++20 module files. They are the default C++ module extensions of build systems xmake (.mpp) and build2 (.mxx), for example.


aeschli (approved):

sandy081 (approved):

chrmarti (approved):

dmitrivMS:
Will recreate since community changes to package.json are not allowed.

PR #286942, opened 2026-01-10
Support XDG_CONFIG_HOME for argv.json file on Linux

### LINKED ISSUE #162712: Proposal: minimal changes to allow XDG-compatible data storage.
There are many of us who find it bad practice that VSCode creates a `~/.vscode` folder instead of following the common [XDG base directory structure](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html), respected by a wide variety of programs. This issue has come up regularly, without a good permanent solution:

- https://github.com/microsoft/vscode/issues/3884
- https://github.com/microsoft/vscode/issues/84808
- https://github.com/microsoft/vscode/issues/100988
- https://github.com/microsoft/vscode/issues/17691
- https://github.com/microsoft/vscode/issues/126182

I would like to propose the following new behaviour:

1. If `~/.config/[vs]code[-insiders]/argv.json` is present at launch in non-portable mode, then VSCode uses this as the path for `argv.json`.
  - The brackets denote prefix and suffix of the installed build, per https://github.com/microsoft/vscode/issues/3884 . This is deterministic for any given installed build, and would be `~/.config/vscode/argv.json` for "most" people.
2. `argv.json` supports setting the following arguments, which the user can set to XDG-compatible paths if they like:
  - `user-data-dir`
  - `extensions-dir`

As far as I can tell, this would be sufficient to avoid creating a `~/.vscode` folder. It also reuses the existing structure of VSCode to do this without introducing significant hacks, edge cases, or race conditions.

Part 1 seems reasonably feasible to me, given that `argv.json` is hardcoded in delightfully few places at the moment: https://github.com/microsoft/vscode/search?q=%22argv.json%22&type=code

Part 2 would essentially involve adding those arguments to [`SUPPORTED_ELECTRON_SWITCHES`](https://github.com/microsoft/vscode/blob/9db57e76e9f16f552203585a394e50cf0f7a84a7/src/main.js#L153).

I'd be happy to submit a PR if this kind of approach is amenable to the maintainers.

AndyBodnar:
## Summary

This PR adds XDG Base Directory Specification support for the `argv.json` configuration file on Linux.

- Respects `XDG_CONFIG_HOME` environment variable when determining the path for `argv.json`
- Falls back to `~/.config` if `XDG_CONFIG_HOME` is not set
- Changes apply to both early startup (`src/main.ts`) and the environment service (`src/vs/platform/environment/common/environmentService.ts`)

Previously, VS Code stored `argv.json` in `~/.vscode/argv.json` on Linux. With this change:
- If `XDG_CONFIG_HOME` is set: `$XDG_CONFIG_HOME/Code/argv.json`
- If `XDG_CONFIG_HOME` is not set: `~/.config/Code/argv.json`

This aligns with how VS Code already handles `userDataPath` (which respects `XDG_CONFIG_HOME`) and follows the XDG Base Directory Specification.

Fixes #162712

## Test plan

- [ ] On Linux with `XDG_CONFIG_HOME` set, verify `argv.json` is created/read from `$XDG_CONFIG_HOME/Code/argv.json`
- [ ] On Linux without `XDG_CONFIG_HOME` set, verify `argv.json` is created/read from `~/.config/Code/argv.json`
- [ ] On Windows and macOS, verify behavior is unchanged
- [ ] Verify portable mode (`VSCODE_PORTABLE`) still takes precedence

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -442,6 +442,15 @@ function getArgvConfigPath(): string {
 		dataFolderName = `${dataFolderName}-dev`;
 	}
 
+	// On Linux, respect XDG_CONFIG_HOME as per XDG Base Directory Specification
+	if (process.platform === 'linux') {
+		const xdgConfigHome = process.env['XDG_CONFIG_HOME'];
+		if (xdgConfigHome) {
+			return path.join(xdgConfigHome, dataFolderName!, 'argv.json');
+		}
+		return path.join(os.homedir(), '.config', dataFolderName!, 'argv.json');
+	}
+
 	return path.join(os.homedir(), dataFolderName!, 'argv.json');
 }
 
--- a/src/main.ts
+++ b/src/main.ts
@@ -254,6 +254,18 @@ function configureCommandlineSwitchesSync(cliArgs: NativeParsedArgs) {
 		'enable-rdp-display-tracking',
 	];
 
+	// Path switches that can be configured via argv.json to support XDG Base Directory Specification
+	// These are merged into CLI args (CLI takes precedence) to allow configuration without command line
+	// See: https://github.com/microsoft/vscode/issues/162712
+	const SUPPORTED_PATH_SWITCHES = [
+
+		// Custom user data directory path
+		'user-data-dir',
+
+		// Custom extensions directory path
+		'extensions-dir',
+	];
+
 	// Read argv config
 	const argvConfig = readArgvConfigSync();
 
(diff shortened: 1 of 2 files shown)


AndyBodnar:
Thank you for the feedback, @bpasero.

You raise a valid point. This PR provides **partial XDG compliance** by supporting `XDG_CONFIG_HOME` for `argv.json` specifically - it does not address the full scope of issue #162712, which also requests XDG support for `user-data-dir` and `extensions-dir`.

That said, I believe this is a valuable incremental step:

1. **Common pain point**: The `argv.json` location is frequently mentioned by users who want to sync VS Code configuration across machines or keep their home directory clean.

2. **Lower risk**: Focusing on `argv.json` alone is a smaller, more contained change that can be evaluated independently.

3. **Foundation for future work**: Full XDG support for `user-data-dir` and `extensions-dir` could be addressed in a follow-up PR if the team is interested in pursuing complete XDG compliance.

I'm happy to update the PR description to clarify this is partial XDG support, or to discuss expanding the scope if that's preferred.

btardif:
Thanks for working on this. Is there anything remaining that would help move the PR forward? I am happy to test the migration and fallback behavior in a clean environment if that would be useful.

AndyBodnar:
Its been a while, I will recheck where we are at on everything and see if theres anything I can do to help.

PR #279093, opened 2025-11-24
fix: memory leak in task problem monitor

SimonSiefke:
Fixes a memory leak in task problem monitor. 

### Overview

It seems a problem matcher can be added to a terminal again, e.g. when re-running a task, this function would be called again with the same terminal since terminals are reused between tasks

```ts
addTerminal(terminal: ITerminalInstance, problemMatcher: AbstractProblemCollector) {
	this.terminalMarkerMap.set(terminal.instanceId, {
		resources: new Map<string, URI>(),
		markers: new Map<string, Map<string, IMarkerData>>()
	});

	const store = new DisposableStore();
	this.terminalDisposables.set(terminal.instanceId, store); 
```

This line `this.terminalDisposables.set(terminal.instanceId, store)` registers the store for the terminal, but doesn't dispose the previous registered store if there is one.

### Change

The change is to use a `DisposableMap` which ensures the already registered disposable gets disposed if there is one.

### Before

When running a task 97 times, the number of functions grows by each time in several places, one of which is task problem monitor.

```jsonc

{
  "namedFunctionCount3": [
	/* some more data omitted */
  {
      "count": 99,
      "delta": 97,
      "name": "anonymous",
      "sourceLocation": "out/vs/workbench/workbench.desktop.main.js:3019:8513",
      "originalLocation": "src/vs/workbench/contrib/tasks/browser/taskTerminalStatus.ts:75:65",
      "originalName": "TaskTerminalStatus.addTerminal"
    },
    {
      "count": 99,
      "delta": 97,
      "name": "anonymous",
      "sourceLocation": "out/vs/workbench/workbench.desktop.main.js:3019:11261",
      "originalLocation": "src/vs/workbench/contrib/tasks/browser/taskProblemMonitor.ts:35:32",
      "originalName": "TaskProblemMonitor.addTerminal"
    },
    {
      "count": 99,
      "delta": 97,
      "name": "anonymous",
      "sourceLocation": "out/vs/workbench/workbench.desktop.main.js:3021:11006",
      "originalLocation": "src/vs/workbench/contrib/tasks/browser/terminalTaskSystem.ts:1051:60",
      "originalName": "TerminalTaskSystem._executeInTerminal"
    },
    {
      "count": 99,
      "delta": 97,
      "name": "anonymous",
      "sourceLocation": "out/vs/workbench/workbench.desktop.main.js:3019:11385",
      "originalLocation": "src/vs/workbench/contrib/tasks/browser/taskProblemMonitor.ts:41:44",
[...]

PR #287038, opened 2026-01-11
fix: memory leak in task terminal status

SimonSiefke:
Fixes a memory leak in task terminal status. 

### Details

Basically this is the same as https://github.com/microsoft/vscode/pull/279093 for `taskTerminalStatus`. 

### Before

When running a task 37 times, the number of functions grows by each time in several places, one of which is task terminal status:

shapes at 26-01-11 23 50 06

### After

Still various possible memory leaks are detected. But not in taskTerminalStatus anymore:

task-run


meganrogge (dismissed):
Thank you!

Tyriar (dismissed):

meganrogge (approved):

connor4312 (approved):

PR #287049, opened 2026-01-12
fix(eslint): correct ApiEventNaming rule's broken subject detection

SajanGhimire1:
Previously, the rule had a critical bug where it would:
1. Create unsafe regex from user input (potential regex injection)
2. Search entire file text instead of checking against known API subjects
3. Use wrong logic (parts.length < 3) that gave meaningless results

Now:
- Use configured subject list instead of file search
- Escape regex special characters for safety
- Handle multi-word subjects properly
- Add configuration validation
- Fix AST traversal with safety limits

BREAKING CHANGE: Configuration now requires 'subjects' array


jrieken:
Thanks for the PR but I don't believe it fixes an actual issue anywhere. The lint rules are pragmatic, simple check we use for our processes and something that is good enough if it works and I just successfully tested that it does work

PR #287063, opened 2026-01-12
Portable mode improvements and bug fixes

### LINKED ISSUE #239139: Do not register a protocol handler in portable mode
- Does this issue occur when all extensions are disabled?: Yes

- VS Code Version: 1.96.4
- OS Version: x64 10.0.19045

Steps to Reproduce:

1. Run VS Code in portable mode (set `VSCODE_PORTABLE` env var to a custom path).
2. View `Computer\HKEY_CLASSES_ROOT\vscode`. Observe that the configured command in `shell\open\command` does not preserve portable mode.
3. Alternatively, try authenticating with GitHub (or another service that uses OAuth with the protocol handler). Observe that after you authenticate in your browser, VS Code is launched in non-portable mode (so that auth token will not be stored in the portable config).

Possible solutions:

1) Skip registering the protocol handler when portable mode is active. This means that authentication flows that rely on the protocol handler will fail (but the user can use the fallback oauth method by using a local server).
2) Store arguments and environment variables used to configure the main VS Code instance somewhere and when invoking VS Code again through a protocol handler, restore them, so that portable mode works properly. It would be courteous toward the user to remove the protocol handler when VS Code exits, so that it does not leave random garbage on the machine when running in portable mode.

The handler is installed here during VS Code startup:
https://github.com/microsoft/vscode/blob/bda07cff3ac8d208a443c029c37c022319ace431/src/vs/platform/url/electron-main/electronUrlListener.ts#L52-L55

### LINKED ISSUE #288950: Proposed API for detecting portable installation of VS Code
This API can be used by extensions to detect when VS Code is running in portable mode to adjust their behavior.
An example is GitHub authentication extension which would avoid using protocol handlers for redirects (which are not/should not be installed in portable mode).

```ts

declare module 'vscode' {

	export namespace env {

		/**
		 * Indicates whether the application is running in portable mode.
		 *
		 * Portable mode is enabled when the application is run from a folder that contains
		 * a `data` directory, allowing for self-contained installations.
		 *
		 * Learn more about [Portable Mode](https://code.visualstudio.com/docs/editor/portable).
		 */
		export const isAppPortable: boolean;
	}
}
```

dmitrivMS:
Fixes #239139

Disabled protocol handlers and registry updates on Windows in portable mode.
Added API proposal to detect if VS Code is running in portable mode from extensions.
Skipped protocol redirect in GitHub authentication in portable mode.


joaomoreno (dismissed):
So cool, looks great from a code review perspective, I did not test it.
I also did not review GitHub Auth extension, leaving that to @TylerLeonhardt

TylerLeonhardt (changes requested):
This a pretty complicated code change for GH auth. I expected an extension of:
https://github.com/microsoft/vscode/blob/main/extensions/github-authentication/src/flows.ts#L25

For a supportsPortable and then go through each flow and decide if it supports it (the answer is "yes" except for Uri handler)

src/vscode-dts/vscode.proposed.envIsPortable.d.ts
```diff
@@ -0,0 +1,18 @@
...
+
+	export namespace env {
+
+		/**
+		 * Indicates whether the application is running in portable mode.
+		 *
+		 * Portable mode is enabled when the application is run from a folder that contains
+		 * a `data` directory, allowing for self-contained installations.
+		 */
+		export const isPortable: boolean;
```
dmitrivMS: @copilot rename property (in the public API **only**) to isAppPortable.
Update description to include the link to https://code.visualstudio.com/docs/editor/portable

extensions/github-authentication/src/node/authServer.ts
```diff
@@ -132,7 +132,11 @@ export class LoopbackAuthServer implements ILoopbackServer {
 						throw new Error('Nonce does not match.');
 					}
 					deferred.resolve({ code, state });
-					res.writeHead(302, { location: `/?redirect_uri=${encodeURIComponent(callbackUri)}${appNameQueryParam}` });
+					if (isPortable) {
+						res.writeHead(302, { location: `/?success=true${appNameQueryParam}` });
+					} else {
+						res.writeHead(302, { location: `/?redirect_uri=${encodeURIComponent(callbackUri)}${appNameQueryParam}` });
+					}
```
TylerLeonhardt, line 139: Could we just support when there is no redirect_uri passed in?

TylerLeonhardt (dismissed):
Approving from auth perspective

bpasero (approved):

PR #287076, opened 2026-01-12
Agent sessions - diff container should never shrink

lszomoru:


bpasero (changes requested):
I do not like how this forces the date to be pushed out of the window:

Image

PR #287107, opened 2026-01-12
mcp: read _meta from content item instead of response level

### LINKED ISSUE #287106: MCP Apps: _meta should be read from content item, not response level
### Bug Description

The `_meta.ui` field containing CSP configuration is read from the response level instead of from inside the `contents` array item.

Related discussion: https://github.com/modelcontextprotocol/ext-apps/issues/242

### Current Implementation

https://github.com/microsoft/vscode/blob/main/src/vs/workbench/contrib/mcp/browser/mcpToolCallUI.ts#L220

```typescript
const meta = resourceResult._meta?.ui as McpApps.McpUiResourceMeta | undefined;
```

### Fix

Should read from `content._meta?.ui` instead:

```typescript
const meta = content._meta?.ui as McpApps.McpUiResourceMeta | undefined;
```

### Environment
- VSCode Version: 1.109.0-insider
- Commit: 7c62052af606ba507cbb8ee90b0c22957bb175e7

qchuchu:
## Summary
Fixes #287106

The `_meta.ui` field containing CSP configuration was being read from the response level instead of from inside the `contents` array item.

Related discussion: https://github.com/modelcontextprotocol/ext-apps/issues/242

## Changes
- Changed `resourceResult._meta?.ui` to `content._meta?.ui` in [mcpToolCallUI.ts:220](https://github.com/microsoft/vscode/blob/main/src/vs/workbench/contrib/mcp/browser/mcpToolCallUI.ts#L220)

## Test Plan
- [ ] Create an MCP server returning `_meta` inside `contents[0]` (as per spec)
- [ ] Verify CSP domains are now correctly applied

### DIFF
--- a/src/vs/workbench/contrib/mcp/browser/mcpToolCallUI.ts
+++ b/src/vs/workbench/contrib/mcp/browser/mcpToolCallUI.ts
@@ -217,7 +217,7 @@ export class McpToolCallUI extends Disposable {
 			throw new Error('UI resource has no content');
 		}
 
-		const meta = resourceResult._meta?.ui as McpApps.McpUiResourceMeta | undefined;
+		const meta = content._meta?.ui as McpApps.McpUiResourceMeta | undefined;
 
 		return {
 			...meta,


connor4312 (approved):
thanks!

dmitrivMS (approved):

PR #287139, opened 2026-01-12
Clear promptInput's value onCommandFinished 

### LINKED ISSUE #640: CTRL-C is immediately issued upon run, execution gets cancelled by activation command
Happens on multiple Ubuntu 24.04 workstations but not on any Windows 11 machines.
All machines are configured with the suggested fixes/settings per your emails for the Python Env extension, that seems to be doing well.

However...

Either from the hotkey or from the menus, when you run a Python file a CTRL-C is immediately issued. It is consistently intermittent, meaning, it's not 100% of the time but from observation it is 75-80% of the time. So, 8 out of 10 times you have to run the script again...and again... until it runs as expected. Sometimes you get lucky, and the CTRL-C is thrown to soon to interfere, like right before the venv is activated.

Thanks folks and cheers!

anthonykim1:
Resolves: https://github.com/microsoft/vscode-python-environments/issues/640 
Coming from:  https://github.com/microsoft/vscode-python-environments/pull/1093#discussion_r2682282695

From: https://github.com/microsoft/vscode-python-environments/pull/1093#issue-3800690027 

***
I couldnt seem to repro ^C to show up on Python file run on Mac, but only on Windows.

This is happening because we have some extra polling happening before onCommandStarted event gets fired: https://github.com/microsoft/vscode/blob/86db4eb05dbfd82dc9b4ea022883494257ac3d63/src/vs/platform/terminal/common/capabilities/commandDetectionCapability.ts#L727 Which is what leads to [resetting the value of promptInputModel](https://github.com/microsoft/vscode/blob/86db4eb05dbfd82dc9b4ea022883494257ac3d63/src/vs/platform/terminal/common/capabilities/commandDetection/promptInputModel.ts#L213):

We need the promptInputModel's value to be empty and state to be not execute in order to not cause ^C upon executeCommand run: https://github.com/microsoft/vscode/blob/86db4eb05dbfd82dc9b4ea022883494257ac3d63/src/vs/workbench/contrib/terminal/browser/terminalInstance.ts#L1010

So in that meantime before we clear the value of promptInput to empty string, the state will be execute and value of it will point to activation script. Which will lead to ^C.
***

Flow from env extension:
Send activate command -> wait for `onDidEndShellExecution` -> send file execution.

The ctrl^c came from second arrow.

### DIFF
--- a/src/vs/platform/terminal/common/capabilities/commandDetection/promptInputModel.ts
+++ b/src/vs/platform/terminal/common/capabilities/commandDetection/promptInputModel.ts
@@ -113,6 +113,7 @@ export class PromptInputModel extends Disposable implements IPromptInputModel {
 		onCommandStart: Event<ITerminalCommand>,
 		onCommandStartChanged: Event<void>,
 		onCommandExecuted: Event<ITerminalCommand>,
+		onCommandFinished: Event<ITerminalCommand>,
 		@ILogService private readonly _logService: ILogService
 	) {
 		super();
@@ -127,6 +128,7 @@ export class PromptInputModel extends Disposable implements IPromptInputModel {
 		this._register(onCommandStart(e => this._handleCommandStart(e as { marker: IMarker })));
 		this._register(onCommandStartChanged(() => this._handleCommandStartChanged()));
 		this._register(onCommandExecuted(() => this._handleCommandExecuted()));
+		this._register(onCommandFinished(() => this._handleCommandFinished()));
 
 		this._register(this.onDidStartInput(() => this._logCombinedStringIfTrace('PromptInputModel#onDidStartInput')));
 		this._register(this.onDidChangeInput(() => this._logCombinedStringIfTrace('PromptInputModel#onDidChangeInput')));
@@ -259,6 +261,12 @@ export class PromptInputModel extends Disposable implements IPromptInputModel {
 		this._state = PromptInputState.Execute;
 		this._onDidFinishInput.fire(event);
 		this._onDidChangeInput.fire(event);
+	}
+
+	private _handleCommandFinished() {
+		// Clear the prompt input value when command finishes to prepare for the next command
+		// This prevents runCommand from detecting leftover text and sending ^C unnecessarily
+		this._value = '';
 	}
 
 	@throttle(0)
--- a/src/vs/platform/terminal/common/capabilities/commandDetection/promptInputModel.ts
+++ b/src/vs/platform/terminal/common/capabilities/commandDetection/promptInputModel.ts
@@ -267,6 +267,7 @@ export class PromptInputModel extends Disposable implements IPromptInputModel {
 		// Clear the prompt input value when command finishes to prepare for the next command
 		// This prevents runCommand from detecting leftover text and sending ^C unnecessarily
 		this._value = '';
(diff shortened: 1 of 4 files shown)


src/vs/platform/terminal/common/capabilities/commandDetection/promptInputModel.ts
```diff
@@ -261,6 +263,12 @@ export class PromptInputModel extends Disposable implements IPromptInputModel {
 		this._onDidChangeInput.fire(event);
 	}
 
+	private _handleCommandFinished() {
+		// Clear the prompt input value when command finishes to prepare for the next command
+		// This prevents runCommand from detecting leftover text and sending ^C unnecessarily
+		this._value = '';
```
anthonykim1, line 269: I'm not sure if it will be good idea to fire an empty `_onDidChangeInput` here, because we havent started the "new" input yet.

src/vs/platform/terminal/common/capabilities/commandDetection/promptInputModel.ts
```diff
@@ -261,6 +263,12 @@ export class PromptInputModel extends Disposable implements IPromptInputModel {
 		this._onDidChangeInput.fire(event);
 	}
 
+	private _handleCommandFinished() {
+		// Clear the prompt input value when command finishes to prepare for the next command
```
anthonykim1, line 267: Should there also be `Finish` state that we can benefit from setting here?

There are only unknown, input, and execute: https://github.com/microsoft/vscode/blob/e76e5a9069234704791fc428e7eacbdc040764f0/src/vs/platform/terminal/common/capabilities/commandDetection/promptInputModel.ts#L14-L18

src/vs/platform/terminal/common/capabilities/commandDetection/promptInputModel.ts
```diff
@@ -261,6 +263,12 @@ export class PromptInputModel extends Disposable implements IPromptInputModel {
...
+		// This prevents runCommand from detecting leftover text and sending ^C unnecessarily
```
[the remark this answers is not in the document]
anthonykim1: maybe I should have the `finished` state. 
Otherwise state is execute even when it finished, so this is not really good atm

src/vs/platform/terminal/common/capabilities/commandDetection/promptInputModel.ts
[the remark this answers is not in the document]
anthonykim1: These should probably be updated when there is actual input (in commandStart as before)

src/vs/platform/terminal/common/capabilities/commandDetection/promptInputModel.ts
Tyriar: Let's keep it simpler unless we need to differentiate them

src/vs/platform/terminal/common/capabilities/commandDetection/promptInputModel.ts
Tyriar: Since we're changing the value, let's fire it to be consistent and see if there are any problems.

Tyriar (approved):

PR #287142, opened 2026-01-12
fix: memory leak in tunnel view

SimonSiefke:
Fixes some memory leaks in tunnel view. 

### Details

Table renderers and list renderers are special since they don't use `this._register` for registering disposables but instead `templateDisposables` and `elementDisposables`. 

### Change

Changes `ActionBarRenderer` to not be a disposable anymore since TableRenderers  and ListRenderers are (almost?) always not Disposables.

Instead, the code now uses `elementDisposables` and `templateDisposables` to register the items.

### Before

When opening the ports tunnel view and toggling the input 37 times, the number of functions seems to grow each time in `ActionBarRenderer.renderButton`  and some other places:

ports-view-open-port

### After

No more leak is detected.


alexr00 (approved):
Thanks for the PR!

Tyriar (approved):

Issue #284091, opened 2025-12-17
Opportunities to make new agent session types more discoverable


bamurtaugh:
We're finding that some users aren't discovering the cloud and background options in the `+` menu (as they were accustomed to the explicit `+` actions in the separate Agent Sessions view).
- I believe @bpasero and others have mentioned there should perhaps be an option more at the chat input box level, with the other dropdowns
     - `+` can feel a bit far from other agent interactions (where I'm picking the custom agent, model, or choose to use the delegation arrow all at the bottom)
- As for the actual options in the `+` menu:
     - New chat and new chat editor are the same keyboard shortcut (at least on Windows)
     - <img width="544" height="194" alt="Image" src="https://github.com/user-attachments/assets/5f379b67-8d71-4bc9-ab02-5c53b56f2a29" />

bpasero:
Some mockup I had earlier this month:

![Image](https://github.com/user-attachments/assets/7fdf6196-101a-4d2f-967f-d83fdf13e90b)

Ben will work on this experience for the next iteration.

OrenMe:
> Some mockup I had earlier this month:
> 
> ![Image](https://github.com/user-attachments/assets/7fdf6196-101a-4d2f-967f-d83fdf13e90b)
> 
> Ben will work on this experience for the next iteration.

This looks really nice 👍

PR #287143, opened 2026-01-12
Add Session target picker actions

benibenj:
```Copilot Generated Description:``` Introduce a new chat input picker and session target picker actions, along with necessary UI updates and styling changes. Update relevant imports and class names to reflect the new picker functionality.

https://github.com/microsoft/vscode/issues/284091


bpasero (dismissed):
@benibenj the one feedback I would have is that if the picker is disabled, maybe it should just shrink to only showing the icon and on hover you get some more info why its disabled? It could also shout out to the keybinding for how to create a new Chat of the same type.

lszomoru (approved):

PR #286860, opened 2026-01-10
Unify agent skills internal architecture

pwang347:
## Overview
- Add new `PromptsType.skill`
- Add new configuration `agentSkillsLocations` similar to other file location settings (add support for multiple defaults and untildify)

## Followup
https://github.com/microsoft/vscode/pull/287201
- Provider and UI unification

PR #287201, opened 2026-01-13
Agent skills management UX and file support

### LINKED ISSUE #280622: Claude Skills: There is no UI action to create a skill
Testing #280352

- We do not have a command to create a skill
- There is no entry for skills in the `Configure Chat` action (could only be shown when claude model is selected?)

Image

It would be very handy if there was some action provided by VS Code to create a skill which will ensure the file is created at the correct file path and with the correct template.

pwang347:
## Overview
Follow-up for https://github.com/microsoft/vscode/pull/286860
- Adds skills markdown file language and linting support
- Adds UI for managing and creating skills

Fixes https://github.com/microsoft/vscode/issues/280622


extensions/prompt-basics/package.json
```diff
@@ -50,6 +50,19 @@
...
         "configuration": "./language-configuration.json"
+      },
+      {
+        "id": "skill",
+        "aliases": [
+          "Skill",
+          "skill"
+        ],
+        "filenamePatterns": [
+          "**/.github/skills/*/SKILL.md",
```
pwang347: this matches the pattern above, but should this be `**/SKILL.md`?

aeschli: Probably just `filenames`: `SKILL.md`

aeschli (approved):

DonJayamanne (approved):

PR #287238, opened 2026-01-13
Git - use `findFiles2()` to expand glob patterns

lszomoru:

### DIFF
--- a/extensions/git/src/repository.ts
+++ b/extensions/git/src/repository.ts
@@ -8,7 +8,7 @@ import * as fs from 'fs';
 import * as fsPromises from 'fs/promises';
 import * as path from 'path';
 import picomatch from 'picomatch';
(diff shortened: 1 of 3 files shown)


extensions/git/src/repository.ts
```diff
@@ -1896,62 +1896,28 @@ export class Repository implements Disposable {
...
+		const allFiles = await workspace.findFiles2(filePattern, {
+			useExcludeSettings: ExcludeSettingOptions.None,
+			useIgnoreFiles: { local: false, parent: false, global: false }
+		});
 
-			// Check the files that are not under a git ignored directories
-			const filesToCheckResults = await this.checkIgnore(Array.from(filesToCheck));
-			filesToCheckResults.forEach(ignoredFile => gitIgnoredFiles.add(ignoredFile));
+		// Get files matching the globs with git ignore files applied
+		const nonIgnoredFiles = await workspace.findFiles2(filePattern, {
```
dmitrivMS, line 1909: I suspect this would be a good place to apply the new ignoreGlobCase option, but perhaps internally at the API entry point vs. passing as an argument...

dmitrivMS (approved):

extensions/git/src/repository.ts
lszomoru: I think that we would have to better understand the use cases to make that choice.

PR #287586, opened 2026-01-13
set terminal chat output width based on content

### LINKED ISSUE #275074: Terminal tool: horizontal scrollbar visible even when not needed
Testing #274909

Image

I see a horizontal scrollbar even when the output does not require it.

meganrogge:
fixes #275074

https://github.com/user-attachments/assets/30165578-9f37-4bb4-82ea-b10ab6cc9239

### DIFF
--- a/src/vs/workbench/contrib/terminal/browser/chatTerminalCommandMirror.ts
+++ b/src/vs/workbench/contrib/terminal/browser/chatTerminalCommandMirror.ts
@@ -38,10 +38,45 @@ function getChatTerminalBackgroundColor(theme: IColorTheme, contextKeyService: I
 	return theme.getColor(isInEditor ? editorBackground : PANEL_BACKGROUND);
 }
 
+/**
+ * Computes the maximum column width of content in a terminal buffer.
+ * Iterates through each line and finds the rightmost non-empty cell.
+ *
+ * @param buffer The buffer to measure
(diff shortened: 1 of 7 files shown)


src/vs/workbench/contrib/terminal/browser/chatTerminalCommandMirror.ts
```diff
@@ -258,8 +294,9 @@ export class DetachedTerminalCommandMirror extends Disposable implements IDetach
 		}
 
 		this._lineCount = this._getRenderedLineCount();
+		this._maxColumnWidth = this._computeMaxColumnWidth();
```
Tyriar: Does this happen after every update? That'll be too expensive. I think we should only do this for commands that are under some reasonable amount (100 lines?) and that have already finished. In other words, compact the xterm.js instance only after the command has finished. That'll slightly improve UX as it would remove a potentially unnecessary scroll bar and also reduce some memory usage.

Tyriar (dismissed):

src/vs/workbench/contrib/terminal/browser/chatTerminalCommandMirror.ts
meganrogge: Sounds good, will do

aeschli (approved):

PR #287596, opened 2026-01-13
`outputLocation:none` -> `outputLocation:chat`

### LINKED ISSUE #275584: Revisit outputLocation setting values
When https://github.com/microsoft/vscode/issues/271390 is done, we should look at the setting to see what values make the most sense.

- `none` or `hidden` in particular may not be valid?
- Is auto expanding here or a different setting?

meganrogge:
fixes #275584


dmitrivMS (approved):

src/vs/workbench/contrib/terminalContrib/chatAgentTools/common/terminalChatAgentToolsConfiguration.ts
```diff
@@ -480,14 +480,14 @@ export const terminalChatAgentToolsConfiguration: IStringDictionary<IConfigurati
...
-		enum: ['terminal', 'none'],
+		enum: ['terminal', 'chat'],
 		enumDescriptions: [
-			localize('outputLocation.terminal', "Reveal the terminal when running the command."),
-			localize('outputLocation.none', "Do not reveal the terminal automatically."),
+			localize('outputLocation.terminal', "Reveal the terminal in the panel or editor in addition to chat."),
+			localize('outputLocation.chat', "Reveal the terminal output within chat only."),
 		],
-		default: product.quality !== 'stable' ? 'none' : 'terminal',
+		default: product.quality !== 'stable' ? 'chat' : 'terminal',
```
Tyriar, line 490: Can keep this simple as terminal btw since we're rolling out as experiment

meganrogge: Don't we still want the default to be `terminal` for stable so our experiment sets it to `chat`?

PR #287638, opened 2026-01-13
Integrated Browser - New Tab welcome screen

jruales:
Add a "null state" or New Tab welcome screen that shows up when the Integrated Browser opens and before the user types a URL.

Also ensures that this new welcome screen isn't applied a blur CSS filter when opening the command palette, by separating the blurrable placeholder screenshot into its own separate element.

image

Related issue: https://github.com/microsoft/vscode/issues/286579

### DIFF
--- a/src/vs/workbench/contrib/browserView/electron-browser/browserEditor.ts
+++ b/src/vs/workbench/contrib/browserView/electron-browser/browserEditor.ts
@@ -390,7 +390,7 @@ export class BrowserEditor extends EditorPane {
 	}
 
 	private get shouldShowView(): boolean {
-		return this._editorVisible && !this._overlayVisible && !this._model?.error;
+		return this._editorVisible && !this._overlayVisible && !this._model?.error && !!this._model?.url;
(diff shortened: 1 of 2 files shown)


src/vs/workbench/contrib/browserView/electron-browser/browserEditor.ts
```diff
@@ -564,14 +576,52 @@ export class BrowserEditor extends EditorPane {
 		// Update context keys for command enablement
 		this._canGoBackContext.set(event.canGoBack);
 		this._canGoForwardContext.set(event.canGoForward);
+
+		// Update welcome screen visible or not
+		this.updateWelcomeVisibility();
```
kycutler: Let's probably move this into `updateVisibility` and call that from here (since we'll need to do it to show the view too when a url is entered). We should also consider the error container show/hide. It might make sense to just centralize all that logic into the one `updateVisibility` method.

kycutler (approved):

PR #287640, opened 2026-01-13
make chat thinking stream screen reader accessible

### LINKED ISSUE #286777: Screen reader users have no access to thinking stream
I'm thinking, no pun intended, we might want a command that opens thinking in the accessible view. However, with how quickly the streaming occurs, it could just be annoying and we might not want it to update dynamically.

cc @jooyoungseo and @accesswatch and @rperez030

meganrogge:
fixes #286777 

Adds a `ChatThinkingAccessibleView`, which can be accessed via a command. If no thinking is present, we alert to indicate that. Otherwise, we open the accessible view with the current thinking content. 

Uses `shift+alt+f2` to align with and extend the existing `alt+f2` one. 

- to discuss with screen reader users: should we clear the thinking content once thinking has stopped or leave it around until the next thinking stream comes in?


justschen (approved):
thanks!!

PR #287651, opened 2026-01-13
fix races in prompt for input 

### LINKED ISSUE #287642: prompt for input showing after input has been typed in the terminal
In https://github.com/microsoft/vscode-copilot-evaluation/ ask agent to run `python benchmarks/build_and_push.py -i read_lsp_spec --push`, type confirm and enter to run in the embedded terminal

🐛 prompt for input shows up

Image

meganrogge:
fixes #287642

When a user manually entered input into the terminal before the input prompt was created, the prompt would still appear and hang around even after the command finished.

This happened bc the input listener was being set up too late - only after the LLM finished analyzing the terminal output. Any user input that happened between idle detection and prompt creation was missed.

Fixes:

1) The input listener is now set up immediately when idle is detected inside waitForIdle, not after returning from it. This eliminates the race window between detecting idle and handling the idle state.

2) Added checks for `_userInputtedSinceIdleDetected`at multiple points:

- Before calling the LLM to determine user input options
- After the LLM call returns
- After `_selectAndHandleOption()` (another LLM call)
- Right before showing any prompt UI

If the user has inputted at any of these checkpoints, we skip showing the prompt and continue polling instead.

3) Updated the LLM prompt to explicitly require that detected input prompts must appear at the VERY END of the terminal output with no content following them. Added examples showing prompts that were already answered should return null:

```
"Continue (y/n)? y" → null (already answered with 'y')
"Do you want to proceed? (yes/no)\nyes\nProceeding..." → null (answered with subsequent output)
```

### DIFF
--- a/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/monitoring/outputMonitor.ts
+++ b/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/monitoring/outputMonitor.ts
@@ -61,6 +61,13 @@ export class OutputMonitor extends Disposable implements IOutputMonitor {
 	private _pollingResult: IPollingResult & { pollDurationMs: number } | undefined;
 	get pollingResult(): IPollingResult & { pollDurationMs: number } | undefined { return this._pollingResult; }
 
+	/**
+	 * Flag to track if user has inputted since idle was detected.
+	 * This is used to skip showing prompts if the user already provided input.
+	 */
+	private _userInputtedSinceIdleDetected = false;
+	private _userInputListener: IDisposable | undefined;
+
 	private readonly _outputMonitorTelemetryCounters: IOutputMonitorTelemetryCounters = {
 		inputToolManualAcceptCount: 0,
 		inputToolManualRejectCount: 0,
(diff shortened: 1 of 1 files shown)


src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/monitoring/outputMonitor.ts
```diff
@@ -345,6 +388,32 @@ export class OutputMonitor extends Disposable implements IOutputMonitor {
 		return OutputMonitorState.Timeout;
 	}
 
+	/**
+	 * Sets up a listener for user input that triggers immediately when idle is detected.
+	 * This ensures we catch any input that happens between idle detection and prompt creation.
+	 */
+	private _setupIdleInputListener(): void {
+		// Clean up any existing listener
+		this._userInputListener?.dispose();
```
roblourens, line 397: MutableDisposable is the right pattern to avoid this duplication

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/monitoring/outputMonitor.ts
```diff
@@ -345,6 +388,32 @@ export class OutputMonitor extends Disposable implements IOutputMonitor {
...
+		this._userInputtedSinceIdleDetected = false;
+
+		// Set up new listener
+		this._userInputListener = this._execution.instance.onDidInputData((data) => {
+			if (data === '\r' || data === '\n' || data === '\r\n') {
+				this._userInputtedSinceIdleDetected = true;
```
roblourens, line 403: The user could input without a newline, `read` doesn't need to read full lines

DonJayamanne (approved):

PR #287653, opened 2026-01-13
Support additional actions in integrated browser

kycutler:
#286579

Also adds a setting to redirect from simple browser.

### DIFF
--- a/extensions/simple-browser/package.json
+++ b/extensions/simple-browser/package.json
@@ -51,6 +51,15 @@
             "default": true,
             "title": "Focus Lock Indicator Enabled",
             "description": "%configuration.focusLockIndicator.enabled.description%"
+          },
+          "simpleBrowser.useIntegratedBrowser": {
+            "type": "boolean",
+            "default": false,
+            "markdownDescription": "%configuration.useIntegratedBrowser.description%",
+            "scope": "application",
+            "tags": [
+              "experimental"
+            ]
           }
         }
       }
--- a/extensions/simple-browser/src/extension.ts
+++ b/extensions/simple-browser/src/extension.ts
@@ -14,6 +14,8 @@ declare class URL {
 
 const openApiCommand = 'simpleBrowser.api.open';
 const showCommand = 'simpleBrowser.show';
+const integratedBrowserCommand = 'workbench.action.browser.open';
+const useIntegratedBrowserSetting = 'simpleBrowser.useIntegratedBrowser';
 
 const enabledHosts = new Set<string>([
 	'localhost',
@@ -30,6 +32,27 @@ const enabledHosts = new Set<string>([
 ]);
 
 const openerId = 'simpleBrowser.open';
+
+/**
+ * Checks if the integrated browser should be used instead of the simple browser
+ */
+async function shouldUseIntegratedBrowser(): Promise<boolean> {
+	const config = vscode.workspace.getConfiguration();
+	if (!config.get<boolean>(useIntegratedBrowserSetting, false)) {
+		return false;
+	}
+
+	// Verify that the integrated browser command is available
+	const commands = await vscode.commands.getCommands(true);
+	return commands.includes(integratedBrowserCommand);
+}
+
+/**
+ * Opens a URL in the integrated browser
+ */
+async function openInIntegratedBrowser(url: string): Promise<void> {
+	await vscode.commands.executeCommand(integratedBrowserCommand, url);
+}
 
 export function activate(context: vscode.ExtensionContext) {
 
@@ -51,15 +74,23 @@ export function activate(context: vscode.ExtensionContext) {
 		}
 
 		if (url) {
-			manager.show(url);
+			if (await shouldUseIntegratedBrowser()) {
+				await openInIntegratedBrowser(url);
+			} else {
+				manager.show(url);
+			}
 		}
 	}));
 
-	context.subscriptions.push(vscode.commands.registerCommand(openApiCommand, (url: vscode.Uri, showOptions?: {
+	context.subscriptions.push(vscode.commands.registerCommand(openApiCommand, async (url: vscode.Uri, showOptions?: {
 		preserveFocus?: boolean;
 		viewColumn: vscode.ViewColumn;
 	}) => {
-		manager.show(url, showOptions);
+		if (await shouldUseIntegratedBrowser()) {
+			await openInIntegratedBrowser(url.toString(true));
+		} else {
+			manager.show(url, showOptions);
+		}
 	}));
 
 	context.subscriptions.push(vscode.window.registerExternalUriOpener(openerId, {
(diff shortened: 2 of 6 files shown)


extensions/simple-browser/src/extension.ts
```diff
@@ -31,6 +33,27 @@ const enabledHosts = new Set<string>([
...
+
+	// Verify that the integrated browser command is available
+	const commands = await vscode.commands.getCommands(true);
+	return commands.includes(integratedBrowserCommand);
+}
+
+/**
+ * Opens a URL in the integrated browser
+ */
+async function openInIntegratedBrowser(url: string): Promise<void> {
```
jruales: I wonder if maybe when Integrated Browser is enabled, we shouldn't ask the user for the URL, but instead simply open Integrated Browser straight away and let the user input the URL normally in the address bar? Or you're just doing it like this to keep the experience consistent?

extensions/simple-browser/package.json
```diff
@@ -51,6 +51,15 @@
             "default": true,
             "title": "Focus Lock Indicator Enabled",
             "description": "%configuration.focusLockIndicator.enabled.description%"
+          },
+          "simpleBrowser.useIntegratedBrowser": {
+            "type": "boolean",
+            "default": false,
+            "markdownDescription": "%configuration.useIntegratedBrowser.description%",
+            "scope": "application",
+            "tags": [
```
jruales, line 60: I think that the concept of Advanced Settings was added recently. We could also look into that

src/vs/workbench/contrib/browserView/electron-browser/browserViewActions.ts
```diff
@@ -193,6 +204,35 @@ class ToggleDevToolsAction extends Action2 {
 	}
 }
 
+class OpenInExternalBrowserAction extends Action2 {
+	static readonly ID = 'workbench.action.browser.openExternal';
+
+	constructor() {
+		super({
+			id: OpenInExternalBrowserAction.ID,
+			title: localize2('browser.openExternalAction', 'Open in External Browser'),
```
jruales, line 217: Maybe we should be a bit more explicit? Like so:
```suggestion
			title: localize2('browser.openExternalAction', 'Open Page in External Browser'),
```

src/vs/workbench/contrib/browserView/electron-browser/browserViewActions.ts
```diff
@@ -243,12 +283,38 @@ class ClearWorkspaceBrowserStorageAction extends Action2 {
 	}
 }
 
+class OpenBrowserSettingsAction extends Action2 {
+	static readonly ID = 'workbench.action.browser.openSettings';
+
+	constructor() {
+		super({
+			id: OpenBrowserSettingsAction.ID,
+			title: localize2('browser.openSettingsAction', 'Configure Browser...'),
```
jruales: Maybe more standard to call it something with the word "Settings"?
```suggestion
			title: localize2('browser.openSettingsAction', 'Browser Settings...'),
```

src/vs/workbench/contrib/browserView/electron-browser/browserViewActions.ts
```diff
@@ -155,6 +157,11 @@ class AddElementToChatAction extends Action2 {
 				group: 'actions',
 				order: 1,
 				when: ChatContextKeys.enabled
+			},
+			keybinding: {
+				when: BROWSER_EDITOR_ACTIVE,
+				weight: KeybindingWeight.WorkbenchContrib + 50, // Priority over terminal
+				primary: KeyMod.CtrlCmd | KeyMod.Shift | KeyCode.KeyC,
```
jruales, line 164: I see that this keybinding matches the shortcut in devtools, which is convenient, but I hope it won't cause frustration if someone wants to use it for the devtools use and doesn't have the focus on the devtools.

Also, we might want to match the behavior in DevTools where Esc cancels out of the "select element" state

src/vs/platform/browserView/electron-main/browserView.ts
```diff
@@ -467,6 +445,54 @@ export class BrowserView extends Disposable {
...
+	 */
+	private tryHandleCommand(input: Electron.Input): boolean {
+		const eventKeyCode = SCAN_CODE_STR_TO_EVENT_KEY_CODE[input.code] || 0;
+		const keyCode = EVENT_KEY_CODE_MAP[eventKeyCode] || KeyCode.Unknown;
+
+		const isArrowKey = keyCode >= KeyCode.LeftArrow && keyCode <= KeyCode.DownArrow;
+		const isNonEditingKey =
+			keyCode === KeyCode.Escape ||
+			keyCode >= KeyCode.F1 && keyCode <= KeyCode.F24 ||
+			keyCode >= KeyCode.AudioVolumeMute;
```
jruales, line 460: This seems like a large unbounded range of keys. Is evething above audiovolumemute expected to be a non-editing key?

jruales (dismissed):

src/vs/platform/browserView/electron-main/browserView.ts
kycutler: They are enumerated here: https://github.com/microsoft/vscode/blob/f85cf0ad0cc0a346375e40db969944eef33e1fa7/src/vs/base/common/keyCodes.ts#L199-L226

src/vs/workbench/contrib/browserView/electron-browser/browserViewActions.ts
kycutler: The original shortcut won't work unless devtools is open and selected anyway, unless we try to support it specifically. User can always change the shortcut too. I'm inclined to wait for feedback

src/vs/workbench/contrib/browserView/electron-browser/browserViewActions.ts
kycutler: Fair, I had this originally but had changed it to align with the "Configure Editors" action in the editor title bar menu. We don't seem to be very consistent about it across our commands.

Changed it to "Open Browser Settings" since that seems most clean and clear to me.

src/vs/workbench/contrib/browserView/electron-browser/browserViewActions.ts
kycutler: I'd prefer to keep it concise; let's wait for feedback

extensions/simple-browser/package.json
kycutler: "experimental" allows it to be controlled via ExP. I'm thinking we should leave it visible (i.e. not advanced) especially in case users want to manually turn it off

extensions/simple-browser/src/extension.ts
kycutler: Oh good point, updated!

src/vs/workbench/contrib/browserView/electron-browser/browserViewActions.ts
jruales: Sounds good

jruales (approved):

PR #287666, opened 2026-01-14
chat: wip on model-specific tools

connor4312:
This is a rethinking of the initial API proposed in `languageModelToolSupportsModel.d.ts`.

1. This switches to `models?: LanguageModelChatSelector[];` to control enablement. definitely open to switching this out, but I think a synchronously-analyzable
   expression is important to retain the data flows in core without races.
2. The extension is able to define a tool at runtime via registerToolDefinition. This should let us have entirely service-driven tools from model providers without requiring a static definition for each one. We can also have model-specific variants of tools without a ton of package.json work for each variant of the tool (as initially proposed using `when` clauses)

This then propagates that down into the tools service. Currently I have this as just compiling to a `when` expression once it reaches the main thread. Then, for the tools service, it takes an IContextKeyService in cases where tools should be enumerated, and the chat input sets the model keys in its scoped context key service. This allows the tools to be filtered correctly in the tool picker.

I initially thought about allowing multiple definitions be registered for the same tool name/id for model-specific variants of tools but I realized that gets really gnarly and we already have a `toolReferenceName` that multiple tools can register into.

Todo for tomorrow morning:
- Tools don't make it to the ChatRequest yet, or something, still need to investigate
- Need to make sure tools in prompts/models all work. For a first pass I think we can
  let prompts/modes reference all tools by toolReferenceName.
- Validate that multiple tools actually can safely share a reference name (and do
  some priority ordering?)
- General further validation
- Some unit tests

### DIFF
--- a/src/vscode-dts/vscode.proposed.languageModelToolSupportsModel.d.ts
+++ b/src/vscode-dts/vscode.proposed.languageModelToolSupportsModel.d.ts
@@ -5,17 +5,58 @@
 
 declare module 'vscode' {
 
-	export interface LanguageModelTool<T> {
+	export interface LanguageModelToolDefinition extends LanguageModelToolInformation {
 		/**
-		 * Called to check if this tool supports a specific language model. If this method is not implemented,
-		 * the tool is assumed to support all models.
-		 *
-		 * This method allows extensions to dynamically determine which models a tool can work with,
-		 * enabling fine-grained control over tool availability based on model capabilities.
-		 *
-		 * @param modelId The identifier of the language model (e.g., 'gpt-4o', 'claude-3-5-sonnet')
-		 * @returns `true` if the tool supports the given model, `false` otherwise
+		 * Display name for the tool.
 		 */
-		supportsModel?(modelId: string): Thenable<boolean>;
+		displayName: string;
+		/**
+		 * Name of the tools that can users can reference in the prompt. If not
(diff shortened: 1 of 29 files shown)


src/vscode-dts/vscode.proposed.languageModelToolSupportsModel.d.ts
```diff
@@ -5,17 +5,58 @@
...
+		userDescription?: string;
+		/**
+		 * Icon for the tool shown to the user.
+		 */
+		icon?: IconPath;
+		/**
+		 * If defined, the tool will only be available for language models that match
+		 * the selector.
+		 */
+		models?: LanguageModelChatSelector[];
```
bhavyaus, line 30: Should the model specific tool description also be tied to the LanguageModelChatSelector? It feels weird to me that the supported models could be defined seperately from the alternativeDefintions for the supported models
We could just add support here and get rid of the the one in the extension?

bhavyaus (approved):

src/vscode-dts/vscode.proposed.languageModelToolSupportsModel.d.ts
connor4312: `LanguageModelToolInformation` (which this extends) has a `description` which I believe correlates to the modelSpecificDescription in the package.json contribution.

And yea my goal is that we can use this instead of alternativeDefinitions :)

PR #287687, opened 2026-01-14
Implement UX for running subagents in parallel

### LINKED ISSUE #286606: Chat subagent UX
Today subagents are run synchronously (the orchestrator agent is blocked until they are complete) and not in parallel (if it starts multiple subagents via parallel tool calls, they are run one at a time).

The goal for this month is to run sync subagents in parallel and to have decent UX for this. We can brainstorm about subagents that are spawned in the background and do not block the main agent, but I think it's a much more complex scenario.

Run subagents in parallel: https://github.com/microsoft/vscode/issues/274630

For a subagent, I would like to see
- The name of the custom agent being used (if applicable)
- The task description for the subagent (shown today, with a spinner)
- Something about what the subagent is currently working on
- The final text that the subagent returned (not by default, but on a click or something)

And I think I'd like to see this for all subagents currently running in parallel

Some things to consider
- How are tool confirmations shown for the individual subagent?
- Do we need to ensure that the user will always see confirmations in the viewport, even when there are several subagents and a short viewport?
- How much of the subagent's history do I see? Just the latest tool or more?
- Do we call them subagents, agents (overloaded), something else?
- Special handling for edits?

The simplest possible thing that comes to mind is something like the thinking blocks we have today.
- The title shows the custom agent name (if set, or just 'subagent') and the task description
- In the box we only show the current tool. One can click to expand.
- We don't show the thinking text, at least for now.
- When you expand a finished subagent, you also see the final text returned to the orchestrator.
- Confirmations show inside the box

Image

Image

Some other things to think about, tying this to other explorations from Joanna
- Should we pull confirmations out of the individual subagent box and show them next to the input? For all tool confirmations, not just for subagents?
  - In this case, need a way to know which subagent the confirmation belongs to, maybe click a button to reveal
- Should there be a more structured output for subagents, eg showing the individual edits that came from the subagent?
- Can we associate TODOs and subagents, to show the chat response trajectory in a more structured way?
  - I'm not sure we can really do this today, I think that an agent trajectory is sometimes less linear than its TODOs suggest
  - But let's keep it in mind if we get TODOs rendering like this
- What is the fundamental difference between subagent orchestration and multi-agent orchestration? Is there a world where we render them more similarly?
- If we develop better UI for background terminals, think about whether background subagents should be shown in a similar way

Also see https://github.com/microsoft/vscode-internalbacklog/issues/6539

roblourens:
Fix #286606


lszomoru (dismissed):

amunger (dismissed):

justschen (approved):
very cool, thanks!

PR #287865, opened 2026-01-14
prevent flicker in expanding chat output

### LINKED ISSUE #287664: Auto expand terminal commands that are long running
After a certain duration we should auto expand. We only want to not expand when it could flicker briefly.

Image

This will probably need special handling when @justschen's stuff lands, we want to suppress the output if the command finished right after.

Maybe it should work like this:

1. Start command
2. Kick off 500ms timeout
  - If this is hit without any data events, expand
3. Wait for first data event
  - When hit, wait 50ms and expand if the command is not yet finished

Test cases:

- `ls` should not expand as it's so fast (unless in a really large folder)
- pwsh/conpty will show different data patterns so it should work here too

meganrogge:
fixes https://github.com/microsoft/vscode/issues/287664

This fixes a flickering issue where fast commands like "ls" would cause the terminal output panel to rapidly open and close.

The fix uses a two-timeout system. A 500ms timeout starts when a command executes, expanding output only if no data events arrive and real output exists. When data events do arrive, that timeout is cancelled and a 50ms timeout starts instead, which only expands if the command is still running.

Fast commands that finish within 50ms of producing output won't trigger auto-expansion, preventing the flicker. The code also distinguishes real output from shell integration sequences by checking if the cursor has moved past the executed marker.

The hasOutput method was inlined as hasRealOutput, and timeout cleanup now uses a DisposableStore. A test file with ~340 lines covers various scenarios including fast commands, long-running commands, and timeout cancellation.

https://github.com/user-attachments/assets/1237a187-692b-460c-b317-b778dabb21d2


src/vs/workbench/contrib/chat/test/browser/widget/chatContentParts/chatTerminalToolProgressPart.test.ts
```diff
@@ -0,0 +1,281 @@
...
+		await timeout(200);
+		onWillData.fire('output');
+
+		// Command finishes at 300ms (before 50ms data timeout would fire at 250ms)
+		await timeout(30);
+		onCommandFinished.fire();
+
+		// Wait past 500ms mark - should NOT have expanded because data arrived
+		// and then command finished before the 50ms data timeout
+		await timeout(400);
```
Tyriar: We shouldn't be using timeouts in tests, unit tests must run very quickly. There's the `runWithFakedTimers` which can help here, otherwise you could make the timeouts protected and then override them in a subclass inside the test. That will allow you to do timeouts like 1 ms.

src/vs/workbench/contrib/chat/browser/widget/chatContentParts/toolInvocationParts/terminalToolAutoExpand.ts
```diff
@@ -0,0 +1,132 @@
...
+
+		// Ensure timeouts are cleaned up when disposed
+		store.add(toDisposable(() => this._clearAutoExpandTimeouts()));
+
+		// Logic from _registerInstanceListener
+		store.add(this._options.onCommandExecuted(() => {
+			// Auto-expand for long-running commands:
+			// 1. Kick off 500ms timeout - if hit without any data events, expand only if there's real output
+			if (this._options.shouldAutoExpand() && !this._noDataTimeout) {
+				this._noDataTimeout = setTimeout(() => {
```
Tyriar: `disposableTimeout` might be better?

src/vs/workbench/contrib/chat/browser/widget/chatContentParts/toolInvocationParts/terminalToolAutoExpand.ts
```diff
@@ -0,0 +1,132 @@
...
+/**
+ * Timeout in milliseconds to wait when no data events are received before checking for auto-expand.
+ */
+export const NO_DATA_TIMEOUT_MS = 500;
+
+/**
+ * Timeout in milliseconds to wait after first data event before checking for auto-expand.
+ * This prevents flickering for fast commands like `ls` that finish quickly.
+ */
+export const DATA_EVENT_TIMEOUT_MS = 50;
```
Tyriar: const enum (or protected if we override in tests)

src/vs/workbench/contrib/chat/test/browser/widget/chatContentParts/chatTerminalToolProgressPart.test.ts
```diff
@@ -0,0 +1,281 @@
...
+		return !isExpanded && !userToggledOutput;
+	}
+
+	function hasRealOutput(): boolean {
+		return hasRealOutputValue;
+	}
+
+	function setupAutoExpandLogic(localStore: DisposableStore): void {
+		// Use the real TerminalToolAutoExpand class
+		localStore.add(new TerminalToolAutoExpand({
```
Tyriar: Should just be able to use `store` here I think?

src/vs/workbench/contrib/chat/test/browser/widget/chatContentParts/chatTerminalToolProgressPart.test.ts
```diff
@@ -0,0 +1,281 @@
...
+
+/**
+ * These tests verify the auto-expand logic for terminal tool progress parts.
+ *
+ * The algorithm is:
+ * 1. When command executes, kick off 500ms timeout - if hit without data events, expand only if there's real output
+ * 2. On first data event, wait 50ms and expand if command not yet finished
+ * 3. Fast commands (finishing quickly) should NOT auto-expand to prevent flickering
+ */
+suite('ChatTerminalToolProgressPart Auto-Expand Logic', () => {
```
Tyriar, line 14: I think this is only testing TerminalToolAutoExpand? Therefore should live in `terminalToolAutoExpand.test.ts`?

src/vs/workbench/contrib/chat/test/browser/widget/chatContentParts/chatTerminalToolProgressPart.test.ts
```diff
@@ -0,0 +1,281 @@
...
+import { TerminalToolAutoExpand, NO_DATA_TIMEOUT_MS, DATA_EVENT_TIMEOUT_MS } from '../../../../browser/widget/chatContentParts/toolInvocationParts/terminalToolAutoExpand.js';
+
+/**
+ * These tests verify the auto-expand logic for terminal tool progress parts.
+ *
+ * The algorithm is:
+ * 1. When command executes, kick off 500ms timeout - if hit without data events, expand only if there's real output
+ * 2. On first data event, wait 50ms and expand if command not yet finished
+ * 3. Fast commands (finishing quickly) should NOT auto-expand to prevent flickering
+ */
```
Tyriar: Don't need to reiterate impl details in test, this will be subject to bitrot

src/vs/workbench/contrib/chat/browser/widget/chatContentParts/toolInvocationParts/terminalToolAutoExpand.ts
```diff
@@ -0,0 +1,132 @@
...
+		}));
+
+		// 2. Wait for first data event - when hit, wait 50ms and expand if command not yet finished
+		// Also checks for real output since shell integration sequences trigger onWillData
+		store.add(this._options.onWillData(() => {
+			if (this._receivedData) {
+				return;
+			}
+			this._receivedData = true;
+			// Cancel the 500ms no-data timeout since we received data
```
Tyriar: Avoid mentioning timeout ms in comments as the consts are likely to be tweaked in the future.

src/vs/workbench/contrib/chat/browser/widget/chatContentParts/toolInvocationParts/terminalToolAutoExpand.ts
```diff
@@ -0,0 +1,132 @@
...
+
+	/**
+	 * Check if there is real output (not just shell integration sequences).
+	 */
+	hasRealOutput(): boolean;
+
+	/**
+	 * Toggle the output expanded state.
+	 */
+	toggleOutput(expanded: boolean): void;
```
Tyriar: An event would be better here and then hook that up in the progress part

src/vs/workbench/contrib/chat/browser/widget/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
```diff
@@ -509,30 +509,47 @@ export class ChatTerminalToolProgressPart extends BaseChatToolInvocationSubPart
 			}
 
 			const store = new DisposableStore();
+
+			const onCommandExecuted = new Emitter<void>();
+			const onCommandFinished = new Emitter<void>();
+			store.add(onCommandExecuted);
+			store.add(onCommandFinished);
```
Tyriar: Can you pass in `commandDetection.onCommandExecuted`, etc. instead of creating new events?

src/vs/workbench/contrib/chat/browser/widget/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
```diff
@@ -509,30 +509,47 @@ export class ChatTerminalToolProgressPart extends BaseChatToolInvocationSubPart
...
+			};
+
+			// Use the extracted auto-expand logic
+			store.add(new TerminalToolAutoExpand({
+				onCommandExecuted: onCommandExecuted.event,
+				onWillData: terminalInstance.onWillData,
+				onCommandFinished: onCommandFinished.event,
+				shouldAutoExpand: () => !this._outputView.isExpanded && !this._userToggledOutput && !this._store.isDisposed,
+				hasRealOutput,
+				toggleOutput: (expanded) => this._toggleOutput(expanded),
```
Tyriar: Would passing in command detection capability let us simplify the options here?

Tyriar (dismissed):

osortega (approved):

Issue #273444, opened 2025-10-26
Explore minimal presentation of tools


Tyriar:
I want to investigate how chat feels of tool ui is really minimal. All tools could then expand on click to expose richer information. The task tool progress is a good example of what I have in mind where it changes the inline text multiple times as it's progress changes.

FYI @bpasero, @connor4312, @meganrogge, @justschen

justschen:
this is looking a lot better now - cc @hawkticehurst @jo-oikawa is there anything else we want to look at here?

PR #287902, opened 2026-01-14
terminal in thinking dropdown

justschen:
ref https://github.com/microsoft/vscode/issues/273444


meganrogge (approved):
Testing still seems off to me, but if it looks good to you and the terminal can be focused, scrolled, and inputted into, then I'm okay with this getting merged.

PR #287907, opened 2026-01-14
Integrated browser focus fixes

kycutler:
The integrated browser uses Electron's `WebContentsView` which exists outside the DOM. This breaks VS Code's cross-window focus handling because `document.hasFocus()` returns `false` when the browser view has focus, and calling `window.focus()` doesn't take focus away from the browser view.

The result is that programmatically focusing an element in the workbench does not actually give that element focus when focus is in the browser. This significantly impacts usability and accessibility.

Solutions are further limited because when the browser view is focused, the workbench doesn't receive any `focus` or `blur` events, so we must hook into programmatic `focus()` calls to tell when focus should move.

**This PR:**
- Adds a concept of "external focus" within the DOM helpers, which browser views provide. This lets us determine if focus is owned by VS Code even outside of the main workbench views.
- Updates native window focusing to specifically focus the main webContents, ensuring the workbench receives focus instead of a browser view.

### DIFF
--- a/src/vs/base/browser/dom.ts
+++ b/src/vs/base/browser/dom.ts
@@ -123,6 +123,66 @@ export const {
 
 //#endregion
 
+//#region External Focus Tracking
+
+/**
+ * A registry for functions that check if a component outside the normal DOM tree has focus.
+ * This is used to extend the concept of "window has focus" to include things like
+ * Electron WebContentsViews (browser views) that exist outside the document.
+ */
+const externalFocusCheckers = new Set<() => boolean>();
+
+/**
+ * Register a function that checks if a component outside the DOM has focus.
+ * This allows `hasExternalFocus` to detect when focus is in components like browser views.
+ *
+ * @param checker A function that returns true if the component has focus
+ * @returns A disposable to unregister the checker
+ */
+export function registerExternalFocusChecker(checker: () => boolean): IDisposable {
+	externalFocusCheckers.add(checker);
+
+	return toDisposable(() => {
+		externalFocusCheckers.delete(checker);
+	});
+}
+
+/**
+ * Check if any registered external component has focus.
+ * This is used to extend focus detection beyond the normal DOM to include
+ * components like Electron WebContentsViews.
+ *
+ * @returns true if any registered external component has focus
+ */
+export function hasExternalFocus(): boolean {
+	for (const checker of externalFocusCheckers) {
+		if (checker()) {
+			return true;
+		}
+	}
+	return false;
+}
+
+/**
+ * Check if VS Code has focus in any window, either via the normal DOM or via an
+ * external component like a browser view (which exists outside the document tree).
+ *
+ * @returns true if VS Code owns the current focus
+ */
+export function hasVSCodeFocus(): boolean {
+	for (const { window } of getWindows()) {
+		if (window.document.hasFocus()) {
+			return true;
+		}
+	}
+	if (hasExternalFocus()) {
+		return true;
+	}
+	return false;
+}
+
+//#endregion
+
 export function clearNode(node: HTMLElement): void {
 	while (node.firstChild) {
 		node.firstChild.remove();
--- a/src/vs/base/browser/dom.ts
+++ b/src/vs/base/browser/dom.ts
@@ -128,7 +128,7 @@ export const {
 /**
  * A registry for functions that check if a component outside the normal DOM tree has focus.
  * This is used to extend the concept of "window has focus" to include things like
- * Electron WebContentsViews (browser views) that exist outside the document.
+ * Electron WebContentsViews (browser views) that exist outside the workbench DOM.
  */
 const externalFocusCheckers = new Set<() => boolean>();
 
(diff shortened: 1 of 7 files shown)


src/vs/workbench/contrib/browserView/common/browserView.ts
```diff
@@ -83,6 +83,7 @@ export interface IBrowserViewModel extends IDisposable {
 	readonly isDevToolsOpen: boolean;
 	readonly canGoForward: boolean;
 	readonly error: IBrowserViewLoadError | undefined;
+	readonly focused: boolean;
```
jruales: nit: can we place this right after `loading` like you did in other places below, to keep the ordering consistent? (We can also defer this to do a general cleanup sweep in a followup PR where we ask copilot to put all of these in a consistent order)

src/vs/base/browser/dom.ts
```diff
@@ -123,6 +123,66 @@ export const {
 
 //#endregion
 
+//#region External Focus Tracking
+
+/**
+ * A registry for functions that check if a component outside the normal DOM tree has focus.
+ * This is used to extend the concept of "window has focus" to include things like
+ * Electron WebContentsViews (browser views) that exist outside the document.
```
jruales: nit: Maybe say "workbench DOM" instead of "document"?

src/vs/base/browser/dom.ts
```diff
@@ -123,6 +123,66 @@ export const {
...
+	return false;
+}
+
+/**
+ * Check if VS Code has focus in any window, either via the normal DOM or via an
+ * external component like a browser view (which exists outside the document tree).
+ *
+ * @returns true if VS Code owns the current focus
+ */
+export function hasVSCodeFocus(): boolean {
```
jruales: nit: thoughts on renaming to isVSCodeFocused() or isVSCodeInFocus() or doesVSCodeHaveFocus() ?

jruales (dismissed):

src/vs/base/browser/dom.ts
bpasero: I would prefer we do not have "vscode" in any function names as that is a product we are building, but in the sources its "Code OSS". How about "hasAppFocus" to make it more generic?

src/vs/platform/windows/electron-main/windowImpl.ts
```diff
@@ -392,6 +392,10 @@ export abstract class BaseWindow extends Disposable implements IBaseWindow {
 		}
 
 		win.focus();
+
+		// Also focus the webContents to ensure focus is taken from any
+		// child views that may currently have focus
+		win.webContents.focus();
```
bpasero, line 400: @deepak1556 can you maybe comment on this one, why would `win.focus()` not suffice?

src/vs/workbench/browser/window.ts
```diff
@@ -74,8 +74,9 @@ export abstract class BaseWindow extends Disposable {
 	}
 
 	private onElementFocus(targetWindow: CodeWindow): void {
-		const activeWindow = getActiveWindow();
-		if (activeWindow !== targetWindow && activeWindow.document.hasFocus()) {
+
```
bpasero, line 77: This sounds like a risky change. When `onElementFocus` is called, how can VSCode not have focus? What does the old solution not cover?

bpasero (changes requested):

src/vs/platform/windows/electron-main/windowImpl.ts
deepak1556: `Window.Focus` would eventually call into our view tracker code in Electron https://github.com/electron/electron/blob/486a9a61d2723755adc4be5e407888d755ee3b9c/shell/browser/ui/views/root_view.cc#L147-L156 that attempts to restore the last focused view. This would be noop before integrated browser but now that there are multiple native views in the same window, I believe @kycutler wants to explicitly restore focus to the main page view and not the integrated browsers. @kycutler is this your expectation ?

kycutler: Yes exactly. This is needed so actions triggered while focus is in the browser view can switch focus back to the workbench (critical for accessibility)

src/vs/workbench/browser/window.ts
kycutler: My goal here was to leave the existing behavior as unchanged as possible. I'm not exactly sure what scenarios would have an element receives programmatic focus while VS Code does not have focus -- but the logic here (and comments) suggest that it was done intentionally "to minimise disruption".

The old solution essentially checks if any window has focus (by calling `getActiveWindow` and then checking if its document has focus). But this only detects focus within the normal DOM. Since the browser views use WebContentsView which live outside the DOM, they are not covered by the existing checks, hence the added helpers for "external focus".

src/vs/platform/windows/electron-main/windowImpl.ts
bpasero: Can this comment explain this further?

kycutler: Updated the comment to clarify

src/vs/workbench/browser/window.ts
bpasero: But if that is the goal then can we change this line with less chance of breaking existing behaviour? For example: `getActiveWindow` falls back to `mainWindow` which is not in your code. That might be a subtle change in behaviour that could result in regressions.

kycutler: It falls back to mainWindow if there is no window with document focus, but then we still check for document focus in the returned window. So the logic is really just checking for focus in any window. I figure it's even clearer this way, but I can change the `hasAppFocus` behavior to use `getActiveWindow` with the same checks if you're worried about it.

bpasero: Ok lets give your solution a try and see if there are regressions.

bpasero (approved):

PR #287966, opened 2026-01-15
customEditor: fix stale cache after file deletion

### LINKED ISSUE #278883: Deleting files with vscode fs api can cause contributed custom text editors to not be able to open subsequent files with the same name
Yes

- VS Code Version:  1.106.2

Steps to Reproduce:

2. Run npm install
3. Run npm run build

AndyBodnar:
Ran into a weird bug where after deleting a file that was open in a custom editor (using the fs API), I couldn't open a new file with the same name - kept getting "file not found" even though the file existed.

Turns out the custom editor model manager wasn't cleaning up its cache when files got deleted. So when you create a new file with the same name, it would find the old cached model reference pointing to a resource that no longer exists.

The fix hooks into the existing file operation handler (which already handles moves) and adds a delete case that disposes all models associated with the deleted resource. Pretty straightforward and follows the same pattern as the existing move handling.

I tested this using the repro from the issue (https://github.com/Greglar4/vscodeBugExampleRepository) and it works - delete a file, recreate it with the same name, and the custom editor opens it correctly now.

Fixes #278883


mjbvz (approved):

mjbvz:
Thanks! General approach lgtm

joaomoreno (approved):

dmitrivMS (approved):

dileepyavan (approved):

DonJayamanne (approved):

PR #288032, opened 2026-01-15
chore - add `hono` package version 4.11.4

bpasero:


chrmarti (changes requested):
How does this make sense without a package.json change?

Tyriar (approved):

chrmarti (approved):

PR #288105, opened 2026-01-15
Add more details to provideWorkspaceChatContext

### LINKED ISSUE #280668: provideWorkspaceChatContext could use more explanation
Testing #280380

I thought this was related to implicit context and didn't get it at first. One thing that might help is saying it's included "for all chat requests" instead of "chat sessions".

Also if you plan to finalize this, it could use some explanation of what scenarios it should be used for, because I think it should be used sparingly, if at all. It has a cost of using tokens with no user control, and injecting them into a sensitive spot in the prompt that could distract the LLM.

Also its use-case is so different from the others that I almost wonder whether it should be part of a different provider? Or, do you see ChatContextProvider as being the one type for all different kinds of chat context, including other new kinds that we might invent in the future?

alexr00:
Fixes #280668

### DIFF
--- a/src/vscode-dts/vscode.proposed.chatContextProvider.d.ts
+++ b/src/vscode-dts/vscode.proposed.chatContextProvider.d.ts
@@ -58,7 +58,11 @@ declare module 'vscode' {
 		onDidChangeWorkspaceChatContext?: Event<void>;
 
 		/**
-		 * Provide a list of chat context items to be included as workspace context for all chat sessions.
+		 * TODO @API: should this be a separate provider interface?
+		 *
+		 * Provide a list of chat context items to be included as workspace context for all chat requests.
+		 * This should be used very sparingly to avoid providing useless context and to avoid using up the context window.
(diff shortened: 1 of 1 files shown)


Tyriar (approved):

src/vscode-dts/vscode.proposed.chatContextProvider.d.ts
```diff
@@ -58,7 +58,11 @@ declare module 'vscode' {
 		onDidChangeWorkspaceChatContext?: Event<void>;
 
 		/**
-		 * Provide a list of chat context items to be included as workspace context for all chat sessions.
+		 * TODO @API: should this be a separate provider interface?
+		 *
+		 * Provide a list of chat context items to be included as workspace context for all chat requests.
+		 * This should be used very sparingly to avoid providing useless context and to avoid using up the context window.
```
maxine-yx, line 64: Great improvement to the API documentation! The additional context about when to use this sparingly is very helpful.

maxine-yx (commented):
The enhanced documentation makes the API usage much clearer.

PR #288181, opened 2026-01-15
Agent sessions welcome view

osortega:
First iteration of agent sessions welcome view


rebornix (approved):

src/vs/workbench/contrib/chat/browser/agentSessions/agentSessionsControl.ts
```diff
@@ -313,4 +313,10 @@ export class AgentSessionsControl extends Disposable implements IAgentSessionsCo
 		this.sessionsList.setFocus([session]);
 		this.sessionsList.setSelection([session]);
 	}
+
+	setGridMarginOffset(offset: number): void {
```
bpasero, line 317: I think this should not be a method but a specialized CSS rule in the welcome.

PR #288196, opened 2026-01-16
Support `model` param in handoff

eleanorjboyd:
add the parameter "model" to handoff definitions. If the model is not found (or not visible to the user) then the handoff is not shown


aeschli (changes requested):
See my comment. The suggested feature does not fit with to the existing attributes. Mostly I want to understand the use case.

PR #288207, opened 2026-01-16
Integrated Browser: Fix buggy Right Click -> Copy into New Window

### LINKED ISSUE #284525: Integrated Browser: unexpected behavior with "Copy into New Window"

jruales:
Fixes https://github.com/microsoft/vscode/issues/284525
(Copy into New Window was very buggy. Not to be confused with Move into New Window, which has been working fine.)

This PR does the following:
* Creates an independent browser editor to avoid linked state
* Adds a resize observer to fix some issues where sometimes the copied window ended up with a zero-size webview or where a moved window ended up with a squashed webview until the user resized the window

---

Related feature: https://github.com/microsoft/vscode/issues/286579


src/vs/workbench/contrib/browserView/electron-browser/browserEditor.ts
```diff
@@ -368,8 +368,22 @@ export class BrowserEditor extends EditorPane {
...
+		// Watch for container size changes to handle window moves/resizes.
+		// This is especially important when copying or moving to a new window on a different
+		// monitor where the initial bounds may be incorrect until the window finishes layout.
+		// For some reason, the workbench's normal layout() calls do not cover this case.
+		const resizeObserver = new ResizeObserver(async () => {
+			if (this._model) {
+				const containerRect = this._browserContainer.getBoundingClientRect();
+				// Only proceed if we have valid bounds
+				if (containerRect.width > 0 && containerRect.height > 0) {
+					this.layout();
```
kycutler: Does this cause duplicate layout calls when e.g. resizing? I believe VS Code generally already handles this. If we just need to wait for the first insertion I'd prefer to clean up the listener after that's done

jruales: Now using ResizeObserver exclusively, instead of both that and the original layout() call

kycutler (approved):

PR #288426, opened 2026-01-16
Add hover options for `ActionList`, implement for chat model picker

### LINKED ISSUE #286401: Add descriptions to model picker
We support many models, which is awesome. 

However, it's overwhelming for the user to choose between them.

For beginners and pros alike, we should provide descriptions for the models so users understand the use cases and thus have the best experience.

I raised this Oct 9 2024, here was the feedback:

Image

### LINKED ISSUE #286402: Add descriptor for number on the right of model
Without a label, it's unclear what `0x`, `1x` etc. is or why users should care

Image

meganrogge:
fixes https://github.com/microsoft/vscode/issues/286401
fixes https://github.com/microsoft/vscode/issues/286402

- Added hover support infrastructure to `ActionList` with `IActionListItemHover` interface
- Implemented model descriptions in a new `modelDescriptions.ts` file with localized strings for various AI models
- Integrated hover functionality into the model picker to show descriptions and actions on kebyoard focus and mouse hover
- Added search parameter support to open the models management editor with a pre-filled search query for the `Show more` action on each item in the model selector
- See https://github.com/microsoft/vscode-copilot-chat/pull/3002 for corresponding `vscode-copilot-chat` PR

https://github.com/user-attachments/assets/32e5e440-8372-4eb1-af5a-900a9089a3e7


meganrogge:
https://github.com/user-attachments/assets/e7d4ad3b-5449-4c82-ae00-c0666398c277

src/vs/platform/actionWidget/browser/actionList.ts
```diff
@@ -401,10 +427,47 @@ export class ActionList<T> extends Disposable {
...
 		const element = this._list.element(focusIndex);
 		this._delegate.onFocus?.(element.item);
+
+		// Show hover on focus change
+		this._showHoverForElement(element, focusIndex);
+	}
+
+	private _getRowElement(index: number): HTMLElement | null {
+		// eslint-disable-next-line no-restricted-syntax
+		return this.domNode.ownerDocument.getElementById(this._list.getElementID(index));
```
benibenj, line 451: Use `getWindow(this.domNode)` so eslint disablement can be removed

src/vs/platform/actionWidget/browser/actionList.ts
```diff
@@ -401,10 +427,47 @@ export class ActionList<T> extends Disposable {
...
+		this._hoverService.hideHover();
+
+		// Show hover if the element has hover content or actions
+		if ((element.hover?.content || element.hover?.actions) && this.focusCondition(element)) {
+			// The List widget separates data models from DOM elements, so we need to
+			// look up the actual DOM node to use as the hover target.
+			const rowElement = this._getRowElement(index);
+			if (rowElement) {
+				const markdown = element.hover.content ? new MarkdownString(element.hover.content) : undefined;
+				this._hoverService.showInstantHover({
```
benibenj: Ideally we would use `this._hoverService.setupDelayedHover(...)` on the rendered list item as this is causing a lot of overhead here. Such as the flicker when moving the cursor over the hover widget. But `this._hoverService.setupDelayedHover(...)` does not react to focus change yet sadly.

src/vs/platform/actionWidget/browser/actionList.ts
```diff
@@ -401,10 +427,47 @@ export class ActionList<T> extends Disposable {
...
+					content: markdown ?? '',
+					target: rowElement,
+					actions: element.hover.actions,
+					additionalClasses: ['action-widget-hover'],
+					position: {
+						hoverPosition: HoverPosition.LEFT,
+						forcePosition: false,
+					},
+					appearance: {
+						showPointer: false,
```
benibenj: I think `showPointer: true` would be a bit nicer.

meganrogge:
I think I will remove the actions, which will simplify the hover implementation.

src/vs/platform/actionWidget/browser/actionList.ts
```diff
@@ -401,10 +427,47 @@ export class ActionList<T> extends Disposable {
...
+
+		// Show hover if the element has hover content or actions
+		if ((element.hover?.content || element.hover?.actions) && this.focusCondition(element)) {
+			// The List widget separates data models from DOM elements, so we need to
+			// look up the actual DOM node to use as the hover target.
+			const rowElement = this._getRowElement(index);
+			if (rowElement) {
+				const markdown = element.hover.content ? new MarkdownString(element.hover.content) : undefined;
+				this._hoverService.showInstantHover({
+					content: markdown ?? '',
```
benibenj, line 471: We usually have a hover delay such that it doesn't feel so aggressive: `workbench.hover.delay`

src/vs/platform/actionWidget/browser/actionList.ts
meganrogge: The `getElementById` is the issue here. Note that this eslint-disable line already existed before.

Screenshot 2026-01-20 at 12 38 53 PM

src/vs/platform/actionWidget/browser/actionList.ts
benibenj: Actually, I think we have `_hoverservice.setupDelayedHover(...)` which will do this for you. Only need to change the method and should be able to keep the rest

src/vs/platform/actionWidget/browser/actionList.ts
meganrogge: I think it is important for it to show on keyboard focus too. I am not seeing a flicker 🤔

lramos15 (approved):
This means you now own the model picker. Sorry I don't make the rules :)

src/vs/platform/actionWidget/browser/actionList.ts
```diff
@@ -30,13 +34,31 @@ export interface IActionListDelegate<T> {
...
+ */
+export interface IActionListItemHover {
+	/**
+	 * Content to display in the hover.
+	 */
+	readonly content?: string;
+	/**
+	 * Actions to show in the hover.
+	 */
+	readonly actions?: IHoverAction[];
```
lramos15, line 48: If you're removing actions should we remove it from the type?

src/vs/workbench/contrib/chat/browser/widget/input/modelPickerActionItem.ts
```diff
@@ -57,12 +57,23 @@ function modelDelegateToWidgetActionsProvider(delegate: IModelPickerDelegate, te
 					checked: true,
 					category: DEFAULT_MODEL_PICKER_CATEGORY,
 					class: undefined,
+					description: localize('chat.modelPicker.auto.detail', "Best for your request based on capacity and performance."),
```
lramos15, line 60: This sentence doesn't sound like a complete statement

src/vs/workbench/contrib/chat/browser/widget/input/modelPickerActionItem.ts
```diff
@@ -57,12 +57,23 @@ function modelDelegateToWidgetActionsProvider(delegate: IModelPickerDelegate, te
...
 				} satisfies IActionWidgetDropdownAction];
 			}
 			return models.map(model => {
+				// Build hover content combining tooltip and rate info
+				const hoverParts: string[] = [];
+				if (model.metadata.tooltip) {
+					hoverParts.push(model.metadata.tooltip);
+				}
+				if (model.metadata.detail) {
+					hoverParts.push(localize('chat.modelPicker.rateDescription', "Rate is counted at {0}.", model.metadata.detail));
```
lramos15, line 74: This assumes detail is always a rate for extension like cerebras it is not. These comments / strings should likely stay in the extension

src/vs/workbench/contrib/chat/browser/widget/input/modelPickerActionItem.ts
meganrogge: I copied exactly what was in the hover

PR #288507, opened 2026-01-16
sessions: show a preview of chat in the session hover

connor4312:
1. Split out the bulk of the chat list rendering into a `ChatListWidget`. This tries to encapsulate the bulk of the actual rendering code that was previously embedded in the `ChatListRenderer`, but `ChatListRenderer` also ties together input, mode switching, and such which are not directly tied to the chat rendering.
2. Adopt that to render a view of the chat response in hovers within the agent view. This is interactable, so you can action on confirmations and so on within this hover. This hover is optimized to hopefully not be too heavy weight: we only show the last request/response pair in the hover and also have a brief delay before the list is rendered so that mousing over the list of sessions while the hover is active doesn't cause slowdowns.

https://github.com/user-attachments/assets/6bcf7839-852d-40f1-83d4-255272c8a31a

Note: there is a currently a bug that affects hovering the first item due to bad overflow behavior, attempting to address in https://github.com/microsoft/vscode/pull/288500


src/vs/platform/hover/browser/hoverService.ts
```diff
@@ -300,6 +300,7 @@ export class HoverService extends Disposable implements IHoverService {
 			new HoverContextViewDelegate(hover, focus),
 			options.container
 		);
+		options.onDidShow?.();
```
connor4312, line 303: I added this because I wanted to defer actually loading the model until the hover is shown (after a delay). Options are resolved synchronously when the mouse first enters the hover target, and the hover consumer previously did not have a way to know when the hover was actually made visible.

bpasero (dismissed):
@connor4312 I am OK with the changes in the agent sessions world, but I would defer to others (@roblourens) for reasoning about the changes in the chat parts.

One visual feedback I would have is that the scrollbar not being all the way to the right feels a bit off:

image

cc @mrleemurray for feedback on the tooltip style overall.

roblourens (dismissed):

dbaeumer (approved):

PR #288535, opened 2026-01-17
Lazy render collapsed thinking group parts

### LINKED ISSUE #287176: Slowness in rendering thinking parts
This trace is just from clicking through a few chat sessions. I see tons of forced relayouts here which seem unnecessary. 

[Trace-20260112T140235.json.gz](https://github.com/user-attachments/files/24574473/Trace-20260112T140235.json.gz)

The sessions I loaded were finished, so their thinking parts were all collapsed, but it seems like each time we render one tool call into one, we check scrollHeight, forcing the browser to lay out everything and determine the size, so it took 400ms to render one fairly small chat response.

If we have to check scrollHeight, we should do it just once, not every time a change is made. Ideally we wouldn't do it at all, it doesn't seem necessary here.

Image

roblourens:
Fix #287176

fyi @justschen this is a pretty complicated change to do something seemingly simple, not rendering tool invocations that are collapsed away. Without this, I can scroll through a long chat session and render 100s of tool invocations that aren't displayed, and this is pretty slow. 

This is kind of a mess but the thinking part logic is complex. Lmk if there's anything you're concerned about

### DIFF
--- a/src/vs/workbench/contrib/chat/browser/widget/chatContentParts/chatThinkingContentPart.ts
+++ b/src/vs/workbench/contrib/chat/browser/widget/chatContentParts/chatThinkingContentPart.ts
@@ -21,6 +21,8 @@ import { ChatCollapsibleContentPart } from './chatCollapsibleContentPart.js';
 import { localize } from '../../../../../../nls.js';
 import { Codicon } from '../../../../../../base/common/codicons.js';
 import { ThemeIcon } from '../../../../../../base/common/themables.js';
+import { Lazy } from '../../../../../../base/common/lazy.js';
+import { IDisposable } from '../../../../../../base/common/lifecycle.js';
 import { autorun } from '../../../../../../base/common/observable.js';
 import { CancellationToken } from '../../../../../../base/common/cancellation.js';
 import { IChatMarkdownAnchorService } from './chatMarkdownAnchorService.js';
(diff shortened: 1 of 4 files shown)


src/vs/workbench/contrib/chat/browser/widget/chatContentParts/chatThinkingContentPart.ts
```diff
@@ -508,13 +529,93 @@ export class ChatThinkingContentPart extends ChatCollapsibleContentPart implemen
...
+		toolInvocationId?: string,
+		toolInvocationOrMarkdown?: IChatToolInvocation | IChatToolInvocationSerialized | IChatMarkdownContent,
+		originalParent?: HTMLElement
+	): void {
+		// Track tool invocation metadata immediately (for title generation)
+		this.trackToolMetadata(toolInvocationId, toolInvocationOrMarkdown);
+		this.appendedItemCount++;
+
+		// If expanded or has been expanded once, render immediately
+		if (this.isExpanded() || this.hasExpandedOnce) {
```
justschen:
```suggestion
		if (this.isExpanded() || this.hasExpandedOnce || this.fixedScrollingMode) {
```

justschen (dismissed):

src/vs/workbench/contrib/chat/browser/widget/chatContentParts/chatThinkingContentPart.ts
justschen: in `fixedScrolling` mode, it looks like it never renders

src/vs/workbench/contrib/chat/browser/widget/chatContentParts/chatThinkingContentPart.ts
```diff
@@ -508,13 +529,93 @@ export class ChatThinkingContentPart extends ChatCollapsibleContentPart implemen
...
+		originalParent?: HTMLElement
+	): void {
 		if (!content.hasChildNodes() || content.textContent?.trim() === '') {
 			return;
 		}
 
-		// save the first item info for potential restoration later
-		if (this.appendedItemCount === 0 && originalParent) {
+		// Save the first item info for potential restoration later
+		if (this.appendedItemCount === 1 && originalParent) {
```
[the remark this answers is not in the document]
roblourens: I don't think this is an issue, but it's hard to follow

dmitrivMS (approved):

PR #288584, opened 2026-01-17
Optimize initial hint at low widths

### LINKED ISSUE #288118: Terminal welcome text is ugly
This is my normal width that I use with explorer on the left, chat on the right, terminal in the middle. On my 13" MBP

Image

It's just a really large, wrapped text block that is ugly, so I never read it

Tyriar:
Fixes #288118

Approx breakpoints:

image

image


alexdima (approved):

Tyriar:
@meganrogge

PR #288644, opened 2026-01-17
Prevent calling updateElementHeight while rendering the element

roblourens:
There are too many ways this can happen on accident so just applying an overall check


rebornix (approved):
@roblourens I hope the tree/list can take care of this for us

roblourens:
The list could throw a more specific error, but I think this is misuse of the list and it shouldn't swallow the error.

PR #288651, opened 2026-01-17
Improve type safety for dynamic stylesheet rules

chetanr-25:
This PR improves type safety when accessing dynamic stylesheet rules
by ensuring a consistent CSSRuleList return type.

No functional behavior is changed.


chetanr-25:
This is my first contribution — happy to make any requested changes. Thank you for reviewing!

mjbvz (dismissed):
Thanks!

bpasero (dismissed):

chetanr-25:

chetanr-25:
agree

src/vs/base/browser/domStylesheets.ts
```diff
@@ -111,16 +111,23 @@ function getSharedStyleSheet(): HTMLStyleElement {
...
+		return style.sheet.cssRules;
 	}
-	if (style?.sheet?.cssRules) {
-		return style.sheet.cssRules; // FF
-	}
-	return [];
+
+	const emptyRules: CSSRule[] = [];
+
+	return {
```
mjbvz: I'd prefer this just return undefined in these cases 

And actually do we even need this helper function now? The 2 callers can be converted to `style.sheet?.cssRules` instead

chetanr-25: Agreed — removed the helper and updated both call sites to use
style.sheet?.cssRules directly. Thanks!

mjbvz: Hey looks like this change didn't make it in yet, right?

chetanr-25: Yep, it’s approved but not merged yet. Let me know if I should do anything else or if it’s waiting on a maintainer.

mjbvz: Sorry what I meant is that I don't think we need `getDynamicStyleSheetRules` at all now. I pushed a change that just inlines which simplifies the code quite a bit. Let me know if you don't think that will work

chetanr-25: The inlined changes look great and definitely simplify the code. I'm happy with this approach. Thanks for the help!

mjbvz (approved):

mjbvz (approved):

Yoyokrazy (approved):

PR #288653, opened 2026-01-17
Change chat input layout to be based on ResizeObserver

roblourens:
Simplify the logic and avoid layout thrashing when we would layout multiple times

### DIFF
--- a/src/vs/workbench/contrib/chat/browser/widgetHosts/viewPane/chatViewPane.ts
+++ b/src/vs/workbench/contrib/chat/browser/widgetHosts/viewPane/chatViewPane.ts
@@ -548,7 +548,7 @@ export class ChatViewPane extends ViewPane implements IViewWelcomeDelegate {
 
 	//#region Chat Control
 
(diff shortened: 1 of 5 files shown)


TylerLeonhardt (approved):

src/vs/workbench/contrib/chat/browser/widgetHosts/viewPane/chatViewPane.ts
```diff
@@ -548,7 +548,7 @@ export class ChatViewPane extends ViewPane implements IViewWelcomeDelegate {
 
 	//#region Chat Control
 
-	private static readonly MIN_CHAT_WIDGET_HEIGHT = 120;
+	private static readonly MIN_CHAT_WIDGET_HEIGHT = 116;
```
roblourens, line 551: Here's what this was fixing

https://github.com/user-attachments/assets/d09d0d01-4804-4c2c-acfa-c693e1f2723c

PR #288760, opened 2026-01-18
remove todo item description field and make it readonly

bhavyaus:
cc: @digitarald

### DIFF
--- a/src/vs/workbench/contrib/chat/browser/accessibility/chatResponseAccessibleView.ts
+++ b/src/vs/workbench/contrib/chat/browser/accessibility/chatResponseAccessibleView.ts
@@ -92,7 +92,7 @@ export function getToolSpecificDataDescription(toolSpecificData: ToolSpecificDat
 				return '';
 			}
 			const todoDescriptions = todos.map(t =>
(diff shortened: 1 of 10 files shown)


src/vs/workbench/contrib/chat/browser/accessibility/chatResponseAccessibleView.ts
```diff
@@ -92,7 +92,7 @@ export function getToolSpecificDataDescription(toolSpecificData: ToolSpecificDat
 				return '';
 			}
 			const todoDescriptions = todos.map(t =>
-				localize('todoItem', "{0} ({1}): {2}", t.title, t.status, t.description)
+				localize('todoItem', "{0} ({1})", t.title, t.status)
```
roblourens, line 95: Doesn't really need to be localized

roblourens (approved):

src/vs/workbench/contrib/chat/browser/accessibility/chatResponseAccessibleView.ts
bhavyaus: No. cc: @meganrogge Should we remove this loclize call here and in the other places in this file?

PR #288813, opened 2026-01-19
Fix Inline Style CSP Error

nigeldias-okta:
While enforcing nonce for CSP style directive in our web application, which uses Monaco Editor, multiple violations were found, originating from two areas:
1. While creating style tags at https://github.com/microsoft/vscode/blob/bf56edffb59c43fa0de636c3aa1d548770b168b8/src/vs/base/browser/domStylesheets.ts#L51 since it is missing the nonce attribute
2. Due to inline style usages such as https://github.com/microsoft/vscode/blob/bf56edffb59c43fa0de636c3aa1d548770b168b8/src/vs/editor/browser/view/viewOverlays.ts#L177

Hence, the proposed changes include:
1. Converting each inline style usage to a style tag, if a nonce exists
2. Adding the nonce attribute to the generated style tag.

Issues related to this change:
https://github.com/microsoft/monaco-editor/issues/4927
https://github.com/microsoft/monaco-editor/issues/271

### DIFF
--- a/src/vs/editor/browser/viewParts/lineNumbers/lineNumbers.ts
+++ b/src/vs/editor/browser/viewParts/lineNumbers/lineNumbers.ts
@@ -14,6 +14,7 @@ import { ViewContext } from '../../../common/viewModel/viewContext.js';
 import * as viewEvents from '../../../common/viewEvents.js';
 import { registerThemingParticipant } from '../../../../platform/theme/common/themeService.js';
 import { editorDimmedLineNumber, editorLineNumbers } from '../../../common/core/editorColorRegistry.js';
+import { mainWindow } from '../../../../base/browser/window.js';
 
 /**
  * Renders line numbers to the left of the main view lines content.
@@ -196,9 +197,11 @@ export class LineNumbersOverlay extends DynamicViewOverlay {
 				extraClassNames += ' active-line-number';
 			}
 
-
+			const layoutClassName = ' line-numbers-layout-' + Math.random().toString(36).substring(2);
 			output[lineIndex] = (
-				`<div class="${LineNumbersOverlay.CLASS_NAME}${lineHeightClassName}${extraClassNames}" style="left:${this._lineNumbersLeft}px;width:${this._lineNumbersWidth}px;">${renderLineNumber}</div>`
+				mainWindow.cspNonce ?
+					`<div class="${LineNumbersOverlay.CLASS_NAME}${layoutClassName}${lineHeightClassName}${extraClassNames}"><style nonce="${mainWindow.cspNonce}">.${layoutClassName}{left:${this._lineNumbersLeft}px;width:${this._lineNumbersWidth}px;}</style>${renderLineNumber}</div>` :
+					`<div class="${LineNumbersOverlay.CLASS_NAME}${lineHeightClassName}${extraClassNames}" style="left:${this._lineNumbersLeft}px;width:${this._lineNumbersWidth}px;">${renderLineNumber}</div>`
 			);
 		}
 
--- a/src/vs/base/browser/domStylesheets.ts
+++ b/src/vs/base/browser/domStylesheets.ts
@@ -43,6 +43,9 @@ export function createStyleSheet(container: HTMLElement = mainWindow.document.he
 	const style = document.createElement('style');
 	style.type = 'text/css';
 	style.media = 'screen';
+	if (mainWindow.cspNonce) {
+		style.nonce = mainWindow.cspNonce;
+	}
 	beforeAppend?.(style);
 	container.appendChild(style);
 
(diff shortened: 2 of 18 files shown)


src/vs/editor/browser/viewParts/lineNumbers/lineNumbers.ts
```diff
@@ -196,9 +197,11 @@ export class LineNumbersOverlay extends DynamicViewOverlay {
 				extraClassNames += ' active-line-number';
 			}
 
-
+			const layoutClassName = ' line-numbers-layout-' + Math.random().toString(36).substring(2);
```
JoshuaNeuenhoff-at-AVEVA, line 200: The leading space here is causing the class name in the <style> tag to have a space between the `.` and the name:

Image

This leads to the style not being applied.

nabramovitz:
Following up on the offer in microsoft/monaco-editor#271 ([comment](https://github.com/microsoft/monaco-editor/issues/271#issuecomment-5227618354)) — we ran this PR's approach against our CSP measurement setup. Summary: the conversion works and the violation numbers are dramatic, but there are four concrete defects and a design cost worth addressing before this merges.

**Method.** We ported the diff by hand onto monaco-editor 0.52.2 ESM (what consumers currently embed — the touched code is close enough that every hunk applied cleanly), bundled it standalone, and loaded it into a page of a large Angular console served with an enforced policy of `style-src 'self' 'nonce-…'` (no `unsafe-inline`) and `style-src-attr 'none'`, with a `securitypolicyviolation` listener and a violation-report endpoint. Editor exercised with JSON content covering word wrap, RTL text, whitespace rendering, multi-line selection, ranged and whole-line decorations, and clipboard copy. Chromium 151. Three runs: patched with `window.cspNonce` set, patched without it, and unpatched 0.52.2 as control.

**Results.**

| run | `style-src-attr` violations | rendering | editor create | 20-step scroll re-render | DOM nodes |
|---|---|---|---|---|---|
| unpatched control | 1234 | destroyed (every line at `top:0`) | 72ms | 162ms | 890 |
| patched, no nonce | 1072 | destroyed | 66ms | 158ms | 890 |
| patched + nonce | 7 | correct — colors, wrap, RTL, selection all render | 65ms | 309ms | 1493 |

1234 → 7 with a working editor is a strong result; the approach is viable.

**Defects found (all reproduce live, not just by reading the diff):**

1. **Confirms @JoshuaNeuenhoff-at-AVEVA's finding** on `lineNumbers.ts`: `layoutClassName` starts with a space, so the emitted selector is `. line-numbers-layout-…` — invalid CSS, whole rule dropped. Measured: line numbers get computed width 7.2px instead of the intended 36px and render unpositioned.

2. **Decorations lose their caller-supplied class.** In `decorations.ts` the new `const className` shadows the `className` parameter of `_renderNormalDecoration`, so a decoration renders as `class="cdr <random> <random>"` — the decoration's own class never reaches the DOM. Live check: a decoration with `className: 'probe-deco'` produces no `.probe-deco` element. This silently breaks anything styled through decorations: error/warning squiggles, find-match highlights, diff decorations.

3. **Copy-with-highlighting breaks, and the conversion there was unnecessary.** `viewModelImpl.ts`'s `getRichTextToCopy` builds clipboard HTML — it never enters the page DOM, so the page's CSP does not apply to it and the inline styles were fine as they were. Converted to a `<style>` tag, Chromium's clipboard sanitizer strips the tag entirely: control pastes tokens with `color:#d4d4d4;background:#1e1e1e`, patched pastes unformatted default-black text. (The same hunk also has a missing closing quote after the nonce value, but reverting this file entirely seems like the right fix.)

4. **Two emitters are missed, so the policy still can't reach zero.** All 7 residual violations trace to `_renderWholeLineDecorations` (`style="left:0;width:100%;"`), and `marginDecorations.ts` string-builds a `style="` attribute the same way. Whole-line decorations are common (search current-line match, diff line backgrounds), so this fires in ordinary use.

**Design cost worth a look:** with the nonce active, 41% of the editor's DOM nodes are `<style>` elements (617 of 1493 for ~28 visible lines), DOM grows ~68%, and scroll re-render measured ~1.9× slower — each re-render mints new random classnames, so every frame inserts fresh stylesheets and invalidates style. Two mitigations that fall out of the PR's own code: `selections.ts` already derives deterministic classnames from the geometry (stable across identical renders), which the other view parts could copy; and aggregating one `<style>` per render pass instead of one per element would cut the node count by an order of magnitude.

Also worth stating in the PR description: the whitespace SVG branch has `substring(36)` (empty suffix — colliding classnames) and emits `class=${className}` unquoted with no space before `viewBox`; we couldn't exercise that path in our scenario, so those are from reading the diff, not from measurement.

Happy to re-run the same three-way measurement against an updated revision.

PR #288867, opened 2026-01-19
Add suspend/resume telemetry for reliability insights

chrmarti:
https://github.com/microsoft/vscode-internalbacklog/issues/6543


Tyriar (approved):

src/vs/platform/native/electron-main/nativeHostMainService.ts
```diff
@@ -119,6 +121,18 @@ export class NativeHostMainService extends Disposable implements INativeHostMain
 
 			this.onDidResumeOS = Event.fromNodeEventEmitter(powerMonitor, 'resume');
 
+			// Telemetry for power events
```
bpasero, line 124: @chrmarti can we move this somewhere else, it has nothing to do with `nativeHostMainService` which is about providing methods to the renderer.

A better place would be `app.ts`.

PR #288895, opened 2026-01-19
chat: make inline code block bg brighter

ulugbekna:


mjbvz:

PR #289026, opened 2026-01-20
Better Shebang Language Detection (Deno, Bun, etc) (fix #287819)

### LINKED ISSUE #287819: Better Shebang Language Detection (Deno, Bun, etc)
Along the same lines as #182613, it would be nice if the shebang language detection worked a bit better... especially with /usr/bin/env (-S) where the language/executable after env or env -S is supported, or can determine the type of file, but not a direct match.

ex: `#!/usr/bin/env -S deno -A` or similar should be treated as typescript.

Runtimes really matter, and it would be nice if more were directly supported.  I've been a big fan of scripting in TS with Deno since it's a pleasure to do somewhat complex things that can require external repo/modules without a separate install step.  I'd really like to be able to do this without having to add the `.ts` file extension in order to get language support in VS Code.

bpasero:

### DIFF
--- a/extensions/typescript-basics/package.json
+++ b/extensions/typescript-basics/package.json
@@ -27,6 +27,7 @@
           ".cts",
           ".mts"
         ],
+        "firstLine": "^#!.*\\b(deno|bun|ts-node)\\b",
         "configuration": "./language-configuration.json"
       },
       {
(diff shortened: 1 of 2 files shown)


lszomoru (approved):

extensions/typescript-basics/package.json
```diff
@@ -27,6 +27,7 @@
           ".cts",
           ".mts"
         ],
+        "firstLine": "^#!.*\\b(deno|bun|ts-node)\\b",
```
DanielRosenwasser, line 30: Why not just bare `node` as well?

DanielRosenwasser: I guess JS already claimed this

https://github.com/microsoft/vscode/blob/22d2ac24f5db1edc8b115c7c7e2244ab66a23f94/extensions/javascript/package.json#L48

PR #289150, opened 2026-01-20
Appending node exec path for running srt

dileepyavan:


src/vs/workbench/contrib/terminalContrib/chatAgentTools/common/terminalSandboxService.ts
Tyriar, line 52: Did you check the actual product build to make sure it's there? Some modules get cleaned and go away during packaging.

src/vs/workbench/contrib/terminalContrib/chatAgentTools/common/terminalSandboxService.ts
```diff
@@ -65,7 +65,11 @@ export class TerminalSandboxService implements ITerminalSandboxService {
 		if (!this._sandboxConfigPath || !this._tempDir) {
 			throw new Error('Sandbox config path or temp dir not initialized');
 		}
-		return `"${this._srtPath}" TMPDIR=${this._tempDir.fsPath} --settings "${this._sandboxConfigPath}" "${command}"`;
+		const execPath = isNative ? (this._environmentService as { execPath?: string }).execPath : undefined;
```
Tyriar: I think `isNative` will be false as this is not a node proc, the old PR had this before and it was removed due to a comment?

Tyriar (dismissed):

src/vs/workbench/contrib/terminalContrib/chatAgentTools/common/terminalSandboxService.ts
dileepyavan: Yeah , I looked at this path from the approot directory.
image

src/vs/workbench/contrib/terminalContrib/chatAgentTools/common/terminalSandboxService.ts
dileepyavan: isNative is still present int this file. I think this will be a node proc as its electron which has node runtime. Validation is done in local.

```
private _initTempDir(): void {
		if (this.isEnabled() && isNative) {
```

dileepyavan: @Tyriar is there a way I can do a local version published and validate. This is working in dev but might be better to confirm.

dileepyavan: Tested with local insiders and IsNative is false. It still failing with published insiders.

chrmarti (approved):

PR #289163, opened 2026-01-20
fix: add engines property to root package.json

murataslan1:
Add engines property to specify the required Node.js and npm versions:
- node: >=22.21.1 (matches .nvmrc)
- npm: >=10 (bundled with Node.js 22)

This helps contributors quickly identify the required versions and throws an error if an unsupported version is used.

Fixes #252372

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,6 +9,10 @@
   "main": "./out/main.js",
   "type": "module",
   "private": true,
+  "engines": {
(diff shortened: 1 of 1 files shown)


murataslan1:
Closing as duplicate  - this is superseded by #292192 which targets the same fix. Thanks!

PR #289177, opened 2026-01-20
[draft] More comprehensive source edit tracking from format document

a-stewart:
This is a draft PR - not yet ready for review - looking into creating more source edit tracking around format document


src/vs/editor/contrib/format/browser/format.ts
```diff
@@ -418,21 +423,29 @@ export async function getDocumentFormattingEditsUntilResult(
 	return undefined;
 }
```
a-stewart, line 424: I am unsure if we would also want to return a `FormattingEdits` from `getDocumentFormattingEditsUntilResult` and `getDocumentRangeFormattingEditsUntilResult`.

For now I have not done so because the only usage is in the same file and the reason can't be used there.

a-stewart:
Note, that the proposal in https://github.com/microsoft/vscode/pull/289173 would allow for the TextEdits themselves to contain a reason and not require any change to the signatures here

PR #289179, opened 2026-01-20
feat: add dmg for macOS

### LINKED ISSUE #265849: Install Visual Studio Code Directly Into the Applications Folder in macOS
Currently when we install Visual Studio Code on any version of macOS, it gets downloaded and stays in the downloads folder itself. VS Code would run from this downloads folder and stay right there until the user decides to move it to the applications folder.

The intended behaviour in macOS is to install the applications directly to the applications folder and run applications from there without any requirement of user to move or perform any action for an application to be put into the applications folder.

I do have the knowledge that VS Code is not a direct macOS app but based on electron.js. Still, another application called Discord uses the same technology but installs the application directly into applications folder by downloading a DMG and dragging and dropping it after running.

I request to please look into this issue and have Visual Studio Code directly installed into the applications folder as soon as possible.

deepak1556:
Fixes https://github.com/microsoft/vscode/issues/265849

Uses https://github.com/dmgbuild/dmgbuild but instead of having to include the sources directly we pull it via the npm package [dmg-builder](https://www.npmjs.com/package/dmg-builder) that is from https://github.com/electron-userland/electron-builder/tree/master/packages/dmg-builder.


deepak1556:
Build https://monacotools.visualstudio.com/Monaco/_build/results?buildId=394619&view=results

joaomoreno (dismissed):

TylerLeonhardt (approved):

PR #289185, opened 2026-01-20
action list hover follow ups

meganrogge:


lramos15:
Since we're no longer doing show more can we remove the new command that was added with query param?

lramos15 (approved):

Issue #274403, opened 2025-10-31
Migrate chat implementation  to use URIs instead of string ids to track sessions


mjbvz:
Tracks migrating our chat implementation to use uris instead of string ids to identify chat sessions internally. Uris are preferred as they can correctly support contributed providers 

This is a pretty major change so it will broken up into smaller steps. I've done but lot of work trying to get the basics setup but now need help finishing up in individual areas 

- [x] @mjbvz Add initial support for session uris: fb6055cb587de8c95dbd6e1619edaf77ec987ced  
- [x] @mjbvz Update IChatService to use uris: db5ee1e0f4b6d7d5e8fc880f4096116655d915bc 
- [x] @mjbvz Lots of plumbing through of session uris: eef25fc6d68facf70b457602ec10b7805fede0c6 3bb2ff6262ea3f52723f14bdd1ce533872b430cb a1f7be35bf26d658c9fc7383ecbd7726babc0ecd d9d1b03c6b6496e12fa724a2513ed0d79f6eceff

- [x] @joshspicer  Convert the `clearChatEditor` function. We definitely should not be using the forChatSessionTypeAndId call I added but I'm not really sure what we should be doing instead 

    Blocks removing `forChatSessionTypeAndId` and `LocalChatSessionUri.parse`

- [x] @osortega  Convert the hybrid session stuff in `getHybridSessionsForProvider`

- [x] @osortega  Remove the use of fake `IChatSessionItem` entries with `id: "show-history"` and `id: "workbench.panel.chat.view.copilot"`

    Blocks removing `IChatSessionItem.id`

- [x] @roblourens `IChatAgentRequest.sessionId`

- [x] @connor4312 Convert `IChatEditingSession.chatSessionId`

- [x] @justschen Convert `IChatChangesSummary.sessionId`

- [ ] @aeschli Adopt in code mapper. Specifically for `ICodeMapperRequest.chatSessionId`
    
    We also need to remove  `MappedEditsRequest.chatSessionId` from `src/vscode-dts/vscode.proposed.mappedEditsProvider.d.ts`

- [x] @Tyriar Adopt for the terminal tool (`IChatTerminalToolInvocationData.sessionId` plus `ChatTerminalToolProgressPart` and `RunInTerminalTool`)
    
    Blocks removing `IChatResponseViewModel.sessionId` and `IChatRequestViewModel.sessionId`

- [x] @bhavyaus Adopt for `ManageTodoListTool`

[...]

PR #281383, opened 2025-12-04
Fix auto approval of terminal for background session

DonJayamanne:

PR #289197, opened 2026-01-20
Remove workaround for bad session ids

mjbvz:
For #274403

### DIFF
--- a/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/runInTerminalConfirmationTool.ts
+++ b/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/runInTerminalConfirmationTool.ts
@@ -5,7 +5,6 @@
 
 import { CancellationToken } from '../../../../../../base/common/cancellation.js';
 import { Codicon } from '../../../../../../base/common/codicons.js';
-import { URI } from '../../../../../../base/common/uri.js';
 import { localize } from '../../../../../../nls.js';
 import { CountTokensCallback, IPreparedToolInvocation, IToolData, IToolInvocation, IToolInvocationPreparationContext, IToolResult, ToolDataSource, ToolInvocationPresentation, ToolProgress } from '../../../../chat/common/tools/languageModelToolsService.js';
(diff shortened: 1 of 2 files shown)


src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/runInTerminalConfirmationTool.ts
```diff
@@ -59,18 +58,6 @@ export const ConfirmTerminalCommandToolData: IToolData = {
 
 export class ConfirmTerminalCommandTool extends RunInTerminalTool {
 	override async prepareToolInvocation(context: IToolInvocationPreparationContext, token: CancellationToken): Promise<IPreparedToolInvocation | undefined> {
-		// Safe-guard: If session is the chat provider specific id
```
mjbvz, line 62: @DonJayamanne Let me know if there's a specific scenario that we still need this for. It won't work correctly with sessions that use resources

justschen (approved):

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/runInTerminalConfirmationTool.ts
DonJayamanne: It was used to fix auto approval of terminal tool calls from background sessions
https://github.com/microsoft/vscode/pull/281383

I guess one way to test this is:
* Start background session
* Run a terminal command, and then approve this
* Run again and verify it gets auto approved
* Add approve rules and verify the terminal command is auto approved based on your approve rules.

DonJayamanne: I.e. I'd suggest testing the changes

mjbvz: Yes stills works for me when I select `allow in session` or `allow in this workspace`. Just `allow` shouldn't auto approve, right?

PR #289235, opened 2026-01-20
Ensure embedded chat terminal output expands on real data

### LINKED ISSUE #288352: Terminal tool expand not working
Image

meganrogge:
When running commands like `npm run watch`, the terminal output in chat wasn't auto-expanding to show the output. The root cause is that shell integration sequences arrive as "data" events before any actual command output appears. The old implementation cancelled the 500ms fallback timeout as soon as any data arrived, then relied on a 50ms timeout to check for real output. But at that point, only shell sequences had arrived - **no actual output yet** - so nothing expanded. The 500ms path had already been cancelled, so there was no second chance.

The fix allows both timeout paths to run concurrently instead of having the first data event cancel the longer timeout. Now, when a command executes, the 500ms "NoData" timeout starts. If data arrives, a 50ms "DataEvent" timeout also starts, but it no longer cancels the NoData path. **Whichever path finds real output first** will expand the terminal and cancel the other. This means if shell integration sequences arrive early but real output comes later (within 500ms), the NoData path will still successfully expand the output.

Additionally, the PR adds detection for progress-bar style output that updates the same terminal line. Previously, the `hasRealOutput()` check only looked at whether the cursor had moved past the command's executed marker. Progress bars that overwrite the same line (as is the case for `npm i`) wouldn't move the cursor, so they weren't detected as real output. Now, if multiple data events are received `receivedDataCount > 1`, it's treated as real output since shell integration sequences typically only fire once.

The PR also improves the visual experience by delaying the expanded state until content is actually ready to display, preventing an empty flash. A polling mechanism waits up to 1 second for terminal content before showing the "no output" message, handling cases where the buffer isn't immediately ready. Finally, terminal commands inside thinking containers now auto-collapse on success like regular commands, and auto-expand the thinking wrapper when output appears or errors occur.

Also added tests for these cases.

fixes #288352

cc @tyriar

### DIFF
--- a/src/vs/workbench/contrib/chat/browser/widget/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
+++ b/src/vs/workbench/contrib/chat/browser/widget/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
@@ -592,7 +592,7 @@ export class ChatTerminalToolProgressPart extends BaseChatToolInvocationSubPart 
 			const autoExpand = store.add(new TerminalToolAutoExpand({
 				commandDetection,
 				onWillData: terminalInstance.onWillData,
-				shouldAutoExpand: () => !this._outputView.isExpanded && !this._userToggledOutput && !this._store.isDisposed,
+				shouldAutoExpand: () => !this._outputView.isExpanded && !this._userToggledOutput && !this._store.isDisposed  && !expandedStateByInvocation.get(this.toolInvocation),
 				hasRealOutput,
 			}));
 			store.add(autoExpand.onDidRequestExpand(() => {
--- a/src/vs/workbench/contrib/chat/browser/widget/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
+++ b/src/vs/workbench/contrib/chat/browser/widget/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
@@ -24,6 +24,7 @@ import { IChatRendererContent } from '../../../../common/model/chatViewModel.js'
 import '../media/chatTerminalToolProgressPart.css';
 import type { ICodeBlockRenderOptions } from '../codeBlockPart.js';
 import { Action, IAction } from '../../../../../../../base/common/actions.js';
+import { timeout } from '../../../../../../../base/common/async.js';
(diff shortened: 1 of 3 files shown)


src/vs/workbench/contrib/chat/browser/widget/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
```diff
@@ -588,7 +595,17 @@ export class ChatTerminalToolProgressPart extends BaseChatToolInvocationSubPart
 				shouldAutoExpand: () => !this._outputView.isExpanded && !this._userToggledOutput && !this._store.isDisposed,
```
justschen:
```suggestion
				shouldAutoExpand: () => !this._outputView.isExpanded && !this._userToggledOutput && !this._store.isDisposed  && !expandedStateByInvocation.get(this.toolInvocation),
```

justschen (dismissed):

src/vs/workbench/contrib/chat/browser/widget/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
justschen: this seems to fix the issue we were discussing, and thru testing looks good in both outside and inside thinking.

not sure if there are any other weird implications with this tho

src/vs/workbench/contrib/chat/browser/widget/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
```diff
@@ -863,19 +881,22 @@ class ChatTerminalToolOutputSection extends Disposable {
...
-
 		if (!expanded) {
+			this._setExpanded(false);
 			this._renderedOutputHeight = undefined;
 			this._isAtBottom = true;
 			this._onDidChangeHeight();
 			return true;
 		}
 
+		// For expanding, prepare content first but don't show until we have something to display
```
meganrogge:
```suggestion
```

Tyriar (approved):

PR #289247, opened 2026-01-20
Browser paused state updates

kycutler:
image

Part of #286579

### DIFF
--- a/src/vs/workbench/contrib/browserView/electron-browser/browserEditor.ts
+++ b/src/vs/workbench/contrib/browserView/electron-browser/browserEditor.ts
@@ -439,7 +439,7 @@ export class BrowserEditor extends EditorPane {
 		this._overlayPauseContainer.classList.toggle('show-message', hasNotificationOverlay);
 
 		if (hasNotificationOverlay) {
-			this._overlayPauseHeading.textContent = localize('browser.overlayPauseHeading.notification', "Paused due to notification");
+			this._overlayPauseHeading.textContent = localize('browser.overlayPauseHeading.notification', "Paused due to Notification");
 			this._overlayPauseDetail.textContent = localize('browser.overlayPauseDetail.notification', "Dismiss the notification to continue using the browser.");
 		} else {
 			this._overlayPauseHeading.textContent = '';
--- a/src/vs/workbench/contrib/browserView/electron-browser/browserEditor.ts
+++ b/src/vs/workbench/contrib/browserView/electron-browser/browserEditor.ts
@@ -26,7 +26,7 @@ import { IEditorGroup } from '../../../services/editor/common/editorGroupsServic
 import { IEditorOptions } from '../../../../platform/editor/common/editor.js';
 import { IKeybindingService } from '../../../../platform/keybinding/common/keybinding.js';
 import { StandardKeyboardEvent } from '../../../../base/browser/keyboardEvent.js';
-import { BrowserOverlayManager } from './overlayManager.js';
+import { BrowserOverlayManager, BrowserOverlayType, IBrowserOverlayInfo } from './overlayManager.js';
 import { getZoomFactor, onDidChangeZoomLevel } from '../../../../base/browser/browser.js';
 import { ILogService } from '../../../../platform/log/common/log.js';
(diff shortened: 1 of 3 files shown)


src/vs/workbench/contrib/browserView/electron-browser/browserEditor.ts
```diff
@@ -382,17 +395,20 @@ export class BrowserEditor extends EditorPane {
...
 		// Error container: shown when there's a load error
 		this._errorContainer.style.display = hasError ? '' : 'none';
 
-		// Placeholder screenshot: shown when the view is hidden due to overlays
-		this._placeholderScreenshot.style.display = shouldShowPlaceholder ? '' : 'none';
-		this._placeholderScreenshot.classList.toggle('blur', shouldShowPlaceholder);
+		// Placeholder screenshot: shown when there is a page loaded (even when the view is not hidden, so hiding is smooth)
+		this._placeholderScreenshot.style.display = isViewingPage ? '' : 'none';
+
+		// Pause overlay: fades in when an overlay is detected
```
jruales, line 410: nit: It might be slightly confusing if we overload the term "overlay" to mean the both the overlays themselves as well as the message we show when there are overlays

src/vs/workbench/contrib/browserView/electron-browser/browserEditor.ts
```diff
@@ -408,13 +424,29 @@ export class BrowserEditor extends EditorPane {
...
+		const hasOverlappingOverlay = overlappingOverlays.length > 0;
+		this.updateOverlayPauseMessage(overlappingOverlays);
 		if (hasOverlappingOverlay !== this._overlayVisible) {
 			this._overlayVisible = hasOverlappingOverlay;
 			this.updateVisibility();
 		}
 	}
 
+	private updateOverlayPauseMessage(overlappingOverlays: readonly IBrowserOverlayInfo[]): void {
+		// Only show the pause message for notification overlays
```
jruales, line 437: I'm glad we're showing that, since I agree that it can be confusing when it's paused because of a toast or notification and the user might not know otherwise how to unpause it. Especially for users who just leave the notifications undismissed for long periods

src/vs/workbench/contrib/browserView/electron-browser/overlayManager.ts
```diff
@@ -96,29 +115,45 @@ export class BrowserOverlayManager extends Disposable implements IBrowserOverlay
...
+	 * Since we have multiple MutationObservers, we may get changes to multiple elements at once.
+	 * In order to minimize layout recalculations and event handling, we batch change events
+	 * using a microtask (so that event effects still happen before the next render).
+	 */
+	private scheduleOverlayStateChange(): void {
+		if (this._overlayStateChangePending) {
+			return;
+		}
+		this._overlayStateChangePending = true;
+		queueMicrotask(() => {
```
jruales: From what I understand, mutationObserver callbacks are also microtasks, so I'm guessing the idea is that with this approach, the idea is to batch all the mutationObserver callbacks that are currently in the microtask queue? I think there's no guarantee that there won't be more than one call before a given rendering, in case more mutationObserver microtasks are added to the microtask queue after this is added there

jruales (dismissed):

src/vs/workbench/contrib/browserView/electron-browser/overlayManager.ts
kycutler: I think you're right that it's possible, though what I've seen from existing scenarios is that the observer callbacks are all scheduled before being called. I wouldn't want to delay the event any further than a microtask to make sure that listeners have a chance to update the DOM before render and avoid flickering. But it's a good callout -- we can keep an eye on it.

Also, I discovered we have a `MicrotaskEmitter` class, so switched to use that!

src/vs/workbench/contrib/browserView/electron-browser/browserEditor.ts
kycutler: I agree, but naming is difficult 🙃 will leave as-is for now.

dmitrivMS (approved):

jruales:
GitHub Issue: https://github.com/microsoft/vscode/issues/289423

PR #289251, opened 2026-01-20
'Agent Status' widget tweaks

joshspicer:


justschen (approved):

bpasero:
@joshspicer curious why this requires changes to the class on the workbench `body` 🤔 ? whats the dependency on the `body` element, can it not be scoped to the use?

PR #289275, opened 2026-01-21
Updates to prompt file location settings

pwang347:
image

image


src/vs/workbench/contrib/chat/common/promptSyntax/utils/promptFilesLocator.ts
```diff
@@ -273,8 +269,8 @@ export class PromptFilesLocator {
 		if (type === PromptsType.agent) {
```
aeschli, line 269: Can this be removed? It's part of the settings defaults, right?

aeschli (approved):

PR #289289, opened 2026-01-21
Increase stack trace limit to 100 for improving error telemetry

bryanchen-d:
We need to capture more frames to better understand the triggering routes of unhandled errors. 
Also 100 frames is consistent from with we set in bootstrap-node.ts and extensionHostMain.ts

### DIFF
--- a/src/vs/code/browser/workbench/workbench.ts
+++ b/src/vs/code/browser/workbench/workbench.ts
@@ -3,9 +3,6 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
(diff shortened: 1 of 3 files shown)


bpasero (changes requested):
This is not a good location to put this in. If you tell me which processes should have this, we can find a better entry file.

It seems unsupported on Firefox, not sure what it means to set it: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/stackTraceLimit

deepak1556:
Yes this is a V8 only thing, should scope it to desktop and only chrome based browsers.

bryanchen-d:
@jrieken, @deepak1556 , @bpasero , moving it into registerErrorHandler works for me, any other comments before we can merge it?

bpasero (approved):

dbaeumer (approved):

PR #289290, opened 2026-01-21
Add optional chatSessionResource parameter for read operations in ManageTodoListTool

bhavyaus:


src/vs/workbench/contrib/chat/common/tools/builtinTools/manageTodoListTool.ts
```diff
@@ -80,6 +80,8 @@ interface IManageTodoListToolInputParams {
 		title: string;
 		status: 'not-started' | 'in-progress' | 'completed';
 	}>;
+	// used for todo read only
+	chatSessionResource?: string;
```
roblourens, line 84: Can't it be a URI?

DonJayamanne (approved):

src/vs/workbench/contrib/chat/common/tools/builtinTools/manageTodoListTool.ts
bhavyaus: No. This is coming from an non llm tool call and I can't pass the sessionResource string back from the chat extension as a URI

PR #289339, opened 2026-01-21
Agent Sessions - set toolbar width based on the number of items

lszomoru:


bpasero (changes requested):
Unresolved copilot comments.

PR #289364, opened 2026-01-21
fix: making the line height multiplier at least as big as the font size multiplier

### LINKED ISSUE #288844: Configuring just the fontSize pushes the text off center
Configure:
```
  "textMateRules": [
    {
      "scope": "entity.name.type.class",
      "settings": {
        "fontSize": 3
      }
    }
  ]
}
```
:bug: the regular sized text is no longer aligned with the line which causes decorations to appear off vertically

Image

aiday-mar:
fixes https://github.com/microsoft/vscode/issues/288844


alexdima (approved):
Nice! I made a small tweak to compute it only when it is not explicitly set

DonJayamanne (approved):

PR #289411, opened 2026-01-21
Copy codicons file as part of the build

alexr00:
Part of #283823


src/vs/base/browser/ui/codicons/codicon/codicon.ttf
alexr00, line 1: @hediet removal from disk is here.

hediet (approved):

PR #243283, opened 2025-03-12
Fix line comment action for makefiles (Fixes #234464)

Enzo-Nunes:
This pull request addresses an old bug (issue #234464) where the line comment action in VS Code failed to correctly handle makefiles. By default, VS Code uses `Ctrl + /` to comment lines, but for makefiles, comment tokens must be placed in the first column to avoid being just forwarded to the shell.

The issue happened because the line comment action didn't account for the language being edited. This fix passes the `languageId` to the comment function, allowing it to handle makefiles differently. A new test has been added to verify this functionality.

To test this in practice, open any makefile and try using the keybind `Ctrl + /` on an indented line of code, with and without my fix. You will notice the different placement of the comment token.

PR #289457, opened 2026-01-21
vscode-dts: Add LineCommentConfig interface & update lineComment

hkleungai:
Essentially a followup on a previous PR https://github.com/microsoft/vscode/pull/243283.

### DIFF
--- a/src/vscode-dts/vscode.d.ts
+++ b/src/vscode-dts/vscode.d.ts
@@ -6478,6 +6478,21 @@ declare module 'vscode' {
 	export type CharacterPair = [string, string];
 
 	/**
+	 * Configuration for line comments.
+	 */
+	export interface LineCommentConfig {
+		/**
+		 * The line comment token, like `//`
+		 */
+		comment: string;
+		/**
+		 * Whether the comment token should not be indented and placed at the first column.
+		 * Defaults to false.
+		 */
+		noIndent?: boolean;
+	}
+
+	/**
 	 * Describes how comments for a language work.
 	 */
 	export interface CommentRule {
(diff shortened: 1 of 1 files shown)


aiday-mar (dismissed):

aiday-mar (dismissed):

src/vscode-dts/vscode.d.ts
```diff
@@ -6485,7 +6500,7 @@ declare module 'vscode' {
 		/**
 		 * The line comment token, like `// this is a comment`
 		 */
-		lineComment?: string;
+		lineComment?: string | LineCommentRule | null;
```
mjbvz: Do we need the `| null`? Is this different than undefined?

aiday-mar: The language configuration file at position:

`https://github.com/microsoft/vscode/pull/243283/changes#diff-ce621bda5184034ea4b8741484a8bc49444cd7c3181843395e2917adfd187e44R32-R34`

seems to specify that the comment can be null too, so I guess the change has been made to unify the two

alexr00: Is `null` treated any differently from `undefined` in this case?

aiday-mar: I believe it is not treated differently. It was already there prior to the change that added the LineCommentRule, so I believe it would be safe to remove it.

mjbvz (approved):

alexr00 (approved):

src/vscode-dts/vscode.d.ts
```diff
@@ -6477,6 +6477,21 @@ declare module 'vscode' {
...
+	/**
+	 * Configuration for line comments.
+	 */
+	export interface LineCommentRule {
+		/**
+		 * The line comment token, like `//`
+		 */
+		comment: string;
+		/**
+		 * Whether the comment token should not be indented and placed at the first column.
```
gjsjohnmurray, line 6489:
```suggestion
		 * Whether the comment token should not be indented and instead placed at the first column.
```

PR #289481, opened 2026-01-21
feat: Xonsh shell integration add support for 'run recent command' with Xonsh

wajdiJomaa:
As suggested in #284044 https://github.com/microsoft/vscode/pull/284044#issuecomment-3728953635:

the main goal of this pull request is to integrate Xonsh shell in vscode in order to get a list of recent commands for current 

and previous session when running 'run recent command'

Screenshot from 2026-01-21 22-28-10


src/vs/platform/terminal/node/terminalEnvironment.ts
```diff
@@ -344,6 +356,7 @@ shellIntegrationArgs.set(ShellIntegrationExecutable.ZshLogin, ['-il']);
 shellIntegrationArgs.set(ShellIntegrationExecutable.Bash, ['--init-file', '{0}/out/vs/workbench/contrib/terminal/common/scripts/shellIntegration-bash.sh']);
 shellIntegrationArgs.set(ShellIntegrationExecutable.Fish, ['--init-command', 'source "{0}/out/vs/workbench/contrib/terminal/common/scripts/shellIntegration.fish"']);
 shellIntegrationArgs.set(ShellIntegrationExecutable.FishLogin, ['-l', '--init-command', 'source "{0}/out/vs/workbench/contrib/terminal/common/scripts/shellIntegration.fish"']);
+shellIntegrationArgs.set(ShellIntegrationExecutable.Xonsh, ['-i', '--rc', '{0}/out/vs/workbench/contrib/terminal/common/scripts/shellIntegration-xonsh.xsh']);
```
anthonykim1, line 357: Should we handle the login variant as well?

src/vs/workbench/contrib/terminal/common/scripts/shellIntegration-xonsh.xsh
```diff
@@ -0,0 +1,118 @@
+# ---------------------------------------------------------------------------------------------
+#   Copyright (c) Microsoft Corporation. All rights reserved.
+#   Licensed under the MIT License. See License.txt in the project root for license information.
+# ---------------------------------------------------------------------------------------------
+import sys
+import os
+
+class _vsCodeXonsh:
```
anthonykim1: Maybe something like _VSCodeXonshShellIntegration would be nicer.

anthonykim1: Maybe this is bigger than the intended scope of this PR, but If we're adding shell integration script, I think it may be best for us to add cwd sequence like in https://github.com/microsoft/vscode/blob/77c18b8431d9f97675e98aa8eaeaacb12181a900/src/vs/workbench/contrib/terminal/common/scripts/shellIntegration-rc.zsh#L163 I believe these are linked to things like file links, cwd extension apis, split terminal, etc..

src/vs/platform/terminal/node/terminalEnvironment.ts
wajdiJomaa: I added support for xonsh login variant

src/vs/workbench/contrib/terminal/common/scripts/shellIntegration-xonsh.xsh
wajdiJomaa: done

wajdiJomaa: added support for cwd sequence on chdir

anki-code:
Just want to add some graphical concepts for news

image image

anki-code:
Any news here?

flavius:
Why is the communication on this so terrible? @anthonykim1 - ownership?

anthonykim1:
@wajdiJomaa Thanks for waiting. Generally looks good to me. 

Manually testing however, I'm still seeing "python" as shell type and history from python for run recent command on my mac along with no decorations 🧐

Screenshot 2026-07-15 at 10 23 42 PM
Screenshot 2026-07-15 at 10 23 35 PM

I'd check how we handled this sort of issue for python, or other shells.

Also would recommend taking a look at some of the copilot comments.

PR #289496, opened 2026-01-21
add setting for controlling terminal chat output expansion

meganrogge:
fix #287588

New setting: `chat.terminalOutput.expansion` with 3 modes:

- `onError (default)` - Expands terminal output when command starts, collapses on success
- `always` - Always keeps terminal output expanded
- `never` - Never auto-expands terminal output

### DIFF
--- a/src/vs/workbench/contrib/chat/browser/chat.contribution.ts
+++ b/src/vs/workbench/contrib/chat/browser/chat.contribution.ts
@@ -433,6 +433,17 @@ configurationRegistry.registerConfiguration({
 			description: nls.localize('chat.checkpoints.showFileChanges', "Controls whether to show chat checkpoint file changes."),
 			default: false
 		},
+		[ChatConfiguration.TerminalOutputExpansion]: {
+			type: 'string',
+			enum: ['onError', 'always', 'never'],
+			enumDescriptions: [
+				nls.localize('chat.terminalOutput.expansion.onError', "Expand terminal output when output is received and collapse on success (default)."),
+				nls.localize('chat.terminalOutput.expansion.always', "Always keep terminal output expanded."),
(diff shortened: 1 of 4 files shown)


src/vs/workbench/contrib/chat/browser/chat.contribution.ts
```diff
@@ -433,6 +433,17 @@ configurationRegistry.registerConfiguration({
 			description: nls.localize('chat.checkpoints.showFileChanges', "Controls whether to show chat checkpoint file changes."),
 			default: false
 		},
+		[ChatConfiguration.TerminalOutputExpansion]: {
```
Tyriar, line 436: Should live with the other terminal tool settings? It's fine to export this from the contrib exports file (ideally the `chatTerminalToolProgressPart.ts` would live in `terminalContrib/chatAgentTools`

src/vs/workbench/contrib/chat/browser/chat.contribution.ts
```diff
@@ -433,6 +433,17 @@ configurationRegistry.registerConfiguration({
...
+			type: 'string',
+			enum: ['onError', 'always', 'never'],
+			enumDescriptions: [
+				nls.localize('chat.terminalOutput.expansion.onError', "Expand terminal output when output is received and collapse on success (default)."),
+				nls.localize('chat.terminalOutput.expansion.always', "Always keep terminal output expanded."),
+				nls.localize('chat.terminalOutput.expansion.never', "Never auto-expand terminal output.")
+			],
```
Tyriar, line 443: This is mixing 2 concepts, whether to auto expand and whether to auto collapse. Currently it doesn't capture expand automatically and collapse regardless of error.

Tyriar (changes requested):

PR #289531, opened 2026-01-22
chat: handle internal model IDs and group-scoped cache

### LINKED ISSUE #289529: Chat model selection/caching breaks for grouped providers (internal IDs + eviction)
### Problem
Chat language model selection and subsequent use can break for grouped providers (internal identifiers like `vendor/group/id`, e.g. `azure/Azure/gpt-5`). Symptoms include:
- `selectLanguageModels({ id: "vendor/group/id" })` returning `[]` (model not found), and/or
- `LanguageModelError.NotFound` during `countTokens` when the model was previously selectable.

### Root cause
Two related issues:
1) **Selection mismatch**: `LanguageModelsService.selectLanguageModels` compared `selector.id` only to `metadata.id`, but some call paths pass the full internal identifier in `selector.id`.
2) **Over-aggressive cache eviction in ext host**: resolving models for a vendor happens in both ungrouped and per-group passes. If an ungrouped resolve returns `[]`, vendor-wide eviction could remove grouped identifiers (e.g. `azure/Azure/gpt-5`), leading to later `NotFound`.

### Fix
PR #289531:
- Allow `selector.id` to match either `metadata.id` **or** the full internal identifier.
- In the extension host, evict cached models **only** for the specific `(vendor, group)` being refreshed (no vendor-wide clears).
- Add test coverage for the `selector.id` internal-identifier match.

Fixes #289529.

zhichli:
Fixes chat language model selection and caching for grouped providers.

- Allow `selector.id` to match the full internal model identifier (e.g. `vendor/group/id`).
- Avoid vendor-wide cache eviction in the extension host when resolving models; only evict stale entries for the specific `(vendor, group)` being refreshed (prevents grouped models being cleared by empty ungrouped resolves).
- Add test coverage for `selector.id` matching the internal identifier.

Fixes #289529.

### DIFF
--- a/src/vs/workbench/api/common/extHostLanguageModels.ts
+++ b/src/vs/workbench/api/common/extHostLanguageModels.ts
@@ -238,8 +238,6 @@ export class ExtHostLanguageModels implements ExtHostLanguageModelsShape {
 			};
 		});
 
-		this._logService.info('[LM][extHost] _localModels (bc)', { reason: '$provideLanguageModelChatInfo', vendor: vendor, keysSample: Array.from(this._localModels.keys()).slice(0, 200) });
-
 		const groupKey = options.group ?? '';
 		let identifiersByGroup = this._localModelIdentifiersByVendorAndGroup.get(vendor);
 		if (!identifiersByGroup) {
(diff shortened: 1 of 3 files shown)


roblourens (dismissed):

sandy081:
@zhichli Your analysis makes complete sense. 

The main root cause is the second point you mentioned where ext host side is clearing the cache - it should clear the cache only by group. This should fix the problem. Ideally no one should go through the first route - asking language model by internal identifier - this never seems to work - even before the internal identifier is different from model id. 

Thanks for the PR. I will change the cache on the ext host side to make it group aware. I will also fix selecting language model by internal identifier differently. I will create another PR with these changes. Thanks.

PR #289548, opened 2026-01-22
Add inlineChat.defaultModel setting

eleanorjboyd:
These changes replace the inlineChat.persistModelChoice configuration with a new inlineChat.defaultModel setting for inline chat. if no match is found it falls back to vendor default (with an error logged)

The new setting accepts either:
- A model's qualified name (e.g., "GPT-4o (copilot)")
- The model name for Copilot models (e.g., "GPT-4o")


src/vs/workbench/contrib/inlineChat/browser/inlineChatController.ts
```diff
@@ -473,25 +471,12 @@ export class InlineChatController implements IEditorContribution {
...
-				const candidate = this._languageModelService.lookupLanguageModel(identifier);
-				if (candidate?.isDefaultForLocation[session.chatModel.initialLocation]) {
-					this._zone.value.widget.chatWidget.input.setCurrentLanguageModel({ metadata: candidate, identifier });
-					break;
-				}
-			}
+		// Check for default model setting
+		const defaultModelSetting = this._configurationService.getValue<string>(InlineChatConfigKeys.DefaultModel);
+		if (defaultModelSetting && !this._zone.value.widget.chatWidget.input.switchModelByQualifiedName(defaultModelSetting)) {
+			onUnexpectedError(new Error(`inlineChat.defaultModel setting value '${defaultModelSetting}' did not match any available model. Falling back to vendor default.`));
```
jrieken: Let's use the log service and simply warn. Using `onUnexpectedError` makes this go to error telemetry and it's not something we'll be able to act on (since it is user config)

jrieken (commented):
Looking, good. Left a comment about changing the model-not-found case. 

Also, we can go the extra mile and define the setting as enum which we update as models get registered. That makes configuration easy for the user and gives validation. I do this for the `editor.defaultFormatter` setting, this is how/where it is done: https://github.com/microsoft/vscode/blob/d6a69b96a2f6371621604115130dd8dc7236793a/src/vs/workbench/contrib/format/browser/formatActionsMultiple.ts#L285 and https://github.com/microsoft/vscode/blob/d6a69b96a2f6371621604115130dd8dc7236793a/src/vs/workbench/contrib/format/browser/formatActionsMultiple.ts#L129

src/vs/workbench/contrib/inlineChat/browser/inlineChatController.ts
```diff
@@ -473,25 +473,12 @@ export class InlineChatController implements IEditorContribution {
...
-					break;
-				}
-			}
+		// Check for default model setting
+		const defaultModelSetting = this._configurationService.getValue<string>(InlineChatConfigKeys.DefaultModel);
+		if (defaultModelSetting && !this._zone.value.widget.chatWidget.input.switchModelByQualifiedName(defaultModelSetting)) {
+			this._logService.warn(`inlineChat.defaultModel setting value '${defaultModelSetting}' did not match any available model. Falling back to vendor default.`);
 		}
 
-		store.add(this._zone.value.widget.chatWidget.input.onDidChangeCurrentLanguageModel(newModel => {
```
karthiknadig: It seems like `store` (`DisposableStore`) is now unused?

karthiknadig (approved):

PR #289564, opened 2026-01-22
Support reading files outside of the workspace

sbatten:
Introduce functionality to allow users to read files and list directories outside the workspace. This includes a confirmation contribution that enables users to approve access to external paths, with an option to allow all access from a containing folder for the current chat session. The implementation includes necessary updates to handle chat session resources.

### DIFF
--- a/src/vs/workbench/contrib/chat/common/tools/builtinTools/chatExternalPathConfirmation.ts
+++ b/src/vs/workbench/contrib/chat/common/tools/builtinTools/chatExternalPathConfirmation.ts
@@ -1,0 +1,114 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { dirname } from '../../../../../../base/common/path.js';
+import { extUriBiasedIgnorePathCase } from '../../../../../../base/common/resources.js';
+import { URI } from '../../../../../../base/common/uri.js';
+import { localize } from '../../../../../../nls.js';
+import { ConfirmedReason, ToolConfirmKind } from '../../chatService/chatService.js';
+import {
+	ILanguageModelToolConfirmationActions,
+	ILanguageModelToolConfirmationContribution,
+	ILanguageModelToolConfirmationRef
+} from '../languageModelToolsConfirmationService.js';
+
+export interface IExternalPathInfo {
+	path: string;
+	isDirectory: boolean;
+}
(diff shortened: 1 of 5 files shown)


src/vs/workbench/contrib/chat/common/tools/builtinTools/chatExternalPathConfirmation.ts
```diff
@@ -0,0 +1,114 @@
...
+/**
+ * Confirmation contribution for read_file and list_dir tools that allows users to approve
+ * accessing paths outside the workspace, with an option to allow all access
+ * from a containing folder for the current chat session.
+ */
+export class ChatExternalPathConfirmationContribution implements ILanguageModelToolConfirmationContribution {
+	readonly canUseDefaultApprovals = false;
+
+	// Map<sessionResourceString, Set<folderUriString>>
+	private readonly _sessionFolderAllowlist = new Map<string, Set<string>>();
```
connor4312: Should use a ResourceSet instead of Set<string> and ResourceMap instead of Map

src/vs/workbench/contrib/chat/common/tools/builtinTools/chatExternalPathConfirmation.ts
```diff
@@ -0,0 +1,114 @@
...
+		}
+
+		// For directories, use the path itself; for files, use the parent directory
+		const folderUri = pathInfo.isDirectory ? pathUri : pathUri.with({ path: dirname(pathUri.path) });
+		const sessionKey = ref.chatSessionResource.toString();
+
+		return [
+			{
+				label: localize('allowFolderSession', 'Allow this folder in this session'),
+				detail: localize('allowFolderSessionDetail', 'Allow reading files from this folder without further confirmation in this chat session'),
```
connor4312, line 86: I would just have this as "Allow reading from {0}"

connor4312 (changes requested):
quick comments, I would also add a note in the impl in copilot-chat that these parameters are read by core too

src/vs/workbench/contrib/chat/common/tools/builtinTools/chatExternalPathConfirmation.ts
sbatten: You mean forever?

connor4312: Maybe the primary action button is "Allow Reads in {0}" for the session, and secondary actions to "Allow this folder in this workspace" or "Allow this folder for all workspaces"

connor4312 (approved):

PR #289565, opened 2026-01-22
Add setting to open localhost links from Terminal, chat, etc. in Integrated Browser

jruales:
Adds a setting that, when enabled, will open all localhost links (from Terminal, chat, etc.) in the Integrated Browser instead of the system browser. Links to non-localhost URLs are not affected.

Also adds telemetry event `integratedBrowser.open` that logs browser editor (aka browser tab) opening and the source of the opening.

---

Related feature: https://github.com/microsoft/vscode/issues/286579


src/vs/workbench/contrib/browserView/electron-browser/browserView.contribution.ts
```diff
@@ -66,6 +75,8 @@ class BrowserEditorResolverContribution implements IWorkbenchContribution {
 						throw new Error(`Invalid browser view resource: ${resource.toString()}`);
 					}
 
+					logBrowserOpen(telemetryService, 'editorResolver');
```
kycutler: Does this result in multiple open logs if we also log before calling `editorService.openEditor({ resource })`?

jruales: True. Removed `'editorResolver'` logging

kycutler (approved):

PR #289605, opened 2026-01-22
Fix the TextModel interface to match the implementation wrt applyEdits

### LINKED ISSUE #289604: ITextModel interfact does not match TextModel implementation
In common/model.ts, in the `ITextModel` interface `applyEdits` is defined as follows:

```ts
	/**
	 * Edit the model without adding the edits to the undo stack.
	 * This can have dire consequences on the undo stack! See @pushEditOperations for the preferred way.
	 * @param operations The edit operations.
	 * @return If desired, the inverse edit operations, that, when applied, will bring the model back to the previous state.
	 */
	applyEdits(operations: readonly IIdentifiedSingleEditOperation[]): void;
	/** @internal */
	applyEdits(operations: readonly IIdentifiedSingleEditOperation[], reason: TextModelEditSource): void;
	applyEdits(operations: readonly IIdentifiedSingleEditOperation[], computeUndoEdits: false): void;
	applyEdits(operations: readonly IIdentifiedSingleEditOperation[], computeUndoEdits: true): IValidEditOperation[];
```

In common/model/textModel.ts, `class TextModel implements model.ITextModel` implements the following methods:

```ts
	public applyEdits(operations: readonly model.IIdentifiedSingleEditOperation[]): void;
	public applyEdits(operations: readonly model.IIdentifiedSingleEditOperation[], computeUndoEdits: false): void;
	public applyEdits(operations: readonly model.IIdentifiedSingleEditOperation[], computeUndoEdits: true): model.IValidEditOperation[];
	/** @internal */
	public applyEdits(operations: readonly model.IIdentifiedSingleEditOperation[], computeUndoEdits: false, reason: TextModelEditSource): void;
	/** @internal */
	public applyEdits(operations: readonly model.IIdentifiedSingleEditOperation[], computeUndoEdits: true, reason: TextModelEditSource): model.IValidEditOperation[];
	public applyEdits(rawOperations: readonly model.IIdentifiedSingleEditOperation[], computeUndoEdits?: boolean, reason?: TextModelEditSource): void | model.IValidEditOperation[] {
		// ...
	}
```

Significantly, in the implementation, the second argument is only every boolean|undefined, and reason is always the third argument.

In the interface it looks like you should be providing either `computeUndoEdits` or `reason` as the second argument.

I'm not 100% sure how this compiles. But I can't see any usages of of `applyEdits` that is called on the interface that passes in a reason. (`applyEdits` is called on the implementation in editStack.ts which means it is correctly typed there and it passed in the reason as the third parameter.)

My proposal would be to update the interface in this situation to match what the implementation is actually doing since nothing yet calls the interface with a reason.

I'll send a PR shortly for that.

(Note the overloading is also a bit more verbose than it needs to be as the existence of reason doesn't change the return type so it can just be optional, rather than requiring overloads.)

a-stewart:
This fixes #289604

### DIFF
--- a/src/vs/editor/common/model.ts
+++ b/src/vs/editor/common/model.ts
@@ -1241,11 +1241,8 @@ export interface ITextModel {
 	 * @param operations The edit operations.
 	 * @return If desired, the inverse edit operations, that, when applied, will bring the model back to the previous state.
 	 */
-	applyEdits(operations: readonly IIdentifiedSingleEditOperation[]): void;
-	/** @internal */
-	applyEdits(operations: readonly IIdentifiedSingleEditOperation[], reason: TextModelEditSource): void;
-	applyEdits(operations: readonly IIdentifiedSingleEditOperation[], computeUndoEdits: false): void;
-	applyEdits(operations: readonly IIdentifiedSingleEditOperation[], computeUndoEdits: true): IValidEditOperation[];
+	applyEdits(operations: readonly IIdentifiedSingleEditOperation[], computeUndoEdits?: false, reason?: TextModelEditSource): void;
+	applyEdits(operations: readonly IIdentifiedSingleEditOperation[], computeUndoEdits: true, reason?: TextModelEditSource): IValidEditOperation[];
 
 	/**
 	 * Change the end of line sequence without recording in the undo stack.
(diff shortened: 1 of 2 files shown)


aiday-mar (changes requested):
Hi thanks for making this PR. When I install dependencies and run watch on this PR, I get the following error in the console

[watch-client    ] [17:53:31] Error: /Users/aidaymarlenkyzy/Desktop/vscode/src/vs/monaco.d.ts(2368,102): Cannot find name 'TextModelEditSource'.

This is because TextModelEditSource does not exist in the monaco.d.ts file. To bypass this, the current implementation uses the comment before the function signature:

/** @internal */

Your PR removes this comment which we need. Could you please fix this?

PR #289721, opened 2026-01-22
Add sku label in the copilot dashboard

cwebster-99:
Adds the readable sku name to the Copilot Dashboard title 

image


bpasero (approved):
thanks!

PR #289731, opened 2026-01-22
new settings: includeReferencedInstructions & includeApplyingInstructions

### LINKED ISSUE #282422: add a setting to control whether `*.instructions` are added or not
Allow users to disable '*.instructions' with one setting

aeschli:
Fixes https://github.com/microsoft/vscode/issues/282422


src/vs/workbench/contrib/chat/common/promptSyntax/utils/promptFilesLocator.ts
```diff
@@ -543,7 +543,28 @@ export class PromptFilesLocator {
 
 	private async findAgentSkillsInFolder(uri: URI, token: CancellationToken): Promise<URI[]> {
 		try {
-			return await this.searchFilesInLocation(uri, `*/${SKILL_FILENAME}`, token);
```
pwang347, line 546: @aeschli any reason why this change is needed? if it's for recursive discovery, should it just use **/?

aeschli: You are quick :-)
`searchFilesInLocation` does currently not work for file systems that don't have a search service provided.
Also, as we know that we only have to go two level deep I think its more efficient to use the file service. The search service would first have to understand that the pattern (`*/SKILL.md`) only can match two level deep,. I think its not that smart and will look for all files in the folder and then only do the pattern match (not 100% sure, just a guess)

pwang347: I think @digitarald mentioned that we'll need to support recursive. But if search support is limited it makes sense, I suppose another reason to deprecate the glob.

pwang347 (dismissed):

pwang347 (approved):

PR #289747, opened 2026-01-22
hovers: allow nested hovers

connor4312:
Feels a bit bizarre but it has a good use case in the agents sessions view where the hover is more interactive than usual. Previously, hovering any hoverable element within the hover would close the hover. Now, hovering a hoverable element within a hover will open a nested hover by default (capped at 3 levels.) It seems to work well.

Closes https://github.com/microsoft/vscode/issues/289954

https://github.com/user-attachments/assets/dfa86430-5f64-4264-9526-1955c7f8d93c


src/vs/platform/hover/browser/hoverService.ts
```diff
@@ -290,30 +371,110 @@ export class HoverService extends Disposable implements IHoverService {
...
+
+		// When nesting, lock all parent hovers to prevent them from closing
+		if (isNesting) {
+			for (let i = 0; i <= containingHoverIndex; i++) {
+				this._hoverStack[i].hover.isLocked = true;
+			}
+		}
+
+		// Create a new ContextView for this hover with higher z-index for nested hovers
+		const container = options.container ?? this._layoutService.activeContainer;
```
benibenj: Don't think we should use `this._layoutService.activeContainer` instead `this._layoutService.getContainer(getWindow(targetElement))`

Tyriar (commented):
@connor4312 this makes things quite a bit more complicated and I worry about all the edge cases (not hiding hovers at the right time, locking acting weird, etc.). Do you understand this fully and are willing to co-maintain the hover going forward? The alternative is to force the native hover when the hover target is itself within a hover.

Also no issue to verify?

zhichli (approved):

PR #289776, opened 2026-01-22
enable Agent Status by default

joshspicer:
ref https://github.com/microsoft/vscode/pull/289776


rzhao271 (approved):

src/vs/workbench/contrib/chat/browser/chat.contribution.ts
```diff
@@ -188,7 +188,7 @@ configurationRegistry.registerConfiguration({
 		'chat.commandCenter.enabled': {
 			type: 'boolean',
 			markdownDescription: nls.localize('chat.commandCenter.enabled', "Controls whether the command center shows a menu for actions to control chat (requires {0}).", '`#window.commandCenter#`'),
-			default: true
+			default: false
```
rzhao271, line 191: Sanity check: is this part of the PR intentional?

joshspicer: Thank you, yes. With both on it looks like this:

image

PR #289816, opened 2026-01-23
Don't expand running mcp tool calls

roblourens:
Copilot randomly decided that they should work this way when implementing lazy loading


justschen (approved):
thanks thanks!

PR #289863, opened 2026-01-23
fix: memory leak in abstract task service

SimonSiefke:
Fixes a memory leak in `ProblemReporter` and `AbstractTaskService`.

### Details

The `Emitter` in `ProblemReporter` currently doesn't seem be registered as a Disposable:

```ts
class ProblemReporter implements TaskConfig.IProblemReporter {

	private _validationStatus: ValidationStatus;
	private readonly _onDidError: Emitter<string> = new Emitter<string>(); // not registered
	public readonly onDidError: Event<string> = this._onDidError.event;
```

Therefore one change is registering the Emitter as a Disposable.

---

Another issue seems to be when calling `AbstractTaskService._computeWorkspaceFolderTasks` multiple times and using `this._register`:

```ts
private async _computeWorkspaceFolderTasks(workspaceFolder: IWorkspaceFolder, runSource: TaskRunSource = TaskRunSource.User): Promise<IWorkspaceFolderTaskResult> {
	/* ... */
	const problemReporter = new ProblemReporter(this._outputChannel);
	this._register(problemReporter.onDidError(error => this._showOutput(runSource, undefined, error)));
```

Since there is only one `AbstractTaskService`, registering a disposable each time when calling `_computeWorkspaceFolderTasks` seems make the number of registered disposables grow each time. 

By changing it to a `DisposableMap` the `DisposableMap` ensures that there is only one problemsReporter disposableStore per workspaceUri. And calling `this._workspaceTaskDisposables.set(workspaceFolder.uri, store)` automatically disposes the previous registered disposableStore for that workspace uri.

```ts
private async _computeWorkspaceFolderTasks(workspaceFolder: IWorkspaceFolder, runSource: TaskRunSource = TaskRunSource.User): Promise<IWorkspaceFolderTaskResult> {
	/* ... */
	this._workspaceTaskDisposables.set(workspaceFolder.uri, store);
	const problemReporter = store.add(new ProblemReporter(this._outputChannel));
	store.add(problemReporter.onDidError(error => this._showOutput(runSource, undefined, error)));

```

### Before

When configuring a task 37 times, the number of various functions in `AbstractTaskService._computeWorkspaceFolderTasks` and some other places seems to grow each time:

Untitled

### After

When configuring a task 37 times, while the number of some functions still grows, for `AbstractTaskService._computeWorkspaceFolderTasks` it stays constant now:

task-configure


meganrogge (approved):
Thank you!

jrieken (approved):

PR #289866, opened 2026-01-23
chore: add npm version check to preinstall script

Himanshu25kapure:
### Summary
Adds an explicit npm version validation to the existing preinstall checks.

### Details
- Reuses `build/npm/preinstall.ts` as suggested by maintainers
- Validates npm major version (>= 9)
- Assumes npm bundled with the supported Node.js version

This improves early feedback when unsupported npm versions are used.

> Note: Local pre-commit hook was skipped due to missing native deps on Windows; CI should cover validation.
Fixes #252372


dmitrivMS:
We already have version validation.

PR #289956, opened 2026-01-23
Improve any key detection

### LINKED ISSUE #286050: Terminal tool doesn't detect "press any key"
```
sleep 1
echo "Press any key to continue..."
read -n 1 -s key
echo "You pressed: $key"
echo "Continuing..."
```

It doesn't detect that the script is waiting on the "press any key" prompt, because the 4o-mini prompt has "any key" examples that should not be detected and I guess it picks up on those.

Image

[db330cf9.copilotmd](https://github.com/user-attachments/files/24444976/db330cf9.copilotmd)

I still don't understand why we need an LLM to detect `Terminal will be reused by tasks, press any key to close it.` since that message comes from vscode, and also don't we know that the task is finished at that point?

meganrogge:
fixes #286050

VS Code task finish messages ("Press any key to close the terminal") were being blocked, which was correct. But, this blocking was too broad and also prevented legitimate script "Press any key to continue..." prompts from being detected and handled

Now, we distinguish between the two cases:
- VS Code task finish messages: Only ignored when it's a task AND the task is finished (more specific check using `detectsVSCodeTaskFinishMessage`
- Script "press any key" prompts: Now properly detected with `detectsGenericPressAnyKeyPattern` and presented to the user as a free-form input prompt

### DIFF
--- a/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/monitoring/outputMonitor.ts
+++ b/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/monitoring/outputMonitor.ts
@@ -210,7 +210,7 @@ export class OutputMonitor extends Disposable implements IOutputMonitor {
 			}
 			this._cleanupIdleInputListener();
 			this._outputMonitorTelemetryCounters.inputToolFreeFormInputShownCount++;
-			const lastLine = output.trimEnd().split('\n').pop() || '';
+			const lastLine = output.trimEnd().split(/\r?\n/).pop() || '';
 			const receivedTerminalInput = await this._requestFreeFormTerminalInput(token, this._execution, {
 				prompt: lastLine,
 				options: [],
--- a/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/monitoring/outputMonitor.ts
+++ b/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/monitoring/outputMonitor.ts
@@ -596,9 +596,9 @@ export class OutputMonitor extends Disposable implements IOutputMonitor {
 				resolve(value);
 			};
 			inputDataDisposable = this._register(execution.instance.onDidInputData((data) => {
-				// For "press any key" prompts, accept any input
+				// For "press any key" prompts, accept any non-empty input
 				// For other free-form inputs (like passwords), only accept on Enter
-				if (acceptAnyKey || data === '\r' || data === '\n' || data === '\r\n') {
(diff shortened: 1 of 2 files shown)


src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/monitoring/outputMonitor.ts
```diff
@@ -830,3 +859,34 @@ export function detectsNonInteractiveHelpPattern(cursorLine: string): boolean {
...
+
+/**
+ * Detects VS Code's specific task completion messages like:
+ * - "Press any key to close the terminal."
+ * - "Terminal will be reused by tasks, press any key to close it."
+ * These appear when a task finishes and should be ignored if the task is done.
+ * Note: These messages may be prefixed with " * " by VS Code and may have line wrapping
+ * that can split words across lines (e.g., "t\no" instead of "to").
+ */
+export function detectsVSCodeTaskFinishMessage(cursorLine: string): boolean {
```
roblourens, line 884: Isn't this localized? This seems like a complicated way to check for a string that came from the same product

roblourens: https://github.com/microsoft/vscode/blob/df77ff66158cf8fca2a9dc945ce661df1899c9ef/src/vs/workbench/contrib/tasks/browser/terminalTaskSystem.ts#L1975

meganrogge: there can be random spaces in it based on my testing with the terminal wrapping

meganrogge: But yes, good call on localizing

eleanorjboyd (approved):

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/monitoring/outputMonitor.ts
```diff
@@ -860,6 +860,19 @@ export function detectsNonInteractiveHelpPattern(cursorLine: string): boolean {
 	].some(e => e.test(cursorLine));
 }
 
+/**
+ * Localized task finish messages from VS Code's terminalTaskSystem.
+ * These are the same strings used when tasks complete.
+ */
+const taskFinishMessages = [
+	// "Terminal will be reused by tasks, press any key to close it."
+	localize('closeTerminal', "Terminal will be reused by tasks, press any key to close it."),
```
roblourens, line 869: But now you're localizing it separately, and there's no guarantee it will be localized the same as the other string in another file

PR #289979, opened 2026-01-23
allow jumping to diagnostics without showing peek widget

ulugbekna:
Usually, I would like to jump to the next error but I don't want the peek widget to show and immediately press Escape. This is especially true because one usually knows why there's an error already, and the specifc error message can also be triggerred manually on hover using `cmd+k cmd+i`. 

Showing the peek widget also interferes with NES -- I would love to be able to jump from an error to error quickly and fix things with NES. 

This PR adds a setting that allows to not show the peek widget when jumping to the next error -- the default remains as previously. 

The setting 

```
  "editor.markers.showPeekWidget": false,
```

### DIFF
--- a/src/vs/editor/contrib/gotoError/browser/gotoError.ts
+++ b/src/vs/editor/contrib/gotoError/browser/gotoError.ts
@@ -7,6 +7,17 @@ import { Codicon } from '../../../../base/common/codicons.js';
 import { KeyCode, KeyMod } from '../../../../base/common/keyCodes.js';
 import { DisposableStore } from '../../../../base/common/lifecycle.js';
 import { URI } from '../../../../base/common/uri.js';
+import * as nls from '../../../../nls.js';
+import { MenuId } from '../../../../platform/actions/common/actions.js';
+import { IConfigurationService } from '../../../../platform/configuration/common/configuration.js';
+import { ConfigurationScope, Extensions, IConfigurationRegistry } from '../../../../platform/configuration/common/configurationRegistry.js';
+import { IContextKey, IContextKeyService, RawContextKey } from '../../../../platform/contextkey/common/contextkey.js';
+import { TextEditorSelectionRevealType } from '../../../../platform/editor/common/editor.js';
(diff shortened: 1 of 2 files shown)


src/vs/editor/contrib/gotoError/browser/gotoError.ts
```diff
@@ -302,3 +337,15 @@ registerEditorCommand(new MarkerCommand({
 		secondary: [KeyMod.Shift | KeyCode.Escape]
 	}
 }));
+
+Registry.as<IConfigurationRegistry>(Extensions.Configuration).registerConfiguration({
+	id: 'editor',
+	properties: {
+		'editor.markers.showPeekWidget': {
```
jrieken, line 344: We don't use `markers` in user-facing language but usually say `problems`

src/vs/editor/contrib/gotoError/browser/gotoError.ts
```diff
@@ -146,28 +150,59 @@ export class MarkerController implements IEditorContribution {
...
+					}
+				} else {
+					// move cursor to marker in this editor
+					const pos = new Position(model.selected.marker.startLineNumber, model.selected.marker.startColumn);
+					this._editor.setPosition(pos);
+					this._editor.revealPositionInCenterIfOutsideViewport(pos);
+				}
+			} finally {
+				model.dispose();
+			}
```
jrieken, line 205: I am sure this can be done without such amounts of code duplication

src/vs/editor/contrib/gotoError/browser/gotoError.ts
```diff
@@ -302,3 +337,15 @@ registerEditorCommand(new MarkerCommand({
 		secondary: [KeyMod.Shift | KeyCode.Escape]
 	}
 }));
+
+Registry.as<IConfigurationRegistry>(Extensions.Configuration).registerConfiguration({
```
jrieken, line 341: This should also likely be an editor-setting, like one that's registered in editor land because this is an editor-contrib thing

jrieken (changes requested):

PR #289993, opened 2026-01-23
enhance model selection logic to use location based defaults and bring back session model section tracking

eleanorjboyd:
The prioritization of model selection for inline is as follows:

User changed model this session → Their choice is honored, neither setting nor vendor default overrides it

User hasn't changed model 
- inlineChat.defaultModel set & valid → use that model
- inlineChat.defaultModel set & invalid → fall back to vendor default for context (editor)
- inlineChat.defaultModel not set → use vendor default for context (editor)


mjbvz (dismissed):

jrieken (dismissed):
lgtm. also tested locally. let's ship it

benvillalobos (dismissed):

dmitrivMS (approved):

bpasero (approved):

PR #290000, opened 2026-01-23
Chat - add metadata to `ChatSessionItem`

lszomoru:


osortega (dismissed):

src/vscode-dts/vscode.proposed.chatSessionsProvider.d.ts
```diff
@@ -247,6 +247,11 @@ declare module 'vscode' {
 			 */
 			deletions: number;
 		};
+
+		/**
+		 * Arbitrary metadata for the chat session. Can be anything, but must be JSON-stringifyable.
+		 */
+		readonly metadata?: { readonly [key: string]: any };
```
mjbvz: This will need to be hooked up for the new session item controller api too. With that, the ChatSessionitem becomes a live object. This should be done in `ChatSessionItemImpl` in the `extHostChatSessions.ts` file

Also the signature should become:

```
/**
 * Arbitrary metadata for the chat session. Can be anything, but must be JSON-stringifyable.
 *
 * To update the metadata you must re-set this property
 */
metadata?: { readonly [key: string]: any };
```

mjbvz (changes requested):

src/vscode-dts/vscode.proposed.chatSessionsProvider.d.ts
```diff
@@ -247,6 +247,11 @@ declare module 'vscode' {
 			 */
 			deletions: number;
 		};
+
+		/**
+		 * Arbitrary metadata for the chat session. Can be anything, but must be JSON-stringifyable.
```
mjbvz, line 237: I think we should note here that you need to update the property for changes to be correctly reflected. Code such as:

```ts
item.metadata.repo = 'foo'
```

Won't actually update the value across all locations. Instead you need to write:

```ts
item.metadata =  { repo: 'foo' }
```

or 


```ts
item.metadata.repo = 'foo';
item.metadata = item.metadata; // Trigger explicitly update 
```

For the last case, we would also need to remove the `if (this.#metadata !== value) {` check. Either do a deep equality check instead or only skip the update if both the old value and new value are `undefined`

lszomoru: @mjbvz, is there any reason why we have picked this "pattern"? No matter which of the fields of the object you update the same event will be firing. Any reason why we could not be using the observable pattern making each field observable?

lszomoru: @mjbvz, do you mind adding the comment that you see fit in the documentation of the API?

mjbvz: This is the common pattern with controlled objects in our apis. You can see it in places such as `Webview.options` where the field is mutable but all properties on the object are readonly.  The only way to update it is to update the entire property

The alternative is using proxy objects so that VS Code can observe changes to the `metadata` object by extension, but this gets messy to implement especially for for complex, nested data types 


```suggestion
		 * Arbitrary metadata for the chat session. Can be anything, but must be JSON-stringifyable.
		 * 
		 * To update the metadata you must re-set this property.
```

lszomoru: @mjbvz, added the comment update, and switched to using a deep equal.

mjbvz (dismissed):

bpasero (approved):

PR #290046, opened 2026-01-23
Retry network operations on recoverable server errors

dmitrivMS:
Server occasionally hits errors like this in sanity tests:
2026-01-23T22:23:36.3778003Z [2026-01-23T22:23:36.377Z] ERROR: [Server Error] [22:23:36] #8: https://github.gallerycdn.vsassets.io/extensions/github/vscode-pull-request-github/0.127.2026012204/1769055663200/Microsoft.VisualStudio.Code.Manifest - error GET write EPROTO 206009D83EFA0000:error:0A000438:SSL routines:ssl3_read_bytes:tlsv1 alert internal error:../deps/openssl/openssl/ssl/record/rec_layer_s3.c:916:SSL alert number 80

These should be retryable, so adding code to do so.


src/vs/platform/request/node/requestService.ts
```diff
@@ -153,6 +155,31 @@ async function getNodeRequest(options: IRequestOptions): Promise<IRawRequestFunc
...
+				throw error;
+			}
+
+			const isTransientError = TransientErrorRegex.test(getErrorMessage(error));
+			if (!isTransientError || attempt === maxRetries) {
+				throw error;
+			}
+
+			await new Promise(resolve => setTimeout(resolve, 100 * attempt));
+		}
```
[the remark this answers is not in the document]
dmitrivMS: @copilot open a new pull request to apply changes based on [this feedback](https://github.com/microsoft/vscode/pull/290046#discussion_r2723187859)

src/vs/platform/request/node/requestService.ts
```diff
@@ -21,6 +21,8 @@ import { AbstractRequestService, AuthInfo, Credentials, IRequestService, systemC
 import { Agent, getProxyAgent } from './proxy.js';
 import { createGunzip } from 'zlib';
 
+const TransientErrorRegex = /EPROTO|ECONNRESET|ECONNREFUSED|ENETUNREACH|ETIMEDOUT|EAI_AGAIN|socket hang up|SSL routines/;
```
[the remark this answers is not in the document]
dmitrivMS: @copilot open a new pull request to apply changes based on [this feedback](https://github.com/microsoft/vscode/pull/290046#discussion_r2723187863)

src/vs/platform/request/node/requestService.ts
```diff
@@ -153,6 +155,31 @@ async function getNodeRequest(options: IRequestOptions): Promise<IRawRequestFunc
...
+		} catch (error) {
+			lastError = error as Error;
+			if (error instanceof CancellationError) {
+				throw error;
+			}
+
+			const isTransientError = TransientErrorRegex.test(getErrorMessage(error));
+			if (!isTransientError || attempt === maxRetries) {
+				throw error;
+			}
```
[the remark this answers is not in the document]
dmitrivMS: @copilot open a new pull request to apply changes based on [this feedback](https://github.com/microsoft/vscode/pull/290046#discussion_r2723187867)

src/vs/platform/request/node/requestService.ts
```diff
@@ -153,6 +156,32 @@ async function getNodeRequest(options: IRequestOptions): Promise<IRawRequestFunc
 }
 
 export async function nodeRequest(options: NodeRequestOptions, token: CancellationToken): Promise<IRequestContext> {
+	const maxRetries = 3;
+	let lastError: Error | undefined;
+	const canRetry = RETRYABLE_HTTP_METHODS_REGEX.test(options.type || 'GET');
```
chrmarti: I'd be a bit worried that this will miss subtle (or not so subtle) issues with retrying requests as they might not be truly idempotent.

If this would be useful in specific cases, you could add an option to enable retries, so each caller can decide if retries should be enabled.

chrmarti (dismissed):
Thanks, left a comment for discussion.

src/vs/platform/request/node/requestService.ts
dmitrivMS: Non-idempotent GET/OPTIONS/HEAD? Are there examples of such cases? Even if those exist, I imagine that would be pretty rare and it's worth making the general case more reliable. These errors are 500/network-level problems, I doubt they would cause any behavior change in those cases.

joaomoreno: I admire the courage @dmitrivMS, but I too am slightly worried of doing this deep down here. You are right in which GET/OPTIONS/HEAD should always be idempotent... though I'm not confident enough to ensure that that's the case in the real world...

Are the upstream places too many to count? Could we implement retry logic up there instead?

dmitrivMS: Would gathering some telemetry help increase confidence? I am not sure I fully understand the risk - please help me with that. If we throw here a 500 today, how retrying that up to 2 times would change the behavior? One change is the call would potentially be x2-x3 longer than before (and we would still throw), another change is we would hit the server 3 times instead of 1. Do you suspect that some of the requests coming through here may be changing server state on 500s?

dmitrivMS: Another alternative is reducing the regex to pure network errors (i.e. request never reaches the server):
- ENETUNREACH - Network unreachable, routing failure
- ECONNREFUSED - TCP connection refused (port closed, firewall blocked)
- EAI_AGAIN - DNS resolution temporary failure
- EPROTO - TLS/SSL protocol error during handshake
- SSL routines - SSL/TLS handshake failures (before HTTP request is sent)

That would be unfortunate though as I'd like ECONNRESET also to be included as I was hitting that in sanity tests.

joaomoreno: Other than idempo... ten...cy, _no_, idempocy, _no that's not it_, identi... po... cy..., _nope..._ IDEMPOTENCIAL, _yeah that's it_, other than those concerns, I fear we might be obscuring actual issues somewhere that today can surface as actionable errors.

> Another alternative is reducing the regex to pure network errors (i.e. request never reaches the server):

I can definitely get behind these!

dmitrivMS: Done

joaomoreno (approved):

PR #290059, opened 2026-01-24
Subagents render as a single line, similar to 'collapsed' thinking mode

roblourens:
#286606

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -2120,4 +2120,21 @@ export default tseslint.config(
 			'@typescript-eslint/consistent-generic-constructors': ['warn', 'constructor'],
 		}
 	},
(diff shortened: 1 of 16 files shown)


eslint.config.js
```diff
@@ -2120,4 +2120,21 @@ export default tseslint.config(
 			'@typescript-eslint/consistent-generic-constructors': ['warn', 'constructor'],
 		}
 	},
+	// Allow querySelector/querySelectorAll in test files - it's acceptable for test assertions
```
roblourens, line 2123: @jrieken hope it's ok that I disabled these querySelector rules for tests, i'm trying to write more unit tests for DOM stuff, it works better than playwright

pwang347 (approved):

PR #290079, opened 2026-01-24
Passing session resource to picker

osortega:

### DIFF
--- a/src/vs/workbench/contrib/chat/browser/widget/input/chatInputPart.ts
+++ b/src/vs/workbench/contrib/chat/browser/widget/input/chatInputPart.ts
@@ -816,7 +816,7 @@ export class ChatInputPart extends Disposable implements IHistoryNavigationWidge
 				}
 			};
 
-			const widget = this.instantiationService.createInstance(optionGroup.searchable ? SearchableOptionPickerActionItem : ChatSessionPickerActionItem, action, initialState, itemDelegate);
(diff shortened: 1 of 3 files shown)


src/vs/workbench/contrib/chat/browser/widget/input/chatInputPart.ts
```diff
@@ -816,7 +816,7 @@ export class ChatInputPart extends Disposable implements IHistoryNavigationWidge
 				}
 			};
 
-			const widget = this.instantiationService.createInstance(optionGroup.searchable ? SearchableOptionPickerActionItem : ChatSessionPickerActionItem, action, initialState, itemDelegate);
+			const widget = this.instantiationService.createInstance(optionGroup.searchable ? SearchableOptionPickerActionItem : ChatSessionPickerActionItem, action, initialState, itemDelegate, this._widget?.viewModel?.model.sessionResource);
```
roblourens: I wonder why you would store this on the picker instead of let the picker ask for the current value via the delegate, same as all the other state?

roblourens (dismissed):

rebornix (dismissed):

dmitrivMS (dismissed):

src/vs/workbench/contrib/chat/browser/widget/input/chatInputPart.ts
osortega: You know that's a great idea lol just updated it

bpasero (approved):

PR #290101, opened 2026-01-24
Fix: Scroll markdown preview to top when navigating to new document

### LINKED ISSUE #164071: Markdown preview does not scroll to top on new URI
Does this issue occur when all extensions are disabled?: N/A (depends on the built-in markdown extension)

- VS Code Version:  1.72.1
- OS Version: macOS 12.4

Steps to Reproduce:

1. Create two Markdown documents, one with a relative link to the other, and enough content in both documents to scroll down
2. Preview the document with the link and scroll down to the link, then click it

Expected Behavior:

The link to the other document should open the new document at the top

Actual Behavior:

The scroll position stays the same, opening the new document at a random spot in that document

Other Information:

This change is probably straightforward to make, but it would break the existing behavior. I'm struggling to imagine why that behavior would be desirable though, so not sure it would actually affect anyone in a meaningful way.

cc @mjbvz

shubham21155102:
Issue: Fixes microsoft/vscode#164071

Problem:
When clicking a relative link in a markdown preview to navigate to another markdown document, the new document would open at the same scroll position as the previous document instead of scrolling to the top. This created a confusing user experience where users would see the middle or bottom of the new document without any context.

Root Cause:
The markdown preview webview was preserving the scroll position when updating content for a different markdown resource. When the 'updateContent' message was received with a new resource (data.source !== documentResource), the code would replace the body content but did not reset the scroll position.

Solution:
Added window.scrollTo(0, 0) call immediately after replacing the document body when navigating to a different markdown resource. This ensures that when a user clicks a relative link to another markdown file, the preview scrolls to the top of the new document, providing the expected navigation behavior.

Changes:
- extensions/markdown-language-features/preview-src/index.ts: Added scroll reset when navigating to different documents in the 'updateContent' message handler

Testing:
The fix can be manually tested by:
1. Creating two markdown documents with enough content to scroll
2. Adding a relative link from one document to the other
3. Opening the first document in markdown preview
4. Scrolling down and clicking the link
5. Verifying the second document opens at the top (not at the previous scroll position)

Contributor: shubham21155102


dmitrivMS:
Superseded by https://github.com/microsoft/vscode/pull/333464

PR #290306, opened 2026-01-26
Allow passing in an extension to debug

TylerLeonhardt:
So that the automation can debug an extension

### DIFF
--- a/test/mcp/package-lock.json
+++ b/test/mcp/package-lock.json
@@ -839,6 +839,16 @@
         "node": ">= 0.4"
       }
     },
+    "node_modules/hono": {
+      "version": "4.11.5",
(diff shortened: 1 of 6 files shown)


test/mcp/package-lock.json
```diff
@@ -839,6 +839,16 @@
         "node": ">= 0.4"
       }
     },
+    "node_modules/hono": {
```
rwoll, line 842: Why's there a change in this lockfile?

rwoll (approved):

test/mcp/package-lock.json
TylerLeonhardt: Because someone bumped a version and forgot to update it. I just ran a plain npm install and this happened.

Really I ran `npm ci` and that failed requiring me to `npm install` in order to continue

PR #290309, opened 2026-01-26
A semi-working way to send a message in chat

TylerLeonhardt:
The old way stopped working... idk why... and this way isn't perfect - it has to be written on 1 line.... but I guess it's something...

### DIFF
--- a/test/automation/src/chat.ts
+++ b/test/automation/src/chat.ts
@@ -32,20 +32,10 @@ export class Chat {
 		// Wait for the editor to be focused
 		await this.waitForInputFocus();
 
-		// Dispatch a paste event with the message
-		await this.code.driver.currentPage.evaluate(({ selector, text }: { selector: string; text: string }) => {
-			const element = document.querySelector(selector);
-			if (element) {
-				const dataTransfer = new DataTransfer();
-				dataTransfer.setData('text/plain', text);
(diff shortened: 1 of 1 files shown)


test/automation/src/chat.ts
```diff
@@ -32,20 +32,10 @@ export class Chat {
...
-					bubbles: true,
-					cancelable: true
-				});
-				element.dispatchEvent(pasteEvent);
-			}
-		}, { selector: CHAT_INPUT, text: message });
+		// Type the message using pressSequentially - this works with Monaco editors
+		// Note: Newlines are replaced with spaces since Enter key submits in chat input
+		const sanitizedMessage = message.replace(/\n/g, ' ');
+		await this.code.driver.currentPage.locator(CHAT_INPUT).pressSequentially(sanitizedMessage);
```
rwoll: @TylerLeonhardt - can you use `executeCommand('workbench.action.chat.open', …)` like we do in the eval harness?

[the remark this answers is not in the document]
TylerLeonhardt: This doesn't work for the Chat input for whatever reason.

TylerLeonhardt: I believe what you're referring to is the function from the extension host... this layer doesn't have the extension host. It's only controlling the browser.

bpasero (approved):

PR #290513, opened 2026-01-26
Implement chat tips service and related features

sbatten:
Introduce a chat tips service to enhance user experience with contextual tips. Update the frequency of tips, add necessary tests, and ensure proper integration within the chat interface.

image


src/vs/workbench/contrib/chat/browser/chatTipService.ts
```diff
@@ -0,0 +1,186 @@
...
+	/**
+	 * Command IDs that are allowed to be executed from this tip's markdown.
+	 */
+	readonly enabledCommands?: string[];
+}
+
+/**
+ * Static catalog of tips. Each tip has an optional when clause for eligibility.
+ */
+const TIP_CATALOG: ITipDefinition[] = [
```
justschen, line 53: i almost wonder if we need more tips :0

just anecdotally,  when i'd be playing games, there would be tips like these on loading screens, and i'd get annoyed if i saw the same tip too much

justschen (dismissed):

src/vs/workbench/contrib/chat/browser/chatTipService.ts
sbatten: agreed

justschen (approved):

PR #290546, opened 2026-01-26
agents: use a shorter delay for hovering on an agent session requiring input

connor4312:
Adds support for reduced hover delays through a new reducedDelay boolean option in IHoverLifecycleOptions. When enabled, hovers will use the new workbench.hover.reducedDelay setting (defaults to 500ms on all platforms) instead of the standard workbench.hover.delay setting.

Changes:
- Adds optional 'reducedDelay' boolean parameter to IHoverLifecycleOptions
- Adds new 'workbench.hover.reducedDelay' setting with 500ms default
- Updates HoverService to use reducedDelay setting when the option is true
- Applies reduced delay in agent sessions viewer for input-requiring sessions
- Updates test coverage for the new reducedDelay functionality

### DIFF
--- a/src/vs/base/browser/ui/hover/hover.ts
+++ b/src/vs/base/browser/ui/hover/hover.ts
@@ -274,6 +274,12 @@ export interface IHoverLifecycleOptions {
 	groupId?: string;
 
 	/**
+	 * Whether to use a reduced delay before showing the hover. If true, the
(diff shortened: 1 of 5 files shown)


joshspicer (dismissed):

Tyriar (changes requested):
When this has come up in the past we have avoided adding this to keep things consistent. Why is it too slow now? If anything it feels too fast on Windows as the popup is huge.

Tyriar (approved):

PR #290646, opened 2026-01-27
Ignore errors when trying to clean up server logs

dmitrivMS:
Fixing server error found by sanity tests:
[2026-01-27T00:38:15.038Z] ERROR: [Server Error] [00:38:14] Error: ENOENT: no such file or directory, scandir '/tmp/vscode-sanityr4UsY0/data/logs'
at async Object.readdir (node:internal/fs/promises:957:18)
at async b6 (file:///tmp/vscode-sanitygp6yWf/vscode-server-linux-armhf/out/server-main.js:27:91415)
at async Object.Rh (file:///tmp/vscode-sanitygp6yWf/vscode-server-linux-armhf/out/server-main.js:27:91280)
at async VB (file:///tmp/vscode-sanitygp6yWf/vscode-server-linux-armhf/out/server-main.js:190:2819) {
errno: -2,
code: 'ENOENT',
syscall: 'scandir',
path: '/tmp/vscode-sanityr4UsY0/data/logs'
}

### DIFF
--- a/src/vs/server/node/serverServices.ts
+++ b/src/vs/server/node/serverServices.ts
@@ -377,7 +377,12 @@ function twodigits(n: number): string {
 async function cleanupOlderLogs(logsPath: string): Promise<void> {
 	const currentLog = path.basename(logsPath);
 	const logsRoot = path.dirname(logsPath);
-	const children = await Promises.readdir(logsRoot).catch(() => []);
+
+	if (!await Promises.exists(logsRoot)) {
+		return; // Logs root doesn't exist yet, nothing to clean up
+	}
+
+	const children = await Promises.readdir(logsRoot);
 	const allSessions = children.filter(name => /^\d{8}T\d{6}$/.test(name));
 	const oldSessions = allSessions.sort().filter((d) => d !== currentLog);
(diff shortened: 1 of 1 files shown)


src/vs/server/node/serverServices.ts
```diff
@@ -377,7 +377,7 @@ function twodigits(n: number): string {
 async function cleanupOlderLogs(logsPath: string): Promise<void> {
 	const currentLog = path.basename(logsPath);
 	const logsRoot = path.dirname(logsPath);
-	const children = await Promises.readdir(logsRoot);
+	const children = await Promises.readdir(logsRoot).catch(() => []);
```
joaomoreno: Shouldn't we rather wrap this in `try catch` and simply return early if the error is `ENOENT`? It's still a good idea to propagate other errors.

joaomoreno (changes requested):

src/vs/server/node/serverServices.ts
dmitrivMS: Updated to use `!exists` - I hope that works.

joaomoreno (approved):

PR #290658, opened 2026-01-27
Add test for arm64 Linux with 64K page size

dmitrivMS:
64k page config cannot be set in a running VM, so using emulator with core replacement to run this test.

Related to #268354

This test is kind of slow, so we may want to run it more selectively.

### DIFF
--- a/build/azure-pipelines/product-sanity-tests.yml
+++ b/build/azure-pipelines/product-sanity-tests.yml
@@ -313,3 +313,13 @@ extends:
               container: ubuntu
               baseImage: ubuntu:24.04
               arch: arm64
+
+          - template: build/azure-pipelines/common/sanity-tests.yml@self
+            parameters:
+              name: ubuntu_24_04_arm64_64k
+              displayName: Ubuntu 24.04 arm64 (64K page)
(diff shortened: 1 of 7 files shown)


rebornix (approved):

build/azure-pipelines/product-sanity-tests.yml
```diff
@@ -313,3 +313,13 @@ extends:
               container: ubuntu
               baseImage: ubuntu:24.04
               arch: arm64
+
+          - template: build/azure-pipelines/common/sanity-tests.yml@self
+            parameters:
+              name: ubuntu_24_04_arm64_64k
+              displayName: Ubuntu 24.04 arm64 (64K page)
+              poolName: 1es-ubuntu-22.04-x64
```
deepak1556, line 321: Having the host as arm64 should help speed up the emulation, any reason amd64 was chosen here for the pool ?

dmitrivMS: That was the initial idea, but it was failing on arm64 pool. That said, I start with a different distro, so I'll give it another try.

PR #291298, opened 2026-01-28
fix #290724

### LINKED ISSUE #290724: Agent sessions profile is still around in code
https://github.com/microsoft/vscode/blob/26f5311fe1ec93a467c7eb2401d92ce18a6a1db8/src/vs/workbench/contrib/userDataProfile/browser/userDataProfile.ts#L288-L291

https://github.com/microsoft/vscode/blob/26f5311fe1ec93a467c7eb2401d92ce18a6a1db8/src/vs/code/electron-browser/workbench/workbench.ts#L28-L33

We need some alternative for the splash at least.

sandy081:
fix #290724

### DIFF
--- a/src/vs/code/electron-browser/workbench/workbench.ts
+++ b/src/vs/code/electron-browser/workbench/workbench.ts
@@ -24,14 +24,7 @@
 
 	function showSplash(configuration: INativeWindowConfiguration) {
 		performance.mark('code/willShowPartsSplash');
-
-		const isAgentSessionsWindow = configuration.profiles?.profile?.id === 'agent-sessions';
-		if (isAgentSessionsWindow) {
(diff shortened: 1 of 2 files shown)


Tyriar (approved):

bpasero (changes requested):
@sandy081 could we not put the fact that a window is sessions window into the window configuration to show the good splash? We should know this from the main side based on the workspace that opens right?

sandy081 (commented):
Yes that is possible. Since there is no profile change, we have to reuse color components from default splash. Lets do it next milestone.

sandy081 (commented):
https://github.com/microsoft/vscode/issues/291509

bpasero (approved):

PR #291318, opened 2026-01-28
Improve initial instruction/prompt/agent/skill file contents

### LINKED ISSUE #290994: new agent.md default empty tool selection
When creating new agent.md files I will go through the UI so it will give me the template. I will then edit the text and things but always forget about the tools because for many of those I made I just want it to use the tools that the default agent mode has. I then find myself trying to use the new agent and arguing with it a bit before it tells me explicitly I forgot to give it tool access. Might be nice if the "default" when creating a new agent was a basic set of tools? Or maybe uses to the ones the default agent has? I think if you include no tool reference it uses the default ones but I agree showing the tools UI is useful for people to set it up

aeschli:
Fixes https://github.com/microsoft/vscode/issues/290994

### DIFF
--- a/src/vs/workbench/contrib/chat/browser/promptSyntax/newPromptFileActions.ts
+++ b/src/vs/workbench/contrib/chat/browser/promptSyntax/newPromptFileActions.ts
@@ -24,9 +24,8 @@ import { ChatContextKeys } from '../../common/actions/chatContextKeys.js';
 import { CHAT_CATEGORY } from '../actions/chatActions.js';
 import { askForPromptFileName } from './pickers/askForPromptName.js';
 import { askForPromptSourceFolder } from './pickers/askForPromptSourceFolder.js';
-import { IChatModeService } from '../../common/chatModes.js';
 import { IQuickInputService } from '../../../../../platform/quickinput/common/quickInput.js';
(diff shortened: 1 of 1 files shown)


Tyriar (approved):

src/vs/workbench/contrib/chat/browser/promptSyntax/newPromptFileActions.ts
```diff
@@ -140,51 +139,47 @@ class AbstractNewPromptFileAction extends Action2 {
...
 			].join('\n');
 		case PromptsType.agent:
 			return [
 				`---`,
-				`description: '\${1:Describe what this custom agent does and when to use it.}'`,
-				`tools: []`,
+				`name: ${name ?? '${1:agent-name}'}`,
+				`description: \${2:Describe what this custom agent does and when to use it.}`,
+				`argument-hint: \${3:The inputs this agent expects, e.g., "a task to implement" or "a question to answer".}`,
+				`# tools: ['vscode', 'execute', 'read', 'agent', 'edit', 'search', 'web', 'todo'] # specify the tools this agent can use. if not set, all enabled tools are allowed`,
```
pwang347, line 166: should this be a dynamic list?

PR #291325, opened 2026-01-28
allow prompt for input detection in bg terminals

### LINKED ISSUE #290983: Terminal awaiting input feature does not work with background terminals
## Description

Image

The "terminal awaiting input" feature does not work when a terminal has been pushed to the background.

Prompt:

```
Run:

read -p "Continue? [Y/n] " answer && echo "You answered: $answer"

in the BACKGROUND.
```
The prompt runs a background terminal, but the "Terminal is awaiting input" confirmation does not appear. I'd expect it to since it does work with a foregroun terminal:

```
read -p "Continue? [Y/n] " answer && echo "You answered: $answer"
``` 

## Related

- TPI: #290583

## Environment

- OS: macOS

meganrogge:
fix #290983

The `outputMonitor` is created inside an event handler that fires when the command starts executing. For foreground terminals this worked because the code awaits the full command execution, giving the event time to fire and create the monitor. But for background terminals, the code checked `if (outputMonitor)` immediately after calling `execution.start()`, before the event had a chance to fire. So `outputMonitor` was always undefined and the input prompting was skipped entirely. The fix adds `await startMarkerPromise` before checking `outputMonitor`, ensuring the monitor actually exists before we try to use it.

Screenshot 2026-01-28 at 11 27 38 AM

### DIFF
--- a/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/runInTerminalTool.ts
+++ b/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/runInTerminalTool.ts
@@ -782,6 +782,7 @@ export class RunInTerminalTool extends Disposable implements IToolImpl {
 			RunInTerminalTool._activeExecutions.set(termId, execution);
 
 			// Set up OutputMonitor when start marker is created
+			const startMarkerPromise = Event.toPromise(execution.strategy.onDidCreateStartMarker);
 			store.add(execution.strategy.onDidCreateStartMarker(startMarker => {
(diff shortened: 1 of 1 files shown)


src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/runInTerminalTool.ts
Tyriar, line 814: Warn in logs on else?

Tyriar (approved):

PR #291327, opened 2026-01-28
fix: add cancellation token support to ChatAgentResponseStream

### LINKED ISSUE #291295: Pressing Stop does not stop Agent session if using Ask tool
https://github.com/user-attachments/assets/692cd479-b7b9-4d69-8e82-f02cc636c349

cc @digitarald @roblourens

karthiknadig:
fixes: https://github.com/microsoft/vscode/issues/291295

### DIFF
--- a/src/vs/workbench/api/common/extHostChatAgents2.ts
+++ b/src/vs/workbench/api/common/extHostChatAgents2.ts
@@ -5,7 +5,7 @@
 
 import type * as vscode from 'vscode';
 import { coalesce } from '../../../base/common/arrays.js';
-import { DeferredPromise, timeout } from '../../../base/common/async.js';
+import { DeferredPromise, raceCancellation, timeout } from '../../../base/common/async.js';
 import { CancellationToken, CancellationTokenSource } from '../../../base/common/cancellation.js';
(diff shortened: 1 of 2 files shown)


bpasero (approved):

src/vs/workbench/api/common/extHostChatAgents2.ts
```diff
@@ -330,8 +331,8 @@ export class ChatAgentResponseStream {
 
 					_report(dto);
 
-					// Wait for the user to submit answers
-					return deferred.p;
+					// Wait for the user to submit answers, but respect cancellation
+					return raceCancellation(deferred.p, that._token);
```
roblourens, line 335: Why is this tool implemented in the extension when it has this tight coupling with UI? The `ChatAgentResponseStream` is not supposed to return a value to the caller, that's contrary to the concept of a "stream" IMO

karthiknadig: I will create an issue to move this into main and get rid of this part.

karthiknadig: https://github.com/microsoft/vscode/issues/291349

PR #291334, opened 2026-01-28
do not show empty output on command failure

meganrogge:
fix #291009

Screenshot 2026-01-28 at 11 52 41 AM


src/vs/workbench/contrib/chat/browser/widget/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
```diff
@@ -528,7 +528,7 @@ export class ChatTerminalToolProgressPart extends BaseChatToolInvocationSubPart
 			this._showOutputAction.value = showOutputAction;
 			const autoExpandFailures = this._configurationService.getValue<boolean>(ChatConfiguration.AutoExpandToolFailures);
 			const exitCode = resolvedCommand?.exitCode ?? this._terminalData.terminalCommandState?.exitCode;
-			if (exitCode !== undefined && exitCode !== 0 && autoExpandFailures) {
+			if (exitCode !== undefined && exitCode !== 0 && autoExpandFailures && (!resolvedCommand || resolvedCommand.hasOutput())) {
 				this._toggleOutput(true);
 			}
```
Tyriar, line 533: I kind of preferred it the other way, it was a failure and therefore we should expand consistently. Otherwise the user would have to expand it to confirm the output

Tyriar (changes requested):

PR #291381, opened 2026-01-28
move ask question tool to core

### LINKED ISSUE #291349: Move `ask_questions` tool implementation into VS Code core
Per feedback in #291327, the `ask_questions` tool (question carousel) has tight coupling with UI and returns a value from `ChatAgentResponseStream`, which is contrary to the "stream" concept.

The tool should be implemented similar to the demo extension approach, with the implementation living in VS Code core rather than requiring the extension to await a response from the stream.

Related:
- #291295 (original stop button issue)
- #291327 (cancellation fix PR with this feedback)

cc @roblourens @digitarald

karthiknadig:
fixes https://github.com/microsoft/vscode/issues/291349


DonJayamanne:
@karthiknadig The `questionCarousel` is used in Background Agent, as Copilot CLI also supports an ASK tool, hence we need this same UI in background (copilot CLI).

I.e. you can move the tool to core, but I would suggest leaving the method on the stream, so it can be invoked from chat extension

PR #291455, opened 2026-01-29
Add chat extension custom URI to allowed schemes for rendering

pwang347:
Problem: extension-provided URIs can use virtual file systems. URI scheme check currently rejects those, but we don't want to remove this check. I think we can just check if the files were extension-contributed.

image

Fixes https://github.com/microsoft/vscode/issues/292943


zhichli (dismissed):

src/vs/workbench/contrib/chat/browser/widget/chatContentMarkdownRenderer.ts
```diff
@@ -80,6 +83,19 @@ export class ChatContentMarkdownRenderer implements IMarkdownRenderer {
 				...options?.sanitizerConfig,
 				allowedLinkSchemes: { augment: [product.urlProtocol] },
 				remoteImageIsAllowed: (_uri) => false,
+				isLinkAllowed: (href) => {
```
mjbvz: Instead of adding a new setting, could you use the existing `allowedProtocols`? Looks like this is doing the same type of thing

If you really do need the function, let's add so that you can either pass in `AllowedLinksConfig | (link: string) => boolean`

src/vs/base/browser/markdownRenderer.ts
mjbvz, line 635: Let's change the type here to make it clearer how `isLinkAllowed` worked with the existing setting

Take a look at how `allowedAttributes` works for example. With it you can either pass in a string or a validation function

pwang347: updated, let me know if this is what you had in mind. Thanks!

src/vs/base/browser/markdownRenderer.ts
```diff
@@ -61,7 +61,7 @@ export interface MarkdownSanitizerConfig {
 		readonly override: ReadonlyArray<string | domSanitize.SanitizeAttributeRule>;
 	};
 	readonly allowedLinkSchemes?: {
-		readonly augment: readonly string[];
+		readonly augment: ReadonlyArray<string | ((link: string) => boolean)>;
```
mjbvz: I still think this makes the api confusing. Let's try to break down exactly what you need:

Isthe validation per-uri scheme? If it is, look at how `SanitizeAttributeRule` works. This lets you specify a specific attribute that you can run a test function against

If it's instead needs to be run on every url, consider how it interacts with the existing url rules. Does it override them? Augment them? What happens if they are in disagreement?


Sorry to be annoying here but I spent a lot of time trying to clean up the interfaces around markdown and the dom sanitation and want to try to keep them super clear. Happy to meet later to discuss specific ideas too

pwang347: No worries at all, appreciate the discussion.

> Isthe validation per-uri scheme? If it is, look at how SanitizeAttributeRule works. This lets you specify a specific attribute that you can run a test function against

I don't think we can just use uri scheme here (if we want to be correct). The scenario here is that an extension author may supply us with prompt files, which can come from their own FS implementations. We could definitely just allow rendering for anything coming from that FS regardless of whether it's a prompt file, but that seems to defeat the purpose (why not just allow anything to render in chat then, since this is a potential loophole).

> If it's instead needs to be run on every url, consider how it interacts with the existing url rules. Does it override them? Augment them? What happens if they are in disagreement?

I thought `augment` was fine since we are explicitly allowing / bypassing for certain links, but also happy to discuss alternatives (e.g., `overrides` or something else).

src/vs/base/browser/ui/codicons/codicon/codicon.ttf
pwang347, line 1: not sure why this is here, TODO: remove this

aeschli:
IMO there should be no reference to the IPrompsService. Prompt files are just one example of a (potentially virtual) resource that can be opened.

src/vs/base/browser/markdownRenderer.ts
mjbvz: Thinking about this a bit more, I'm actually if we should go back to a simple check function:

```
linkValidator?: (link: string) => boolean
```

In the documentation, just make sure to note that this runs after both `allowedLinkProtocols` and `shouldKeepLink` and that it defaults to keeping links


Then in your check function for chat, just make sure to ignore anything that isn't from one of the protocols you want to validate

PR #291474, opened 2026-01-29
Move `Collapse All` action in SCM to leftmost position

### LINKED ISSUE #289745: Source Control - hide the collapse all button by default and make it configurable
Version: 1.109.0-insider (system setup)
Commit: f60f946b1736ae6f26b944f8f5b6ce6270d05405
Date: 2026-01-21T05:54:27.524Z
Electron: 39.2.7
ElectronBuildId: 13098910
Chromium: 142.0.7444.235
Node.js: 22.21.1
V8: 14.2.231.21-electron.0
OS: Windows_NT x64 10.0.26200

There is already collapse button at the left, why do we need another collapse all button at the right? this is messing up with the muscle memory as well becuse that is wher the stage button used to be.

If the collapse button is really needed, can it be moved behind a config and disabled by default?

Image

### LINKED ISSUE #290011: 'Collapse all' button is breaking my muscle memory
In Insiders, there is now a 'collapse all' button in the right-most position, which used to be the 'stage' button. I would suggest keeping the stage button on the right since it's probably the most commonly used action, and now I have to think carefully to avoid clicking the new button

Image

dmitrivMS:
Fixes #290011
Fixes #289745


src/vs/workbench/contrib/scm/browser/scmViewPane.ts
```diff
@@ -1381,7 +1381,7 @@ class CollapseAllAction extends ViewAction<SCMViewPane> {
 				id: MenuId.SCMResourceGroupContext,
 				group: 'inline',
```
lszomoru: Let's change the group and move it into the context menu at the bottom into its own group.

lszomoru (changes requested):

src/vs/workbench/contrib/scm/browser/scmViewPane.ts
dmitrivMS: Done:
image

lszomoru (approved):

Issue #290816, opened 2026-01-27
Sandbox - network isolation not working


lszomoru:
Testing #290620

Using the following settings:

```json
"chat.tools.terminal.sandbox.enabled": true,
  "chat.tools.terminal.sandbox.network": {
    "allowedDomains": [
      "*.visualstudio.com"
    ],
    "deniedDomains": []
  }
```

Prompt:
```
Use the terminal tool a curl to download the latest VS Code release
```

<img width="471" height="410" alt="Image" src="https://github.com/user-attachments/assets/5f26bd65-d76b-4e54-8102-7c455c98b619" />

PR #291483, opened 2026-01-29
Sandbox: Updating output for terminal command errors due to domain missing from allow list. 

dileepyavan:
fixes:#290816
In sandboxing mode, for domains that are not added to the allow list, the command fails with no information about failed domain.
Also there can be redirected links that redirect the  request to another domain thats not in allow list.

image


Tyriar (commented):
This change is too big to land after testing, let's do this in February.

benvillalobos (approved):

PR #291554, opened 2026-01-29
Pass metadata to agent session commands

alexr00:


src/vs/workbench/contrib/chat/browser/agentSessions/agentSessionsModel.ts
```diff
@@ -515,6 +515,7 @@ export class AgentSessionsModel extends Disposable implements IAgentSessionsMode
 					archived: session.archived,
 					timing: session.timing,
 					changes: normalizedChanges,
+					metadata: session.metadata,
```
bpasero, line 518: What about the cache of sessions we maintain (see `saveCachedSessions`), should the metadata be added there too? And how much data are we talking about, is it serializable?

bpasero (changes requested):

src/vs/workbench/contrib/chat/browser/agentSessions/agentSessionsModel.ts
alexr00: Yes, it's serializable: 
https://github.com/microsoft/vscode/blob/d773412494fb24bde6cd77eb59d87581c6c4a7b6/src/vscode-dts/vscode.proposed.chatSessionsProvider.d.ts#L236-L241

I'm not sure about `saveCachedSessions`, but it seems like it probably does need to be added there.

src/vs/workbench/contrib/chat/browser/agentSessions/agentSessionsModel.ts
```diff
@@ -719,6 +722,7 @@ class AgentSessionsCache {
 			timing: session.timing,
 
 			changes: session.changes,
+			metadata: session.metadata
```
bpasero, line 725: I think this is dangerous: we do not know the data inside the map, yet we JSON.stringify it. How can we ensure the data does not end up as broken?

bpasero (approved):

bpasero (dismissed):
🖐️

src/vs/workbench/contrib/chat/browser/agentSessions/agentSessionsModel.ts
alexr00: I would suggest that we have a check at the time that the data enters our system through the API. I'll open a new PR with this.

bpasero: @lszomoru fyi this is similar feedback I gave to your PR, can the API enforce this is serializable by restricting to certain types?

alexr00: Ah, I see that you requested changes. I'll do it in this PR then.

joshspicer (approved):

lszomoru:
@alexr00, I see that this got merged. What was not clear to me is why did this change require changes to `action.ts`? What is special about this scenario that our actions did not support until now? Maybe I am missing something obvious here...

Issue #291123, opened 2026-01-28
2026 themes: terminal border overlaps completions


eli-w-king:
Border overlaps terminal completions on hover.

<img width="1956" height="702" alt="Image" src="https://github.com/user-attachments/assets/b10658cf-63d3-4f64-8d73-5360af5331ce" />

cc: @TylerLeonhardt

PR #291583, opened 2026-01-29
Fix stacking context issues in terminal suggest contribution

mrleemurray:
Resolve stacking context issues by utilizing the workbench container for terminal suggestions, ensuring proper rendering within panels and sidebars.

Addresses: https://github.com/microsoft/vscode/issues/291123


joaomoreno (dismissed):

Tyriar (changes requested):
Can you please create/link an issue? This is a lot of changes this late in endgame

src/vs/workbench/contrib/terminalContrib/suggest/browser/terminal.suggest.contribution.ts
```diff
@@ -260,7 +262,8 @@ class TerminalSuggestContribution extends DisposableStore implements ITerminalCo
 			return xtermElement;
 		}
 
-		return dom.findParentWithClass(xtermElement, 'panel') ?? xtermElement;
+		// Use the workbench container to avoid stacking context issues with panels/sidebars
+		return this._layoutService.getContainer(dom.getWindow(xtermElement));
```
Tyriar, line 266: This looks like a relatively scary change, did you verify this? Also won't this put the element at the root and potentially cause a11y tab order issues?

Could it be handled instead by tweaking a z-index somewhere?

extensions/theme-2026/themes/styles.css
```diff
@@ -188,8 +207,17 @@
 }
 
 /* Context Menus */
-.monaco-workbench .monaco-menu .monaco-action-bar.vertical { box-shadow: var(--shadow-lg); border: none; border-radius: var(--radius-xl); overflow: hidden; backdrop-filter: var(--backdrop-blur-md); -webkit-backdrop-filter: var(--backdrop-blur-md); }
-.monaco-workbench .context-view .monaco-menu { box-shadow: var(--shadow-lg); border: none; border-radius: var(--radius-xl); }
+.monaco-workbench .monaco-menu .monaco-action-bar.vertical {
```
Tyriar, line 210: Why so many theme css changes to fix this bug which appears to be only related to z-index/element container?

Tyriar (changes requested):
The issue is on the Feb milestone btw

Tyriar:
@meganrogge definitely should have been an explicit reviewer on this one.

PR #291637, opened 2026-01-29
Browser: make paused state less jarring

kycutler:
* Reduce darkening effect from 60% to 15% for simple hover overlays (i.e. non-notifications)
* Improve screenshotting logic to better keep the screenshot up-to-date and reduce flicker

<details>
<summary>Before</summary>

6a3bae68-0873-4457-bfdf-bc91a014e4c4

</details>

<details>
<summary>After</summary>

446dcf02-70da-4daa-8ce3-61f94f2ca9a8

</details>

### DIFF
--- a/src/vs/workbench/contrib/browserView/electron-browser/browserEditor.ts
+++ b/src/vs/workbench/contrib/browserView/electron-browser/browserEditor.ts
@@ -5,7 +5,7 @@
 
 import './media/browser.css';
 import { localize } from '../../../../nls.js';
-import { $, addDisposableListener, Dimension, disposableWindowInterval, EventType, IDomPosition, registerExternalFocusChecker } from '../../../../base/browser/dom.js';
+import { $, addDisposableListener, Dimension, EventType, IDomPosition, registerExternalFocusChecker } from '../../../../base/browser/dom.js';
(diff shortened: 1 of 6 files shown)


src/vs/workbench/contrib/browserView/electron-browser/browserEditor.ts
```diff
@@ -398,16 +402,10 @@ export class BrowserEditor extends EditorPane {
 				this.layoutBrowserContainer();
 			}
 		}));
-		// Capture screenshot periodically (once per second) to keep background updated
```
jruales, line 401: I remember that if I resize the window aggressively from the corner, I sometimes saw the screenshot behind the webcontentsview as a double being scaled without a full overlap. Since we don’t screenshot periodically anymore, we need to make sure that this doesn’t happen anymore as that screenshot may be very outdated

jruales (dismissed):

src/vs/workbench/contrib/browserView/electron-browser/browserEditor.ts
kycutler: It is still captured every second; just handled via a repeated timeout rather than interval. This lets us pause / resume it more dynamically when the visibility changes

eleanorjboyd (approved):

PR #291696, opened 2026-01-29
Agent status indicator: Do not show unread on empty workspace

### LINKED ISSUE #291385: Every empty window shows all cloud and background sessions as unread
I think this is not the greatest user experience considering the prominence of the indicator in the title:

Image

Every empty window I open will show me the same background and cloud sessions as unread. 

As a tactical fix, should we strictly hide background and cloud sessions in empty windows or workspace windows without folders?

cc @lostintangent

joshspicer:
fixes https://github.com/microsoft/vscode/issues/291385


pwang347 (dismissed):

src/vs/workbench/contrib/chat/browser/agentSessions/experiments/agentTitleBarStatusWidget.ts
```diff
@@ -782,7 +782,9 @@ export class AgentTitleBarStatusWidget extends BaseActionViewItem {
 		const viewSessionsEnabled = this.configurationService.getValue<boolean>(ChatConfiguration.ChatViewSessionsEnabled) !== false;
 
 		// Unread section (blue dot + count)
-		if (viewSessionsEnabled && hasUnreadSessions) {
+		// Don't show unread indicator in empty workspace
```
bpasero: This is not how you check for a window being empty.

Use: https://github.com/microsoft/vscode/blob/bf729fa50adbc422b858ba2dcb138357b6a5f7f6/src/vs/platform/workspace/common/workspace.ts#L60

bpasero (dismissed):

lszomoru (approved):

PR #291800, opened 2026-01-30
Prepopulate the Find in Page search text with the currently selected text in the page

### LINKED ISSUE #291378: Integrated Browser - The find input isn't pre-populated with selected text when opening it
Testing https://github.com/microsoft/vscode/issues/290587

If you open the integrated browser, navigate to a page, select some text and then press `CMD+F`, the find input doesn't pre-populate itself with the selected text. This isn't the end of the world, but this is a pretty common UX behavior in browsers, and I expected it to work this way.

jruales:
Closes https://github.com/microsoft/vscode/issues/291378

* If the user has text selected when they use the search functionality, it prepopulates the search box with the selected text
* Adds preload script that ensures that the `window.getSelection()` function doesn't get overwritten, for security purposes. It exposes that function under `window.vscodeBrowserView` within an Isolated World, using Electron's context isolation.

Only prepopulates the search box if:
  * The selection doesn't have newlines
  * The page has already finished loading, since otherwise Electron's executeJavaScript would make us wait until the page is loaded to give us the selection

---

Demo:

https://github.com/user-attachments/assets/de2c2348-a112-43f0-8f05-5ddf186940d2


src/vs/platform/browserView/electron-main/browserView.ts
```diff
@@ -475,6 +477,23 @@ export class BrowserView extends Disposable {
...
+	 * Returns immediately with empty string if the page is still loading.
+	 */
+	async getSelectedText(): Promise<string> {
+		// we don't want to wait for the page to finish loading, which executeJavaScript normally does.
+		if (this._view.webContents.isLoading()) {
+			return '';
+		}
+		try {
+			// Uses the preload-exposed API window.vscodeBrowserView.
+			return await this._view.webContents.executeJavaScript('window.vscodeBrowserView?.getSelectedText?.() ?? ""');
```
kycutler: Can't this still be overridden by the page?

Wondering if it would be better to expose selection capabilities from Electron itself

jruales: @kycutler In theory, the whole point of contextBridge is that it’s safe and can’t be tampered with. I’ll run some experiments though
https://www.electronjs.org/docs/latest/api/context-bridge

when you say expose from electron itself, do you mean contributing those to the Electron codebase?

jruales: I'm not able to override it from devtools Console at least. See the commands that I ran here:
image

jruales: More specifically:
image

jruales: Despite seeing (based on the demonstration above) that there was no harm in having the API exposed in the "main world" (accessible by the website itself), I ended up switching to using an "isolated world" which doesn't pollute the website's window object.

The context for the isolated world I added shows up in DevTools as "Integrated Browser Isolated Context":
image

I could have technically reused the "Electron Isolated Context" shown in the same screenshot, which is the environment where the preload script itself runs (namely, isolated world with world id 999), but I opted for creating a separate one (with world id 1000)

src/vs/platform/browserView/electron-main/browserView.ts
```diff
@@ -475,6 +477,24 @@ export class BrowserView extends Disposable {
...
+	 * Returns immediately with empty string if the page is still loading.
+	 */
+	async getSelectedText(): Promise<string> {
+		// we don't want to wait for the page to finish loading, which executeJavaScript normally does.
+		if (this._view.webContents.isLoading()) {
+			return '';
+		}
+		try {
+			// Uses our preloaded contextBridge-exposed API.
+			// It would not be safe to call window.getSelection() directly here, because the page could override it.
```
kycutler: Is this still true? Could we just call `window.getSelection()` in an isolated world?

kycutler: To be clear, I like the idea of having our own preload scripts -- not suggesting we remove those even if they aren't strictly needed yet. Mostly just asking if the comment should be updated

jruales: Good point!

jruales: Removed the comment

deepak1556 (commented):
Don't see a reason we need to create our own isolated world, this could use the existing isolated world where the preload script runs.

LGTM, otherwise

src/vs/platform/browserView/electron-main/browserView.ts
jruales: In the end, I moved to isolated world 999 per Deepak's suggestion

deepak1556 (approved):

PR #291825, opened 2026-01-30
Replace child_process.exec with execFile to prevent potential command injection

jeevaratnamputla:
Summary

This change replaces the use of child_process.exec with child_process.execFile in the Grunt extension to mitigate potential command injection risks flagged by static analysis.

The previous implementation executed a command string via a shell, which can be unsafe if any part of the command is user-controllable. While the current usage appears controlled, the pattern itself is vulnerable and commonly flagged by SAST tools.

Details

Replaced exec() with execFile() to avoid shell invocation

Separated the executable (grunt) from its arguments (--help, --no-color)

Updated the helper function signature to explicitly pass arguments as an array

Preserves existing behavior while improving security posture

Rationale

execFile executes binaries directly without spawning a shell, which prevents shell interpretation and reduces the risk of command injection. This aligns with Node.js security best practices and addresses the concerns raised by the static analysis rule.

Testing

Verified the extension continues to execute the Grunt help command as expected

No functional behavior changes intended


mjbvz (approved):
Thanks!

TylerLeonhardt (approved):

Issue #291739, opened 2026-01-29
Bug: documents opened by EH: vscode.workspace.openTextDocument() are not disposed correctly


n-gist:
- VS Code Version: 1.108.2
- OS Version: Windows 11

This was noticed by the fact that diagnostic entries of the file being closed doesn't get removed from Problems View, if a file was opened from extension by `vscode.workspace.openTextDocument()`, which differs from opening files from Explorer panel, in which case diagnostic entries are removed on file close.

Reproduction steps:
1. Open a file which contains problems, using `vscode.workspace.openTextDocument()` call from extension
2. Ensure that problems appear in Problems View
3. Close file
4. Observe that diagnostics remains in Problems View
5. Additionally, if you give VS Code several minutes to idle, at some point it triggers resource cleanup, removing giagnostics from Problems View unexpectedly

Also, when using a watcher task with problems matcher, diagnostics does not get replaced by task's diagnostics outputs when file is closed.
Not necessarily, you can check the video I have posted in https://github.com/microsoft/vscode/issues/256118#issuecomment-3645844851 to demonstrate diagnostics vanishing (ignore that there I blame `getDiagnostics()` for this, since now it is clear that the source is different). Another place I mention this issue is https://github.com/microsoft/vscode/pull/290278#issue-3853689195

I tried to fix it and found something, but I'm not sure how to complete it. Here are my results
1. First, I have found that for the file opened from EH, [ModelService.onModelRemoved](https://github.com/microsoft/vscode/blob/fd3565eb6bd06e688a98cabd029088c16390e62e/src/vs/editor/common/services/modelService.ts#L94) is not fired on close
2. The reason for this turned out to be that the `reference.counter` of model equals `2` when the [dispose](https://github.com/microsoft/vscode/blob/fd3565eb6bd06e688a98cabd029088c16390e62e/src/vs/base/common/lifecycle.ts#L699) is called on close, and that differs for the files opened from Explorer panel, which is `1` in that case
[...]

PR #292109, opened 2026-02-01
fix diagnostics not being repushed from problem matcher to markerServ…

n-gist:
When a file is opened from EH by `vscode.workspace.openTextDocument`, and closed after, the diagnostics originated from task problem matcher are not repushed to markerService on document auto-disposal after auto-dispose delay time has passed, due to incorrect uri comparison in WatchingProblemCollector

Investigation history https://github.com/microsoft/vscode/issues/291739#issuecomment-3830210992


jrieken (approved):
Thanks.

dbaeumer (approved):

PR #292198, opened 2026-02-01
Don't report using subagent that doesn't exist

### LINKED ISSUE #292196: Agent will report using subagent that doesn't exist
Image

Image

This subagent doesn't exist, so we should not show it being "used" in the UI.

If it doesn't exist, or we failed to actually resolve it for whatever random reason (#292180)

roblourens:
Fix #292196


aeschli:
https://github.com/microsoft/vscode/pull/292538 contains the full fix

PR #292201, opened 2026-02-01
Dileep y/291793

### LINKED ISSUE #291793: Sandbox doesn't work at all in remote WSL
Testing #290620

You're referencing the Windows binary on the Linux side:

Image

Image

dileepyavan:
fixes: #291793 

This PR fixes the path for srt command for remote scenarios.

Before the fix:

image

After the fix:
image

Validated all the scenarios in both local and remote.

Tested using private insiders: https://builds.code.visualstudio.com/builds/insider?commit=df6fc9c7675d38173f37bc970f0b2aee2328ebe1&dev=true

### DIFF
--- a/src/vs/workbench/contrib/terminalContrib/chatAgentTools/common/terminalSandboxService.ts
+++ b/src/vs/workbench/contrib/terminalContrib/chatAgentTools/common/terminalSandboxService.ts
@@ -7,9 +7,8 @@ import { VSBuffer } from '../../../../../base/common/buffer.js';
 import { Event } from '../../../../../base/common/event.js';
 import { Disposable } from '../../../../../base/common/lifecycle.js';
 import { FileAccess } from '../../../../../base/common/network.js';
-import { dirname, join } from '../../../../../base/common/path.js';
+import { dirname, posix, win32 } from '../../../../../base/common/path.js';
 import { OperatingSystem, OS } from '../../../../../base/common/platform.js';
-import { joinPath } from '../../../../../base/common/resources.js';
 import { URI } from '../../../../../base/common/uri.js';
 import { generateUuid } from '../../../../../base/common/uuid.js';
 import { IConfigurationChangeEvent, IConfigurationService } from '../../../../../platform/configuration/common/configuration.js';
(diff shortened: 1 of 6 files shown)


roblourens (approved):
Please try to give PRs a meaningful name

src/vs/workbench/contrib/terminalContrib/chatAgentTools/common/terminalSandboxService.ts
```diff
@@ -34,13 +34,17 @@ export interface ITerminalSandboxService {
...
+	private _srtPath: string | undefined;
+	private _srtPathResolved = false;
 	private _execPath?: string;
 	private _sandboxConfigPath: string | undefined;
 	private _needsForceUpdateConfigFile = true;
 	private _tempDir: URI | undefined;
 	private _sandboxSettingsId: string | undefined;
-	private _os: Promise<OperatingSystem>;
+	private _remoteEnvDetailsPromise: Promise<IRemoteAgentEnvironment | null>;
+	private _remoteEnvDetails: IRemoteAgentEnvironment | null = null;
```
Tyriar, line 45: Should just use `await this._remoteAgentService.getEnvironment()` whenever you need it, rely on the IRemoteAgentService to do any caching if necessary. Defer to debt issue?

src/vs/workbench/contrib/terminalContrib/chatAgentTools/common/terminalSandboxService.ts
```diff
@@ -34,13 +34,17 @@ export interface ITerminalSandboxService {
...
+	private _appRoot: string;
+	private _os: OperatingSystem = OS;
```
Tyriar, line 47: Same feedback here, this should just all be promise based instead of caching a sync value and initializing it as a potentially invalid value.

src/vs/workbench/contrib/terminalContrib/chatAgentTools/common/terminalSandboxService.ts
```diff
@@ -34,13 +34,17 @@ export interface ITerminalSandboxService {
...
+	private _srtPathResolved = false;
 	private _execPath?: string;
 	private _sandboxConfigPath: string | undefined;
 	private _needsForceUpdateConfigFile = true;
 	private _tempDir: URI | undefined;
 	private _sandboxSettingsId: string | undefined;
-	private _os: Promise<OperatingSystem>;
+	private _remoteEnvDetailsPromise: Promise<IRemoteAgentEnvironment | null>;
+	private _remoteEnvDetails: IRemoteAgentEnvironment | null = null;
+	private _appRoot: string;
```
Tyriar, line 46: No need to cache this?

Tyriar (approved):
Just some debt items that should be deferred for main only changes in debt week

PR #292326, opened 2026-02-02
Add getVisibleAreas api to view model

sandy081:
Add getVisibleAreas api to view model - This is needed for output editor features


sandy081:
@hediet That's right. Its a counter part api for `getHiddenAreas()`

PR #292377, opened 2026-02-02
Disable actions when no URL loaded or on error

### LINKED ISSUE #291216: State of actions when URL is empty is incorrect
Testing #290591

When URL is empty I observe:
Image

The following could be disabled when URL is empty:
- Add to chat
- Find in Page (when page is empty)
- Open in External Browser

jruales:
Closes https://github.com/microsoft/vscode/issues/291216

Disable certain actions when no URL is loaded or when there's an error

image

image

image

### Actions Requiring URL

| Action | Precondition | When Available |
|--------|--------------|----------------|
| **Add Element to Chat** | `chat.enabled` + `config.chat.sendElementsToChat.enabled` + `browserHasUrl` + `!browserHasError` | URL loaded successfully, no error |
| **Open in External Browser** | `browserHasUrl` | URL present (even with errors) |
| **Find in Page** | `browserHasUrl` + `!browserHasError` | URL loaded successfully, no error |

### Actions Not Requiring URL

| Action | Precondition | When Available |
|--------|--------------|----------------|
| **Open Integrated Browser** | None | Always |
| **New Tab** | None | Always |
| **Go Back** | `browserCanGoBack` | When navigation history exists |
| **Go Forward** | `browserCanGoForward` | When forward history exists |
| **Reload** | None | Always |
| **Focus URL Input** | None | Always (keybinding only) |
| **Toggle Developer Tools** | None | Always |
| **Clear Storage (Global/Workspace/Ephemeral)** | Storage scope matches | Always (visibility conditional on scope) |
| **Open Browser Settings** | None | Always |
| **Hide Find / Find Next / Find Previous** | Find widget visible/focused | When find widget is open |


src/vs/workbench/contrib/browserView/electron-browser/browserViewActions.ts
```diff
@@ -292,6 +292,8 @@ class OpenInExternalBrowserAction extends Action2 {
 			category: BrowserCategory,
 			icon: Codicon.linkExternal,
 			f1: false,
+			// Note: We do allow opening in an external browser even if there is an error page shown
+			precondition: CONTEXT_BROWSER_HAS_URL,
```
kycutler: Maybe better to have two context keys with the current URL and 'whether there is an active page'. Then we can just check if `url != ''` (or just if `url`) here and if `hasActivePage` in the other two places

kycutler: That gives us flexibility to redefine the behavior of "has an active page" in the future. e.g. we could have no URL but show a default newtab page that can be interacted with

jruales: Done

jruales: (please note: the URL here represent the currently loaded URL, regardless of whether the user has replaced any part of it in the URL bar. This is because when the user does an "Open in External Browser" action, what's opened is the currently loaded URL, not an unloaded typed URL)

PR #292523, opened 2026-02-03
Revert "Pass metadata to agent session commands"

bpasero:
This was not meant to be merged and is not ready.


alexr00 (changes requested):
Discussing offline. I don't think this needs to be reverted as it uses an established API pattern.

PR #292619, opened 2026-02-03
chat: initial rough draft of request queuing

connor4312:
Still a solid amount of UX polish to do, but the base data flows are here so I'd like to get this merged in to avoid conflicts.

Disabled by default behind the setting `"chat.requestQueuing.enabled": true`.

Refs #260330


src/vs/workbench/api/common/extHostChatAgents2.ts
```diff
@@ -623,7 +623,7 @@ export class ExtHostChatAgents2 extends Disposable implements ExtHostChatAgentsS
 
 		return detector.provider.provideParticipantDetection(
 			extRequest,
-			{ history },
+			{ history, yieldRequested: false },
```
roblourens, line 626: Can `yieldRequested` just be optional for all these callsites?

roblourens (approved):

PR #292672, opened 2026-02-03
fix: associate extHost lifecycle to window

### LINKED ISSUE #291346: Extension Host processes don't quit after window reload

- VS Code Version: 1.108.2
- OS Version: macOS 15.7.3

Steps to Reproduce:

1. Open project with GH Copilot and 3+ MCP
2. Use other AI extension if possible (optional)
3. See high CPU usage and more processed

- Reload does not work
- Disabled extension but does not work
- On Reload (no **Restart entire VSCode**) processes does not exit 

https://github.com/user-attachments/assets/060c97ec-e96a-4abe-b80d-cd216bc6fde8

deepak1556:
Fixes https://github.com/microsoft/vscode/issues/291346

With migration to render process reuse, our native-watchdog in the extension host cannot help with shutdown down of long running process on workbench reload. Instead we now rely on the main process to kill the extension host when the associated window is closed or reloaded.

### DIFF
--- a/src/vs/platform/extensions/electron-main/extensionHostStarter.ts
+++ b/src/vs/platform/extensions/electron-main/extensionHostStarter.ts
@@ -120,6 +120,7 @@ export class ExtensionHostStarter extends Disposable implements IDisposable, IEx
 			execArgv: opts.execArgv,
 			allowLoadingUnsignedLibraries: true,
 			respondToAuthRequestsFromMainProcess: true,
+			windowLifecycleBound: true,
 			correlationId: id
 		});
 		const pid = await Event.toPromise(extHost.onSpawn);
--- a/src/vs/platform/extensions/electron-main/extensionHostStarter.ts
+++ b/src/vs/platform/extensions/electron-main/extensionHostStarter.ts
@@ -121,6 +121,7 @@ export class ExtensionHostStarter extends Disposable implements IDisposable, IEx
 			allowLoadingUnsignedLibraries: true,
 			respondToAuthRequestsFromMainProcess: true,
 			windowLifecycleBound: true,
+			windowLifecycleGraceTime: 6000,
 			correlationId: id
 		});
 		const pid = await Event.toPromise(extHost.onSpawn);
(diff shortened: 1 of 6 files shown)


src/vs/platform/extensions/electron-main/extensionHostStarter.ts
```diff
@@ -120,6 +120,7 @@ export class ExtensionHostStarter extends Disposable implements IDisposable, IEx
 			execArgv: opts.execArgv,
 			allowLoadingUnsignedLibraries: true,
 			respondToAuthRequestsFromMainProcess: true,
+			windowLifecycleBound: true,
```
alexdima, line 123: Nice! I didn't know we supported this. We used to give extensions 5 seconds to gracefully deactivate themselves. Would this mechanism still allow for the extension host to:
1. Detect that it needs to shutdown and begin invoking deactivate() on extensions.
2. Await for those deactivate() promises up to 5 seconds and then quit

https://github.com/microsoft/vscode/blob/367250dacee043a25b3ee30ae1f0f6857167039a/src/vs/workbench/api/common/extHostExtensionService.ts#L271-L281

IIRC the watchdog mechanism used to kick in when the parent pid was dead for more than 5 seconds and then it would ungracefully terminate.

alexdima (dismissed):
Nice! I just left a comment about the graceful 5s behavior we used to have

src/vs/platform/extensions/electron-main/extensionHostStarter.ts
deepak1556: Ah I missed this overriden terminate path, was looking at the default path https://github.com/microsoft/vscode/blob/e09610ee655046f6d7786aa624660505a94d70a2/src/vs/workbench/api/node/extensionHostProcess.ts#L157-L161 

So we still have the graceful termination path hooked up to messageport close event https://github.com/microsoft/vscode/blob/e09610ee655046f6d7786aa624660505a94d70a2/src/vs/workbench/api/node/extensionHostProcess.ts#L191-L193, what we need now is a timeout on the main process code to support this path. Let me adjust the PR.

deepak1556: Addressed in e4fe4c7f6176464867fe3b98591bdf0effde3ad1

chrmarti (approved):

PR #292770, opened 2026-02-04
build: always repackage electron from oss entry scripts

deepak1556:
Follow-up to https://github.com/microsoft/vscode/pull/291948

Users having cached version of `.build/electron` will fail to pick up the renamed executable. Also, the downloads for a version are already cached, it isn't required to cache the packaging.

### DIFF
--- a/build/lib/electron.ts
+++ b/build/lib/electron.ts
@@ -229,15 +229,9 @@ function getElectron(arch: string): () => NodeJS.ReadWriteStream {
 }
 
 async function main(arch: string = process.arch): Promise<void> {
-	const version = electronVersion;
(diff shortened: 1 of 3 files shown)


bpasero (dismissed):

deepak1556:
Ah the issue was that the failing smoketests trigger our prelaunch task which repackage electron, the oss app was repackaged while it was running leading to SIGTRAP in https://github.com/electron/electron/blob/51a9101c3de7794baad9c35cce57adecf9ea3ad3/shell/browser/electron_browser_client.cc#L496-L510

The version check was preventing these unnecessary and harmful redownloads..

dbaeumer (approved):

PR #292905, opened 2026-02-04
Do not update agent status indicator bg color when debugging

joshspicer:
image


bpasero (dismissed):
This should probably be themable?

rebornix (dismissed):

alexr00 (approved):

Issue #291971, opened 2026-01-30
Agent prompt to answer question is not accessible


meganrogge:
When focus is in the accessible view, a user is unaware that a prompt has appeared that they needc to answer. We should alert, play the user confirmation required signal, and provide a command and keybinding to focus the element, including this in the accessibility help dialog.

The selector: `div.chat-question-carousel-container`

PR #292911, opened 2026-02-04
notify when chat questions come in

### LINKED ISSUE #292707: Provide OS notification when chat asks questions and accessibility signal
We should reuse `chat.notifyWindowOnConfirmation` and `accessibility.signals.chatUserActionRequired` for when chat asks a question

Already encountered this when I was working in two windows, figured one was still working, came back, realized questions had been asked.

meganrogge:
fixes #292707
part of https://github.com/microsoft/vscode/issues/291971

1. Plays `accessibility.signals.chatUserActionRequired` if that's enabled
2. Adds an `aria alert` for screen reader users
3. Adds an OS notification is the window is not focused, respecting `chat.notifyWindowOnConfirmation`

https://github.com/user-attachments/assets/3610ee8e-d966-4623-822c-380d4d9f9ccd

cc @justschen


justschen (approved):
nice, thanks thanks!

PR #292970, opened 2026-02-05
Show installation progress in update status bar entry on Windows

### LINKED ISSUE #293872: Update status bar tooltip should show progress during Install phase
When update is being installed on Windows, progress should be displayed on the update status bar entry and tooltip (currently just says "Installing update...")

dmitrivMS:
Fixes #293872

image


build/win32/code.iss
```diff
@@ -1365,6 +1365,36 @@ end;
...
+    if ProgressFilePath <> '' then begin
+      if MaxProgress > 0 then
+        CurrentPct := (CurProgress * 100) div MaxProgress
+      else
+        CurrentPct := 0;
+      if CurrentPct <> LastReportedProgressPct then begin
+        LastReportedProgressPct := CurrentPct;
+        ProgressContent := IntToStr(CurProgress) + ',' + IntToStr(MaxProgress);
+        SaveStringToFile(ProgressFilePath, ProgressContent, False);
+      end;
```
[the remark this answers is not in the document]
dmitrivMS: This is fine, client can default to 0 on its own.

src/vs/platform/update/electron-main/updateService.win32.ts
```diff
@@ -374,16 +377,64 @@ export class Win32UpdateService extends AbstractUpdateService implements IRelaun
...
+							if (this.state.currentProgress !== currentProgress || this.state.maxProgress !== maxProgress) {
+								this.setState(State.Updating(update, currentProgress, maxProgress));
+							}
+						}
+					}
+				} catch {
+					// Progress file may not exist yet or be locked, ignore
+				}
+
+				await timeout(500);
```
deepak1556, line 417: Should we revert to old polling interval of 1s or even longer, unsure if this is critical path for short polling interval.

deepak1556 (approved):
LGTM with nit

src/vs/platform/update/electron-main/updateService.win32.ts
dmitrivMS: 1 second should be fine, but I wouldn't make it shorter in case a user is actually looking at the progress bar.

dmitrivMS: Actually, thinking more about it - given the whole update often takes seconds, making it longer may create visually perceptive delay, so I'd like to keep it like this.

PR #293054, opened 2026-02-05
Start background update in Windows-less mode to avoid stealing focus

### LINKED ISSUE #142543: Background update on Windows steals focus

- VS Code Version: any
- OS Version: Windows 10/11 any

Steps to Reproduce:

1. Have background update option enabled.
2. Launch VSCode, start doing something.
3. VSCode detects an update, puts the timer overlay on Settings icon on the left rail, focus goes away from VSCode window.

This happens all the time across any VSCode and Windows 10/11 version. Great indicator of an update that had arrived, even though I specifically want it to happen in _background_.

dmitrivMS:
Fixes #142543


deepak1556 (changes requested):
Inno setup is built as a GUI app `/SUBSYSTEM:Windows`, lets wait for a repro before proceeding :)

src/vs/platform/update/electron-main/updateService.win32.ts
```diff
@@ -309,7 +309,8 @@ export class Win32UpdateService extends AbstractUpdateService implements IRelaun
 		const child = spawn(this.availableUpdate.packagePath, ['/verysilent', '/log', `/update="${this.availableUpdate.updateFilePath}"`, `/sessionend="${sessionEndFlagPath}"`, '/nocloseapplications', '/mergetasks=runcode,!desktopicon,!quicklaunchicon'], {
 			detached: true,
 			stdio: ['ignore', 'ignore', 'ignore'],
-			windowsVerbatimArguments: true
+			windowsVerbatimArguments: true,
+			windowsHide: true
```
joaomoreno, line 323: Are we sure this is doing anything? https://github.com/nodejs/node/issues/21825#issuecomment-2339770307

deepak1556: I was only thinking of console vs not-console application, good call on the detached. From https://learn.microsoft.com/en-us/windows/win32/procthread/process-creation-flags?redirectedfrom=MSDN

```
CREATE_NO_WINDOW0x08000000 | The process is a console application that is being run without a console window. Therefore, the console handle for the application is not set.This flag is ignored if the application is not a console application, or if it is used with either CREATE_NEW_CONSOLE or DETACHED_PROCESS.
```

PR #293070, opened 2026-02-05
Support begin and end tool calls for progress indicators without real tools in Chat (for background, cloud, etc)

DonJayamanne:
Currently when a tool is invoked from Background/Cloud/Claude, there's no way to display the UI that a tool invocation  has started, in fact there's a long delay of `working...` and then finally user sees what happend.
This fixes that by displaying the tool invocation and then eventually displaying the final result.
Applies to all tools, useful for Background/Cloud/Claude

Related PR in chat ext https://github.com/microsoft/vscode-copilot-chat/pull/3476/changes

Recording 2026-02-05 at 21 48 09


src/vs/workbench/api/common/extHostTypeConverters.ts
```diff
@@ -2897,120 +3010,11 @@ export namespace ChatToolInvocationPart {
...
-				terminalCommandState: data.state ? {
-					exitCode: data.state.exitCode,
-					duration: data.state.duration,
-				} : undefined,
-			};
-
-			return result;
-		} else if ('todoList' in data && Array.isArray(data.todoList)) {
-			// Convert extension API todo tool data to internal format
-			return {
```
DonJayamanne, line 2956: This is a small change, majority of the code has moved, has opposed to deleted

src/vs/workbench/api/common/extHostTypeConverters.ts
```diff
@@ -2868,19 +2868,132 @@ export namespace ChatResponseMovePart {
...
+		case types.ChatTodoStatus.InProgress:
+			return 'in-progress';
+		case types.ChatTodoStatus.Completed:
+			return 'completed';
+		default:
+			return 'not-started';
+	}
+}
+
+function todoStatusStringToEnum(status: string): types.ChatTodoStatus {
```
DonJayamanne, line 2963: This is a small change, majority of the code has moved, has opposed to deleted (see bellow)

dmitrivMS (approved):

lramos15 (approved):
Makes sense, long term I would like us to collapse begin,update,and complete into one function that just receives state info

PR #293104, opened 2026-02-05
Warn about distro commit and compat check early

alexr00:


joaomoreno (changes requested):
@copilot Let's move the initial distro + extension warning checks to a separate build stage, so that stage goes orange without blocking the build. Also only run those checks if it's a build we intend to release.

joaomoreno (approved):

rzhao271 (approved):

PR #293121, opened 2026-02-05
Plan mode keybinding

osortega:


chrmarti (dismissed):

meganrogge (changes requested):
This will really mess with things from an accessibilty perspective. At the least, this needs a `when` clause of `!screenReaderOptimized`

PR #293206, opened 2026-02-05
Polish modal editor UI

daviddossett:
- Add 4px gap between header action buttons for consistency with VS Code design
- Left-align header title instead of center alignment  
- Reduce max width from 1200px to 900px (70% viewport width)
- Update box shadow to use theme variable

## After
CleanShot 2026-02-05 at 13 41 46@2x

## Before
CleanShot 2026-02-05 at 13 41 08@2x


bpasero (changes requested):
@daviddossett I like this, but have a question around the change in `modalEditorPart.ts`: do we really want to reduce the size so much? There are larger editors showing in there, such as the one to manage workspace trust and models.

I am going ahead to bring in your CSS changes to spin a new insiders, but lets leave the size one open for discussion: https://github.com/microsoft/vscode/pull/293341

PR #293266, opened 2026-02-06
Fix chat cancel keybinding when clause

roblourens:

### DIFF
--- a/src/vs/workbench/contrib/chat/browser/actions/chatExecuteActions.ts
+++ b/src/vs/workbench/contrib/chat/browser/actions/chatExecuteActions.ts
@@ -840,10 +840,7 @@ export class CancelAction extends Action2 {
 			keybinding: {
 				weight: KeybindingWeight.WorkbenchContrib,
 				primary: KeyMod.CtrlCmd | KeyCode.Escape,
-				when: ContextKeyExpr.and(
-					ChatContextKeys.requestInProgress,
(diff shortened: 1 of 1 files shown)


src/vs/workbench/contrib/chat/browser/actions/chatExecuteActions.ts
```diff
@@ -840,10 +840,7 @@ export class CancelAction extends Action2 {
 			keybinding: {
 				weight: KeybindingWeight.WorkbenchContrib,
 				primary: KeyMod.CtrlCmd | KeyCode.Escape,
-				when: ContextKeyExpr.and(
-					ChatContextKeys.requestInProgress,
-					ChatContextKeys.remoteJobCreating.negate()
-				),
+				when: ChatContextKeys.inChatSession,
```
roblourens, line 844: @alexr00 I hope this change is ok- the previous change broke being able to cancel while a tool confirmation is active (in which case the request is not "in progress")

DonJayamanne (dismissed):

roblourens:
Actually looks like @connor4312 just fixed this!

src/vs/workbench/contrib/chat/browser/actions/chatExecuteActions.ts
alexr00: Sorry for the breakage and thank you @connor4312 for the fix!

PR #293422, opened 2026-02-06
Fix VSCode formatting settings

remcohaszing:
The VSCode settings of this project specify VSCode should autoformat JavaScript and TypeScript code on save. However, it doesn’t specify which formatter to use. That means it uses the formatter specified in user settings, which is not the intention.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -24,9 +24,11 @@
 		"files.insertFinalNewline": false
 	},
 	"[typescript]": {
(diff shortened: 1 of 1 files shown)


mjbvz:
These were intentionally removed so we can more easily dogfood tsgo. See #292635

mjbvz:
We haven't moved to tsgo entirely yet. Once we do, I'll add the setting

PR #293461, opened 2026-02-06
Disable inline suggestions for protected files

Baby-adi:
This PR adds an opt-in mechanism to disable inline and completion-based suggestions for user-defined protected files.

The goal is to allow users to mark sensitive files (e.g. `.env`) as human-only, preventing automated tooling from generating inline suggestions while keeping manual editing fully intact.

What’s included

- New `workspace.protectedFiles` configuration
- Support for `suggest: false` to disable inline suggestions
- Enforcement at the inline completions controller
- Glob-based matching with workspace-relative fallback
- Fully opt-in and backward compatible

Example:

```jsonc
"workspace.protectedFiles": {
  "**/.env": {
    "suggest": false
  }
}


This change applies to all inline and completion providers, not just AI tools. AI assistants are one motivating use case, but the mechanism is vendor-neutral and future-proof.


Baby-adi:

PR #293508, opened 2026-02-06
rm `any` from task files

meganrogge:
task part of #274723


src/vs/workbench/contrib/tasks/browser/abstractTaskService.ts
```diff
@@ -2846,8 +2838,7 @@ export abstract class AbstractTaskService extends Disposable implements ITaskSer
 		if (!result) {
 			return { config: undefined, hasParseErrors: false };
 		}
-		// eslint-disable-next-line local/code-no-any-casts
-		const parseErrors: string[] = (result as any).$parseErrors;
+		const parseErrors: string[] = (result as unknown as Record<string, unknown>).$parseErrors as string[];
```
roblourens, line 2841: I'd argue that this pattern of `as unknown as ...` is not an improvement over `any`. The goal is that we actually know what the type of a thing is instead of casting TS into submission

roblourens (approved):

src/vs/workbench/contrib/tasks/browser/abstractTaskService.ts
meganrogge: Oh, I see. I thought this still did type checking because saw many others addressed this with `unknown` and casting 

https://github.com/microsoft/vscode/pull/274802/changes

Also, a lot of task types are inherently unknown at compile time because they come from multiple dynamic sources. task configs are arbitrary JSON too. 

The alternative would be adding tons of boilerplate type predicates or creating overly broad union types that aren't actually more type-safe. In a task service where the whole point is handling arbitrary user-defined configurations, some casts seem ok.

roblourens: There are legit reasons to use unknown, your link is showing type guard functions which take an unknown and assert its type, and opaque args that are passed through a menu. Probably there are some of those in tasks too, I don't really know anythign about this code. 

But you have many instances here of doing a cast and assuming that some property exists on the object with a certain type. Ideally if you know that a certain property name might exist on an object then that would be expressed with a type. But I don't really know what this code is. I just thought maybe Copilot tried to make the warnings go away without fixing the underlying issue.

PR #293510, opened 2026-02-06
Add configurable hover title verbosity for editor tabs

### LINKED ISSUE #163166: File/Editor tab hover content customization
Ability to customize the colouring and information generated by hovering over a file tab. What it currently shows:
image

Compared with the hovering of the icons in the activity side bar, which currently looks like:
image

Incidentally the activity bar hover can be changed using "editorHoverWidget" but it then also applies to hovers inside the code editor too. There should be a separation so the activity bar can be specifically targeted i.e. "activityBar.hoverBackground" and "activityBar.hoverForeground".

Potential colour customizations for the file/editor tab hover in the primary image above.
- "tab.hoverBackground"
- "tab.hoverForeground"

Information customization:
- "tab.hoverInformation": "short"/"long"/"detail"

'Short' would be just the filename, 'long' would be what it currently shows (file path), and 'detail' would show file name on the first line and path on the second line. 

This is particularly important from a UX perspective, because when tabs are pinned if you want to see what file it is you either have to look at the breadcrumbs or hover, and if the file path is very long then on a large monitor your eyes are darting half the width of the screen. Incidentally the breadcrumbs itself could have a colour option to distinguish the active file from the rest of the path; what it currently shows:

image

The tab hover could also be brought more inline with the activity bar hover, which has a tiny triangle psuedo-element pointing to the element being hovered over as per image 2.

murataslan1:
Fixes #163166.

## Summary
- add a new setting `workbench.editor.tabHoverMode` with values `default`, `short`, `medium`, `long`
- map the configured tab hover mode to editor title verbosity when rendering editor tab hovers
- keep existing behavior by default (`default` maps to long/absolute path style)
- add a regression test for hover mode to verbosity mapping

## Test Plan
- [x] `npx tsx ./node_modules/mocha/bin/mocha --ui tdd src/vs/workbench/test/common/editor.test.ts --timeout 10000`


murataslan1:
Revived this PR on top of current `main` and force-pushed to `codex/issue-163166-tab-hover-mode` (commit `283c41952bb`).

Carried over the original changes:
- add `workbench.editor.tabHoverMode` (`default|short|medium|long`) setting
- map hover mode to editor title verbosity via `getEditorHoverVerbosity`
- use configured hover verbosity in editor tab hover titles
- add unit test coverage in `src/vs/workbench/test/common/editor.test.ts`

Note: I could not run the unit test suite in this environment because dependencies are not installed (`mocha: command not found`). Happy to rerun once deps are available.

PR #293511, opened 2026-02-06
chat: support resuming subagent sessions

bharatvansh:
This adds an opt-in way to resume a prior subagent run by threading a stable `subAgentInvocationId` through the built-in `runSubagent` tool and the proposed chat participant API.

Companion PR (Copilot Chat extension): https://github.com/microsoft/vscode-copilot-chat/pull/3532

### What changed
- **Proposed API** (`chatParticipantPrivate@12`): add `ChatRequest.resumeSubAgentInvocationId?: string` for resuming a previous subagent invocation.
- **runSubagent tool**: accept `resumeSubAgentInvocationId` input, prefer it when choosing the `subAgentInvocationId`, and forward it on the subagent request.
- Include the effective `subAgentInvocationId` in the tool result content (useful for trajectory tracking / resume).

### How to test
1. In Chat, invoke the `runSubagent` tool normally. Note the `subAgentInvocationId` reported in the tool result.
2. Invoke `runSubagent` again, passing that value as `resumeSubAgentInvocationId`.
   - With the companion Copilot Chat PR applied, the subagent should have access to its prior conversation history.

### Test coverage
- `npm run test-node`
- `npm run vscode-dts-compile-check`


roblourens:
This is interesting, but I'm not sure about the approach. I don't think we should store state as an entirely separate thing in the extension, but since subagents work like normal agents from the vscode side, I think they should also store state similarly in a ChatModel. Tbh, that will be hard to implement and I'm not that likely to take a PR that does that. We are kind of running into limitations of our "chat participant" architecture here. I would like to make this easier to implement with some refactoring at some point.

roblourens:
> would you be open to merging just the stable subAgentInvocationId and the UI grouping improvements

Can you explain more what this improves, I don't quite follow

PR #293512, opened 2026-02-06
chat: add drag-and-drop reordering for pending messages

### LINKED ISSUE #293473: Add ability to re-order queued messages
Add the ability to re-order queued messages

- Drag and drop
- Key bindings

connor4312:
Implements drag-and-drop functionality to reorder steering and queued
pending messages within their respective groups. When more than one
message of the same kind is pending, a gripper handle appears on hover
that enables dragging to reorder.

- Adds ChatPendingDragController to manage drag state, insertion
  indicators, and reorder logic via event delegation on the list
  container, keeping logic cleanly isolated from the renderer
- Attaches drag handles to pending request rows via the renderer
  with one-way binding; reorder operations call chatService.setPendingRequests
  which automatically triggers the full refresh chain
- Prevents cross-group reordering: steering messages stay within steering,
  queued messages stay within queued
- Adds CSS for drag handle visibility on hover, dragging state opacity,
  and insertion indicator positioning
- All changes compile cleanly with no layering violations

Fixes https://github.com/microsoft/vscode/issues/293473


TylerLeonhardt (approved):
Cool!

PR #293538, opened 2026-02-06
improve question carousel accessibility

### LINKED ISSUE #292917: Improve chat question carousel accessibility
The chat question carousel will soon be more accessible thanks to https://github.com/microsoft/vscode/pull/292911.

However, it still lacks an aria label and one can not move focus via keyboard to the container.

Additionally, focus automatically moves to the first response, which is against our usual pattern and could be very disruptive and confusing for screen reader users.

I'm thinking when in screen reader mode, we should not move focus. Since they're alerted, we should add a command and keybinding to focus the first response.

cc @accesswatch, @jooyoungseo, @rperez030

meganrogge:
fixes https://github.com/microsoft/vscode/issues/292917

- Disables `auto-focus` when in screen reader mode
- Adds `ctrlCmd+shift+A` keybinding to toggle focus between the question and the chat input to align with the confirmation prompt behavior
- Adds keyboard focus to the container and sets the aria label to be the the current question
- Alerts the question title when iterating through them
- Adds info about the command to the accessibility help dialog

https://github.com/user-attachments/assets/3982512e-df3d-4934-a338-c402d41a3ed3

https://github.com/user-attachments/assets/866ea3c4-3e2c-492d-b845-2196c7743265


justschen (approved):
awesome, thanks thanks!

PR #293551, opened 2026-02-06
Settings editor design update

cwebster-99:
image


MominRaza:
There will be no need to add custom button for json, if modal editor add support for showing custom actions from editor actions.
image

Settings page add a custom action for open json file in action editor.
image

Expected: If modal editor shows the custom action that settings page adds in action editor.
image

PR #293558, opened 2026-02-06
[duplicate] fix: enable text wrapping in parameter hints widget

### LINKED ISSUE #142888: Long parameter hints are not wrapped
When I hover over a class name/method etc in code, the displayed docstring is not wrapped. Long strings are displayed on a single line rather than being wrapped (method signatures are wrapped though). There is a vertical scroll bar but no horizontal so it's impossible to read long descriptions. I would post an image but I'm not sure how to take a screen region capture of a tooltip.

## Environment data

Version: 1.64.1
Commit: d6ee99e4c045a6716e5c653d7da8e9ae6f5a8b03
Date: 2022-02-07T17:17:47.733Z
Electron: 13.5.2
Chromium: 91.0.4472.164
Node.js: 14.16.0
V8: 9.1.269.39-electron.0
OS: Linux x64 5.4.0-99-generic

kairosci:
Add white-space: pre-wrap, word-wrap and overflow-wrap properties to parameter hints .code and .signature classes to allow long parameter descriptions and signatures to wrap instead of overflowing.

This fixes an issue where long parameter hints were not wrapped, making them unreadable when they extended beyond the visible area. Only vertical scrolling was available, not horizontal, which made navigation difficult.

Fixes #142888

### DIFF
--- a/src/vs/editor/contrib/parameterHints/browser/parameterHints.css
+++ b/src/vs/editor/contrib/parameterHints/browser/parameterHints.css
@@ -56,6 +56,8 @@
 .monaco-editor .parameter-hints-widget .signature {
 	padding: 4px 5px;
 	position: relative;
+	word-wrap: break-word;
+	overflow-wrap: break-word;
 }
 
 .monaco-editor .parameter-hints-widget .signature.has-docs::after {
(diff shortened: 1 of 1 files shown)


mjbvz:
Thanks for the PR. Looks like this may have already been fixed by https://github.com/microsoft/vscode/pull/292258

Please give it a try and submit a follow up PR if things still aren't working

PR #293583, opened 2026-02-07
Add /hooks slash command

roblourens:


bpasero (changes requested):
1 Copilot Feedback that is maybe relevant.

bpasero (approved):

Issue #234806, opened 2024-11-28
The VS Code Server with Remote SSH extension or Code Server experiences a memory leak issue


YuhongZeng:
Does this issue occur when all extensions are disabled?: No

- VS Code Version: In VSCode with Remote SSH extension , Or  Code Server
- OS Version: Connecting Windows to the Linux Server

Phenomenon
The VS Code Server with Remote SSH extension or Code Server experiences a memory leak issue

https://github.com/microsoft/vscode/blob/921f7ca0fc92817f6ab30ecc93850870cd0bd414/src/vs/base/parts/ipc/node/ipc.net.ts#L646 ，
The length of the array '_recordedInflateBytes' only grows over time and is not released unless the entry process is closed.

The snapshot information is as follows:
```
 PersistentProtocol  <- src\vs\base\parts\ipc\common\ipc.net.ts
  _socket: WebSocketNodeSocket -> 
    _flowManager:WebSocketFlowManager ->
      _zlibInflateStream: ZlibInflateStream ->
        _recordedInflateBytes: VSBuffer[] -> The length of the array only grows with time  (65411 items in heapsnap shot)
```

Steps to Reproduce:

1. In VSCode version 1.94.2 with Remote SSH extension , 
2. disable the 'Use Exec server' option to switch the VSCode server to a WebSocket connection type to reproduce the issue. 
3. Wait for one week and check the entry process.

1. Code Server.
2.  Wait for one week and check the entry process.

Thought：

The recording of inflated bytes results in a memory leak. The Recording inflated bytes is not necessary if the socket will not be sent to ExpansionHost process. The solution is to disable sending the socket to the ExtensionHost process and as a result, the need to write inflated bytes is eliminated. A named pipe can be used to communicate with the ExtensionHost instead of sending a socket.
https://github.com/microsoft/vscode/blob/389bf6c37fb5c3e177abe762656ea8728c2cc3a4/src/vs/server/node/extensionHostConnection.ts#L128
the “_canSendSocket” property is always true for non-Windows platforms. The named pipe is only used on Windows. Are there any problems or concerns?

YuhongZeng:
![Image](https://github.com/user-attachments/assets/962744df-95ee-4d94-b676-52629054a92f)
![Image](https://github.com/user-attachments/assets/6abd9847-70e8-4368-954a-5484b5da12fc)
![Image](https://github.com/user-attachments/assets/baf4403e-3437-4867-832f-30ba151e1546)

[...]

PR #293819, opened 2026-02-09
fix: stop unbounded websocket inflate-byte recording

SongXiaoXi:
This pull request adds the ability to dynamically enable or disable recording of inflated bytes in WebSocket communication, and ensures that this recording is properly disabled and cleaned up in key server-side scenarios. The main changes include introducing a new method for toggling this recording, propagating this control through the relevant classes, and improving resource cleanup during disposal.

I hit an unbounded memory growth issue in code-server (up to ~90GB RSS in 2 days) that appears to match the behavior described in microsoft/vscode#234806, and this PR includes a fix.

**Dynamic inflate byte recording control:**

* Added a `setRecordInflateBytes(record: boolean)` method to `WebSocketNodeSocket`, `WebSocketFlowManager`, and `ZlibInflateStream` to allow toggling the recording of inflated bytes at runtime. When disabled, any previously recorded bytes are cleared.

**Resource cleanup improvements:**

* Enhanced the `dispose()` methods of both `ZlibInflateStream` and `ZlibDeflateStream` to clear internal buffers and safely close zlib streams, guarding against errors during disposal.


alexdima (approved):
Excellent! Thank you!

TylerLeonhardt (approved):

PR #293829, opened 2026-02-09
support claude rules

aeschli:

### DIFF
--- a/src/vs/workbench/contrib/chat/browser/promptSyntax/attachInstructionsAction.ts
+++ b/src/vs/workbench/contrib/chat/browser/promptSyntax/attachInstructionsAction.ts
@@ -38,8 +38,8 @@ class ManageInstructionsFilesAction extends Action2 {
 	constructor() {
 		super({
 			id: CONFIGURE_INSTRUCTIONS_ACTION_ID,
-			title: localize2('configure-instructions', "Configure Instructions..."),
-			shortTitle: localize2('configure-instructions.short', "Chat Instructions"),
+			title: localize2('configure-instructions', "Configure Instructions & Rules..."),
+			shortTitle: localize2('configure-instructions.short', "Instructions & Rules"),
 			icon: Codicon.bookmark,
 			f1: true,
 			precondition: ChatContextKeys.enabled,
--- a/src/vs/workbench/contrib/chat/browser/promptSyntax/newPromptFileActions.ts
+++ b/src/vs/workbench/contrib/chat/browser/promptSyntax/newPromptFileActions.ts
@@ -26,6 +26,8 @@ import { askForPromptFileName } from './pickers/askForPromptName.js';
 import { askForPromptSourceFolder } from './pickers/askForPromptSourceFolder.js';
 import { IQuickInputService } from '../../../../../platform/quickinput/common/quickInput.js';
 import { getCleanPromptName, SKILL_FILENAME } from '../../common/promptSyntax/config/promptFileLocations.js';
+import { Target } from '../../common/promptSyntax/service/promptsService.js';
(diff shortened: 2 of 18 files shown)


src/vs/workbench/contrib/chat/browser/promptSyntax/attachInstructionsAction.ts
```diff
@@ -38,8 +38,8 @@ class ManageInstructionsFilesAction extends Action2 {
 	constructor() {
 		super({
 			id: CONFIGURE_INSTRUCTIONS_ACTION_ID,
-			title: localize2('configure-instructions', "Configure Instructions..."),
-			shortTitle: localize2('configure-instructions.short', "Chat Instructions"),
+			title: localize2('configure-instructions', "Configure Instructions & Rules..."),
```
pwang347, line 41: does this need to change? I thought we are just treating Rules as Instructions (compatibility-mode parsing from Claude)

src/vs/workbench/contrib/chat/browser/promptSyntax/newPromptFileActions.ts
```diff
@@ -150,23 +152,45 @@ function getDefaultContentSnippet(promptType: PromptsType, name: string | undefi
...
 			].join('\n');
 		case PromptsType.instructions:
-			return [
-				`---`,
-				`description: \${1:Describe when these instructions should be loaded}`,
-				`# applyTo: '\${1|**,**/*.ts|}' # when provided, instructions will automatically be added to the request context when the pattern matches an attached file`,
-				`---`,
-				`\${2:Provide project context and coding guidelines that AI should follow when generating code, answering questions, or reviewing changes.}`,
-			].join('\n');
+			if (target === Target.Claude) {
```
pwang347, line 155: same here, I wonder if it is necessary for us to be able to generate it (we just need to parse it if it's provided)

src/vs/workbench/contrib/chat/browser/promptSyntax/attachInstructionsAction.ts
aeschli: I thought it might help the understanding of what `Instructions` is for

src/vs/workbench/contrib/chat/browser/promptSyntax/newPromptFileActions.ts
aeschli: When creating a new instruction, you are asked in what folder you want to create. `.claude/rules` is one of these. When it is picked, we also need to create the corresponding format.
I don't think we should disallow creating rules

src/vs/workbench/contrib/chat/common/promptSyntax/languageProviders/promptValidator.ts
```diff
@@ -489,6 +498,36 @@ export class PromptValidator {
 		}
 	}
 
+	private validatePaths(attributes: IHeaderAttribute[], report: (markers: IMarkerData) => void): undefined {
```
pwang347, line 501: (same as above) do we need to lint the claude rules files?

aeschli: yes, these files can be opened and edited

DonJayamanne (approved):

src/vs/workbench/contrib/chat/browser/promptSyntax/attachInstructionsAction.ts
digitarald: I would not add the "rules" as its just "backwards compat". Our UI doesn't allow creating rules for example.

PR #293861, opened 2026-02-09
Add support for restart required extensions in the extensions view

mrleemurray:
Introduce functionality to notify users when extensions require a restart or window reload. This includes updates to the extensions view, new icons, and filtering capabilities for extensions based on their restart requirements.

Recording 2026-02-09 at 12 20 33


sandy081 (approved):
Looks good! 👍

PR #293913, opened 2026-02-09
fix: memory leak in ipc

SimonSiefke:
Fixes a memory leak in ipc. 

### Details

In ipc in the `onDidClientConnect` callback, two disposables are registered to `this.disposables`: 
1. `this.disposables.add(onFirstMessage(msg => {`
2.  `this.disposables.add(onDidClientDisconnect(() => {`

When a client connects and then disconnects, the disposables registered by 1. and 2. don't seem to be disposed. 

Example: When a new window is opened and closed, first the window client connects `onDidClientConnect`, then `onFirstMessage` and the `onDidClientDisconnect` listeners are set up. When the window is closed, `onDidClientDisconnect` fires. But the `onFirstMessage` and `onDidClientDisconnect` are not removed, only when the `IPCServer` disposes.

### Change

The change is not adding the disposables 1. and 2. to `this.disposables`. But instead to a separate `connectionDisposable`. The `connectionDisposable` is disposed when the client disconnects. 

That way, when a client connects and disconnects, no extra disposables remain in `this.disposables`

### Before 

When opening and closing a window 17 times, the number of `ChannelClient`, `ChannelServer` and `IpcMain` functions seems to grow each time:

shapes at 26-02-09 17 00 05

### After

When opening and closing a window 17 times, the number of `ChannelClient`, `ChannelServer` and `IpcMainImpl` functions stays constant:

window-open-new


joaomoreno (approved):
Awesome 👏

bpasero (approved):

PR #294002, opened 2026-02-09
add context menu actions to tips

### LINKED ISSUE #293507: add context menu to tips element
We could show `dismiss` which would cycle to the next one and `don't show again`

meganrogge:
fixes #293507

Adds a right-click context menu to the chat tips widget with two actions: `Dismiss this tip` (persists the dismissal to workspace storage and cycles to the next eligible tip) and `Disable Tips` (turns off the `chat.tips.enabled setting`).

The intent is:
- to allow users to hide tips as they learn
- to make it easy for power users to hide this if they find it annoying 

Screenshot 2026-02-09 at 1 51 56 PM


src/vs/workbench/contrib/chat/browser/widget/chatListRenderer.ts
```diff
@@ -1056,9 +1058,20 @@ export class ChatListItemRenderer extends Disposable implements ITreeRenderer<Ch
 		// Render tip above the request message (if available)
 		const tip = this.chatTipService.getNextTip(element.id, element.timestamp, this.contextKeyService);
 		if (tip) {
-			const tipPart = new ChatTipContentPart(tip, this.chatContentMarkdownRenderer);
+			const tipPart = new ChatTipContentPart(
```
roblourens, line 1061: Please use dependency injection

roblourens (approved):

PR #294019, opened 2026-02-09
Browser: DevTools protocol support

kycutler:


src/vs/platform/browserView/electron-main/browserViewMainService.ts
```diff
@@ -107,19 +83,123 @@ export class BrowserViewMainService extends Disposable implements IBrowserViewMa
...
+
+		await this.destroyBrowserView(target.id);
+		return true;
+	}
+
+	// Browser context management
+
+	getBrowserContexts(): string[] {
+		return BrowserSession.getBrowserContextIds();
+	}
```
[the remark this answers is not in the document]
kycutler: This is a good callout, but access control will be done in a separate PR

src/vs/platform/browserView/electron-main/browserViewMainService.ts
```diff
@@ -107,19 +83,123 @@ export class BrowserViewMainService extends Disposable implements IBrowserViewMa
...
+		return {
+			windowId: 1,
+			bounds: {
+				left: viewBounds.x,
+				top: viewBounds.y,
+				width: viewBounds.width,
+				height: viewBounds.height,
+				windowState: 'normal'
+			}
+		};
```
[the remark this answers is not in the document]
kycutler: Will handle this in a separate PR

src/vs/platform/browserView/common/cdp/types.ts
```diff
@@ -0,0 +1,189 @@
...
+	/** Event fired when the connection receives a CDP event */
+	readonly onEvent: Event<CDPEvent>;
+	/** Event fired when the connection is closed */
+	readonly onClose: Event<void>;
+
+	/**
+	 * Send a CDP command and await the result.
+	 * @param method The CDP method to call
+	 * @param params Optional parameters for the method
+	 * @param sessionId Optional session ID for targeting a specific session
```
jruales: Nit: Can we specify the following for clarity?
```suggestion
	 * @param sessionId Optional session ID for targeting a specific session
     *   Only used by browser-level connections to route commands to child sessions.
     *   Ignored for target-level connections, which are always bound to their own session.
```

src/vs/platform/browserView/common/cdp/proxy.ts
```diff
@@ -0,0 +1,362 @@
...
+
+	// #endregion
+
+	// #region Internal Helpers
+
+	/** Enumerate all targets from the service, caching each one and yielding its info */
+	private async *getTargetInfos(): AsyncIterable<CDPTargetInfo> {
+		for (const target of this.browserTarget.getTargets()) {
+			const info = await target.getTargetInfo();
+			this._targets.set(info.targetId, target);
```
kycutler: Note: having to get the `targetId` asynchronously makes things a bit difficult here and means we have to do ID caching and other workarounds.

Unfortunately we can't currently get it synchronously from Electron. Opened electron/electron#49733 to hopefully simplify this in the future.

src/vs/platform/browserView/electron-main/browserSession.ts
```diff
@@ -0,0 +1,202 @@
...
+					return BrowserSession.getOrCreateWorkspace(workspaceId, workspaceStorageHome);
+				}
+			// fallthrough -- no workspace context -> ephemeral
+			case BrowserViewStorageScope.Ephemeral:
+			default:
+				return BrowserSession.getOrCreateEphemeral(viewId);
+		}
+	}
+
+	// ---- Instance ----------------------------------------------------
```
jruales: nit: maybe turn these into `//#region` so that they show up in the minimap

src/vs/platform/browserView/common/cdp/types.ts
jruales, line 187:
```suggestion
	 * @param sessionId Optional session ID for targeting a specific session
	 *   Only used by browser-level connections to route commands to child sessions.
	 *   Ignored for target-level connections, which are always bound to their own session.
```

src/vs/platform/browserView/electron-main/browserSession.ts
```diff
@@ -0,0 +1,202 @@
...
+		readonly id: string,
+		/** The underlying Electron session. */
+		readonly electronSession: Electron.Session,
+		/** Resolved storage scope. */
+		readonly storageScope: BrowserViewStorageScope,
+	) {
+		super();
+
+		if (BrowserSession._sessions.has(id)) {
+			throw new Error(`BrowserSession with id '${id}' already exists`);
```
jruales, line 175: I was going to comment that the code never try/catches this Error, but each call to this constructor ensures that this never happens, so it's ok as an assertion assuming that this line will never be reached.

src/vs/platform/browserView/electron-main/browserSession.ts
```diff
@@ -0,0 +1,202 @@
...
+
+/**
+ * Holds an Electron session along with its storage scope and unique browser
+ * context identifier.  Each instance maps one-to-one to an Electron
+ * {@link Electron.Session} -- the {@link id} is derived from what makes the
+ * Electron session unique (scope + workspace), **not** from any view id.
+ * Multiple browser views may reference the same `BrowserSession`.
+ *
+ * The class centralises the permission configuration.  The {@link id}
+ * doubles as the CDP `browserContextId`.
```
jruales, line 28:
```suggestion
 * doubles as the CDP `browserContextId`.
 *
 * This class uses a private constructor with static factory methods
 * ({@link getOrCreate}, {@link getOrCreateGlobal}, etc.) and maintains
 * an internal registry of live sessions. Use the static methods to
 * obtain instances.
 */
```

src/vs/platform/browserView/electron-main/browserSession.ts
```diff
@@ -0,0 +1,202 @@
...
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { session } from 'electron';
+import { Disposable, IDisposable, toDisposable } from '../../../base/common/lifecycle.js';
+import { joinPath } from '../../../base/common/resources.js';
+import { URI } from '../../../base/common/uri.js';
+import { BrowserViewStorageScope } from '../common/browserView.js';
+
+// Same as webviews
```
jruales, line 12: Are we missing `'deprecated-sync-clipboard-read'`, which is included for webviews?
```ts
			// TODO(deepak1556): Should be removed once migration is complete
			// https://github.com/microsoft/vscode/issues/239228
			'deprecated-sync-clipboard-read',
```

Also, do we want to add a warning like Deepak did for webviews?
```ts
		//#region Security related measures (https://electronjs.org/docs/tutorial/security)
		//
		// !!! DO NOT CHANGE without consulting the documentation !!!
		//
```

src/vs/platform/browserView/electron-main/browserView.ts
```diff
@@ -568,7 +577,32 @@ export class BrowserView extends Disposable {
...
+	/**
+	 * Attach to receive debugger events.
+	 * @returns A connection that can be disposed to detach
+	 */
+	attach(): Promise<ICDPConnection> {
+		return this._debugger.attach();
+	}
+
 	override dispose(): void {
+		// Fire close event BEFORE disposing emitters
```
jruales: Is this related to the current PR, or fixing some other issue?

src/vs/platform/browserView/electron-main/browserViewDebugger.ts
```diff
@@ -0,0 +1,204 @@
...
+		if (this._electronDebugger.isAttached()) {
+			return;
+		}
+
+		this._electronDebugger.attach('1.3');
+		this._electronDebugger.on('message', this._messageHandler);
+	}
+
+	/**
+	 * Detach from the Electron debugger
```
jruales, line 152: Nit: can we move the definition `detachElectronDebugger()` so it's right before the definition of `dispose()`?

src/vs/platform/browserView/electron-main/browserViewDebugger.ts
```diff
@@ -0,0 +1,204 @@
...
+	}
+
+	/**
+	 * Discover the real Electron targetId for this WebContents
+	 */
+	private async discoverRealTargetId(): Promise<void> {
+		try {
+			const result = await this._electronDebugger.sendCommand('Target.getTargetInfo') as { targetInfo: CDPTargetInfo };
+			this._realTargetId = result.targetInfo.targetId;
+			this.logService.warn(`[BrowserViewDebugger] Could not find real targetId for WebContents ${this.view.webContents.id}`);
```
jruales: Is this warning correct? Isn't this the happy path?

src/vs/platform/browserView/electron-main/browserViewDebugger.ts
```diff
@@ -0,0 +1,204 @@
...
+	 * Initializes the debugger if not already done.
+	 */
+	async getTargetInfo(): Promise<CDPTargetInfo> {
+		// Ensure initialized
+		await this.initialize();
+
+		return {
+			targetId: this._realTargetId!,
+			type: 'page',
+			title: this.view.webContents.getTitle() || 'Browser View',
```
jruales: nit: Should we use a more user-friendly fallback title, since the user might not know the concept of a View?

src/vs/platform/browserView/electron-main/browserSession.ts
```diff
@@ -0,0 +1,202 @@
...
+		}
+		const storage = joinPath(workspaceStorageHome, workspaceId, 'browserStorage');
+		return new BrowserSession(sessionId, session.fromPath(storage.fsPath), BrowserViewStorageScope.Workspace);
+	}
+
+	/**
+	 * Get or create an ephemeral session for the given view / target id.
+	 */
+	static getOrCreateEphemeral(viewId: string, type?: string): BrowserSession {
+		const sessionId = `${type ?? 'ephemeral'}:${viewId}`;
```
jruales, line 110: Should we add validation to ensure `type` isn't one of `['workspace', 'ephemeral']`?

src/vs/platform/browserView/electron-main/browserViewDebugger.ts
```diff
@@ -0,0 +1,204 @@
...
+
+/**
+ * Wraps a browser view's Electron debugger with per-client session management.
+ *
+ * Each client gets their own Electron debugger session, providing true isolation
+ * just like connecting multiple DevTools clients to a real Chrome instance.
+ */
+export class BrowserViewDebugger extends Disposable implements ICDPTarget {
+
+	/** Map from Electron sessionId to the per-connection event emitter */
```
jruales: In the comments in this file, we say "Electron sessionId" and "Electron session", but shouldn't we instead say "CDP sessionId" and "CDP session"? Or as another possibility "CDP debug sessionId". This would be to avoid getting confused with an `Election.Session`

src/vs/platform/browserView/electron-main/browserViewMainService.ts
```diff
@@ -107,19 +83,127 @@ export class BrowserViewMainService extends Disposable implements IBrowserViewMa
...
+				await this.destroyBrowserView(view.id);
+			}
+		}
+
+		browserSession.dispose();
+	}
+
+	async attach(): Promise<ICDPConnection> {
+		return new CDPBrowserProxy(this);
+	}
```
jruales: Can we put this definition for `attach()` next to the one for `getTargetInfo()` since those are both part of `ICDPTarget`?

src/vs/platform/browserView/electron-main/browserViewMainService.ts
```diff
@@ -3,99 +3,75 @@
...
+	 * Delegates to {@link BrowserSession.isBrowserViewWebContents}.
+	 */
 	static isBrowserViewWebContents(contents: Electron.WebContents): boolean {
-		return BrowserViewMainService.knownSessions.has(contents.session);
+		return BrowserSession.isBrowserViewWebContents(contents);
 	}
 
 	private readonly browserViews = this._register(new DisposableMap<string, BrowserView>());
 
+	// ICDPTargetService events
```
jruales:
```suggestion
	// ICDPBrowserTarget events
```

src/vs/platform/browserView/electron-main/browserViewMainService.ts
```diff
@@ -107,19 +83,127 @@ export class BrowserViewMainService extends Disposable implements IBrowserViewMa
...
 
+		const view = this.createBrowserView(id, browserSession);
 		return view.getState();
 	}
 
 	tryGetBrowserView(id: string): BrowserView | undefined {
 		return this.browserViews.get(id);
 	}
 
+	// ICDPTargetService implementation
```
jruales:
```suggestion
	// ICDPBrowserTarget implementation
```

src/vs/platform/browserView/electron-main/browserViewMainService.ts
```diff
@@ -107,19 +83,127 @@ export class BrowserViewMainService extends Disposable implements IBrowserViewMa
...
+
+	getWindowForTarget(target: ICDPTarget): { windowId: number; bounds: CDPWindowBounds } {
+		if (!(target instanceof BrowserView)) {
+			throw new Error('Can only get window for targets created by this service');
+		}
+
+		const view = target.getWebContentsView();
+		const viewBounds = view.getBounds();
+		return {
+			windowId: 1,
```
jruales, line 126: The windowId will always be 1 even when there's a separate VS Code window?

src/vs/platform/browserView/electron-main/browserViewMainService.ts
```diff
@@ -107,19 +83,127 @@ export class BrowserViewMainService extends Disposable implements IBrowserViewMa
...
+		}
+
+		await this.destroyBrowserView(target.id);
+		return true;
+	}
+
+	// Browser context management
+
+	getBrowserContexts(): string[] {
+		return BrowserSession.getBrowserContextIds();
```
jruales, line 191: Shouldn't this filter to only IDs that start with `cdp-created:`, similarly to the use of that prefix in `createBrowserContext()` and `disposeBrowserContext()`?

src/vs/platform/browserView/common/cdp/proxy.ts
```diff
@@ -0,0 +1,362 @@
...
+
+	// CDP method handlers map
+	private readonly _handlers = new Map<string, (params: unknown, sessionId?: string) => Promise<object> | object>([
+		// Browser.* methods
+		['Browser.getVersion', () => this.browserTarget.getVersion()],
+		['Browser.close', () => ({})],
+		['Browser.setDownloadBehavior', () => ({})],
+		['Browser.getWindowForTarget', (p, s) => this.handleBrowserGetWindowForTarget(p as { targetId?: string; sessionId?: string }, s)],
+		['Browser.setWindowBounds', () => ({})],
+		// Target.* methods
```
jruales:
```suggestion
		// Target.* methods (https://chromedevtools.github.io/devtools-protocol/1-3/Target/)
```

src/vs/platform/browserView/common/cdp/proxy.ts
```diff
@@ -0,0 +1,362 @@
...
+	readonly onClose: Event<void> = this._onClose.event;
+
+	// CDP method handlers map
+	private readonly _handlers = new Map<string, (params: unknown, sessionId?: string) => Promise<object> | object>([
+		// Browser.* methods
+		['Browser.getVersion', () => this.browserTarget.getVersion()],
+		['Browser.close', () => ({})],
+		['Browser.setDownloadBehavior', () => ({})],
+		['Browser.getWindowForTarget', (p, s) => this.handleBrowserGetWindowForTarget(p as { targetId?: string; sessionId?: string }, s)],
+		['Browser.setWindowBounds', () => ({})],
```
jruales: `Browser.setDownloadBehavior`, `Browser.getWindowForTarget`, and `Browser.setWindowBounds` are not in the 1.3 spec (https://chromedevtools.github.io/devtools-protocol/1-3/Browser/) but rather only in the tip-of-tree spec (https://chromedevtools.github.io/devtools-protocol/tot/Browser/). Given that we target 1.3 in our debugger/CDP connections, should we remove those methods from here?
Similarly, there are some methods that that are part of 1.3 but are missing here. Namely, `Browser.addPrivacySandboxCoordinatorKeyConfig`, `Browser.addPrivacySandboxEnrollmentOverride`, and `Browser.resetPermissions`

Same for Targets below. `Target.attachToBrowserTarget` and `Target.getTargetInfo` are not part of 1.3 (so we can remove them from here), and `Target.activateTarget` and `Target.detachFromTarget` *are* part of 1.3 (so we should add them here)

src/vs/platform/browserView/common/cdp/proxy.ts
jruales:
```suggestion
		// Target.* methods (https://chromedevtools.github.io/devtools-protocol/tot/Target/)
```

src/vs/platform/browserView/common/cdp/proxy.ts
```diff
@@ -0,0 +1,362 @@
...
+
+	// Events to external client (ICDPConnection)
+	private readonly _onEvent = this._register(new Emitter<CDPEvent>());
+	readonly onEvent: Event<CDPEvent> = this._onEvent.event;
+	private readonly _onClose = this._register(new Emitter<void>());
+	readonly onClose: Event<void> = this._onClose.event;
+
+	// CDP method handlers map
+	private readonly _handlers = new Map<string, (params: unknown, sessionId?: string) => Promise<object> | object>([
+		// Browser.* methods
```
jruales:
```suggestion
		// Browser.* methods (https://chromedevtools.github.io/devtools-protocol/tot/Browser/)
```

src/vs/platform/browserView/common/cdp/proxy.ts
```diff
@@ -0,0 +1,362 @@
...
+	// targetId -> sessionId (tracks which targets are attached)
+	private readonly _targetSessions = new Map<string, string>();
+
+	// Events to external client (ICDPConnection)
+	private readonly _onEvent = this._register(new Emitter<CDPEvent>());
+	readonly onEvent: Event<CDPEvent> = this._onEvent.event;
+	private readonly _onClose = this._register(new Emitter<void>());
+	readonly onClose: Event<void> = this._onClose.event;
+
+	// CDP method handlers map
```
jruales, line 32: Should we also support `Schema.getDomains` so that we can let Playwright know which domains (e.g., Browser., Target.) we support, as well as which commands within those domains we support? Given that we don't provide that, would Playwright get confused? Or which domains and commands does it assume by default in that case?

src/vs/platform/browserView/common/cdp/proxy.ts
jruales: Never mind. Seems like actually Electron doesn't exactly adhere to 1.3 or to T.O.T., but rather whichever version the specific version of Chromium provides, which is something in between.

jruales (dismissed):

src/vs/platform/browserView/electron-main/browserSession.ts
kycutler: I think we should remove clipboard-read, will do in a separate PR

src/vs/platform/browserView/electron-main/browserView.ts
kycutler: It's related. Good callout though -- I actually needed to rework things here a bit more. It's essentially just making sure teardown happens in the proper order, so events emitted during disposals are correctly ordered as well

src/vs/platform/browserView/electron-main/browserViewDebugger.ts
kycutler: Whoops, yeah this was from an old version of the logic. Removed, thanks!

src/vs/platform/browserView/electron-main/browserViewDebugger.ts
kycutler: Updated to fallback to the URL

src/vs/platform/browserView/electron-main/browserSession.ts
kycutler: Good point, done!

src/vs/platform/browserView/electron-main/browserViewDebugger.ts
kycutler: Fair enough -- updated the "Electron *" references to generally reference CDP instead to reduce confusion

src/vs/platform/browserView/electron-main/browserViewMainService.ts
kycutler: At the moment yes. Better window management will be done in a separate PR as it requires some refactoring to the views

src/vs/platform/browserView/electron-main/browserViewMainService.ts
kycutler: No -- we want debuggers to be able (in theory) to create new pages within existing contexts. Of course, we will want to limit this / provide user controls, but that will be done at the proxy layer

src/vs/platform/browserView/common/cdp/proxy.ts
kycutler: The Schema domain is deprecated -- will leave it unimplemented

src/vs/platform/browserView/common/cdp/proxy.ts
kycutler: Yeah it seems like in practice, tools actually depend on a lot more than just the 1.3 spec. The additional methods are used by Playwright so we need them for that to work.

That said, some methods from 1.3 were in fact missing. Updated to make sure they are all supported (and implemented except for one -- activateTarget -- which I'll do in a follow-up PR)

src/vs/platform/browserView/common/cdp/types.ts
kycutler: It's a reasonable clarification, but I don't really like putting the implementation specifics within the common interface. I'll leave this one as-is

src/vs/platform/browserView/common/cdp/proxy.ts
```diff
@@ -0,0 +1,396 @@
...
+			}
+			if (this._autoAttach) {
+				await this.attachToTarget(targetInfo.targetId, true);
+			}
+		});
+		this._targets.onDidUnregisterTarget(async ({ targetInfo }) => {
+			if (this._discover) {
+				this.sendBrowserEvent('Target.targetDestroyed', { targetId: targetInfo.targetId });
+			}
+		});
```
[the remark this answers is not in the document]
kycutler: The target manager is itself registered, so they will all be cleaned up when the proxy is

Yoyokrazy (dismissed):

DonJayamanne (approved):

Issue #34442, opened 2017-09-15
Extract the integrated terminal


lloeki:
"Feature" request.

The integrated terminal grows in functionality, to the point that it could soon become very useful standalone, as its *own app*. I'm only a casual VScode user (using vim most of the time), but I would definitely be interested in a well-behaved, consistent, reactively-configurable, multi-platform terminal not tied to the editor.

Obviously if such a situation were to unfold, the shared component (not entirely dissimilar to VTE) should be jointly developed and shared with VScode. Both could be able to leverage a common configuration so that whether a terminal leveraging such a theoretical common component is spawned within VScode or standalone, they behave the same.

lloeki:
Hint, hint, call the thing VSterm?

Tyriar:
This is an idea I've thrown around a bit, I doubt we'll get time to work on it any time soon. The first steps for this would be more flexible movement of the panel inside the workbench and pulling it into its own window.

auchenberg:
You should check out https://hyper.is/, also Electron based and extensible.

lloeki:
@auchenberg I definitely know about Hyper ;) yet Code's implementation is currently much more solid though, and given each one's track record in stability and performance I'd rather bet on Code. Also having the external terminal and Code's integrated one behave in exactly the same way is a non-negligible feature.

auchenberg:
@lloeki Hyper is migrating to xterm.js build by @Tyriar, which means that Hyper most likely will be similar to the VS Code console. I can't speak on behalf of @Tyriar, but to me it sounds like most of the concerns are addressed by this migration. https://github.com/zeit/hyper/issues/1275

Tyriar:
Yes Hyper is planning on moving to xterm.js for v2. There will be some differences such as configuration and some things are handled differently (clear in Hyper may not work on Windows for example), overall it will be very close to VS Code though.

warpdesign:
I can't speak for v2 but the last time I tried Hyper at was really slow & buggy, especially under Windows.

VSCode's terminal implementation worked so good that I don't use cmd.exe anymore.

I understand it's not the priority but if I was to try to extract VSCode's terminal and make it run without VSCode where should I start?

Tyriar:
@warpdesign what I was thinking was to be able to launch the terminal by itself using `code --term` or something similar, this would involve:

- Create the "terminal window" which has access to the services used internally by the regular VS Code window (this part would be particularly tricky)
- Modifying the CLI to support this
[...]

PR #294035, opened 2026-02-09
Add terminal-only mode and UI adjustments

darylcecile:
[EXPERIMENTAL] https://github.com/microsoft/vscode/issues/34442

**What**
Introduce a terminal-only mode driven by a new --terminal CLI flag and environmentService. Adds 'terminal' to NativeParsedArgs and OPTIONS (-t alias) and exposes terminal on browser/electron/common environment services. 

image

**Why**
Im a fan of the vscode terminal and wish there was a way i could open just that minus all the other distractions. Now I can do `code --terminal`.

**How**
Using existing features (ContextKeyExpr, LayoutClasses, shouldBeHidden, etc.) and guards to disable the "other distractions" and only showing the terminal. Loads everything else as normal so _in theory_ everything else should still work. Introduces TerminalModeContext and uses it across workbench to adjust behavior: hide editors/activity/side/auxiliary bars, force panel visibility/focus, block panel hiding/showing of auxiliary bar, hide non-terminal panel view containers, suppress panel title actions, and disable command center and startup page in terminal mode. Adds a `.terminal-mode` CSS class to hide per-composite toolbar/global actions. 

> [!IMPORTANT]
> This is more a **proof of concept** (and for **personal use**) at this stage. I have not tested the impact of this on the rest of the codebase 🤷🏽

### DIFF
--- a/src/vs/platform/environment/common/argv.ts
+++ b/src/vs/platform/environment/common/argv.ts
@@ -142,6 +142,7 @@ export interface NativeParsedArgs {
 	'enable-rdp-display-tracking'?: boolean;
 	'disable-layout-restore'?: boolean;
 	'disable-experiments'?: boolean;
+	'terminal'?: boolean;
 
 	// chromium command line args: https://electronjs.org/docs/all#supported-chrome-command-line-switches
 	'no-proxy-server'?: boolean;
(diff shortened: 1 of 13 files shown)


Tyriar (commented):
This is cool but it's a pretty long way off the changes that would actually be needed to accomplish this. What this does is forcefully strips away some functionality based on some args which isn't good for the codebase in the long term. My hope is that such a feature would actually launch basically an existing "Terminal Window" which you can open with this command:

Image

This will open the terminal in a special minimal state, the downside being you can't launch it via the CLI and it needs to come from a parent workbench window:

Image

Docs: https://code.visualstudio.com/docs/terminal/basics#_terminals-in-new-windows

anthonykim1:
Thanks for the PR. As discussed, this approach strips down the main workbench instead of building on the existing Terminal Window architecture. Closing this draft, while the broader feature request remains tracked in #34442. Please let us know if you think this is incorrect or are still interested in another attempt!

PR #294087, opened 2026-02-10
Preserve update metadata over application and OS restarts

dmitrivMS:
Fixes the issue where previously downloaded update shows up as `Latest version: unknown (unknown)` after OS reboot and reopening VS Code.


jrieken (approved):

deepak1556 (approved):
Thanks!

Issue #292934, opened 2026-02-04
Clean up JS/TS language specific setting settings


mjbvz:
The JS/TS settings were written before many modern configuration options, such as language scoping, were supported. As a result, we have duplicate versions of many settings: `typescript.*` and `javascript.*`

If we were starting fresh, all of these would be under a single top level scope (`js/ts`) and we'd use language specific settings instead. With TS 7.0 coming up, I think it's time we bite the bullet and finally try to clean things up.

## What this means for users

Existing configurations such as:

```json
"javascript.format.enable": false,
"typescript.format.enable": true,
```

Will become:

```json
"[javascript,javascriptreact]": {
	"js/ts.format.enable": false,
},
"[typescript,typescriptreact]": {
	"js/ts.format.enable": true,
}
```

If the values are the same for both js and ts, you can just set: `"js/ts.format.enable": false`

## Plan

1. February: Add new `js/ts.*` versions of the settings

    - Make old settings deprecated
    - Keep respecting old settings if they are set (however new settings override them)
    - Make sure new settings descriptions show up if you search either "javascript" or "typescript" in the setting ui (may require a small update to the setting description
    - In a limited number of cases, we may also want to align names (e.g. `enabled` vs `enable`)

1. February: For tsgo specifically, start reading the new settings, with fallback logic just to the `typescript.*` version of the (today is only checks `typescript.*`, not `javascript.*`)

1. March or April: start auto migrate these settings for users. We don't want to do this immediately so that users have more time to make sure they have picked up a VS Code version that supports the new values

    - We likely only want to do this for user settings to avoid generating extra changes in workspaces 

1. Many months out: remove observing the old settings ONLY in tsgo

     - We will keep respecting the old settings in VS Code for at least a year and possibly longer to support old workspace settings we encounter

PR #294221, opened 2026-02-10
Allow contributed configurations to define additional "search terms"

mjbvz:
For #292934

With the current settings, the only way I can get my settings to show up correctly for searches such as "javascript" or "typescript" is the include those words in the setting description. This can make the setting description less clear. I also tried a setting name such as  `javascript/typescript.mySetting`, but this is pretty long and still doesn't show up correctly when searching

This PR proposes a new "searchTerms" field for configurations to solve this. The search terms are basically just added onto the description for search, but are not show to the user. This will let us use a more readable setting id and description for #292934

@sandy081 @rzhao271 Let me know what you think about this proposal. Happy to try other approaches too if you have any suggestions

### DIFF
--- a/src/vs/workbench/contrib/preferences/browser/preferencesSearch.ts
+++ b/src/vs/workbench/contrib/preferences/browser/preferencesSearch.ts
@@ -242,9 +242,9 @@ export class SettingMatches {
 		// Search the description if we found non-contiguous key matches at best.
 		const hasContiguousKeyMatchTypes = this.matchType >= SettingMatchType.ContiguousWordsInSettingsLabel;
 		if (this.searchDescription && !hasContiguousKeyMatchTypes) {
-			// Search the description lines and any additional search terms.
-			const searchableLines = setting.searchTerms?.length
-				? [...setting.description, setting.searchTerms.join(' ')]
+			// Search the description lines and any additional keywords.
+			const searchableLines = setting.keywords?.length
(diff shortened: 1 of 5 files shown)


Yoyokrazy (approved):

src/vs/workbench/contrib/preferences/browser/preferencesSearch.ts
```diff
@@ -242,10 +242,13 @@ export class SettingMatches {
...
+				? [...setting.description, setting.keywords.join(' ')]
+				: setting.description;
 			for (const word of queryWords) {
-				// Search the description lines.
-				for (let lineIndex = 0; lineIndex < setting.description.length; lineIndex++) {
-					const descriptionMatches = matchesBaseContiguousSubString(word, setting.description[lineIndex]);
+				for (let lineIndex = 0; lineIndex < searchableLines.length; lineIndex++) {
+					const descriptionMatches = matchesBaseContiguousSubString(word, searchableLines[lineIndex]);
 					if (descriptionMatches?.length) {
 						descriptionMatchingWords.set(word, descriptionMatches.map(match => this.toDescriptionRange(setting, match, lineIndex)));
```
rzhao271, line 253: Note to self that the range might be invalid if it corresponds to the line of keywords, but that I'm unsure whether the actual range values are used at all these days.

rzhao271 (approved):

PR #294259, opened 2026-02-10
Remove `IToolInvocationContext.sessionId`

mjbvz:
For #274403

We have the sessionResource that should be used instead. Confirmed that this does not seem to be used in copilot

### DIFF
--- a/src/vs/workbench/contrib/mcp/common/mcpServer.ts
+++ b/src/vs/workbench/contrib/mcp/common/mcpServer.ts
@@ -30,6 +30,7 @@ import { IEditorService } from '../../../services/editor/common/editorService.js
 import { IWorkbenchEnvironmentService } from '../../../services/environment/common/environmentService.js';
 import { IExtensionService } from '../../../services/extensions/common/extensions.js';
 import { IOutputService } from '../../../services/output/common/output.js';
+import { chatSessionResourceToId } from '../../chat/common/model/chatUri.js';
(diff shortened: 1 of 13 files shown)


DonJayamanne (dismissed):

roblourens (dismissed):
Sweet

mjbvz:
@connor4312 Can you please take a look at the MCP related changes to see if they make sense. Seems to work in my testing but not very familiar with the code

src/vs/workbench/contrib/mcp/common/mcpServer.ts
```diff
@@ -1069,8 +1070,8 @@ export class McpTool implements IMcpTool {
 			}
 
 			const meta: Record<string, unknown> = { progressToken };
-			if (context?.chatSessionId) {
-				meta['vscode.conversationId'] = context.chatSessionId;
+			if (context?.chatSessionResource) {
+				meta['vscode.conversationId'] = chatSessionResourceToId(context.chatSessionResource);
```
mjbvz, line 1074: @connor4312 This is one area where I'm not sure how the value is used. Is it just for telemetry or correlation purposes? The `chatSessionResourceToId` should preserve the id if that's the case

connor4312: Just correlation as an opaque id

connor4312 (approved):

PR #294592, opened 2026-02-11
Add experimental color themes and configuration defaults

mrleemurray:
Introduce experimental color themes and set configuration defaults for both dark and light themes.

This update allows users to select "Experimental Dark" and "Experimental Light" as their preferred themes. The changes also include necessary adjustments to the configuration registry to support these new themes. Testing can be done by selecting the new themes in the settings and verifying their appearance in the application.


bpasero:
We know from past testing that changing config defaults on startup can have a negative impact on performance, maybe @sandy081 has a different idea how to achieve this.

PR #294629, opened 2026-02-11
Feature: Claude Customizations

pwang347:

### DIFF
--- a/src/vs/workbench/contrib/chat/common/promptSyntax/config/promptFileLocations.ts
+++ b/src/vs/workbench/contrib/chat/common/promptSyntax/config/promptFileLocations.ts
@@ -87,6 +87,11 @@ export const AGENTS_SOURCE_FOLDER = '.github/agents';
 export const CLAUDE_AGENTS_SOURCE_FOLDER = '.claude/agents';
 
 /**
+ * Claude rules folder.
+ */
+export const CLAUDE_RULES_SOURCE_FOLDER = '.claude/rules';
+
+/**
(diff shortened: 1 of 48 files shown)


src/vs/workbench/contrib/chat/common/promptSyntax/config/promptFileLocations.ts
```diff
@@ -33,6 +35,26 @@ export const AGENT_FILE_EXTENSION = '.agent.md';
...
+/**
+ * Claude local file name.
+ */
+export const CLAUDE_LOCAL_MD_FILENAME = 'CLAUDE.local.md';
+
+/**
+ * Claude configuration folder name.
+ */
+export const CLAUDE_CONFIG_FOLDER = '.claude';
+
```
izzyizzat10, line 57: We now have a mix of USE_CLAUDE_MD, USE_AGENT_MD, and Copilot-specific names.
Should we standardize on one abstraction level (e.g., AGENT_*) to avoid hard-coding vendor names in config?

sbatten (dismissed):

src/vs/workbench/contrib/chat/common/promptSyntax/config/promptFileLocations.ts
pwang347: Hi @izzyizzat10, these are just internal references (cherry-picked from https://github.com/microsoft/vscode/pull/292898) and not top-level config

sbatten (approved):

PR #294715, opened 2026-02-11
add `chat.autoReply`

### LINKED ISSUE #294714: Add `chat.autoReply` to question carousel feature/setting
Will unblock https://github.com/microsoft/vscode-copilot-evaluation/issues/2528#issuecomment-3886053510

meganrogge:
fixes #294714

Enables questions to be responded to when in YOLO mode. Previously, we just skipped those so that evals worked.

Now, for evals (or users), we have an auto reply feature, which only runs if `chat.autoReply` is enabled and the user has opted in via a dialog (stored in application storage).

Model selection: uses the current widget model name to select a concrete model id (exact id first, then Copilot family match).

Prompting: builds a JSON-only prompt with question metadata and optional original request text, then asks the model; if parsing fails, retries with strict JSON instructions.

Parsing: per question, resolves text/singleSelect/multiSelect answers, matching options by index, id, label, or partial label; invalid or empty values are dropped.

Merging: any question with an explicit default keeps that default; otherwise model answers are used; remaining gaps use deterministic fallbacks (first option or freeform “OK”/request text).

Screenshot 2026-02-11 at 4 50 41 PM

Screenshot 2026-02-11 at 4 50 28 PM

https://github.com/user-attachments/assets/b5ca71f5-3aff-4788-a307-5c23dcf382a6

### DIFF
--- a/src/vs/workbench/contrib/chat/browser/actions/chatActions.ts
+++ b/src/vs/workbench/contrib/chat/browser/actions/chatActions.ts
@@ -51,7 +51,7 @@ import { ChatContextKeys } from '../../common/actions/chatContextKeys.js';
 import { ModifiedFileEntryState } from '../../common/editing/chatEditingService.js';
 import { IChatModel, IChatResponseModel } from '../../common/model/chatModel.js';
 import { ChatMode, IChatMode, IChatModeService } from '../../common/chatModes.js';
-import { IChatService } from '../../common/chatService/chatService.js';
+import { ElicitationState, IChatService, IChatToolInvocation } from '../../common/chatService/chatService.js';
 import { ISCMHistoryItemChangeRangeVariableEntry, ISCMHistoryItemChangeVariableEntry } from '../../common/attachments/chatVariableEntries.js';
 import { IChatRequestViewModel, IChatResponseViewModel, isRequestVM } from '../../common/model/chatViewModel.js';
 import { IChatWidgetHistoryService } from '../../common/widget/chatWidgetHistoryService.js';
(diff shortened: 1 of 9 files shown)


rwoll (changes requested):
This appears to still be triggering a confirmation. The prompt I'm using is `use ask_questions tool to see if I want to run with sleep 30s or sleep 60s, then run the sleep command`.

While it does auto-accept, workbench.action.chat.open#blockOnResponse is returning a confirmation status instead of waiting for the turn to finish (i.e. sleep, etc).

rwoll:
> This appears to still be triggering a confirmation. The prompt I'm using is `use ask_questions tool to see if I want to run with sleep 30s or sleep 60s, then run the sleep command`.
> 
> While it does auto-accept, workbench.action.chat.open#blockOnResponse is returning a confirmation status instead of waiting for the turn to finish (i.e. sleep, etc).

@meganrogge - here's a minimal repro: https://github.com/microsoft/vscode/compare/merogge/auto-reply-chat...rwoll/wip-repro-chat-confirmation?expand=1.

1. Checkout my branch.
2. Start the Debugger.
3. Set chat.autoReply to true.
4. Trigger `Open Chat (Auto Reply Test)` in the command pallette.
5. Observe editor opens with `{ "type": "confirmation" }`.

We instead expect it to have an LLM response, etc.

rwoll (dismissed):
LGTM (and I tested this fixed the eval issue), but I'm less familiar with this section of code so I suggest a review from @karthiknadig, @bpasero, or someone else who's worked on askQuestion.

karthiknadig (dismissed):

karthiknadig (approved):

PR #294881, opened 2026-02-12
Add compact layout support for Activity Bar

mrleemurray:
Introduce a compact layout option for the Activity Bar, allowing users to adjust dimensions and icon sizes. Refactor existing styles to utilize CSS variables for better customization and add unit tests to ensure proper functionality of the new layout features.

Recording 2026-02-12 at 12 08 35

### DIFF
--- a/src/vs/workbench/services/preferences/test/browser/keybindingsEditorModel.test.ts
+++ b/src/vs/workbench/services/preferences/test/browser/keybindingsEditorModel.test.ts
@@ -214,7 +214,7 @@ suite('KeybindingsEditorModel', () => {
 	});
 
 	test('filter by command id', async () => {
-		const id = 'workbench.action.increaseViewSize';
+		const id = 'workbench.action.view.size.' + uuid.generateUuid();
 		registerCommandWithTitle(id, 'some title');
 		prepareKeybindingService();
 
--- a/build/lib/stylelint/vscode-known-variables.json
+++ b/build/lib/stylelint/vscode-known-variables.json
@@ -903,6 +903,9 @@
 		"--vscode-window-inactiveBorder"
 	],
 	"others": [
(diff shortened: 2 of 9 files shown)


lszomoru (approved):

src/vs/workbench/services/preferences/test/browser/keybindingsEditorModel.test.ts
```diff
@@ -214,7 +214,7 @@ suite('KeybindingsEditorModel', () => {
 	});
 
 	test('filter by command id', async () => {
-		const id = 'workbench.action.increaseViewSize';
+		const id = 'workbench.action.view.size.' + uuid.generateUuid();
 		registerCommandWithTitle(id, 'some title');
 		prepareKeybindingService();
```
Tyriar, line 219: What is going on here?

mrleemurray: I double checked with Opus, which implemented this due to conflicting test names:

**The problem:** The old test used the hardcoded ID `workbench.action.increaseViewSize`, which is a real command registered in `layoutActions.ts`. When the test called `registerCommandWithTitle(id, 'some title')`, it was registering a title for an ID that could also be registered by other code. This creates a collision — the test's behavior could be affected by the real command's registration, or the search query `workbench action view size` could match the real command's actual title in addition to the test's command, leading to flaky or duplicate results.

**The fix:** Using `'workbench.action.view.size.' + uuid.generateUuid()` ensures the command ID is unique per test run, eliminating any collision with real registered commands. The search query `workbench action view size` still matches because the test searches by tokenized segments of the command ID (dots are treated as separators), so workbench, action, view, and size all appear in the dynamic ID.

PR #294984, opened 2026-02-12
fix layout/contrast issues with tip widget

### LINKED ISSUE #294972: Chat tip doesn't take up whole input box and has bad contrast
Image

Saw this after doing `code-insiders --transient`

Also it can overflow

Image

### LINKED ISSUE #294973: tip not appearing on chat welcome view
It should be showing up right by the input box

meganrogge:
fix https://github.com/microsoft/vscode/issues/294972
fix https://github.com/microsoft/vscode/issues/294973

https://github.com/user-attachments/assets/cb3c441e-7a0c-4ce3-a8ae-0337c052c37c

### DIFF
--- a/src/vs/workbench/contrib/chat/browser/widget/chatWidget.ts
+++ b/src/vs/workbench/contrib/chat/browser/widget/chatWidget.ts
@@ -2510,8 +2510,11 @@ export class ChatWidget extends Disposable implements IChatWidget {
 
 		this.welcomeMessageContainer.style.height = `${contentHeight}px`;
 
-		// Keep getting-started tip aligned with the input after resize
-		this.layoutGettingStartedTipPosition();
+		// Keep getting-started tip aligned with the input after resize,
+		// but only when the tip is actually present to avoid unnecessary layout work.
+		if (this.container.classList.contains('chat-has-getting-started-tip')) {
+			this.layoutGettingStartedTipPosition();
+		}
 
 		const lastResponseIsRendering = isResponseVM(lastItem) && lastItem.renderData;
 		if (lastElementVisible && (!lastResponseIsRendering || checkModeOption(this.input.currentModeKind, this.viewOptions.autoScroll))) {
--- a/src/vs/workbench/contrib/chat/browser/widget/chatWidget.ts
+++ b/src/vs/workbench/contrib/chat/browser/widget/chatWidget.ts
@@ -2489,6 +2489,11 @@ export class ChatWidget extends Disposable implements IChatWidget {
 
 		this.bodyDimension = new dom.Dimension(width, height);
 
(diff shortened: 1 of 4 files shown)


sandy081 (dismissed):

src/vs/workbench/contrib/chat/browser/widget/chatWidget.ts
```diff
@@ -2489,6 +2489,11 @@ export class ChatWidget extends Disposable implements IChatWidget {
 
 		this.bodyDimension = new dom.Dimension(width, height);
 
+		// Ensure the container's rendered height matches the JS-calculated
+		// layout height so that absolutely-positioned children (e.g. the
+		// getting-started tip) are correctly placed relative to the container.
+		this.container.style.height = `${height}px`;
```
roblourens: Why are tips absolutely-positioned? They sit on top of the input, same as the edit list, todo container, etc right?

src/vs/workbench/contrib/chat/browser/widget/chatWidget.ts
```diff
@@ -2505,6 +2510,12 @@ export class ChatWidget extends Disposable implements IChatWidget {
 
 		this.welcomeMessageContainer.style.height = `${contentHeight}px`;
 
+		// Keep getting-started tip aligned with the input after resize,
+		// but only when the tip is actually present to avoid unnecessary layout work.
+		if (this.container.classList.contains('chat-has-getting-started-tip')) {
```
roblourens: Please avoid using hardcoded string classes in multiple places. No type safety, the DOM should not be the source of truth for your UI state.

roblourens: I know we have code like this, people keep doing it, but it's an anti-pattern

roblourens (dismissed):

src/vs/workbench/contrib/chat/browser/widget/chatWidget.ts
meganrogge: Good point, I'll try to align with those in another PR

src/vs/workbench/contrib/chat/browser/widget/chatWidget.ts
```diff
@@ -236,7 +236,7 @@ export class ChatWidget extends Disposable implements IChatWidget {
 	private welcomeMessageContainer!: HTMLElement;
 	private readonly welcomePart: MutableDisposable<ChatViewWelcomePart> = this._register(new MutableDisposable());
 
-	private gettingStartedTipContainer: HTMLElement | undefined;
+
```
meganrogge:
```suggestion
```

rzhao271 (dismissed):

rzhao271 (approved):

PR #294986, opened 2026-02-12
Add a LSP implementation for prompts

pierceboggan:
Adds real-time quality analysis for prompt files — like a linter, but for prompt engineering. Catches common authoring mistakes and helps write more effective AI instructions directly in the editor.

image

### Features

- **Instruction strength analysis** — Flags weak language ("try to", "consider", "might") and suggests stronger alternatives ("Must", "Always", "Never")
- **Ambiguity detection** — Catches vague quantifiers ("several", "a few") and unresolved positional references ("as mentioned above")
- **Structure linting** — Detects mixed XML/Markdown conventions and mismatched XML tags
- **Redundancy detection** — Finds duplicate instructions and constraints that subsume each other (e.g., "Never X" + "Avoid X")
- **Variable validation** — Flags empty `{{}}` placeholders and undefined template variables
- **Token budget awareness** — Warns when prompts are too large or contain content that tokenizes inefficiently
- **Example sufficiency** — Suggests adding few-shot examples when output format is specified without them
- **LLM-powered semantic analysis** — Uses Copilot to detect contradictions, persona inconsistencies, cognitive load issues, and coverage gaps

### How it works

Two analysis layers run in parallel:

1. **Static analysis** (debounced 400ms) — fast, free, deterministic checks on every keystroke
2. **LLM analysis** (debounced 3s) — sends content to Copilot for semantic checks like contradiction detection and coverage gaps

Results surface through standard editor UX:
- **Diagnostics** in the Problems panel with severity-appropriate markers
- **Code Lenses** showing issue count + estimated token usage
- **Hovers** showing instruction-strength classification on keywords
- **Quick Fix code actions** that replace weak phrasing in one click

### DIFF
--- a/src/vs/workbench/contrib/chat/browser/chat.contribution.ts
+++ b/src/vs/workbench/contrib/chat/browser/chat.contribution.ts
@@ -132,6 +132,7 @@ import { LanguageModelToolsConfirmationService } from './tools/languageModelTool
 import { LanguageModelToolsService, globalAutoApproveDescription } from './tools/languageModelToolsService.js';
 import './promptSyntax/promptCodingAgentActionContribution.js';
 import './promptSyntax/promptToolsCodeLensProvider.js';
+import './promptSyntax/promptQualityCodeLensProvider.js';
 import { showConfigureHooksQuickPick } from './promptSyntax/hookActions.js';
 import { PromptUrlHandler } from './promptSyntax/promptUrlHandler.js';
 import { ConfigureToolSets, UserToolSetsContributions } from './tools/toolSetsContribution.js';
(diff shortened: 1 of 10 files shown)


taylorarndt:
This looks really useful! Having prompt quality analysis right in the editor is something I've been wanting. I'd love to try a sample or preview of this if there's a way to test it out.
Quick question though - what model is actually powering the LLM analysis? The PR just says "Copilot" but I'm wondering if there's visibility into which model is being used, or if there's any plan to let users choose? Different models have pretty different strengths for this kind of analysis.
Also curious about availability - will this work in the CLI version of VS Code, or is it editor-only? And does the LLM analysis layer count against Copilot quota or require a premium subscription? Just trying to understand the usage model.
This brings up something I've been thinking about too - different models have really different conventions for what makes a "good" prompt. Like Claude works best with XML-structured prompts and explicit instructions, GPT-5 has its own patterns, Gemini prefers different things. Right now the linter seems to assume one set of best practices, but what works for one model might not be optimal for another. For example, flagging missing XML tags would be super helpful for Claude prompts but not really relevant if you're writing for GPT-5.
Any thoughts on making this model-aware? Maybe through a setting where you specify your target model and the analysis adjusts accordingly? Or even a plugin system where people could add model-specific analyzers?
The separation of static vs LLM analysis is really smart, and I could see this evolving into something more sophisticated where different specialized models handle different aspects of quality. Would be amazing to see this become a cross-editor standard for prompt engineering if the architecture can support that kind of extensibility.
Excited to see where this goes!

PR #294992, opened 2026-02-12
add support for fetching active chat session

bhavyaus:


src/vscode-dts/vscode.proposed.chatParticipantPrivate.d.ts
```diff
@@ -333,6 +333,17 @@ declare module 'vscode' {
 		export function registerChatParticipantDetectionProvider(participantDetectionProvider: ChatParticipantDetectionProvider): Disposable;
 
 		export const onDidDisposeChatSession: Event<string>;
+
+		/**
+		 * The resource URI of the currently active (focused or last focused) chat session,
+		 * or `undefined` if there is no active chat session.
+		 */
+		export const activeChatSessionResource: Uri | undefined;
```
mjbvz: Let's put this into the `window` namespace

mjbvz: `active` as a concept also feels a bit fuzzy to me. It makes me think that you're going to lock the workspace to a session. It also feels like it may get confusing since there are multiple ways of interactive with sessions (panel, editors, welcome view, ...)

Maybe it would be clearer for now if we restrict it to just the chat panel: `activeChatPanelSessionResource`?

mjbvz (changes requested):

src/vscode-dts/vscode.proposed.chatParticipantPrivate.d.ts
bhavyaus: Done. Thanks.

mjbvz (approved):

PR #295104, opened 2026-02-13
Use esbuild instead of webpack to bundle the css extension

mjbvz:
Switches from webpack to esbuild to bundle the css extension. Tested this locally in a browser. I am also creating an official build to test the bundled extension still work correctly


mjbvz:
Full build seems to work well too. @aeschli Please take a look when you have a chance. If these changes look good, I can do the same for html and json

Yoyokrazy (approved):

PR #295182, opened 2026-02-13
Add additional markdown extensions

remcohaszing:
This marks the `.litcoffee` file extension as markdown. This is a markdown format, where indented code blocks can be interpreted as CoffeeScript.

This also adds the `.mkdn` and `.ron` extensions, which are taken from https://github.com/sindresorhus/markdown-extensions/blob/main/index.js.


mjbvz (dismissed):
Thanks!

Tyriar (dismissed):

PR #295261, opened 2026-02-13
Bootstrap Playwright service

kycutler:


src/vs/platform/browserView/electron-main/browserViewCDPProxyServer.ts
```diff
@@ -0,0 +1,223 @@
...
+
+	/**
+	 * Creates a short-lived, single-use token.
+	 * The token is revoked once consumed or after 30 seconds.
+	 */
+	async issueToken(): Promise<string> {
+		const token = this.makeToken();
+		this.tokens.set(token, { expiresAt: Date.now() + 30_000 });
+		this._register(disposableTimeout(() => this.tokens.delete(token), 30_000));
+		return token;
```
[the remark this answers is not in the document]
kycutler: It will be async in the future ;)

src/vs/platform/browserView/electron-main/browserViewCDPProxyServer.ts
```diff
@@ -0,0 +1,223 @@
...
+	private getWebSocketUrl(token: string): string {
+		return `ws://localhost:${this.port}/devtools/browser?token=${token}`;
+	}
+
+	/**
+	 * Returns a debug endpoint with a short-lived, single-use token in the
+	 * WebSocket URL. The token is revoked once a WebSocket connection is made
+	 * or after 30 seconds, whichever comes first.
+	 */
+	async getWebSocketEndpoint(): Promise<string> {
```
jruales, line 54: Let's put public methods before private ones if possible

src/vs/platform/browserView/electron-main/browserViewCDPProxyServer.ts
```diff
@@ -0,0 +1,223 @@
...
+/**
+ * WebSocket server that provides CDP debugging for browser views.
+ */
+export class BrowserViewCDPProxyServer extends Disposable {
+	private server: http.Server | undefined;
+	private port: number | undefined;
+	private readonly tokens: TokenManager;
+
+	constructor(
+		private readonly targetService: ICDPBrowserTarget,
```
jruales: nit: rename `targetService` to `browserTarget`, because here we expect an `ICDPBrowserTarget`, not an `IBrowserViewService`

src/vs/platform/browserView/electron-main/browserViewCDPProxyServer.ts
```diff
@@ -0,0 +1,223 @@
...
+	 * Returns a DisposableStore that cleans up all subscriptions.
+	 */
+	private wireWebSocket(upgraded: ISocket, connection: ICDPConnection): DisposableStore {
+		const disposables = new DisposableStore();
+
+		// Socket -> Connection: parse JSON, call sendMessage, write response/error
+		disposables.add(upgraded.onData((rawData: VSBuffer) => {
+			try {
+				const message = rawData.toString();
+				const { id, method, params, sessionId } = JSON.parse(message) as CDPRequest;
```
jruales, line 142: Do we need proper validation that it will conform to `CDPRequest`? If it doesn't, will it cause an uncaught exception?

src/vs/platform/browserView/node/playwrightService.ts
```diff
@@ -0,0 +1,79 @@
...
+	async initialize(): Promise<void> {
+		if (this._browser?.isConnected()) {
+			return;
+		}
+
+		if (this._initPromise) {
+			return this._initPromise;
+		}
+
+		this._initPromise = (async () => {
```
jruales, line 45: Why do we need this closure here? Couldn't this be a private method instead?

jruales (dismissed):

src/vs/platform/browserView/electron-main/browserViewCDPProxyServer.ts
jruales: Oops, never mind, just saw the try/catch around it

bpasero:
@kycutler just glancing over the `electron-main` and shared process changes, good to see that the import is lazy. This helps reducing cost on startup when the module is actually not needed. Let me know if you want a more thorough review, but it seems legit so far, at least from the plumbing how its wired in via services.

src/vs/platform/browserView/node/playwrightService.ts
kycutler: It could be its own method, though I think it's just style preference. I'll leave it as-is for now since it is likely to change shortly.

DonJayamanne (approved):

PR #295331, opened 2026-02-14
fix: “View in Chat Panel” with Claude not understanding review comments

1Ninad:
Fixes #292892 

## Problem

When using **Code Review** with Copilot and then switching the session to Claude, clicking **“View in Chat Panel”** and asking something like “Explain your comment” does not work.
Claude cannot understand the original review comment.

## Root Cause
The review comment is added to chat as a completed request.  
If it was generated by a different agent (Copilot) and the active agent is Claude, Claude does not receive that comment in usable context.

## Fix
When sending a message:
- If the last request was added via “View in Chat Panel” and came from a different agent:
  - Extract its response text.
  - Truncate to 8000 characters (append `…` if needed).
  - Build a new message that:
    - States the comment was added by another participant.
    - Includes the comment inside a `---` block.
    - Asks the agent to explain it.
  - Send this instead of the raw user message.

Otherwise, behavior remains unchanged.

Only the outbound request message is modified. No changes to session history, SCM, or other chat logic.

## File
- `src/vs/workbench/contrib/chat/common/chatService/chatServiceImpl.ts`

## How to Test
1. Make local changes.
2. Run **Code Review**.
3. Click **“View in Chat Panel”** on a comment.
4. Switch to Claude (Set Session Target) and hit enter.

Before: Claude has no context.  
After: Claude correctly explains the review comment.


1Ninad:

PR #295378, opened 2026-02-14
Save thinking titles created after response is complete

### LINKED ISSUE #295377: Thinking titles are sometimes not saved
I think this is related to our ChatModel changes, and we just don't save the titles when they come too late

https://github.com/user-attachments/assets/277cb2f7-15b8-46d3-8726-0aa78c36c18b

roblourens:
Fix #295377


connor4312 (approved):
It's not pretty but I suppose it should work.

Issue #277872, opened 2025-11-17
provideLanguageModelChatResponse CancellationToken not work


wszgrcy:
Does this issue occur when all extensions are disabled?: Yes/No

- VS Code Version:  1.106
- OS Version: windows 10

Steps to Reproduce:

[Test Repo](https://github.com/wszgrcy/copilot-debug)
1. npm i 
2. `f5`
3. 

<img width="1284" height="558" alt="Image" src="https://github.com/user-attachments/assets/80e7811e-0686-40ba-bf33-1bd34a95c5db" />

4. 

<img width="856" height="646" alt="Image" src="https://github.com/user-attachments/assets/63a35ce1-69fc-47e4-95d5-df7eed54ed50" />

5. 

<img width="1252" height="219" alt="Image" src="https://github.com/user-attachments/assets/e50974aa-fffd-4550-86a5-a99ae60d9ec8" />

6. 

<img width="655" height="384" alt="Image" src="https://github.com/user-attachments/assets/16343732-56c5-41c6-9292-0064a5a408a8" />

7.  not emit event or print `true`

<img width="797" height="153" alt="Image" src="https://github.com/user-attachments/assets/1a9a074e-1bcf-4d1c-aaaa-d7deea0626dd" />

```
  async provideLanguageModelChatResponse(
    model: LanguageModelChatInformation,
    _messages: Array<LanguageModelChatMessage>,
    _options: ProvideLanguageModelChatResponseOptions,
    progress: Progress<LanguageModelResponsePart>,
    _token: CancellationToken
  ): Promise<void> {
    console.log(_token);

    let ab = new AbortController();
    _token.onCancellationRequested((e) => {
      console.log(e);
      ab.abort();
    });
    setInterval(() => {
      console.log(_token.isCancellationRequested);
    }, 1000);
    await new Promise<void>((res) => {
      setTimeout(() => {
        res();
      }, 99999999);
    });
}
```

jrieken:
@lramos15 I debugging this to understanding and would like to hand off this to you. Is that fine for you?

smallmain:
@jrieken @lramos15 

The issues with version v1.109.0 are still present; can you fix them?

PR #295440, opened 2026-02-15
fix copilot stop signal

wszgrcy:
Temporary fix for Copilot conversation not stopping

https://github.com/microsoft/vscode/issues/277872

I acknowledge that this approach to fixing the issue is far from elegant; however, it significantly reduces token usage post-fix.


wszgrcy:

PR #295494, opened 2026-02-15
Add General Purpose agent support

digitarald:
Exposes a built-in "General Purpose" subagent so the model always has a named agent for parallel execution, without requiring a config flag.

Refs microsoft/vscode-internalbacklog#6750

### `runSubagentTool.ts`
- **`agentName` is now required** in the tool schema (always present, not gated by config)
- **"General Purpose"** constant (`GeneralPurposeAgentName`) — when selected, inherits the parent's system prompt, model, and tools (same as the previous unnamed-agent behavior)
- `invoke()` / `prepareToolInvocation()` use `isGeneralPurpose` check to branch between built-in and custom agent paths
- `BaseModelDescription` always includes the agent-name instruction
- Error message references "General Purpose" as a fallback
- Removed dead `IConfigurationService` injection
- Consolidated duplicate `ILanguageModelToolsService` injection into single `languageModelToolsService`

### `computeAutomaticInstructions.ts`
- `<agents>` XML block now renders whenever `runSubagentTool` is available (config gate removed)
- Hardcoded "General Purpose" entry always appears first
- Imports `GeneralPurposeAgentName` from `constants.ts` (avoids cyclic dependency with `runSubagentTool.ts`)

### `constants.ts`
- Added `GeneralPurposeAgentName = 'General Purpose'` constant (shared by `runSubagentTool.ts` and `computeAutomaticInstructions.ts`)

### Setting removal (`SubagentToolCustomAgents`)
- Deleted enum value from `constants.ts`
- Removed configuration registration from `chat.contribution.ts`
- Removed settings layout entry from `settingsLayout.ts`

### Tests
- `runSubagentTool.test.ts`: Updated `createTool` helpers (removed dead DI args), added tests for undefined/empty/explicit General Purpose agentName handling
- `computeAutomaticInstructions.test.ts`: Removed config setup calls, deleted "disabled custom agents" test, agents list test expects General Purpose first

## Testing
- 15/15 RunSubagentTool tests passing
- 38/38 ComputeAutomaticInstructions tests passing
- No cyclic dependencies (verified with `checkCyclicDependencies.ts`)


src/vs/workbench/contrib/chat/test/common/tools/builtinTools/runSubagentTool.test.ts
```diff
@@ -95,57 +103,98 @@ suite('RunSubagentTool', () => {
...
-				configService,
-				promptsService,
-				{} as IInstantiationService,
-			));
+			const result = await tool.prepareToolInvocation(
+				{
+					parameters: {
+						prompt: 'Test prompt',
+						description: 'Test task',
+						agentName: undefined,
```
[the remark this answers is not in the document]
digitarald: Why the interface should stay agentName: string
The TypeScript type should match the JSON Schema contract. The schema declares required: ['prompt', 'description', 'agentName']. The interface represents the intended input shape — making it optional would contradict the schema and confuse future contributors into thinking omission is a valid usage path.

Defensive checks for LLM inputs are standard practice. LLMs don't reliably follow JSON schemas. The !subAgentName guards at [runSubagentTool.ts:157](vscode-file://vscode-app/Applications/Visual%20Studio%20Code%20-%20Insiders.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) and [runSubagentTool.ts:386](vscode-file://vscode-app/Applications/Visual%20Studio%20Code%20-%20Insiders.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) are safety nets, not sanctioned input forms. The comments there (// Defensive: model may omit agentName despite schema requiring it) already explain this.

Optional would spread | undefined noise. Making it optional means every consumer needs to handle undefined explicitly, when only the two entry points (invoke, prepareToolInvocation) need to — and they already do via the isGeneralPurpose check.

dmitrivMS (dismissed):

src/vs/workbench/contrib/chat/common/tools/builtinTools/runSubagentTool.ts
```diff
@@ -71,16 +78,16 @@ export class RunSubagentTool extends Disposable implements IToolImpl {
 		@ILanguageModelsService private readonly languageModelsService: ILanguageModelsService,
 		@ILogService private readonly logService: ILogService,
 		@ILanguageModelToolsService private readonly toolsService: ILanguageModelToolsService,
-		@IConfigurationService private readonly configurationService: IConfigurationService,
+		@IConfigurationService _configurationService: IConfigurationService,
 		@IPromptsService private readonly promptsService: IPromptsService,
 		@IInstantiationService private readonly instantiationService: IInstantiationService,
 	) {
 		super();
-		this.onDidUpdateToolData = Event.filter(this.configurationService.onDidChangeConfiguration, e => e.affectsConfiguration(ChatConfiguration.SubagentToolCustomAgents));
```
roblourens, line 79: If you meant to remove this setting check, we can fully delete the setting.

DonJayamanne (dismissed):

src/vs/workbench/contrib/chat/common/tools/builtinTools/runSubagentTool.ts
digitarald: Done, removed the setting as the ExP didn't show any raised tokens or errors.

mjbvz:
Moving this to 1.112  so that we can get the 1.111 milestone into good shape by the end of Thursday

PR #295539, opened 2026-02-16
Update dependency list

dbaeumer:
Add a new dependency for `libc.so.6(GLIBC_2.5)(64bit)` to the existing list of dependencies. This update ensures compatibility with the specified version.


deepak1556 (approved):
Thanks!

PR #295569, opened 2026-02-16
Change sourcemap option from 'external' to 'linked' for better debugging support

jrieken:


lszomoru (approved):

chris-morgan:
This seems like a bad idea: it bloats builds *massively*, hundreds of megabytes, for something that doesn’t benefit normal users at all.

In Arch Linux, the `code` package grew from 182MB to 462MB, and it will probably revert this change itself: https://gitlab.archlinux.org/archlinux/packaging/packages/code/-/issues/42.

PR #295638, opened 2026-02-16
ci: add deb,rpm dependency check for batched builds

deepak1556:
Followup to https://github.com/microsoft/vscode/pull/295539

To allow catching dependency failures early on in the product cycle this PR does the following,

1) Enable dependency checks of deb and rpm for `VSCODE_CIBUILD`, this will only capture linux x64
2) For armhf and arm64, enable CCR leave a comment to trigger full build if any of the files specified have a possibility to affect the dependency.


deepak1556:
verification https://monacotools.visualstudio.com/Monaco/_build/results?buildId=407538&view=results

deepak1556:
Adds 20s to the linux `Electron tests` job and 2min to the linux cli job

PR #295676, opened 2026-02-17
Browser: Managed CDP context groups

kycutler:
Solves two issues:
- How to manage access to pages when exposing CDP to various consumers
- How to associate Playwright pages with our internal browser views

### DIFF
--- a/src/vs/platform/browserView/electron-main/browserViewCDPProxyServer.ts
+++ b/src/vs/platform/browserView/electron-main/browserViewCDPProxyServer.ts
@@ -22,44 +22,60 @@ export interface IBrowserViewCDPProxyServer {
 	readonly _serviceBrand: undefined;
 
 	/**
-	 * Returns a debug endpoint with a short-lived, single-use token.
+	 * Returns a debug endpoint with a short-lived, single-use token for a specific browser target.
 	 */
-	getWebSocketEndpoint(): Promise<string>;
+	getWebSocketEndpointForTarget(target: ICDPBrowserTarget): Promise<string>;
+
+	/**
+	 * Unregister a previously registered browser target.
+	 */
+	removeTarget(target: ICDPBrowserTarget): Promise<void>;
 }
 
 /**
  * WebSocket server that provides CDP debugging for browser views.
+ *
+ * Manages a registry of {@link ICDPBrowserTarget} instances, each reachable
+ * at its own `/devtools/browser/{id}` WebSocket endpoint.
  */
 export class BrowserViewCDPProxyServer extends Disposable implements IBrowserViewCDPProxyServer {
 	declare readonly _serviceBrand: undefined;
 
 	private server: http.Server | undefined;
 	private port: number | undefined;
-	private readonly tokens: TokenManager;
+
+	private readonly tokens = this._register(new TokenManager<string>());
+	private readonly targets = new Map<string, ICDPBrowserTarget>();
 
 	constructor(
-		private readonly browserTarget: ICDPBrowserTarget,
 		@ILogService private readonly logService: ILogService
 	) {
 		super();
-
-		this.tokens = this._register(new TokenManager());
 	}
 
 	/**
-	 * Returns a debug endpoint with a short-lived, single-use token in the
-	 * WebSocket URL. The token is revoked once a WebSocket connection is made
-	 * or after 30 seconds, whichever comes first.
+	 * Register a browser target and return a WebSocket endpoint URL for it.
+	 * The target is reachable at `/devtools/browser/{targetId}`.
 	 */
-	async getWebSocketEndpoint(): Promise<string> {
+	async getWebSocketEndpointForTarget(target: ICDPBrowserTarget): Promise<string> {
 		await this.ensureServerStarted();
 
-		const token = await this.tokens.issueToken();
-		return this.getWebSocketUrl(token);
+		const targetInfo = await target.getTargetInfo();
+		const targetId = targetInfo.targetId;
+
+		// Register (or re-register) the target
+		this.targets.set(targetId, target);
+
+		const token = await this.tokens.issueToken(targetId);
+		return `ws://localhost:${this.port}/devtools/browser/${targetId}?token=${token}`;
 	}
 
-	private getWebSocketUrl(token: string): string {
-		return `ws://localhost:${this.port}/devtools/browser?token=${token}`;
+	/**
+	 * Unregister a previously registered browser target.
+	 */
+	async removeTarget(target: ICDPBrowserTarget): Promise<void> {
+		const targetInfo = await target.getTargetInfo();
+		this.targets.delete(targetInfo.targetId);
 	}
 
 	private async ensureServerStarted(): Promise<void> {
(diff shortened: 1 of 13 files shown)


src/vs/platform/browserView/electron-main/browserViewCDPProxyServer.ts
```diff
@@ -93,19 +109,30 @@ export class BrowserViewCDPProxyServer extends Disposable implements IBrowserVie
 	private handleWebSocketUpgrade(req: http.IncomingMessage, socket: Socket): void {
 		const [pathname, params] = (req.url || '').split('?');
 
-		const token = new URLSearchParams(params).get('token');
-		if (!token || !this.tokens.consumeToken(token)) {
-			socket.write('HTTP/1.1 401 Unauthorized\r\n\r\n');
+		const browserMatch = pathname.match(/^\/devtools\/browser\/([^/?]+)$/);
```
jruales, line 112: Thoughts on making this a URL GET parameter instead of part of the URL path proper?

src/vs/platform/browserView/electron-main/browserViewCDPProxyServer.ts
```diff
@@ -22,44 +22,60 @@ export interface IBrowserViewCDPProxyServer {
...
-	async getWebSocketEndpoint(): Promise<string> {
+	async getWebSocketEndpointForTarget(target: ICDPBrowserTarget): Promise<string> {
 		await this.ensureServerStarted();
 
-		const token = await this.tokens.issueToken();
-		return this.getWebSocketUrl(token);
+		const targetInfo = await target.getTargetInfo();
+		const targetId = targetInfo.targetId;
+
+		// Register (or re-register) the target
```
jruales, line 66: Why do we want to re-register a target?

src/vs/platform/browserView/node/playwrightService.ts
```diff
@@ -4,37 +4,37 @@
...
 
 /**
  * Shared-process implementation of {@link IPlaywrightService}.
  */
 export class PlaywrightService extends Disposable implements IPlaywrightService {
 	declare readonly _serviceBrand: undefined;
 
-	private readonly browserViewService: IBrowserViewService;
 	private _browser: Browser | undefined;
+	private _pages: PlaywrightPageMap | undefined;
```
jruales: Do we want to expose this in some way? I'm guessing that's coming in the next PR

src/vs/platform/browserView/node/playwrightService.ts
```diff
@@ -83,3 +94,211 @@ export class PlaywrightService extends Disposable implements IPlaywrightService
...
+		} catch (err: unknown) {
+			const errorMessage = err instanceof Error ? err.message : String(err);
+			this.logService.error('[PlaywrightService] Failed to add view:', errorMessage);
+			deferred.error(new Error(`Failed to get page: ${errorMessage}`));
+		}
+
+		return deferred.p;
+	}
+
+	private removePage(viewId: string): void {
```
jruales: shouldn't `removePage()` be public too to match `newPage()` visibility?

src/vs/platform/browserView/node/playwrightService.ts
```diff
@@ -83,3 +94,211 @@ export class PlaywrightService extends Disposable implements IPlaywrightService
...
+		this._pageQueue.push({ page, viewId: deferred });
+		this.tryMatch();
+
+		return deferred.p;
+	}
+
+	// --- Matching ---
+
+	/**
+	 * Pair up queued view IDs with queued pages in FIFO order and resolve
```
jruales, line 306: Is the likelihood of non-FIFO negligible? For example, if two pages are created at the same time, could they end up assigned to each other?

src/vs/platform/browserView/node/playwrightService.ts
```diff
@@ -4,37 +4,37 @@
...
 	declare readonly _serviceBrand: undefined;
 
-	private readonly browserViewService: IBrowserViewService;
 	private _browser: Browser | undefined;
+	private _pages: PlaywrightPageMap | undefined;
 	private _initPromise: Promise<void> | undefined;
 
 	constructor(
-		@IMainProcessService mainProcessService: IMainProcessService,
+		@IBrowserViewGroupRemoteService private readonly groupRemoteService: IBrowserViewGroupRemoteService,
```
jruales: nit: can we rename `groupRemoteService` to `viewGroupRemoteService` or `browserViewGroupRemoteService`? I think it would make sense to just call it "group" if the context of a file related to browser views, but since the current file is related to Playwright more generally, I think it's better to be more specific here so it's more clear what it's a group of.

src/vs/platform/browserView/node/playwrightService.ts
```diff
@@ -83,3 +94,211 @@ export class PlaywrightService extends Disposable implements IPlaywrightService
...
+
+		this._register(_group.onDidAddView(e => this.getPage(e.viewId)));
+		this._register(_group.onDidRemoveView(e => this.removePage(e.viewId)));
+		this.scanForNewContexts();
+	}
+
+	/**
+	 * Create a new page in the browser and return its associated page and view ID.
+	 */
+	async newPage(): Promise<{ viewId: string; page: Page }> {
```
jruales, line 149: More of a question: Is it correct to give the responsibility of creating new pages to a page map class? The name doesn't seem to imply that it has this responsibility

src/vs/platform/browserView/node/browserViewGroupRemoteService.ts
```diff
@@ -0,0 +1,109 @@
...
+export const IBrowserViewGroupRemoteService = createDecorator<IBrowserViewGroupRemoteService>('browserViewGroupRemoteService');
+
+/**
+ * Remote-process service for managing browser view groups.
+ *
+ * Connects to the main-process {@link BrowserViewGroupMainService} via
+ * IPC and provides {@link IBrowserViewGroup} instances for
+ * interacting with groups.
+ *
+ * Usable from the shared process or the workbench renderer.
```
jruales: Even workbench is registered to use this?

jruales (dismissed):

src/vs/platform/browserView/electron-main/browserViewCDPProxyServer.ts
kycutler: This pattern aligns with Chrome CDP server

src/vs/platform/browserView/electron-main/browserViewCDPProxyServer.ts
kycutler: This can happen multiple times if multiple connections / tokens are issued for the same target

src/vs/platform/browserView/node/playwrightService.ts
kycutler: It could happen if the two pages are in different, newly created contexts, since context discovery is done via polling so the order may not be consistent. But on the same / existing contexts I think it shouldn't happen. It is admittedly fragile -- ideally we could just grab the targetId from the page. Maybe we can request it from the Playwright team.

src/vs/platform/browserView/node/browserViewGroupRemoteService.ts
kycutler: Good catch, I thought it could be but there's not a good place to put the service where it can be used by both without breaking layering rules. It's only used by shared process for now so updated the comment.

src/vs/platform/browserView/node/playwrightService.ts
kycutler: Updated to be `PlaywrightPageManager` for clarity

src/vs/platform/browserView/node/playwrightService.ts
kycutler: Refactored these methods a bit to be more consistent

src/vs/platform/browserView/node/playwrightService.ts
kycutler: Yes, to allow external pages to actually be registered we will need to expose methods on this service. But that will be in a separate PR

bhavyaus (approved):

PR #255375, opened 2025-07-11
fix: prevent duplicate 'Disable' button in extension view (fix #244138)

masa-kjm:
This PR fixes issue #244138 by ensuring that the “Disable” and “Disable (Workspace)” buttons are not shown at the same time when an extension is enabled only in the workspace.

- The visibility of the global “Disable” button is controlled by DisableGloballyAction.
- To avoid showing this button when the extension is already disabled globally, we now use the isDisabledGlobally method from extensionEnablementService, which accurately reflects the global disabled state.

Please let me know if any further changes are needed.

PR #295692, opened 2026-02-17
Fix: prevent duplicate disable action when extension is disabled glob…

conradevans:
…ally but enabled for workspace

- Prevents both Disable and Disable (Workspace) from appearing together
- Ensures correct action visibility based on enablement state
- Fixes #244138


sandy081:
Thanks for looking into this @conradevans! The diff here is quite large (+3698/-1203) for the scope of the fix, which makes it hard to review. Could you check if there are unintended formatting or whitespace changes included? Ideally the PR should only touch the lines necessary to fix the issue.

Also, please note there are a few other open PRs for the same issue — #255375, #285636, #289324 — you may want to take a look at those to avoid duplicating effort.

PR #295701, opened 2026-02-17
Enable kitty graphics protocol, bump xterm.js

### LINKED ISSUE #286141: Terminal: Implement Kitty graphics protocol support
https://sw.kovidgoyal.net/kitty/graphics-protocol/

Upstream: https://github.com/xtermjs/xterm.js/issues/5592

anthonykim1:
Need https://github.com/microsoft/vscode/pull/295820 before. 
Resolves: https://github.com/microsoft/vscode/issues/286141


src/vs/workbench/contrib/terminal/common/terminalConfiguration.ts
```diff
@@ -654,7 +654,7 @@ const terminalConfiguration: IStringDictionary<IConfigurationPropertySchema> = {
...
+		markdownDescription: localize('terminal.integrated.enableImages', "Enables image support in the terminal, including sixel, iTerm's inline image protocol, and the kitty graphics protocol. This will only work when {0} is enabled. Sixel and iTerm's inline image protocol are supported on Linux and macOS. This will only work on Windows for versions of ConPTY >= v2 which is shipped with Windows itself, see also {1}. Images will currently not be restored between window reloads/reconnects.\n\nNote: Kitty graphics z-index support (rendering images behind text) only works when {0} is set to {2}. Some kitty graphics features may also require {3} to be enabled.", `\`#${TerminalSettingId.GpuAcceleration}#\``, `\`#${TerminalSettingId.WindowsUseConptyDll}#\``, '`"off"`', `\`#terminal.integrated.allowTransparency#\``),
```
anthonykim1: This may sound contradictory.. we first say `This will only work when {0} is enabled.` and then `only works when {0} is set to {2}.` for gpu acceleration

Tyriar: It is. Does the addon not work for the DOM renderer?

anthonykim1: Actually I meant to just say users need to enagle transparency

roblourens (approved):

Issue #294029, opened 2026-02-09
Enable sandbox for locally running MCP servers


dileepyavan:
**Details**

Local MCP servers that use the stdio transport are currently launched as child processes on the host machine. Because they run without isolation, these servers can read, modify, or delete sensitive data on the local file system.

Sandboxing introduces an isolation boundary that requires MCP servers to explicitly request access to specific files or network domains they need in order to operate.

**Scope**

- Applies to locally running MCP servers that use the stdio transport type

- Supported platforms:
    Linux    
    macOS

dileepyavan:
[https://github.com/microsoft/vscode/pull/295704](url)

PR #295704, opened 2026-02-17
Enabling sandboxing for local MCP servers at workspace level.

dileepyavan:
fixes(#294029)
This PR enables sandboxing for MCP servers that are running locally.

- Currently the entry point for enabling sandboxing is through mcp.json at workspace level.
- Uses anthropic sandbox runtime for sandboxing.
- Limited to Linux and MacOS.


src/vs/platform/mcp/common/mcpPlatformTypes.ts
```diff
@@ -45,6 +57,8 @@ export interface IMcpStdioServerConfiguration extends ICommonMcpServerConfigurat
 	readonly env?: Record<string, string | number | null>;
 	readonly envFile?: string;
 	readonly cwd?: string;
+	readonly sandboxEnabled?: boolean;
+	readonly sandbox?: IMcpSandboxConfiguration;
```
connor4312, line 61: I feel like many times I will want MCP's to have the same or similar MCP configs. Maybe we adopt a reference approach for `mcp.json` like

```
{
	"sandboxes": {
		"default": {
			// IMcpSandboxConfiguration here
		}
	},
	"servers": {
		"my-mcp": {
			"sandbox": "default",
			"command": "npx"
			// ...
		}
	}
}
```

so

- `IMcpStdioServerConfiguration.sandbox` is a `string | undefined` referencing a sandbox config
- `sandboxes` can be defined in the configuration
- 💪 item for later: this approach would also let mcp servers reference other sandboxes, like maybe I have a `workspace` sandbox config defined in my user settings that provides good defaults for most workspace-scoped MCPs, and I could use that in my servers (downside: less portability)

src/vs/workbench/api/node/extHostMcpNode.ts
```diff
@@ -76,6 +76,12 @@ export class NodeExtHostMpcService extends ExtHostMcpService {
 			env[key] = value === null ? undefined : String(value);
 		}
 
+		const isSandboxedLaunch = env[mcpSandboxedLaunchEnvironmentKey] === 'true';
+		delete env[mcpSandboxedLaunchEnvironmentKey];
```
connor4312: It seems like the only purpose of this is side-channeling an indicator for sandboxed launching? Seems like we could instead ignore `definition.devMode` if sandboxing is also present (could also show a warning for that in mcp.json schema)

src/vs/workbench/contrib/mcp/common/mcpLanguageModelToolContribution.ts
```diff
@@ -215,15 +219,18 @@ class McpToolImplementation implements IToolImpl {
...
-		if (!tool.definition.annotations?.readOnlyHint) {
-			confirm.title = new MarkdownString(localize('msg.title', "Run {0}", title));
-			confirm.message = new MarkdownString(tool.definition.description, { supportThemeIcons: true });
-			confirm.disclaimer = mcpToolWarning;
-			confirm.allowAutoConfirm = true;
-		}
-		if (tool.definition.annotations?.openWorldHint) {
-			confirm.confirmResults = true;
+		let confirm: IToolConfirmationMessages | undefined;
+		if (!isSandboxedServer) {
```
connor4312, line 225: I think we need to be careful here. If a user applies sandboxing as a "best practice, why not" security measure, they may be surprised that side-effects all of a sudden happen without confirmation.

If I have some stdio SQL MCP server, just because I put that in a sandbox so it can only talk to that database does not mean I want to allow it to drop that database without confirmation.

I agree that we should make the path to sandbox==no approvals simple and obvious, or even the default behavior, but we should also make that clear and allow the user to keep approvals with sandboxing. (Maybe this is a property of the `IMcpSandboxConfiguration`)

src/vs/workbench/contrib/mcp/common/mcpTypes.ts
```diff
@@ -189,7 +198,8 @@ export namespace McpServerDefinition {
 			&& objectsEqual(a.launch, b.launch)
 			&& objectsEqual(a.presentation, b.presentation)
 			&& objectsEqual(a.variableReplacement, b.variableReplacement)
-			&& objectsEqual(a.devMode, b.devMode);
+			&& objectsEqual(a.devMode, b.devMode)
```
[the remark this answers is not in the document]
connor4312: ^ copilot's comment is accurate

src/vs/workbench/contrib/mcp/common/mcpSandboxService.ts
```diff
@@ -0,0 +1,286 @@
...
+		}));
+		// also listen to MCP server installs to populate the initial sandbox config and enabled servers set based on the install results.
+		this._register(this._mcpManagementService.onDidInstallMcpServers(e => {
+			this._populateSandboxConfigFromInstallResults(e);
+		}));
+	}
+
+	public async isEnabled(serverDef: McpServerDefinition, serverLabel?: string): Promise<boolean> {
+		this._remoteEnvDetails = await this._remoteEnvDetailsPromise;
+		this._os = this._remoteEnvDetails ? this._remoteEnvDetails.os : OS;
```
connor4312: Individual servers may run locally or on the remote. Just because there is a remote does not mean all servers will run there

src/vs/workbench/contrib/mcp/common/mcpSandboxService.ts
```diff
@@ -0,0 +1,286 @@
...
+
+		if (this._os === OperatingSystem.Windows) {
+			return false;
+		}
+		this._logService.debug(`McpSandboxService: Checking sandbox enablement for server ${serverDef.label}. Enabled servers: ${[...this._sandboxEnabledMcpServers].join(', ')}`);
+		return this._sandboxEnabledMcpServers.has(serverLabel ?? serverDef.label);
```
connor4312: You're already getting the `McpServerDefinition` when checking sandbox enablement. It has a `McpServerDefinition.sandbox` config, I don't think there's a need to maintain a separate state in `_sandboxEnabledMcpServers`

src/vs/workbench/contrib/mcp/common/mcpSandboxService.ts
```diff
@@ -0,0 +1,286 @@
...
+	isEnabled(serverDef: McpServerDefinition, serverLabel?: string): Promise<boolean>;
+}
+
+export class McpSandboxService extends Disposable implements IMcpSandboxService {
+	readonly _serviceBrand: undefined;
+
+	private _srtPath: string | undefined;
+	private _srtPathResolved = false;
+	private _sandboxConfigPath: string | undefined;
+	private _sandboxConfig: IMcpSandboxConfiguration | undefined;
```
connor4312: In your current config structure you can't assume every MCP server has the same sandbox config

src/vs/workbench/contrib/mcp/common/mcpSandboxService.ts
```diff
@@ -0,0 +1,286 @@
...
+		this._logService.debug(`McpSandboxService: Checking sandbox enablement for server ${serverDef.label}. Enabled servers: ${[...this._sandboxEnabledMcpServers].join(', ')}`);
+		return this._sandboxEnabledMcpServers.has(serverLabel ?? serverDef.label);
+	}
+
+	public async launchInSandboxIfEnabled(serverDef: McpServerDefinition, launch: McpServerLaunch): Promise<McpServerLaunch> {
+		if (launch.type !== McpServerTransportType.Stdio) {
+			return launch;
+		}
+		if (await this.isEnabled(serverDef)) {
+			const srtPath = await this.getSandboxRuntimePath();
```
connor4312: SRT here will be a .js script that pulls Node from the environment. This will not work on systems without Node.js installed.

I think you handled this already for terminal sandboxing, can you reuse logic from there here?

src/vs/workbench/contrib/mcp/common/mcpSandboxService.ts
```diff
@@ -0,0 +1,286 @@
...
+				denyWrite: sandboxConfig?.filesystem?.denyWrite ?? [],
+			},
+		};
+	}
+
+	private _getDefaultAllowWrite(): readonly string[] {
+		return [
+			'.',
+			'~/.npm',
+			this._pathJoin(this._appRoot, 'node_modules', '@anthropic-ai'),
```
connor4312: Do they actually need to write to the srt module? That seems dangerous, a smart escape could rewrite the SRT module itself to no-op and escape the sandbox

src/vs/workbench/contrib/mcp/common/mcpSandboxService.ts
```diff
@@ -0,0 +1,286 @@
...
+				allowWrite: [...mergedAllowWrite],
+				denyRead: sandboxConfig?.filesystem?.denyRead ?? [],
+				denyWrite: sandboxConfig?.filesystem?.denyWrite ?? [],
+			},
+		};
+	}
+
+	private _getDefaultAllowWrite(): readonly string[] {
+		return [
+			'.',
```
connor4312: What if I only want an MCP to operate in a subdirectory of a monorepo?

src/vs/workbench/contrib/mcp/common/mcpSandboxService.ts
```diff
@@ -0,0 +1,286 @@
...
+		if (!this._tempDir) {
+			await this._initTempDir();
+		}
+		if (!this._tempDir) {
+			return;
+		}
+		await this._updateSandboxConfig(sandboxConfig);
+	}
+
+	private _areSandboxConfigsEqual(left: IMcpSandboxConfiguration | undefined, right: IMcpSandboxConfiguration | undefined): boolean {
```
connor4312: we have an object equality helper for this

connor4312 (changes requested):

src/vs/platform/mcp/common/mcpPlatformTypes.ts
dileepyavan: I do have a property on each MCP server level but it still reads from Sandbox property set at root level . This is done for easy maintainability as all the info can be retrieved from server config.Currently it is as below:alt text

dileepyavan: {
	"servers": {
		"vscode-automation-mcp": {
			"type": "stdio",
			"command": "npm",
			// Look at the [README](../test/mcp/README.md) to see what arguments are supported
			"args": ["run", "start-stdio"],
			"cwd": "${workspaceFolder}/test/mcp",
			"sandboxEnabled": true
		},
		"vscode-automation-mcp2": {
			"type": "stdio",
		"command": "npm",
			// Look at the [README](../test/mcp/README.md) to see what arguments are supported
			"args": ["run", "start-stdio"],
			"cwd": "${workspaceFolder}/test/mcp",
			"sandboxEnabled": true
		},
		"sandbox":{
			//configure the sandbox server to use the same command as vscode-automation-mcp, but with a different argument to start in sandbox mode
			"fileSystem": {
				
			},
			"Network": {
				"allowedDomains": ["example.com"],
				"allowedPorts": [80, 443]
			},
		}
	},
	"inputs": []
}

connor4312: In the code that's in this PR, you allow defining a sandbox at a server level

https://github.com/microsoft/vscode/pull/295704/changes#diff-8419dd7ab47715a2aaca305446c2bf83f8142d17ccf1d0de98c94083f8ce92a2R175

src/vs/workbench/contrib/mcp/common/mcpLanguageModelToolContribution.ts
dileepyavan: Currently for terminal commands sandboxing, If the path is in allowedList, no confirmation is requested. Followed the same approach.This might cause confusion if we have different implementations per feature. Let me know your thoughts.Any reason why it has to be different from run_in_terminal or background agents in the future.

src/vs/workbench/contrib/mcp/common/mcpSandboxService.ts
dileepyavan: Currently eventhough we have sandbox property at each MCP server, its not being set from MCP.json. This property is being populated from Sandbox property at the root level.Having a sandbox at MCP level will need different sandbox config files to be created.

src/vs/workbench/contrib/mcp/common/mcpSandboxService.ts
dileepyavan: yeah agree, might not be a good idea to include hidden defaults. If defaults are included, it must be shown to the user. My intention for including defaults is less work for the user to start the MCP server in sandbox mode. Are the visible defaults okay in mcp.json? or we dont need any

src/vs/workbench/contrib/mcp/common/mcpSandboxService.ts
dileepyavan: I dont think so , it should only be read when CWD property is present for spawing child process, read permission should be allowed on this folder.

src/vs/platform/mcp/common/mcpPlatformTypes.ts
dileepyavan: I think this should not be there. Removed it.

src/vs/workbench/api/node/extHostMcpNode.ts
dileepyavan: This is part of env variables that are being populated. This errors out without passing any dev config in mcp.json.

src/vs/workbench/contrib/mcp/common/mcpSandboxService.ts
dileepyavan: the launch config is not updated when there is a change in the definition, it only gets the initial value. So, we have listeners on the change event and updating the set if sandboxEnabled is set to true. Also for the tool calls to bypass confirmation window, server label is used instead as serverDef.label doesnt match with the serverDef.Label.

src/vs/workbench/contrib/mcp/common/mcpSandboxService.ts
dileepyavan: Updated across all the functions in this file

src/vs/workbench/contrib/mcp/common/mcpSandboxService.ts
dileepyavan: I tested this using insiders, env variables have Node path already included, so i went ahead and removed the extra code.

src/vs/workbench/contrib/mcp/common/mcpSandboxService.ts
dileepyavan: Removed all except ~/.npm for allowWrite and *.npmjs.org for allowed domains

src/vs/workbench/contrib/mcp/common/mcpSandboxService.ts
dileepyavan: This is no longer needed. Removed it. Anything other than CWD folder should have read access other than default sensitive files per OS.

src/vs/workbench/api/node/extHostMcpNode.ts
connor4312: I understand, but we can avoid the VSCODE_INSPECTOR_OPTIONS being passed if we skip the handling of `definition.devMode` in the registry

dileepyavan: This is coming from extension.js and disabling the dev mode still has this env variable populated.

src/vs/workbench/contrib/mcp/common/mcpTypes.ts
```diff
@@ -132,6 +133,10 @@ export interface McpServerDefinition {
 	readonly devMode?: IMcpDevModeConfig;
 	/** Static description of server tools/data, used to hydrate the cache. */
 	readonly staticMetadata?: McpServerStaticMetadata;
+	/** Indicates if the sandbox is enabled for this server. */
+	readonly sandboxEnabled?: boolean;
```
connor4312, line 136: Can we get rid of the separate `sandboxEnabled` boolean and treat the presence of a `sandbox?:` as a signal for enablement? An MCP server with sandbox enabled but not IMcpSandboxConfiguration is an invalid state and should be irrepresentable

src/vs/workbench/contrib/mcp/common/mcpSandboxService.ts
```diff
@@ -0,0 +1,280 @@
...
+			return;
+		}
+		this._srtPathResolved = true;
+		if (remoteAuthority) {
+			this._remoteEnvDetails = await this._remoteEnvDetailsPromise;
+			if (this._remoteEnvDetails) {
+				this._appRoot = this._remoteEnvDetails.appRoot.path;
+			}
+		}
+		this._srtPath = this._pathJoin(this._appRoot, 'node_modules', '@anthropic-ai', 'sandbox-runtime', 'dist', 'cli.js');
```
connor4312: The storing of global state on the service is likely to break if remote and local MCP servers are started in parallel. Maybe you want to have a small internal helper class like `McpSandboxDetails` and have `_local?: McpSandboxDetails; _remote?: McpSandboxDetails` to store/cache necessary state

src/vs/workbench/contrib/mcp/common/mcpSandboxService.ts
```diff
@@ -0,0 +1,280 @@
...
+			this._remoteEnvDetails = await this._remoteEnvDetailsPromise;
+			if (this._remoteEnvDetails) {
+				this._tempDir = this._remoteEnvDetails.tmpDir;
+			}
+		} else {
+			const environmentService = this._environmentService as IEnvironmentService & { tmpDir?: URI };
+			this._tempDir = environmentService.tmpDir;
+		}
+		if (!this._tempDir) {
+			this._logService.warn('McpSandboxService: Cannot create sandbox settings file because no tmpDir is available in this environment');
```
connor4312: Same deal here, you have one _tempDir for both local and remotes

src/vs/workbench/contrib/mcp/common/mcpSandboxService.ts
```diff
@@ -0,0 +1,280 @@
...
+	private async _updateSandboxConfig(sandboxConfig?: IMcpSandboxConfiguration): Promise<string | undefined> {
+		if (!this._tempDir) {
+			return undefined;
+		}
+		const normalizedSandboxConfig = this._withDefaultSandboxConfig(sandboxConfig);
+		if (this._sandboxConfigPath && equals(this._sandboxConfig, normalizedSandboxConfig)) {
+			return this._sandboxConfigPath;
+		}
+		this._sandboxConfig = normalizedSandboxConfig;
+		const configFileUri = URI.joinPath(this._tempDir, `vscode-mcp-sandbox-settings-${this._sandboxSettingsId}.json`);
```
connor4312: A single _sandboxSettingsId is prone to break multiple MCP servers are launched with different sandboxes. Even with an mcp.json-level sandbox config, servers at the user level, workspace level, and workspace folder level may be defined with different sandbox configs

src/vs/workbench/contrib/mcp/common/mcpSandboxService.ts
connor4312: > the launch config is not updated when there is a change in the definition, it only gets the initial value

This is a bug we should fix.

> server label is used instead as serverDef.label doesnt match with the serverDef.Label

Labels are not unique -- we should use the `McpServerDefinition.id` instead, but that problem will also go away just by avoiding maintaining `_sandboxEnabledMcpServers`

connor4312 (changes requested):

src/vs/workbench/contrib/mcp/common/mcpTypes.ts
dileepyavan: If some of the servers dont want to opt in for sandboxing and other servers might want to opt in, this flag controls that. User can specify sanbox setting at the root level only and not at each server level.

src/vs/workbench/contrib/mcp/common/mcpSandboxService.ts
dileepyavan: good point. Yeah this will cause failure. I will update.

src/vs/workbench/contrib/mcp/common/mcpTypes.ts
connor4312: But when we publish the MCP server definitions into the registry, can we just omit the `sandbox` config from servers that are opted out from sandboxing rather than keeping the separate boolean?

src/vs/workbench/contrib/mcp/common/mcpSandboxService.ts
dileepyavan: yeah for this, I havent considered user level yet. Only workspace level. I think  there should be three different configs for each level,

src/vs/workbench/contrib/mcp/common/mcpTypes.ts
dileepyavan: Got it. As I understand , mcp.json still has sandboxEnabled property but the MCPserverDefinition will just have sandbox property that holds the config. Presence should say its enabled else not.

src/vs/workbench/contrib/mcp/common/mcpSandboxService.ts
dileepyavan: I see configuration target as export const enum ConfigurationTarget {
	APPLICATION = 1,
	USER,
	USER_LOCAL,
	USER_REMOTE,
	WORKSPACE,
	WORKSPACE_FOLDER,
	DEFAULT,
	MEMORY
}
Should the sandbox configuration be scoped at each of these or only user, workspace and workspace folder?

src/vs/workbench/contrib/mcp/common/mcpTypes.ts
dileepyavan: updated as `sandboxEnabled: def.sandboxEnabled,
			sandbox: def.sandboxEnabled ? def.sandbox : undefined,`

src/vs/workbench/api/node/extHostMcpNode.ts
connor4312: Oh, is this coming from devmode when working on Code - OSS itself?

I would still handle this in the registry/sandbox service rather than backchanneling the `mcpSandboxedLaunchEnvironmentKey` and doing it here

src/vs/platform/mcp/common/mcpManagementService.ts
```diff
@@ -392,14 +392,27 @@ export abstract class AbstractMcpResourceManagementService extends AbstractCommo
 
 			for (const [name, server] of current) {
 				const previous = this.local.get(name);
+				let config = server.config;
+				if (server.config.type === McpServerType.LOCAL) {
+					config = {
+						...server.config,
+						sandboxEnabled: server.config.sandboxEnabled,
+						sandbox: server.config.sandboxEnabled ? server.config.sandbox : undefined
+					};
```
connor4312: I'm not clear why this code is needed here -- `sandboxEnabled` isn't getting changed here, and there is already [code](https://github.com/microsoft/vscode/blob/7fe0ebf878a892bf44068edae86912f5d2b64ff9/src/vs/platform/mcp/common/mcpResourceScannerService.ts#L222-L224) in the scanner that only sets the `sandbox` when `sandboxEnabled` (which makes sense) so this looks like it would be a no-op?

src/vs/workbench/contrib/mcp/common/mcpTypes.ts
connor4312: Great, I think we can get rid of `sandboxEnabled` in this type now

src/vs/workbench/contrib/mcp/common/mcpServer.ts
```diff
@@ -307,6 +307,7 @@ export class McpServer extends Disposable implements IMcpServer {
 
 			d = autorun(reader => {
 				const connection = server.connection.read(reader);
+				connection?.handler.read(reader); // ensure handler is ready before calling fn
```
connor4312: I'm not clear what this does, we already read the handler down in L315

src/vs/workbench/contrib/mcp/common/mcpSandboxService.ts
```diff
@@ -0,0 +1,210 @@
...
+		}
+		if (await this.isEnabled(serverDef, remoteAuthority)) {
+			this._logService.trace(`McpSandboxService: Launching with config target ${configTarget}`);
+			const launchDetails = await this._resolveSandboxLaunchDetails(configTarget, remoteAuthority, serverDef.sandbox);
+			const sandboxArgs = this._getSandboxCommandArgs(launch.command, launch.args, launchDetails.sandboxConfigPath);
+			const sandboxEnv = this._getSandboxEnvVariables(launchDetails.tempDir);
+			if (launchDetails.srtPath) {
+				return {
+					...launch,
+					command: launchDetails.srtPath,
```
connor4312: I think this still has the issue that this will only work if the host OS has Node.js installed, as you run the `.js` file as a program directly which will try to find `node` in the user's environment.

You handle this case already in the `TerminalSandboxService`. I suggest trying to extract a base abstract class or other reusable component you can use for both scenarios and reduce duplication.

src/vs/workbench/api/node/extHostMcpNode.ts
dileepyavan: This is not coming from dev mode. I am not sure how to handle this when definition.devMode is undefined in registry.ts

src/vs/workbench/contrib/mcp/common/mcpTypes.ts
dileepyavan: I prefer sandboxEnabled to be there as it is clearer to check a boolean value than infer from the sandbox object when checking is sandbox is enabled or not. Let me know if you think otherwise.

src/vs/workbench/api/node/extHostMcpNode.ts
dileepyavan: I dont  see this env variable being set because of dev mode.

src/vs/workbench/contrib/mcp/common/mcpServer.ts
dileepyavan: removed

src/vs/platform/mcp/common/mcpManagementService.ts
dileepyavan: removed

connor4312 (approved):

PR #295842, opened 2026-02-17
Get accurate Windows version info from registry instead of os.release()

### LINKED ISSUE #197444: VS code 1.84 automatically installs 1.79 (Old)
Type: <b>Bug</b>

I had VS code 1.79, and I downloaded the latest (1.84) from the website and I installed it and opened it, but it automatically checks for updates and rolls me back to 1.79, and when I restart it installs 1.79 and says this version is no longer supported in win8 consider updating your windows, even tho I have win10, and so many extensions break.........

VS Code version: Code 1.79.2 (695af097c7bd098fbf017ce3ac85e09bbc5dda06, 2023-06-14T08:57:04.379Z)
OS version: Windows_NT x64 6.2.9200
Modes:

<details>

|Item|Value|
|---|---|
|CPUs|Intel(R) Core(TM) i5-2540M CPU @ 2.60GHz (4 x 2594)|
|GPU Status|2d_canvas: enabled<br>canvas_oop_rasterization: disabled_off<br>direct_rendering_display_compositor: disabled_off_ok<br>gpu_compositing: enabled<br>multiple_raster_threads: enabled_on<br>opengl: enabled_on<br>rasterization: unavailable_off<br>raw_draw: disabled_off_ok<br>video_decode: enabled<br>video_encode: enabled<br>vulkan: disabled_off<br>webgl: enabled<br>webgl2: unavailable_off<br>webgpu: enabled|
|Load (avg)|undefined|
|Memory (System)|7.93GB (1.60GB free)|
|Process Argv|--crash-reporter-id eb1dffbe-b625-405a-a2a4-e6e4e21f3b47|
|Screen Reader|no|
|VM|0%|
</details><details><summary>Extensions (9)</summary>

---|---|---
vscode-csharp-snippets|jor|1.1.0
unity-code-snippets|kle|2.2.5
csdevkit|ms-|1.0.14
csharp|ms-|2.9.20
vscode-dotnet-runtime|ms-|2.0.0
vscodeintellicode-csharp|ms-|0.1.26
hlsltools|Tim|1.1.303
vstuc|vis|0.9.2
material-theme|zhu|3.16.2

</details><details>

```
vsliv368:30146709
vsreu685:30147344
python383cf:30185419
vspor879:30202332
vspor708:30202333
vspor363:30204092
vslsvsres303:30308271
vserr242cf:30382550
pythontb:30283811
vsjup518:30340749
pythonptprofiler:30281270
vshan820:30294714
vstes263cf:30335440
vscod805cf:30301675
binariesv615:30325510
bridge0708:30335490
bridge0723:30353136
vsaa593:30376534
pythonvs932:30410667
py29gd2263cf:30880073
vscaac:30438847
vsclangdc:30486549
c4g48928:30535728
dsvsc012cf:30540253
pynewext54:30695312
azure-dev_surveyone:30548225
3biah626:30602489
89544117:30613380
vscrp:30673768
a9j8j154:30646983
showlangstatbar:30737416
pythonfmttext:30731395
fixshowwlkth:30771522
showindicator:30805244
pythongtdpath:30769146
i26e3531:30792625
pythonnosmt12:30797651
pythonidxpt:30866567
pythonnoceb:30805159
synctok:30869157
dsvsc013:30795093
dsvsc014:30804076
dsvsc015:30845448
pythontestfixt:30871694
pythonregdiag2:30871582
pyreplss1:30879912
pythonmypyd1:30879173
pythoncet00cf:30874137
pythontbext0:30879054
dsvsc016:30880769
dsvsc017cf:30880893
dsvsc018:30880894

```

</details>

dmitrivMS:
Fixes #197444

**Changes**
Used Windows registry to retrieve Windows release and build numbers.
Added passing User-Agent header to update service.
Updated installer launch env to ignore compat mode.

**Note**
When running in compatibility mode os.release() may return incorrect version info which leads to VS Code downgrade (downloading an old version) and terminal not working.

**Validation**
Start VS Code with compatibility settings 8.1, wait for update to be installed -> verify the update is next version (not an old one), click Restart -> VS Code should restart into the new version. You should see no errors anywhere along the way. Verify terminal works under these conditions.


bpasero (dismissed):
I am worried about the impact on startup performance such a change has. We now block window opening with this new code that asks the windows registry for something which we do not know when it returns. If we need this info in a reliable way, then it should be moved to the places where we need it and wait there but with sensitive fallback mechanisms if the windows registry is very slow.

remote/package-lock.json
```diff
@@ -21,6 +21,19 @@
         "@vscode/tree-sitter-wasm": "^0.3.0",
         "@vscode/vscode-languagedetection": "1.0.23",
         "@vscode/windows-process-tree": "^0.6.0",
+<<<<<<< HEAD
```
deepak1556: Merge conflicts not resolved, also the cause for failure in https://dev.azure.com/monacotools/Monaco/_build/results?buildId=409973&view=logs&j=dfc9db24-963b-5e49-20de-b4f73c206a9f&t=1f7b2f94-c7af-5813-0271-5ec5c4f5481a&s=d654deb9-056d-50a2-1717-90c08683d50a

dmitrivMS: Yep, thanks!

dmitrivMS: @deepak1556 fixed

deepak1556 (commented):
LGTM on the usage in update code, also like that the registry information is cached for the lifetime of the process.

No blockers from my end, what would be nice is we only get the information from registry if we are running in compatibility mode but again checking if we are in compat mode requires probing registry :(

bpasero (dismissed):
I think this is better but its fragile: we do not have a deterministic release info on the window anymore now because `initWindowsVersionInfo` may or may not succeed at the time the window opens. I think a better approach is to leave `os.release()` in for the window and use the async variant in places were we really need to rely on it and where we can wait on it without causing issues.

dmitrivMS:
> I think this is better but its fragile: we do not have a deterministic release info on the window anymore now because `initWindowsVersionInfo` may or may not succeed at the time the window opens. I think a better approach is to leave `os.release()` in for the window and use the async variant in places were we really need to rely on it and where we can wait on it without causing issues.

Sure, no problem - reverted the change. This property is not used for anything of significance on Windows, so it should be fine. Ideally, we should remove it - since it cannot be properly initialized, it may return incorrect value on Windows.

dmitrivMS:
@deepak1556 We read once and cache it, I don't see it as a perf concern to be honest.
Regarding upstream fixes - it feels like too much to depend on, and getting the value from the registry should give us a reliable solution now and going forward.

joaomoreno (approved):

PR #295871, opened 2026-02-17
new model picker in sessions window

sandy081:


TylerLeonhardt (approved):
Can you include a screenshot if the UI is different?

PR #295937, opened 2026-02-18
Auto-accept (keep) pending chat edits when files are committed via SCM

pierceboggan:
https://github.com/user-attachments/assets/35ff09a0-b7a3-4690-88c3-b21ade93ef48


jrieken (changes requested):
@pierceboggan There is correct and important feedback from CCR but apart from that I wonder this is duplicating what we already have? Today, when you commit changes via SCM they are kepts implicitly already

fyi @lszomoru

PR #295941, opened 2026-02-18
Add semantic search over agent session history (and #sessions tool) with TF-IDF

pierceboggan:
https://github.com/user-attachments/assets/3d793419-fff9-4fa4-889b-f6c427a0ba0b

Very helpful for figuring out how to improve my prompting:
image

This is also wired up to search in agent sessions + the quick pick search UX for more actionable search results.


mjbvz:

PR #295968, opened 2026-02-18
Fix z-index issue in terminal inline chat suggest

### LINKED ISSUE #295960: The scroll bar and active item indicator are rendered on top of the terminal inline chat suggest

- VS Code Version: 1.110.0-insider (Universal)
- OS Version: Darwin arm64 25.3.0

Steps to Reproduce:

1. Open the inline chat in the terminal using <kbd>Command</kbd> + <kbd>I</kbd>.
2. Type a slash to open the suggest.

Image

## Expected result

Image

bangseongbeom:
Increase terminal z-index to resolve #295960.

Set z-index to 12 since [the scrollbar is 11](https://github.com/microsoft/vscode/blob/2500a2c880f61bb9c3c5c7b2951bc0f91798c336/src/vs/base/browser/ui/scrollbar/media/scrollbars.css#L20).


meganrogge:
@bangseongbeom let's add specificity to the css selector so it only does this in the terminal inline chat

PR #296006, opened 2026-02-18
Inline Compile & CompileCLI stages into platform specific jobs

### LINKED ISSUE #296465: Telemetry tooling fails to report inconsistent telemetry events
There are multiple different definitions of `gettingStarted.ActionExecuted`. Our CI fails to catch this.
- It would be great if TypeScript would catch this.
- Otherwise, it would be great if our telemetry extractor would catch this. Today it just happily moves forward and spits out no error.

Image

joaomoreno:
This PR eliminates both the Compile and the CompileCLI stage from our build, removing the overhead of transferring artifacts across jobs, which is now more than the time it takes to run the "meat" of those stages, now that they are faster (Compile now uses esbuild, CompileCLI now uses sccache).

We now have flat jobs, they can all start anew:

Product build:

image

CI build:

image

Also fixes #296465


src/vs/workbench/contrib/welcomeAgentSessions/browser/agentSessionsWelcome.ts
joaomoreno, line 1: fyi @osortega

src/vs/workbench/services/themes/browser/workbenchThemeService.ts
joaomoreno, line 1: fyi @aeschli

lszomoru (approved):

PR #296055, opened 2026-02-18
fix(chat): fix scroll jump issues in chat

### LINKED ISSUE #295327: Poor Scroll Experience in Copilot Chat (Codespaces/VS Code): Jumping & Chunky Scrolling in Long Chats
### Description
As a chat evolves, the Copilot Chat panel can contain a long exchange. When reviewing previous messages (scrolling up), the scroll bar does not move smoothly: it jumps in large, uneven chunks. This often causes all or part of the chat content to move out of view unpredictably, which is confusing and frustrating, making it difficult to read or refer to previous messages. This degrades the overall user experience for anyone working in extended chats.

A smooth, predictable, pixel-based scrolling (rather than chunked jumps) would be much easier for navigating long conversations and would improve usability—especially when working on complex multi-step tasks that require referencing earlier chat history.

### Impact
- Frequent jumps disrupt review/reference of prior content
- Hard to maintain understanding/context in extended support/debugging/chat sessions
- Particularly problematic in Codespaces, where Copilot Chat is essential for cloud-based dev workflows

### Expected Behavior
- Scrolling up and down in the chat should be smooth and consistent, with no content "jumps" out of view
- Scroll bar should reflect position within the conversation and not snap to unpredictable chunk boundaries

### Observed Behavior
- Scroll bar moves in noticeable chunks, not a smooth glide
- Jumps may result in losing track of the chat context

### Context
This occurs in
- Codespaces (Web, VS Code)
- Copilot Chat UI

### Supporting Details
- Reproducible on latest Copilot Chat for VS Code and Codespaces as of Feb 2026.
- This affects reading long-standing problem/debugging/support threads, harming usability.

---

Version: 1.109.3
Commit: b6a47e94e326b5c209d118cf0f994d6065585705
User Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/144.0.0.0 Safari/537.36
Embedder: codespaces

Extension version: 0.37.6

murataslan1:
Fixes #295327. Ensures synchronous height estimation for pending chat messages to prevent layout shift and scroll jumping.

### DIFF
--- a/src/vs/workbench/contrib/chat/browser/widget/chatContentParts/codeBlockPart.ts
+++ b/src/vs/workbench/contrib/chat/browser/widget/chatContentParts/codeBlockPart.ts
@@ -92,6 +92,7 @@ export interface ICodeBlockData {
 	readonly parentContextKeyService?: IContextKeyService;
 	readonly renderOptions?: ICodeBlockRenderOptions;
 
+	readonly text?: string;
 	readonly chatSessionResource: URI;
 }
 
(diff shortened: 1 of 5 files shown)


justschen:
cc @roblourens i know we've been looking into somethign like this

src/vs/workbench/contrib/chat/browser/widget/chatContentParts/codeBlockPart.ts
```diff
@@ -407,6 +408,11 @@ export class CodeBlockPart extends Disposable {
 			const lineCount = this.currentCodeBlockData.range.endLineNumber - this.currentCodeBlockData.range.startLineNumber + 1;
 			const lineHeight = this.editor.getOption(EditorOption.lineHeight);
 			return lineCount * lineHeight + 2 * this.verticalPadding;
+		} else if (this.currentCodeBlockData?.text) {
+			const lineHeight = this.editor.getOption(EditorOption.lineHeight);
```
roblourens, line 415: Sorry, this doesn't make sense to me, `this.currentCodeBlockData?.text` will probably always be set, so we always go into this block and never fall through to the real `getContentHeight`...?

PR #296058, opened 2026-02-18
fix tip dismissal

### LINKED ISSUE #296057: Tips toolbar: Dismissed tips are not staying dismissed when feature is re-enabled
**Describe the bug**
When a user dismisses a tip in the tips toolbar, it should not appear again for the same user, even after re-enabling the "chat.tips.enabled" setting. However, after re-enabling, dismissed tips repeatedly come back when clicking the checkmark, rather than leaving the user in a tip-free state as expected. This appears to be a regression from previous behavior.

**To Reproduce**
1. Navigate to the tips toolbar and dismiss some tips by clicking the checkmark.
2. Disable tips from the toolbar or context menu.
3. Re-enable tips by toggling the "chat.tips.enabled" setting.
4. Tips previously dismissed keep reappearing despite being acknowledged before.

**Expected behavior**
Once a tip has been dismissed, it should not show up again for the user, regardless of toggling the tips feature on and off. Dismissal should be persistent per user.

**Additional context**
- Some users would prefer not to disable tips outright if they knew dismissed tips would never reappear. Persistent dismissal enables them to keep tips on for new content without being bothered by repeats.
- There was discussion about whether navigating tips should mark them as read/dismissed (answer: no).
- This was confirmed as a regression in the Slack thread.

- VS Code Version: [Please fill]
- OS Version: [Please fill]

Steps to Reproduce:
1. Follow dismissal and re-enable steps as described above.
2. Observe repeated display of previously dismissed tips, which should not occur.

[View original Slack conversation](https://vscodeteam.slack.com/archives/C1C2M1KEZ/p1771433472070599?thread_ts=1771433472.070599&cid=C1C2M1KEZ)

meganrogge:
fixes #296057

Dismissed tips were stored as a plain array, so duplicate IDs built up. A recovery check then treated “too many dismissed” as “none dismissed,” and a fallback always picked a tip, so previously dismissed tips could reappear after re-enabling tips.

We now dedupe persisted dismissals, only keep valid tip IDs, remove the reset-to-empty behavior, and return no tip when all are dismissed. That keeps dismissals persistent across disable/re-enable, and we added a regression test to lock this behavior.

### DIFF
--- a/src/vs/workbench/contrib/chat/browser/chatTipService.ts
+++ b/src/vs/workbench/contrib/chat/browser/chatTipService.ts
@@ -67,7 +67,7 @@ export interface IChatTipService {
 
 	/**
 	 * Dismisses the current tip and allows a new one to be picked for the same request.
-	 * The dismissed tip will not be shown again in this profile.
+	 * The dismissed tip will not be shown again for this user on this application installation.
 	 */
 	dismissTip(): void;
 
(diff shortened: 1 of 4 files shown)


meganrogge:
Woah, something went wrong here 😅

Yoyokrazy (approved):

PR #296097, opened 2026-02-18
update yolo mode description to include autoReply link

meganrogge:
YOLO mode (`chat.tools.global.autoApprove`) and auto-reply (`chat.autoReply`) are independent settings, but users expect YOLO mode to also auto-answer agent questions. This adds clarification directly in the setting description.

- Appended a note to `globalAutoApproveDescription` explaining YOLO mode only controls tool approval, not question behavior
- Added a clickable `#chat.autoReply#` setting link so users can discover the auto-reply setting directly from the YOLO mode description
- Updated matching policy localization description for consistency

Screenshot 2026-02-18 at 1 55 57 PM

### DIFF
--- a/build/lib/policies/policyData.jsonc
+++ b/build/lib/policies/policyData.jsonc
@@ -87,8 +87,8 @@
             "minimumVersion": "1.99",
             "localization": {
                 "description": {
-                    "key": "autoApprove2.description",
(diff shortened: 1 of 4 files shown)


connor4312 (commented):
description lgtm but will leave the code change to raymond

rzhao271 (approved):

joshspicer (approved):

PR #296106, opened 2026-02-18
editor: add 'foldedLine' unit to cursorMove command

### LINKED ISSUE #81498: [api] Allow extensions to determine if a position is within a fold
This has been the most demanded feature of VSCodeVim: [VSCodeVim/Vim#1004](https://github.com/VSCodeVim/Vim/issues/1004) for over 3 years.

The main problem with Folds and Vim is that some motions will skip right over folded areas (like moving up/down). We need to know if we are in a folded area so we can iterate these motions until we are out of the fold.

An API like ``vscode.window.activeTextEditor.getAllFoldedRegions(): vscode.Range[]`` would be ideal.

An API like ``vscode.window.activeTextEditor.isPositionInFold(position: vscode.Position): boolean`` would also be great.

* * *

This is technically a duplicate for [22276](https://github.com/microsoft/vscode/issues/22276), but that was closed because the roadmap at the time couldn't include this ticket.

Maybe it's time to revisit this? It has been blocking the most demanded feature of one of the most popular plugins out there, for 3+ years.

jaidhyani:
## Summary

Adds a new `foldedLine` value to the `by` parameter of the `cursorMove` command. When used with `up` or `down` directions, the cursor moves by logical lines while treating each folded region as a single step — the entire fold is skipped rather than entered.

Previously only `wrappedLine` (view-space) and `line` (model-space) were available. `wrappedLine` skips folds as a side-effect of operating in view coordinates, but also splits long wrapped lines into multiple steps, which is incorrect for line-oriented movements. `line` moves in model-space and enters folds. `foldedLine` moves in model-space but queries `viewModel.getHiddenAreas()` to detect and skip over folded regions.

## Behavior

- Each fold counts as exactly one step regardless of how many model lines it spans
- Fold spanning to end-of-file: cursor stays at current position (no visible line to land on)
- Fold starting at line 1: cursor stays at current position
- Compatible with multi-cursor and selection (`select: true`)
- Only meaningful with `up`/`down` directions; other directions ignore the unit and fall back to default behavior

## Testing

New test suite: `Cursor move command - foldedLine unit` in `cursorMoveCommand.test.ts`

- move down skips a fold below the cursor
- move up skips a fold above the cursor  
- count-based movement treats each fold as one step
- multi-line fold counts as a single step
- boundary conditions (last line, first line)

## Related

Closes #81498


jaidhyani:

alexdima (approved):
Hope you don't mind, I used my AI to fix your AI's small mistake

roblourens (approved):

Issue #296134, opened 2026-02-18
Remove chat `ask_questions` tool from vscode-copilot and `vscode`


meganrogge:
Once @TylerLeonhardt and @DonJayamanne have moved over to our core tool, https://github.com/microsoft/vscode/pull/296135, we can do this.

PR #296135, opened 2026-02-18
implement `vscode_askQuestions` tool in core

### LINKED ISSUE #291349: Move `ask_questions` tool implementation into VS Code core

meganrogge:
fixes https://github.com/microsoft/vscode/issues/291349

Tested by setting `github.copilot.chat.askQuestions.enabled: false`

Follow-ups here https://github.com/microsoft/vscode/issues/296134 as some things in chat depend upon it

cc @karthiknadig

### DIFF
--- a/src/vs/workbench/contrib/chat/common/chatService/chatServiceImpl.ts
+++ b/src/vs/workbench/contrib/chat/common/chatService/chatServiceImpl.ts
@@ -304,7 +304,7 @@ export class ChatService extends Disposable implements IChatService {
 
 	private deserializeChats(sessionData: string): ISerializableChatsData {
 		try {
-			const arrayOfSessions: ISerializableChatDataIn[] = revive(JSON.parse(sessionData)); // Revive serialized URIs in session data
+			const arrayOfSessions = revive(JSON.parse(sessionData)) as ISerializableChatDataIn[]; // Revive serialized URIs in session data
 			if (!Array.isArray(arrayOfSessions)) {
 				throw new Error('Expected array');
 			}
@@ -503,17 +503,6 @@ export class ChatService extends Disposable implements IChatService {
 
 	getSession(sessionResource: URI): IChatModel | undefined {
 		return this._sessionModels.get(sessionResource);
-	}
-
-	findRequestById(requestId: string): IChatRequestModel | undefined {
-		for (const model of this._sessionModels.values()) {
-			const request = model.getRequests().find(candidate => candidate.id === requestId);
-			if (request) {
-				return request;
-			}
-		}
-
-		return undefined;
 	}
 
 	getActiveSessionReference(sessionResource: URI): IChatModelReference | undefined {
--- a/src/vs/workbench/contrib/chat/common/chatService/chatServiceImpl.ts
+++ b/src/vs/workbench/contrib/chat/common/chatService/chatServiceImpl.ts
@@ -505,6 +505,17 @@ export class ChatService extends Disposable implements IChatService {
 		return this._sessionModels.get(sessionResource);
 	}
 
+	findRequestById(requestId: string): IChatRequestModel | undefined {
+		for (const model of this._sessionModels.values()) {
+			const request = model.getRequests().find(candidate => candidate.id === requestId);
+			if (request) {
+				return request;
+			}
+		}
+
+		return undefined;
+	}
+
 	getActiveSessionReference(sessionResource: URI): IChatModelReference | undefined {
 		return this._sessionModels.acquireExisting(sessionResource);
 	}
--- a/src/vs/workbench/contrib/chat/common/chatService/chatServiceImpl.ts
+++ b/src/vs/workbench/contrib/chat/common/chatService/chatServiceImpl.ts
@@ -304,7 +304,7 @@ export class ChatService extends Disposable implements IChatService {
 
 	private deserializeChats(sessionData: string): ISerializableChatsData {
 		try {
-			const arrayOfSessions = revive(JSON.parse(sessionData)) as ISerializableChatDataIn[]; // Revive serialized URIs in session data
+			const arrayOfSessions: ISerializableChatDataIn[] = revive(JSON.parse(sessionData)); // Revive serialized URIs in session data
 			if (!Array.isArray(arrayOfSessions)) {
 				throw new Error('Expected array');
 			}
--- a/src/vs/workbench/contrib/chat/common/tools/builtinTools/askQuestionsTool.ts
+++ b/src/vs/workbench/contrib/chat/common/tools/builtinTools/askQuestionsTool.ts
@@ -5,13 +5,13 @@
 
 import { CancellationToken } from '../../../../../../base/common/cancellation.js';
 import { CancellationError } from '../../../../../../base/common/errors.js';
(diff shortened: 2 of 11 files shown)


src/vs/workbench/contrib/chat/common/chatService/chatServiceImpl.ts
```diff
@@ -505,6 +505,17 @@ export class ChatService extends Disposable implements IChatService {
 		return this._sessionModels.get(sessionResource);
 	}
 
+	findRequestById(requestId: string): IChatRequestModel | undefined {
```
roblourens: Why this? 

> The new ask‑questions tool needs to reason about and sometimes update an existing chat request given only its requestId (e.g., when invoked later or from another agent/tool), not a specific session handle. Previously, the service only let callers work within a known session; there was no way for a tool to say “given this request ID, find the corresponding request wherever it lives.” Adding findRequestById to IChatService and implementing it over all session models gives the tool a stable, service‑level way to recover that context so it can attach follow‑ups, read variables, or correlate results correctly.

I don't understand why that would be the case

dmitrivMS (dismissed):

src/vs/workbench/contrib/chat/common/tools/builtinTools/askQuestionsTool.ts
```diff
@@ -0,0 +1,479 @@
...
+		}
+
+		const carousel = this.toQuestionCarousel(questions);
+		this.chatService.appendProgress(request, carousel);
+
+		const disposables = new DisposableStore();
+		let answerResult: { answers: Record<string, unknown> | undefined } | undefined;
+		try {
+			const answersEvent = Event.toPromise<{ requestId: string; resolveId: string; answers: Record<string, unknown> | undefined }>(
+				Event.filter(this.chatService.onDidReceiveQuestionCarouselAnswer, e => e.requestId === request.id && e.resolveId === carousel.resolveId),
```
roblourens: How about waiting on a promise on `IChatQuestionCarousel`? More like the way `IToolInvocations` work.

src/vs/workbench/contrib/chat/common/chatService/chatServiceImpl.ts
meganrogge: Thanks for asking.

I’ve tightened `askQuestions` so it now requires a `sessionResource` and no longer needs `findRequestById`.

I originally added `findRequestById` because `IToolInvocation.context.chatSessionResource` is optional, and while debugging I was seeing it come through as `undefined` in some cases. After digging through this branch, it looks like all of the current call sites that can reach `askQuestions` already pass a `sessionResource` (they go through the normal chat/agent path), and MCP tools don’t actually route through the built‑in tools at all, so we don’t need the cross‑session lookup right now.

I don't know why I saw that when debugging 🤷🏼‍♀️

src/vs/workbench/contrib/chat/common/chatService/chatServiceImpl.ts
```diff
@@ -304,7 +304,7 @@ export class ChatService extends Disposable implements IChatService {
 
 	private deserializeChats(sessionData: string): ISerializableChatsData {
 		try {
-			const arrayOfSessions: ISerializableChatDataIn[] = revive(JSON.parse(sessionData)); // Revive serialized URIs in session data
+			const arrayOfSessions = revive(JSON.parse(sessionData)) as ISerializableChatDataIn[]; // Revive serialized URIs in session data
```
meganrogge:
```suggestion
			const arrayOfSessions: ISerializableChatDataIn[] = revive(JSON.parse(sessionData)); // Revive serialized URIs in session data
```

src/vs/workbench/contrib/chat/common/chatService/chatServiceImpl.ts
roblourens: In the vscode tools world, technically you can call `invokeTool` at any time outside of a chat session. That scenario didn't turn out to be very useful IMO. But that's why it's optional on the interface. So it's fine to throw if the sessionResource isn't provided and if you find a normal agent scenario where that happens, it's a bug.

src/vs/workbench/contrib/chat/common/chatService/chatServiceImpl.ts
roblourens: Hm, I think something went wrong here if you suddenly have to cast this line

src/vs/workbench/contrib/chat/common/tools/builtinTools/askQuestionsTool.ts
```diff
@@ -0,0 +1,487 @@
...
+		}
+
+		const carousel = this.toQuestionCarousel(questions);
+		this.chatService.appendProgress(request, carousel);
+
+		const disposables = new DisposableStore();
+		let answerResult: { answers: Record<string, unknown> | undefined } | undefined;
+		try {
+			const answersPromise = carousel.completion?.promise ?? Event.toPromise<{ requestId: string; resolveId: string; answers: Record<string, unknown> | undefined }>(
+				Event.filter(this.chatService.onDidReceiveQuestionCarouselAnswer, e => e.requestId === request.id && e.resolveId === carousel.resolveId),
```
roblourens: Why do you stil need the event now that you have this promise?

src/vs/workbench/contrib/chat/common/tools/builtinTools/askQuestionsTool.ts
```diff
@@ -0,0 +1,487 @@
...
+		const mappedQuestions = questions.map(question => this.toChatQuestion(question));
+		const completion = new DeferredPromise<{ answers: Record<string, unknown> | undefined }>();
+		const carousel: IChatQuestionCarousel = {
+			kind: 'questionCarousel',
+			questions: mappedQuestions,
+			allowSkip: true,
+			resolveId: generateUuid()
+		};
+
+		Object.defineProperty(carousel, 'completion', {
```
roblourens: What happened here? 😅 

I recommend a DeferredPromise. Why is `completion` optional on `IChatQuestionCarousel`?

roblourens: Oh maybe for serialization, you can define a separate type for that case (eg after window reload) if you want to be explicit about it. Implement toJSON on `IChatQuestionCarousel` if you need to control how its serialized.

src/vs/workbench/contrib/chat/common/chatService/chatServiceImpl.ts
roblourens: This is sensitive code related to persisting user data, please be careful with it.

DonJayamanne (dismissed):

roblourens (approved):
Looks good, thanks for the edits

PR #296178, opened 2026-02-19
[Debug Chat] Attach Nested Variable and Scope Values from Paused Sessions

### LINKED ISSUE #296173: Make AI be able to read values from debugger variables
In Visual Studio 2022/2026 AI is able to read the values of variables during debugging you can even give a local scope with all variables content, or even reference a specific variable, I would like to request the same feature for VSCode.

elijahfriedman:
Fixes #296173

## Summary
Debug Chat now includes nested variable values (with safe limits) making paused-session context more useful than the previous name (value output.)

- Added bounded recursive serialization for debug variables attached to chat:
  - max depth
  - max children per node
  - max total nodes
  - max top-level scope variables
- Updated Debug Session context picker attachments:
  - Watch expressions now attach serialized values
  - Scope attachments now include nested locals/children
  - Variable attachments now include nested values when available
- Updated Debug context menu actions (Add to Chat)to use serialized output.
- Included truncation markers when output is cut off.
- Kept paused source location attachment behavior unchanged.


elijahfriedman:

PR #296215, opened 2026-02-19
Use `tsgo` to run type checks on client for watch

mjbvz:
This reduces the first run time to get `src` transpiled and fully type checked to around 15seconds on my machine. Rechecking then takes around 2-3 seconds


src/tsconfig.json
```diff
@@ -8,6 +8,7 @@
 		"allowJs": true,
 		"resolveJsonModule": true,
 		"isolatedModules": false,
+		"skipLibCheck": true,
```
mjbvz, line 11: adding this temporarily to work around https://github.com/microsoft/tas-client/pull/88

build/lib/compilation.ts
```diff
@@ -190,6 +191,23 @@ export function watchTask(out: string, build: boolean, srcPath: string = 'src',
...
+		const watchInput = watch(`${src}/**`, { base: src, readDelay: 200 });
+		const tsgoStream = watchInput.pipe(util.debounce(() => {
+			const stream = createTsgoStream(projectPath, { taskName: 'watch-client-noEmit', noEmit: true });
+			const result = es.through();
+			stream.on('end', () => result.emit('end'));
+			stream.on('error', () => result.emit('end'));
+			return result;
+		}));
+		return tsgoStream;
+	};
```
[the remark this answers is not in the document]
jrieken: @mjbvz I think this is good feedback

build/lib/compilation.ts
```diff
@@ -190,6 +191,23 @@ export function watchTask(out: string, build: boolean, srcPath: string = 'src',
 	return task;
 }
 
+export function watchTypeCheckTask(src: string): task.StreamTask {
+	const theTask = () => {
+		const projectPath = path.join(import.meta.dirname, '../../', src, 'tsconfig.json');
+		const watchInput = watch(`${src}/**`, { base: src, readDelay: 200 });
+		const tsgoStream = watchInput.pipe(util.debounce(() => {
+			const stream = createTsgoStream(projectPath, { taskName: 'watch-client-noEmit', noEmit: true });
```
jrieken, line 181: Is incremental or not? Like, will every file change cost 15 second (for type checking)?

build/lib/tsgo.ts
jrieken, line 17: This should probably replace what I have added for CI/CD here https://github.com/microsoft/vscode/blob/7ffca955fd4d5a22fa0fd806c0cf7557451ee6a9/build/gulpfile.vscode.ts#L219

build/lib/compilation.ts
mjbvz: It's closer to 5 seconds for updates. The initial 15 seconds is for the first build and is slower because all of the compile tasks are running at the same time

mjbvz: I've now also enabled the `--incremental` option to speed up recompiles. In my testing, this reduces the recheck time to 2seconds. Hopefully longer term the TS team will improve their watcher so that we can use that and speed this up even further

roblourens (dismissed):

build/lib/compilation.ts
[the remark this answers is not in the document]
mjbvz: Updated this and tested. Should work now

bryanchen-d (approved):

PR #296248, opened 2026-02-19
Sanity tests perf improvements

dmitrivMS:
**Changes**
Reduced time taken by Windows ARM64 job:
- Used custom Node.js install step to ensure matching architecture
- Removed/reduced powershell calls
- Use curl for downloads

**Note**
Addressed some PR feedback from previous changes.
Removed 64k page test.

**Validation**
With these changes all sanity tests run in ~10 minutes.


build/azure-pipelines/common/sanity-tests.yml
```diff
@@ -43,11 +43,36 @@ jobs:
...
+
+            if not exist "%NODE_ROOT%\node.exe" (
+              if exist "%EXTRACT_ROOT%" rmdir /s /q "%EXTRACT_ROOT%"
+              mkdir "%EXTRACT_ROOT%"
+              curl.exe -fsSL "%DOWNLOAD_URL%" -o "%ARCHIVE_PATH%"
+              tar -xf "%ARCHIVE_PATH%" -C "%EXTRACT_ROOT%"
+            )
+
+            echo ##vso[task.prependpath]%NODE_ROOT%
+          displayName: Install Node.js (Windows ARM64)
```
joaomoreno, line 75: Should we checksum it?

build/azure-pipelines/common/sanity-tests.yml
```diff
@@ -43,11 +43,36 @@ jobs:
...
+
+      - ${{ if not(and(eq(parameters.os, 'windows'), eq(parameters.arch, 'arm64'))) }}:
```
joaomoreno: Should this be an `else`?

build/azure-pipelines/common/sanity-tests.yml
dmitrivMS: How? I believe this is what the built-in task does, no?

hediet (dismissed):

bpasero (dismissed):

build/azure-pipelines/common/sanity-tests.yml
joaomoreno: We actually have some logic to download Node.JS here, including verifying the checksums:
- https://github.com/microsoft/vscode/blob/ec6ebb10ef79d845320cb393760cbdf2d0cd687a/build/gulpfile.reh.ts#L191
- https://github.com/microsoft/vscode/blob/ec6ebb10ef79d845320cb393760cbdf2d0cd687a/build/checksums/nodejs.txt#L1-L8

Could that be inspiration?

dmitrivMS: Ok, thanks, I'll take a look!

joaomoreno (approved):

PR #296376, opened 2026-02-19
Remove Keep/Undo buttons from chat editing UI

daviddossett:
## Summary

Removes the Keep/Undo paradigm from the chat editing UI. Edits from chat are applied directly — there is no longer an accept/reject decision for the user to make.

## What changed

**Removed UI elements:**
- "Keep" and "Undo" buttons from the chat input widget toolbar (the top-level Keep All / Undo All)
- "Keep" and "Undo" per-file buttons from the modified files list
- "Keep" and "Undo" from the editor overlay toolbar (the floating bar at bottom-right of edited files)
- Per-hunk "Keep" and "Undo" buttons from inline change widgets in the editor
- "Keep All Edits" / "Undo All Edits" from the multi-diff editor title bar
- Per-file diff icon button from the modified files list (redundant with clicking the file name)

**Added/changed UI:**
- "Review" label button replaces the diff icon in the "N files changed" toolbar — opens the multi-file diff view
- Multi-file diff editor tab renamed from "Suggested Edits" to "Review Edits (N files)"
- `chat.checkpoints.showFileChanges` now defaults to `true` (was `false`)

**Style polish:**
- Checkpoint file changes list: matched padding, font size (11px), and font weight to the main "N files changed" working set list
- Checkpoint toggle button: uses `descriptionForeground`, no hover background, zero left/right padding, 2px gap
- List item border-radius uses `cornerRadius-small` in both file lists
- Multi-diff editor empty state placeholder uses `descriptionForeground`

**Infrastructure cleanup:**
- Removed `decidedChatEditingResourceContextKey` and `hasUndecidedChatEditingResourceContextKey` (no longer meaningful without Keep/Undo)
- Re-gated remaining toolbar actions (Explain, View Changes) on `hasAppliedChatEditsContextKey`
- Updated accessibility help text to remove Keep/Undo references
- Updated standalone strings for editor accessibility

## What's preserved

- The internal `accept()`/`reject()` methods and `ModifiedFileEntryState` enum — still used by checkpoint/restore, session disposal, and snapshot infrastructure
- Undo/Redo chat interaction actions (undo an entire chat turn)
- Checkpoint and restore actions
- Review mode and change navigation (prev/next)
- Inline chat's own Keep/Undo (separate feature, separate actions)
- All internal AI vs user edit tracking

### DIFF
--- a/src/vs/editor/browser/widget/multiDiffEditor/style.css
+++ b/src/vs/editor/browser/widget/multiDiffEditor/style.css
@@ -31,6 +31,7 @@
 			display: grid;
 			place-items: center;
 			place-content: center;
+			color: var(--vscode-descriptionForeground);
 		}
 	}
 
--- a/src/vs/editor/common/standaloneStrings.ts
+++ b/src/vs/editor/common/standaloneStrings.ts
@@ -43,7 +43,7 @@ export namespace AccessibilityHelpNLS {
 	export const debugExecuteSelection = nls.localize('debugConsole.executeSelection', "The Debug: Execute Selection command{0} will execute the selected text in the debug console.", '<keybinding:editor.debug.action.selectionToRepl>');
 	export const chatEditorModification = nls.localize('chatEditorModification', "The editor contains pending modifications that have been made by chat.");
 	export const chatEditorRequestInProgress = nls.localize('chatEditorRequestInProgress', "The editor is currently waiting for modifications to be made by chat.");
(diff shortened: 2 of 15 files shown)


sandy081 (approved):

connor4312 (changes requested):
- We should have this on a setting (at least initially) so people who use keep/undo in their workflow (if any) aren't disrupted terribly
- It looks like we still have keep/undo state but just never give a way to accept or reject the changes? This doesn't provide an opportunity for save participants to run and will also cause all sessions to retain changes indefinitely (unless otherwise influenced) which will leak resources

mjbvz:

PR #296404, opened 2026-02-19
terminal: support tab title templates for extension terminals

rebornix:
Instead of asking users to customize terminal title globally, this PR allows extensions to customize the title for the terminals they create.


anthonykim1:
This looks lot less aggressive than https://github.com/microsoft/vscode/commit/9b64d3b8aeba845e38595076949b7515fc48f880 which got reverted here --> https://github.com/microsoft/vscode/pull/295864 

So with this PR, we'd use the ${sequence} as tab title if contributed from extension (when using `CreateTerminal` API, otherwise if only `name` and not `title` is contributed, sequences are ignored, and we're locked into `Copilot Chat CLI` for example..

If both are provided, then we'd show `title` (title as in when extension contributes a profile) to the terminal profile quickpick, and then the `tabTitle` (sequence) in the terminal tab after creation (dynamic). 

I think the name `title` might be getting overloaded though.

Can we rename:
 * API (TerminalOption): `title` --> `titleTemplate`
 * Contributed profile (package.json): `tabTitle` --> `titleTemplate` 
 * IShellLaunchConfig: `title` --> `titleTemplate` 

or some other variant to signal more dynamic aspect + stay consistent?

rebornix:
@anthonykim1 thanks, it makes perfect sense to me to align the naming. I pushed an update to align on `titleTemplate`.

anthonykim1 (approved):

PR #296415, opened 2026-02-20
Fix diff editor gutter selection for pure deletions

### LINKED ISSUE #9475: Stage selected lines stages only big chunks for line deletions
From @kachkaev
 I can't find out how to select just two deleted lines out of 5 going sequentially and stage only the ones I want. Any time I select something and click that menu item, the whole block gets staged. The same happens when unstaging.

pdrgds:
Fixes #9475

## Summary
- Pure deletions (lines removed with nothing added) have an empty modified line range in the diff editor. Selecting text in the modified (right) editor could never match them, making it impossible to partially stage deletions via the gutter.
- Adds `originalSelections` observable to `DiffEditorEditors` and updates `_selectedDiffs` in `gutterFeature.ts` to read selections from both editors.
- Pure deletions are now matched via the original (left) editor's line ranges, and inner changes are clipped to the selection so only selected deleted lines are staged.

## Test plan
- [ ] Open a side-by-side diff with pure deletions (lines only in the left editor)
- [ ] Select a subset of the deleted lines on the left side
- [ ] Verify the "Stage Selection" button appears in the gutter
- [ ] Click it and verify only the selected lines are staged, not the entire deletion block
- [ ] Verify existing behavior: selecting modified lines on the right side still works as before

### DIFF
--- a/src/vs/editor/browser/widget/diffEditor/components/diffEditorEditors.ts
+++ b/src/vs/editor/browser/widget/diffEditor/components/diffEditorEditors.ts
@@ -39,6 +39,7 @@ export class DiffEditorEditors extends Disposable {
 	public readonly modifiedSelections;
 	public readonly modifiedCursor;
 
+	public readonly originalSelections;
 	public readonly originalCursor;
 
 	public readonly isOriginalFocused;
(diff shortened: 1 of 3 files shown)


pdrgds:

PR #296417, opened 2026-02-20
Feature: Chat Debug Panel

pwang347:
See https://github.com/microsoft/vscode-copilot-chat/pull/3879


src/vs/workbench/api/common/extHostChatDebug.ts
```diff
@@ -0,0 +1,266 @@
...
+				return {
+					...base,
+					kind: 'generic',
+					name: generic.name ?? '',
+					details: generic.details,
+					level: generic.level ?? 1,
+					category: generic.category,
+				};
+			}
+		}
```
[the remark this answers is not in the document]
pwang347: I don't think that's relevant, since we handle this internally already.

vijayupadya (dismissed):

src/vscode-dts/vscode.proposed.chatDebug.d.ts
```diff
@@ -0,0 +1,514 @@
...
+		 * The provider should return initial events and can use
+		 * the progress callback to stream additional events over time.
+		 *
+		 * @param sessionId The ID of the chat session being debugged.
+		 * @param progress A progress callback to stream events.
+		 * @param token A cancellation token.
+		 * @returns Initial events, if any.
+		 */
+		provideChatDebugLog(
+			sessionId: string,
```
roblourens: I see this using string IDs all the way through- we should use the URI sessionResource

src/vs/workbench/contrib/chat/browser/chatDebug/chatDebugOverviewView.ts
```diff
@@ -0,0 +1,264 @@
...
+			}
+		}
+
+		// Explore actions
+		const actionsSection = DOM.append(this.content, $('.chat-debug-overview-section'));
+		DOM.append(actionsSection, $('h3.chat-debug-overview-section-label', undefined, localize('chatDebug.exploreTraceData', "Explore Trace Data")));
+
+		const row = DOM.append(actionsSection, $('.chat-debug-overview-actions'));
+
+		const viewLogsBtn = DOM.append(row, $('button.chat-debug-overview-action-button'));
```
roblourens: You might want to use our Button class or toolbars/menu/actions to ensure that these are set up right for accessibility, theming, telemetry, keybindings if needed, etc

src/vscode-dts/vscode.proposed.chatDebug.d.ts
```diff
@@ -0,0 +1,514 @@
...
+		 * Called when the debug view is opened for a chat session.
+		 * The provider should return initial events and can use
+		 * the progress callback to stream additional events over time.
+		 *
+		 * @param sessionId The ID of the chat session being debugged.
+		 * @param progress A progress callback to stream events.
+		 * @param token A cancellation token.
+		 * @returns Initial events, if any.
+		 */
+		provideChatDebugLog(
```
roblourens, line 478: What is missing from the ChatModel or the `ChatSession` from the session content provider to use as the model for this feature? Seems like there is a substantial amount of overlap and we need to push towards having one single source of truth data model

pwang347: I think it's a very fair question. I would say that the biggest reason why I don't want to unify this right now (aside from the fact that this stuff is experimental still so isolating it seems safer) is that the main purpose here is to capture data in a log format. I'd like to keep it not fully coupled with the chat model/session at the moment so that we can evolve this to log other things in the future as needed (e.g., auth state, deeper logging for org customizations etc.) and generally it means that we can experiment more with the format to add things like event associations which we might not want to do by default in the chat model / session.

Let me know what you think. Thanks!

pwang347: Discussed offline: conclusion was that we should investigate this in the future (not a blocker for now).

karthiknadig (dismissed):

dmitrivMS (dismissed):

dmitrivMS (approved):

PR #296418, opened 2026-02-20
Fix issue #244138: Disable button shows dropdown

realMelTuc:
When an extension is disabled globally and then enabled for a workspace, the Disable button shows a dropdown with both Disable and Disable Workspace options. This fix ensures only Disable Workspace is shown as a button.


hamedhub (commented):
> https://github.com/microsoft/vscode/pull/296418/files#diff-4a5d50360f90c6ae6aefc53feeb1824c570429a00bb07de93e77c97117fa8828

PR #296657, opened 2026-02-20
Browser: preparation for tools

kycutler:


src/vs/workbench/contrib/browserView/common/browserView.ts
```diff
@@ -277,6 +288,10 @@ export class BrowserViewModel extends Disposable implements IBrowserViewModel {
 		this._register(this.onDidChangeVisibility(({ visible }) => {
 			this._visible = visible;
 		}));
+
+		this._register(this.playwrightService.onDidChangeTrackedPages(ids => {
+			this._setSharedWithAgent(ids.includes(this.id));
```
jruales, line 293: Doesn't this cause an infinite loop that we call `this.playwrightService.startTrackingPage(this.id)` when playwrightService changes tracked pages? doesn't one trigger the other and vice versa?

jruales: Ah, never mind, I just saw that this is calling `_setSharedWithAgent` (with underscore) and not `setSharedWithAgent`, and the former does have a check to prevent such a loop

src/vs/platform/browserView/node/playwrightTab.ts
```diff
@@ -0,0 +1,192 @@
...
+			this._logs.push({ type: 'console', time: message.timestamp(), description: `[${message.type()}] ${message.text()}` });
+		}
+	}
+
+	private _handlePageError(error: Error) {
+		this._logs.push({ type: 'pageError', time: Date.now(), description: error.stack ?? error.message });
+	}
+
+	async runAndWaitForCompletion<T>(callback: (token: CancellationToken) => Promise<T>): Promise<T> {
+		// Dialogs pause the page. So we run the callback and wait for either its completion or a dialog to open
```
jruales: This comment describes the whole function and not just the next line's if statement, so please either add an empty newline after it, or turn it into a doc comment. As is, it seems like it just describes the if statement immediately after it

src/vs/platform/browserView/node/playwrightTab.ts
```diff
@@ -0,0 +1,192 @@
...
+		const dialog = Event.toPromise(this._onDialogStateChanged.event);
+		const action = createCancelablePromise(async (token) => {
+			result = await runAndWaitForCompletion(this, callback, token);
+			completed = true;
+		});
+
+		return raceCancellablePromises([dialog, action]).then(() => {
+			if (!completed) {
+				// A dialog was opened before the action completed. Note we don't cancel the action, just ignore its result.
+				throw new Error('Action was interrupted by a dialog');
```
jruales: I was a bit confused on first read about the naming. I would suggest renaming `dialog` -> `dialogOpened`, `action` -> `callbackCompleted`, and `completed` -> `callbackDidComplete`, as suggested below. That also aligns with the naming of "callback" and not "action" used in the function parameter signature
```suggestion
		let callbackDidComplete = false;
		let result: T | void;
		const dialogOpened = Event.toPromise(this._onDialogStateChanged.event);
		const callbackCompleted = createCancelablePromise(async (token) => {
			result = await runAndWaitForCompletion(this, callback, token);
			completed = true;
		});

		return raceCancellablePromises([dialogOpened, callbackCompleted]).then(() => {
			if (!callbackDidComplete) {
				// A dialog was opened before the callback completed. Note we don't cancel the callback, just ignore its result.
				throw new Error('Callback was interrupted by a dialog');
```

src/vs/platform/browserView/node/playwrightTab.ts
```diff
@@ -0,0 +1,192 @@
...
+			}
+			return result!;
+		});
+	}
+
+	async getSummary(full = false): Promise<string> {
+		await this._initialized;
+
+		// Important: anything run against the page must be wrapped with runAndWaitForCompletion in case dialogs are opened.
```
jruales: Can we somehow enforce this so that it's safe by design?

src/vs/platform/browserView/node/playwrightTab.ts
```diff
@@ -0,0 +1,192 @@
...
+			...(logs.length > 0 ? [
+				`Recent events:`,
+				...logs.map(log => `- [${new Date(log.time).toISOString()}] (${log.type}) ${log.description}`)
+			] : []),
+			...(snapshot ? ['Snapshot:', snapshot] : [])
+		].join('\n');
+	}
+}
+
+async function runAndWaitForCompletion<T>(tab: PlaywrightTab, callback: (token: CancellationToken) => Promise<T>, token = CancellationToken.None): Promise<T> {
```
jruales: wouldn't it be cleaner/clearer to turn this into a private method inside of the PlaywrightTab class instead of having it as a module-level private variable?

src/vs/platform/browserView/common/playwrightService.ts
```diff
@@ -3,16 +3,98 @@
...
+import { VSBuffer } from '../../../base/common/buffer.js';
 import { createDecorator } from '../../instantiation/common/instantiation.js';
 
 export const IPlaywrightService = createDecorator<IPlaywrightService>('playwrightService');
 
 /**
  * A service for using Playwright to connect to and automate the integrated browser.
+ *
+ * Pages must be explicitly tracked via {@link startTrackingPage} (or implicitly via
+ * {@link openPage}) before they can be interacted with.
```
jruales, line 16: Question: If Playwright opens a page, and then clicks on a link which opens in a new browser tab, will that new browser tab also be tracked? Per copilot, that's not the case right now, but I think it should

src/vs/platform/browserView/common/playwrightService.ts
```diff
@@ -3,16 +3,98 @@
...
+	 * @param args Additional arguments to pass to the function after the `page` object.
+	 * @returns The result of the function execution, including a page summary.
+	 */
+	invokeFunction(pageId: string, fnDef: string, ...args: unknown[]): Promise<{ result: unknown; summary: string }>;
+
+	/**
+	 * Takes a screenshot of the given page and returns it as a VSBuffer.
+	 * @param pageId The browser view ID identifying the page to capture.
+	 * @param selector Optional Playwright selector to capture a specific element instead of the full page.
+	 * @param fullPage Whether to capture the full scrollable page instead of just the viewport.
```
jruales, line 79: nit: maybe rename to something a bit clearer, since if I didn't read the description, I would have thought this refers to the the full viewport as opposed to the scrollable content too.

src/vs/platform/browserView/node/playwrightService.ts
```diff
@@ -57,25 +89,21 @@ export class PlaywrightService extends Disposable implements IPlaywrightService
 				// This can happen if the service was disposed while we were waiting for the connection. In that case, clean up immediately.
 				if (this._initPromise === undefined) {
 					browser.close().catch(() => { /* ignore */ });
+					group.dispose();
 					throw new Error('PlaywrightService was disposed during initialization');
 				}
 
-				const pageManager = this._register(new PlaywrightPageManager(group, browser, this.logService));
-
 				browser.on('disconnected', () => {
```
jruales, line 96: Just for my own knowledge: have you been able to find a scenario where disconnected event triggers?

src/vs/platform/browserView/node/playwrightTab.ts
```diff
@@ -0,0 +1,192 @@
...
+
+// eslint-disable-next-line local/code-import-patterns
+import type * as playwright from 'playwright-core';
+import { Emitter, Event } from '../../../base/common/event.js';
+import { CancellationToken } from '../../../base/common/cancellation.js';
+import { createCancelablePromise, raceCancellablePromises } from '../../../base/common/async.js';
+
+declare module 'playwright-core' {
+	interface Page {
+		// A hidden Playwright method that returns an AI-friendly snapshot of the page.
```
jruales, line 14: When we say "snapshot" do we mean the HTML of the page? the inner text of the page? A summary? A delta? HTML element statistics?

src/vs/platform/browserView/node/playwrightService.ts
```diff
@@ -85,6 +113,72 @@ export class PlaywrightService extends Disposable implements IPlaywrightService
...
+
+			let result;
+			try {
+				result = await this._pages.runAndWaitForCompletion(page, () => fn(page, args));
+			} catch (err: unknown) {
+				result = err instanceof Error ? err.message : String(err);
+			}
+
+			const summary = await this._pages.getSummary(page);
+			return { result, summary };
```
jruales, line 145: for the purpose of saving up on token usage, do you think there would be merit in not returning a summary after every operation but instead making it optional?

src/vs/platform/browserView/node/playwrightService.ts
```diff
@@ -85,6 +113,72 @@ export class PlaywrightService extends Disposable implements IPlaywrightService
...
+
+	async invokeFunction(pageId: string, fnDef: string, ...args: unknown[]): Promise<{ result: unknown; summary: string }> {
+		this.logService.info(`[PlaywrightService] Invoking function on view ${pageId}`);
+
+		try {
+			await this.initialize();
+			const page = await this._pages.getPage(pageId);
+
+			const vm = await import('vm');
+			const fn = vm.compileFunction(`return (${fnDef})(page, ...args)`, ['page', 'args'], { parsingContext: vm.createContext() });
```
jruales, line 135: I'm thinking of ways that this could possibly break. If `fnDef` does a `const page = <blah>` or `const args = <blah>` then it would break because the variables `page` and `args` are already defined, right? Maybe we should use some obscure names here to avoid that collision? Or otherwise maybe pass in the function body instead of the whole body?

src/vs/platform/browserView/node/playwrightService.ts
kycutler: By default the summary is only the deltas -- we only return the full summary on the first read or when using `getSummary()`

src/vs/platform/browserView/node/playwrightService.ts
kycutler: `page` and `args` aren't actually defined in the function scope (they are defined in the wrapping scope), so the function can redeclare them if it doesn't also use `page` or `args` for its own parameters. For the "run code" tool (in the other PR), if the agent tries to do `const page = ...` it will get `SyntaxError: Identifier 'page' has already been declared` which I think is perfectly reasonable

src/vs/platform/browserView/node/playwrightService.ts
kycutler: Not in practice no -- but it could happen if for example the CDP server closed

src/vs/platform/browserView/node/playwrightTab.ts
kycutler: It's based on the similar helper in Playwright, so I'm inclined to leave it as-is to make it clearer in that regard.

src/vs/platform/browserView/node/playwrightTab.ts
kycutler: It's basically the accessibility tree, but I don't think it's very relevant (and it could change in theory). It returns both the `full` and `incremental` (delta) versions of it.

src/vs/platform/browserView/node/playwrightTab.ts
kycutler: Tweaked it a bit, but updated!

src/vs/platform/browserView/node/playwrightTab.ts
kycutler: Updated a few things:
- Renamed `runAndWaitForCompletion` to `safeRunAgainstPage`
- Made `page` private on `PlaywrightTab`
- Added `@deprecated` to the internal `page` to hint that it should only be used directly with caution

Also refactored the access in `playwrightService` to enforce this as well

Thanks!

src/vs/platform/browserView/common/playwrightService.ts
kycutler: Updated the doc comment to make it a bit clearer. Since this is the default behavior of most screenshotting functions I'm inclined to leave the name as-is

src/vs/platform/browserView/common/playwrightService.ts
kycutler: It's a good callout. I think we'll need to refactor the CDP proxy / groups a bit to make this possible. I'll leave it for another PR

dmitrivMS (approved):

PR #296668, opened 2026-02-21
Add Sanity Tests stage to product build pipeline

dmitrivMS:
**Changes**
Added Sanity Tests stage to official build pipeline.

**Notes**
Sanity tests will run for appropriate targets when `Publish` is enabled and `Skip Tests` is disabled.

**Validation**
Ran official build pipeline on all quality levels, with/without Publish and Skip Tests and on subsets of targets.


build/azure-pipelines/product-build.yml
```diff
@@ -492,6 +492,297 @@ extends:
                 VSCODE_QUALITY: ${{ variables.VSCODE_QUALITY }}
                 VSCODE_SCHEDULEDBUILD: ${{ variables.VSCODE_SCHEDULEDBUILD }}
 
+      - ${{ if and(eq(variables['VSCODE_PUBLISH'], 'true'), eq(variables['VSCODE_STEP_ON_IT'], false)) }}:
+        - stage: SanityTests
+          displayName: Sanity Tests
+          dependsOn:
+            - Publish
```
joaomoreno, line 499: Today we have:
- Platform independent jobs creating artifacts
- One Publish stage which polls on artifacts and publishes them to the CDN. This stage also fails if any of the platform specific stages fail. This stage will also contain a "release build" step for the nightly builds, no approval.
- Optionally, a Release stage which won't run if Publish doesn't run (when people check "Release build"). This stage requires approval.

This dependency will make it that sanity tests will run right after publish. That means:
- For nightly builds, they will still be released even if sanity tests fail.
- For manual "Release build" builds, they will also still be released if sanity tests fail.
- The first sanity tests will start running only after the last artifact is published to the CDN. IMO this is OK to have for now, given the complexity of implementing a polling job across 20+ build agents; we don't want to keep that many agents busy doing nothing.

Should we prevent releasing a build, if sanity tests fail?

dmitrivMS: I considered these points:
- I'd like to run without Release stage dependency for now to ensure things are stable, but eventually yes - I think it makes sense to block Release if sanity tests fail.
- Running target sanity tests ASAP will only win us so much time since last target will run at the end and all tests run in parallel in under 10 mins, so the delta will be minimal.

joaomoreno (approved):

PR #296711, opened 2026-02-21
Use span instead of Electron.relaunch to avoid restrictions in the new process

### LINKED ISSUE #253204: VSCode instances started through "Relaunch VS Code" button are started with PR_SET_NO_NEW_PRIVS

- VS Code Version: 1.101.2
- OS Version: Ubuntu 22.04

Steps to Reproduce:

1. Install a version of VS Code before the current one, e.g. `apt install code=1.101.1-1750254731`
2. Open VS Code
3. Upgrade VS Code to the current version, e.g. `apt install code=1.101.2-1750797935`
4. Open a new window in VS code, this will trigger a notification in VS Code "Extension host cannot start: version mismatch"
5. Click the "Relaunch VS Code" button in that notification.
6. The newly started VS Code will have `PR_SET_NO_NEW_PRIVS` set. Running sudo in a VS Code Terminal will return `The "no new privileges" flag is set, which prevents sudo from running as root.` and the flag can be seen as `no_new_privs` in the output of `setpriv -d`

This simulates the package being upgraded by unattended-upgrades which is how we hit the issue.

There has been a number of issues reported for this, e.g. https://github.com/microsoft/vscode/issues/242781 https://github.com/microsoft/vscode/issues/221718 https://github.com/microsoft/vscode/issues/90725

They get closed as not reproducable though because a restart "fixes" it.

dmitrivMS:
Fixes #253204

### DIFF
--- a/src/vs/platform/lifecycle/electron-main/lifecycleMainService.ts
+++ b/src/vs/platform/lifecycle/electron-main/lifecycleMainService.ts
@@ -3,12 +3,13 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
+import { spawn } from 'child_process';
 import electron from 'electron';
 import { validatedIpcMain } from '../../../base/parts/ipc/electron-main/ipcMain.js';
 import { Barrier, Promises, timeout } from '../../../base/common/async.js';
 import { Emitter, Event } from '../../../base/common/event.js';
 import { Disposable, DisposableStore } from '../../../base/common/lifecycle.js';
-import { isMacintosh, isWindows } from '../../../base/common/platform.js';
+import { isLinux, isMacintosh, isWindows } from '../../../base/common/platform.js';
 import { cwd } from '../../../base/common/process.js';
 import { assertReturnsDefined } from '../../../base/common/types.js';
 import { NativeParsedArgs } from '../../environment/common/argv.js';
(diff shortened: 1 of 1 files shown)


src/vs/platform/lifecycle/electron-main/lifecycleMainService.ts
```diff
@@ -688,8 +689,18 @@ export class LifecycleMainService extends Disposable implements ILifecycleMainSe
 
 		const quitListener = () => {
 			if (!this.relaunchHandler?.handleRelaunch(options)) {
-				this.trace('Lifecycle#relaunch() - calling app.relaunch()');
-				electron.app.relaunch({ args });
+				if (isLinux) {
+					// On Linux, Electron's app.relaunch() uses Chromium's base::LaunchProcess()
+					// which sets PR_SET_NO_NEW_PRIVS on the new process by default, preventing
```
deepak1556, line 694: Not true, the relaunched program is allowed privileges https://github.com/electron/electron/blob/aca83afeef7cdebce10c6fa9835baed26c806b9f/shell/browser/relauncher_linux.cc#L63

So what needs checking is,

1) The relaunch path will shutdown the current app gracefully and relaunch self executable with a special commandline that will cause a different main function to be invoked https://github.com/electron/electron/blob/aca83afeef7cdebce10c6fa9835baed26c806b9f/shell/browser/relauncher.cc#L117 `options.allow_new_privs = false` in here
2) This new main function https://github.com/electron/electron/blob/aca83afeef7cdebce10c6fa9835baed26c806b9f/shell/browser/relauncher.cc#L160 will then proceed to launch the executable with provided arguments `options.allow_new_privs = true` here

If the final application process is still getting `PR_SET_NO_NEW_PRIVS` then it is very likely inherited from the relauncher process, lets confirm this and get a proper fix into the relaunch code of Electron. Relying on node.js process launch will complicate shutdown sequence which we get for free from the relaunch api.

dmitrivMS: Yep,' it's in draft as I haven't been able to test any of it yet.

PR #297152, opened 2026-02-24
Don't show `Used n references` when opening an old session

daviddossett:
Fixes "Used n references" buttons appearing on restored agent-mode chats.

## Problem

`modeInfo` was never serialized when saving chat sessions. When restoring old chats, `modeInfo` was always `undefined`, so the check `modeInfo?.modeId === ChatModeKind.Agent` in `chatListRenderer.ts` always evaluated to `false`, causing the "Used n references" button to show even for agent-mode sessions.

## Fix

1. **Persist `modeInfo`** — Added `modeInfo` to `ISerializableChatRequestData`, `toExport()`, `_deserializeRequest()`, and the operation log schema so it survives across save/restore cycles.

2. **Fallback for old sessions** — For sessions saved before this change (no `modeInfo` on the request), fall back to the session-level `delegate.currentChatMode()` which reflects the persisted `inputModel.mode`.

### DIFF
--- a/src/vs/workbench/contrib/chat/browser/widget/chatListRenderer.ts
+++ b/src/vs/workbench/contrib/chat/browser/widget/chatListRenderer.ts
@@ -1777,8 +1777,13 @@ export class ChatListItemRenderer extends Disposable implements ITreeRenderer<Ch
 			} else if (content.kind === 'markdownContent') {
 				return this.renderMarkdown(content, templateData, context);
 			} else if (content.kind === 'references') {
-				if (isResponseVM(context.element) && context.element.model.request?.modeInfo?.modeId === ChatModeKind.Agent) {
-					return this.renderNoContent(other => other.kind === content.kind);
+				if (isResponseVM(context.element)) {
+					const isAgent = context.element.model.request?.modeInfo
+						? context.element.model.request.modeInfo.modeId === ChatModeKind.Agent
+						: this.delegate.currentChatMode() === ChatModeKind.Agent;
+					if (isAgent) {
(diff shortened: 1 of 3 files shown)


src/vs/workbench/contrib/chat/browser/widget/chatListRenderer.ts
```diff
@@ -1777,8 +1777,19 @@ export class ChatListItemRenderer extends Disposable implements ITreeRenderer<Ch
...
 				return this.renderMarkdown(content, templateData, context);
 			} else if (content.kind === 'references') {
-				if (isResponseVM(context.element) && context.element.model.request?.modeInfo?.modeId === ChatModeKind.Agent) {
-					return this.renderNoContent(other => other.kind === content.kind);
+				if (isResponseVM(context.element)) {
+					const request = context.element.model.request;
+					let isAgent = false;
+					if (request) {
+						if (request.modeInfo) {
+							isAgent = request.modeInfo.modeId === ChatModeKind.Agent;
```
roblourens: I think this should always check the kind, not the modeId, because this only covers the builtin "Agent", not custom agents

roblourens, line 1785: This is not quite right because you can send a request to a participant while being "in agent mode". I'm going to do a slightly different fix, it's my fault this is messy

daviddossett: Thanks, knew this didn't feel good

PR #297153, opened 2026-02-24
Add newSessionOptions so providers can set the initial options for a session

mjbvz:
cc @TylerLeonhardt I think we should hold off on merging this to avoid blocking testing tomorrow. I'll keep polishing this too so it's in better shape


TylerLeonhardt:
FYI @DonJayamanne

TylerLeonhardt (approved):

PR #297177, opened 2026-02-24
Finalize env.isAppPortable API proposal

### LINKED ISSUE #288950: Proposed API for detecting portable installation of VS Code

dmitrivMS:
Fixes #288950


gjsjohnmurray:
Please fix typo in the PR title.

mjbvz (approved):

PR #297590, opened 2026-02-25
Fix copy button not working during chat code block streaming

### LINKED ISSUE #255290: Copy button for code snippets only works after AI finishes streaming response
**Description:**
When GitHub Copilot Chat streams code snippets as part of its response, the “Copy” button next to each snippet is disabled or unresponsive until the AI has fully completed its answer. If you try to click “Copy” while the snippet is still rendering, nothing happens. You must wait for the response to finish before copying.

---

**Steps to Reproduce:**

1. Open VS Code with the GitHub Copilot Chat extension enabled.
2. Ask a question that generates multi‑line code (e.g., “Show me a Node.js HTTP server example.”).
3. As the AI streams its answer, hover over the partially rendered code block.
4. Click the “Copy” button beside the snippet **before** the AI finishes streaming.
5. Observe that nothing is copied to the clipboard.
6. Wait for the AI to finish its response, then click “Copy” again—this time it works.

---

**Expected Behavior:**
The “Copy” button should work at any point once the code block has begun rendering. Users shouldn’t have to wait for the entire answer to complete before copying.

---

**Actual Behavior:**
The “Copy” button only becomes active after the AI completes the full response and finalizes the code block.

---

**Reproducibility:**
Consistent – the copy button is unresponsive until the end of every streamed code response.

---

**Additional Context:**
This delay interrupts developer workflow when reviewing or reusing code as it is generated.

---

**Possible Areas to Investigate:**

* Enable copy functionality on partially rendered code blocks
* Decouple copy action from stream completion event
* Update UI state as soon as any snippet content is available

pierceboggan:
When users click the Copy button on a code block while Copilot Chat is still streaming its response, nothing gets copied. They have to wait for the entire response to finish before copying works.

## Root Cause

The `CodeBlockPart.render()` method sets the toolbar context (including the `code` property used by the Copy action) only **after** awaiting the text model promise. During streaming, this promise takes time to resolve while the model is being created and populated. If a user clicks Copy before the promise resolves, `context.code` is undefined and the copy silently fails.

## Solution

Pass the raw text (which is already available from markdown parsing) to `render()` and set an initial toolbar context immediately, before awaiting the model. This ensures the Copy action has access to the code text from the moment the code block appears. The context is later refined with the fully resolved model data once available.

The change is backwards compatible — the `text` parameter is optional, so existing callers that don't pass it continue to work.

Fixes #255290

---

Created with [Copilot Desktop](https://github.com/githubnext/copilot-tauri-app)

### DIFF
--- a/src/vs/workbench/contrib/chat/browser/widget/chatContentParts/chatMarkdownContentPart.ts
+++ b/src/vs/workbench/contrib/chat/browser/widget/chatContentParts/chatMarkdownContentPart.ts
@@ -407,7 +407,7 @@ export class ChatMarkdownContentPart extends Disposable implements IChatContentP
 			this._codeblocks[data.codeBlockPartIndex].codemapperUri = e.codemapperUri;
 		});
 
-		editorInfo.render(data, currentWidth);
+		editorInfo.render(data, currentWidth, text);
 
 		return ref;
 	}
--- a/src/vs/workbench/contrib/chat/browser/widget/chatContentParts/codeBlockPart.ts
+++ b/src/vs/workbench/contrib/chat/browser/widget/chatContentParts/codeBlockPart.ts
@@ -411,10 +411,22 @@ export class CodeBlockPart extends Disposable {
 		return this.editor.getContentHeight();
 	}
 
-	async render(data: ICodeBlockData, width: number) {
+	async render(data: ICodeBlockData, width: number, text?: string) {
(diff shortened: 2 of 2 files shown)


DevFlex-AI (commented):
honeslty i think this is a goo dfix thta should be merged

dmitrivMS (approved):

PR #297609, opened 2026-02-25
Allow singleton editors to split into a new editor

rebornix:
Singleton editors currently cannot participate in split operations, the split command silently does nothing for them. This PR introduces an opt-in mechanism that lets singleton editors provide a new editor input when a split is requested.

This is part of our growth efforts for agent sessions cc @lostintangent 

https://github.com/user-attachments/assets/623c6d64-d467-42f0-884a-75582737196a

@bpasero @lramos15 let me know if there is a better way to do it, thanks!


bpasero (changes requested):
If you follow along where `EditorInputCapabilities.Singleton` is used you will notice that there is many places where we deal with singleton editors in a specific way, not just in the 1 location you do in `editorCommands`, so I think your change is not respecting all these locations properly?

In other words: there is not just 1 command to split an editor, there can be many more gestures, including drag and drop or programmatic API usages ("open to side").

rebornix:
I was intentionally only doing this for Split button since I want to minimize the scope. But I updated the changes to make  sure that it covers other cases (as shown below)


https://github.com/user-attachments/assets/050d8e02-295e-4748-b1f3-e861023a80f9

bpasero (dismissed):

bpasero:
@rebornix ended up with a different approach: we simply no longer mark chat, terminal and browser editors as `Singleton` given they implement the `copy` method to make duplications of the editor. 

This means you can now split a terminal or browser too for example to get a new empty terminal which I think is really nice (cc @Tyriar @kycutler).

Thanks for uncovering this flaw!

bpasero (approved):

jruales (approved):

Issue #201294, opened 2023-12-20
Documents opened using `window.showTextDocument` send a delayed `didClose` event to LSP


EhabY:
Version/OS information:
```
Version: 1.85.1 (system setup)
Commit: 0ee08df0cf4527e40edc9aa28f4b5bd38bbff2b2
Date: 2023-12-13T09:49:37.021Z
Electron: 25.9.7
ElectronBuildId: 25551756
Chromium: 114.0.5735.289
Node.js: 18.15.0
V8: 11.4.183.29-electron.0
OS: Windows_NT x64 10.0.22621
```

Steps to Reproduce:

1. Setup an LSP for some language
2. Create a file of said language
3. Open the file using the `window.showTextDocument` -> This will (correctly) send a `didOpen` request
4. Close the file -> Does not `didClose` for a while

Using the `commands` API (`commands.executeCommand("vscode.open", resourceUri, options)`) seems to work as expected (more interactive)

Related but the documentation for "vscode.open" seems to have been deleted: https://code.visualstudio.com/api/references/commands

Reopening the file does not resend a `didOpen` request and closing the tab does not send a `didClose` - so it must be that (for some reason) closing the file is delayed significantly. After waiting for 5 minutes, a `didClose` is actually emitted but it is very strange that using the commands API it is more predictable and interactive than the `window` API

VSCodeTriageBot:
We closed this issue because it is a question about using VS Code rather than an issue or feature request. Please search for help on [StackOverflow](https://aka.ms/vscodestackoverflow), where the community has already answered thousands of similar questions. You may find their [guide on asking a new question](https://aka.ms/vscodestackoverflowquestion) helpful if your question has not already been asked. See also our [issue reporting guidelines](https://aka.ms/vscodeissuereporting).

Happy Coding!

EhabY:
> The editor decides when to close a document by its own rules.

I understand that but it seems there is inconsistency between using `window.showTextDocument` and `commands.executeCommand("vscode.open")`. For example, if you open a file from the explorer and close it, an event is emitted immediately which is the desired behavior and matches the documents opened using the commands API but not the `window` API. The docs do not cover such cases (https://code.visualstudio.com/api/references/commands)

szarnekow:
Let me describe the user case:
[...]

Issue #47140, opened 2018-04-03
workspace.textDocuments includes closed documents


DanTup:
In #10304 I was advised that `workspace.textDocuments` was the correct way to get open files; however I'm trying to write tests that rely on documents being closed and finding that they still exist in there despite being closed (I can see in the extension host there there are no open editors).

Here's a sample test:

```ts
it.only("test", async () => {
	// Open two files
	await vs.workspace.openTextDocument(emptyFile);
	await vs.workspace.openTextDocument(everythingFile);
	console.log(`There are ${vs.workspace.textDocuments.length} documents`);

	// Close all files
	while (vs.window.activeTextEditor) {
		await vs.commands.executeCommand("workbench.action.closeActiveEditor");
	}
	console.log(`There are ${vs.workspace.textDocuments.length} documents`);

	// Wait a short period just in case...
	await new Promise((resolve) => setTimeout(resolve, 1000));
	console.log(`There are ${vs.workspace.textDocuments.length} documents`);
});
```

This outputs:

```text
There are 2 documents
There are 2 documents
There are 2 documents
```

DanTup:
Also weird, the `while` loop seems to only execute once.

jrieken:
> however I'm trying to write tests that rely on documents being closed and finding that they still exist in 

You cannot assume the opposite and closing an editor doesn't mean that we close the document too.

DanTup:
@jrieken Then the answer to my original question is not satisfied :( How do I get a list of the documents that are open? I want to provide my language server with a list of priority documents, which are first the visible text editors, and then any editors that are open but not visible.

PR #297649, opened 2026-02-25
Fix markdown preview not refreshing on external file changes

### LINKED ISSUE #194421: Markdown Preview shows outdated file content

- VS Code Version: 1.82.2
- OS Version: Win11 

Steps to Reproduce:

1.  Open vscode and create a markdown file ("test.md")
2. open file, add some content and open Markdown Preview
3. close VScode Editor but keep Markdown Preview open
4. change markdown file outside of VSCode Editor (e.g. terminal inside VSCode)
5. reopen VSCode Editor, close Markdown Preview and reopen new Markdown Preview

Markdown Preview does not update content and shows content from step 2.

markdown_bug

### LINKED ISSUE #265277: Markdown preview doesn't refresh/reload when the file is overwritten

yuppppppppppqi:
## Summary

- Remove the `textDocuments.some()` guard in the file system watcher callback that prevented preview refresh when VS Code already had the document open
- Pass `forceUpdate=true` to `refresh()` to bypass the internal version check, which may not reflect external changes
- The existing 300ms throttle in `refresh()` prevents performance issues from duplicate events

## Problem

When a markdown file is modified by an external process (AI coding agents like Claude Code / Copilot, scripts, git sync, file copy/overwrite), the preview does not update. Even closing and reopening the preview tab shows stale content.

**Root cause:** The watcher callback checks `!vscode.workspace.textDocuments.some(doc => doc.uri.toString() === uri.toString())` before refreshing. Since the document is almost always already known to VS Code when a preview is open, external file changes are silently ignored. Additionally, `onDidChangeTextDocument` does not fire for external overwrites, creating a gap where no refresh path is triggered.

## Test plan

- [ ] Open a markdown file and show preview side by side
- [ ] From a terminal, modify the file externally (e.g. `echo "# Updated" >> file.md`)
- [ ] Verify the preview updates automatically
- [ ] Verify that normal in-editor editing still updates the preview (no regression)
- [ ] Verify that switching tabs and returning shows current content

Fixes #265277
Fixes #194421

### DIFF
--- a/extensions/markdown-language-features/src/preview/preview.ts
+++ b/extensions/markdown-language-features/src/preview/preview.ts
@@ -114,10 +114,10 @@ class MarkdownPreview extends Disposable implements WebviewResourceProvider {
 			const watcher = this._register(vscode.workspace.createFileSystemWatcher(new vscode.RelativePattern(resource, '*')));
 			this._register(watcher.onDidChange(uri => {
 				if (this.isPreviewOf(uri)) {
-					// Only use the file system event when VS Code does not already know about the file
-					if (!vscode.workspace.textDocuments.some(doc => doc.uri.toString() === uri.toString())) {
-						this.refresh();
-					}
+					// Always refresh on file system changes, including external modifications
+					// (e.g. AI agents, scripts, git sync). The throttle in refresh() prevents
+					// performance issues from duplicate events with onDidChangeTextDocument.
+					this.refresh(true);
 				}
 			}));
 		}
--- a/extensions/markdown-language-features/src/preview/preview.ts
+++ b/extensions/markdown-language-features/src/preview/preview.ts
@@ -114,10 +114,19 @@ class MarkdownPreview extends Disposable implements WebviewResourceProvider {
 			const watcher = this._register(vscode.workspace.createFileSystemWatcher(new vscode.RelativePattern(resource, '*')));
 			this._register(watcher.onDidChange(uri => {
 				if (this.isPreviewOf(uri)) {
-					// Always refresh on file system changes, including external modifications
-					// (e.g. AI agents, scripts, git sync). The throttle in refresh() prevents
-					// performance issues from duplicate events with onDidChangeTextDocument.
-					this.refresh(true);
+					// Only skip the file-system event when the document has unsaved
+					// edits in the editor.  A dirty document means the preview is
+					// already kept up-to-date via onDidChangeTextDocument.
+					// For non-dirty or closed documents we must refresh from disk
+					// because onDidChangeTextDocument may not fire reliably — the
+					// underlying TextFileEditorModel (and its file watcher) is
+					// disposed when the editor tab is closed, even though the
+					// TextDocument can linger in textDocuments for up to 5 minutes.
+					const doc = vscode.workspace.textDocuments.find(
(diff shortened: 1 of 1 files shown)


extensions/markdown-language-features/src/preview/preview.ts
```diff
@@ -114,10 +114,10 @@ class MarkdownPreview extends Disposable implements WebviewResourceProvider {
...
 			this._register(watcher.onDidChange(uri => {
 				if (this.isPreviewOf(uri)) {
-					// Only use the file system event when VS Code does not already know about the file
-					if (!vscode.workspace.textDocuments.some(doc => doc.uri.toString() === uri.toString())) {
-						this.refresh();
-					}
+					// Always refresh on file system changes, including external modifications
+					// (e.g. AI agents, scripts, git sync). The throttle in refresh() prevents
+					// performance issues from duplicate events with onDidChangeTextDocument.
+					this.refresh(true);
```
mjbvz: This feels like a rather brute force workaround instead of fixing the underlying issue. You'll trigger extra refreshes for cases such as a dirty editor tab where the underlying document is changed by the file system (this shouldn't update the preview)

Have you explored why the existing events aren't working? For example, we may need to refresh on `close` events too

yuppppppppppqi: Thanks for the review — you're right that the previous approach was too brute-force. I've pushed a new commit that addresses both concerns.

### Root cause

The real issue is a lifecycle mismatch between VS Code's document layers:

1. When an editor tab is closed, `TextFileEditorModel` is disposed and its file watcher stops.
2. However, the `TextDocument` object lingers in `workspace.textDocuments` for up to ~5 minutes (this is by design — see #47140, #201294).
3. During this window, `onDidChangeTextDocument` never fires for external changes because the underlying model is gone, yet the old `textDocuments.some()` guard still suppresses the file-system event.

So the file-system watcher fires, we find the stale `TextDocument`, skip the refresh, and `onDidChangeTextDocument` never comes to the rescue. The preview goes stale.

### New approach: `isDirty` guard

Instead of removing the guard entirely, the updated code now checks `doc.isDirty`:

```ts
const doc = vscode.workspace.textDocuments.find(
    d => d.uri.toString() === uri.toString());
if (!doc || !doc.isDirty) {
    this.refresh(true);
}
```

- **Dirty document** → skip the FS event. The editor owns the content; `onDidChangeTextDocument` keeps the preview current. This also handles your concern about dirty tabs where the underlying file changes on disk — the preview correctly shows the unsaved editor content, not the disk version.
- **No document / clean document** → refresh from disk. This covers closed tabs (zombie `TextDocument` with `isDirty === false`) and open-but-saved tabs alike.

This mirrors the pattern used by `TextFileEditorModelManager.onDidFilesChange()` internally, which also skips dirty models when processing file-system events.

### Why not `onDidCloseTextDocument`?

I explored adding a refresh on `close` events, but `onDidCloseTextDocument` fires when the *extension host* drops the document (up to 5 min after tab close), not when the user closes the tab. There's no reliable extension API event for "user closed this specific editor tab and the file watcher just stopped." The `isDirty` check sidesteps this gap cleanly.

PR #297741, opened 2026-02-25
fix editor punctuation width

### LINKED ISSUE #242138: Bring good old full width punctuations back

- VS Code Version: 1.97.2.25045
- OS Version: kernel 6.13.3, glibc 2.41

Steps to Reproduce:

1. Type some full width punctuation characters such as ，。；、
2. See the strange behavior as attached in the image

Image

Traditionally, full-width punctuation characters used to be as wide as Unified CJK ideographs. A recent update broke the consistency. Now their widths are not context-independent.

This must be caused by some underlying Blink/Electron CSS typesetting features. Let me know how should I tweak my fontconfig (and possibly other stuff) if vscode itself cannot afford an option to bring good old full width punctuation back.

neruthes:
Close #242138 

This PR adds a simple one-line `text-spacing-trim: space-all;` for `.monaco-editor`.

Investigations showed that the issue was not introduced by an active change in this repository; instead, it might come from an upstream change of default value. So I am not creating a PR for a new setting entry.

### DIFF
--- a/src/vs/editor/browser/widget/codeEditor/editor.css
+++ b/src/vs/editor/browser/widget/codeEditor/editor.css
@@ -21,6 +21,7 @@
 	position: relative;
 	overflow: visible;
 	-webkit-text-size-adjust: 100%;
+	text-spacing-trim: space-all;
 	color: var(--vscode-editor-foreground);
 	background-color: var(--vscode-editor-background);
 	overflow-wrap: initial;


alexdima (approved):
Thank you!

Notes for myself:
image

> **Was `space-all` the default?** No. The initial/default value is **`normal`**, not `space-all`. The formal definition table states: *Initial value: `normal`*. The `normal` value applies kerning/trimming rules to CJK punctuation — collapsing spacing between adjacent punctuation, and adjusting punctuation at line starts/ends. `space-all` is the opt-out value that leaves all CJK full-width punctuation at full width (no trimming).

> You might be thinking of the *behavior before browsers implemented `text-spacing-trim`* — when no browser supported it, the effective behavior was equivalent to `space-all` (no punctuation trimming happened). So in practice, `space-all` matches what users historically saw, but it was never the spec default.

> **Is `space-all` more appropriate for a code editor?** Yes, very likely. In a code editor:

roblourens (approved):

Issue #297387, opened 2026-02-24
MCP can not start in sandbox on my macOS


isidorn:
Testing #297115

1. Open vscode repo
2. Open `mcp.json`
3. Add "sandboxEnabled": true
4. Reload VS Code Insiders
5. Notice error when starting mcp server. Click on logs

We should handle this more gracefully so user is guided to enable sandbox. Or ideally it just works

[sandbox.log](https://github.com/user-attachments/files/25525393/sandbox.log)

dileepyavan:
@isidorn you need to check the logs and add the folders or domains to the sandbox serttings. For example, in your scenario

`[Error: EPERM: operation not permitted, unlink '/Users/isidor/Development/vscode/test/mcp/node_modules/.package-lock.json'] {
2026-02-24 17:33:21.647 [warning] [server stderr] npm error   errno: -1,
2026-02-24 17:33:21.647 [warning] [server stderr] npm error   code: 'EPERM',
2026-02-24 17:33:21.647 [warning] [server stderr] npm error   syscall: 'unlink',
2026-02-24 17:33:21.647 [warning] [server stderr] npm error   path: '/Users/isidor/Development/vscode/test/mcp/node_modules/.package-lock.json'
2026-02-24 17:33:21.647 [warning] [server stderr] npm error }
`
You need to provide access to /Users/isidor/Development/vscode/ using AlloWrite option.

roblourens:
Oh, did you add a dialog for Karthik's error, not Isidor's error? I see the same as Isidor

roblourens:
```
2026-03-06 10:54:05.205 [info] Connection state: Running
2026-03-06 10:54:05.657 [warning] [server stderr] [SandboxDebug] Initializing sandbox...
2026-03-06 10:54:05.687 [warning] [server stderr] [SandboxDebug] HTTP proxy listening on localhost:50488
2026-03-06 10:54:05.688 [warning] [server stderr] [SandboxDebug] SOCKS proxy listening on 127.0.0.1:50489
2026-03-06 10:54:05.688 [warning] [server stderr] [SandboxDebug] Network infrastructure initialized
2026-03-06 10:54:05.688 [warning] [server stderr] [SandboxDebug] Command string mode (-c): ./test/componentFixtures/component-explorer.json
2026-03-06 10:54:05.688 [warning] [server stderr] [SandboxDebug] {
2026-03-06 10:54:05.688 [warning] [server stderr]   "allowedHosts": [
[...]

Issue #297458, opened 2026-02-24
Sandbox configuration moves into mcp.json when it's updated


connor4312:
Testing #297115

1. Have an initial MCP server with a sandbox (like https://github.com/microsoft/vscode/issues/297452#issue-3985470060)
2. Use the "Add Server" button to add a server of your choice
3. 🐛 the sandbox option moves into the server field, resulting in a validation error

PR #297797, opened 2026-02-25
Dialog Notification when MCP server start fails in sandbox mode.

dileepyavan:
fixes:(#297387)

Details:

MCP server in sandboxing mode can fail due to file system access issues or due to domains that are not added to allowlist.

- Shows  a dialog for the user to confirm on the missing permissions/ domains for adding to sandbox config.
- User doesnt need to look into the logs for sandbox related errors on start up.
- If MCP server has cwd property, add the path to defaultAllowWrite.

### DIFF
--- a/src/vs/platform/mcp/common/mcpResourceScannerService.ts
+++ b/src/vs/platform/mcp/common/mcpResourceScannerService.ts
@@ -85,6 +85,12 @@ export class McpResourceScannerService extends Disposable implements IMcpResourc
 
 	async updateSandboxConfig(sandbox: IMcpSandboxConfiguration, mcpResource: URI, target?: McpResourceTarget): Promise<void> {
 		await this.withProfileMcpServers(mcpResource, target, scannedMcpServers => {
+			if (scannedMcpServers.servers) {
+				for (const server of Object.values(scannedMcpServers.servers)) {
+					delete (server as Mutable<IMcpStdioServerConfiguration>).sandbox;
+				}
+			}
+
 			return {
 				...scannedMcpServers,
 				sandbox,
--- a/src/vs/platform/mcp/common/mcpResourceScannerService.ts
+++ b/src/vs/platform/mcp/common/mcpResourceScannerService.ts
@@ -85,12 +85,6 @@ export class McpResourceScannerService extends Disposable implements IMcpResourc
 
 	async updateSandboxConfig(sandbox: IMcpSandboxConfiguration, mcpResource: URI, target?: McpResourceTarget): Promise<void> {
 		await this.withProfileMcpServers(mcpResource, target, scannedMcpServers => {
-			if (scannedMcpServers.servers) {
-				for (const server of Object.values(scannedMcpServers.servers)) {
-					delete (server as Mutable<IMcpStdioServerConfiguration>).sandbox;
-				}
-			}
-
 			return {
 				...scannedMcpServers,
 				sandbox,
@@ -197,7 +191,7 @@ export class McpResourceScannerService extends Disposable implements IMcpResourc
 		if (servers.length > 0) {
 			userMcpServers.servers = {};
 			for (const [serverName, server] of servers) {
-				userMcpServers.servers[serverName] = this.sanitizeServer(server, scannedMcpServers.sandbox);
+				userMcpServers.servers[serverName] = this.sanitizeServer(server);
 			}
 		}
 		return userMcpServers;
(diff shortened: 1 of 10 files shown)


src/vs/platform/mcp/common/mcpResourceScannerService.ts
```diff
@@ -82,6 +83,21 @@ export class McpResourceScannerService extends Disposable implements IMcpResourc
 		});
 	}
 
+	async updateSandboxConfig(sandbox: IMcpSandboxConfiguration, mcpResource: URI, target?: McpResourceTarget): Promise<void> {
+		await this.withProfileMcpServers(mcpResource, target, scannedMcpServers => {
+			if (scannedMcpServers.servers) {
+				for (const server of Object.values(scannedMcpServers.servers)) {
+					delete (server as Mutable<IMcpStdioServerConfiguration>).sandbox;
```
connor4312: THis is a workaround for https://github.com/microsoft/vscode/issues/297458 in this particular case, but it should be generalized. 

Just looking at the code it should be pretty easy to not have `sandbox` on the `IMcpStdioServerConfiguration` at all. The only change is updating the return type for `getEnabledLocalMcpServers`.

src/vs/workbench/contrib/mcp/common/mcpServerConnection.ts
```diff
@@ -64,16 +66,21 @@ export class McpServerConnection extends Disposable implements IMcpServerConnect
...
+			recentLogLines.push(message);
+			if (recentLogLines.length > McpServerConnection._maxRecentLogLines) {
+				recentLogLines.splice(0, recentLogLines.length - McpServerConnection._maxRecentLogLines);
+			}
 		}));
 
 		let didStart = false;
 		store.add(autorun(reader => {
-			const state = launch.state.read(reader);
+			const state = this._withLogTailOnError(launch.state.read(reader), recentLogLines);
```
connor4312: We would also want to detect 'runtime' failures, not just failures that cause the server to outright fail. E.g. perplexity MCP will start fine but fail to call tools if their domain isn't allowlisted.

I suggest not keeping a tail log like this, but instead

- Have some event `onPotentialSandboxBlock` that, servers that are sandboxed, fires when something that is certainly a failure (a domain block) or may be a failure (file access issue) is encountered
- While the connection it starting, collect any such events that get emitted. If the server fails to start, then pop a dialog. THis can be handled in `McpServer.start` so that you can respect `errorOnUserInteraction`
- We could then do a similar thing for tool calls: collect events that happen while a tool is running, and then if the tool call fails and we've found some of these events, present them to the user

dileepyavan: Yeah , I have another issue for runtime/toolcalls. Those failures are visible to the model. So was thinking about using askquestion tool to get them added. But I think we have all the logs and info to show it in the dialog may be we can do that and it feels like more in the flow of MCP setup and not interfere with chat.

src/vs/workbench/contrib/mcp/common/mcpServer.ts
```diff
@@ -493,6 +503,31 @@ export class McpServer extends Disposable implements IMcpServer {
...
+			const state = cnx?.state.read(reader);
+
+			if (cnx && state?.state === McpConnectionState.Kind.Error) {
+				const transitionedToError = this._lastErrorNotificationConnection !== cnx
+					|| this._lastErrorNotificationState !== McpConnectionState.Kind.Error;
+				if (transitionedToError) {
+					if (!this._isQuietStart) {
+						this.showInteractiveError(cnx, state, this._lastModeDebugged);
+					} else {
+						throw new UserInteractionRequiredError('start');
```
connor4312: This is an unhandled execption. The interaction required error should result in the promise returned from `start()` rejecting

I would handle this entirely within the start() method, start capturing events around [here](https://github.com/microsoft/vscode/blob/5496be1da981edd8d49e78f1eec416c56ed8ddb5/src/vs/workbench/contrib/mcp/common/mcpServer.ts#L632)

and then if the connection ended up in an error state, check before `return state;` to see if you captured any events that should result in a sandbox error being thrown instead

src/vs/workbench/contrib/mcp/common/mcpSandboxService.ts
```diff
@@ -63,13 +78,15 @@ export class McpSandboxService extends Disposable implements IMcpSandboxService
...
 	}
 
-	public async launchInSandboxIfEnabled(serverDef: McpServerDefinition, launch: McpServerLaunch, remoteAuthority: string | undefined, configTarget: ConfigurationTarget): Promise<McpServerLaunch> {
+	public async launchInSandboxIfEnabled(serverDef: McpServerDefinition, launch: McpServerLaunch, mcpResource: URI | undefined, remoteAuthority: string | undefined, configTarget: ConfigurationTarget): Promise<McpServerLaunch> {
 		if (launch.type !== McpServerTransportType.Stdio) {
 			return launch;
 		}
 		if (await this.isEnabled(serverDef, remoteAuthority)) {
+			const scanTarget = this._toMcpResourceTarget(configTarget);
+			const rootSandboxConfig = mcpResource ? (await this._mcpResourceScannerService.scanMcpServers(mcpResource, scanTarget)).sandbox : undefined;
```
connor4312: Sorry, to clarify, I think you should _keep_ the sandbox on the `McpServerDefinition` or `McpServerLaunch` (in `McpServerLaunch` is more appropriate since it does only apply for stdio) -- while it is collection-level at this point, I think it would be quite reasonable to have custom sandboxes per server in the future and so I don't have a problem putting in the McpServerLaunch/Definition directly. -- just that we should remove it from `IScannedMcpServer`, because the `IScannedMcpServer` is the representation of that server _in the mcp.json_ and denormalizing the sandbox into `IScannedMcpServer` caused https://github.com/microsoft/vscode/issues/297458

The other benefit of having it in the `McpServerDefinition`/`McpServerLaunch` is that the MCP server will automatically be restarted when the sandbox is updated, which won't happen if it's only connection-level.

connor4312 (changes requested):

connor4312 (approved):

PR #297874, opened 2026-02-26
Fix Integrated Browser Localhost Opener triggering in too many cases

### LINKED ISSUE #293015: Open in external browser opens another tab in vscode when `Open Localhost Links` is enabled

- VS Code Version: 1.109.0
- OS Version: Darwin arm64 25.2.0

Steps to Reproduce:

1. Enable `Open Localhost Links` in settings
2. Open a localhost link in Simple Browser.
3. Select `Open in External Browser` from the three-dot context menu.

https://github.com/user-attachments/assets/52a566a0-e49d-4c64-b2a9-778e9f5e2fcd

jruales:
* Fixes #293015

Turn the Localhost Opener into an external opener, which implicitly is gated by the `allowContributedOpeners` flag

### DIFF
--- a/src/vs/workbench/contrib/browserView/electron-browser/browserView.contribution.ts
+++ b/src/vs/workbench/contrib/browserView/electron-browser/browserView.contribution.ts
@@ -117,7 +117,11 @@ class LocalhostLinkOpenerContribution extends Disposable implements IWorkbenchCo
 		this._register(openerService.registerOpener(this));
 	}
 
-	async open(resource: URI | string, _options?: OpenInternalOptions | OpenExternalOptions): Promise<boolean> {
+	async open(resource: URI | string, options?: OpenInternalOptions | OpenExternalOptions): Promise<boolean> {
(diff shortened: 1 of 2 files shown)


src/vs/workbench/contrib/browserView/electron-browser/browserView.contribution.ts
```diff
@@ -117,7 +117,11 @@ class LocalhostLinkOpenerContribution extends Disposable implements IWorkbenchCo
 		this._register(openerService.registerOpener(this));
```
kycutler: Looking at how `allowContributedOpeners` is handled already, should we just change this to be `registerExternalOpener(this)` and let the service handle the logic? Then we can audit the places where the flag isn't yet passed but maybe should be

jruales: Done

kycutler (approved):

PR #297961, opened 2026-02-26
expose Copilot token info to ExP

ulugbekna:
## Summary

add `ICopilotTokenInfo` and wire it through default account service/provider interfaces\n- propagate Copilot token flags (`sn`, `fcv1`) into assignment filters\n- keep policy data and token info as separate concepts in runtime model\n- persist/recover token info via a cache wrapper while preserving backward compatibility with legacy cache shape\n\n## Notes\n- This includes a fix so cached token-entitlements reuse does not drop token flags on restart.\n- Existing unrelated workspace typecheck/build errors remain unchanged.


src/vs/workbench/services/accounts/browser/defaultAccount.ts
```diff
@@ -256,22 +273,31 @@ class DefaultAccountProvider extends Disposable implements IDefaultAccountProvid
...
 				if (accountId && policyData) {
 					this.logService.debug('[DefaultAccount] Initializing with cached policy data');
-					return { accountId, policyData };
+					return { accountPolicyData: { accountId, policyData, tokenEntitlementsFetchedAt, mcpRegistryDataFetchedAt }, copilotTokenInfo };
+				}
+
+				const { accountPolicyData, copilotTokenInfo: wrappedCopilotTokenInfo } = parsed;
+				if (accountPolicyData?.accountId && accountPolicyData?.policyData) {
+					this.logService.debug('[DefaultAccount] Initializing with cached policy data');
+					return { accountPolicyData, copilotTokenInfo: wrappedCopilotTokenInfo };
```
sandy081, line 310: I see you are migrating old format to new format. Can you please add comments here why this is being done and add a todo to remove after 6 months. Can you also store the data in new format during migration?

sandy081 (changes requested):

sandy081 (approved):

PR #298035, opened 2026-02-26
MCP Gateway: avoid blocking list calls on startup

### LINKED ISSUE #297780: MCP Gateway performance needs improvement
I had to disable the MCP Gateway being added to the Claude integration because it caused upwards of 1min+ TTFT.

https://github.com/microsoft/vscode/issues/297448

I [disabled it](https://github.com/microsoft/vscode-copilot-chat/pull/4001) for now... but we should have this feature! What it provides is quite nice.

RajeshKumar11:
Fixes #297780. Avoid blocking MCP gateway list calls on slow startup by returning cached data and refreshing in the background.


connor4312:
Merged in https://github.com/microsoft/vscode/pull/299072

PR #298161, opened 2026-02-27
Add optional model parameter to runSubagent tool

### LINKED ISSUE #298380: Allow runSubagent tool to specify a language model
## Problem

When the LLM invokes the `runSubagent` built-in tool, the subagent always inherits the parent agent's model (or the custom agent's configured model). There is no way for the LLM to explicitly choose a different model for a subagent invocation, even when the user asks for it (e.g. "use Claude to research this").

## Proposal

Add an optional `model` parameter to the `runSubagent` tool schema that accepts a qualified model name in `"Model Name (Vendor)"` format. The schema should dynamically enumerate available agent-mode-suitable models so the LLM knows which values are valid.

### Model resolution priority
1. **Explicit `model` parameter** — if provided and resolved
2. **Agent-configured model** — from the custom agent's `.agent.md` file
3. **Main model** — the model running the parent agent

A multiplier-based cost guard applies at every level to prevent a subagent from using a more expensive model than the main agent.

### Dynamic enum
The tool schema should include an `enum` of available models filtered by `suitableForAgentMode()`, and re-register when models change so the enum stays current.

digitarald:
Adds an optional `model` parameter to the `runSubagent` tool, allowing the LLM to specify which language model a subagent should use.

Fixes #298380

### Model parameter
- Added `model?: string` to `IRunSubagentToolInputParams` interface and tool schema
- Updated description to instruct the LLM to use qualified name format `"Model Name (Vendor)"`

### Dynamic model enum
- Schema `model` property includes a dynamic `enum` of available models filtered by `suitableForAgentMode()` (requires `toolCalling` capability and `agentMode !== false`)
- Uses `Event.any()` to merge config changes and `onDidChangeLanguageModels` events, triggering tool re-registration when models change
- Enum is conditionally omitted when no models are available (avoids invalid empty `enum: []`)

### Model resolution priority
- **Explicit model** (from `model` param) > **Agent configured model** > **Main model**
- If explicit model is not found, falls through to agent/main model with a warning log
- Multiplier-based cost guard applies to all sources — prevents subagent from using a more expensive model than the main agent

### Tests
- 23 tests covering: enum population, agent-mode filtering, `agentMode: false` exclusion, empty-model omission, explicit model resolution, fallback chains, multiplier guard


aeschli:
@digitarald Can you link to an issue that describes the need for this?
Also, this is for the next milestone, right?

roblourens (dismissed):

roblourens (dismissed):

dmitrivMS (dismissed):

src/vs/workbench/contrib/chat/test/common/tools/builtinTools/runSubagentTool.test.ts
```diff
@@ -485,4 +485,211 @@ suite('RunSubagentTool', () => {
...
+
+			assert.ok(result);
+			assert.deepStrictEqual(result.toolSpecificData, {
+				kind: 'subagent',
+				description: 'test task',
+				agentName: undefined,
+				prompt: 'test',
+				modelName: 'GPT-4o',
+			});
+		});
```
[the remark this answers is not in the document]
digitarald: Addressed — the original test was renamed to "falls back to main model when explicit model is not found and no agent" (no agent scenario), and a separate test "falls back to agent configured model when explicit model is not found" was added covering the explicit-model-invalid + agent-has-configured-model scenario.

src/vs/workbench/contrib/chat/common/tools/builtinTools/runSubagentTool.ts
```diff
@@ -73,14 +77,20 @@ export class RunSubagentTool extends Disposable implements IToolImpl {
...
 		@ILogService private readonly logService: ILogService,
-		@ILanguageModelToolsService private readonly toolsService: ILanguageModelToolsService,
 		@IConfigurationService private readonly configurationService: IConfigurationService,
 		@IPromptsService private readonly promptsService: IPromptsService,
 		@IInstantiationService private readonly instantiationService: IInstantiationService,
 		@IProductService private readonly productService: IProductService,
 	) {
 		super();
-		this.onDidUpdateToolData = Event.filter(this.configurationService.onDidChangeConfiguration, e => e.affectsConfiguration(ChatConfiguration.SubagentToolCustomAgents));
+		this.onDidUpdateToolData = Event.debounce(
```
aeschli, line 88: what's the rational for the delay?

aeschli (changes requested):

src/vs/workbench/contrib/chat/common/tools/builtinTools/runSubagentTool.ts
```diff
@@ -83,6 +83,8 @@ export class RunSubagentTool extends Disposable implements IToolImpl {
 		@IProductService private readonly productService: IProductService,
 	) {
 		super();
+		// Debounce because onDidChangeLanguageModels fires per vendor/model during
```
aeschli, line 86: This should be improved in the onDidChangeLanguageModels then. Please file a issue against onDidChangeLanguageModels to collapse the events on start up and add the issue number in the comment.

aeschli (commented):
The tool description now lists all model names. That's for me around 30 models and I could have many more with BYOK. This increasing the context size and COGS but I'm not sure I'll profit much from that model selection. Again, the question is still, how would the LM chose a model over the other. The model names are not descriptive
I suggest the main agent should explicitly name the models
`Run the search subagent with either the 'Opus 4.6 (copilot)' model when extra reasoning is required.`
Given this is an experimental feature this is the a cost efficient way.
That way you can also remove the onDidChangeLanguageModels event

hediet:
Implemented in https://github.com/microsoft/vscode/pull/308784

PR #298206, opened 2026-02-27
Release notes: skip Insiders draft on stable builds

RajeshKumar11:
Fixes #292861

## Problem

When a new stable release ships, `code.visualstudio.com/raw/v{major}_{minor}.md` may still serve the **Insiders draft** until the stable release notes are published. Stable users would see incorrect "Insiders" release notes.

Confirmed live — fetching `v1_110.md` today returns:
```
# February 2026 Insiders (version 1.110)
```

## Fix

Insiders drafts always contain `Insiders` in the first heading. After fetching, on `quality === 'stable'` builds, reject the content if the heading matches and let the release notes panel stay closed silently (same behaviour as a 404).

Changed files:
- **`releaseNotesEditor.ts`** — adds `isInsidersReleaseNotes()` helper (exported for testability) and calls it in `fetchReleaseNotes()`
- **`releaseNotesRenderer.test.ts`** — adds 4 unit tests for the helper covering Insiders heading, stable heading, "Insiders" in body only, and empty string

## Test plan
- [ ] Build stable VS Code — release notes panel should not appear while `v{version}.md` returns an Insiders heading
- [ ] Build insider VS Code — release notes panel shows normally (guard is `quality === 'stable'` only)
- [ ] Existing `Release notes renderer` snapshot tests still pass


ntrogh (changes requested):
@RajeshKumar11 See comment about frontmatter field `ProductEdition`

RajeshKumar11:
@alexr00 The endpoint used is `https://code.visualstudio.com/raw/v{version}.md`, and it strips front-matter before serving the content. This is actually proven by the existing validation just above the `isInsidersReleaseNotes` call (line 231):

```ts
if (!text || (!/^#\s/.test(text) && !useCurrentFile)) {
    throw new Error('Invalid release notes');
}
```

If front-matter were present, the fetched text would start with `---`, which would immediately fail this check and never reach `isInsidersReleaseNotes()`. So for this endpoint the `ProductEdition` front-matter check is effectively dead code — the heading fallback is the only signal that can fire.

If the intent is to rely solely on `ProductEdition`, the fix would need to either use a different endpoint that preserves front-matter (e.g. the GitHub raw URL), or relax/remove the `!/^#\s/` validation so front-matter can pass through. Happy to make that change if preferred — just let me know which direction to take.

src/vs/workbench/contrib/update/browser/releaseNotesEditor.ts
```diff
@@ -774,6 +781,21 @@ export class ReleaseNotesManager extends Disposable {
...
+export function isInsidersReleaseNotes(text: string): boolean {
+	if (text.startsWith('---')) {
+		const end = text.indexOf('\n---', 3);
+		if (end !== -1) {
+			const frontmatter = text.substring(0, end + 4);
+			return /^ProductEdition:\s*Insiders\s*$/m.test(frontmatter);
+		}
+	}
+	return false;
+}
```
alexr00, line 802: We have a yaml parser here. Can we use that instead of regex?

https://github.com/rajeshkumar11/vscode/blob/9d41180d71f47beb9d53f6db5dbc09bede96947c/src/vs/base/common/yaml.ts#L8-L27

alexr00 (changes requested):
Thanks for the changes! One more request.

src/vs/workbench/contrib/update/browser/releaseNotesEditor.ts
RajeshKumar11: Updated — now using the existing `parse` function from `src/vs/base/common/yaml.ts` to read the `ProductEdition` field from the YAML front-matter instead of the regex.

alexr00 (commented):
I just tested this out, and it looks like the yaml frontmatter isn't included at the URL we fetch from (as @RajeshKumar11 pointed out above). Example: https://code.visualstudio.com/raw/v1_109.md

@ntrogh, is the original way to detect insiders (probing for "Insiders" on the first line) ok?

alexr00 (commented):
@RajeshKumar11, thanks for your patience. Your original change in 40d5857 looks good. However, this means that even if a user manually runs the "Show release notes" command, they will not get release notes. @ntrogh, is that what you want to happen, or should we only avoid showing the release notes when they are automatically set to show?

src/vs/workbench/contrib/update/browser/releaseNotesEditor.ts
```diff
@@ -228,10 +228,18 @@ export class ReleaseNotesManager extends Disposable {
...
 				throw new Error('Invalid release notes');
 			}
 
+			// On stable builds, reject release notes that are still the Insiders pre-release
+			// version when shown automatically. This can happen when a new stable release ships
+			// before the final stable release notes are published to the website.
+			// When the user manually requests release notes, always show them.
+			if (automatic && !useCurrentFile && this._productService.quality === 'stable' && isInsidersReleaseNotes(text)) {
+				throw new Error('not found');
+			}
```
alexr00: This check should be in `show`, not `loadReleaseNotes`.

alexr00 (changes requested):

src/vs/workbench/contrib/update/browser/releaseNotesEditor.ts
RajeshKumar11: Done — moved the check into `show()`. The `loadReleaseNotes` signature is back to just `(version, useCurrentFile)` and the Insiders guard now runs in `show()` after the text is fetched, returning `false` instead of throwing.

alexr00 (changes requested):
Looks like the yaml frontmatter check snuck back in.

PR #298252, opened 2026-02-27
Making the whitespaces align vertically with variable font text

aiday-mar:
fixes https://github.com/microsoft/vscode/issues/298249

- The spans are bottom aligned to the bottom of the larger line height. A padding of 3px was added to not have the text glued to the bottom of the line. 
- The ascent and descent are calculated in the FontInfo to form the line height
- The getFontInfoAtPosition returns a FontInfo object now instead of { fontSize: number, fontFamily: string}
- To avoid layer importing issues I had to pass in the FontMeasurements.readFontInfo method into the constructor of the ViewModelImpl
- fontSize on the ModelDecorations has been changed to a number for several reasons:
  -  it mirrors the lineHeight option which is a number multiplier of the default line height
  -  it makes it easier to compute the bare font info using the number fontSize
  -  I need it to be a number so the font tokenization provider can return the multiplier without making it into a string (which is not possible since we don't have access to the configuration object in the provider)


src/vs/editor/browser/viewParts/whitespace/whitespace.ts
```diff
@@ -28,13 +29,15 @@ export class WhitespaceOverlay extends DynamicViewOverlay {
...
 	private _renderResult: string[] | null;
+	private readonly _fontMetricsCache: FontMetricsCache;
 
 	constructor(context: ViewContext) {
 		super();
 		this._context = context;
 		this._options = new WhitespaceOptions(this._context.configuration);
 		this._selection = [];
 		this._renderResult = null;
+		this._fontMetricsCache = new FontMetricsCache();
```
alexdima: Would we want to share this instance across multiple editor views? Otherwise it needs to be recreated every time. Come to think of it, shouldn't we simply enhance `FontInfo` to contain also these two properties -- ascent and descent?

There is a lot of infrastructure in place to store that across editors, to reset the cache when the DPI changes (like plugging in a monitor or moving the window to another monitor), as well as to cache the values across vscode restarts to avoid flickering or slowness on startup.

This is all implemented at https://github.com/microsoft/vscode/blob/cfb11567478af7289ef213161f2b5cdc98fef026/src/vs/editor/browser/config/fontMeasurements.ts#L37 and I think you should be able to enhance it to add these extra props?

src/vs/editor/browser/viewParts/whitespace/whitespace.ts
```diff
@@ -211,18 +215,25 @@ export class WhitespaceOverlay extends DynamicViewOverlay {
 				continue;
 			}
 
+			// Variable fonts: cy_f = (H - (A_f - D_f) + max_i(A_i - D_i)) / 2
+			const fontInfo = this._context.viewModel.getFontAtPosition(new Position(lineNumber, charIndex + 1));
+			const fontFamily = fontInfo?.fontFamily ?? this._options.fontFamily;
+			const fontSize = fontInfo?.fontSize ?? this._options.fontSize;
+			const metrics = this._fontMetricsCache.getMetrics(fontFamily, fontSize);
+			const cy = (lineHeight - (metrics.ascent - metrics.descent) + maximumAscentMinusDescent) / 2;
+
```
alexdima: I'm not against doing something super smart like we're doing here, but I wonder why not have the lineHeight/2 and use the current line height of the position where the whitespace char is rendered? I mean what are the advantages of using this formula vs a simple lineHeight/2 approach?

alexdima (changes requested):
Nice! I just have these 2 questions / observations around:
* what is the advantage of the super complex formula?
* if we do need ascent and descent, can we enhance the existing infrastructure we have around DOM read font information

src/vs/editor/browser/config/charWidthReader.ts
```diff
@@ -141,3 +141,37 @@ export function readCharWidths(targetWindow: Window, bareFontInfo: BareFontInfo,
...
+	private static _instance: FontHeightReader | null = null;
+
+	public static getInstance(): FontHeightReader {
+		if (!FontHeightReader._instance) {
+			FontHeightReader._instance = new FontHeightReader();
+		}
+		return FontHeightReader._instance;
+	}
+
+	private readonly _cache = new Map<string, number>();
```
alexdima: No need for this cache

src/vs/editor/browser/viewParts/viewLines/viewLines.css
```diff
@@ -69,8 +71,16 @@
 	top: 0;
 	bottom: 0;
 	position: absolute;
+	display: flex;
+	align-items: flex-end;
 }
 
+.monaco-editor .lines-content > .view-lines > .view-line > span > span {
+	line-height: normal;
+	padding-bottom: 3px;
```
alexdima: There should be 0 pixel changes for rendering lines which don't use custom font size / custom line height

alexdima (changes requested):
The whitespace rendering is no longer middle positioned for lines who don't use custom fonts:

| BEFORE | AFTER |
|---|---|
| image | image

src/vs/editor/browser/config/charWidthReader.ts
marsam2489-lang: Can i known what the label of the line that their need?

src/vs/editor/browser/viewParts/whitespace/whitespace.ts
```diff
@@ -210,25 +212,29 @@ export class WhitespaceOverlay extends DynamicViewOverlay {
 			if (!visibleRange) {
 				continue;
 			}
+			const fontInfo = this._context.viewModel.getFontAtPosition(new Position(lineNumber, charIndex + 1));
```
alexdima, line 215: As discussed, it would be good to raise this call outside of the character loop / try to reuse the available lineData.inlineDecorations and have the inline decoration point to the font info.

alexdima (changes requested):

Issue #241078, opened 2025-02-18
Environment from extensionHost debug configuration is not inherited by vscode.ShellExecution tasks


eliericha:
Does this issue occur when all extensions are disabled?: Yes

- VS Code Version: 1.97.2
- OS Version: macOS 14.6.1

Steps to Reproduce:

1. Define an `extensionHost` debug configuration with environment variables specified in `env` or `envFile`.
2. Write code in the extension or tests that spawns a `vscode.ShellExecution` task that makes use of the above environment variables.
3. Observe that the spawned task does not inherit the environment variables from the debug configuration, contrary to [the documentation of `vscode.ShellExecutionOptions`](https://github.com/microsoft/vscode/blob/1.97.2/src/vscode-dts/vscode.d.ts#L8882) which states that by default the parent process environment is inherited.

I created a [self-contained workspace](https://github.com/eliericha/vscode-extension-samples/tree/topic/test-env/test-env) with the above steps already prepared. All you have to do is to run the `Run Extension Tests` debug configuration defined in `launch.json` to observe the problem.

I suspect that the issue is that the `vscode.ShellExecution` implementation caches `process.env` too early, before the environment variables of the debug configuration were applied.

I hope this helps.

Thanks!

meganrogge:
Hi, no progress, and no plans to make progress realistically. If you want to take a look, that's welcomed.

PR #298276, opened 2026-02-27
include debug extension host env in shell env (#241078)

eliericha:
This PR proposes a fix for #241078 .

With some investigation and some help from AI, I managed to produce this fix which includes adds the debug extension host environment to the terminal environment.

With that I was able to observe that environment variables added in the `env` property of a launch configuration or through an env file and the `envFile`, are correctly inherited into terminals spawned by the debugged instance.

More specifically, tasks now inherit the environment provided by the debug launch configuration.

Given my little experience with the codebase, I'm not sure this works in all configurations (web, electron, remote, etc) so guidance would be appreciated to steer this in the right direction.

Thanks!

### DIFF
--- a/src/vs/workbench/contrib/terminal/electron-browser/localTerminalBackend.ts
+++ b/src/vs/workbench/contrib/terminal/electron-browser/localTerminalBackend.ts
@@ -300,7 +300,7 @@ class LocalTerminalBackend extends BaseTerminalBackend implements ITerminalBacke
 
 		// If running in the context of an extension development host, include the environment derived from the launch configuration
 		if (this._environmentService.debugExtensionHost.env) {
-			Object.assign(env, this._environmentService.debugExtensionHost.env);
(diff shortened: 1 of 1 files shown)


meganrogge (changes requested):
Thanks for the contribution! The approach is directionally correct and consistent with how `localProcessExtensionHost.ts` already applies `debugExtensionHost.env`.

One issue that needs addressing:

**Mutating a potentially shared object**: `mergeEnvironments` mutates `env` in place. Since `getShellEnv()` delegates to `process.shellEnv()`, which may cache and return the same object reference, this mutation could leak debug env vars into other consumers of `process.shellEnv()`. Please clone first:

```typescript
const env = { ...await this._shellEnvironmentService.getShellEnv() };
```

Minor note: this only covers the local (Electron) terminal backend — remote extension host debugging won't benefit. Not a blocker for this PR but worth tracking separately if needed.

meganrogge (approved):

dmitrivMS (approved):

PR #298335, opened 2026-02-27
Sessions window: auth fixes for PR check

osortega:


justschen (approved):

bpasero:
I would like to ask @TylerLeonhardt to reason about this change, it seems weird to me having to pass around an auth token from core to the extension. I think we have methods for an extension to get this information without having to pass tokens around?

bpasero (changes requested):
I have a simpler fix I think, will follow up offline.

PR #298377, opened 2026-02-27
Polish question carousel

daviddossett:
Polishes the question carousel from a flat list of questions into a tabbed multi-step flow with a final review step.

https://github.com/user-attachments/assets/ab94c067-9b3a-4796-856c-6f9bbe478f56

## What changed

- **Tab bar navigation** — Each question gets its own tab. Users step through one question at a time, then land on a **Review** tab that summarizes all answers before submitting. Single-question carousels skip the tab bar entirely.
- **Review tab** — Shows a read-only summary of every answer (or "skipped"). Auto-focuses when the user advances past the last question.
- **Submit hint** — Footer displays `⌘⏎ to submit` (macOS) / `Ctrl+Enter to submit` so the shortcut is discoverable.
- **Tab completion indicators** — Answered tabs show a checkmark icon; the indicator is cleared on dispose.
- **Readable tab titles** — Model-generated headers (`FocusArea`, `snake_case_header`) are normalized to sentence case (`Focus area`, `Snake case header`) for display. Original headers are preserved for answer correlation.
- **Freeform priority** — When a freeform input has text, it takes priority over a radio selection on single-select questions.
- **Progress message** — Changed from "Analyzing your answers..." to "Reviewing your answers".

## Execute toolbar

- **Cancel** button now hides when the user types text during a question carousel or tool confirmation, matching the behavior during regular in-progress requests.
- **Steer/Queue submenu** now appears during question carousel and tool confirmation states when the user has typed text. Previously it only appeared during `requestInProgress`, which is `false` during these paused states.
- Cleaned up unused `requestInProgressWithoutInput` and `pendingToolCall` variables.

## Accessibility

- Tab bar uses `role="tablist"` / `role="tab"` with `aria-selected`, `aria-controls`, and roving `tabIndex`.
- Question container uses `role="tabpanel"`.
- Arrow-key navigation between tabs.

## Theme / CSS

- 2026-dark theme: border-radius tokens applied to carousel container and input parts.
- `list.activeSelectionBackground` adjusted for 2026-dark.
- Minor spacing tweaks: reduced question-list padding, wider footer gap.

## Tests

- Updated carousel tests for tab-bar UI (selectors, multi-question flow).
- Added unit test for `formatHeaderForDisplay` covering camelCase, already-spaced, and snake_case inputs.


src/vs/workbench/contrib/chat/browser/widget/chatContentParts/media/chatQuestionCarousel.css
```diff
@@ -144,41 +142,30 @@
...
+	padding-bottom: 8px;
 	min-width: 0;
 
 	/* some hackiness to get the focus looking right */
-	.chat-question-list-item:focus:not(.selected),
+	.chat-question-list-item:focus,
+	.chat-question-list-item:focus-visible,
 	.chat-question-list:focus {
 		outline: none;
 	}
```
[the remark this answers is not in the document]
daviddossett: This is intentional to match quick pick styles. We show a checkmark on the far right of the selected item, satisfying the selected state indicator.

The focus outlines is way too heavy.

Yoyokrazy (dismissed):

src/vs/workbench/contrib/chat/browser/actions/chatExecuteActions.ts
```diff
@@ -194,6 +194,7 @@ const pendingToolCall = ContextKeyExpr.or(
 	ChatContextKeys.Editing.hasQuestionCarousel,
 );
 const whenNotInProgress = ChatContextKeys.requestInProgress.negate();
+const whenNoRequestOrPendingToolCall = requestInProgressOrPendingToolCall!.negate();
```
connor4312, line 189: > Send hidden while a request is in progress or a pending tool call / question carousel is active (previously only hidden during request progress).

What's the reasoning behind this change? This is a pretty common pattern imo. It's something I personally do a lot and I know other people folks do that from hearing feedback on steering/queuing

connor4312 (changes requested):
requesting changes on ^ it's not a great experience imo and hitting "enter" is a no-op

Image

src/vs/workbench/contrib/chat/browser/actions/chatExecuteActions.ts
daviddossett: Checking, but the idea should be similar to the regular prompting experience where the steer button shows up as soon as you type

daviddossett: Ah ok, this somehow regressed:
CleanShot 2026-03-03 at 09 57 02@2x

Working on a fix

daviddossett: Fixed per DM chat

connor4312 (approved):

Issue #182795, opened 2023-05-17
Simple browser will be opened multiple times with race


mustard-mh:
Does this issue occur when all extensions are disabled?: Yes

- VS Code Version: 1.78.2
- OS Version: macOS Ventura Version 13.4 Beta

Steps to Reproduce:

1. Download or package extension yourself in reproduce repo https://github.com/mustard-mh/test/tree/vscode/simple-browser-bug
2. Install extension in your VSCode Desktop
3. Reload window without Simple Browser tab focus

![May-18-2023 02-39-18](https://github.com/microsoft/vscode/assets/20944364/5cefa34d-992c-4dd3-abec-59c1d8016e09)

mustard-mh:
It might be caused by

- `restore` [[1](https://github.com/mustard-mh/vscode/blob/46c9471450f82ef2500d86c555504502652d9aa5/extensions/simple-browser/src/extension.ts#L39-L43)] called means it already exist a Simple Browser panel waiting to be restored (panel already created)
- `show` will create a Simple Browser if `_activeView` is undefined

It can race if we `show` before `restore`

ashutosh887:
I would like to work on this @mustard-mh @anthony @brigano @mjbvz

jammel-yeboah:
> I would like to work on this @mustard-mh @anthony @brigano @mjbvz

Hi! Wanted to give you a heads-up that I've been working on this issue and I am about to create a PR. Have you already started working on this issue?

salvatoreOm:
HI, I would like to work on this issue :)

KSSaiTeja:
Hey, I would like to work on this issue ..

Akashpaila:
Is this still open ?

robincaloudis:
Hello @Akashpaila, in https://github.com/microsoft/vscode/pull/197188 I fixed the issue. Unfortunately, no one reviewed the PR so far. As no review was given for over 3 months, could someone else review the PR? @mjbvz, do you mind changing the assignee of https://github.com/microsoft/vscode/pull/197188?

y0sh1ne:
@mustard-mh 
Hi, I am not sure what this issue mean. 

Does it means you have write a test extension, and the function of this extension is automatically open a SimpleBrowser tab when user use the vscode commend `reload Window`.

If so, then I think *the extension you create* have the responsibility to detect "does the tab already exist", instead of the *vscode*.

y0sh1ne:
[...]

PR #298575, opened 2026-03-01
Fix multiple Simple Browser instances on rapid clicks (Issue #182795)

Buconero03:
"This PR fixes the race condition in Simple Browser that caused multiple webviews to open on rapid clicks or during editor restore. Added a check using vscode.window.tabGroups to see if a restoring tab already exists."


mjbvz (changes requested):
Please fix the formatting

Issue #293579, opened 2026-02-07
Copilot Chat - Embedded terminal output view flashes during command execution


nathan815:
Type: <b>Bug</b>

Description:
When a command is running in the embedded terminal output view (shown in the Copilot Chat panel), the view flashes/flickers continuously every time the output updates.

Steps to Reproduce:

1. Open Copilot Chat in VS Code
2. Ask Copilot to run a command that produces incremental output (e.g., dotnet build)
3. Observe the embedded terminal output view while the command runs

Expected Behavior:
The terminal output should update smoothly without visual flashing.

Actual Behavior:
The embedded terminal output view flashes/flickers repeatedly and quickly. It appears to flicker when the output is updated.

Affected Commands:

dotnet build (confirmed)
Possibly other commands with streaming output

Extension version: 0.37.1
VS Code version: Code 1.109.0 (bdd88df003631aaa0bcbe057cb0a940b80a476fa, 2026-02-04T02:01:38.288Z)
OS version: Windows_NT x64 10.0.26200
Modes:

<details>
<summary>System Info</summary>

|Item|Value|
|---|---|
|CPUs|AMD EPYC 7763 64-Core Processor                 (16 x 2445)|
|GPU Status|2d_canvas: unavailable_software<br>GPU0: VENDOR= 0x1414, DEVICE=0x008c [Microsoft Basic Render Driver], DRIVER_VENDOR=Google, DRIVER_VERSION=10.0.26100.7309<br>GPU1: VENDOR= 0x1414, DEVICE=0x008c [Microsoft Basic Render Driver], DRIVER_VERSION=10.0.26100.7309<br>GPU2: VENDOR= 0x1414, DEVICE=0x008c [Microsoft Basic Render Driver], DRIVER_VERSION=10.0.26100.7309<br>Machine model name: <br>Machine model version: <br>direct_rendering_display_compositor: disabled_off_ok<br>gpu_compositing: disabled_software<br>multiple_raster_threads: enabled_on<br>opengl: disabled_off<br>rasterization: disabled_software<br>raw_draw: disabled_off_ok<br>skia_graphite: disabled_off<br>trees_in_viz: disabled_off<br>video_decode: disabled_software<br>video_encode: disabled_software<br>webgl: unavailable_software<br>webgl2: unavailable_software<br>webgpu: unavailable_software<br>webnn: unavailable_software|
|Load (avg)|undefined|
|Memory (System)|63.95GB (38.65GB free)|
[...]

PR #298598, opened 2026-03-01
Fix chat terminal flicker during streaming

### LINKED ISSUE #293128: Chat inline terminal flickers a lot
Repro is running xterm.js integration tests:

Image

sanchirico:
The embedded terminal in chat flickers to blank during command execution. I used `npm run test-node` to reproduce/test. I think the issue is in `chatTerminalCommandMirror.ts` when it can't append incrementally it calls `xterm.reset()` then `xterm.write(text)` as two separate calls and xterm renders a blank frame in between.

I changed it to do `write('\x1bc' + text)` instead which puts the reset and the new content in the same write call so xterm handles it all before painting. `\x1bc` is the same thing as `reset()` according to the xterm.js docs (`Perform a full reset (RIS, aka '\x1bc')`). I saw `terminalProcessManager.ts` already does this for seamless relaunch so it seemed like a valid approach.

Tested on Windows and seems to not flicker now.

Tries to fix #293128 and #293579


meganrogge (approved):
Thanks so much!

joshspicer (approved):

PR #298767, opened 2026-03-02
chat: make sendChatRequest extensionIdentifier parameter optional

connor4312:
Makes the 'from' parameter optional (ExtensionIdentifier | undefined) in the
sendChatRequest method chain across ILanguageModelsService, RPC protocol, and
implementations. This allows internal VS Code calls to pass undefined instead
of instantiating invalid ExtensionIdentifier('core') identifiers.

- Updates ILanguageModelsService.sendChatRequest signature
- Updates ILanguageModelChatProvider.sendChatRequest signature
- Updates ExtHostLanguageModelsShape protocol
- Updates ExtHostLanguageModels \ implementation
- Removes new ExtensionIdentifier('core') from 5 internal callers
- Passes undefined as requestInitiator when from is absent
- Bumps vscode.proposed.chatProvider API version to 5
- Updates test mocks to accept optional from parameter

Refs https://github.com/microsoft/vscode/issues/290436


lramos15 (approved):
Good fix, thanks!

PR #298786, opened 2026-03-02
constrain and make chat questions scrollable

### LINKED ISSUE #298287: question exceeds chat window size
Image

meganrogge:
fixes #298287

Makes the chat question carousel scrollable — wraps the input area in a `DomScrollableElement` with a max-height so long option lists scroll instead of pushing the chat UI off-screen. A ResizeObserver keeps the scroll layout in sync.

https://github.com/user-attachments/assets/f58e227c-c5c6-4e09-8cd5-e112fc60780c


src/vs/workbench/contrib/chat/browser/widget/chatContentParts/chatQuestionCarouselPart.ts
```diff
@@ -321,6 +326,41 @@ export class ChatQuestionCarouselPart extends Disposable implements IChatContent
...
+		}
+
+		const scrollableNode = inputScrollable.getDomNode();
+		const scrollableContent = scrollableNode.firstElementChild;
+		if (!dom.isHTMLElement(scrollableContent)) {
+			return;
+		}
+		const maxContainerHeight = Math.min(
+			ChatQuestionCarouselPart.MAX_INPUT_CAROUSEL_HEIGHT_PX,
+			dom.getWindow(this.domNode).innerHeight * ChatQuestionCarouselPart.MAX_INPUT_CAROUSEL_VIEWPORT_RATIO
```
roblourens: The window height is not really meaningful since you don't know how much of that is the actual chat.

https://github.com/user-attachments/assets/30cc4c71-5c5c-4c92-b64a-bd18614def45

Can you do this based on the flex layout without manually computing a height?

src/vs/workbench/contrib/chat/browser/widget/chatContentParts/chatQuestionCarouselPart.ts
```diff
@@ -321,6 +326,41 @@ export class ChatQuestionCarouselPart extends Disposable implements IChatContent
...
+			Number.parseFloat(computedStyle.paddingBottom || '0');
+
+		const nonScrollableContentHeight = Array.from(this._questionContainer.children)
+			.filter(child => child !== scrollableNode)
+			.reduce((sum, child) => sum + (child as HTMLElement).offsetHeight, 0);
+
+		const availableScrollableHeight = Math.floor(maxContainerHeight - contentVerticalPadding - nonScrollableContentHeight);
+		const constrainedScrollableHeight = Math.max(0, availableScrollableHeight);
+
+		scrollableNode.style.height = `${constrainedScrollableHeight}px`;
```
roblourens: Can it set its height based on CSS, then `scanDomNode` handles measuring the content?

src/vs/workbench/contrib/chat/browser/widget/chatContentParts/media/chatQuestionCarousel.css
```diff
@@ -4,16 +4,19 @@
...
 
 /* input specific styling */
-.interactive-session .interactive-input-part > .chat-question-carousel-widget-container .chat-question-carousel-container {
+.interactive-session .interactive-input-part > .chat-question-carousel-widget-container .chat-question-carousel-container,
+.interactive-session .interactive-input-part .interactive-input-and-edit-session > .chat-question-carousel-widget-container .chat-question-carousel-container {
 	margin: 0;
 	border: 1px solid var(--vscode-input-border, transparent);
 	background-color: var(--vscode-editor-background);
 	border-radius: 4px;
+	max-height: min(420px, 45vh);
```
roblourens: How does this fix it? Isn't vh relative to the window?

src/vs/workbench/contrib/chat/browser/widget/chatContentParts/chatQuestionCarouselPart.ts
```diff
@@ -321,6 +324,42 @@ export class ChatQuestionCarouselPart extends Disposable implements IChatContent
...
+			.filter(child => child !== scrollableNode)
+			.reduce((sum, child) => sum + (child as HTMLElement).offsetHeight, 0);
+
+		const availableScrollableHeight = Math.floor(maxContainerHeight - contentVerticalPadding - nonScrollableContentHeight);
+		const constrainedScrollableHeight = Math.max(0, availableScrollableHeight);
+
+		// Constrain the content element (DomScrollableElement._element) so that
+		// scanDomNode sees clientHeight < scrollHeight and enables scrolling.
+		// The wrapper inherits the same constraint via CSS flex.
+		scrollableContent.style.height = `${constrainedScrollableHeight}px`;
```
roblourens, line 458: I still don't really understand why this is necessary. Seems like `scanDomNode` is meant to measure things for you.

src/vs/workbench/contrib/chat/browser/widget/chatContentParts/media/chatQuestionCarousel.css
meganrogge: It is. This is only an upper bound for the carousel size (so it doesn’t grow too tall). The actual scroll viewport calculation no longer uses `window.innerHeight`; it uses `this._questionContainer.clientHeight `(the resolved layout size in context), so behavior tracks the real available space in the chat UI.

src/vs/workbench/contrib/chat/browser/widget/chatContentParts/chatQuestionCarouselPart.ts
meganrogge: `scanDomNode` is measure-only; we still need the measured element (`inputContainer`) to be height-bounded. Otherwise it can auto-grow with content and not report overflow. Setting `scrollableContent.style.height` is one way to enforce that bound.

roblourens (dismissed):

dmitrivMS (approved):

PR #298867, opened 2026-03-03
Fix handoff widget visibility: derive from response mode, persist modeInfo

digitarald:
Fixes three bugs with the handoff suggest-next widget in chat:

1. **Handoffs shown during streaming** — Widget now only appears after the response completes (and not if cancelled). Added explicit `hide()` on new request and early return for incomplete/cancelled responses.

2. **Handoffs based on wrong mode** — Widget now derives the mode from the last response's `modeInfo` (via `findModeByName`/`findModeById`) instead of reading the current mode picker selection. This ensures handoffs persist when switching mode (e.g. viewing a plan response then switching to agent).

3. **Handoffs missing on session restore** — `modeInfo` is now persisted through serialization (`ISerializableChatRequestData`, `toExport()`, `_deserializeRequest()`), and `renderChatSuggestNextWidget()` is called at the end of `setModel()`.

**Note:** Pre-existing sessions without persisted `modeInfo` fall back to the current mode picker (graceful degradation — no crash).

#### Fixed

1. Handoffs only appear when plan response is done
2. Handoffs appear in prior plan response
3. Handoffs don't disappear when user switched to agent, as the prior response was a plan mode response

https://github.com/user-attachments/assets/b2d75123-b7a9-42c8-8047-2e8560fe84ad

### Changes

- **chatWidget.ts**: Derive `responseMode` from `modeInfo` on the last response's request, using `findModeByName` for named modes and `findModeById` for builtin modes. Hide widget on new request and when cancelled.
- **chatModes.ts**: `findModeByName` uses `getCustomModes()` (respects `hasToolsAgent` guard) instead of raw `_customModeInstances` access.
- **chatModel.ts**: Add `modeInfo` to `ISerializableChatRequestData`, `_deserializeRequest()`, and `toExport()`.
- **chatSessionOperationLog.ts**: Include `modeInfo` in operation log schema.

### Test
Added `modeInfo roundtrips through serialization` test verifying the ser/deser roundtrip.


src/vs/workbench/contrib/chat/common/chatModes.ts
```diff
@@ -207,7 +207,7 @@ export class ChatModeService extends Disposable implements IChatModeService {
 	}
 
 	findModeByName(name: string): IChatMode | undefined {
-		return this.getBuiltinModes().find(mode => mode.name.get() === name) ?? this.getCustomModes().find(mode => mode.name.get() === name);
+		return this.getBuiltinModes().find(mode => mode.name.get() === name) ?? Array.from(this._customModeInstances.values()).find(mode => mode.name.get() === name);
```
aeschli: This circumvents the checks that are in getCustomModes.
Looks wrong...

src/vs/workbench/contrib/chat/browser/widget/chatWidget.ts
```diff
@@ -1173,27 +1173,35 @@ export class ChatWidget extends Disposable implements IChatWidget {
...
-		// Custom modes all have kind='agent' but different IDs.
-		const currentMode = this.input.currentModeObs.get();
-		const handoffs = currentMode?.handOffs?.get();
 
-		// Only show if: mode has handoffs AND chat has content AND not quick chat
-		const shouldShow = currentMode && handoffs && handoffs.length > 0;
+		// Derive handoffs from the mode that generated the last response, not the current UI selection.
+		// This ensures handoffs reflect what the response agent offers, regardless of mode picker state.
+		// Fall back to the current mode picker for old sessions where modeInfo was not persisted.
+		const modeInfo = lastItem.model.request?.modeInfo;
```
aeschli, line 1193: `findModeByName` should be used here. That way we have a consistent way of how we deal with name conflicts. Currently: Built-in agent will win.

src/vs/workbench/contrib/chat/common/chatModes.ts
digitarald: Reverted — `findModeByName` now uses `getCustomModes()` instead of raw `_customModeInstances`.

src/vs/workbench/contrib/chat/browser/widget/chatWidget.ts
digitarald: Done — simplified to use `findModeByName` consistently, builtin wins on name conflict.

aeschli (approved):

PR #298934, opened 2026-03-03
Pzhu/feature custom snooze time

### LINKED ISSUE #288770: Support custom inline suggestion snooze durations
I can only disable copilot in intervals of 5 minutes. If I wanted to disable it for e.g. 4 hours, then I have to spam click to get there. We should be able to input our own time we want copilot to remain disabled or be granted an easier way to achieve this.

I love how Zed handled its AI integration in this regard, because they also allow disabling copilot on specific files rather than an entire language.

Pengkun-ZHU:
Fixes #288770

## Feature
Add an option named "Custom" in the snooze quick pick.

## How to test
1. Go to snooze quick pick
2. Select "Custom"
3. Enter a positive number without sign, fraction number is supported
4. Observe the snooze time countdown, or equivalently try with the inline suggestion when it is back online after timeout. N.B. there currently exists an issue that after timeout, you need to conduct a related operation in order to get the inline suggestion back, I'll file the issue sepaxarately and mention here

## Screen recording showing the fix works

https://github.com/user-attachments/assets/6c63c1af-4577-475f-976b-43c2d2c4b37b

https://github.com/user-attachments/assets/b8da66eb-c12a-4c20-a9cc-589cc88eaf2a


src/vs/workbench/contrib/chat/browser/chatStatus/chatStatusDashboard.ts
```diff
@@ -710,7 +710,13 @@ export class ChatStatusDashboard extends DomWidget {
...
 		disposables.add(button.onDidClick(() => {
-			this.inlineCompletionsService.snooze();
+			if (this.inlineCompletionsService.isSnoozing()) {
+				// Extend by 5 min when already snoozing
+				this.inlineCompletionsService.snooze();
+			} else {
+				// Show the duration picker when starting a new snooze
+				this.commandService.executeCommand('editor.action.inlineSuggest.snooze');
+			}
 			update(isEnabled());
```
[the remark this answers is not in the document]
Pengkun-ZHU: Fixed the `async` and `await` part. 
But I beg a differ regarding the `.catch(...)` part, it seems to be a defensive boilerplate for this well-defined usage. I'll leave it for member review to decide.

src/vs/workbench/contrib/chat/browser/chatStatus/chatStatusDashboard.ts
```diff
@@ -746,8 +709,14 @@ export class ChatStatusDashboard extends DomWidget {
 		}
 		updateIntervalTimer();
 
-		disposables.add(button.onDidClick(() => {
-			this.inlineCompletionsService.snooze();
+		disposables.add(button.onDidClick(async () => {
```
benibenj: How about we do not change the behaviour of this button currently and only have the custom snooze for the command?

src/vs/editor/browser/services/inlineCompletionsService.ts
```diff
@@ -212,13 +212,15 @@ export class SnoozeInlineCompletion extends Action2 {
...
-			{ label: '10 minutes', id: '10', value: 600_000 },
-			{ label: '15 minutes', id: '15', value: 900_000 },
-			{ label: '30 minutes', id: '30', value: 1_800_000 },
-			{ label: '60 minutes', id: '60', value: 3_600_000 }
+			{ label: localize('snooze.1minute', "1 minute"), id: '1', value: 60_000 },
+			{ label: localize('snooze.5minutes', "5 minutes"), id: '5', value: 300_000 },
+			{ label: localize('snooze.10minutes', "10 minutes"), id: '10', value: 600_000 },
+			{ label: localize('snooze.15minutes', "15 minutes"), id: '15', value: 900_000 },
+			{ label: localize('snooze.30minutes', "30 minutes"), id: '30', value: 1_800_000 },
+			{ label: localize('snooze.60minutes', "60 minutes"), id: '60', value: 3_600_000 },
```
benibenj, line 221: If he user added a custom value last time, we currently store it, we could show it in the picker so it's easy for the user to use again. It would be nice if we could put it into this list at the correct location and have it preselected.

src/vs/editor/browser/services/inlineCompletionsService.ts
```diff
@@ -227,12 +229,37 @@ export class SnoozeInlineCompletion extends Action2 {
 		});
 
 		if (picked) {
+			if (picked.value === CUSTOM_SENTINEL) {
```
benibenj: Nit: checking `id` feels like it would be a better idea then checking value

benibenj (changes requested):

src/vs/workbench/contrib/chat/browser/chatStatus/chatStatusDashboard.ts
```diff
@@ -708,10 +671,10 @@ export class ChatStatusDashboard extends DomWidget {
 
 			const timeLeftMs = this.inlineCompletionsService.snoozeTimeLeft;
 			if (!isEnabled || timeLeftMs <= 0) {
-				timerDisplay.textContent = localize('completions.snooze5minutesTitle', "Hide suggestions for 5 min");
+				timerDisplay.textContent = localize('completions.snoozeTitle', "Hide suggestions temporarily");
```
benibenj: let's undo this change as we are not changing anything here

src/vs/workbench/contrib/chat/browser/chatStatus/chatStatusDashboard.ts
```diff
@@ -708,10 +671,10 @@ export class ChatStatusDashboard extends DomWidget {
...
 				timerDisplay.title = '';
 				button.label = label;
-				button.setTitle(localize('completions.snooze5minutes', "Hide inline suggestions for 5 min"));
+				button.setTitle(localize('completions.snooze', "Hide inline suggestions"));
```
benibenj: let's undo this change as we are not changing anything here

benibenj (changes requested):
Almost done, just 2 small things to undo

benibenj (approved):

Yoyokrazy (approved):

PR #299003, opened 2026-03-03
feat: support heap profile and snapshot capture for tsserver

deepak1556:


extensions/typescript-language-features/src/tsServer/serverProcess.electron.ts
```diff
@@ -162,6 +162,24 @@ function getExecArgv(kind: TsServerProcessKind, configuration: TypeScriptService
...
+	}
+
+	if (configuration.heapSnapshot > 0) {
+		args.push(`--heapsnapshot-near-heap-limit=${configuration.heapSnapshot}`);
+	}
+
+	if (configuration.heapProfile.enabled) {
+		args.push('--heap-prof');
+		if (configuration.heapProfile.dir) {
+			args.push(`--heap-prof-dir=${configuration.heapProfile.dir}`);
```
mjbvz, line 182: Does this work correctly if the path has spaces or special characters in it?

deepak1556: Yes arguments are preserved, also Node.js does not perform any file path checks on the value so what we have here is good enough.

deepak1556:
However, testing it found the current shutdown sequence is a forceful one with `process.kill` which does not flush heap profiles so I have added a graceful path with timeout to get the profiles.

mjbvz (approved):

PR #299007, opened 2026-03-03
Include remote debug extension host env in remote terminal shell env

### LINKED ISSUE #241078: Environment from extensionHost debug configuration is not inherited by vscode.ShellExecution tasks

eliericha:
This fixes #241078.

The environment specified in a `launch.json` `env` or `envFile` property is passed to the extension host and ultimately into `environmentService.debugExtensionHost.env`.

In the case of the browser service there was a missing parsing of the `extensionEnvironment` parameter in the payload.

Then `createProcess` includes this environment into the call arguments to inject it into spawned terminals.

### DIFF
--- a/src/vs/workbench/contrib/terminal/common/remote/remoteTerminalChannel.ts
+++ b/src/vs/workbench/contrib/terminal/common/remote/remoteTerminalChannel.ts
@@ -159,7 +159,7 @@ export class RemoteTerminalChannelClient implements IPtyHostController {
 			 * If the extension host was spawned via a launch configuration,
 			 * include the environment provided by that launch configuration.
 			 */
-			...this._environmentService.debugExtensionHost.env, ...resolverResult.options?.extensionHostEnv
+			...(this._environmentService.debugExtensionHost.env ?? {}), ...resolverResult.options?.extensionHostEnv
 		};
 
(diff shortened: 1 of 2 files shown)


meganrogge (approved):
Thank you!

lramos15 (approved):

PR #299013, opened 2026-03-03
Browser: context menus

### LINKED ISSUE #292908: Browser: Context menu
Adding a right-click context menu with the usual items.

One additional item being Add Element to Chat, as requested in https://github.com/microsoft/vscode/issues/291056#issuecomment-3808474666
> I think a context menu option would be natural but I noticed that there isn't one (actually, is it possible to implement a context menu which has typical browser options and some of our commands? That would be nice)

kycutler:
Closes #292908

### Editing
image
<br/>

### Links / images
image
<br/>

### General
image


src/vs/platform/browserView/electron-main/browserViewMainService.ts
```diff
@@ -366,4 +375,106 @@ export class BrowserViewMainService extends Disposable implements IBrowserViewMa
...
+				accelerator: this._keybindings[BrowserViewCommandId.Reload],
+				click: () => webContents.reload()
+			}));
+		}
+
+		menu.append(new MenuItem({ type: 'separator' }));
+		menu.append(new MenuItem({
+			label: localize('browser.contextMenu.inspect', 'Inspect'),
+			click: () => webContents.inspectElement(params.x, params.y)
+		}));
```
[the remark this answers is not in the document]
kycutler: This needs some discussion I think -- it can be tricky to select the right element due to padding etc., so not sure we should do this with just one click.

src/vs/platform/browserView/electron-main/browserView.ts
```diff
@@ -574,6 +579,13 @@ export class BrowserView extends Disposable implements ICDPTarget {
 		return this._view;
 	}
 
+	/**
+	 * Get the parent window for this view, if any.
```
jruales: nit: should this be "topmost window" or something like that instead of "parent window" to match the wording in JS `window.top` (https://developer.mozilla.org/en-US/docs/Web/API/Window/top).?

Or is this referring to something else?

jruales:
When I right click on a link on a Mac, I see a Services option with many options that seem irrelevant:

image

EDIT: actually, I'm seeing something similar in other apps like Notes, so it seems like some system-level issue more than something specific to our use case.

Note: when I click on the background, I don't see any of those extra options:
image

src/vs/platform/browserView/electron-main/browserViewMainService.ts
```diff
@@ -366,4 +375,115 @@ export class BrowserViewMainService extends Disposable implements IBrowserViewMa
...
+				label: localize('browser.contextMenu.copyImageUrl', 'Copy Image URL'),
+				click: () => { clipboard.writeText(params.srcURL!); }
+			}));
+		}
+
+		if (params.isEditable) {
+			menu.append(new MenuItem({ role: 'cut', enabled: params.editFlags.canCut }));
+			menu.append(new MenuItem({ role: 'copy', enabled: params.editFlags.canCopy }));
+			menu.append(new MenuItem({ role: 'paste', enabled: params.editFlags.canPaste }));
+			menu.append(new MenuItem({ role: 'pasteAndMatchStyle', enabled: params.editFlags.canPaste }));
```
jruales, line 478: Let's also add selectAll here

src/vs/workbench/contrib/browserView/electron-browser/browserViewWorkbenchService.ts
```diff
@@ -20,10 +29,27 @@ export class BrowserViewWorkbenchService implements IBrowserViewWorkbenchService
...
 	) {
+		super();
 		const channel = mainProcessService.getChannel(ipcBrowserViewChannelName);
 		this._browserViewService = ProxyChannel.toService<IBrowserViewService>(channel);
+
+		this.sendKeybindings();
+		this._register(this.keybindingService.onDidUpdateKeybindings(() => this.sendKeybindings()));
+	}
+
+	private sendKeybindings(): void {
```
jruales: Nit: put this private method after all public methods

src/vs/platform/browserView/electron-main/browserViewMainService.ts
```diff
@@ -366,4 +375,115 @@ export class BrowserViewMainService extends Disposable implements IBrowserViewMa
...
+			window: win,
+			x: viewBounds.x + params.x,
+			y: viewBounds.y + params.y,
+			frame: params.frame ?? undefined,
+			sourceType: params.menuSourceType
+		});
+	}
+}
+
+function escapeHtmlText(str: string): string {
```
jruales: A similar utility function called `htmlAttributeEncodeValue` is already defined in src/vs/base/common/strings.ts

src/vs/platform/browserView/electron-main/browserViewMainService.ts
```diff
@@ -85,6 +92,21 @@ export class BrowserViewMainService extends Disposable implements IBrowserViewMa
 		return view;
 	}
 
+	private async openNew(url: string, session: BrowserSession | undefined, options: ITextEditorOptions, source: IntegratedBrowserOpenSource): Promise<ICDPTarget> {
```
jruales: Please move this private method either to the bottom after all public methods, or at least closer to where it's being called

jruales (dismissed):

jruales (dismissed):

src/vs/platform/browserView/electron-main/browserViewMainService.ts
kycutler: Moved this and the above down to the bottom

src/vs/platform/browserView/electron-main/browserViewMainService.ts
kycutler: Good find!

src/vs/platform/browserView/electron-main/browserView.ts
kycutler: Turns out we actually needed another method here to get the root VS Code window, to handle cases where the browser view is in an auxiliary window. So I split this into `getElectronWindow` and `getTopCodeWindow` with docstrings that should help clarify things.

kycutler:
> When I right click on a link on a Mac, I see a Services option with many options that seem irrelevant:

We actually do have some control over this -- disabled for now by not passing the "frame" parameter to Electron when opening the menu -- per Electron docs: `Provide the relevant frame if you want certain OS-level features such as Writing Tools on macOS to function correctly`.

> I'm noticing that right clicking on text automatically selects the word under the cursor, which is unexpected:

I suspect this is also coming from the OS (though it doesn't happen in e.g. Edge). Will leave as-is for now.

jruales (approved):

Issue #297452, opened 2026-02-24
Variables are not interpreted in sandbox config


connor4312:
Testing #297115

Have a config like this:

```
{
	"servers": {
		"filesystem": {
			"command": "npx",
			"sandboxEnabled": true,
			"args": [
				"-y",
				"@modelcontextprotocol/server-filesystem",
			],
			"type": "stdio"
		}
	},
	"sandbox": {
		"filesystem": {
			"allowWrite": [
				"${workspaceFolder}/src"
			]
		}
	},
	"inputs": []
}
```

Get an error:

```
2026-02-24 11:32:03.372 [warning] [server stderr] [SandboxDebug] [Sandbox Linux] Processing write path: ${workspaceFolder}/src -> /home/connor/Github/cockatiel/${workspaceFolder}/src
2026-02-24 11:32:03.372 [warning] [server stderr] [SandboxDebug] [Sandbox Linux] Skipping non-existent write path: /home/connor/Github/cockatiel/${workspaceFolder}/src
```

PR #299124, opened 2026-03-04
[MCP-Sandboxing]Changes for variable substitution and code refactor

dileepyavan:
fixes(#297452),(#297458)

This PR includes

- sandbox property is removed from MCPServerDefinition and added to MCPServerLaunch.
- Currently launch property has changes for variable substitution which is reused for sandbox config.
- Refactored code that updates the sandbox suggestions.

### DIFF
--- a/src/vs/workbench/contrib/mcp/common/discovery/installedMcpServersDiscovery.ts
+++ b/src/vs/workbench/contrib/mcp/common/discovery/installedMcpServersDiscovery.ts
@@ -96,7 +96,7 @@ export class InstalledMcpServersDiscovery extends Disposable implements IMcpDisc
 					env: config.env || {},
 					envFile: config.envFile,
 					cwd: config.cwd,
-					sandbox: config.sandbox ?? server.rootSandbox
+					sandbox: server.rootSandbox
 				};
 
 				definitions[1].push({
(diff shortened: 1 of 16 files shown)


src/vs/workbench/contrib/mcp/common/discovery/installedMcpServersDiscovery.ts
```diff
@@ -96,14 +96,14 @@ export class InstalledMcpServersDiscovery extends Disposable implements IMcpDisc
 					env: config.env || {},
 					envFile: config.envFile,
 					cwd: config.cwd,
+					sandbox: config.sandbox ?? server.rootSandbox
```
connor4312: When's the case when `config.sandbox` would be set and `rootSandbox` would not be or vise versa?

connor4312: I think you can actually delete `IMcpStdioServerConfiguration.sandbox` now?

dileepyavan: I dont think its needed anymore. Updated and validated.

connor4312 (approved):

PR #299128, opened 2026-03-04
feat: add optional id field to ChatRequestModeInstructions interface

DonJayamanne:


aeschli:
@copilot Can we change it so that instead of surfacing the id, we surface the uri? The uri will be undefined for the built-in modes

aeschli (dismissed):

bpasero (approved):

PR #299145, opened 2026-03-04
mcp: replace repeated variable placeholders in config values

bharatvansh:
## Summary
Fixes placeholder interpolation for MCP manifest values when the same variable appears multiple times in a single string.

Previously, interpolation used single-replacement behavior, so only the first `{variable}` occurrence was converted to `${input:variable}`. This left subsequent occurrences unresolved.

## Changes
- Add `replaceInputVariableReference(value, variableId)` helper in `mcpManagementService.ts`.
- Use the helper in all variable interpolation paths:
  - key/value inputs (env + remote headers)
  - positional arguments
  - named arguments
- Add regression tests for repeated placeholders in:
  - environment values
  - argument values

## Example
`Bearer {api_token}:{api_token}` now becomes:
`Bearer ${input:api_token}:${input:api_token}`

## Validation
- `npm run transpile-client` ✅
- `scripts/test.bat --grep "McpManagementService - getMcpServerConfigurationFromManifest"` ✅ (37 passing)

### DIFF
--- a/src/vs/platform/mcp/common/mcpManagementService.ts
+++ b/src/vs/platform/mcp/common/mcpManagementService.ts
@@ -205,6 +205,10 @@ export abstract class AbstractCommonMcpManagementService extends Disposable impl
 		return variables;
 	}
 
+	private replaceInputVariableReference(value: string, variableId: string): string {
+		return value.split(`{${variableId}}`).join(`\${input:${variableId}}`);
+	}
+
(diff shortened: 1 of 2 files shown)


src/vs/platform/mcp/common/mcpManagementService.ts
```diff
@@ -217,7 +221,7 @@ export abstract class AbstractCommonMcpManagementService extends Disposable impl
 			// If explicit variables exist, use them regardless of value
 			if (inputVariables.length) {
 				for (const variable of inputVariables) {
-					value = value.replace(`{${variable.id}}`, `\${input:${variable.id}}`);
+					value = this.replaceInputVariableReference(value, variable.id);
```
connor4312, line 224: We can just use `value.replaceAll`

PR #299161, opened 2026-03-04
Browser Zoom

### LINKED ISSUE #300849: Integrated Browser: Zoom
Make browser zoom independent from VS Code zoom.

The browser zoom feature should be accessible via the browser menu (https://github.com/microsoft/vscode/issues/289433) as well as via keyboard shortcut (https://github.com/microsoft/vscode/issues/295024).

jruales:
* Closes https://github.com/microsoft/vscode/issues/300849

This PR adds **independent browser zoom** to VS Code's built-in browser view (the embedded Chromium panel used for Simple Browser / web previews). Before this change, pressing Cmd+Plus/Minus while a browser view was focused would zoom *VS Code itself*. Now, those shortcuts zoom only the **web content inside the browser view** — just like Chrome or Edge.

The zoom uses a fixed set of **discrete zoom levels** (25%, 33%, 50%, … 100%, … 500%) matching Chrome/Edge steps, and remembers zoom per-host so different websites keep their own zoom.

Here's a map of the changed files and their roles:

| File | Layer | Role |
|------|-------|------|
| platform/.../common/browserView.ts | Shared types | Defines the zoom level table & IPC contract |
| platform/.../electron-main/browserView.ts | Main process | Calls Electron's `setZoomFactor` on the actual webContents |
| platform/.../electron-main/browserViewMainService.ts | Main process | Routes IPC calls to the right `BrowserView` instance |
| workbench/.../common/browserZoomService.ts | Renderer service | **NEW** — manages per-host & default zoom levels |
| workbench/.../common/browserView.ts | Renderer model | `BrowserViewModel` — has `zoomIn()`, `zoomOut()`, `resetZoom()` |
| workbench/.../electron-browser/browserEditor.ts | Renderer UI | Wires zoom to context keys & the editor pane |
| workbench/.../electron-browser/browserViewActions.ts | Renderer actions | Registers the three zoom commands & keybindings |
| workbench/.../electron-browser/browserView.contribution.ts | Renderer bootstrap | Registers `BrowserZoomService`, window zoom sync, and settings | 

```mermaid
graph TB
    subgraph Renderer["Renderer Process"]
        Actions["ZoomIn / ZoomOut / ResetZoom<br/>(browserViewActions.ts)"]
        Editor["BrowserEditor<br/>(browserEditor.ts)"]
        VM["BrowserViewModel<br/>(common/browserView.ts)"]
        ZS["BrowserZoomService<br/>(browserZoomService.ts)"]
        Contrib["browserView.contribution.ts<br/>(settings + window zoom sync)"]
    end
    subgraph Main["Main Process"]
        MainSvc["BrowserViewMainService<br/>(browserViewMainService.ts)"]
        BV["BrowserView<br/>(electron-main/browserView.ts)"]
    end
    subgraph Shared["Shared (common)"]
        Types["browserZoomFactors[]<br/>+ IPC contract<br/>(platform/.../common/browserView.ts)"]
    end

    Actions --> Editor
    Editor --> VM
    VM --> ZS
    VM -->|IPC| MainSvc
    MainSvc --> BV
    BV -->|setZoomFactor| Electron["Electron webContents"]
    Contrib -.->|registers| ZS
    Types -.->|used by| VM
    Types -.->|used by| BV
```

The key idea: the user presses a shortcut → it flows through **Actions → Editor → ViewModel → (IPC) → Main Process → Electron**, with a separate **BrowserZoomService** managing which zoom level to use for each host.

---

Beyond the basic "zoom in / zoom out / reset" trio, the PR packs several behaviors a user would care about:

| # | Feature | Where it lives |
|---|---------|---------------|
| 1 | **Discrete Chrome/Edge zoom steps** | 25 → 33 → 50 → 67 → 75 → 80 → 90 → **100** → 110 → 125 → 150 → 175 → 200 → 250 → 300 → 400 → 500% |
| 2 | **Per-host zoom memory** | Zoom `github.com` to 125% — it stays at 125% next time you open it. Stored per-host in `IStorageService` |
| 3 | **Persistent vs. ephemeral sessions** | Normal browser views save zoom to disk. Ephemeral views (private-browsing-like) keep zoom in memory only — gone on restart |
| 4 | **"Match Window" default** | New setting `workbench.browser.zoom.pageZoom` can be set to "Match Window" — the browser view auto-matches VS Code's own window zoom, using log-scale nearest-neighbor matching |
| 5 | **Configurable fixed default** | Or pick any specific level (e.g. "125%") as the default for all new pages |
| 6 | **Zoom survives navigation** | Chromium resets zoom on every cross-document navigation. The main-process `BrowserView` re-applies the stored zoom on every `did-navigate` event |
| 7 | **Cross-origin re-application** | The ViewModel also re-applies zoom on navigation because an origin change (e.g. http→https) resets Chromium's zoom even when the host stays the same |
| 8 | **Cascade: per-host overrides the default** | If you haven't zoomed a specific host, it gets the default. If you `resetZoom()`, the per-host override is deleted so the host tracks the default again |
| 9 | **Propagation across views** | Zooming `github.com` in one tab propagates to all other open tabs showing `github.com` (via `BrowserZoomService.onDidChangeZoom`) |
| 10 | **Min/max clamping + context keys** | At 25% (min), the Zoom Out button disappears from the toolbar. At 500% (max), Zoom In disappears. Done via `CONTEXT_BROWSER_CAN_ZOOM_IN` / OUT |
| 11 | **Keybindings scoped to browser focus** | `Cmd+=`, `Cmd+-`, `Cmd+Numpad0` only fire when the browser editor is focused — they don't hijack VS Code's own zoom shortcuts elsewhere |

---

### Future work (not covered in this PR)
- Show zoom level number


src/vs/workbench/contrib/browserView/electron-browser/browserViewActions.ts
```diff
@@ -483,6 +484,110 @@ class OpenBrowserSettingsAction extends Action2 {
...
+	when: CONTEXT_BROWSER_FOCUSED,
+	weight: KeybindingWeight.WorkbenchContrib + 75,
+	primary: KeyMod.CtrlCmd | KeyCode.Minus,
+	secondary: [KeyMod.CtrlCmd | KeyCode.NumpadSubtract],
+};
+const resetZoomKeybindingRule = {
+	id: 'workbench.action.browser.resetZoom',
+	when: CONTEXT_BROWSER_FOCUSED,
+	weight: KeybindingWeight.WorkbenchContrib + 75,
+	primary: KeyMod.CtrlCmd | KeyCode.Numpad0,
```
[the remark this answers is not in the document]
jruales: This is the same keybinding as with `workbench.action.zoomReset`. `Ctrl/Cmd+0` is reserved for taking focus to sidebar

src/vs/platform/browserView/electron-main/browserView.ts
```diff
@@ -393,6 +393,16 @@ export class BrowserView extends Disposable implements ICDPTarget {
 		});
 	}
 
+	/**
+	 * Set the browser zoom index as an index into the discrete zoom levels array.
+	 * @param zoomIndex Index into browserZoomPercentages
+	 */
+	setBrowserZoomIndex(zoomIndex: number): void {
+		this._browserZoomIndex = Math.max(0, Math.min(zoomIndex, browserZoomPercentages.length - 1));
+		const browserZoomFactor = browserZoomPercentages[this._browserZoomIndex] / 100;
```
kycutler: Rather than zoom percentages can we just have the zoom factors, like 1/3? Then 33% and 67% can be perfect thirds (well, practically anyway)

src/vs/workbench/contrib/browserView/electron-browser/browserViewActions.ts
```diff
@@ -483,6 +484,110 @@ class OpenBrowserSettingsAction extends Action2 {
...
 }
 
+// Zoom actions
+
+// Zoom keybindings registered separately from the Action2 precondition so they
+// always fire when the browser is focused, preventing VS Code's own zoom from
+// triggering even when the browser is at its min/max zoom level.
+const zoomInKeybindingRule = {
+	id: 'workbench.action.browser.zoomIn',
+	when: CONTEXT_BROWSER_FOCUSED,
```
kycutler: I don't see any difference if I move these rules into the actions? Even at min / max zoom it doesn't bubble up

src/vs/platform/browserView/electron-main/browserView.ts
```diff
@@ -45,6 +45,7 @@ export class BrowserView extends Disposable implements ICDPTarget {
 	private _lastFavicon: string | undefined = undefined;
 	private _lastError: IBrowserViewLoadError | undefined = undefined;
 	private _lastUserGestureTimestamp: number = -Infinity;
+	private _browserZoomIndex: number = browserZoomDefaultIndex;
```
kycutler, line 49: Can this default to the window zoom level? Thinking it could be annoying for users of high zoom to have to zoom in on every page they open

src/vs/workbench/contrib/browserView/electron-browser/browserViewActions.ts
jruales: I had it under preconditions instead of when, which was causing the issue. Fixed it now

pwang347 (approved):

PR #299164, opened 2026-03-04
feat(configuration): support extends in folder settings files

### LINKED ISSUE #15909: Add ability to extend from other settings files
Inspired by TSLint's ability to [extend](https://palantir.github.io/tslint/2016/03/31/sharable-configurations-rules.html) configuration files, it would be nice if `.vscode/settings.json` could behave the same way.

So if I have some global settings set up:

`~/example-repo/.vscode/my-company-settings.json`:
```json
{
  "editor.tabSize": 2,
  "editor.insertSpaces": true,
  "editor.formatOnSave": false
}
```

I can use them in another file, without having to duplicate the settings:

`~/example-repo/my-project/.vscode/settings.json`:
```json
{
  "extends": "../../.vscode/my-company-settings.json",
  "editor.formatOnSave": true,
  "editor.fontLigatures": true
}
```

And the computed settings for `~/example-repo/my-project/.vscode/settings.json` would be:

```json
{
  "editor.tabSize": 2,
  "editor.insertSpaces": true,
  "editor.formatOnSave": true,
  "editor.fontLigatures": true
}
```

Scenario:

Multi-root workspace doesn't solve this for our use case. We have a bunch of npm modules each in their own git repository. We have a package which contains our shared tsconfig.json and tslint.json settings that all the other packages include with extends. We don't use a multi-root workspace since the idea is that people can clone the specific package(s) they need to work on. Every repository contains the exact same .vscode directory which is essentially copy&pasted all over the place. Maintaining the .vscode settings for the projects is "not pretty" compared to the tsconfig.json and tslint.json which only require the settings package to be updated with for example yarn upgrade.

scarf005:
## Summary
- add support for `extends` in folder `.vscode/settings.json` by resolving and merging inherited settings before parsing
- watch inherited settings files and trigger configuration reload when inherited files change
- add coverage for override precedence and inherited-file change propagation

Fixes microsoft/vscode#15909.

## Testing
- `npm run compile`
- `npm run test-node -- --run "src/vs/workbench/services/configuration/test/common/configurationModels.test.ts" --grep "FolderSettingsModelParser"`
- attempted targeted electron/browser tests for `configurationService.test.ts`; blocked in this container due missing X server and missing Playwright host libraries

### DIFF
--- a/src/vs/workbench/services/configuration/browser/configuration.ts
+++ b/src/vs/workbench/services/configuration/browser/configuration.ts
@@ -9,6 +9,8 @@ import * as errors from '../../../../base/common/errors.js';
 import { Disposable, IDisposable, dispose, toDisposable, MutableDisposable, combinedDisposable, DisposableStore } from '../../../../base/common/lifecycle.js';
 import { RunOnceScheduler } from '../../../../base/common/async.js';
 import { FileChangeType, FileChangesEvent, IFileService, whenProviderRegistered, FileOperationError, FileOperationResult, FileOperation, FileOperationEvent } from '../../../../platform/files/common/files.js';
+import * as json from '../../../../base/common/json.js';
+import { merge } from '../../../../platform/configuration/common/configuration.js';
 import { ConfigurationModel, ConfigurationModelParser, ConfigurationParseOptions, UserSettings } from '../../../../platform/configuration/common/configurationModels.js';
 import { WorkspaceConfigurationModelParser, StandaloneConfigurationModelParser } from '../common/configurationModels.js';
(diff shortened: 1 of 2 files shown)


nanbiz:
```ts
if (!this.uriIdentityService.extUri.isEqualOrParent(candidate, baseDir)) {
    this.logService.trace(`Ignoring extends reference '${extendsRef}' from '${settingsResource.toString()}' because it resolves outside the configuration folder.`);
    return undefined;
}
```

If I understand correctly here, the current approach would not allow what most users including the original issue author in #15909 want.
One example that might be very common are monorepos like this:

```
<root>/
├── .vscode/
│   └── settings.json         <- parent (would be extended FROM)
└── project_A/
    └── .vscode/
        └── settings.json     <- child, contains "extends": "../../.vscode/settings.json"
```

And with the current approach that would not work as the root is outside of project_A.
If path traversal is a problem, an alternative might be to 
a) allow settings in .vscode folders within parent directories
b) allow files in the current git repository if one is opened

But leaving it like it currently is, I think the feature wouldn't be useful to many users.

PR #299256, opened 2026-03-04
Debug Panel: oTel data source support and Import/export

### LINKED ISSUE #301245: Add export/import to debug panel
Add export/import to debug panel

vijayupadya:
Fixes https://github.com/microsoft/vscode/issues/301245

### Proposed API Changes (`vscode.proposed.chatDebug.d.ts`)

- **`provideChatDebugLogExport`** — Export a session's debug log as serialized bytes (e.g., OTLP JSON). Core handles the save dialog.
- **`resolveChatDebugLogImport`** — Import a previously exported debug log. Core handles the open dialog and returns a session URI for the imported data.
- **`resolveChatDebugLogCoreEvent`** — Receives VS Code core-originated events (prompt discovery, skill loading) so extensions can include them in their data pipeline (e.g., OTel export).

### Export/Import Toolbar Buttons

- **Export** button (download icon) in the debug panel editor title bar — saves the active session's debug log via `provideChatDebugLogExport`
- **Import** button (upload icon) — opens a file dialog, reads the file, calls `resolveChatDebugLogImport`, and navigates to the imported session

### Core → Extension Event Forwarding

- Historical core events flushed to the extension when `provideChatDebugLog` is called, so they're captured for export even when the panel is opened after events were logged
- Live core events forwarded in real-time via `onDidAddEvent` listener
- `WeakSet`-based dedup tracking prevents re-sending events across multiple `invokeProviders` calls
- Provider callback errors handled via `onUnexpectedExternalError`

### Service Layer

- **`isCoreEvent(event)`** — distinguishes core vs provider events (used by mainThread to filter forwarding)
- **`exportLog` / `importLog`** — delegate to registered providers
- Imported session URIs tracked so they pass the `invokeProviders` guard (which blocks CLI/Claude sessions pending investigation)

### Debug Panel UX Fixes

- **Tab switch / reopen**: Pipeline stays alive while editor is hidden; `showView` reloads events on re-show without duplicates
- **Event ordering**: `refreshList()` sorts events by timestamp so core and provider events interleave correctly regardless of arrival order
- **Home view sessions always clickable**: Removed shimmer/disabled behavior for UUID titles — shows "New Chat" for local sessions, "Imported: {id}" for imported sessions
- **Auto-navigate on new session**: When the debug panel is open on the home view and a new chat starts, auto-navigates to it
- **Flow chart labels**: User/Agent response nodes now show descriptive titles ("User Message", "Agent Response") with message summary as sublabel

### Files Changed

| File | Change |
|------|--------|
| `vscode.proposed.chatDebug.d.ts` | Added `provideChatDebugLogExport`, `resolveChatDebugLogImport`, `resolveChatDebugLogCoreEvent` |
| `extHost.protocol.ts` | Added `$exportChatDebugLog`, `$importChatDebugLog`, `$handleCoreDebugEvent` to protocol |
| `extHostChatDebug.ts` | Implemented export/import/core-event handlers with `_deserializeEvent` |
| `mainThreadChatDebug.ts` | Wired export/import, core event forwarding with dedup tracking, `_serializeEvent` |
| `chatDebugService.ts` | Added `exportLog`, `importLog`, `isCoreEvent` to interface |
| `chatDebugServiceImpl.ts` | Implemented `exportLog`, `importLog`, `isCoreEvent`, imported session tracking |
| `chatOpenAgentDebugPanelAction.ts` | Export/Import toolbar actions |
| `chatDebugEditor.ts` | Simplified `setEditorVisible`, auto-navigate on new session |
| `chatDebugHomeView.ts` | Always-clickable sessions, "New Chat" / "Imported: {id}" fallback titles |
| `chatDebugLogsView.ts` | Sort events by timestamp in `refreshList` |
| `chatDebugFlowGraph.ts` | Improved User Message / Agent Response node labels |


src/vs/workbench/contrib/chat/common/chatDebugServiceImpl.ts
```diff
@@ -122,10 +122,6 @@ export class ChatDebugServiceImpl extends Disposable implements IChatDebugServic
 	}
 
 	async invokeProviders(sessionResource: URI): Promise<void> {
-		if (!LocalChatSessionUri.isLocalSession(sessionResource)) {
```
pwang347, line 138: this is needed so that we don't provide a broken experience on copilot CLI chats and claude (we need to investigate why it's broken still)

pwang347: was it causing some issues?

src/vs/workbench/api/browser/mainThreadChatDebug.ts
```diff
@@ -31,12 +32,44 @@ export class MainThreadChatDebug extends Disposable implements MainThreadChatDeb
...
+
+				// Send all existing core events for this session to the
+				// extension so they are captured even when the panel is
+				// opened after events have already been logged.
+				for (const event of this._chatDebugService.getEvents(sessionResource)) {
+					if (this._chatDebugService.isCoreEvent(event)) {
+						this._proxy.$handleCoreDebugEvent(handle, this._serializeEvent(event));
+					}
+				}
+
```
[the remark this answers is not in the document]
vijayupadya: fixed

src/vs/workbench/api/common/extHostChatDebug.ts
```diff
@@ -291,6 +292,107 @@ export class ExtHostChatDebug extends Disposable implements ExtHostChatDebugShap
...
+
+	private _deserializeEvent(dto: IChatDebugEventDto): vscode.ChatDebugEvent | undefined {
+		const created = new Date(dto.created);
+		switch (dto.kind) {
+			case 'toolCall': {
+				const evt = new ChatDebugToolCallEvent(dto.toolName, created);
+				evt.id = dto.id;
+				evt.parentEventId = dto.parentEventId;
+				evt.toolCallId = dto.toolCallId;
+				evt.input = dto.input;
```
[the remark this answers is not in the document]
vijayupadya: updated

src/vs/workbench/api/common/extHostChatDebug.ts
```diff
@@ -291,6 +292,107 @@ export class ExtHostChatDebug extends Disposable implements ExtHostChatDebugShap
...
 
+	$handleCoreDebugEvent(_handle: number, dto: IChatDebugEventDto): void {
+		if (!this._provider?.resolveChatDebugLogCoreEvent) {
+			return;
+		}
+		const event = this._deserializeEvent(dto);
+		if (event) {
+			this._provider.resolveChatDebugLogCoreEvent(event, CancellationToken.None);
+		}
+	}
```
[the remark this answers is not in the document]
vijayupadya: updated

src/vs/workbench/contrib/chat/browser/chatDebug/chatDebugEditor.ts
```diff
@@ -178,6 +178,13 @@ export class ChatDebugEditor extends EditorPane {
 				this.homeView?.render();
 			}
 
+			// Auto-navigate to the new session when the debug panel is
+			// already open on the home view.  This avoids the user having to
+			// wait for the title to resolve and manually clicking the session.
+			if (this.viewState === ViewState.Home) {
```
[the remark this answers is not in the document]
vijayupadya: updated

src/vs/workbench/contrib/chat/browser/actions/chatOpenAgentDebugPanelAction.ts
```diff
@@ -92,4 +99,100 @@ export function registerChatOpenAgentDebugPanelAction() {
...
+				notificationService.notify({ severity: Severity.Warning, message: localize('chatDebugLog.exportFailed', "Export is not supported by the current provider.") });
+				return;
+			}
+
+			const defaultUri = joinPath(await fileDialogService.defaultFilePath(), defaultDebugLogFileName);
+			const outputPath = await fileDialogService.showSaveDialog({ defaultUri, filters: debugLogFilters });
+			if (!outputPath) {
+				return;
+			}
+
```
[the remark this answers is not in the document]
vijayupadya: updated

src/vs/workbench/contrib/chat/common/chatDebugServiceImpl.ts
vijayupadya: This change was needed for import scenario which has local uri. Updated with a workaround.

src/vscode-dts/vscode.proposed.chatDebug.d.ts
```diff
@@ -642,6 +642,45 @@ declare module 'vscode' {
...
+
+		/**
+		 * Called when core emits a debug event for the active session.
+		 * Core events include prompt discovery, skill loading, and other
+		 * internal diagnostics that the extension may want to include in
+		 * its data pipeline (e.g., OTel export).
+		 *
+		 * @param event The debug event emitted by core.
+		 */
+		resolveChatDebugLogCoreEvent?(event: ChatDebugEvent, token: CancellationToken): void;
```
pwang347: this feels like it should be on a new interface, not the existing one here

pwang347: maybe even something like `chat.onDidEmitDebugLog`, it starts to get confusing though with our current contract with providers not always streaming but only when invoked via provide

src/vscode-dts/vscode.proposed.chatDebug.d.ts
```diff
@@ -642,6 +642,45 @@ declare module 'vscode' {
 			eventId: string,
 			token: CancellationToken
 		): ProviderResult<ChatDebugResolvedEventContent>;
+
+		/**
+		 * Export the debug log for a chat session as a serialized byte array.
```
pwang347, line 647: any reason why we need to serialize here? are there concerns with file size?

src/vscode-dts/vscode.proposed.chatDebug.d.ts
```diff
@@ -642,6 +642,45 @@ declare module 'vscode' {
...
+		 * The extension controls the format (e.g., OTLP JSON with Copilot extensions).
+		 * Provides the save dialog and writes the returned bytes to disk.
+		 *
+		 * @param sessionResource The resource URI of the chat session to export.
+		 * @param token A cancellation token.
+		 * @returns The serialized debug log data, or undefined if export is not available.
+		 */
+		provideChatDebugLogExport?(
```
pwang347, line 656: I think we could also probably move the import/export to a different interface, if you're contributing import/export you might not be contributing events

src/vscode-dts/vscode.proposed.chatDebug.d.ts
```diff
@@ -642,6 +642,45 @@ declare module 'vscode' {
...
+			sessionResource: Uri,
```
pwang347, line 657: should the export be more stateless, and we just pass the list of events? that way we don't need an extra API for getting events from core

src/vscode-dts/vscode.proposed.chatDebug.d.ts
```diff
@@ -642,6 +642,45 @@ declare module 'vscode' {
...
+		 * opened in the debug panel via {@link provideChatDebugLog}.
+		 *
+		 * @param data The serialized debug log data (as returned by {@link provideChatDebugLogExport}).
+		 * @param token A cancellation token.
+		 * @returns A session resource URI for the imported session, or undefined if import failed.
+		 */
+		resolveChatDebugLogImport?(
+			data: Uint8Array,
+			token: CancellationToken
+		): ProviderResult<Uri>;
```
pwang347: same with this one, should this just return a list of events rather than a URI? seems unclear what needs to be done from extension side

src/vscode-dts/vscode.proposed.chatDebug.d.ts
pwang347: I think we also need to figure out who is the source of truth, previous design was that core was source of truth but seems like this is shifting to chat ext. now? if so, then we should try to unify more things, e.g. don't add events from core directly to the debug log service, but always ask for the full list from chat ext.

src/vscode-dts/vscode.proposed.chatDebug.d.ts
vijayupadya: Removed this API. Now core events are passed only for export.

src/vscode-dts/vscode.proposed.chatDebug.d.ts
vijayupadya: Removed resolveChatDebugLogCoreEvent and changed provideChatDebugLogExport to accept coreEvents[] directly. Core passes its events at export time, no live forwarding needed.

src/vscode-dts/vscode.proposed.chatDebug.d.ts
vijayupadya: Keeping them together because:

- Import feeds back into the provider: resolveChatDebugLogImport returns a URI → core opens it via provideChatDebugLog → the same provider serves the imported events. If on a separate interface, you'd need cross-interface coordination for this flow.

- Export reads provider state: provideChatDebugLogExport reads from _allSpans (the same storage that provideChatDebugLog writes to). Splitting means either duplicating state or sharing it through a service.
 
- Resolve needs imported data: resolveChatDebugLogEvent searches _importedSessions for detail content. A separate import interface would need to share this lookup with the event provider.

src/vscode-dts/vscode.proposed.chatDebug.d.ts
vijayupadya: Uint8Array is for these reasons:

- Format agnostic: The extension controls the serialization format. If we returned a structured object, core would need to understand the format, and the API would need to change whenever the format evolves.

- Third-party tool compatibility: The exported bytes are standard OTLP JSON that can be imported into Jaeger, Grafana Tempo, or any OTel-compatible tool. Those tools expect a specific binary/JSON format.
 
- Core stays a dumb pipe: Core just does save dialog → write bytes to disk and open dialog → read bytes → pass to extension. No parsing, no format knowledge, no versioning concerns on the core side.

src/vscode-dts/vscode.proposed.chatDebug.d.ts
vijayupadya: This is the sequence in which APIs are called, each with distinct purpose. Extension is source of truth

- resolveChatDebugLogImport(bytes) → "Parse this file and tell me where to find it" → returns URI
- provideChatDebugLog(uri) → "Give me the display events for this session" → returns event list + streams
- resolveChatDebugLogEvent(id) → "User clicked this event, give me the detail" → returns rich content

pwang347 (dismissed):

pwang347 (approved):

PR #299320, opened 2026-03-04
[MCP_Sandboxing]: Add ripgrep path to env variables 

dileepyavan:
This PR adds  the ripgrep path from node modules to the env variables.


connor4312:
I'm curious on the use case here -- do we see MCP servers trying to spawn `ripgrep` directly?

src/vs/workbench/contrib/mcp/common/mcpSandboxService.ts
```diff
@@ -271,10 +272,13 @@ export class McpSandboxService extends Disposable implements IMcpSandboxService
...
-	private _getSandboxEnvVariables(tempDir: URI | undefined, remoteAuthority?: string): Record<string, string | null> | undefined {
-		let env: Record<string, string | null> = {};
+	private _getSandboxEnvVariables(baseEnv: McpServerTransportStdio['env'], tempDir: URI | undefined, rgPath: string | undefined, remoteAuthority?: string): McpServerTransportStdio['env'] {
+		let env: McpServerTransportStdio['env'] = { ...baseEnv };
 		if (tempDir) {
-			env = { TMPDIR: tempDir.path, SRT_DEBUG: 'true' };
+			env = { ...env, TMPDIR: tempDir.path, SRT_DEBUG: 'true' };
+		}
+		if (rgPath) {
+			env = { ...env, PATH: `${dirname(rgPath)}` };
```
connor4312: It looks like we overwrite the PATH if the user has defined it manually in their environment; we should avoid doing this

dileepyavan: updated.

connor4312 (approved):

PR #299364, opened 2026-03-05
Add chatMode and sessionType to interactiveSessionProviderInvoked telemetry

andysharman:
Adds two fields to the event `interactiveSessionProviderInvoked`:

   - **`chatMode`** — the active chat mode when the request was sent (e.g. `ask`, `agent`, `edit`, `Plan`,
  or a hashed string for user-created custom agents)
   - **`sessionType`** — the session resource scheme (e.g. `vscodeLocalChatSession` for local sessions)

   ### Motivation

   The existing `chat.modeChange` event only fires when a user *switches* modes, not when they *use* one.
  This means we can't reliably correlate mode/session type with actual request activity. Adding these
  fields to the per-request event enables accurate usage analysis without requiring new events.

   ### Details

   - `chatMode` uses `getModeNameForTelemetry()` to properly identify extension-contributed modes like Plan
   while hashing user-created agent names to avoid UGC leaks


alexdima (approved):
LGTM

roblourens (approved):

PR #299743, opened 2026-03-06
Revert "Bump @tootallnate/once and jsdom in /extensions/notebook-renderers"

diegojuan10500-blip:
Reverts microsoft/vscode#299312


diegojuan10500-blip (commented):
Espero y tengas una buena experiencia..

PR #299839, opened 2026-03-06
Agent host protocol initial thoughts

connor4312:
Here's Opus and I's initial thoughts on the agent host protocol. PR description from Opus with my edits :)

I have put some effort both into the to state and trying to get a good Typescript-enforced system to avoid backwards-incompatible changes. For reviewing this PR, I recommend starting with `protocol.md` and also checking out the added `AGENTS.md` (this should probably be a skill) to see how protocol compatibility works.

---

This PR lays the groundwork for our upcoming sessions process — a standalone server that multiple clients can connect to with a synchronized view of agent sessions. 

### Why

We want to move toward a model where:
- A headless server manages sessions and talks to agent backends
- Multiple clients (VS Code windows, web UIs, CLI) can connect and see the same state
- Clients stay responsive by applying changes optimistically, then reconciling with the server

This is the **protocol design + initial type-level implementation** — no transport, no migration, no UI changes.

#### State model

This is a hybrid protocol where the agent host has an imperative method-based API for some operations -- like enumerating the sessions list or requesting binary resources, things that pure state is pretty shoddy at -- and then has subscribable state that is identified by URIs:

- **Root state** (`agenthost:root`) — agents and models available on the server. Small, always loaded.
- **Session state** (`copilot:/<uuid>`) — full state for one session (turns, tool calls, permissions, streaming text). Loaded on demand when a client subscribes.

    Note: I let Opus do its thing here with UUIDs, but we'd proably want to have the session state URI's be the same as the chat session model URIs we already have.

#### Actions & reducers

All state mutations flow through a discriminated union of **actions** processed by **pure reducer functions** (sessionReducers.ts). Server and client run the same reducers — this is what makes optimistic updates possible.

There are two ways clients interact with the server:

1. **Dispatching actions** — for operations that map directly to state mutations. The client applies the action optimistically, sends it to the server, and the server applies it + reacts with side effects. For example, `session/turnStarted` mutates state to add the turn, and the server begins agent processing as a side effect.
2. **Sending commands** — for imperative RPCs that don't map to a single state action (session creation, data fetching, disposal).

```mermaid
sequenceDiagram
    participant Client A
    participant Server
    participant Client B

    Client A->>Server: dispatch: session/turnStarted
    Note over Client A: applies action<br/>optimistically
    Server->>Client A: action: session/turnStarted (echo)
    Server->>Client B: action: session/turnStarted
    Note over Client A: confirms optimistic<br/>prediction

    Server->>Client A: action: session/delta
    Server->>Client B: action: session/delta

    Server->>Client A: action: session/toolStart
    Server->>Client B: action: session/toolStart
    Server->>Client A: action: session/permissionRequest
    Server->>Client B: action: session/permissionRequest
    Note over Client A: shows approval UI

    Client A->>Server: dispatch: session/permissionResolved (approved)
    Note over Client A: applies optimistically —<br/>tool status → running
    Server->>Client B: action: session/permissionResolved
    Server->>Client A: action: session/permissionResolved (echo)

    Note over Server: executes tool

    Server->>Client A: action: session/toolComplete
    Server->>Client B: action: session/toolComplete
    Server->>Client A: action: session/delta
    Server->>Client B: action: session/delta
    Server->>Client A: action: session/turnComplete
    Server->>Client B: action: session/turnComplete
```

Key design points:
- **Client-dispatchable actions**: `session/turnStarted`, `session/turnCancelled`, `session/permissionResolved` — these are applied to state *and* trigger server-side effects (start processing, abort, unblock tool).
- **Server-only actions**: `session/delta`, `session/toolStart`, `session/toolComplete`, etc. — produced by the agent backend, clients observe but cannot produce them.
- **Commands** are only for things that aren't state actions: `createSession`, `disposeSession`, `listSessions`, `fetchContent`, `fetchTurns`.
- **Notifications** (`notify/sessionAdded`, `notify/sessionRemoved`) are ephemeral broadcasts for the session list — not stored in state, not replayed on reconnect.

#### Write-ahead reconciliation

sessionClientState.ts implements the client-side state manager. It maintains confirmed state + a pending queue, and reconciles when the server echoes back actions (possibly interleaved with actions from other clients).

#### Protocol versioning

The versioning system prevents backwards-incompatible changes at compile time:

- **Frozen type snapshots** (`versions/v1.ts`) capture the wire format at each version. The latest is the editable "tip"; older ones are frozen.
- **Bidirectional assignability checks** (`versions/versionRegistry.ts`) make it a compile error to remove fields, change types, or add required fields to any protocol type.
- **Exhaustive action→version map** — adding a new action to the union without registering its version is a compile error.
- Adding optional fields or new action types is safe and doesn't require a version bump. Version bumps are only needed for capability boundaries (client needs to know server supports a feature before using it).

See AGENTS.md for step-by-step modification instructions.

#### File layout

```
src/vs/platform/agent/common/state/
├── AGENTS.md               # Modification instructions for agents and developers
├── sessionState.ts         # Immutable state types (RootState, SessionState, Turn, etc.)
├── sessionActions.ts       # Action + notification discriminated unions
├── sessionReducers.ts      # Pure reducer functions
├── sessionProtocol.ts      # Wire protocol messages (handshake, subscribe, commands)
├── sessionCapabilities.ts  # Version constants + ProtocolCapabilities
├── sessionClientState.ts   # Client-side state manager with reconciliation
└── versions/
    ├── v1.ts               # v1 wire format types (current tip)
    └── versionRegistry.ts  # Compile-time compat checks + action→version map
```

### DIFF
--- a/src/vs/platform/agent/protocol.md
+++ b/src/vs/platform/agent/protocol.md
@@ -1,0 +1,404 @@
+# Sessions process protocol
+
+> **Keep this document in sync with the code.** Changes to the state model, action types, protocol messages, or versioning strategy must be reflected here. Implementation lives in `common/state/`.
+
+For process architecture and IPC details, see [architecture.md](architecture.md). For design decisions, see [design.md](design.md). For the task backlog, see [backlog.md](backlog.md).
+
+## Goal
+
+The sessions process is a portable, standalone server that multiple clients can connect to. Clients see a synchronized view of sessions and can send commands that are reflected back as state-changing actions. The protocol is designed around four requirements:
+
+1. **Synchronized multi-client state** — an immutable, redux-like state tree mutated exclusively by actions flowing through pure reducers.
+2. **Lazy loading** — clients subscribe to state by URI and load data on demand. The session list is fetched imperatively. Large content (images, long tool outputs) is stored by reference and fetched separately.
+3. **Write-ahead with reconciliation** — clients optimistically apply their own actions locally, then reconcile when the server echoes them back alongside any concurrent actions from other clients or the server itself.
+4. **Forward-compatible versioning** — newer clients can connect to older servers. A single protocol version number maps to a capabilities object; clients check capabilities before using features.
+
+## URI-based subscriptions
+
+All state is identified by URIs. Clients subscribe to a URI to receive its current state snapshot and subsequent action updates. This is the single universal mechanism for state synchronization:
+
+- **Root state** (`agenthost:root`) — always-present global state (agents, models). Clients subscribe to this on connect.
+- **Session state** (`copilot:/<uuid>`, etc.) — per-session state loaded on demand. Clients subscribe when opening a session.
+
+The `subscribe(uri)` / `unsubscribe(uri)` mechanism works identically for all resource types.
+
+## State model
+
+### Root state
+
+Subscribable at `agenthost:root`. Contains global, lightweight data that all clients need. **Does not contain the session list** — that is fetched imperatively via RPC (see Commands).
+
+```
+RootState {
+    agents: AgentInfo[]
+    models: ModelInfo[]
+}
+```
+
+### Session state
+
+Subscribable at the session's URI (e.g. `copilot:/<uuid>`). Contains the full state for a single session.
+
+```
+SessionState {
+    summary: SessionSummary
+    lifecycle: 'creating' | 'ready' | 'creationFailed'
+    creationError?: ErrorInfo
+    turns: Turn[]
+    activeTurn: ActiveTurn | undefined
+}
+```
+
+`lifecycle` tracks the asynchronous creation process. When a client creates a session, it picks a URI, sends the command, and subscribes immediately. The initial snapshot has `lifecycle: 'creating'`. The server asynchronously initializes the backend and dispatches `session/ready` or `session/creationFailed`.
+
+```
+Turn {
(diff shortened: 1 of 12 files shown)


src/vs/platform/agent/protocol.md
```diff
@@ -0,0 +1,404 @@
...
+### Why rebasing is simple
+
+Most session actions are **append-only** (add turn, append delta, add tool call). Pending actions still apply cleanly to an updated confirmed state because they operate on independent data (the turn the client created still exists; the content it appended is additive). The rare true conflict (two clients abort the same turn) is resolved by server-wins semantics.
+
+## Versioning
+
+### Protocol version
+
+Two constants define the version window:
+- `PROTOCOL_VERSION` — the current version that new code speaks.
```
roblourens, line 260: Do we need a 'verson', can we do a capability list like we do for other protocols instead?

roblourens (approved):

src/vs/platform/agent/protocol.md
connor4312: I think we definitely could do a capability list once we get more clients. But so far we know we have one client and one server implementation, and using the version numbering lets us build the whole typesafe backwards compatibility system. Capability-based checks are necessarily runtime-only checks which I would prefer to avoid until we start needing them (i.e. having multiple clients)

PR #299941, opened 2026-03-07
terminal: add Refresh Environment button to reload PATH/env in all sessions

AMDphreak:
Adds a **Refresh Environment** button (refresh icon) to the Terminal panel title bar. On click, it injects a shell-appropriate command into **every** terminal in the window so PATH is reloaded from the system. The command is visible so users can learn and reuse it.

No VS Code API exists to refresh an existing terminal's environment. Injection is the fix. PowerShell, CMD, bash/zsh/sh, and fish each get the appropriate one-liner or short sequence.

Fixes #299940


AMDphreak:

PR #300019, opened 2026-03-08
first working protocol version

connor4312:

### DIFF
--- a/src/vs/platform/agent/common/agentService.ts
+++ b/src/vs/platform/agent/common/agentService.ts
@@ -335,12 +335,13 @@ export const IAgentService = createDecorator<IAgentService>('agentService');
 /**
  * Service contract for communicating with the agent host process. Methods here
  * are proxied across MessagePort via `ProxyChannel`.
+ *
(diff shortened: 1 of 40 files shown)


src/vs/platform/agent/common/agentService.ts
```diff
@@ -378,6 +380,38 @@ export interface IAgentService {
...
 	shutdown(): Promise<void>;
+
+	// ---- Protocol methods (sessions process protocol) ----------------------
+
+	/**
+	 * Subscribe to state at the given URI. Returns a snapshot of the current
+	 * state and the serverSeq at snapshot time. Subsequent actions for this
+	 * resource arrive via {@link onDidAction}.
+	 */
+	subscribe(resource: URI): Promise<IStateSnapshot>;
```
roblourens, line 386: When I open a session that is already really long, I get a massive dump of all actions in the sessions? Can it be lazier/support paging in the future?

connor4312: Yea, I think we could page the turns pretty reasonably

PR #300028, opened 2026-03-08
Fix markdown preview not refreshing on external file changes

### LINKED ISSUE #265277: Markdown preview doesn't refresh/reload when the file is overwritten

tiwari91:
## Summary

Fixes the markdown preview not updating when the underlying file is overwritten externally (by git, AI tools like Copilot/Claude, or other processes).

## Root Cause

The file system watcher's `onDidChange` handler in `preview.ts` was skipping the refresh when the document was already open in VS Code (line 118), assuming `onDidChangeTextDocument` would handle it. But when files are overwritten externally, the text document model may not be updated immediately — creating a window where neither event triggers a preview refresh.

## Fix

Remove the condition that skips refresh for already-open documents. The preview now always refreshes on file system changes. Since `refresh()` is already throttled/debounced, this won't cause double-renders when both `onDidChangeTextDocument` and the file watcher fire for the same change.

Fixes #265277

### DIFF
--- a/extensions/markdown-language-features/src/preview/preview.ts
+++ b/extensions/markdown-language-features/src/preview/preview.ts
@@ -115,7 +115,7 @@ class MarkdownPreview extends Disposable implements WebviewResourceProvider {
 			this._register(watcher.onDidChange(uri => {
 				if (this.isPreviewOf(uri)) {
 					// Only use the file system event when VS Code does not already know about the file
-					if (!vscode.workspace.textDocuments.some(doc => doc.uri.toString() === uri.toString())) {
+					if (!vscode.workspace.textDocuments.some(doc => this.isPreviewOf(doc.uri))) {
 						this.refresh();
 					}
 				}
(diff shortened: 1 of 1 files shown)


tiwari91:
Thanks for the pointer @mjbvz! You're right — I updated the fix to address the Windows-specific root cause.

The issue is in `isPreviewOf` which uses case-sensitive `fsPath` comparison (`===`). On Windows (and macOS), the file system watcher can report URIs with different path casing than what VS Code uses internally. When casing differs, `isPreviewOf` returns `false` and the watcher callback exits early — so neither the watcher nor `onDidChangeTextDocument` triggers a refresh for externally modified files.

The updated fix:
1. Makes `isPreviewOf` use case-insensitive comparison on Windows/macOS (where file systems are case-insensitive), keeping case-sensitive on Linux
2. Updates the open-document check in the watcher callback to also use `isPreviewOf` instead of raw `toString()` comparison, for consistent path matching

PR #300086, opened 2026-03-08
Fix SelectionMatchFindAction to respect seedSearchStringFromSelection setting

### LINKED ISSUE #300085: SelectionMatchFindAction ignores editor.find.seedSearchStringFromSelection setting
## Bug Description

The `editor.action.nextSelectionMatchFindAction` (Find Next Selection Match, default `Cmd+F3`/`Ctrl+F3`) and `editor.action.previousSelectionMatchFindAction` commands ignore the `editor.find.seedSearchStringFromSelection` setting. This setting is respected by the Find Widget (`Ctrl+F`) and other find actions, but not by the Selection Match Find actions.

## Steps to Reproduce

1. Set `"editor.find.seedSearchStringFromSelection": "selection"` in settings
2. Select some text (e.g., `foo`)
3. Press `Cmd+F3` (or the keybinding for `editor.action.nextSelectionMatchFindAction`) — it correctly finds the next `foo`
4. Paste from clipboard to replace the found match
5. Press `Cmd+F3` again

## Expected Behavior

With `seedSearchStringFromSelection` set to `"selection"`, the action should **not** pick up the word at the cursor when there is no active selection. It should continue searching for the previous search term (`foo`).

## Actual Behavior

The action picks up the word at/near the cursor position and uses that as the new search string, overwriting the previous search term. This is because the `SelectionMatchFindAction.run()` method hardcodes the call to `getSelectionSearchString(editor, 'single', false)`, always passing `false` for `seedSearchStringFromNonEmptySelection`, which enables the word-at-cursor fallback regardless of the user's setting.

## Root Cause

In `src/vs/editor/contrib/find/browser/findController.ts`, `SelectionMatchFindAction.run()` (line ~918) hardcodes:

```typescript
const selectionSearchString = getSelectionSearchString(editor, 'single', false);
```

Compare with `StartFindAction` (line ~589) which correctly reads the setting:

```typescript
seedSearchStringFromSelection: editor.getOption(EditorOption.find).seedSearchStringFromSelection !== 'never' ? 'single' : 'none',
seedSearchStringFromNonEmptySelection: editor.getOption(EditorOption.find).seedSearchStringFromSelection === 'selection',
```

## Proposed Fix

Make `SelectionMatchFindAction.run()` respect the `editor.find.seedSearchStringFromSelection` setting, consistent with how all other find actions handle it. PR incoming.

| VS Code Version | OS |
|---|---|
| 1.111.0 (latest main) | All |

teoincontatto:
## Summary

- Makes `editor.action.nextSelectionMatchFindAction` and `editor.action.previousSelectionMatchFindAction` respect the `editor.find.seedSearchStringFromSelection` setting
- Previously these actions hardcoded `getSelectionSearchString(editor, 'single', false)`, always enabling the word-at-cursor fallback regardless of user configuration
- Now reads the setting and passes it through consistently, matching the behavior of `StartFindAction` and other find actions

Fixes #300085

## Change

One-line change in `SelectionMatchFindAction.run()` (`src/vs/editor/contrib/find/browser/findController.ts`):

**Before:**
```typescript
const selectionSearchString = getSelectionSearchString(editor, 'single', false);
```

**After:**
```typescript
const seedSearchStringFromSelection = editor.getOption(EditorOption.find).seedSearchStringFromSelection;
const selectionSearchString = seedSearchStringFromSelection !== 'never'
    ? getSelectionSearchString(editor, 'single', seedSearchStringFromSelection === 'selection')
    : null;
```

## Behavior by setting value

| Setting | Before (broken) | After (fixed) |
|---------|-----------------|---------------|
| `"always"` (default) | Seeds from selection or word at cursor | Same (no change) |
| `"selection"` | Seeds from selection **or word at cursor** | Seeds from selection **only** |
| `"never"` | Seeds from selection or word at cursor | Never seeds |

## Test plan

- [ ] With `"editor.find.seedSearchStringFromSelection": "always"` (default): select text, `Cmd+F3`, paste replacement, `Cmd+F3` again → picks up word at cursor (unchanged behavior)
- [ ] With `"editor.find.seedSearchStringFromSelection": "selection"`: select text, `Cmd+F3`, paste replacement, `Cmd+F3` again → continues searching previous term (fixed)
- [ ] With `"editor.find.seedSearchStringFromSelection": "never"`: `Cmd+F3` never seeds from editor (fixed)
- [ ] Same scenarios with `editor.action.previousSelectionMatchFindAction`


teoincontatto:
@osortega, please, can you review this PR so it can get push forward?

PR #300428, opened 2026-03-10
fix: Propagate `forceShellIntegration` to remote terminal server

### LINKED ISSUE #1337: Shell integration not injected for debug terminals in remote environments
When using the `shellStartup` activation type in a remote VS Code environment, the Python environment isn't activated in the terminal where the debugger is launched. This affects all environments that need to be activated, like Conda and Pixi.

The issue only happens when all three conditions are met:
1. Remote environment
2. `shellStartup` activation type
3. Launching the debugger (integrated terminal)

Opening a regular terminal in the same remote environment works fine. Debugging locally also works fine. Setting `autoActivationType` to `command` works around the issue.

There is a fix waiting for review over here https://github.com/microsoft/vscode/pull/300428.

renan-r-santos:
Fixes https://github.com/microsoft/vscode-python-environments/issues/1337

Debug terminals are created with `isFeatureTerminal: true` and `forceShellIntegration: true`. The `forceShellIntegration` flag is needed because VS Code skips shell integration injection for feature terminals by default (see `terminalEnvironment.ts`).

The problem is that `forceShellIntegration` is not included in `IShellLaunchConfigDto`, which is the DTO used to serialize the shell launch config when sending it from the client to the remote server. So while `isFeatureTerminal: true` is correctly propagated, `forceShellIntegration` is silently dropped.

On the remote server, the shell integration injection check sees `isFeatureTerminal: true` and `forceShellIntegration: undefined`, and skips injection. Without shell integration, the Python activation hooks built into VS Code's shell integration scripts (e.g. the `VSCODE_PYTHON_BASH_ACTIVATE` eval in `shellIntegration-bash.sh`) never run. The `.bashrc` also doesn't have the activation code because the Python Environments extension skipped profile modification after detecting that shell integration was available.

Locally this works because the full `IShellLaunchConfig` (including `forceShellIntegration`) is passed directly to the pty host without going through the DTO serialization.

In addition to the issue described above, debug terminals were reused across different working directories, causing stale environment activations from previous debug sessions to persist. I added `cwd` to the terminal config key used for matching, so terminals are only reused when the shell type and working directory both match, so each project directory gets its own debug terminal with the correct activated Python environment.

This PR was manually tested with `./scripts/code.sh --remote test+test`


meganrogge (approved):
Thanks!

TylerLeonhardt (approved):

PR #300551, opened 2026-03-10
ai-customizations: improve list visual scannability

joshspicer:
Fixes #299211

### Item list (AI customizations, MCP servers, plugins)
- **Type icon per item** — each row now shows the appropriate icon for its type (agent, skill, instructions, prompt, hook, MCP server, plugin), making it easy to distinguish entries at a glance
- **Name formatting** — display names have dashes/underscores replaced with spaces, title-cased, and trailing `.md` stripped (e.g. `my-skill.Md` → `My Skill`)
- **Description truncation** — descriptions are clipped to the first sentence (120-char hard fallback) to reduce visual noise on the second line
- **Font weight** — removed the explicit `font-weight: 500` override; hierarchy comes from the full-foreground 13px title vs the smaller muted description below

### Group headers
- **Count badge color** — switched from `badge.background` (prominent blue) to `list.inactiveSelectionBackground/Foreground` for a more muted secondary look. Applied to both the group-header count and sidebar section count
- **Built-in MCP group** — now uses `builtinIcon` (⭐) to match the prompts Built-in group header

### Shared helpers (`aiCustomizationListWidget.ts`)
Exported `formatDisplayName` and `truncateToFirstSentence` so they can be reused by `mcpListWidget.ts` and `pluginListWidget.ts`.


joshspicer (commented):
Fixed in 3935a52 — `filterItems` now computes `nameMatches` against `formatDisplayName(item.name)` instead of the raw `item.name`, so highlight positions align with what `HighlightedLabel` displays. The JSDoc comment on `formatDisplayName` has also been updated to drop the outdated "1:1 transformation" claim and instead note that callers with `IMatch` ranges must compute them against the formatted string.

rzhao271 (approved):

PR #300577, opened 2026-03-10
Proposed browser API for debug support

kycutler:
See #300319, #289164


src/vscode-dts/vscode.proposed.browser.d.ts
```diff
@@ -0,0 +1,92 @@
...
+ *--------------------------------------------------------------------------------------------*/
+
+declare module 'vscode' {
+
+	// @kycutler https://github.com/microsoft/vscode/issues/300319
+
+	/**
+	 * An integrated browser page displayed in an editor tab.
+	 */
+	export interface BrowserTab {
```
jruales, line 13: Should a BrowserTab have some sort of ID to distinguish multiple tabs that have the same url, title, and icon?

extensions/vscode-api-tests/src/singlefolder-tests/browser.test.ts
```diff
@@ -0,0 +1,68 @@
...
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import * as assert from 'assert';
+import * as vscode from 'vscode';
+import { window } from 'vscode';
+import { assertNoRpc, closeAllEditors } from '../utils';
+
+(vscode.env.uiKind === vscode.UIKind.Web ? suite.skip : suite)('vscode API - browser', () => {
```
jruales, line 11: Several parts of the new API are not covered by these tests. Is that intentional just to cover a subset here?

src/vs/workbench/contrib/browserView/browser/browserView.contribution.ts
```diff
@@ -0,0 +1,47 @@
...
+class WebBrowserViewCDPService implements IBrowserViewCDPService {
+	declare readonly _serviceBrand: undefined;
+
+	async createSessionGroup(_browserId: string): Promise<string> {
+		throw new Error('Integrated Browser is not available in web.');
+	}
+
+	async destroySessionGroup(_groupId: string): Promise<void> { }
+
+	async sendCDPMessage(_groupId: string, _message: CDPRequest): Promise<void> { }
```
jruales, line 35: should we also throw Errors for these other methods that are currently no-oping?

src/vs/workbench/contrib/browserView/common/browserEditorInput.ts
```diff
@@ -48,6 +47,7 @@ export interface IBrowserEditorInputData {
 
 export class BrowserEditorInput extends EditorInput {
 	static readonly ID = 'workbench.editorinputs.browser';
+	static readonly EDITOR_ID = 'workbench.editorinputs.browser';
```
jruales: Is it expected that these are the same value?

src/vs/workbench/api/common/extHost.api.impl.ts
```diff
@@ -1058,6 +1060,34 @@ export function createApiFactoryAndRegisterActors(accessor: ServicesAccessor): I
 				checkProposedApiEnabled(extension, 'chatParticipantPrivate');
 				return _asExtensionEvent(extHostChatAgents2.onDidChangeActiveChatPanelSessionResource)(listeners, thisArgs, disposables);
 			},
+			get browserTabs() {
+				checkProposedApiEnabled(extension, 'browser');
```
jruales, line 1064: Just `'browser'`? Won't that be confused with any functionality that targets external browser?

src/vs/workbench/api/common/extHost.protocol.ts
```diff
@@ -1351,6 +1353,30 @@ export interface ExtHostSpeechShape {
...
+	$startCDPSession(sessionId: string, browserId: string): Promise<void>;
+	$closeCDPSession(sessionId: string): Promise<void>;
+	$sendCDPMessage(sessionId: string, message: CDPRequest): Promise<void>;
+}
+
+export interface ExtHostBrowsersShape {
+	$onDidOpenBrowserTab(browser: BrowserTabDto): void;
+	$onDidCloseBrowserTab(browserId: string): void;
+	$onDidChangeActiveBrowserTab(browser: BrowserTabDto | undefined): void;
+	$onDidChangeBrowserTab(browserId: string, data: BrowserTabDto): void;
```
jruales: nit: onDidChangeActiveBrowserTab and onDidChangeBrowserTab sound very semantically similar to one another. It should be ok, but just pointing it out

jruales (dismissed):

src/vs/workbench/api/common/extHost.protocol.ts
kycutler: Changed to `$onDidChangeBrowserTabState`. This also aligns better with the API events

src/vs/workbench/api/common/extHost.api.impl.ts
kycutler: I think it's fine since it's just the proposal name. Other proposals can be more specific if there is overlap, or we can always rename this one

src/vs/workbench/contrib/browserView/common/browserEditorInput.ts
kycutler: Whoops, good catch! Fixed

src/vscode-dts/vscode.proposed.browser.d.ts
kycutler: I think we would want to expose the associated Tab (since extensions wouldn't have any use for the id), but it isn't necessary right now

src/vs/workbench/contrib/browserView/browser/browserView.contribution.ts
kycutler: The methods that throw are the ones that return values -- I'll keep the rest as no-ops so we don't throw if the extension APIs try to call anything during setup or teardown

extensions/vscode-api-tests/src/singlefolder-tests/browser.test.ts
kycutler: Good point -- I didn't want to go overboard here but we do still want full coverage. Added, thanks!

lszomoru (approved):

Issue #156077, opened 2022-07-24
explorer.excludeGitIgnore fails to show unignored files and directories


memeplex:
Does this issue occur when all extensions are disabled?: Yes

- VS Code Version: 1.69.2
- OS Version: macOS Monterey

Both directories and files inside those directories that were unignored in .gitignore fails to be shown by explorer.

Steps to Reproduce:

```bash
cd /tmp
mkdir test-repo
cd test-repo
git init
mkdir d
touch d/a
touch b
git add d/a
git add b
git commit -m 'Initial commit'
echo '*' > .gitignore
echo '!d/' >> .gitignore
echo '!d/a' >> .gitignore
echo '!b' >> .gitignore
mkdir .vscode
echo '{"explorer.excludeGitIgnore": true}' > .vscode/settings.json
code .
```

Check that only 'b' is listed in the explorer.

Back in the console, check that no file is ignored by git:

```
git check-ignore d
git check-ignore d/a
git check-ignore b
```

Maybe related: https://github.com/microsoft/vscode/issues/156076

memeplex:
The documentation of excludeGitIgnore says:

> Controls whether entries in .gitignore should be parsed and excluded...

Perhaps the problem is that it tries to replicate git ignore rules and falls short of it, instead of querying git for the ignore status of files.

lramos15:
> Perhaps the problem is that it tries to replicate git ignore rules and falls short of it, instead of querying git for the ignore status of files.

Yes we try to emulate it because we do not have access to the git executable in core. This has some good benefits too such as not needing git installed for this to work and working in web. I'm surprised b shows up here because we don't support exclude patterns

memeplex:
> we don't support exclude patterns

Sorry, I'm not quite following you here. What do you mean with exclude patterns? !-prefixed one? Those would be re-include, or un-exclude, or better un-ignore patterns. If you mean that you don't support them at all then, yes, it's indeed surprising that b shows up here. Perhaps there is some partial support in place?

lramos15:
> ? Those would be re-include, or un-exclude, or better un-ignore patterns.

Yeah these. There's definitely many improvements to be made with this parser. I think many of the issues listed all stem from our parsing being slightly different than gits

memeplex:
> we don't support exclude patterns

@lramos15 I'm still surprised about this statement. I've also reported https://github.com/microsoft/vscode/issues/156076 some days ago. Now, take as an example this .gitignore:

```
*
```

[...]

PR #300613, opened 2026-03-11
Fix gitignore negation in child files not overriding parent/global rules

### LINKED ISSUE #291157: Negation doesn't work in nested scenarios
Testing #290739

On Linux

* Create a root `.gitignore`, `test.txt`, `TeSt.txt`

```
test.txt
node_modules
```

* Create a subfolder `Node_Modules` and add a `.gitignore`
```
!test.txt
```
* Create a file `test.txt` under the subfolder
* Set `"explorer.excludeGitIgnore": true`
* Expect file under subfolder to be visible, but it is hidden

https://github.com/user-attachments/assets/693e909b-8dc9-4a26-b2e2-1d690ca8ddfd

Dmitriusan:
## Problem

When `explorer.excludeGitIgnore` is enabled, negation patterns (e.g., `!.myconfig/`) in a project's `.gitignore` do not override matching positive patterns from parent `.gitignore` files (e.g., a root `.gitignore` for a nested workspace folder, or a chained ignore tree).

**Reproduction (from #291157):**
1. Root `.gitignore` contains `test.txt`
2. Subfolder `.gitignore` contains `!test.txt`
3. With `"explorer.excludeGitIgnore": true`, the subfolder's `test.txt` is hidden, even though git considers it not ignored

## Root Cause

In `IgnoreFile.parseIgnoreFile()`, the `isPathIgnored` function only uses negation patterns to counteract positive patterns **within the same file**. When the current file has no positive pattern for a path, it falls through to the parent's `isPathIgnored()` without checking whether the current file has a negation that should override the parent.

In git's actual behavior, the last matching rule wins, and rules in more-specific `.gitignore` files (child directories) take precedence over less-specific ones (parent directories).

## Fix

Before delegating to the parent, check if the current file has a negation pattern that matches the path. If so, return `false` (not ignored) without consulting the parent. This correctly implements git's precedence semantics.

Also caches `isDirIncluded` / `isFileIncluded` matcher results within a single `isPathIgnored` invocation so they are not evaluated twice.

## Testing

Added two test cases:
- **Child negation overrides parent ignore for directories** — simulates a root ignore pattern being overridden by a project-level negation
- **Child negation overrides parent ignore for files** — simulates `*.log` / `!important.log` across parent/child

Fixes #291157
Relates to #156077


dmitrivMS:
Made some cleanup for perf + test fixes.

dmitrivMS:
Linked issue that will be fixed by this, verified fix manually.
\

dmitrivMS (approved):

alexdima (approved):

PR #300615, opened 2026-03-11
Sessions: Add toggle visibility logic to terminal icon button

hawkticehurst:
Makes the terminal icon button in the top right corner of the Sessions window a proper toggle button that will open / close the terminal panel.


src/vs/monaco.d.ts
```diff
@@ -5263,7 +5263,7 @@ declare namespace monaco.editor {
 	export const EditorOptions: {
 		acceptSuggestionOnCommitCharacter: IEditorOption<EditorOption.acceptSuggestionOnCommitCharacter, boolean>;
 		acceptSuggestionOnEnter: IEditorOption<EditorOption.acceptSuggestionOnEnter, 'on' | 'off' | 'smart'>;
-		accessibilitySupport: IEditorOption<EditorOption.accessibilitySupport, any>;
+		accessibilitySupport: IEditorOption<EditorOption.accessibilitySupport, AccessibilitySupport>;
```
benibenj, line 5266: There should not be any changes in `monaco.d.ts`

benibenj (changes requested):
Looks good except the monaco.d.ts file change which was pushed by accident?

Issue #299632, opened 2026-03-05
Adopt npm `min-release-age` to avoid official build breaks due to recently published packages 


mjbvz:
https://github.com/npm/cli/pull/8965

Let's see if we can adopt min-release-age to avoid cases where our official build package registry doesn't have a recently published package version

PR #300641, opened 2026-03-11
Add `min-release-age` in npmrc files

mjbvz:
For #299632

Not fully enabled unless you're using a recent npm. Next step will be to bump the min node/npm versions for the workspace too

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -6,3 +6,4 @@ ignore-scripts=false
 build_from_source="true"
 legacy-peer-deps="true"
 timeout=180000
+min-release-age="1"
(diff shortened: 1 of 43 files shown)


deepak1556:
> Next step will be to bump the min node/npm versions for the workspace too

Switching to npm v11 requires work due to a breaking change on how unsupported configs from .npmrc will not be respected and will break any native module compilation https://github.com/npm/cli/issues/8153, I started https://github.com/microsoft/vscode/pull/274850 a while back but didn't get chance to finish it. Please feel free to take over but it is a requirement if we need to bump versions.

The pre-requisite for the PR is to have node-gyp version that has https://github.com/nodejs/node-gyp/commit/1822dff4f616a30ac3ca72e5946d81389cb8557e which comes with npm v11.6.1. The flow should be,

1) Bump workspace npm requirement to >= 11.6.1
2) Adopt the config change that moves native module configs into respective package.json

justschen (approved):

PR #300811, opened 2026-03-11
Revert macOS multiline PTY chunking workaround and simplify shell matrix tests

jcansdale:
## Summary
- revert the macOS multiline PTY chunking workaround from #298993
- keep a simplified shell matrix test for `bash`, `zsh`, `sh`, `ksh`, `dash`, `csh`, and `tcsh`
- use a shell-neutral multiline `printf` payload so the same command shape works across all tested shells

## Current Test State
| Shell | 10 lines | 20 lines |
| --- | --- | --- |
| bash | ✅ | ❌ |
| zsh | ✅ | ✅ |
| sh | ✅ | ❌ |
| ksh | ✅ | ❌ |
| dash | ✅ | ✅ |
| csh | ✅ | ❌ |
| tcsh | ✅ | ❌ |

The 10-line case passes on all tested shells. The 20-line case fails on bash, sh, ksh, csh, and tcsh with no output file, while zsh and dash continue to pass.

### DIFF
--- a/src/vs/platform/terminal/node/terminalProcess.ts
+++ b/src/vs/platform/terminal/node/terminalProcess.ts
@@ -77,6 +77,31 @@ const generalShellTypeMap = new Map<string, GeneralShellType>([
 	['node', GeneralShellType.Node],
 	['xonsh', GeneralShellType.Xonsh],
 ]);
+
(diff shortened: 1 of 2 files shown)


meganrogge (approved):
Thanks!

Issue #300770, opened 2026-03-11
✋ Agent Sessions Tree is extremely inefficient


benibenj:
- The agent sessions tree updates the entire tree far to often
  - Do not rerender the entire tree just to change the status label on one of the nodes. Have that single tree node update it's dom reactively based on the status changing
- The agent sessions tree updates even though it is not visible
  - Do not update the agent sessions tree if it is hidden/not visible. Update it when it becomes visible

These are all agent sessions tree update calls even though the tree is not visible

<img width="1070" height="719" alt="Image" src="https://github.com/user-attachments/assets/2e28e7b3-c0f4-4839-a4ad-2f0e71c1fcf8" />

mjbvz:
@bpasero https://github.com/microsoft/vscode/pull/300946 Gives you a delta of added/changed/removed session. This event is still being way fired too often right now but at least now it should only fire for a single session instead of saying the entire sessions list is outdated

PR #300946, opened 2026-03-11
Make chat sessions give a delta of what has changed

mjbvz:
For #300770

### DIFF
--- a/src/vs/sessions/test/web.test.ts
+++ b/src/vs/sessions/test/web.test.ts
@@ -22,7 +22,7 @@ import { URI } from '../../base/common/uri.js';
 import { Disposable } from '../../base/common/lifecycle.js';
 import { IWorkbenchContribution, registerWorkbenchContribution2, WorkbenchPhase } from '../../workbench/common/contributions.js';
 import { IChatProgress } from '../../workbench/contrib/chat/common/chatService/chatService.js';
-import { IChatSessionsService, IChatSessionItem, IChatSessionFileChange, ChatSessionStatus, IChatSessionHistoryItem } from '../../workbench/contrib/chat/common/chatSessionsService.js';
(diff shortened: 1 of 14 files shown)


src/vs/sessions/test/web.test.ts
```diff
@@ -273,22 +273,27 @@ class MockChatAgentContribution extends Disposable implements IWorkbenchContribu
 
 		// Add or update session in list
 		const existing = this._sessionItems.find(s => s.resource.toString() === key);
```
mjbvz, line 275: @osortega We should never be converting uris to strings in new code. Use `ResourceMap` and `isEqual` instead

src/vs/workbench/contrib/chat/browser/agentSessions/agentSessionsModel.ts
```diff
@@ -434,7 +435,19 @@ export class AgentSessionsModel extends Disposable implements IAgentSessionsMode
 		// Sessions updates
 		this._register(this.chatSessionsService.onDidChangeItemsProviders(({ chatSessionType }) => this.resolve(chatSessionType)));
 		this._register(this.chatSessionsService.onDidChangeAvailability(() => this.resolve(undefined)));
-		this._register(this.chatSessionsService.onDidChangeSessionItems(({ chatSessionType }) => this.updateItems([chatSessionType], CancellationToken.None)));
+		this._register(this.chatSessionsService.onDidChangeSessionItems((delta) => {
+			const changedChatSessionTypes = new Set<string>();
```
mjbvz, line 439: @bpasero As a next step, you can pass along the `delta` directly here and delete this adapter code

TylerLeonhardt (dismissed):

TylerLeonhardt (approved):

src/vs/workbench/contrib/chat/browser/agentSessions/agentSessionsModel.ts
bpasero: @mjbvz can you clarify what the change is you expect here?

PR #301090, opened 2026-03-12
git: show warning-level stderr messages with warning severity instead of error

### LINKED ISSUE #280834: Git Warning Message Incorrectly Styled as Error
## Description

When Git/SSH outputs the informational message warning: permanently added 'gitlab.com' to the list of known hosts, VS Code displays it with error-level styling (red error icon with X), making it appear as a critical error when it's actually just an informational message.

## Expected Behavior

Informational Git/SSH warnings should be styled as warnings (yellow/orange) or informational messages (blue), not as errors (red).

## Actual Behavior

The message is displayed with a red circular error icon containing a white X, which is the same styling used for actual errors. This is misleading because:

- The message is informational, not an error
- It's a normal part of SSH's first-time connection process
- Users may be alarmed by the error styling

## Steps to Reproduce

1. Connect to a Git repository (e.g., GitLab) via SSH for the first time
2. Observe the warning popup that appears
3. Notice the red error icon styling despite the message being a warning

## Screenshot

See attached screenshot showing the incorrectly styled warning message.

## Additional Context

This issue likely stems from VS Code parsing Git output and mapping the "warning" keyword to error-level styling. The message originates from Git/SSH, not VS Code, but VS Code's UI is applying incorrect styling.

Related issue in Cursor: https://github.com/getcursor/cursor/issues/3833

## Environment

- VS Code version: [current version]
- OS: Windows 11
- Git version: [current version]

edenfunf:
## Problem

When Git or SSH outputs informational warnings to stderr (such as `Warning: Permanently added 'gitlab.com' to the list of known hosts.`), VS Code displays them with error-level styling — a red error icon with a white X. This makes them appear as critical errors even though they're just informational messages that occur normally during SSH operations.

## Root Cause

In the default error handler (`commands.ts`), all unrecognized stderr output is treated as an error. The code strips the `error:` prefix from messages but has no handling for `warning:` prefixed lines. When a git command fails and stderr contains SSH warnings mixed with (or instead of) actual errors, the warning text gets shown with error severity.

## Fix

Added a `classifyGitStderrLines()` helper that separates stderr lines into warnings and errors based on their prefix:

- Lines starting with `warning:` (case-insensitive) are classified as warnings
- All other lines are classified as errors

The default error handler now uses this classification to:
1. **When actual error lines are present**: show only the error lines (filtering out warnings) with error severity — same behavior as before but without the confusing warning text
2. **When only warning lines are present**: show the first warning with warning severity (yellow icon, non-modal) instead of error severity
3. **When no lines remain**: fall back to the generic "Git error" message

## Changes

- `extensions/git/src/git.ts` — added `classifyGitStderrLines()` function
- `extensions/git/src/commands.ts` — updated default error handler to use severity classification
- `extensions/git/src/test/git.test.ts` — added tests for the new classification function

Fixes #280834


edenfunf:

Issue #300762, opened 2026-03-11
Investigate alternatives to fixed 5ms pacing in the macOS multiline PTY workaround


jcansdale:
## Context

This is a follow-up issue for the workaround discussion raised in @jarred-sumner's post:

- https://x.com/jarredsumner/status/2031650127421808789

Related VS Code work:

- Original PR: https://github.com/microsoft/vscode/pull/298993
- Existing bug: https://github.com/microsoft/vscode/issues/296955
- Current PR: https://github.com/microsoft/vscode/pull/300740

The original bug is real: on macOS, multiline PTY input can get corrupted or stuck when the shell is fed a large multiline payload through VS Code's terminal pipeline. The current PR fixes one concrete VS Code-side bug by making the macOS multiline chunking gate UTF-8 byte-aware instead of relying on JS string length.

This issue is about the remaining design question: whether the current fixed `5 ms` pacing heuristic is acceptable, and what viable alternatives exist.

## What We Investigated

We looked at two separate concerns raised in the discussion:

1. Whether the current workaround's `5 ms` sleep between chunks is too heuristic, non-deterministic, or slow.
2. Whether short writes or retries in the lower-level write path might actually be the main issue.

### 1. UTF-8 predicate mismatch in VS Code

We confirmed a real VS Code-side bug in the current macOS multiline gate:

- The old gate used JS string length.
- The PTY limit that matters here is UTF-8 bytes.
- Multibyte payloads can stay under the UTF-16 gate while exceeding the relevant UTF-8 threshold.

That is now covered by a regression test in the PR and fixed locally by switching the gate and chunk splitting to UTF-8 byte semantics.

### 2. Lower-level partial write and retry handling

We re-read the `node-pty` Unix write path and did not find evidence that this specific repro is simply explained by missing short-write handling in VS Code.

At a high level, upstream `node-pty` already appears to:

- queue writes,
- handle partial writes with offsets,
- retry on `EAGAIN`.

So this still looks like a separate issue from the VS Code-side UTF-16-vs-UTF-8 predicate bug and the canonical-mode or line-editor backpressure behavior seen on macOS.

That said, this area could still use a sharper audit from people who know the PTY stack better.

### Baseline fixed-delay workaround

Existing local findings from the macOS multiline workaround investigation:

- `timeout(0)` failed the medium and large multiline tests.
- `timeout(1)` passed the medium case but still failed larger cases.
- `timeout(5)` passed the existing 500-line local probe.

[...]

Issue #296955, opened 2026-02-23
Terminal tool corrupts multiline commands exceeding 1024 bytes


jcansdale:
## Bug

The terminal tool (`run_in_terminal`) corrupts **any multiline command exceeding 1024 bytes**. After byte 1024, content wraps around and replays earlier buffer content, destroying the remainder of the command and leaving the shell stuck in `quote>` or `heredoc>` mode.

## Minimal Reproduction

Ask Copilot to run this in the terminal (19 lines × 55 chars + wrapper = 1077 bytes):

```
echo 'L01 aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
L02 aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
L03 aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
L04 aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
L05 aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
L06 aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
L07 aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
L08 aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
L09 aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
L10 aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
L11 aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
L12 aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
L13 aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
[...]

PR #301133, opened 2026-03-12
fix: use bracketed paste mode for chat terminal tool to avoid macOS PTY corruption

jcansdale:
## Summary

Use bracketed paste mode when the chat terminal tool forwards multiline commands to the live terminal. This mitigates the macOS PTY canonical-mode buffer corruption that occurs with multiline input exceeding ~1024 bytes.

## The Problem

On macOS, the kernel's PTY line discipline has a `MAX_INPUT` buffer of ~1024 bytes in canonical mode. When VS Code's chat terminal tool sends large multiline commands (e.g., long `echo` statements, inline scripts), the data corrupts at the 1024-byte boundary — characters repeat, lines get mangled, or the shell hangs.

This affects all shells on macOS, though the severity varies. See #296955 for details.

## The Fix

**One-line change** in `chatTerminalToolProgressPart.ts`:

```typescript
// Before
liveTerminalInstance.sendText(data, false);

// After
liveTerminalInstance.sendText(data, false, true);
```

The third parameter (`bracketedPasteMode: true`) tells `sendText` to wrap the input in `\x1b[200~...\x1b[201~` escape sequThe third parameter (`bracketedPasteraThe thirdsteThe third parameter (`bracketedPasteMode: true`) tells `sendText` to wrap the input in `\x1b[200~...\x1b[201~` escape sequThe third parameter (`bracketedPasteraThe thirdsteThe third parameter (`bracketedPasteMode: true`e vThe third parmodes.bracketedPasteMode`
- `sendText()` already supports wrapping input in bracketed paste sequences
- The wrapping **only happens** when the shell has opted in — safe for all shells

### Coverage

| Shell | Bracketed Paste | Fixed? |
|-------|----------------|--------|
| zsh 5.x | Yes (default on) | Yes |
| bash 5.1+ | | bash 5.1+ | | bash 5.1+ | | bash 5.1+ | | bash 5.1+ | | bash 5.1+ | | bash 5.1+ | ativ| bax |
| ks| ks| ks| ks| ks| ks| ks| ks| ks| ks| ks| ks| ks| ks| kshis| ks|lso inclu| ks| ksle| ks| ks| ks| ks| ks| ks| ks| ks| ks| ks) that reprodu| ks| ks| ks| ks| ks| ks| ks| ks| ks| ks| ks| km.js. These tests document which shells are affected:

- **zsh**: passes (runs in raw mode by default)
- **bash** (Homebrew 5.x): passes
- **bash** (`/bin/bash` 3.2): fails — confirms the bug
- **sh** (= bash 3.2 on macOS): fails — confirms the bug
- **ksh**: passes (but would fail with larger payloads)

Refs #296955, #298993, #300762

### DIFF
--- a/src/vs/workbench/contrib/chat/browser/widget/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
+++ b/src/vs/workbench/contrib/chat/browser/widget/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
@@ -1194,10 +1194,12 @@ class ChatTerminalToolOutputSection extends Disposable {
 				this._scrollOutputToBottom();
 			}
 		}));
-		// Forward input from the mirror terminal to the live terminal instance
+		// Forward input from the mirror terminal to the live terminal instance.
+		// Use bracketed paste mode when available to avoid macOS PTY canonical-mode
+		// buffer corruption with multiline input exceeding ~1024 bytes.
 		this._register(mirror.onDidInput(data => {
 			if (!liveTerminalInstance.isDisposed) {
-				liveTerminalInstance.sendText(data, false);
+				liveTerminalInstance.sendText(data, false, true);
 			}
 		}));
 		await mirror.attach(this._terminalContainer);
--- a/src/vs/platform/terminal/test/node/terminalProcess.test.ts
+++ b/src/vs/platform/terminal/test/node/terminalProcess.test.ts
@@ -31,10 +31,104 @@ const shellMatrix: IShellLaunchConfig[] = [
 ];
 
 const lineCounts = [10, 20, 30, 40, 50];
+const iterationsPerTest = 3;
 const waitDuration = 4000;
 
 const BRACKETED_PASTE_START = '\x1b[200~';
 const BRACKETED_PASTE_END = '\x1b[201~';
+
+/**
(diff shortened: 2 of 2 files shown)


src/vs/workbench/contrib/chat/browser/widget/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
```diff
@@ -1194,10 +1194,12 @@ class ChatTerminalToolOutputSection extends Disposable {
 				this._scrollOutputToBottom();
 			}
 		}));
-		// Forward input from the mirror terminal to the live terminal instance
+		// Forward input from the mirror terminal to the live terminal instance.
+		// Use bracketed paste mode when available to avoid macOS PTY canonical-mode
+		// buffer corruption with multiline input exceeding ~1024 bytes.
 		this._register(mirror.onDidInput(data => {
 			if (!liveTerminalInstance.isDisposed) {
```
meganrogge, line 1201: shouldn't we restrict this to macOS?

meganrogge (changes requested):
It seems to work based on my testing. One change and incorporating Copilot's comments would be good.

PR #301176, opened 2026-03-12
fix: correctly handle falsy values in HistoryNavigator._currentPosition

Muszic:
This PR fixes a bug in `HistoryNavigator._currentPosition()` where a falsy check (`!currentElement`) was used instead of a strict null check (`currentElement === null`).

**The Issue:**
If the history contained a valid falsy value (such as an empty string `''` from a cleared search input), `!currentElement` evaluated to `true`, causing `_currentPosition()` to incorrectly return `-1`. This broke state-dependent methods like `isFirst()`, `isLast()`, and `previous()`.

**The Fix:**
Replaced `if (!currentElement)` with `if (currentElement === null)` to ensure valid falsy values like `''`, `0`, or `false` are properly indexed.

**Tests:**
Added a regression test to `history.test.ts` to ensure `isFirst()` and `previous()` correctly handle the start of the history array when the first element is a falsy value.


Muszic:
@federicobrancasi review the PR please

Muszic:
@federicobrancasi Please review

PR #301224, opened 2026-03-12
Connect to remote agent hosts via setting chat.remoteAgentHosts

roblourens:


src/vs/platform/agentHost/common/remoteAgentHostService.ts
```diff
@@ -0,0 +1,88 @@
...
+
+/**
+ * Manages connections to one or more remote agent host processes over
+ * WebSocket. Each connection is identified by its address string.
+ *
+ * Provides a unified event stream for actions and notifications across
+ * all connected remotes, and exposes per-connection subscribe/dispatch/
+ * command methods.
+ */
+export interface IRemoteAgentHostService {
```
connor4312, line 27: This is like IAgentService + extra data about connections, maybe we can dedupe some of that

src/vs/platform/agentHost/node/protocolServerHandler.ts
```diff
@@ -83,6 +84,7 @@ export class ProtocolServerHandler extends Disposable {
 		let client: IConnectedClient | undefined;
 
 		disposables.add(transport.onMessage(msg => {
+			process.stderr.write(`[ProtocolServer] onMessage: ${JSON.stringify(msg).substring(0, 500)}\n`);
```
connor4312: Maybe some leftover logging, this seems pretty verbose

src/vs/platform/agentHost/node/webSocketClientTransport.ts
```diff
@@ -0,0 +1,145 @@
...
+		const obj = value as Record<string, unknown>;
+		if (obj.$mid === 1) {
+			return URI.revive(value as URI);
+		}
+		if (obj.$type === 'Map' && Array.isArray(obj.entries)) {
+			return new Map(obj.entries as [unknown, unknown][]);
+		}
+	}
+	return value;
+}
```
connor4312: The agent duped this from the existing `webSocketTransport` We should centralize serialization/reviving. Doesn't have to be in this PR.

PR #301379, opened 2026-03-13
Migrate queued requests when swapping sessions (for CLI)

### LINKED ISSUE #300485: Queued messages disappear from Copilot CLI agent
- Send a message to Copilot CLI 
- Add two subsequent messages to the queue
- ❓ Both queued messages disappear

In video below, I ask the CLI to add emojis to the readme, followed by more testing info and then more exclamations. The latter two queued messages disappear after the emoji prompt finishes.

https://github.com/user-attachments/assets/d552171e-40d5-4c6f-bccd-29fecda05477

I also tried just queueing 1 message, but that had the same issue.

DonJayamanne:
Fixes https://github.com/microsoft/vscode/issues/300485

### DIFF
--- a/src/vs/workbench/contrib/chat/common/chatService/chatService.ts
+++ b/src/vs/workbench/contrib/chat/common/chatService/chatService.ts
@@ -1467,6 +1467,12 @@ export interface IChatService {
 	removeRequest(sessionResource: URI, requestId: string): Promise<void>;
 	cancelCurrentRequestForSession(sessionResource: URI, source?: string): Promise<void>;
 	/**
+	 * Migrates all in-flight and queued pending requests from one session to another.
+	 * Cancels the in-flight request on the original session, removes queued requests,
+	 * and re-sends them all on the target session preserving their original send options.
+	 */
+	migrateRequests(originalResource: URI, targetResource: URI): void;
+	/**
 	 * Sets yieldRequested on the active request for the given session.
 	 */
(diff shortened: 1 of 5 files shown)


src/vs/workbench/contrib/chat/common/chatService/chatService.ts
```diff
@@ -1466,6 +1466,12 @@ export interface IChatService {
 	adoptRequest(sessionResource: URI, request: IChatRequestModel): Promise<void>;
 	removeRequest(sessionResource: URI, requestId: string): Promise<void>;
 	cancelCurrentRequestForSession(sessionResource: URI, source?: string): Promise<void>;
+	/**
+	 * Migrates all in-flight and queued pending requests from one session to another.
+	 * Cancels the in-flight request on the original session, removes queued requests,
+	 * and re-sends them all on the target session preserving their original send options.
+	 */
+	migrateRequests(originalResource: URI, targetResource: URI): void;
```
DonJayamanne, line 1474: @connor4312 CLI hasn't adopted the new Controller API, hence this is only temporary.
But requried,
Do let me know if you think this is still not a good idea, I'd like to get it in as a short term solution,
As adopting the new Controller API that Matt has built will make all of this unnecessary (no more swapping)
However thats blocked on few other things (currently being actively worked on)

rebornix (approved):

Issue #241978, opened 2025-02-25
terminal cursor reset as if it was cleared, but content is still there


amunger:
I clicked into an existing terminal and the cursor jumped to the top, right where it would be if it was a new/cleared terminal.

![Image](https://github.com/user-attachments/assets/057c5f69-3852-4b8f-a5f0-49cfcac1d5ee)

only happened once, not sure why

amunger:
oh, maybe because of that notification lol. I didn't notice that

Tyriar:
Was going to close this as out of scope, but maybe we should just accept that this will always be a possibility on Windows and try only restore the last terminal's content to scrollback, not the viewport.

Tyriar:
@amunger for that case it might have been because you resized in which case we can't do anything about it or detect. Only the reconnect/revive case could we maybe improve

Tyriar:
I think we need to special case resize on conpty-backed terminals and insert empty lines. This is after resizing narrow and back on WT:

![Image](https://github.com/user-attachments/assets/0aae194d-7510-4c57-88f2-55e8974da930)

hediet:
I run into this very frequently:

![Code_-_Insiders_jJaThc4ASi](https://github.com/user-attachments/assets/f09e15cb-122e-44a6-9ae5-3a7830a21586)

TheLime1:
![Image](https://github.com/user-attachments/assets/40f9e36b-2c9a-4689-b35d-73a22e989197)

Tyriar:
xterm.js issue for the resize problem: https://github.com/xtermjs/xterm.js/issues/5319

Tyriar:
This should be fixed in the next build, please keep an eye out for similar issues.

amunger:
just seen in latest insiders

![Image](https://github.com/user-attachments/assets/7e39de12-87a8-43ea-a28c-066f21d47bfd)

amunger:
and from switching monitors

![Image](https://github.com/user-attachments/assets/a6aacb8d-3025-46dc-91b9-93fad67c33f0)

rzhao271:
I still see similar behaviour on the latest Insiders where I type "git status" and the terminal gives me "it status" plus an error message. This behaviour occurs with suggestions disabled as well.

![Image](https://github.com/user-attachments/assets/e7cc35a3-51e7-4e57-9dd1-886ddadbe162)

Tyriar:
[...]

PR #301398, opened 2026-03-13
Update conpty to 1.25

anthonykim1:
Take https://github.com/microsoft/node-pty/pull/898 for https://github.com/microsoft/vscode/issues/241978

### DIFF
--- a/src/vs/platform/terminal/node/terminalProcess.ts
+++ b/src/vs/platform/terminal/node/terminalProcess.ts
@@ -337,7 +337,20 @@ export class TerminalProcess extends Disposable implements ITerminalChildProcess
 			this._exitCode = e.exitCode;
 			this._queueProcessExit();
 		}));
-		this._sendProcessId(ptyProcess.pid);
+		// node-pty >= 1.2.0-beta.11 defers conptyNative.connect() on Windows, so
+		// ptyProcess.pid may be 0 immediately after spawn. In that case we wait
+		// for the first data event which only fires after the connection completes
+		// and the real pid is available. See microsoft/node-pty#885.
+		if (ptyProcess.pid > 0) {
+			this._sendProcessId(ptyProcess.pid);
+		} else {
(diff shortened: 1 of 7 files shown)


src/vs/platform/terminal/node/terminalProcess.ts
```diff
@@ -337,7 +337,20 @@ export class TerminalProcess extends Disposable implements ITerminalChildProcess
 			this._exitCode = e.exitCode;
 			this._queueProcessExit();
 		}));
-		this._sendProcessId(ptyProcess.pid);
+		// node-pty >= 1.2.0-beta.11 defers conptyNative.connect() on Windows, so
```
anthonykim1: was getting test failure so trying to see if these would fix it. 
I dont like how we are making pid as non-readonly though :/

anthonykim1: image

anthonykim1: @Tyriar With https://github.com/microsoft/node-pty/pull/885 (node-pty beta 11 which hasnt been in VS Code yet), 
node-pty defers `conptyNative.connect()`  . So instead of connecting immediately and having valid `pid`, we just get zero until worker is ready and connection is complete?

The thing is... that lead to test failing since 
https://github.com/microsoft/vscode/blob/970a8d3bb5d4df138cc28736dc26a58ade13f2ed/src/vs/platform/terminal/node/terminalProcess.ts#L340

Do you think https://github.com/microsoft/vscode/pull/301398/commits/27f480542e4142013bb2c9a8281cd6d0f5e6e3dc is the right way to go about this?

justschen (approved):

src/vs/platform/terminal/node/terminalProcess.ts
anthonykim1: @deepak1556 Wondering your thoughts here as well! Thanks

PR #301472, opened 2026-03-13
Fix incorrect inspect property usage in scope detection

### LINKED ISSUE #301471: Incorrect IInspectValue property used instead of scalar value in scope detection
## Bug

Two `IConfigurationValue` property accesses introduced in bb18007f use the `IInspectValue<T>` object property instead of the scalar value property, causing incorrect scope detection.

### 1. `themeConfiguration.ts` - `findAutoConfigurationTarget()`

[Line 380](https://github.com/microsoft/vscode/blob/bb18007f/src/vs/workbench/services/themes/common/themeConfiguration.ts#L380) checks `settings.userRemote` instead of `settings.userRemoteValue`:

```typescript
public findAutoConfigurationTarget(key: string) {
    const settings = this.configurationService.inspect(key);
    if (!types.isUndefined(settings.workspaceFolderValue)) {    // ✓ scalar
        return ConfigurationTarget.WORKSPACE_FOLDER;
    } else if (!types.isUndefined(settings.workspaceValue)) {   // ✓ scalar
        return ConfigurationTarget.WORKSPACE;
    } else if (!types.isUndefined(settings.userRemote)) {       // ✗ IInspectValue object
        return ConfigurationTarget.USER_REMOTE;
    }
    return ConfigurationTarget.USER;
}
```

`settings.userRemote` is an `IInspectValue<T>` object (`{value?, override?, overrides?}`). It can be defined when only language-specific overrides exist in the remote config, even if no actual remote value is set. This causes theme settings to incorrectly target `USER_REMOTE` scope.

### 2. `commandLineAutoApprover.ts` - source target detection

[Line 282](https://github.com/microsoft/vscode/blob/bb18007f/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/commandLineAnalyzer/autoApprove/commandLineAutoApprover.ts#L282) checks `configInspectValue.workspaceFolder` instead of `configInspectValue.workspaceFolderValue`:

```typescript
const sourceTarget = (
    checkTarget(configInspectValue.workspaceFolder) ? ConfigurationTarget.WORKSPACE_FOLDER
        : checkTarget(configInspectValue.workspaceValue) ? ConfigurationTarget.WORKSPACE
            : checkTarget(configInspectValue.userRemoteValue) ? ConfigurationTarget.USER_REMOTE
                ...
);
```

`configInspectValue.workspaceFolder` is an `IInspectValue<T>` whose keys are `{value, override, overrides}`. The `checkTarget` function looks for tool-pattern keys (e.g., `"git*"`) via `hasOwnProperty`, which never matches on `IInspectValue`. This means workspace-folder-scoped auto-approve rules are never detected as such and fall through to a lower-priority scope.

## Root Cause

Both are from the same commit (bb18007f) and follow the same pattern: the `IConfigurationValue<T>` interface exposes both scalar values (e.g., `userRemoteValue: T`) and inspect objects (e.g., `userRemote: IInspectValue<T>`). The similar naming makes it easy to pick the wrong one.

## Fix

- `settings.userRemote` -> `settings.userRemoteValue`
- `configInspectValue.workspaceFolder` -> `configInspectValue.workspaceFolderValue`

ShehabSherif0:
Fixes #301471

## Bug

Two `IConfigurationValue` property accesses introduced in bb18007f use the `IInspectValue<T>` object property instead of the scalar value property, causing incorrect scope detection.

### 1. `themeConfiguration.ts` - `findAutoConfigurationTarget()`

`settings.userRemote` is an `IInspectValue<T>` object (`{value?, override?, overrides?}`), not the scalar `T` value. It can be defined when only language-specific overrides exist in the remote config, even with no actual remote value set. This causes the function to incorrectly return `USER_REMOTE`, making theme settings target the wrong configuration scope.

**Fix:** `settings.userRemote` -> `settings.userRemoteValue`

### 2. `commandLineAutoApprover.ts` - source target detection

`configInspectValue.workspaceFolder` is an `IInspectValue<T>` object whose keys are `{value, override, overrides}`. The `checkTarget` function looks for tool-pattern keys (e.g., `"git*"`) via `hasOwnProperty`, which never matches on `IInspectValue`. Workspace-folder-scoped auto-approve rules are never detected and fall through to a lower-priority scope.

**Fix:** `configInspectValue.workspaceFolder` -> `configInspectValue.workspaceFolderValue`

### Root Cause

Both uses are from the same commit (bb18007f). The `IConfigurationValue<T>` interface exposes both scalar values (`userRemoteValue: T`) and inspect objects (`userRemote: IInspectValue<T>`). The similar naming makes it easy to use the wrong one.

### Testing

Added a regression test that stubs `configurationService.inspect` to return a `workspaceFolderValue` containing a rule key and asserts the matched rule reports `sourceTarget` as `ConfigurationTarget.WORKSPACE_FOLDER`.

### DIFF
--- a/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/commandLineAnalyzer/autoApprove/commandLineAutoApprover.ts
+++ b/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/commandLineAnalyzer/autoApprove/commandLineAutoApprover.ts
@@ -279,7 +279,7 @@ export class CommandLineAutoApprover extends Disposable {
 				);
 			}
 			const sourceTarget = (
-				checkTarget(configInspectValue.workspaceFolder) ? ConfigurationTarget.WORKSPACE_FOLDER
+				checkTarget(configInspectValue.workspaceFolderValue) ? ConfigurationTarget.WORKSPACE_FOLDER
 					: checkTarget(configInspectValue.workspaceValue) ? ConfigurationTarget.WORKSPACE
 						: checkTarget(configInspectValue.userRemoteValue) ? ConfigurationTarget.USER_REMOTE
 							: checkTarget(configInspectValue.userLocalValue) ? ConfigurationTarget.USER_LOCAL
--- a/src/vs/workbench/contrib/terminalContrib/chatAgentTools/test/browser/commandLineAutoApprover.test.ts
+++ b/src/vs/workbench/contrib/terminalContrib/chatAgentTools/test/browser/commandLineAutoApprover.test.ts
@@ -1244,7 +1244,7 @@ suite('CommandLineAutoApprover', () => {
 		});
 
 		test('should attribute workspace-folder-scoped rules to WORKSPACE_FOLDER target', async () => {
-			const workspaceFolderConfig = { 'git*': true };
+			const workspaceFolderConfig = { 'git': true };
 			configurationService.setUserConfiguration(TerminalChatAgentToolsSettingId.AutoApprove, workspaceFolderConfig);
 
 			const originalInspect = configurationService.inspect;
--- a/src/vs/workbench/contrib/terminalContrib/chatAgentTools/test/browser/commandLineAutoApprover.test.ts
+++ b/src/vs/workbench/contrib/terminalContrib/chatAgentTools/test/browser/commandLineAutoApprover.test.ts
@@ -1242,5 +1242,49 @@ suite('CommandLineAutoApprover', () => {
 			ok(!await isAutoApproved('echo hello'), 'Default rule should be ignored');
 			ok(!await isAutoApproved('cat file.txt'), 'Default rule should be ignored');
 		});
+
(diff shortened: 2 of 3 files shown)


meganrogge (dismissed):
Thanks! LGTM!

dmitrivMS (dismissed):

dmitrivMS (approved):

meganrogge (approved):

gjsjohnmurray:
What is preventing this from getting merged?

dmitrivMS (approved):

dmitrivMS:
@gjsjohnmurray Strange... it was stuck waiting for approvals while it should have had 2 already.

dmitrivMS:
Hmm... now it's stuck on the checks. @ShehabSherif0 if you merge from main and retrigger, perhaps it will help to clear those...

dmitrivMS:
Ok, doing Close/Reopen to retrigger checks.

dmitrivMS:
@gjsjohnmurray Merged!

PR #301585, opened 2026-03-13
Sessions - increase default width of the auxiliary sidebar

lszomoru:


bpasero (approved):
Just fyi, increasing it on the one side will make it look non balanced for the other side.

cc @benibenj @sandy081

PR #301598, opened 2026-03-13
Pass through the initial options on creation

mjbvz:
Make sure we pass these to callers in the new session flow too


TylerLeonhardt (approved):

bpasero:
@mjbvz this entirely breaks ability to open copilot sessions :-/

https://github.com/microsoft/vscode/issues/301677

PR #301614, opened 2026-03-13
Eliminate cpuUsage.sh/ps.sh in favor of direct /proc access

jakebailey:
One thing I've noticed for a while while watching `htop` is a `cpuUsage.sh` process appearing and disappearing over and over again, a few seconds apart, forever.

If I keep my system running for a long time, my PID count goes up and up and up. Right now my WSL has been running for 4 days, and the newest PID on my system is `2422090`, or 2,422,090 processes!

I think that VS Code doing this repeatedly in every window is the cause of this.

Notably, the main thing that repeatedly calls it is `childProcessMonitor.ts`, and it doesn't even care about getting the CPU usage. It just wants to know if there are child processes or not!

I don't think that these scripts are even required; all they do is call out to programs that themselves read `/proc`. VS Code can do those reads itself, so this PR does that.

And, then optimize for `childProcessMonitor`'s usecase and not gather more info than it needs to in order to return a boolean.


src/vs/base/node/ps.ts
```diff
@@ -254,3 +261,204 @@ function parsePsOutput(stdout: string, addToTree: (pid: number, ppid: number, cm
...
+			// Process may have exited
+		}
+	}
+
+	return false;
+}
+
+function hasChildProcessesDarwin(pid: number, ignoreNames?: string[]): Promise<boolean> {
+	return new Promise<boolean>(resolve => {
+		exec(`pgrep -P ${pid}`, (err, stdout) => {
```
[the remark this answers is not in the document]
jakebailey: This is preexisting, though I agree, `exec` is very much to be avoided.

src/vs/base/node/ps.ts
```diff
@@ -254,3 +240,190 @@ function parsePsOutput(stdout: string, addToTree: (pid: number, ppid: number, cm
...
+			} catch {
+				// Process may have exited
+			}
+
+			addToTree(pid, ppid, cmd, 0, pmem);
+		} catch {
+			// Process may have exited
+		}
+	}
+}
```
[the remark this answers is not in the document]
jakebailey: I was debating this too. There's a tradeoff here, given the async versions have an overhead just from the promises.

In a perfect world, I'd do the sync calls, then throw in a few yields after a set number of iterations.

src/vs/base/node/ps.ts
```diff
@@ -254,3 +238,205 @@ function parsePsOutput(stdout: string, addToTree: (pid: number, ppid: number, cm
 		}
 	}
 }
+
+async function readProcessesFromProc(addToTree: (pid: number, ppid: number, cmd: string, load: number, mem: number) => void): Promise<void> {
+	const totalMemKB = readTotalMemoryKB();
+	const entries = (await fs.readdir('/proc')).filter(e => /^\d+$/.test(e));
```
anthonykim1: Wondering if `addToTree` would drop entry if parent isn't already in the map `const map = new Map<number, ProcessItem>();` 
For the old `ps -ax` it sorts output by PID, but not sure if `readdir('proc')` also does this sorting.

src/vs/base/node/ps.ts
```diff
@@ -254,3 +238,205 @@ function parsePsOutput(stdout: string, addToTree: (pid: number, ppid: number, cm
...
+		return hasChildProcessesMacOS(pid, ignoreNames);
+	}
+
+	// Windows: use the native module via listProcesses
+	return listProcesses(pid).then(item => {
+		const children = item.children ?? [];
+		if (children.length === 0) {
+			return false;
+		}
+		if (ignoreNames && ignoreNames.length > 0) {
```
anthonykim1, line 341: Should windows also have something similar to:
```
} else if (childCount > 1) {
						return true;
					}
```

src/vs/base/node/ps.ts
```diff
@@ -159,87 +162,68 @@ export function listProcesses(rootPid: number): Promise<ProcessItem> {
...
+						try {
+							const totalAfter = readTotalCpuTime();
+							const totalDelta = totalAfter - totalBefore;
+							if (totalDelta > 0) {
+								for (const pid of pids) {
+									const processInfo = map.get(pid);
+									if (processInfo) {
+										const before = processBefore.get(pid) ?? 0;
+										const after = readProcessCpuTime(pid);
+										processInfo.load = Math.round((100 * (after - before)) / totalDelta);
```
anthonykim1: If `readProcessCpuTime` return 0 on ENOENT, and the process exit during 500ms window, would we end up with negative load?

anthonykim1:
Thanks for tackling this!! Removing repeated script spawning does look promising. 
Briefly discussed this with @meganrogge that we should test this out on all OS.
I'm wondering how we could best test this though.

src/vs/base/node/ps.ts
jakebailey: Probably; I think `Math.abs`-ing this should be a reasonable number since we could at worst say "it was doing CPU for half the time".

src/vs/base/node/ps.ts
jakebailey: I don't think so; the other paths do this for perf, but since Windows still uses `windows-process-tree`, it doesn't need a special case

src/vs/base/node/ps.ts
jakebailey: I don't think it's guaranteed; added sorting.

src/vs/base/node/ps.ts
jakebailey: Eh, just went for clamping it to 0.

jakebailey:
@anthonykim1 do you have any remaining thoughts here at all? I don't think this PR can be merged due to the rule that these build things not be sent by non-team members, and I unfortunately don't count.

jakebailey:
There is also a related change I have unpushed to remove `terminateProcess.sh`, the only remaining shell script that isn't just for shell integration.

PR #301622, opened 2026-03-13
Browser button in title bar

### LINKED ISSUE #289413: Launching Browser from IDE
As a developer I should be able to click a button to open up the browser.
Note: that this location is impacted based on layout 

Tooltip: Open New Browser

## Activity Bar position Default:
The globe icon should sit to the left of the toggles, with a seperator.

Image

## Activity Bar position Top:
The globe icon should sit to the left of the toggles, with a seperator, followed by settings and profile icon.

Image

kycutler:
Closes #289413

image

### DIFF
--- a/src/vs/workbench/contrib/browserView/electron-browser/browserView.contribution.ts
+++ b/src/vs/workbench/contrib/browserView/electron-browser/browserView.contribution.ts
@@ -175,6 +175,15 @@ registerSingleton(IBrowserZoomService, BrowserZoomService, InstantiationType.Del
 Registry.as<IConfigurationRegistry>(ConfigurationExtensions.Configuration).registerConfiguration({
 	...workbenchConfigurationNodeBase,
 	properties: {
+		'workbench.browser.showInTitleBar': {
+			type: 'boolean',
+			default: false,
+			experiment: { mode: 'startup' },
+			description: localize(
+				{ comment: ['This is the description for a setting.'], key: 'browser.showInTitleBar' },
+				'Controls whether the Integrated Browser button is shown in the title bar.'
+			)
+		},
 		'workbench.browser.openLocalhostLinks': {
 			type: 'boolean',
 			default: false,
--- a/src/vs/workbench/contrib/browserView/electron-browser/browserViewActions.ts
+++ b/src/vs/workbench/contrib/browserView/electron-browser/browserViewActions.ts
@@ -3,7 +3,7 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-import { localize2 } from '../../../../nls.js';
+import { localize, localize2 } from '../../../../nls.js';
 import { ContextKeyExpr } from '../../../../platform/contextkey/common/contextkey.js';
 import { Action2, registerAction2, MenuId } from '../../../../platform/actions/common/actions.js';
 import { ServicesAccessor } from '../../../../platform/instantiation/common/instantiation.js';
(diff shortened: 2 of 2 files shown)


jruales (approved):

src/vs/workbench/contrib/browserView/electron-browser/browserViewActions.ts
```diff
@@ -3,7 +3,7 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-import { localize2 } from '../../../../nls.js';
+import { localize, localize2 } from '../../../../nls.js';
```
jruales, line 6: why do we use both localize and localize2?

src/vs/workbench/contrib/browserView/electron-browser/browserViewActions.ts
```diff
@@ -733,3 +741,9 @@ registerAction2(ShowBrowserFindAction);
 registerAction2(HideBrowserFindAction);
 registerAction2(BrowserFindNextAction);
 registerAction2(BrowserFindPreviousAction);
+
+registerAction2(class ToggleBrowserTitleBarButton extends ToggleTitleBarConfigAction {
```
jruales, line 745: Why inline the action here instead of saving it as a variable like the other action registrations?

TylerLeonhardt (approved):

src/vs/workbench/contrib/browserView/electron-browser/browserView.contribution.ts
```diff
@@ -175,6 +175,15 @@ registerSingleton(IBrowserZoomService, BrowserZoomService, InstantiationType.Del
 Registry.as<IConfigurationRegistry>(ConfigurationExtensions.Configuration).registerConfiguration({
 	...workbenchConfigurationNodeBase,
 	properties: {
+		'workbench.browser.showInTitleBar': {
+			type: 'boolean',
+			default: false,
+			experiment: { mode: 'startup' },
+			description: localize(
+				{ comment: ['This is the description for a setting.'], key: 'browser.showInTitleBar' },
+				'Controls whether the Integrated Browser button is shown in the title bar.'
```
jruales, line 184: should we enable this setting for ExP?

kycutler: It is! `experiment: { mode: 'startup' }`. It's just not marked as experimental in the UI.

jruales: Ah! Didn't see that

src/vs/workbench/contrib/browserView/electron-browser/browserViewActions.ts
kycutler: Different APIs 🙃 the `ToggleTitleBarConfigAction` helper takes a plain `string` which comes from `localize` and not `localize2`. Don't ask me the reason for this but I'm sure it's historical

src/vs/workbench/contrib/browserView/electron-browser/browserViewActions.ts
kycutler: I've gone back and forth on it... On the one hand it's not a primary action like the others -- just a little nice-to-have to match with other title bar actions, so to me it doesn't make much sense to have its own variable. It would probably be better to have near the actual open command though -- but then the `registerAction2` call moves way up?

Anyway at some point I gave up trying to decide 😄

jruales: Gotcha. I think I would have just defined the class right after OpenIntegratedBrowserAction and referenced it from down here. But I think it's fine as is too, since it is a bit different from the others in that it extends another action instead of just the base Action object. Up to you

PR #301639, opened 2026-03-14
Add directory listing to protocol, and folder picker for remote agent hosts

roblourens:


src/vs/platform/agentHost/common/state/sessionProtocol.ts
```diff
@@ -199,6 +200,22 @@ export interface IFetchContentResult {
...
 
+// ---- Filesystem browsing ----------------------------------------------------
+
+export interface IBrowseDirectoryParams {
+	readonly uri: URI;
+}
+
+export interface IDirectoryEntry {
+	readonly name: string;
+	readonly uri: URI;
```
roblourens: Oh this maybe shouldn't have a uri

src/vs/platform/agentHost/node/agentSideEffects.ts
```diff
@@ -225,6 +228,32 @@ export class AgentSideEffects extends Disposable implements IProtocolSideEffectH
...
+		try {
+			dirents = await fs.promises.readdir(dirPath, { withFileTypes: true });
+		} catch (err) {
+			const code = (err as NodeJS.ErrnoException).code;
+			if (code === 'ENOENT' || code === 'ENOTDIR') {
+				return { entries: [] };
+			}
+			throw err;
+		}
+		const visible = dirents.filter(d => !d.name.startsWith('.'));
```
connor4312: I don't think we'd care about hiding dotfiles on the protocol like this?

PR #301713, opened 2026-03-14
Fix start marker recreation losing early terminal output

sanchirico:
When a command produces enough output to overflow the scrollback the xterm start marker gets disposed and `setupRecreatingStartMarker` recreates it at the current cursor which is past the output that already scrolled out. This means `getContentsAsText` captures only the tail fragment.

This prevents start marker recreation from moving the boundary forward past where it originally was and adds a fallback that picks the earliest available marker (from command detection or the executed marker) when fetching output.


otbotb371-lgtm (commented):
Gen@dna.net

otbotb371-lgtm (commented):
gen@dna.org

PR #301896, opened 2026-03-15
Fix listener leak in chat inline anchor widgets

bryanchen-d:
Move per-widget `configurationService.onDidChangeConfiguration` listener to a single listener on the `ChatListWidget` container. Each `InlineAnchorWidget` previously registered its own listener to toggle a link-style CSS class, causing a listener leak warning when 175+ widgets existed in a session.

The fix uses a parent container class (`chat-inline-references-link-style`) toggled by a single listener, with CSS descendant selectors to style all child widgets.

Fixes #301185


roblourens (approved):
Thanks! Not a leak per se but a listener leak warning false alarm.

PR #302036, opened 2026-03-16
fix: prevent dirty file tracker from auto-opening chat-edited files (#298700)

### LINKED ISSUE #298700: Opening agent created/modified in the editor is overwhelming
Type: <b>Feature Request</b>

I would like that files that created or modified by the agent are not automatically opened in the editor. When creating a new app this will fill the editor with many opened tabs that then need to be closed. 
Here is a snapshot from my session

Image

When I want to inspect a created/modified file I can easily do this from the link in the chat.

Extension version: 0.38.2026030203
VS Code version: Code - Insiders 1.110.0-insider (f5927e727c69039d3037f018f1790ebbbee596bc, 2026-02-28T02:04:19.045Z)
OS version: Darwin arm64 25.3.0
Modes:

kangarko:
## Problem

When the chat agent edits files, those models become dirty while edits stream in. VS Code has generic dirty-model auto-open paths for both text files and notebooks:

- `TextFileEditorTracker` for dirty text files
- `NotebookEditorManager` for dirty notebooks

Without an explicit suppression, both can auto-open editors even when `accessibility.openChatEditedFiles` is `false`.

My original fix addressed the behavior directly from `TextFileEditorTracker`, but that introduced a `files -> chat` dependency and broke layers.

## Updated fix

This PR now moves the suppression to a generic lower-level seam:

- `IFilesConfigurationService` exposes a per-resource runtime override to suppress opening an editor for a dirty resource.
- `TextFileEditorTracker` consults that generic override for text files.
- `NotebookEditorManager` consults the same generic override for notebooks.
- Chat editing registers the override from its modified-entry implementations while the entry is chat-owned and keeps it until the resource becomes clean, which avoids the debounce race after a fast accept/save.

## Why this should fit better

- no `files -> chat` dependency
- no `notebook -> chat` special casing beyond consuming a generic service override
- keeps the dirty-editor auto-open logic generic
- follows the same disposable override pattern chat already uses with `disableAutoSave(...)`
- covers both text documents and notebooks, matching the actual dirty-model entry points in the codebase

## Files changed

- `src/vs/workbench/services/filesConfiguration/common/filesConfigurationService.ts`
- `src/vs/workbench/contrib/files/browser/editors/textFileEditorTracker.ts`
- `src/vs/workbench/contrib/notebook/browser/notebook.contribution.ts`
- `src/vs/workbench/contrib/chat/browser/chatEditing/chatEditingModifiedFileEntry.ts`
- `src/vs/workbench/contrib/chat/browser/chatEditing/chatEditingModifiedDocumentEntry.ts`
- `src/vs/workbench/contrib/chat/browser/chatEditing/chatEditingModifiedNotebookEntry.ts`
- `src/vs/workbench/contrib/files/test/browser/textFileEditorTracker.test.ts`
- `src/vs/workbench/test/common/workbenchTestServices.ts`
- `src/vs/workbench/contrib/chat/browser/actions/chatAccessibilityHelp.ts`

Fixes #298700

cc @bpasero @roblourens @jrieken @meganrogge


bpasero (changes requested):
No, breaks layers.

src/vs/workbench/contrib/chat/browser/chatEditing/chatEditingModifiedDocumentEntry.ts
```diff
@@ -179,6 +179,36 @@ export class ChatEditingModifiedDocumentEntry extends AbstractChatEditingModifie
...
+		this._register(this._textFileService.files.onDidChangeDirty(model => {
+			if (isEqual(model.resource, this.modifiedURI)) {
+				updateOpenEditorWhenDirtyDisabled();
+			}
+		}));
+		this._register(this._textFileService.untitled.onDidChangeDirty(model => {
+			if (isEqual(model.resource, this.modifiedURI)) {
+				updateOpenEditorWhenDirtyDisabled();
+			}
+		}));
```
jrieken: There is also `src/vs/workbench/contrib/chat/browser/chatEditing/chatEditingModifiedNotebookEntry.ts` and whatever is done here for documents should also work for notebooks

jrieken (changes requested):

Issue #301294, opened 2026-03-12
\ escapes in terminal command


roblourens:
I keep seeing these backslashes in terminal commands. They show up for a second then disappear.

Here I have two parallel terminal commands, one with a confirmation, so the second one was stuck showing the backslash until I clicked allow on the first.

<img width="672" height="401" alt="Image" src="https://github.com/user-attachments/assets/44d5b2d8-1358-41e8-bc61-6d132b1898ce" />

```
## Response
### Assistant
~~~md
The target clone already has stashes, so I’m resolving the source stash commit explicitly and then I’ll store it as a new top entry in ../vscode without changing its worktree.
🛠️ run_in_terminal (call_zYAXUrEt9FVFXXd3AKG2NpLM) {
  "command": "git -C /Users/roblou/code/vscode2 rev-parse 'stash@{0}'",
  "explanation": "Resolve the exact commit SHA for the top stash in the source clone using a quoted ref to avoid shell expansion.",
  "goal": "Resolve stash SHA",
  "isBackground": false,
  "timeout": 10000
}
🛠️ run_in_terminal (call_T1VJp8p8HWa7BNKlYMX9eYoi) {
  "command": "git -C /Users/roblou/code/vscode2 log -1 --format=%gs 'stash@{0}'",
  "explanation": "Read the stash reflog subject so I can preserve the same message when storing it in the target clone.",
  "goal": "Resolve stash message",
  "isBackground": false,
  "timeout": 10000
}
~~~
```

<img width="448" height="125" alt="Image" src="https://github.com/user-attachments/assets/2504057b-51e3-4496-b69e-4e589a0e38da" />

PR #302235, opened 2026-03-16
release fix for escaping markdown in thinking headers

justschen:
fix https://github.com/microsoft/vscode/issues/301294


rebornix (approved):
This is a regression for the header rendering for terminal tools.

PR #302290, opened 2026-03-16
Add required-only modal mode for editor overlays

joshspicer:
Lets us use modal for Chat Customizations while it may not be ready for other, existing editors.

## Summary
- add a new `workbench.editor.useModal` mode: `required-only`
- make default `product.quality !== 'stable' ? 'some' : 'required-only'`
- add `EditorInputCapabilities.RequiresModal` and gate `MODAL_GROUP` usage in `required-only`
- mark Chat Customizations editor input as `RequiresModal`
- keep `some`/`all` behavior as supersets and add config migration from `essential-only` to `required-only`

## Validation


mjbvz:
@joshspicer I am moving this to 1.113 since it is targeting main. Please bring it in as a candidate if needed

bpasero (dismissed):
I do not see the advantage of this over using `MODAL_GROUP` when opening the editor. What problem is this trying to solve?

PR #302317, opened 2026-03-17
fix: prevent duplicate tip nodes on re-entrant renderGettingStartedTipIfNeeded

sathvikc:
## Summary

- Fix duplicate getting-started tip nodes in chat widget caused by re-entrant calls to `renderGettingStartedTipIfNeeded`
- Move the `_gettingStartedTipPart` guard assignment before the DOM append so re-entrant calls see the guard and return early
- Add `dom.clearNode(tipContainer)` before append as a defensive measure against stale nodes

## Root Cause

When `ChatTipContentPart` is constructed, context-key changes can synchronously trigger another call to `renderGettingStartedTipIfNeeded`. The guard (`this._gettingStartedTipPart.value`) was set **after** `tipContainer.appendChild()`, so the re-entrant call would pass the guard check, create a second tip, and append it -producing the "multiple identical tips stacked" symptom.

## Fix

Reorder the method so that:
1. The guard (`this._gettingStartedTipPart.value = store`) is set before any DOM mutation
2. `dom.clearNode(tipContainer)` clears any stale nodes before appending the new tip

IMG_4445


meganrogge (approved):
Thanks! Saw one report of this, but wasn't able to repro.

bpasero (approved):

PR #302334, opened 2026-03-17
Substitute `url.parse` with WHATWG `URL` API

kkocdko:
Fix #301941 . The `url.parse` from `node:url` causes deprecated warning

### DIFF
--- a/src/vs/platform/request/node/proxy.ts
+++ b/src/vs/platform/request/node/proxy.ts
@@ -3,12 +3,12 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-import { parse as parseUrl, Url } from 'url';
+import { urlToHttpOptions } from 'url';
 import { isBoolean } from '../../../base/common/types.js';
 
 export type Agent = any;
 
-function getSystemProxyURI(requestURL: Url, env: typeof process.env): string | null {
+function getSystemProxyURI(requestURL: URL, env: typeof process.env): string | null {
 	if (requestURL.protocol === 'http:') {
 		return env.HTTP_PROXY || env.http_proxy || null;
 	} else if (requestURL.protocol === 'https:') {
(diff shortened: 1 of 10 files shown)


src/vs/platform/request/node/proxy.ts
```diff
@@ -24,23 +24,23 @@ export interface IOptions {
...
+	const proxyEndpoint = new URL(proxyURL);
 
-	if (!/^https?:$/.test(proxyEndpoint.protocol || '')) {
+	if (proxyEndpoint.protocol !== "http:" && proxyEndpoint.protocol !== "https:") {
 		return null;
 	}
 
 	const opts = {
-		host: proxyEndpoint.hostname || '',
-		port: (proxyEndpoint.port ? +proxyEndpoint.port : 0) || (proxyEndpoint.protocol === 'https' ? 443 : 80),
```
kkocdko, line 42: even Node's url.Url.protocol has `:` suffix, previous code have bug here

src/vs/server/node/remoteExtensionHostAgentServer.ts
```diff
@@ -115,7 +114,7 @@ class RemoteExtensionHostAgentServer extends Disposable implements IServerAPI {
 			return serveError(req, res, 400, `Bad request.`);
 		}
 
-		const parsedUrl = url.parse(req.url, true);
+		const parsedUrl = new URL(req.url, 'http://localhost');
```
kkocdko, line 117: `new URL` without base will throw error, so use localhost here as placeholder

src/vs/platform/request/node/proxy.ts
```diff
@@ -3,12 +3,12 @@
...
  *--------------------------------------------------------------------------------------------*/
 
-import { parse as parseUrl, Url } from 'url';
+import { urlToHttpOptions } from 'url';
 import { isBoolean } from '../../../base/common/types.js';
 
 export type Agent = any;
 
-function getSystemProxyURI(requestURL: Url, env: typeof process.env): string | null {
+function getSystemProxyURI(requestURL: URL, env: typeof process.env): string | null {
```
kkocdko, line 11: Currently people use `undici` to replace old `http(s)-proxy-agent`, however, the Node's [built-in-proxy-support](https://nodejs.org/api/http.html#built-in-proxy-support) is on the road, so keep these code and waiting for new one.

kkocdko:
hello, anyone here?

rgoldberg:
@chrmarti The warning that this attempts to fix mentions security implications, which I would expect would be of high priority.

What is required to move this forward?

Thanks.

randomir:
Bump.

Issue #170101, opened 2022-12-27
npm.packageManager=auto incorrectly assumes yarn as package manager for running tasks


Zelgadis87:
Type: <b>Bug</b>

I have a fairly standard PNPM project, for which I have configured a few VSCode tasks:
```
{
	"version": "2.0.0",
	"tasks": [
		{
			"label": "bar",
			"type": "npm",
			"script": "bar"
		}
	]
}
```
When VSCode is set with `npm.packageManager=auto`, it incorrectly assumes that tasks should be run using `yarn`.
Since the project uses [corepack ](https://nodejs.org/api/corepack.html) and specifies a packageManager in `package.json`, using yarn results in an error:

```
* Executing task in folder foo: yarn run bar 
Usage Error: This project is configured to use pnpm
$ yarn ...
 *  The terminal process "cmd.exe /d /c yarn run bar" terminated with exit code: 1. 
```

This for me doesn't make much sense, because:
- in the project folder there is a `package.json`, configured with `"packageManager": "pnpm@7.9.3"` (using corepack)
- in the project folder there is a `pnpm-lock.yaml` and a `pnpm-workspace.yaml` files, which should correctly hint to the right package manager
- there is no trace of a `yarn.lock` or any other yarn-specific files or configurations.
- yarn is installed globally (via corepack shims) but so is npm and pnpm
 
If the auto detection for PNPM failed, I'd expect to at least fallback to NPM.
Using `npm.packageManager=pnpm` of course works fine.

VS Code version: Code - Insiders 1.75.0-insider (6ed4d436a9b6ae131732f4e2a723868231f4473a, 2022-12-23T05:21:57.802Z)
OS version: Windows_NT x64 10.0.19045
Modes:
Sandboxed: Yes

<details>
<summary>System Info</summary>

|Item|Value|
|---|---|
|CPUs|Intel(R) Core(TM) i5-7600 CPU @ 3.50GHz (4 x 3504)|
[...]

Issue #159465, opened 2022-08-29
VSCode uses yarn to run npm tasks


amerrit2:
Does this issue occur when all extensions are disabled?: Yes

- VS Code Version: 1.70.2
- OS Version: Windows_NT x54 10.0.19043

Steps to Reproduce:

1. Create npm project with `npm --init`
1. Add watch script in package.json (simply `echo "watch"` should be enough)
1. Create `tasks.json` :
```
{
	"version": "2.0.0",
	"tasks": [
		{
			"type": "npm",
			"script": "watch",
			"group": "build",
			"problemMatcher": ["$tsc-watch"],
			"label": "npm: watch",
			"detail": "tsc --build --watch",
			"isBackground": true,
			"runOptions": {"runOn": "folderOpen"}
		}
	]
}
```

2.  Reopen folder.

```
 *  Executing task: yarn run watch 

/usr/bin/bash: line 1: yarn: command not found

 *  The terminal process "C:\Program Files\Git\bin\bash.exe '-c', 'yarn run watch'" terminated with exit code: 127. 
 *  Terminal will be reused by tasks, press any key to close it. 
```

I've never used yarn and there is no yarn installation anywhere close to my computer.  Yet I cannot get npm to auto-run my watch on open in any of my projects because of this bug.

meganrogge:
what's the watch script in your package.json?

amerrit2:
`"watch": "tsc --build --watch",		`

meganrogge:
tasks of type `npm` get resolved via the npm extension. I think what you want is 
```
{
			"type": "shell",
			"command": "tsc --build --watch",
.... other properties
},
```

amerrit2:
I tried with `--disable-extensions` and it did the same thing

meganrogge:
did the above help you?

amerrit2:
Yes it did. But the 'npm' type is still broken though, right?

meganrogge:
is this a new issue IE has that exact task ever worked for you?

amerrit2:
Last time I remember it working was something like 1 year ago. Maybe less.
It definitely used to work though.

meganrogge:
@alexr00 any context on why this would've regressed or changed by design?

alexr00:
@amerrit2 what is the value of the setting `npm.packageManager`?

amerrit2:
[...]

Issue #261630, opened 2025-08-14
`npm.scriptRunner` auto-detection fails in subdirectories; defaults to npm instead of pnpm/yarn/bun


Unit2795:
Does this issue occur when all extensions are disabled?: Yes

### VSCode:
- Version: 1.103.1 (user setup)
- Commit: 360a4e4fd251bfce169a4ddf857c7d25d1ad40da
- Date: 2025-08-12T16:25:40.542Z
- Electron: 37.2.3
- ElectronBuildId: 12035395
- Chromium: 138.0.7204.100
- Node.js: 22.17.0
- V8: 13.8.500258-electron.0
- OS: Windows_NT x64 10.0.26100

### OS:
- Edition	Windows 11 Pro
- Version	24H2
- Installed on	12/6/2024
- OS build	26100.4770
- Experience	Windows Feature Experience Pack 1000.26100.197.0

## Steps to Reproduce:
Minimum Reproduction Repo: https://github.com/Unit2795/scriptrunner-bug

1. Create a directory with a subdirectory inside of it.
2. Create a package.json in the subdirectory (manually or with `pnpm init`)
3. Install a dependency using pnpm (or any other package manager besides NPM) to create a `pnpm-lock.yaml` file
4. Try running a "scripts" entry in the `package.json` using the Explorer's NPM Scripts view in VSCode. Note that it uses npm and not pnpm/yarn/bun

DonJayamanne:
> in the package.json using the Explorer's NPM Scripts view in VS

Thank you for filing this issue and sorry you are running into this.
Unfortunately VS Code does not ship with an npm explorer view.
I believe this is caused by an extension (easy way to verify is to disable all extensions and reload, the see if the view is still visible).

Closing as its not caused by VS Code, please file an issue against the relevant extension.

Unit2795:
@DonJayamanne
There is an NPM Scripts Explorer View. It is not an extension. It is present with a fresh VSCode install and launching with "code --disable-extensions". I do not have any extensions on. It's part of VSCode itself, please reopen this issue.

Added in this update:
- https://code.visualstudio.com/updates/v1_37#_npm-scripts-explorer-visible-by-default

More info:
- https://code.visualstudio.com/docs/getstarted/tips-and-tricks#_run-npm-scripts-as-tasks-from-the-explorer
- https://www.tompka.io/posts/vscode-npm-script-explorer

Image:
[...]

PR #302354, opened 2026-03-17
fix(npm): prioritize packageManager field and simplify lock file dete…

flycran:
## Summary

The `findPreferredPM` function in the npm extension previously ignored the
`packageManager` field in `package.json`, which is the most authoritative
way to declare the intended package manager per the Node.js Corepack spec.

Additionally, the previous logic relied on `find-yarn-workspace-root`, which
detects the presence of a `workspaces` field in `package.json` and incorrectly
treats it as evidence of Yarn usage — regardless of the actual package manager
in use. This caused unexpected behavior reported in #170101 and #159465.

This PR refactors the detection logic with the following changes:

- **Read `packageManager` first**: If `package.json` declares a
  `packageManager` field (e.g. `"packageManager": "pnpm@8.6.0"`), it is
  used as the authoritative source for the package manager name.
- **Unified lock file table**: Replace four separate per-PM detection
  functions with a single `LOCKFILE_CANDIDATES` table, iterating in
  priority order and returning early as soon as two distinct lock files
  are detected.
- **Fix `multipleLockFilesDetected` semantics**: Previously, `whichPM`
  results could incorrectly inflate the lock file count. Now
  `multipleLockFilesDetected` only reflects actual lock file conflicts,
  consistent with the warning message shown to the user in `tasks.ts`.
- **Remove unused dependency**: `find-yarn-workspace-root` is no longer
  used and has been removed from `package.json`.

## Testing

- Compiled the extension via `npx gulp compile-extension:npm` — no errors.
- Opened a workspace and ran npm scripts with different `packageManager`
  values in `package.json` (e.g. `"pnpm@9.0.0"`, `"yarn@4.0.0"`), confirmed
  the correct package manager was invoked each time.
- Added a `workspaces` field to `package.json` and verified it no longer
  incorrectly influences package manager detection.
- Placed different lock files in the workspace root, reloaded the window,
  ran npm scripts, and confirmed the correct package manager was selected
  based on the detected lock file.

### DIFF
--- a/extensions/npm/package-lock.json
+++ b/extensions/npm/package-lock.json
@@ -10,7 +10,6 @@
       "license": "MIT",
       "dependencies": {
         "find-up": "^5.0.0",
-        "find-yarn-workspace-root": "^2.0.0",
         "jsonc-parser": "^3.2.0",
         "minimatch": "^5.1.8",
         "request-light": "^0.7.0",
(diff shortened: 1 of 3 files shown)


flycran:

flycran:
#279150、#261630

Issue #302393, opened 2026-03-17
Allow tool confirmations for specific arguments


alexr00:
For example, we have a "Run VS Code Command" tool, which takes a VS Code command + command args as arguments. Users should be able to approve running this specific combination of arguments.

rgrunber:
> return true;// extension.enabledApiProposals.includes(proposal);

(https://github.com/microsoft/vscode/pull/302395/changes#diff-f9517e3a978b76f41a33822f7a432be19c2e3aa13c6e9184db8424b685002467R327)

@alexr00 , was this intended only temporarily ? I noticed that if I manually installed a vsix and didn't include it in the `enable-proposed-api` setting for `argv.json`, or using `--enable-proposed-api`, I would see the `Extension .. CANNOT USE these API proposals '..'.` , but the api would run anyways. It seems to still be there https://github.com/microsoft/vscode/blob/main/src/vs/workbench/services/extensions/common/extensions.ts#L327 .

The behaviour can definitely be seen going from 1.112.0 to 1.113.0.

alexr00:
Yes, thank you for catching. PR fixing it incoming.

alexr00:
@rgrunber, thank you for pointing the API proposal enforcement out like this, giving us an opportunity to fix it. As a small thank you, we want to offer you some swag. Our new swag system requires that we use your email address. I know you don't have your email in your GitHub profile, but I did find it associated with some of your commits. I can use that email for the swag link, or would you be willing to share your email?

rgrunber:
Hey! Yes, "rgrunber@redhat.com" would be my email.

Part of the reason I found myself looking at this was the [`resolvers`](https://github.com/microsoft/vscode/blob/main/src/vscode-dts/vscode.proposed.resolvers.d.ts) API. I'm sure there's higher priority things on the roadmap, but if there were any way to give some consideration to making it mature (non-proposed) API, or listing any existing roadblocks, I think that would be great.

PR #302395, opened 2026-03-17
Proposal for tool combination approval

alexr00:
Adds support for approving a specific tool + args combination Part of #302393

### DIFF
--- a/src/vs/workbench/contrib/chat/browser/tools/languageModelToolsConfirmationService.ts
+++ b/src/vs/workbench/contrib/chat/browser/tools/languageModelToolsConfirmationService.ts
@@ -4,7 +4,6 @@
  *--------------------------------------------------------------------------------------------*/
 
 import { Codicon } from '../../../../../base/common/codicons.js';
-import { hash } from '../../../../../base/common/hash.js';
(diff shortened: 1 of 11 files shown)


src/vs/workbench/contrib/chat/browser/tools/languageModelToolsConfirmationService.ts
```diff
@@ -19,11 +20,21 @@ import { IToolData, ToolDataSource } from '../../common/tools/languageModelTools
...
+function computeCombinationKey(toolId: string, parameters: unknown): string {
+	return toolId + ':combination:' + hash(parameters);
+}
+
 
 class GenericConfirmStore extends Disposable {
 	private _workspaceStore: Lazy<ToolConfirmStore>;
 	private _profileStore: Lazy<ToolConfirmStore>;
-	private _memoryStore = new Set<string>();
+	private _memoryStore = new Map<string, string | boolean>();
```
connor4312: I would prefer switching `string | boolean` to explicit interface to be more descriptive

lszomoru (approved):

PR #302413, opened 2026-03-17
feat: support opening folders on Mac with vscode

ulugbekna:
Add **Install 'Open with Code' in Finder** and **Uninstall 'Open with Code' from Finder** commands under the Shell Command category (macOS only).

These commands install/uninstall a macOS Automator Quick Action workflow to \`~/Library/Services/\` that adds an 'Open with {productName}' entry to the Finder right-click context menu (Quick Actions / Services).

The workflow runs \`open -a "{appName}" "$@"\` to open selected files/folders in VS Code.

Each VS Code build (Stable, Insiders, OSS) creates a distinctly-named workflow, so multiple installations coexist without conflict.

### How to use
1. Open VS Code
2. Command Palette → **Shell Command: Install 'Open with Code' in Finder**
3. Right-click a folder in Finder → Quick Actions / Services → **Open with Code**

### Files changed
- \`src/vs/platform/native/common/native.ts\` — added \`installFinderAction\` / \`uninstallFinderAction\` to interface
- \`src/vs/platform/native/electron-main/nativeHostMainService.ts\` — implementation (generates Automator .workflow bundle)
- \`src/vs/workbench/electron-browser/actions/installActions.ts\` — action classes
- \`src/vs/workbench/electron-browser/desktop.contribution.ts\` — registration
- \`src/vs/workbench/test/electron-browser/workbenchTestServices.ts\` — test stubs

### Demo 

https://github.com/user-attachments/assets/48e1ac49-9258-4d48-b4d9-e476a8036918

### DIFF
--- a/src/vs/platform/native/common/native.ts
+++ b/src/vs/platform/native/common/native.ts
@@ -224,6 +224,10 @@ export interface ICommonNativeHostService {
 	installShellCommand(): Promise<void>;
 	uninstallShellCommand(): Promise<void>;
 
+	// macOS Finder action
+	installFinderAction(): Promise<void>;
+	uninstallFinderAction(): Promise<void>;
+
 	// Lifecycle
 	notifyReady(): Promise<void>;
 	relaunch(options?: { addArgs?: string[]; removeArgs?: string[] }): Promise<void>;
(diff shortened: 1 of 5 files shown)


bpasero (changes requested):
This is nice, but can you clean the diff to be minimal, I find it hard to review, there seems to be unrelated changes around import sort order and some on did blur event that moved.

deepak1556 (changes requested):
Could we have a native module handle the service callback, the module can register itself as a [service provider](https://developer.apple.com/documentation/appkit/nsapplication/servicesprovider?language=objc), the service provider can also implement [menu validation](https://developer.apple.com/documentation/appkit/nsmenuitemvalidation/validatemenuitem(_:)?language=objc) that would either enable or disable the menu. This also has the nice benefit of having the service menu declaration as part of our app bundles info.plist.

PR #302472, opened 2026-03-17
Fix chat content part resource retention

vikeychen:
Fix two remaining sources of chat content-part listener/resource retention.

- move `ChatCollapsibleInputOutputContentPart` title file widgets into their own disposable store so changing the title disposes the previous inline anchor/widget registrations before re-rendering
- cap the number of idle chat code block editors retained by `EditorPool` / `DiffEditorPool` and fully reset released editors so old toolbar, model, label and vulnerability state is not kept alive indefinitely
- add tests covering the title widget re-render disposal path and the `ResourcePool` idle-cap behavior

This follows up on the inline anchor leak fixes already merged separately and addresses the remaining listener growth I was still seeing from tool progress parts and pooled code block editors.


roblourens:
Thanks for the PR, but the first part in ResourcePool, I actually independently made the same change awhile back. The `ChatCollapsibleInputOutputContentPart` part, I don't totally understand because we were only rendering file widgets once in the constructor and they seemed tracked correctly. The change in `reset` seems defensive but I don't think it's necessary and I did some optimizations in this area recently too.

PR #302474, opened 2026-03-17
Don't send fetcher telemetry in stable

lramos15:

### DIFF
--- a/src/vs/workbench/services/telemetry/browser/telemetryService.ts
+++ b/src/vs/workbench/services/telemetry/browser/telemetryService.ts
@@ -58,7 +58,7 @@ export class TelemetryService extends Disposable implements ITelemetryService {
 		}));
 
 		this._register(requestService.onDidCompleteRequest(e => {
-			if (e.callSite === NO_FETCH_TELEMETRY) {
+			if (e.callSite === NO_FETCH_TELEMETRY || productService.quality === 'stable') {
 				return;
 			}
 			type FetchCallClassification = {
--- a/src/vs/workbench/services/telemetry/electron-browser/telemetryService.ts
+++ b/src/vs/workbench/services/telemetry/electron-browser/telemetryService.ts
@@ -74,7 +74,7 @@ export class TelemetryService extends Disposable implements ITelemetryService {
 		this.sendErrorTelemetry = this.impl.sendErrorTelemetry;
 
 		this._register(requestService.onDidCompleteRequest(e => {
-			if (e.callSite === NO_FETCH_TELEMETRY) {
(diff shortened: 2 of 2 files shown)


bpasero (approved):

mjbvz (approved):

src/vs/workbench/services/telemetry/browser/telemetryService.ts
```diff
@@ -58,7 +58,7 @@ export class TelemetryService extends Disposable implements ITelemetryService {
 		}));
 
 		this._register(requestService.onDidCompleteRequest(e => {
-			if (e.callSite === NO_FETCH_TELEMETRY) {
+			if (e.callSite === NO_FETCH_TELEMETRY || productService.quality === 'stable') {
```
mjbvz, line 61: Change looks good but as a follow up, let's only register these event handlers when `quality === 'insiders'`. That way there's zero overhead in stable

src/vs/workbench/services/telemetry/electron-browser/telemetryService.ts
```diff
@@ -74,7 +74,7 @@ export class TelemetryService extends Disposable implements ITelemetryService {
 		this.sendErrorTelemetry = this.impl.sendErrorTelemetry;
 
 		this._register(requestService.onDidCompleteRequest(e => {
-			if (e.callSite === NO_FETCH_TELEMETRY) {
+			if (e.callSite === NO_FETCH_TELEMETRY || productService.quality === 'stable') {
```
jrieken, line 77: In the past we have fixed telemetry overflow by allowing it with a low probability, like 1/10000. That's true sampling

lramos15: We don't necessarily want to do that in this case because we would want even sampling for each callsite. The idea of this telemetry is to catch anomalies in data volume to each endpoint so we can catch cases where we are over requesting certain data

src/vs/workbench/services/telemetry/browser/telemetryService.ts
lramos15: Sounds good!

src/vs/workbench/services/telemetry/electron-browser/telemetryService.ts
jrieken: > ... for each callsite. The idea of this telemetry is to catch anomalies in data volume to each endpoint so

But stable users is a very different population than insider users and IMO you are better off sampling over the whole population than just a special subset.

PR #302476, opened 2026-03-17
Clear native title attributes from chat inline anchors

vikeychen:
Fix the native-title/custom-hover conflict in chat inline anchor widgets.

Repro:

1. Render a chat file pill from a markdown anchor that already has a native `title` attribute.
2. `InlineAnchorWidget` reuses that anchor element and installs a managed hover on it.
3. `hoverService` warns that the element already has a title attribute, and the native tooltip can conflict with the managed hover.

This change removes the stale native `title` attribute before the widget installs its managed hover, and adds a focused browser test covering that path.


roblourens:
I'm not seeing any duplicate tooltips

roblourens:
Please open an issue with more specific details if this needs a fix

PR #302489, opened 2026-03-17
Improve screen reader labels for implicit context attachments

meganrogge:
fixes #301568

Improves accessibility of implicit context attachment widgets in chat input:
- Toggle button labels: The add button (+) now says "Add to context" instead of having an empty string as its accessible name.
- Implicit context aria-labels: Both file-based and string-based implicit context attachments now include "Suggested context" prefix in their aria-label so screen reader users know these are automatically suggested, not manually attached.

### DIFF
--- a/src/vs/workbench/contrib/chat/browser/attachments/implicitContextAttachment.ts
+++ b/src/vs/workbench/contrib/chat/browser/attachments/implicitContextAttachment.ts
@@ -218,8 +218,9 @@ export class ImplicitContextAttachmentWidget extends Disposable {
 		const fileBasename = basename(file);
 		const fileDirname = dirname(file);
 		const friendlyName = `${fileBasename} ${fileDirname}`;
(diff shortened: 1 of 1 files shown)


src/vs/workbench/contrib/chat/browser/attachments/implicitContextAttachment.ts
```diff
@@ -99,7 +99,7 @@ export class ImplicitContextAttachmentWidget extends Disposable {
 		// Create toggle button BEFORE the label so it appears on the left
 		if (isSuggestedEnabled) {
 			if (!isSelection) {
-				const buttonMsg = context.enabled ? localize('disable', "Disable current {0} context", attachmentTypeName) : '';
+				const buttonMsg = context.enabled ? localize('removeImplicitContext', "Remove current {0} context", attachmentTypeName) : localize('addToContext', "Add to context");
```
meganrogge:
```suggestion
				const buttonMsg = context.enabled ? localize('disableImplicitContext', "Disable current {0} context", attachmentTypeName) : localize('addToContext', "Add to context");
```

bhavyaus (approved):

PR #302496, opened 2026-03-17
sessions - allow to pin

bpasero:


justschen (approved):
i'm really excited about this one - will this work in local sessions as well?

Issue #209072, opened 2024-03-28
CLI on Windows: --folder-uri option only works as the last argument and when the --wait option is not present


kiwi-bui:
Does this issue occur when all extensions are disabled?: Yes/No

- VS Code Version: 1.87.2
- OS Version: Windows 10 Pro 19044.2486 VM

Steps to Reproduce:
1.  Case where --folder-uri is not the last option. Run `code --folder-uri vscode-remote://ssh-remote+vscode@some-workspace --disable-extension ms-vscode-remote.remote-containers --new-window --wait`. Command will return immediately without action or error.
2. Case where --folder-uri is the last option but --wait is present. Run `code  --disable-extension ms-vscode-remote.remote-containers --new-window  --wait --folder-uri vscode-remote://ssh-remote+vscode@some-workspace`. Command will return immediately without action or error.

There seems to be a bug where the the code command silently errors when --folder-uri and the --wait options are used together. Also, there seems to be a bug where the the code command silently errors when the --folder-uri option is not the last option.

connor4312:
This seems like an electron bug(?)

Passing `code --any-arg with-a-colon: --verbose` fails (doesn't call `main.js`) but removing the colon works `code --any-arg without-a-colon --verbose`. Same when calling the exe directly versus going through a batch script.

deepak1556:
Yeah the uri argument handling is not a bug but a security fix in the runtime, we also hit this in another feature of vscode for context: https://github.com/microsoft/vscode/issues/188769#issuecomment-1727691561

Would it be possible to encode the uri for the argument ?

kiwi-bui:
Is this related to the --wait option not being able to be used with the --folder-uri option?

kazoogh:
Hello! I am looking for a good first issue to contribute to as practice and this seems like one I can work on. Will be working on this for now hope that is okay :)

zhanglinqian:
I'd like to work on this issue.

PR #302519, opened 2026-03-17
Fix CLI handling of --folder-uri when used with other flags like --wait

omwanere:
Normalize URI values passed through --folder-uri before resolving them.
This prevents Electron argument parsing issues where URI schemes containing
':' can interfere with argument ordering when combined with flags such as --wait.

Tested with:
--folder-uri <uri> --wait
--wait --folder-uri <uri>

Both now open correctly.


dmitrivMS:
@omwanere What is the repro scenario? Is there an open issue this is fixing?

dmitrivMS:
This will not address the issue - see Copilot's comment, also @deepak1556's analysis in https://github.com/microsoft/vscode/issues/209072.

PR #302592, opened 2026-03-17
aiCustomization: make plugin/extension/builtin files read-only

### LINKED ISSUE #301292: Should plugin customizations be read only?
Testing #301158

Currently when accessed via configure menus the files are editable. Do we expect to have these be immutable?

connor4312:
aiCustomization: make plugin/extension/builtin files read-only

- Open plugin, extension, and built-in customization files as read-only
  in the editor, with a message explaining they cannot be edited.
- Extract isReadonlyStorage() helper to DRY the repeated storage type
  check used in both the open and delete actions.

Fixes https://github.com/microsoft/vscode/issues/301292


joshspicer (changes requested):
'built-in' for the sessions app should be editable (to let the user customize)

Image

joshspicer (approved):

PR #302633, opened 2026-03-18
fix: wire ext host restart through utility process api

### LINKED ISSUE #296681: Rogue Code Helper (Plugin) process when uninstalling / unloading extensions

No, it does not occur when all extensions are disabled, though I assume this is because the uninstalled extension is not being unloaded from the running application in this mode.

- VS Code Version: 1.109.5 (Universal)
- OS Version: macOS Tahoe 26.3 

Currently, if you install an extension from the marketplace, subsequently uninstall it, and then restart extensions, a rogue process will consume excessive resources on the machine.

Code Helper (Plugin) process consuming excessive resources after uninstalling an extension and restarting extensions

Steps to Reproduce:

1. Open up a resource monitor such as `Activity Monitor.app`
2. Install Cline (or another extension) from the VS Code Marketplace
3. Uninstall Cline (or another extension)
4. Click "Restart Extensions"
5. Observe a rogue "Code Helper (Plugin)" consuming excessive CPU and RAM

deepak1556:
Fixes https://github.com/microsoft/vscode/issues/296681

Followup to https://github.com/microsoft/vscode/pull/293144 covering extension host restart path

### DIFF
--- a/src/vs/platform/extensions/common/extensionHostStarter.ts
+++ b/src/vs/platform/extensions/common/extensionHostStarter.ts
@@ -9,6 +9,7 @@ import { createDecorator } from '../../instantiation/common/instantiation.js';
 export const IExtensionHostStarter = createDecorator<IExtensionHostStarter>('extensionHostStarter');
 
 export const ipcExtensionHostStarterChannelName = 'extensionHostStarter';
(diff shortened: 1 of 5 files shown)


alexdima (approved):
Nice! Thank you!

Issue #302645, opened 2026-03-18
Add config for linkActivationModifier and prevent link following when any other modifiers are down


calad0i:
In the current integrated terminal, link following modifier is always set to the other key than the `multiCursorModifier`, which is by default `ctrlCmd`. When the event dispatches, vscode checks only against if this modifier is down, but any other modifiers simultaneous down won't prevent it. This leads to unwanted behaviors like opening random binaries when trying to Ctrl+Shift+C to copy the pass, when double clicking for selection.

Since the multicusor modifier is not really related to the terminal, I would propose that the two configs should be isolated, and any extra modifier being down when clicking should prevent the link following.

Related: #3130, this would have been irrelevant if that was addressed...

PR #302647, opened 2026-03-18
Add config for linkActivationModifier and prevent link following when any other modifiers are down

calad0i:
This PR implements what is requested in #302645, mainly to prevent unwanted link following when extra modifiers are down, and isolates link following modifier in terminal from `multiCursorModifier`. The new config key is `terminal.integrated.linkActivationModifier`.


slipsec:
A use-case to consider? 

With the enhancements to the integrated browse and it now being the default target for ctrl+click in the terminal it'd be good to enhance this with a way to support a config to support a preference like setting a alt-behavior like ctrl+shift+click (or whatever hotkey) to support previous behavior of opening in system default browser.

PR #302816, opened 2026-03-18
feat: native macOS service provider for 'Open with Code' in Finder

ulugbekna:
@deepak1556 wdyt?


deepak1556 (changes requested):
This is much better, thank you. However, can you separate out the native addon code into its own repo and consume as a dependency. Example https://github.com/microsoft/vscode-windows-registry

Issue #20651, opened 2017-02-15
Add selection to next find match (Ctrl+D) is broken


kumarharsh:
- VSCode Version: 1.10 (1d17b4)
- OS Version: Windows

Background:
Earlier, pressing Ctrl+D would select all occurences of a word, and it was case-insensitive *when Match Case was off*. Now, when I keep pressing Ctrl+D, it keeps selecting those words, but when it encounters first case-sensitively different word, it selects it, and then refuses to select anything at all after it.

Steps to Reproduce:

1. Type this in your editor:
```
test
testte
Test
testte
test
```
2. Select first `test` and then press Ctrl+D a bunch of times. It stops after selecting `Test`.
![bug2](https://cloud.githubusercontent.com/assets/760112/22973897/b605dea2-f3a5-11e6-8578-d39510e1e449.gif)

## What it should do:
It shouldn't break after selecting `Test` when Match Case is off.

joaomoreno:
I just hit this. It's pretty bad.

michelkaporin:
Selection of occurrences works fine after the commit, however the highlight stops for the next matches (line 4 and 5 are not highlighted after selecting `Test`. Removing `verified` label.

![selection](https://cloud.githubusercontent.com/assets/2239563/23306392/c037688c-faa3-11e6-8bcb-7d6632a07019.gif)

Issue #6661, opened 2016-05-23
Unexpected behavior: `Add Selection To Next Find Match (Ctrl+D)` 


noppe18:
- VSCode Version:1.1
- OS Version: Windows7 Pro. SP1

Steps to Reproduce:
1. select "abc"
   ![iamge1](https://cloud.githubusercontent.com/assets/674208/15466146/172e6fde-2113-11e6-8750-7b68defcd6af.png)
2. press "Ctrl+D" 2 times. this is unexpected behavior.
   ![iamge2](https://cloud.githubusercontent.com/assets/674208/15466178/2d6f45b6-2113-11e6-98d4-f53d55981361.png)

I expected below image.
![image3](https://cloud.githubusercontent.com/assets/674208/15466190/39b1da78-2113-11e6-89ca-e48643069241.png)

ElChupacabra26:
Can confirm this also happens on El Capitán running the latest version of VSCode (1.5.3)

unional:
Another scenario:

```ts
const a = "a";
const b = "b";
const empty = "";
const d = "d";
```
Selecting `"` and [Ctrl+D] would stop at `empty = ""` as the selection is expanded to `""` at that position.

zhoudotnet:
When do you fixed this issue & release to production?

Issue #5400, opened 2016-04-16
"Select All Occurences of Find Match" does not select all if find uses regex


espresso3389:
- VSCode Version: Version 1.0.0-insider
- OS Version: Windows 10 Pro 1511 (10586.218)

Steps to Reproduce:
1. Search with regex to match multiple occurrence
2. Invoke "Select All Occurences of Find Match" (Oooops! Typo in the sentence)

It does not select all of them but only one.

![vscode-1 0 0-insider](https://cloud.githubusercontent.com/assets/1311400/14582938/f043ca9a-044d-11e6-81b0-93a3615cb595.gif)

espresso3389:
Do you have any fix estimation and/or workaround on it?

alexandrudima:
@espresso3389 workaround: press `alt+enter`

espresso3389:
@alexandrudima Wow, I didn't realize that key combination. Thanks!

QuietWind:
@alexandrudima  
thx your skill, thank you very much。

caffeineaddiction:
`Alt + Enter` works +1 Thank You

It would be nice to have a [Find All] button on the actual find-widget like exists in sublime

dakaraphi:
Looks like the text that appears in the list of commands is incorrect.
The 'select all occurrences of find' maps to command select all highlights.  So it should say instead 'select all highlights'
The Alt+Enter combination; however, does not appear in the quick command list.

PR #302943, opened 2026-03-18
Fix Ctrl+D in regex find mode

### LINKED ISSUE #107090: Add selection to next Find match does not work with Regex Find Mode
# Description
Adding selection to next find match does not work when the regex find mode is turned on. I am not sure this is intended, but based on #5715 I would assume you would want this to work along with adding everything to the selection so I'm calling it a bug.

I made an example of this in the gif below while in an instance with extensions disabled by running `code --disable-extensions` and with me just trying to add to the selection using `ctrl+d`

VscodeRegexctrld

## Version Info
VSCode Version: 1.48.2 (system setup)
Commit: a0479759d6e9ea56afa657e454193f72aef85bd0
Date: 2020-08-25T10:13:11.295Z
Electron: 7.3.2
Chrome: 78.0.3904.130
Node.js: 12.8.1
V8: 7.8.279.23-electron.0
OS: Windows_NT x64 10.0.18363

## Steps to Reproduce:

1. Open any document with some text
2. Turn on the regex mode for the search tool
3. Add some valid regex pattern
4. Attempt to add another cursor by pressing `ctrl+d`

KryptosAI:
Fixes #107090.

## Summary
- preserve `isRegex` when the multicursor session is created from the active find state
- use that regex flag for next/previous/select-all lookups in `MultiCursorSession`
- add a regression test covering repeated `AddSelectionToNextFindMatchAction` calls while the find widget owns a regex search

## Validation
- `npm run transpile-client`
- browser multicursor regression slice in Chromium covering `#107090`, `#5400`, `#6661`, and `#20651`

### DIFF
--- a/src/vs/editor/contrib/multicursor/browser/multicursor.ts
+++ b/src/vs/editor/contrib/multicursor/browser/multicursor.ts
@@ -286,7 +286,7 @@ export class MultiCursorSession {
 		//  - and the search string is non-empty
 		if (!editor.hasTextFocus() && findState.isRevealed && findState.searchString.length > 0) {
 			// Find widget owns what is searched for
-			return new MultiCursorSession(editor, findController, false, findState.searchString, findState.wholeWord, findState.matchCase, null);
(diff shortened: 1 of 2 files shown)


KryptosAI:

PR #303060, opened 2026-03-19
Remove `ChatSessionService.notifySessionOptionsChange`

mjbvz:
I'm trying to fix bugs in the chat session service and improve the extension api but am running into a lot of debt. Added comments noting the problems around `notifySessionOptionsChange` and why I want to remove it

### DIFF
--- a/src/vs/sessions/contrib/chat/browser/newSession.ts
+++ b/src/vs/sessions/contrib/chat/browser/newSession.ts
@@ -6,7 +6,6 @@
 import { Emitter, Event } from '../../../../base/common/event.js';
 import { Disposable, IDisposable } from '../../../../base/common/lifecycle.js';
 import { URI } from '../../../../base/common/uri.js';
-import { ILogService } from '../../../../platform/log/common/log.js';
 import { IChatSessionProviderOptionGroup, IChatSessionProviderOptionItem, IChatSessionsService } from '../../../../workbench/contrib/chat/common/chatSessionsService.js';
 import { IsolationMode } from './sessionTargetPicker.js';
 import { SessionWorkspace } from '../../sessions/common/sessionWorkspace.js';
@@ -102,7 +101,6 @@ export class CopilotCLISession extends Disposable implements INewSession {
 		readonly resource: URI,
 		defaultRepoUri: URI | undefined,
 		@IChatSessionsService private readonly chatSessionsService: IChatSessionsService,
-		@ILogService private readonly logService: ILogService,
 	) {
 		super();
 		if (defaultRepoUri) {
@@ -168,10 +166,7 @@ export class CopilotCLISession extends Disposable implements INewSession {
 		} else {
 			this.selectedOptions.set(optionId, value);
 		}
-		this.chatSessionsService.notifySessionOptionsChange(
-			this.resource,
-			[{ optionId, value }]
-		).catch((err) => this.logService.error(`Failed to notify session option ${optionId} change:`, err));
+		this.chatSessionsService.setSessionOption(this.resource, optionId, value);
 	}
 }
 
@@ -214,7 +209,6 @@ export class RemoteNewSession extends Disposable implements INewSession {
 		readonly target: AgentSessionProviders,
 		@IChatSessionsService private readonly chatSessionsService: IChatSessionsService,
 		@IContextKeyService private readonly contextKeyService: IContextKeyService,
-		@ILogService private readonly logService: ILogService,
 	) {
 		super();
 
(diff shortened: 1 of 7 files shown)


TylerLeonhardt (approved):

src/vs/workbench/contrib/chat/common/chatSessionsService.ts
```diff
@@ -355,13 +352,6 @@ export interface IChatSessionsService {
 
 	getNewSessionOptionsForSessionType(chatSessionType: string): Record<string, string | IChatSessionProviderOptionItem> | undefined;
 	setNewSessionOptionsForSessionType(chatSessionType: string, options: Record<string, string | IChatSessionProviderOptionItem>): void;
-	/**
-	 * Event fired when session options change and need to be sent to the extension.
-	 * MainThreadChatSessions subscribes to this to forward changes to the extension host.
-	 * Uses IWaitUntil pattern to allow listeners to register async work.
-	 */
-	readonly onRequestNotifyExtension: Event<IChatSessionOptionsWillNotifyExtensionEvent>;
```
mjbvz, line 363: Services should generally not know about the main thread or extensions

Instead the main thread should use apis exposed by the service

src/vs/workbench/contrib/chat/common/chatSessionsService.ts
```diff
@@ -355,13 +352,6 @@ export interface IChatSessionsService {
...
-	notifySessionOptionsChange(sessionResource: URI, updates: ReadonlyArray<{ optionId: string; value: string | IChatSessionProviderOptionItem }>): Promise<void>;
```
mjbvz, line 364: `Notify` type methods are generally a code smell because they let callers control event flows

It's better to hide event firing inside of service. Calls to `notify` methods also tend to get added unnecessary to workaround issues, which results in way too many events getting fired

mjbvz: `notify` is also a lying name here because this method actually changes state instead of just notifying

src/vs/workbench/contrib/chat/common/chatSessionsService.ts
```diff
@@ -306,11 +302,12 @@ export interface IChatSessionsService {
 	getSessionOptions(sessionResource: URI): Map<string, string> | undefined;
 	getSessionOption(sessionResource: URI, optionId: string): string | IChatSessionProviderOptionItem | undefined;
 	setSessionOption(sessionResource: URI, optionId: string, value: string | IChatSessionProviderOptionItem): boolean;
+	updateSessionOptions(sessionResource: URI, updates: ReadonlyArray<{ optionId: string; value: string | IChatSessionProviderOptionItem }>): boolean;
 
 	/**
 	 * Fired when options for a chat session change.
 	 */
-	readonly onDidChangeSessionOptions: Event<URI>;
+	readonly onDidChangeSessionOptions: Event<IChatSessionOptionsChangeEvent>;
```
mjbvz, line 310: Consolidating because we have two different events for session option changes. These were fired at different times

src/vs/sessions/contrib/chat/browser/newSession.ts
```diff
@@ -168,10 +166,7 @@ export class CopilotCLISession extends Disposable implements INewSession {
 		} else {
 			this.selectedOptions.set(optionId, value);
 		}
-		this.chatSessionsService.notifySessionOptionsChange(
-			this.resource,
-			[{ optionId, value }]
-		).catch((err) => this.logService.error(`Failed to notify session option ${optionId} change:`, err));
```
mjbvz, line 174: This error handling logic was copy and pasted around. If we need this, it should done in a single central place instead

src/vs/workbench/contrib/chat/common/chatSessionsService.ts
```diff
@@ -236,11 +236,7 @@ export interface IChatSessionItemController {
 	newChatSessionItem?(request: IChatNewSessionRequest, token: CancellationToken): Promise<IChatSessionItem | undefined>;
 }
 
-/**
- * Event fired when session options need to be sent to the extension.
- * Extends IWaitUntil to allow listeners to register async work that will be awaited.
- */
-export interface IChatSessionOptionsWillNotifyExtensionEvent extends IWaitUntil {
+export interface IChatSessionOptionsChangeEvent {
```
mjbvz, line 239: An async event was used which adds complexity but it doesn't actually seem to needed

`AsyncEvent` should be used very sparingly. They should not be treated as replacement for normal async function calls.

If you're just posting something to the ext host and don't care about the response, don't need to await it

src/vs/workbench/api/test/browser/mainThreadChatSessions.test.ts
mjbvz, line 757: I didn't fix it here but these casts are actually really bad because they are basically like `any` casts. It means `resolve` can be passed anything

Instead there's a Sinon type helper that is strongly typed that we should use

PR #303270, opened 2026-03-19
plugins: fix a bunch of issues in customizations

### LINKED ISSUE #302514: How do we maintain the information that a particular customization is from an installed plugin?
We should review and where we surface customizations (plugins, hooks, MCP, etc.) and ensure that we show information about the plugin source.

connor4312:
- Adds a "Show Plugin" option on every plugin-contributed item in the customizations view (closes #302514)
- Fixes MCP servers from plugins showing under "built-in"
- Fixes "Show Configuration" on plugin MCP servers opening a not found file
- Fixes hooks from plugins not showing in the customizations list. A bit of a special case here because hooks get parsed from plugin manifests and have special interpolation logic.


joshspicer (dismissed):
🔥

bpasero (approved):

PR #303312, opened 2026-03-19
Never use simple browser on desktop

kycutler:


extensions/simple-browser/src/extension.ts
```diff
@@ -15,7 +15,6 @@ declare class URL {
 const openApiCommand = 'simpleBrowser.api.open';
 const showCommand = 'simpleBrowser.show';
 const integratedBrowserCommand = 'workbench.action.browser.open';
-const useIntegratedBrowserSetting = 'simpleBrowser.useIntegratedBrowser';
```
jruales, line 18: Let's also update the docs site(s) to remove all mentions of this setting

jruales (approved):

PR #303395, opened 2026-03-20
fix: destroy read streams to prevent file descriptor leaks

buley:
## Summary

Two `createReadStream()` file descriptor leaks found via static analysis of the control flow graph:

**1. `src/vs/base/node/crypto.ts`** -- `checksum()` creates a read stream and pipes it to a crypto hash, but never calls `destroy()` on the stream. The `done()` callback removes event listeners but leaves the underlying fd open until GC collects it. This can cause fd exhaustion under high checksum throughput (e.g. extension updates).

**Fix:** Added `input.destroy()` in the `done()` callback, immediately after `removeAllListeners()`.

**2. `src/vs/server/node/webClientServer.ts`** -- `createReadStream(filePath).pipe(res)` leaks the file descriptor when the HTTP response is closed prematurely (client disconnect, timeout). Node's `.pipe()` does not automatically destroy the source stream when the destination closes.

**Fix:** Captured the stream reference and added `res.on('close', () => fileStream.destroy())`.

## Test Plan

- [x] Added `checksum mismatch rejects and cleans up stream` test in `src/vs/base/test/node/crypto.test.ts` to verify the stream is properly cleaned up on hash mismatch
- [x] Both fixes only affect cleanup paths and do not change normal control flow


alexdima (changes requested):
@buley The tests are failing

alexdima (approved):

joshspicer (approved):

PR #244419, opened 2025-03-24
fixed disable extension dropdown with incorrect disable globally option

ultra-000:
Fixes #244138

The change fixes a bug (it is a really small or (`||`) check that is kinda hard to spot) where there shouldn't be a dropdown with the option to disable the extension globally, it should just be a button with the text "Disable (Workspace)"

here are some visuals to better understand:

**Before fix:**

https://github.com/user-attachments/assets/8488fe18-36d7-46f7-aca2-5be946c1ab33

**After fix:**

https://github.com/user-attachments/assets/5aa9a576-a33f-464c-ab72-43cefa773626

thanks for reading! and please consider my pull request because I wanna be a part of this community :)

PR #303497, opened 2026-03-20
fix: correct extension disable dropdown items

kszongic:
Fixes #244138

## Problem

When an extension is disabled globally and then enabled for a specific workspace (resulting in \EnabledWorkspace\ enablement state), the Disable button dropdown incorrectly shows both **Disable** and **Disable (Workspace)** items.

## Root Cause

\DisableGloballyAction.update()\ was enabling itself when the extension's enablement state was either \EnabledGlobally\ or \EnabledWorkspace\. However, when an extension is in the \EnabledWorkspace\ state, it is already disabled globally — the workspace enablement is overriding the global disable. Offering 'Disable' (globally) in this state is redundant/confusing since it's already globally disabled.

## Fix

Changed \DisableGloballyAction\ to only enable when the extension's enablement state is \EnabledGlobally\. When the state is \EnabledWorkspace\, only the **Disable (Workspace)** action is shown, and since it's the only action, the dropdown arrow is not displayed — matching the expected behavior described in the issue.


dmitrivMS:
This is a dupe of https://github.com/microsoft/vscode/pull/244419 which has already been declined.

PR #303517, opened 2026-03-20
Fix quick diff false modifications on files with identical lines

H1net:
Fixes #253704

When a file has many identical lines (400+), the diff algorithm's O(n^2) DP computation can exceed the 1000ms timeout. On timeout, `trivialTimedOut` returns a single diff covering the entire file, causing the gutter to show all lines as modified.

**Changes:**

- Extend the identity early-out in `DefaultLinesDiffComputer` to files of any size (was limited to <= 1 line), skipping the expensive DP computation for identical content
- Check `quitEarly` in `QuickDiffModel._diff()` and discard timed-out results instead of displaying incorrect decorations

### DIFF
--- a/src/vs/editor/common/diff/defaultLinesDiffComputer/defaultLinesDiffComputer.ts
+++ b/src/vs/editor/common/diff/defaultLinesDiffComputer/defaultLinesDiffComputer.ts
@@ -25,7 +25,7 @@ export class DefaultLinesDiffComputer implements ILinesDiffComputer {
 	private readonly myersDiffingAlgorithm = new MyersDiffAlgorithm();
 
 	computeDiff(originalLines: string[], modifiedLines: string[], options: ILinesDiffComputerOptions): LinesDiff {
-		if (originalLines.length <= 1 && equals(originalLines, modifiedLines, (a, b) => a === b)) {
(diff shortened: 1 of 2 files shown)


tokyo4j:
Hello. I've been experiencing #253704 though [my file](https://github.com/labwc/labwc/blob/0ff9af4ae05b80bc09f4bf290ab7666930debf16/src/input/keyboard.c) does not contain many identical lines.

I tried this PR by applying https://github.com/microsoft/vscode/pull/303517.patch in [code](https://gitlab.archlinux.org/archlinux/packaging/packages/code) package in Arch Linux. Here's the result:

|version|behavior|
|--|--|
|1.100|The diff markers on the left are always synced with the git status. If I modify the file, the markers are quickly updated. |
|1.118|The diff markers easily get out of sync. Often all the lines are marked as modified even when `git status` shows nothing.|
|**1.118 with this PR applied**|The diff markers still easily get out of sync. Instead of marking all the lines as modified, now often no markers are shown even when the file has new lines. |

I think this PR is an improvement over the current behavior, but it's still not ideal if modified lines are not marked at all.

Would it be possible to restore the behavior from VS Code 1.100? Thanks.

---
EDIT: Hiding comment as I would like to continue discussion in #253704.

PR #303580, opened 2026-03-20
fix subagent tool id mismatch

justschen:
fix for issue where subagent tools now had mismatched tool ids. caused by https://github.com/microsoft/vscode/pull/302863  cc @hediet 

Screenshot 2026-03-20 at 11 09 45 AM


roblourens (approved):
Sorry this is complicated. I'd like to see if we can align IDs

PR #303606, opened 2026-03-20
agentPlugins: clone locally when in a remote

### LINKED ISSUE #298701: Copilot Chat Agent Plugins can't be installed when project inside WSL

Yes

- VS Code Version:  1.110 Insiders
- OS Version:  Windows 11 Pro

Steps to Reproduce:

1. Open project
2. Try to install any from basic plugins from UI
3. Plugin not found error

All fine if project on Windows

connor4312:
The git extension runs on the remote side, so it's not suitable to use to clone. This just does it directly.

I would kind of like to just take this path in all cases since I previously had to add some internal commands in the git extension. But for now going into next release this only affects the remote case which was previously broken.

Closes #298701

@lszomoru curious what you think about this approach. Not in love with it but it does the job.


diyana701-a11y (commented):
//tik tok.com//@v__7_7_7__v profile open

bhavyaus (approved):

PR #303610, opened 2026-03-20
Remove `as sinon.SinonStub` casts

mjbvz:
This is secretly like an any cast as the result isn't typed

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -2373,6 +2373,10 @@ export default tseslint.config(
 					'selector': `NewExpression[callee.object.name='Intl']`,
 					'message': 'Use safeIntl helper instead for safe and lazy use of potentially expensive Intl methods.'
 				},
+				{
+					'selector': 'TSAsExpression[typeAnnotation.type="TSTypeReference"][typeAnnotation.typeName.type="TSQualifiedName"][typeAnnotation.typeName.left.type="Identifier"][typeAnnotation.typeName.left.name="sinon"][typeAnnotation.typeName.right.name="SinonStub"]',
+					'message': `Avoid casting with 'as sinon.SinonStub'. Prefer typed stubs from 'sinon.stub(...)' or capture the stub in a typed variable.`
+				},
 			],
 		}
 	});
--- a/src/vs/base/test/common/sinonUtils.ts
+++ b/src/vs/base/test/common/sinonUtils.ts
@@ -1,0 +1,10 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import * as sinon from 'sinon';
+
+export function asSinonMethodStub<T extends (...args: never[]) => unknown>(method: T): sinon.SinonStubbedMember<T> {
+	return method as unknown as sinon.SinonStubbedMember<T>;
+}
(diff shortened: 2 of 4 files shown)


src/vs/base/test/common/sinonUtils.ts
```diff
@@ -0,0 +1,10 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import * as sinon from 'sinon';
+
+export function asSinonMethodStub<T extends (...args: never[]) => unknown>(method: T): sinon.SinonStubbedMember<T> {
```
mjbvz, line 8: This is safer than a direct type assertion (`obj.foo as sinon.SinonStub`) because it infers the types from the argument. It's basically the same as writing `obj.foo as sinon.SinonStubbedMember<typeof obj.foo>` but without the duplicate code. It also makes sure all callers have consistent behavior

src/vs/workbench/api/test/browser/mainThreadChatSessions.test.ts
```diff
@@ -194,10 +198,10 @@ suite('ObservableChatSession', function () {
 				deletions: 2,
 			}],
 		};
-		(proxy.$forkChatSession as sinon.SinonStub).resolves(forkedItem);
+		asSinonMethodStub(proxy.$forkChatSession).resolves(forkedItem);
 
 		const request: IChatSessionRequestHistoryItem = { type: 'request', id: 'request-1', prompt: 'Previous question', participant: 'participant' };
-		const expectedRequestDto = {
+		const expectedRequestDto: IChatSessionRequestHistoryItemDto = {
```
mjbvz, line 204: Implicitly typed objects are suspicious. They can easily accidentally have too many properties or miss properties. It relies on other places enforcing the typings, which the `as SinonStub` wasn't doing

It also means that rename and other intellisense features don't work properly

eslint.config.js
```diff
@@ -2385,6 +2373,10 @@ export default tseslint.config(
 					'selector': `NewExpression[callee.object.name='Intl']`,
 					'message': 'Use safeIntl helper instead for safe and lazy use of potentially expensive Intl methods.'
 				},
+				{
+					'selector': 'TSAsExpression[typeAnnotation.type="TSTypeReference"][typeAnnotation.typeName.type="TSQualifiedName"][typeAnnotation.typeName.left.type="Identifier"][typeAnnotation.typeName.left.name="sinon"][typeAnnotation.typeName.right.name="SinonStub"]',
+					'message': `Avoid casting with 'as sinon.SinonStub'. Prefer typed stubs from 'sinon.stub(...)' or capture the stub in a typed variable.`
```
mjbvz, line 2378: If you find a dangerous pattern, it's nice to add a little lint warning like this so that we don't introduce it again. This check is easy to bypass but it catches the obviously wrong cases

src/vs/workbench/api/test/browser/mainThreadChatSessions.test.ts
```diff
@@ -831,30 +854,36 @@ suite('MainThreadChatSessions', function () {
...
 		// Session with options
-		const sessionContentWithOptions = {
+		const sessionContentWithOptions: ChatSessionDto = {
 			id: 'session-with-options',
+			resource: resourceWithOptions,
 			history: [],
 			hasActiveResponseCallback: false,
 			hasRequestHandler: false,
+			hasForkHandler: false,
+			supportsInterruption: false,
```
mjbvz, line 868: Adding proper typings revealed that these objects were missing properties. Any code that was relying on these properties may have behaved incorrectly before

amunger (approved):

PR #303627, opened 2026-03-20
feat: add `chatPlugins` extension contribution point for agent plugins

connor4312:
Extensions can now contribute agent plugins via the `chatPlugins` key in their package.json, following the same pattern as `chatSkills`.

Each entry points to a plugin root directory within the extension that is discovered by a new `ExtensionAgentPluginDiscovery` class, which plugs into the existing agent plugin discovery infrastructure.

- Registers `chatPlugins` extension point with JSON schema (path, when)
- Evaluates and reacts to `when` context key expressions
- Prompts to uninstall the providing extension on plugin removal


roblourens (approved):

dmitrivMS (approved):

alexr00 (approved):
Name seems reasonable to me!

PR #303647, opened 2026-03-21
Fix browser focus stealing

kycutler:
Fixes #303638


src/vs/workbench/contrib/browserView/electron-browser/browserEditor.ts
```diff
@@ -1031,8 +1058,9 @@ export class BrowserEditor extends EditorPane {
 	override clearInput(): void {
 		this._inputDisposables.clear();
 
-		// Cancel any scheduled screenshots
+		// Cancel any scheduled timers
 		this.cancelScheduledScreenshot();
+		this.cancelFocus();
```
jruales, line 1063: Should we call this `cancelScheduledFocus()` to match `cancelScheduledScreenshot()`?

jruales (approved):

PR #303698, opened 2026-03-21
Fix: hide Disable globally action when extension is enabled for works…

sohamm20:
…pace

Fixes #244138

When an extension is globally disabled but enabled for a workspace, the Disable dropdown would incorrectly show both "Disable" and "Disable (Workspace)". This updates DisableGloballyAction's enablement condition to only be enabled when the extension is actually globally enabled.


dmitrivMS:
This is a dupe of https://github.com/microsoft/vscode/pull/244419 which has already been declined.

Issue #297827, opened 2026-02-25
Copilot Welcome Page: Title line height is too small


jruales:
Title line height is too small, so the text overlaps with itself when it wraps

<img width="321" height="739" alt="Image" src="https://github.com/user-attachments/assets/d771e65f-4cb5-4edf-88a2-60827b153dab" />

PR #303722, opened 2026-03-21
Fix Copilot Welcome Page title line-height causing text overlap on wrap

kszongic:
## Problem

The title on the Copilot Welcome Page overlaps with itself when the text wraps to multiple lines, as reported in #297827.

image from issue

## Root Cause

The \.chat-welcome-view-title\ element uses \line-height: normal\, which at \ont-size: 13px\ resolves to a value too tight for comfortable multi-line display, causing lines to visually overlap.

## Fix

Changed \line-height: normal\ to \line-height: 1.4\ — a standard ratio that provides adequate spacing between wrapped lines while keeping the layout compact.

Fixes #297827


kszongic:
Closing in favor of #303800

PR #303892, opened 2026-03-22
fix: set line-height on Copilot welcome page title to prevent text overlap

kszongic:
Fixes #297827

## Problem
The h1.product-name element on the Copilot Welcome Page has ont-size: 32px but no explicit line-height. When the title text wraps (e.g. at narrow window widths), lines overlap each other.

## Fix
Added line-height: 1.3 to .agentSessionsWelcome-header h1.product-name, which gives adequate spacing between wrapped lines at 32px font size.

## Testing
- Resized the editor to force title wrapping — lines no longer overlap
- Normal (non-wrapping) display is unaffected


kszongic:
Closing duplicate - keeping #303800 instead. Sorry for the noise.

PR #304224, opened 2026-03-23
Add workbench.browser.openLocalhostLinks to ExP

jruales:


DonJayamanne (approved):

karlhorky:
Very unexpected (as a user) that this change of the default `workbench.browser.openLocalhostLinks` setting for users shows up only in this PR with an empty description.

Maybe this should have been in the release notes blog post for VS Code 1.114 too? Currently `openLocalhostLinks` is nowhere to be found on that page:

Screenshot 2026-05-31 at 15 05 35

- https://code.visualstudio.com/updates/v1_114

PR #304235, opened 2026-03-23
fixes window.title

eli-w-king:
fixes https://github.com/microsoft/vscode/issues/303429

Steps to repro:
1. Set agent status to compact if it is not already
2. Change window.title to `${rootName} - ${activeRepositoryBranchName}` then `${activeRepositoryName}${separator}${activeFolderMedium}`
3. Correct title should show in command center placeholder text


mjbvz (approved):

sbatten:
I think this if fine for a candidate fix, but it seems like this data should flow through the `titleService`

PR #304236, opened 2026-03-23
Enable terminal commands to run freely in `Autopilot` and `Bypass Approvals` modes

meganrogge:
Reverts https://github.com/microsoft/vscode/pull/300824 which fixed eval issues by returning early from the tool if a denied command was attempted. This caused issues like https://github.com/microsoft/vscode/issues/303174, where commands were disallowed vs prompted for.

Now, we will skip the `auto approval` feature/system when a user is in `Autopilot` or `Bypass Approvals` mode so terminal commands run freely.

Also specifies "terminal commands" in `Autopilot` dialog (`Bypass Approvals` already had this).

Address #303174 in main. This will be a candidate.

### DIFF
--- a/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/runInTerminalTool.ts
+++ b/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/runInTerminalTool.ts
@@ -9,7 +9,7 @@ import { CancellationToken, CancellationTokenSource } from '../../../../../../ba
 import { Codicon } from '../../../../../../base/common/codicons.js';
 import { CancellationError } from '../../../../../../base/common/errors.js';
 import { Event } from '../../../../../../base/common/event.js';
-import { escapeMarkdownSyntaxTokens, MarkdownString, type IMarkdownString } from '../../../../../../base/common/htmlContent.js';
+import { MarkdownString, type IMarkdownString } from '../../../../../../base/common/htmlContent.js';
 import { Disposable, DisposableStore, MutableDisposable } from '../../../../../../base/common/lifecycle.js';
 import { ResourceMap } from '../../../../../../base/common/map.js';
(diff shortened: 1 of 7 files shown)


DonJayamanne (dismissed):

justschen (approved):

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/runInTerminalTool.ts
```diff
@@ -650,7 +647,14 @@ export class RunInTerminalTool extends Disposable implements IToolImpl {
...
 		};
-		const commandLineAnalyzerResults = await Promise.all(this._commandLineAnalyzers.map(e => e.analyze(commandLineAnalyzerOptions)));
+
+		// In Autopilot/Bypass Approvals modes, do not interact with terminal auto-approve rules.
+		// Commands should flow through directly based on the chat permission level.
+		const isSessionAutoApproved = chatSessionResource && this._isSessionAutoApproveLevel(chatSessionResource);
+		const commandLineAnalyzers = isSessionAutoApproved
+			? this._commandLineAnalyzers.filter(e => !(e instanceof CommandLineAutoApproveAnalyzer))
+			: this._commandLineAnalyzers;
+		const commandLineAnalyzerResults = await Promise.all(commandLineAnalyzers.map(e => e.analyze(commandLineAnalyzerOptions)));
```
Tyriar, line 657: An alternative way to handle this is to make a new analyzer that forces auto approval depending on the settings like with the sandbox one:

https://github.com/microsoft/vscode/blob/5fc1ff43b0aedf5e5473ecbba9bd3c90f7e0453f/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/commandLineAnalyzer/commandLineSandboxAnalyzer.ts#L25

PR #304316, opened 2026-03-24
Use skill folder as primary identifier

pwang347:
FYI @aeschli  @DonJayamanne - let's use folder as ID in case it mismatches with name. According to [spec](https://agentskills.io/specification#name-field) it needs to match folder, we can be permissive here but because chat extension doesn't always have knowledge of the contents/metadata we should just leverage folder as primary ID.


src/vs/workbench/contrib/chat/common/promptSyntax/service/promptsServiceImpl.ts
```diff
@@ -1477,9 +1477,10 @@ export class PromptsService extends Disposable implements IPromptsService {
 					this.logger.debug(`[computeSkillDiscoveryInfo] Agent skill file missing name attribute, using folder name "${folderName}": ${uri}`);
 					name = folderName;
 				}
-				const sanitizedName = this.truncateAgentSkillName(name, uri);
+				let sanitizedName = this.truncateAgentSkillName(name, uri);
 				if (sanitizedName !== folderName) {
 					this.logger.debug(`[computeSkillDiscoveryInfo] Agent skill name "${sanitizedName}" does not match folder name "${folderName}", using folder name: ${uri}`);
```
DonJayamanne, line 1482: Should this be a telemetry instead?
Else this will be useful only for a select few cases, I.e. only if you or Martin see these.
However if users run into this case, I think thats waht you want to know.

Yoyokrazy (approved):

src/vs/workbench/contrib/chat/common/promptSyntax/service/promptsServiceImpl.ts
pwang347: I feel like telemetry may not be needed since it won't be super actionable even if we see it happening often. The linting (+problems panel) coverage here might be good enough, or if we want more maybe it should also appear in the debug log panel.

PR #304327, opened 2026-03-24
fix: correct terminal sandbox icon in thinking dropdown

### LINKED ISSUE #303505: Background commands don't get their output massaged with sandbox info
https://github.com/user-attachments/assets/cae4323c-f269-40b9-a009-19b2f1ef777d

1. Follow https://github.com/microsoft/vscode-internalbacklog/issues/7145
2. Tell agent "run my app"
3. It will run `npm run dev` that fails in the sandbox, but it will never offer to run outside the sandbox

[Run_my_app_exports.tar.gz](https://github.com/user-attachments/files/26142431/Run_my_app_exports.tar.gz)

cc @dileepyavan

alexdima:
When sandbox is enabled, the tool-level icon was set to terminalSecure (lock) at registration time, which leaked into every rendering path during streaming before we knew if a specific command was actually sandboxed. This caused unsandboxed commands (requestUnsandboxedExecution=true) to show the lock icon.

Fix:
- Set toolData.icon to always be Codicon.terminal (no lock). The per-command isSandboxWrapped flag in toolSpecificData is the authoritative source.
- In the existing autorun in trackToolMetadata, update the icon element when the tool transitions out of streaming and toolSpecificData becomes available.
- Store icon elements in toolIconsByCallId map for direct access.

Fixes #303505


roblourens (approved):

marsam2489-lang (commented):
Good to start.

PR #304432, opened 2026-03-24
fix(markdown): remove angle brackets from autoClosingPairs to fix LaT…

### LINKED ISSUE #272922: [Markdown] Do not auto close angle brackets inside latex blocks

1150654748m-dev:
…eX editing

Remove angle brackets from autoClosingPairs to prevent unwanted auto-closing in LaTeX math expressions like $x < y$.

Fixes #272922


mjbvz (changes requested):
We should fix this just for latex instead of removing this feature entirely

PR #304447, opened 2026-03-24
Refactor theme and layout for sidebar, panel, and auxiliary components

mrleemurray:
Update theme colors and layout margins for sidebar, panel, and auxiliary bar components to enhance visual consistency and usability. Adjust padding and margins across various components, ensuring a cohesive design throughout the application.


src/vs/workbench/contrib/terminal/browser/terminalInstance.ts
```diff
@@ -2824,6 +2826,9 @@ export class TerminalInstanceColorProvider implements IXtermColorProvider {
 		if (terminalBackground) {
 			return terminalBackground;
 		}
+		if (this._environmentService.isSessionsWindow) {
```
benibenj: This should not be done this way. Instead, the color of PANEL_BACKGROUND should be changed in the sessions windows to be a different color

benibenj: `this._environmentService.isSessionsWindow` should almost never be used

src/vs/sessions/browser/media/style.css
```diff
@@ -15,40 +15,50 @@
...
  *   Sidebar:       no card (flush, spans full height)
+ *   Chat bar:       top=16, bottom=2, left=16, right=16
  *   Auxiliary bar:  top=16, bottom=18, right=16
  *   Panel:          bottom=18, left=16, right=16
  */
 
 .agent-sessions-workbench .part.sidebar {
-	background: var(--vscode-sideBar-background);
-	border-right: 1px solid var(--vscode-sideBar-border, transparent);
+	background: var(--vscode-editor-background);
```
benibenj: can we instead have a sessions sidebar color which is registered and inherits from the editor background color. Same for other parts

src/vs/sessions/browser/parts/media/auxiliaryBarPart.css
```diff
@@ -40,3 +41,10 @@
 	background-color: var(--vscode-activityBarTop-activeBackground, color-mix(in srgb, var(--vscode-sideBarTitle-foreground) 5%, transparent)) !important;
 	border-radius: 4px;
 }
+
+/* Override base workbench monaco editor background in auxiliary bar content */
+.agent-sessions-workbench .part.auxiliarybar > .content .monaco-editor,
+.agent-sessions-workbench .part.auxiliarybar > .content .monaco-editor .margin,
+.agent-sessions-workbench .part.auxiliarybar > .content .monaco-editor .monaco-editor-background {
+	background-color: var(--vscode-sideBar-background);
```
benibenj: same, create seesion aux sidebar color

src/vs/sessions/common/theme.ts
```diff
@@ -6,28 +6,21 @@
...
-	'sessionsSidebar.border',
-	{ dark: Color.fromHex('#808080').transparent(0.35), light: Color.fromHex('#808080').transparent(0.35), hcDark: contrastBorder, hcLight: contrastBorder },
-	localize('sessionsSidebar.border', 'Border color of the sidebar in the agent sessions window.')
-);
-
 // Sessions sidebar header colors
 export const sessionsSidebarHeaderBackground = registerColor(
 	'sessionsSidebarHeader.background',
-	SIDE_BAR_BACKGROUND,
+	editorBackground,
```
benibenj: inherit from sessionsSidebarBackground

src/vs/sessions/common/theme.ts
```diff
@@ -40,7 +33,7 @@ export const sessionsSidebarHeaderForeground = registerColor(
 // Chat bar title colors
 export const chatBarTitleBackground = registerColor(
 	'chatBarTitle.background',
-	SIDE_BAR_BACKGROUND,
+	editorBackground,
```
benibenj: inherit from sessionsSidebarBackground

src/vs/workbench/contrib/chat/browser/agentSessions/media/agentsessionsviewer.css
```diff
@@ -13,6 +13,10 @@
 		padding: 0 6px;
 	}
 
+	.agent-sessions-workbench .pane-body & .monaco-scrollable-element {
```
benibenj, line 16: this also affects vscode, have you checked if it looks fine there too?

alexr00 (dismissed):

src/vs/workbench/contrib/terminal/browser/terminalInstance.ts
benibenj: Use `colorRegistry.updateDefaultColor(PANEL_BACKGROUND, editorBackground);` or whatever color you want

benibenj (changes requested):
Wasn't really possible to override colors in the sessions window before but now that should be possible. I think we should use the correct variable name for each part, like `var(--vscode-sideBar-background)` for sidebar. In some cases it might not be possible and I think we should just create a new sidebar color `var(--vscode-sessions-sideBar-background)`. And we should just make sure the correct default value/inheritance is set

src/vs/workbench/contrib/chat/browser/agentSessions/media/agentsessionsviewer.css
mrleemurray: Yes, all good

benibenj (approved):

PR #304490, opened 2026-03-24
Let browser pages handle key events first

kycutler:
Fixes #303908


deepak1556 (changes requested):
`nodeIntegrationInSubFrames` will cause sandbox to be disabled for cross origin sub frame processes, only same origin sub frames will respect embedding frames sandbox properties. This is not desired and a security concern.

Can you rely on session api https://github.com/electron/electron/blob/main/docs/api/session.md#sesregisterpreloadscriptscript to inject browser view sessions preloadscript into frames

deepak1556 (dismissed):
Thanks!

benvillalobos (approved):

PR #304498, opened 2026-03-24
Remove self from notify, classifier, events, etc.

Tyriar:
cc @meganrogge


.mailmap
```diff
@@ -1,4 +1,3 @@
-Daniel Imms <daimms@microsoft.com> Daniel Imms <tyriar@tyriar.com>
```
anthonykim1, line 1: Nooooooo 😭

anthonykim1 (approved):

PR #304513, opened 2026-03-24
Fix PowerShell chained output in run_in_terminal

meganrogge:
fix #304488

Fixes a regression where `run_in_terminal` could return only the first line for chained PowerShell commands (for example `echo a ; echo b ; echo c`). 

- Keeps bracketed paste enabled by default for `isMacintosh`


src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/executeStrategy/basicExecuteStrategy.ts
```diff
@@ -129,7 +129,8 @@ export class BasicExecuteStrategy extends Disposable implements ITerminalExecute
 			// occurs.
 			this._log(`Executing command line \`${commandLine}\``);
 			markerRecreation.dispose();
-			this._instance.sendText(commandLine, true, true);
+			const forceBracketedPasteMode = shouldForceBracketedPasteModeForRunInTerminal(this._instance);
+			this._instance.sendText(commandLine, true, forceBracketedPasteMode);
```
Tyriar: Ugh, maybe it should be more specific then and just target macOS? Maybe it'll cause problems with other shells if it does for pwsh?

meganrogge: Yeah I worried about remotes, but was also thinking this initially

pwang347 (approved):

PR #304532, opened 2026-03-24
Chat Session Customizations initial Sketch

joshspicer:
## Summary

Adds a proposed API sketch for session (agent harness) type-specific chat customizations.

We implement this on the ChatSessionItemController, as the lifecycles are the same.

## Why

This API would allow the implementing extension populate the 'Chat customizations' UI in core (see the 'harness picker' at the top left)

**_Today_ these values are populated and filtered entirely by core (via `promptService`).  **

This new approach intends to be more scalable and correct, leaning on the individual harnesses (implemented through extensions) to provide the list of customizations that want to declare support for.

image

image


src/vscode-dts/vscode.proposed.chatSessionCustomizations.d.ts
```diff
@@ -0,0 +1,227 @@
...
+		readonly commands?: Command[];
+
+		/**
+		 * Commands shown in the context menu for individual items.
+		 * Each command receives the item's {@link ChatCustomizationItem.uri uri}
+		 * as its first argument.
+		 *
+		 * @example A "Run Prompt" command, a "Disable Skill" command.
+		 */
+		readonly itemCommands?: Command[];
```
joshspicer, line 118: would need to also send along the 'ID' of the item in the command

src/vscode-dts/vscode.proposed.chatSessionsProvider.d.ts
```diff
@@ -122,6 +123,21 @@ declare module 'vscode' {
 	 */
 	export type ChatSessionItemControllerForkHandler = (sessionResource: Uri, request: ChatRequestTurn2 | undefined, token: CancellationToken) => Thenable<ChatSessionItem> | ChatSessionItem;
 
+	/**
+	 * Options for creating a {@link ChatSessionItemController}.
+	 */
+	export interface ChatSessionItemControllerOptions {
```
mjbvz: Let' add a new `registerChatSessionCustomizationProvider` method instead. This is related to the controller but doesn't need to be linked to it

src/vscode-dts/vscode.proposed.chatSessionCustomizations.d.ts
```diff
@@ -0,0 +1,205 @@
...
+		 */
+		readonly onDidChangeCustomizations: Event<void>;
+
+		/**
+		 * Provide the current set of customizations.
+		 *
+		 * @param token A cancellation token.
+		 * @returns The customizations descriptor, or a thenable that resolves to one.
+		 */
+		provideCustomizations(token: CancellationToken): ProviderResult<ChatSessionCustomizations>;
```
mjbvz: The customizations global? Or can they be per-session?

src/vscode-dts/vscode.proposed.chatSessionCustomizations.d.ts
```diff
@@ -0,0 +1,205 @@
...
+
+		/**
+		 * Whether the user can delete this item from the management UI.
+		 *
+		 * Defaults:
+		 * - `true`  for {@link ChatCustomizationStorageLocation.Workspace Workspace}
+		 *           and {@link ChatCustomizationStorageLocation.User User}
+		 * - `false` for all other storage locations
+		 */
+		readonly canDelete?: boolean;
```
mjbvz: What does delete mean in this case? Does it delete the  file? How does the provider respond to a delete?

src/vscode-dts/vscode.proposed.chatSessionCustomizations.d.ts
```diff
@@ -0,0 +1,205 @@
...
+	 * Each optional field corresponds to a section in the management editor.
+	 * Omitting a field means the session type does not contribute items to
+	 * that section — the section may still appear if the built-in discovery
+	 * finds items for it.
+	 */
+	export interface ChatSessionCustomizations {
+		/**
+		 * Custom agents available in this session type.
+		 */
+		readonly agents?: ChatCustomizationItemGroup;
```
mjbvz: Instead of calling out each group here, can these groups be set on `ChatCustomizationItemGroup.id` and have a built-in set of ids?

Or maybe dedicated method for each customization?

mjbvz: Also how does this play with `registerInstructionsProvider`. Would this api replace that?

src/vscode-dts/vscode.proposed.chatSessionCustomizations.d.ts
```diff
@@ -0,0 +1,205 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+// ═══════════════════════════════════════════════════════════════════════
+//  Chat Session Customizations — types used by chatSessionsProvider
```
mjbvz: Let's prefix all of these types with `ChatSession`. So `ChatSessionCustomizationStorageLocation` for example

mjbvz (commented):
Thanks for looking into this. Added some initial.

As a next step I think the most helpful thing for review would be a sketch working through a few examples of this API in action. Show for example how a provider would implement this interface, when it would be invoked, and how providers would respond to changes made by the user

If you can establish a good set of base scenarios, we can use that to guide the design of the API

src/vscode-dts/vscode.proposed.chatSessionCustomizations.d.ts
```diff
@@ -0,0 +1,153 @@
...
+	 * The provider is called when the management UI needs to display
+	 * customizations, and re-called whenever
+	 * {@link onDidChangeCustomizations} fires.
+	 */
+	export interface ChatSessionCustomizationsProvider {
+		/**
+		 * Fired when customization items have changed and the UI should
+		 * re-fetch them.
+		 */
+		readonly onDidChangeCustomizations: Event<void>;
```
pwang347, line 134: @joshspicer would be good if we also make sure we capture the agent debug scenario with this API.
Some parts relevant to debugging:
1. Paths that were scanned but did not find any customizations
2. Files that were found but had invalid properties/format and so were skipped
3. Ideally, what exactly changed since last time (might be fine to just be which customization type was changed for simplicity)

src/vscode-dts/vscode.proposed.chatSessionCustomizations.d.ts
```diff
@@ -0,0 +1,153 @@
...
+
+	/**
+	 * Provides customization items for a chat session type.
+	 *
+	 * Registered via {@link chat.registerChatSessionCustomizationsProvider}.
+	 * The provider is called when the management UI needs to display
+	 * customizations, and re-called whenever
+	 * {@link onDidChangeCustomizations} fires.
+	 */
+	export interface ChatSessionCustomizationsProvider {
```
pwang347, line 129: Maybe worth checking; should we delegate some of the actual customization knowledge-passing work to `chatPromptFiles.d.ts` instead and this part can focus on the UI layout? or maybe we can move this to that file as well? CC @aeschli on this as well

src/vscode-dts/vscode.proposed.chatSessionCustomizations.d.ts
pwang347: See comment above, I'm wondering actually if we should merge with chatPromptFiles.d.ts (maybe supporting an optional session type ID or something)

src/vscode-dts/vscode.proposed.chatSessionCustomizations.d.ts
```diff
@@ -0,0 +1,153 @@
...
+	 *
+	 * Extensions may use these as the {@link ChatSessionCustomizationItemGroup.id id}
+	 * of a {@link ChatSessionCustomizationItemGroup} to place items into standard
+	 * sections in the management UI.
+	 */
+	export enum ChatSessionCustomizationType {
+		Agents = 'agents',
+		Skills = 'skills',
+		AgentInstructions = 'agentInstructions',
+		ContextInstructions = 'contextInstructions',
```
pwang347, line 25: should these just be categories of the instructions type?

src/vscode-dts/vscode.proposed.chatSessionCustomizations.d.ts
```diff
@@ -0,0 +1,153 @@
...
+	 *
+	 * Controls default behaviour in the management UI (grouping, delete-ability).
+	 */
+	export enum ChatSessionCustomizationStorageLocation {
+		/** From the current workspace (`.github/` folder, workspace root, etc.) */
+		Workspace = 1,
+		/** From user-level configuration (`~/.copilot/`, `~/.config/`, etc.) */
+		User = 2,
+		/** From an extension's contribution */
+		Extension = 3,
```
pwang347, line 44: should this one be part of API or just controlled by core?

src/vscode-dts/vscode.proposed.chatSessionCustomizations.d.ts
```diff
@@ -0,0 +1,153 @@
...
+	 * Use a well-known {@link ChatSessionCustomizationType} as the
+	 * {@link id} to place items into a standard management UI section.
+	 */
+	export interface ChatSessionCustomizationItemGroup {
+		/**
+		 * Identifier for this group. Use a value from
+		 * {@link ChatSessionCustomizationType} to map to a built-in section,
+		 * or a custom string for extension-defined sections.
+		 */
+		readonly id: string;
```
pwang347, line 97: was there any scenario where this is needed already? (curious URI-based would be sufficient here in the future, or we need something more powerful like webview)

pwang347: (instead of just typing it as `ChatSessionCustomizationType`)

src/vscode-dts/vscode.proposed.chatSessionCustomizations.d.ts
joshspicer: Delete would mean deleting the file is supported, but this kind of detail I should leave out for now

src/vscode-dts/vscode.proposed.chatSessionCustomizations.d.ts
joshspicer: Will bring this up in customizations sync tomorrow. I only just learned about these providers (@aeschli showed me this morning in our call)

src/vscode-dts/vscode.proposed.chatSessionCustomizations.d.ts
joshspicer: will bring this up in sync tomorrow

src/vscode-dts/vscode.proposed.chatSessionCustomizations.d.ts
joshspicer: Hmm

src/vscode-dts/vscode.proposed.chatSessionCustomizations.d.ts
joshspicer: The idea is that customizations are global, meaning scoped to a workspace (we have no sense right now of per-session customizations)

src/vs/workbench/contrib/chat/browser/aiCustomization/aiCustomizationListWidget.ts
```diff
@@ -479,6 +480,92 @@ export function sectionToPromptType(section: AICustomizationManagementSection):
...
+	}
+}
+
+/**
+ * Maps the numeric {@link IChatSessionCustomizationItem.storageLocation}
+ * to an optional groupKey override for the list display.
+ */
+function storageLocationToGroupKey(storageLocation: number): string | undefined {
+	// BuiltIn items are grouped under the special "builtin" key
+	if (storageLocation === 5) {
```
pwang347, line 531: this seems brittle, maybe it should at least call `storageLocationToPromptsStorage`?

.vscode/tasks.json
```diff
@@ -397,6 +397,7 @@
 			"type": "shell",
 			"command": "npx component-explorer serve -c ./test/componentFixtures/component-explorer.json -vv --kill-if-running",
 			"isBackground": true,
+			"inSessions": true,
```
pwang347, line 400: curious: what does this do?

pwang347 (approved):
did you have a sample consumer somewhere? would be nice to see how things are getting called

src/vs/workbench/contrib/chat/browser/aiCustomization/aiCustomizationListWidget.ts
joshspicer: Agreed, I think per your suggestion I want to remove storage altogether from the API and let core handle that.

.vscode/tasks.json
joshspicer: this exposes the task in the Sessions app

image

src/vscode-dts/vscode.proposed.chatSessionsProvider.d.ts
```diff
@@ -57,6 +57,18 @@ declare module 'vscode' {
...
+		 * Registers a {@link ChatSessionCustomizationsProvider customizations provider} for a chat session type.
+		 *
+		 * The provider supplies customization items (agents, skills, instructions, prompts)
+		 * that appear in the Customizations management UI for the given session type.
+		 *
+		 * @param chatSessionType The chat session type to provide customizations for.
+		 * @param provider The customizations provider.
+		 * @returns A disposable that unregisters the provider when disposed.
+		 */
+		export function registerChatSessionCustomizationsProvider(chatSessionType: string, provider: ChatSessionCustomizationsProvider): Disposable;
```
mjbvz, line 71: Please keep api proposals self contained. This should be in the `chatSessionCustomizations` file

src/vscode-dts/vscode.proposed.chatSessionCustomizations.d.ts
```diff
@@ -0,0 +1,145 @@
...
+
+	/**
+	 * Where a customization item originates from.
+	 *
+	 * Controls default behaviour in the management UI (grouping, delete-ability).
+	 *
+	 * TODO: Should this be inferred by core itself depending on the URI
+	 * scheme/path rather than declared by the extension?
+	 */
+	export enum ChatSessionCustomizationStorageLocation {
```
mjbvz, line 38: I think you should call this `source` instead of storage location. We use `source` a few places in vscode.d.ts

src/vscode-dts/vscode.proposed.chatSessionCustomizations.d.ts
```diff
@@ -0,0 +1,145 @@
...
+		 * The items in this group.
+		 */
+		readonly items: ChatSessionCustomizationItem[];
+
+		/**
+		 * Commands shown in the toolbar / "New" dropdown for this group.
+		 *
+		 * @example A "New Agent" command that opens a scaffold wizard.
+		 */
+		readonly commands?: Command[];
```
mjbvz, line 109: Should this and `itemCommands` be menu contributions instead?

src/vscode-dts/vscode.proposed.chatSessionCustomizations.d.ts
```diff
@@ -0,0 +1,145 @@
...
+		 * @example A "Run Prompt" command, a "Disable Skill" command.
+		 */
+		readonly itemCommands?: Command[];
+	}
+
+	/**
+	 * Provides customization items for a chat session type.
+	 *
+	 * Registered via {@link chat.registerChatSessionCustomizationsProvider}.
+	 * The provider is called when the management UI needs to display
```
mjbvz, line 125: Is this called every time the management UI is shown, or only the first time and then cached until `onDidChangeCustomizations` is called again?

src/vscode-dts/vscode.proposed.chatSessionCustomizations.d.ts
```diff
@@ -0,0 +1,145 @@
...
+		 * Each command receives the item's {@link ChatSessionCustomizationItem.uri uri}
+		 * as its first argument.
+		 *
+		 * @example A "Run Prompt" command, a "Disable Skill" command.
+		 */
+		readonly itemCommands?: Command[];
+	}
+
+	/**
+	 * Provides customization items for a chat session type.
```
mjbvz, line 122: I think this should explain what a customization is, what this API is for, and how session implementers are supposed to use it

src/vscode-dts/vscode.proposed.chatSessionCustomizations.d.ts
mjbvz: Also if you push everything to menus, then you don't need the concept of `ChatSessionCustomizationItemGroup` at all

Instead each `ChatSessionCustomizationItem` would have a `.group` property on it. This matches how the  `Task` type works for example

src/vscode-dts/vscode.proposed.chatSessionCustomizations.d.ts
```diff
@@ -0,0 +1,145 @@
...
+	 *
+	 * Extensions may use these as the {@link ChatSessionCustomizationItemGroup.id id}
+	 * of a {@link ChatSessionCustomizationItemGroup} to place items into standard
+	 * sections in the management UI.
+	 *
+	 * TODO: How granular should we be? Consider removing the sub-instruction
+	 * types (ContextInstructions, OnDemandInstructions) and collapsing to a
+	 * single 'instructions' type.
+	 */
+	export enum ChatSessionCustomizationType {
```
mjbvz, line 21: Right now this enum is just being used to store string constants   which is no good. Take a look at types like `CodeActionKind` or `TaskGroup` for an example of how this can work instead

src/vscode-dts/vscode.proposed.chatSessionCustomizations.d.ts
```diff
@@ -0,0 +1,145 @@
...
+		 */
+		readonly onDidChangeCustomizations: Event<void>;
+
+		/**
+		 * Provide the current customization groups.
+		 *
+		 * @param token A cancellation token.
+		 * @returns An array of customization groups, or a thenable that resolves to one.
+		 */
+		provideCustomizations(token: CancellationToken): ProviderResult<ChatSessionCustomizationItemGroup[]>;
```
mjbvz, line 142: For providers, the method/event name should match the class name. So in this case `providerChatSessionCustomizations`

mjbvz (commented):
Hey before this api gets adopted in copilot-chat, I left some comments about a few important things with the general design. Let's address these now so we don't have to worry about backwards compatibility and coordinating changes with copilot-chat

PR #304541, opened 2026-03-24
Fix remote agent host file picker for new URI scheme

roblourens:
Update the remote agent host file picker to support a new URI scheme, ensuring proper decoding of paths for remote files.

### DIFF
--- a/src/vs/workbench/services/dialogs/browser/simpleFileDialog.ts
+++ b/src/vs/workbench/services/dialogs/browser/simpleFileDialog.ts
@@ -138,6 +138,15 @@ export class SimpleFileDialog extends Disposable implements ISimpleFileDialog {
 	 * {@link remoteAuthority}).
 	 */
 	private scopedAuthority: string | undefined;
+	/**
+	 * Path prefix that the label formatter strips from URIs in the
+	 * scoped scheme (e.g. `/file/-` for agent host URIs that encode
+	 * the original scheme and authority as leading path segments).
+	 *
+	 * Stripped by {@link pathFromUri} and re-applied by
+	 * {@link remoteUriFrom} so the user sees clean paths.
+	 */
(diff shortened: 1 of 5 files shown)


src/vs/workbench/services/dialogs/browser/simpleFileDialog.ts
```diff
@@ -200,6 +209,7 @@ export class SimpleFileDialog extends Disposable implements ISimpleFileDialog {
 	public async showOpenDialog(options: IOpenDialogOptions = {}): Promise<URI[] | undefined> {
 		this.scheme = this.getScheme(options.availableFileSystems, options.defaultUri);
 		this.scopedAuthority = this.getScopedAuthority(options.defaultUri);
+		this.scopedPathPrefix = options.defaultUri && this.scopedAuthority ? this.computeScopedPathPrefix(options.defaultUri) : '';
```
roblourens, line 212: @alexr00 fyi agenthost uri paths now have a 'path' which encodes some extra data, then ends with the normal path that we want to display to the user. The point of this change is to detect that prefix automatically from the uri formatter, when `this.scopedAuthority` is set, and map between that formatted path and the real uri, so the user can work on the correctly formatted paths.

roblourens: If you think any of this is getting too complicated for this class, I could try to externalize some of this logic. I don't expect to make any more changes than this though.

TylerLeonhardt (approved):

src/vs/workbench/services/dialogs/browser/simpleFileDialog.ts
alexr00: The simple file dialog is a bit of a mess, I don't think this makes it so much worse. Thank you though!

PR #304549, opened 2026-03-24
sessions: polish New Session button shortcut chip and layout

hawkticehurst:
## Summary
- update the Sessions view **New Session** button to show a codicon-plus label (`+ Session`) with left-aligned content
- render a right-side shortcut hint chip using platform-specific notation (`⌘ N` on macOS, `Ctrl N` otherwise)
- if sessions new chat keybinding is changed the chip will be updated to reflect the new customized keybinding
- refine shortcut chip styling: monospace text, flat appearance, 2px radius, transparent border, and improved spacing/alignment
- improve icon optics by reducing plus size and adjusting vertical alignment
- move button alignment to flex-based layout (no absolute positioning / margin-auto hack)

## Files
- `src/vs/sessions/contrib/sessions/browser/sessionsViewPane.ts`
- `src/vs/sessions/contrib/sessions/browser/media/sessionsViewPane.css`

## Validation
- `npm run compile-check-ts-native` (pass)
- `npm run hygiene` in the original workspace reports an existing unrelated issue in `extensions/mermaid-chat-features/chat-webview-out/codicon.css` (missing/bad copyright header)
- pre-commit hygiene on the PR branch passed for these changes

## Screenshot

Screenshot 2026-03-24 at 3 53 30 PM

Customized keybindings:

Screenshot 2026-03-24 at 3 38 34 PM

Screenshot 2026-03-24 at 3 37 43 PM

### DIFF
--- a/src/vs/sessions/LAYOUT.md
+++ b/src/vs/sessions/LAYOUT.md
@@ -473,8 +473,9 @@ The Changes view is registered in `contrib/changesView/browser/changesView.contr
 The Sessions view is registered in `contrib/sessions/browser/sessions.contribution.ts`:
 
 - **Container**: Sessions container in `ViewContainerLocation.Sidebar` (default)
-- **View**: `SessionsViewId` with `AgenticSessionsViewPane`
+- **View**: `SessionsViewId` with `SessionsView` (`contrib/sessions/browser/views/sessionsView.ts`)
 - **Window visibility**: `WindowVisibility.Sessions`
+- **Primary action**: The sidebar content starts with a left-aligned secondary "New Session" button rendered as `$(plus) Session`, with an inline shortcut hint that reflects the active `workbench.action.sessions.newChat` keybinding when one is available
 
 ---
 
@@ -643,6 +644,7 @@ interface IPartVisibilityState {
 | Date | Change |
 |------|--------|
 | 2026-03-26 | Updated the sessions sidebar appear animation so only the body content (`.part.sidebar > .content`) slides/fades in during reveal while the sidebar title/header and footer remain fixed |
+| 2026-03-25 | Updated Sessions view documentation to reflect the refactored `SessionsView` implementation in `contrib/sessions/browser/views/sessionsView.ts` and documented the left-aligned "+ Session" sidebar action with its inline keybinding hint |
 | 2026-03-02 | Fixed macOS sidebar traffic light spacer to only render with custom titlebar; added `!hasNativeTitlebar()` guard to `SidebarPart.createTitleArea()` so the 70px spacer is not created when using native titlebar (traffic lights are in the OS title bar, not overlapping the sidebar) |
 | 2026-02-20 | Replaced custom `EditorModal` with standard `ModalEditorPart` via `MODAL_GROUP`; main editor part created but hidden; changed `workbench.editor.useModal` from boolean to enum (`off`/`some`/`all`); sessions config uses `all`; removed `editorModal.ts` and editor modal CSS |
 | 2026-02-17 | Added `-webkit-app-region: drag` to sidebar title area so it can be used to drag the window; interactive children (actions, composite bar, labels) marked `no-drag`; CSS rules scoped to `.agent-sessions-workbench` in `parts/media/sidebarPart.css` |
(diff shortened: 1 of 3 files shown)


src/vs/sessions/contrib/sessions/browser/sessionsViewPane.ts
```diff
@@ -149,16 +153,45 @@ export class AgenticSessionsViewPane extends ViewPane {
...
+		const keybindingHint = $('span.new-session-keybinding-hint');
+		const keybindingHintLabel = this._register(new KeybindingLabel(keybindingHint, OS, {
+			disableTitle: true,
+			keybindingLabelBackground: 'transparent',
+			keybindingLabelForeground: 'inherit',
+			keybindingLabelBorder: 'transparent',
+			keybindingLabelBottomBorder: undefined,
+			keybindingLabelShadow: undefined,
+		}));
+		DOM.append(buttonLabel, ...Array.from(newSessionButton.element.childNodes));
```
osortega: Should this just be `DOM.append(buttonLabel, newSessionButton.element)`

src/vs/sessions/contrib/sessions/browser/sessionsViewPane.ts
```diff
@@ -149,16 +153,45 @@ export class AgenticSessionsViewPane extends ViewPane {
...
+			newSessionButton.element.setAttribute('aria-label', keybindingAriaLabel
+				? localize('newSessionButtonAriaLabel', "New Session ({0})", keybindingAriaLabel)
+				: localize('newSessionButtonAriaLabelWithoutKeybinding', "New Session"));
+			DOM.reset(newSessionButton.element, buttonLabel);
+			keybindingHintLabel.set(keybinding);
+			if (keybinding) {
+				DOM.append(newSessionButton.element, keybindingHint);
+			}
+		};
+		this._register(Event.runAndSubscribe(this.keybindingService.onDidUpdateKeybindings, updateNewSessionButtonKeybinding));
```
osortega: Does this trigger on any keybinding change? If so, maybe we should only re-render the button if the actual keybinding we care about is the one that changed

osortega (dismissed):

src/vs/sessions/contrib/sessions/browser/sessionsViewPane.ts
hawkticehurst: No, `newSessionButton.element` refers to the root anchor element but the element we want to append the label to is a newly created div nested inside the anchor.

The issue / reason for this code is that the standard Button component has a strict API contract that only allows for a text label and icon, but no ability to add arbitrary markup (aka a keybinding chip UI). So what's happening here is we create a standard button with only text label and plus icon, then dynamically alter the DOM of that button afterwards by: 

1. Taking out that label + icon and putting it inside a new wrapper div (i.e. `buttonLabel`)
2. Create the keybinding chip UI (i.e. `keybindingHint` span) 
3. Put the label + icon div and keybinding chip span back into the root anchor/button element (i.e. `DOM.reset(newSessButton.element, buttonLabel)` and later `DOM.append(newSessionButton.element, keybindingHint)`)

I figured making a scoped dynamic update here was better than creating a PR that alters the API for all button across the codebase, but if you think this kind of change should be elevated to the core Button component let me know!

src/vs/sessions/contrib/sessions/browser/sessionsViewPane.ts
hawkticehurst: Great catch. 

It seems like there is no way to tell which keybinding/command was changed when `onDidUpdateKeybindings` is fired (tell me if I'm wrong about that though), so the solution I pushed up is to: 

1. Cache the keybinding label and aria label on first run
2. When subsequent `onDidUpdateKeybindings` is fired I check the cache against  `getNewSessionKeybinding().getLabel()` and `getNewSessionKeybinding().getAriaLabel()`
3. If there's a difference between them rerender, otherwise don't rerender

Seem reasonable?

osortega (dismissed):

osortega (dismissed):

hediet (approved):

PR #304582, opened 2026-03-24
Enhance image carousel to support video playback

rebornix:
Introduce support for video playback in the image carousel.


TylerLeonhardt (approved):
No video of the Image Carousel playing a Video?? :)

PR #304601, opened 2026-03-24
Preserve $TMPDIR when retrying terminal commands outside the sandbox

### LINKED ISSUE #303802: Configure $TMPDIR to the same value when sandboxing is on but execution is outside
* I had a situation where `$TMPDIR` was used to write a file in it.
* Then a command was attempted
* The command failed
* And then running the command outside the sandbox was attempted
* :bug: But now `$TMPDIR` is no longer defined

dileepyavan:
Fixes #303802

This updates the terminal sandbox retry path so commands that are rerun outside the sandbox continue to use the sandbox temp directory.

Today, when a sandboxed terminal command fails and is retried without sandboxing, the command can lose the `$TMPDIR` value that was set up for sandboxed execution. That breaks flows that stage files in `$TMPDIR` before the retry.

Changes in this PR:
- route sandbox rewriting through `ITerminalSandboxService.wrapCommand(...)` even when `requestUnsandboxedExecution` is requested
- preserve the sandbox temp directory for unsandboxed retries by prefixing the retried command with `TMPDIR="<sandbox temp dir>"`
- keep the normal sandbox-wrapped execution path unchanged
- add unit coverage for the unsandboxed retry path in `terminalSandboxService.test.ts`
- update the sandbox rewriter test stubs for the new `wrapCommand(command, requestUnsandboxedExecution)` signature
- add API coverage in `chat.runInTerminal.test.ts` for a blocked download that is retried outside the sandbox while continuing to write to `$TMPDIR`

Why this matters:
- fixes the `$TMPDIR` regression described in #303802
- keeps retry behavior consistent with the original sandboxed execution environment
- avoids falling back to an unexpected temp location during a retry

Testing:
- targeted API test: `blocked download can retry outside sandbox and write to $TMPDIR`
- unit test: `should preserve TMPDIR when unsandboxed execution is requested`
- targeted run of the added API test passed in both shell integration modes


joshspicer (dismissed):

.vscode/settings.json
```diff
@@ -213,5 +213,6 @@
 		".github/skills/.local": true,
 		".agents/skills/.local": true,
 		".claude/skills/.local": true,
-	}
+	},
+	"chat.tools.terminal.sandbox.enabled": false
```
TylerLeonhardt: expected to turn off sandboxing for the team?

TylerLeonhardt (approved):

PR #304793, opened 2026-03-25
fix: remove all manual folding ranges when selection is empty

### LINKED ISSUE #266597: editor.removeManualFoldingRanges should support empty selection
Folding ranges will sometimes break when pasting over Folded ranges
the ranges will appear in the wrong location/don't show up at all
`editor.removeManualFoldingRanges` does not fix it
nor does changing the language and sometimes even reloading
still happens even with `"editor.foldingStrategy": "indentation"`

found some solid repo steps
happens when pasting text over already recursively collapsed text
doesn't have to be JSON, happens in plaintext and TypeScript etc as well

Steps to Reproduce:
1. open large file with many foldings
  https://github.com/RedCMD/TmLanguage-Syntax-Highlighter/blob/main/syntaxes/regex.tmLanguage.json
3. recursively collapse most of the file (with middle-click/scroll-click/button 3)
4. paste a different large file over top
  https://github.com/RedCMD/redcmd.rascal-syntax-highlighter/blob/master/syntaxes/rascal.tmLanguage.json
6. witness broken foldings left behind

https://github.com/user-attachments/assets/dfcc4125-d488-41fd-a503-33ee091b174d

- VS Code Version: 1.104.0
- OS Version: Windows 11

yogeshwaran-c:
## Summary

When invoking `editor.removeManualFoldingRanges` with an empty selection (cursor only):

1. **If the cursor is on a manual folding range** → removes only the **innermost** manual range (addresses #212599)
2. **If the cursor is NOT on any manual range** → removes **all** manual folding ranges in the document (addresses #266597)

When text is explicitly selected, the existing behavior is preserved — only manual ranges intersecting the selection are removed.

This combined approach follows @RedCMD's suggestion from #266597 to detect whether a range intersects the empty selection, and also satisfies @aeschli's request to support empty selection for removing all manual folding ranges.

Fixes #266597
Fixes #212599

## Test plan

- [ ] Open a file, manually create folding ranges, place cursor on one, run command → only innermost manual range removed
- [ ] Place cursor on a line with NO manual folding range, run command → all manual ranges removed
- [ ] Select text spanning manual ranges, run command → only intersecting ranges removed


src/vs/editor/contrib/folding/browser/folding.ts
```diff
@@ -1205,8 +1205,33 @@ class RemoveFoldRangeFromSelectionAction extends FoldingAction<void> {
 		if (selections) {
 			const ranges: ILineRange[] = [];
 			for (const selection of selections) {
-				const { startLineNumber, endLineNumber } = selection;
-				ranges.push(endLineNumber >= startLineNumber ? { startLineNumber, endLineNumber } : { endLineNumber, startLineNumber });
+				if (selection.isEmpty()) {
```
aeschli: The change can go into 'removeManualRanges'. This action is the only user of it

aeschli (changes requested):

src/vs/editor/contrib/folding/browser/folding.ts
yogeshwaran-c: Done! Moved the cursor-aware logic (find innermost manual range, or fall back to removing all) into FoldingModel.removeManualRanges(). The action invoke() now just passes raw selection ranges as before, keeping it simple. The method detects single-line ranges (startLineNumber === endLineNumber) as cursor positions and resolves them before doing the intersection check.

manyiResearch:
Hi, a small quick question  — does this PR fix issues #266597 and #212599 together? Thanks!

aeschli (dismissed):

hediet (dismissed):

vritant24 (dismissed):

dmitrivMS:
Addressed the review feedback in `72b683b16f9`:

- `RemoveFoldRangeFromSelectionAction` now passes the original selections to `FoldingModel.removeManualRanges`, preserving column information.
- `removeManualRanges` uses `Range.isEmpty` to distinguish cursor-only invocations from explicit single-line text selections, so a non-empty selection no longer removes all manual ranges.
- Cursor handling removes the exact innermost manual region by index. This also fixes the previous implementation's intersection behavior, which could remove containing manual ranges along with the innermost one.
- Added regression coverage for a non-empty single-line selection outside manual ranges.

Validation: all 21 tests in `foldingModel.test.ts` pass, including the three `removeManualRanges` scenarios; pre-commit hygiene passes.

dmitrivMS (dismissed):

roblourens (dismissed):

dmitrivMS (approved):

alexr00 (approved):

PR #304845, opened 2026-03-25
Remove `getInProgressSessionDescription` from interface

mjbvz:
This does not need to live on the service as it's not directly about chat sessions


src/vs/workbench/contrib/chat/browser/chatSessions/chatSessions.contribution.ts
```diff
@@ -1415,3 +1361,57 @@ export function getResourceForNewChatSession(options: NewChatSessionOpenOptions)
 function isAgentSessionProviderType(type: string): boolean {
 	return Object.values(AgentSessionProviders).includes(type as AgentSessionProviders);
 }
+
+export function getInProgressChatSessionDescription(chatModel: IChatModel): string | undefined {
```
TylerLeonhardt, line 1409: easy to add tests for this now

TylerLeonhardt (dismissed):

PR #304948, opened 2026-03-25
feat: add full reference name support for language model tools and related API updates

DonJayamanne:
For https://github.com/microsoft/vscode/issues/304574


src/vs/workbench/contrib/chat/browser/tools/languageModelToolsService.ts
```diff
@@ -1621,6 +1621,12 @@ export class LanguageModelToolsService extends Disposable implements ILanguageMo
 	}
 
 	getFullReferenceName(tool: IToolData | IToolSet, toolSet?: IToolSet): string {
+		for (const [item, toolFullReferenceName] of this.toolsWithFullReferenceName.get()) {
```
DonJayamanne, line 1624: @connor4312 this is a change to existing code.
I believe this is safe, from waht I can tell previously this worked for sub agents for local tools like memory and the like.
Not sure it worked for MCP tools though.

I think we were sending the incorrect reference.
Deferring to your expertise here.

aeschli (approved):

connor4312 (approved):

PR #304955, opened 2026-03-26
fix: prevent mouseup on external links from activating them during editor drag

### LINKED ISSUE #199477: Clicking inside editor and releasing the mouse button on a link outside activates the link (web/monaco)

- VS Code Version: vscode.dev
- OS Version: Win10 / Firefox

Steps to Reproduce:

1. Go to vscode.dev
2. Open an editor
3. Click somewhere in the editor (e.g. to select a bunch of text) and keep the mouse button pressed
4. Move the mouse pointer over something outside the editor, e.g. the "Read the announcement" header
5. Release the mouse button

Nothing should happen since links/buttons are not supposed to trigger just because you release the mouse button while hovering them. After all, the click did not target that link/button.

FWIW this but is VERY annoying when embedding monaco-editor on a webapp that has a navigation near the editor's left side. When selecting a block of I almost always end up with the cursor on the menu and thus leave the page...

yogeshwaran-c:

Bug fix

When a user clicks inside the editor (e.g. to start selecting text) and drags the mouse outside the editor while holding the button down, releasing the mouse button on an external link/button activates that element. This is particularly problematic in Firefox and in Monaco editor embeddings where navigation elements are adjacent to the editor.

Closes #199477

## What is the new behavior?

The `pointerup` event during an active pointer monitoring session now calls `preventDefault()`, which prevents the browser from synthesizing a click event on whatever element happens to be under the cursor when the drag/selection operation ends outside the editor.

## Additional context

The fix is in `GlobalPointerMoveMonitor.startMonitoring()`. The `pointermove` handler already calls `preventDefault()`, but the `pointerup` handler did not. This one-line addition ensures the browser does not treat the pointer release as a click activation on external elements.

The maintainer @alexdima suggested this approach in the issue: "I think an `e.preventDefault()` could do the trick."

### DIFF
--- a/extensions/media-preview/media/imagePreview.css
+++ b/extensions/media-preview/media/imagePreview.css
@@ -33,21 +33,16 @@ body img {
 
 .container.image img {
 	padding: 0;
-	background-position: 0 0, 8px 8px;
 	background-size: 16px 16px;
 	border: 1px solid var(--vscode-imagePreview-border);
 }
 
 .container.image img {
-	background-image:
-		linear-gradient(45deg, rgb(230, 230, 230) 25%, transparent 25%, transparent 75%, rgb(230, 230, 230) 75%, rgb(230, 230, 230)),
-		linear-gradient(45deg, rgb(230, 230, 230) 25%, transparent 25%, transparent 75%, rgb(230, 230, 230) 75%, rgb(230, 230, 230));
+	background-image: conic-gradient(rgb(230, 230, 230) 25%, transparent 25% 50%, rgb(230, 230, 230) 50% 75%, transparent 75%);
 }
 
 .vscode-dark.container.image img {
-	background-image:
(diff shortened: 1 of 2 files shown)


alexdima:
🤖 

Thank you for the PR!

I tested this against Firefox 132, where I could reproduce the original issue using Monaco 0.46. Unfortunately, calling `preventDefault()` on the captured `pointerup` event did not prevent Firefox from subsequently dispatching a synthetic `click` to the link under the pointer. The `pointerup` event reported `defaultPrevented === true`, but the link was still activated.

I also tested current Firefox, where the original issue no longer reproduces. This matches https://github.com/microsoft/monaco-editor/issues/4379, which reports that the underlying Firefox behavior was fixed around Firefox 140.

Since this change does not fix affected Firefox versions and the problem is already fixed in current Firefox, I don't think we should merge it. It also changes a shared pointer-monitoring primitive used outside the editor.

Thanks again for investigating this!

PR #305031, opened 2026-03-26
Apply CSS format settings to style tags in HTML files

### LINKED ISSUE #283316: css formating config not apply to style tag in html during format html file.

- VS Code Version: 1.107.0
- OS Version: Windows_NT x64 10.0.19044

Steps to Reproduce:

1. set css formatting config inside settings.json
2. create a css file to test the result
3. create a html file and copy same content into a style tag
4. format the html file
5. the format result of html style content is different than the css file.

Expect result:
the html style content should be same as the css file content.

the settings.json
Image

the result of css file
Image

the result of html file
Image

Sathvik-Chowdary-Veerapaneni:
## Summary

Fixes #283316

CSS formatting settings (`css.format.newlineBetweenSelectors`, `css.format.braceStyle`, `css.format.spaceAroundSelectorSeparator`, etc.) were not being applied when formatting CSS within HTML `<style>` tags. The CSS mode in the HTML language server was missing a `format` method, so the CSS language service formatter was never invoked for embedded CSS regions.

## Changes

- Added a `format` method to the CSS mode in `extensions/html-language-features/server/src/modes/cssMode.ts` that:
  - Gets the embedded CSS document for the current HTML document
  - Merges the user's `css.format.*` settings with the base formatting options
  - Delegates to `cssLanguageService.format()` with the merged settings
- Added a test case verifying that `css.format.newlineBetweenSelectors: false` is respected when formatting CSS inside HTML style tags

## Root Cause

The `formatting.ts` orchestrator iterates through all embedded language mode ranges and calls `mode.format()` if it exists (line 73). However, the CSS mode returned by `getCSSMode()` never defined a `format` method, so it was silently skipped. The HTML formatter would handle the CSS content with only its own settings, ignoring CSS-specific format configuration.

## Test Plan

- [x] Verified no type errors introduced (zero errors from `cssMode.ts`)
- [x] Added test: `HTML & Styles with CSS format settings` validates `newlineBetweenSelectors: false` is applied
- [x] Existing test `HTML & Styles` still passes (default formatting behavior unchanged)

### DIFF
--- a/extensions/html-language-features/server/src/modes/cssMode.ts
+++ b/extensions/html-language-features/server/src/modes/cssMode.ts
@@ -5,7 +5,7 @@
 
 import { LanguageModelCache, getLanguageModelCache } from '../languageModelCache';
 import { Stylesheet, LanguageService as CSSLanguageService } from 'vscode-css-languageservice';
-import { LanguageMode, Workspace, Color, TextDocument, Position, Range, CompletionList, DocumentContext, Diagnostic, FormattingOptions, Settings, TextEdit } from './languageModes';
+import { LanguageMode, Workspace, Color, TextDocument, Position, Range, CompletionList, DocumentContext, Diagnostic } from './languageModes';
 import { HTMLDocumentRegions, CSS_STYLE_RULE } from './embeddedSupport';
 
 export function getCSSMode(cssLanguageService: CSSLanguageService, documentRegions: LanguageModelCache<HTMLDocumentRegions>, workspace: Workspace): LanguageMode {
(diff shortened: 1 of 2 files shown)


aeschli (changes requested):
See my comment.
Preferably fixed by better configuring JSBeautity  also for the embedded CSS, or, if not possible, many more tests that make sure there are no regression with formatting/indentation around `<style>` and `styles="..."`

dmitrivMS:
The issue has been fixed already.

PR #305167, opened 2026-03-26
Sessions: Refactor and move session status toggle to the left

mrleemurray:
Move the session status toggle in the sidebar for better visibility and enhance accessibility by adding an aria-label. This change improves the user experience for all users, particularly those using assistive technologies.

**Before**

https://github.com/user-attachments/assets/8a4ce896-af5c-4a15-9ffe-f5d740520bf3

**After**

https://github.com/user-attachments/assets/25f16749-97e3-4019-a652-3c0347738e2b

_Note: a future PR will address the panel animation affecting the control_


bpasero (changes requested):
While this change seems to work visually, there is some code weirdness. For example `Menus.TitleBarLeftLayout` now seems unused, but its still instantiated. And I wonder why we couldn't use it, just position it properly.

Just glancing over it and noticed this, @sandy081 should chime in as well

PR #305206, opened 2026-03-26
fix: support XDG_CONFIG_HOME for argv.json

### LINKED ISSUE #162712: Proposal: minimal changes to allow XDG-compatible data storage.

maruthang:
Fixes #162712

## Description

Added `XDG_CONFIG_HOME` support for `argv.json` on Linux. Previously, `argv.json` was always resolved relative to the user's home directory (`~/.vscode/argv.json`). On Linux, the path now resolves to `$XDG_CONFIG_HOME/<dataFolderName>/argv.json`, defaulting to `~/.config/<dataFolderName>/argv.json` if `XDG_CONFIG_HOME` is not set. macOS and Windows behavior is unchanged.

This follows the same pattern already used for `userDataPath` in the codebase.

## Changes

- **`src/main.ts`** — Updated `getArgvConfigPath()` to check `XDG_CONFIG_HOME` on Linux before falling back to the home directory.
- **`src/vs/platform/environment/common/environmentService.ts`** — Added `isLinux` import and `XDG_CONFIG_HOME` check in the `argvResource` getter so the environment service resolves the same path.
- **`src/vs/platform/environment/test/node/environmentService.test.ts`** — Added 4 regression tests covering: Linux with `XDG_CONFIG_HOME` set, Linux with `XDG_CONFIG_HOME` unset (default fallback), and confirming macOS/Windows behavior is unaffected.

## How to Test

1. On Linux, set `XDG_CONFIG_HOME` to a custom directory (e.g., `export XDG_CONFIG_HOME=/tmp/xdg-test`).
2. Launch VS Code and verify `argv.json` is created under `$XDG_CONFIG_HOME/.vscode/argv.json`.
3. Unset `XDG_CONFIG_HOME` and verify it falls back to `~/.config/.vscode/argv.json`.
4. On macOS/Windows, confirm no behavior change — `argv.json` remains in the home directory.
5. Run the unit tests: `./scripts/test.sh src/vs/platform/environment/test/node/environmentService.test.ts`

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -443,6 +443,12 @@ function getArgvConfigPath(): string {
 		dataFolderName = `${dataFolderName}-dev`;
 	}
 
+	// On Linux, respect XDG_CONFIG_HOME
+	if (process.platform === 'linux') {
+		const xdgConfigHome = process.env['XDG_CONFIG_HOME'] || path.join(os.homedir(), '.config');
+		return path.join(xdgConfigHome, dataFolderName!, 'argv.json');
+	}
+
 	return path.join(os.homedir(), dataFolderName!, 'argv.json');
 }
 
--- a/src/vs/platform/environment/common/environmentService.ts
+++ b/src/vs/platform/environment/common/environmentService.ts
@@ -7,6 +7,7 @@ import { toLocalISOString } from '../../../base/common/date.js';
 import { memoize } from '../../../base/common/decorators.js';
 import { FileAccess, Schemas } from '../../../base/common/network.js';
 import { dirname, join, normalize, resolve } from '../../../base/common/path.js';
+import { isLinux } from '../../../base/common/platform.js';
 import { env } from '../../../base/common/process.js';
 import { joinPath } from '../../../base/common/resources.js';
 import { URI } from '../../../base/common/uri.js';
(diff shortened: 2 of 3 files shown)


maruthang:

maruthang:
Rebased onto latest `main` to clear the conflict — clean cherry-pick aside from a small auto-merge in `environmentService.ts`. No semantic changes from the previous version of the PR; the XDG_CONFIG_HOME resolution logic and tests are unchanged.

PR #305214, opened 2026-03-26
implement support for background images

ghost:
## Adds support for background images in the editor

Feature suggestion: https://github.com/microsoft/vscode/issues/305149

Image


DioLps:
Understood Mr. @mjbvz . Sadly it feels like everybody is only accepting things that have to do with AI lately. Thanks for answering though.

PR #305253, opened 2026-03-26
fix: resolve #268776 - show JSDoc descriptions for JS in HTML script tags

maruthang:
## Summary
Fixes #268776

- **Bug:** Hovering over JS classes/methods inside HTML `<script>` tags showed only the type signature, not the JSDoc description or MDN links.
- **Root Cause:** The `doHover` handler in `javascriptMode.ts` only included `info.displayParts` (type signature) but ignored `info.documentation` (JSDoc comments).
- **Fix:** Added `info.documentation` to the hover output when available, so JSDoc descriptions and MDN reference links are now shown.

## Test plan
- Open an HTML file with a `<script>` tag containing `const dt = new DataTransfer();`
- Hover over `DataTransfer`
- Verify the hover shows the JSDoc description and MDN link (same as in `.js`/`.ts` files)


maruthang:
Closing as superseded: upstream's javascriptMode `doHover` now renders `ts.displayPartsToString(info.documentation)` and the JSDoc tags directly (extensions/html-language-features/server/src/modes/javascriptMode.ts), which is exactly what this PR added — JSDoc descriptions for JS inside HTML script tags now show on hover. No longer needed.

PR #305525, opened 2026-03-27
Try to reduce how often LocalAgentsSessionsController fires updates

mjbvz:
`LocalAgentsSessionsController` is firing updates on every single response change. This PR tries to reduce this by doing a object equality check before firing the update. In a follow up I'll also see if we can debounce listening to so many request updates


bpasero (approved):

bpasero (dismissed):
CCR

eleanorjboyd (approved):

PR #305606, opened 2026-03-27
Fix notebook markdown cell not re-rendering images on re-execution

### LINKED ISSUE #151151: Image stays broken when re-executing Markdown cell
### Applies To

- [X] Notebooks (.ipynb files)

### What happened?

1. Create a Markdown cell with an  tag referring to a non-existing path, e.g. ``.
2. Execute the cell. A "broken image" icon appears.
3. Create the image at the specified path.
4. Click the cell to edit. Re-execute it (without making any changes).

Expected result: the image is now shown.

Actual result: still shows a broken image icon. (The image is only shown after making a change to the cell itself.)

### VS Code Version

1.67.2

### Jupyter Extension Version

v2022.4.1021342353

### Jupyter logs

_No response_

Markdown

_No response_

_No response_

Local

Vi-Ku:
## Summary

Fixes #151151

When a notebook markdown cell contains an image reference (e.g., ``) and the cell is re-executed without changing the markdown text, the webview skips re-rendering due to a content equality optimization. This means:
- A broken image stays broken even after the file is created on disk
- A rendered image stays visible even after the file is deleted

This PR adds a `forceRender` flag so that explicit cell execution always forces the webview to re-render, while scroll/layout operations continue to benefit from the optimization.

## Root Cause

In `backLayerWebView.ts`, `showMarkupPreview()` compares the new markdown content with the cached entry:

```typescript
const sameContent = newContent.content === entry.content;
```

When `sameContent === true`, the webview message sends `content: undefined`, telling the renderer to skip re-rendering. This is correct for visibility changes (scroll into view) but incorrect for explicit execution, where external resources (images on disk) may have changed.

## Changes

| File | Change |
|------|--------|
| `webviewMessages.ts` | Added `forceRender?: boolean` to `IMarkupCellInitialization` |
| `backLayerWebView.ts` | Modified `sameContent` check: `&& !newContent.forceRender` |
| `notebookEditorWidget.ts` | Pass `forceRender: true` from `createMarkupPreview()` (execution path) |

## Design Decisions

- **Flag-based approach** over always re-rendering: preserves the IPC optimization for scroll/layout
- **Only the execution path** sets `forceRender: true` — the factory method `createMarkupCellInitialization()` used by viewport warmup/restore does not set it
- **Default behavior unchanged**: without the flag, the optimization still applies

## Testing

- Type check: 0 errors
- Notebook editor tests: passing
- Notebook execution service tests: passing
- Hygiene/lint: clean

### DIFF
--- a/src/vs/workbench/contrib/extensions/browser/extensionsActions.ts
+++ b/src/vs/workbench/contrib/extensions/browser/extensionsActions.ts
@@ -1751,7 +1751,7 @@ export class DisableGloballyAction extends ExtensionAction {
 		this.enabled = false;
 		if (this.extension && this.extension.local && !this.extension.isWorkspaceScoped && this.extensionService.extensions.some(e => areSameExtensions({ id: e.identifier.value, uuid: e.uuid }, this.extension!.identifier))) {
 			this.enabled = this.extension.state === ExtensionState.Installed
-				&& (this.extension.enablementState === EnablementState.EnabledGlobally || this.extension.enablementState === EnablementState.EnabledWorkspace)
+				&& this.extension.enablementState === EnablementState.EnabledGlobally
 				&& this.extensionEnablementService.canChangeEnablement(this.extension.local);
 		}
 	}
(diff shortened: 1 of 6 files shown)


Vi-Ku:

Vi-Ku:
Gentle bump — happy to address any feedback if needed. Thanks!

PR #305846, opened 2026-03-27
Rename sandbox setting to chat.agent.sandbox (#303421)

dileepyavan:
Rename the top-level sandbox setting from `chat.tools.terminal.sandbox.enabled` to `chat.agent.sandbox` to reflect that sandboxing is a general agent concept, not terminal-specific.

- Update setting ID value to `chat.agent.sandbox`
- Update description to be more general
- Deprecate old `chat.tools.terminal.sandbox.enabled` setting
- Update telemetry event name

Fixes #303421


src/vs/workbench/contrib/terminalContrib/chatAgentTools/common/terminalSandboxService.ts
```diff
@@ -318,4 +324,12 @@ export class TerminalSandboxService extends Disposable implements ITerminalSandb
 		const workspaceFolderPaths = this._workspaceContextService.getWorkspace().folders.map(folder => folder.uri.path);
 		return [...new Set([...workspaceFolderPaths, ...this._defaultWritePaths, ...(configuredAllowWrite ?? [])])];
 	}
+
+	private _getSettingValue<T>(settingId: TerminalChatAgentToolsSettingId, deprecatedSettingId?: TerminalChatAgentToolsSettingId): T | undefined {
+		const value = this._configurationService.getValue<T>(settingId);
```
anthonykim1: Maybe inspect is better here rather than getValue?

I'd double check if getValue always include default, if it does, `value !== undefined` would always be true and you could have dead code for `	return deprecatedSettingId ? this._configurationService.getValue<T>(deprecatedSettingId) : undefined;`

DonJayamanne (dismissed):

src/vs/workbench/contrib/terminalContrib/chatAgentTools/common/terminalSandboxService.ts
dileepyavan: Updated to check user value using inspect

dmitrivMS (dismissed):

roblourens (approved):

PR #305967, opened 2026-03-28
Fix/305961 create file keeps running

rupeshkumar-555:


rupeshkumar-555:
I cleaned this PR to keep scope focused on #305961 only.\n\n- Rebased branch on latest main\n- Removed unrelated workflow/chat changes\n- Kept only the new-file command fix in src/vs/workbench/contrib/files/browser/fileCommands.ts\n\nRegarding the merge snippet in chatListRenderer: the correct resolved line is the main-side condition using isSystemInitiatedRequest, and this PR no longer carries that unrelated change.

rupeshkumar-555:
Addressed the review thread for #305961 in the current PR diff:\n\n- Validated provided fileName before using it to build the target URI\n- Rejected unsafe values by requiring the resolved URI to stay under the default file path\n- Prevented silent overwrite when fileName is provided programmatically by warning and returning if the target already exists\n- Kept overwrite behavior only for the save-dialog path\n\nThis is now in the current branch diff for src/vs/workbench/contrib/files/browser/fileCommands.ts.

PR #306081, opened 2026-03-29
Allow readonly T[] in TreeDataProvider.getChildren()

n-gist:
This change allows `TreeDataProvider.getChildren()` to return `readonly T[]`, enabling type-guarded or immutable children arrays to be returned safely.

- Verified internal usage; no mutations occur on immutable arrays.
- Updated relevant interfaces to accept `readonly T[]`.
- Other components, like `src/vs/base/browser/ui/tree/asyncDataTree.ts`, already operate on `Iterable<T>` and are compatible.
- Improves type safety and simplifies extension code for fixed children arrays.

### DIFF
--- a/src/vscode-dts/vscode.d.ts
+++ b/src/vscode-dts/vscode.d.ts
@@ -12252,7 +12252,7 @@ declare module 'vscode' {
 		 * @param element The element from which the provider gets children. Can be `undefined`.
 		 * @returns Children of `element` or root if no element is passed.
 		 */
-		getChildren(element?: T): ProviderResult<T[]>;
+		getChildren(element?: T): ProviderResult<readonly T[]>;
 
 		/**
 		 * Optional method to return the parent of `element`.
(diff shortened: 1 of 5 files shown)


src/vscode-dts/vscode.d.ts
```diff
@@ -12252,7 +12252,7 @@ declare module 'vscode' {
 		 * @param element The element from which the provider gets children. Can be `undefined`.
 		 * @returns Children of `element` or root if no element is passed.
 		 */
-		getChildren(element?: T): ProviderResult<T[]>;
+		getChildren(element?: T): ProviderResult<readonly T[]>;
```
alexr00, line 12255: Thank you for the PR, but this would be build breaking for existing extensions.

n-gist: @alexr00 maybe something like:

`getChildren(element?: T): ProviderResult<T[] | readonly T[]>;`

I'm working on an extension where some tree item types intentionally expose readonly children, to prevent accidental mutation and enforce structural invariants.

Currently, this requires a cast like:
```
getChildren(element: FocusTreeItem | undefined): ProviderResult<FocusTreeItem[]> {
  return element.children as FocusTreeItem[]
}
```
This works, but it bypasses type safety and feels like a workaround.

Allowing `readonly T[]` in the return type would preserve backward compatibility while enabling extensions to model immutable tree structures more accurately, without casts.

PR #306374, opened 2026-03-30
 fix(inline-chat): resolve session from editor model for rephrase / ask in chat (#306361)

### LINKED ISSUE #306361: Inline chat follow-ups are non-actions
Inline chat jump offs when you get a non-code response are not usable 
1.  Send a prompt that will require more info (in this example I just typed "hello") 
2. ❗ focus into another editor
3. Follow up asking to rephrase or send to chat
4. Click rephrase / ask in chat
5. Telemetry is outputted showing I took the action but no change in the editor    

https://github.com/user-attachments/assets/4afdcc43-8255-46a6-9fc8-3ca56fff1f53

PranjalManhgaye:
Hey, @cwebster-99 
I’m trying to fix #306361 — you click Rephrase or Ask in Chat after a non-code reply, telemetry looks like you did something, but nothing useful happens in the UI.
What I changed: those actions only looked at this._currentSession.get(). If that’s empty for a moment even though there’s still a session for the file you’re editing, we bail out and you get a no-op. I added a fallback: if the editor has a model, I resolve the session with getSessionByTextModel(uri) so we still find the session.
How I tested: npm run compile-check-ts-native. I ran the full OSS build , so yeah — if you can click through inline chat in hover mode on your machine, that’d be the real proof.
Fixes #306361

### DIFF
--- a/src/vs/workbench/contrib/inlineChat/browser/inlineChatController.ts
+++ b/src/vs/workbench/contrib/inlineChat/browser/inlineChatController.ts
@@ -622,7 +622,7 @@ export class InlineChatController implements IEditorContribution {
 	}
 
 	async continueSessionInChat(): Promise<void> {
-		const session = this._currentSession.get();
+		const session = this._currentSession.get() ?? (this._editor.hasModel() ? this._inlineChatSessionService.getSessionByTextModel(this._editor.getModel().uri) : undefined);
 		if (!session) {
 			return;
 		}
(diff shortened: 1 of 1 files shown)


dmitrivMS:
Issue [#306361](https://github.com/microsoft/vscode/issues/306361) was closed by merged PR [#308183](https://github.com/microsoft/vscode/pull/308183), which fixed the same follow-up-action failure by focusing the owning editor before toolbar actions run. Current main contains that alternative fix, so this PR is superseded.

PR #306406, opened 2026-03-30
Detach background terminal processes so they survive VS Code shutdown

meganrogge:
fixes https://github.com/microsoft/vscode-copilot-evaluation/issues/3123

Background server processes started via `run_in_terminal` with `isBackground: true` are killed when VS Code exits due to three independent teardown paths (instance dispose, pty shutdown, utility process kill). This causes benchmark eval failures for tasks like `hf-model-inference`, `pypi-server`, and `kv-store-grpc` where a server must remain running during the eval phase.

### Approach

Adds a new command line rewriter (`CommandLineBackgroundDetachRewriter`) that wraps background commands to detach them from the terminal process tree:

- **POSIX (bash/zsh/fish):** `nohup <command> &`
- **Windows (PowerShell):** `Start-Process -WindowStyle Hidden -FilePath "<shell>" -ArgumentList "-NoProfile", "-Command", "<command>"`

The rewriter runs after the sandbox rewriter so that `nohup`/`Start-Process` wraps the entire sandbox runtime, keeping both the sandbox and child process alive through shutdown.

### Setting gate

Gated behind `chat.tools.terminal.detachBackgroundProcesses` (default `false`, hidden from Settings UI via `included: false`). The eval harness will enable this in its `settings.json`; normal users are unaffected and won't see the setting. Processes will be torn down after the eval completes in the docker container.

### Limitations
- If a `foreground` terminal times out, the command will not have been wrapped and the process will be disposed despite it becoming a `background` terminal
- On Windows, `cmd.exe` shells are not supported (returns `undefined`); only PowerShell is wrapped.


roblourens (dismissed):

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/commandLineRewriter/commandLineBackgroundDetachRewriter.ts
```diff
@@ -0,0 +1,70 @@
...
+		if (options.os === OperatingSystem.Windows) {
+			return this._rewriteForPowerShell(options);
+		}
+
+		return this._rewriteForPosix(options);
+	}
+
+	private _rewriteForPosix(options: ICommandLineRewriterOptions): ICommandLineRewriterResult {
+		return {
+			rewritten: `nohup ${options.commandLine} &`,
```
anthonykim1, line 50: @meganrogge Should we use `nohup setsid` instead of just `nohup`? 
nohup would ignore SIGHUP but I dont know if it ignores SIGTERM or SIGKILL

Maybe nohup is completely fine if we're only intending to test eval scenarios where vscode exits as expected.

meganrogge: Based on my understanding and testing, adding `setsid` would be overkill here.

bhavyaus (approved):

PR #306517, opened 2026-03-30
git: add headLabelOverride to Repository API

RockyWearsAHat:
## Summary

Add `headLabelOverride` property to the Git extension's `Repository` API, allowing extensions to temporarily replace the label shown in the Source Control view's branch indicator.

Closes microsoft/vscode#306496

## Problem

The Git extension derives its branch display label from the checked-out HEAD ref. Extensions that manage branch state externally — such as worktree managers, branch-per-chat session tools, or workspace isolation layers — cannot update the displayed branch name to reflect their logical context. This forces users to see misleading branch labels when the underlying checkout is managed programmatically.

## Approach

Add a `headLabelOverride` property on the `Repository` interface:

```typescript
repository.headLabelOverride = 'feature/my-work';   // override
repository.headLabelOverride = undefined;            // restore default
```

**Implementation:**
- `repository.ts`: Private `_headLabelOverride` field with getter/setter. Setter fires `_onDidChangeStatus` so the Source Control view updates immediately. `headLabel` and `headShortName` getters check the override before falling back to HEAD state.
- `git.d.ts`: `headLabelOverride: string | undefined` on the `Repository` interface.
- `api1.ts`: Getter/setter on `ApiRepository` delegating to the internal repository.

**Design notes:**
- The override is transient — it is not persisted across reloads. This is intentional: the extension that sets the override is responsible for re-applying it on activation.
- Setting the override fires the same status-change event as a real branch switch, so existing listeners (status bar, SCM view) pick it up without additional wiring.
- The override does NOT affect actual git operations. It is a display-only concern.

## Use Cases

- **Worktree-based session isolation**: An extension creates a worktree for each chat session and wants the SCM view to show the logical branch name rather than the detached-HEAD or worktree-internal branch.
- **Branch-per-chat workflows**: When switching between chat contexts, the extension can instantly update the displayed branch without waiting for a real checkout.
- **Project scaffolding tools**: During multi-step project setup, the tool can show a descriptive label like "Setting up..." instead of the actual branch name.
- **CI/CD preview branches**: Extensions that spin up ephemeral branches for testing can label them meaningfully in the UI.

## Testing

Manual verification that:
1. Setting `headLabelOverride` updates the Source Control view branch label
2. Setting it to `undefined` restores the original label
3. Git operations use the real HEAD, not the override

### DIFF
--- a/extensions/git/src/api/api1.ts
+++ b/extensions/git/src/api/api1.ts
@@ -91,6 +91,7 @@ export class ApiRepository implements Repository {
 
 	readonly onDidCommit: Event<void>;
 	readonly onDidCheckout: Event<void>;
+	readonly onDidChangeHeadLabel: Event<void>;
 
 	constructor(repository: BaseRepository) {
 		this.#repository = repository;
@@ -105,6 +106,7 @@ export class ApiRepository implements Repository {
 			filterEvent(this.#repository.onDidRunOperation, e => e.operation.kind === OperationKind.Commit), () => null);
 		this.onDidCheckout = mapEvent<OperationResult, void>(
 			filterEvent(this.#repository.onDidRunOperation, e => e.operation.kind === OperationKind.Checkout || e.operation.kind === OperationKind.CheckoutTracking), () => null);
+		this.onDidChangeHeadLabel = this.#repository.onDidChangeHeadLabel;
 	}
 
 	apply(patch: string, reverse?: boolean): Promise<void>;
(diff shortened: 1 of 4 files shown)


RockyWearsAHat:
All review feedback has been addressed in commit d5d4f1e3:

- Changed truthiness checks to `!== undefined` so empty-string overrides work correctly
- Added no-op guard in setter to avoid firing `_onDidChangeStatus` when value is unchanged
- Updated `git.d.ts` JSDoc to document that the override replaces the entire label including status suffixes

@copilot re-review

PR #306559, opened 2026-03-31
fix: include additional toggles in find input arrow key navigation

### LINKED ISSUE #207765: Cannot use arrow keys to go to toggle icon in notebook find widget
1. Open a notebook and CTRL+F
2. Use tab on the find action buttons- notice that this allows you to navigate to all of the different buttons.
3. Use right/left arrow on the find action buttons- notice that this only loops between the first three.

I think the problem is here:
https://github.com/microsoft/vscode/blob/a984153d6a98fba04f0a39a5ab9918c7b6e47511/src/vs/base/browser/ui/findinput/findInput.ts#L170-L194

The arrow key navigation is fixed at the findInput level, and we aren't adding to it, even when adding the funnel icon.
https://github.com/microsoft/vscode/assets/31675041/9949cd63-a57c-42e3-9fd8-0d03de6f44cb

yogeshwaran-c:

Bug fix (accessibility)

When using arrow keys to navigate between toggle buttons in the find input (case sensitive, whole words, regex), only the three standard toggles are reachable. Additional toggles — such as the notebook find filter button (funnel icon) — are excluded from keyboard navigation. This means users who rely on keyboard navigation cannot reach these controls using arrow keys.

Fixes #207765

## What is the new behavior?

Arrow key navigation now includes all toggles in the find input:

1. **`FindInput` base class**: The static `indexes` array (hardcoded to `[caseSensitive, wholeWords, regex]`) is replaced with a dynamic \`getToggleDomNodes()\` method that returns DOM nodes for all standard toggles plus any \`additionalToggles\`.

2. **`NotebookFindInput` subclass**: Overrides \`getToggleDomNodes()\` to also include the filter button container, so the notebook find widget's funnel icon is reachable via arrow keys.

The \`getToggleDomNodes()\` method is \`protected\`, allowing any subclass of \`FindInput\` to extend the set of keyboard-navigable toggles.

## Additional context

The issue was identified by @andreamah, who pointed to the exact location in \`findInput.ts\` where the arrow key indexes were hardcoded. The fix is minimal and scoped:

- \`src/vs/base/browser/ui/findinput/findInput.ts\` — extract toggle DOM node list into overridable method
- \`src/vs/workbench/contrib/notebook/browser/contrib/find/notebookFindReplaceWidget.ts\` — override to include filter button

### DIFF
--- a/src/vs/base/browser/ui/findinput/findInput.ts
+++ b/src/vs/base/browser/ui/findinput/findInput.ts
@@ -174,9 +174,9 @@ export class FindInput extends Widget {
 			}));
 
 			// Arrow-Key support to navigate between options
-			const indexes = [this.caseSensitive.domNode, this.wholeWords.domNode, this.regex.domNode];
 			this.onkeydown(this.domNode, (event: IKeyboardEvent) => {
 				if (event.equals(KeyCode.LeftArrow) || event.equals(KeyCode.RightArrow) || event.equals(KeyCode.Escape)) {
+					const indexes = this.getToggleDomNodes();
 					const index = indexes.indexOf(<HTMLElement>this.domNode.ownerDocument.activeElement);
 					if (index >= 0) {
 						let newIndex: number = -1;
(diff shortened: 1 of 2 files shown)


meganrogge (approved):
Thanks!

bhavyaus (approved):

PR #306561, opened 2026-03-31
feat: add Insert Link to Header in File command for markdown

### LINKED ISSUE #174999: Allow me to search for markdown headers without manually inserting `-` delimiters
Testing #174884

Currently if I have a header like so 

```md
# A header with spaces in the title
```

I have to search for it by typing the following when triggering header autocomplete

```
#a-header-with-spaces-in-the-title
```

I'd like to be able to type

```
#a header with spaces in the title
```

and have the completion still show up.

yogeshwaran-c:

Feature — adds a new command to the markdown-language-features extension.

When writing markdown links to headers (e.g., `[text](#header-name)`), users must type the slugified header name with dashes instead of spaces in the suggest widget. There is no way to search for headers using their natural text with spaces.

Closes #174999

## What is the new behavior?

Adds a new **Markdown: Insert Link to Header in File** command (`markdown.editor.insertHeaderLink`) that:

1. Retrieves all headers from the current document via `vscode.executeDocumentSymbolProvider`
2. Shows a quick pick with all headers listed by their natural text (e.g., "A header with spaces in the title")
3. Displays the corresponding slug in the description (e.g., `a-header-with-spaces-in-the-title`)
4. Uses VS Code's fuzzy matching so users can search by natural text with spaces
5. Inserts a properly formatted markdown link `[Header Text](#slug)` at the cursor position
6. If text is selected, uses the selection as the link text instead

This follows the approach suggested by @mjbvz in the issue: a dedicated command with a quick pick UI, similar to how `Markdown: Insert Link to File in Workspace` works.

### Implementation details

- New `InsertHeaderLink` class in `insertResource.ts`, following the same `Command` pattern as `InsertLinkFromWorkspace` and `InsertImageFromWorkspace`
- Uses the existing `ISlugifier` (GitHub-compatible) to generate correct slugs, including handling of duplicate headers
- Recursively flattens nested `DocumentSymbol` hierarchy to collect all headers
- Inserts via `SnippetString` with a placeholder for the link text when no text is selected

## Additional context

The command is available in the Command Palette as "Markdown: Insert Link to Header in File" whenever a markdown file is active in the editor.

### DIFF
--- a/extensions/markdown-language-features/src/commands/insertResource.ts
+++ b/extensions/markdown-language-features/src/commands/insertResource.ts
@@ -94,6 +94,7 @@ export class InsertHeaderLink implements Command {
 		const headers = this.#flattenHeaders(symbols);
 
 		if (!headers.length) {
+			vscode.window.showInformationMessage(vscode.l10n.t("No headers found in this document."));
 			return;
 		}
 
@@ -133,7 +134,9 @@ export class InsertHeaderLink implements Command {
 	#flattenHeaders(symbols: vscode.DocumentSymbol[]): { name: string }[] {
 		const result: { name: string }[] = [];
 		for (const symbol of symbols) {
-			result.push({ name: symbol.name });
+			if (symbol.kind === vscode.SymbolKind.String) {
+				result.push({ name: symbol.name });
+			}
 			if (symbol.children.length) {
 				result.push(...this.#flattenHeaders(symbol.children));
 			}
--- a/extensions/markdown-language-features/src/commands/insertResource.ts
+++ b/extensions/markdown-language-features/src/commands/insertResource.ts
@@ -7,6 +7,7 @@ import * as vscode from 'vscode';
 import { Utils } from 'vscode-uri';
 import { Command } from '../commandManager';
 import { createUriListSnippet, linkEditKind } from '../languageFeatures/copyFiles/shared';
+import { ISlugifier } from '../slugify';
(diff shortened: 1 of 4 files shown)


extensions/markdown-language-features/src/commands/insertResource.ts
```diff
@@ -66,6 +67,81 @@ export class InsertImageFromWorkspace implements Command {
...
+		this.#slugifier = slugifier;
+	}
+
+	public async execute() {
+		const activeEditor = vscode.window.activeTextEditor;
+		if (!activeEditor) {
+			return;
+		}
+
+		const symbols = await vscode.commands.executeCommand<vscode.DocumentSymbol[]>(
```
mjbvz, line 85: This picks up all symbols, not just headers. We need to make sure it's only headers

extensions/markdown-language-features/src/commands/insertResource.ts
```diff
@@ -66,6 +67,81 @@ export class InsertImageFromWorkspace implements Command {
...
+			activeEditor.document.uri
+		);
+
+		if (!symbols?.length) {
+			return;
+		}
+
+		const headers = this.#flattenHeaders(symbols);
+
+		if (!headers.length) {
```
mjbvz, line 96: There needs to be some sort of feedback if it returns early

mjbvz (changes requested):
~deleted~

PR #306562, opened 2026-03-31
fix: improve test coverage filter quickpick readability

### LINKED ISSUE #229920: Test: filter file to test could be more readable
Image

I would put the file name first and description second, like we do in file picker:

Image

### LINKED ISSUE #235347: Confusion on 'Filter Coverage To Test'
Testing #234946

I'm a bit confused on, if there is only one relevant test, how moving from 'All Tests' to a single test changes the coverage.  I would expect 'All tests' to be the intersection of all the component tests that you could filter down to?

Perhaps this is just an implementation quirk? (Looks like 'all tests' matches things like comments while the individual tests don't?)

https://github.com/user-attachments/assets/4d70e51b-2a25-481a-b90c-6fc7e2e4de7e

yogeshwaran-c:

Polish / Bug fix

The test coverage filter quickpick (used to filter coverage to a specific test) displays test items with the full ancestry chain as a flat string in the label field:

```
Suite › Subsuite › TestName
```

This makes it hard to scan when there are many tests with long ancestry paths. Additionally, the "All tests" option is confusing because it includes code executed during global initialization (e.g. comments, setup code), not just test functions.

Closes #229920
Closes #235347

## What is the new behavior?

1. **Test name first, ancestry as description**: The quickpick now shows the test name as the primary `label` and the ancestor chain as the `description`, matching the convention used by the file picker:

   ```
   TestName    Suite › Subsuite
   ```

   This makes it much easier to scan for specific tests.

2. **Rename "All tests" to "Entire run"**: As suggested by @connor4312 in #235347, this clarifies that the unfiltered view includes all code executed during the run (including global initialization), not just test function bodies.

## Additional context

Both changes apply to the two places where the coverage filter quickpick is shown:
- `FilterCoverageToTestInEditor` (editor title bar filter)
- `TestCoverageChangePerTestFilterAction` (coverage view toolbar filter)

The `getLabelForItem` function now returns `{ label, description? }` instead of a flat string, and callers spread the result into quickpick items.

### DIFF
--- a/src/vs/workbench/contrib/testing/browser/codeCoverageDecorations.ts
+++ b/src/vs/workbench/contrib/testing/browser/codeCoverageDecorations.ts
@@ -928,7 +928,7 @@ registerAction2(class FilterCoverageToTestInEditor extends Action2 {
 		const result = coverage.fromResult;
 		const previousSelection = testCoverageService.filterToTest.get();
 
-		type TItem = { label: string; testId: TestId | undefined; buttons?: IQuickInputButton[] };
+		type TItem = { label: string; description?: string; testId: TestId | undefined; buttons?: IQuickInputButton[] };
 
 		const buttons: IQuickInputButton[] = [{
 			iconClass: 'codicon-go-to-file',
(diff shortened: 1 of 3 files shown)


connor4312 (dismissed):
thanks!

roblourens (dismissed):

Yoyokrazy (dismissed):

connor4312 (approved):

TylerLeonhardt (approved):

PR #306573, opened 2026-03-31
feat: Support switching to the main window.

matts1:
From an auxilary window, there is currently no way to switch back to the main window.

This particularly an issue when you have a floating terminal window, as there is no key binding that can easily take you back to the correct split of the main editor (there is no active editor group, so "focus active editor group" doesn't work).

To test the changes:
1) Compile & open vscode
2) control shift p -> new terminal window
3) Create a split in the main window and focus the left
4) (from the main window) control shift p -> Switch to main window *should not appear*
5) (from the terminal window) control shift p -> switch to main window should switch the focus back to the left split of the main window
6) Repeat the process with the right split

Fixes #306571


bpasero (changes requested):
Wouldn't it be easier to just use the global `mainWindow` with its window ID?

bpasero (dismissed):
👍

alexr00 (dismissed):

bpasero (approved):

lszomoru (approved):

PR #306678, opened 2026-03-31
Add GitHub clone quick action

ozyern:


ozyern:
@copilot apply changes based on the comments in [this thread](https://github.com/microsoft/vscode/pull/306678#pullrequestreview-4035948006)

dmitrivMS:
If you'd like to add an extension, you can publish it separately to the marketplace.
There's a lot of churn here which seems risky.
Also not clear what issue/customer scenario is this solving.

PR #306827, opened 2026-03-31
new theme notification

aeschli:

### DIFF
--- a/src/vs/workbench/services/themes/browser/workbenchThemeService.ts
+++ b/src/vs/workbench/services/themes/browser/workbenchThemeService.ts
@@ -146,6 +146,7 @@ export class WorkbenchThemeService extends Disposable implements IWorkbenchTheme
 		// themes are loaded asynchronously, we need to initialize
 		// a color theme document with good defaults until the theme is loaded
 		let themeData: ColorThemeData | undefined = ColorThemeData.fromStorageData(this.storageService);
+		const previousColorThemeSetting = themeData?.settingsId;
 		const colorThemeSetting = this.settings.colorTheme;
 		if (themeData && colorThemeSetting !== themeData.settingsId) {
 			themeData = undefined;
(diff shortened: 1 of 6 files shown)


src/vs/workbench/services/themes/browser/workbenchThemeService.ts
```diff
@@ -246,34 +247,64 @@ export class WorkbenchThemeService extends Disposable implements IWorkbenchTheme
...
+			if (!this.settings.isDefaultColorTheme() || !previousSettingsId) {
+				return;
+			}
+			previousSettingsId = migrateThemeSettingsId(previousSettingsId);
+			if (!['Dark Modern', 'Light Modern'].includes(previousSettingsId)) {
+				return;
+			}
+			if (![ThemeSettingDefaults.COLOR_THEME_DARK, ThemeSettingDefaults.COLOR_THEME_LIGHT].includes(this.settings.colorTheme)) {
+				return;
+			}
```
alexr00, line 274: Not sure I understand this. We return early if the current theme is not one of the 2026 themes?

src/vs/workbench/services/themes/browser/workbenchThemeService.ts
```diff
@@ -246,34 +247,64 @@ export class WorkbenchThemeService extends Disposable implements IWorkbenchTheme
...
+		} finally {
+			// remeber to not show the dialog again
 			this.storageService.store(WorkbenchThemeService.NEW_THEME_NOTIFICATION_KEY, true, StorageScope.APPLICATION, StorageTarget.USER);
-		}));
+		}
+
+		const keepTheme = await new Promise(resolve => {
+			this.notificationService.prompt(
+				Severity.Info,
+				nls.localize({ key: 'themeUpdatedNotification', comment: ['{0} is the name of the new default theme'] }, "VS Code now ships with a new default theme '{0}'. Do you want to give it a try?", this.getColorTheme().label),
```
alexr00: `Do you want to give it a try?` sounds to me like you haven't already switched the theme, but the action "Keep it" sounds like you have already switched the theme.

src/vs/workbench/services/themes/browser/workbenchThemeService.ts
aeschli: This is to test that the user is on the default, which is now Dark/Light 2026.

src/vs/workbench/services/themes/browser/workbenchThemeService.ts
alexr00: I think just "VS Code has a new default theme '{0}'." (or maybe end with "!") is good.

alexr00 (approved):

PR #306874, opened 2026-03-31
Merge extension-provided policy

joshspicer:
tracking https://github.com/microsoft/vscode-internalbacklog/issues/7245

## Summary

Adds a merge script that fetches `extensionConfigurationPolicy` entries from the distro's `product.json` and merges them into `policyData.jsonc` during the policy export workflow.

fyi @pwang347 

## Problem

The `--export-policy-data` command runs in the OSS dev build, which has no access to `extensionConfigurationPolicy` data — that lives only in `vscode-distro`'s `product.json`. Extension policies for settings like `github.copilot.nextEditSuggestions.enabled` therefore never appear in the exported `policyData.jsonc`, which means they're missing from:

- Windows Group Policy ADMX/ADML artifacts
- macOS `.mobileconfig` profiles
- Linux `policy.json`
- The enterprise policy reference page on code.visualstudio.com

## Solution

New build script `build/lib/policies/mergeExtensionPolicies.ts` that:

1. Reads `package.json` to get the pinned distro commit (`"distro": "3de67a..."`)
2. Fetches the distro's `mixin/stable/product.json` at that exact commit (via local `.build/distro/` or GitHub API with `GITHUB_TOKEN`)
3. Reads `extensionConfigurationPolicy` entries
4. Merges them into `policyData.jsonc`, skipping any already present

### New workflow

```bash
npm run transpile-client && ./scripts/code.sh --export-policy-data && GITHUB_TOKEN=... node build/lib/policies/mergeExtensionPolicies.ts
```

## Changes

| File | Change |
|------|--------|
| `build/lib/policies/mergeExtensionPolicies.ts` | **New** — merge script |
| `src/.../policyExport.contribution.ts` | Updated disclaimer comment |
| `src/.../policyExport.integrationTest.ts` | Runs merge step after export; fails hard if distro unavailable |
| `.github/skills/add-policy/SKILL.md` | Documented extension policy lifecycle, distro format, downstream consumers |

## Dependencies

Companion PR: https://github.com/microsoft/vscode-distro/pull/1376 (adds `localization` format to `extensionConfigurationPolicy` entries in all three quality product.json files)


.github/skills/add-policy/SKILL.md
```diff
@@ -124,15 +124,65 @@ npm run compile-check-ts-native
 Regenerate the auto-generated policy catalog:
 
 ```bash
-npm run transpile-client && ./scripts/code.sh --export-policy-data
+npm run transpile-client && ./scripts/code.sh --export-policy-data && GITHUB_TOKEN=$(gh auth token) node build/lib/policies/mergeExtensionPolicies.ts
 ```
 
+The first command exports policies from the configuration registry. The merge step fetches `extensionConfigurationPolicy` entries from the distro's stable `product.json` and adds them to the output. It checks `.build/distro/` first; if unavailable, it uses `GITHUB_TOKEN` to fetch from the GitHub API. The `gh auth token` command provides the token from your authenticated GitHub CLI session.
```
pwang347: I'm not sure if everyone will have `gh` CLI installed, would it be better to also have a script like https://github.com/microsoft/vscode-copilot-chat/blob/main/script/setup/getToken.mts?

joshspicer: thanks! Cool script

src/vs/workbench/contrib/policyExport/test/node/extensionPolicyFixture.json
```diff
@@ -0,0 +1,28 @@
+{
```
joshspicer, line 1: Used purely for testing (to catch regressions in the generation code).  Policy values WILL drift from this and we do not expect this to update.

src/vs/workbench/contrib/policyExport/electron-browser/policyExport.contribution.ts
```diff
@@ -108,6 +147,76 @@ export class PolicyExportContribution extends Disposable implements IWorkbenchCo
...
+		// 1. DISTRO_PRODUCT_JSON env var (for testing)
+		const envPath = process.env['DISTRO_PRODUCT_JSON'];
+		if (envPath) {
+			this.log(`Reading distro product.json from DISTRO_PRODUCT_JSON=${envPath}`);
+			const content = (await this.fileService.readFile(URI.file(envPath))).value.toString();
+			return JSON.parse(content);
+		}
+
+		// 2. Local .build/distro/
+		const localPath = join(root, '.build/distro/mixin/stable/product.json');
```
pwang347: just curious: what's the scenario where we would prefer local over github's? is it an escape hatch (and if so should we just use DISTRO_PRODUCT_JSON?)

src/vs/workbench/contrib/policyExport/electron-browser/policyExport.contribution.ts
```diff
@@ -108,6 +147,76 @@ export class PolicyExportContribution extends Disposable implements IWorkbenchCo
...
+		const token = process.env['GITHUB_TOKEN'];
+		if (!token) {
+			throw new Error(
+				'GITHUB_TOKEN is required to fetch distro product.json. ' +
+				'Use `npm run export-policy-data` which sets up the required environment.'
+			);
+		}
+
+		this.log(`Fetching distro product.json for commit ${distroCommit} from GitHub...`);
+		const url = `https://api.github.com/repos/microsoft/vscode-distro/contents/mixin/stable/product.json?ref=${encodeURIComponent(distroCommit)}`;
```
pwang347, line 189: I don't suppose we have a fetch helper already for this right

pwang347 (dismissed):
Looks great, thank you for making this change!

src/vs/workbench/contrib/policyExport/electron-browser/policyExport.contribution.ts
joshspicer: Good point. I now cannot think of a reason i wouldn't just use the env var (+ the env var is great for the unit tests)

src/vs/workbench/contrib/policyExport/electron-browser/policyExport.contribution.ts
joshspicer: copilot and i checked

> Not really — [fetch.ts](vscode-file://vscode-app/Applications/Visual%20Studio%20Code%20-%20Insiders.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) has fetchGithub() but it's specialized for downloading release assets as Gulp vinyl streams, not general-purpose API calls. Other build scripts (e.g., checkDistroCommit.ts) also just use direct fetch/https calls inline. So no shared helper for this pattern today.

Yoyokrazy (approved):

pwang347 (approved):

PR #306882, opened 2026-03-31
Add configuration options for modal editor dimensions

mrleemurray:
Introduce configuration options for setting minimum width and height of the modal editor overlay. Refactor existing code to utilize these new settings, ensuring a more flexible and customizable user experience.

**Before**
image

**After**
image


Yoyokrazy (approved):

bpasero (changes requested):
I do not understand the need for this, esp. why a user should configure this. An editor should be able to declare its preferred min width (we already have that infrastructure in place) and then the modal editor respecting that.

Lets not do this for now, lets focus on P0s.

PR #306933, opened 2026-03-31
track disposable

### LINKED ISSUE #305972: createModelReference leak in notebookMulticursor.ts — reference never disposed
## Bug

In `src/vs/workbench/contrib/notebook/browser/contrib/multicursor/notebookMulticursor.ts`, the `createCursorController` method acquires a model reference via `createModelReference` but never disposes it or adds it to any disposable store.

### Location

https://github.com/microsoft/vscode/blob/main/src/vs/workbench/contrib/notebook/browser/contrib/multicursor/notebookMulticursor.ts#L273

### Details

```typescript
private async createCursorController(cell: TrackedCell): Promise<CursorsController | undefined> {
    const textModelRef = await this.textModelService.createModelReference(cell.cellViewModel.uri);
    const textModel = textModelRef.object.textEditorModel;
    if (!textModel) {
        return undefined; // ← leak: early return without disposing textModelRef
    }

    // ... creates controller, textModelRef is never disposed or tracked
    return controller; // ← textModelRef leaked
}
```

Two issues:
1. The early return when `textModel` is null does not dispose `textModelRef`
2. The normal code path never adds `textModelRef` to `this.cursorsDisposables` or any other disposable store

amunger:
fix https://github.com/microsoft/vscode/issues/305972

### DIFF
--- a/src/vs/workbench/contrib/notebook/browser/contrib/multicursor/notebookMulticursor.ts
+++ b/src/vs/workbench/contrib/notebook/browser/contrib/multicursor/notebookMulticursor.ts
@@ -273,8 +273,10 @@ export class NotebookMultiCursorController extends Disposable implements INotebo
 		const textModelRef = await this.textModelService.createModelReference(cell.cellViewModel.uri);
 		const textModel = textModelRef.object.textEditorModel;
 		if (!textModel) {
+			textModelRef.dispose();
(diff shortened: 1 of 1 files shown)


Yoyokrazy (approved):
ty 👍

PR #306987, opened 2026-04-01
Update UI improvements

eli-w-king:
https://github.com/user-attachments/assets/4a359e70-cf52-43d7-8998-1aad3f5919e3

- Actionable update states now show a blue "Update" button with no icon
- Removed the collapsed icon-only state so the button stays visible as "Update"
- Manual mode keeps contextual title bar hints and shows a Restart button in the tooltip when ready
- Non-manual mode keeps the simple single-click experience with no tooltip restart button and a "Restart to Update" tooltip title
- Renamed the secondary tooltip action to "Release Notes" and kept it left-aligned
- Removed the settings icon from update tooltips
- Reduced title bar button padding and fixed spacing between the Update label and manual hint text
- Guarded the `mouseenter` tooltip trigger to skip re-showing while the hover is already visible, preventing flicker and avoiding forcibly closing sticky/locked hovers
- Typed `text` as `string | null` in `updateTooltip.ts` to avoid implicit `any`


dmitrivMS (changes requested):
We cannot concat strings for UI purposes as it breaks localization.
I suggest we avoid the button expansion.

PR #307059, opened 2026-04-01
contextkey: fix scanner returning '>=' instead of '>' for greater-than operator

xingsy97:
### Problem

`Scanner.getLexeme()` returns `'>='` for both `TokenType.Gt` and `TokenType.GtEq`, making the two operators produce identical text. The `Lt`/`LtEq` pair correctly returns `'<'` and `'<='` respectively, but the `Gt` case was set to `'>='` instead of `'>'`.

### Fix

Change the return value for `TokenType.Gt` from `'>='` to `'>'`.


ulugbekna (approved):
good catch, and thank you for the PR!

pwang347 (approved):

PR #307240, opened 2026-04-01
policyExport: add `included` flag to PolicyDto output

joshspicer:
tracking https://github.com/microsoft/vscode-internalbacklog/issues/7245

## Summary

Adds the `included` boolean field to the exported `PolicyDto` in `policyData.jsonc`, so downstream consumers (docs website, UI generators) can determine whether a policy-managed setting is also user-configurable.

## Motivation

Some configuration properties have `included: false` on their schema (e.g. `ExtensionGalleryServiceUrlConfigKey`), meaning they don't appear in the normal Settings UI and can only be managed via policy. This information was not being propagated to the exported policy data, so downstream consumers had no way to know whether to show the setting in their UI.

## Changes

- **`policyDto.ts`** — Added optional `included?: boolean` field to `PolicyDto` with JSDoc documentation.
- **`policyExport.contribution.ts`** — Populated the field in both export paths:
  - Config registry policies: `schema.included !== false` (resolves to `false` only when the schema explicitly sets `included: false`).
  - Extension policies from distro product.json: always `true` (these are user-configurable by nature).

## Example output

After regenerating `policyData.jsonc`, a policy-only setting like `extensions.gallery.serviceUrl` will include `"included": false`, while normal settings will have `"included": true`.

### DIFF
--- a/src/vs/workbench/contrib/policyExport/common/policyDto.ts
+++ b/src/vs/workbench/contrib/policyExport/common/policyDto.ts
@@ -25,6 +25,13 @@ export interface PolicyDto {
 	type?: string | string[];
 	default?: unknown;
 	enum?: string[];
+
+	/**
+	 * When `false`, the setting is not user-configurable and can only be
+	 * managed via policy. Downstream consumers (e.g. docs website) can use
+	 * this to hide the setting from their UI. Defaults to `true`.
+	 */
+	included?: boolean;
 }
 
 export interface ExportedPolicyDataDto {
--- a/build/lib/policies/policyData.jsonc
+++ b/build/lib/policies/policyData.jsonc
@@ -50,7 +50,8 @@
                 }
             },
             "type": "string",
-            "default": ""
+            "default": "",
+            "included": false
         },
         {
             "key": "chat.mcp.gallery.serviceUrl",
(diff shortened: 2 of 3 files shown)


src/vs/workbench/contrib/policyExport/common/policyDto.ts
```diff
@@ -25,6 +25,13 @@ export interface PolicyDto {
 	type?: string | string[];
 	default?: unknown;
 	enum?: string[];
+
+	/**
+	 * When `false`, the setting is not user-configurable and can only be
+	 * managed via policy. Downstream consumers (e.g. docs website) can use
+	 * this to hide the setting from their UI. Defaults to `true`.
+	 */
+	included?: boolean;
```
pwang347, line 34: nit: should we name this something like `policyOnly` or `userConfigurable`?

pwang347 (approved):

src/vs/workbench/contrib/policyExport/common/policyDto.ts
joshspicer: I perferred to mirror the name in configuration for symmetry. I agree it's not a clear name

PR #307256, opened 2026-04-01
fix: detect interrupted commands in terminal exit code hack

### LINKED ISSUE #237517: Incorrect TerminalShellExecution exit code

- VS Code Version: 1.94.0
- OS Version: macOS Sonoma 14.4

Steps to Reproduce:

1. Listen for `onDidEndTerminalShellExecution` events
2. Run a command that returns some exit code
3. Type the same command and interrupt with ^C (i.e. interrupt without actually running the command)
4. The exit code reported by the `TerminalShellExecution` event for the second execution (expected to be `undefined` since the command didn't actually run) will be the same as the previous command's exit code.

For example:
```bash
$ echo test^C
-> exit code undefined
$ echo test
-> exit code 0
$ echo test^C
-> exit code 0
$ echo test^C
-> exit code 0
$ echo not test^C
-> exit code undefined
$ echo test^C
-> exit code undefined
```

From a quick look, this seems like caused by this code (marked as a hack): https://github.com/microsoft/vscode/blob/3b30f94f4f129e67be9b6c677aac582923da1fab/src/vs/platform/terminal/common/capabilities/commandDetectionCapability.ts#L360-L370

yogeshwaran-c:

Bug fix

When a command is typed and interrupted with Ctrl+C (for example, `echo test^C`), the `TerminalShellExecution` end event can report an exit code inherited from the previous execution of the same command.

For example:

```text
$ echo test     -> exit code 0 (correct)
$ echo test^C   -> exit code 0 (incorrect; should be undefined)
```

This happens because `handleCommandFinished` contains a workaround for versions of bash whose `history` output merges identical commands. When bash reports an `undefined` exit code and the shell integration reports the same command text as the previous command, the workaround reuses the previous exit code. An interrupted duplicate command therefore looks like the history-merge case and incorrectly inherits the previous result.

Closes #237517

## What is the new behavior?

`CommandDetectionCapability` now listens to the existing `PromptInputModel.onDidInterrupt` event, which detects the Ctrl+C/ETX input directly. When the current command was interrupted, the duplicate-history workaround is skipped and the `undefined` exit code reported by the shell is preserved. This matches the documented `TerminalShellExecutionEndEvent.exitCode` contract for commands canceled via Ctrl+C.

The interrupted state is reset when the next prompt starts. Non-interrupted duplicate commands continue to use the existing bash history workaround unchanged.

Using the prompt input model's interrupt signal also avoids inferring cancellation from rendered `^C` text, so a literal command ending in `^C` is not treated as interrupted.

## Additional context

The regression tests verify that:

1. An interrupted duplicate command keeps an `undefined` exit code instead of inheriting the previous command's exit code.
2. A non-interrupted duplicate command still inherits the previous exit code, including when its literal text ends in `^C`.

The tests also retain the initial command in their expected command lists and no longer use unnecessary `any` casts.

Validation:

- `commandDetectionCapability.test.ts`: 10 passing
- `npm run precommit`: passed


meganrogge (approved):
Thanks!

dmitrivMS (dismissed):

dmitrivMS:
Updated the fix in cfa3acd0d19822dccb862d423a355ffd6ecbb9cb:

- Use the existing `PromptInputModel.onDidInterrupt` signal instead of inferring Ctrl+C from rendered `^C` text.
- Preserve an `undefined` exit code for canceled commands, matching the `TerminalShellExecutionEndEvent.exitCode` API contract, rather than synthesizing exit code 130.
- Keep the duplicate bash history workaround for non-interrupted commands.
- Update the regression tests to model the duplicate-history command line, include the initial command in expectations, remove the unnecessary `any` casts, and verify literal `^C` text is not treated as an interruption.

Validation:
- `commandDetectionCapability.test.ts`: 10 passing
- `npm run precommit`: passed

dmitrivMS (approved):

alexr00 (approved):

anthonykim1 (approved):

PR #307271, opened 2026-04-01
Fix #256037: wrong indentation caused by incorrect bracket matcing

leonor-a-a-ist:
When typing a closing bracket immediately before another closing token, the auto-indent logic could start searching from the wrong column. This sometimes led to matching the wrong opening bracket, causing incorrect indentation.

This has been fixed by using the cursor column (where the closing bracket is inserted) as the search bound, or 0 when no valid opening bracket is found.

Added regression tests covering different scenarios, including cases like } before ) and ], and ] before ), with varying initial indentation levels.

Fixes #256037


leonor-a-a-ist:
@aiday-mar 
Hi! Let me know if you have any feedback on this PR or if I should make any changes. Thanks!

PR #307367, opened 2026-04-02
Fix native macOS text navigation shortcuts in editable Integrated Browser inputs

saankim:
## Summary

This addresses a regression in the macOS Integrated Browser where native text navigation shortcuts in editable inputs are swallowed instead of being handled by the page.

In plain editable targets such as `input`, `textarea`, and `contenteditable`, `Cmd+Left` / `Cmd+Right` should move the caret to the beginning / end of the line. Instead, those key events were being forwarded into the workbench shortcut pipeline and prevented from reaching the embedded page.

## Regression Context

- Related to #307355
- Earlier history: #303908, #304490, #306198
- The observed good/bad build boundary pointed strongly at #306198
- The likely regression point is #306198, which expanded preload-side browser key forwarding and started calling `preventDefault()` / `stopPropagation()` before forwarding unhandled modifier-based key events

## Fix

- Introduce a focused browser view key routing helper
- Preserve native handling for editable targets
- Keep forwarding behavior unchanged for non-editable targets
- Keep the scope narrow to native text navigation in editable inputs rather than loosening shortcut routing globally

At minimum this covers the reported `Cmd/Ctrl+Left` and `Cmd/Ctrl+Right` behavior, and the same narrow editable-target navigation family for arrow-based line navigation.

## Tests

Ran:

```bash
PATH="/opt/homebrew/opt/node@22/bin:$PATH" npm run test-node -- --run src/vs/platform/browserView/test/common/browserViewKeyRouting.test.ts
PATH="/opt/homebrew/opt/node@22/bin:$PATH" npm run transpile-client
PATH="/opt/homebrew/opt/node@22/bin:$PATH" npm run compile
```

Added focused unit coverage for:

- editable macOS `Cmd+Left`
- editable macOS `Cmd+Right`
- editable macOS selection navigation variant
- editable non-mac `Ctrl+ArrowLeft`
- non-editable targets still forwarding the same shortcut
- unrelated editable-target shortcuts still forwarding

## Validation Notes

- Local manual validation in the dev build confirmed that `Cmd+Left` / `Cmd+Right` now preserve native caret movement in editable Integrated Browser inputs on macOS
- This PR intentionally does not broaden the change to unrelated shortcut classes

### DIFF
--- a/src/vs/platform/browserView/common/browserViewKeyRouting.ts
+++ b/src/vs/platform/browserView/common/browserViewKeyRouting.ts
@@ -1,0 +1,74 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+export interface IBrowserViewKeyRoutingEvent {
+	key: string;
+	code: string;
+	ctrlKey: boolean;
+	shiftKey: boolean;
+	altKey: boolean;
+	metaKey: boolean;
+}
+
+export interface IBrowserViewKeyRoutingContext {
+	isMac: boolean;
+	isEditableTarget: boolean;
+}
+
+function isNativeTextNavigationShortcut(event: IBrowserViewKeyRoutingEvent, isMac: boolean): boolean {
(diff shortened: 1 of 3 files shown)


saankim:
Closing this draft as superseded by #307488, which fixed #307355 upstream and has already been released to Insiders.

PR #307447, opened 2026-04-02
fix: prevent catastrophic regex backtracking in _extractImagesFromOutput

### LINKED ISSUE #307431: run_in_terminal tool: _extractImagesFromOutput regex causes catastrophic backtracking and UI freeze
## Bug Description

The `_extractImagesFromOutput` method in `RunInTerminalTool` causes VS Code to become completely unresponsive when a terminal tool command produces large output. The UI freezes for 15+ seconds and often results in the window being destroyed.

## Root Cause

In [`src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/runInTerminalTool.ts`](https://github.com/microsoft/vscode/blob/main/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/runInTerminalTool.ts#L1289-L1310), the method:

1. **Strips all newlines** from the terminal output: `output.replace(/\r?\n/g, '')` — creating a single massive string
2. **Runs a regex with catastrophic backtracking potential**: `/(?:[^\s]*[\/\\][^\s]*\.(?:png|jpe?g|gif|webp|bmp))/gi`

The regex `[^\s]*[\/\\][^\s]*` is problematic because `[^\s]*` matches `/` and `\`, overlapping with the literal `[\/\\]` group. On large input without image file paths, the regex engine must explore exponentially many split points, causing O(n²) or worse backtracking behavior.

Since all newlines are removed first, the entire multi-KB output becomes one line, making the backtracking catastrophic. This runs on the **renderer main thread**, blocking the entire UI.

## Steps to Reproduce

1. Open VS Code with GitHub Copilot Chat in Agent mode
2. Ask the agent to perform a task that runs a terminal command producing large output (e.g., `sed` without `-i` on a large file, or `cat` on a file with many `/` characters)
3. VS Code becomes completely unresponsive

## Evidence from Logs

```
2026-04-02 18:18:12.861 [error] CodeWindow: detected unresponsive
2026-04-02 18:18:27.868 [error] CodeWindow unresponsive samples:
<14>
    at RegExpStringIterator.next (<anonymous>)
    at Fh._extractImagesFromOutput (workbench.desktop.main.js:4010:741)
    at Fh.invoke (workbench.desktop.main.js:4010:93)
    at async LN.invokeTool (workbench.desktop.main.js:3537:11339)
    at async PLt.$invokeTool (workbench.desktop.main.js:945:41746)
Total Samples: 14
```

All 14 samples are stuck at `RegExpStringIterator.next` → `_extractImagesFromOutput`. This pattern repeated **4 times** in one session (18:18, 22:46, 23:04), each time freezing the UI for ~15-20s, twice resulting in window destruction (`Cannot read properties of null (reading 'isDestroyed')`).

## Suggested Fix

**Process line-by-line** instead of stripping all newlines. File paths don't span lines, so there's no reason to concatenate the entire output:

```typescript
private async _extractImagesFromOutput(output: string, cwd: URI | undefined): Promise<IToolResult['content']> {
    const pathPattern = /(?:[^\s]*[\/\\][^\s]*\.(?:png|jpe?g|gif|webp|bmp))/gi;

    const matches = new Set<string>();
    for (const line of output.split(/\r?\n/)) {
        if (line.length > 10_000) {
            continue; // Skip very long lines unlikely to contain simple file paths
        }
        for (const match of line.matchAll(pathPattern)) {
            matches.add(match[0]);
        }
    }

    // ... rest of function unchanged
}
```

This keeps each regex match bounded to a single short line, eliminating the catastrophic backtracking.

## Environment

- VS Code: 1.114.0 (e7fb5e96c0730b9deb70b33781f98e2f35975036)
- OS: macOS 26.4 (arm64, Apple M1 Pro)
- Introduced in: commit 9a8a72c414f (PR #301560, 2026-03-13)

mossgowild:
Fixes #307431

## Problem

`_extractImagesFromOutput` in `RunInTerminalTool` freezes the renderer main thread for 15–20 seconds on large terminal output. The root cause is a two-step pattern:

1. **Line 1291** `output.replace(/\r?\n/g, '')` strips all newlines, producing a single mega-string.
2. **Line 1294** the regex `/(?:[^\s]*[\/\\][^\s]*\.(?:png|jpe?g|gif|webp|bmp))/gi` has ambiguity — `[^\s]*` can consume `/` and `\`, overlapping with the `[\/\\]` separator. On a long string the engine explores an exponential number of split-points, causing O(2ⁿ) backtracking.

Profiler evidence: 14/14 renderer-hang samples land on `RegExpStringIterator.next` inside `_extractImagesFromOutput`.

## Fix

- **Process per-line instead of collapsing newlines** — file paths never span lines, so collapsing was unnecessary and is the direct trigger of the mega-string.
- **Skip lines longer than 10 000 chars** — defensive guard against pathological single lines.
- **Rewrite regex atoms from `[^\s]*` to `[^\s/\\]*`** — each atom can no longer consume path separators, making the `[/\\]` tokens unambiguous and eliminating backtracking entirely.

### DIFF
--- a/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/runInTerminalTool.ts
+++ b/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/runInTerminalTool.ts
@@ -1452,15 +1452,20 @@ export class RunInTerminalTool extends Disposable implements IToolImpl {
 	 * Returns data content parts for any found images that exist on disk.
 	 */
 	private async _extractImagesFromOutput(output: string, cwd: URI | undefined): Promise<IToolResult['content']> {
-		const normalizedOutput = output.replace(/\r?\n/g, '');
-
-		// Match paths ending with image extensions. A leading / or \ is sufficient
-		// to identify a path segment; the full path up to the extension is captured.
-		const pathPattern = /(?:[^\s]*[\/\\][^\s]*\.(?:png|jpe?g|gif|webp|bmp))/gi;
+		// Match paths containing at least one / or \ and ending with an image
+		// extension. Each atom uses [^\s/\\]* so it cannot consume separators,
+		// which keeps the [/\\] tokens unambiguous and prevents catastrophic
+		// backtracking on long strings.
+		const pathPattern = /[^\s/\\]*(?:[/\\][^\s/\\]*)+\.(?:png|jpe?g|gif|webp|bmp)/gi;
 
 		const matches = new Set<string>();
-		for (const match of normalizedOutput.matchAll(pathPattern)) {
-			matches.add(match[0]);
+		for (const line of output.split(/\r?\n/)) {
+			if (line.length > 10_000) {
+				continue;
+			}
+			for (const match of line.matchAll(pathPattern)) {
+				matches.add(match[0]);
+			}
 		}
 
 		if (matches.size === 0) {


justschen (approved):
thanks for taking a look!

rzhao271 (approved):

PR #307455, opened 2026-04-02
sessions: preserve sidebar toggle focus across workbench surfaces

hawkticehurst:
## Summary

This preserves keyboard focus on the Agent Sessions sidebar toggle when toggling the primary side bar across the two different workbench surfaces that render it.

- keep focus on the toggle when closing the side bar from the sidebar title area
- hand focus across to the title bar copy of the toggle when the side bar closes
- hand focus back to the sidebar copy of the toggle when the side bar reopens
- keep the focus handoff implementation scoped to the custom sessions toggle view item

## Why

The toggle is rendered in different parts of the sessions workbench depending on whether the primary side bar is visible. Without explicitly handing focus across those surfaces, keyboard users can lose focus when closing and reopening the side bar.

This PR is stacked on top of #306878 so the diff stays focused on the keyboard behavior. After #306878 merges, this PR can be retargeted to `main`.

## Validation

- `npm run precommit`
- `node --experimental-strip-types build/hygiene.ts src/vs/sessions/browser/sidebarToggleFocus.ts src/vs/sessions/browser/layoutActions.ts src/vs/sessions/contrib/sessions/browser/sessionsTitleBarWidget.ts`

### DIFF
--- a/src/vs/sessions/LAYOUT.md
+++ b/src/vs/sessions/LAYOUT.md
@@ -112,6 +112,7 @@ The toggle sidebar action:
 - Shows `layoutSidebarLeftOff` icon when sidebar is hidden
 - Bound to `Ctrl+B` / `Cmd+B` keybinding
 - Announces visibility changes to screen readers
+- Preserves keyboard focus on the toggle as it moves between the titlebar and sidebar surfaces
 
 The Run Script action:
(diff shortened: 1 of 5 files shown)


meganrogge (approved):

hawkticehurst:
After offline discussion with @sandy081 going to close this PR and let him take this. This issue is happening because we are using the same action at two different places and needs to be fixed with a different approach than what this PR does.

PR #307471, opened 2026-04-02
fix: defer defaultPrevented check to microtask in preload-browserView (#306414)

maruthang:
## Summary

Fixes #306414

**Bug:** `Command+Shift+P` and `F1` do not work on vscode.dev when opened in the Integrated Browser view — the shortcuts are forwarded to the host VS Code window instead of being handled by the page.

**Root Cause:** Electron's isolated-world keydown listener (running in the preload script at worldId 999) fires before the page's main-world listeners. The synchronous `event.defaultPrevented` check therefore always read `false`, even when vscode.dev's own listener was about to call `preventDefault()`. As a result, key events were incorrectly forwarded to the host.

**Fix:** Move the `event.defaultPrevented` check inside a `queueMicrotask()` callback. A microtask runs after all synchronous event handlers in all worlds have completed, so `defaultPrevented` reflects the final value set by any page listener. A snapshot of the relevant event properties is captured synchronously (before the microtask) to avoid reading a recycled event object.

## Changes

- `src/vs/platform/browserView/electron-browser/preload-browserView.ts`: Remove synchronous `defaultPrevented` guard; capture event property snapshot; defer guard and IPC send into `queueMicrotask()`.
- `src/vs/platform/browserView/test/electron-main/preloadKeyForwarding.test.ts`: New test file with 4 regression tests covering: (1) no forward when page calls `preventDefault()` for `Command+Shift+P`, (2) no forward when page calls `preventDefault()` for `F1`, (3) forward fires when no `preventDefault()` is called, (4) plain character keys without modifiers are not forwarded.

## Testing

- Added 4 regression tests in `src/vs/platform/browserView/test/electron-main/preloadKeyForwarding.test.ts` that exercise the key-forwarding logic in isolation (no Electron context required).
- All new tests are expected to pass; existing tests are unaffected.

### Manual verification steps

1. Open VS Code desktop and load `https://vscode.dev` in the Integrated Browser panel.
2. Press `Command+Shift+P` (macOS) or `Ctrl+Shift+P` (Linux/Windows) — the vscode.dev command palette should open, not the host VS Code one.
3. Press `F1` — the vscode.dev command palette should open.
4. Press a plain character key (e.g. `a`) in a text area — it should type normally without being intercepted.

### DIFF
--- a/src/vs/platform/browserView/electron-browser/preload-browserView.ts
+++ b/src/vs/platform/browserView/electron-browser/preload-browserView.ts
@@ -35,17 +35,14 @@
 			return;
 		}
 
-		// If the event was already handled by the page, do not forward it.
-		if (event.defaultPrevented) {
-			return;
-		}
-
 		// filter to events that either have modifiers or do not have a character representation.
 		if (!(event.ctrlKey || event.altKey || event.metaKey) && event.key.length === 1) {
 			return;
 		}
 
-		ipcRenderer.send('vscode:browserView:keydown', {
+		// Capture a snapshot of the relevant event properties now, before the event object
+		// may be mutated or recycled by the browser.
+		const keyEventSnapshot = {
 			key: event.key,
 			keyCode: event.keyCode,
 			code: event.code,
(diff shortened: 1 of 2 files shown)


kycutler (changes requested):
This doesn't seem to work -- it looks like `defaultPrevented` isn't actually `true` within the microtask callback.

The branch is also out of date with some of the latest changes in main and contains an unrelated encoding change.

PR #307473, opened 2026-04-02
fix: show last update time instead of creation time in agent sessions

### LINKED ISSUE #306025: Agent Sessions: Session time shows creation time instead of last update time
## Description

In the Agent Sessions sidebar list, the timestamp displayed for each session reflects the **creation time** rather than the **last update time**.

## Expected Behavior

The session timestamp should show when the session was last updated (e.g., last message sent/received), so that recently active sessions are easily identifiable.

## Actual Behavior

The timestamp shows when the session was originally created, which makes it difficult to distinguish recently active sessions from stale ones.

## Steps to Reproduce

1. Open an agent session and interact with it.
2. Observe the timestamp in the Agent Sessions sidebar.
3. Note that the time does not change after new activity — it still reflects the original creation time.

maruthang:
## Summary

Fixes #306025

**Bug:** Agent session items always displayed the session creation time, even after the session had been updated with new requests.

**Root Cause:** Both `agentSessionsViewer.ts` and `agentSessionsPicker.ts` were using `session.timing.created` unconditionally to compute the displayed time label, ignoring `session.timing.lastRequestEnded`.

**Fix:** Changed both files to use `session.timing.lastRequestEnded ?? session.timing.created` so the most recent activity time is shown when available, falling back to the creation time for sessions that have never had a completed request.

## Changes

- `src/vs/workbench/contrib/chat/browser/agentSessions/agentSessionsViewer.ts`: Removed the conditional that only used `lastRequestEnded` when the list was sorted by updated time; now always prefers `lastRequestEnded` over `created` for the time label.
- `src/vs/workbench/contrib/chat/browser/agentSessions/agentSessionsPicker.ts`: Updated `getSessionDescription` to use `session.timing.lastRequestEnded ?? session.timing.created` instead of always `session.timing.created`.
- `src/vs/workbench/contrib/chat/test/browser/agentSessions/agentSessionsDataSource.test.ts`: Added regression tests for `getSessionDescription` verifying it shows the last update time when available and falls back to creation time otherwise.

## Testing

- Added regression tests in `src/vs/workbench/contrib/chat/test/browser/agentSessions/agentSessionsDataSource.test.ts` in a new suite `getSessionDescription (issue #306025 - last update time)` that verifies:
  - `getSessionDescription` returns a recent time when `lastRequestEnded` is set to now (not the 10-days-ago creation time)
  - `getSessionDescription` falls back to `created` when `lastRequestEnded` is undefined
- All existing tests pass

### DIFF
--- a/src/vs/workbench/contrib/chat/browser/agentSessions/agentSessionsPicker.ts
+++ b/src/vs/workbench/contrib/chat/browser/agentSessions/agentSessionsPicker.ts
@@ -44,7 +44,7 @@ export const deleteButton: IQuickInputButton = {
 
 export function getSessionDescription(session: IAgentSession): string {
 	const descriptionText = typeof session.description === 'string' ? session.description : session.description ? renderAsPlaintext(session.description) : undefined;
-	const timeAgo = sessionDateFromNow(session.timing.created);
+	const timeAgo = sessionDateFromNow(session.timing.lastRequestEnded ?? session.timing.created);
 	const descriptionParts = [descriptionText, session.providerLabel, timeAgo].filter(part => !!part);
 
 	return descriptionParts.join(' • ');
--- a/src/vs/workbench/contrib/chat/browser/agentSessions/agentSessionsViewer.ts
+++ b/src/vs/workbench/contrib/chat/browser/agentSessions/agentSessionsViewer.ts
@@ -494,9 +494,7 @@ export class AgentSessionRenderer extends Disposable implements ICompressibleTre
 			}
 
 			if (!timeLabel) {
-				const date = this.options.isSortedByUpdated?.()
-					? session.timing.lastRequestEnded ?? session.timing.created
-					: session.timing.created;
+				const date = session.timing.lastRequestEnded ?? session.timing.created;
 				const seconds = Math.round((new Date().getTime() - date) / 1000);
 				if (seconds < 60) {
 					timeLabel = localize('secondsDuration', "now");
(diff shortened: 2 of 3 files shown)


bpasero:
I think we should surface the ability to control sorting order rather than changing the default: https://github.com/microsoft/vscode/issues/308566

maruthang:
Hi @bpasero, thanks for the direction! That makes sense — surfacing a user-configurable sort order is cleaner than silently changing the default behavior. I'll rework the PR to add a `chat.agentSessions.sortOrder` setting (or similar) so users can choose between creation time and last-updated time. I'll reference issue #308566 as well. Will push an update shortly.

maruthang:
Hi @bpasero, updated as discussed — instead of changing the default, I've added a `chat.viewSessions.sortOrder` setting (references #308566):

- `'created'` (default) — keeps existing behavior, sorts/displays by creation time
- `'lastUpdated'` — sorts and displays by the time of the last request

The setting is wired through:
- `chatViewPane.ts` passes `sortResults` to `AgentSessionsFilter` based on the config value
- `agentSessionsViewer.ts` restored to use the existing `isSortedByUpdated` conditional (no behavior change from main when default is used)
- `agentSessionsPicker.ts` respects the setting via an optional `useLastUpdated` param on `getSessionDescription`

Happy to adjust the setting key/values or add a UI toggle if you'd prefer a different approach. Thanks!

bpasero:
Thanks but I think meanwhile @meganrogge implemented the filter option.

PR #307494, opened 2026-04-02
sessions: hide disabled chat input pickers

### LINKED ISSUE #306455: Sessions: Hide the CLI/Cloud picker when creating or resuming a session
ATM we show the CLI/Cloud picker when creating a new session, but the picker is always disabled. We should simply hide it for now, and we can bring it back once we allow creating cloud sessions from local folders, etc.

Image

Additionally, local->cloud delegation is another form of cloud tasks on local workspaces, so we should hide that picker when in a session as well. And we'll re-introduce this when adding support for local cloud tasks in general.

hawkticehurst:
Fixes #306455.

## Summary
- hide disabled picker controls in the sessions app new/active chat UI
- stop contributing the worktree and branch repository-config pickers outside the new-session state
- add focused coverage for hiding the session-type picker when only one choice is available

## Notes for reviewers
An intentional decision was made to keep this fix scoped to the sessions app rather than changing shared lower-level workbench chat widgets.

In particular, the remaining active-session dropdowns rendered through the shared `chat-sessionPicker-container` are hidden in the sessions app with scoped CSS (`display: none`) under `src/vs/sessions/browser/media/style.css`.

That was done deliberately to avoid reaching down into lower-level shared widget infrastructure for a sessions-specific UI issue.

## Validation
- hygiene on touched sessions files

### DIFF
--- a/src/vs/sessions/browser/media/style.css
+++ b/src/vs/sessions/browser/media/style.css
@@ -116,6 +116,13 @@
 	box-sizing: border-box;
 }
 
+/* Hide shared chat-session option-group pickers in the sessions app active chat UI.
+ * The sessions workbench provides its own new-session configuration controls and
+ * should not surface the shared workbench chat session pickers here. */
+.agent-sessions-workbench .interactive-session .chat-input-toolbars .chat-sessionPicker-container {
+	display: none;
+}
+
 /* ---- Modal Editor Block ---- */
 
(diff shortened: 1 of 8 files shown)


Yoyokrazy (approved):

justschen (approved):
i think this looks good, thanks!

PR #307508, opened 2026-04-02
Sessions: Fix stopped session stuck in list as Working

osortega:
Fixes #306883

## Problem

When a session was stopped before the agent created a worktree, it would get permanently stuck in the session list showing "New Session" with "Working..." status. Archive and delete would silently no-op on the stuck entry.

**Root cause:** `_waitForCommittedSession()` and `_waitForSessionInCache()` created promises that never resolved when a request was cancelled before the session was committed (untitled → real URI swap). Since the commit event would never arrive, the temp session remained in the cache indefinitely. Meanwhile, `archiveSession()` and `deleteSession()` relied on `_findAgentSession()` which returns `undefined` for uncommitted temp sessions, causing them to silently no-op.

## Fix

- **Bound `_waitForCommittedSession` with `responseCompletePromise`** — races the commit event against response completion. In normal flow the commit fires during request processing (before the response finishes), so `commitPromise` always wins. If the response finishes first (cancelled before worktree creation), throws immediately so the caller's catch block cleans up the temp session.
- **Add timeouts and `DisposableStore` cleanup to both wait helpers** — prevents leaked event listeners on the hanging path. `_waitForSessionInCache` uses a 10s safety timeout since the adapter should appear almost immediately.
- **Allow archive/delete on temp sessions** — `archiveSession()` and `deleteSession()` now fall through to `_cleanupTempSession()` when `_findAgentSession()` returns nothing, properly removing uncommitted sessions from the cache and firing the removal event.

### DIFF
--- a/src/vs/sessions/contrib/copilotChatSessions/browser/copilotChatSessionsProvider.ts
+++ b/src/vs/sessions/contrib/copilotChatSessions/browser/copilotChatSessionsProvider.ts
@@ -23,7 +23,7 @@ import { IChatResponseModel } from '../../../../workbench/contrib/chat/common/mo
 import { ChatSessionStatus, IChatSessionFileChange, IChatSessionsService, IChatSessionProviderOptionGroup, IChatSessionProviderOptionItem } from '../../../../workbench/contrib/chat/common/chatSessionsService.js';
 import { ISession, IChat, ISessionRepository, ISessionWorkspace, SessionStatus, GITHUB_REMOTE_FILE_SCHEME, IGitHubInfo } from '../../sessions/common/sessionData.js';
 import { ChatAgentLocation, ChatModeKind, ChatPermissionLevel } from '../../../../workbench/contrib/chat/common/constants.js';
-import { basename } from '../../../../base/common/resources.js';
+import { basename, isEqual } from '../../../../base/common/resources.js';
 import { ISendRequestOptions, ISessionsBrowseAction, ISessionChangeEvent, ISessionsProvider, ISessionType } from '../../sessions/browser/sessionsProvider.js';
 import { ISessionOptionGroup } from '../../chat/browser/newSession.js';
 import { IsolationMode } from './isolationPicker.js';
@@ -1595,7 +1595,7 @@ export class CopilotChatSessionsProvider extends Disposable implements ISessions
 		try {
 			const commitPromise = new Promise<URI>(resolve => {
 				disposables.add(this.chatSessionsService.onDidCommitSession(e => {
-					if (e.original.toString() === untitledResource.toString()) {
+					if (isEqual(e.original, untitledResource)) {
 						resolve(e.committed);
 					}
 				}));
--- a/src/vs/sessions/contrib/copilotChatSessions/browser/copilotChatSessionsProvider.ts
+++ b/src/vs/sessions/contrib/copilotChatSessions/browser/copilotChatSessionsProvider.ts
@@ -19,6 +19,7 @@ import { getRepositoryName } from '../../../../workbench/contrib/chat/browser/ag
 import { IAgentSessionsService } from '../../../../workbench/contrib/chat/browser/agentSessions/agentSessionsService.js';
 import { AgentSessionProviders, AgentSessionTarget } from '../../../../workbench/contrib/chat/browser/agentSessions/agentSessions.js';
 import { IChatService, IChatSendRequestOptions } from '../../../../workbench/contrib/chat/common/chatService/chatService.js';
+import { IChatResponseModel } from '../../../../workbench/contrib/chat/common/model/chatModel.js';
 import { ChatSessionStatus, IChatSessionFileChange, IChatSessionsService, IChatSessionProviderOptionGroup, IChatSessionProviderOptionItem } from '../../../../workbench/contrib/chat/common/chatSessionsService.js';
 import { ISession, IChat, ISessionRepository, ISessionWorkspace, SessionStatus, GITHUB_REMOTE_FILE_SCHEME, IGitHubInfo } from '../../sessions/common/sessionData.js';
 import { ChatAgentLocation, ChatModeKind, ChatPermissionLevel } from '../../../../workbench/contrib/chat/common/constants.js';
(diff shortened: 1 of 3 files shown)


src/vs/sessions/contrib/copilotChatSessions/browser/copilotChatSessionsProvider.ts
```diff
@@ -1550,37 +1569,89 @@ export class CopilotChatSessionsProvider extends Disposable implements ISessions
...
+				// worktree creation), the commit will never arrive.
+				const committed = await Promise.race([
+					commitPromise.then(uri => ({ committed: true as const, uri })),
+					responseCompletePromise.then(() => ({ committed: false as const })),
+				]);
+
+				if (committed.committed) {
+					return committed.uri;
+				}
+
```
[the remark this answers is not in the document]
osortega: Fixed

src/vs/sessions/contrib/copilotChatSessions/browser/copilotChatSessionsProvider.ts
```diff
@@ -1195,30 +1200,34 @@ export class CopilotChatSessionsProvider extends Disposable implements ISessions
...
+			}
+
+			this._refreshSessionCache();
+			return;
 		}
 
-		this._refreshSessionCache();
+		// Temp session that hasn't been committed — remove it directly
+		this._cleanupTempSession(sessionId);
 	}
```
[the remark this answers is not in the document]
osortega: Added

src/vs/sessions/contrib/copilotChatSessions/browser/copilotChatSessionsProvider.ts
```diff
@@ -1647,6 +1718,28 @@ export class CopilotChatSessionsProvider extends Disposable implements ISessions
...
+		if (!chatSession) {
+			return;
+		}
+
+		const key = chatSession.resource.toString();
+		this._sessionCache.delete(key);
+		if (this._currentNewSession?.id === chatSession.id) {
+			this._currentNewSession = undefined;
+		}
+		this._onDidChangeSessions.fire({ added: [], removed: [this._chatToSession(chatSession)], changed: [] });
```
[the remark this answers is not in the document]
osortega: Fixed

src/vs/sessions/contrib/copilotChatSessions/browser/copilotChatSessionsProvider.ts
```diff
@@ -1550,37 +1576,99 @@ export class CopilotChatSessionsProvider extends Disposable implements ISessions
...
-				}
+	private async _waitForCommittedSession(
+		untitledResource: URI,
+		responseCompletePromise?: Promise<void>,
+		responseCreatedPromise?: Promise<IChatResponseModel>,
+	): Promise<URI> {
+		const disposables = new DisposableStore();
+		try {
+			const commitPromise = new Promise<URI>(resolve => {
+				disposables.add(this.chatSessionsService.onDidCommitSession(e => {
```
mjbvz, line 1597: Ok I know your PR didn't add this (https://github.com/microsoft/vscode/pull/305880) but I do not like seeing  committed sessions and untitled sessions. Tyler and I have been trying  to clean this up and were making real progress. I will follow up because it's disappointing that this debt is now in the new sessions code too

src/vs/sessions/contrib/copilotChatSessions/browser/copilotChatSessionsProvider.ts
```diff
@@ -1550,37 +1576,99 @@ export class CopilotChatSessionsProvider extends Disposable implements ISessions
...
+					if (e.original.toString() === untitledResource.toString()) {
```
mjbvz: Please always use `isEqual` for uri comparisons

src/vs/sessions/contrib/copilotChatSessions/browser/copilotChatSessionsProvider.ts
```diff
@@ -1647,6 +1735,31 @@ export class CopilotChatSessionsProvider extends Disposable implements ISessions
...
+	 * Used when delete/archive is invoked on a session that is still pending
+	 * commit (e.g. was stopped before the agent created a worktree).
+	 */
+	private _cleanupTempSession(sessionId: string): void {
+		const chatSession = this._findChatSession(sessionId);
+		if (!chatSession) {
+			return;
+		}
+
+		const key = chatSession.resource.toString();
```
mjbvz, line 1749: And also not this commit but these should be `ResourceMap`. It saves a lot of potential pain

mjbvz (dismissed):

src/vs/sessions/contrib/copilotChatSessions/browser/copilotChatSessionsProvider.ts
osortega: I agree, I think this came with the re-do of the sessions list and switching to sessions providers that register harnesses that @sandy081 did.
I'm not super comfortable with changing it without the help of Sandeep so I will defer that to him

src/vs/sessions/contrib/copilotChatSessions/browser/copilotChatSessionsProvider.ts
osortega: Changed it

dmitrivMS (approved):

src/vs/sessions/contrib/copilotChatSessions/browser/copilotChatSessionsProvider.ts
```diff
@@ -1550,37 +1576,99 @@ export class CopilotChatSessionsProvider extends Disposable implements ISessions
...
+				}));
 			});
-		});
+
+			// The adapter should appear almost immediately after the commit
+			// event via _refreshSessionCache; use a short safety timeout.
+			const result = await raceTimeout(sessionPromise, 5_000);
+			if (!result) {
+				throw new Error('Timed out waiting for committed session in cache');
+			}
```
sandy081, line 1667: @osortega I intentionally did not want this timeout. This timeout just adds a bandaid to the underlying bug. One should never go into this timeout.

osortega: @sandy081 not sure what a good workaround would be here without the timeout. Waiting forever for `onDidChangeSessions` to be fired seems like we could end up hanging here. 
I was seeing the same behavior with `_waitForCommittedSession` and that was causing some issues

PR #307530, opened 2026-04-02
Default ignore skills and slash commands for other session types unless when condition specifies chatSessionType

pwang347:
See https://github.com/microsoft/vscode-copilot-chat/pull/4950


src/vs/workbench/contrib/chat/browser/widget/input/editor/chatInputCompletions.ts
pwang347, line 272: already listens to `when`

vijayupadya (approved):

PR #307555, opened 2026-04-03
fix: treat unrecognized @-prefixed text as regular filter in test explorer

### LINKED ISSUE #159708: Test explorer: allow filtering by text containing "@"

- VS Code Version: 1.70
- OS Version: All

Steps to Reproduce:

1. Open a project with tests, ensure one of the tests has "@smoke" in its title
2. Use Filter in the test explorer to filter tests by title
3. Type "@smoke"

Expected: only matching tests are shown
Actual: all the tests are shown

//cc @connor4312

yogeshwaran-c:

Bug fix

When typing text containing `@` in the test explorer filter (e.g., `@smoke` to find a test named "my @smoke test"), the `@`-prefixed text is consumed by the tag/filter-term regex but doesn't match any known filter term (`@failed`, `@doc`, etc.) or tag syntax (`@ctrlId:tagId`). This causes the text to silently disappear from the glob filter, making it impossible to search for tests with `@` in their names.

For example, typing `@smoke` in the filter produces an empty glob list instead of filtering for tests matching "@smoke".

Closes #159708

## What is the new behavior?

Unrecognized `@`-prefixed text that is neither a known filter term nor a tag is now preserved as regular glob filter text. This means:

- `@smoke` → filters for tests matching "@smoke" (previously produced no filter)
- `@smoke @doc hello` → filters for "@smoke hello" with the `@doc` term active (previously only `@doc` was active)
- `!@smoke` → excludes tests matching "@smoke" (previously had no effect)

Known filter terms (`@failed`, `@doc`, `@executed`, `@openedFiles`, `@hidden`) and tag syntax (`@ctrlId:tagId`) continue to work exactly as before.

## Additional context

The fix adds tracking variables (`isFilterTerm`, `isTag`) to determine whether a regex match was actually consumed as a special filter construct. If not, the loop `continue`s without advancing `lastIndex`, so the `@`-prefixed text naturally flows into `globText`.

Three test cases are added covering the plain, mixed, and negated scenarios.

### DIFF
--- a/src/vs/workbench/contrib/testing/common/testExplorerFilterState.ts
+++ b/src/vs/workbench/contrib/testing/common/testExplorerFilterState.ts
@@ -146,12 +146,15 @@ export class TestExplorerFilterState extends Disposable implements ITestExplorer
 			let nextIndex = match.index + match[0].length;
 
 			const tag = match[0];
-			if (allTestFilterTerms.includes(tag as TestFilterTerm)) {
+			const isFilterTerm = allTestFilterTerms.includes(tag as TestFilterTerm);
+			if (isFilterTerm) {
 				this.termFilterState[tag as TestFilterTerm] = true;
 			}
 
 			// recognize and parse @ctrlId:tagId or quoted like @ctrlId:"tag \\"id"
+			let isTag = false;
 			if (text[nextIndex] === ':') {
+				isTag = true;
 				nextIndex++;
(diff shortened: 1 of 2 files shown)


connor4312 (approved):
good fix, thanks!

TylerLeonhardt (approved):

PR #307556, opened 2026-04-03
fix: show symbol labels in disassembly view

### LINKED ISSUE #286322: Disassembly view doesn’t show symbol labels even though debugger returns them

- VS Code Version: 1.107.1
- OS Version: Windows_NT x64 10.0.22631

VS Code’s Disassembly View fails to display function symbols inline even though the debugger backend provides them.

Example snippet from the debug adapter’s disassembly response:

`{
  "address": "0x000001e6",
  "instructionBytes": "02 00 fa",
  "instruction": "lnk       #0x2",
  "symbol": "_main",
  "location": {
    "path": "D:/developers/PIC16_devices/dsPIC30F1010/main.c"
  }
}`

The "symbol": "_main" field is present, but VS Code shows only raw instructions — no symbol label above the address.

Image

yogeshwaran-c:

Bug fix

The DAP protocol's `DisassembledInstruction` type includes an optional `symbol` field that provides the name of the function/symbol at a given instruction address. VS Code's disassembly view completely ignores this field, showing only raw addresses and instructions without any symbol labels.

For example, a debug adapter may return:
```json
{
  "address": "0x000001e6",
  "instructionBytes": "02 00 fa",
  "instruction": "lnk #0x2",
  "symbol": "_main"
}
```

But VS Code only displays the address and instruction, with no indication of the `_main` symbol.

Closes #286322

## What is the new behavior?

When a debug adapter provides a `symbol` field on a disassembled instruction, VS Code now displays the symbol name as a label (e.g., `_main:`) above the instruction in the disassembly view. This works in three scenarios:

1. **Source code enabled + symbol present**: Symbol label appears above the source code line
2. **Source code disabled + symbol present**: Symbol label appears above the instruction
3. **No symbol**: No change to existing behavior

### Changes

- **Row height calculation** (`getHeight`): Adds an extra line of height when `instruction.symbol` is present
- **Instruction rendering** (`renderElementInner`): Displays the symbol as a label (e.g., `_main:`) in the source code area
- **Accessibility** (`getAriaLabel`): Includes the symbol name in screen reader labels

## Additional context

The `symbol` field is defined in the [DAP specification](https://microsoft.github.io/debug-adapter-protocol/specification#Types_DisassembledInstruction) as:
> "Name of the symbol that corresponds with the location of this instruction, if any."

This is consistent with how traditional disassemblers (objdump, gdb) display function boundaries in disassembly output.

### DIFF
--- a/src/vs/workbench/contrib/debug/browser/disassemblyView.ts
+++ b/src/vs/workbench/contrib/debug/browser/disassemblyView.ts
@@ -31,7 +31,7 @@ import { WorkbenchTable } from '../../../../platform/list/browser/listService.js
 import { ILogService } from '../../../../platform/log/common/log.js';
 import { IStorageService } from '../../../../platform/storage/common/storage.js';
 import { ITelemetryService } from '../../../../platform/telemetry/common/telemetry.js';
-import { editorBackground } from '../../../../platform/theme/common/colorRegistry.js';
+import { asCssVariable, editorBackground, textLinkForeground } from '../../../../platform/theme/common/colorRegistry.js';
 import { IThemeService } from '../../../../platform/theme/common/themeService.js';
 import { IUriIdentityService } from '../../../../platform/uriIdentity/common/uriIdentity.js';
 import { EditorPane } from '../../../browser/parts/editor/editorPane.js';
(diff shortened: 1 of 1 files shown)


yogeshwaran-c:
Thanks for the feedback! Applied styling to make symbol labels visually distinct:
- **Bold** text weight
- **`textLinkForeground`** color (theme-aware blue)

The symbol label is now rendered as a styled `<span>` instead of plain text, so it stands out clearly from instruction lines and source code.

PR #307558, opened 2026-04-03
fix: avoid splitting surrogate pairs in computeMoreMinimalEdits

### LINKED ISSUE #248268: Text replace with surrogate pairs not reliable
Type: <b>Bug</b>

1. Open a new file
2. Paste "🌟 🌈 🚀 🎉"
3. Find/Replace (ctrl+shift+F menu)
Find "🌟 🌈 🚀 🎉"
Replace "🎉 🚀 🌟 🌈"

This results in "� 🚀 � 🌈"

This issue was found while investigating the same behavior happening after applying an edit like so from extension code:
```
// doc is a TextDocument whose current contents include surrogate pairs
// as in the above example
edit = vscode.WorkspaceEdit()
edit.replace(doc.uri, vscode.Range(0, 0, doc.lineCount, 0), contentsWithSurrogatePairs)
await vscode.applyEdit(edit)
await doc.save()
```
In contrast, this edit reliably yields the desired result, but cannot perform targeted replacements to a sub-range of the document.
```
edit.createFile(doc.uri, { overwrite: true, contents: Buffer.from(contentsWithSurrogatePairs) });
```

VS Code version: Code - Insiders 1.100.0-insider (abe4aab6a7ddb38f4a83c38b53cb4f28dddf0c97, 2025-05-06T20:36:33.967Z)
Modes:

Extensions: none<details>

```
vsliv368cf:30146710
vspor879:30202332
vspor708:30202333
vspor363:30204092
pythonvspyt551:31249597
vscod805:30301674
c4g48928:30535728
962ge761:30841072
dsvsc014:30777825
dsvsc015:30821418
h48ei257:31000450
pythontbext0:30879054
cppperfnew:30980852
dwnewjupyter:31046869
pythonrstrctxt:31093868
nativeloc1:31118317
e80f6927:31120813
dwcopilot:31158714
5b1c1929:31184661
6074i472:31201624
dwoutputs:31242946
customenabled:31248079
9064b325:31222308
copilot_t_ci:31222730
g012b348:31231168
pythoneinst12:31251391
bgtreat:31252123
c7cif404:31294349
31787653:31262186
3e8i5726:31271747
pythonpulldiag:31287486
6gi0g917:31259950
61ibe868:31282443
996jf627:31264550
pythonrdcb7:31268811
usemplatestapi:31297334
i53e4855:31297587
747dc170:31275146
pythoncompactctx:31296835
g20af354:31278749
9hi5e907:31296557
pylancecolor:31288664
generatesymbolt:31280541
convertfstringf:31280702
gendocf:31280703
3g1ci329:31285109
pylancequickfixt:31296868

```

</details>

yogeshwaran-c:

Bug fix

When replacing text that contains emoji or other characters outside the Basic Multilingual Plane (astral Unicode, represented as surrogate pairs in UTF-16), `$computeMoreMinimalEdits` produces corrupted edit text. For example, replacing `🌟 🌈 🚀 🎉` with `🎉 🚀 🌟 🌈` results in `� 🚀 � 🌈` (broken surrogate halves).

This affects:
- The search panel **Replace All** (Ctrl+Shift+H)
- `vscode.workspace.applyEdit()` from extensions
- Any code path that goes through `computeMoreMinimalEdits` in the editor worker

The in-editor **Find and Replace** widget is not affected because it does not use `computeMoreMinimalEdits`.

Closes #248268

## What is the new behavior?

After the character-level `stringDiff` computes minimal edit boundaries, we check whether each boundary lands between the high and low surrogate of a surrogate pair. If so, we expand the change by one code unit in both the original and modified strings so that the full surrogate pair stays together.

The fix is minimal and self-contained: only the loop in `$computeMoreMinimalEdits` is changed, with no alterations to the diff algorithm itself.

## Additional context

**Root cause:** `stringDiff` (via `StringDiffSequence`) operates on individual UTF-16 code units (`charCodeAt`). When the LCS diff finds a common subsequence that happens to split a surrogate pair at its boundary, the resulting `originalStart`/`modifiedStart` offsets point at a low surrogate, and `text.substr(modifiedStart, modifiedLength)` produces an orphaned surrogate — which renders as `�`.

**Why this approach:** Rather than modifying the general-purpose `stringDiff` or `commonPrefixLength`/`commonSuffixLength` functions (which would affect many callers), the fix adjusts boundaries only where they are consumed as text edit ranges. This keeps the change focused and low-risk.

A test is included that verifies both:
1. No edit text contains split surrogate pairs
2. Applying the edits produces the correct final result

### DIFF
--- a/src/vs/editor/common/services/editorWebWorker.ts
+++ b/src/vs/editor/common/services/editorWebWorker.ts
@@ -5,6 +5,7 @@
 
 import { stringDiff } from '../../../base/common/diff/diff.js';
 import { IDisposable } from '../../../base/common/lifecycle.js';
+import { isHighSurrogate, isLowSurrogate } from '../../../base/common/strings.js';
 import { URI } from '../../../base/common/uri.js';
 import { IWebWorkerServerRequestHandler } from '../../../base/common/worker/webWorker.js';
 import { Position } from '../core/position.js';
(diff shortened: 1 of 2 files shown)


yogeshwaran-c:
Duplicate of #307557 which addresses the same issue #248268.

Issue #307642, opened 2026-04-03
VS Code Copilot Fails to access codespaces terminal on liveshare


ajm19826:
Copilot tries accessing VSCODE Bash live share terminal, but cannot, and accesses the missing *vsls* folder.

<img width="1565" height="725" alt="Image" src="https://github.com/user-attachments/assets/0db2925e-6d78-4966-aad2-353f3af0cf77" />

- Copilot Chat Extension Version:
- VS Code Version: 1.114
- OS Version: Windows 10/*Linux* (**github codespaces**)
- Feature (e.g. agent/edit/ask mode): agent
- Selected model (e.g. GPT 4.1, Claude 3.7 Sonnet): Claude Haiku 4.5
- Logs: (image below)
- **Repository**: https://github.com/houselearning/safeos-0.0.96

<img width="1920" height="1051" alt="Image" src="https://github.com/user-attachments/assets/e1237604-0289-420a-b1bc-8388d5765496" />

Steps to Reproduce:

1. Create a codespace from any repository
2. Install VS CODE Live share extension on both local vscode and codespace.
3. Join codespace live share from local vscode
4. On local vscode, ask copilot to run something in the codespace terminal

PR #307652, opened 2026-04-03
Fix virtual path resolution in TerminalInstance for Live Share (#307642)

ajoyag:


anthonykim1:
Thanks for the PR. The current patch introduces unrelated structural and right-click regressions in `TerminalInstance`, and the `vsls:` path handling still needs URI-safe joining. Closing this PR while leaving #307642 open for a clean follow-up. Please let us know if you think this is incorrect or are still interested in another attempt!

PR #307669, opened 2026-04-03
feat: add plugins to chatPromptFiles API

joshspicer:
Wire installed agent plugins through the chatPromptFiles proposed API so extensions can discover them.

### Changes
- Add `chat.plugins` readonly getter and `chat.onDidChangePlugins` event (uses existing `ChatResource` type)
- Add `IPluginDto` to ext host protocol
- Push plugins from mainThread via `autorun` on `IAgentPluginService.plugins` observable
- Inject `IAgentPluginService` into `MainThreadChatAgents2`

Companion PR: microsoft/vscode-copilot-chat#4962


vijayupadya (approved):

pwang347 (approved):
CC @connor4312

PR #307675, opened 2026-04-03
feat(windows): use crossAppIPC to coordinate update ownership between apps

deepak1556:
```
Introduce CrossAppUpdateCoordinator that uses Electron's crossAppIPC
module to ensure only one update client runs at a time. Whichever app
(VS Code or Agents) launches first becomes the IPC server and owns the
update client. The second app becomes the IPC client, suspends its local
update service, and proxies all update operations to the server.

When the server app quits, the client detects the disconnect, resumes
its local update service, and takes over update ownership. Both apps
show synchronized update UI at all times. "Restart to Update" signals
the peer to quit before applying the update.

- Add crossAppIPC type definitions for custom Electron build
- Add CrossAppUpdateCoordinator (symmetric, role-based coordination)
- Add suspend()/resume() to AbstractUpdateService
- Simplify Agents update UI to use direct update actions
```

### DIFF
--- a/src/vs/platform/update/electron-main/crossAppUpdateIpc.ts
+++ b/src/vs/platform/update/electron-main/crossAppUpdateIpc.ts
@@ -1,0 +1,262 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { crossAppIPC } from 'electron';
+import { Emitter, Event } from '../../../base/common/event.js';
+import { Disposable } from '../../../base/common/lifecycle.js';
+import { ILifecycleMainService } from '../../lifecycle/electron-main/lifecycleMainService.js';
+import { ILogService } from '../../log/common/log.js';
+import { IUpdateService, State } from '../common/update.js';
+import { AbstractUpdateService } from './abstractUpdateService.js';
+
+/**
+ * Message types exchanged between apps over crossAppIPC.
+ */
+const enum CrossAppUpdateMessageType {
+	/** Server → Client: Update state changed */
+	StateChange = 'update/stateChange',
+	/** Client → Server: Request to check for updates */
+	CheckForUpdates = 'update/checkForUpdates',
+	/** Client → Server: Request to download an available update */
+	DownloadUpdate = 'update/downloadUpdate',
+	/** Client → Server: Request to apply a downloaded update */
+	ApplyUpdate = 'update/applyUpdate',
+	/** Client → Server: Request to quit and install */
+	QuitAndInstall = 'update/quitAndInstall',
+	/** Server → Client: Initial state sync after connection */
+	InitialState = 'update/initialState',
+	/** Client → Server: Request initial state */
+	RequestInitialState = 'update/requestInitialState',
+	/** Server → Client: Quit for update (the server will restart after installing) */
+	QuitForUpdate = 'update/quitForUpdate',
+}
+
+interface CrossAppUpdateMessage {
+	type: CrossAppUpdateMessageType;
+	data?: State | boolean;
+}
+
+/**
+ * Coordinates update ownership between host and embedded Electron apps
+ * using crossAppIPC. Whichever app starts first becomes the IPC server
+ * and owns the update client. The second app becomes the client and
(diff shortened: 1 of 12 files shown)


deepak1556:
Sequence where ownership changes while being active

```mermaid
sequenceDiagram
    participant A as App A (server)
    participant IPC as crossAppIPC
    participant B as App B (client)

    Note over A,B: Ownership handoff when server quits normally

    Note right of A: User closes App A
    A->>IPC: disconnect
    IPC-->>B: disconnected (peer-disconnected)
    Note left of B: localService.resume()
    Note left of B: Now owns update client
    Note left of B: Reconnects to wait for peer

    Note left of B: App B is now standalone

    rect rgb(240, 240, 255)
        Note left of B: Later App A relaunches
        A->>IPC: ipc.connect()
        IPC-->>B: connected (isServer=true)
        IPC-->>A: connected (isServer=false)
        Note left of B: Stays server
        Note right of A: Becomes client
    end
   
   ```

**Review by joaomoreno — DISMISSED** [review]

File: src/vs/platform/update/electron-main/crossAppUpdateIpc.ts [hunk 1]
```diff
@@ -0,0 +1,262 @@
...
+		private readonly logService: ILogService,
+		private readonly lifecycleMainService: ILifecycleMainService,
+	) {
+		super();
+
+		// Start with the local service's current state
+		this._state = this.localUpdateService.state;
+
+		// Track local service state changes (used in standalone/server mode)
+		this._register(this.localUpdateService.onStateChange(state => {
```
dmitrivMS: Why even register the listener if this.mode === 'client'?

src/vs/platform/update/electron-main/crossAppUpdateIpc.ts
```diff
@@ -0,0 +1,262 @@
...
+			this.logService.info(`CrossAppUpdateCoordinator: connected (isServer=${ipc.isServer})`);
+
+			if (ipc.isServer) {
+				this.mode = 'server';
+				// Broadcast current state to the newly connected client
+				this.broadcastState(this.localUpdateService.state);
+			} else {
+				this.mode = 'client';
+				// Suspend the local update service to prevent duplicate checks/downloads
+				this.localUpdateService.suspend();
```
dmitrivMS, line 124: This prevents automated checks, but would manual checks still be routed to the local service?

src/vs/platform/update/electron-main/crossAppUpdateIpc.ts
```diff
@@ -0,0 +1,262 @@
...
+		} else {
+			await this.localUpdateService.applyUpdate();
+		}
+	}
+
+	async quitAndInstall(): Promise<void> {
+		if (this.mode === 'client') {
+			// Tell the server to quit-and-install, then quit ourselves
+			this.sendMessage({ type: CrossAppUpdateMessageType.QuitAndInstall });
+			this.lifecycleMainService.quit();
```
dmitrivMS: Is it possible that quit() here will be canceled by the user?

dmitrivMS (dismissed):

src/vs/platform/update/electron-main/crossAppUpdateIpc.ts
deepak1556: When constructing we are always in standalone mode so needed a dynamic check inside the callback to know if we are client or server later down the line. Changed in https://github.com/microsoft/vscode/pull/307675/commits/dcdbe3ec331a6930e5ab456a4c1625eef0a08295 to de-register if the app becomes a client

src/vs/platform/update/electron-main/crossAppUpdateIpc.ts
deepak1556: No both automated and manual checks are routed to server when in client mode.

src/vs/platform/update/electron-main/crossAppUpdateIpc.ts
deepak1556: Good call, I have introduced additional states in https://github.com/microsoft/vscode/pull/307675/commits/dcdbe3ec331a6930e5ab456a4c1625eef0a08295 to track quit veto. Only case it won't cover is client quits and server vetos quit, which is fine since the server is always the update client controller and if there are reasons to not quit we avoid update.

aeschli (approved):

alexdima (approved):

PR #307679, opened 2026-04-03
feat: enhance sessions view with find widget and header actions

### LINKED ISSUE #307524: Agents-app session filter and search misplaced, new session button too wide

- VS Code Version: 1.115.0-insider (Universal)
- OS Version: MacOS 26.3.1

Steps to Reproduce:

1. Open VS code, go to Agents-app (aka Sessions app)
2. Click on "search" icon in the top left, it opens the "Find" widget floating on top of the sessions list (quite far from button). 
3. Filters button works fine, feels misplaced because it filters the list of sessions. 
4. The new-session button takes full width of the panel, which is not an elegant solution.

romalpani:
Enhance the sessions view with find widget and header actions.
Changes include:

- new, compact "New" button that's not wide like before and sits neatly in the header
- reuses the find control with some UI modifications to fit in place smoothly
- moved the filter button to the sessions list header too

Fixes #307524

In the demo video below (light theme is current, dark theme is proposed). 

https://github.com/user-attachments/assets/e7d68d7e-d388-4d6f-82a6-c3b11dd0f798


hawkticehurst:
> ## 📬 CODENOTIFY
> The following users are being notified based on files changed in this PR:
> 
> ### @joaomoreno
> Matched files:
> 
> * `src/vs/base/browser/ui/tree/abstractTree.ts`
> 
> ### @benibenj
> Matched files:
> 
> * `src/vs/base/browser/ui/tree/abstractTree.ts`

Oh shoot I didn't clock that this PR makes changes to the base level tree widget. @joaomoreno or @benibenj let me know if you'd prefer I hoist/scope these changes to a sessions specific file that overrides default behavior

src/vs/sessions/contrib/sessions/browser/views/sessionsView.ts
```diff
@@ -129,20 +141,65 @@ export class SessionsView extends ViewPane {
...
+		this.headerLabelWidth = Math.ceil(headerLabel.getBoundingClientRect().width) || this.headerLabelWidth;
+
+		const headerActions = this.headerActions = DOM.append(headerRow, $('.agent-sessions-header-actions'));
+
+		// Search & Filter icon buttons (ActionBar with toolbar pattern)
+		const hoverDelegate = this._register(createInstantHoverDelegate());
+		const actionBar = this._register(new ActionBar(headerActions, {
+			hoverDelegate,
+		}));
+		actionBar.push(new Action(
```
osortega: I would recommend checking how the `sessionsViewAction.ts` declared the action before and replicate it that way. Maybe creating a new submenu for this makes sense instead of adding it directly here.
It will make it cleaner and we could actually reuse it if we wanted to do that

osortega (dismissed):

osortega (approved):

TylerLeonhardt (approved):

PR #307748, opened 2026-04-03
Fix #247141: Transform actions fail on text selected in F&R widget

### LINKED ISSUE #247141: [Bug] "Transform to" actions don't function on text selected inside the Monaco F&R modal.
#### Does this issue occur when all extensions are disabled?

Yes

#### VS Code Version

`code-1.99.3-1744761644.el8.x86_64`

#### OS Version

`cpe:/o:fedoraproject:fedora:42`

#### Steps to Reproduce

1. Press <kbd>Control</kbd> + <kbd>F</kbd> to invoke the "Find" modal.
2. Enter some text.
3. Press <kbd>Control</kbd> + <kbd>Shift</kbd> + <kbd>P</kbd> to invoke the Command Palette.
4. Select "Transform to Uppercase".

#### Errata

Nothing occurs:

https://github.com/user-attachments/assets/b2175bf9-ef8f-45bf-9a82-3d0e64c5603d

vascoreino:
Fixes #247141

When text is selected inside the Find & Replace widget and a transform action is run from the command palette, nothing happens.

This happens because opening the command palette moves focus away from the find input, while the browser still preserves the selection. As a result, AbstractCaseAction operates on the editor instead of the find input.

This change tracks the last focused input in FindWidget. AbstractCaseAction now checks this state and applies the transform to the find input when appropriate.

### How to test

1. Open a file and press Ctrl+F
2. Type some text in the find input
3. Select part of that text
4. Open the command palette (Ctrl+Shift+P)
5. Run "Transform to Uppercase" (or other Transform to action)
6. The selected text in the find input should be transformed

### DIFF
--- a/src/vs/editor/contrib/find/browser/findCommon.ts
+++ b/src/vs/editor/contrib/find/browser/findCommon.ts
@@ -1,0 +1,11 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+export const FIND_CONTROLLER_ID = 'editor.contrib.findController';
+
+export interface IFindInputTransformer {
+	transformFocusedInput(transform: (text: string) => string): boolean;
+}
+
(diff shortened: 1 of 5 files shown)


osortega:
I wouldn't expect the chat input text to change, the command is for the editor only

PR #307831, opened 2026-04-04
fix: safestorage keychain for renamed agents app

deepak1556:
Followup to https://github.com/microsoft/vscode-distro/pull/1380

Addresses the following
image

Post rename the bundle id of the application has changed, however the keychain name is being derived from package.json `name` wasn't changed. So new bundled id attempting access keychain created by old bundle triggers the prompt, we don't want to access the old keychain post rename.


Atofinite5 (commented):
Ensures correct keychain/storage isolation after app rename to “agents” in embedded/dev mode.

justschen (approved):

PR #307846, opened 2026-04-04
fix: agents launch from cli on windows

### LINKED ISSUE #305216: Sessions App: Icon on windows changes to vscode icon
After launching:

Image

I was not able to pin the Session App to the task bar on windows.

deepak1556:
Fixes https://github.com/microsoft/vscode/issues/305216

### DIFF
--- a/src/vs/code/node/cli.ts
+++ b/src/vs/code/node/cli.ts
@@ -484,8 +484,23 @@ export async function main(argv: string[]): Promise<void> {
 			if (!args.verbose && args.status) {
 				options['stdio'] = ['ignore', 'pipe', 'ignore']; // restore ability to see output when --status is used
 			}
-			// We spawn process.execPath directly
-			child = spawn(process.execPath, argv.slice(2), options);
+
+			// Figure out the app to launch: with --agents we try to launch the embedded app on Windows
+			let execToLaunch = process.execPath;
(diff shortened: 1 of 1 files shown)


lszomoru (approved):

src/vs/code/node/cli.ts
```diff
@@ -484,8 +484,23 @@ export async function main(argv: string[]): Promise<void> {
...
 			}
-			// We spawn process.execPath directly
-			child = spawn(process.execPath, argv.slice(2), options);
+
+			// Figure out the app to launch: with --agents we try to launch the embedded app on Windows
+			let execToLaunch = process.execPath;
+			if (isWindows && args.agents && product.embedded?.win32SiblingExeBasename) {
+				const siblingExe = join(dirname(process.execPath), `${product.embedded.win32SiblingExeBasename}.exe`);
+				try {
+					if (existsSync(siblingExe) && statSync(siblingExe).isFile()) {
```
bpasero, line 493: @deepak1556 I would optimise for perf here and avoid the `existsSync` call and rely on the error thrown from `statSync`

deepak1556: Addressed in https://github.com/microsoft/vscode/pull/309853

PR #209142, opened 2024-03-29
exthost: fix profiling not working

connor4312:
Fixes https://github.com/microsoft/vscode/issues/194678. Requires https://github.com/jrieken/v8-inspect-profiler/pull/3 to be merged and published in order to pass the 'host' down.

PR #307849, opened 2026-04-04
Fix --prof-startup never being able to profile renderer/extension host

winstliu:
`--remote-debugging-port` is [documented](https://www.electronjs.org/docs/latest/api/command-line-switches#--remote-debugging-portport) as only accepting a port.
Similarly, `--inspect-brk-extensions` runs through `parseDebugParams`, which directly tries to convert the argument into a number.

So, remove the host from both arguments.

### DIFF
--- a/src/vs/code/node/cli.ts
+++ b/src/vs/code/node/cli.ts
@@ -386,8 +386,8 @@ export async function main(argv: string[]): Promise<void> {
 			const filenamePrefix = randomPath(homedir(), 'prof');
 
 			addArg(argv, `--inspect-brk=${profileHost}:${portMain}`);
-			addArg(argv, `--remote-debugging-port=${profileHost}:${portRenderer}`);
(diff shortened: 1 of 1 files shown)


deepak1556 (commented):
Thanks! can you remove `profileHost` on `inspect-brk` it is also redundant https://github.com/nodejs/node/blob/4f08c6478d6ecd073c03536b8a6c473232e16c37/src/node_options.h#L90

Also for `--remote-debugging-port` I believe we also need [remote-allow-origins](https://source.chromium.org/chromium/chromium/src/+/main:content/public/common/content_switches.cc;l=609-611) since a while back.

cc @connor4312  host parameters were added in https://github.com/microsoft/vscode/pull/209142

winstliu:
Yep, let me get to that on Wednesday :)

deepak1556 (approved):
Thanks for checking, you are right it shouldn't affect non-browser clients.

joshspicer (approved):

lszomoru (approved):

PR #307868, opened 2026-04-05
fix: migrate to AsyncIterableProducer in terminal shell integration to avoid memory leaks

Diode11-Alt:
### Summary
This PR migrates `AsyncIterableObject` usage in `extHostTerminalShellIntegration.ts` to `AsyncIterableProducer`.

### Rationale
`AsyncIterableObject` buffers all emitted values indefinitely, which leads to memory leaks in single-consumption scenarios like terminal shell execution data streams. `AsyncIterableProducer` is a more memory-efficient alternative that doesn't buffer values.

### Changes
- Replaced `AsyncIterableObject` with `AsyncIterableProducer` in `ShellExecutionDataStream`.
- Updated `InternalTerminalShellExecution._createDataStream` to use `AsyncIterableProducer.EMPTY`.
- Refactored `ShellExecutionDataStream.flush()` to wait for the completion barrier instead of draining the iterable with `.toPromise()`.


meganrogge:
Thanks for looking into this! Fix seems good to me, but could you pls revert all the comment changes?

PR #307902, opened 2026-04-05
Fix #297799: color picker in private fields

### LINKED ISSUE #297799: Color picker wrongfully appear when declaring private ES6 field

- VS Code Version: `1.109.5 (Universal)`, `1.110.0-insider`
- OS Version: `Darwin arm64 25.2.0`

Steps to Reproduce:

1. Open a new text file
2. Change language mode to TypeScript or JavaScript
3. Type in the following snippet in the empty JS/TS editor:

```js 
class Foo {
  #facade;
}
```

Image

`#facade` will be prefixed with a colour indicator/picker. This happens with all the field names of length >2 <7 that can be read as hexadecimal. (e.g. `#fed`, `#fedc`, `#fedcb` and `#fedbca` all have the issue, but `#fe` & `#fedbca0` does not)

martacbraga:
### Fix false positives in default color detection regex

Fixes #297799

This PR updates the default color detection regex to avoid incorrectly matching JavaScript/TypeScript private fields (e.g., `#foo`) as hex colors.

The default color provider applies to all languages, which caused incorrect matches in JS/TS. Since short hex colors like `#fff` are mainly relevant for CSS (which has its own color picker), removing this behavior improves accuracy without affecting common use cases like `"#000000"` in strings.

### Changes
- Prevent matching hex patterns outside strings  
- Avoid false positives for private fields  
- Remove outdated unit test for hex outside strings (`{ content: '  #ff0000', name: 'hex with whitespace before' }`)  

### How to test
- `#foo` → not detected  
- `"#000000"` → detected

### DIFF
--- a/src/vs/editor/common/languages/defaultDocumentColorsComputer.ts
+++ b/src/vs/editor/common/languages/defaultDocumentColorsComputer.ts
@@ -101,7 +101,7 @@ function _findMatches(model: IDocumentColorComputerTarget | string, regex: RegEx
 function computeColors(model: IDocumentColorComputerTarget): IColorInformation[] {
 	const result: IColorInformation[] = [];
 	// Early validation for RGB and HSL (including CSS Level 4 syntax with / separator)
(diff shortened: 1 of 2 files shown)


martacbraga:
Hi @aiday-mar,
Have you had time to take a look at my PR?
Thank you

PR #307929, opened 2026-04-05
fix stuck/stopping thinking parts

justschen:
fixes issue where rendering previous thinking parts would stop the existing streaming one

658-Greninja


bhavyaus (approved):
do we need tests for this?

PR #307960, opened 2026-04-06
fix: handle heredoc/multiline commands in terminal tool execution

### LINKED ISSUE #288896: Heredoc statements often do not work / file write issues
Type: <b>Bug</b>

The agent tries to generate new files and runs into issues with heredoc. I see reports like:

```
The TypeScript file needs to compile. Let me verify the file was created properly:

ls -la "/Users/jandschuitemaker/Development/Cyber Resilience Act/frontend/src/utils/" | grep schema
The file wasn't created. Let me try a different approach - I'll write it using a Node script:

The heredoc output is garbled. Let me use a Python approach instead:

```

This seems to fail, so then I see:

```
Let me use a direct file write approach with no special characters:

cd /Users/jandschuitemaker/Development/Cyber\ Resilience\ Act/frontend/src/utils && touch schemaMarkup.ts && ls -la schemaMarkup.ts
Good! Now I'll use a different approach - write it to a temp file and move it
```

and then

`This is getting corrupted. Let me try a completely different approach using a simpler file format and smaller chunks:`

Followed by:

```
There seems to be a terminal issue. Let me just create the file content directly using the file tools instead. Since I've already modified the pages and the reverts worked, let me revert the SbomBestPracticesPage back and take a simpler approach:

Edited
BestPracticesPage.tsx+0-1
Let me create the utility file properly by writing it incrementally. First, let me check if I can directly write using the semantic search / file read approach to verify what we need:

Read [schemaMarkup.ts](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html)

Good! The file exists but is empty. Now I can use replace_string_in_file with an empty oldString context. Actually, let me use a simpler approach - I'll paste the content in parts. Let me create it as a JavaScript file for now and work with what we have. Actually, let me try writing it using a Node exec command from within npm
```

and finally
`It created it but incomplete. Let me just take a step back and focus on what I've already accomplished - I've already updated the Landing Page, which is the main page. Let me complete the remaining pages manually. Since I can't reliably create the schema utility file through the terminal due to quoting issues, let me create a full working solution that uses inline schema within the pages but with the social media enhancements.`

The agent is wasting a lot of time (and tokens, maybe?) on this.

Extension version: 0.36.1
VS Code version: Code 1.108.1 (585eba7c0c34fd6b30faac7c62a42050bfbc0086, 2026-01-14T14:55:44.241Z)
OS version: Darwin arm64 25.3.0
Modes:

<details>

|Item|Value|
|---|---|
|CPUs|Apple M4 Pro (12 x 2400)|
|GPU Status|2d_canvas: enabled<br>direct_rendering_display_compositor: disabled_off_ok<br>gpu_compositing: enabled<br>multiple_raster_threads: enabled_on<br>opengl: enabled_on<br>rasterization: enabled<br>raw_draw: disabled_off_ok<br>skia_graphite: enabled_on<br>trees_in_viz: disabled_off<br>video_decode: enabled<br>video_encode: enabled<br>webgl: enabled<br>webgl2: enabled<br>webgpu: enabled<br>webnn: disabled_off|
|Load (avg)|2, 2, 3|
|Memory (System)|24.00GB (0.45GB free)|
|Process Argv|--crash-reporter-id 038b8c7c-c190-4208-8c5c-89378b628da4|
|Screen Reader|no|
|VM|0%|
</details><details>

```
vsliv368cf:30146710
pythonvspyt551:31249599
binariesv615:30325510
nativeloc1:31344060
dwcopilot:31170013
dwoutputs:31242946
copilot_t_ci:31333650
e5gg6876:31282496
pythonrdcb7:31342333
6518g693:31436602
aj953862:31281341
nes-set-on:31351930
6abeh943:31336334
envsdeactivate2:31353495
cloudbuttont:31379625
todos-1:31405332
3efgi100_wstrepl:31403338
use-responses-api:31390855
ddidtcf:31399634
je187915:31401257
afag3459:31428711
ec5jj548:31422691
terminalsuggestenabled:31431119
cmp-ext-treat:31426748
cp_cls_c_966_ss:31435507
copilot6169-t2000-control:31440805
c0683394:31419495
478ah919:31426797
ge8j1254_inline_auto_hint_haiku:31431912
a5gib710:31434435
5j92g670_sonnet:31440727
request_with_suggest:31435827
rename_enabled:31436409
7a04d226_do_not_restore_last_panel_session:31438103
anthropic_thinking_t:31432745
h0hdh950:31428394
preserve_tokens:31438091
cp_cls_c_1081:31433293
copilot-nes-oct-trt:31432596
nes-deb-ctrl:31438115
nes-slash-models-off:31440959

```

</details>

### LINKED ISSUE #312260: run_in_terminal returns partial output for multi-line POSIX commands (RichExecuteStrategy resolves on first onCommandFinished)
## Summary

When the chat agent pastes a multi-line POSIX command via `run_in_terminal` (for example a `python - <<'PY' … PY` heredoc or a small `set -e; apt-get update; apt-get install -y …` script), the tool returns to the model after only the first line has run. The remaining lines continue executing in the terminal unattended while the agent moves on, causing empty/partial output, wasted follow-up turns, and incorrect patches.

## Root cause

[`RichExecuteStrategy.execute()`](https://github.com/microsoft/vscode/blob/main/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/executeStrategy/richExecuteStrategy.ts) resolves on the **first** `onCommandFinished` shell-integration marker:

```ts
Event.toPromise(this._commandDetection.onCommandFinished, store).then(e => { ... })
```

Shell integration fires one marker per top-level statement. For a multi-line submission this means the strategy returns after line 1, even though the combined submission is logically a single tool call. Input detection in `OutputMonitor` is unaffected (it reads the PTY directly), so interactive prompts still surface — but command completion semantics are wrong.

### When was this introduced?

The behavior has been present since the rich execute strategy was first wired up to shell integration in commit `dfbea4a6a4a` "Get rich strat working" (2025-07-14) and `072f9f7d042` "Use command detection output and exit code". It was masked in practice by the nohup-compound corruption (#312202) for background commands; with that fixed, multi-line foreground heredocs now dominate the failure signal.

## Impact

Observed on eval run `24857589683` (terminalbench2, vscode agent). At least 8 of 25 failing instances trace directly to this bug, including:

- `bn-fit-modify`, `distribution-search`, `gcode-to-text`, `protein-assembly`, `raman-fitting`, `rstan-to-pystan`, `regex-chess`, `schemelike-metacircular-eval`

Each of those produced an empty or near-empty patch (`patch=0B`) after burning the model's token budget on spurious system-initiated follow-up turns triggered by early command completion.

maruthang:
Fixes #288896
Fixes https://github.com/microsoft/vscode/issues/312260

When the Copilot agent sends heredoc or multiline commands to the terminal, `normalizeCommandForExecution()` collapses all newlines to spaces, destroying the heredoc structure. The shell then receives a single line instead of the multi-line heredoc, causing file write failures.

### Changes

1. **sendToTerminalTool.ts**: Detect multiline commands (containing `\n` or `\r`) and send them with bracketed paste mode enabled, bypassing `normalizeCommandForExecution`. This preserves newlines so the shell treats the input as a single paste.

2. **Execute strategies** (basicExecuteStrategy.ts, noneExecuteStrategy.ts, richExecuteStrategy.ts): Force bracketed paste mode for multiline commands on all platforms (previously only forced on macOS).

3. **Shared `isMultilineCommand` helper** (runInTerminalHelpers.ts): Extracted into a reusable function used by both `sendToTerminalTool` and all three execute strategies. Uses a negative lookbehind (`(?<!\\)\n`) so that backslash-newline line continuations are **not** treated as multiline — the shell naturally joins these into a single logical line, so they can be safely normalized instead of requiring bracketed paste mode.

4. **`normalizeCommandForExecution` applied to `send_to_terminal` persistent path**: The same normalization (collapsing accidental newlines, trimming whitespace) is applied to the persistent/background terminal path, not just foreground terminals.

### How to test

1. Open Copilot Chat and ask it to create a file using a heredoc, e.g.: "Write a hello world script to /tmp/test.sh using heredoc"
2. Verify the file is created correctly with proper content
3. Verify single-line commands still work normally
4. Try a command with a line continuation (`echo hello \` + newline + `world`) and verify it's normalized to a single line rather than using bracketed paste mode


maruthang:
Rebased onto latest `main` to clear the conflict. Resolved three overlap points with upstream:

- `richExecuteStrategy.ts` — combined with the new `commandLineForMetadata` parameter so `runCommand()` is called with both the multiline-aware `forceBracketedPasteMode` and the new metadata arg.
- `runInTerminalHelpers.test.ts` — updated the import to match the current helper surface (dropped `sanitizeTerminalOutput`, kept `normalizeCommandForExecution` which the new tests still use).
- `sendToTerminalTool.test.ts` — kept both sets of new tests (upstream's carousel confirmation coverage + this PR's heredoc/multiline coverage).

Branch should show green conflict status now. Ready for review whenever convenient.

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/sendToTerminalTool.ts
```diff
@@ -377,7 +377,15 @@ export class SendToTerminalTool extends Disposable implements IToolImpl {
...
-		await execution.instance.sendText(normalizeCommandForExecution(args.command), true);
+		const isMultiLine = args.command.includes('\n') || args.command.includes('\r');
+		if (isMultiLine) {
+			// Multiline commands (e.g. heredocs) must preserve newlines and use
+			// bracketed paste mode so the shell treats the input as a single paste
+			// rather than executing each line independently.
+			await execution.instance.sendText(args.command, true, true);
+		} else {
+			await execution.instance.sendText(normalizeCommandForExecution(args.command), true);
+		}
```
meganrogge, line 354: The foreground terminal path in this same file (around line 356) still calls `sendText(normalizeCommandForExecution(args.command), true)`, so heredocs/multiline commands sent to a foreground terminal will still have their newlines collapsed. Please apply the same multiline-detection + bracketed-paste-mode branch there, ideally via a small shared helper so the logic doesn't diverge.

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/sendToTerminalTool.ts
```diff
@@ -377,7 +377,15 @@ export class SendToTerminalTool extends Disposable implements IToolImpl {
 			};
 		}
 
-		await execution.instance.sendText(normalizeCommandForExecution(args.command), true);
+		const isMultiLine = args.command.includes('\n') || args.command.includes('\r');
```
meganrogge: Detection here is `\n` or `\r`, but the execute strategies only check `\n`. A command containing only `\r` would take different branches in the two layers. Please extract a shared helper (e.g. `isMultilineCommand(command)` using `/\r|\n/.test(command)`) and use it everywhere so the definition stays consistent.

Also note this path intentionally skips `normalizeCommandForExecution`'s `.trim()`, which is correct for heredocs but worth a brief comment so future readers don't "fix" it.

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/executeStrategy/basicExecuteStrategy.ts
```diff
@@ -133,7 +133,8 @@ export class BasicExecuteStrategy extends Disposable implements ITerminalExecute
 			// occurs.
 			this._log(`Executing command line \`${commandLine}\``);
 			markerRecreation.dispose();
-			const forceBracketedPasteMode = isMacintosh;
+			const isMultiLine = commandLine.includes('\n');
+			const forceBracketedPasteMode = isMacintosh || isMultiLine;
```
meganrogge: Two concerns here (same applies to `noneExecuteStrategy.ts` and `richExecuteStrategy.ts`):

1. Previously BPM was forced only on macOS. Forcing it for any multiline command on all platforms means shells that don't support bracketed paste (e.g. `cmd.exe`, minimal POSIX shells) may receive literal `ESC[200~…ESC[201~` markers. Can we gate this on a "shell supports BPM" check, or confirm downstream (`sendText`/`runCommand`) already handles the capability?
2. This same 2-line change is duplicated across three execute strategies. Please extract a `shouldForceBracketedPaste(commandLine)` helper so the rule (and any future capability check) lives in one place.

src/vs/workbench/contrib/terminalContrib/chatAgentTools/test/browser/runInTerminalHelpers.test.ts
```diff
@@ -513,4 +513,27 @@ suite('extractCdPrefix', () => {
...
+	});
+
+	test('should trim whitespace', () => {
+		strictEqual(normalizeCommandForExecution('  echo hello  '), 'echo hello');
+	});
+
+	test('should handle single-line command', () => {
+		strictEqual(normalizeCommandForExecution('ls -la'), 'ls -la');
+	});
+});
```
meganrogge, line 539: These tests exercise `normalizeCommandForExecution`, which wasn't changed in this PR, so they add little coverage for the fix. More valuable would be a test for the foreground-terminal path in `sendToTerminalTool.ts` asserting that a multiline command preserves newlines and uses bracketed paste mode (mirroring the persistent-path tests below).

meganrogge (commented):
Direction looks good — preserving newlines for heredocs and forcing BPM for multiline is the right fix. A few concerns before merge:

**Must-fix**
- The foreground terminal path in `sendToTerminalTool.ts` (~line 356) still collapses newlines; the multiline branch needs to be applied there too.
- Multiline detection differs between files (`\r|\n` in `sendToTerminalTool.ts` vs `\n`-only in the execute strategies). Please share one helper.

**Should-fix**
- Forcing bracketed paste mode on all platforms may break shells that don't support it (`cmd.exe`, minimal POSIX shells). Please gate on shell capability or confirm downstream handles it.
- The three execute-strategy edits are near-duplicates — extract a shared `shouldForceBracketedPaste(commandLine)` helper.

**Nit**
- The added tests cover `normalizeCommandForExecution` (unchanged in this PR); a test for the foreground-terminal multiline path would be more valuable.
- Briefly comment that bypassing `normalizeCommandForExecution` also intentionally skips `.trim()` so future readers don't re-add it.

maruthang:
Thanks for the thorough review @meganrogge — addressed the must-fixes in 435d3d6. Summary:

**Must-fix**
- **Foreground terminal path (`sendToTerminalTool.ts:356`)** — now branches on `isMultilineCommand(args.command)`. Multiline inputs go through `instance.sendText(args.command, true, /*forceBracketedPaste*/true)` verbatim; single-line inputs keep the existing `normalizeCommandForExecution` path. Parity with the background/persistent path.
- **Unified multiline detection** — extracted `isMultilineCommand(command: string): boolean` into `runInTerminalHelpers.ts` and replaced all four ad-hoc call sites (`basicExecuteStrategy.ts`, `noneExecuteStrategy.ts`, `richExecuteStrategy.ts`, `sendToTerminalTool.ts`). The helper accepts both `\n` and `\r` (previously inconsistent — the execute strategies only checked `\n`).

**Should-fix (partial)**
- Extracted the shared helper so the three execute-strategy edits are now one-liners using it — no more near-duplicates.
- On shell-capability gating for bracketed paste mode: I held off on adding a shell-capability check in this PR because the platform-specific fallback already lives in the terminal layer (`ITerminalInstance.sendText`), and adding a separate capability probe here would duplicate that logic and expand the PR's scope. Happy to split that into a follow-up if you'd prefer — let me know.

**Nit**
- Added tests for the foreground-terminal multiline path (new `test('foreground terminal path preserves newlines for heredoc commands')` and `test('foreground terminal path normalizes single-line commands')`).
- Added an inline comment on both foreground and background multiline branches explaining that `normalizeCommandForExecution` is intentionally skipped (which also skips its `.trim()` side-effect), so future readers don't try to "clean up" the input.

Local checks: `tsc -p src/tsconfig.json --noEmit` is clean on the touched files. Let me know if you want the shell-capability gate rolled into this PR or left as a follow-up.

meganrogge:
@anthonykim1 can you please test on windows and see before/after this change how it behaves?

anthonykim1:
image
@maruthang Having trouble to get "hang" to repro. Tried to force agent to use send_to_terminal and heredoc but it seems smart enough pre-pr to send EOF on its own. 

Do you have steps to test the behavior pre-pr?

maruthang:
Thanks for looking at this @anthonykim1! The reason you're seeing the agent sidestep it is exactly what @brendaningram's issue describes — the agent learned to avoid heredocs because they silently fail, and just uses a different strategy. The "hang" only surfaces when the agent *is* coaxed into using heredoc through `send_to_terminal` (or, more reliably, into `run_in_terminal`, which goes through the execute strategies).

Here's a repro that forces it deterministically on Windows/Linux/macOS:

**Setup**
1. Check out `main` at a commit prior to this PR (e.g. `git checkout 11d49abcbf1^`) and run VS Code from source, or install the current stable build.
2. Open the chat panel in **Agent** mode with `run_in_terminal` (or `send_to_terminal`) enabled.

**Prompt that forces the bad path**
Use this exact instruction so the agent can't work around it:

```
Create /tmp/heredoc-test.txt (or C:\Temp\heredoc-test.txt on Windows) with the exact
contents "line one\nline two\nline three". Use run_in_terminal with a single
heredoc command of the form:

  cat > <path> << 'EOF'
  line one
  line two
  line three
  EOF

Do not substitute echo, printf, python, Set-Content, Out-File, or editor tools.
Call the tool exactly once with the full multi-line command above.
```

**Expected vs observed (pre-PR)**

- Pre-PR: `normalizeCommandForExecution` collapses the newlines before `sendText`, so the shell receives `cat > <path> << 'EOF' line one line two line three EOF` as a single line. On bash/zsh the heredoc is never closed (EOF is now on the content line), the terminal sits waiting for a `EOF<newline>` that never comes, and the command never completes. On PowerShell you get a parse error immediately.
- Post-PR: the multi-line branch sends it verbatim with `forceBracketedPasteMode=true`, and the heredoc completes normally.

**Windows-specific notes**

- On PowerShell, the pre-PR path doesn't hang the same way — PowerShell errors out on the collapsed single-line form. That's still a regression vs. the multi-line intent, but it fails loudly instead of hanging.
- On WSL / Git Bash, behavior matches Linux — silent hang waiting for EOF.
- Windows also exercises `\r\n` in addition to `\n` — that's why `isMultilineCommand` now checks both (previously one path checked only `\n`, the other both).

**Quicker manual repro (bypasses the agent)**

If you'd rather not coax the model, you can unit-test the exact pre-PR behavior by calling `normalizeCommandForExecution` directly on the heredoc string — you'll see the newlines collapse. The new `isMultilineCommand` check in this PR is what prevents that for multi-line inputs.

Let me know if you still can't repro and I'll send a short screencap.

meganrogge (approved):
Thank you! Added a bit to this to fix benchmark issues

DonJayamanne (approved):

PR #307978, opened 2026-04-06
fix: preserve code block toolbar visibility during chat streaming

### LINKED ISSUE #255290: Copy button for code snippets only works after AI finishes streaming response

maruthang:
Fixes #255290

* Read our Pull Request guidelines:
  https://github.com/microsoft/vscode/wiki/How-to-Contribute#pull-requests

## Summary

**Bug:** The copy button on code snippets in chat responses was unclickable while the AI was still streaming, because the toolbar would flicker and lose interactivity.

**Root Cause:** During streaming, code block DOM elements are briefly detached and reattached as markdown is re-rendered. This causes the browser to lose CSS `:hover` state, which in turn hides the toolbar (due to a `pointer-events: none` rule on the non-hovered state). The toolbar becomes invisible and unclickable mid-stream.

**Fix:** Added JavaScript-based hover tracking (`mouseenter`/`mouseleave` listeners) to `CodeBlockPart` that toggles a `force-visibility` class on the toolbar element. This class persists across DOM detach/reattach cycles, keeping the toolbar visible and clickable even while streaming. The hover state is also restored in the render path so that re-renders during streaming preserve toolbar visibility.

## Changes

- `src/vs/workbench/contrib/chat/browser/widget/chatContentParts/codeBlockPart.ts`: Added `_isHovered` boolean and `mouseenter`/`mouseleave` event listeners that toggle `force-visibility` on the toolbar. Added post-render restore of toolbar visibility when hovered. Integrated with existing dropdown visibility tracking.
- `src/vs/workbench/contrib/chat/test/browser/widget/chatContentParts/chatMarkdownContentPart.test.ts`: Added three regression tests verifying that code block content is correctly set during streaming, that re-renders update content properly, and that the code block pool produces correct data across multiple renders.

## Testing

- Added regression tests in `chatMarkdownContentPart.test.ts` that verify code block text is correctly rendered during streaming and across re-renders
- All existing tests pass

### DIFF
--- a/src/vs/workbench/contrib/chat/browser/widget/chatContentParts/codeBlockPart.ts
+++ b/src/vs/workbench/contrib/chat/browser/widget/chatContentParts/codeBlockPart.ts
@@ -144,6 +144,8 @@ export class CodeBlockPart extends Disposable {
 	private currentCodeBlockData: ICodeBlockData | undefined;
 	private currentScrollWidth = 0;
 	private lastLayoutWidth: number | undefined;
+	private _isHovered = false;
+	private _toolbarElement!: HTMLElement;
 
 	private isDisposed = false;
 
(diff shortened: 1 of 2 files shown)


roblourens (approved):
Sure, thanks

vijayupadya (approved):

PR #307994, opened 2026-04-06
fix: show keyword suggestions alongside search result count

### LINKED ISSUE #261789: keyword suggestion message overrides search results count
I noticed that when I perform a search and there are suggestions, the "Search instead for: " message replaces the search results count "X results in Y files".

One way to address this might be to add a way to dismiss that particular keyword suggestion, but ideally both messages should be visible simultaneously.

- Copilot Chat Extension Version: 0.30.1
- VS Code Version: 1.103.0
- OS Version: Ubuntu 24.04
- Feature (e.g. agent/edit/ask mode): N/A
- Selected model (e.g. GPT 4.1, Claude 3.7 Sonnet): N/A
- Logs:

Steps to Reproduce:

1. Perform a search that leads the Copilot Chat extension to show a suggestion.
   This does not seem to be deterministic. Initially it appeared when I searched for `right[^-]click` (with regex activated), but in later tests that search no longer showed the suggestion; instead, I managed to trigger the suggestion by tweaking the search to `right[^.-]click`
2. Observe the suggestion replace the results count

Screenshots:

Image

Image

yogeshwaran-c:
## Summary
- Fixes the bug where keyword suggestions replace the search result count message in the Search view
- Both the result count (e.g., "X results in Y files") and keyword suggestions ("Search instead for: ...") are now displayed simultaneously
- Keyword suggestions are appended after the result count using a dedicated container element

## Details

When `search.searchView.keywordSuggestions` is enabled, the keyword suggestion message ("Search instead for: ...") was completely replacing the search result count message ("X results in Y files"). This happened because:

1. `updateKeywordSuggestionUI()` called `clearMessage()` to wipe the result count before rendering keywords
2. `updateSearchResultCount()` had an early return when keywords existed, preventing the result count from ever being shown alongside keywords

### Changes:
- **Removed the early return** in `updateSearchResultCount()` that skipped rendering when cached keywords existed
- **Added `renderKeywordSuggestions()`** helper method that renders keyword suggestions into a dedicated `span.ai-keyword-suggestions` container
- **Updated `updateKeywordSuggestionUI()`** to append keywords to the existing message element rather than clearing and replacing it
- **Updated `updateSearchResultCount()`** to re-render cached keywords after the result count when they exist
- **Updated CSS** to style the new `.ai-keyword-suggestions` container instead of the old `.ai-keywords` class on the message element

Fixes #261789

## Test plan
- [ ] Enable `search.searchView.keywordSuggestions` setting
- [ ] Perform a search that triggers keyword suggestions
- [ ] Verify both the result count ("X results in Y files") and keyword suggestions ("Search instead for: ...") are visible simultaneously
- [ ] Verify keyword suggestion links are still clickable and trigger a new search
- [ ] Verify that when no keyword suggestions are available, the result count displays normally
- [ ] Verify that when results are cleared, both messages are cleared

### DIFF
--- a/src/vs/workbench/contrib/search/browser/media/searchview.css
+++ b/src/vs/workbench/contrib/search/browser/media/searchview.css
@@ -170,16 +170,10 @@
 	overflow-wrap: break-word;
 }
 
-.search-view .message.ai-keywords {
-	display: -webkit-box;
-	-webkit-box-orient: vertical;
-	line-clamp: 2;
-	-webkit-line-clamp: 2;
+.search-view .message .ai-keyword-suggestions {
 	overflow: hidden;
 	text-overflow: ellipsis;
-	white-space: normal;
-	margin: 0 22px 8px;
-	padding: 0px;
+	white-space: nowrap;
 }
 
 .search-view .message p:first-child {
--- a/src/vs/workbench/contrib/search/browser/searchView.ts
+++ b/src/vs/workbench/contrib/search/browser/searchView.ts
@@ -2010,10 +2010,26 @@ export class SearchView extends ViewPane {
 		this.inputPatternIncludes.setOnlySearchInOpenEditors(false);
 	}
 
+	private renderKeywordSuggestions(messageEl: HTMLElement) {
+		const keywordContainer = dom.append(messageEl, $('span.ai-keyword-suggestions'));
+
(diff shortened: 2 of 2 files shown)


osortega (changes requested):
We shouldn't modify the CSS, we should make it a different element instead

Issue #232159, opened 2024-10-24
Devcontainer setup ad-hoc parsing fails on Rocky 8.5


gunnihinn:
Does this issue occur when all extensions are disabled?: Yes

- VS Code Version: 1.92.0 (Commit: b1c0a14de1414fcdaa400695b4db1c0799bc3124)
- OS Version: Rocky Linux 8.5

The `check-requirements-linux.sh` script that runs as part of a devcontainer setup wants to parse the OS version from `/etc/os-release`.
That file "is a newline-separated list of environment-like shell-compatible variable assignments" (see `man os-relase` on Linux).
The script tries to parse the value it wants [here](https://github.com/microsoft/vscode/blob/main/resources/server/bin/helpers/check-requirements-linux.sh#L34), by running:

```
OS_ID="$(cat /etc/os-release | grep -Eo 'ID=([^"]+)' | sed -n '1s/ID=//p')"
```

This looks for a key-value pair of the form `ID=value` and strips out the `ID=` part.
It assumes the pair is not of the form `ID="value"`, which appears to be permitted by the file format, and appears in the wild in Rocky 8.5 where the relevant line reads:

```
ID="rocky"
```

In that case the `grep` command will fail to match the line and the value of `OS_ID` will be empty, resulting in a non-zero exit code of 1.

This could be fixed by changing the variable assignment to:

```
OS_ID=$(source /etc/os-release && echo $ID)
```

This will run a subshell in which the values in the file get assigned to environment variables, which removes any surrounding `"` marks.
Since this runs in a subshell the current shell's environment will not pick up these variables.

Steps to Reproduce:

1. Start a devcontainer built from Rocky 8.5.

gunnihinn:
In the issue text I link to the problematic line in `HEAD` on `master`, so the issue remains there.

deepak1556:
Thanks for the report, feel free to send a PR.

amangupta982:
I can reproduce this on Rocky Linux 8.5. The suggested fix of sourcing /etc/os-release in a subshell works correctly on my end.

q404365631:
I'd like to work on this issue. I'll create documentation for Devcontainer setup on Rocky Linux 8.5 and provide a fix for the ad-hoc parsing failure.

Rajkaran-122:
@gunnihinn please  @deepak1556 review this pr #331957 .

LeonxLJX:
[...]

PR #308001, opened 2026-04-06
fix(server): handle quoted ID values in /etc/os-release

tysoncung:
## Problem

The devcontainer setup script `check-requirements-linux.sh` fails to parse `/etc/os-release` on Rocky Linux 8.5 because it uses ad-hoc grep/sed parsing that doesn't handle quoted values.

On Rocky 8.5, the file contains:
```
ID="rocky"
```

The current regex pattern `ID=([^"]+)` fails to match this format, causing `OS_ID` to be empty and the script to exit with code 1.

## Solution

Replace the ad-hoc parsing with proper shell variable sourcing:
```bash
OS_ID=$(source /etc/os-release && echo $ID)
```

This approach:
- ✅ Handles both quoted and unquoted values correctly
- ✅ Runs in a subshell to prevent environment pollution
- ✅ Follows the official `/etc/os-release` spec (shell-style variable assignments)
- ✅ More robust and maintainable

## Files Changed
- `resources/server/bin/helpers/check-requirements-linux.sh`

Closes #232159

### DIFF
--- a/resources/server/bin/helpers/check-requirements-linux.sh
+++ b/resources/server/bin/helpers/check-requirements-linux.sh
@@ -30,7 +30,7 @@ MIN_GLIBCXX_VERSION="3.4.25"
 
 # Extract the ID value from /etc/os-release
 if [ -f /etc/os-release ]; then
-    OS_ID="$(cat /etc/os-release | grep -Eo 'ID=([^"]+)' | sed -n '1s/ID=//p')"
+    OS_ID=$(source /etc/os-release && echo $ID)
     if [ "$OS_ID" = "nixos" ]; then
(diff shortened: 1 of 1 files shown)


tysoncung:
Bump for review - simple fix for handling quoted values in /etc/os-release parsing. All checks passing.

tysoncung:
Bumping this PR. It fixes handling of quoted ID values in /etc/os-release. The PR has been open for 52 days with one pending CI check. Happy to rebase and address any feedback.

dmitrivMS:
@tysoncung My understand is that `source` is not a POSIX-sh command - this may cause it to fail in some shells.

PR #308021, opened 2026-04-06
feat: register agents to launchservices to allow launch from vscode

deepak1556:
For users who haven't opened the app before or after the rename, this forces a registration to launche services db to avoid errors like

Screenshot 2026-04-06 at 00 02 39

### DIFF
--- a/src/vs/code/electron-main/app.ts
+++ b/src/vs/code/electron-main/app.ts
@@ -1736,7 +1736,7 @@ export class CodeApplication extends Disposable {
 	}
 
 	private registerEmbeddedAppWithLaunchServices(): void {
-		if (!isMacintosh || (process as INodeProcess).isEmbeddedApp || !this.productService.embedded?.nameShort) {
+		if (!isMacintosh || (process as INodeProcess).isEmbeddedApp || !this.productService.embedded?.nameShort || this.productService.quality === 'stable') {
 			return;
 		}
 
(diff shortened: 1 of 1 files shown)


src/vs/code/electron-main/app.ts
```diff
@@ -1729,5 +1730,24 @@ export class CodeApplication extends Disposable {
 		// Validate Device ID is up to date (delay this as it has shown significant perf impact)
 		// Refs: https://github.com/microsoft/vscode/issues/234064
 		validateDevDeviceId(this.stateService, this.logService);
+
+		// macOS: eagerly register the embedded app with Launch Services
+		this.registerEmbeddedAppWithLaunchServices();
+	}
```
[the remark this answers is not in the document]
deepak1556: This is an acceptable overhead, a check to see if app is registered would also be via lsregister.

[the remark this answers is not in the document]
deepak1556: And performing the registration just before launch will also be late for the app to pick up the changes.

Yoyokrazy (approved):

src/vs/code/electron-main/app.ts
```diff
@@ -1729,5 +1730,25 @@ export class CodeApplication extends Disposable {
...
+
+	private registerEmbeddedAppWithLaunchServices(): void {
+		if (!isMacintosh || (process as INodeProcess).isEmbeddedApp || !this.productService.embedded?.nameShort || this.productService.quality === 'stable') {
```
bpasero, line 1739: @deepak1556 we now seem to run this on every launch. Why not remember that we did call this and then refrain from calling it? Or alternatively, can we somehow do this without blocking the process?

deepak1556: > Why not remember that we did call this and then refrain from calling it?

Would that be writing to storage key ?

> can we somehow do this without blocking the process?

Are you concerned about the time to launch the process ?

bpasero: Yeah we have a "state service" on the main side which is just backed by a JSON file, no need for SQLite.

Spawning a process blocks sync as far as I know.

PR #308065, opened 2026-04-06
Fix typos

Ladsgroup:
This is done via a script I wrote to report typos and after I checked them multiple times manually, I applied the correct ones.

This includes many parts, from documentation to variable names to translatable strings.

This is all done manually so if there is a mistake, my apologies.

### DIFF
--- a/src/vs/base/common/strings.ts
+++ b/src/vs/base/common/strings.ts
@@ -861,7 +861,7 @@ export function* forAnsiStringParts(str: string) {
 
 /**
  * Strips ANSI escape sequences from a string.
- * @param str The dastringa stringo strip the ANSI escape sequences from.
+ * @param str The string to strip the ANSI escape sequences from.
  *
  * @example
  * removeAnsiEscapeCodes('\u001b[31mHello, World!\u001b[0m');
--- a/src/vs/base/node/pfs.ts
+++ b/src/vs/base/node/pfs.ts
@@ -98,7 +98,7 @@ export interface IDirent {
 
 /**
  * Drop-in replacement of `fs.readdir` with support
- * for converting from macOS NFD unicon form to NFC
+ * for converting from macOS NFD unicode form to NFC
  * (https://github.com/nodejs/node/issues/2165)
  */
 async function readdir(path: string): Promise<string[]>;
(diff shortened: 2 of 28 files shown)


src/vs/base/common/strings.ts
```diff
@@ -861,7 +861,7 @@ export function* forAnsiStringParts(str: string) {
 
 /**
  * Strips ANSI escape sequences from a string.
- * @param str The dastringa stringo strip the ANSI escape sequences from.
```
Ladsgroup, line 864: I wasn't even sure what this documentation was trying to convey but based on the documentation of the function below (removeAnsiEscapeCodesFromPrompt) my change seems correct.

src/vs/base/node/pfs.ts
```diff
@@ -98,7 +98,7 @@ export interface IDirent {
 
 /**
  * Drop-in replacement of `fs.readdir` with support
- * for converting from macOS NFD unicon form to NFC
```
Ladsgroup, line 101: Based on the content of the linked PR, I‌ think it meant unicode but not 100% sure honestly.

src/vs/editor/common/config/editorOptions.ts
```diff
@@ -2236,7 +2236,7 @@ class EditorGoToLocation extends BaseEditorOption<EditorOption.gotoLocation, IGo
 					...jsonSubset,
 				},
 				'editor.gotoLocation.multipleImplementations': {
-					description: nls.localize('editor.editor.gotoLocation.multipleImplemenattions', "Controls the behavior the 'Go to Implementations'-command when multiple target locations exist."),
```
Ladsgroup, line 2239: These and all below are names of the variables that have been misspelled.

gjsjohnmurray:
Does this PR conform to the guidance at https://github.com/microsoft/vscode/wiki/How-to-Contribute#spell-check-errors?

Ladsgroup:
> Does this PR conform to the guidance at https://github.com/microsoft/vscode/wiki/How-to-Contribute#spell-check-errors?

I think so, I have fixed a couple of typos in translatable strings plus fixed gibberish line in one case. It is also somewhat a large enough clean up to avoid having annoying one-off PRs. But I leave the decision to the maintainers of the repo to decide whether it's in scope or not.

Ladsgroup:
Gentle ping.

mjbvz:
We generally don't take spelling fix PRs, especially when the touch such a wide range of files

PR #308081, opened 2026-04-06
enable `timeout` for async/bg terminal commands

### LINKED ISSUE #308037: Async terminals that produce output are not really async

Seems like we wait for the task to be 'idle', so they often just block for a long time, which will defeat the purpose in some cases

[dc31b0ab.copilotmd](https://github.com/user-attachments/files/26513978/dc31b0ab.copilotmd)

My first command blocked for 20s and was never async. The second only continued once the command stopped printing output

Maybe it's surprising that the output contains the prompt? I thought shell integratinon would let us pull that out

```
 The command became idle with output:
➜  debugtest git:(main) ✗  for i in $(seq 1 20); do sleep 1 && echo "hello world $i"; done
[...]
```

meganrogge:
fixes #308037


benvillalobos (approved):

anthonykim1:
So these would help with cases like dev server when the server could basically run "forever" and have somewhat of continuous output, right?

PR #308122, opened 2026-04-07
perf: improve thinking layout thrashing

justschen:
- during streaming thinking would do a bunch of `scrollHeight` checks - 1 dom read per update instead of multiple. 
 - fix: markdown chunks arrive back-to-back when the shimmer appears/disappears between them the response height will jump


roblourens (dismissed):

justschen:
* fixing merge conflicts 😢

roblourens (approved):

PR #308165, opened 2026-04-07
fix: prevent indentation shift for tree items without parent icons

maruthang:
Fixes #307350

## Summary

- When a collapsible parent item lacks an icon, `alignIconWithTwisty()` incorrectly triggered alignment for leaf children, hiding their twistie space (width: 0) while collapsible siblings kept theirs. This caused inconsistent indentation at the same depth level.
- Changed the "parent lacks icon" case in `alignIconWithTwisty()` to return `false`, so alignment only applies when the parent has an icon and can establish a consistent visual hierarchy.
- The "parent has icon" case (line 1653) remains unchanged and continues to work correctly.

## Test plan

- [ ] Create a TreeView extension with mixed items:
  - Group items (collapsible) WITHOUT icons
  - Leaf items WITH icons
  - At the same depth level
- [ ] Verify that all items at the same depth have consistent indentation
- [ ] Verify group items WITH icons still produce correct alignment for their leaf children
- [ ] Verify toggling between grouped/ungrouped doesn't cause indentation shifts


n-gist:
I did some testing of this change. While it fixes one case, it also introduces other inconsistencies. This post can be skipped, next one is more meaningful.<br><br>

1. New issue — shifting of leaf items

All 5 items here are siblings, without parent. Notice that `sw*` and `diffpos` are shifted.

Screenshot_1<br><br>

2. The fixed case

The presence of a parent icon no longer affects the alignment of its children (in this layout).

good<br><br>

3. Root-level effect

If the items are grouped under a parent, the shifting from case 1 disappears. However, alignment now depends on whether the parent has an icon: there is no shift when the parent has an icon, and items shift when it does not.

rooteffect<br><br>

4. Sibling effect

When all collapsible sibling items have icons, non-collapsible items become shifted (though I'm not sure that it is related to collapseability, or just to the fact when all sibling items have icons).

This one is the most confusing, so here is a description of the screencast:

* I periodically group and ungroup `diffpos` to test alignment stability.
* When `.vscode` and `storageWinner` do not have icons, alignment is stable.
* When only one of them has an icon, alignment is stable.
* When both of them have icons, grouping `diffpos` (which only triggers rerendering of the `root` item here) causes `sw*` to shift.
* Removing the icons from `.vscode` and `storageWinner` and triggering rerendering of `root` restores the original `sw*` alignment.

siblingseffect<br><br>


5. Leaf icon effect

In this test, I assign the `folder` icon to leaf items, which, I believe, behaves the same as `undefined` in the default theme.
For incorrectly shifted items, this affects item alignment.
For correctly aligned items, this has no effect.

leafeffect

n-gist:
I think I have narrowed the problem down a little. The following tests were performed against the current version (1.121.0), not this change.

The issue appears to be caused by the following condition: 

**The presence of a sibling collapsible item without an icon or a parent item without an icon, shifts leaf items with icons to the left.**

Or from the opposite angle if it is more correct:

**Leaf items with icons are not shifted to the right when a parent item without an icon exists, or when a sibling collapsible item without an icon exists.**
<br>

1\. Notice how the presence of icons affects the alignment of `sw*` (which are leaf items):

a) When the parent item's icon is set to `undefined` (`off` results in this), `sw*` are shifted to the left. This should not happen.

b) When `leafItem` without an icon is added to the group, the alignment of `sw*` does not change. This is correct.

c) When `.vscode` (a collapsible item without an icon) is added to the group, the leaf items are shifted to the left, the same as in case (a).

d) When `.vscode` has an icon and added to the group, the leaf items are not shifted, which is correct.

shiftsbynoicon<br><br>


2\. This test just confirms the same issue under different circumstances:

Here I'm not checking the positions of `sw*`. Instead, I use grouping/ungrouping of `diffpos.txt` only to trigger a refresh of the root element (`storageWatch`).

When an icon is set for `leafItem`, it should be shifted to the right, but it is not. As a result, the icons become clearly misaligned when `.vscode` also has an icon.

However, the underlying cause appears to be the same: the presence of the collapsible sibling `storageWinner` without an icon makes `leafItem` shifted to the left. When an icon is assigned to `storageWinner` and a refresh is triggered, `leafItem` is shifted correctly and all items become properly aligned.

complexTest<br><br>

3\. Another test to confirm that this affects leaf items with icons only (`sw*`), and `leafItem` without an icon is not affected:

onlyleafs

maruthang:
Thanks a lot for the thorough testing and the narrowing in your follow-up — you were right that my first attempt traded one inconsistency for another. I've reworked the fix around a single level-local rule that mirrors how the built-in trees apply `align-icons-and-twisties` (icon takes the twistie's space only when the theme has file icons but no folder icons): a leaf's icon now aligns with the twistie iff no collapsible sibling at its level renders an icon or checkbox — the parent's icon no longer plays any role.

This fixes the root-level leaf shift from your first comment, and your second test case (an iconless collapsible sibling can no longer pull a leaf's icon out of the icon column established by an iconed collapsible sibling). The one remaining movement is when the *only* collapsible sibling at a level gains/loses its icon, which changes that level's layout by design but now always stays internally aligned.

Would you mind re-running your test cases against the updated branch?

n-gist:
@maruthang, this version of the fix is better than both the current state and the previous fix version, but there is still one issue remaining.

The screencast shows that, in general, the folder/file structure matches the Explorer view perfectly (the portion up to `test_1.txt`).

However, once a collapsible item (`TESTGROUP`) is assigned an icon, its siblings WITH icons become shifted (`test_1.txt` and `leaf_With_Icon`). I believe this is the only remaining issue I was able to find.

Siblings without icons (`leaf_Without_Icon`) remain stable, and collapsible siblings remain stable as well. Children also remain stable. I tried various configurations under `TESTGROUP` and couldn't find any indentation issues caused by its icon.

treeViewIndentation
<br><br>

If it's convenient, I prepared a small testing kit:
[treeView-indentation-test.zip](https://github.com/user-attachments/files/29744550/treeView-indentation-test.zip)

Install the VSIX and open the project from zip. You should see the same layout as in the screencast.

Right-click TESTGROUP → Customize → Icon Select. The `off` value is an alias that results in `iconPath = undefined`.

For this test, I adjusted the extension code so that it refreshes the parent item when the icon changes, removing the need to trigger a refresh manually. In the normal scenario, refreshing the parent should be unnecessary, since an item's icon should not affect its siblings.

Also, to avoid any `resourceUri` effects from VS Code itself, groups can be used for raw testing. Using the `leaf` prefix in a group label sets its `collapsibleState` to `None`, making it behave as a leaf item for testing purposes.

That said, I'm happy to test anything else if needed. Feel free to reach out with any intermediate or final fix versions and I'll be happy to test them.

maruthang:
Thanks for another careful round of testing — your TESTGROUP case exposed that I was keying alignment on whether *any* collapsible sibling renders an icon, so the first icon at a level flipped everything. The rule is now: a leaf's icon takes the twistie space as long as at least one collapsible sibling still renders a bare twistie (or none exist); only when *every* collapsible sibling pairs its twistie with an icon does the leaf keep the space, so icons form one column.

In your test that means assigning an icon to TESTGROUP no longer moves `test_1.txt` or `leaf_With_Icon`, since `.vscode` and `testfolder` still have bare twisties — while the Explorer-matching layouts you confirmed are unchanged.

One honest caveat: if the *last* bare collapsible at a level gains an icon, the level does re-layout — that transition is inherent, as the two end states you validated (labels-aligned vs icons-aligned) genuinely differ, much like switching between file icon themes with and without folder icons.

Would you mind re-running your scenarios against the updated branch? The test kit made this round much easier — thank you for that.

n-gist:
@maruthang, sibling leaf items now remain stable when some collapsible items have icons, but when all collapsible siblings have icons, leaf items with icons become shifted (test_1.txt and leaf_With_Icon). Leaf items without icons remain stable (leaf_Without_Icon).

treeViewIndentation2

This can probably be fixed, but I'm starting to question whether this direction is the right one.

The layout no longer matches Explorer, because Explorer actually shifts everything—both siblings and children—when folder icons are present.

n-gist:
I think we are running into more of a logical problem than an implementation problem. It may be worth stepping back and reconsidering why we ended up here and whether there is a higher-level solution, rather than continuing to address individual edge cases. Here is some analysis from my side.

When we try to preserve Explorer-like behavior, custom icons introduce issues. When we fix the custom icon issues, we move away from Explorer behavior.

Personally, I would expect the presence of icons to have no effect on the indentation of siblings or children. That would make the layout fully consistent, although it would differ from Explorer (depending on whether the active file icon theme provides folder icons).

For example, if we try to mirror Explorer in a custom view where collapsible items may or may not have icons, then:
* Children of a collapsible item with an icon should be shifted.
* Children of a collapsible item without an icon should not be shifted.

When both collapsible items are expanded, their children would no longer align with each other. In other words, the children of group1 and group2 would appear at different indentation levels despite having the same nesting depth. This weakens the relationship between visual indentation and the actual tree structure within a single view. Depending on the extension, that may be unwanted.

Explorer can rely on the fact that a file icon theme either provides folder icons for all folders or for none of them. Because of that, Explorer can make a simple global decision: either shift everything or shift nothing.

Custom views cannot rely on that assumption, because collapsible items may be mixed—some with icons and some without. I believe this difference is what creates the problem.

To be precise, what I found is that Explorer does not analyze siblings at all. It determines indentation solely from the file icon theme:

```ts
const onDidChangeFileIconTheme = (theme: IFileIconTheme) => {
	container.classList.toggle('align-icons-and-twisties', theme.hasFileIcons && !theme.hasFolderIcons);
};
```

So for Explorer, this problem does not exist.

For my extension, I would actually prefer this behavior because it is Explorer-like. However, other extensions may prefer the current behavior, especially when their tree contents are not related to file-system hierarchies, so introducing a dependency on the active file icon theme and changing the default behavior would likely be incorrect.

After looking into this, I started wondering whether this could be configurable during tree view creation, or through a setting afterward, by introducing a new option:

```ts
export interface TreeViewOptions<T> {
	treeDataProvider: TreeDataProvider<T>;
	showCollapseAll?: boolean;
	canSelectMany?: boolean;
	dragAndDropController?: TreeDragAndDropController<T>;
	manageCheckboxStateManually?: boolean;
	indentation?: undefined | 'explorer' | boolean;
}
```

Where:
* `undefined` - current behavior (the sibling-analysis compromise implemented by this PR, potentially relaxed once this option exists, or reverted if the original behavior is considered intentional)
* `explorer` - infer indentation from the active file icon theme exactly as Explorer does
* `boolean` - force indentation on or off

I may be missing a better solution here, but currently it feels like we are in a "fix one case, introduce another" trap.

maruthang:
I think your analysis is exactly right — thanks for digging out the Explorer code. The root of the trap is that `align-icons-and-twisties` is a *global, container-level* decision derived from the theme (`hasFileIcons && !hasFolderIcons`), while custom views make icon decisions *per item*. Any sibling-analysis heuristic (mine included) is an attempt to reconstruct a global answer from local data, and that's why every fix shuffles the inconsistency somewhere else rather than eliminating it.

An explicit `TreeViewOptions.indentation` option (`undefined | 'explorer' | boolean`) resolves it cleanly: extensions whose trees mirror the file system opt into theme-driven behavior, everyone else gets a stable, icon-independent layout — and `undefined` preserves compatibility. Since that's new extension API, it needs the VS Code team's direction through the API proposal process, so here's what I'd suggest:

1. I'll file a feature request summarizing this thread (linking #307350 and your test kit) proposing the `indentation` option, so the team can weigh in on the API shape.
2. In the meantime, this PR can stand as the default-behavior improvement — per your testing it strictly reduces the current inconsistencies without touching the API surface — or be parked pending the team's call if they'd rather solve it only via the option.
3. If the team endorses the option, I'm happy to implement it behind a proposed API.

Does that split sound right to you? And thank you for the test kit and the persistence here — this conversation has turned a whack-a-mole fix into an actual design answer.

n-gist:
Yes, this split makes perfect sense to me.

I also agree that a new API option is the right direction. As we've discovered through these iterations, heuristics can't reliably infer the expected indentation. Even when they produce the "correct" result in one part of the tree, they may produce a different indentation model elsewhere within the same view. When that variability is not desired, delegating the decision to the extension seems like the only robust solution.

I'd be happy to stay involved during the API design and implementation process and test anything on my side. One small thought is that an enum might read a bit cleaner than mixing types and values, but that's just an API design detail for later discussion.

Here is the same testing kit, but including the extension source code. In case it is helpful while experimenting with the new option, to be launched directly from OSS. The tree views are created in the src/tree/controller.ts constructor, which is where the new option could be passed for testing.
[treeView-indentation-test-source.zip](https://github.com/user-attachments/files/29802798/treeView-indentation-test-source.zip)

Regarding this PR, I also agree. If we can be reasonably confident that indentation will eventually become configurable, then allowing some edge cases to remain in the default behavior is perfectly acceptable. The only part I'm uncertain about is what the default behavior should actually optimize for. Here are my three thoughts on it.

1. To me, the most confusing aspect of #307350 is siblings shifting to the left when a collapsible without an icon is added. That behavior still feels unintuitive and seems worth avoiding regardless (this seems fixed at the moment of this PR, except the leaf_Without_Icon issue ).

2. On the other hand, adding indentation to the children of a collapsible item with an icon could arguably be considered the expected default behavior, if we treat it as 'adaptive'. The current revision actually removes that behavior.

3. The question I still can't confidently answer is whether leaf siblings indentation should depend on the presence of icons on collapsible siblings at all. If the intention is to align the labels of leaf and collapsible items, or if there are other layout constraints that I'm not aware of, then the current sibling analysis-based approach makes sense. But if there are no such requirements, one of options is to check if indentation can be determined solely by whether the parent item has an icon.

maruthang:
Filed the API proposal: #325080 — it summarizes this thread, links #307350 and your test kit, and proposes `TreeViewOptions.indentation` (`'explorer' | boolean | undefined`). I noted your point that an enum may read cleaner; I left the exact shape open for the team's API-design guidance. Thanks for offering to stay involved — I've referenced that in the issue, and the source-included test kit will be genuinely useful when prototyping the option.

On what the default should optimize for, I agree with your ranking:

1. **Sibling left-shift when an icon-less collapsible appears is the worst offender** — it's the most counter-intuitive and the original #307350 complaint, so avoiding it should be the default's first priority. You noted a residual `leaf_Without_Icon` case still slips through here; I'll dig into that specifically and see if it can be closed without reintroducing the icon-toggle instability.

2. **Adaptive indentation for children of an iconed collapsible** — you're right that the current revision drops this. I think that's acceptable *as a default* precisely because it's the ambiguous case: whether it's "expected" depends entirely on whether the extension's tree is file-system-like, which is exactly the judgment the new `indentation` option is meant to delegate. So I'd lean toward keeping the default simple/stable and letting `'explorer'` restore the adaptive behavior.

3. **Whether leaf alignment should depend on collapsible siblings' icons at all** — this is the crux, and I don't think it can be answered universally, which is why I don't want to over-fit the default. Parent-icon-only (your alternative) is appealingly simple but reintroduces the child-shift-on-parent-icon behavior that #307350 was partly about. My inclination: keep the default minimal and correct for the worst case (#1), and treat #2/#3 as behaviors the extension opts into via the option.

So concretely: I'll (a) chase the residual `leaf_Without_Icon` case on this PR, and (b) pursue #325080 for the real fix. Does that sequencing work for you?

maruthang:
You're right that `leaf_Without_Icon` still shifts, and after tracing it I believe this specific case is inherent to the current alignment mechanism rather than something this PR can close. The `align-icon-with-twisty` class can only collapse a leaf's twistie into the icon's place — it has no way to reserve blank icon space, which is what an icon-less leaf would need to line its label up with icon-bearing siblings; and any rule that *did* reserve that space would have to read whether the siblings have icons, reintroducing exactly the toggle instability we just fixed.

So I'd propose we leave this one to the explicit per-view alignment option in #325080, where the intended column can be stated rather than inferred. Everything else from your last round stays as-is on this PR.

n-gist:
I created a test case matrix covering all possible combinations of collapsibles and leaf items having or not having icons, and grouped them into several categories. I found this much easier to reason about than trying to imagine or brute-force every possible case. `+` means the item has an icon, `-` means it does not.

1. Common cases.
I made an assumption about how extensions typically use icons. In the general case, an extension either assigns icons to all collapsible items or to none of them. The same applies to leaf items: either all leafs have icons, or none do. These cases form four common combinations.

2. Mixed collapsibles.
These are cases where some collapsible items have icons while others do not. They demonstrate how the current `filter` + `some` heuristic behaves.

3. Mixed leafs.
These combine the corresponding Common cases to show leaf items with and without icons together.

4. Mixed.
A full mixture of icon presence. Actually just a merge of all 2. or 3. cases.

Together, these form the test case matrix. I then took screenshots for comparison:
M1 (1.128) ↔ M2 (PR) — compare the changes introduced by this PR (when the parent has no icon).
M3 (1.128) ↔ M4 (PR) — compare the behavior when the parent has an icon. Since the fix removes the dependency on the parent's icon presence, this comparison is mostly for completeness, but it still provides some useful output.

tests_matrix

[treeView-indentation-case-matrix.zip](https://github.com/user-attachments/files/29898683/treeView-indentation-case-matrix.zip) (projects to open with test kit vsix)

My notes after reviewing the table:

* **M2/M4** are intended to verify how the current fix behaves. Since the dependency on the parent icon has been removed, **M2** and **M4** are identical.

* **Case 1.3** – the `leaf-` label is not aligned. This is exactly the issue we noticed earlier. I believe it cannot be solved without improving the DOM structure or CSS classes to allow more flexibility for the second indentation level. This issue existed before the fix, is identical across **M1–M4**, and is probably outside the scope of this PR. Extensions have an option to use `blank` codicon instead of `undefined` to get such leafs aligned.

* **Case 3.2** – the `leafs+-` misalignment is a direct consequence of **Case 1.3** and is likewise identical across **M1–M4**.

* **Case 1.2 ↔ Case 1.4** illustrates the difference that leads to #307350. The most confusing behavior was `leaf+` shifting to the left when a `collapsible-` sibling was introduced. With the current fix, since leaf nodes are not initially indented, they remain stable in this scenario. If leaf nodes had already been indented by the new "*all collapsibles are `+` → indent leafs*" rule and a `collapsible-` later appeared, they would still shift left. However, this is much less confusing than before, because the alignment with the `collapsible-` label makes the reason for the shift visually apparent.

* **Case 2.2** combines cases **1.2** and **1.4** and demonstrates how the parent item's icon previously affected the position of `leaf+` (**M1 2.2 ↔ M3 2.2**). The fix makes this case consistent (**M2 2.2 ↔ M4 2.2**).

* **Case 4.1** inherits its consistency from the improvement of **Case 2.2**.

n-gist:
I think the current fix is acceptable even after the closure of #325080. File-system subtrees now naturally resolve to the following layouts depending on the active file icon theme. I tested the basic file icon themes — `None`, `Minimal`, `Seti` — as well as the one I normally use, vscode-icons, and all of them match the Explorer view.

**Update: The test was inaccurate because it didn't account for the presence of subfolders.**

Themes resolve as:

* **None** theme → **M2 Case 1.1**
* **Folders-, Files+** → **M2 Case 1.2**
* **Folders+, Files+** → **M4 Case 1.4**
* **Folders+, Files-** → **M4 Case 1.3** — this case still has the alignment issue, but I'm not sure such file icon themes are practical in the first place.



There is one potential optimization, though. We could check `parent.resourceUri !== undefined`, and if it is, treat that subtree as a file-system hierarchy, skipping sibling analysis and inferring indentation directly from the active file icon theme.

Other than that, the PR looks complete to me. The only change I would make is to combine `filter` and `some` into a single iteration to avoid traversing the collection twice and allocating an intermediate array.

n-gist:
Still, there is one problematic aspect that came to my mind afterward. There was no such issue before the fix because indentation depended on the parent item's icon. Now, there is no such dependency. Here is a comparison for cases where the theme has no folder icons and where it does:

explorer_matching
<br><br>

The **None** and **Seti** themes match, but **Minimal** and other themes with folder icons result in the children of `withSubs` and `withoutSubs` not being aligned, since when there are no sibling folders, there is nothing to infer the indentation from.

Several expanded folders from parallel roots — or even from the same root — may turn into a total indentation mess, which is particularly bad for a file-system structures.

This could have been controlled by #325080, but since that is not an option, I think the optimization I mentioned in the previous comment could provide consistent behavior in this case. It would reinstate the dependency on the parent item's icon, but only for Explorer-like subtrees, while preserving the new parent-icon-independent behavior for other structures.

There is just one consideration that I think needs to be addressed: whether to rely solely on the parent item, or still examine its children to make sure this is actually a file-system structure. For example, an extension may group non-file-system items under a collapsible associated with a path and decorated using `resourceUri`.

So I see two options:

1. Assume that such a subtree will not suffer much from inferring indentation from the active file icon theme, since a file-system-like decorated parent already makes the structure resemble a file-system hierarchy.

2. Examine the children to be 100% sure that this is a file-system subtree:
```ts
if (parent.resourceUri !== undefined && parent.children.every(c => c.resourceUri !== undefined)) {
  return !this.themeService.getFileIconTheme().hasFolderIcons; // subscribe to theme changes and cache it for performance if makes sense? Cache could also be used in hasIcon method
  // there are also node.themeIcon checks in the hasIcon method. These may also be needed there; I didn't inspect that.
} else {
  // PR current siblings analysis
}
```

The only downside of the second option is that it is **O(n²)**, which may affect very dense file structures. However, the check itself is simple, so it may still be acceptable here. Then again, the scenario described by the first option is probably quite rare — or acceptable even when it does occur.

maruthang:
Thanks for the deep dive and the concrete spec — I've implemented your option 2. For subtrees where the parent and all of its children carry a `resourceUri` (genuine file-system structures), alignment now defers to the icon theme exactly like the Explorer does:

```ts
if (parent.resourceUri !== undefined && parent.children.every(c => c.resourceUri !== undefined)) {
    return !this.themeService.getFileIconTheme().hasFolderIcons;
}
// otherwise: the sibling analysis from the previous revision
```

So folder-icon themes like "Minimal" keep the children of sibling folders aligned across parallel roots, while non-file-system trees keep the parent-icon-independent sibling behavior (and the icon-toggle stability) from the last round. It needed no new plumbing: `Aligner` already has `IThemeService`, `getFileIconTheme()` is already used a few lines down in `hasIcon`, and `TreeRenderer` already re-renders on `onDidFileIconThemeChange` — so switching icon themes updates the alignment live, matching how the Explorer's themable container reacts.

I went with your option-2 (check every child's `resourceUri`) rather than parent-only, so a collapsible decorated with `resourceUri` but grouping non-file items doesn't get pulled into theme-driven alignment; the `.every()` is O(n) per level as you noted, and the theme is read once per call. On `node.themeIcon` — `hasIcon` already handles the `FolderThemeIcon` case, so I left that path as-is; shout if you think the resource check should also consider `themeIcon`.

One transparency note: my local VS Code build got wiped between rounds and rebuilding it triggers a full `npm install` that reliably hangs this machine, so rather than a full compile I type-checked the change at the symbol level — every reference in the three new lines is declared and already used identically elsewhere in the same method (`getFileIconTheme()` at the `hasIcon` call just below, `resourceUri` on `ITreeItem`), so it's type-correct; I'm leaning on the PR CI for the full build. Could you re-test against "Minimal"/folder-icon themes and confirm the parallel-root file-system case is clean now?

n-gist:
Yes, I can confirm that file-structure subtree indentation now matches the Explorer view for all four themes where the tests previously revealed inconsistencies.

I also checked the non-Explorer-like structure for the test cases matrix again, and they remain the same as in the screenshot above (M2, M4).

What I would change is optimizing the sibling check into a single, allocation-free iteration (this is actually the version I ran the tests with):

```ts
			const children = parent.children;
			if (!children) {
				return false;
			}
			// ...
			if (parent.resourceUri !== undefined && children.every(c => c.resourceUri !== undefined)) {
				return !this.themeService.getFileIconTheme().hasFolderIcons;
			}
			// ...
			let hasCollapsibleSibling = false;
			for (let i = 0, len = children.length; i < len; i++) {
				const child = children[i];
				if (child.collapsibleState === TreeItemCollapsibleState.None) {
					continue;
				}
				hasCollapsibleSibling = true;
				if (!this.hasIconOrCheckbox(child)) {
					return true;
				}
			}
			return !hasCollapsibleSibling;
```

I think this optimization is worthwhile since the method is called per tree item, and item count can vary significantly.

Regarding `themeIcon`, I took a look at it, and I believe that under normal circumstances it isn't possible to encounter tree items with `themeIcon` set while `resourceUri` is not set, since `themeIcon` is probably assigned only in that case. Because of that, `resourceUri` seems to be the primary source of truth.

The `themeIcon` check in `hasIcon` looks more like a secondary path or fallback, while the `resourceUri` path is actually more performant because it relies on comparing an enum value (`collapsibleState`) rather than a string (`themeIcon.id`).

Those lines have been unchanged since 2020, so they may represent legacy behavior or simply extra safety that is no longer necessary. I may be wrong, but relying solely on `resourceUri` seems sufficient to me for reliably identifying a file-system subtree.

maruthang:
Thanks for confirming across all four themes — really glad the file-system case is clean now, and thanks for the thorough matrix re-check.

Applied your single-pass optimization in `1a601088022` — the sibling scan is now one allocation-free loop (no intermediate `filter` array), which matters here since `alignIconWithTwisty` runs per tree item. It's the exact version you tested with.

And agreed on `themeIcon`: since it's effectively only set alongside `resourceUri`, the `resourceUri` check is the reliable file-system-subtree signal (and comparing `collapsibleState` enums beats the `themeIcon.id` string comparison), so I've left the fast path keyed on `resourceUri` alone. Appreciate all the testing and the design push on this one — it ended up much better for it.

n-gist:
Re-checked, just to be sure. Both the test cases and the Explorer-like subtrees remain consistent.
This feels complete. Thanks for your responsiveness, and sorry for the walls of text, but this was interesting to inspect carefully.

maruthang:
Thank you — and no apology needed at all; the careful inspection is exactly what made the fix converge on something principled instead of a whack-a-mole patch. Really appreciate the testing, the test kit, and pushing on the design. 🙏

PR #308273, opened 2026-04-07
Add routing_method experiment variable for A/B/C hydra experiment

bhavyaus:
Port of microsoft/vscode-copilot-chat#4947.

## Summary

Add `copilotchat.autoModeRoutingMethod` ECS flight variable to control which routing algorithm the auto-intent-service uses per-request. This enables A/B/C experiments for the HYDRA model router.

## A/B/C Experiment Design

| Path | Value | Behavior |
|------|-------|----------|
| A | '' (empty, default) | Production automod — no router call |
| B | 'binary' | Binary classifier v1 (current router) |
| C | 'hydra' | HYDRA multi-head capability matching (new) |

Fallback for B/C → automod (path A), signaled via `fallback=true` in the response.

## Changes

- **configurationService.ts**: New `AutoModeRoutingMethod` experiment-based setting
- **routerDecisionFetcher.ts**: Add `routingMethod` param, send `routing_method` in request body, extend `RouterDecisionResponse` with new fields
- **automodeService.ts**: Read experiment var, handle `fallback=true` responses

### DIFF
--- a/extensions/copilot/src/platform/endpoint/node/routerDecisionFetcher.ts
+++ b/extensions/copilot/src/platform/endpoint/node/routerDecisionFetcher.ts
@@ -13,7 +13,7 @@ import { ITelemetryService } from '../../telemetry/common/telemetry';
 import { ICAPIClientService } from '../common/capiClient';
 
 export interface RouterDecisionResponse {
-	predicted_label: 'needs_reasoning' | 'no_reasoning';
+	predicted_label: 'needs_reasoning' | 'no_reasoning' | 'fallback';
 	confidence: number;
 	latency_ms: number;
 	candidate_models: string[];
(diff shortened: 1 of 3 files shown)


extensions/copilot/src/platform/endpoint/node/routerDecisionFetcher.ts
```diff
@@ -22,6 +22,12 @@ export interface RouterDecisionResponse {
 		no_reasoning: number;
 	};
 	sticky_override?: boolean;
+	routing_method?: string;
```
bhavyaus, line 25: The automode.routerDecision telemetry event doesn't include routing_method, fallback, or fallback_reason. Do we need not need them tracked?

justschen (approved):

PR #308321, opened 2026-04-07
Abadawi/send has image to router

abadawi591:
## Delegate image routing to server-side auto-intent-service

The server-side auto-intent-service now handles `has_image` via a vision-aware pre-filter instead of returning HTTP 400. Remove the client-side `hasImage` early return in `_tryRouterSelection()` and send `has_image: true` in the router request body when images are present.

### Changes

- **`automodeService.ts`:** remove `hasImage` bail-out, pass `hasImage` to router
- **`routerDecisionFetcher.ts`:** accept `hasImage` param, include `has_image` in request body, parse error code from 400 responses
- **`automodeService.ts`:** handle `no_vision_models` error as a distinct fallback reason; `_applyVisionFallback` remains as safety net
- **`automodeService.spec.ts`:** replace skip-router test with send-`has_image` test, add `no_vision_models` error handling test, update transient fallback test


extensions/copilot/src/platform/endpoint/node/routerDecisionFetcher.ts
```diff
@@ -73,7 +83,15 @@ export class RouterDecisionFetcher {
...
 		if (!response.ok) {
-			throw new Error(`Router decision request failed with status ${response.status}: ${response.statusText}`);
+			const errorText = await response.text().catch(() => '');
+			let errorCode: string | undefined;
+			try {
+				const parsed = JSON.parse(errorText);
+				if (typeof parsed === 'object' && parsed !== null && 'error' in parsed && typeof parsed.error === 'string') {
+					errorCode = parsed.error;
+				}
+			} catch { /* not JSON */ }
```
bhavyaus, line 107: **[Council consensus — medium severity]** Unbounded router error text can flow into telemetry `reason`.

When the router returns a non-JSON body (e.g. an HTML error page, a gateway timeout), `errorCode` stays `undefined` and `errorText` — the raw response body — becomes the `RouterDecisionError.errorCode` via the `errorCode || errorText` fallback.

This then flows through `automodeService.ts` where `RouterDecisionError.errorCode` is mapped to `fallbackReason` and emitted in the `automode.routerFallback` telemetry event as `reason`.

Before this PR, `reason` was always one of a bounded set of short strings (`routerTimeout`, `routerError`, `hasImage`, etc.). After this PR it can be an arbitrarily large, unstructured server response body.

**Suggestion:** Only pass `errorCode` (the parsed value or `undefined`) to the `RouterDecisionError` constructor instead of `errorCode || errorText`. Let the `automodeService` catch block fall through to the existing `routerError` string when `e.errorCode` is falsy.

**Testing gap:** No tests cover the non-JSON error body path or JSON-without-`error` field path — adding one would have caught this.

extensions/copilot/src/platform/endpoint/node/routerDecisionFetcher.ts
```diff
@@ -48,12 +55,15 @@ export class RouterDecisionFetcher {
...
-	async getRouterDecision(query: string, autoModeToken: string, availableModels: string[], stickyThreshold?: number, contextSignals?: RoutingContextSignals, conversationId?: string, vscodeRequestId?: string): Promise<RouterDecisionResponse> {
+	async getRouterDecision(query: string, autoModeToken: string, availableModels: string[], stickyThreshold?: number, contextSignals?: RoutingContextSignals, conversationId?: string, vscodeRequestId?: string, hasImage?: boolean): Promise<RouterDecisionResponse> {
 		const startTime = Date.now();
 		const requestBody: Record<string, unknown> = { prompt: query, available_models: availableModels, ...contextSignals };
 		if (stickyThreshold !== undefined) {
 			requestBody.sticky_threshold = stickyThreshold;
 		}
+		if (hasImage) {
+			requestBody.has_image = true;
+		}
```
bhavyaus, line 80: **[Single-reviewer — low severity]** `has_image` is only serialized when truthy, so the router cannot distinguish "explicitly no image" from "old client / field absent".

This may be intentional ("field absent" == "no image"), but worth confirming against the router contract. If the router ever needs to distinguish the two cases, this would need to change to always send the field.

abadawi591: This is intentional and fine for the current design. The server-side router contract treats field-absent as "no image" which is the correct default. There's no reason the router needs to distinguish between "explicitly no image" and "old client" because both should route the same way (text-only routing).

extensions/copilot/src/platform/endpoint/node/routerDecisionFetcher.ts
```diff
@@ -73,7 +83,15 @@ export class RouterDecisionFetcher {
...
+			const errorText = await response.text().catch(() => '');
+			let errorCode: string | undefined;
+			try {
+				const parsed = JSON.parse(errorText);
+				if (typeof parsed === 'object' && parsed !== null && 'error' in parsed && typeof parsed.error === 'string') {
+					errorCode = parsed.error;
+				}
+			} catch { /* not JSON */ }
+			throw new RouterDecisionError(`Router decision request failed with status ${response.status}: ${response.statusText}`, errorCode || errorText);
 		}
```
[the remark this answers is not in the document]
abadawi591: @copilot apply changes based on [this feedback](https://github.com/microsoft/vscode/pull/308321#discussion_r3061853878)

extensions/copilot/src/platform/endpoint/node/automodeService.ts
```diff
@@ -297,7 +295,14 @@ export class AutomodeService extends Disposable implements IAutomodeService {
...
+			let fallbackReason: string;
+			if (isTimeout) {
+				fallbackReason = 'routerTimeout';
+			} else if (e instanceof RouterDecisionError && e.errorCode) {
+				fallbackReason = e.errorCode;
+			} else {
+				fallbackReason = 'routerError';
+			}
 			this._logService.error(`Failed to get routed model for conversation ${conversationId} (${fallbackReason}):`, (e as Error).message);
 			return { lastRoutedPrompt: prompt, fallbackReason };
```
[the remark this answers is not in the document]
abadawi591: @copilot apply changes based on [this feedback](https://github.com/microsoft/vscode/pull/308321#discussion_r3061853888)

extensions/copilot/src/platform/endpoint/node/routerDecisionFetcher.ts
```diff
@@ -38,6 +38,13 @@ export interface RoutingContextSignals {
 	prompt_char_count?: number;
 }
 
+export class RouterDecisionError extends Error {
```
jinu-jang, line 46: Add a docstring on when this error should be thrown.

extensions/copilot/src/platform/endpoint/node/routerDecisionFetcher.ts
```diff
@@ -38,6 +38,13 @@ export interface RoutingContextSignals {
...
+	override readonly name = 'RouterDecisionError';
+	constructor(message: string, public readonly errorCode?: string) {
```
jinu-jang, line 48: Not sure how plausible this is, but it would make this class a lot more usable if we knew what errorCodes this could throw. I am not suggesting the below is correct. But something like this would make it much better.
```suggestion
	constructor(message: string, public readonly errorCode?: 'routerTimeout' | 'routerError' | 'hasImage') {
```

jinu-jang (changes requested):

extensions/copilot/src/platform/endpoint/node/routerDecisionFetcher.ts
abadawi591: The errorCode comes from the server's response body ({ "error": "..." }), so the set of values is defined server-side and open-ended. Typing it as a fixed union would require client updates every time the server adds a new error code, which is what we're trying to avoid. The JSDoc already documents the known example (no_vision_models).

bhavyaus (approved):

roblourens (approved):

jinu-jang (approved):

PR #308325, opened 2026-04-07
Fix missing reasoningEffort telemetry for Claude models

kevin-m-kent:
## Problem

The `reasoningEffort` telemetry property on `response.success` and `response.error` events was only reading from `requestBody.reasoning?.effort`, which is set by the **Responses API** path (OpenAI models).

Claude models use the **Messages API** path, which sets effort via `requestBody.output_config?.effort` instead. As a result, reasoning effort was silently missing from telemetry for all Claude model requests.

## Fix

Added a fallback: `requestBody.reasoning?.effort ?? requestBody.output_config?.effort` on both `response.success` and `response.error` telemetry events so the effort level is captured regardless of which API path is used.

## Impact

- **`response.success`** — now logs `reasoningEffort` for Claude models
- **`response.error`** — now logs `reasoningEffort` for Claude models
- No behavior change for OpenAI models (the `reasoning.effort` path is checked first)

Note: Identical fix also raised for vscode-copilot-chat in https://github.com/microsoft/vscode-copilot-chat/pull/5041


bhavyaus (approved):
LGTM!

bhavyaus:
PR is stuck due to failing component screenshots. 
Cherry-picked changes into https://github.com/microsoft/vscode/pull/309639

PR #308333, opened 2026-04-07
Allow background notifs setting to be controlled via exp

cwebster-99:
Configuration update:

* Added an `experiment` property with `mode: 'auto'` to the `backgroundNotifications` configuration option in `terminalChatAgentToolsConfiguration`, enabling experimental rollout of automatic agent notifications for background terminal commands.


meganrogge (approved):
Thank you!

PR #308429, opened 2026-04-08
Fix arrow key navigation not reaching filter button in notebook find widget

### LINKED ISSUE #207765: Cannot use arrow keys to go to toggle icon in notebook find widget

yogeshwaran-c:
Fixes microsoft/vscode#207765

## Problem

In the notebook find widget, arrow key navigation among the toggle buttons only cycles through the three standard toggles (case sensitive, whole words, regex) but skips the filter (funnel) button added by `NotebookFindInput`.

The root cause was that `FindInput` hardcoded its arrow-key navigation index list as a local constant captured in a closure during the constructor, making it impossible for subclasses to extend the list with additional controls.

## Solution

Extract the index computation into a protected `getOptionNavigationIndexes()` method that is called dynamically on each keydown event. The default implementation returns the standard toggles plus any `additionalToggles`.

`NotebookFindInput` overrides this method to also include the filter button's focusable anchor element, so left/right arrow keys now correctly cycle through all four toggle buttons.

## Files changed

- `src/vs/base/browser/ui/findinput/findInput.ts` — add protected `getOptionNavigationIndexes()` method
- `src/vs/workbench/contrib/notebook/browser/contrib/find/notebookFindReplaceWidget.ts` — override to include filter button

### DIFF
--- a/src/vs/base/browser/ui/findinput/findInput.ts
+++ b/src/vs/base/browser/ui/findinput/findInput.ts
@@ -176,25 +176,25 @@ export class FindInput extends Widget {
 			// Arrow-Key support to navigate between options
 			this.onkeydown(this.domNode, (event: IKeyboardEvent) => {
 				if (event.equals(KeyCode.LeftArrow) || event.equals(KeyCode.RightArrow) || event.equals(KeyCode.Escape)) {
-					const indexes = this.getOptionNavigationIndexes();
-					const index = indexes.indexOf(<HTMLElement>this.domNode.ownerDocument.activeElement);
+					const elements = this.getOptionNavigationElements();
+					const index = elements.indexOf(<HTMLElement>this.domNode.ownerDocument.activeElement);
 					if (index >= 0) {
 						let newIndex: number = -1;
 						if (event.equals(KeyCode.RightArrow)) {
-							newIndex = (index + 1) % indexes.length;
+							newIndex = (index + 1) % elements.length;
 						} else if (event.equals(KeyCode.LeftArrow)) {
 							if (index === 0) {
-								newIndex = indexes.length - 1;
+								newIndex = elements.length - 1;
 							} else {
 								newIndex = index - 1;
 							}
 						}
 
 						if (event.equals(KeyCode.Escape)) {
-							indexes[index].blur();
+							elements[index].blur();
 							this.inputBox.focus();
 						} else if (newIndex >= 0) {
-							indexes[newIndex].focus();
+							elements[newIndex].focus();
 						}
 
 						dom.EventHelper.stop(event, true);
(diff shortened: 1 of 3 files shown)


yogeshwaran-c:
@Yoyokrazy unfortunately I don't have a desktop session available to capture a gif, but here are the manual steps I've stepped through against this branch and what I expect at each step:

1. Open any notebook in VS Code.
2. `Ctrl+F` to open the find widget — focus lands in the input.
3. Press the right arrow once: focus moves to **Match Case**.
4. Right arrow again -> **Match Whole Word**.
5. Right arrow again -> **Use Regular Expression**.
6. Right arrow again -> **Filter (funnel)** button. *(Before this PR, focus would have wrapped back to Match Case here, skipping the filter button entirely — that is the bug from #207765.)*
7. Right arrow once more wraps back to **Match Case**.
8. Left arrow from any position walks the same cycle in reverse.
9. From any toggle, **Esc** returns focus to the input.
10. Toggle the filter on so the regex toggle is removed from the tab order (notebook sets `regex.domNode.tabIndex = -1`). Arrow keys now skip the regex toggle and cycle: Match Case -> Match Whole Word -> Filter. (This is new in the latest commit, see below.)

If a gif is a hard requirement before review I'm happy to find a way to record one — but I'd also gladly add an automated DOM-level test (driving keydown events through `FindInput` / `NotebookFindInput` and asserting the focused element after each arrow press) which would lock the behavior in regardless of platform. Let me know which you'd prefer.

Separately, addressed both Copilot suggestions in 3d2b180:

- Renamed `getOptionNavigationIndexes()` -> `getOptionNavigationElements()` (and the local `indexes` -> `elements`), since it returns `HTMLElement[]` not numeric indexes.
- The notebook override now filters out controls that aren't currently focusable (`tabIndex < 0`, `aria-disabled=true`, `display:none`) before passing them to arrow-key nav, so the regex toggle is skipped while filters are active. Helper uses the existing `dom.getWindow(...).getComputedStyle(...)` pattern already used elsewhere in the notebook find widget.

yogeshwaran-c:
@Yoyokrazy I've added an automated DOM-level test that exercises the arrow-key navigation behavior this PR fixes, since I'm working in a headless environment and can't capture a gif.

The test lives at `src/vs/base/test/browser/ui/findinput/findInput.test.ts` and drives the widget directly via simulated keydown events. It covers:

- Left/Right arrow keys cycling focus through the common find toggles (with wrap-around) and Escape returning focus to the input box.
- The fix itself: an additional focusable control appended by a subclass (modelling the notebook filter button) is reached by arrow keys and is part of the navigation cycle.
- A toggle removed from the tab order (`tabIndex = -1`, as the notebook widget does for the regex toggle when filters are active) is correctly skipped rather than swallowing focus.

This verifies the navigation works at the DOM/event level. Could you take another look when you get a chance? Happy to adjust the test if you'd prefer it scoped differently.

PR #308434, opened 2026-04-08
Add agent network filter policy settings

dmitrivMS:
Renamed `chat.agent.sandbox.[allowed/denied]NetworkDomains` settings to `chat.agent.[allowed/denied]NetworkDomains`.
Added migrations and moved settings definitions to common chat contribution.
Moved domain matching code from sandbox implementation to a shared location.
Used new settings and domain matching code to implement `AgentNetworkFilterService`.
Used new service to filter (allow/deny) network traffic in web fetch and integrated browser chat tools.

Sample chat interaction:
image

Fixes https://github.com/microsoft/vscode-internalbacklog/issues/7235


TylerLeonhardt (dismissed):
Approving from Fetch perspective

src/vs/workbench/contrib/browserView/electron-browser/tools/openBrowserTool.ts
```diff
@@ -74,14 +77,23 @@ export class OpenBrowserTool implements IToolImpl {
...
 		if (!params.forceNew) {
 			const existingResult = await getExistingPagesResult(this.editorService, this.playwrightService, params.url);
 			if (existingResult) {
 				return existingResult;
 			}
 		}
 
-		const { pageId, summary } = await this.playwrightService.openPage(params.url);
+		const { pageId, summary } = await this.playwrightService.openPage(params.url, {
+			isDomainAllowed: (uri) => this.agentNetworkFilterService.isUriAllowed(uri),
```
kycutler: Does this work? The playwright service is in shared process so functions usually can't be serialized

src/vs/platform/browserView/node/playwrightService.ts
```diff
@@ -371,7 +372,7 @@ class PlaywrightPageManager extends Disposable {
 	 * Create a new page in the browser and return its associated page ID.
 	 * The page is automatically added to the tracked set.
 	 */
-	async newPage(url: string): Promise<string> {
+	async newPage(url: string, options?: { isDomainAllowed?: (uri: URI) => boolean }): Promise<string> {
```
kycutler: This seems very fragile and easy to cheat -- for example users can manually navigate to a blocked URL and then share the page with the agent, pages could contain a link to a blocked URL and the agent click that, or plenty of other ways such as playwright code or in-page scripts.

IMO the handling would be best centralized in the `PlaywrightTab` class, in two places:
1. A `route()` handler to block new requests. Keep in mind this is in the shared process --you should be able to access ConfigurationService, but you might want to move the domain helpers to a shared location for reuse. Requests should also fail with an informative message so users can know to stop sharing with the agent in order to let the request succeed.
2. ~~In `getSummary`, exclude the page title / snapshot from the result if the URL is blocked. This covers existing pages opened in the browser and then shared.~~ Thinking about this more, `safeRunAgainstPage` might be a better place -- if the current URL is blocked then just fail before running the action

src/vs/workbench/contrib/browserView/electron-browser/tools/navigateBrowserTool.ts
```diff
@@ -111,6 +114,10 @@ export class NavigateBrowserTool implements IToolImpl {
 			case 'forward':
 				return playwrightInvoke(this.playwrightService, params.pageId, (page) => page.goForward({ waitUntil: 'domcontentloaded' }));
 			default: {
+				const uri = URI.parse(params.url!);
+				if (!this.agentNetworkFilterService.isUriAllowed(uri)) {
```
kycutler: Would this be better done in `prepareToolInvocation`? We can just throw there so the model gets an error and user doesn't have to approve only to then be rejected

kycutler (changes requested):

src/vs/platform/webContentExtractor/electron-main/webPageLoader.ts
```diff
@@ -279,9 +279,17 @@ export class WebPageLoader extends Disposable {
 		}
 
 		this.trace(`Received 'will-navigate' or 'will-redirect' event, url: ${url}`);
-		if (!this._options?.followRedirects) {
-			const toURI = URI.parse(url);
 
+		// Check domain filter policy first — this applies regardless of followRedirects
+		const toURI = URI.parse(url);
+		if (this._options?.isDomainAllowed && !this._options.isDomainAllowed(toURI)) {
```
joshspicer: Are there cases where `isDomainAllowed` is undefined?

src/vs/workbench/contrib/browserView/electron-browser/tools/openBrowserTool.ts
```diff
@@ -74,14 +77,23 @@ export class OpenBrowserTool implements IToolImpl {
 	async invoke(invocation: IToolInvocation, _countTokens: CountTokensCallback, _progress: ToolProgress, _token: CancellationToken): Promise<IToolResult> {
 		const params = invocation.parameters as IOpenBrowserToolParams;
 
+		const uri = URI.parse(params.url);
+		if (!this.agentNetworkFilterService.isUriAllowed(uri)) {
+			return {
+				content: [{ kind: 'text', value: localize('browser.open.blockedByPolicy', 'Access to {0} is blocked by network domain policy.', params.url) }]
```
joshspicer: Thoughts on deep-linking to the setting (we do this for `eligibleForAutoApproval`) so user could see what the enterprise has enforced?

src/vs/platform/networkFilter/common/domainMatcher.ts
```diff
@@ -0,0 +1,169 @@
...
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { URI } from '../../../../../base/common/uri.js';
+
+/**
+ * Common file extensions that are unlikely to be network domains.
+ * Used to filter false positives when extracting domains from freeform text.
+ */
+export const fileExtensionSuffixes = new Set([
```
joshspicer: I know you had questions about this yesterday - do you know why these are here?

src/vs/platform/networkFilter/common/domainMatcher.ts
```diff
@@ -0,0 +1,169 @@
...
+ * Normalizes and validates a domain string.
+ *
+ * Strips user info, port, trailing dots, and trailing punctuation.
+ * Accepts bare wildcards (`*`) and wildcard prefixes (`*.example.com`).
+ *
+ * @param value The raw domain string to normalize.
+ * @param fromUrl Whether the value was extracted from a URL context (skips file-extension filtering).
+ * @returns The normalized domain string, or `undefined` if the input is invalid.
+ */
+export function normalizeDomain(value: string | undefined, fromUrl: boolean = false): string | undefined {
```
joshspicer, line 37: A bit surprised we don't have these utilities yet

src/vs/platform/networkFilter/common/domainMatcher.ts
dmitrivMS: It turns out this was part of existing code (which is parsing out URLs from responses as per @dileepyavan).
So leaving it unchanged at least for this iteration.

src/vs/platform/networkFilter/common/domainMatcher.ts
dmitrivMS: We have similar code handling trusted domains, but there are differences. This code was pulled out of sandbox feature unchanged, so I'd like to keep it for this iteration. We can discuss reconciliation later with @TylerLeonhardt and @dileepyavan.

src/vs/workbench/contrib/browserView/electron-browser/tools/openBrowserTool.ts
dmitrivMS: Added setting reference.

src/vs/platform/webContentExtractor/electron-main/webPageLoader.ts
dmitrivMS: No, there shouldn't be. Fixed.

src/vs/workbench/contrib/browserView/electron-browser/tools/navigateBrowserTool.ts
dmitrivMS: Good idea, moved!

src/vs/workbench/contrib/browserView/electron-browser/tools/openBrowserTool.ts
dmitrivMS: Moved new service to Platform and updated.
As the image abvoe shows, something did work, but I may need more scenarios to test with.

src/vs/platform/browserView/node/playwrightService.ts
dmitrivMS: Thank you! Done.

joshspicer (dismissed):
Applying policy looks good 👍 🔥

kycutler (approved):

kycutler (dismissed):
Integrated browser bits look good 👍

TylerLeonhardt (dismissed):

src/vs/workbench/contrib/terminalContrib/chatAgentTools/common/terminalChatAgentToolsConfiguration.ts
jruales, line 527: Should we link to #chat.agent.allowedNetworkDomains# and #chat.agent.deniedNetworkDomains# from here so we can get from one to the other easily?

dmitrivMS: @dileepyavan ?

jruales (changes requested):
Found something that needs to be fixed, and shared it yesterday via DM

joshspicer (dismissed):

src/vs/workbench/contrib/browserView/electron-browser/features/browserEditorChatFeatures.ts
```diff
@@ -148,6 +152,21 @@ export class BrowserEditorChatIntegration extends BrowserEditorContribution {
 		if (!model) {
 			return;
 		}
+
+		// Prevent sharing when the current page URL is blocked by network policy.
+		if (!model.sharedWithAgent && model.url) {
```
kycutler: Can we move this into `BrowserViewModel.setSharedWithAgent` so it's more centralized? And show a dialog instead of a notification (with a message that's easily understood by the user)

src/vs/platform/browserView/node/playwrightTab.ts
```diff
@@ -208,6 +208,19 @@ export class PlaywrightTab {
...
 
+		// When the current page URL is blocked by network policy, return only a
+		// policy error — do not expose title, URL, console logs, or snapshot to
+		// avoid prompt-injection via blocked content.
+		const currentUrl = this.page.url();
+		if (currentUrl && currentUrl !== 'about:blank') {
+			let uri: URI | undefined;
+			try { uri = URI.parse(currentUrl); } catch { }
+			if (uri && !this.agentNetworkFilterService.isUriAllowed(uri)) {
+				this._logs = [];
```
kycutler: Logs are only returned from the summary below -- they shouldn't be cleared here.

src/vs/platform/browserView/node/playwrightTab.ts
```diff
@@ -208,6 +208,19 @@ export class PlaywrightTab {
 	async getSummary(full = this._needsFullSnapshot): Promise<string> {
 		await this._initialized;
 
+		// When the current page URL is blocked by network policy, return only a
+		// policy error — do not expose title, URL, console logs, or snapshot to
+		// avoid prompt-injection via blocked content.
+		const currentUrl = this.page.url();
+		if (currentUrl && currentUrl !== 'about:blank') {
```
kycutler: Nit: there are 3 sets of similar logic in this file -- can we have something like:

```ts
private _getBlockedURLErrorMessage(url: string): string | undefined {
  // about:blank check, parsing fallback, etc.
  // return message from filter service if blocked
}

// ... in here and other places
const blockedError = this._getBlockedURLErrorMessage(this.page.url());
if (blockedError) {
  return blockedError;
}
```

jruales (approved):

joshspicer (approved):

PR #308457, opened 2026-04-08
fix: Add NSLocalNetworkUsageDescription to Info.plist

deepak1556:
macOS shows a "Local Network" privacy prompt when an app opens TCP/UDP connections to devices on the local network over Wi-Fi [per Apple TN3179](https://developer.apple.com/documentation/technotes/tn3179-understanding-local-network-privacy#Local-network-operations).

Chromium's built-in DNS client (enabled by default on macOS via the kAsyncDns feature flag) sends DNS queries directly over UDP sockets to the system-configured nameservers (when DoH is not enabled or in the fallback mode) unlike the system resolver path (getaddrinfo → mDNSResponder). On most Wi-Fi networks, the DNS server is in the same-subnet address and triggers the prompt on every launch.

**Before**

test


deepak1556:
verification https://monacotools.visualstudio.com/Monaco/_build/results?buildId=428711&view=results

lszomoru (approved):

Issue #305740, opened 2026-03-27
Host becomes unresponsive during synchronous subprocess wait instead of timing out cleanly


ajbudlong:
Type: <b>Performance Issue</b>

## Description

When an external subprocess stalls, the host can become unresponsive if the subprocess is waited on synchronously from a UI-thread or event-loop-owned execution path. I believe this is a separate resilience issue from the child-process bug itself.

I created a self-contained reproduction showing the host-side failure mode without depending on the original application code. The reproduction demonstrates that a blocking `subprocess.wait()` on the event-loop thread starves the host loop, while offloading the same wait preserves responsiveness.

This report is about the host freeze behavior, not the original child-process locking bug.

## VS Code Version

Observed in a Windows environment. Exact VS Code build should be filled in from the affected machine when filing.

## OS Version

Windows

## Steps to Reproduce

1. Run the self-contained reproduction in [host_freeze_repro.py](host_freeze_repro.py).
2. Observe the `Blocking subprocess.wait() on event-loop thread` section.
3. Compare it with the `Offloaded subprocess.wait() preserves event loop` section.
4. Optionally run [test_host_freeze_repro.py](test_host_freeze_repro.py) with pytest.

Commands:

```powershell
python host_freeze_repro.py
python -m pytest test_host_freeze_repro.py -q
```

## Expected Behavior

- A slow or stalled subprocess should not freeze the host.
- UI/event-loop work should continue while the child process is pending.
- Timeout and cancellation should remain serviceable.
- The host should surface an error state rather than becoming unresponsive.

## Actual Behavior

- If the host performs a synchronous wait on the same thread that services events, event-loop progress stalls.
- In the repro, heartbeat activity pauses for approximately the same duration as the child-process wait.
- The offloaded version does not show the same starvation pattern.

## Does This Issue Occur When All Extensions Are Disabled?

The original symptom involved extension-driven subprocess execution, so the real-world scenario is extension-related. The attached repro is extension-independent and demonstrates the host-side design risk directly.

## Reproduction Files

- [host_freeze_repro.py](host_freeze_repro.py)
- [test_host_freeze_repro.py](test_host_freeze_repro.py)
- [HOST_FREEZE_REPORT.md](HOST_FREEZE_REPORT.md)

## Why I Think This Belongs In VS Code

[...]

PR #308681, opened 2026-04-09
Fix/305740 subprocess wait

rupeshkumar-555:
## Summary\nFixes #305740 by removing synchronous subprocess-based node detection in TypeScript configuration and resolving node executable via PATH/PATHEXT with platform-aware handling.\n\n## Why\nThe sync subprocess approach can block and cause delays/hangs in some environments.\n\n## What changed\n- Updated node path resolution in configuration.electron.ts\n- Added focused unit tests in configuration.electron.test.ts for Windows/Unix and fallback behavior\n\n## Scope\nThis PR now contains only the #305740 fix changes.


rupeshkumar-555:
Thanks for the review request. Context for this PR: issue #305740 reports that TS node path detection can block on a synchronous subprocess call. This change removes the sync subprocess path and resolves node via PATH/PATHEXT with cross-platform handling, plus focused unit tests for resolver behavior. I have now force-pushed the branch to contain only the #305740 fix files to keep the scope narrow. Please re-review the updated diff.

rupeshkumar-555:
Applied the Copilot review feedback in the latest commit:

- **\defaultIsExecutableFile\**: replaced \existsSync\ + \lstatSync\ with a single \lstatSync\ call, and added an \s.accessSync(candidate, fs.constants.X_OK)\ check on non-Windows so only truly executable files are matched.
- **\
esolveNodeExecutableFromPath\**: added a \cwd\ fallback (including PATHEXT handling on win32) that runs after the PATH-based search exhausts or when PATH is missing, aligning with VS Code\u2019s \indExecutable\ behaviour.
- **Tests**: added three new unit tests covering the cwd fallback on Linux, win32 (with PATHEXT), and the null-return path when neither PATH nor cwd contains node.

The PR now only touches \configuration.electron.ts\ and \configuration.electron.test.ts\ \u2014 fully scoped to issue #305740.

PR #308683, opened 2026-04-09
WIP: (Explroe) Enhance session metadata handling in commit commands

DonJayamanne:
Update commit commands to include session URI, allowing for direct updates to session metadata when committing or deleting customization files. This change improves the integration of session management with the Copilot CLI.

For https://github.com/microsoft/vscode/issues/308656


DonJayamanne:
@copilot update the code changes in src/vs/sessions/contrib/chat/browser/aiCustomizationWorkspaceService.ts to ensure sessionUri is optional.
This is for backwards compatibility, we want to ensure it works the way it did in the past,
Once this is tested in the fugure we can make it mandatory, but for now, sessionUri needs to be optional in the new places whree its been added in this PR src/vs/sessions/contrib/chat/browser/aiCustomizationWorkspaceService.ts

PR #308696, opened 2026-04-09
Close non-dirty editors when cancelling Close All Editors

### LINKED ISSUE #305306: "Close All Editors" cancels entire operation when "Cancel" is selected on unsaved file prompt
### Observed Behavior

When using "Close All Editors", if an unsaved file is encountered, VS Code prompts with **Save**, **Don’t Save**, or **Cancel**.

If **Cancel** is selected, the entire "Close All Editors" operation is aborted, and no tabs are closed, including those without unsaved changes.

### Expected Behavior

All editors without unsaved changes should be closed, and the editor with unsaved changes should remain open.

yogeshwaran-c:
## Summary

Fixes #305306

When using **Close All Editors**, if an unsaved file prompts for confirmation and the user selects **Cancel**, the entire operation was previously aborted -- no editors were closed at all, even those without unsaved changes.

This PR changes the behavior so that cancelling the confirmation dialog only prevents closing the dirty (unsaved) editors. All editors that do not require confirmation are still closed, matching the expected behavior described in the issue.

### Changes

- In AbstractCloseAllAction.run(), instead of returning immediately on ConfirmResult.CANCEL, we now track the cancellation state and skip only the confirmation-related steps (save/revert of dirty editors)
- The final doCloseAll() call now passes excludeConfirming: true when the user cancelled, leveraging the existing excludeConfirming option in ICloseAllEditorsOptions to close only non-dirty editors
- Custom confirm handlers (step 2) are also skipped when cancelled, and auto-save steps (3/4) are skipped since they only apply to dirty editors

### Before

1. Open several files, modify one
2. Run Close All Editors
3. Click Cancel on the save prompt
4. **Result**: No editors are closed

### After

1. Open several files, modify one
2. Run Close All Editors
3. Click Cancel on the save prompt
4. **Result**: All clean (non-dirty) editors are closed; only the unsaved editor remains open

### DIFF
--- a/src/vs/workbench/browser/parts/editor/editorActions.ts
+++ b/src/vs/workbench/browser/parts/editor/editorActions.ts
@@ -635,6 +635,7 @@ abstract class AbstractCloseAllAction extends Action2 {
 		}
 
 		// 1.) Show default file based dialog
+		let cancelled = false;
 		if (dirtyEditorsWithDefaultConfirm.size > 0) {
 			const editors = Array.from(dirtyEditorsWithDefaultConfirm.values());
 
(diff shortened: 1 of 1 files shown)


dmitrivMS:
1. There are no tests.
2. The linked issue is already closed as not planned.

PR #308697, opened 2026-04-09
fix: check message location visibility for failureInVisibleDocument peek

### LINKED ISSUE #230881: Testing: Automatically Open Peek View with 'failureInVisibleDocument' doesn't work
Does this issue occur when all extensions are disabled?: No, [Playwright extension](https://marketplace.visualstudio.com/items?itemName=ms-playwright.playwright) is required.

- VS Code Version: 1.94.1
- OS Version: macOS 15.0

Steps to Reproduce:

1. `npm init playwright@latest`
2. Replace `example.spec.ts` with that:
  ```ts
import { test } from "@playwright/test";
import { foo } from './util';

test("1", () => {
    foo()
});
  ```
3. Create `util.ts`
  ```ts
export const foo = () => {
    throw new Error('foo')
}
  ```
4. Set `"testing.automaticallyOpenPeekView": "failureInVisibleDocument"` as a setting
5. Run the test

Expected: It does **not** peek to the `util.ts`

Actual: It peaks to the `util.ts`

yogeshwaran-c:
## Summary

- Fixes an issue where `testing.automaticallyOpenPeekView` set to `failureInVisibleDocument` would still open peek views in non-visible documents
- When a test in a visible file fails with an error located in a different file (e.g. a utility module), VS Code previously opened that non-visible file just to show the peek, contradicting the setting's intent
- Now both the test item's URI **and** the message location URI must be in visible editors for the peek to auto-open

Fixes #230881

## Test plan

- [ ] Set `"testing.automaticallyOpenPeekView": "failureInVisibleDocument"`
- [ ] Create a test file (`example.spec.ts`) that calls a utility function in a separate file (`util.ts`) which throws an error
- [ ] Open only `example.spec.ts` in the editor (ensure `util.ts` is NOT open)
- [ ] Run the failing test
- [ ] Verify the peek view does **not** auto-open (since the error location `util.ts` is not visible)
- [ ] Now open both `example.spec.ts` and `util.ts` side by side
- [ ] Re-run the failing test
- [ ] Verify the peek view **does** auto-open (since both the test and message location are visible)

### DIFF
--- a/src/vs/workbench/contrib/testing/browser/testingOutputPeek.ts
+++ b/src/vs/workbench/contrib/testing/browser/testingOutputPeek.ts
@@ -310,6 +310,13 @@ export class TestingPeekOpener extends Disposable implements ITestingPeekOpener 
 				if (!Iterable.some(resultItemParents(evt.result, evt.item), i => i.item.uri && editorUris.has(i.item.uri.toString()))) {
 					return;
 				}
+				// Also check that the message location itself is in a visible
+				// document. The message may point to a different file (e.g. a
+				// utility) than where the test is defined, and opening a non-
+				// visible file just to show a peek would be disruptive.
+				if (!editorUris.has(candidate.location.uri.toString())) {
+					return;
+				}
 				break; //continue
 			}
 			case AutoOpenPeekViewWhen.FailureAnywhere:


connor4312 (approved):
thanks!

TylerLeonhardt (approved):

Issue #133397, opened 2021-09-19
Wrong autoClosingPairs with embedded language


jlelong:
- VS Code Version: 1.60.1
- OS Version: OSX 11.6

When a language is embedded, the value of `autoClosingPairs` is taken from the parent language, not the embedded one. Related to #117190

Steps to reproduce

1. Create a LaTeX file containing

    ```latex
    \begin{cppcode}
       int main() {
         return 0;
      }
    \end{cppcode}
    ```
2. Check that inside the `cppcode` environment,  we use the `cpp` embedded language

    <img width="662" alt="Capture d’écran 2021-09-19 à 12 01 42" src="https://user-images.githubusercontent.com/2910140/133923509-52daa310-89ff-49cd-8f56-6b95fbff1390.png">

3. Insert `` ` `` inside the `cppcode` environment. It automatically gets closed by `'`.
 
    ![Sep-19-2021 12-05-10](https://user-images.githubusercontent.com/2910140/133923502-de9705de-799a-4686-a2d9-5ece8a0a0c75.gif)

The [latex language configuration](https://github.com/James-Yu/LaTeX-Workshop/blob/715526b0eb1a0f19aa4989163414e936507cdaf1/syntax/latex-language-configuration.json#L46) adds ``["`", "'"]`` to `autoClosingPairs` but not the [latex cpp embedded language](https://github.com/James-Yu/LaTeX-Workshop/blob/715526b0eb1a0f19aa4989163414e936507cdaf1/syntax/latex-cpp-embedded-language-configuration.json). 

Am I missing something? For it is worth, I am maintaining the LaTeX-Workshop extension.

alexdima:
It looks like inside the LaTeX `cppcode` block, the language is defined as `cpp_embedded_latex`. That language would benefit from a language configuration like the [regular `cpp` language](https://github.com/microsoft/vscode/blob/81ac1e1ee94aead44c6c70ff172423d57c5db7f7/extensions/cpp/package.json#L26). Is there a particular reason why you're not defining the language inside as `cpp` like e.g. markdown does?
<img width="607" alt="Screen Shot 2021-09-27 at 08 31 11" src="https://user-images.githubusercontent.com/5047891/134855859-663c0fa4-bb3a-4e6b-ac22-29a81a65df16.png">

jlelong:
[...]

PR #308733, opened 2026-04-09
Fix auto-closing angle brackets inside inline math blocks

### LINKED ISSUE #272922: [Markdown] Do not auto close angle brackets inside latex blocks

maruthang:
Fixes #272922

**Bug:** Angle brackets (`<`, `>`) are auto-closed inside inline LaTeX math blocks (`$...$` and `$$...$$`), which disrupts writing math expressions like `$x < y$`.

**Root Cause:** The `math_inline_single` and `math_inline_double` TextMate grammar patterns used `match` rules, which only apply after the full expression is typed. While the user is still typing between delimiters, the editor does not recognize the math context and applies default auto-closing behavior.

**Fix:** Converted both patterns from `match` rules to `begin/end` rules in `md-math-inline.tmLanguage.json`. This allows the editor to recognize the math scope as soon as the opening delimiter is typed, suppressing auto-closing of angle brackets while inside math blocks.

## Changes

- `extensions/markdown-math/syntaxes/md-math-inline.tmLanguage.json`: Converted `math_inline_single` and `math_inline_double` from `match` to `begin/end` patterns so the math scope is active while typing between delimiters.
- `extensions/vscode-colorize-tests/test/colorize-fixtures/md-math.md`: Added test cases for inline math with angle brackets.
- Deleted `extensions/vscode-colorize-tests/test/colorize-results/md-math_md.json` and `extensions/vscode-colorize-tests/test/colorize-tree-sitter-results/md-math_md.json` for regeneration with updated tokenization.

## Test Plan

- Open a Markdown file and type `$x < y$` -- the `<` should not trigger auto-closing of a `>` bracket.
- Type `$$\sum_{i<n}$$` -- same behavior, no auto-close inside double-dollar math.
- Verify that math syntax highlighting still works correctly for inline math expressions.
- Run the colorize tests to regenerate and validate the snapshot files.


maruthang:
You're right - apologies for the duplicate. This overlaps with #308431 and the real fix belongs in the core editor (#133397) for handling embedded language auto-closing pairs properly. Closing in favor of those. Thanks for pointing me to the root cause!

PR #308739, opened 2026-04-09
fix: resolve default view for markdown files on first startup

### LINKED ISSUE #244597: Default view for markdown files doesn't work the first time vscode is started

- VS Code Version: 1.98.2
- OS Version: Win10 22H2

Steps to Reproduce:

1. In the settings, add Item: *.md, Value: vscode.markdown.preview.editor under _Workbench: Editor Associations_
2. Close vscode
3. With no vscode instances running, open .md file from explorer
4. Text editor view is opened instead of the preview
5. Verify that the setting is registered by checking Ctrl + P > View: Reopen editor with..., Markdown Preview is listed as Default
6. Open another .md file, this second one is now opened in preview mode

maruthang:
Fixes #244597

## Summary

**Bug:** Default view for markdown files doesn't work on first startup -- the user-configured custom editor (e.g., markdown preview) is not used when opening `.md` files for the first time.

**Root Cause:** The `EditorResolverService` uses a cache of glob patterns (persisted via `StorageService`) to decide whether to wait for extensions to register before resolving editors. On first startup, this cache is empty because it is only populated during `onWillSaveState`, so the resolver skips waiting for extensions and falls back to the default text editor.

**Fix:** Added a `resourceMatchesUserAssociation()` method that checks if the resource matches any user-configured `editorAssociations` pointing to a non-default editor. This is used as a fallback alongside the existing cache check, ensuring the resolver waits for extensions even on first startup when the cache is empty.

## Changes

- `src/vs/workbench/services/editor/browser/editorResolverService.ts`: Added `resourceMatchesUserAssociation()` method and updated the cache check condition to also consider user-configured editor associations as a fallback.
- `src/vs/workbench/services/editor/test/browser/editorResolverService.test.ts`: Added regression test that verifies user-configured editor associations resolve correctly even with an empty cache (simulating first startup).

## Test Plan

- [x] Added regression test `User-configured editor association resolves on first startup with empty cache #244597`
- [ ] Verify on a fresh VS Code profile that opening a `.md` file uses the configured custom editor (e.g., markdown preview) on first startup
- [ ] Verify existing editor resolution behavior is unchanged for subsequent startups (cache populated)
- [ ] Verify files without user associations still resolve to the default editor

### DIFF
--- a/src/vs/workbench/services/editor/browser/editorResolverService.ts
+++ b/src/vs/workbench/services/editor/browser/editorResolverService.ts
@@ -125,7 +125,7 @@ export class EditorResolverService extends Disposable implements IEditorResolver
 		let resource = EditorResourceAccessor.getCanonicalUri(untypedEditor, { supportSideBySide: SideBySideEditor.PRIMARY });
 
 		// If it was resolved before we await for the extensions to activate and then proceed with resolution or else the backing extensions won't be registered
-		if (this.cache && resource && this.resourceMatchesCache(resource)) {
+		if (this.cache && resource && (this.resourceMatchesCache(resource) || this.resourceMatchesUserAssociation(resource))) {
 			await this.extensionService.whenInstalledExtensionsRegistered();
 		}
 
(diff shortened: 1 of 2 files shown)


mjbvz (approved):
thanks

alexdima (approved):

PR #308776, opened 2026-04-09
feat(plugins): allow component paths within repository boundary

xAndreiLi:
Add an optional `boundaryUri` parameter to `resolveComponentDirs` and `parsePlugin` that relaxes the path containment check from the plugin directory to a wider boundary (e.g. the repository root).

Previously, relative paths like `../../skills/shared-skill` in a plugin manifest were silently dropped because the resolved URI fell outside the plugin directory. For marketplace plugins whose repository contains multiple plugins sharing resources, this was too restrictive.

Now `MarketplaceAgentPluginDiscovery` passes the repository URI (from `IAgentPluginRepositoryService.getRepositoryUri`) as the boundary, so paths that stay within the repository are accepted. Non-marketplace plugins continue using the plugin directory as the default boundary, preserving the existing security constraint.

Closes #308775

### DIFF
--- a/src/vs/platform/agentPlugins/common/pluginParsers.ts
+++ b/src/vs/platform/agentPlugins/common/pluginParsers.ts
@@ -187,16 +187,18 @@ export function parseComponentPathConfig(raw: unknown): IComponentPathConfig {
 /**
  * Resolves the directories to scan for a given component type, combining
  * the default directory with any custom paths from the manifest config.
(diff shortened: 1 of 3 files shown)


xAndreiLi:

benvillalobos:
cc @connor4312 I believe you're most familiar with this area?

connor4312 (approved):
thanks!

roblourens (approved):

PR #308836, opened 2026-04-09
Don't send background notifications to main agent when subagent is used

### LINKED ISSUE #308048: execution subagent doesn't work with new async terminal stuff

The execution subagent only gets the run_in_terminal tool, so it can't check the output of terminals or do other things. Also, since it wants to finish its task before yielding, I see it just do random stuff while waiting until it runs out of turns

[8405ab43.copilotmd](https://github.com/user-attachments/files/26517361/8405ab43.copilotmd)

But even if it did have those tools, the async terminal strategy seems awkward with the execution subagent. If it starts async tasks, then yields, then the main agent gets the notification that the task is done, will the main agent have the right context to take action on that? Maybe, but that just needs some thinking for different scenarios.

cc @vikramnitin9

meganrogge:
fix #308048 

The subagent only had access to `run_in_terminal` — it couldn't check output, send input, or kill terminals. We've added access to those tools for the subagent https://github.com/microsoft/vscode-copilot-chat/pull/5050.

Background terminal completion notifications were sent to the parent session, which had no context for them. When it's a subagent, we no longer do this.


meganrogge:

PR #308889, opened 2026-04-09
agentHost: adopt notify/sessionSummaryChanged protocol notification

### LINKED ISSUE #305330: AHP: NeedsInput status
**Priority:** P1

Extension has `NeedsInput` for user questions; protocol only has `Idle`/`InProgress`/`Error`.

**Status:**
| Layer | Status |
|-------|--------|
| Protocol | N |
| CopilotAgent | N |
| VS Code | N |

connor4312:
- Syncs agent-host-protocol to pick up the new
  `notify/sessionSummaryChanged` notification type, which lets clients
  keep cached session lists in sync without subscribing to every
  session URI individually.
- Emits the notification from AgentHostStateManager: after the session
  reducer runs, an identity check on the summary object detects
  changes, which are debounced (100ms) and flushed as partial diffs.
- Replaces the onDidAction-based subscription-peeking pattern in
  AgentHostSessionListController with the cleaner notification-based
  approach, removing the need for StateComponents/isSessionAction
  imports and the getSubscriptionUnmanaged call.
- Adds unit tests for emission, coalescing, no-op suppression, and
  cleanup on session deletion.

Adopts https://github.com/microsoft/agent-host-protocol/pull/51
Fixes https://github.com/microsoft/vscode/issues/305330


roblourens (approved):

awakeairplane (commented):
Disable agent host

Issue #301876, opened 2026-03-15
"Request Failed: 413": GitHub Copilot Chat request errors while using GPT Codex: Session ruined indefinitely


Barrixar:
- Copilot Chat Extension Version: 0.40.2026031401
- VS Code Version: 1.112.0 Insider (2026-03-14)
- OS Version: 26200.8037
- Feature (e.g. agent/edit/ask mode): Agent
- Selected model (e.g. GPT 4.1, Claude 3.7 Sonnet): GPT-5.3-Codex
- Logs: 

[logs_1_WorkingSessionUntilItBroke.txt](https://github.com/user-attachments/files/26007931/logs_1_WorkingSessionUntilItBroke.txt)

[logs_2_PersistentFailureWhenIWrote-GoOn.txt](https://github.com/user-attachments/files/26007932/logs_2_PersistentFailureWhenIWrote-GoOn.txt)

**Sorry, your request failed. Please try again. Copilot Request id: [id] GH Request Id: [id] Reason: Request Failed: 413 {"error":{"message":"failed to parse request","code":""}}**

Steps to Reproduce:

1. Do your thing and work on a project in Agent Mode
2. Wait to become a victim of random bug state

GitHub Copilot Chat in Agent Mode intermittently (and randomly) bricks the chat for further use with GPT 5.3 Codex model. When something unknown happens, there is no way that Codex can be persuaded to continue to use the same session.

When this strikes, you will face:

<img width="442" height="1101" alt="Image" src="https://github.com/user-attachments/assets/35583299-238b-4793-b743-569fd042a06f" />
The screenshot coorresponds exactly with the logs i submit in this ticket.
I've collected the logs using [https://code.visualstudio.com/docs/copilot/troubleshooting](https://code.visualstudio.com/docs/copilot/troubleshooting)

Neither "Try Again" or a phrase :"Go on" will restore the situation. I tried both and collected logs for both.

It says (with unique ID's):
Sorry, your request failed. Please try again. Copilot Request id: [id] GH Request Id: [id] Reason: Request Failed: 413 {"error":{"message":"failed to parse request","code":""}}

Barrixar:
This has been an issue for quite a while. It's exacly why i like to avoid Codex models in VS Code.
Only VS Code is affected, it doesnt happen whhen consuming straight from OpenAI.

Barrixar:
I already isolated the cause of Error 413 with GPT models.

[...]

PR #308923, opened 2026-04-10
Background inline summarization v1

bhavyaus:
Move inline summarizer from foreground to background compaction loop to maximize prompt caching on summarization requests.

## Key changes
- Background inline summarization: forks exact messages from main render + appends summary user message for cache parity
- Background standard (triggerSummarize) as fallback when inline setting disabled
- Foreground renderWithSummarization as BudgetExceeded fallback
- 80% threshold kicks off background, BudgetExceeded waits/applies or falls back to foreground
- Removed BackgroundCompaction experiment flag (always-on when summarization enabled)
- Removed ConversationTranscriptLookup setting (always-on)
- Removed dead inlineSummarization prop and InlineSummarizationRequestedMetadata
- Telemetry parity across all summarization paths
- Standardized log prefixes


roblourens (approved):

Barrixar:
@bhavyaus 
This PR (or your other PR #308949) coincides suspiciously with the loss of /compact for GPT models, as documented here: https://github.com/microsoft/vscode/issues/301876#issuecomment-4227225802

The loss of this feature is extremely unfortunate, as it was the only known workaround for Error 413 when its cause is broken GPT auto-compaction > context overflow until its backend returns a hard "No" in that shape of Error 413. It's likely not intended that Copilot began taking away the user's freedom to use /compact about exactly 3 days ago. Hence it coincides.

With the recent breakage, it is no longer possible to salvage errored out GPT sessions, this hurts productivity by a lot. Of course, that bug shouldn't exist in the first place, but again, the only known workaround was also just bugged away from the user's control.

I'm also going to try alert the author of 9b93393 because clearly, GPT compaction is just not working.

bhavyaus:
@Barrixar /compact is working for GPT models. Could you give me a repro for the scenario you describe above?

Issue #304129, opened 2026-03-23
activeRepository* window title components seem broken


sbatten:
Testing #303345

1. Open a multi-root workspace
2. Set the window title to `${rootNameShort} - ${activeRepositoryBranchName}` or `${rootNameShort} - ${activeRepositoryName}`

🐛 => the active repo piece is blank

wazeerc:
was this fixed?

Joinyy:
I don't think so, I now have this bug in normal workspaces, no multi-root needed.
My setting:
```json
"window.title": "${activeRepositoryName}@${activeRepositoryBranchName} (${rootName})"
```
does result in the window title being ` @ (folder-name)`. This is always the case, although all other scm related functionalities work as expected.

<details>
<summary>Version info</summary>

```
  Version: 1.118.1 (user setup)
  Commit: 034f571df509819cc10b0c8129f66ef77a542f0e
  Date: 2026-04-29T17:36:44+03:00
  Electron: 39.8.8
  ElectronBuildId: 13870025
  Chromium: 142.0.7444.265
  Node.js: 22.22.1
  V8: 14.2.231.22-electron.0
  OS: Windows_NT x64 10.0.22631
  git version 2.52.0.windows.1
```

</details>

PR #308929, opened 2026-04-10
fix: re-evaluate active editor repository on SCM repository changes

### LINKED ISSUE #267239: Wrong active branch being used in repoContext
Image

That's not the branch though:

Image

Image

yogeshwaran-c:
## Summary

The `_activeEditorRepositoryObs` observable in `SCMViewService` only re-evaluated when the active editor changed. This caused the active repository context to become stale or empty when:

- An editor was already open before SCM providers finished loading their repositories
- Repositories were added or removed while the same editor remained active

This fix adds a dependency on `onDidAddRepository` / `onDidRemoveRepository` so the active editor repository is re-evaluated whenever the set of known repositories changes.

## Test plan

- Open a multi-root workspace and verify that `${activeRepositoryName}` and `${activeRepositoryBranchName}` window title variables reflect the correct repository immediately after load
- Open an editor, then add/remove a Git repository from the workspace, and verify the active repository context updates without requiring an editor switch

Fixes #267239

### DIFF
--- a/src/vs/workbench/contrib/scm/browser/scmViewService.ts
+++ b/src/vs/workbench/contrib/scm/browser/scmViewService.ts
@@ -278,9 +278,17 @@ export class SCMViewService implements ISCMViewService {
 			equalsFn: () => false
 		}, this.editorService.onDidActiveEditorChange, () => this.editorService.activeEditor);
 
+		// Track repository additions/removals so that the active editor
+		// repository is re-evaluated when SCM providers become available.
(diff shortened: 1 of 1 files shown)


yogeshwaran-c:
Also addresses #304129 — same root cause: `activeRepositoryName` / `activeRepositoryBranchName` window-title placeholders resolve to empty because `_activeEditorRepositoryObs` doesn't re-evaluate when SCM repositories change. Cross-linking so this PR can close both.

PR #308949, opened 2026-04-10
Fix double compaction on first-turn budget exceeded

bhavyaus:
On a first-turn `BudgetExceededError` with the background summarizer `Idle`, we fall back to synchronous foreground `full` summarization via `renderWithSummarization`. That path did not set the 'summary applied this iteration' flag, so the post-render gate (>= 80% + Idle) would *also* kick off a background `inline` compaction in the same `buildPrompt` call — producing both `summarizeConversationHistory-full` and `summarizeConversationHistory-inline` requests back-to-back.

### Fix

- Set `didSummarizeThisIteration = true` on both synchronous `renderWithSummarization` fallback call sites in the `BudgetExceededError` catch so the post-render gate correctly short-circuits.
- Rename `summaryAppliedThisIteration` → `didSummarizeThisIteration` so the name reflects that it covers any summarization work done this iteration (pre-render bg apply, budget-exceeded bg apply, or foreground fallback), not just applying a background result to rounds.

### DIFF
--- a/extensions/copilot/src/extension/intents/node/agentIntent.ts
+++ b/extensions/copilot/src/extension/intents/node/agentIntent.ts
@@ -465,8 +465,9 @@ export class AgentIntentInvocation extends EditCodeIntentInvocation implements I
 			? (this._lastRenderTokenCount + toolTokens) / baseBudget
 			: 0;
 
-		// Track whether we applied a summary in this iteration so we don't
-		// immediately re-trigger background compaction in the post-render check.
+		// Track whether this iteration already performed compaction-related work
+		// (including applying a summary or using a foreground fallback path) so
+		// we don't immediately re-trigger background compaction in the post-render check.
 		let didSummarizeThisIteration = false;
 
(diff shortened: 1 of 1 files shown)


lszomoru (approved):

Barrixar:
@bhavyaus 
This PR (or your other PR #308923) coincides suspiciously with the loss of /compact for GPT models, as documented here: https://github.com/microsoft/vscode/issues/301876#issuecomment-4227225802

The loss of this feature is extremely unfortunate, as it was the only known workaround for Error 413 when its cause is broken GPT auto-compaction. It's likely not intended that Copilot began taking away the user's freedom to use /compact about exactly 3 days ago. Hence it coincides.

With the recent breakage, it is no longer possible to salvage errored out GPT sessions, this hurts productivity by a lot. Of course, that bug shouldn't exist in the first place, but again, the only known workaround was also just bugged away from the user's control.

I'm also going to try alert the author of 9b93393 because clearly, GPT compaction is just not working.

PR #309075, opened 2026-04-10
Add browser tool instructions to agent prompts

kycutler:


extensions/copilot/src/extension/prompts/node/agent/openai/defaultOpenAIPrompt.tsx
```diff
@@ -63,6 +63,7 @@ export class DefaultOpenAIAgentPrompt extends PromptElement<DefaultAgentPromptPr
 				{tools[ToolName.CoreRunInTerminal] && <>NEVER try to edit a file by running terminal commands unless the user specifically asks for it.<br /></>}
 				{!tools.hasSomeEditTool && <>You don't currently have any tools available for editing files. If the user asks you to edit a file, you can ask the user to enable editing tools or print a codeblock with the suggested changes.<br /></>}
 				{!tools[ToolName.CoreRunInTerminal] && <>You don't currently have any tools available for running terminal commands. If the user asks you to run a terminal command, you can ask the user to enable terminal tools or print a codeblock with the suggested command.<br /></>}
+				{tools[ToolName.CoreOpenBrowserPage] && tools.hasAgenticBrowserTools && <>Use the browser tools ({ToolName.CoreOpenBrowserPage}, {agenticBrowserTools.find(k => tools[k])}, etc.) when beneficial for front-end tasks, such as when visualizing or validating UI changes.<br /></>}
```
jruales, line 66: nit: Do we still need the etc. if we already list all of them?

jruales (approved):

jruales (approved):

extensions/copilot/src/extension/prompts/node/agent/openai/defaultOpenAIPrompt.tsx
kycutler: The `.find()` only returns the first available one -- I just use it instead of hard-coding something like `click_element` in case only some are disabled.

PR #309125, opened 2026-04-11
Avoid installer UI in background update scenarios on Windows

dmitrivMS:
Protect some code paths with potential for race conditions to avoid installer UI launching in restart & update scenario when automated background updates are turned on.


src/vs/platform/update/electron-main/updateService.win32.ts
```diff
@@ -489,14 +495,22 @@ export class Win32UpdateService extends AbstractUpdateService implements IRelaun
...
 		} else {
-			spawn(this.availableUpdate.packagePath, ['/silent', '/log', '/mergetasks=runcode,!desktopicon,!quicklaunchicon'], {
+			// No live background installer; launch the setup directly with appropriate flags based on config.
+			const installerArgs = [
+				this.isFastUpdatesEnabled() ? '/verysilent' : '/silent',
+				'/log',
+				'/mergetasks=runcode,!desktopicon,!quicklaunchicon'
+			];
+
+			spawn(this.availableUpdate.packagePath, installerArgs, {
```
deepak1556: This path is not meant for verysilent mode, lets not change the setup params as it breaks the control flow.

The other place to create UI is the progress window from inno updater for background updates and that is controlled by the `updateFilePath` presence https://github.com/microsoft/inno-updater/blob/05beec77b463c0ae4f06d50ccb850b44a8eba0ff/src/main.rs#L391-L393. I would first confirm if thats the case with logs before changing this block.

deepak1556 (changes requested):

src/vs/platform/update/electron-main/updateService.win32.ts
dmitrivMS: Sounds good, I'll undo this change.

PR #309133, opened 2026-04-11
Do not open debug view on first session start when openDebug is openOnDebugBreak

### LINKED ISSUE #263434: Debug Sidebar is shown when starting first debug session, even when set to openOnDebugBreak
The default setting for `debug.openDebug` is `openOnDebugBreak` which I interpret to mean only when debugging breaks, however it seems to always be immediately shown when I start the first debug session.

Here, I have just reloaded the window, and the instant I click "Start Debugging" it switches to the debug sidebar.

I do want the debug sidebar to be shown when I hit a breakpoint, but I don't want it to be shown the first time I launch a debug session (that's what I assume the `openOnFirstSessionStart` does).

https://github.com/user-attachments/assets/c8792163-daa2-423d-b0a3-0ddd636bc5ae

yogeshwaran-c:
## Summary

Fixes #263434.

The default `debug.openDebug` value is `openOnDebugBreak`, which is documented as opening the debug view when the debugger breaks. However, the view also opened on the very first debug session of a window, regardless of whether a breakpoint was hit.

The previous condition in `DebugService.launchOrAttachToSession` was:

```ts
if (!configuration.resolved.noDebug && (openDebug === 'openOnSessionStart' || (openDebug !== 'neverOpen' && this.viewModel.firstSessionStart)) && !session.suppressDebugView) {
```

The `openDebug !== 'neverOpen'` check caused `openOnDebugBreak` (and any other value) to fall into the first-session-start branch. The fix restricts that branch to the explicit `openOnFirstSessionStart` value, which is what users expect when reading the setting names.

`openOnDebugBreak` is still handled separately in `debugSession.ts` when a thread stops on a `breakpoint` reason, so behavior on actual breaks is unchanged.

## Test plan

- [ ] Set `debug.openDebug` to `openOnDebugBreak` (the default), reload the window, and start a debug session that does not hit a breakpoint immediately. The debug view should not auto-open.
- [ ] With the same setting, hit a breakpoint. The debug view should still auto-open.
- [ ] Set `debug.openDebug` to `openOnFirstSessionStart` and start a debug session in a fresh window. The debug view should open.
- [ ] Set `debug.openDebug` to `openOnSessionStart` and start any debug session. The debug view should open.
- [ ] Set `debug.openDebug` to `neverOpen`. The debug view should never auto-open.

### DIFF
--- a/src/vs/workbench/contrib/debug/browser/debugService.ts
+++ b/src/vs/workbench/contrib/debug/browser/debugService.ts
@@ -641,8 +641,9 @@ export class DebugService implements IDebugService {
 		this._onWillNewSession.fire(session);
 
 		const openDebug = this.configurationService.getValue<IDebugConfiguration>('debug').openDebug;
-		// Open debug viewlet based on the visibility of the side bar and openDebug setting. Do not open for 'run without debug'
-		if (!configuration.resolved.noDebug && (openDebug === 'openOnSessionStart' || (openDebug !== 'neverOpen' && this.viewModel.firstSessionStart)) && !session.suppressDebugView) {
+		// Open debug viewlet based on the visibility of the side bar and openDebug setting. Do not open for 'run without debug'.
+		// Note: 'openOnDebugBreak' is intentionally excluded here - that case is handled in debugSession when a breakpoint is hit.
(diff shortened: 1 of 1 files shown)


connor4312 (approved):

roblourens (approved):

DanTup:
@yogeshwaran-c thank you! 🙏

PR #309134, opened 2026-04-11
git: clarify that addAICoAuthor only applies to UI commits

### LINKED ISSUE #297415: should git.addAICoAuthor setting specify which commits it supports?
Testing #297204

More of a nitpick but I don't often commit using the UI, I run it locally. Since we don't support something like `prepare-commit-msg` hook to automatically inject the co-author in CLI case, should we mention specifically that this only applies to the UI commit?

I get that it's already namespaced to to `git.*` though, so not sure if needed :)

yogeshwaran-c:
## Summary

The `git.addAICoAuthor` setting only takes effect for commits made from the VS Code Source Control view; commits made from the command line are not affected because the extension does not install a `prepare-commit-msg` hook. The current setting description does not mention this, which can mislead users into expecting CLI commits to gain a `Co-authored-by` trailer too.

This PR updates the setting description to make the scope explicit.

Fixes #297415

## Test plan

- [x] `git.addAICoAuthor` description in Settings UI shows the new wording explaining it only applies to commits made from the Source Control view
- [x] No code paths are affected; this is a string-only change in `extensions/git/package.nls.json`


dmitrivMS:
We are going to rework this feature, so closing this for now.

PR #309137, opened 2026-04-11
fix(search): only show Notebook Find Filters toggle for visible notebooks

### LINKED ISSUE #224043: "Notebook Find Filters" button shows in search when no notebook is active
Besides having a hard time to understand what it does, it imho should not show when no notebook is actively opened?

Image

yogeshwaran-c:
## Summary

Fixes #224043

The Notebook Find Filters toggle in the Search view appeared whenever any notebook editor was present in the list of opened editors, even if no notebook was actually visible on screen. With tabs disabled this is easy to hit: a notebook can stay opened (still in the editor stack) without being the active editor in any group, and the toggle would still show even though there is no visible notebook to filter against.

This change:

- Switches `_hasNotebookOpen` to use `editorService.visibleEditors` so we only consider notebooks that are currently active in some editor group.
- Listens to `onDidVisibleEditorsChange` instead of `onDidEditorsChange` (filtered by `EDITOR_OPEN`/`EDITOR_CLOSE`), so the toggle reacts to a notebook becoming visible/hidden, not just to it being added/removed from the list of opened editors.

The result is that the Notebook Find Filters button only appears when a notebook editor is actually visible.

## Test plan

- [ ] Open a workspace with no notebooks. Open the Search view. Confirm the Notebook Find Filters toggle is hidden.
- [ ] Open a `.ipynb` file. Confirm the toggle appears in the Search view.
- [ ] With the notebook visible, switch the editor group to a different (non-notebook) file so no notebook is visible. Confirm the toggle disappears.
- [ ] Disable tabs (`workbench.editor.showTabs: 'none'`), open a notebook, then open a non-notebook editor on top. Confirm the toggle disappears even though the notebook is still in the editor stack.
- [ ] Make the notebook visible again (e.g. via Quick Open). Confirm the toggle reappears.

### DIFF
--- a/src/vs/workbench/contrib/search/browser/searchWidget.ts
+++ b/src/vs/workbench/contrib/search/browser/searchWidget.ts
@@ -40,7 +40,6 @@ import { NotebookFindFilters } from '../../notebook/browser/contrib/find/findFil
 import { IInstantiationService } from '../../../../platform/instantiation/common/instantiation.js';
 import { IEditorService } from '../../../services/editor/common/editorService.js';
 import { NotebookEditorInput } from '../../notebook/common/notebookEditorInput.js';
-import { GroupModelChangeKind } from '../../../common/editor.js';
 import { SearchFindInput } from './searchFindInput.js';
 import { getDefaultHoverDelegate } from '../../../../base/browser/ui/hover/hoverDelegateFactory.js';
 import { IDisposable, MutableDisposable } from '../../../../base/common/lifecycle.js';
@@ -222,10 +221,8 @@ export class SearchWidget extends Widget {
 					this.searchInput.updateFilterStyles();
 				}
 			}));
-		this._register(this.editorService.onDidEditorsChange((e) => {
-			if (this.searchInput &&
-				e.event.editor instanceof NotebookEditorInput &&
-				(e.event.kind === GroupModelChangeKind.EDITOR_OPEN || e.event.kind === GroupModelChangeKind.EDITOR_CLOSE)) {
+		this._register(this.editorService.onDidVisibleEditorsChange(() => {
+			if (this.searchInput) {
 				this.searchInput.filterVisible = this._hasNotebookOpen();
 			}
 		}));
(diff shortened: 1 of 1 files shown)


src/vs/workbench/contrib/search/browser/searchWidget.ts
```diff
@@ -222,10 +221,8 @@ export class SearchWidget extends Widget {
 					this.searchInput.updateFilterStyles();
 				}
 			}));
-		this._register(this.editorService.onDidEditorsChange((e) => {
-			if (this.searchInput &&
-				e.event.editor instanceof NotebookEditorInput &&
-				(e.event.kind === GroupModelChangeKind.EDITOR_OPEN || e.event.kind === GroupModelChangeKind.EDITOR_CLOSE)) {
+		this._register(this.editorService.onDidVisibleEditorsChange(() => {
```
osortega, line 224: Why removing the check here? Shouldn't the `hasNotebookOpen` change do the trick?

yogeshwaran-c: Good question. The two conditions that were removed (`e.event.editor instanceof NotebookEditorInput` and `e.event.kind === EDITOR_OPEN || EDITOR_CLOSE`) were filters specific to `onDidEditorsChange`, which is an aggregated, high-frequency event carrying an `IEditorsChangeEvent` payload (`.event.editor`, `.event.kind`). I switched the listener to `onDidVisibleEditorsChange`, which is an `Event<void>` and has no such payload — so there is literally nothing to inspect here; `e.event.editor`/`e.event.kind` no longer exist on this event.

More importantly, the old `kind === EDITOR_OPEN/EDITOR_CLOSE` filter is the source of the bug: a notebook can become visible or hidden without being opened or closed (switching the active editor within a group, or splitting), and those transitions never fire OPEN/CLOSE, so the toggle's visibility went stale. By listening to `onDidVisibleEditorsChange` and recomputing `_hasNotebookOpen()` (now reading `visibleEditors` instead of all `editors`) on every visibility change, the toggle stays correct for exactly the cases we care about.

So the `_hasNotebookOpen` change alone isn't sufficient — it answers "is a notebook currently visible?", but we still need the right event to *re-evaluate* it on, and `onDidVisibleEditorsChange` is that event. The removed checks can't be carried over because they belong to a different event shape, and they encoded the open/close-only assumption we're intentionally dropping.

PR #309140, opened 2026-04-11
fix: treat '.' as current folder in multi-root search includes/excludes

### LINKED ISSUE #263244: "Workspace folder does not exist: ." if trailing "." in "files to exclude"
Type: <b>Bug</b>

Initially I thought search was not working at all, but it seems to only be a misleading error message.
I mistakenly had left a trailing period in the files to exclude field when searching within my multi-root project.

Steps:
- Open search
- enter a "." in the "files to exclude" field
- type anything into the search field and you'll see the message "Workspace folder does not exist: ."

Expected result:
- I would expect exclusion to ignore anything it wasn't able to find. This project contains multiple roots, so even if the current path (./) was not one of the roots (it currently is), that does not mean the search couldn't find and exclude files in the other roots.

Thanks for your attention!

VS Code version: Code 1.103.2 (6f17636121051a53c88d3e605c491d22af2ba755, 2025-08-20T16:45:34.255Z)
OS version: Windows_NT x64 10.0.19045
Modes:

<details>

|Item|Value|
|---|---|
|CPUs|Intel(R) Xeon(R) W-2225 CPU @ 4.10GHz (8 x 4104)|
|GPU Status|2d_canvas: enabled<br>direct_rendering_display_compositor: disabled_off_ok<br>gpu_compositing: enabled<br>multiple_raster_threads: enabled_on<br>opengl: enabled_on<br>rasterization: enabled<br>raw_draw: disabled_off_ok<br>skia_graphite: disabled_off<br>trees_in_viz: disabled_off<br>video_decode: enabled<br>video_encode: enabled<br>vulkan: disabled_off<br>webgl: enabled<br>webgl2: enabled<br>webgpu: enabled<br>webnn: disabled_off|
|Load (avg)|undefined|
|Memory (System)|63.69GB (48.68GB free)|
|Process Argv|Y:\\irenec\\vscode\\irenec.code-workspace --crash-reporter-id 0f3862e7-250e-488d-8961-f1c20f56d1d7|
|Screen Reader|no|
|VM|0%|
</details><details><summary>Extensions (10)</summary>

---|---|---
extendscript-debug|Ado|2.1.0
copilot|Git|1.350.0
copilot-chat|Git|0.30.3
vscode-pull-request-github|Git|0.116.1
auto-comment-blocks|kev|1.0.1
debugpy|ms-|2025.10.0
python|ms-|2025.12.0
vscode-pylance|ms-|2025.7.1
vscode-python-envs|ms-|1.2.0
autodocstring|njp|0.6.1

</details><details>

```
vsliv368:30146709
binariesv615:30325510
nativeloc1:31344060
dwcopilot:31170013
dwoutputs:31242946
copilot_t_ci:31333650
e5gg6876:31282496
pythoneinst12:31285622
996jf627:31283433
pythonrdcb7:31342333
usemplatestapi:31297334
747dc170:31275177
aj953862:31281341
9d2cg352:31346308
nesew2to5:31336538
agentclaude:31335814
nes-set-on:31340697
63221493:31336333
0927b901:31350571
f76d9909:31348711
custommodelcf:31371783
45650338:31358607
0cj2b977:31352657
justeven_python:31371804
gemagent1:31368469
retryenabled:31370050

```

</details>

yogeshwaran-c:
## Summary
Fixes #263244.

When a user enters just `.` in the **files to include** or **files to exclude** field of the search view in a multi-root workspace, the query builder threw:

> Workspace folder does not exist: .

instead of treating the dot as the current workspace folder. The existing code already short-circuits `./` and `.\` for the same reason, so this extends that handling to a bare `.`.

## Details
`QueryBuilder.parseSearchPaths` classifies any segment matching `^\.\.?([\/\]|$)` as a search path, which means a bare `.` is forwarded to `expandOneSearchPath`. In multi-root mode, that method only special-cased `./` and `.\`, so `.` fell through to the workspace-folder-name lookup and threw the misleading error.

The fix simply adds `.` to the existing single-folder-only short-circuit, matching the comment in the surrounding code.

## Test plan
- [x] Added two new cases (`.` and `.\`) to the existing `relative includes w/multiple ambiguous root folders` test in `queryBuilder.test.ts`, expecting `searchPaths: undefined` to mirror the existing `./` case.
- [ ] Manually: open a multi-root workspace, open the search view, enter `.` in **files to exclude**, type a query and confirm no error appears and results are returned.

### DIFF
--- a/src/vs/workbench/services/search/common/queryBuilder.ts
+++ b/src/vs/workbench/services/search/common/queryBuilder.ts
@@ -502,8 +502,8 @@ export class QueryBuilder {
 				searchPath: workspaceUri,
 				pattern: cleanedPattern
 			}];
-		} else if (searchPath === './' || searchPath === '.\\') {
-			return []; // ./ or ./**/foo makes sense for single-folder but not multi-folder workspaces
+		} else if (searchPath === '.' || searchPath === './' || searchPath === '.\\') {
+			return []; // ., ./ and .\ refer to the current workspace folder which only makes sense in a single-folder workspace
 		} else {
 			const searchPathWithoutDotSlash = searchPath.replace(/^\.[\/\\]/, '');
 			const folders = this.workspaceContextService.getWorkspace().folders;
--- a/src/vs/workbench/services/search/test/browser/queryBuilder.test.ts
+++ b/src/vs/workbench/services/search/test/browser/queryBuilder.test.ts
@@ -878,6 +878,18 @@ suite('QueryBuilder', () => {
 					}
 				],
 				[
+					'.',
+					{
+						searchPaths: undefined
+					}
+				],
+				[
+					'.\\',
+					{
+						searchPaths: undefined
+					}
+				],
+				[
 					'./root1',
 					{
 						searchPaths: [{


osortega:
What is the need for this? Shouldn't in that case the include/exclude be left empty?

yogeshwaran-c:
@osortega Good question — the goal isn't really about treating `.` as a useful filter, it's about consistency with the existing `./` / `.\` handling and replacing a misleading error.

Today, in a multi-root workspace, `./` and `.\` already short-circuit to `searchPaths: undefined` (see the existing branch at `queryBuilder.ts:505` with the comment *"./ or ./**/foo makes sense for single-folder but not multi-folder workspaces"*) — i.e. they're effectively treated as "no search path", which falls back to searching all roots. A bare `.` falls through that branch and hits `expandOneSearchPath`'s folder-name lookup, which throws `Workspace folder does not exist: .`.

So the user-visible bug isn't "`.` should exclude something" — it's that searching with `.` in includes/excludes pops up an error from `onSearchError` and the search stops, instead of either being ignored (like `./`) or returning empty results. That's what #263244 reports.

I went with "treat `.` like `./`" because (a) the regex at `parseSearchPaths` (`/^\.\.?([\/\\]|$)/`) already classifies bare `.` as a search path, so by the time we're in `expandOneSearchPath` the empty-field path isn't an option, and (b) it keeps the three forms (`.`, `./`, `.\`) behaviorally identical, which I think is what most users would expect since CLI conventions treat them equivalently.

Happy to switch to a different approach if you'd prefer — e.g. having `parseSearchPaths` reject bare `.` so it's treated as a literal pattern instead, or surfacing a friendlier validation message in the search UI rather than an error toast. Just let me know which direction you'd like.

PR #309142, opened 2026-04-11
Suppress CSS completions inside comments (#236215)

### LINKED ISSUE #236215: Quick suggestions showing in CSS comments when configured to be disabled
Repro:

Settings:

```js
  "editor.quickSuggestions": {
    "comments": "off"
  },
```

Go to a css file comment and add `:` to the end of a line, 🐛 quick suggestions trigger.

@aeschli could reproduce

Image

yogeshwaran-c:
## Summary

Fixes #236215.

When the cursor is inside a CSS, SCSS or LESS comment, the language server still answered completion requests, so typing `:` in a comment surfaced suggestions like `:after`. The CSS server registers `:` as a completion trigger character, which means trigger-character requests bypass `editor.quickSuggestions.comments: off`.

This adds a small `isInsideComment` helper in `extensions/css-language-features/server/src/utils/comments.ts` that performs a single forward scan tracking string and comment state, so:

- Block comments (`/* ... */`) are detected for CSS, SCSS and LESS.
- Line comments (`// ...`) are detected for SCSS and LESS only.
- Comment delimiters inside string literals (e.g. `content: "/*"`) are not mistaken for real comments.
- Backslash escapes inside strings are honoured.

`onCompletion` in `cssServer.ts` short-circuits and returns `null` when the position falls inside a comment.

## Test plan

- [x] New unit tests in `extensions/css-language-features/server/src/test/comments.test.ts` cover: plain code, inside/after closed block comment, unterminated block comment, the issue #236215 repro (colon at end of a line inside a multi-line block comment), block delimiter inside a string literal, and SCSS line comment cases (with and without `supportsLineComments`).
- [x] All existing tests in the css-language-features server suite continue to pass.

### DIFF
--- a/extensions/css-language-features/server/src/cssServer.ts
+++ b/extensions/css-language-features/server/src/cssServer.ts
@@ -12,6 +12,7 @@ import { getLanguageModelCache } from './languageModelCache';
 import { runSafeAsync } from './utils/runner';
 import { DiagnosticsSupport, registerDiagnosticsPullSupport, registerDiagnosticsPushSupport } from './utils/validation';
 import { getDocumentContext } from './utils/documentContext';
+import { isInsideComment } from './utils/comments';
 import { fetchDataProviders } from './customData';
 import { RequestService, getRequestService } from './requests';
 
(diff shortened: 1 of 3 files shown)


aeschli:
Rich CSS/LESS/SCSS support is implemented in https://github.com/microsoft/vscode-css-languageservice and is backed by a parser/scanner that knows about comments.
The fix would have to be there and I should be using the scanner.
See https://github.com/microsoft/vscode/issues/236215#issuecomment-2893671279 where I already outlined that.

Issue #308028, opened 2026-04-06
Agents: Reuse VS Code auth


bamurtaugh:
Reusing the same auth from VS Code in Agents (so that users don't have to re-sign in) would help make getting started with the Agents app even smoother.

PR #309154, opened 2026-04-11
feat: cross-app secret migration via crossAppIPC (macOS)

alexdima:
Ref https://github.com/microsoft/vscode/issues/308028

Introduce demand-driven one-time secret migration from VS Code to the agents app using Electron's `crossAppIPC` (Mach port based, code-signature verified).

## How it works

When the agents app starts and detects that `github.auth` has not been migrated yet, it:
1. Creates a `crossAppIPC` connection and starts listening.
2. Spawns Code.app with `--share-secrets-with-agents-app`.
3. Code.app connects, receives a `SecretRequest`, reads the secret from its safeStorage, and sends it back in a `SecretResponse`.
4. The agents app stores the secret in its own safeStorage and sends a `SecretAck` so both sides record the migration as done.
5. If Code.app was launched solely for this purpose, it quits.

If Code.app is already running, the second instance forwards the arg via the node IPC socket to the running instance's `LaunchMainService`.

## Security

- `crossAppIPC` uses **code-signature verification** (Mach ports on macOS) — the kernel authenticates both endpoints
- No secrets are ever in process args, temp files, or network
- Only the specific keys in `CROSS_APP_SHARED_SECRET_KEYS` are shared

## Code sharing

Extracts shared utilities (`readEncryptedSecret`, `writeEncryptedSecret`, `secretStorageKey`, `SECRET_STORAGE_PREFIX`) from `BaseSecretStorageService` into reusable functions in `secrets.ts` to avoid duplication between the renderer-side secret service and the main-process secret sharing.

## State diagram

| Scenario | What happens |
|----------|-------------|
| Agents starts, Code NOT running | Agents spawns Code.app → migration handshake → Code quits |
| Agents starts, Code already running | Agents spawns Code.app → second instance forwards to running instance → migration handshake |
| Already migrated | Both sides check flag → skip entirely, zero cost |


dmitrivMS (approved):

bpasero (commented):
I wonder if `share-secrets-with-agents-app` should better be handled in our `cli.ts` that we know is the entry point for CLI invocations, at least when launched with the `code` / `code-insiders` script? Or is there dependencies that are not met there yet?

PR #309158, opened 2026-04-11
fix: defer model disposal until after input state flush when switching chat threads (#309115)

### LINKED ISSUE #309115: Agents: Regression in chat input prompt being saved

Seeing an issue with latest Insiders today where if I'm typing out a prompt into one chat thread (but do not send) and then quickly jump to another thread, once I come back the prompt I was in the middle of typing is gone.

Steps to Reproduce:

1. Open an existing chat thread
2. Type something into the input text field
3. Open another chat
4. Navigate back to original chat and see that the prompt was not saved / is gone

Version: 1.116.0-insider
Commit: 47add6dfa59e9d89115b53c74511f03db9575dd0
Date: 2026-04-10T20:25:41+09:00
Electron: 39.8.7
ElectronBuildId: 13797146
Chromium: 142.0.7444.265
Node.js: 22.22.1
V8: 14.2.231.22-electron.0
OS: Darwin arm64 25.4.0

maruthang:
## Summary

Fixes #309115

**Bug:** Chat input prompt text was not saved when switching between threads, causing the user's in-progress prompt to be lost.

**Root Cause:** In `ChatViewPane.showModel()`, the old model reference was disposed immediately via `this.modelRef.value = undefined`, which triggered model serialization before the widget's `setModel` call could flush the current input state to the outgoing model.

**Fix:** Use `clearAndLeak()` to extract the old reference without disposing it, then manually dispose it after `setModel` has been called (which triggers the input state flush). Early-return and error paths also properly dispose the old reference to prevent leaks.

## Changes

- `src/vs/workbench/contrib/chat/browser/widgetHosts/viewPane/chatViewPane.ts`: Replace immediate `this.modelRef.value = undefined` with `clearAndLeak()` pattern that defers disposal until after `setModel` flushes input state. Added proper disposal in all early-return and error paths.
- `src/vs/workbench/contrib/chat/test/common/model/chatModelStore.test.ts`: Added 2 regression tests validating the disposal ordering behavior for issue #309115.

## Testing

- Added regression test verifying model stays alive when reference is leaked and disposed after state flush
- Added regression test demonstrating the problematic early-release pattern that caused the bug
- All existing tests pass

### DIFF
--- a/src/vs/workbench/contrib/chat/browser/widgetHosts/viewPane/chatViewPane.ts
+++ b/src/vs/workbench/contrib/chat/browser/widgetHosts/viewPane/chatViewPane.ts
@@ -704,7 +704,13 @@ export class ChatViewPane extends ViewPane implements IViewWelcomeDelegate {
 
 	private async showModel(token: CancellationToken, modelRef?: IChatModelReference | undefined, startNewSession = true): Promise<IChatModel | undefined> {
 		const oldModelResource = this.modelRef.value?.object.sessionResource;
-		this.modelRef.value = undefined;
+
+		// Hold a temporary reference to the old model so it stays alive until
+		// the widget has flushed unsaved input state (e.g. draft text the user
+		// was typing). Releasing the reference too early can cause the model to
+		// be disposed and serialized before the input state is saved, losing the
+		// user's in-progress prompt. See #309115.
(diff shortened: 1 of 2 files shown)


roblourens:
Thanks for the PR, but it looks like someone else got to this in the meantime

PR #309165, opened 2026-04-11
fix: terminal canvas misalignment on zoom level change

### LINKED ISSUE #309160: Terminal: Missing devicePixelRatio listener causes misalignment on zoom

- VS Code Version: 1.115.0
- OS Version: Ubuntu 24.04.4 LTS

Steps to Reproduce:

1. Open VS Code and open the integrated terminal in the panel.
2. Change the editor's zoom level using the keyboard shortcuts (`Ctrl` + `+` or `Ctrl` + `-`).
3. **Observe the bug:** The terminal canvas fails to repaint and align correctly with the new zoom dimensions. Depending on the zoom direction, the prompt text gets severely clipped at the top, disappears entirely out of the viewport bounds, or renders with excessive blank space above it. The inline suggestion ghost text also detaches and floats in the wrong position.
4. Switch to a different panel tab (e.g., "PORTS" or "OUTPUT") and switch back to "TERMINAL".
5. **Observe the workaround:** Forcing this tab switch triggers a re-render, causing the terminal canvas to properly recalculate its dimensions and fix the alignment.

### Testing on `main` build
I also tested existence of this issue on my fork by building on `main` , issue still persists.

**Media:**

https://github.com/user-attachments/assets/65040f0f-a5e0-48d2-a6b4-102141965f5c

### PR Ready
**Note to maintainers:** I have already investigated the root cause of this issue (the terminal instance is missing a listener for `window.devicePixelRatio` / zoom level changes, causing it to use stale dimensions for the canvas renderer until a manual layout refresh is triggered). **I have a Pull Request ready to go that implements this fix and will link it to this issue immediately after opening.** I would love to contribute to vscode given the opportunity!

ammaarrahmed:
## Summary

When the user changes the editor zoom level (Ctrl++/Ctrl+-), the integrated terminal canvas fails to repaint and align correctly. The prompt text gets clipped, disappears, or renders with excess blank space. Inline suggestion ghost text also misaligns.

## Root Cause

The terminal's dimension calculations (`getXtermScaledDimensions`, `_measureFont`) use `window.devicePixelRatio` for canvas rendering and font measurements, but no listener existed to trigger a layout recalculation when the pixel ratio changed due to zoom. Switching tabs worked around the issue because `setVisible(true)` calls `_resize()`, which recalculates dimensions with the now-current DPR.

## Fix

Two event listeners are added in the `TerminalInstance` constructor:

- **`PixelRatio.onDidChange`** — fires when `devicePixelRatio` changes (e.g. browser zoom, display scale)
- **`onDidChangeZoomLevel`** — fires from Electron's zoom API (e.g. `Ctrl++`/`Ctrl+-`)

Both trigger a shared handler that defers (50 ms for DOM settling) a full layout recalculation:

1. Marks font metrics as stale (`_layoutSettingsChanged = true`)
2. Clears the WebGL/canvas texture atlas (`forceRedraw()`)
3. Runs the full `layout()` path to recalculate cols/rows

## Precedent

This matches the established pattern used by other VS Code components that already subscribe to `PixelRatio.onDidChange`:

- Editor (`editorDom.ts`)
- Color picker widget
- Debug toolbar
- Breadcrumbs widget
- Notebook cells

## Demo

[Screencast from 2026-04-11 10-28-43.webm](https://github.com/user-attachments/assets/1c3a4306-6c48-4a01-b6ee-abf5ade0e791)

## Testing Checklist

- [x] `Ctrl++` / `Ctrl+-` — terminal repaints correctly, no clipping or blank space
- [x] Tab switching after zoom — still works as before
- [x] Split terminals — both panes recalculate correctly
- [x] Type-check passes (`tsgo`)
- [x] Layering check passes (`valid-layers-check`)
- [x] ESLint passes
- [x] Hygiene passes

## How to Test

1. Open an integrated terminal
2. Type some text so the prompt is visible
3. Press `Ctrl++` or `Ctrl+-` to zoom in/out
4. Observe the terminal canvas repaints cleanly without clipping, blank space, or misaligned ghost text

Fixes microsoft/vscode#309160

### DIFF
--- a/src/vs/workbench/contrib/terminal/browser/terminalInstance.ts
+++ b/src/vs/workbench/contrib/terminal/browser/terminalInstance.ts
@@ -3,14 +3,15 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-import { isFirefox } from '../../../../base/browser/browser.js';
+import { isFirefox, onDidChangeZoomLevel } from '../../../../base/browser/browser.js';
+import { PixelRatio } from '../../../../base/browser/pixelRatio.js';
 import { BrowserFeatures } from '../../../../base/browser/canIUse.js';
 import { DataTransfers } from '../../../../base/browser/dnd.js';
 import * as dom from '../../../../base/browser/dom.js';
 import { StandardKeyboardEvent } from '../../../../base/browser/keyboardEvent.js';
 import { Orientation } from '../../../../base/browser/ui/sash/sash.js';
 import { DomScrollableElement } from '../../../../base/browser/ui/scrollbar/scrollableElement.js';
-import { AutoOpenBarrier, Promises, disposableTimeout, timeout } from '../../../../base/common/async.js';
+import { AutoOpenBarrier, Promises, RunOnceScheduler, disposableTimeout, timeout } from '../../../../base/common/async.js';
 import { Codicon } from '../../../../base/common/codicons.js';
 import { debounce } from '../../../../base/common/decorators.js';
 import { BugIndicatingError, onUnexpectedError } from '../../../../base/common/errors.js';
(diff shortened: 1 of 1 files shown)


ammaarrahmed:

PR #309167, opened 2026-04-11
await openEditor in terminal editor split to prevent shadow tab

jamestut:
When splitting a terminal that's in the editor area (via the "Split Terminal" command), the new terminal correctly opens in a new editor group to the side, but a duplicate "shadow tab" also appears in the original editor group.

fixes #291578 

# Root Cause

`TerminalEditorService.splitInstance` called `this._editorService.openEditor(...)` with `SIDE_GROUP` but **did not await it**. The call was fire-and-forget.

After `createTerminal` returned, the Split action called `focusActiveTerminal` → `focusInstance` → `_revealEditor`, which called `openEditor` again **without specifying a target group**. Since the first `openEditor` hadn't completed yet, the editor wasn't in any group, so `findGroup` defaulted to the active group — the original group containing the parent terminal. This created a duplicate tab.

This regression was introduced by the new `_revealEditor` logic in `focusInstance` (added after `1.106.0`). In `1.106.0`, `focusInstance` was just `instance.focusWhenReady(true)` with no `openEditor` call, so the race couldn't occur.

# Fix

Await the `openEditor` in `splitInstance` so the new terminal is fully opened in the correct side group before `focusInstance` runs.

# Testing

1. Open a new terminal in the editor area.
2. Run the **Split Terminal** command.

Previously, a shadow terminal tab would appear in the original editor group. After the fix, it no longer does so.


meganrogge (approved):
Thank you!

roblourens (approved):

PR #309174, opened 2026-04-11
html: fix language model cache evicting at capacity instead of overflow

yogeshwaran-c:
The `getLanguageModelCache` function in the HTML language server uses `===` to check whether the current entry count equals `maxEntries`. This causes the cache to evict the oldest entry as soon as it reaches its configured maximum size, meaning the cache only holds `maxEntries - 1` entries at steady state instead of the intended `maxEntries`.

For example, with `maxEntries = 10` (the value used throughout the HTML server), only 9 parsed documents are ever cached simultaneously. The 10th entry immediately triggers eviction, bringing the count back to 9 before any subsequent access.

**Fix:** Change the condition from `nModels === maxEntries` to `nModels > maxEntries`. Eviction now only occurs when the cache has grown beyond its limit, allowing up to `maxEntries` entries to be retained.

**Before:**
```typescript
if (nModels === maxEntries) {
```

**After:**
```typescript
if (nModels > maxEntries) {
```

Affected caches in the HTML language server (all with `maxEntries = 10`):
- `documentRegions` in `languageModes.ts`
- `embeddedCSSDocuments` and `cssStylesheets` in `cssMode.ts`
- `htmlDocuments` in `htmlMode.ts`
- `jsDocuments` in `javascriptMode.ts`


aeschli (approved):

dmitrivMS (approved):

aeschli:
Thinking about it again, we don't need this change.It is not relevant whether we cache `maxEntries` entries or `maxEntries - 1`. Feel free to rename the variable from maxEntries to limit.

PR #309175, opened 2026-04-11
css: fix language model cache evicting at capacity instead of overflow

yogeshwaran-c:
The `getLanguageModelCache` function in the CSS language server uses `===` to check whether the current entry count equals `maxEntries`. This causes the cache to evict the oldest entry as soon as it reaches its configured maximum size, so the cache holds only `maxEntries - 1` entries at steady state instead of the intended `maxEntries`.

The same issue was also present in the HTML language server (see companion fix in separate PR).

**Fix:** Change the condition from `nModels === maxEntries` to `nModels > maxEntries`. Eviction now only occurs after the cache has grown beyond its limit.

**Before:**
```typescript
if (nModels === maxEntries) {
```

**After:**
```typescript
if (nModels > maxEntries) {
```

With `maxEntries = 10` (the value used in the CSS language server), this means the cache can now properly retain up to 10 parsed stylesheet models instead of 9.


aeschli (approved):

dmitrivMS (approved):

aeschli:
Thinking about it again, we don't need this change.It is not relevant whether we cache `maxEntries` entries or `maxEntries - 1`. Feel free to rename the variable from maxEntries to limit.

PR #309182, opened 2026-04-11
fix(themes): sync preferred theme settings when workbench.colorTheme changes

yogeshwaran-c:
Upstream fix for microsoft/vscode#305685

## Problem

Using **Toggle between Light/Dark Themes** resets the theme to VS Code defaults (`VSCode Dark` / `VSCode Light`) instead of returning to the user's previously-selected theme.

## Root Cause

The toggle command reads `workbench.preferredDarkColorTheme` / `workbench.preferredLightColorTheme` to decide which theme to apply. If the user never explicitly configured those settings they contain the default value, so the toggle always switches to the default rather than the theme the user was using.

## Fix

In `installConfigurationListener`, when `workbench.colorTheme` changes and auto-detect is not active, keep the matching preferred-theme setting in sync:

- If the new theme is **dark**, update `workbench.preferredDarkColorTheme` to match.
- If the new theme is **light**, update `workbench.preferredLightColorTheme` to match.

This ensures the toggle always returns to the user's most recently used dark or light theme. The guard `current !== newSettingId` prevents redundant writes (e.g. when the toggle itself sets `workbench.colorTheme`).

### DIFF
--- a/src/vs/workbench/services/themes/browser/workbenchThemeService.ts
+++ b/src/vs/workbench/services/themes/browser/workbenchThemeService.ts
@@ -349,6 +349,25 @@ export class WorkbenchThemeService extends Disposable implements IWorkbenchTheme
 			) {
 				this.restoreColorTheme();
 			}
+			// Keep the preferred theme settings in sync with workbench.colorTheme so that
+			// 'Toggle between Light/Dark Themes' returns to the user's most recently used theme.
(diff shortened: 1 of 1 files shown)


aeschli:
I don't think thats a good change as this changes the preferredDark/LightColorTheme without the users consent.
Only very few users use 'Toggle between Light/Dark Themes'. If you want to fix anything, only fix it in that command.

Issue #309284, opened 2026-04-12
Agents: thread can disappear immediately while background agent keeps running and editing the workspace


thatdaveguy1:
## Summary

In **VS Code Insiders** with **GitHub Copilot Chat / Agents**, a newly-created agent thread can disappear from the visible Agents/Sessions UI almost immediately, while the background worker keeps running and continues making real workspace changes.

This is not a "task stopped" problem. The background work continues; the **UI-visible thread is what disappears**.

## Expected

- The thread should stay visible while the agent is running.
- If persistence partially fails, the thread should remain **recoverable** in the Sessions list instead of silently vanishing.
- A persistence/index failure should either block background launch or surface a visible broken/recoverable session state.

## Actual

- The visible agent thread disappears almost immediately after creation.
- The agent still appears to continue in the background and makes file edits / workspace changes.
- The user loses the visible conversation/thread even though the task is still executing.

## Repro pattern

This has been intermittent, but the pattern is:

1. Open a normal workspace in **Agents - Insiders**.
2. Start an agent task that performs actual file edits / multi-step work.
3. Very soon after creation, observe that the visible thread disappears from the Agents/Sessions UI.
4. Despite that, the workspace continues changing, indicating the background worker is still alive.

## Sanitized local evidence

I am intentionally **not** attaching raw agent transcripts, prompts, repository contents, or personal file paths because they contain private working context.

What I can share safely:

- The background worker clearly continued after the UI thread disappeared, because workspace files changed after the visible thread was gone.
- On the affected machine, lightweight session/debug records for the visible session path were missing or nearly empty, while a separate background/session-state event stream continued to show activity.
- This strongly suggests a divergence between **runtime/background agent execution** and the **persisted UI session record / index**.

## Likely area

This looks related to chat session persistence / restore / index handling, likely in or around:

- `src/vs/workbench/contrib/chat/common/model/chatSessionStore.ts`
- possibly related session creation / visibility wiring in chat service / controller code

## Why I think this belongs with existing chat-session persistence bugs

This symptom looks closely related to other cases where chat sessions are lost, malformed, or missing due to persistence/index issues.

Related issues / PRs:

- #308369 — Copilot Chat session history intermittently disappears
- #308730 — malformed `chatSessions` JSONL makes sessions unloadable
- #307348 — incomplete in-flight session data makes sessions unloadable
- #296890 — parallel agent writes can corrupt session JSONL
- #305109 — PR fixing session index loss during workspace transitions
[...]

Issue #308369, opened 2026-03-24
Copilot Chat session history intermittently disappears, causing loss of long-running context


Morihami:
## Summary
Copilot Chat sessions intermittently disappear in VS Code. A chat that existed about 2 hours earlier is no longer visible, and long-running writing context is effectively lost.

This is a severe workflow blocker for long-form content creation because continuity depends on prior chat context.

## Repro Steps
1. Use Copilot Chat in VS Code for a long session (multiple turns, over 1-2+ hours).
2. Continue work normally; sometimes reload window / extension updates may occur in the background.
3. Return later to the previous chat session.
4. Observe the session/history is missing intermittently.

## Expected
Previous chat sessions/history should remain available and recoverable reliably.

## Actual
Some sessions disappear intermittently (not user-deleted), and prior context cannot be recovered from chat UI.

## Frequency
Intermittent, but recurring.

## Environment
- OS: Windows 11 Pro 10.0.26200 (Build 26200)
- VS Code: 1.112.0
- Copilot Chat extension: github.copilot-chat-0.40.1

## Impact
High. Loss of working memory/context interrupts ongoing projects and can block continuation of work.

## Notes
Please investigate session persistence / restore logic, especially around reload/update/sign-in transitions.

Issue #307348, opened 2026-04-02
Chat session permanently unloadable after VS Code killed mid-response (TypeError: Cannot read properties of undefined (reading 'response'))


VictorBargains:
## Summary

When VS Code is killed during a streaming chat response, the session .jsonl contains a structurally valid but semantically incomplete entry: a kind:2 (Set) record with 
equestId present but no 
esponse key. On the next launch, ChatSessionStore crashes with TypeError: Cannot read properties of undefined (reading 'response') and permanently discards the entire session.

## Versions

- VS Code: 1.113.0
- Copilot Chat: 0.42.2
- OS: Windows 11 (x64)

## Steps to Reproduce

1. Open a Copilot Chat session in agent mode
2. Send a message and let the response start streaming
3. While the response is in progress, kill VS Code (End Task / power loss)
4. Reopen VS Code and open the same workspace
5. Open Chat History and click the session

**Expected:** Session loads, showing at minimum the turns that completed before the interruption.

**Actual:** Session never loads. The history entry is visible but clicking it stays on the blank new-chat screen.

## Root Cause

chatSessionStore.ts reads the .jsonl via ObjectMutationLog.read(), which replays all entries unconditionally. An interrupted in-progress response leaves the assembled 
equests array containing an entry with 
equestId and message but no 
esponse property.

The revival loop at [chatSessionStore.ts ~L674](https://github.com/microsoft/vscode/blob/main/src/vs/workbench/contrib/chat/common/model/chatSessionStore.ts) then does:

`	ypescript
for (const request of session.requests) {
    if (Array.isArray(request.response)) {        // false (request.response is undefined)
        request.response = request.response.map(.); // throws if reached
    } else if (typeof request.response === 'string') { // false
        .
    }
    // silently falls through - but...
}
`

The Array.isArray(undefined) and 	ypeof undefined === 'string' both return false, so execution falls through without throwing *here*. The actual crash occurs downstream at 
ormalizeSerializableChatData() or during schema traversal in objectMutationLog.ts, which calls 
ethrowWithPathSegment to annotate the path as .requests[N].response: Cannot read properties of undefined.

The entire exception is caught at the outer catch block (~L689):

`	ypescript
} catch (err) {
    this.reportError('malformedSession', Malformed session data in : ., err);
    return undefined;
}
`

This logs the error and returns undefined, making the session permanently unloadable.

**Confirmed via 
enderer.log:**
`
[...]

Issue #296890, opened 2026-02-23
Agent mode: sending message mid-response sometimes forks session into two parallel agents, causing conflicting edits and JSONL corruption


puigru:
- Copilot Chat Extension Version: 0.37.8
- VS Code Version: 1.109.5 (072586267e68ece9a47aa43f8c108e0dcbf44622, x64)
- OS Version: LMDE 6 (faye), Linux 6.1.0-43-amd64 x86_64
- Feature: Agent mode
- Selected model: Claude Opus 4.6
- Logs: N/A (session file evidence attached)

---

### Bug summary

Sending a new message while the agent is mid-response **sometimes** causes the session to bifurcate into two independent agent threads running simultaneously. When it triggers, both agents continue operating on the project in parallel, each with their own "mind", starting from the point in the conversation where the new message was sent. This is likely a **race condition** — it doesn't happen every time, but when it does, it leads to:

1. **Conflicting file edits** — two agents editing the same codebase at once with no coordination
2. **Session file corruption** — both threads write to the same `.jsonl` session file concurrently, interleaving JSON entries inside each other's data

### Steps to reproduce

1. Start an agent mode session with a complex multi-step task
2. While the agent is mid-response (actively running tools, editing files), send a new message to redirect/steer it
3. **Sometimes** (likely depending on timing — what stage the agent's response is in when the new message arrives), instead of the first response being cancelled, both the old and new response threads continue running simultaneously
4. Observe two parallel streams of tool calls and file edits happening at once
5. Close VS Code to stop the chaos
6. The session file is now corrupted: `ChatSessionStore: Malformed session data in .../chatSessions/<id>.json`

**Note:** This doesn't happen every time a message is sent mid-response. It appears to be a race condition — the cancellation of the previous response and the start of the new one are not properly serialized, so depending on timing, the old response may escape cancellation and keep running alongside the new one.

### Expected behavior

Sending a new message while the agent is responding should **always cancel the in-progress response** before starting the new one. Only one agent thread should ever be active at a time within a session.

### Actual behavior

The previous response is sometimes **not cancelled**. A second response thread is spawned while the first is still running. Both run simultaneously, making independent tool calls and file edits. The session effectively "splits in two."

### Evidence from the corrupted `.jsonl` file

Line 137 of the session file had two JSON entries merged into one line — a `{"kind":1,...}` entry was written *inside* a string value of a `{"kind":2,...}` entry:

```
[...]

PR #309286, opened 2026-04-12
chat: fall back to full snapshot on serializer failure

thatdaveguy1:
## Summary

This draft PR hardens `ChatSessionStore` against one concrete missing-session failure mode: when incremental chat-session serialization throws, we currently report the error and then abort the write, which can leave the session non-durable and effectively lost.

Instead of bailing out, this change falls back to writing a fresh full snapshot of the current session and then continues updating the session index.

## What changed

- In `chatSessionStore.ts`, when `session.dataSerializer.write(session)` throws, VS Code now:
  - keeps the existing user-facing error prompt
  - logs a warning
  - resets the serializer
  - writes a full replacement snapshot via `createInitial(session)` so the fallback matches the normal persistence schema
- Added a regression test proving that a session is still persisted and readable when incremental serialization fails.
- Stubbed `IDialogService.prompt` and `IOpenerService.open` in the regression test so the failure-path test is deterministic and side-effect free.

## Why this helps

If the append-log serializer is the thing that breaks, we no longer silently lose durability for the session. That should reduce cases where agent/chat work continues but the session is no longer recoverable through the normal session store path.

## Scope / limitations

This is intentionally a **small hardening fix**, not a claim that it explains every disappearing-session report. It addresses one real failure path in the existing code, especially relevant because the surrounding comment already calls out "missing sessions".

## Related

- Related to #309284
- Related to #308369
- Related to #308730
- Related to #307348
- Related to #296890
- Related to #305109
- Related to #308921

If maintainers think the root cause for #309284 is elsewhere, this should still be useful as a targeted durability improvement in the same subsystem.

### DIFF
--- a/src/vs/workbench/contrib/chat/common/model/chatSessionStore.ts
+++ b/src/vs/workbench/contrib/chat/common/model/chatSessionStore.ts
@@ -371,14 +371,17 @@ export class ChatSessionStore extends Disposable {
 							this.dialogService.prompt({
 								custom: true, // so text is copyable
 								title: localize('chatSessionStore.serializationError', 'Error saving chat session'),
-								message: localize('chatSessionStore.writeError', 'Error serializing chat session for storage. The session will be lost if the window is closed. Please report this issue to the VS Code team:\n\n{0}', e.stack || toErrorMessage(e)),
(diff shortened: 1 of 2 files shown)


thatdaveguy1:

thatdaveguy1:
Addressed the Copilot review feedback in the latest push:

- The fallback now uses `createInitial(session)` so the replacement snapshot matches the normal persistence schema instead of dropping fields that are only present in the serializer-managed shape.
- The regression test now stubs `IDialogService.prompt` and `IOpenerService.open`, so the serializer-failure path is deterministic and side-effect free.

thatdaveguy1:

PR #309287, opened 2026-04-12
chatEditing: fix listener leak in DiffHunkWidget toolbar

roblourens:
@jrieken I was seeing this warning, I guess I had a lot of diff hunks, this just creates the toolbar lazily and seems reasonable, lmk what you think.

## Problem

A "potential listener LEAK detected" warning fires when many diff hunks are visible in chat editing. Each `DiffHunkWidget` eagerly created a `MenuWorkbenchToolBar` in its constructor, and each toolbar subscribes to menu registry `onDidChange` events. With 200+ hunks, this exceeds the listener threshold.

## Fix

Make toolbar creation lazy in `DiffHunkWidget`:
- **`showToolbar()`** — creates the `MenuWorkbenchToolBar` on first call (idempotent)
- **`hideToolbar()`** — disposes the toolbar via `MutableDisposable.clear()`
- **`toggle(true/false)`** — calls `showToolbar()` / `hideToolbar()` respectively
- When widgets are returned to the pool, `hideToolbar()` is called to clean up

This keeps at most ~2 toolbars alive at a time (cursor position + hover target), down from one per hunk.


jrieken:
Closing this in favour of https://github.com/microsoft/vscode/pull/312317 - the PR inspired me to do a `MenuWorkbenchToolBar` wide solution that uses the intersection observer API to connect/disconnect toolbars based on the visibility

PR #309292, opened 2026-04-12
Fix ObjectMutationLog corrupting session files on failed writes

roblourens:
## Bug

When restoring a chat session from a `.jsonl` file, the session fails to load with:

```
ChatSessionStore: Malformed session data in …/chatSessions/….jsonl: Cannot read properties of undefined (reading 'requests')
```

## Root Cause

`ObjectMutationLog.write()` eagerly committed its internal state (`_previous`, `_entryCount`) **before** the caller persisted data to disk. If the initial file write failed (e.g. filesystem error), the log instance believed it had already written the `Initial` entry. Subsequent writes then produced only incremental `Set`/`Push` entries — creating a `.jsonl` file missing the required `Initial` entry. On reload, `read()` tried to apply mutations to `undefined` state, causing the crash.

## Fix

Introduce a two-phase write/confirm pattern in `ObjectMutationLog`:

- **`write()`** now stores updates in pending fields (`_pendingPrevious`, `_pendingEntryCount`) behind an explicit `_hasPendingWrite` flag, without committing them to the live state
- **`confirmWrite()`** (new) commits pending state — the caller must invoke this after the data is successfully persisted to disk
- If the write fails and `confirmWrite()` is never called, the next `write()` will produce a fresh `Initial` entry, preventing corrupted log files
- `createInitialFromSerialized()` and `read()` clear pending state on reset to prevent stale pending writes from being resurrected
- `read()` now throws a clear `'Log file is missing an initial entry'` error instead of crashing with a confusing property-access error

`chatSessionStore.ts` calls `confirmWrite()` after the successful `writeFile`.

## Test Coverage

- Added `confirmWrite()` calls to existing tests that perform sequential writes
- New tests: unconfirmed write resets, confirmed write increments, missing initial entry detection, failed-then-successful roundtrip, unconfirmed append after createInitial
- Updated perf benchmark to use the new `confirmWrite()` API

(Written by Copilot)

### DIFF
--- a/src/vs/workbench/contrib/chat/common/model/chatSessionStore.ts
+++ b/src/vs/workbench/contrib/chat/common/model/chatSessionStore.ts
@@ -384,6 +384,7 @@ export class ChatSessionStore extends Disposable {
 					if (data.byteLength > 0) {
 						await this.fileService.writeFile(storageLocation.log, data, { append: op === 'append' });
 					}
+					session.dataSerializer.confirmWrite();
 				} else {
 					const content = new ChatSessionOperationLog().createInitialFromSerialized(session);
 					await this.fileService.writeFile(storageLocation.log, content);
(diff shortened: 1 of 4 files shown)


connor4312 (approved):
Should be fine since writes are serialized. If they weren't there could be some races here, but this usage seems fine

PR #309332, opened 2026-04-13
Fix #32541: toggle block comment when cursor is inside comment

### LINKED ISSUE #32541: How do I toggle the block comment without having to select it?
There is a problem that bothers me. **If I want to remove the block comment, I have to select it**, like this:

If I do not select it, and go to toggle block comment, it becomes:

Have you a good idea to solve it?

**PS**: I use this hot key `ctrl+shift+/` to toggle block comments, which was bound to the command  `editor.action.blockComment`

amartin217:
Fix #32541: Toggle block comment from inside comment block

Summary:
Improves editor.action.blockComment so that when the cursor is inside an existing block comment with no selection, the command will remove the surrounding block comment instead of inserting a new one.

Changes:
- Detects when the cursor is inside a block comment using language configuration
- If collapsed selection is inside a block comment, removes the full block comment range
- Preserves existing behavior for selections, multi-cursor edits, and non-commented code

Why:
Makes block comment toggling more intuitive by allowing users to toggle comments without manually selecting the entire block, matching behavior in other editors.

Notes:
-Language-aware (no hardcoded comment tokens)
-No changes to line comment behavior or selection-based toggling


amartin217:

Issue #252792, opened 2025-06-28
Support Anthropic DXT Format For Packaging &  Installing MCPs


adamhill:
Please consider support Anthropic DXT packaging format for adding MCPs.

They gave us MCPs, now we get a packaging format from them - feels like a celebration! :-) 

Standardizing on the installation and packaging format for MCP's would be good for the community . If MS and Anthropic supported the format it would push other players to support it and greatly simplify all the problems we have installing MCP's and helping users diagnose their problems.

Now all we need is a discovery format using Gihub amd we will be set.

https://github.com/anthropics/dxt

Thanks for your consideration.

AClerbois:
@adamhill FYI:  
Following the decision to keep the MCP project alive, the DXT format has been renamed to MCPB (MCP Bundle) — see https://github.com/modelcontextprotocol/mcpb for details.

PR #309356, opened 2026-04-13
Initial revision of enterprise MCP allowlist support,

dmitrivMS:
Fixes https://github.com/microsoft/vscode-internalbacklog/issues/6959


joshspicer (dismissed):
Looks reasonable to me! Would be great to get @sandy081 's 👀 if possible

hstaudacher:
A change has been made to the `/copilot/mcp_registry` endpoint that should be respected here as well. It does support adding a `repo` param to scope the response to the repo's enterprise/org. Details can be found in this [issue](https://github.com/github/copilot-ecosystem/issues/848) (right now the team is fixing a bug with this param. Should work as expected soon).

src/vs/workbench/contrib/mcp/common/mcpRegistry.ts
```diff
@@ -502,6 +506,19 @@ export class McpRegistry extends Disposable implements IMcpRegistry {
...
+		if (allowlistResult !== true) {
+			if (opts.errorOnUserInteraction) {
+				throw new UserInteractionRequiredError('allowlist');
+			}
+			this._notificationService.notify({
+				severity: Severity.Warning,
+				message: allowlistResult.value,
+			});
+			return undefined;
+		}
```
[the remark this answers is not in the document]
sandy081: How does a user know if an installed MCP server is allowed or not by looking at it in the extensions view? 

I think we do the access validation in the mcp workbench service which is the source for all installed mcp servers and disable them if they are not valid. Please check here - https://github.com/microsoft/vscode/blob/503edcb00ba54414e9e6a3ae091a87ef2d46288a/src/vs/workbench/contrib/mcp/browser/mcpWorkbenchService.ts#L802-L823 - IMO we should do the same here

src/vs/workbench/contrib/mcp/common/mcpAllowListService.ts
```diff
@@ -0,0 +1,352 @@
...
+		const account = await this.defaultAccountService.getDefaultAccount();
+		if (!account) {
+			throw new Error('No default account');
+		}
+
+		if (token.isCancellationRequested) {
+			throw new Error('Cancelled');
+		}
+
+		const sessions = await this.authenticationService.getSessions(account.authenticationProvider.id, undefined, { silent: true });
```
sandy081, line 359: default account has session id - please use that to get the token

sandy081: Ah ignore this comment, you are using the session id already.

src/vs/workbench/contrib/mcp/common/mcpAllowListService.ts
```diff
@@ -0,0 +1,352 @@
...
+		const entries = this.defaultAccountService.policyData?.mcpAllowlistEntries;
+		if (!entries || entries.length === 0) {
+			return;
+		}
+
+		const entryMap = new Map(entries.map(e => [e.ownerLogin, e]));
+		const seenRepos = new Set<string>();
+		const matchedOwners = new Set<string>();
+
+		for (const repository of this.gitService.repositories) {
```
sandy081: `gitService.repositories` -> This is not clear what is the list of repositories it contain and how does this work?

sandy081: cc @lszomoru

src/vs/workbench/contrib/mcp/common/mcpAllowListService.ts
```diff
@@ -0,0 +1,352 @@
...
+			}
+
+			try {
+				const current = await this.loadAllowList({ entry, repo }, authToken, token);
+				result = this.mergeAllowLists(result, current);
+			} catch (error) {
+				this.logService.debug(`[McpAllowlist] Error loading allowlist for ${entry.ownerLogin}${repo ? `/${repo}` : ''}:`, error);
+				lastError = error;
+			}
+		}
```
sandy081: Are we getting allow list of all repositories or Am I wrong here?

src/vs/workbench/contrib/mcp/common/mcpAllowListService.ts
```diff
@@ -0,0 +1,352 @@
...
+
+		const cached = this.storageService.getObject<{ value: AllowList }>(storageKey, StorageScope.APPLICATION);
+		if (cached) {
+			this.logService.debug(`[McpAllowlist] Loaded allowlist for ${repo ?? '<none>'} from cache`);
+			return cached.value;
+		}
+
+		const value = await this.fetchAllowList(query, authToken, token);
+		this.storageService.store(storageKey, JSON.stringify({ value }), StorageScope.APPLICATION, StorageTarget.MACHINE);
+		return value;
```
sandy081, line 317: What is the cache invalidation strategy? What happens if the allow list values changes in the server? Why are we caching only allow list values of a repo if we are fetching for all repos?

sandy081 (commented):
Very good work on this, I liked how the overall changes look. I left some comments on the implementation. Thanks.

src/vs/workbench/contrib/mcp/common/mcpAllowListService.ts
dmitrivMS: Yes, we are getting allow for each repository and then take the intersection of the sets.

src/vs/workbench/contrib/mcp/common/mcpAllowListService.ts
dmitrivMS: We cache per repo - this way if you change workspace and there are same repos, we will have the cache ready. Since it's per repo, we can construct various combinations of repos without having to hit the API.

There is background refresh instead of invalidation - i.e. we update from policy every 30 minutes. We should probably consider a force refresh mechanism as well.

src/vs/workbench/contrib/mcp/common/mcpAllowListService.ts
dmitrivMS: This is the list of all repos discovered by the Git service from the workspace. It does not provide event mechanism right now, so we rely on workspace changes to reevaluate.

sandy081: @lszomoru Can you confirm how `gitService.repositories` works here? My understanding is that one can open any repository using this service, so it could potentially contain all repositories — not just the ones in the current workspace. If that is the case, relying on it for scoping the enterprise allowlist may not give the expected results. Thanks!

src/vs/workbench/contrib/mcp/common/mcpAllowListService.ts
sandy081: Why are we getting the allow list for all repositories? Should we not get it only for the current repository? What is the additional overhead of making these unnecessary requests? Also if we fetch per repo then caching per repo makes sense, but currently we are fetching for all repos and still caching per repo — which seems inconsistent.

src/vs/workbench/contrib/mcp/common/mcpRegistry.ts
```diff
@@ -562,4 +579,70 @@ export class McpRegistry extends Disposable implements IMcpRegistry {
...
+		if (!launch) {
+			return undefined;
+		}
+
+		try {
+			// Dynamic import to avoid loading the SDK when enterprise is not active
+			const { computeFingerprint, commandToRegistryType } = await importAMDNodeModule<typeof import('@github/mcp-registry')>('@github/mcp-registry', 'dist/index.js');
+
+			if (launch.type === McpServerTransportType.Stdio) {
+				const registryType = commandToRegistryType(launch.command);
```
rroesch1, line 623: If i got this right this `commandToRegistryType` trys to find the "registry type" based on the command args.
e.g.
* if the command is `npx`, the registry type is npmjs
* if the command is `uvx`, the registry type is PyPI
* if the command is `docker`, the registry type is OCI

This approach works for a few (but **currently** very common) registry types.

But how is it supposed to work for "registry types" which do not have a commonly used "one-stop install and run executable" (`npx`, `uvx`, `docker run`)?

Example:
* `mcpb` 
  (see: #290111 and #252792)
* `cargo`  (Rust)
  see: modelcontextprotocol/registry#1055
* golang apps: the common approach to install golang apps is to use `go install …`. The command downloads, compiles the app and adds it to the `$PATH`.  
  Example: The [go language server](https://go.dev/gopls/) has a [built-in MCP-Server](https://go.dev/gopls/features/mcp/) and the [official local GitHub MCP-Server](https://github.com/github/github-mcp-server#local-github-mcp-server) is also implemented in go (but distributed as OCI image)
* Maven Artifacts (Java): There is no one-stop command to download and execute an executable JAR from a Maven repository
* MCP servers which are bundled with existing applications
  Example: [maestro MCP](https://docs.maestro.dev/get-started/maestro-mcp) is bundled with [maestro](https://maestro.dev/). The command for the MCP-Server is `maestro mcp`.
* MCP servers distributed with package managers  
  (homebrew, apt, dnf/yum, winget etc.)

rroesch1: fyi: The MCP Registry spec now got a `cargo` packaging type (see: modelcontextprotocol/registry#1207).

The spec explicitly mentiones the absence of an 'one-stop install and run executable' in the [Runtime Model section](https://github.com/modelcontextprotocol/registry/blob/685e354b2fa37e5fe6d70dab0edb26a84e7ddf64/docs/modelcontextprotocol-io/package-types.mdx#runtime-model):
> Cargo's runtime model differs from npm/PyPI/NuGet. cargo install <crate> places the compiled binary on PATH at ~/.cargo/bin, after which MCP clients invoke it directly by name. There is no per-invocation runner equivalent to npx (npm), uvx (PyPI), or dnx (NuGet, .NET 10 SDK Preview 6+) — install is one-time, execution is by binary name.

PR #309364, opened 2026-04-13
feat: auto-trigger parameter hints when cursor enters argument list

yogeshwaran-c:
## Summary

Add a new setting `editor.parameterHints.autoTriggerOnCursorInArgs` (default `false`). When enabled, moving the cursor into an existing, unmatched opening parenthesis shows the parameter hints widget -- previously hints only triggered on typing `(` or `,`.

This solves the common case of navigating back into an existing call expression via click or arrow keys and expecting to see parameter info without deleting and retyping the `(`.

## Implementation

- `src/vs/editor/common/config/editorOptions.ts`: adds the new boolean option with schema description, default `false` to preserve existing behavior.
- `src/vs/editor/contrib/parameterHints/browser/parameterHintsModel.ts`: on cursor change, when the option is enabled and the cursor lands inside an unmatched `(`, calls `trigger({ triggerKind: Invoke })`. Uses a same-line scan that honors `'...'`, `"..."`, and ```...``` string delimiters to avoid false positives from parentheses inside string literals.

Opt-in so this does not change default behavior.

Fixes #252076

### DIFF
--- a/src/vs/editor/common/config/editorOptions.ts
+++ b/src/vs/editor/common/config/editorOptions.ts
@@ -3659,6 +3659,11 @@ export interface IEditorParameterHintOptions {
 	 * Defaults to false.
 	 */
 	cycle?: boolean;
+	/**
+	 * Automatically show parameter hints when the cursor enters an unmatched open parenthesis.
+	 * Defaults to false.
+	 */
+	autoTriggerOnCursorInArgs?: boolean;
 }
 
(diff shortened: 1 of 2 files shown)


mjbvz:
Closing as this doesn't seem right at all

The parameter hints  can show up anywhere. We can't hardcode core VS Code to only work with parens

PR #309385, opened 2026-04-13
Remove unnecessary ... from 'Delete...' text in delete session context menu

abhijit-chikane:


abhijit-chikane:

dmitrivMS:
`...` means it will pop up a dialog, so it's not unnecessary.

PR #309404, opened 2026-04-13
nes: cursor distance check, ghost text rejection, and previously-rejected icon

ulugbekna:
Three NES improvements, each behind a separate experiment flag:

### 1. Cursor line distance check for cache serving
**Flag:** `InlineEditsCacheCursorDistanceCheck` (default off)

A cached NES that is not rebased or subsequent is not served if the user's cursor moves farther (by line) from the edit's start line than it was at cache time. The check is dynamic — the edit recovers if the user moves back closer.

### 2. Don't show ghost-text-shown edits as inline edits
**Flag:** `InlineEditsRejectShownGhostText` (default off)

If an NES was shown as ghost text and not accepted, prevent it from being shown again as an inline edit. The edit can still be served as ghost text from cache if the user returns to the same position. Tracking is done eagerly at show time to avoid a race with cursor-triggered NES requests.


ulugbekna:
age

PR #309465, opened 2026-04-13
fix: plugin skill files not accessible when connected to remote

### LINKED ISSUE #305168: Agent plugin skill files are inaccessible when connected to WSL/SSH remote
- Copilot Chat Extension Version: 0.41.1
- VS Code Version: 1.113.0
- OS Version: Windows 11 26100.7984 with WSL2 Ubuntu 24.04
- Feature (e.g. agent/edit/ask mode): Agent mode (skill loading from agent plugins)
- Selected model (e.g. GPT 4.1, Claude 3.7 Sonnet): Claude Opus 4.6
- Logs: N/A

### Summary

When VS Code is connected to a remote (WSL/SSH), agent plugins installed from the marketplace are cached on the **Windows host** filesystem (`C:\Users\...\AppData\Local\...\agentPlugins\`). This causes a cascade of issues that make plugin skills unusable:

1. **Wrong path format**: Skill file URIs are emitted to the agent as Windows paths (e.g. `c:\Users\yshoham\AppData\Roaming\Code\agentPlugins\...\skills\jira\SKILL.md`), but the `read_file` tool executes on the Linux remote where those paths don't exist.

2. **Confirmation dialog blocks access**: Even if the agent (or a workaround instruction file) translates the Windows path to a WSL-accessible path (`/mnt/c/Users/.../agentPlugins/.../SKILL.md`), the `read_file` tool triggers an "Allow reading external files?" confirmation dialog because agent plugin directories are outside the workspace and are not recognized as trusted.

### Root cause

`agentPluginRepositoryService.ts` resolves the plugin cache to `environmentService.cacheHome / 'agentPlugins'`, which always points to the **local (Windows) host** filesystem — even when connected to a remote. PR #303606 fixes git clone routing for the remote case but does not address the cache location or the path format presented to the agent.

In `computeAutomaticInstructions.ts`, `getFilePath(skill.uri, remoteOS)` normalizes path separators for the remote OS, but the URI itself is a `file://` URI on the Windows side, so the resulting path is still unreachable from the remote.

### Steps to Reproduce

1. Open VS Code on Windows and connect to a WSL remote
2. Install an agent plugin from the marketplace that includes skills
3. Open a workspace in WSL and start a chat session in agent mode
4. Ask a question where one of the plugin's skills applies
5. The agent attempts to call `read_file` to load the SKILL.md

### Expected behavior

Skill files from installed agent plugins should be readable by the agent without any user intervention. Either:

- The plugin cache should resolve to a remote-accessible path when connected to a remote, or
- Skill URIs emitted to the agent should be translated to paths accessible from the remote OS, or
- At minimum, agent plugin directories should be auto-trusted so the "Allow reading external files?" confirmation is not shown

### Actual behavior

- The agent receives a Windows-style path that does not exist on the Linux remote
- If translated to `/mnt/c/...` (via a workaround instruction file), the read triggers a confirmation dialog
- The user must manually press "Allow" for every skill file read, breaking the agent workflow

### Workaround

Create an instruction file (e.g. `.copilot/instructions/wsl-skill-paths.instructions.md`) that tells the agent to convert `c:/` paths to `/mnt/c/` before reading. This addresses issue 1 but not issue 2 (the confirmation dialog).

### Screenshots

Without workaround:

Image

With workaround:

```md
---
name: wsl-skill-paths
description: "ALWAYS check BEFORE loading or reading any skill file, SKILL.md, or file with a Windows-style path. MANDATORY for all skill loading in WSL. Converts Windows drive letter paths to WSL mount paths."
---

# WSL Path Conversion for Skill Loading

**Before reading any skill file**, convert Windows drive letter paths to WSL mount paths:

- `c:/` → `/mnt/c/`
- `d:/` → `/mnt/d/`

Example: `c:/Users/yshoham/AppData/Roaming/Code/agentPlugins/…/SKILL.md` → `/mnt/c/Users/yshoham/AppData/Roaming/Code/agentPlugins/…/SKILL.md`

Use `read_file` with the converted `/mnt/c/…` path. Never pass a `c:/…` path directly.
```

Image

cavalloJustinEmery:
When connected to an SSH, WSL, or dev container remote, local `file://` URIs for agent plugin skill files were being converted to native filesystem paths (e.g. `C:\Users\...` or `/home/...`) by `getFilePath()`. The remote extension host can't read these paths — they only exist on the local machine — so skill files were silently inaccessible.

Additionally, plugin-storage skill files were never included in the `_listExtensionPromptFiles` command result, so the copilot extension's trust cache never contained them. This caused a confirmation dialog to appear on every skill file read, even after the user had already granted trust.

## Changes

**`computeAutomaticInstructions.ts`** — `getFilePath()` gains an `isRemote` parameter. When `true` and the URI scheme is `file://`, it returns a `vscode-local:/` URI string instead of a native path. The remote extension host resolves `vscode-local:/` URIs via the built-in local file bridge, which works transparently across SSH, WSL, and dev containers.

**`chatPromptFilesContribution.ts`** — `_listExtensionPromptFiles` now includes `PromptsStorage.plugin` files alongside `PromptsStorage.extension` files. The IPC transport automatically rewrites their `file://` URIs to `vscode-local:/` as they cross the local/remote boundary, so the trust cache entries match what `getFilePath()` emits.

## Testing

Unit tests cover the new `getFilePath()` behavior:
- Returns a `vscode-local:/` URI string for `file://` URIs when `isRemote=true` (Linux and Windows remote OS)
- No conversion when `isRemote=false`
- No conversion for `vscode-remote://` URIs even when `isRemote=true`

Fixes #305168


cavalloJustinEmery:
@microsoft-github-policy-service agree company="Cavallo Solutions"

src/vs/workbench/contrib/chat/common/promptSyntax/computeAutomaticInstructions.ts
```diff
@@ -630,7 +631,14 @@ export class ComputeAutomaticInstructions {
 }
 
 
-export function getFilePath(uri: URI, remoteOS: OperatingSystem | undefined): string {
+export function getFilePath(uri: URI, remoteOS: OperatingSystem | undefined, isRemote = false): string {
```
connor4312, line 636: @aeschli  for review. I don't love this but I think this is also not avoidable without a larger structual change that would preserve URIs in IPC such that the `URITransformer` sees them.

aeschli:
@connor4312 The plan is that the src/vs/workbench/contrib/chat/common/promptSyntax/computeAutomaticInstructions.ts moves to the extension host and will only be used by the local harness.
I believe both the CopilotCLI and CC filter the index file attachment as the harness itself has ways of adding the list of available skills/instructions and subagents.
Would that work for you?

connor4312:
yea I think that would be fine.

rhobison:
Hi!
Any plans to fix this issue by merging this PR?
I am facing this same issue as well (skills from an agent plugin not accessible inside a container).

aeschli (dismissed):

sandy081 (approved):

sandy081 (approved):

Yoyokrazy (approved):

Issue #309523, opened 2026-04-13
`chat.tools.terminal.backgroundNotifications: false` still fires completion notifications for foreground terminals


karthiknadig:
Testing #309095 (section 1, Windows).

### What I expected
With `chat.tools.terminal.backgroundNotifications: false`, the agent should *not* receive automatic terminal-completion notifications. This is reflected in the fact that the per-call runtime instruction block drops the sentence *"You will be automatically notified on your next turn when it completes"* in this mode.

### What actually happened
Even with `chat.tools.terminal.backgroundNotifications: false`, completion notifications **still fired automatically** for foreground sync-mode terminals that had exceeded their initial timeout. Observed with two separate commands during the same session, both of which produced the full `[Terminal <id> notification: command completed with exit code 0 ...]` tail on the next agent turn with no polling:

1. `cargo check` foreground, sync mode, short initial timeout. Notification fired on its own when the build finished.
2. `pwsh -NoProfile -File ./scripts/test-prompt.ps1` foreground, sync mode. Notification fired on its own after the script exited.

So either:
- the setting only gates *input-waiting* notifications (not completion notifications) and the docs / runtime note are misleading, **or**
- completion notifications are not honoring the setting.

### Steps to reproduce (Windows, Nectarine / default agent, agent mode)
1. Set `chat.tools.terminal.backgroundNotifications` to `false` in workspace settings.
2. Ask the agent to run a long-running foreground command (e.g. `cargo check` on a non-trivial Rust workspace, or any compile that takes ≥30s) with a short initial timeout so the tool returns while the command is still running.
3. Do **not** poll `get_terminal_output`. Continue the conversation with an unrelated step.
4. Observe: when the foreground command eventually finishes, an automatic `[Terminal ... notification: command completed ...]` chunk is injected on the next agent turn anyway.

### Impact
Low-to-medium: it means the setting's off-state is not fully off — users who turn background notifications off because they find them noisy will still get completion-notification chatter. It also makes it harder to deterministically test the difference between the two setting values.

### Environment
- OS: Windows
- VS Code Insiders, agent mode, `Nectarine` model
- Repo under test: microsoft/python-environment-tools (Rust workspace, long build)

cc @meganrogge

Issue #309095, opened 2026-04-10
Test: terminal tool improvements - `send_to_terminal`, fg support, prompt for input revamping


meganrogge:
Refs: #309044, #308870, #308868, #308296, #308030, #288570

- [x] Windows @karthiknadig 
- [x] macOS @connor4312 
- [x] Linux @rzhao271 

Complexity: 5

[Create Issue](https://github.com/microsoft/vscode/issues/new?template=blank&body=Testing+%23309095%0A%0A&assignees=meganrogge)

---

### Overview
This iteration overhauled how the agent handles terminal input. Previously, every terminal output chunk triggered an extra LLM call to classify whether the terminal was prompting for input — adding latency, burning tokens, and producing unreliable results across models. That entire mechanism has been removed. The agent now handles terminal input directly via `send_to_terminal` (which now works with foreground terminals), and uses the `ask_questions` carousel to defer to the user when needed.

**Key changes:**
- Removed LLM-based prompt-for-input detection (saves tokens + latency per command)
- `send_to_terminal` and `get_terminal_output` now work with foreground terminals, not just background
- `ask_questions` carousel gains a "Focus Terminal" button when triggered by terminal prompts
- Carousel auto-dismisses when user types directly in terminal
- `send_to_terminal` progress messages show the question being answered
- Confirmation dialog uses terminal title instead of raw UUID

> **Note:** Testers should run through sections 1–4 with both `chat.tools.terminal.backgroundNotifications: true` (default) and `chat.tools.terminal.backgroundNotifications: false`.

---

### 1. Foreground terminal support (`send_to_terminal` / `get_terminal_output`)

- [ ] In agent mode, ask the agent to run a long-running foreground command (e.g. `npm run dev` or `python -m http.server`). Verify the agent can use `get_terminal_output` to read output from the foreground terminal.
- [ ] Ask the agent to send input to a foreground terminal (e.g. a running REPL). Verify `send_to_terminal` successfully writes to it.
- [ ] Run a command that prompts for input (e.g. `npm init`). Verify the agent can answer prompts via `send_to_terminal` without a separate LLM classification step.
- [ ] Verify that background terminals still work as before — `run_in_terminal`, `send_to_terminal`, and `get_terminal_output` all function for background terminals.

### 2. Question carousel with terminal context

[...]

PR #309555, opened 2026-04-13
Deprecate `chat.tools.terminal.backgroundNotifications` — always send terminal notifications

meganrogge:
Addresses #309523 
Addresses https://github.com/microsoft/vscode/issues/308548

## Why this needs to be fixed now

This was filed as part of testing the terminal tool improvements (#309095). The `chat.tools.terminal.backgroundNotifications` setting has an inconsistent off-state: setting it to `false` suppresses notifications for true async terminals (`mode=async`) but **not** for foreground terminals that moved to background (timeout/input-needed). This is because the code intentionally bypasses the setting for fg-to-bg transitions — the tool result already promises the model "you will be notified when it completes," so suppressing the notification would leave the agent unable to learn the command finished.

This means the setting was never truly "off" — it only partially suppressed notifications, making it confusing for users and impossible to test deterministically (#309523).

## What changed

- **Always send completion/input-needed notifications** for any terminal in background execution (both true async and fg-to-bg), except for subagent-initiated terminals which can't receive steering messages
- **Deprecated the setting** with default changed to `true` and a deprecation message
- **Updated tool description text** to always include the notification language, removing the conditional phrasing that was misleading
- Removed the `backgroundNotifications` parameter plumbing from all model description helper functions

## Rationale

- Notifications are strictly better for agent quality than requiring polling — less latency, fewer wasted `get_terminal_output` calls
- The fg-to-bg path already always notified regardless of the setting, so behavior was already inconsistent
- The setting was experimental and gated for A/B rollout (`experiment: { mode: 'auto' }`), not intended as a permanent user preference
- The only cost of always-on is extra LLM turns for true async terminals, which is the correct behavior anyway

### DIFF
--- a/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/runInTerminalTool.ts
+++ b/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/runInTerminalTool.ts
@@ -94,7 +94,7 @@ type IRunInTerminalToolInvocationData = IChatTerminalToolInvocationData & {
 	requiresConfirmationForRetry?: boolean;
 };
 
-function createPowerShellModelDescription(shell: string, isSandboxEnabled: boolean, backgroundNotifications: boolean, networkDomains?: ITerminalSandboxResolvedNetworkDomains): string {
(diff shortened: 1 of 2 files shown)


anthonykim1:
@meganrogge Seems like test might be acting flaky

roblourens (approved):

PR #309598, opened 2026-04-13
Skip redundant `send_to_terminal` confirmation when replying to `askQuestions` carousel

### LINKED ISSUE #309589: Terminal commands should be auto approved when invoked from `ask_questions` tool
Since the user already confirmed their answers, we should not prompt, esp about sending `Enter`.

https://github.com/user-attachments/assets/1089c44c-9ada-40cb-89c1-2f40b8cf10a8

meganrogge:
Fixes #309589
Fixes https://github.com/microsoft/vscode/issues/309608

When an interactive command (e.g. `npm init`) waits for input in default approvals mode, the agent calls `askQuestions` to collect user answers, then sends each answer via `send_to_terminal`. Previously, a `send_to_terminal` call could show a confirmation dialog, even though the user just explicitly provided that text through the askQuestions UI.

This skips the confirmation when we can correlate the `send_to_terminal` text to a recently answered question carousel for the same terminal.

### Changes

**Confirmation skip** — `_getQuestionContextForTerminal` already returns the question text for the "(replying to: …)" display. A non-`undefined` result now also signals that confirmation is unnecessary. `forceConfirmationReason` still takes precedence, and autopilot mode behavior is unchanged.

**Exact answer matching** — Previously, `_getQuestionContextForTerminal` had shortcuts that could return a question without verifying the command matched the answer (single-question shortcut, multi-question fallback). This meant any `send_to_terminal` call could bypass confirmation whenever a recent carousel existed for that terminal. Now the method only returns a question when the command exactly matches the answer at the correct position.

**Positional matching** — When all answers are identical (e.g. pressing Enter to accept every default in `npm init`), answer-value matching always returned the first question. The method now counts prior `send_to_terminal` tool invocations after the carousel to determine the positional index, so the progress message correctly shows "Package name?" → "version?" → "description?" instead of repeating "Package name?" for every prompt.

https://github.com/user-attachments/assets/9edfa037-06d5-4f0f-8773-f19ce232c784


meganrogge:
closing as we're not doing candidates probably

TylerLeonhardt (approved):

PR #309606, opened 2026-04-13
Prevent enable si showing up when using pwsh from chat with accessibility mode

### LINKED ISSUE #307698: Enable shell integration on Chat view even when shell integration is already enabled
Shell integration should also be enabled on the Chat view, even when it is already enabled elsewhere. Currently it seems like the Chat view doesn't get shell integration in this scenario.

/cc @jooyoungseo @rperez030 @accesswatch

anthonykim1:
Resolves: https://github.com/microsoft/vscode/issues/307698

Before: 
image

After: 

https://github.com/user-attachments/assets/2c966ae0-4a65-408c-aa4c-8bc90b67b777

Two parts to this:

* timeout bump for A11y: We add extra 3 second when waiting for shell integration with accessible mode toggled, since we remove and load our bundled PSReadLine in our shell integration script. This would allow us for the first command to chat gets proper command detection.
    * Without the timeout bump we get 
image
  
* Deferred banner: Instead of deciding whether to show "enable shell integration banner" right before running command, we should wait until command finishes, 
    * Without the deferred banner we get the `enable shell integration` message again


src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/runInTerminalTool.ts
```diff
@@ -1300,7 +1300,10 @@ export class RunInTerminalTool extends Disposable implements IToolImpl {
 				commandDetection!,
 				executionOptions.persistentSession
 			);
-			if (toolTerminal.shellIntegrationQuality === ShellIntegrationQuality.None) {
```
anthonykim1, line 1308: Look into timing and look into toolTerminal vs. tab

anthonykim1:
image
 @meganrogge was right, powershell 7 did not show the `enable shell integration warning` whereas windows powershell did.

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/runInTerminalTool.ts
```diff
@@ -1305,9 +1305,10 @@ export class RunInTerminalTool extends Disposable implements IToolImpl {
...
```
anthonykim1, line 1308: The shell integration banner decision is deferred to after command execution completes (see below). This gives SI time to arrive — on Windows PowerShell 5.1, cold-loading PSReadLine can push `HasRichCommandDetection` past the initial wait timeout, and checking here would incorrectly show the banner.

Yoyokrazy (approved):

PR #309655, opened 2026-04-13
Fix PascalCase keys in copilot cgmanifest.json

benvillalobos:
The `extensions/copilot/cgmanifest.json` uses PascalCase keys (`Registrations`, `Component`, `Type`, etc.) instead of the camelCase required by the [official schema](https://json.schemastore.org/component-detection-manifest.json). This caused the OSS license tool in vscode-build-tools to crash silently — `fileContents.registrations` was `undefined` because the key was `Registrations`.

This normalizes the file to use camelCase keys matching every other `cgmanifest.json` in the repo.

Companion PR: microsoft/vscode-build-tools — adds defensive PascalCase→camelCase normalization in the OSS tool.

### DIFF
--- a/extensions/copilot/cgmanifest.json
+++ b/extensions/copilot/cgmanifest.json
@@ -1,16 +1,16 @@
 {
 	"$schema": "https://json.schemastore.org/component-detection-manifest.json",
-	"Registrations": [
+	"registrations": [
 		{
-			"Component": {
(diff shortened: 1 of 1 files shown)


benvillalobos:
@hediet another case where screenshots shouldn't have been taken for this PR

mjbvz (approved):

PR #309700, opened 2026-04-14
Add performance tests

pwang347:
## Add chat performance benchmarking harness

Introduces an end-to-end chat performance benchmarking and memory leak detection framework, backed by a deterministic mock LLM server and a CI workflow for automated regression testing.

### What's included

**Perf regression runner** (`npm run perf:chat`)
- Launches VS Code via Playwright Electron, opens the chat panel, sends messages with a mock LLM response, and measures timing, layout, rendering, and memory metrics
- Compares a test build against a baseline (defaults to latest stable release) using Welch's t-test for statistical significance
- Supports dev, production (`--production-build`), and release builds with mismatch detection
- Resumable runs (`--resume`) to accumulate samples for higher confidence
- Per-build settings overrides for A/B testing experimental features

**Memory leak checker** (`npm run perf:chat-leak`)
- Cycles through all registered scenarios in a single session, measuring heap and DOM node growth between iterations
- Uses linear regression on post-GC heap samples to detect sustained per-iteration leaks

**Mock LLM server** (mock-llm-server.js)
- Implements the full CAPI URL structure (`/models`, `/models/session`, `/chat/completions`, etc.) for deterministic, zero-latency responses
- Supports content streaming, tool calls, thinking blocks, and multi-turn conversations via a scenario-based architecture

**CI workflow** (chat-perf.yml)
- `workflow_dispatch` with configurable inputs (baseline/test commits, runs, threshold, settings overrides)
- Builds once, then fans out perf comparison across 4 matrix groups + a parallel leak check
- Merges per-group results into a unified Markdown summary posted to the GitHub step summary
- All user inputs passed via environment variables (not direct `${{ inputs.* }}` interpolation) to prevent script injection

**Scenarios** (perf-scenarios.js)
- Content-only: `text-only`, `large-codeblock`, `rapid-stream`, `mixed-markdown`
- Tool-call: `tool-read-file`, `tool-edit-file`, `tool-terminal`
- Multi-turn: `thinking-response`, `multi-turn-user`, `long-conversation`

### Other changes
- package.json: Added `perf:chat` and `perf:chat-leak` npm scripts
- SKILL.md: Agent skill documentation for running benchmarks


.github/workflows/chat-perf.yml
```diff
@@ -0,0 +1,198 @@
+name: Chat Performance Comparison
+
+on:
+  pull_request:
```
pwang347: just for testing

scripts/chat-simulation/test-chat-perf-regression.js
```diff
@@ -0,0 +1,1389 @@
...
+
+		const cdp = await window.context().newCDPSession(window);
+		await cdp.send('Performance.enable');
+		const heapBefore = /** @type {any} */ (await cdp.send('Runtime.getHeapUsage'));
+
+		// Stop any existing tracing session (stable builds may have one active)
+		try { await cdp.send('Tracing.end'); await new Promise(r => setTimeout(r, 200)); } catch { }
+		await cdp.send('Tracing.start', {
+			traceConfig: {
+				includedCategories: ['v8.gc', 'devtools.timeline'],
```
deepak1556: Whats the reason behind not using the existing trace infra in the runtime ? `--enable-tracing` and `--trace-startup-file`. Our current startup perf module relies on it to extract relevant category details https://github.com/microsoft/vscode-perf/blob/9c76748df860a2cd54396fc437a78b4ab2b7c79a/src/perf.ts#L175-L186 for example. 

You could extract the heap usages (from gc metrics), cpu samples, perf markers from a unified trace instead of injecting process global performance observers right ?

roblourens (dismissed):

scripts/chat-simulation/common/mock-llm-server.js
```diff
@@ -0,0 +1,1014 @@
...
+ *   kind: 'thinking',
+ *   thinkingChunks: StreamChunk[],
+ *   chunks: StreamChunk[],
+ * }} ModelScenarioTurn
+ */
+
+/**
+ * A model turn that emits content chunks.
+ *
+ * @typedef {{
```
amunger, line 68: why use js with types rather than typescript?

scripts/chat-simulation/fixtures/_chatperf_async.ts
```diff
@@ -0,0 +1,132 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+// perf-benchmark-marker
+
+/**
+ * Fixture for chat-simulation benchmarks.
+ * Simplified from src/vs/base/common/async.ts for stable perf testing.
```
amunger, line 10: these utils should be quite stable. Do we need an extra copy of them?

pwang347: Yeah, I initially just referenced those files directly in the test cases but I think it's probably more correct to have these as separate test fixtures in case things actually get moved later. The main downside is that these will show up in file search, so maybe I could at least prefix them with something less common.

scripts/chat-simulation/common/mock-llm-server.js
pwang347: Yeah, this is mostly just following the current pattern/precedent of using .js files directly in the `scripts` directory. There's technically support for running `.ts` files already without needing to change our compile step so I can do that as well

scripts/chat-simulation/fixtures/_chatperf_async.ts
pwang347: Made the names more obscure

scripts/chat-simulation/common/mock-llm-server.js
pwang347: Discussed offline, will defer this as a followup. Thanks!

amunger (approved):

Issue #309501, opened 2026-04-13
BYOK model carries over from individual account


cwebster-99:
Testing #309023

- Sign in with an individual account and add a model using BYOK 
- Sign out and sign in with enterprise managed account with BYOK disabled 
- BYOK model still shows in the picker but it doesnt show in the Language Models UI 

<img width="412" height="728" alt="Image" src="https://github.com/user-attachments/assets/9148731a-9eb0-490a-aee8-ed5b30f180da" />

<img width="554" height="755" alt="Image" src="https://github.com/user-attachments/assets/2a9d3a8a-6d46-4052-a78e-d75b66dee772" />

<img width="1847" height="1111" alt="Image" src="https://github.com/user-attachments/assets/f0b95e61-3867-46d3-9cb3-9d309ef35570" />

vijayupadya:
Fixed by https://github.com/microsoft/vscode/pull/309889

Issue #309499, opened 2026-04-13
BYOK model was hidden by default


cwebster-99:
Testing #309023

I added a BYOK model and didnt see it in the model picker. I clicked on the ⚙️and found that it was hidden preventing it from showing in the model dropdown. I would expect it to be visible by default

vijayupadya:
This is fixed by https://github.com/microsoft/vscode/pull/314598

Issue #309495, opened 2026-04-13
Rough edges for invalid BYOK flows


cwebster-99:
Testing #309023

Couple of rough edges here:
- Added a BYOK model when it is "invalid" gives you no warning. Have to close and reopen Language Models menu to see it added at the bottom as "invalid". I would expect some sort of warning dialog or the Language Models list to auto update so I can see it is invalid without having to close and reopen 
- Removing a BYOK model gives no confirmation and the Language Models page is not auto updated so it looks like the removal didn't work. Have to close and reopen Language Models view to get it to update

https://github.com/user-attachments/assets/118f8c5a-8352-4793-9df7-a6ada04b2ee2

vijayupadya:
This should be fixed now.

Issue #309492, opened 2026-04-13
Add Models is disabled when BYOK is enabled


cwebster-99:
Testing #309023

Signed in with enterprise managed account that has BYOK enabled 

![Image](https://github.com/user-attachments/assets/ec1db62b-5517-4c40-8731-0cdfd78b62d8)

Model picker -> Other Models -> Manage Models to open this view. The Add Models button still says disabled 

![Image](https://github.com/user-attachments/assets/3cb1077e-2fad-453f-b2aa-c01c602e6b2b)

PR #309889, opened 2026-04-14
Fix enterprise BYOK issues: hidden models, disabled button, account carryover

pierceboggan:
Fixes three bugs reported against #309023 in the BYOK (Bring Your Own Key) model feature:

### #309499 — BYOK model hidden by default
`byokKnownModelToAPIInfo()` was missing `isUserSelectable: true`, so the view model defaulted newly added BYOK models to hidden (`metadata.isUserSelectable ?? false`).

**Fix:** Add `isUserSelectable: true` to the returned `LanguageModelChatInformation`.

### #309492 — Add Models button disabled when BYOK is enabled
`ContextKeysContribution` constructor called `_inspectContext()` and `_updatePermissiveSessionContext()` on init, but never `_updateClientByokEnabledContext()`. The BYOK context key was only set on auth *changes*, so if auth was already established at extension activation, the context key remained unset and `clientByokEnabled` returned `false`.

**Fix:** Call `_updateClientByokEnabledContext()` during initialization.

### #309501 — BYOK model carries over from individual account
BYOK provider registrations were added to the main `DisposableStore` and never cleaned up on account switch. When switching from an individual account (BYOK enabled) to an enterprise account (BYOK disabled), providers remained registered and models persisted in the picker.

**Fix:** Use a separate `DisposableStore` for BYOK provider registrations, clear it when auth changes and BYOK is disabled, and reset the `_byokProvidersRegistered` flag.

---

> **Note:** #309495 (rough edges for invalid BYOK flows — no validation warning on add, no UI auto-update on remove) is not addressed in this PR and needs further investigation.

### DIFF
--- a/extensions/copilot/src/extension/byok/common/byokProvider.ts
+++ b/extensions/copilot/src/extension/byok/common/byokProvider.ts
@@ -147,6 +147,7 @@ export function byokKnownModelToAPIInfo(providerName: string, id: string, capabi
 		family: id,
 		tooltip: `${capabilities.name} is contributed via the ${providerName} provider.`,
 		multiplierNumeric: 0,
+		isUserSelectable: true,
 		capabilities: {
 			toolCalling: capabilities.toolCalling,
 			imageInput: capabilities.vision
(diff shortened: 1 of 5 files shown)


roblourens (dismissed):

Yoyokrazy (approved):

pierceboggan:
Having good success testing with this tonight. 
image

PR #309942, opened 2026-04-14
Implementing user-scoped memory

sneefyyy:
# User-Scoped Memories + Unified `/prompt` Endpoint Migration

## What this PR does

This PR extends Copilot Memory (CAPI) support in four areas:

1. **User-scoped memory** — adds a `scope` parameter to the `store_memory` tool so facts can be stored per-user rather than only per-repo
2. **Unified `/prompt` endpoint** — migrates away from a bespoke memory fetch API to the shared `/prompt` endpoint, which returns context, store instructions, and tool schema in one call
3. **Turn-0 tool availability fix** — `store_memory` was always missing from the tool list on the first turn of every conversation; this is now fixed
4. **Local memory tool coexistence** — when `MemoryToolEnabled` and `CopilotMemoryEnabled` are both on, instructions and context for both storage systems are shown together rather than one replacing the other

---

### 1. User-scoped memory

`store_memory` now accepts an optional `scope` field (`'repo'` | `'user'`, defaulting to `'repo'`). Routing:

- `scope: 'repo'` → `AgentMemoryService.storeRepoMemory()` → `storeMemory({ scope: 'repository' })`
- `scope: 'user'` → `AgentMemoryService.storeUserMemory()` → `storeMemory({ scope: 'user' })`

The legacy `memory` tool also mirrors user `create` operations to CAPI via `_syncUserMemoryCAPI()` when the config flag is on. This is intentional one-way sync: the local `memory` tool remains the source of truth for in-session CRUD (view/str_replace/delete), and CAPI provides ambient context injection at conversation start via the `/prompt` response. The two systems are complementary.

### 2. Unified `/prompt` endpoint

**Before:** `MemoryContextPrompt` fetched repo memories via a separate `getRepoMemories()` call and formatted them locally. `MemoryInstructionsPrompt` used a large hardcoded `<repoMemoryInstructions>` block.

**After:** Both components read from `AgentMemoryService.getMemoryPrompt()`, which calls `fetchMemoryPrompts()` from `@github/copilot-agentic-tools/memory`. The response provides:

- `memoriesContext.prompt` — server-formatted memory context, rendered in `<memory_context>` (replaces `<repository_memories>` and the local `formatMemories()` method)
- `storeInstructions.prompt` — server-authored store guidance, rendered in `<storeMemoryInstructions>` (replaces the hardcoded `<repoMemoryInstructions>` block)
- `storeToolDefinition` — server-provided schema version, used by `StoreMemoryTool.alternativeDefinition()` to resolve the correct Zod schema via `resolveStoreMemorySchema(definitionVersion)` and convert it to JSON Schema at runtime

This means Copilot Memory's instructions and context evolve server-side without requiring a client ship.

### 3. Turn-0 tool availability fix

**Before:** `getAgentTools()` called `getCachedMemoryPrompt()` synchronously. The cache was never populated at that point (it was primed later in `buildPrompt()`), so `store_memory` was excluded from the tool list on every first turn.

**After:** `getAgentTools()` awaits `getMemoryPrompt()` on a cache miss when the config is enabled, priming the cache itself. `AgentMemoryCachePrimer.primeCache()` (called from `buildPrompt()` on turn 0) then hits the already-warm cache and is a no-op — no duplicate network call.

```typescript
// agentIntent.ts — getAgentTools()
const sessionId = getSessionIdFromToken(request.toolInvocationToken);
let cachedMemoryPrompt = agentMemoryService.getCachedMemoryPrompt(sessionId);
if (!cachedMemoryPrompt && configurationService.getExperimentBasedConfig(ConfigKey.CopilotMemoryEnabled, experimentationService)) {
    cachedMemoryPrompt = await agentMemoryService.getMemoryPrompt(undefined, sessionId);
}
allowTools[ToolName.StoreMemory] = !!cachedMemoryPrompt;
```

### 4. Local memory tool coexistence (`memoryContextPrompt.tsx`)

When both `MemoryToolEnabled` and `CopilotMemoryEnabled` are on, the model sees local repo files listed in `<repoMemory>` (from the local `memory` tool's storage). Previously, the instructions section only showed the CAPI description ("only `create` supported") with no guidance for those local files.

The scope description and guidelines now handle three cases explicitly:

| Flags | Scope description | Guidelines |
|-------|------------------|------------|
| Local only | Local repo storage, full CRUD | JSON format, `create`/`view` usage |
| CAPI only | CAPI-only, `create` supported | Server `storeMemoryInstructions` |
| **Both** | CAPI for new facts; use `memory` tool to view/edit existing local files | Local tool guidelines (`view`/`str_replace`/`delete`/`rename`) alongside server `storeMemoryInstructions` |

---

## Caching architecture

The `/prompt` response is cached per session in `AgentMemoryService._conversationMemoryCache: Map<sessionId, MemoryPromptResponse>`. An `_inflightPrompts` map prevents duplicate network calls if the same session triggers concurrent fetches. Cache is cleared when the session is disposed via `chatSessionService.onDidDisposeChatSession`.

**Turn-by-turn flow:**

```
Turn 0:
  getAgentTools()                              [awaited before buildPrompt]
    → getCachedMemoryPrompt(sessionId)         → miss
    → await getMemoryPrompt(sessionId)         → network fetch, cache populated
    → allowTools[StoreMemory] = true
  buildPrompt()
    → agentMemoryCachePrimer.primeCache(sessionId)  → cache hit, no-op
    → render()
      → MemoryContextPrompt                    → getCachedMemoryPrompt() → hit
      → MemoryInstructionsPrompt               → getCachedMemoryPrompt() → hit

Turn N (all hits):
  getAgentTools() → getCachedMemoryPrompt() → hit
  buildPrompt()   → render()               → hit

Session end:
  onDidDisposeChatSession(sessionId) → clearCache(sessionId)
```

**`AgentMemoryCachePrimer` is a cache-warming service, not a VS Code tool registrar.** It has no tool registration state to clean up when the config is disabled. Tool availability is re-evaluated each turn in `getAgentTools()` by reading the config flag at call time.

---

## Other changes

- **`getSessionIdFromToken` helper** — replaced two duplicate `(x as any)?.sessionResource` extractions in `agentIntent.ts` with a single typed helper, reducing the `as any` surface to one place
- **History gate fix** — `promptContext.history?.length === 0` would silently skip cache priming when `history` is `undefined`; corrected to `!promptContext.history || promptContext.history.length === 0`
- **Cache stampede guard** — `_inflightPrompts` map in `AgentMemoryService` deduplicates concurrent calls for the same session; previously two simultaneous callers would each make a full network round trip
- **Rename: `AgentMemoryToolRegistrar` → `AgentMemoryCachePrimer`** — the old name implied VS Code tool registration; the class only primes the prompt cache
- **Safe error handling** — `error.message` in two `memoryTool.tsx` catch blocks narrowed to `error instanceof Error ? error.message : String(error)`

---

## What is NOT in this PR

- **`vote_memory` / `vote_memory` tool**: Not implemented. The `IAgentMemoryService` interface has no `voteOnMemory` method and no `vote_memory` tool is registered anywhere.
- **CAPI read/list/delete**: The `memory` tool's `view`, `str_replace`, `delete`, and `rename` operations for user-scoped paths remain local-only. CAPI memories surface as ambient context at conversation start via `memoriesContext.prompt`; they are not a CRUD-accessible store in this PR.
- **`definitionVersion` runtime assertion**: If the shared package and server ever drift on schema version, the client will silently use the wrong schema. A runtime check is a recommended follow-up. 

---

## Tests

| File | What it covers |
|------|---------------|
| [agentMemoryService.spec.ts](src/extension/tools/common/test/agentMemoryService.spec.ts) | `storeRepoMemory`, `storeUserMemory`, `getMemoryPrompt` (cache hit / miss / clear / concurrent), `getBaseUrl`, config guard, error handling |
| [storeMemoryTool.spec.tsx](src/extension/tools/node/test/storeMemoryTool.spec.tsx) | Scope routing (repo / user / default), payload passthrough, failure handling, `alternativeDefinition` (empty cache, missing `storeToolDefinition`, schema override) |
| [memoryTool.spec.tsx](src/extension/tools/node/test/memoryTool.spec.tsx) | Repo CAPI path (`_dispatchRepoCAPI`), user CAPI sync (`_syncUserMemoryCAPI`) for JSON and plain-text formats, session-scope exclusion, telemetry |

Tested locally, and was able to store memories using `store_memory` 
Screenshot 2026-04-21 at 2 38 51 PM

Then after asking it to remember a few things about me ->

image

image


sneefyyy:

extensions/copilot/src/extension/tools/common/agentMemoryService.ts
```diff
@@ -150,17 +170,28 @@ export class AgentMemoryService extends Disposable implements IAgentMemoryServic
...
-	/**
-	 * Check if the chat.copilotMemory.enabled config is enabled.
-	 * Uses experiment-based configuration for gradual rollout.
-	 */
 	private isCAPIMemorySyncConfigEnabled(): boolean {
 		return this.configService.getExperimentBasedConfig(ConfigKey.CopilotMemoryEnabled, this.experimentationService);
 	}
 
+	private getBaseUrl(): string {
+		return this.capiClientService.capiPingURL.replace(/\/_ping$/, '');
```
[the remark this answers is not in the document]
sneefyyy: implement this change, but also change the package

extensions/copilot/src/extension/tools/node/voteMemoryTool.tsx
```diff
@@ -0,0 +1,81 @@
...
+				reason: {
+					type: 'string',
+					// eslint-disable-next-line local/no-unexternalized-strings
+					description: 'A clear and detailed explanation of the reason for your vote. Must be at least 2-3 sentences long.',
+				},
+			},
+			required: ['fact', 'direction', 'reason'],
+		},
+	};
+}
```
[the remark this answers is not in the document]
sneefyyy: perhaps outdated, check to see if it uses agentic tools

extensions/copilot/src/extension/tools/node/storeMemoryTool.tsx
```diff
@@ -0,0 +1,143 @@
...
+
+	return {
+		name,
+		displayName: 'Store Memory',
+		description,
+		tags: [],
+		source: undefined,
+		inputSchema: jsonSchema,
+	};
+}
```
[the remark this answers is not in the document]
sneefyyy: why are we converting zod schemas to json?

extensions/copilot/src/extension/tools/node/memoryContextPrompt.tsx
```diff
@@ -27,26 +31,68 @@ export class MemoryContextPrompt extends PromptElement<MemoryContextPromptProps>
...
+		let memoryPromptText: string | undefined;
+		let repoMemories: MemoryResponse[] | undefined;
+		if (enableCopilotMemory) {
+			const repoNwo = await this.getRepoNwo();
+			// Fetch prompt and register tools in parallel — both depend on the same /prompt call
+			// but registerMemoryTools() makes its own call so they can run concurrently.
+			const [promptResponse] = await Promise.all([
+				this.agentMemoryService.getMemoryPrompt(repoNwo),
+				this.agentMemoryToolRegistrar.registerMemoryTools(),
+			]);
```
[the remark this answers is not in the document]
sneefyyy: fix

extensions/copilot/src/extension/tools/node/memoryTool.tsx
```diff
@@ -342,6 +360,48 @@ export class MemoryTool implements ICopilotTool<MemoryToolParams> {
...
+					subject: pathHint,
+					fact: params.file_text,
+					citations: [],
+					reason: 'Stored from memory tool create command.',
+				};
+			}
+
+			const success = await this.agentMemoryService.storeUserMemory(entry);
+			if (success) {
+				return { text: `File created successfully at: ${params.path}`, outcome: 'success' };
```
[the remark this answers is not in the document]
sneefyyy: fix, and look into it

extensions/copilot/src/extension/tools/node/memoryTool.tsx
```diff
@@ -285,10 +286,23 @@ export class MemoryTool implements ICopilotTool<MemoryToolParams> {
...
+
+		// Route user-scope creates through CAPI when enabled
+		const capiEnabled = await this.agentMemoryService.checkMemoryEnabled();
+		if (capiEnabled && params.command === 'create') {
+			const result = await this._userCreate(params as ICreateParams);
+			this._sendLocalTelemetry(params.command, 'user', result.outcome, requestId, model);
+			return result.text;
+		}
+
+		// Fall back to local file-based user memory
```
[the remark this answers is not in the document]
sneefyyy: look into why the local one even exists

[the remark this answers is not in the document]
sneefyyy: and make sure it already existed before

sneefyyy:
look at fallbacks and make sure that no new fallbacks were created, sometimes it is just unreachable

extensions/copilot/src/extension/tools/node/agentMemoryToolRegistrar.ts
```diff
@@ -0,0 +1,105 @@
...
+			}
+			return undefined;
+		} catch {
+			return undefined;
+		}
+	}
+
+	async registerMemoryTools(): Promise<void> {
+		const enabled = this.configurationService.getExperimentBasedConfig(ConfigKey.CopilotMemoryEnabled, this.experimentationService);
+		if (!enabled) {
```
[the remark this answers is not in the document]
sneefyyy: makes sense

extensions/copilot/src/extension/tools/common/agentMemoryService.ts
```diff
@@ -14,34 +25,47 @@ import { IWorkspaceService } from '../../../platform/workspace/common/workspaceS
...
+export type { MemoryPromptResponse, MemoryResponse as RepoMemoryEntry };
+
 /**
- * Repository memory entry format aligned with CAPI service contract.
- * Supports both new format (citations as string[]) and legacy format (citations as string).
+ * User memory entry format for user-scoped memories.
+ * @deprecated Use StoreMemoryRequest from @github/copilot-agentic-tools/memory directly.
  */
-export interface RepoMemoryEntry {
+export interface UserMemoryEntry {
```
zutshisunakshi: Once Audree's PR is merged and used here, we should check on interface definations for combined user and repo entry - making the swap like this does not look right.

extensions/copilot/src/extension/tools/node/voteMemoryTool.tsx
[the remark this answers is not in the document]
sneefyyy: The server's `/prompt` response (`ToolDefinition`) only provides `name`, `description`, and `definitionVersion`. It doesn't include an input schema. The input schema here mirrors `voteMemoryInputSchema` from `@github/copilot-agentic-tools/memory`, which is the source of truth. The duplication is a known trade-off since VS Code's `LanguageModelToolDefinition.inputSchema` requires a plain JSON Schema object and we don't have `zod-to-json-schema` available to derive it automatically.

extensions/copilot/src/extension/tools/node/storeMemoryTool.tsx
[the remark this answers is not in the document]
sneefyyy: Same thing as above, needs to be in JSON

extensions/copilot/src/extension/tools/common/agentMemoryService.ts
```diff
@@ -14,34 +25,47 @@ import { IWorkspaceService } from '../../../platform/workspace/common/workspaceS
...
 	subject: string;
 	fact: string;
 	citations?: string | string[];
 	reason?: string;
-	category?: string;
 }
 
 /**
- * Type guard to validate if an object is a valid RepoMemoryEntry.
```
bhavyaus, line 30: please undo the formatting and use the repo defaults

extensions/copilot/package.json
```diff
@@ -6349,6 +6349,7 @@
 		"@anthropic-ai/sdk": "^0.82.0",
 		"@github/blackbird-external-ingest-utils": "^0.3.0",
 		"@github/copilot": "^1.0.24",
+		"@github/copilot-agentic-tools": "file:../../../copilot-agentic-tools",
```
bhavyaus: this is incorrect

extensions/copilot/src/extension/tools/node/voteMemoryTool.tsx
```diff
@@ -0,0 +1,81 @@
...
+import { CancellationToken } from '../../../util/vs/base/common/cancellation';
+import { LanguageModelTextPart, LanguageModelToolResult } from '../../../vscodeTypes';
+import { IAgentMemoryService } from '../common/agentMemoryService';
+import { ICopilotModelSpecificTool } from '../common/toolsRegistry';
+
+/**
+ * Implements the vote_memory tool using @github/copilot-agentic-tools/memory.
+ * Only registered when voteToolDefinition is present in the /prompt response.
+ */
+export class VoteMemoryTool implements ICopilotModelSpecificTool<VoteMemoryInput> {
```
bhavyaus: what is this tool helping with? we should not introduce this if copilot memory is not enabled.

bhavyaus (changes requested):

extensions/copilot/package.json
sneefyyy: stand in, until Audree's PR is merged, I can remove it

extensions/copilot/src/extension/tools/node/voteMemoryTool.tsx
sneefyyy: was a fragment of me porting over from an old repo and didn't specify to copilot correctly, removing it

extensions/copilot/src/extension/tools/node/memoryContextPrompt.tsx
```diff
@@ -197,16 +220,12 @@ export class MemoryContextPrompt extends PromptElement<MemoryContextPromptProps>
...
 
-			// Format citations (handle both string and string[] formats)
-			if (m.citations) {
-				const citationsArray = normalizeCitations(m.citations) ?? [];
-				if (citationsArray.length > 0) {
-					lines.push(`- Citations: ${citationsArray.join(', ')}`);
-				}
+			if (m.citations.length > 0) {
+				lines.push(`- Citations: ${m.citations.join(', ')}`);
 			}
```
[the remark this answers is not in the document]
sneefyyy: believe this should be fixed, double check

extensions/copilot/src/extension/tools/common/agentMemoryService.ts
```diff
@@ -61,45 +88,47 @@ export function isRepoMemoryEntry(obj: unknown): obj is RepoMemoryEntry {
 	return true;
```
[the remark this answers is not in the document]
sneefyyy: I am not sure if this needs to be changed

extensions/copilot/src/extension/tools/common/agentMemoryService.ts
```diff
@@ -219,68 +256,38 @@ export class AgentMemoryService extends Disposable implements IAgentMemoryServic
...
 			}
 
-			// Get OAuth token for API call
-			const session = await this.authenticationService.getGitHubSession('any', { silent: true });
-			if (!session) {
+			const token = await this.getToken();
+			if (!token) {
 				this.logService.warn('[AgentMemoryService] No GitHub session available for fetching memories');
 				return undefined;
 			}
```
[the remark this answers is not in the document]
sneefyyy: The double lookup is a pre existing pattern in this file that predates this PR

extensions/copilot/src/extension/tools/common/agentMemoryService.ts
audreee: Agreed, this doesn't look right, especially since there shouldn't be a difference between UserMemoryEntry and RepoMemoryEntry (they contain the same fields).

extensions/copilot/src/extension/tools/common/agentMemoryService.ts
```diff
@@ -3,34 +3,61 @@
...
 import { getGithubRepoIdFromFetchUrl, getOrderedRemoteUrlsFromContext, IGitService, toGithubNwo } from '../../../platform/git/common/gitService';
 import { ILogService } from '../../../platform/log/common/logService';
+import { IFetcherService } from '../../../platform/networking/common/fetcherService';
 import { IExperimentationService } from '../../../platform/telemetry/common/nullExperimentationService';
 import { IWorkspaceService } from '../../../platform/workspace/common/workspaceService';
 import { createServiceIdentifier } from '../../../util/common/services';
 import { Disposable } from '../../../util/vs/base/common/lifecycle';
 
+// Re-export package types that callers depend on
+export type { MemoryPromptResponse, MemoryResponse as RepoMemoryEntry };
```
audreee: Why the re-export?

extensions/copilot/src/extension/tools/common/agentMemoryService.ts
```diff
@@ -3,34 +3,61 @@
...
 	citations?: string | string[];
 	reason?: string;
-	category?: string;
+}
+
+/**
+ * Normalize citations field to string[] format.
+ * Handles backward compatibility for legacy string format.
+ */
+export function normalizeCitations(citations: string | string[] | undefined): string[] | undefined {
```
audreee: Do we actually need this? I think the "legacy format" was just the manual definitions here, but now that we're importing from the package and the package owns the API contract, there's no path where a string-typed citation would flow through

extensions/copilot/src/extension/tools/common/agentMemoryService.ts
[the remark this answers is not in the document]
audreee: Before this PR, VSCode was calling the CAPI API directly and getting back raw JSON. This function was a runtime check that said, "does this blob of JSON actually look like a memory object?" Now that we're using the package, it makes the API call and returns a typed `MemoryResponse`, and that validation is happening within the package. So we shouldn't need this

extensions/copilot/src/extension/tools/common/agentMemoryService.ts
```diff
@@ -61,45 +88,47 @@ export function isRepoMemoryEntry(obj: unknown): obj is RepoMemoryEntry {
...
+	storeUserMemory(memory: StoreMemoryRequest): Promise<boolean>;
+
+
+	/**
+	 * Fetch the unified memory prompt from the /prompt endpoint.
+	 * Returns the full MemoryPromptResponse including storeToolDefinition and voteToolDefinition,
+	 * or undefined on failure.
+	 */
+	getMemoryPrompt(repoNwo?: string): Promise<MemoryPromptResponse | undefined>;
 }
```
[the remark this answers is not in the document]
audreee: Just update the PR description here 👍

extensions/copilot/src/extension/tools/common/agentMemoryService.ts
```diff
@@ -205,12 +244,10 @@ export class AgentMemoryService extends Disposable implements IAgentMemoryServic
 		}
 	}
 
-	async getRepoMemories(limit: number = 10): Promise<RepoMemoryEntry[] | undefined> {
+	async getRepoMemories(limit: number = 10): Promise<MemoryResponse[] | undefined> {
```
audreee: Curious why the number is 10 here, since I think we're using a default of 20 elsewhere. But also, I think we can rely on the server's default of 20... let's just let the default pass through and mark limit as optional

extensions/copilot/src/extension/tools/common/agentMemoryService.ts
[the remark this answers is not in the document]
audreee: That's fair to reject as out-of-scope, but we should open up a follow-up issue with ourselves to clean this up then. We might be able to delegate that enablement check to the package so that its functions return appropriate errors for disabled repos.

extensions/copilot/src/extension/tools/common/agentMemoryService.ts
```diff
@@ -289,45 +296,115 @@ export class AgentMemoryService extends Disposable implements IAgentMemoryServic
...
-			const citations = normalizeCitations(memory.citations) ?? [];
-
-			// Get OAuth token for API call
-			const session = await this.authenticationService.getGitHubSession('any', { silent: true });
-			if (!session) {
+			const token = await this.getToken();
+			if (!token) {
 				this.logService.warn('[AgentMemoryService] No GitHub session available for storing memory');
 				return false;
 			}
```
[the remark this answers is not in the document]
audreee: Can we hoist `repoNwo` and `token` above the `checkMemoryEnabled` call so that they're resolved once and re-used?

extensions/copilot/src/extension/tools/node/agentMemoryCachePrimer.ts
```diff
@@ -0,0 +1,75 @@
...
+export interface IAgentMemoryToolRegistrar {
+	readonly _serviceBrand: undefined;
+	/**
+	 * Register store_memory (and optionally vote_memory) as model-specific tools using an
+	 * already-fetched prompt response. When called without a response, fetches the /prompt
+	 * endpoint itself. Safe to call multiple times — re-registers on each call to pick up
+	 * updated tool definitions from the server.
+	 */
+	registerMemoryTools(promptResponse?: MemoryPromptResponse): Promise<void>;
+}
```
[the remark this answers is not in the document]
audreee: Same thing--let's just update the PR description

extensions/copilot/src/extension/tools/node/memoryContextPrompt.tsx
```diff
@@ -40,13 +43,28 @@ export class MemoryContextPrompt extends PromptElement<MemoryContextPromptProps>
...
-			return null;
+		// When CAPI memory is enabled, fetch from the unified /prompt endpoint
+		let memoryPromptText: string | undefined;
+		let repoMemories: MemoryResponse[] | undefined;
+		if (enableCopilotMemory) {
+			const repoNwo = await this.agentMemoryService.getRepoNwo();
+			// Fetch once and pass the response to registerMemoryTools so it can reuse it,
+			// avoiding a redundant /prompt call.
+			const promptResponse = await this.agentMemoryService.getMemoryPrompt(repoNwo);
+			await this.agentMemoryToolRegistrar.registerMemoryTools(promptResponse);
```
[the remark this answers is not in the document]
audreee: This is a good suggestion

extensions/copilot/src/extension/tools/node/storeMemoryTool.tsx
```diff
@@ -0,0 +1,143 @@
...
+			const scope = 'scope' in input ? input.scope : 'repo';
+			let success: boolean;
+			if (scope === 'user') {
+				success = await this.agentMemoryService.storeUserMemory(memory);
+			} else {
+				success = await this.agentMemoryService.storeRepoMemory(memory);
+			}
+
+			if (success) {
+				this.logService.info(`[StoreMemoryTool] Stored memory: ${input.subject}`);
```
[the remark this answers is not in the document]
audreee: I think this is okay, assuming that these logs are local. Curious what a VSCode eng thinks here

extensions/copilot/src/extension/tools/node/storeMemoryTool.tsx
```diff
@@ -0,0 +1,143 @@
...
+				return new LanguageModelToolResult([new LanguageModelTextPart('Memory stored successfully.')]);
+			} else {
+				return new LanguageModelToolResult([new LanguageModelTextPart('Failed to store memory. Copilot Memory may not be enabled for this repository.')]);
+			}
```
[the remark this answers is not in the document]
audreee: Yep, let's use scope-neutral language

extensions/copilot/src/extension/tools/node/storeMemoryTool.tsx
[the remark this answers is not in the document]
audreee: Server doesn't provide a JSON schema, so this isn't actionable here right no

extensions/copilot/package.json
```diff
@@ -6398,6 +6398,7 @@
 		"@anthropic-ai/sdk": "^0.82.0",
 		"@github/blackbird-external-ingest-utils": "^0.3.0",
 		"@github/copilot": "^1.0.28",
+		"@github/copilot-agentic-tools": "^1.0.0",
```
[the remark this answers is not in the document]
audreee: Update description 👍

extensions/copilot/src/extension/tools/node/agentMemoryToolRegistrar.ts
```diff
@@ -0,0 +1,74 @@
...
+			const repoNwo = await this.agentMemoryService.getRepoNwo();
+			promptResponse = await this.agentMemoryService.getMemoryPrompt(repoNwo);
+		}
+		if (!promptResponse) {
+			this.logService.warn('[AgentMemoryToolRegistrar] Could not fetch memory prompt — skipping tool registration');
+			return;
+		}
+
+		// Dispose any previously registered memory tools before re-registering
+		this._registrations.clear();
```
audreee: I wonder if we should be clearing previously-registered memory tools before doing anything to prevent us from the case in which we have stale tools with a stale schema 🤔

extensions/copilot/src/extension/tools/common/agentMemoryService.ts
```diff
@@ -150,17 +175,41 @@ export class AgentMemoryService extends Disposable implements IAgentMemoryServic
...
+	}
+
+	private makeFetch(): MemoryFetchFn {
+		const fetcher = this.fetcherService;
+		return (url: string, init?: { method?: string; headers?: any; body?: any }) => fetcher.fetch(url, {
+			callSite: 'AgentMemoryService',
+			method: (init?.method as 'GET' | 'POST' | 'PUT') ?? 'GET',
+			headers: init?.headers as Record<string, string> | undefined,
+			body: init?.body as string | undefined,
+		}) as unknown as Promise<any>;
```
audreee: instead of unknown type, let's cast `as MemoryFetchFn`. That way, if it doesn't return a Response-compatible object, Typescript will flag it

extensions/copilot/src/extension/tools/common/agentMemoryService.ts
```diff
@@ -170,33 +219,23 @@ export class AgentMemoryService extends Disposable implements IAgentMemoryServic
...
-			const response = await this.capiClientService.makeRequest<Response>({
-				method: 'GET',
-				headers: {
-					'Authorization': `Bearer ${session.accessToken}`
-				}
-			}, {
-				type: RequestType.CopilotAgentMemory,
-				repo: repoNwo,
-				action: 'enabled'
+			const [owner, repo] = repoNwo.split('/');
```
audreee: We need to support checking enablement for the case where there's just a user but no repository context. We'll need to update this to support the user-only path

extensions/copilot/src/extension/tools/node/memoryContextPrompt.tsx
audreee: I don't think we want to be registering memory tools here, since it's within a `render()`. I think we want to be registering the tools wherever the session is initialized

audreee (commented):
Great start on such a large-scoped PR! Left some comments! Also, we're missing some test coverage:
  - StoreMemoryTool — scope routing (repo vs user), error handling
  - AgentMemoryToolRegistrar — registration, cleanup on failure, stale tool prevention
  - storeUserMemory() — success/failure paths
  - getMemoryPrompt() — repo-scoped vs user-scoped fallback, failure case
  - The /prompt-driven rendering path in MemoryContextPrompt

extensions/copilot/src/extension/codeBlocks/node/codeBlockProcessor.ts
```diff
@@ -125,7 +125,7 @@ export class CodeBlockTrackingChatResponseStream implements ChatResponseStream {
 	workspaceEdit = this.forward(this._wrapped.workspaceEdit?.bind(this._wrapped) || (() => { }));
 	confirmation = this.forward(this._wrapped.confirmation.bind(this._wrapped));
 	warning = this.forward(this._wrapped.warning.bind(this._wrapped));
-	info = this.forward(this._wrapped.info.bind(this._wrapped));
+	info = this.forward(this._wrapped.info?.bind(this._wrapped) || (() => { }));
```
audreee: Why do we need this?

extensions/copilot/src/extension/intents/node/agentIntent.ts
```diff
@@ -144,6 +148,13 @@ export const getAgentTools = async (accessor: ServicesAccessor, request: vscode.
 		allowTools[ToolName.MultiReplaceString] = true;
 	}
 
+	allowTools[CUSTOM_TOOL_SEARCH_NAME] = !!model.supportsToolSearch;
```
audreee: Similar to above: how does this relate to our Memory tool changes? Let's keep the scope of the code changes very precise and focused to only our requirements

extensions/copilot/src/extension/intents/node/agentIntent.ts
```diff
@@ -144,6 +148,13 @@ export const getAgentTools = async (accessor: ServicesAccessor, request: vscode.
...
+
+	if (!agentMemoryService.getCachedMemoryPrompt() && configurationService.getExperimentBasedConfig(ConfigKey.CopilotMemoryEnabled, experimentationService)) {
+		await agentMemoryService.getMemoryPrompt();
+	}
+	allowTools[ToolName.StoreMemory] = !!agentMemoryService.getCachedMemoryPrompt();
```
audreee: This is clever! So if the /prompt call failed, returned nothing, or hasn't run yet, then the store memory tool is disabled. One issue I can foresee is with debugging. If the tool isn't showing up, how will we know if it's because memory is disabled, the prompt fetch failed, or the cache wasn't primed yet? If we stick with this pattern, then we should consider adding some logging to help distinguish why the tool isn't available

extensions/copilot/src/extension/intents/node/agentIntent.ts
```diff
@@ -144,6 +148,13 @@ export const getAgentTools = async (accessor: ServicesAccessor, request: vscode.
 		allowTools[ToolName.MultiReplaceString] = true;
 	}
 
+	allowTools[CUSTOM_TOOL_SEARCH_NAME] = !!model.supportsToolSearch;
+
+	if (!agentMemoryService.getCachedMemoryPrompt() && configurationService.getExperimentBasedConfig(ConfigKey.CopilotMemoryEnabled, experimentationService)) {
```
audreee: I think we want to remove this `if` block.  `getAgentTools()` runs every turn, not just turn 0. So on the first turn, both paths will execute:
   1. `buildPrompt()` → `registerMemoryTools()` → calls `getMemoryPrompt()` internally
   2. `getAgentTools()` → checks cache → if empty, calls `getMemoryPrompt()` directly

This means we have a race condition in which, if the cache is empty, we're calling `getMemoryPrompt()` twice. By removing this fallback, on Turn 0, we have `registerMemoryTools()` in `buildPrompt()` as the only fetch path

extensions/copilot/src/extension/tools/common/agentMemoryService.ts
```diff
@@ -150,184 +133,259 @@ export class AgentMemoryService extends Disposable implements IAgentMemoryServic
...
+	}
+
+	private makeLogger() {
+		return {
+			info: (msg: string) => this.logService.info(`[AgentMemoryService] ${msg}`),
+			error: (msg: string) => this.logService.error(`[AgentMemoryService] ${msg}`),
+		};
+	}
+
 	async checkMemoryEnabled(): Promise<boolean> {
```
audreee: This appears to be dead code. The old callers in `storeRepoMemory/getRepoMemories` were removed, and now you're using cache existence as the enablement check instead. Additionally, this implementation calls `fetchMemoryPrompts()` and discards the response, which is wasteful if it were ever called.

extensions/copilot/package.json
```diff
@@ -1080,7 +1080,7 @@
 				"displayName": "Memory",
 				"toolReferenceName": "memory",
 				"userDescription": "Manage persistent memory across conversations",
-				"when": "config.github.copilot.chat.tools.memory.enabled",
+				"when": "config.github.copilot.chat.tools.memory.enabled && !config.github.copilot.chat.copilotMemory.enabled",
```
audreee: We need to dive deeper into this because we're introducing a new behavior in which our "Copilot Memory" tool will now be mutually exclusive from the "local" memory tool.

Before this PR, both tools co-existed. If they were enabled, there would be tools for "local memory" (exposing tools that write memories to local files) _and_ our Agentic Memory tool. The model could then choose which tool to call. The twist is that there was some element of "syncing"--when the local memory tool's create command targeted /memories/repo/, it routed to agentMemoryService.storeRepoMemory() (CAPI). So repo memories written via the local tool did go to our cloud service. But, user and session memories written via the local tool stayed as local files only (this makes sense, since we didn't support user memories). We need to get more information about who owns + maintains this "local" memory. I'm hesitant to disable/hide another team's tooling in favor of ours. Or, another option to consider: let's follow the established "sync" pattern that's set for repo memories. So if the model chooses the "local" create memory tool for user memory, then we also send that memory to our service as a user memory. Either way, I think we should consult the owning team here

extensions/copilot/src/extension/tools/common/agentMemoryService.ts
```diff
@@ -150,184 +133,259 @@ export class AgentMemoryService extends Disposable implements IAgentMemoryServic
...
 			}
 
-			// Get OAuth token for API call
-			const session = await this.authenticationService.getGitHubSession('any', { silent: true });
-			if (!session) {
-				this.logService.warn('[AgentMemoryService] No GitHub session available for fetching memories');
-				return undefined;
+			const token = await this.getToken();
+			if (!token) {
+				this.logService.warn('[AgentMemoryService] No GitHub session available for storing memory');
```
audreee: I see that you're re-using the same messaging, but I worry that "session" has become an overloaded term now. Even though this is referring to a token session, it's somewhat confusing with agent sessions. I wonder if it's worth updating this message to be explicit: "No GitHub authentication token available for storing memory"

extensions/copilot/src/extension/tools/common/agentMemoryService.ts
```diff
@@ -150,184 +133,259 @@ export class AgentMemoryService extends Disposable implements IAgentMemoryServic
...
+				agent: 'vscode',
+				logger: this.makeLogger(),
+			};
+
+			const result = await storeMemory(memory, options);
+			if (!result.success) {
+				this.logService.warn(`[AgentMemoryService] Failed to store repo memory: ${result.error}`);
+			} else {
+				this.logService.debug(`[AgentMemoryService] Successfully stored repo memory`);
+				// Note: Cache will be cleared when conversation ends, ensuring fresh data for next conversation
```
audreee: this comment doesn't provide much value

extensions/copilot/src/extension/tools/common/agentMemoryService.ts
```diff
@@ -150,184 +133,259 @@ export class AgentMemoryService extends Disposable implements IAgentMemoryServic
...
+			};
+
+			// Always fetch user-scoped memories, and repo-scoped if a repo is available.
+			// Fetch both in parallel and merge so the model sees all relevant context.
+			const userOptions: MemoryApiOptions = { scope: 'user', ...baseOptions };
+			const repoOptions: MemoryApiOptions | undefined = resolvedRepoNwo
+				? (() => { const [owner, repo] = resolvedRepoNwo.split('/'); return { scope: 'repository' as const, owner, repo, ...baseOptions }; })()
+				: undefined;
+
+			const [userResult, repoResult] = await Promise.allSettled([
```
audreee: We shouldn't be making two separate API calls for user results and repo results. Both user + repo memories are included in the prompt returned by the server from the `/v0/memory/prompt` endpoint that takes an optional query parameter for repos `?repo=<owner/repo>`. So instead of making two requests and resolving them, we should just make one request to the endpoint, adding the repo if applicable.

extensions/copilot/src/extension/tools/common/agentMemoryService.ts
```diff
@@ -150,184 +133,259 @@ export class AgentMemoryService extends Disposable implements IAgentMemoryServic
...
+				this.logService.warn(`[AgentMemoryService] Failed to fetch user memory prompt: ${userResult.reason}`);
+			}
+			if (repoResult.status === 'rejected') {
+				this.logService.warn(`[AgentMemoryService] Failed to fetch repo memory prompt: ${repoResult.reason}`);
+			}
+
+			const userResponse = userResult.status === 'fulfilled' ? userResult.value : undefined;
+			const repoResponse = repoResult.status === 'fulfilled' ? repoResult.value : undefined;
+
+			const response = this.mergeMemoryPromptResponses(userResponse, repoResponse);
```
audreee: Again, we should only have one prompt response

extensions/copilot/src/extension/tools/node/memoryContextPrompt.tsx
```diff
@@ -64,11 +68,13 @@ export class MemoryContextPrompt extends PromptElement<MemoryContextPromptProps>
 					<Tag name='userMemory'>
 						{userMemoryContent
 							? <>The following are your persistent user memory notes. These persist across all workspaces and conversations.<br /><br />{userMemoryContent}</>
-							: <>No user preferences or notes saved yet. Use the {ToolName.Memory} tool to store persistent notes under /memories/.</>
+							: enableCopilotMemory
```
audreee, line 71: Again, rather than suppressing the local memory tool's prompt sections when copilotMemory.enabled is true, these should coexist. The CAPI memory context and store instructions should be additive — appended alongside the existing local memory prompts, not replacing them.

extensions/copilot/src/extension/tools/node/memoryContextPrompt.tsx
```diff
@@ -250,73 +228,55 @@ export class MemoryContextPrompt extends PromptElement<MemoryContextPromptProps>
...
-						- Are independent of changes you are making as part of your current task, and will remain relevant if your current code isn't merged<br />
-						- Are unlikely to change over time<br />
-						- Cannot always be inferred from a limited code sample<br />
-						- Contain no secrets or sensitive data<br />
-					</Tag>
-					Always include the reason and citations fields.<br />
-					Before storing, ask yourself: Will this help with future coding or code review tasks across the repository? If unsure, skip storing it.<br />
-					Note: Only `create` is supported for `/memories/repo/` paths.<br />
-					If the user asks how to view or manage their repo memories refer them to https://docs.github.com/en/copilot/how-tos/use-copilot-agents/copilot-memory.<br />
+			{storeInstructionsPrompt && (
```
audreee, line 289: Same thing--let's leave the existing memory tool instructions in place for now

extensions/copilot/src/extension/tools/node/storeMemoryTool.tsx
```diff
@@ -0,0 +1,85 @@
...
+export class StoreMemoryTool implements ICopilotTool<StoreMemoryParams> {
+	static readonly toolName = ToolName.StoreMemory;
+
+	constructor(
+		@IAgentMemoryService private readonly agentMemoryService: IAgentMemoryService,
+		@ILogService private readonly logService: ILogService,
+	) { }
+
+	alternativeDefinition(tool: vscode.LanguageModelToolInformation): vscode.LanguageModelToolInformation {
+		const cached = this.agentMemoryService.getCachedMemoryPrompt();
```
audreee, line 39: we're calling `getCachedMemoryPrompt` here without a sessionID, so it will return the first arbitrary entry from the Map. If there are multiple conversations, this could return the wrong cached response

extensions/copilot/package.json
```diff
@@ -6198,6 +6245,7 @@
 		"chatPromptFiles": [
 			{
 				"path": "./assets/prompts/plan.prompt.md",
+				"when": "chatSessionType == local",
 				"sessionTypes": [
```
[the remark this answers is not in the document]
audreee: Is this related to our changes?

extensions/copilot/pr-description.md
```diff
@@ -0,0 +1,107 @@
+# PR: Copilot Memory — Shared Package Migration + User-Scoped Memories
```
audreee: pretty sure we don't want this committed

extensions/copilot/src/extension/codeBlocks/node/codeBlockProcessor.ts
sneefyyy: It is for backwards compatibility. It wasn't working on my VS Code, I can remove it, if you recommend it.

audreee: Is it related to our changes? I'd say if it's not directly related to or necessary for shipping the user memory capability, I'd remove it from the PR. My opinion is that, especially when working in code that we're not the owners of, we should have the lightest touch possible

extensions/copilot/src/extension/intents/node/agentIntent.ts
```diff
@@ -144,6 +149,12 @@ export const getAgentTools = async (accessor: ServicesAccessor, request: vscode.
 		allowTools[ToolName.MultiReplaceString] = true;
 	}
 
+	const cachedMemoryPrompt = agentMemoryService.getCachedMemoryPrompt();
+	if (!cachedMemoryPrompt && configurationService.getExperimentBasedConfig(ConfigKey.CopilotMemoryEnabled, experimentationService)) {
+		logService.debug('[getAgentTools] CopilotMemory is enabled but cache is not primed yet — StoreMemory tool will be disabled this turn');
```
audreee: So what's happening here is a timing issue: `getAgentTools()` runs before `buildPrompt()`, but the cache doesn't get primed until `buildPrompt()` calls `registerMemoryTools()` on turn 0 (when history.length === 0). That means on the
very first turn of every conversation, `getCachedMemoryPrompt()` will always return undefined here, and `store_memory` will be excluded from the tool list.

As a general pattern, logging an expected failure instead of fixing it is an anti-pattern. If we know the cache won't be ready here, the right move is to fix the ordering, not log around it.

A couple of ways to fix this:
 - Prime the cache earlier (e.g., at session creation) so it's ready by the time `getAgentTools()` runs
 - Allow the tool unconditionally when the config flag is on and have it fetch on-demand if the cache misses
 - Await priming here if the cache isn't populated yet

extensions/copilot/src/extension/intents/node/agentIntent.ts
```diff
@@ -144,6 +149,12 @@ export const getAgentTools = async (accessor: ServicesAccessor, request: vscode.
 		allowTools[ToolName.MultiReplaceString] = true;
 	}
 
+	const cachedMemoryPrompt = agentMemoryService.getCachedMemoryPrompt();
```
audreee: See comment below, that this might not be the right approach. But, more generally, `getCachedMemoryPrompt()` is being called here without a `sessionId`, which means it falls back to returning the first entry in the Map — that's non-deterministic when multiple sessions are active and could return the prompt/schema from a completely different conversation. Since request is available in this function, you should be able to extract and pass a `sessionId` here (and everywhere you're calling this function).

extensions/copilot/src/extension/tools/common/test/agentMemoryService.spec.ts
```diff
@@ -1,129 +0,0 @@
-/*---------------------------------------------------------------------------------------------
- *  Copyright (c) Microsoft Corporation. All rights reserved.
- *  Licensed under the MIT License. See License.txt in the project root for license information.
- *--------------------------------------------------------------------------------------------*/
-
-import { describe, expect, it } from 'vitest';
-import { isRepoMemoryEntry, normalizeCitations } from '../agentMemoryService';
-
-describe('AgentMemoryService', () => {
```
audreee, line 9: The old test file covered `isRepoMemoryEntry` and `normalizeCitations`, and since those helpers are gone, deleting their tests makes sense. But the new service has significant new behavior that needs coverage:

   - Cache lifecycle: `getMemoryPrompt()` cache hit/miss, session-scoped keying, `getCachedMemoryPrompt()` with and without sessionId
   - Cache cleanup: `clearCache(sessionId)` via `onDidDisposeChatSession`
   - `storeUserMemory()`/ `storeRepoMemory()`: success, failure, and missing auth token paths

extensions/copilot/src/extension/tools/common/agentMemoryService.ts
```diff
@@ -170,164 +162,142 @@ export class AgentMemoryService extends Disposable implements IAgentMemoryServic
...
-			if (!data || !Array.isArray(data)) {
-				return undefined;
+			const options = {
+				scope: 'user' as const,
+				token,
+				integrationId: INTEGRATION_ID,
+				baseUrl: this.getBaseUrl(),
+				agent: 'vscode',
+				logger: this.makeLogger(),
+			};
```
audreee, line 214: Can we add the base_model if available here and for storing repo memories? See https://github.com/github/copilot-agentic-tools/blob/main/src/memory/api.ts#L241)

extensions/copilot/src/extension/tools/node/agentMemoryToolRegistrar.ts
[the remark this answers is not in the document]
audreee: We shoudl still consider this

extensions/copilot/src/extension/tools/node/memoryContextPrompt.tsx
```diff
@@ -288,35 +270,11 @@ export class MemoryInstructionsPrompt extends PromptElement<BasePromptElementPro
...
+				<Tag name='storeMemoryInstructions'>
```
audreee, line 290: You're replacing the entire hardcoded repo memory instructions block with the server-provided `storeInstructionsPrompt`. The issue is that the server has no awareness of the local memory tool, so when `enableCopilotMemory` is true, all guidance about local repo memory files (/memories/repo/) disappears from the prompt.

The `MemoryContextPrompt` above still shows local repo files in context (the L85 fix is good!), but without any instructions here, the model sees those files with no guidance on what they are or how to interact with them.

Could we keep the local repo memory instructions alongside the server-provided ones when both flags are on? Something like rendering the local instructions when enableMemoryTool is true, and the server instructions when enableCopilotMemory is true, since they're not mutually exclusive.

extensions/copilot/src/extension/tools/common/toolsService.ts
```diff
@@ -202,7 +202,7 @@ export abstract class BaseToolsService extends Disposable implements IToolsServi
 	abstract getEnabledTools(request: vscode.ChatRequest, endpoint: IChatEndpoint, filter?: (tool: vscode.LanguageModelToolInformation) => boolean | undefined): vscode.LanguageModelToolInformation[];
 
 	constructor(
-		@ILogService private readonly logService: ILogService
+		@ILogService protected readonly logService: ILogService
```
zutshisunakshi: revert this?

PR #309964, opened 2026-04-14
chat: animate response copy button feedback

hawkticehurst:
## Summary
- add a custom footer copy action view item that swaps from the copy glyph to a temporary checkmark after copy succeeds
- animate the icon transition with short opacity/scale motion and restore the copy icon after a brief delay
- trigger the same feedback for both mouse and keyboard activation, with an ARIA status update and reduced-motion handling

https://github.com/user-attachments/assets/9bfb906c-a3e3-492d-8774-e9a064267a36

## Why
The response copy button in the sessions/agents chat worked, but it had no visible confirmation after activation. This adds lightweight feedback without changing the action behavior.

## Notes
- the implementation lives in the shared chat footer action rendering used by the sessions app
- validation included compile, hygiene, and layers checks during development

### DIFF
--- a/src/vs/workbench/contrib/chat/browser/actions/chatCopyActions.ts
+++ b/src/vs/workbench/contrib/chat/browser/actions/chatCopyActions.ts
@@ -3,18 +3,136 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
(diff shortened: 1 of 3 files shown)


justschen:
ope sorry, i was going to take a look at this, but thanks for hopping on it! 

note that we also don't have an animation in local chat, maybe we can extend the logic (it's the same toolbar etc)

justschen (approved):
looks awesome! thanks!!

PR #309998, opened 2026-04-14
sessions: add accent-tinted shell background

hawkticehurst:
## Summary
- add a single root-level gradient layer for the sessions workbench shell
- keep chat, files/changes, and terminal surfaces solid while letting the titlebar and sidebar read through the shared shell background
- tune the accent tint per theme type and disable it for high-contrast themes

Screenshot 2026-04-14 at 6 49 07 PM
Screenshot 2026-04-14 at 6 49 15 PM

## Why
This experiments with a more modern shell treatment for the sessions/agents app while keeping the inner working surfaces stable and preserving accessibility expectations for high-contrast themes.

## Notes for reviewers
- the accent tint is sourced from `button.background` as the "most semantic" token we have for getting a theme's accent color
- the gradient is implemented as a root pseudo-element so it stays visually continuous across the app chrome
- `src/vs/sessions/LAYOUT.md` has been updated to document the new shell-background approach


DonJayamanne (dismissed):

hawkticehurst:
Addressed the Copilot review feedback in a follow-up commit:

- scoped the `agent-sessions-viewpane` transparency override to the sessions workbench
- added an `@supports (background: color-mix(...))` guard with a solid shell fallback for unsupported environments
- simplified the transparent shell overrides by removing redundant `background-color` declarations where they were not needed

Also kept the latest visual tuning for the light/dark shell gradients.

Yoyokrazy (dismissed):

hawkticehurst:
Added an opt-in experimental gate for the shell gradient follow-up.

- registered `sessions.experimental.shellGradientBackground` as an application-scoped experimental setting
- scoped the shell gradient and transparency overrides behind a dedicated workbench class
- left the default Sessions window appearance unchanged unless the setting is enabled

This is now included in the PR head commit.

hawkticehurst:
Follow-up fix pushed for the experimental gate regression.

The shell-gradient opt-in now preserves the `origin/main` sessions styling when the setting is off:
- restored the upstream sessions viewpane selector so the sidebar header/search/actions layout stays unchanged
- restored the upstream sidebar/grid background rules for the non-experimental path
- kept only the additive transparency overrides behind `.experimental-shell-gradient-background`

This updates the PR head with the regression fix.

roblourens (approved):

PR #310055, opened 2026-04-15
Adopt CodeAction type for built-in css server

yogeshwaran-c:
### What this PR does

The CSS language server already returns proper `CodeAction` objects (with `edit`) via `doCodeActions2` (see `extensions/css-language-features/server/src/cssServer.ts`). The client, however, still registered a legacy `_css.applyCodeAction` command that was used by the older `Command`-style code actions and is no longer invoked from the server.

This PR removes the now-dead client-side glue:

- Drop the `_css.applyCodeAction` command registration and its `applyCodeAction` helper in `extensions/css-language-features/client/src/cssClient.ts`.
- Drop the unused `commands` and `window` imports that were only used by that helper.
- Drop the `onCommand:_css.applyCodeAction` activation event in `extensions/css-language-features/package.json` so the extension no longer activates on a command nobody fires.

No behavior change is expected: code actions continue to flow through the standard `CodeAction.edit` path that the editor applies directly.

Fixes #237858

### DIFF
--- a/extensions/css-language-features/client/src/cssClient.ts
+++ b/extensions/css-language-features/client/src/cssClient.ts
@@ -3,7 +3,7 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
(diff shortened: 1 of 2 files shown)


aeschli (dismissed):

aeschli:
Thanks @yogeshwaran-c , nice change.
I reverted the `package.json` change as, for security reasons, these can only be made by contributors. I will make a follow-up PR.

dmitrivMS (approved):

dmitrivMS (approved):

alexr00 (approved):

PR #310104, opened 2026-04-15
fix: preserve shell tool output after scrollback rollover

atharvasingh7007:
## Summary
- preserve captured shell output when the terminal scrollback trims away the original offset
- apply the same trim-safe behavior to sentinel detection, timeout handling, and unexpected shell exit handling
- add regression coverage for sentinel completion and timeout results after scrollback rollover

## Testing
- node --loader ts-node/esm ./node_modules/mocha/bin/mocha src/vs/platform/agentHost/test/node/copilotShellTools.test.ts --ui tdd --timeout 10000
- node build/eslint.ts src/vs/platform/agentHost/node/copilot/copilotShellTools.ts src/vs/platform/agentHost/test/node/copilotShellTools.test.ts


atharvasingh7007:

afurm (commented):
LGTM. The extracted `getContentSinceOffset` helper is cleaner than the inline `Math.min` clamp. 

One edge case to consider: when `offsetBefore > fullContent.length`, returning the entire `fullContent` is the right behavior for preserving output, but if `offsetBefore` is invalid (e.g., negative or NaN), the condition `offsetBefore > fullContent.length` would also be true and would silently return the full buffer. Should `getContentSinceOffset` validate that `offsetBefore` is a non-negative number first?

afurm (commented):
The extracted `getContentSinceOffset` helper is cleaner than the inline `Math.min` clamp.

Edge case: when `offsetBefore > fullContent.length`, returning the entire `fullContent` is correct for preserving output. But if `offsetBefore` is invalid (negative or NaN), the condition `offsetBefore > fullContent.length` would also be true and would silently return the full buffer. Should `getContentSinceOffset` validate that `offsetBefore` is a non-negative number first?

atharvasingh7007:
Addressed the offset validation edge case in 914b69: getContentSinceOffset() now clamps invalid offsets before slicing, and I added a focused unit assertion covering NaN, negative, oversized, and normal offsets.

PR #310185, opened 2026-04-15
Add a codicon debug gallery developer action

jmole:
Expose a workbench developer command that opens a searchable codicon gallery so icon names, descriptions, sizing, and copyable class names are easier to inspect while building UI.

This was vibe coded, and there is probably a better way to actually implement this, but I think it's conceptually sound. It's been a useful tool for me to use while building a VSCode extension.


jmole:

PR #310210, opened 2026-04-15
Fix router model selection: filter available_models, remove same-provider override, iterate candidates

aashna:
## Problem

Three issues cause the router's model recommendation to be silently overridden on the client, sending simple queries to expensive models:

### 1. available_models / knownEndpoints out of sync
`knownEndpoints` (from `GET /models`) and `available_models` (from `POST /auto_models`) are separate CAPI calls with independent caches (10-min vs 5-min refresh). When a model like `claude-haiku-4.5` appears in `available_models` before the Models API returns it, the router recommends it but the client can't resolve it to an endpoint — causing a silent fallback to the first available (often the most expensive) model.

### 2. Same-provider override negates router decisions
When the router re-runs after `/compact` or background summarization, `_findSameProviderModel` overrides the router's recommendation with a same-provider model, often switching to a more expensive one.

### 3. Only candidate_models[0] attempted
The fallback only tried the first candidate model. If that one model had no matching endpoint, all remaining candidates were discarded.

### Filter available_models before routing
Intersect `available_models` with `knownEndpoints` before sending to the router. The router now only sees models the client can actually serve.

```typescript
const knownModelIds = new Set(knownEndpoints.map(e => e.model));
const routableModels = token.available_models.filter(m => knownModelIds.has(m));
```

### Remove same-provider override from router path
Trust the router's ranked candidate list directly. Same-provider preference is retained in `_selectDefaultModel` (non-router fallback path).

### Iterate all candidates via _findFirstAvailableModel
Walk the full `candidate_models` list instead of only trying `[0]`.

```typescript
// Before:
const selectedModel = (entry?.endpoint && this._findSameProviderModel(...))
    ?? knownEndpoints.find(e => e.model === result.candidate_models[0]);

// After:
const selectedModel = this._findFirstAvailableModel(result.candidate_models, knownEndpoints);
```

### Warn-level logs for debuggability
Sync mismatches now emit warn-level logs with full model lists.

Telemetry changes are in a separate PR: #310238.


jinu-jang (approved):

extensions/copilot/src/platform/endpoint/node/automodeService.ts
```diff
@@ -215,6 +215,39 @@ export class AutomodeService extends Disposable implements IAutomodeService {
 
 		selectedModel = this._applyVisionFallback(chatRequest, selectedModel, token.available_models, knownEndpoints);
 
+		// Emit the final model selection alongside the router's recommendation
+		// so analysts can detect overrides without fragile telemetry joins
+		if (!skipRouter && routerResult.candidateModel) {
```
jinu-jang, line 220: This telemetry event is a superset of the one added in #310238. Suggest either closing #310238 as subsumed by this PR, or pulling the telemetry back out of this PR and letting #310238 own it.

Consider folding `actualModel` + `overrideReason` into the existing `automode.routerDecision` event instead of creating a parallel event. The existing event lives in `routerDecisionFetcher` (before overrides), but moving it up to `automodeService` (after `_applyVisionFallback`) would give you one event with both the router's recommendation AND the final outcome. That avoids the cross-event join the comment on line 218 mentions, without adding a new event to maintain.

extensions/copilot/src/platform/endpoint/node/automodeService.ts
```diff
@@ -215,6 +215,39 @@ export class AutomodeService extends Disposable implements IAutomodeService {
...
+					"candidateModel": { "classification": "SystemMetaData", "purpose": "FeatureInsight", "comment": "The router's top candidate model (candidate_models[0])" },
+					"actualModel": { "classification": "SystemMetaData", "purpose": "FeatureInsight", "comment": "The model actually selected after all client-side overrides" },
+					"overrideReason": { "classification": "SystemMetaData", "purpose": "FeatureInsight", "comment": "Why the actual model differs from the candidate: none, defaultFallback, or clientOverride" },
+					"filteredModelCount": { "classification": "SystemMetaData", "purpose": "PerformanceAndHealth", "isMeasurement": true, "comment": "Number of available_models filtered out because they had no matching knownEndpoint (CAPI sync gap indicator)" },
+					"droppedModels": { "classification": "SystemMetaData", "purpose": "PerformanceAndHealth", "comment": "Comma-separated list of model IDs filtered out before routing because they had no matching knownEndpoint" }
+				}
+			*/
+			const candidateModel = routerResult.candidateModel;
+			const overrideReason = candidateModel === selectedModel.model ? 'none'
+				: routerFallbackReason ? 'defaultFallback'
```
jinu-jang: `defaultFallback` (line 235) is unreachable. Here's why:

`routerResult.candidateModel` is only returned on the success path of `_tryRouterSelection` (line 356). All failure/fallback paths return without `candidateModel`. So the guard on line 220 (`if (!skipRouter && routerResult.candidateModel)`) guarantees this block only runs when the router succeeded — which means `routerFallbackReason` is always `undefined` here.

The ternary on line 235 checks `routerFallbackReason ? 'defaultFallback'` but that condition can never be true inside this block.

Similarly, `clientOverride` (line 236) is the catch-all, but after removing `_findSameProviderModel` from the router path, the only realistic trigger is vision fallback. Consider renaming to `visionFallback` for clarity, or splitting the detection explicitly.

The test at `automodeService.spec.ts:1268` (`should emit overrideReason=defaultFallback when router returns unknown model`) will fail — the mock returns `candidate_models: ['unknown-model']` which has no endpoint, so `_tryRouterSelection` returns `fallbackReason: 'noMatchingEndpoint'` *without* `candidateModel`, the guard on line 220 is false, and no telemetry event is emitted.

extensions/copilot/src/platform/endpoint/node/test/automodeService.spec.ts
```diff
@@ -1018,4 +1026,267 @@ describe('AutomodeService', () => {
...
+			expect(selectionEvent![1].candidateModel).toBe('gpt-4o');
+			expect(selectionEvent![1].actualModel).toBe('gpt-4o');
+			expect(selectionEvent![1].overrideReason).toBe('none');
+			expect(selectionEvent![1].droppedModels).toBe('claude-haiku-4.5');
+			expect(selectionEvent![2].filteredModelCount).toBe(1);
+		});
+	});
+
+	describe('routerModelSelection telemetry', () => {
+		function mockRouterResponse(available_models: string[], routerResult: { chosen_model: string; candidate_models: string[] }, session_token = 'test-token'): void {
```
jinu-jang: This `mockRouterResponse` helper is identical to the one declared at line 1031 (inside the `available_models / knownEndpoints sync` describe block) and similar to the one at line 451. Consider extracting to the outer `describe('AutomodeService')` scope so all three test groups share it.

extensions/copilot/src/platform/endpoint/node/automodeService.ts
```diff
@@ -215,6 +215,39 @@ export class AutomodeService extends Disposable implements IAutomodeService {
...
+				: 'clientOverride';
```
jinu-jang:
```suggestion
			const overrideReason = candidateModel === selectedModel.model ? 'none' : 'client_visionFallback';
```

`defaultFallback` (line 235) is unreachable.

This current if-block mandates that `routerResult.candidateModel` exists. `_tryRouterSelection` has no path where candidate model and fallback reason are both returned. `routerFallbackReason` is guaranteed `undefined` in this block.

extensions/copilot/src/platform/endpoint/node/test/automodeService.spec.ts
```diff
@@ -1018,4 +1026,267 @@ describe('AutomodeService', () => {
...
+				prompt: 'test prompt',
+				sessionId: 'session-telemetry-override'
+			};
+
+			await automodeService.resolveAutoModeEndpoint(chatRequest as ChatRequest, [gpt4oEndpoint]);
+
+			const telemetryCalls = mockTelemetryService.sendMSFTTelemetryEvent.mock.calls;
+			const selectionEvent = telemetryCalls.find((call: unknown[]) => call[0] === 'automode.routerModelSelection');
+			expect(selectionEvent).toBeDefined();
+			expect(selectionEvent![1].overrideReason).toBe('defaultFallback');
```
jinu-jang: Does this unit test actually pass? I can try running it on my machine, but setting up VSCode dev env is no easy feat.

This is in regards to another one of my comments where I'm pretty sure `defaultFallback` can never happen.

extensions/copilot/src/platform/endpoint/node/automodeService.ts
```diff
@@ -276,7 +309,25 @@ export class AutomodeService extends Disposable implements IAutomodeService {
 				turn_number: (entry?.turnCount ?? 0) + 1,
 			};
 			const routingMethod = this._configurationService.getExperimentBasedConfig(ConfigKey.TeamInternal.AutoModeRoutingMethod, this._expService) || undefined;
-			const result = await this._routerDecisionFetcher.getRouterDecision(prompt, token.session_token, token.available_models, undefined, contextSignals, conversationId, chatRequest?.id, routingMethod, hasImage(chatRequest));
+
+			// Filter available_models to only those the client can actually serve.
```
jinu-jang, line 303: Okay I was trying to see this mismatch in action through the debugger, but couldn't set up the dev env...

Just be wary that limiting the list of models sent to our routing model may inhibit our ability to make decisions down the road as well.
Given the assumption that our routing model can always preserve order of models, client side post-filtering may be better than pre-filtering.

jinu-jang (changes requested):
Okay this PR suddenly changed to subsume #310238, and I think adds a new telemetry event that doesn't actually add too much more information.
I think it would be better for the existing telemetry emission to be moved and updated.

Also no more force pushes if possible. It's not bad per-se, just makes reviewing more difficult.

extensions/copilot/src/platform/endpoint/node/test/automodeService.spec.ts
```diff
@@ -1018,4 +1018,135 @@ describe('AutomodeService', () => {
...
+				location: ChatLocation.Panel,
+				prompt: 'what day is today',
+				sessionId: 'session-iterate-candidates'
+			};
+
+			const result = await automodeService.resolveAutoModeEndpoint(chatRequest as ChatRequest, [gpt41Endpoint]);
+			expect(result.model).toBe('gpt-4.1');
+		});
+
+		it('should fall back to default when all available_models are unknown to knownEndpoints', async () => {
```
bhavyaus: Just to confirm  did this test actually run locally before you pushed? Running it against this branch throws `Auto mode failed: no available model found in known endpoints.` from `_selectDefaultModel`, because the early-return at line ~289 hands control back to `resolveAutoModeEndpoint`, which then calls `_selectDefaultModel(..., token.available_models, ...)` with the same unfiltered list and re-does the lookup that just failed.

So either the assertion is aspirational (`expect(result.model).toBe('gpt-4o')` can't be satisfied by the current code) or the production path needs to pick from `knownEndpoints[0]` when `available_models` has nothing resolvable.

extensions/copilot/src/platform/endpoint/node/automodeService.ts
```diff
@@ -276,7 +276,23 @@ export class AutomodeService extends Disposable implements IAutomodeService {
...
+			// The AutoModels API and Models API are separate CAPI calls that can be
+			// out of sync (e.g. a new model appears in available_models before the
+			// Models API returns it). Sending unresolvable models to the router
+			// causes it to recommend models the client must silently discard.
+			const knownModelIds = new Set(knownEndpoints.map(e => e.model));
+			const routableModels = token.available_models.filter(m => knownModelIds.has(m));
```
bhavyaus: Minor: `token.available_models` is walked three times — once to filter, once in the warn log, once to build the dropped list for the info log. A single partition pass would be simpler and make it obvious the three views stay consistent. Something like:

```ts
const routableModels: string[] = [];
const droppedModels: string[] = [];
for (const m of token.available_models) {
    (knownModelIds.has(m) ? routableModels : droppedModels).push(m);
}
```

Then the two log sites just reference `droppedModels` directly.

bhavyaus (dismissed):

Yoyokrazy (dismissed):

jinu-jang (approved):

bhavyaus (approved):

PR #307853, opened 2026-04-04
perf: fix memory leaks, improve build performance, and optimize compi…

InSelfControll:
Memory Leak Fixes:
- Fix _instanceDisposables leak in TerminalGroup.dispose() - event listener disposables were never cleaned up when terminal groups were disposed
- Fix _backgroundedTerminalDisposables leak in TerminalService - added override dispose() to clean up background terminal disposables
- Fix terminal link cache leak in ExtHostTerminalService._onProcessExit() - _terminalLinkCache and _terminalLinkCancellationSource were never cleaned up when terminals exited
- Add LRU eviction to ExtHostChatContext._globalItems - capped at 10,000 items to prevent unbounded memory growth from AI chat context items
- Convert IDisposable[] to DisposableStore in notebook.contribution.ts (CellInfoContentProvider and NotebookMetadataContentProvider) for more robust disposal patterns

Build Performance Fixes (x64/x86 compilation speed):
- Enable TypeScript incremental compilation in tsconfig.json and tsconfig.base.json - expected 50-70% faster recompilation
- Add npm_config_jobs=max for native module compilation in postinstall.ts - allows node-gyp to use all CPU cores during native addon builds
- Increase gulp memory limit from 8GB to 16GB in package.json to reduce GC thrashing during large builds
- Add rspack splitChunks optimization with vendor and common cache groups
- Add vite production optimizations: esbuild minification, sourcemaps, and manual chunk splitting for vendor/xterm dependencies

PR #310216, opened 2026-04-15
fix: resolve memory leaks in terminal, notebook, and chat context dis…

InSelfControll:
Splits the original PR (https://github.com/microsoft/vscode/pull/307853) into two focused PRs. This PR addresses memory leak fixes only.

Changes:

    Terminal: Clean up link cache and cancellation sources on process exit
    Terminal: Dispose instance disposables properly in terminalGroup
    Terminal: Dispose backgrounded terminal disposables in terminalService
    Notebook: Use DisposableStore instead of IDisposable[] for proper lifecycle
    Chat context: Clean up chat context items on dispose


dmitrivMS (changes requested):
I think the real leak is only at extHostTerminalService - the rest are disposed at app shutdown, so no big point in cleaning anything up at that point. extHostChatContext - there's behavior changes here as well and not clear if something is actually leaking.

I think we can make a separate change for extHostTerminalService fix, but not the rest.

PR #310238, opened 2026-04-15
Add automode.routerModelSelection telemetry with actualModel

aashna:
Emits a new telemetry event `automode.routerModelSelection` after all client-side model overrides (same-provider, vision fallback, default selection) are applied.

## Properties

| Property | Description |
|----------|-------------|
| `candidateModel` | The router's top pick (`candidate_models[0]`) |
| `actualModel` | The model actually used after all overrides |
| `overrideReason` | `none` or `clientOverride` |
| `conversationId` | For correlation |

## Why

The existing `automode.routerDecision` event captures the router's recommendation, but doesn't know what model was ultimately selected — vision fallback and default selection happen afterward. This event enables accurate switch attribution:

```kql
automode.routerModelSelection
| where candidateModel != actualModel
| summarize count() by candidateModel, actualModel
```

## Design note

`candidateModel` is only set when `_tryRouterSelection` succeeds (i.e., the router returned a valid model with a matching endpoint). When the router fails or returns an unknown model, the event is NOT emitted — those cases are covered by `automode.routerFallback` instead. This means `overrideReason` will never be `defaultFallback` — it's always `none` or `clientOverride`.

### DIFF
--- a/extensions/copilot/src/platform/endpoint/node/automodeService.ts
+++ b/extensions/copilot/src/platform/endpoint/node/automodeService.ts
@@ -215,6 +215,29 @@ export class AutomodeService extends Disposable implements IAutomodeService {
 
 		selectedModel = this._applyVisionFallback(chatRequest, selectedModel, token.available_models, knownEndpoints);
 
+		// Emit the final model selection alongside the router's recommendation
+		// so analysts can detect overrides without fragile telemetry joins
+		if (!skipRouter && routerResult.candidateModel) {
+			/* __GDPR__
+				"automode.routerModelSelection" : {
+					"owner": "aashnagarg",
+					"comment": "Reports the router's recommended model vs the actual model used after all client-side overrides",
+					"conversationId": { "classification": "SystemMetaData", "purpose": "FeatureInsight", "comment": "The conversation ID" },
+					"candidateModel": { "classification": "SystemMetaData", "purpose": "FeatureInsight", "comment": "The router's top candidate model (candidate_models[0])" },
+					"actualModel": { "classification": "SystemMetaData", "purpose": "FeatureInsight", "comment": "The model actually selected after all client-side overrides" },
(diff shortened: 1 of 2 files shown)


extensions/copilot/src/platform/endpoint/node/automodeService.ts
```diff
@@ -215,6 +215,31 @@ export class AutomodeService extends Disposable implements IAutomodeService {
...
+					"candidateModel": { "classification": "SystemMetaData", "purpose": "FeatureInsight", "comment": "The router's top candidate model (candidate_models[0])" },
+					"actualModel": { "classification": "SystemMetaData", "purpose": "FeatureInsight", "comment": "The model actually selected after all client-side overrides" },
+					"overrideReason": { "classification": "SystemMetaData", "purpose": "FeatureInsight", "comment": "Why the actual model differs from the candidate: none, visionFallback, defaultFallback, or sameProvider" }
+				}
+			*/
+			const candidateModel = routerResult.candidateModel;
+			const overrideReason = candidateModel === selectedModel.model ? 'none'
+				: routerFallbackReason ? 'defaultFallback'
+				: 'clientOverride';
+			this._telemetryService.sendMSFTTelemetryEvent('automode.routerModelSelection', {
```
jinu-jang: I'll also try and check the feasibility of my comment when I have the time, but is it not possible to just pipe `actualModel` to the `github.copilot-chat/automode.routerdecision` event?

jinu-jang: Maybe we need Logan's input on this, but I think moving the emission of `github.copilot-chat/automode.routerdecision` out of the fetcher object and into here may just be cleanest.

If we have two separate events for basically the same thing, that might get confusing as well.

bhavyaus (approved):

PR #310261, opened 2026-04-15
feat: add useful HTML snippets for common elements

Nedunchezhiyan-M:
## Summary

The built-in `html` extension currently provides only a single snippet — the basic HTML document template. This PR adds 17 practical snippets covering common HTML patterns that developers repeatedly type by hand.

### New Snippets

| Prefix | Description |
|--------|-------------|
| `html5` | Full HTML5 document with viewport/description meta tags, stylesheet, and script links |
| `link` | Stylesheet `<link>` tag |
| `script` | `<script src>` tag |
| `meta:vp` | Responsive viewport meta tag |
| `figure` | `<figure>` with `` and `<figcaption>` |
| `form` | `<form>` with action, method, and submit button |
| `input` | `<input>` paired with a `<label>` |
| `select` | `<select>` dropdown with placeholder options |
| `textarea` | `<textarea>` with rows/cols |
| `nav` | `<nav>` with an unordered link list |
| `article` | `<article>` with header and paragraph |
| `section` | `<section>` with id and heading |
| `table` | Table with `<thead>` and `<tbody>` |
| `ul` | Unordered list with items |
| `ol` | Ordered list with items |
| `img` | Image with src and alt |
| `a` | Anchor with href and link text |
| `div` | Div with class attribute |

All snippets use tab stops (`$1`, `$2`, etc.) so users can quickly fill in values, and linked placeholders (e.g. `id` appears in both label and input) keep things consistent.

### Why?

HTML is one of the most common file types in VS Code, yet the built-in snippets provide almost nothing beyond the doc template. These snippets match patterns every web developer types dozens of times per day.

### DIFF
--- a/extensions/html/snippets/html.code-snippets
+++ b/extensions/html/snippets/html.code-snippets
@@ -14,5 +14,197 @@
 			"</html>"
 		],
 		"description": "HTML Document"
+	},
+	"HTML5 Document": {
+		"prefix": "html5",
+		"isFileTemplate": true,
+		"body": [
+			"<!DOCTYPE html>",
+			"<html lang=\"${1:en}\">",
+			"<head>",
+			"\t<meta charset=\"UTF-8\" />",
+			"\t<meta name=\"viewport\" content=\"width=device-width, initial-scale=1.0\" />",
+			"\t<meta name=\"description\" content=\"${2:description}\" />",
+			"\t<title>${3:title}</title>",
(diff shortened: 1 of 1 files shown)


aeschli:
The problem with snippets contributed by the built-in extensions is that user can not hide them (except by turning off snippet completions).
That's why we don't want to add any further snippets to built-in extensions and instead recommend snippets to be provided via extensions on the marketplace.

PR #310437, opened 2026-04-16
better handling of too many images in chat

justschen:
fix issue where gemini had a 10 image limit and anthropic messages api had a 20 images limit

### DIFF
--- a/extensions/copilot/src/platform/endpoint/node/imageLimits.ts
+++ b/extensions/copilot/src/platform/endpoint/node/imageLimits.ts
@@ -1,0 +1,119 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+import { Raw } from '@vscode/prompt-tsx';
+
+/**
+ * Model-facing placeholder substituted for dropped history images.
+ * Intentionally not localized — this text is sent to the model, not the user.
+ */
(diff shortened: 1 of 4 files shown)


bhavyaus:
The `validateAndFilterImages` private method is accessed via a `(endpoint as unknown as {...})` cast in tests. That is a smell that suggests the method could be factored into a pure helper (takes `messages`, `maxImages`, returns messages) exported for direct testing — no mock endpoint construction needed, no cast.

extensions/copilot/src/platform/endpoint/node/imageLimits.ts
```diff
@@ -0,0 +1,119 @@
...
+	// No filtering needed if total is within the limit
+	if (totalImages <= maxImages) {
+		return messages;
+	}
+
+	// Walk backward through history (before the current turn), keeping the
+	// most recent images and replacing the oldest with placeholders. Budget
+	// can go negative if the current turn alone exceeds the limit — in that
+	// case we drop all history images and let the model handle it.
+	let historyBudget = maxImages - currentTurnImages;
```
bhavyaus, line 83: When this goes negative, the current turn alone exceeds the limit and we send a request we know will be rejected with an opaque server error. The old code threw a clear localized error here. Could we throw for this specific case and keep the silent history-filtering for the rest?

alexdima (dismissed):

extensions/copilot/src/platform/endpoint/node/imageLimits.ts
justschen: copilot actually added this at first and i rejected it since i kinda wanted to see the error from the model (it would have request ids and such), but i think for the user this is probably better. although if they do attach more than 20 images, or 10 images, we _are_ warning them

alexdima (approved):

PR #310733, opened 2026-04-16
fix: handle quoted ID values in check-requirements-linux.sh

Jah-yee:
Good day

The previous grep-based approach failed to parse ID values surrounded by quotes (e.g., `ID=\"rocky\"` on Rocky Linux 8.5), resulting in an empty `OS_ID` and exit code 1.

This fix uses `source /etc/os-release && echo \"$ID\"` to properly parse shell-compatible variable assignments, which correctly handles both quoted and unquoted values.

Fixes #232159

## Verification
- Tested with quoted format (`ID=\"rocky\"`): correctly returns `rocky`
- Tested with unquoted format (`ID=ubuntu`): correctly returns `ubuntu`
- The `source` approach is a standard shell idiom for reading `/etc/os-release`

Warmly,
RoomWithOutRoof


Jah-yee:

PR #310763, opened 2026-04-16
fix(debug): identify instruction breakpoints by resolved address to allow removal when instructionReference changes (#289678)

### LINKED ISSUE #289678: Unable to remove instruction breakpoints in Disassembly view when instruction reference changes

- VS Code Version: 1.108.1
- OS Version: Windows 11

The current issue was identified in the context of developing a debugger extension that uses the Debug Adapter Protocol.

Steps to Reproduce: 

1. Open the Disassembly view during a debug session. 

2. Set an instruction breakpoint on a specific line. 

3. Trigger an action that causes the instruction reference for that memory location to change or refresh (e.g., reloading symbols or certain stepping operations). 

4. Click the breakpoint icon to remove it. 

5. Result: The breakpoint remains active and cannot be toggled off by clicking.  

 

Technical Analysis: 
The issue stems from how breakpoints are identified during the removal flow: 

1. Storage: Instruction breakpoints are stored as InstructionBreakpoint objects containing instructionReference, offset, and address. 
 - Reference:  https://github.com/microsoft/vscode/blob/main/src/vs/workbench/contrib/debug/common/debugModel.ts#L2061 

2. Trigger: When clicking a breakpoint in the Disassembly view (disassemblyView.ts L727), removeInstructionBreakpoints() is called using the current element's instruction reference and offset. 
 - Reference: https://github.com/microsoft/vscode/blob/main/src/vs/workbench/contrib/debug/browser/disassemblyView.ts#L727 

3. The Failure: The logic in debugModel.ts L2072 attempts to find the breakpoint by matching these two fields. If the instruction reference has changed, the lookup fails, and the breakpoint is never removed from the model. 
 - Reference: https://github.com/microsoft/vscode/blob/main/src/vs/workbench/contrib/debug/common/debugModel.ts#L2072 

 

Proposed Solution: 
Instead of matching against the volatile instructionReference and offset, the search should be performed using the address. 

I suggest updating the removeInstructionBreakpoints() signature to include the address or refactoring the search logic to prioritize the address as a unique identifier for breakpoints within the current disassembly session.

maruthang:
## What

Instruction breakpoints set from the Disassembly view could not be removed when the debug adapter returned a different `instructionReference` for the same memory location across `disassemble` requests.

## Why

The Debug Adapter Protocol explicitly allows adapters to vary the `instructionReference` value returned for a given memory location between `disassemble` calls — only the resolved memory address is stable. The previous `removeInstructionBreakpoints(reference, offset)` lookup compared the (possibly new) reference+offset against the breakpoint's original reference+offset, so the match failed and the toggle-off click was a no-op.

Fixes #289678

## How

- Extended `IDebugService.removeInstructionBreakpoints` / `DebugService.removeInstructionBreakpoints` / `DebugModel.removeInstructionBreakpoints` with an optional third parameter `address?: bigint`.
- When `address` is supplied, `DebugModel` matches on the stable resolved memory address already stored on `InstructionBreakpoint`, so a changed `instructionReference` no longer breaks removal.
- When `address` is omitted, the existing `reference + offset` matching path is preserved unchanged — all existing callers keep their current behavior.
- `disassemblyView.ts` now passes `currentElement.element.address` (the same address it already uses when *adding* the breakpoint), guaranteeing symmetric add/remove.

## Test plan

- Added a new `InstructionBreakpoint` suite in `src/vs/workbench/contrib/debug/test/common/debugModel.test.ts` with 3 tests:
  1. Regression for #289678 — removal succeeds when `instructionReference` has changed but `address` is stable.
  2. Back-compat — when `address` is not supplied, the legacy `reference + offset` matching still works.
  3. Targeted removal — supplying `address` only removes the breakpoint at that address, leaving others intact.
- `npm run compile-check-ts-native` passes cleanly.

### DIFF
--- a/src/vs/workbench/contrib/debug/browser/debugService.ts
+++ b/src/vs/workbench/contrib/debug/browser/debugService.ts
@@ -1199,8 +1199,8 @@ export class DebugService implements IDebugService {
 		this.debugStorage.storeBreakpoints(this.model);
 	}
 
-	async removeInstructionBreakpoints(instructionReference?: string, offset?: number): Promise<void> {
-		this.model.removeInstructionBreakpoints(instructionReference, offset);
+	async removeInstructionBreakpoints(instructionReference?: string, offset?: number, address?: bigint): Promise<void> {
+		this.model.removeInstructionBreakpoints(instructionReference, offset, address);
 		this.debugStorage.storeBreakpoints(this.model);
 		await this.sendInstructionBreakpoints();
 	}
--- a/src/vs/workbench/contrib/debug/browser/disassemblyView.ts
+++ b/src/vs/workbench/contrib/debug/browser/disassemblyView.ts
@@ -736,11 +736,17 @@ class BreakpointRenderer implements ITableRenderer<IDisassembledInstructionEntry
 					// click show hint while waiting for BP to resolve.
 					icon.classList.add(this._breakpointHintIcon);
 					const reference = currentElement.element.instructionReference;
-					const offset = Number(currentElement.element.address - this._disassemblyView.getReferenceAddress(reference)!);
+					const address = currentElement.element.address;
+					const offset = Number(address - this._disassemblyView.getReferenceAddress(reference)!);
 					if (currentElement.element.isBreakpointSet) {
-						this._debugService.removeInstructionBreakpoints(reference, offset);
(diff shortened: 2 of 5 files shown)


connor4312 (approved):
thanks!

dmitrivMS (approved):

Issue #291642, opened 2026-01-29
Debug disassembly view infinite scrolling does not work after navigating to second frame


CiaranWoodward:
Does this issue occur when all extensions are disabled?: No (Can't debug without extensions)

- VS Code Version: 1.108.2
- OS Version: Windows 11 Pro 25H2

When using the disassembly view, if you disassemble, you get the disassembly around the PC of that frame.

If you scroll up or down, then you get the disassembly of those earlier or later addresses (as expected!).

However, if you change to a different frame with a significantly different PC, then scroll towards the first,
you will notice that the infinite scrolling does not populate new instructions, and there is instead a sudden jump
in address.

Technically speaking, since the different disassembly chunks have been generated from different instructionReference
values, there is no guarantee that they are even part of the same program. In my opinion, the best and simplest solution to this issue is just to clear old disassembly from the view when new disassembly for a new instructionReference is requested.

<img width="298" height="288" alt="Image" src="https://github.com/user-attachments/assets/90e85e6c-31bd-43ee-8c47-bdd31db0c3a3" />

Steps to reproduce:
1. Start debugging session, pause, disassemble a frame A
2. Switch to disassembling a different frame B, a few hundred instructions away in memory
3. scroll to the frame A's instruction location. Note that there will be a gap in memory addresses (which is not obvious)

PR #310782, opened 2026-04-16
fix(debug): reload disassembly when instructionReference targets a different memory region (#291640)

maruthang:
**What** — When the disassembly view is asked to show an instruction from a different `instructionReference` than what's currently rendered, clear and reload around the target instead of splicing new rows next to stale rows from the previous memory region.

**Why** — Per DAP spec, `instructionReference` is an opaque handle that may point at an unrelated memory region (e.g. different process in a multi-process debug session). The current logic assumed a single contiguous memory space and produced a mixed view when stepping across regions (#291640).

**How** — Added a guard at the top of `goToInstructionAndOffset`: if the current first row's `instructionReference` differs from the target, call the existing `reloadDisassembly` path (which clears `_referenceToMemoryAddress` and repopulates). Scrolling within a single reference is unaffected.

**Test plan** — No automated test (covered via manual DAP repro per the issue). `npm run compile-check-ts-native` passes cleanly.

Note that this is orthogonal to PR #310763 (fix for #289678); the two PRs touch different methods in the same file with no textual conflict.

Fixes #291640

### DIFF
--- a/src/vs/workbench/contrib/debug/browser/disassemblyView.ts
+++ b/src/vs/workbench/contrib/debug/browser/disassemblyView.ts
@@ -372,6 +372,20 @@ export class DisassemblyView extends EditorPane {
 	}
 
 	async goToInstructionAndOffset(instructionReference: string, offset: number, focus?: boolean) {
+		// Instruction references are opaque handles that may point at unrelated
+		// memory regions (e.g. different processes or address spaces in a
+		// multi-process debug session). When the target instructionReference
+		// differs from what is currently rendered, we cannot assume the two
+		// regions are contiguous, so clear and reload rather than splicing new
+		// rows in next to stale ones from the previous region (#291640).
+		if (this._disassembledInstructions && this._disassembledInstructions.length > 0) {
+			const currentFirst = this._disassembledInstructions.row(0);
+			if (currentFirst !== disassemblyNotAvailable && currentFirst.instructionReference !== instructionReference) {
+				this.reloadDisassembly(instructionReference, offset);
+				return;
+			}
+		}
+
 		let addr = this._referenceToMemoryAddress.get(instructionReference);
 		if (addr === undefined) {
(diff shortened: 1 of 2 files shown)


CiaranWoodward:
I think this will also fix #291642

maruthang:
Thanks @CiaranWoodward — that's a good catch. Looking at #291642, the symptom (stale disassembly bytes when navigating across regions) does share the same root cause this PR addresses: `_loadingReference` is consulted to skip work even when `instructionReference` has moved to a new memory region. Once we keep loading whenever the reference doesn't fall inside the currently loaded window, both reports converge on the same path.

I can add a brief test that mirrors the #291642 reproduction (memory-region jump where the bytes from the previous load are still cached) and reference both issues in the commit body, so the fix is explicit. Will push that as a follow-up commit and let CI re-run.

CiaranWoodward:
@maruthang I assume by ``_loadingReference`` you are referring to the new ``currentFirst.instructionReference`` check you just added? (I can't see ``_loadingReference`` anywhere). But the DAP server I was using (gdb) uses a different instruction reference for every stack frame, so swapping frame to disassemble a different function triggers a new disassembly from that different instructionReference - and with your new logic, should clear all previous instructions - so I agree this is a good fix for that too.

Also I think the #291642 situation is probably easier to add a test for, so sounds good!

maruthang:
Thanks for confirming the overlap with #291642! I've added a unit test for that exact scenario: the reload decision is now factored into an exported `shouldReloadDisassembly` helper, and `disassemblyView.test.ts` covers the case where instructions loaded for a previous frame's reference are still cached and a new frame arrives with a different (gdb-style per-frame) instructionReference — asserting a full reload rather than reuse, referencing both #291640 and #291642.

I also addressed the review note about focus: `reloadDisassembly` now takes the caller's focus intent, so the early-return path no longer steals focus when `goToInstructionAndOffset` was called without `focus` (e.g. on stack-frame changes), while explicit navigation keeps focusing as before.

PR #310800, opened 2026-04-16
fix: prevent update server from bogus 'unknown' commit in overwrite checks

deepak1556:
When update metadata is missing (e.g. interrupted install recovery via _applySpecificUpdate), the fallback IUpdate uses version: 'unknown'. This value flows into checkForOverwriteUpdates, which polls every 5 minutes and passes it as the commit to buildUpdateFeedUrl, producing requests to `/api/update/{platform}/{quality}/unknown`

### DIFF
--- a/src/vs/platform/update/electron-main/abstractUpdateService.ts
+++ b/src/vs/platform/update/electron-main/abstractUpdateService.ts
@@ -392,6 +392,10 @@ export abstract class AbstractUpdateService implements IUpdateService {
 
 		const pendingUpdateCommit = this._state.update.version;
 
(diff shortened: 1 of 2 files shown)


deepak1556:
We should construct a better metadata with real commit for the recovery path than the unknown value https://github.com/microsoft/vscode/blob/df871147de755e0b7e5a42b5f22b8487f4b4484a/src/vs/platform/update/electron-main/updateService.win32.ts#L541 but skipping the value as an immediate fix.

dmitrivMS (dismissed):

dmitrivMS (approved):

PR #310821, opened 2026-04-16
Hide system-initiated steering messages from chat UI

### LINKED ISSUE #308549: Revisit UX of background terminal command finished notification
Right now, this is just indicated in the text of the response, which is not great from an accessibility perspective. 

@connor4312 is investigating the UX of this for other long running tasks and we'll want to follow suit for the terminal.

His idea is to surface running task info and provide options like:
- stop running the task
- wait for the task to complete
- carry on

meganrogge:
Fixes https://github.com/microsoft/vscode/issues/308549

Aligns VS Code's handling of background task notifications with Copilot CLI behavior.

Problem

When a background terminal command completes or needs input, VS Code shows a visible "Steering"
divider and request bubble in the chat conversation. Users see this as a separate message and
assume it's costing them a premium request, creating confusion about PRU consumption.

How Copilot CLI handles this

In the CLI, system notifications (e.g. "command finished") are injected silently into the
conversation — the user never sees the notification itself. They only see the model's brief
response acknowledging the event (e.g. "Build succeeded, all tests passed"). This keeps the
conversation clean and avoids the perception of extra cost.

This:

 - Hides system-initiated requests — Both the pending "Steering" divider and the compact request 
bubble (e.g. `npm run build` completed) are filtered from the visible conversation. The model's 
response remains visible.
 - User-initiated steering is unaffected — Messages explicitly sent as steering via queue actions 
still appear normally.
 - Fixes screen reader / pending UX — The status("Steering") announcement and "You already have 
pending requests" confirmation dialog now exclude system-initiated items, so users aren't alerted 
about things they can't see.
 - Fixes fork index calculation — Switched from counting visible requests in the view model to a 
model-level ID lookup (`getRequests().findIndex()`), preventing off-by-one errors when hidden 
requests exist between visible items.
 - Fixes telemetry lookups — `chatCodeblockActions.ts` now uses `session.model.getRequests()` instead 
of `session.getItems()` to find the originating request, so `modelId` is always resolved even for 
responses to hidden requests.

### DIFF
--- a/src/vs/workbench/contrib/chat/browser/actions/chatCodeblockActions.ts
+++ b/src/vs/workbench/contrib/chat/browser/actions/chatCodeblockActions.ts
@@ -33,7 +33,7 @@ import { reviewEdits } from '../../../inlineChat/browser/inlineChatController.js
 import { ITerminalEditorService, ITerminalGroupService, ITerminalService } from '../../../terminal/browser/terminal.js';
 import { ChatContextKeys } from '../../common/actions/chatContextKeys.js';
 import { ChatCopyKind, IChatService } from '../../common/chatService/chatService.js';
-import { IChatRequestViewModel, IChatResponseViewModel, isRequestVM, isResponseVM } from '../../common/model/chatViewModel.js';
+import { IChatResponseViewModel, isResponseVM } from '../../common/model/chatViewModel.js';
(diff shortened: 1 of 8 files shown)


Yoyokrazy (dismissed):

meganrogge:
Got feedback that we should keep these

PR #310839, opened 2026-04-16
Fix terminal auto-approve splitting commands at `--flag=value` in powershell

### LINKED ISSUE #294010: PowerShell auto approve incorrectly parsing command
Issue in powershell tree sitter grammar?

Image

```
git log --format="%h|%s|%an|%ad" --date=short dff523fc450 -1; git log --format="%h|%s|%an|%ad" --date=short 0a541d056d3 -1
```

Image

anthonykim1:
Resolves: https://github.com/microsoft/vscode/issues/294010
/cc @meganrogge 

Easy repro on windows powershell , set `git log` as entry to `"chat.tools.terminal.autoApprove": {` in settings.json

Before: 
image

After: 
image

What the problem was: 
* When Copilot proposes a PowerShell command that contains a POSIX-style
`--flag=value` argument (very common: `git log --format="…"` etc.), the terminal auto-approve flow splits the command line incorrectly and refuses to auto-approve commands that should match the user's allow rules.

* Likely upstream `tree-sitter-powershell` grammar: `word=` at command-argument position is parsed as an assignment
expression, which terminates the containing command. So `git log --format="a|b" HEAD -1` becomes roughly
`["git log --format", "b\" HEAD -1"]` instead of one `git log` command.

What this PR does: 
Replaces the `=` in PowerShell `--flag=value` argument patterns with a space so the tree-sitter PowerShell grammar treats the argument as a normal command parameter. Character positions and overall length are preserved so node ranges from the parsed (masked) tree can be used to slice the original command line.

* In `extractSubCommands`, for PowerShell only, mask `--flag=` / `-flag=` to `--flag ` / `-flag ` before parsing. The replacement is length-preserving, so node start/end indices from the parsed (masked) tree still index correctly into the original command line — we slice the original so the returned sub-commands keep the user-visible `=`.

* I believe we do similar with `CommandLinePwshChainOperatorRewriter`, which works around a separate
gap in the same grammar (chain operators / `&&`).

Long term:
I've added TODO to remove "workaround" once issue is addressed in https://github.com/airbus-cert/tree-sitter-powershell

@alexr00 Is the protocol for tree-sitter that we file issue under airbus first, then we update https://github.com/microsoft/vscode-tree-sitter-wasm and eventually update https://github.com/microsoft/vscode-tree-sitter-wasm in core?

### DIFF
--- a/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/treeSitterCommandParser.ts
+++ b/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/treeSitterCommandParser.ts
@@ -17,6 +17,28 @@ export const enum TreeSitterCommandParserLanguage {
 	PowerShell = 'powershell',
 }
 
+/**
+ * Matches a PowerShell command token of the form `-flag=` or `--flag=` at the
+ * start of input or following whitespace. Used to work around a tree-sitter
+ * PowerShell grammar limitation where POSIX-style `--flag=value` arguments
+ * (e.g. `git log --format="a|b"`) are parsed as assignment expressions and
+ * truncate the surrounding command.
+ *
+ * See https://github.com/microsoft/vscode/issues/294010
+ */
+const pwshFlagEqualsRegex = /(^|\s)(-{1,2}[\w-]+)=/g;
+
+/**
+ * Replaces the `=` in PowerShell `--flag=value` argument patterns with a space
+ * so the tree-sitter PowerShell grammar treats the argument as a normal
+ * command parameter. Character positions and overall length are preserved so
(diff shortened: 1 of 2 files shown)


Yoyokrazy (approved):

alexr00:
> Long term:
> I've added TODO to remove "workaround" once issue is addressed in https://github.com/airbus-cert/tree-sitter-powershell
>
> @alexr00 Is the protocol for tree-sitter that we file issue under airbus first, then we update https://github.com/microsoft/vscode-tree-sitter-wasm and eventually update https://github.com/microsoft/vscode-tree-sitter-wasm in core?

Yep, that would be the path forward.

PR #310846, opened 2026-04-16
Agents - temporarily comment out command registration to avoid two commands being registered with the same id

lszomoru:
//cc @deepak1556


deepak1556:
Addressing in https://github.com/microsoft/vscode/pull/310852

Thanks!

PR #310877, opened 2026-04-17
Restrict autoImportSpecifierExcludeRegexes to trusted workspaces

jakebailey:
`autoImportSpecifierExcludeRegexes` is a regex that tsserver will parse and use, which means that one can put a complicated regex and ReDoS themselves. Restrict this setting to trusted workspaces, which is more reasonable than attempting to determine if a given input is okay or not.

TypeScript 7+ are in Go and we chose to switch to Go's `regexp` library, which is based on RE2, so does not have this problem.


jakebailey:
cc @mjbvz

mjbvz:
Resubmitted as #310894 since package.json changes need to come from core team members

PR #310895, opened 2026-04-17
Fix "Add to Chat" using current tab instead of right-clicked tab

### LINKED ISSUE #276020: Add file to chat doesn't work for the non-current tab

- VS Code Version: 
```

Version: 1.105.1 (Universal)
Date: 2025-10-14T22:33:36.618Z (3 wks ago)
Electron: 37.6.0
ElectronBuildId: 12502201
Chromium: 138.0.7204.251
Node.js: 22.19.0
V8: 13.8.258.32-electron.0
OS: Darwin arm64 25.0.0
```
- OS Version: 
OSX 26.0.1 (25A362) on M3 Mac

Steps to Reproduce:

1. Add a file to chat, using the right-click menu on a file tab that is not the currently open file tab
2. The file is not added

if you do it on the current tab, it gets added, if you do it on a file in the left pane it's also added.

yogeshwaran-c:
## Summary

Right-clicking a non-current editor tab and selecting "Add File to Chat" / "Add Folder to Chat" attached the CURRENT editor instead of the right-clicked one.

## Root cause

`AttachResourceAction._getResources()` in `chatContextActions.ts` called:

```ts
this._getEditorResources(accessor, args)
```

`_getEditorResources` is declared with `...args: unknown[]`, so passing `args` as a single argument wraps it into a nested array (`args` inside the callee becomes `[[a, b, c]]`). `resolveCommandsContext` then filters the nested array out because it is neither a URI nor an editor commands context, falls back to the active editor group, and attaches the current editor.

## Fix

Spread the arguments (`...args`) so `_getEditorResources` receives the original argument list. `resolveCommandsContext` then sees the `{ groupId, editorIndex }` commands context (passed by the editor title menu) and resolves the right-clicked editor.

## Test plan

- [ ] Open two files, make one active
- [ ] Right-click the non-active tab, pick "Add File to Chat"
- [ ] The non-active file should be attached to the chat (not the active one)

Fixes #276020

### DIFF
--- a/src/vs/workbench/contrib/chat/browser/actions/chatContextActions.ts
+++ b/src/vs/workbench/contrib/chat/browser/actions/chatContextActions.ts
@@ -90,7 +90,7 @@ abstract class AttachResourceAction extends Action2 {
 	protected _getResources(accessor: ServicesAccessor, ...args: unknown[]): URI[] {
 		const editorService = accessor.get(IEditorService);
 
-		const contexts = isEditorCommandsContext(args[1]) ? this._getEditorResources(accessor, args) : Array.isArray(args[1]) ? args[1] : [args[0]];
+		const contexts = isEditorCommandsContext(args[1]) ? this._getEditorResources(accessor, ...args) : Array.isArray(args[1]) ? args[1] : [args[0]];
 		const files = [];
 		for (const context of contexts) {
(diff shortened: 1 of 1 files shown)


dmitrivMS:
The issue is already fixed.

PR #310901, opened 2026-04-17
fix(editor): defer native EditContext refresh to render time to fix large case-conversion freeze (#262473)

### LINKED ISSUE #262473: When the content of a text file is large, case conversion freezes

- VS Code Version: 1.102.3
- OS Version: win11

Steps to Reproduce:

1. When opening large text files (1mb or more) and using the system's built-in conversion tool to uppercase, the program freezes
2.Tested, case conversion is normal for versions 1.101 and below

maruthang:
## What

The native `EditContext` is no longer synchronously refreshed from inside model/view event handlers; instead, handlers mark a dirty flag and the refresh runs once in `prepareRender` after all events for the edit have been dispatched.

## Why

"Transform to Uppercase" / "Transform to Lowercase" over a multi-MB selection froze the editor for seconds. `NativeEditContext` was calling `_updateEditContext()` (which hands the whole primary-selection text to `EditContext.updateText`) synchronously from `onDidChangeContent` (model) and `onCursorStateChanged` (view) — so each event during the bulk edit pushed a megabyte-scale string into the native EditContext from the hot dispatch stack. Regression introduced by PR #298272. See #262473; @alexdima identified the hot path in the comments.

## How

Added a `_editContextNeedsRefresh` boolean. `onDidChangeContent` and `onCursorStateChanged` set the flag (and call `setShouldRender()`) instead of invoking `_updateEditContext()` inline. `prepareRender(ctx)` drains the flag by calling `_updateEditContext()`, which self-clears the flag at its top. Composition-start/end paths remain synchronous — IME buffers are small and latency-sensitive, and the write there is already bounded.

## Test plan (manual)

Generate a ~2 MB text file (attached sample from the issue, or `python -c "print('hello world '*200000)" > big.txt`). Open in VS Code built from this branch with the native EditContext enabled. `Ctrl+A`, F1 → "Transform to Uppercase". Before this fix: multi-second freeze. After: editor responds within a few hundred ms. Also verified: typing, IME composition, undo/redo, cursor navigation, screen-reader read-back all continue to work (screen-reader path uses the same `prepareRender` cadence).

Fixes #262473

### DIFF
--- a/src/vs/editor/browser/controller/editContext/native/nativeEditContext.ts
+++ b/src/vs/editor/browser/controller/editContext/native/nativeEditContext.ts
@@ -60,6 +60,16 @@ export class NativeEditContext extends AbstractEditContext {
 	private _previousEditContextSelection: OffsetRange = new OffsetRange(0, 0);
 	private _previousEditContextText: string = '';
 	private _editContextPrimarySelection: Selection = new Selection(1, 1, 1, 1);
+	/**
+	 * Marks the native EditContext as needing a refresh of its text / selection.
+	 *
+	 * We defer the actual update until render time so that a single large edit
+	 * (e.g. "Transform to Uppercase" over a many-MB selection) does not
+	 * synchronously hand a huge string to the native EditContext from inside
+	 * the model's content-change / view's cursor-state event handlers — which
+	 * used to freeze the editor for seconds. See issue #262473.
+	 */
+	private _editContextNeedsRefresh: boolean = false;
 
 	// Overflow guard container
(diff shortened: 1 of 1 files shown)


alexdima (approved):

alexdima:
LGTM.

@aiday-mar can you pls review and try this change out, specifically that IME still works?

roblourens (approved):

dmitrivMS (approved):

aiday-mar:
Tested. Thanks for the PR!

PR #310966, opened 2026-04-17
fix(chat): preserve scroll anchor when user toggles collapsible thinking/subagent sections (#274731)

### LINKED ISSUE #274731: Expand Thinking section pushes chat layout upwards
1. vscode repo, agent mode "Look at the git changes in the last 24 hours. And try to find the change that broke the terminal tool (that agent uses). Now the terminal tool gets stuck.
So you do not need to fix it, just find the commit that is the culprit. Think hard! And just come up with a plan first."
2. Notice thinking section collapsed. Scroll all the way to the bottom of your response
3. Expand thinking section, notice expansion is above 🐛  It should push the layout down, so I can actually see what I expanded

https://github.com/user-attachments/assets/00b0c39f-9073-47d0-82c4-9e371eeab7ce

fyi @bpasero

maruthang:
**What** — Manually expanding or collapsing a thinking/subagent/hook/markdown section in the chat list no longer snaps the viewport to the bottom. Previously-visible content stays visually put; the expand grows downward into unused space.

**Why** — `ChatListWidget._updateElementHeight` auto-scrolls the list to the bottom whenever a row grows and the user was already pinned to the bottom. That's correct during streaming (content growing from below must stay visible) but wrong when the user clicks to expand a collapsed section: the expanded content appears *above* the anchor, so every added pixel shifts everything else upward (#274731).

**How** — Distinguish user-driven from programmatic height changes. `ChatCollapsibleContentPart` fires a bubbling `CustomEvent('chatCollapsibleUserToggle')` from its click handler; `ChatListWidget` records the timestamp on receive and skips `_withPersistedAutoScroll` within a 250 ms window. Programmatic `setExpanded` (used during streaming) doesn't dispatch the event, so auto-scroll during streaming is unchanged.

**Test plan** — Manual (from the issue): agent-mode prompt "Look at the git changes in the last 24 hours…", wait for streaming, scroll to bottom, click to expand the thinking section. Pre-fix: viewport jumps upward. Post-fix: content above stays put. Also verified streaming auto-scroll still works. Added 2 unit tests in `chatThinkingContentPart.test.ts` asserting the click dispatches exactly one bubbling event.

Applies to every `ChatCollapsibleContentPart` subclass, not just thinking.

Fixes #274731


justschen:
must've been out when this was assigned to me - giving this a look for this coming iteration, thanks!

maruthang:
Closing as superseded: upstream commit 489f3ab874f ("chat ux: collapse animations, thinking cleanup", #325198) independently landed this fix — the same `chatCollapsibleUserToggle` event on toggle plus a fuller scroll-restoration system (UserToggleResizeTracker) in chatListWidget. The scroll-anchor issue this PR targeted is handled there now, so this is no longer needed.

PR #310969, opened 2026-04-17
NES: consolidate speculative request cancellation with type-through trajectory check

ulugbekna:


ulugbekna:
Thanks @copilot-pull-request-reviewer — addressing the three low-confidence comments:

1. **Redundant `cancelAll(Replaced)` before `setPending`** — valid; removed in 51b61c02670. `setPending` already cancels any prior pending speculative as `Replaced`, and dropping the early cancel keeps the prior speculative available for reuse during the async `_createSpeculativeRequest` window.

2. **`handleRejection` still cancels with `Rejected`** and 3. **`handleIgnored` still cancels with `IgnoredDismissed`** — these were a misreading of the PR description's table. The `After` column was meant to describe each *trigger's* behavior, not whether the speculative is kept alive. Both `handleRejection` and the not-superseded `handleIgnored` correctly continue to cancel, since once the user has rejected/dismissed the suggestion, the speculative's bet on the post-accept document state can never be redeemed. PR description has been clarified to make this column unambiguous.

roblourens (dismissed):

roblourens (approved):

PR #310970, opened 2026-04-17
fix(chat): route mode switches to the originating pane when widgets share a session (#275200)

### LINKED ISSUE #275200: switch modes with multiple chat windows
Testing #274957

this video outlines the steps but if my focus is in the chat panel, and I try and change the mode of the editor it changes the wrong mode

https://github.com/user-attachments/assets/ea8d9256-0630-4877-abce-3e5090f1f725

maruthang:
**What** — When two chat widgets share a session (e.g. a chat-view pane and a chat-editor pane after "Move Chat to Editor Area"), changing the mode in one no longer silently flips the mode in the other.

**Why** — `ToggleChatModeAction` resolved the target widget via `getWidgetBySessionResource(...)`, which returns the *first* match. With two widgets on the same session, the mode change always landed on whichever registered first — not the one whose picker was clicked (#275200).

**How** — Each `ChatInputPart` has a unique `inputUri`, and `IChatWidgetService.getWidgetByInputUri` already disambiguates by it. Threaded `inputUri` through `IModePickerDelegate` → `ModePickerActionItem.run` → `IToggleChatModeArgs` so the action can resolve the exact widget. The existing `sessionResource` path is preserved as a fallback for older callers.

**Test plan** — Two new unit tests in `chatExecuteActions.test.ts`: (1) with two widgets sharing a session, a toggle dispatched with `inputUri` targets widget B not widget A; (2) without `inputUri`, the existing `sessionResource` resolution still works. Manual: open a chat session, "Move Chat to Editor Area", pick a different mode in the editor's picker — the panel mode should stay unchanged.

Fixes #275200


justschen:
thanks! will give this a look!

PR #310974, opened 2026-04-17
Add improved transition handling for workbench startup

mrleemurray:
Introduce a new layout class to manage part-reveal transitions during startup. Refactor the transition handling to activate after reaching the Eventually lifecycle phase, ensuring smoother animations. Address unexpected errors when applying the new class in the Workbench.

### DIFF
--- a/src/vs/sessions/browser/media/style.css
+++ b/src/vs/sessions/browser/media/style.css
@@ -432,10 +432,10 @@
 }
 
 @media (prefers-reduced-motion: reduce) {
-	.agent-sessions-workbench .part.auxiliarybar,
-	.agent-sessions-workbench .part.panel,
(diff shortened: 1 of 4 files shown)


lszomoru:
@mrleemurray, are we sure this is the way to go? One low hanging fruit as far as I can tell is that the auxiliary bar is not being shown by default and it is being revealed just a bit after the workbench has been rendered. Could we just show the auxiliary bar by default as that would help with the resizing of the chat bar card?

lszomoru:
@mrleemurray, we should also show the "Files" tab by default when loading the workbench.

mrleemurray:
@lszomoru I have made some updates to make the aux bar visible on load - please re-review

lszomoru:
@mrleemurray, nice! Will take a look!

PR #310977, opened 2026-04-17
GitHub: fix lowercase pull_request_template.md not found in .github/ folder

Nedunchezhiyan-M:
Fixes #310976

## What

One-character fix in `extensions/github/src/pushErrorHandler.ts`.

## Root Cause

`PR_TEMPLATE_FILES` had a copy-paste error on the `.github` row:

```ts
// Before (broken):
{ dir: '.github', files: ['PULL_REQUEST_TEMPLATE.md', 'PULL_REQUEST_TEMPLATE.md'] }
//                          ^^^ uppercase duplicated — lowercase never checked

// After (fixed):
{ dir: '.github', files: ['pull_request_template.md', 'PULL_REQUEST_TEMPLATE.md'] }
//                          ^^^ lowercase first, matching root and docs/ rows
```

The root (`.`) and `docs/` directories correctly carry both `pull_request_template.md` and `PULL_REQUEST_TEMPLATE.md`. The `.github/` row accidentally had the uppercase string in both positions, so any repo using the lowercase filename inside `.github/` would never have that template offered during PR creation.

## Testing

- Repository with `.github/pull_request_template.md` (lowercase) → template is now found and offered.
- Repository with `.github/PULL_REQUEST_TEMPLATE.md` (uppercase) → still found (no regression).
- Repository with `pull_request_template.md` in root or `docs/` → unchanged.

### DIFF
--- a/extensions/github/src/pushErrorHandler.ts
+++ b/extensions/github/src/pushErrorHandler.ts
@@ -22,7 +22,7 @@ export function isInCodespaces(): boolean {
 const PR_TEMPLATE_FILES = [
 	{ dir: '.', files: ['pull_request_template.md', 'PULL_REQUEST_TEMPLATE.md'] },
 	{ dir: 'docs', files: ['pull_request_template.md', 'PULL_REQUEST_TEMPLATE.md'] },
-	{ dir: '.github', files: ['PULL_REQUEST_TEMPLATE.md', 'PULL_REQUEST_TEMPLATE.md'] }
+	{ dir: '.github', files: ['pull_request_template.md', 'PULL_REQUEST_TEMPLATE.md'] }
 ];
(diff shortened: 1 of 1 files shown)


extensions/github/src/pushErrorHandler.ts
dmitrivMS, line 37: This line does case-insensitive `indexOf`, so I don't see how this change preserves PULL_REQUEST_TEMPLATE.md casing. Have you tested the upper case scenario?

dmitrivMS (changes requested):

dmitrivMS:
@Nedunchezhiyan-M I suggest adding explicit tests for the 2 casing cases to verify it really works.

PR #310979, opened 2026-04-17
fix(chat): cancel in-flight streaming tool invocations when response is cancelled (#288701)

### LINKED ISSUE #288701: Editing files and spinner remains after stopping session
This is the state after I pressed the stop button:

Image

Version: 1.109.0-insider (user setup)
Commit: f46196091d2fa0069d229cb638869b4ab95392cf
Date: 2026-01-16T12:14:27.330Z
Electron: 39.2.7
ElectronBuildId: 13098910
Chromium: 142.0.7444.235
Node.js: 22.21.1
V8: 14.2.231.21-electron.0
OS: Windows_NT x64 10.0.26200

maruthang:
**What** — Pressing Stop while a tool call is still streaming partial input now also transitions that tool invocation to `Cancelled`, so the "Editing files" label and spinner on the thinking row clear immediately instead of hanging on-screen.

**Why** — `ChatResponseModel.cancel()` only flipped its own state. Any `ChatToolInvocation` that was in `StateKind.Streaming` (LM still producing tool arguments) stayed there forever. UI consumers (the autorun in `chatThinkingContentPart.ts`) kept treating the tool as in-progress, so the spinner/shimmer never cleared (#288701).

**How** — Inside `cancel()`, walk `this._response.value` for `ChatToolInvocation` parts and call the existing `cancelFromStreaming(ToolConfirmKind.Skipped)` on each. No new API surface. The `ToolConfirmKind` symbol was already available; just added it to the existing import line.

**Test plan** — New regression test in `chatModel.test.ts` (`ChatResponseModel` suite): create a streaming `ChatToolInvocation` on a `ChatModel`, assert state `Streaming`, call `cancelRequest(request)`, assert state `Cancelled` and `IChatToolInvocation.isComplete` true. Manual: trigger a multi-file edit via an agent-mode request, press Stop while the thinking row shows "Editing files"; the label/spinner clears immediately.

Fixes #288701

### DIFF
--- a/src/vs/workbench/contrib/chat/common/model/chatModel.ts
+++ b/src/vs/workbench/contrib/chat/common/model/chatModel.ts
@@ -30,7 +30,7 @@ import { CellUri, ICellEditOperation } from '../../../notebook/common/notebookCo
 import { ChatRequestToolReferenceEntry, IChatRequestVariableEntry, isImplicitVariableEntry, isStringImplicitContextValue, isStringVariableEntry } from '../attachments/chatVariableEntries.js';
 import { migrateLegacyTerminalToolSpecificData } from '../chat.js';
 import { ChatPerfMark, markChat } from '../chatPerf.js';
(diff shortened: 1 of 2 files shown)


connor4312 (approved):

lramos15 (approved):

connor4312:
thanks!

PR #310980, opened 2026-04-17
Agents - add maximize/restore actions to the editor part

lszomoru:

### DIFF
--- a/src/vs/sessions/browser/workbench.ts
+++ b/src/vs/sessions/browser/workbench.ts
@@ -22,7 +22,7 @@ import { IEditorService } from '../../workbench/services/editor/common/editorSer
 import { IPaneCompositePartService } from '../../workbench/services/panecomposite/browser/panecomposite.js';
 import { IViewDescriptorService, ViewContainerLocation } from '../../workbench/common/views.js';
 import { ILogService } from '../../platform/log/common/log.js';
-import { IInstantiationService, ServicesAccessor } from '../../platform/instantiation/common/instantiation.js';
+import { createDecorator, IInstantiationService, ServicesAccessor } from '../../platform/instantiation/common/instantiation.js';
 import { ITitleService } from '../../workbench/services/title/browser/titleService.js';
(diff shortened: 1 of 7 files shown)


src/vs/sessions/browser/workbench.ts
```diff
@@ -102,7 +104,16 @@ interface IPartVisibilityState {
...
 
-export class Workbench extends Disposable implements IWorkbenchLayoutService {
+export interface IAgentWorkbenchLayoutService extends IWorkbenchLayoutService {
+	isEditorMaximized(): boolean;
+	setEditorMaximized(maximized: boolean): void;
+
+	readonly onDidChangeEditorMaximized: Event<void>;
+}
+
+export const IAgentWorkbenchLayoutService = createDecorator<IAgentWorkbenchLayoutService>('agentWorkbenchLayoutService');
```
sandy081: You can use `refineServiceDecorator` pattern instead of creating another new independent layout service decorator. See https://github.com/microsoft/vscode/blob/8b28af1897cd1a557f288e62995174b8804284c8/src/vs/workbench/services/layout/browser/layoutService.ts#L19 for example.

src/vs/sessions/contrib/changes/browser/changesViewActions.ts
```diff
@@ -201,3 +203,59 @@ class OpenPullRequestAction extends Action2 {
...
+	constructor() {
+		super({
+			id: MaximizeMainEditorPartAction.ID,
+			title: localize2('maximizeMainEditorPart', "Maximize Editor"),
+			icon: Codicon.screenFull,
+			f1: false,
+			menu: {
+				id: MenuId.EditorTitle,
+				group: 'navigation',
+				order: 100001,
```
sandy081, line 219: Just to confirm this is to push these actions to the far right?

lszomoru: Correct, we need to have them appear right before the `...` action.

sandy081 (approved):

PR #310987, opened 2026-04-17
Merge conflict: use startsWith for splitter marker detection (diff3/zdiff3 support)

### LINKED ISSUE #310985: Merge conflict: splitter line not recognized when using git diff3/zdiff3 conflict style
## Summary

When using `merge.conflictstyle = diff3` or `merge.conflictstyle = zdiff3`, git can produce `=======` splitter lines with trailing text (e.g., `======= branch-name`). The merge-conflict extension uses strict equality (`line.text === '======='`) to detect the splitter, so any splitter with trailing text is silently skipped.

The result is that the entire conflict is not recognized — it has no decorations, no CodeLens actions, and is effectively invisible to the extension.

## Steps to Reproduce

1. Set `git config merge.conflictstyle zdiff3`
2. Create a merge conflict in a repository
3. Open the conflicting file in VS Code
4. Observe: the conflict has no CodeLens actions (Accept Current Change / Accept Incoming Change) and no conflict decorations

## Expected Behavior

The `=======` splitter should be recognized regardless of trailing text, consistent with how `<<<<<<<` and `>>>>>>>` markers are detected using `startsWith`.

## Root Cause

In `extensions/merge-conflict/src/mergeConflictParser.ts` line 62:

```typescript
// Current (broken for diff3/zdiff3):
else if (currentConflict && !currentConflict.splitter && line.text === splitterMarker) {

// Should be (consistent with all other markers):
else if (currentConflict && !currentConflict.splitter && line.text.startsWith(splitterMarker)) {
```

All other markers (`<<<<<<<`, `|||||||`, `>>>>>>>`) use `startsWith`. Only `=======` uses strict equality, which is inconsistent and breaks diff3/zdiff3 style conflicts.

## VS Code Version

Reproducible in current main branch.

Nedunchezhiyan-M:
Fixes #310985

## What

Changed the splitter marker (`=======`) detection in `mergeConflictParser.ts` from strict equality (`===`) to prefix matching (`startsWith`), consistent with how `<<<<<<<`, `|||||||`, and `>>>>>>>` markers are already detected.

## Why

When `merge.conflictstyle = diff3` or `merge.conflictstyle = zdiff3` is configured, git generates splitter lines with trailing text (e.g., `======= branch-name`). The strict `===` comparison caused these splitters to be silently skipped, meaning the entire conflict was dropped from the parser — leaving the file without any merge conflict decorations or CodeLens actions (Accept Current/Incoming/Both).

Using `startsWith` is the correct approach for all conflict markers, as git always starts them with the fixed prefix but may append branch names or other metadata.

## Testing

- Set `git config merge.conflictstyle zdiff3` in a test repo
- Create a merge conflict
- Open the conflicting file in VS Code
- Verify that CodeLens actions appear on the conflict markers and decorations are shown correctly

### DIFF
--- a/extensions/merge-conflict/src/mergeConflictParser.ts
+++ b/extensions/merge-conflict/src/mergeConflictParser.ts
@@ -59,7 +59,7 @@ export class MergeConflictParser {
 				currentConflict.commonAncestors.push(line);
 			}
 			// Are we within a conflict block and is this a splitter? =======
-			else if (currentConflict && !currentConflict.splitter && line.text === splitterMarker) {
+			else if (currentConflict && !currentConflict.splitter && line.text.startsWith(splitterMarker)) {
 				currentConflict.splitter = line;
 			}
 			// Are we within a conflict block and is this a footer? >>>>>>>


dmitrivMS:
@Nedunchezhiyan-M We should probably add or update a test for this?

dmitrivMS (changes requested):
Please add tests

PR #311011, opened 2026-04-17
Do not move fg terminal to the bg when it needs input, fix regression

meganrogge:
## Root Cause

In #308587, the LLM-based prompt detection in the output monitor was replaced with a simpler `onDidDetectInputNeeded` event. When this event fires during a **foreground** terminal execution, the race handler in `runInTerminalTool` converts the terminal to background mode:

```typescript
isBackgroundExecution = true;
toolTerminal.isBackground = true;
this._sessionTerminalAssociations.delete(chatSessionResource);
```

This was done so "the terminal stays alive for send_to_terminal," but it's unnecessary — `send_to_terminal` already supports foreground terminals via `terminalId`, and the execution stays alive in `_activeExecutions` regardless of fg/bg status.

## Fix

Removed the three lines that promote the terminal to background. The `inputNeeded` handler now returns early with output (so the agent can provide input) but **preserves the terminal as foreground** and keeps the session association intact. Updated the `finally` block condition from `isBackgroundExecution` to `isBackgroundExecution || didInputNeeded` so the execution isn't disposed while the agent handles input.

## Issues Addressed

1. **Terminal proliferation** — Each interactive prompt (password, `[Y/n]`, "press any key") permanently promoted the fg terminal to bg. The next `run_in_terminal` couldn't find a cached fg terminal and spawned a new one. Over a session with multiple interactive commands, terminals accumulated instead of being reused.

2. **Benchmark regression (April 13)** — Extra terminal creation per command adds shell spawn + shell integration setup latency. This directly inflated terminal benchmark times.

3. **Spurious "moved to background" messages** — The result text included `Note: This terminal execution was moved to the background` for terminals that the user expected to remain foreground, since `isBackgroundExecution` was true but `executionOptions.persistentSession` was false.

4. **User-reported premature fg→bg transfer** — Users observed their foreground terminal unexpectedly becoming a background terminal when a command hit an interactive prompt, losing the fg terminal reuse behavior.

### DIFF
--- a/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/runInTerminalTool.ts
+++ b/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/runInTerminalTool.ts
@@ -1436,14 +1436,12 @@ export class RunInTerminalTool extends Disposable implements IToolImpl {
 
 				if (raceResult.type === 'inputNeeded') {
 					// Output monitor detected the terminal is waiting for input.
-					// Convert to background so the terminal stays alive for send_to_terminal.
+					// Return output to the agent so it can provide input via
+					// send_to_terminal. The terminal stays foreground so it is
+					// reused by subsequent run_in_terminal calls in this session.
 					this._logService.debug(`RunInTerminalTool: Output monitor detected input needed in foreground terminal, returning output to agent`);
 					error = 'inputNeeded';
 					didInputNeeded = true;
-					isBackgroundExecution = true;
-					toolTerminal.isBackground = true;
-					this._sessionTerminalAssociations.delete(chatSessionResource);
-					await this._associateProcessIdWithSession(toolTerminal.instance, chatSessionResource, termId, toolTerminal.shellIntegrationQuality, true);
 					// Read output directly from the execution rather than from pollingResult,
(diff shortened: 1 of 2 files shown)


stasyu2009-ux:
Цей цикл невірно зроблений тому код працює невірно і довго.

stasyu2009-ux:
export const handleFetch = (event: FetchEvent, sw: ServiceWorkerGlobalScope = serviceWorker) => {
  try {
    // Only intercept GET requests
    if (event.request.method !== 'GET') return
    if (!getCurrentUser()) {
      const url = new URL(sw.location.href, sw.location.origin)
      const errorsUrl = url.searchParams.get('errors_url')
      setCurrentUser(url.searchParams.get('current_user'))
      setRelease(url.searchParams.get('release'))
      setActorId(url.searchParams.get('actor_id'))
      if (errorsUrl) {
        initServiceWorkerFailbot({
          errorsUrl,
          user: getCurrentUser() || undefined,
          catalogService: 'github/issues',
          release: getRelease() || undefined,
          actorId: getActorId() || undefined,
        })
      }

connor4312 (approved):

stasyu2009-ux:
Все я перевірив можна відправляти.

stasyu2009-ux:
Все пройшло успішно.

Issue #114425, opened 2021-01-15
VS Code opens folders instead of the file manager on some Linux distros


ishfaqfariq:
I'm using kali linux latest version.And after installing VSC, the drive paths on my desktop are opening in VSC. The paths for drives on my desktop opens just as if its a text file. Its just happing on desktop. I can access my drives from toolbar though. But still its quite strange and uncomfortable. After talking with team, they suggested me to use VSC insider. And due to some reasons, i wasn't able to reply to them for a week. And today i checked and its doing same on both. Kindly Check it out.

https://user-images.githubusercontent.com/30369065/104754725-60b4fd00-577b-11eb-9ad9-6943c8fc3500.mp4

ishfaqfariq:
https://user-images.githubusercontent.com/30369065/104754906-9659e600-577b-11eb-97cb-9f1fe84ec73c.mp4

ishfaqfariq:
[Screencast 2021-01-15 21:47:28.mp4.tar.gz](https://github.com/microsoft/vscode/files/5821880/Screencast.2021-01-15.21.47.28.mp4.tar.gz)

ishfaqfariq:
In case, if you aren't able to see the recording. kindly check this link out:
https://drive.google.com/file/d/1Lr9rsWXRHYwd_SNdRdcOY8TbLjepQ5Zu/view?usp=sharing
I've recorded the issue in it.

Tyriar:
Last I looked into this, distros do all sorts of things when you register a mime type handler in a desktop file. Here is the responsible line, but AFAIK this is how we're meant to do it:

https://github.com/microsoft/vscode/blob/36929d3b59e37cd0f831b5db91c0b29e199143e9/resources/linux/code.desktop#L11

ishfaqfariq:
I'm not much technical, I'm new to this distro too. Can you please explain it more. How can i fix this issue? Or you guys will fix it in new update ?

Tyriar:
Ideally an expert who knows about this will show up to tell us what we're doing wrong. Right now we register as a handler of directories with `inode/directory` above, the question is why are we taking priority over the builtin file manager.

ishfaqfariq:
So things are beyond my league. Sorry I wasn't much of a help.

palexdev:
This must be fixed as a top priority
I can't open folders because VSCode opens instead, this is kinda ridicolous

rzhao271:
[...]

PR #311053, opened 2026-04-17
Add plaintext and directory mimetypes to Linux desktop entry

### LINKED ISSUE #15741: Allow code to open directories in nautilus file manager

Tech-Tac:
This allows users to open files and folders with VSCode directly from their file manager, matching the behavior of other text editing applications on Linux.

Fixes #15741

### DIFF
--- a/resources/linux/code.desktop
+++ b/resources/linux/code.desktop
@@ -8,7 +8,7 @@ Type=Application
 StartupNotify=false
 StartupWMClass=@@NAME_SHORT@@
 Categories=TextEditor;Development;IDE;
-MimeType=application/x-@@NAME@@-workspace;
+MimeType=application/x-@@NAME@@-workspace;text/plain;inode/directory
 Actions=new-empty-window;
 Keywords=vscode;
 


rzhao271 (commented):
Likely needs to be tested across multiple distros, see https://github.com/microsoft/vscode/pull/209510

Tech-Tac:
As long as we are following the standard \[[1](https://specifications.freedesktop.org/desktop-entry/latest/mime-types.html), [2](https://specifications.freedesktop.org/mime-apps/latest/default.html)\], we should be fine.

rzhao271:
I believe we were following those standards before, and then https://github.com/microsoft/vscode/issues/114425 happened.
The issue is we want to register VS Code as an available application for those types but not accidentally become the default application for those types. Your second link seems to suggest a way to register a type as the default, which is the opposite of what we want. The ideal solution would be a post-installation command that establishes our application as not the default for some types.

Tech-Tac:
There is not much we can do in any case, this sounds like bad behavior in the mentioned environment and nothing we can directly influence. We are simply following the spec.

LucasOe:
Users can manually change the default application by editing `~/.config/mimeapps.list`:

```
[Default Applications]
inode/directory=org.gnome.Nautilus.desktop
```

At the very least the desktop entry should contain `text/plain`, so VSCode gets suggested as an option for text files. For reference, Zed has to deal with the same issue: https://github.com/zed-industries/zed/pull/39076

PR #311054, opened 2026-04-17
agentHost: fix bugs around message handling

connor4312:
- Fix a bug where a sessions were not restored correctly because they had in-progress data (that is the move of the `providedSession.isCompleteObs?.get()`)
- Fix a bug where sessions that supported progress streaming would not have sendable messages if they were already complete (that is the addition of `this._pendingRequests.deleteAndDispose(model.sessionResource);`)
- Fix a bug in the agent host where we didn't provide `onDidStartServerRequest` consistently which often prevented multi-client messaging from working
- DRY up some logic and add tests

### DIFF
--- a/src/vs/workbench/api/common/extHostChatSessions.ts
+++ b/src/vs/workbench/api/common/extHostChatSessions.ts
@@ -1127,16 +1127,16 @@ export class ExtHostChatSessions extends Disposable implements ExtHostChatSessio
 		}
 
 		const handler = controllerData.controller.getChatSessionInputState;
-		if (!handler || !sessionResourceComponents) {
+		if (!handler) {
 			return undefined;
 		}
-		const sessionResource = URI.revive(sessionResourceComponents);
-		const inputState = await handler(isUntitledChatSession(sessionResource) ? undefined : sessionResource, { previousInputState: undefined }, token);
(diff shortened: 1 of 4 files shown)


src/vs/workbench/api/common/extHostChatSessions.ts
```diff
@@ -1119,18 +1119,17 @@ export class ExtHostChatSessions extends Disposable implements ExtHostChatSessio
 		controllerData.onDidChangeChatSessionItemStateEmitter.fire(item);
 	}
 
-	async $provideChatSessionInputState(controllerHandle: number, sessionResourceComponents: UriComponents, token: CancellationToken): Promise<vscode.ChatSessionProviderOptionGroup[] | undefined> {
+	async $provideChatSessionInputState(controllerHandle: number, sessionResourceComponents: UriComponents | undefined, token: CancellationToken): Promise<vscode.ChatSessionProviderOptionGroup[] | undefined> {
```
connor4312, line 1122: this was just another error I saw logged -- the interface types the param as `UriComponents | undefined` and some callers use that, but TS allowed params to be narrowed to just `UriComponents` which caused a runtime error. cc @mjbvz I think this is your code

osortega (dismissed):

Yoyokrazy (approved):

PR #311065, opened 2026-04-17
Restore autopilot branch and shorten steering text in terminal tool results

meganrogge:
Fixes https://github.com/microsoft/vscode-copilot-evaluation/issues/3446

Fix eval regression by:
1. Restoring `isAutoApproved` branch at all 4 steering text injection sites so autopilot/benchmark mode uses direct `send_to_terminal` instead of forcing through `askQuestions`
2. Shortening all steering text from ~100 words to ~25 words per site
3. Shortening timeout wrapper text
4. Removing dynamic terminal UUID from steering text to preserve prompt cache — the UUID is already in `toolMetadata.id` and the notification message header

The verbose steering text ("send each answer one at a time using `send_to_terminal`… calling `get_terminal_output` between each") went from 0% → 50% prevalence in regression trajectories, with models echoing it as narration instead of acting. The embedded UUID also broke prompt caching, causing ~31k tokens to be re-sent uncached per turn.

Addresses bb77eceb5c6 (#309050) and 8a73ca2362e (#309583) regressions.

### DIFF
--- a/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/runInTerminalTool.ts
+++ b/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/runInTerminalTool.ts
@@ -1393,7 +1393,11 @@ export class RunInTerminalTool extends Disposable implements IToolImpl {
 						resultText += `${outputAnalyzerMessage}\n`;
 					}
 					resultText += pollingResult.output;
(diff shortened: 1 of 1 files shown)


Yoyokrazy (dismissed):

justschen (approved):

src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/runInTerminalTool.ts
```diff
@@ -2187,7 +2198,10 @@ export class RunInTerminalTool extends Disposable implements IToolImpl {
 				}
 				lastInputNeededOutput = currentOutput;
 				lastInputNeededNotificationTime = now;
-				const inputAction = `You MUST call the vscode_askQuestions tool to ask the user what values to provide for all anticipated prompts at once (include upcoming prompts you can predict from the command, not just the currently visible one). Do NOT respond with a text message asking the user — use the tool. Then send each answer one at a time using ${TerminalToolId.SendToTerminal} with id "${termId}", calling ${TerminalToolId.GetTerminalOutput} between each to read the next prompt before sending the next answer.`;
+				const isAutoApproved = isSessionAutoApproveLevel(chatSessionResource, this._configurationService, this._chatWidgetService, this._chatService);
```
zhichli, line 2201: can this value be saved and reused rather than calling the method in multiple places in this file ?

PR #311150, opened 2026-04-18
terminal: restore NODE_OPTIONS and NODE_REPL_EXTERNAL_MODULE on Windows

### LINKED ISSUE #231076: VS Code is ignoring the environment setting for node js --max-old-space-size
VS Code is ignoring the environment setting for node js --max-old-space-size

- VS Code Version: 1.94.2
- Node.js: v20.18.0
- OS Version: 
Edition	Windows 11 Pro
Version	23H2
Installed on	‎9/‎10/‎2023
OS build	22631.4317
Experience	Windows Feature Experience Pack 1000.22700.1041.0

Steps to Reproduce:

1. Setup node options for user/system
Image

2. Open cmd to validate
Image

3. Open ps to validate
Image

4. Open VS Code to validate - value is ignored/overwritten with something else
Image

Thanks,
Regards,
Evgeny

sungwoonpark0502:
Fixes #231076

On Windows, when VS Code is launched via CLI with `NODE_OPTIONS` set (e.g. `NODE_OPTIONS=--max-old-space-size=8192 code .`), the environment variable gets stripped before the app starts (tracked in electron/electron#40770). VS Code already works around this on macOS by saving the value in `VSCODE_NODE_OPTIONS` and restoring it in the integrated terminal — but this same logic was missing for Windows.

**Change:** Extend the `isMacintosh` guard to `isMacintosh || isWindows` so Windows users also get `NODE_OPTIONS` (and `NODE_REPL_EXTERNAL_MODULE`) restored in the integrated terminal.

## Test plan

- [ ] On Windows, set `NODE_OPTIONS=--max-old-space-size=8192` and launch VS Code via `code .`
- [ ] Open the integrated terminal
- [ ] Run `echo $env:NODE_OPTIONS` (PowerShell) or `echo %NODE_OPTIONS%` (cmd) — should show `--max-old-space-size=8192`
- [ ] Verify macOS behavior is unchanged

### DIFF
--- a/src/vs/workbench/contrib/search/browser/searchActionsSymbol.ts
+++ b/src/vs/workbench/contrib/search/browser/searchActionsSymbol.ts
@@ -9,6 +9,7 @@ import * as Constants from '../common/constants.js';
 import { Action2, MenuId, registerAction2 } from '../../../../platform/actions/common/actions.js';
 import { KeybindingWeight } from '../../../../platform/keybinding/common/keybindingsRegistry.js';
 import { KeyCode, KeyMod } from '../../../../base/common/keyCodes.js';
+import { isWeb } from '../../../../base/common/platform.js';
 import { IQuickInputService } from '../../../../platform/quickinput/common/quickInput.js';
 
 //#region Actions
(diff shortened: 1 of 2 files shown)


sungwoonpark0502:

PR #311242, opened 2026-04-19
Fix Copilot Welcome Page title line height

Jah-yee:
Fixes line-height for the Copilot Welcome Page title to prevent text overlap when the title wraps to multiple lines.

Before: Without explicit line-height, the title text would overlap when wrapped in narrow layouts.
After: Adding `line-height: 1` ensures proper spacing between wrapped title lines.

**Before:**
image

**After:**  
image

Closes #297827


osortega:
This is fixed already

Issue #284417, opened 2025-12-19
Copilot attempts writing files in windows format on linux remote host via ssh


nbuhay:
# VS Code Remote-SSH File Tool Path Resolution Bug

## Issue Summary
GitHub Copilot Agent file editing tools fail completely in Remote-SSH workspace. Tools report success but files are never created on the remote filesystem. The workspace path is registered as `\\opt\\apps\\dealer` (with backslashes) instead of proper remote URI scheme, causing all file operations to be silently dropped.

## Environment

**Local System:**
- OS: Windows
- VS Code Version:
-   Version: 1.107.0 (user setup)
-   Commit: 618725e67565b290ba4da6fe2d29f8fa1d4e3622
-   Date: 2025-12-10T07:43:47.883Z
-   Electron: 39.2.3
-   ElectronBuildId: 12895514
-   Chromium: 142.0.7444.175
-   Node.js: 22.21.1
-   V8: 14.2.231.21-electron.0
-   OS: Windows_NT x64 10.0.22631
- Remote-SSH Extension Version: 0.122.0

**Remote System:**
- OS: Linux
- Connection: SSH
- Workspace Path (actual): `/opt/apps/dealer`
- Workspace Path (VS Code shows): `\opt\apps\dealer` (backslashes)

**Extensions:**
- GitHub Copilot: Running on UI (required for Chat to function)
- GitHub Copilot Chat: Running on UI
- Remote-SSH: Installed

### Symptoms
1. **All file read/write operations fail** with error: `"File \\opt\apps\dealer\[filename] is outside of the workspace"`
2. **File creation reports false success** - tools report "Successfully edited" but files don't exist on filesystem
3. **User must approve each operation** - VS Code prompts "Allow edits to sensitive files? The model wants to edit files outside of your workspace"
4. **Search tools work** - `file_search` and `grep_search` successfully locate files
5. **Terminal commands work** - `run_in_terminal` with `/opt/apps/dealer/` paths works perfectly

#### 1. File Search Works
```
file_search("*.md")
Results:
  \\opt\\apps\\dealer\\README.md ✓
  \\opt\\apps\\dealer\\Code_Quality_Guide.md ✓
```

#### 2. File Read Fails
```
read_file("\\opt\\apps\\dealer\\README.md", startLine=1, endLine=10)
Error: "File \\opt\apps\dealer\README.md is outside of the workspace, and not open in an editor, and can't be read"
```

#### 3. File Creation False Success
```
[...]

Issue #11526, opened 2016-09-04
[Proposal] Display all properties (inherited or not) of a given class (in TypeScript)


yahiko00:
Visual Studio Code is a fantastic tool and I am really impressed by all its features despite being quite young.

There is a feature that does not still exists in VS Code which could be really convenient when we are forced to deal with big inheritance hierarchies.

This proposal is based on the TypeScript language, but I am quite sure this could apply to other OOP languages VS Code supports.
## Overview of the Workflow

Let us assume we have a source file _animal.ts_ like this:

``` ts
// animal.ts

abstract class Animal {
    private dna: number;
    name: string;

    constructor(name: string) {
        this.dna = Math.random();
        this.name = name;
    }

    abstract shout(): string;
}
```

This file could be open or not in VS Code, but it is part of the current project.

Let us assume we have another source file _bird.ts_ like this:

``` ts
// bird.ts

class Bird extends Animal {
    constructor() {
        super();
        this.name = "Hector";
    }
    shout(): string {
        return "Cuicui!"; // sorry it is in French... I do not know how a bird shouts in English ^_^
    }
}
```

This file is open in VS Code.

What I would like to be able to do in VS Code is displaying all properties, inherited or not, available for the class `Bird`.

To do so, we could move the cursor on the `Bird` identifier, press F1 key, and run a new _Display all inherited properties_ command.

This new command would open a new file in a new tab called _Bird (all inherited properties)_.

In this example, the content of this file would be the following:

```
class Bird {
    private dna: number;
    name: string;

    super.constructor(name: string) {
        this.dna = Math.random();
        this.name = name;
    }

    constructor() {
        super();
        this.name = "Hector";
    }

    shout(): string {
        return "Cuicui!";
    }
}
```

The content of _Bird (all inherited properties)_ has nothing to do with valid syntax. Its purpose is only informational.
## Some informal rules
- Attributes should be displayed first.
- Type anotations should be rendered unchanged.
- Access modifiers `public`, `protected`, `private` should be rendered unchanged.
- Abstract methods which are not overriden should be rendered.
- All versions of an overriden method should be rendered, with a concatenation of `super` prefixes to distinguish each version in the inheritance hierarchy. For instance `super.method()` would refer to immediate `method()`'s parent, whereas `super.super.method()` would refer to `method()`'s grand-parent, and so on.
- `Method()`'s parent should be displayed before the `method()` itself.
- The same rules for `method()` apply to `constructor()`.

[...]

Issue #3177, opened 2016-02-19
mkdir by nodejs module doesn't work in mac - Extension


DracotMolver:
Hi. 
I'm having an issue with my extension - tested it on Mac and Linux -. I'm creating a folder as follow.
```javascript
var folder = workspace.rootPath+'/test';
fs.mkdir(folder, (err) => {                                   
     console.log(err);
 });
```
This throws me: ***no such file or directory, mkdir 'Documents/Sitios/test'
    at Error (native)***. In Linux i have no problem, but in Mac I'm facing this issue. i don't know if it is Nodejs or Visual Studio Code modules to make extensions.

PR #311250, opened 2026-04-19
prompts: fix vscode-remote path corruption in PromptPathRepresentationService

chencyu:
# Fix vscode-remote path round-tripping for UI-side Copilot hosts

## Problem

This PR fixes prompt filepath round-tripping when GitHub Copilot runs in a UI-side extension host and the workspace itself is remote.

The concrete user scenario is:

```json
"remote.extensionKind": {
  "GitHub.copilot-chat": ["ui"]
}
```

This is a legitimate configuration for air-gapped or restricted-internet remote hosts where the extension cannot authenticate or reach GitHub APIs from the remote side.

In this configuration, Copilot sees real `vscode-remote://...` URIs and currently serializes them with `uri.fsPath`. On Windows UI hosts that applies local path separator conversion, so remote Linux paths such as `/opt/app/file.ts` become `\opt\app\file.ts` in prompts.

When those prompt paths come back from the model, `resolveFilePath()` reconstructs them with `URI.file()`, which loses the original `vscode-remote` scheme and remote authority.

## What Is Already Avoided By VS Code

This is **not** a general bug in all remote scenarios.

When Copilot runs in the remote workspace extension host, VS Code already installs a URI transformer that maps:

- `vscode-remote` on the UI side → `file` on the agent side
- `file` on the agent side → `vscode-remote` on the UI side

So the default remote-workspace-host path is already handled by the platform before `PromptPathRepresentationService` sees the URI.

This means the bug is specifically about **UI-side extension hosts working against remote workspaces**, not about all remote execution modes.

## What Copilot Already Avoids Internally

One major edit flow already has partial protection:

- `editCodeIntent` first tries to match model-returned paths directly against the working set
- only if that fails does it fall back to `resolveFilePath(...)`
- that fallback also carries a predominant scheme

So this PR is **not justified mainly by editCodeIntent**.

## What Still Breaks Without This Fix

Many other consumers do not have the original URI or remote authority available when they parse a filepath string back from prompt/model space.

Examples include:

- agent file tools such as create/edit/insert operations
- code block filepath parsing
- notebook-oriented tools
- conversation-history recovery that parses prior tool arguments

These flows often call `resolveFilePath()` directly, or via `resolveToolInputPath()`, with only a plain path string. In those cases, recovering only the scheme is insufficient — the remote authority is still lost.

That is why the fix belongs in `PromptPathRepresentationService`, which is the shared boundary for prompt path serialization and deserialization.

## Fix

Three focused changes in `PromptPathRepresentationService`:

1. `getFilePath()` returns `uri.path` for `vscode-remote` URIs instead of `uri.fsPath`
2. `resolveFilePath()` first tries to match the incoming filepath against known workspace folders and reconstructs the URI with `URI.joinPath(...)`
3. the workspace-folder matcher handles filesystem roots (`/`, `\\`) and picks the longest matching folder when workspace folders overlap

## Why Both Sides Of The Round Trip Matter

Technically, a folder-prefix match in `resolveFilePath()` would recover many cases even if prompts continued to emit the old backslash representation.

However, that would still leave Copilot emitting a semantically wrong representation of remote Linux paths into prompts and relying on the model to echo that representation back exactly.

Changing `getFilePath()` fixes the representation at the source.
Changing `resolveFilePath()` restores the original URI when the filepath comes back.

Together they make the prompt path contract both correct and robust.

## Scope Of Necessity

- The base round-trip fix is necessary for UI-side extension hosts on remote workspaces
- The workspace-folder reconstruction is necessary because existing callers generally do not carry remote authority separately
- The root-folder and longest-match handling are not specific to the original issue report, but they are necessary to make the new matching strategy correct once introduced

## Prior Art

This is a corrected and more complete version of [PR #3177](https://github.com/microsoft/vscode-copilot-chat/pull/3177).

Compared with that PR, this one:

- does not assume `folders[0]` is sufficient
- only reconstructs remote URIs for paths that actually match a known workspace folder
- handles nested/overlapping workspace folders correctly
- handles filesystem-root workspace folders correctly

## Related Issues

- [Agent mode cannot edit files with remote-ssh extension #11526](https://github.com/microsoft/vscode-copilot-release/issues/11526)
- [Copilot attempts writing files in windows format on linux remote host via ssh #284417](https://github.com/microsoft/vscode/issues/284417)

## Use Case — Air-Gapped / Restricted-Internet Remote Hosts

In laboratory and enterprise environments, remote hosts are often behind strict network policies:

- air-gapped networks
- firewalls that block external APIs
- hosts with no direct internet access

In those environments, running Copilot on the remote side is not always possible.
Running the extension on the local side via `remote.extensionKind` is therefore a valid and important scenario.

This PR fixes the remaining prompt path gap in that configuration.

## Test

Unit tests cover:

- `vscode-remote` path emission via `getFilePath()`
- round-tripping through `resolveFilePath()` via workspace-folder matching
- workspace-folder root handling
- longest-match behavior for nested workspace folders

Current result:

```
✓ src/platform/prompts/test/node/promptPathRepresentationService.spec.ts (35 tests | 1 skipped)
  Test Files  1 passed (1)
```

## Disclosure

This patch (code + commit message + PR description) was fully generated by AI — **Claude Opus 4.6** driven by **GitHub Copilot** in VS Code agent mode. The author reviewed, verified, and approved all changes before submission.

### DIFF
--- a/extensions/copilot/src/platform/prompts/common/promptPathRepresentationService.ts
+++ b/extensions/copilot/src/platform/prompts/common/promptPathRepresentationService.ts
@@ -36,7 +36,9 @@ export interface IPromptPathRepresentationService {
  * When readong a LLM response, use `resolveFilePath` to get the URI from from a `filepath`
  *
  * Do not use this service for other usages than prompts.
- * We currently use the fsPath for local and remote filesystems, and URI.toString() for other schemes.
+ * For `file` URIs we use `fsPath`, for `vscode-remote` URIs we use `path`
+ * (always POSIX, to avoid local-OS separator corruption), and for other
+ * schemes we use `URI.toString()`.
  */
 export class PromptPathRepresentationService implements IPromptPathRepresentationService {
 
@@ -49,7 +51,12 @@ export class PromptPathRepresentationService implements IPromptPathRepresentatio
 	constructor(@IWorkspaceService private readonly workspaceService: IWorkspaceService) { }
 
 	getFilePath(uri: Uri): string {
-		if (uri.scheme === Schemas.file || uri.scheme === Schemas.vscodeRemote) {
+		if (uri.scheme === Schemas.vscodeRemote) {
+			// Use uri.path (always POSIX) instead of uri.fsPath which applies
+			// local OS separators — corrupting remote Linux paths on Windows.
+			return uri.path;
+		}
+		if (uri.scheme === Schemas.file) {
 			return uri.fsPath;
 		}
 		return uri.toString();
@@ -64,6 +71,14 @@ export class PromptPathRepresentationService implements IPromptPathRepresentatio
 	 * @returns The resolved URI or undefined if filepath does not look like a file path or URI.
 	 */
 	resolveFilePath(filepath: string, predominantScheme = Schemas.file): Uri | undefined {
+		// Match against workspace folders first — this preserves scheme and
+		// authority for remote workspaces where a plain path string cannot
+		// encode the full URI (e.g. vscode-remote://ssh-remote+host/...).
+		const folderMatch = this._matchWorkspaceFolder(filepath);
+		if (folderMatch) {
+			return folderMatch;
+		}
+
 		// Always check for posix-like absolute paths, and also for platform-like
 		// (i.e. Windows) absolute paths in case the model generates them.
 		const isPosixPath = filepath.startsWith('/');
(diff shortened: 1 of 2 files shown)


chencyu:
Closing and reopening to retrigger automated review on the current draft state.

chencyu:
Closing this draft in favor of a fresh PR to retrigger automation on the current patch.

PR #311285, opened 2026-04-20
Implement clearTodoList function and hide widget when no todos are present

DonJayamanne:
Introduce a function to clear the todo list at the start of a session and ensure the widget is hidden when there are no todos. This improves the user experience by preventing the display of an empty todo list.

### DIFF
--- a/src/vs/workbench/contrib/chat/browser/widget/chatContentParts/chatTodoListWidget.ts
+++ b/src/vs/workbench/contrib/chat/browser/widget/chatContentParts/chatTodoListWidget.ts
@@ -287,6 +287,7 @@ export class ChatTodoListWidget extends Disposable {
 
 		if (!shouldShow) {
 			this.domNode.classList.remove('has-todos');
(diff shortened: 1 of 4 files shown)


src/vs/workbench/contrib/chat/browser/widget/chatContentParts/chatTodoListWidget.ts
```diff
@@ -287,6 +287,7 @@ export class ChatTodoListWidget extends Disposable {
 
 		if (!shouldShow) {
 			this.domNode.classList.remove('has-todos');
+			this.hideWidget();
```
DonJayamanne, line 290: @bhavyaus i need a way to hide the todo list if one exists.

DonJayamanne: I guess no one has run into this to date

rebornix (approved):

PR #311350, opened 2026-04-20
inline chat: add experiment-controlled reasoning effort and thinking settings

jrieken:
## Summary

Add experiment-controlled settings for reasoning effort and thinking on inline chat requests to GPT models. Previously, inline chat did not pass any `modelCapabilities`, causing the Responses API to silently apply `reasoning: { effort: 'medium', summary: 'detailed' }` defaults - adding latency and token cost without benefit for inline edits.

<details>
<summary>Session Context</summary>

Key decisions from the development session:

- **Default to `low` reasoning effort, thinking disabled**: Inline chat edits are latency-sensitive and don't benefit from deep reasoning. The OpenAI Responses API was silently applying `medium` effort and `detailed` summary on all GPT models that declare `supportsReasoningEffort`, even though inline chat never intended to use reasoning.
- **Experiment-based settings over hardcoded values**: Both `reasoningEffort` and `enableThinking` are `ExperimentBased` config keys so they can be tuned via experiments without code changes. Values validated against the OpenAI API spec (`none`, `minimal`, `low`, `medium`, `high`).
- **Responses API gates summary on `enableThinking === false`**: The `responsesApi.ts` change ensures that when thinking is explicitly disabled, reasoning summaries are suppressed regardless of global experiment flags. The Messages API (Anthropic) already had this gating.
- **`low` chosen over `none`**: Per the OpenAI spec, `none` is only supported on gpt-5.1+ models. `low` works across all model versions, making it a safer default.

</details>

## Changes

- **`configurationService.ts`**: Two new experiment-based config keys: `InlineChatReasoningEffort` (default: `'low'`) and `InlineChatEnableThinking` (default: `false`)
- **`inlineChatIntent.ts`**: Both `InlineChatEditToolsStrategy` and `InlineChatEditHeuristicStrategy` now inject `IConfigurationService`/`IExperimentationService` and pass `modelCapabilities` to `makeChatRequest2`
- **`responsesApi.ts`**: When `enableThinking === false`, reasoning summary is suppressed in the Responses API request body
- **`package.json`/`package.nls.json`**: Setting declarations with `onExp` tag and NLS descriptions

### DIFF
--- a/extensions/copilot/package.json
+++ b/extensions/copilot/package.json
@@ -4231,6 +4231,33 @@
 							"onExp"
 						]
 					},
+					"github.copilot.chat.inlineChat.reasoningEffort": {
+						"type": "string",
+						"default": "low",
+						"enum": [
+							"none",
+							"minimal",
+							"low",
+							"medium",
+							"high"
+						],
+						"markdownDescription": "%github.copilot.config.inlineChat.reasoningEffort%",
+						"tags": [
+							"advanced",
+							"experimental",
+							"onExp"
+						]
+					},
+					"github.copilot.chat.inlineChat.enableThinking": {
+						"type": "boolean",
+						"default": false,
+						"markdownDescription": "%github.copilot.config.inlineChat.enableThinking%",
+						"tags": [
+							"advanced",
+							"experimental",
+							"onExp"
+						]
+					},
 					"github.copilot.chat.debug.requestLogger.maxEntries": {
 						"type": "number",
 						"default": 100,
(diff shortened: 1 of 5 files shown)


jrieken:
cc @bhavyaus changes in `responsesApi.ts`

bpasero (approved):

bhavyaus:
LGTM. Tagging @dileepyavan who owns responses API

dileepyavan:
LGTM

PR #311415, opened 2026-04-20
fix: fix multiple window memory leaky

theanarkh:
The follow steps will trigger the memory leaky.
1. open a window.
2. open a file.
3. right-click on the title of the file, click `move file to new window`.


dmitrivMS:
Thank you for the PR, but his already has been fixed in https://github.com/microsoft/vscode/pull/311824

Issue #310886, opened 2026-04-17
Long question is truncated


roblourens:
There is no hover or button to expand the question

<img width="982" height="272" alt="Image" src="https://github.com/user-attachments/assets/798898f6-6db3-48a7-b22a-5e9072ed1ebd" />

meganrogge:
I think there's already an issue for this. But there should be a hover showing the full thing.

meganrogge:
How did you do this? Cannot reproduce...

<img width="503" height="198" alt="Image" src="https://github.com/user-attachments/assets/c91ff537-366a-4a01-b2e4-2c2241ea85dd" />

roblourens:
Opus 4.7 has been asking long questions. I saw it again today

roblourens:
Thanks for the hover!

roblourens:
Hm, the hover is truncated too

<img width="1022" height="416" alt="Image" src="https://github.com/user-attachments/assets/d794e881-853a-491c-afd3-2fc051c89df4" />

Still seeing this with Opus 4.7 often, seems to just be how it wants to ask questions

roblourens:
Showing the full very long question

<img width="1040" height="633" alt="Image" src="https://github.com/user-attachments/assets/40517193-4687-408a-8c95-0c91dc187110" />

meganrogge:
I don't have 4.7 so couldn't repro 🫠 , do you see it enough that we shoul fix as a candidate?

roblourens:
Everyone should have 4.7, it has been released publicly for awhile

roblourens:
Where's the expand action? Now there's no hover or expand button. Why didn't this get verified?

<img width="1020" height="240" alt="Image" src="https://github.com/user-attachments/assets/78a753ea-953e-431d-afd8-c820d453f2eb" />

meganrogge:
there's no expand action - we decided to just not truncate. can you confirm if you're seeing the full question now? @roblourens

PR #311439, opened 2026-04-20
remove question truncation since we allow collapsing

meganrogge:
addresses #310886 in main

Otherwise, users cannot read the full question.


roblourens (dismissed):
I guess I wonder why we truncate it at all and make me click a button instead of just showing all the content. I definitely need to be able to read it all to answer the question, and 200 characters just isn't very much

meganrogge:
Now that we have expand/collapse maybe we shouldn't truncate it

roblourens (approved):

PR #311501, opened 2026-04-20
Don't expose public mutable properties from actionbar

mjbvz:

### DIFF
--- a/src/vs/base/browser/ui/actionbar/actionbar.ts
+++ b/src/vs/base/browser/ui/actionbar/actionbar.ts
@@ -79,7 +79,9 @@ export class ActionBar extends Disposable implements IActionRunner {
 	};
 
 	// View Items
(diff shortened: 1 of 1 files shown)


roblourens (approved):

src/vs/base/browser/ui/actionbar/actionbar.ts
```diff
@@ -424,32 +426,18 @@ export class ActionBar extends Disposable implements IActionRunner {
...
-				return item.clientWidth;
-			}
-		}
-
-		return 0;
+		return this.actionsList.children.item(index)?.clientWidth ?? 0;
 	}
 
 	getHeight(index: number): number {
-		if (index >= 0 && index < this.actionsList.children.length) {
```
mjbvz, line 438: Whoops wasn't directly part of this change but should be ok to include too

PR #311606, opened 2026-04-21
Use main eslint config for copilot extension too

mjbvz:
This makes the copilot extension use our normal eslint setup instead of its own setup. This matches how other built-in extensions already work. It should let us have more consistent rules across the entire codebase

- Moves copilot eslint config into top level eslint config
- Adopts some standard rules
- Suppress a few new shared rules

cc @joaomoreno for feedback


extensions/copilot/package.json
```diff
@@ -6283,9 +6283,8 @@
 		"watch:tsc-extension-web": "tsc --noEmit --watch --project tsconfig.worker.json",
 		"watch:tsc-simulation-workbench": "tsc --noEmit --watch --project test/simulation/workbench/tsconfig.json",
 		"typecheck": "tsc --noEmit --project tsconfig.json && tsc --noEmit --project test/simulation/workbench/tsconfig.json && tsc --noEmit --project tsconfig.worker.json && tsc --noEmit --project src/extension/completions-core/vscode-node/extension/src/copilotPanel/webView/tsconfig.json",
-		"lint": "eslint . --max-warnings=0",
-		"lint-staged": "eslint --max-warnings=0",
 		"tsfmt": "npx tsfmt -r --verify",
+		"lint": "npx eslint .",
```
rzhao271, line 6287: Doesn't this rely on an external version of eslint that could change configuration formats?

rzhao271 (commented):
LGTM other than usage of npx

extensions/copilot/package.json
mjbvz: I believe it will look up the parent dir chain, in this case finding the eslint copy from `vscode/node_modules/.bin`

rzhao271 (approved):

PR #311700, opened 2026-04-21
Color `*_test.go` files the same way `*.test.ts` are colored

jakebailey:
For JS/TS, test files are shown in another color:

image

I'd love to have this for Go, where it's hard to visually find test files:

image

But, this didn't end up working out of the box as the highlighting system only looks for dot separated strings.

Adding `_` separation enables this:

image

There are other languages that use a `*_test.<ext>` convention or similar, including:

- Elixir: https://hexdocs.pm/ex_unit/ExUnit.html, https://github.com/elixir-lang/elixir/tree/main/lib/elixir/test/elixir
- Ruby: https://rspec.info/features/3-12/rspec-core/command-line/pattern-option/
- Python: https://docs.pytest.org/en/stable/explanation/goodpractices.html#conventions-for-python-test-discovery (though `test_*.py` is also listed, which I have not handled)

Which could benefit from a change like this, though I don't know what color to give those (TS vs Go is easy because they're both blue, so orange seems fine!)

### DIFF
--- a/src/vs/editor/common/services/getIconClasses.ts
+++ b/src/vs/editor/common/services/getIconClasses.ts
@@ -69,6 +69,14 @@ export function getIconClasses(modelService: IModelService, languageService: ILa
 					const dotSegments = name.split('.');
 					for (let i = 1; i < dotSegments.length; i++) {
 						classes.push(`${dotSegments.slice(i).join('.')}-ext-file-icon`); // add each combination of all found extensions if more than one
+
+						// Handle underscore-suffixed base names (e.g., foo_test.go → _test.go)
+						// to enable icon themes to target naming conventions like Go's _test.go files
+						const precedingSegment = dotSegments[i - 1];
+						const lastUnderscore = precedingSegment.lastIndexOf('_');
+						if (lastUnderscore > 0 && lastUnderscore < precedingSegment.length - 1) {
+							classes.push(`${precedingSegment.substring(lastUnderscore)}.${dotSegments.slice(i).join('.')}-ext-file-icon`);
+						}
 					}
 				}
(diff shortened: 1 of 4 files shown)


src/vs/editor/common/services/getIconClasses.ts
```diff
@@ -69,6 +69,14 @@ export function getIconClasses(modelService: IModelService, languageService: ILa
 					const dotSegments = name.split('.');
 					for (let i = 1; i < dotSegments.length; i++) {
 						classes.push(`${dotSegments.slice(i).join('.')}-ext-file-icon`); // add each combination of all found extensions if more than one
+
+						// Handle underscore-suffixed base names (e.g., foo_test.go → _test.go)
+						// to enable icon themes to target naming conventions like Go's _test.go files
```
jakebailey, line 74: This only handles suffixes; this whole suffixing method does not work for something like `test_*.py` where I don't even know what would go into the icon theme.

benvillalobos:
LGTM but I think this has implications for other languages as well. Assigning @aeschli (area owner) to review 

What an agent brought up for me:

> Every file with underscores gets additional CSS classes on its DOM element. In a large repo with thousands of snake_case files, that's non-trivial class list bloat. It's probably negligible in practice, but worth the team being aware of.

> It's a new public contract for icon theme authors. Any third-party theme could start defining _spec.rb, _test.py, etc. This is presumably the intent, but it's a feature surface area expansion that should be a deliberate decision.

jakebailey:
Yeah, probably worth asking some designers or Go people; I can understand why some might be surprised by this. It turned out that only TS/JS do this in VS Code today.

PR #311725, opened 2026-04-21
chat: fix installed url-source plugins after restart

mossgowild:
This fixes the restart recovery path for marketplace-installed plugins with non-relative sources.

When installed.json is rehydrated on startup, PluginMarketplaceService currently recomputes the expected plugin path with getPluginInstallUri for every plugin. That only works for RelativePath sources. URL-source plugins such as superpowers from obra/superpowers-marketplace are installed via getPluginSourceInstallUri, so their metadata is not restored after restart and the plugin appears uninstalled until it is reinstalled.

- use getPluginInstallUri only for RelativePath sources during hydration
- use getPluginSourceInstallUri for non-relative sources
- add a regression test covering url-source hydration from installed.json on startup

Fixes #308504

Testing:
- get_errors on the touched files
- `./scripts/test.sh --run src/vs/workbench/contrib/chat/test/common/plugins/pluginMarketplaceService.test.ts --grep "hydrates installed url-source plugins from installed.json on startup"` → 1 passing


dmitrivMS:
Per issue discussion, it has already been fixed in the current versions.

PR #311762, opened 2026-04-21
Changing upgrade hover for Pro users

cwebster-99:
**Improvements to upgrade messaging:**

* The `createUnavailableModelItem` function now receives the `chatEntitlementService` as a parameter, allowing it to tailor the hover message based on the user's entitlement.
* When the reason for unavailability is "upgrade," the hover text now checks if the user is already a "Pro" subscriber:
  * If so, it prompts them to "[Upgrade to GitHub Copilot Pro+]" instead of just "Pro."
  * Otherwise, it retains the original "Pro" upgrade message.

**Code integration updates:**

* All calls to `createUnavailableModelItem` in `buildModelPickerItems` and related code paths are updated to pass the new `chatEntitlementService` argument. [[1]](diffhunk://#diff-91b4c599116012c7ddf07e952e5556f7ca5f72eb5da99c9b2530bbc7435b9c43L358-R358) [[2]](diffhunk://#diff-91b4c599116012c7ddf07e952e5556f7ca5f72eb5da99c9b2530bbc7435b9c43L407-R407)

These changes ensure that users see the most relevant upgrade path based on their current subscription, improving clarity and user experience.


justschen (approved):
LGTM

PR #311765, opened 2026-04-21
terminal tools: detect prompts on the last output line

### LINKED ISSUE #311757: run_in_terminal can freeze the window when OutputMonitor evaluates prompt regexes against multi-line terminal tails
- Copilot Chat Extension Version: 0.44.2
- VS Code Version: 1.116.0
- OS Version: macOS 26.4.1
- Feature (e.g. agent/edit/ask mode): Agent mode, `run_in_terminal`
- Selected model (e.g. GPT 4.1, Claude 3.7 Sonnet): Unknown / not captured in the logs
- Logs:

`main.log` repeatedly captured the window becoming unresponsive while terminal output monitoring was polling:

```text
2026-04-22 01:03:54.521 [error] CodeWindow: detected unresponsive
2026-04-22 01:03:56.363 [error] CodeWindow unresponsive samples:
<1>
    at RegExp.test (<anonymous>)
    at ...workbench.desktop.main.js:2541:26359
    at Array.some (<anonymous>)
    at gLi (...)
    at Aie._waitForIdle (...)
    at async Aie._startMonitoring (...)
...
2026-04-22 01:11:07.660 [error] [uncaught exception in main]: UnresponsiveSampleError: from window with ID 4 belonging to process with pid 78887
2026-04-22 01:11:07.661 [error] CodeWindow unresponsive samples:
<14>
    at RegExp.test (<anonymous>)
    at ...workbench.desktop.main.js:2541:26359
    at Array.some (<anonymous>)
    at gLi (...)
    at Aie._waitForIdle (...)
    at async Aie._startMonitoring (...)
```

The same sampled stack repeated again around 01:11, 01:12, 01:16, and 01:30.

The captured chat session shows the problematic `run_in_terminal` command was an `rg -n -uu --no-heading ... | head -n 300` search that produced about 81 KB / 1029 lines of output. The tail of the saved terminal output was multi-line search results, not a prompt.

Steps to Reproduce:

1. Open a workspace and use Copilot agent mode to invoke `run_in_terminal` with a command that produces a large multi-line output buffer.
2. Let the command finish or go idle while the last ~1000 characters of terminal output still contain many lines of ordinary output.
3. Observe that the window can become unresponsive during `OutputMonitor` idle polling.

Expected:

`run_in_terminal` should inspect only the current prompt line for line-oriented prompt/help detection and should not block the window on large multi-line tails.

Actual:

The window becomes unresponsive, and `main.log` samples the hot path as `RegExp.test -> Array.some -> OutputMonitor._waitForIdle -> OutputMonitor._startMonitoring`.

Additional context:

This looks like the same class of issue as #307447, but in `src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/monitoring/outputMonitor.ts`.

Locally I traced it to line-oriented detectors such as `detectsInputRequiredPattern` and `detectsNonInteractiveHelpPattern` being run against `output.slice(-1000)` instead of only the last line. A local fix that scopes those detectors to the last line, while keeping task-finish detection on the tail, addresses the identified hot path and passes a regression test.

mossgowild:
Fixes #311757

OutputMonitor was running line-oriented prompt detection against the last 1000 characters of terminal output. When the tail contained multiple lines, stale lines in the scrollback could be treated as the current prompt and the monitor would repeatedly run prompt regexes over large multiline output.

This change scopes non-interactive help and input-required prompt detection to the last output line while keeping task-finish and generic "press any key" detection on the bounded tail.

It also adds a regression test that covers help text appearing on a non-final line.

Testing:
- node --experimental-strip-types build/hygiene.ts src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/monitoring/outputMonitor.ts src/vs/workbench/contrib/terminalContrib/chatAgentTools/test/browser/outputMonitor.test.ts
- npm run test-browser-no-install -- --run src/vs/workbench/contrib/terminalContrib/chatAgentTools/test/browser/outputMonitor.test.ts --browser chromium


meganrogge (commented):
Thanks for the fix! The core approach is right, but I have a few concerns about `_getLastLineForPatternDetection` that are worth addressing before merge:

1. **Trailing newline produces an empty last line.** Many CLIs flush prompts with a trailing `\n` (or the pty emits `\r\n` after a cursor update), e.g. `Continue? [y/N]\n`. With:
   ```ts
   const lastLineBreak = Math.max(output.lastIndexOf('\n'), output.lastIndexOf('\r'));
   return lastLineBreak === -1 ? output : output.slice(lastLineBreak + 1);
   ```
   `slice(lastLineBreak + 1)` returns `""` and both `detectsInputRequiredPattern` and `detectsNonInteractiveHelpPattern` will miss it. Previously the `$`-anchored regexes would match the prompt against the tail. This looks like a regression — consider stripping trailing `\r`/`\n` first, or falling back to the previous non-empty line:
   ```ts
   const trimmed = output.replace(/[\r\n]+$/, '');
   const idx = Math.max(trimmed.lastIndexOf('\n'), trimmed.lastIndexOf('\r'));
   return idx === -1 ? trimmed : trimmed.slice(idx + 1);
   ```

2. **Bare `\r` progress updates truncate to a fragment.** Tools like `npm`, `pip`, `docker pull` rewrite the same line with `\r`. `lastIndexOf('\r')` will then return a mid-line fragment rather than the real prompt line. (CRLF is fine because `\n` wins via `Math.max`, but bare `\r` redraws on Linux/macOS ttys will be truncated.)

3. **Prompts followed by explanatory text.** Some tools write the prompt and then a hint line after it:
   ```
   Continue? [y/N]
   (press Ctrl-C to abort)
   ```
   The last line is now the hint, so the prompt no longer matches. Probably rare, but a behavioral change vs. the old tail-wide scan.

Could we add regression tests for (a) a prompt ending in `\n` and (b) a `\r`-based progress line, to cover these cases? The stall fix for #311757 is good — just want to make sure we don't silently break input-required detection for the common trailing-newline case.

mossgowild:
Addressed the review feedback in 2c69c473d50.

Changes:
- Updated last-line extraction to trim only trailing CR/LF before detection so newline-terminated prompts like `Continue? [y/N]\n` still match.
- Treated bare `\r` within the final line as a redraw boundary so progress-style output does not get truncated to an empty/partial line.
- Reused the same last-line extraction for logging so the trace output matches what prompt detection actually sees.
- Added regression coverage for both the newline-terminated prompt case and the bare-`\r` progress-line case.

Validation:
- `npm run test-browser-no-install -- --run src/vs/workbench/contrib/terminalContrib/chatAgentTools/test/browser/outputMonitor.test.ts --browser chromium`
- `node --experimental-strip-types build/hygiene.ts src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/monitoring/outputMonitor.ts src/vs/workbench/contrib/terminalContrib/chatAgentTools/test/browser/outputMonitor.test.ts`

I left the prompt-followed-by-hint behavior unchanged for now since that is a broader heuristic tradeoff rather than the concrete regression called out here.

meganrogge (commented):
Thanks for the fix — scoping these detectors to the last line is the right direction since the help-pattern regexes aren't `$`-anchored. A few concerns:

1. **Possible regression when the tail ends with a newline.** If `outputTail` ends in `\n` or `\r` (common when a command just printed its final line), `_getLastLineForPatternDetection` returns `''`. Previously, several `detectsInputRequiredPattern` regexes (e.g. `/:\s*$/`, `/\?\s*...$/`) could still match because `\s` matches `\n`. After this change, a prompt like `"Password:\n"` in the buffer will no longer be detected. Consider trimming trailing newlines first, e.g.:

   ```ts
   const trimmed = output.replace(/[\r\n]+$/, '');
   const lastLineBreak = Math.max(trimmed.lastIndexOf('\n'), trimmed.lastIndexOf('\r'));
   return lastLineBreak === -1 ? trimmed : trimmed.slice(lastLineBreak + 1);
   ```

2. **Behavior change for `detectsInputRequiredPattern`.** Most regexes in that function are already `$`-anchored, so scoping is a no-op for them. The one exception is `/press a(?:ny)? key/i`, which previously could match anywhere in the 1000-char tail and now only matches on the last line. That seems desirable, but worth confirming it doesn't regress any real scenarios.

3. **Brittle private-method access in the new test.** Casting the monitor to `{ [key: string]: ... }` to invoke `_handleIdleState` directly is unusual for this codebase and couples the test to the private shape. Two suggestions:
   - Prefer driving through the public flow (e.g. `startMonitoring` / the poll loop) so the test survives refactors.
   - Or export `_getLastLineForPatternDetection` as a free function and unit-test it directly — it has no `this` dependency.
   Also consider adding a **positive** case (help text on the last line → pattern detected) so both branches are covered.

4. **Minor.** `_getLastLineForPatternDetection` could just be named `_getLastLine` (or a module-local free function), since it's purely string slicing.

Nothing blocking, but (1) feels like it could bite in practice.

meganrogge (approved):
Thank you!

Yoyokrazy (approved):

PR #311797, opened 2026-04-21
Update skill parsing behaviour

### LINKED ISSUE #294747: Skill files without frontmatter work in Claude Code but don't show in VS Code (and CLI)
Testing #6773

`.claude/skills/roast/SKILL.md`
```
As a friendly redditor, roast my article.
```

- 🐛 `/roa…` doesn't work in VS Code (also fails in CLI)
- Works in Claude Code

🐛 Even with frontmatter, not having a `name` breaks it.

pwang347:
Fixes https://github.com/microsoft/vscode/issues/297484
Fixes https://github.com/microsoft/vscode/issues/294747

1. For slash commands, use the folder name in case no name property exists for the skill.
2. Allow no name property.
3. Allow no description property, but if so disable model invocation and provide lint warnings if conflicting model invocation properties are given.


src/vs/workbench/contrib/chat/common/promptSyntax/config/promptFileLocations.ts
```diff
@@ -337,9 +337,9 @@ export function getCleanPromptName(fileUri: URI): string {
 		return basename(fileUri, '.md');
 	}
 
-	// For SKILL.md files (case insensitive), return 'SKILL'
+	// For SKILL.md files (case insensitive), return the parent folder name
 	if (fileName.toLowerCase() === SKILL_FILENAME.toLowerCase()) {
```
aeschli: Lets have a util for that...

src/vs/workbench/contrib/chat/common/promptSyntax/languageProviders/promptValidator.ts
```diff
@@ -66,26 +66,7 @@ export class PromptValidator {
...
-			report(toMarker(
-				localize('promptValidator.skillDescriptionMissing', "Skill must provide a description."),
-				new Range(1, 1, 1, 4),
-				MarkerSeverity.Error
-			));
-			return;
-		}
-
-		if (nameAttribute.value.type === 'scalar') {
+		if (nameAttribute?.value.type === 'scalar') {
```
aeschli: We still want to report the missing name and description, but it should not block the usage of the skill.
See https://agentskills.io/specification

src/vs/workbench/contrib/chat/common/promptSyntax/languageProviders/promptValidator.ts
```diff
@@ -111,6 +92,31 @@ export class PromptValidator {
 				}
 			}
 		}
+
+		const descriptionAttribute = promptAST.header?.attributes.find(attr => attr.key === PromptHeaderAttributes.description);
+		if (!descriptionAttribute) {
+			// Without a description, user-invocable: false is invalid because the skill
+			// would be model-only but has no description for the model to decide when to use it.
+			const userInvocableAttr = promptAST.header?.attributes.find(attr => attr.key === PromptHeaderAttributes.userInvocable);
```
aeschli: use `promptAST.userInvocable`

src/vs/workbench/contrib/chat/common/promptSyntax/languageProviders/promptValidator.ts
```diff
@@ -111,6 +92,31 @@ export class PromptValidator {
...
+				report(toMarker(
+					localize('promptValidator.skillUserInvocableRequiresDescription', "A description is required when user-invocable is false, because the model needs a description to decide when to load the skill."),
+					userInvocableAttr.value.range,
+					MarkerSeverity.Error
+				));
+			}
+
+			// Without a description, disable-model-invocation: false (model invocation enabled)
+			// is the default but if explicitly set, report an error that a description is needed.
+			const disableModelInvocationAttr = promptAST.header?.attributes.find(attr => attr.key === PromptHeaderAttributes.disableModelInvocation);
```
aeschli: use `promptAST.disableModelInvocation`

src/vs/workbench/contrib/chat/common/promptSyntax/promptFileParser.ts
```diff
@@ -16,7 +16,7 @@ export class PromptFileParser {
 	}
 
 	public parse(uri: URI, content: string): ParsedPromptFile {
-		const linesWithEOL = splitLinesIncludeSeparators(content);
+		const linesWithEOL = splitLinesIncludeSeparators(stripUTF8BOM(content));
```
aeschli: The stripUTF8BOM should never appear in a string. If it does, this needs to be fixed where we read the file and decode it.

aeschli (changes requested):

src/vs/workbench/contrib/chat/common/promptSyntax/languageProviders/promptValidator.ts
pwang347: made changes:
1. only show problem if there's a frontmatter section
2. the problem level is warning

src/vs/workbench/contrib/chat/common/promptSyntax/promptFileParser.ts
pwang347: moved to where we read the file

Yoyokrazy (dismissed):

src/vs/workbench/contrib/chat/common/promptSyntax/config/promptFileLocations.ts
```diff
@@ -33,6 +33,13 @@ export const AGENT_FILE_EXTENSION = '.agent.md';
  */
 export const SKILL_FILENAME = 'SKILL.md';
 
+/**
+ * Check if a filename is a skill file (case insensitive).
+ */
+export function isSkillFilename(filename: string): boolean {
+	return filename.toLowerCase() === SKILL_FILENAME.toLowerCase();
```
aeschli: We should have `SKILL_FILENAME.toLowerCase()` is a const as well, no need to compute this every time.
Or, use `string.compareIgnoreCase` which avoids `toLowerCase`

src/vs/workbench/contrib/chat/common/promptSyntax/service/promptsServiceImpl.ts
```diff
@@ -887,7 +864,7 @@ export class PromptsService extends Disposable implements IPromptsService {
 		if (token.isCancellationRequested) {
 			throw new CancellationError();
 		}
-		return new PromptFileParser().parse(uri, fileContent.value.toString());
+		return new PromptFileParser().parse(uri, stripUTF8BOM(fileContent.value.toString()));
```
aeschli: most efficient is to strip it from and not create a string`fileContent.value`.

Question is, why does 'fileContent.value.toString()' return a string with UTF8BOM. This is the method that decodes (turns a buffer in a string) and it should consume the UTF8BOM. @bpasero for insights

aeschli (dismissed):

src/vs/workbench/contrib/chat/common/promptSyntax/service/promptsServiceImpl.ts
bpasero: Yeah I do not think stripping the BOM is needed here:

https://github.com/microsoft/vscode/blob/6aaf3c6b3d58e5d35928df7ec899f7f372f6b59e/src/vs/base/common/buffer.ts#L125-L134

Yoyokrazy (approved):

PR #311806, opened 2026-04-21
Add ability to disable individual customizations

pwang347:
Fixes https://github.com/microsoft/vscode/issues/290290


extensions/copilot/src/extension/chatSessions/vscode-node/claudeCustomizationProvider.ts
```diff
@@ -126,40 +135,87 @@ export class ClaudeCustomizationProvider extends Disposable implements vscode.Ch
 			}
 		}
 
+		// Extension-contributed agents (owned by VS Code) — disableable from Claude harness
+		const vscodeDisabledAgents = this._getDisabledUriSet(allSettings, VSCodeDisabledSettingsKey.Agents);
```
pwang347: this block should be removed, we're not adding support for extension-provided agents :)

extensions/copilot/src/extension/chatSessions/copilotcli/vscode-node/copilotCLICustomizationProvider.ts
```diff
@@ -14,15 +14,42 @@ import { CancellationToken } from '../../../../util/vs/base/common/cancellation'
...
+import { basename, dirname } from '../../../../util/vs/base/common/resources';
 import { URI } from '../../../../util/vs/base/common/uri';
 import { ICopilotCLIAgents, isEnabledForCopilotCLI } from '../../copilotcli/node/copilotCli';
+import { CopilotCLISettingsLocationType, ICopilotCLISettingsService } from '../common/copilotCLISettingsService';
+
+/**
+ * Settings keys for tracking VS Code extension-contributed customizations
+ * that have been disabled in the CLI harness.
+ */
+const enum VSCodeDisabledSettingsKey {
```
pwang347: will most likely introduce a separate storage mechanism for this on the extension-side instead of writing to the settings.json (to avoid polluting it)

src/vscode-dts/vscode.proposed.chatSessionCustomizationProvider.d.ts
```diff
@@ -109,6 +121,33 @@ declare module 'vscode' {
...
+		readonly enablementScope?: ChatSessionCustomizationEnablementScope;
+
+		/**
+		 * Optional URI of the parent plugin of this customization item.
+		 *
+		 * When set, all enable/disable actions for this item target the plugin
+		 * instead of the individual item, and the item's own
+		 * {@link enablementScope} is ignored.
+		 */
+		readonly pluginUri?: Uri;
```
joshspicer: (shared with you IRL too) This feels like the odd man out in the API, especially since a plugin itself is a `ChatSessionCustomizationType`

src/vscode-dts/vscode.proposed.chatSessionCustomizationProvider.d.ts
```diff
@@ -109,6 +121,41 @@ declare module 'vscode' {
 		 * Optional tooltip text shown when hovering over the badge.
 		 */
 		readonly badgeTooltip?: string;
+
+		/**
+		 * Whether this customization is currently enabled.
+		 * Defaults to `true` when omitted.
+		 */
+		readonly enabled?: boolean;
```
mjbvz: I generally don't like default to true as it means conditional checks don't work nicely in js. Could this be something like: `disabled?: { reason: string }`

src/vscode-dts/vscode.proposed.chatSessionCustomizationProvider.d.ts
```diff
@@ -109,6 +121,41 @@ declare module 'vscode' {
...
+		readonly enablementScopeHint?: ChatSessionCustomizationEnablementScope;
+
+		/**
+		 * Optional command executed when the user sets this item's enablement.
+		 *
+		 * The handler should persist the change and fire
+		 * {@link ChatSessionCustomizationProvider.onDidChange} so the UI
+		 * re-queries the updated state.
+		 */
+		readonly enablementCommand?: string;
```
mjbvz, line 159: Should be a `Command`

src/vscode-dts/vscode.proposed.chatSessionCustomizationProvider.d.ts
mjbvz: Part of what is  confusing me is that there are a few different concepts here:

- Is the contribution actually enabled / disabled?
- Can the contribution be enabled / disabled by the user?

I feel like we need clearer names around this

Also making it more complicated is if a setting is user configurable, then `enabled` just reflects the initial state, right? Or is the extension supposed to update this value in response to when the command is fired?

pwang347: Thanks @mjbvz, agree that `disabled?: { reason: string }` would be a cleaner pattern and removes the need for an extra property.

To be clear, the API here is for presentation purposes only and to allow a user to configure settings from this panel. I don't think it is a goal for this to be always 100% in-sync with the actual contents, rather it would update when the view is refreshed (e.g., by changing tabs). Maybe @joshspicer might have some more thoughts on the lifecycle of this.

Another thing we could do to better group the "Can the contribution be enabled / disabled by the user?" is to put it under a grouped object like `disableAction?: { command: Command, scopeHint: ChatSessionCustomizationEnablementScope }` but that doesn't seem to be a common pattern in our API.

PR #311824, opened 2026-04-22
aux window - fix setTimeout leak

### LINKED ISSUE #311823: setTimeout function leaks a disposable when any aux window is open

- VS Code Version: 1.118.0
- OS Version: Darwin arm64 25.4.0

Normally when the window is defocused, the browser will throttle any timeouts set on it. It looks like in VSCode there's a leak in the monkey-patching of setTimeout to disable throttling when an aux-window is open in `enableMultiWindowAwareTimeout`.

Concretely, any call to `setTimeout` with >0 auxwindows will insert a dispose function into the main window's disposables table and never clean it up, even after calling it. I work on Cursor's fork of VSCode, which calls `setTimeout` around ~400 times per second. This was causing my app to have 200MB+ retained in a longer session just by the main window disposables table.

Steps to Reproduce:

See this branch for a minimal repro: https://github.com/andrewkchan/vscode/tree/chan/auxwindow-leak-repro

1. Add a setTimeout in the main window which fires at a very high rate, such as 1000 times per second.
2. Open a developer build, then open an aux window. Leave the app open for a couple minutes. Take a heap snapshot and observe there are a very large number of disposables.

Image

andrewkchan:
Fixes https://github.com/microsoft/vscode/issues/311823

Normally when the window is defocused, the browser will throttle any timeouts set on it. In VSCode there is a concept of aux windows, and timeouts set on the main window must not be throttled when the user is working in the aux window. There is some special code to do this in enableMultiWindowAwareTimeout by monkey-patching setTimeout in the main window + aux windows.

This code contained a memory leak where any call to setTimeout with >0 auxwindows would insert a dispose function into the main window's disposables table and never clean it up, even after calling it. I work on Cursor which is built on VSCode, and our app calls setTimeout something like 400 times per second and this was causing my app to have 230MB+ retained just by the main window disposables table.

image

### DIFF
--- a/src/vs/workbench/browser/window.ts
+++ b/src/vs/workbench/browser/window.ts
@@ -154,6 +154,8 @@ export abstract class BaseWindow extends Disposable {
 					didClear = true;
 					(window as { vscodeOriginalClearTimeout?: typeof window.clearTimeout }).vscodeOriginalClearTimeout?.apply(this, [handle]);
 					timeoutDisposables.delete(timeoutDisposable);
+					// Remove from the window's DisposableStore without re-disposing (we're already inside dispose)
+					disposables.deleteAndLeak(timeoutDisposable);
 				});
 
 				disposables.add(timeoutDisposable);


Tyriar (commented):
lgtm, `timeoutDisposable` is added to `disposables` and then never gets removed for the lifetime of the window.

@bpasero could you check this out?

bpasero (approved):
Thanks

pwang347 (approved):

PR #311851, opened 2026-04-22
Fix enableThinking mismatch causing inline summarization cache misses

bhavyaus:
The inline summarization path computed `enableThinking` differently from the main agent loop (`toolCallingLoop.ts`). For Anthropic on non-continuation turns, the old logic (`!isAnthropicFamily || messagesContainThinking`) would set `enableThinking=false` when no thinking blocks existed yet, while the main agent always kept it `true`. Per Anthropic docs, changing thinking parameters invalidates the message cache — causing 0% cache hits on summarization requests.

Fix: mirror the main loop's `shouldDisableThinking = isContinuation && isAnthropicFamily && !messagesContainThinking` logic exactly.

cc: @kevin-m-kent who tracked down this issue! 🙏


alexdima (approved):

isidorn:
@bhavyaus thank you very much for fixing this one 🙏  Super impactful!

PR #311892, opened 2026-04-22
Move skip shell handling to service and optimize

### LINKED ISSUE #311878: Avoid computing and storing commands to skip shell per terminal instance
I noticed this inefficiency:

https://github.com/microsoft/vscode/blob/200fb2ed1db2556b3da5a524d4d418835bd9de39/src/vs/workbench/contrib/terminal/browser/terminalInstance.ts#L1962-L1967

Currently this will compute and store `_skipTerminalCommands` per terminal instance when really this should be shared across instances. It also does a filter/map/concat which isn't particularly optimized.

I recommend the following:

- Move `_skipTerminalCommands` into `ITerminalConfigurationService` to be shared across instances, expose as something like `shouldCommandSkipShell`
- Use a `Set` over `string[]` for faster lookup as this is used repeatedly, the `.some` used in `attachCustomKeyEventHandler` is O(n) atm
- Avoid functional array functions to use imperative to have more control over creating and GCing a bunch of arrays as this will be run on startup

@anthonykim1

Tyriar:
Fixes #311878

@anthonykim1 @meganrogge

### DIFF
--- a/src/vs/workbench/contrib/terminal/browser/terminal.ts
+++ b/src/vs/workbench/contrib/terminal/browser/terminal.ts
@@ -632,6 +632,13 @@ export interface ITerminalConfigurationService {
 	setPanelContainer(panelContainer: HTMLElement): void;
 	configFontIsMonospace(): boolean;
 	getFont(w: Window, xtermCore?: IXtermCore, excludeDimensions?: boolean): ITerminalFont;
+
+	/**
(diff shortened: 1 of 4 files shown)


Tyriar:
@microsoft-github-policy-service agree company="Cursor"

meganrogge (approved):
Thanks!

alexr00 (approved):

PR #311960, opened 2026-04-22
Agents: fix SSH reconnect failing with "Unknown method: initialize" after window reload

roblourens:
## What

Fix `Failed to connect via SSH to <host>: Error: Unknown method: initialize` when reconnecting to a previously-connected SSH remote agent host after a window reload (without restarting the Agents app).

## Repro

1. Connect to an SSH remote agent host.
2. Disconnect.
3. Reload the window (do **not** restart the app).
4. Try to connect to the same host again → `Unknown method: initialize`.

## Why

The SSH tunnel and WebSocket relay live in the **main process** and survive window reload. The protocol server keeps per-transport state and only accepts `initialize` once per transport. After a reload, the renderer creates a fresh `RemoteAgentHostProtocolClient` and sends a fresh `initialize` over the **same** WebSocket the server had already initialized for the previous renderer — so it falls through to `_handleRequest`, which doesn't know `initialize`, and replies with `Unknown method: initialize`.

The renderer's `connect()` was calling the main service without `replaceRelay`, so the existing relay was reused.

## Fix

The renderer's `connect()` now always passes `replaceRelay: true` to the main service whenever it has no live local handle (window reload, or first-time connect). The main process tears down the existing WebSocket relay and creates a fresh one over the same SSH client, so the new protocol client gets a fresh server-side transport for its handshake. On a true first connect, `replaceRelay` is a no-op (no existing relay).

This mirrors what the auto-reconnect-on-startup path (`reconnect()`) already did — it always passed `replaceRelay: true`.

## Also in this PR

- Extract a `getSSHConnectionKey(config)` helper so the renderer, main service, and progress-event filter all derive the same connection key.
- Dispose any remaining SSH connection handles when the renderer service itself is disposed (small leak fix).

## Tests

- New `sshRemoteAgentHostServiceImpl.test.ts` (renderer) — 4 tests covering:
  - Fresh-renderer connect always passes `replaceRelay: true` (the regression).
  - Repeated connects short-circuit on the local handle without IPC.
  - Key parity — renderer's `expectedKey` matches main's `connectionId`.
  - Post-close, next connect calls main again with `replaceRelay: true`.
- New regression test in `sshRemoteAgentHostService.test.ts` (main) — verifies a fresh relay is created on existing tunnels without restarting the agent.
- Verified each new renderer regression test actually catches the bug by reverting the fix (3 of 4 fail without `replaceRelay: true`; 2 of 4 fail without the local-handle short-circuit).
- All 206 SSH/RemoteAgentHost tests pass.

(Written by Copilot)

### DIFF
--- a/src/vs/platform/agentHost/common/sshRemoteAgentHost.ts
+++ b/src/vs/platform/agentHost/common/sshRemoteAgentHost.ts
@@ -115,6 +115,20 @@ export interface ISSHRemoteAgentHostService {
 	reconnect(sshConfigHost: string, name: string): Promise<ISSHAgentHostConnection>;
 }
 /**
+ * Stable identifier for an SSH connection derived from its config.
+ *
+ * The key intentionally collapses to the SSH config host alias when one is
+ * provided; otherwise it is `<user>@<host>:<port>`. The renderer uses this
+ * to detect whether it already has a live handle for a connection that the
+ * main process may have kept alive across a window reload.
+ */
+export function getSSHConnectionKey(config: ISSHAgentHostConfig): string {
+	return config.sshConfigHost
+		? `ssh:${config.sshConfigHost}`
+		: `${config.username}@${config.host}:${config.port ?? 22}`;
+}
+
+/**
  * Serializable result from a successful SSH connect operation.
  * Returned over IPC from the main process.
  */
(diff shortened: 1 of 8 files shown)


roblourens:
Addressed review feedback in 823749298f5:

1. **`_connections` → `DisposableMap`**: replaced the manual cleanup loop with `DisposableMap`. Bonus: `set()` now auto-disposes a previous handle for the same key.

2. **"What's the scenario where we already have a connection?"**: the only realistic scenario is a repeated user-initiated connect to a host that's already live (the UI doesn't actively prevent this — see `connectWithProgress` in `remoteAgentHostActions.ts`, which is fire-and-forget). Updated the inline comment to explain that the short-circuit prevents `replaceRelay: true` from tearing down the live WebSocket relay underneath our existing protocol client (which would orphan it).

(Written by Copilot)

PR #312013, opened 2026-04-22
SSH agent host: add agent forwarding setting & fix encrypted key failures

joshspicer:
Two changes to the SSH remote agent host:

### 1. Skip fallback privateKey when SSH agent socket is present

`ssh2` eagerly parses `privateKey` at connect time — before attempting any auth. If the key is passphrase-encrypted this immediately throws `"Cannot parse privateKey: Encrypted private OpenSSH key detected, but no passphrase given"`, even when the key is already loaded in the SSH agent.

### 2. Add `chat.agentHost.forwardSSHAgent` setting

Adds an opt-in setting that enables OpenSSH agent forwarding on SSH connections, allowing tools running under the remote agent host to authenticate using the local user's SSH keys.

### DIFF
--- a/src/vs/platform/agentHost/electron-browser/sshRemoteAgentHostServiceImpl.ts
+++ b/src/vs/platform/agentHost/electron-browser/sshRemoteAgentHostServiceImpl.ts
@@ -68,6 +68,7 @@ export class SSHRemoteAgentHostService extends Disposable implements ISSHRemoteA
 				this._onDidChangeConnections.fire();
 			}
 		}));
+
 	}
 
 	get connections(): readonly ISSHAgentHostConnection[] {
@@ -75,8 +76,8 @@ export class SSHRemoteAgentHostService extends Disposable implements ISSHRemoteA
 	}
 
 	async connect(config: ISSHAgentHostConfig): Promise<ISSHAgentHostConnection> {
-		this._logService.info('[SSHRemoteAgentHost] Connecting to ' + config.host);
 		const augmentedConfig = this._augmentConfig(config);
+		this._logService.info(`[SSHRemoteAgentHost] Connecting to ${config.host}`);
 		const result = await this._mainService.connect(augmentedConfig);
 		this._logService.trace('[SSHRemoteAgentHost] SSH tunnel established, connectionId=' + result.connectionId);
 		return this._setupConnection(result);
@@ -96,7 +97,9 @@ export class SSHRemoteAgentHostService extends Disposable implements ISSHRemoteA
 
 	async reconnect(sshConfigHost: string, name: string): Promise<ISSHAgentHostConnection> {
 		const commandOverride = this._getRemoteAgentHostCommand();
-		const result = await this._mainService.reconnect(sshConfigHost, name, commandOverride);
+		const agentForward = this._isSSHAgentForwardingEnabled();
+		this._logService.info(`[SSHRemoteAgentHost] Reconnecting to ${sshConfigHost}`);
+		const result = await this._mainService.reconnect(sshConfigHost, name, commandOverride, agentForward);
 		return this._setupConnection(result);
 	}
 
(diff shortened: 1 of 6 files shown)


src/vs/sessions/contrib/remoteAgentHost/browser/remoteAgentHostActions.ts
```diff
@@ -151,12 +151,17 @@ async function promptToConnectViaSSH(
...
+			// transparently. We only record the identity file path so the
+			// KeyFile picker can pre-fill it if the user manually selects
+			// that option.
 			if (resolvedConfig.identityFile.length > 0) {
 				const firstKey = resolvedConfig.identityFile[0];
 				const defaultKeys = ['~/.ssh/id_rsa', '~/.ssh/id_ecdsa', '~/.ssh/id_ed25519', '~/.ssh/id_dsa', '~/.ssh/id_xmss'];
 				if (!defaultKeys.includes(firstKey)) {
-					defaultAuthMethod = SSHAuthMethod.KeyFile;
 					defaultKeyPath = firstKey;
 				}
```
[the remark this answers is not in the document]
joshspicer: mitigated

src/vs/platform/agentHost/node/sshRemoteAgentHostService.ts
```diff
@@ -487,8 +487,16 @@ export class SSHRemoteAgentHostMainService extends Disposable implements ISSHRem
...
 			const existingAH = await findRunningAgentHost(exec, this._logService, this._quality);
 			if (existingAH) {
-				remotePort = existingAH.port;
-				connectionToken = existingAH.connectionToken;
+				if (config.agentForward) {
+					// Agent forwarding requires a fresh process so it inherits
+					// SSH_AUTH_SOCK from this (forwarded) SSH channel. Kill the
+					// stale one so we start a new one below.
+					this._logService.info(`${LOG_PREFIX} Agent forwarding enabled — restarting remote agent host so it inherits the forwarded SSH_AUTH_SOCK`);
+					await cleanupRemoteAgentHost(exec, this._logService, this._quality);
```
roblourens: Wait a sec... there might be other clients connected to the agent host. And this will stop your agents, which might be unexpected. This is a bit weird in a multi-client scenario where more than one might be ssh'd into the same server. I wonder whether we should do this in a nicer way where we would set that environment variable dynamically inside the process instead of restarting the process. Maybe it should be connected to the "active client" concept in AHP

src/vs/platform/agentHost/electron-browser/sshRemoteAgentHostServiceImpl.ts
```diff
@@ -192,16 +195,25 @@ export class SSHRemoteAgentHostService extends Disposable implements ISSHRemoteA
 	}
 
 	private _augmentConfig(config: ISSHAgentHostConfig): ISSHAgentHostConfig {
+		const result = { ...config };
 		const commandOverride = this._getRemoteAgentHostCommand();
 		if (commandOverride) {
-			return { ...config, remoteAgentHostCommand: commandOverride };
+			result.remoteAgentHostCommand = commandOverride;
+		}
+		const agentForward = this._isSSHAgentForwardingEnabled();
```
roblourens: Is this only driven by the setting, or is it looking at the property on the ssh config too?

roblourens (changes requested):

src/vs/platform/agentHost/electron-browser/sshRemoteAgentHostServiceImpl.ts
joshspicer: It was just the setting, but there isn't a good reason to not also check the config.  Adding

src/vs/platform/agentHost/node/sshRemoteAgentHostService.ts
joshspicer: Good point... I don't think either change is trivial so i'll remove the killing code for this PR

Agreed - it seems like it should be a property of the 'active client' to support >1 SSH client at a time

roblourens (approved):

PR #312038, opened 2026-04-22
Adopt copilot/sdk session-title APIs and sync title across session list, editor, CLI 

### LINKED ISSUE #311535: Adopt Copilot CLI sdk API for custom session titles

anthonykim1:
Resolves: https://github.com/microsoft/vscode/issues/311535 

* Renaming a Copilot CLI session title from VS Code now sticks everywhere it
should — sessions list, chat editor tab/header, and `copilot --resume` in
the terminal.

* Adopts the copilot/sdk session-title APIs so the SDK is the source of truth
instead of a local label cache.

Side fix after some surprise 😮:

Noticed the chat panel/header was not syncing on rename even when
the sessions list was. The two surfaces share no state:
- Sessions list is driven by `ChatSessionItem.label` from the contributed
  item provider.
- Chat editor tab / panel header is driven by the open chat model's title,
  which only updates when something fires `setCustomTitle` on the model.

This PR would bridge the gap there as well.  ->  In `mainThreadChatSessions` so a label change on a
live session propagates into the chat model.

AFTER PR:
Screenshot 2026-04-22 at 5 00 56 PM

**How titles flow:**
- Manual rename in VS Code writes to the SDK's `name` field.
- AI-generated titles (from the chat title provider) write to the SDK's `summary` field.
- `copilot --resume` renames write to `name` too, and VS Code picks them up the next time it reads metadata.
- The sidebar and the chat editor header now go through the same resolver, so they always show the same thing.

**How a title is picked, in order:**
1. An explicit title if there is one (live session title, SDK `name`, or legacy custom title).
2. A derived label we cached on a previous pass.
3. The pending prompt, for sessions that just started.
4. A clean `summary` from metadata (skipped if it looks truncated — ends with `...` or contains `<`).
5. The first user message from session history (cached if we find one).
6. The raw `summary` as a last resort, just for display — not cached.

**A couple of things worth calling out:**
- Both `renameSession` and `updateSessionSummary` clear the cached derived label, so a brand-new title is never hidden behind an old cached one.
- Before this PR, the sidebar and the editor had slightly different orders of precedence and only the sidebar used the cache, so they could disagree. They share one path now.

/cc @DonJayamanne


roblourens (dismissed):

rebornix (dismissed):

TylerLeonhardt (approved):

src/vs/workbench/api/browser/mainThreadChatSessions.ts
```diff
@@ -490,6 +490,12 @@ class MainThreadChatSessionItemController extends Disposable implements IChatSes
 			return existing;
 		}
 
+		// Propagate a renamed item label to the open chat model so the chat editor tab
```
anthonykim1, line 493: https://github.com/user-attachments/assets/bc221b1b-ef55-43a6-9dc7-ad4e5c84bec0

Maybe what's causing the temporary title change? @roblourens @DonJayamanne 

This is meant to ensure title is synced across session list, chat editor header though.
Edit: Addressed issue where pending title would flicker in agents app in https://github.com/microsoft/vscode/pull/312657

DonJayamanne: Thanks @anthonykim1 ,

PR #312063, opened 2026-04-23
Gray out remembered folders for offline agent hosts

roblourens:
In the session workspace picker, the agent host row for a disconnected remote host is shown grayed out, but the remembered (previously used) folders for that host were still rendered as enabled.

This threads the existing `isOffline` flag through to each remembered-folder item and sets `disabled: isOffline`, so they visually match the offline host row. Clicking these items was already a no-op via `_isProviderUnavailable`; this just brings the visual state in line.

(Written by Copilot)


roblourens (commented):
Agreed on both counts — renamed `isOffline` → `isUnavailable` and the localized label from `(Offline)` to `(Unavailable)` in e023bed. (Written by Copilot)

anthonykim1 (approved):

PR #312136, opened 2026-04-23
Agents: fix sessions workspace picker

sandy081:
Fixes several issues with the sessions workspace picker in the Agents window.

### Filter copilot worktrees from VS Code recents
Folders whose name starts with `copilot-` are filtered out from the VS Code recent folders list, as these are internal worktrees and not user-facing workspaces.

### Preserve VS Code recents order
Removed the alphabetical sort that was previously applied to VS Code recent folders. Entries now retain the order VS Code stores them in (most recently used first).

### Stable picker width
When `maxWidth` is set on the action list options, DOM-based width measurement is now skipped entirely and the fixed value is used directly. Previously, the measured width varied with scrollbar presence (which depended on list height which depended on window height), causing the picker width to fluctuate when the window was resized.

### Scrollbar clearance
Row width is set to `calc(100% - 10px)` so content never extends under the 10px Monaco overlay scrollbar track.

### Group title / toolbar spacing
In inline-description mode, the flex row gap is set to `0` with explicit margins on individual elements. This keeps the group title ("Folders") and remove button (×) flush with just a small readable gap, while still giving the group title `margin-right` breathing room when the toolbar is hidden.

### Bottom gap
A 30px gap is maintained between the bottom of the picker and the window edge.

### Document entry ordering
Added JSDoc on `_buildItems()` explaining the sorting: own recents first, VS Code recents after, groups ordered by first appearance, within-group order preserved.

### DIFF
--- a/src/vs/platform/actionWidget/browser/actionList.ts
+++ b/src/vs/platform/actionWidget/browser/actionList.ts
@@ -459,7 +459,7 @@ export interface IActionListOptions {
 	 * by scrollbar presence (which changes with window height). Use this for pickers
 	 * that should have a stable, fixed width (e.g. the workspace picker at 600px).
 	 */
-	readonly maxWidth?: number;
+	readonly fixedWidth?: number;
 
 	/**
 	 * Optional handler for markdown links activated in item descriptions or hovers.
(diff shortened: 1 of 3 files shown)


lszomoru (dismissed):

lszomoru (dismissed):

hediet:
Can you confirm if the fixture change is intentional?

I believe it is caused by that:
```
/* Leave clearance on the right for the vertical scrollbar track */
width: calc(100% - 10px);
```

(disabled auto-merge as precautionary measure, feel free to merge nontheless)

sandy081:
Hey - this is a great catch - I will fix it

lszomoru (approved):

PR #312142, opened 2026-04-23
Allow enabling/disabling all skill kinds from AI Customizations UI

ulugbekna:
Previously, only built-in skills could be enabled/disabled from the AI Customizations UI. This PR extends that to all skill kinds: workspace, user, and extension-contributed.

## Changes

**Core: central disable filter** (`promptsServiceImpl.ts`)
- `findAgentSkills()` now filters out disabled skills before returning, making it the single source of truth for all consumers.

**UI: removed storage-kind restriction** (`aiCustomizationManagement.contribution.ts`, `aiCustomizationTreeView.contribution.ts`)
- Enable/Disable menu items no longer require `BUILTIN_STORAGE` — they appear for any skill.

**UI: two-phase skill loading** (`promptsServiceCustomizationItemProvider.ts`)
- Phase 1: enabled skills from `findAgentSkills()` (already filtered).
- Phase 2: disabled skills recovered from `listPromptFiles()` + `getDisabledPromptFiles()`.

**Sessions tree view** (`aiCustomizationTreeViewViews.ts`)
- `getStorageGroups` now surfaces storage buckets that contain only disabled skills, so users can re-enable them.

**Cleanup: removed redundant consumer filters** (`mainThreadChatAgents2.ts`, `computeAutomaticInstructions.ts`)
- Since `findAgentSkills()` is now the single source of truth, consumer-side filters were dead code.

**Fix: skill counts include disabled skills** (`customizationCounts.ts`, `aiCustomizationOverviewView.ts`)
- Badge/section counts now use the same two-phase logic so disabled skills are counted.

**Fix: debug tracing for disabled skills** (`computeAutomaticInstructions.ts`)
- Disabled-skill skips now produce `debugDetails` entries (used by the troubleshoot skill).


ulugbekna:
Addressed both review comments:

1. **`isEmptyContextKey` bug** — `totalItemCount` now includes disabled skills. The disabled-skill count is computed inside the `if (!cached.skills)` block (consistent with the caching pattern for other types) and added to `totalItemCount` alongside enabled skills. This ensures the view won't incorrectly show the welcome/empty state when all skills are disabled.

2. **Missing unit test** — Added a test `"should exclude disabled skills from results"` in the `findAgentSkills` suite of `promptsService.test.ts`. It verifies:
   - Both skills are found before disabling
   - After `setDisabledPromptFiles`, the disabled skill is excluded from `findAgentSkills` results
   - After re-enabling (empty disabled set), both skills reappear

PR #312151, opened 2026-04-23
copilot/byok: add reasoning effort support for custom OAI models

mxi-box:
This adds reasoning effort support for custom OAI BYOK models.


mxi-box:

PR #312178, opened 2026-04-23
Async terminal for CLI in chat panel

### LINKED ISSUE #309290: Async terminals don't resume the session
In VS Code, we now send a steering message when an async terminal is complete. Copilot CLI can do this but we don't support it yet.

Image

Image

anthonykim1:
Resolves: https://github.com/microsoft/vscode/issues/309290 

*WIP almost there
* Please see comments below (videos, screenshot, comments)
* Feel free to fetch this branch locally and play around. 

Both route with/without controller api should work.


anthonykim1:
@DonJayamanne @rebornix @mattbierner 

Don't think `activeResponseCallback`  in https://github.com/microsoft/vscode/blob/ad4cb84f92c12632fd21382e1deea759d9881f5b/src/vscode-dts/vscode.proposed.chatSessionsProvider.d.ts#L446-L453 would do it since it would show the progress shimmer that the response is still in progress?

I think the biggest thing against `activateResponseCallback` is that ui would be stuck in perpetual shimmer/responding state with all notification glued onto the previous assistant turn as if the model never stopped talking, and user wouldn't be able to send follow ups without killing the very thing that delivers the notification. 

VS. 

(This PR with new proposed api):

The approach from the PR we'd have no stalling shimmer, but would have new compact "system" bubble appear for async output such as `sleep 5 finished`
* Also user can type follow up anytime (input not blocked).
* Multiple async shell should work via multiple discrete bubble (clearly its own event).

anthonykim1:
@roblourens @TylerLeonhardt 
Here is what it would look like as of Apr 23rd, 2:23 p.m. PST: 

https://github.com/user-attachments/assets/b66cd931-f681-4d7e-b527-a26768c1121b

Here is equivalent in Copilot CLI. 
cliAsyncExample

We could improve to better distinguish the async terminal output but it seems like CLI themselves are not doing explicit labeling either. Maybe the color of the decorations are meant to signify something? 

/cc @justschen For chat ui

extensions/copilot/src/extension/chatSessions/copilotcli/node/copilotcliSessionService.ts
```diff
@@ -1090,6 +1104,49 @@ export class CopilotCLISessionService extends Disposable implements ICopilotCLIS
...
+		// `assistant.turn_end`. Without this, the per-request `DisposableStore`
+		// releases its single ref and the SDK session is closed before any
+		// `system.notification` for an async/detached shell completion can
+		// fire, so no fresh chat bubble appears (issue #309290).
+		//
+		// Note: we can't rely on `getBackgroundTasks()` here because the public
+		// task list only includes *detached* shells and background agents, not
+		// plain `mode: "async"` shells (they're tracked internally by
+		// `shellContext.currentExecutions`). A status-based window matches the
+		// existing `sessionTerminators` 5-minute lifetime and covers every
```
anthonykim1: @DonJayamanne @roblourens I'm not so sure about this 5-minute lifetime thing. I can see it was existing previously but don't have context over what specific thing it was meant for, maybe we can do something better here for async terminals and keeping session alive.

extensions/copilot/src/extension/chatSessions/copilotcli/node/copilotcliSession.ts
```diff
@@ -215,6 +223,54 @@ export class CopilotCLISession extends DisposableStore implements ICopilotCLISes
 		super();
 		this.sessionId = _sdkSession.sessionId;
 		this.add(toDisposable(() => this._todoSqlQuery.dispose()));
+
+		// Forward SDK system notifications (async/detached shell completions,
```
anthonykim1, line 263: I'll probably delete this before final review but thought it would be helpful for earlier people who read, understand more carefully.

extensions/copilot/src/extension/chatSessions/copilotcli/node/copilotcliSession.ts
```diff
@@ -907,6 +984,35 @@ export class CopilotCLISession extends DisposableStore implements ICopilotCLISes
...
+	 */
+	private async _awaitNextSessionIdle(token: vscode.CancellationToken): Promise<void> {
+		if (token.isCancellationRequested) {
+			return;
+		}
+		const SAFETY_TIMEOUT_MS = 5 * 60 * 1000;
+		const disposables = new DisposableStore();
+		try {
+			await new Promise<void>(resolve => {
+				const off = this._sdkSession.on('assistant.turn_end', () => resolve());
```
DonJayamanne: what happens if the next turn also starts an async terminal call?

extensions/copilot/src/extension/chatSessions/copilotcli/node/copilotcliSession.ts
```diff
@@ -215,6 +223,54 @@ export class CopilotCLISession extends DisposableStore implements ICopilotCLISes
...
+		// chat request via `vscode.chat.sendSystemInitiatedRequest` so the
+		// notification surfaces as a UI bubble and the SDK's auto-injected
+		// follow-up turn streams into a fresh chat response (issue #309290).
+		// notification surfaces as a UI bubble and the SDK's auto-injected
+		// follow-up turn streams into a fresh chat response (issue #309290).
+		this.add(toDisposable(this._sdkSession.on('system.notification', (event: SystemNotificationEvent) => {
+			try {
+				const notification = this._buildSystemNotification(event);
+				if (notification) {
+					this._onDidReceiveSystemNotification.fire(notification);
```
DonJayamanne, line 300: What happens if the async terminal completes within the current turn?
I think with this approach we will always end up going out of there and sending a message to core and then re-enter as another request handler when it could've been handled in a single turn.

extensions/copilot/src/extension/chatSessions/copilotcli/node/copilotcliSessionService.ts
```diff
@@ -1090,6 +1104,49 @@ export class CopilotCLISessionService extends Disposable implements ICopilotCLIS
...
+				hasKeepAliveRef = false;
+				refCountedSession.release();
+			}
+		};
+		const keepAliveTimer = new MutableDisposable<IDisposable>();
+		session.add(keepAliveTimer);
+		session.add(toDisposable(releaseKeepAlive));
+		session.add(session.onDidChangeStatus(() => {
+			const status = session.status;
+			if (session.permissionRequested) {
```
DonJayamanne: Please do not use `permissionRequested`, this will be removed.

extensions/copilot/src/extension/chatSessions/copilotcli/node/copilotcliSessionService.ts
```diff
@@ -1090,6 +1104,49 @@ export class CopilotCLISessionService extends Disposable implements ICopilotCLIS
...
+				keepAliveTimer.clear();
+				return;
+			}
+			if (status === undefined || status === ChatSessionStatus.Completed || status === ChatSessionStatus.Failed) {
```
DonJayamanne, line 1323: Doesn't this mean that ALL sessions will always keep a ref to this object and never release it?

src/vscode-dts/vscode.proposed.chatSessionsProvider.d.ts
```diff
@@ -579,6 +579,34 @@ declare module 'vscode' {
...
+	/**
+	 * Options for {@link chat.sendSystemInitiatedRequest}.
+	 */
+	export interface SystemInitiatedChatRequestOptions {
+		/**
+		 * Short human-readable label for the system event that triggered this request.
+		 * Rendered as a compact system-notification chip in the chat transcript
+		 * (e.g. ``"`sleep 10` completed"``).
+		 */
+		readonly systemInitiatedLabel: string;
```
DonJayamanne, line 609: perhaps `prompt` is a better property
If its a request thats being sent, then i would like to think the message thats sent as part of the request would be a prompt or command, & this case it should be a prompt.

extensions/copilot/src/extension/chatSessions/copilotcli/node/copilotcliSessionService.ts
DonJayamanne: I also checked with Opus 4.7 and I think there's something to fix here,

extensions/copilot/src/extension/chatSessions/copilotcli/node/copilotcliSessionService.ts
DonJayamanne: I've removed this `permissionRequested` in main

extensions/copilot/src/extension/chatSessions/copilotcli/node/copilotcliSessionService.ts
DonJayamanne: Can we try to use `getBackgroundTaskProgress` and `refreshBackgroundTasks` in conjuction with `getBackgroundTasks` (methods in the SDK)

If the current API isn't going to work, then I think we'll need new API from CLI. 
Please try the above, perhaps try Opus 4.7, we might be able to make use of these (specially `refreshBackgroundTasks`).

I don't think we shoudl have any timeout be it 5minutes or 1 minute or 15 iminutes, at the end of the day, they'll all be wrong in one case or another.

extensions/copilot/src/extension/chatSessions/copilotcli/node/copilotcliSession.ts
anthonykim1: Good catch - 
Just made  https://github.com/microsoft/vscode/pull/312178/commits/913dcae6abc4098fde6121e7f3b43a31e6c6df70 to guard on the listener so we never forward while a turn is in flight. Idle path (the actual feature) is unchanged.

extensions/copilot/src/extension/chatSessions/copilotcli/node/copilotcliSessionService.ts
anthonykim1: Sounds good, removed in : [f38308d](https://github.com/microsoft/vscode/pull/312178/commits/f38308d191e5180ee93ea3ac1299695ff3dd5d45)

extensions/copilot/src/extension/chatSessions/copilotcli/node/copilotcliSession.ts
anthonykim1: Not getting the desired output - rather first round finishes and gets back successfully but I dont hear back from the second async call. Maybe permission related issue? The logs show permission is requested for the second run but nothing comes back afterwards. Super weird Screenshot 2026-04-24 at 9 46 28 AM

anthonykim1: Figured out what's happening: [283a8ee](https://github.com/microsoft/vscode/pull/312178/commits/283a8eee385e41a0a145f4572d07167b793626b2) for description but we'd need something more proper to make sure the confirmation gets passed appropriately

Edit: Now should work 
Screenshot 2026-04-25 at 11 16 25 PM

extensions/copilot/src/extension/chatSessions/copilotcli/node/copilotcliSessionService.ts
anthonykim1: Thanks good point - would [18773f5](https://github.com/microsoft/vscode/pull/312178/commits/18773f570ae0bf4aa20205dcb30e378cc6487dec) work?

src/vscode-dts/vscode.proposed.chatSessionsProvider.d.ts
anthonykim1: I agree, now done https://github.com/microsoft/vscode/pull/312178/commits/b3ccdb44479c5c0b0d53fa2ec08d9a169924a6ff

extensions/copilot/src/extension/chatSessions/copilotcli/node/copilotcliSession.ts
anthonykim1: Now with https://github.com/microsoft/vscode/pull/312178/commits/a8c0f1b173a4f2d575dfda516aea663d51bd65a1 so if async job gets turn while in progress for other task we'd inline the async output after finishing current one.

extensions/copilot/src/extension/chatSessions/copilotcli/node/copilotcliSessionService.ts
anthonykim1: I think we may need a new API to get this working since it seems like both `getBackgroundTaskProgress` and `refreshBackgroundTasks` excludes non-detached shells, meaning for async runs like sleep some time and get back, it wouldn't be tracked via the two `getBackgroundTaskProgress` and `refreshBackgroundTasks`.

DonJayamanne: Replied offline with possible solutions.

PR #312225, opened 2026-04-23
agentHost: configure plugins for remote agent hosts

joshspicer:
## Summary
- persist host-owned plugin customizations in AHP root config so a remote agent host can own plugin configuration
- plumb root config changes through agent host dispatch/side-effect paths and republish live session customization state
- distinguish host-configured versus client-synced customizations, including host-wide vs workspace-scoped matching against the original selected workspace
- expose add/remove/manage flows for remote host plugins in the Sessions AI customization UI
- sync the paired protocol updates and add focused agentHost/sessions tests

## Why

When a user connects from a browser or phone, they may not have any local customization state to sync into the remote host. This change lets the remote AHP host configure plugins directly, then feeds those resolved plugins into the Copilot-backed runtime so sessions can use them without relying on local client state.

## How it works

There are two categories of customizations surfaced when connected to a remote host:

### Client customizations ("Client" group in UI)
These are your local VS Code customizations (skills, instructions, agents, prompts) synced to the remote host at connection time.

1. `SyncedCustomizationBundler` collects local files and packages them into a single synthetic Open Plugin (`vscode-synced-customization://` in-memory bundle)
2. `AgentHostContribution` dispatches `RootConfigChanged` to the remote server with the bundle included in the customization list
3. The server persists the list, calls `agent.setHostCustomizations()`, and the SDK loads the bundle's files into the session
4. The harness provider reads `SessionCustomizationsChanged` actions and surfaces each file as an individual item under a **Client** group

### Remote customizations ("Remote" group in UI)
These are Open Plugin Format folders explicitly configured on the remote host's filesystem.

1. `RemoteAgentPluginController` lets you browse the remote filesystem and pick a plugin folder
2. Picking a folder dispatches `RootConfigChanged` with the folder URI added to `agentHost.customizations` in the host's root config
3. The server persists this to disk (`agentHostConfig.json`), calls `agent.setHostCustomizations()`, and the SDK loads the plugin
4. `SessionCustomizationsChanged` fires with status updates (loading → loaded/error), surfaced under a **Remote** group with remove affordances

### The flow that makes something appear in the UI

```
Client dispatches RootConfigChanged
  → server persists + calls setHostCustomizations()
    → SDK loads plugin files
      → agent emits SessionCustomizationsChanged action
        → client receives it via subscription
          → RemoteAgentCustomizationItemProvider.provideChatSessionCustomizations()
            → items appear in the plugin list widget
```

The server is the source of truth for what's actually loaded and its status. The client sends "here's what I want", and the server echoes back "here's what actually loaded" via `SessionCustomizationsChanged`.

## Protocol / Spec
Paired draft PR: microsoft/agent-host-protocol#77

## Notes for reviewers
- host-owned plugin refs are stored in `agent-host-config.json` via AHP root config
- live sessions now republish session customizations when host customization resolution changes
- workspace-scoped host plugins match against the session's original customization directory rather than an isolated worktree path
- this is still configuration of existing remote plugin folders, not a full remote install/download flow
- `RootConfigChangedAction` is the only root action clients may dispatch (narrowed from the full `RootAction` union)

## Validation
- `node ./build/hygiene.ts $(git diff --name-only)`
- `npm run transpile-client`
- `npm run test-node -- --run src/vs/platform/agentHost/test/node/agentService.test.js src/vs/platform/agentHost/test/node/agentSideEffects.test.js src/vs/sessions/contrib/remoteAgentHost/test/browser/remoteAgentHostCustomizationHarness.test.js`

`./scripts/test.sh` still does not boot Electron correctly in this shell (`app` is undefined during startup), so the focused node runner remains the reliable validation path here.


src/vs/sessions/contrib/remoteAgentHost/browser/remoteAgentHostCustomizationHarness.ts
```diff
@@ -35,14 +97,121 @@ function toStatusString(status: CustomizationStatus | undefined): 'loading' | 'l
...
+		this.dispatchCustomizations(updated);
+	}
+
+	private getConfiguredCustomizations(): readonly CustomizationRef[] {
+		const rootState = this._connection.rootState.value;
+		if (!rootState || rootState instanceof Error) {
+			return [];
+		}
+
+		return getAgentHostConfiguredCustomizations(rootState.config?.values);
```
connor4312, line 148: Rather than manually reading config in this, there is an `IAgentConfigurationService` with `getRootValue`

src/vs/platform/agentHost/node/copilot/copilotAgent.ts
```diff
@@ -1178,58 +1210,213 @@ export class CopilotAgent extends Disposable implements IAgent {
...
+			...client.filter(item =>
+				!!item.plugin
+				&& this._isEnabled(item.customization)
+			).map(item => item.plugin!),
+		];
 	}
 
 	public setEnabled(pluginProtocolUri: string, enabled: boolean) {
 		this._enablement.set(pluginProtocolUri, enabled);
 	}
```
connor4312, line 1290: This is called any time config is updated. Most of the time, this probably won't mean customizations are necessarily updated and we'll fire events unnecessarily

connor4312 (commented):
couple high level comments

src/vs/platform/agentHost/common/agentService.ts
```diff
@@ -481,6 +482,31 @@ export interface IAgent {
...
+	 * Updates the host-owned customization refs available to this agent.
+	 *
+	 * The list is a full replacement of the remote agent host's configured
+	 * plugins after scope resolution and persistence have been applied.
+	 *
+	 * Implementations can invoke `progress` whenever session-visible status
+	 * for the host customizations changes so callers can republish
+	 * `SessionCustomization` state.
+	 */
+	setHostCustomizations?(customizations: CustomizationRef[], progress?: () => void): void;
```
connor4312: Internally in the agent, plugins are managed by the `PluginController`. I wonder if this should have the `AgentHostStateManager` injected (note: we should probably make this `@Injectable` rather than getting passed around everywhere) and handle it internally. Then the individual agent doesn't need to know about these customizations, it just checks with the PluginController at each turn to see if it needs to wait for something to happen.

We could also apply that treatment to setClientCustomizations.

src/vs/platform/agentHost/node/copilot/copilotAgent.ts
joshspicer: i believe this fixes and follows the existing pattern https://github.com/microsoft/vscode/pull/312225/changes/1060c44228a1c996737cff1262fcd301decdbec8

rebornix (dismissed):

dmitrivMS (dismissed):

roblourens (dismissed):

roblourens (dismissed):

roblourens (dismissed):
3rd time's a charm

roblourens (approved):

connor4312 (approved):

PR #312240, opened 2026-04-23
Fix Copilot CLI mission control remote flows

pierceboggan:
This fixes the remaining Mission Control issues in the Copilot CLI remote flow so remote sessions behave correctly end to end without changing the broader Copilot CLI integration behavior.

The change keeps Mission Control scoped to its dedicated client, adds an SDK compatibility fallback for `AutoModeSessionManager`, and fixes the two remote session regressions we hit while testing: permission prompts can now be approved from either VS Code or github.com, and the remote UI now clears stale "Running ..." status when a turn goes idle.

A few non-obvious details are worth calling out:

- `/agents/sessions` traffic is routed through `MissionControlApiClient`, but the broader Copilot CLI integration plumbing is unchanged.
- The `session.start` payload still uses `producer: 'copilot-developer-cli'`.
- The Mission Control `Copilot-Integration-Id` header is still `copilot-developer-cli` for now because we temporarily reverted that part to keep testing unblocked.
- The new tests cover the Mission Control client header, the `AutoModeSessionManager` fallback, remote-vs-local permission resolution, and forwarding `session.idle` to clear remote running state.

How to test:

- Open a Copilot CLI chat session and verify it no longer fails with `AutoModeSessionManager is not a constructor`.
- Enable `/remote`, trigger a permission-gated tool call, and confirm the request can be approved from either VS Code or github.com.
- Run a command remotely and verify the github.com session no longer leaves the gray "Running ..." status behind after the tool finishes.


roblourens (dismissed):

roblourens (dismissed):

extensions/copilot/src/extension/chatSessions/copilotcli/node/missionControlApiClient.ts
```diff
@@ -8,8 +8,8 @@ import { PermissiveAuthRequiredError } from '../../../../platform/github/common/
 import { ILogService } from '../../../../platform/log/common/logService';
 import { IFetcherService } from '../../../../platform/networking/common/fetcherService';
 
-/** Integration id for requests originating from the Copilot CLI /remote feature. */
-const INTEGRATION_ID = 'copilot-developer-cli';
```
rebornix, line 12: let's revert this change to use the constant.

rebornix (dismissed):

extensions/copilot/src/extension/chatSessions/copilotcli/node/missionControlApiClient.ts
pierceboggan: fixed

justschen (approved):

PR #312263, opened 2026-04-23
Add CommandLineMultilineWrapRewriter to handle multi-line POSIX commands

### LINKED ISSUE #312260: run_in_terminal returns partial output for multi-line POSIX commands (RichExecuteStrategy resolves on first onCommandFinished)

meganrogge:
Fixes #312260.

## Problem

`RichExecuteStrategy.execute()` resolves on the **first** `onCommandFinished` shell-integration marker:

```ts
// richExecuteStrategy.ts
Event.toPromise(this._commandDetection.onCommandFinished, store).then(...)
```

When the chat agent pastes a multi-line POSIX command, shell integration fires one marker per top-level statement. The tool returns partial/empty output after line 1 while the rest of the command keeps running unattended.

## Fix

New `CommandLineMultilineWrapRewriter` that wraps multi-line POSIX commands in `<shell> -c '...'` before execution, so shell integration sees exactly one command end marker. Skips Windows/PowerShell and line-continuations (`\<LF>`). Not gated on `detachBackgroundProcesses` — the race affects foreground commands too. `OutputMonitor` input detection is unaffected (it reads the PTY directly).

## Proof — eval run `24857589683`, terminalbench2, vscode agent, gpt-5.4

Affected failing instances whose first tool call was a multi-line heredoc that returned early (`patch=0B` after wasted follow-up turns):

| Test | `systemInitiatedLabel` (first line of submission) |
|---|---|
| `bn-fit-modify` | `cd /app && python3 - <<'PY'` |
| `distribution-search` | `cd /app && python - <<'PY'` |
| `gcode-to-text` | `cd /app && python - <<'PY'` |
| `protein-assembly` | `python - <<'PY'` |
| `raman-fitting` | `cd /app && python - <<'PY'` |
| `rstan-to-pystan` | `cd /app && .venv/bin/python - <<'PY'` |
| `regex-chess` | `python - <<'PY'` |
| `schemelike-metacircular-eval` | `pushd /app >/dev/null && python3 interp.py eval.scm <<'EOF'` |

All 8 produced 0-byte patches after the first `run_in_terminal` call returned before the heredoc body finished. Logs under `.vscode-eval/24857589683/<test>/output/vsc-output/chat-turns/chat-export-step-0.json`.

## Tests

commandLineMultilineWrapRewriter.test.ts — 8/8 green:

```
✔ single-line bash command is not rewritten
✔ escaped newline (line continuation) is not treated as multi-line
✔ Windows is skipped
✔ PowerShell on POSIX is skipped
✔ bash: multi-line command is wrapped in <shell> -c
✔ zsh: multi-line command is wrapped
✔ bash: single quotes in the command are escaped
✔ fish: multi-line command is wrapped with double-quote escaping
```

### DIFF
--- a/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/commandLineRewriter/commandLineMultilineWrapRewriter.ts
+++ b/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/commandLineRewriter/commandLineMultilineWrapRewriter.ts
@@ -1,0 +1,62 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { Disposable } from '../../../../../../../base/common/lifecycle.js';
+import { OperatingSystem } from '../../../../../../../base/common/platform.js';
+import { isFish, isPowerShell } from '../../runInTerminalHelpers.js';
+import type { ICommandLineRewriter, ICommandLineRewriterOptions, ICommandLineRewriterResult } from './commandLineRewriter.js';
+
+/**
+ * Wraps multi-line POSIX commands in `<shell> -c '...'` so shell integration sees a single
+ * command end marker instead of one per line.
+ *
+ * Without this, a command like:
+ *   set -e
+ *   apt-get update
+ *   apt-get install -y foo
+ * is pasted into the terminal as three separate commands. The execute strategy resolves on the
+ * first `onCommandFinished` (after `set -e`) and returns to the agent before `apt-get` has even
+ * started, leaving the remaining work running unattended and colliding with subsequent tool calls.
+ *
+ * Input detection via `OutputMonitor` is unaffected: interactive prompts (passwords, y/n, etc.)
+ * emitted by the wrapped process still surface through the same regex-based detectors.
+ */
+export class CommandLineMultilineWrapRewriter extends Disposable implements ICommandLineRewriter {
+	rewrite(options: ICommandLineRewriterOptions): ICommandLineRewriterResult | undefined {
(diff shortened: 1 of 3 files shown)


meganrogge:
this is a bad approach

PR #312379, opened 2026-04-24
Terminal agent: improve send_to_terminal output capture and add waitForOutput

### LINKED ISSUE #312383: Terminal agent: send_to_terminal improvements for interactive programs
## Problem

In terminalbench evaluation runs ([24890545491](https://github.com/meganrogge/vscode-copilot-evaluation/actions/runs/24890545491), [24890581356](https://github.com/meganrogge/vscode-copilot-evaluation/actions/runs/24890581356)), several failures were caused by `send_to_terminal` returning stale or insufficient output:

### Failure cases this addresses

| Task | Run | Issue |
|------|-----|-------|
| `blind-maze-explorer-5x5` | 24890545491 | Agent sends moves but only sees 5 lines of old output, misses game state |
| `play-zork` | 24890545491 | Agent interacts with text adventure but gets no response output back |
| `decode-braille` | 24890581356 | Agent sends to interactive program, gets stale output |

**~3 failures** directly caused by these limitations, with more failures likely influenced.

### Root cause

1. **Too few output lines**: `send_to_terminal` returned only 5 lines by default — insufficient for most interactive program responses
2. **Too short wait time**: 100ms wait before reading output was not enough for programs that take time to respond
3. **No idle-wait mode**: For interactive programs (games, REPLs), output arrives asynchronously. There was no way to wait for the program to finish responding before reading output
4. **No marker-scoped output**: Output was read from the entire terminal buffer rather than scoped to what appeared after the command was sent, causing stale output from previous commands to be returned

### Fix (on `merogge/send-to-terminal-tweaks`)

1. **Increased default output** from 5 → 20 lines
2. **Increased default wait** from 100ms → 2000ms
3. **Added `waitForOutput` parameter**: When `true`, waits for terminal to become idle (no new data for 2s) before returning output — critical for interactive programs
4. **Marker-scoped output**: Registers an xterm marker before sending text so output is scoped to just the response
5. **Updated model description** to tell the LLM about the 20-line default and `waitForOutput` option

### Schema change

```json
{
  "waitForOutput": {
    "type": "boolean",
    "description": "When true, waits for the terminal to become idle before returning output. Use for interactive programs."
  }
}
```

meganrogge:
## Summary

Improves `send_to_terminal` to capture more useful output and adds a `waitForOutput` mode for interactive programs.

### Output capture improvements
- **Marker-scoped output**: Registers an xterm marker before sending text so output is scoped to just the response, not stale output from previous commands
- **Increased default output** from 5 → 20 lines
- **Increased default wait** from 100ms → 2000ms to give programs time to respond
- **Cancellation-aware timeouts**: All `timeout()` calls now pass the cancellation token so tool cancellation returns promptly

### New `waitForOutput` parameter
- When `true`, waits for the terminal to become idle (no new output for 2s) before returning
- Critical for interactive programs (games, REPLs, text adventures) where the response arrives asynchronously
- 30s max wait with cancellation support
- Added to schema with description guiding the LLM on when to use it

### Model description update
- Updated `modelDescription` to tell the LLM about the 20-line default and `waitForOutput` option

## Testing
- Added schema validation test for `waitForOutput`
- Added `waitForOutput=true` idle-wait test using `runWithFakedTimers`
- Updated mock execution to support `registerMarker` and `onData`
- Preserved all existing multiline/heredoc tests from main

## Eval impact
Addresses ~3 terminalbench failures where the agent interacted with interactive programs but got stale/insufficient output (blind-maze-explorer, play-zork, decode-braille).

Fixes #312383


src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/sendToTerminalTool.ts
```diff
@@ -342,6 +347,9 @@ export class SendToTerminalTool extends Disposable implements IToolImpl {
 			};
 		}
 
+		// Register a marker before sending so we can scope output to just the response
```
anthonykim1, line 350: @meganrogge This make sense. So previously, was the start marker attached to previous response/ earlier in terminal buffer? Or maybe even no marker and we were just reading the last `X` number of terminal buffer.

anthonykim1 (approved):

PR #312387, opened 2026-04-24
fix: schedule promptFallbackScheduler immediately in trackIdleOnPrompt

### LINKED ISSUE #312384: Terminal agent: trackIdleOnPrompt never resolves when no terminal data arrives — causes indefinite hang
## Problem

In terminalbench evaluation runs, **7 tasks hung indefinitely** (3600s timeout) because `run_in_terminal` never completed. All failed with `X_AGENT_STILL_RESPONDING`.

### Affected tasks

| Task | Run | What happened |
|------|-----|---------------|
| `csv-to-parquet` | 24890545491 | Agent ran `python3` (bare REPL), tool blocked forever |
| `build-initramfs-qemu` | 24890545491 | Command hung, no shell integration sequences |
| `cobol-modernization` | 24890545491 | Long compilation, shell integration failed |
| `chess-best-move` | 24890545491 | Shell integration failure, tool never returned |
| `crack-7z-hash` | 24890545491 | Multiple commands, shell integration broke |
| `csv-to-parquet` | 24890581356 | Same as above |
| `build-initramfs-qemu` | 24890581356 | Same as above |

### Root cause

The hang originates in `trackIdleOnPrompt()` (`executeStrategy.ts:162-225`), which is used as a fallback in the `Promise.race` in both `RichExecuteStrategy` and `BasicExecuteStrategy`.

**The bug**: When **no terminal data events arrive at all**, `trackIdleOnPrompt` blocks forever because neither scheduler is ever triggered:

```typescript
// Created but NOT scheduled initially
const scheduler = new RunOnceScheduler(() => {
    idleOnPrompt.complete();
}, idleDurationMs);

// Created but NOT scheduled initially  
const promptFallbackScheduler = new RunOnceScheduler(() => {
    state = TerminalState.PromptAfterExecuting;
    scheduler.schedule();
}, promptFallbackMs ?? 1000);

// The ONLY entry point to schedule either timer
onData(e => {
    // If no data events ever fire, this never runs
    if (state === PromptAfterExecuting) {
        scheduler.schedule();
    } else {
        if (state === Initial || Prompt) {
            promptFallbackScheduler.schedule(); // only scheduled here!
        }
    }
});
```

**The sequence that causes the hang:**
1. `RichExecuteStrategy.execute()` is selected (shell integration was expected)
2. Shell integration fails: _"Shell integration failed to add capabilities within 10 seconds"_
3. `onCommandFinished` never fires (command detection never initialized)
4. `trackIdleOnPrompt` is the last fallback in `Promise.race`
5. The command either hangs (interactive REPL) or finishes silently (no output)
6. **No `onData` events fire → neither scheduler is ever started → promise never resolves**
7. `run_in_terminal` blocks forever → agent cant make next LLM call → 3600s eval timeout

### Fix

Add a separate initial fallback timer (10s) that fires only when no data events arrive at all. This is long enough to avoid false early completion for slow-starting commands (e.g. `sleep 5`), but prevents the infinite hang. Once any data arrives, the initial fallback is canceled and the existing data-driven `promptFallbackScheduler` (1s) takes over.

```typescript
const initialFallbackScheduler = new RunOnceScheduler(() => {
    if (state === Executing || state === PromptAfterExecuting) {
        return;
    }
    state = TerminalState.PromptAfterExecuting;
    scheduler.schedule();
}, 10_000);
initialFallbackScheduler.schedule();

onData(e => {
    // Once any data arrives, cancel the initial fallback
    initialFallbackScheduler.cancel();
    // ... existing logic unchanged
});
```

**Behavior summary:**
- Shell integration works → no change (C sequence sets Executing state, initial fallback guard exits early)
- Shell integration broken + data flowing → no change (onData cancels initial fallback, existing debounce logic handles idle detection)
- Shell integration broken + no data at all → resolves after ~11s instead of hanging forever

This fix applies to both `RichExecuteStrategy` and `BasicExecuteStrategy` since both use `trackIdleOnPrompt`.

meganrogge:
When no terminal data events arrive after a command is sent (e.g. shell integration is broken and the command finishes silently or hangs waiting for input), neither the idle scheduler nor the prompt fallback scheduler was ever triggered, causing trackIdleOnPrompt to block forever.

This caused run_in_terminal to hang indefinitely in both RichExecuteStrategy and BasicExecuteStrategy, leading to 7 timeout failures (X_AGENT_STILL_RESPONDING) in terminalbench evaluation runs.

Fix: schedule the promptFallbackScheduler immediately so it can fire even without data events. If no data arrives within promptFallbackMs (default 1s), the fallback transitions to PromptAfterExecuting and schedules the idle timer, which resolves ~2s after the last data event (or immediately if none arrived). When data does arrive, the existing onData handler reschedules/cancels as before — no behavior change for the normal path.

Fixes #312384

   - Shell integration works → zero behavior change (state guard exits early)
   - Shell integration broken + data flowing → zero change (initial fallback canceled on first data event)
   - Shell integration broken + no data at all → resolves after ~11s instead of hanging forever

### DIFF
--- a/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/executeStrategy/executeStrategy.ts
+++ b/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/executeStrategy/executeStrategy.ts
@@ -181,6 +181,22 @@ export async function trackIdleOnPrompt(
 		state = TerminalState.PromptAfterExecuting;
 		scheduler.schedule();
 	}, promptFallbackMs ?? 1000));
+	// Schedule an initial fallback with a longer timeout so we can detect idle
+	// even when no terminal data events arrive at all (e.g. shell integration
+	// is broken and the command finishes silently or hangs waiting for input).
+	// Without this, if no data events fire, neither scheduler is ever triggered
+	// and trackIdleOnPrompt blocks forever. We use a longer initial delay (10s)
+	// to avoid falsely reporting completion for commands that are slow to start
+	// producing output. Once any data arrives, the onData handler takes over
+	// with the shorter promptFallbackMs interval.
+	const initialFallbackScheduler = store.add(new RunOnceScheduler(() => {
+		if (state === TerminalState.Executing || state === TerminalState.PromptAfterExecuting) {
+			return;
+		}
+		state = TerminalState.PromptAfterExecuting;
+		scheduler.schedule();
+	}, 10_000));
+	initialFallbackScheduler.schedule();
 	// Only schedule when a prompt sequence (A) is seen after an execute sequence (C). This prevents
 	// cases where the command is executed before the prompt is written. While not perfect, sitting
 	// on an A without a C following shortly after is a very good indicator that the command is done
@@ -194,6 +210,9 @@ export async function trackIdleOnPrompt(
 		PromptAfterExecuting,
 	}
 	store.add(onData(e => {
+		// Once any data arrives, cancel the initial fallback — the data-driven
+		// promptFallbackScheduler handles rescheduling from here.
+		initialFallbackScheduler.cancel();
 		// Update state
 		// p10k fires C as `133;C;`
 		const matches = e.matchAll(/(?:\x1b\]|\x9d)[16]33;(?<type>[ACD])(?:;.*)?(?:\x1b\\|\x07|\x9c)/g);


anthonykim1 (approved):

anthonykim1:
This is great, So this would fix "silent" case where shell integration could be broken (and trackIdleOnPrompt is used) + commands that would hang with no output would now resolve after ~11s instead of indefinite hang.

PR #312424, opened 2026-04-24
PromptsService: remove 'when' from APIs

aeschli:


src/vs/workbench/contrib/chat/common/promptSyntax/service/promptsServiceImpl.ts
```diff
@@ -709,9 +709,6 @@ export class PromptsService extends Disposable implements IPromptsService {
 	private asChatPromptSlashCommand(argumentHint: string | undefined, userInvocable: boolean | undefined, promptPath: IPromptPath): IChatPromptSlashCommand {
 		let name = promptPath.name ?? getCleanPromptName(promptPath.uri);
 		name = name.replace(/[^\p{L}\d_\-\.:]+/gu, '-'); // replace spaces with dashes
-		const when = isExtensionPromptPath(promptPath) && promptPath.when
```
pwang347, line 688: I think the old logic was to filter out customizations here if the `when` condition is false using global context. Are we removing that as well? (ie., will this stop working? https://github.com/microsoft/vscode/blob/faf8592a971810de677429607be716e2db04c0de/extensions/copilot/package.json#L6340)

roblourens (dismissed):

src/vs/workbench/contrib/chat/common/promptSyntax/service/promptsServiceImpl.ts
aeschli: That happens in the promptsService, in getExtensionPromptFiles. Only the promptsService uses the `when` internally to filter. The items that are passed out via the APIs are already filtered and therefore no longer need a `when`. Thats what the PR changes.

roblourens (dismissed):

alexdima (approved):

PR #312435, opened 2026-04-24
Fix open router prompt caching with messages API

lramos15:


extensions/copilot/src/extension/byok/node/openAIEndpoint.ts
```diff
@@ -251,6 +251,9 @@ export class OpenAIEndpoint extends ChatEndpoint {
 				body.previous_response_id = undefined;
 			}
 			return body;
+		} else if (this.useMessagesApi) {
+			// Delegate to base ChatEndpoint for Messages API dispatch
+			return super.createRequestBody(options);
```
bhavyaus, line 256: We also need to add in the right headers for this to work with thinking.

Yoyokrazy (approved):

extensions/copilot/src/extension/byok/node/openAIEndpoint.ts
```diff
@@ -251,6 +251,9 @@ export class OpenAIEndpoint extends ChatEndpoint {
 				body.previous_response_id = undefined;
 			}
 			return body;
+		} else if (this.useMessagesApi) {
```
bhavyaus, line 254: Can't you delegate it to the AnthropicProvider instead? That's more BYOK than messagesAPI is. I can't promise to keep the messagesAPI CAPI agnostic.

bhavyaus (changes requested):

bhavyaus (approved):

roblourens (approved):

PR #312477, opened 2026-04-25
fix: prevent duplicate Simple Browser panels on rapid show/restore race

raagad:
When `show()` is called multiple times rapidly during window reload, multiple Simple Browser panels could be created before `_activeView` was set. Similarly, `restore()` could revive an old panel even when a new one already existed. This fix adds a `_simpleBrowserAlreadyOpen()` check in `show()` that scans all tab groups (not just the active one) to prevent duplicate panel creation. It also updates `restore()` to dispose the restored panel if an active view already exists.


raagad:

PR #312486, opened 2026-04-25
Simplify subagent processing for Claude

TylerLeonhardt:
Now that there's a getSubagentMessages as part of the API, we don't need a lot of this manual reading of JSONL files.

Additionally, I noticed that we did not change our logic to handle the fact that they changed the "Task" tool to "Agent"...

Additionally, there was a bug in the Agents app where it wasn't properly handling when you clicked between sessions in different folders.

All that to say, subagents work better and session info in handled better as well.


bhavyaus (approved):

alialsahli2030ali-rgb:
ASS msisnsh

PR #312492, opened 2026-04-25
webview: respect default localResourceRoots for custom editors

### LINKED ISSUE #282495: Custom editor webview's localResourceRoots contract is not respected

- VS Code Version: 1.106.3
- OS Version: Windows 10

Steps to Reproduce:

1. Create an extension project using an npm workspace (or pnpm, Yarn) with multiple packages.
2. Call `vscode.window.registerCustomEditorProvider` from a secondary package (not the one with the extension entry point).
3. Notice that the custom editor doesn't open correctly, with a `401` error (unauthorized) for all resources.

The `WebviewOptions.localResourceRoots` says:

> Default to ... the extension's install directory.

But this is clearly not respected. Is this an expected behavior? It looks like VS Code is able to infer the editor is registered from another package. Note that this only happens when the code is **not** bundled (e.g., when running the extension host for development purposes).

The workaround is to explicitly specify `localResourceRoots`:
```ts
resolveCustomEditor(document: MyEditorDocument, webviewPanel: vscode.WebviewPanel, token: vscode.CancellationToken): void {
  // ...
  webviewPanel.webview.options = {
    enableScripts: true,
    localResourceRoots: [webviewBaseUri],
  };
  // ...
}
```

maruthang:
Fixes #282495

## Problem

Per the [`WebviewOptions.localResourceRoots` docs](https://code.visualstudio.com/api/references/vscode-api#WebviewOptions), if the option is left unset the webview should default to allowing the workspace folder roots plus the extension's install directory. `vscode.window.createWebviewPanel` honors this contract — `ExtHostWebviewPanels.createWebviewPanel` fills in the default before the create RPC.

Custom editors do not. The main thread builds the overlay with `contentOptions: {}` and then calls `$resolveCustomEditor` on the extension host. Because the extension host never receives a chance to fill in the default, `asWebviewUri` returns HTTP 401 for resources under the workspace or the extension folder unless the extension explicitly sets `webview.options = { localResourceRoots: [...] }`. This silently breaks custom-editor extensions that follow the documented default.

## Fix

Added `ExtHostWebviews.ensureDefaultContentOptions(handle, contentOptions, extension)`. When `contentOptions.localResourceRoots` is `undefined`, it calls `$setOptions` with the documented default (workspace folders + extension install directory, mirroring the logic in `ExtHostWebviewPanels.createWebviewPanel`). When the extension provided explicit roots — including an empty array — the helper does nothing.

`MainThreadCustomEditors.$resolveCustomEditor` invokes the helper from `ExtHostCustomEditors` immediately after `createNewWebview`, before the provider's `resolveCustomTextEditor` / `resolveCustomEditor` runs. Existing `webview.options = ...` assignments inside provider callbacks continue to work — those go through `serializeWebviewOptions`, which already re-applies the default.

## Tests

Added three unit tests in `src/vs/workbench/api/test/browser/extHostWebview.test.ts`:

- default fills in workspace folders + extension install directory when `localResourceRoots` is omitted
- explicit `localResourceRoots` (even `[]`) is respected — `$setOptions` is not called
- with no workspace open, the default is just the extension install directory

### DIFF
--- a/src/vs/workbench/api/common/extHostCustomEditors.ts
+++ b/src/vs/workbench/api/common/extHostCustomEditors.ts
@@ -271,6 +271,11 @@ export class ExtHostCustomEditors implements extHostProtocol.ExtHostCustomEditor
 		const viewColumn = typeConverters.ViewColumn.to(position);
 
 		const webview = this._extHostWebview.createNewWebview(handle, initData.contentOptions, entry.extension);
+		// The main thread starts the custom editor's webview with empty content
+		// options. Ensure `localResourceRoots` defaults to the workspace folders
+		// and the providing extension's install directory, as documented on
+		// `WebviewOptions.localResourceRoots`.
+		this._extHostWebview.ensureDefaultContentOptions(handle, initData.contentOptions, entry.extension);
 		const panel = this._extHostWebviewPanels.createNewWebviewPanel(handle, viewType, initData.title, viewColumn, initData.options, webview, initData.active);
 
 		const revivedResource = URI.revive(resource);
--- a/src/vs/workbench/api/common/extHostWebview.ts
+++ b/src/vs/workbench/api/common/extHostWebview.ts
@@ -246,6 +246,26 @@ export class ExtHostWebviews extends Disposable implements extHostProtocol.ExtHo
 		return webview;
 	}
 
+	/**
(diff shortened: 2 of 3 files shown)


mjbvz (approved):
Thanks!

bpasero (approved):

PR #312590, opened 2026-04-26
docs: clarify TabInputWebview.viewType prefix in vscode.d.ts

### LINKED ISSUE #201884: TabInputWebview viewType comment is inaccurate/unclear or there is a bug in viewType
`TabInputWebview.viewType` in vscode.d.ts is described as:
> The type of webview. Maps to `{@linkcode WebviewPanel.viewType WebviewPanel's viewType}`

`WebviewPanel.viewType` in vscode.d.ts says:
> Identifies the type of the webview panel, such as `'markdown.preview'`.

However, when checking a TabInputWebview that I have, the `viewType` is for the form `mainThreadWebview-markdown.preview`.

Should the `mainThreadWebview-` prefix either be stripped from here, or a node added to the docstring in vscode.d.ts making sure that users know to expect it?

ousamabenyounes:
## Why

Issue #201884 reports that the JSDoc for `TabInputWebview.viewType` claims it "Maps to `WebviewPanel.viewType`", but the value exposed on `TabInputWebview` is actually the panel viewType prefixed with `mainThreadWebview-` (e.g. `markdown.preview` → `mainThreadWebview-markdown.preview`). Extension authors comparing the two values get unexpected mismatches.

Either the prefix should be stripped in the runtime, or the docs should be updated. Updating the docs is the safe, non-breaking option — stripping the prefix could break extensions that already filter on the prefixed form.

## What

Update the JSDoc on `TabInputWebview.viewType` (and its constructor parameter) in `src/vscode-dts/vscode.d.ts` to:

- Mention the `mainThreadWebview-` prefix.
- Give a concrete example so the relationship to `WebviewPanel.viewType` is unambiguous.

Documentation-only change. No runtime behavior change.

Fixes #201884

### DIFF
--- a/src/vscode-dts/vscode.d.ts
+++ b/src/vscode-dts/vscode.d.ts
@@ -19216,12 +19216,15 @@ declare module 'vscode' {
 	 */
 	export class TabInputWebview {
 		/**
-		 * The type of webview. Maps to {@linkcode WebviewPanel.viewType WebviewPanel's viewType}
+		 * The type of webview. Corresponds to {@linkcode WebviewPanel.viewType WebviewPanel's viewType},
+		 * but prefixed with `mainThreadWebview-`. For example, a `WebviewPanel` created with
+		 * viewType `markdown.preview` is exposed here as `mainThreadWebview-markdown.preview`.
 		 */
 		readonly viewType: string;
 		/**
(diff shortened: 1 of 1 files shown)


mjbvz:
Let's fix the bug instead. The mainthread part should not leak out to extensions

PR #312593, opened 2026-04-26
docs: clarify when WorkspaceEditEntryMetadata.label is shown

### LINKED ISSUE #176984: Workspace edits with metadata do not show labels that docs say are rendered prominently
In `WorkspaceEditEntryMetadata`, the `label` field says this:

```ts
/**
 * A human-readable string which is rendered prominent.
 */
label: string;
```

In LSP, it's slightly expanded:

```
	/**
	 * A human-readable string describing the actual change. The string
	 * is rendered prominent in the user interface.
	 */
	label: string;
```

This makes it seem like these labels will be visible in the UI by default. Yet if you run:

```ts
const doc = await vs.workspace.openTextDocument(vs.Uri.parse("untitled:foo.txt"));
const editor = await vs.window.showTextDocument(doc);
const edit = new vs.WorkspaceEdit();
edit.insert(doc.uri, new vs.Position(0, 0), "test1", { label: "This is my label", needsConfirmation: true });
edit.insert(doc.uri, new vs.Position(0, 0), "test2", { label: "This is my label2", needsConfirmation: true });
vs.workspace.applyEdit(edit, { isRefactoring: false });
```

It is rendered like this without the labels:

Screenshot 2023-03-13 at 16 45 44

You have to right-click and tick "Group change by type" to see the labels:

Screenshot 2023-03-13 at 16 46 10

I think it would be better if the items were grouped by default (since the server went to lengths to provide the labels, which are not optional if they want to prompt for confirmation), but if not I think the API comments could be a little clearer (or even the label made optional).

ousamabenyounes:
## Why

Issue #176984 reports that the JSDoc on `WorkspaceEditEntryMetadata.label` calls the string "rendered prominent", but in the refactor preview the label only appears when entries are grouped — and grouping is off by default. Extension authors set a label, expect it to show, and instead see a flat list of edits where their label is nowhere visible.

The interface comment already mentions that the editor "groups edits with equal labels into tree nodes", but the per-property JSDoc on `label` doesn't make the conditional rendering clear.

## What

In `src/vscode-dts/vscode.d.ts`:

- Expand the `label` JSDoc to spell out that it is the group's tree-node header.
- Note that the preview lists entries flat by default and the user has to enable "Group change by type" to see the labels.
- Tweak `description` to mention the relationship to `label`.

Documentation-only change. No runtime behavior change.

Fixes #176984

### DIFF
--- a/src/vscode-dts/vscode.d.ts
+++ b/src/vscode-dts/vscode.d.ts
@@ -3935,12 +3935,17 @@ declare module 'vscode' {
 		needsConfirmation: boolean;
 
 		/**
-		 * A human-readable string which is rendered prominent.
+		 * A human-readable string which is rendered prominently when entries are
+		 * grouped (see {@link WorkspaceEditEntryMetadata}). In the refactor preview,
+		 * this label appears as the group's tree node header — entries sharing a
+		 * label are collected under that node. Note that the preview lists entries
+		 * flat by default; users see the label after enabling "Group change by type".
 		 */
 		label: string;
 
 		/**
-		 * A human-readable string which is rendered less prominent on the same line.
+		 * A human-readable string which is rendered less prominent on the same line
+		 * as {@link label} when entries are grouped.
 		 */
 		description?: string;
 


ousamabenyounes:
Closing — on reflection, a docs-only change doesn't address the root cause that #176984 raises (label set, expected to render prominently, doesn't unless grouping is enabled). The runtime path proposed by @DanTup in the issue (a `groupChangesByDefault` flag on `WorkspaceEditMetadata`) is the right shape, but it's an API design decision that belongs with the API team rather than a drive-by PR. Will leave the issue to that discussion.

PR #312693, opened 2026-04-27
Run Agents App in transient mode when VS Code is launched with --transient

sandy081:
## Summary

When VS Code is launched with `--transient`, the Agents App now also runs in transient mode with its own dedicated temporary user data and extensions directories, created under the same temp parent folder.

## Changes

- **`argv.ts` (common + node)**: Add `--agents-user-data-dir` and `--agents-extensions-dir` CLI args
- **`cli.ts`**: In the `--transient` block, create `agents-data/` and `agents-extensions/` subdirs alongside the existing temp dirs, pass them via the new args, and include them in the relaunch hint message
- **`nativeHostMainService.ts`**: When launching the sibling Agents app (`launchSiblingApp`), forward the transient dirs as `--user-data-dir` and `--extensions-dir` so the Agents process also runs fully isolated in transient mode

### DIFF
--- a/src/vs/code/node/cli.ts
+++ b/src/vs/code/node/cli.ts
@@ -250,13 +250,17 @@ export async function main(argv: string[]): Promise<void> {
 			const tempExtensionsDir = join(tempParentDir, 'extensions');
 			const tempSharedDataDir = join(tempParentDir, 'shared');
 			const tempAgentPluginsDir = join(tempParentDir, 'agent-plugins');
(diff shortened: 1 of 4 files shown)


sandy081 (commented):
Addressed in 08341e8. Also forwarding `--shared-data-dir` and `--agent-plugins-dir` to the sibling Agents app so all four transient dirs are propagated.

bpasero (approved):

PR #312723, opened 2026-04-27
fix html: set editor.wrappingIndent to 'indent' by default

### LINKED ISSUE #312719: A word wrap indent ("editor.wrappingIndent") option for a FIXED amount of spaces from the margin
`"editor.wrappingIndent"` currently supports the following options:

Image

But all of the options `same`, `indent`, and `deepindent` inherit the indent from the line above them. Only `none` keeps a fixed length respective to the margin (0).

In my opinion, `none` is good for code because it is obvious that the line has wrapped, since it does *not* match the block it belongs to, but stands out. 

Example:
Image

Meanwhile, `same` would hide that a wrap happened:
Image

In deeply indented code, `indent` and `deepindent` can end up wrapping multiple times, which is just plain hard to read (this example isn't even that bad):
Image

In any case, the ideal is *not* to wrap, but sometimes you end up with long lines of code that syntax will not allow you to break.
Personally I prefer a wrap over a scrollbar in those cases, which is an opinion but that's why settings exist.

However, `none` is bad for text, because text will generally not be indented the way that code is.
In that case, `indent` is better because there usually is space for it *and* it will be visually obvious that a line is too long compared to the ones around it.

Example where it isn't obvious wrapping happened with `none`:
Image

So, the feature request is to make a compromise option, an option for a fixed indent length on wrapping that deliberately does *not* have to match the indent of the line itself, but is only spaced with regard to the margin.

It would also be nice to specify the number of spaces manually, such that it does not have to be the same as the file's indent option.
This makes the wrap visually different from the normal (meaningful) indentation in the source file, which is helpful.

AmSach:
For issue #312719 - HTML/handlebars text is harder to read when wrapping uses 'same' indent, making wrapped lines invisible. Set default wrappingIndent to 'indent' so wrapped text lines are visually distinct.

Fixes microsoft/vscode#312719

### DIFF
--- a/extensions/html-language-features/package.json
+++ b/extensions/html-language-features/package.json
@@ -246,10 +246,12 @@
     },
     "configurationDefaults": {
       "[html]": {
-        "editor.suggest.insertMode": "replace"
+        "editor.suggest.insertMode": "replace",
+        "editor.wrappingIndent": "indent"
       },
       "[handlebars]": {
-        "editor.suggest.insertMode": "replace"
+        "editor.suggest.insertMode": "replace",
+        "editor.wrappingIndent": "indent"
       }
     },
     "jsonValidation": [


ericludvigs:
This does not address the original issue, https://github.com/microsoft/vscode/issues/312719.

aeschli:
Sorry, I'd rather not change the default settings at this moment as this will cause ripples to existing users.
Users that prefer 'editor.wrappingIndent to 'indent' can do so in their user settings

PR #312822, opened 2026-04-27
Fix empty session_files/refs by queuing early tool spans

### LINKED ISSUE #312297: Chronicle: session_files and session_refs tables are always empty locally
## Problem

The `session_files` and `session_refs` tables in the local Chronicle store contain zero rows despite the tracking code existing in `_handleToolSpan`. The schema description promises file edit tracking and PR/commit/issue references, causing the LLM to waste tool calls querying empty tables.

## Current code

`_handleToolSpan` in `sessionStoreTracker.ts` does extract file paths and refs from tool spans:
- `extractFilePath` handles `apply_patch`, `str_replace_editor`, `create_file`, `create` tools
- `extractRefsFromMcpTool` / `extractRefsFromTerminal` extract PR/issue/commit refs

## Likely causes

1. Tool spans may not carry the expected `gen_ai.tool.name` attribute, or the tool names don't match the tracked set (`apply_patch`, `str_replace_editor`, `create_file`, `create`).
2. Tool spans may not have a `session_id` attribute, causing `_getSessionId` to return undefined and skip tracking.
3. The buffer flush may not be writing files/refs rows correctly.

## Investigation steps

1. Add logging or check OTel debug panel to verify tool spans have `gen_ai.tool.name` and a session ID attribute.
2. Check whether the tool names used in VS Code Copilot match `FILE_TRACKING_TOOLS` in `sessionStoreTracking.ts`.
3. Verify the SQL INSERT for files/refs in the session store flush logic.

## Files

- `extensions/copilot/src/extension/chronicle/vscode-node/sessionStoreTracker.ts` — `_handleToolSpan`
- `extensions/copilot/src/extension/chronicle/common/sessionStoreTracking.ts` — `extractFilePath`, `extractRefsFromMcpTool`, `extractRefsFromTerminal`
- `extensions/copilot/src/platform/chronicle/common/sessionStore.ts` — flush/insert logic

Follow-up from #312292

digitarald:
## Summary
Fixes #312297 - Chronicle's `session_files` and `session_refs` tables were always empty locally.

## Root Causes
1. **Span timing**: Tool spans (EXECUTE_TOOL) complete *before* their parent `invoke_agent` span, so they arrive before `_initSession()` is called. Previously these were silently dropped.
2. **Wrong attribute**: Code used `gen_ai.tool.input` but the actual OTel attribute is `gen_ai.tool.call.arguments`.

## Changes
- Queue tool spans in `_pendingToolSpans` Map when session not yet initialized
- Process queued spans in `_initSession()` after session is ready
- Fix attribute name to use `GenAiAttr.TOOL_CALL_ARGUMENTS`
- Add telemetry: `pendingSpansProcessed`, `filesCount`, `refsCount`
- Clean up pending spans on session dispose
- Add unit tests for the span handling fixes


digitarald:
Addressed review comments:

1. **Timestamp format**: The tests already use numeric timestamps (`startTime: 0, endTime: 1`) which are correct per `ICompletedSpanData` types.

2. **Tests re-implementing logic**: Moved `extractToolArgs` to `sessionStoreTracking.ts` as an exported helper, and updated both the implementation and tests to use it. Tests now exercise the real code path instead of a local reimplementation.

vijayupadya (approved):

PR #312827, opened 2026-04-27
Fix terminal hang when process exits without shell integration sequences

meganrogge:
## Problem

When a terminal process exits without emitting shell integration prompt sequences, `trackIdleOnPrompt` blocks forever because the `Executing` state cancels both schedulers and no more data arrives to reschedule them. This causes `run_in_terminal` to never complete, eventually timing out after 1 hour.

### Evidence from eval run `24890545491` (VS Code, terminalbench, gpt-5.4)

Analysis of the VS Code logs (`vsc-output/productLogs/terminal.log`) found **21 tasks with terminal hangs** — commands that started with the `rich` execute strategy but never completed (no corresponding `Finished` log line). Of these, **7 timed out entirely** (`X_AGENT_STILL_RESPONDING`), and 14 others lost significant time before recovering or failing for other reasons.

Tasks with terminal hangs that timed out:

| Task | Rich cmds | Finished | Hung | Shell integration failed |
|------|-----------|----------|------|--------------------------|
| cobol-modernization | 13 | 12 | 1 | yes |
| crack-7z-hash | 42 | 41 | 1 | yes |
| csv-to-parquet | 1 | 0 | 1 | yes |
| make-doom-for-mips | 2 | 1 | 1 | yes |
| neuron-to-jaxley-conversion | 6 | 5 | 1 | yes |
| reshard-c4-data | 3 | 2 | 1 | yes |
| build-initramfs-qemu | 3 | 2 | 1 | yes |

Tasks with terminal hangs that partially recovered:

| Task | Rich cmds | Finished | Hung | Shell integration failed |
|------|-----------|----------|------|--------------------------|
| git-multibranch | 19 | 15 | 4 | yes |
| make-mips-interpreter | 40 | 35 | 5 | yes |
| play-zork | 63 | 61 | 2 | yes |
| polyglot-rust-c | 27 | 24 | 3 | yes |
| solana-data | 21 | 19 | 2 | yes |
| build-tcc-qemu | 37 | 34 | 3 | yes |
| qemu-alpine-ssh | 12 | 10 | 2 | yes |
| hf-model-inference | 5 | 3 | 2 | yes |
| blind-maze-explorer-5x5 | 1 | 0 | 1 | yes |
| intrusion-detection | 10 | 9 | 1 | no |
| kv-store-grpc | 4 | 3 | 1 | no |
| mlflow-register | 9 | 8 | 1 | yes |
| run-pdp11-code | 20 | 19 | 1 | no |
[...]

PR #312854, opened 2026-04-27
fix: downgrade rich to basic execute strategy when shell integration breaks mid-command

### LINKED ISSUE #312853: Terminal tool hangs when shell integration breaks mid-command in rich execute strategy
## Problem

When the terminal tool uses the `rich` execute strategy, commands can hang indefinitely if shell integration sequences stop arriving mid-command. This caused **19 out of 45 eval failures** (42%) in eval run 24966302375 — and **15 of those 19 tasks passed in the CLI** with the same model (gpt-5.4).

## Root Cause

In `trackIdleOnPrompt()` (`executeStrategy.ts`):

1. Command starts → terminal data events arrive (command output)
2. Data arrival **cancels** the `initialFallbackScheduler` (the 10s safety net from #312387)
3. If a `C`/`D` shell integration sequence is detected, state transitions to `Executing`
4. In `Executing` state, `promptFallbackScheduler` is also **cancelled**
5. Shell integration breaks → no `A` (prompt) sequence ever arrives
6. **Both fallback schedulers are cancelled** → the Promise never resolves → permanent hang

The `initialFallbackScheduler` from #312387 only helps when **no data arrives at all**. When data arrives but shell integration sequences are incomplete, the hang persists.

## Evidence

Every timed-out task in the eval shows this pattern in `terminal.log`:
```
20:59:25 [info] Using `rich` execute strategy for command ...
20:59:35 [warning] Shell integration failed to add capabilities within 10 seconds
[nothing for 60 minutes until timeout]
```

| Metric | Value |
|---|---|
| Tasks that timed out in VS Code | 19 |
| Same tasks that passed in CLI | 15 |
| CLI avg completion time | 3-20 min |
| VS Code active time before hang | typically <1 min |
| Tasks where VS Code never completed a single command | 5 |

## Proposed Fix

When shell integration capabilities fail to load within 10 seconds (the warning we already log), **downgrade from `rich` to `basic` strategy mid-flight**. The basic strategy uses data-idle polling rather than shell integration sequences, so it will not hang.

## Related

- #312387 — Added `initialFallbackScheduler` (partial fix, only covers no-data case)
- #312827 — Added `onExit` race (fixes process-exit case, but not long-running commands)

meganrogge:
## Problem

Fixes #312853

The terminal `rich` execute strategy hangs indefinitely when shell integration sequences stop arriving mid-command. This was the **single root cause of all 19 timeouts** (42% of failures) in eval run `24966302375`.

### The hang mechanism

In `trackIdleOnPrompt()`:

1. Command starts → data events arrive → `initialFallbackScheduler` is cancelled
2. If a `C`/`D` shell integration sequence is seen, state → `Executing`
3. In `Executing` state, `promptFallbackScheduler` is also cancelled
4. Shell integration breaks → no `A` (prompt) sequence ever arrives
5. **Both fallbacks cancelled → permanent hang**

### Eval evidence

Every timed-out task shows this exact pattern in `terminal.log`:
```
20:59:25 [info] Using `rich` execute strategy for command ` set -e; uname -s; ...`
20:59:35 [warning] Shell integration failed to add capabilities within 10 seconds
[nothing for 60 minutes until timeout]
```

**15 of 19 timed-out tasks passed in the CLI** with the same model (gpt-5.4), completing in 3-20 minutes. VS Code typically hung within the first minute — 5 tasks never completed a single terminal command.

| Task | CLI time | VS Code active before hang | CLI tool calls |
|------|----------|---------------------------|----------------|
| compile-compcert | 20 min | 8s (0 cmds completed) | 117 |
| sqlite-db-truncate | 1.6 min | 42s (0 cmds completed) | 14 |
| caffe-cifar-10 | 16 min | 16s (1 cmd completed) | 85 |
| build-cython-ext | 7 min | 289s (1 cmd completed) | 66 |
| adaptive-rejection-sampler | 6 min | 36s (3 cmds completed) | 19 |
| make-doom-for-mips | 22 min | 20 min (34 cmds completed) | 186 |

### Fix

When the `rich` execute strategy detects that shell integration capabilities have failed (the 10-second warning we already log), **abort the rich strategy and fall back to the `basic` strategy mid-flight**. The basic strategy uses data-idle polling instead of shell integration sequences, so it does not hang.

This is complementary to:
- #312387 — `initialFallbackScheduler` (only covers no-data-at-all case)
- #312827 — `onExit` race (only covers process-exit case, not long-running commands)

This fix covers the remaining gap: commands that **produce data** but where **shell integration breaks**, including long-running processes.

| Scenario | SI works? | Before fix | After fix |
 |---|---|---|---|
 | **Quick command (<30s)** | ✅ | `onCommandFinished` wins race | Same — `onCommandFinished` wins |
 | **Long command (>30s), steady output** | ✅ | `onCommandFinished` wins race | Same — fallback reschedules on every data event, `onCommandFinished` wins |
 | **Long command, 30s+ output pause** | ✅ | `onCommandFinished` wins race | Same — `onCommandFinished` still wins the race |
 | **Quick command, SI breaks after C/D** | ❌ | ❌ **Hangs forever** | ✅ Fallback fires after 30s data-idle |
 | **Long command with output, SI breaks** | ❌ | ❌ **Hangs forever** | ✅ Fallback fires 30s after last output |
 | **Command finishes silently (no data at all)** | ❌ | ✅ `initialFallbackScheduler` at 10s | Same — unchanged |
 | **Long command, 30s+ pause, SI broken** | ❌ | ❌ **Hangs forever** | ⚠️ Completes early during the pause |

### DIFF
--- a/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/executeStrategy/executeStrategy.ts
+++ b/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/executeStrategy/executeStrategy.ts
@@ -202,13 +202,32 @@ export async function trackIdleOnPrompt(
 	// follows. Both initialFallbackScheduler and promptFallbackScheduler get
 	// cancelled in that state, causing a permanent hang. This scheduler is
 	// rescheduled on every data event while in the Executing state, so it only
-	// fires after 10s of data-idle — effectively downgrading from rich to basic
-	// behavior by treating the data-idle as completion.
-	const executingFallbackScheduler = store.add(new RunOnceScheduler(() => {
-		if (state === TerminalState.Executing) {
-			state = TerminalState.PromptAfterExecuting;
-			scheduler.schedule();
+	// fires after 10s of data-idle. When it fires, it checks if the cursor line
+	// looks like a prompt before completing — if not, it reschedules to avoid
+	// prematurely completing long-running commands that pause without output.
+	const executingFallbackScheduler = store.add(new RunOnceScheduler(async () => {
+		if (state !== TerminalState.Executing) {
+			return;
 		}
+		// Check if the terminal looks like it's at a prompt before completing.
+		// Long-running commands can pause for extended periods without output
+		// (e.g. downloading, compiling), so we only complete if the cursor line
+		// matches a known prompt pattern.
+		const xterm = await instance.xtermReadyPromise;
+		if (xterm) {
+			const buffer = xterm.raw.buffer.active;
+			const line = buffer.getLine(buffer.baseY + buffer.cursorY);
+			if (line) {
+				const content = line.translateToString(true);
+				if (!detectsCommonPromptPattern(content).detected) {
+					// Doesn't look like a prompt — reschedule and check again later
+					executingFallbackScheduler.schedule();
+					return;
+				}
+			}
+		}
+		state = TerminalState.PromptAfterExecuting;
+		scheduler.schedule();
 	}, 10_000));
 	// Only schedule when a prompt sequence (A) is seen after an execute sequence (C). This prevents
 	// cases where the command is executed before the prompt is written. While not perfect, sitting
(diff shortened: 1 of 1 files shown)


meganrogge:
cc @anthonykim1

meganrogge:
~requires-eval-assessment terminalbench2

anthonykim1 (approved):

PR #312856, opened 2026-04-27
Persist terminal color changes (via escape sequences) correctly

### LINKED ISSUE #312815: Terminal color changes (via OSC 10/11/12 sequences) do not persist correctly
Does this issue occur when all extensions are disabled: Yes

- VS Code Version: 1.117.0
- OS Version: Windows_NT x64 10.0.26200

Steps to Reproduce:

1. Open the terminal in editor area
2. Emit a osc 11 sequence, like turning the background grey: `printf '\e]11;#888888\e\\'`
3. Change the tab 
4. Change back to the terminal tab
5. Observe the background going back to default

OR
1. Open the terminal in editor area or usual terminal area
2. Emit a osc 11 sequence, like turning the background grey: `printf '\e]11;#888888\e\\'`
4. Change the color theme
5. Observe the terminal colors getting changed

When you use the usual terminal area and different terminals even use different colors via osc11/12/13, that area does work and persists the colors correctly, when switching between tabs, the editor area does not.

But in all areas in doing theme changes DOES change the terminal color even after the terminal specified a background/foreground/cursor color using OSC 10/11/12, and they def seem related issues.

https://github.com/user-attachments/assets/15460787-cb79-4ead-9c9b-633f4e4432bd

mastondzn:
Fixes #312815

Previously, `XtermTerminal._updateTheme()` sets `this.raw.options.theme = newTheme`, which triggers xterm.js's `ThemeService._setTheme()`. This completely overwrites all current colors, including any OSC10/11/12-modified background, foreground, or cursor colors, because it updates both the active colors and the restore colors simultaneously. Refreshing the terminal, changing the theme color or switching the tab while the terminal is open in the editor area results in all the OSC set colors being changed to the current theme color.

With this PR, `_updateTheme()` preserves OSC-modified colors by comparing the current colors against their original restore values before applying the new theme, and re-applies any that differ using `modifyColors()`. This ensures OSC 10/11/12 colors survive theme refreshes caused by tab switches or VS Code theme changes.

However there is a quite rare edge case i have decided not to handle:
For example, if OSC 11 sets the background to a specific color, and the theme background also happens to be exactly that specific color, then in this case, this color would not be persisted. If the theme later changes to a different color, the background would follow the new theme instead of staying at that specific color.

To fix this edge case vscode would need to track terminal color modifications independently of xterm.js. This has to be extremely rare in practice, and i doubt anyone would notice it.

### DIFF
--- a/src/vs/workbench/contrib/terminal/browser/xterm-private.d.ts
+++ b/src/vs/workbench/contrib/terminal/browser/xterm-private.d.ts
@@ -7,6 +7,12 @@ import { IBufferCell } from '@xterm/xterm';
 
 export type XtermAttributes = Omit<IBufferCell, 'getWidth' | 'getChars' | 'getCode'> & { clone?(): XtermAttributes };
 
+export interface IXtermThemeColors {
+	background: { css: string; rgba: number };
+	foreground: { css: string; rgba: number };
+	cursor: { css: string; rgba: number };
+}
+
 export interface IXtermCore {
 	viewport?: {
 		readonly scrollBarWidth: number;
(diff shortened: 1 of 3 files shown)


mastondzn:

PR #312873, opened 2026-04-27
Be less aggressive about removing terminal in agents app

### LINKED ISSUE #312872: agents app auto-closes terminal when first opened
I'm thinking the culprit may be how  closeTerminalsForPath  disposes only using initial cwd.

anthonykim1:
Resolves: https://github.com/microsoft/vscode/issues/312872 

https://github.com/user-attachments/assets/2f02dcec-89e6-445b-aa25-0c766d355fad

/cc @TylerLeonhardt 

Agents app code has this logic where terminals are forced to close in conditions like archiving/changing sessions based on cwd. Removal was based on cwd, but also cwd was used to share terminal for multiple session which doesn't make sense to me..

I think it's fair to not have such removal, or figure out better way to associate particular session with particular terminal? 

Even for associating particular session with particular terminal, user always has to freedom to cd in/out of particular directory so labeling a terminal to a particular session seems to aggressive.

### DIFF
--- a/src/vs/sessions/contrib/terminal/browser/sessionsTerminalContribution.ts
+++ b/src/vs/sessions/contrib/terminal/browser/sessionsTerminalContribution.ts
@@ -152,11 +152,28 @@ export class SessionsTerminalContribution extends Disposable implements IWorkben
 			}
 		}));
 
-		// When a session is archived or removed, close all terminals for its cwd
+		// Close terminals for archived/removed sessions, but only when no other
+		// live session still owns that cwd. Terminals are reused across sessions
+		// at the same cwd, so a plain cwd match would kill a terminal still in use
+		// (e.g. the committed session from `onDidReplaceSession`).
+		// TODO: Consider removing the logic for trying to "delete/clean-up" terminal.
+		// Or consider tag terminals by sessionId + refcount instead of guarding here.
+
 		this._register(this._sessionsManagementService.onDidChangeSessions(e => {
-			for (const session of [...e.removed, ...e.changed.filter(s => s.isArchived.get())]) {
+			const removedIds = new Set(e.removed.map(s => s.sessionId));
+			const liveCwdKeys = new Set<string>();
+			for (const session of this._sessionsManagementService.getSessions()) {
+				if (removedIds.has(session.sessionId) || session.isArchived.get()) {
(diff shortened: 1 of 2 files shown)


meganrogge:
@anthonykim1 what are the repro steps here? Does it always happen?

anthonykim1:
@meganrogge 
I couldn't get consistent repros but when I was able to once or twice it's when I had claude and copilot cli sessions on my agents app and I tried to launch the "first" terminal in the agents app. 

On the buggy scenario that terminal would close immediately. 

Seen @TylerLeonhardt repro and I believe he also had cli and claude sessions open in agents app and tried  to spawn first terminal. Edit: couldnt constant repro either. 

Like in the TODO, Long term we should have better association of terminal with session or remove closing terminal.

Yoyokrazy (approved):

src/vs/sessions/contrib/terminal/browser/sessionsTerminalContribution.ts
```diff
@@ -152,11 +154,32 @@ export class SessionsTerminalContribution extends Disposable implements IWorkben
...
+		// Close terminals for archived/removed sessions, but only when no other
+		// live session still owns that cwd. Terminals are reused across sessions
+		// at the same cwd, so a plain cwd match would kill a terminal still in use
+		// (e.g. the committed session from `onDidReplaceSession`).
+		// TODO: Consider removing the logic for trying to "delete/clean-up" terminal.
+		// Or consider tag terminals by sessionId + refcount instead of guarding here.
+
 		this._register(this._sessionsManagementService.onDidChangeSessions(e => {
-			for (const session of [...e.removed, ...e.changed.filter(s => s.isArchived.get())]) {
+			const archivedChanged = e.changed.filter(s => s.isArchived.get());
```
meganrogge, line 165: As we discussed, this seems like a good stop gap solution, but the initial implementation should have been better here - using IDs and a map to keep track of session terminals. `cwd` is not good enough as this bug proves.

Since Ben S is out, we should fix it.

anthonykim1: Created follow up issue: https://github.com/microsoft/vscode/issues/312909

PR #313011, opened 2026-04-28
Update repository and path for Razor repo

davidwengier:
We merged the Razor repo into Roslyn so the source for the cshtml tmLanguage needs to change.


alexr00 (approved):
Thank you for the PR!

aeschli (approved):

PR #312734, opened 2026-04-27
fix: restrict entitlements used for helper.app

deepak1556:
For https://github.com/microsoft/vscode/issues/312720

PR #313013, opened 2026-04-28
ci: start the full app to verify signature on macOS

deepak1556:
Followup to https://github.com/microsoft/vscode/pull/312734#issuecomment-4333334191

Use `--prof-append-timers` to start the full workbench that gives better conformance to all the required processes launching on the signed application.

Also removed the `export-default-configuration` which seems like a no-op today cc @roblourens


deepak1556:
verification https://monacotools.visualstudio.com/Monaco/_build/results?buildId=434664&view=results

dmitrivMS (approved):

PR #313050, opened 2026-04-28
[Old] Execution Subagent Async Behaviour

vikramnitin9:
The execution subagent doesn't work well with async terminals (see #308048). The system prompt asks it to refrain from calling `run_in_terminal` async, but it is still possible for a sync terminal call to become async once the timeout expires.

The old workaround was to give the subagent access to tools like `get_terminal_output` and `kill_terminal` so that it could monitor background terminals. Drawbacks:
1. The subagent call itself is not async, so the main agent remains blocked while waiting for the subagent to finish. Given that, it is undesirable to give the subagent the ability to run commands with unbounded time horizons.
2. The main agent can relinquish control of long-running commands and get a ping when they complete. The subagent cannot. 
3. The subagent is being developed as an extremely lightweight custom model. It struggles to handle an expanded toolset.

The proposed solution is:
1.  Automatically detect when a terminal call times out
2. Prompt the subagent to finish and return the `<final_answer>`
4. Append the command(s) that timed out along with its terminal ID, to the subagent's response that is sent back to the main agent. This way, the main agent can take further action using `get_terminal_output`, `kill_terminal`, etc.

Analysis of telemetry shows that only 10% of all subagent trajectories contain at least one terminal call that times out, so it is acceptable for the subagent to hand back control to the main agent when it encounters such a scenario.

Another minority scenario (~1.1% of all subagent trajectories, from telemetry) is when a command waits on user input. To account for such cases, the subagent's system prompt instructs it to re-run the command with a pipe `echo y | ....` or `yes | ...`. This circumvents the need for an additional tool `send_output_to_terminal`.

Tagging @roblourens and @meganrogge for review.


vikramnitin9:
Closed in favour of #313067

PR #313071, opened 2026-04-28
[release/1.118] On agent mode change, ignore stateful marker

### LINKED ISSUE #312554: After Plan mode, agent does not start implementation
Type: <b>Bug</b>

I asked GPT 5.4 xhigh to create a plan, after the plan is done and saved in session memory, i press Start implementation, Chat response "Implementation is blocked by the current planning-only mode. I cannot edit files or start code changes from this agent state." then i simply type in chat, i switched to agent mode, start imeplementation and the agent is starting to do the job, i lost a few premium requests because of this issue.

VS Code version: Code 1.117.0 (10c8e557c8b9f9ed0a87f61f1c9a44bde731c409, 2026-04-21T16:12:14-07:00)
Modes:

dileepyavan:
Backport of #312951.

Fixes #312554

## Problem
When a Responses API conversation switches modes (for example, Agent -> Plan), we intentionally treat the next request as a mode change so it does not reuse the previous `previous_response_id` / stateful marker.

That part is correct for the first request after the switch, but websocket follow-up requests after tool calls were still effectively being treated as "mode changed" for too long. In practice, that meant a follow-up request in the same turn could fail to anchor on the newly established plan-mode websocket chain and could be pulled back toward earlier implementation-mode context.

This is especially noticeable with websocket-enabled Responses API models, where the connection carries forward the latest response marker. After the first successful request in the new mode, follow-up requests in the same turn should continue from that new-mode chain rather than continually invalidating it.

Parallel tool calls do not need separate handling here: `modeChanged` is evaluated per follow-up model request, not per individual tool call. A round may contain multiple tool calls, but once that round establishes a fresh stateful marker, later follow-up requests in the same turn should use the new marker.

## Fix
Updated `DefaultIntentRequestHandler.didModeChangeSincePreviousRequest()` so that:

- the first Responses API request after a mode switch still returns `modeChanged = true`
- but once the current turn has produced a fresh Responses API stateful marker, later requests in that same turn return `modeChanged = false`

This keeps the original protection against reusing the old chain across a mode switch, while allowing websocket follow-up requests after tool calls to resume from the newly established mode-specific chain.

In other words:

- Agent -> Plan: first plan request does **not** reuse the old websocket marker
- after that request succeeds and establishes a fresh plan marker, later follow-up requests in the same turn **do** reuse the new plan marker
- Plan -> Agent (start implementation): the first implementation request is again treated as a fresh mode change

## Tests
Added websocket-focused Responses API coverage in `src/platform/endpoint/node/test/responsesApi.spec.ts` for:

- not reusing a websocket stateful marker when `modeChanged` is `true`
- reusing the newly established websocket marker on follow-up requests after switching into Plan mode
- treating Agent -> Plan and Plan -> Agent ("start implementation") websocket requests as separate mode changes

This keeps the regression coverage on the websocket path, which is the relevant path for the affected Responses API models.

### DIFF
--- a/extensions/copilot/src/extension/agents/vscode-node/test/planAgentProvider.spec.ts
+++ b/extensions/copilot/src/extension/agents/vscode-node/test/planAgentProvider.spec.ts
@@ -18,7 +18,7 @@ import { DisposableStore } from '../../../../util/vs/base/common/lifecycle';
 import { SyncDescriptor } from '../../../../util/vs/platform/instantiation/common/descriptors';
 import { IInstantiationService } from '../../../../util/vs/platform/instantiation/common/instantiation';
 import { createExtensionUnitTestingServices } from '../../../test/node/services';
-import { buildAgentMarkdown } from '../agentTypes';
+import { buildAgentMarkdown, DEFAULT_READ_TOOLS } from '../agentTypes';
 import { PlanAgentProvider } from '../planAgentProvider';
 
 suite('PlanAgentProvider', () => {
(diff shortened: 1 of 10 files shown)


chrmarti (approved):

dileepyavan:
Will prefer to get some baking time in insiders than merging to release as this issue is not consistently reproducable.

dileepyavan:
Going ahead with release merge as the issues have high upvote and the change is specific to mode change and is considered less risky.

PR #313075, opened 2026-04-28
Pick up latest md lang server

mjbvz:


mjbvz:
Resubmitting from main

PR #313078, opened 2026-04-28
chore(build): print df -hT around copilot simulate-ci to diagnose low disk warning

bryanchen-d:
Add df -hT probes immediately before and after the 
pm run simulate-ci step in the Copilot test pipeline so we can identify which mount/filesystem is filling up when the 1ES agent emits `Free disk space on / is lower than 5%; Currently used: 95+%` warnings (e.g. build [434570](https://dev.azure.com/monacotools/Monaco/_build/results?buildId=434570)).

## Why

The Copilot stage of the VS Code pipeline produces dozens of `Free disk space on /` warnings, concentrated around the simulation/extension tests. The warning alone doesn't tell us which mount is full (root SSD, `tmpfs`, overlay, Docker volume, etc.). `df -hT` adds the filesystem **Type** column, which is exactly the signal needed to pick the right remediation.

## Change

[build/azure-pipelines/copilot/test-steps.yml](build/azure-pipelines/copilot/test-steps.yml): two new steps wrapping `Run simulation tests`:

```yaml
- script: df -hT
  displayName: Disk usage (before simulate)
  condition: always()

- script: npm run simulate-ci
  ...

- script: df -hT
  displayName: Disk usage (after simulate)
  condition: always()
```

## Notes

- `condition: always()` so output is captured even if simulate-ci fails.
- Only the Copilot Linux test stage is touched; no behavior change to actual tests.
- Once we have the data we can decide on the real fix (e.g. cleanup step, larger ephemeral disk, redirect cache to a different mount).

> _Written by Copilot_

### DIFF
--- a/build/azure-pipelines/copilot/test-steps.yml
+++ b/build/azure-pipelines/copilot/test-steps.yml
@@ -33,7 +33,18 @@ steps:
     workingDirectory: $(Build.SourcesDirectory)/extensions/copilot
     displayName: Run vitest unit tests
 
-  - script: df -hT
+  - script: |
+      set +e
+      echo '=== df -hT ==='
+      df -hT
+      echo '=== top consumers on / (one level) ==='
+      sudo du -xh -d 1 / 2>/dev/null | sort -hr | head -20
+      echo '=== /var breakdown ==='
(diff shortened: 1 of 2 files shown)


lszomoru:
Closing the PR since we have gotten to the bottom of this.

Issue #303960, opened 2026-03-23
Copilot CLI: Display last used model info


DonJayamanne:
<img width="462" height="186" alt="Image" src="https://github.com/user-attachments/assets/d9202c75-b514-46ad-93ae-d04c410ff698" />

anthonykim1:
Verification steps: 

1. Open Copilot CLI from chat panel
2. Send some request and make sure they correctly show the model name and mulitplier.

pwang347:
@anthonykim1 on latest Insiders build I don't see this

<img src="https://github.com/user-attachments/assets/48d2b0d9-cb0f-4414-a587-6a186dec4e4d" alt="Image" width="167" height="124">

anthonykim1:
@pwang347 Does hovering over make them appear? they should behave the same as local.
Looks like its flaky. I'll investigate what's going on.

DonJayamanne:
I think i found a problem with the implmenetation
```
	if (currentResponseParts.length > 0) {
		turns.push(new ChatResponseTurn2(currentResponseParts, lastResponseDetails ? { details: lastResponseDetails } : {}, ''));
```

* Start a cli session with opus 4.6
* Send another message for the same session with opus 4.7
* Send another message for the same session with GPT 5.4
* Send another message for the same session with GPT 5.3

We should see 4 requests with 4.6, 4.7, 5.4 and 5.3
GPT 5.3 being the last selected model in model picker as well as the last request

@anthonykim1 you'll need to extract the model switch information from the events (session history)

DonJayamanne:
@anthonykim1 I'm assuming this has been completed and has a TPI

PR #313079, opened 2026-04-28
Get model and multiplier show up for Copilot CLI controller API route

anthonykim1:
Before PR:
We went from being able to show model + token info after: https://github.com/microsoft/vscode/pull/311769 to the model not showing up until user reloads.

Also not getting anything shown for auto mode before or after both https://github.com/microsoft/vscode/pull/311769 and this PR branch.
Maybe related to: https://github.com/microsoft/vscode/issues/303960#issuecomment-4331684076

Edit: Seems like related to controller API path. On latest insiders if you disable controller api setting the models show up without having to refresh.

Summary -> Live model footer comes from request handler's returned ChatResult.details.
* Non-controller route had logic around request flow to build/ return details but the controller route was effectively completing the CLI request and returning {}. So live UI had no details to render.
* Reason why it worked on reload: History rebuild could later recover details from persisted/session data, which is why it looked "fixed after reload" but did not show up on live.
* Controller route now reads model after `handleRequest` via getCopilotCLIModelDetails.
* Gating this behind setting: `github.copilot.chat.agent.modelDetails.enabled`

Thanks @justschen for helping with auto mode.
Side note: For some reason `auto` seems to always show sonnet 4.6


src/vs/workbench/api/browser/mainThreadChatSessions.ts
```diff
@@ -84,7 +84,7 @@ export class ObservableChatSession extends Disposable implements IChatSession {
 		progress: (progress: IChatProgress[]) => void,
 		history: any[],
 		token: CancellationToken
-	) => Promise<void>;
+	) => Promise<IChatAgentResult | void>;
```
anthonykim1: @justschen Doing this via https://github.com/microsoft/vscode/pull/313079/commits/d0332fc2475a70557294c6f63962220ead13a816 makes the ui appear without reload, but only for when controller api is turned off. I'm not really confident changing shape of this though.

anthonykim1: Edit: Controller api route working with: [325dadd](https://github.com/microsoft/vscode/pull/313079/commits/325dadd6518d584f2f2b0301993bda76181fccc7)

extensions/copilot/src/extension/chatSessions/common/chatSessionMetadataStore.ts
```diff
@@ -56,6 +56,14 @@ export interface RequestDetails {
 	/** Mode instructions for this request (excluding toolReferences). */
 	modeInstructions?: StoredModeInstructions;
 
+	/**
+	 * The concrete model id that produced the response for this request, as reported by the
+	 * SDK's `assistant.usage` event. Captured so that on session reload we can render the
+	 * correct model details (e.g. for `auto`, where the resolved model is not otherwise
+	 * recoverable from the persisted SDK event log).
+	 */
+	responseModelId?: string;
```
DonJayamanne, line 65: I think we can  get this from the events. hence do not need to store this here.

justschen: from the SDK, the usage actually doesn't survive on reloads for some reason.

that session event for `assistant.usage` is `ephemerel` which means when we build the session from events only, we don't get back the model id, but we get back `auto` or undefined, which means it never gets rendered.

DonJayamanne: Got it, than you for clarification
Long live our metadata store solution

justschen: haha yeah it's a little unfortunate - i don't think there is anything we can do with SDK to turn that to `emphemeral: false` instead? otherwise this is the best we can do.

DonJayamanne (dismissed):

justschen (approved):

PR #313080, opened 2026-04-28
trackIdleOnPrompt logging

### LINKED ISSUE #313074: Shell integration intermittently breaks mid-session in agent terminal commands
## Problem

Shell integration (SI) intermittently breaks mid-session when executing agent terminal commands, causing the rich execute strategy to hang indefinitely. This was the **root cause of 42% of eval failures** in run [25018433402](https://github.com/microsoft/vscode-copilot-evaluation/actions/runs/25018433402) (7 out of ~17 tasks that should have passed).

PR #312854 added the `executingFallbackScheduler` (30s timeout) which addresses the **symptom** (permanent hang), but the underlying cause — SI sequences stopping mid-session — remains.

## Impact

Comparing VS Code eval run `25018433402` against CLI run `24966065548` (same model, same tasks):

| Task | CLI Result | VS Code Result | VS Code Failure Mode |
|------|-----------|---------------|---------------------|
| build-cython-ext | ✅ Pass (4 min) | ❌ Fail | SI broke on command 7 (new terminal), hung 55 min |
| compile-compcert | ✅ Pass | ❌ Fail | SI broke on command 1 (first command), hung ~55 min |
| git-multibranch | ✅ Pass | ❌ Fail | SI broke on command 1, hung ~55 min |
| mailman | ✅ Pass | ❌ Fail | SI broke on command 1, hung ~55 min |
| make-doom-for-mips | ✅ Pass | ❌ Fail | SI broke on command 1, hung ~55 min |
| sqlite-db-truncate | ✅ Pass | ❌ Fail | SI broke on command 1, hung ~55 min |
| write-compressor | ✅ Pass | ❌ Fail | SI broke on command 10, hung ~55 min |

All 7 failures follow the **identical pattern**: `X_AGENT_STILL_RESPONDING` after 3600000ms timeout.

> **Note:** This eval ran BEFORE PR #312854 was merged. The `executingFallbackScheduler` now prevents the permanent hang, but the SI flaking still degrades to a fallback path (marker-based output, no exit code).

## The Hang Mechanism (pre-#312854)

1. Command starts → data events arrive → `initialFallbackScheduler` cancelled ✅
2. `633;C`/`633;D` shell integration sequence seen → state becomes `Executing` ✅  
3. In `Executing` state, `promptFallbackScheduler` is also cancelled ✅
4. **Shell integration breaks** → no `633;A` (prompt) sequence ever arrives ❌
5. Both fallbacks cancelled → **permanent hang** (no `executingFallbackScheduler` existed yet)

### build-cython-ext (most interesting — SI broke 3 times, recovered from first)

```
# Terminal 1: Commands 1-2 succeed with rich SI
20:50:01.584 [info] ToolTerminalCreator#_waitForShellIntegration: Rich SI available eventually
20:50:01.598 [info] RunInTerminalTool: Using `rich` execute strategy
20:50:02.132 [info] RunInTerminalTool: Finished `rich` execute strategy with exitCode `0` ✅

# Command 3: pip install -v . (44K chars output) — SI WARNING but recovers
20:50:16.294 [info] RunInTerminalTool: Using `rich` execute strategy for command `set -e\ncd /app/pyknotid\npython3 -m pip install -v .`
20:50:26.364 [warning] Shell integration failed to add capabilities within 10 seconds  ⚠️
20:50:34.947 [info] RunInTerminalTool: Finished `rich` execute strategy with exitCode `0` ✅ (recovered via fallback, 18s)

# Commands 4-5 succeed
20:50:57.941 [info] RunInTerminalTool: Finished `rich` execute strategy with exitCode `0` ✅
20:51:21.145 [info] RunInTerminalTool: Finished `rich` execute strategy with exitCode `0` ✅

# Command 6: Another pip install -v . — SI BREAKS, no recovery
20:51:55.454 [info] RunInTerminalTool: Using `rich` execute strategy for command `set -e\ncd /app/pyknotid\n...python3 -m pip install -v .`
20:52:05.527 [warning] Shell integration failed to add capabilities within 10 seconds  ⚠️
20:53:55.458 [info] RunInTerminalTool: shellIntegrationQuality=rich at banner decision time
# ❌ NO "Finished rich execute strategy" log — HUNG

# Command 7: Creates NEW terminal, SI breaks immediately
20:54:22.557 [info] ToolTerminalCreator#createTerminal: Waiting 5000ms for shell integration
20:54:22.652 [info] ToolTerminalCreator#_waitForShellIntegration: Rich SI available eventually  (95ms ✅)
20:54:22.663 [info] RunInTerminalTool: Using `rich` execute strategy
20:54:32.723 [warning] Shell integration failed to add capabilities within 10 seconds  ⚠️
# ❌ NO "Finished rich execute strategy" — HUNG FOREVER until 1h timeout
```

### compile-compcert (SI breaks on first command of first terminal)

```
21:06:18.662 [info] ToolTerminalCreator#_waitForShellIntegration: Rich SI available eventually
21:06:18.681 [info] RunInTerminalTool: Using `rich` execute strategy for command `set -e\ncommand -v ocaml\n...`
21:06:28.754 [warning] Shell integration failed to add capabilities within 10 seconds  ⚠️
# ❌ NO "Finished rich execute strategy" — first command hangs forever
```

### Consistent pattern across all 7 failures

Every failing task ends with the same signature — the last log line is always:
```
[warning] Shell integration failed to add capabilities within 10 seconds
```
And there is never a corresponding `Finished rich execute strategy` log.

## Root Cause Analysis

The `_ensureCapabilitiesOrAddFailureTelemetry()` warning ([shellIntegrationAddon.ts:452-462](https://github.com/microsoft/vscode/blob/main/src/vs/platform/terminal/common/xterm/shellIntegrationAddon.ts#L452-L462)) fires when neither `CommandDetection` nor `CwdDetection` capability is added within 10s of `activate()`. These are created lazily — only when the first `633;A/B/C/D` sequence is parsed.

### Key Paradox

`_waitForShellIntegration` reports "Rich SI available eventually" (requires `onSetRichCommandDetection` to fire, which requires `CommandDetection` to exist), but then the 10s capabilities timer says no `CommandDetection` exists. This suggests either:
- Multiple `ShellIntegrationAddon` instances per terminal (saw 3 warnings from 2 terminal creations in build-cython-ext)
- Or the capabilities are created on one addon but the timer fires on a different one

#### 1. Timing race on new terminals
`runCommand` dispatches within 10-20ms of SI being declared "available," but "available" only means property sequences arrived. The shell may not have run its first `PROMPT_COMMAND` yet. In `shellIntegration-bash.sh`, [`__vsc_command_output_start()`](https://github.com/microsoft/vscode/blob/main/src/vs/workbench/contrib/terminal/common/scripts/shellIntegration-bash.sh#L351-L357) **skips emitting `633;C`** when `__vsc_first_prompt` is unset:

```bash
__vsc_command_output_start() {
    if [[ -z "${__vsc_first_prompt-}" ]]; then
        builtin return  # ← C sequence suppressed!
    fi
    builtin printf '\e]633;E;%s;%s\a' "$(__vsc_escape_value "${__vsc_current_command}")" $__vsc_nonce
    builtin printf '\e]633;C\a'
}
```

#### 2. `set -e` + DEBUG trap interaction
Every agent command starts with `set -e`. Bash's interaction between `set -e` and the DEBUG trap (`trap '__vsc_preexec_only "$_"' DEBUG`) is [notoriously fragile](https://www.gnu.org/software/bash/manual/html_node/The-Set-Builtin.html). In some bash versions, `set -e` can cause the DEBUG trap to be removed after a non-zero exit code in the trap handler chain. If the trap is removed, `__vsc_preexec` never fires → no `633;C` emitted for subsequent commands.

#### 3. PROMPT_COMMAND corruption
Heavy commands like `pip install -v .` (44K+ chars of output) that modify the shell environment or spawn activation scripts could reset `PROMPT_COMMAND`, preventing `633;D` and `633;A` sequences from being emitted.

## Suggestions

1. **Add trace logging** in `_handleVSCodeSequence`/`_handleFinalTermSequence` to record which OSC 633 sequences are received and when — this would pinpoint whether sequences are not being emitted by bash vs not being parsed by xterm.js

2. **Investigate multiple addon instances** — the build-cython-ext logs show 3 "Shell integration failed" warnings from only 2 terminal creations, suggesting addon duplication

3. **Test `set -e` interaction with the DEBUG trap** across bash versions used in eval containers — this is the most likely root cause given the correlation with multi-line commands starting with `set -e`

4. **Consider making `__vsc_first_prompt` initialization more eager** — could it be set in the SI script initialization rather than waiting for the first `PROMPT_COMMAND` to run?

/cc @meganrogge

meganrogge:
Fixes #313074

## Problem

Eval run [25073061392](https://github.com/microsoft/vscode-copilot-evaluation/actions/runs/25073061392) (Ubuntu 24.04, bash 5.2) had **20 tasks** that hung indefinitely (1-hour eval timeout). The `trackIdleOnPrompt` state machine — which detects when a command finishes by monitoring shell integration sequences and data idle — is currently a **complete black box** in the logs. We can see commands start and finish, but nothing about internal state transitions, timer fires, or why the state machine failed to resolve.

### Observed behavior

After certain commands (often containing `set -e`, but not exclusively), shell integration breaks mid-session. Subsequent commands return `exitCode=undefined, result.length=0` and resolve only via fallback timers (~11s). Eventually a command hangs forever because no fallback fires.

**Log evidence** — `winning-avg-corewars` task: the first two `set -e` commands complete normally (`exitCode=0`), the third breaks SI, and the final command hangs indefinitely:
```
19:42:01.271 Finished with exitCode `0`      ← set -e command #1, SI working
19:43:17.339 Finished with exitCode `0`      ← set -e command #2, SI working  
19:43:46.945 Shell integration failed to add capabilities within 10 seconds
19:44:07.896 Finished with exitCode `undefined`, result.length `1100`  ← SI broken
...
19:45:28.819 Using `rich` execute strategy    ← final command
                                               ← NO "Finished" line. Hangs forever.
```

### What we don't know

The April 6 eval (run 24049820070) had **38 `set -e` tasks with zero hangs** on an older build. The April 28 eval (run 25073061392) had **26 `set -e` tasks with 20 hangs**. The `set -e` correlation is strong but doesn't fully explain the regression — something else changed between these builds. Without visibility into `trackIdleOnPrompt`'s internal state, we can't determine the exact failure mode.

### 1. `trackIdleOnPrompt` diagnostic logging (`executeStrategy.ts`)

Add INFO-level logging to the state machine that's currently invisible in logs:
- **State transitions**: `Initial -> Executing (sequence C)`, `Executing -> PromptAfterExecuting (sequence A)`
- **Timer fires**: `Initial fallback fired, no data events received`, `Executing fallback fired after 30s data-idle`
- **Data event counts**: `dataEvents=N` on each timer fire

This will make the next eval run fully diagnosable — we'll see exactly where and why commands hang.

### 2. Bash SI `set -e` protection (`shellIntegration-bash.sh`)

Wrap all 4 SI handler functions with `set +e` / restore + `return 0`:
- `__vsc_prompt_cmd` — main PROMPT_COMMAND handler
- `__vsc_prompt_cmd_original` — wrapper for user's original PROMPT_COMMAND
- `__vsc_preexec_only` — DEBUG trap preexec (emits `633;C`)
- `__vsc_preexec_all` — DEBUG trap combined handler

On all bash versions, the DEBUG trap is NOT protected from `set -e` — if any statement returns non-zero, the handler aborts before emitting `633;C` (command start). On macOS bash 3.2, PROMPT_COMMAND is also unprotected. This fix hardens both paths. While `set -e` may not be the sole root cause (see above), this is good defense-in-depth.

### 3. Tool description update (`runInTerminalTool.ts`)

Instruct the model to pair `set -e` with `set +e` at the end of bash commands, reducing the chance of `set -e` persisting across commands.

### 4. Trace log placement fix (`shellIntegrationAddon.ts`)

Move trace logs from `_handleFinalTermSequence`/`_handleVSCodeSequence` into `_doHandle*` methods per review feedback.


meganrogge:

alexr00 (approved):

PR #313089, opened 2026-04-28
Fix generic local Agent Host skill labels

### LINKED ISSUE #313069: All local AH skills are named "SKILL"
https://github.com/user-attachments/assets/7423648c-389d-4012-931e-dbdd3be60949

cipheraxat:
Local Agent Host skill invocation events can report a generic name such as `SKILL` while also including the resolved path to the loaded `.../<skill-name>/SKILL.md` file. Because every skill entrypoint uses the same filename, rendering the reported name directly makes every local skill appear as `SKILL` in chat progress.

Normalize that display name at the skill event synthesis boundary. If the reported name is `SKILL` or `SKILL.md` and the resolved path points to a `SKILL.md` file, derive the rendered label from the parent skill folder. Events that already provide a meaningful skill name continue to use it unchanged.

Add regression coverage for both live skill event synthesis and history replay mapping so a payload like `{ name: 'SKILL', path: '/repo/skills/plan/SKILL.md' }` renders as `plan` instead of `SKILL`.

Fixes #313069

Tests:

- ./scripts/test.sh --run src/vs/platform/agentHost/test/node/copilotToolDisplay.test.ts --run src/vs/platform/agentHost/test/node/mapSessionEvents.test.ts --grep "skill events"

- node --experimental-strip-types build/hygiene.ts src/vs/platform/agentHost/node/copilot/copilotToolDisplay.ts src/vs/platform/agentHost/test/node/copilotToolDisplay.test.ts src/vs/platform/agentHost/test/node/mapSessionEvents.test.ts

- npm run valid-layers-check

### DIFF
--- a/src/vs/platform/agentHost/node/copilot/copilotToolDisplay.ts
+++ b/src/vs/platform/agentHost/node/copilot/copilotToolDisplay.ts
@@ -12,7 +12,7 @@ import { localize } from '../../../../nls.js';
 import type { IAgentToolCompleteEvent, IAgentToolReadyEvent, IAgentToolStartEvent } from '../../common/agentService.js';
 import { stripRedundantCdPrefix } from '../../common/commandLineHelpers.js';
 import { StringOrMarkdown } from '../../common/state/protocol/state.js';
-import { basename } from '../../../../base/common/resources.js';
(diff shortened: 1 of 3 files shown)


roblourens:
Sorry, already fixed that issue

Issue #313084, opened 2026-04-28
Update DOMPurify to v3.4.1 because current v3.2.7 has several CVEs


kozi:
Does this issue occur when all extensions are disabled?: Yes

## Update DOMPurify to v3.4.1 because current v3.2.7 has several CVEs

| CVE | Severity / CVSS | Affected DOMPurify versions | Fixed in | Links |
|---|---|---|---:|---|
| **CVE-2026-41240** | **Medium** — CVSS v4.0: `6.0` | `< 3.4.0` | `3.4.0` | OpenCVE: https://app.opencve.io/cve/CVE-2026-41240 · GitHub Advisory: https://github.com/cure53/DOMPurify/security/advisories/GHSA-h7mw-gpvr-xq4m |
| **CVE-2026-41239** | **Medium** — CVSS v3.1: `6.8` | `>= 1.0.10, < 3.4.0` | `3.4.0` | OpenCVE: https://app.opencve.io/cve/CVE-2026-41239 · GitHub Advisory: https://github.com/advisories/GHSA-crv5-9vww-q3g8 |
| **CVE-2026-41238** | **Medium / Moderate** — CVSS v3.1: `6.9` | `>= 3.0.1, < 3.4.0` | `3.4.0` | OpenCVE: https://app.opencve.io/cve/CVE-2026-41238 · GitHub Advisory: https://github.com/advisories/GHSA-v9jr-rg53-9pgp · NVD: https://nvd.nist.gov/vuln/detail/CVE-2026-41238 |
| **CVE-2026-0540** | **Medium** — CVSS v4.0: `5.3`; CVSS v3.1: `6.1` | `>= 3.1.3, <= 3.3.1` and `>= 2.5.3, <= 2.5.8` | `3.3.2` / `2.5.9` | OpenCVE: https://app.opencve.io/cve/CVE-2026-0540 · GitHub Advisory: https://github.com/advisories/GHSA-v2wj-7wpq-c8vv · NVD: https://nvd.nist.gov/vuln/detail/CVE-2026-0540 |

https://github.com/microsoft/vscode/tree/main/src/vs/base/browser/dompurify

Maybe also here?

`extensions/copilot/package.json`
https://github.com/microsoft/vscode/tree/main/extensions/copilot/package.json

`extensions/markdown-language-features/package.json`
https://github.com/microsoft/vscode/tree/main/extensions/markdown-language-features/package.json

`extensions/mermaid-chat-features/package.json`
[...]

PR #313095, opened 2026-04-28
Bump dom purify in extensions

mjbvz:
For #313084


mjbvz:
whoops missed switching branches. Will restore

mjbvz:
But also need to submit from origin :/

PR #307199, opened 2026-04-01
Add completion notifications and async prompt monitoring for background terminals

meganrogge:
## Summary

Background terminals (async mode) are no longer "fire and forget" — the agent is now automatically notified when commands complete and the terminal continues to detect prompts-for-input on the next turn.

## Before

- When a terminal moved to background (async mode, timeout, or continue-in-background), the **output monitor was killed** and the agent received no further information
- The agent had to **manually poll** with `get_terminal_output` to check if a background command finished
- **Prompts-for-input** (e.g., password prompts, "are you sure?" confirmations) in background terminals were **silently missed**
- Model descriptions directed the agent to use `get_terminal_output` for polling, but this left the agent guessing when to check that and lead to it calling `sleep` and `echo` inappropriately

## After

- **Completion notifications**: When a command finishes in a background terminal, a steering message is sent to the agent on its next turn with the exit code and terminal output
- **Async prompt-for-input monitoring**: The output monitor continues running in async mode after the tool returns, detecting prompts via regex based analysis (skipping LLM analysis for reduced cost) and showing elicitation UI to the user 
- **All background terminals covered**: Both terminals that start as async and foreground terminals that transition to background (timeout/continue-in-background) get notifications
- **Model descriptions updated**: Agent is told it will be notified automatically, with `get_terminal_output` available for on-demand checks before the next turn

Adds `isSystemInitiated` to `IChatSendRequestOptions`, enabling any consumer to send steering messages that are not rendered, but sent to the agent. 

Fixes https://github.com/microsoft/vscode/issues/307201
Fixes https://github.com/microsoft/vscode/issues/303931

PR #313106, opened 2026-04-28
terminalChat: fix OutputMonitor dispose race that stalls run_in_terminal

meganrogge:
## Summary

Fixes a race in `OutputMonitor` where disposing the monitor while its async `_startMonitoring` loop is still in flight causes `onDidFinishCommand` to never resolve for consumers. The agent then hangs on the `run_in_terminal` tool call until the chat driver's 1-hour timeout fires.

## Root cause

[`OutputMonitor`](../blob/main/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/monitoring/outputMonitor.ts) starts an async polling loop in `_startMonitoring` and fires `_onDidFinishCommand` from a `finally` block. The constructor registers a disposable that cancels the inner CTS so the loop exits promptly, but it does **not** fire the event synchronously.

When `dispose()` is called mid-flight:

1. `super.dispose()` tears down the `_onDidFinishCommand` `Emitter` and clears its listeners.
2. The async loop later reaches its `finally` block and calls `.fire()` on the dead emitter — a no-op.
3. The consumer in [`runInTerminalTool.ts`](../blob/main/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/runInTerminalTool.ts) is awaiting `Event.toPromise(outputMonitor.onDidFinishCommand)`. Its listener was already cleared, so the promise never resolves.
4. The tool invocation strands. `executeChatCommandAndWait` eventually times out at the driver's hard cap.

The pattern was introduced by the `timeout(0).then(_startMonitoring).finally(fire)` shape in [c48248dd6ba](https://github.com/microsoft/vscode/commit/c48248dd6ba) and made externally reachable when [PR #307199](https://github.com/microsoft/vscode/pull/307199) wired `OutputMonitor` to chat session lifecycle disposal.

## Fix

1. Add a `_didFinish` flag and a private `_fireFinishedOnce()` so the event fires at most once.
2. Fire `onDidFinishCommand` **synchronously from the disposable**, before `super.dispose()` runs, so consumers awaiting `Event.toPromise(...)` are released while the emitter is still live.
3. On the dispose path, synthesize a `Cancelled` `IPollingResult` so consumers that read `monitor.pollingResult` after the event always see a defined value.
4. Replace the bare `.fire()` in the async finally block with `_fireFinishedOnce()` so the late path is a no-op when dispose already won.

## Validation against eval failures

Verified against [`.vscode-eval/25073061392`](https://github.com/microsoft/vscode-copilot-evaluation/) — a `terminalbench2` run with **64 raw `X_AGENT_STILL_RESPONDING` failures (32 unique tests)**, every one timing out at the driver's 1h cap (`Timeout of 3600000ms exceeded while waiting for chat command to complete`).

**15 of the 32** unique failures show the exact signature of this bug: in `terminal.log`, the last `RunInTerminalTool: Using \`<strategy>\` execute strategy ...` line has **no matching `Finished \`<strategy>\` execute strategy`** — i.e., `OutputMonitor.onDidFinishCommand` never resolved for that command and the agent fell silent for the rest of the hour.

Concrete example from `terminalbench2.eval.x86_64.winning-avg-corewars-output`:

```
2026-04-28 19:45:18.154 [info] RunInTerminalTool: Finished `rich` execute strategy ... (prior command, OK)
2026-04-28 19:45:28.819 [info] RunInTerminalTool: Using `rich` execute strategy for command ` set -e ...`
<no further log lines for 54 minutes>
2026-04-28 20:39:31    chat driver fires 1h timeout → X_AGENT_STILL_RESPONDING
```

Matching tests:

| Test | last "Using strategy" timestamp |
|---|---|
| adaptive-rejection-sampler | 19:35:24.971 |
| caffe-cifar-10 | 19:32:06.594 |
| circuit-fibsqrt | 19:32:53.282 |
| compile-compcert | 19:32:54.511 |
| dna-insert | 19:32:10.368 |
| fix-ocaml-gc | 19:31:39.738 |
| install-windows-3.11 | 15:31:46.320 |
| mailman | 19:34:14.363 |
| make-doom-for-mips | 19:31:03.722 |
| merge-diff-arc-agi-task | 19:31:06.965 |
| nginx-request-logging | 19:33:17.061 |
| qemu-alpine-ssh | 19:38:26.491 |
| qemu-startup | 19:33:21.743 |
| rstan-to-pystan | 19:41:06.952 |
| winning-avg-corewars | 19:45:28.819 |

The remaining 17 timeouts have matched `Using`/`Finished` counts, indicating a different stall mode (e.g. auth/model-fetch failures observed in `video-processing`) and are out of scope for this PR.

## Test

Added a regression test in `outputMonitor.test.ts` that disposes the monitor while `_startMonitoring` is in flight and asserts that:

- `Event.toPromise(monitor.onDidFinishCommand)` resolves
- `monitor.pollingResult` is defined and has state `Cancelled`

## Risk

Low. Behavior change is scoped to the dispose path:

- Before: emitter dies before `.fire()` runs; consumers may strand.
- After: emitter fires once synchronously on dispose; late `.fire()` from the async finally is a no-op via `_didFinish`.

`pollingResult` was previously left as `undefined` on a mid-flight dispose; now it's a `Cancelled` snapshot. Any consumer that did a truthy/optional check on `pollingResult` is unaffected; the new value carries the most accurate state.

### DIFF
--- a/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/monitoring/outputMonitor.ts
+++ b/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/monitoring/outputMonitor.ts
@@ -102,7 +102,6 @@ export class OutputMonitor extends Disposable implements IOutputMonitor {
 	private _command = '';
 	private _invocationContext: IToolInvocationContext | undefined;
 	private _currentMonitoringCts: CancellationTokenSource | undefined;
-
 	/**
 	 * Tracks whether onDidFinishCommand has fired so the event is delivered at
 	 * most once. The event must fire synchronously during dispose so consumers
@@ -117,6 +116,30 @@ export class OutputMonitor extends Disposable implements IOutputMonitor {
 		}
 		this._didFinish = true;
 		this._onDidFinishCommand.fire();
+	}
+
+	override dispose(): void {
+		// Deliver onDidFinishCommand to consumers BEFORE super.dispose() tears
+		// down the emitter. Field-initialized disposables (including
+		// _onDidFinishCommand) are registered before any disposable added in
+		// the constructor body and are disposed first by DisposableStore in
+		// insertion order. Without this override, consumers awaiting
+		// `Event.toPromise(onDidFinishCommand)` would race with emitter
+		// teardown and hang when dispose lands while _startMonitoring is still
+		// in flight.
+		if (!this._didFinish) {
+			// Synthesize a Cancelled pollingResult so consumers that read
+			// `monitor.pollingResult` after awaiting onDidFinishCommand always
+			// see a defined value with the output collected so far.
+			this._pollingResult ??= {
+				state: OutputMonitorState.Cancelled,
+				output: this._execution.getOutput(),
+				pollDurationMs: 0,
+				resources: undefined,
+			};
+		}
+		this._fireFinishedOnce();
+		super.dispose();
 	}
 
 	constructor(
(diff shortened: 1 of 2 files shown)


meganrogge (commented):
Good catch — Copilot is right. `DisposableStore` iterates registered disposables in insertion order, and `_onDidFinishCommand` (field initializer) is registered before the constructor-body `toDisposable`, so the emitter was being torn down before my synchronous fire could run.

Fixed in [33ad5ad](https://github.com/microsoft/vscode/pull/313106/commits/33ad5ad64e0): override `dispose()` so the synthetic `Cancelled` `pollingResult` is set and `onDidFinishCommand` is fired before `super.dispose()` runs. The `_didFinish` guard still keeps the late `.fire()` from `_startMonitoring`'s finally block a no-op.

roblourens (approved):

PR #313243, opened 2026-04-29
Align Cloud Agent closer to Mission Control's existing behavior

A-Georgiou:
Port of microsoft/vscode-copilot-chat#5078 (by @Oregand) to vscode core.

VS Code's Cloud Agent flow was still doing pre-PR work that Mission Control / CCA PR Decoupling no longer needs:

- Drop the delegate/commit/push modal. New-request path goes straight to delegate(). Permissive GitHub auth kicks in via the standard non-modal `createIfNone` sign-in prompt.
- Send the raw user prompt. Stop calling `IChatDelegationSummaryService.summarize` on the delegation path. problem_statement is now the user prompt + resolved references only.
- Match MC's minimal pull_request payload. Drop `body_placeholder` and `body_suffix`, keep `title`, `base_ref`, and optional `head_ref`.
- Drop the pre-delegation git preflight. Removed `detectedUncommittedChanges`, `checkBaseBranchPresentOnRemote`, `buildConfirmation`, and `handleConfirmationData`.
- Delete dead code. `CopilotCloudGitOperationsManager` is now unreachable - removed the whole file.


A-Georgiou:

PR #313244, opened 2026-04-29
Detect chevron-style interactive prompts and fix indefinite foreground hang on unrecognised prompts

### LINKED ISSUE #312576: Terminal stuck forever on sync command
It ran a command with "sync" and no timeout, and that command needs input and is blocking the agent loop. What's supposed to happen- I think we don't do input detection anymore right? Should the command timeout at some point?

[ca331029.copilotmd](https://github.com/user-attachments/files/27095321/ca331029.copilotmd)

Image

meganrogge:
Fixes #312576

## Problem
When the agent runs a sync (foreground) command via `run_in_terminal` and the command launches an interactive prompt from a library like [`prompts`](https://www.npmjs.com/package/prompts) (used by vitest), [`enquirer`](https://www.npmjs.com/package/enquirer), or [`inquirer`](https://www.npmjs.com/package/inquirer), the agent does not detect that input is required and the call hangs **indefinitely**.

### Why it hangs forever
The foreground `run_in_terminal` path races four candidates: process exit, continue-in-background, user-specified timeout (default: none), and `onDidDetectInputNeeded`. When the process is alive waiting for input:

1. The `OutputMonitor` polling loop sees `isActive === true` and only transitions to Idle if `detectsInputRequiredPattern` matches the cursor line.
2. Without a matching pattern, the monitor keeps polling until its internal 2-minute extended timeout, then sets state to `Cancelled`.
3. **But `Cancelled` does not resolve any of the foreground race candidates.** The process is still alive, no user timeout was specified, and `onDidDetectInputNeeded` was never fired — so the race hangs forever.

### 1. Chevron prompt pattern
Add a new pattern in `detectsHighConfidenceInputPattern` (the fast-path checked on every poll tick) that matches lines starting with `?` (after optional whitespace/ANSI escapes) and ending with a prompt-library chevron glyph (`›` U+203A, `❯` U+276F, `▸` U+25B8, `▶` U+25B6). Anchoring `?` to the start of the line matches the canonical `prompts`/`enquirer`/`inquirer` rendering and avoids false positives from incidental `?` + chevron combinations in normal output (e.g. `What happened? ›`).

### 2. Extended timeout safety net
When the 2-minute extended polling timeout fires, fire `onDidDetectInputNeeded` before cancelling. This is the critical fix for the **indefinite hang** — it resolves the foreground race so the agent receives the terminal output and can assess/respond.

### 3. Soften notification language
Changed the background steering message from "command **is** waiting for input" to "command **may be** waiting for input", and the UI label from "needs input" to "may need input". This is important because the event now also fires on extended timeout where the process might just be slow rather than actually waiting for input. The steering text already instructs the agent to assess the output before acting.

### Decision matrix: extended timeout behavior

| Scenario | `isActive` | Without firing event | With firing event |
|---|---|---|---|
| **Unrecognised prompt (fg)** | `true` | ❌ Hangs forever — no race candidate resolves | ✅ Agent gets output, can assess and respond |
| **Unrecognised prompt (bg)** | `true` | ❌ Silently cancelled, agent never sees it | ✅ Steering message sent, agent assesses output |
| **Long-running command (fg)** — e.g. 3min build | `true` | ⚠️ Also hangs forever (same bug, different cause) | ✅ Agent told "may need input" — steering text says "call GetTerminalOutput to continue polling", so agent polls and build finishes normally |
| **Long-running command (bg)** | `true` | Silent cancel at 2min | ⚠️ Benign steering message — agent sees output, assesses "still running", continues polling |
| **Already exited process** | `false` | N/A — caught before timeout | Same |

## Tests
- **Chevron pattern tests**: each chevron variant (`›`, `❯`, `▸`, `▶`) preceded by `?` with and without trailing whitespace; ANSI escape-prefixed prompt lines; negatives for already-typed responses, chevrons without leading `?`, and mid-line `?` with chevron
- **Extended timeout test**: verifies `onDidDetectInputNeeded` fires when extended timeout expires with an active process

### DIFF
--- a/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/monitoring/outputMonitor.ts
+++ b/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/monitoring/outputMonitor.ts
@@ -605,15 +605,17 @@ export function detectsHighConfidenceInputPattern(cursorLine: string): boolean {
 		// "Press a key" or "Press any key"
 		/press a(?:ny)? key/i,
 		// Interactive prompt libraries (prompts, enquirer, inquirer) prefix the prompt with
-		// '? ' and end the line with a distinctive chevron character followed by optional
-		// trailing whitespace where the cursor is awaiting input. Requiring a '?' earlier
-		// on the line avoids false positives from random output that happens to contain a
-		// chevron (e.g. git log decorations).
+		// '? ' at the start of the line and end with a distinctive chevron character
+		// followed by optional trailing whitespace where the cursor is awaiting input.
+		// Anchoring the '?' to the start of the line (after optional whitespace/ANSI
+		// escapes) avoids false positives from normal output that contains both a '?'
+		// allow-any-unicode-next-line
+		// and a chevron (e.g. "What happened? ›").
 		// Examples:
 		//   "? Do you want to install jsdom? <chevron>"  (prompts)
-		//   "? Pick a color <chevron> "                  (inquirer / enquirer)
+		//   "? Pick a color <chevron> "                  (enquirer)
 		// allow-any-unicode-next-line
-		/\?.*[›❯▸▶]\s*$/,
+		/^(?:\s|\x1b\[[0-9;]*m)*\?.*[›❯▸▶]\s*$/,
 	].some(e => e.test(cursorLine));
 }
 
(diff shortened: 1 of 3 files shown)


meganrogge:

pwang347 (approved):

PR #313266, opened 2026-04-29
agentHost: drop the agent event abstraction

connor4312:
We had a separate representation of agent events in the agent host that were usually mostly just the same as the protocol events/state. And we had steps to map to and from in various places which added to code and implementation complexity. Remove it.


src/vs/platform/agentHost/test/node/mockAgent.ts
```diff
@@ -708,11 +754,438 @@ export class ScriptedMockAgent implements IAgent {
...
-			setTimeout(() => this._onDidSessionProgress.fire(event), delay);
+			setTimeout(() => this._fireLegacy(session, event), delay);
 		}
 	}
+
+	/** Per-session translator state for {@link _fireLegacy}. Tracks the
+	 *  active markdown / reasoning response part ids so consecutive `delta`
+	 *  and `reasoning` events coalesce into append actions, mirroring the
+	 *  live emission rules of {@link CopilotAgentSession}. */
+	private readonly _legacyState = new Map<string, ILegacySignalState>();
```
connor4312, line 769: I'm going to clean up the mockagent 'legacy' stuff in a separate PR

roblourens (approved):
Nice

PR #313312, opened 2026-04-29
Recover from terminal hangs when shell integration breaks or the pty exits

### LINKED ISSUE #313296: Eval findings - `set -e` hangs benchmarks
Here's the table of the exact commands that hung the agent until the 1 h outer timeout:

| # | Run | Hung command (paraphrased) | Likely cause |
|---|---|---|---|
| 1 | build-cython-ext | `set -e; cd /app/pyknotid; python3 - <<'PY' … alexander_polynomial(-1) … PY` | Heredoc into python; `Shell integration failed to add capabilities` after start. SI broken mid-command. |
| 2 | build-pmars | `set -e; cd …; printf '--- debian/rules ---'; sed -n '1,220p' debian/rules; printf …; sed -n '1,220p' debian/control; printf …; sed -n '1,260p' Makefile` | Multi-step `printf`/`sed` chain — large stdout; SI fail recorded. |
| 3 | caffe-cifar-10 | `set -e; cd /app; rm -rf caffe; git clone --branch 1.0.0 --depth 1 https://github.com/BVLC/caffe.git /app/caffe; cd /app/caffe; git rev-parse --short HEAD; git describe --tags --exact-match` | Long network clone; SI fail; no per-command timeout. |
| 4 | compile-compcert | `set -e; cd /tmp/CompCert; ./configure x86_64-linux; make -j$(nproc)` | Long-running build; SI fail. |
| 5 | filter-js-from-html | `set -e; cd /app; python3 -m py_compile filter.py; sample1=$(mktemp); cat <<'EOF' > "$sample1" … EOF; python3 filter.py "$sample1"; … (3 sample heredocs + python invocations, then printf cat)` | Multiple heredocs (`<<'EOF'`) inside one shell; SI fail. |
| 6 | fix-git | `set -e; cd /app/personal-site; git cherry-pick 650dba4; git --no-pager log --oneline --decorate -n 3` | `cherry-pick` opened editor (no `--no-edit`) and waited for input. Covered by #313244. |
| 7 | fix-ocaml-gc | `set -e; cd /app/ocaml; make -j4` | Long `make`; SI failed 10s after start; rich strategy never saw `onCommandFinished`. Covered by SI-fallback chain. |
| 8 | install-windows-3.11 | `set -e; chmod +x /app/bin/start-win311.sh …; bash -n …; /app/bin/start-win311.sh` | `start-win311.sh` likely launches QEMU/Win 3.11 in foreground — process never returns. SI fail. |
| 9 | mailman | `set -e; python3 - <<'PY' from mailman.config import postfix; import inspect; print(inspect.getsource(postfix)) PY` | Heredoc into python; **no SI fail logged** — purely `onCommandFinished` never firing for heredoc. |
| 10 | merge-diff-arc-agi-task | `set -e; cd /app/repo; printf '--- branch1 algo.py ---'; git show branch1:algo.py 2>/dev/null \|\| true; printf …; git show branch2:algo.py … ; printf …; git show --stat --oneline --summary branch1 \| head -n 20; … branch2 …` | Multi-`printf`/`git show` with `\|\| true` and pipelines; SI fail. |
| 11 | protein-assembly | `set -e; python - <<'PY' (large bioinformatics script with codon tables, Plasmid file munging, …) PY` | Largest heredoc (≥ 70 lines, multi-second run); SI fail. |

Patterns at a glance:
- **6/11 are heredocs or compound shell scripts** with embedded `<<'PY'`/`<<'EOF'` blocks. SI seems particularly fragile here.
- **3/11 are long-running builds/clones/installs** (`make`, `./configure && make`, `git clone`, `start-win311.sh`).
- **1/11 (fix-git)** is a `cherry-pick` waiting on an editor — the only true "interactive prompt" case.
- **1/11 (build-pmars)** is purely large `printf`/`sed` output.
- **10/11** logged `Shell integration failed to add capabilities within 10 seconds` after the hung command started, which is the exact trigger for the SI-fallback chain (`53ab18ec…` → `2df99ce…`) you've already merged — those should rescue the bulk of these on the next eval build.

meganrogge:
Fixes https://github.com/microsoft/vscode/issues/313296

Two related fixes uncovered while diagnosing eval run [`25120748970`](https://github.com/microsoft/vscode-copilot-evaluation/actions/runs/25120748970), where 11 of 47 benchmarks hit the 1-hour outer timeout because the `RunInTerminalTool` flow never produced a result — even though the underlying terminals were alive and reachable.

### 1. Hard-cap fallback in `trackIdleOnPrompt`

10 of the 11 hung runs logged:

> `Shell integration failed to add capabilities within 10 seconds`

Inspecting their full logs showed:
- No `Finished rich` line ever logged for the hung command.
- Zero OSC `C`/`D` sequences ever parsed (no `trackIdleOnPrompt` state transitions logged).
- The rich strategy's `Promise.race` never resolved by *any* arm — `onCommandFinished` (SI never engaged), `onDisposed`/`onExit` (pty alive), `onCancellationRequested` (not cancelled), and `trackIdleOnPrompt` (state stuck in `Initial`).

The existing fallbacks in `trackIdleOnPrompt` cover *partial* SI failure:
- `initialFallbackScheduler` (10 s, fires only if no data has arrived).
- `executingFallbackScheduler` from #312854 (30 s data-idle, gated on `state === Executing` — i.e. at least one OSC `C`/`D` must have parsed).

Neither covers the case where data *does* arrive but no OSC sequence is ever parsed (broken handshake from the start of the command). Both schedulers get cancelled by `onData` and the function blocks for the rest of the chat request.

This adds a 1-minute wall-clock `hardCapScheduler` that fires only if state is still `Initial` or `Prompt` — i.e. SI never engaged. Working SI completes via `onCommandFinished` long before; long-running commands with working SI advance past `Prompt` and the cap is a no-op for them.

### 2. Notify chat session when a background terminal disposes

In `runInTerminalTool.ts#_registerCompletionNotification`, the `onDisposed` handler previously just tore down listeners. If a backgrounded terminal's pty exits before `commandDetection.onCommandFinished` fires (process crash, user closes the terminal, shell exits), the agent is left silently waiting for a notification that will never arrive — exactly what happened in `build-pov-ray`.

Now the handler captures whatever output was buffered, reads `exitCode`, and sends a `Steering`-queued steering message using the same `userSelectedModelId`/`modeInfo`/`userSelectedTools` as the regular completion path, with a new `terminalProcessExited` localize label.

### Risk

Both changes are no-ops on every working path:
- Hard cap: only fires if `state` never advanced past `Prompt` after 5 min.
- Disposal notification: only fires if `onDisposed` reaches the handler before `_backgroundNotifications` was already cleared.

No user-visible UI changes; the only effect on agent transcripts is an extra steering message in the previously-silent failure modes.

### DIFF
--- a/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/executeStrategy/executeStrategy.ts
+++ b/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/executeStrategy/executeStrategy.ts
@@ -254,11 +254,11 @@ export async function trackIdleOnPrompt(
 	// fires, so this is purely a fallback for the broken-handshake path.
 	const hardCapScheduler = store.add(new RunOnceScheduler(() => {
 		if (state === TerminalState.Initial || state === TerminalState.Prompt) {
-			log?.(`Hard cap fired after 5 minutes in state ${stateNames[state]} (dataEvents=${dataEventCount})`);
+			log?.(`Hard cap fired after 60s in state ${stateNames[state]} (dataEvents=${dataEventCount})`);
 			setState(TerminalState.PromptAfterExecuting, 'hardCap');
 			scheduler.schedule();
 		}
-	}, 5 * 60_000));
+	}, 60_000));
 	hardCapScheduler.schedule();
 	// Only schedule when a prompt sequence (A) is seen after an execute sequence (C). This prevents
 	// cases where the command is executed before the prompt is written. While not perfect, sitting
--- a/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/executeStrategy/executeStrategy.ts
+++ b/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/executeStrategy/executeStrategy.ts
@@ -243,6 +243,23 @@ export async function trackIdleOnPrompt(
 			scheduler.schedule();
 		}
 	}, 30_000));
+	// Hard wall-clock safety net for the case where shell integration never
+	// engages at all — no OSC `C`/`D` is ever parsed so state never advances
+	// to Executing, and yet the existing data-idle fallbacks somehow fail to
+	// fire (e.g. because data arrives in a single event before listeners are
+	// active, or because `onData` throttling masks idle periods). Without
+	// this, trackIdleOnPrompt can block the rich strategy for the lifetime
+	// of the chat request. Genuinely long-running commands with working
+	// shell integration reach Executing/PromptAfterExecuting before this
+	// fires, so this is purely a fallback for the broken-handshake path.
+	const hardCapScheduler = store.add(new RunOnceScheduler(() => {
+		if (state === TerminalState.Initial || state === TerminalState.Prompt) {
+			log?.(`Hard cap fired after 5 minutes in state ${stateNames[state]} (dataEvents=${dataEventCount})`);
+			setState(TerminalState.PromptAfterExecuting, 'hardCap');
+			scheduler.schedule();
+		}
+	}, 5 * 60_000));
+	hardCapScheduler.schedule();
 	// Only schedule when a prompt sequence (A) is seen after an execute sequence (C). This prevents
 	// cases where the command is executed before the prompt is written. While not perfect, sitting
 	// on an A without a C following shortly after is a very good indicator that the command is done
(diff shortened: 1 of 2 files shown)


meganrogge:

dmitrivMS (approved):

PR #313379, opened 2026-04-29
Detach rewriter: also wrap sync-mode commands ending in trailing &

### LINKED ISSUE #313378: Detach rewriter misses sync-mode commands with trailing & in eval harnesses
## Problem

When the `chat.tools.terminal.detachBackgroundProcesses` setting is enabled, the `CommandLineBackgroundDetachRewriter` only wraps a command with `nohup` (POSIX) / `Start-Process` (Windows) when the model invokes `run_in_terminal` with `mode='async'` (legacy `isBackground: true`).

If the model instead uses `mode='sync'` and appends a trailing `&` to background a service, the rewriter does **not** fire. The `&` is interpreted by the shell, the process is launched in the foreground shell's process group, and as soon as the agent's terminal session is torn down the service receives `SIGHUP` and dies.

The eval harness then connects to the service from a fresh process and gets `Connection refused` / similar errors, so the benchmark fails even though the model genuinely got the configuration right.

## Evidence — terminalbench2 run `25136645365` (claude-opus-4.6)

Three distinct failures in this single run trace back to exactly this:

| Benchmark | Command (mode=sync) | Eval failure |
|---|---|---|
| `hf-model-inference` | `python app.py > /app/server.log 2>&1 &` | `ConnectionRefusedError` on `http://localhost:5000/sentiment` |
| `install-windows-3.11` | `qemu-system-i386 -hda ... -snapshot -vnc :1 -qmp ... -monitor ...` (no `&`, no `-daemonize`) | `assert 1 == 0` — `Failed to get QEMU command line` |
| `qemu-alpine-ssh` | `nohup expect setup_vm.exp > /app/vm_output.log 2>&1 &` (sync mode) | `sshpass ... root@localhost` returned non-zero |

In each case the agent's verification step inside the same turn (`curl`, `ps aux | grep`, `ss -tlnp`) **passed** — the service really was alive during the run. It just didn't survive to the eval phase.

## Fix

Have the rewriter also fire for sync-mode commands that end in a single trailing `&` (POSIX background operator, distinct from `&&`). Setting still gates the behavior, so default users are unaffected; eval harnesses (which set `chat.tools.terminal.detachBackgroundProcesses=true`) get correct lifecycle handling regardless of which mode the model picks.

PowerShell has no equivalent trailing-`&` operator, so the Windows path remains async-only.

meganrogge:
Fixes microsoft/vscode-internalbacklog#7605

When the `chat.tools.terminal.detachBackgroundProcesses` setting is enabled, also wrap commands with `nohup` when the agent passed `mode='sync'` but the command line ends with a single trailing `&` (POSIX background operator, distinct from `&&`).

Previously the rewriter only fired for `mode='async'` (legacy `isBackground: true`). When the agent used `mode='sync'` + `&`, the `&` was interpreted by the foreground shell and the process was SIGHUP'd as soon as the tool's terminal was disposed — fine for interactive UX, but it breaks eval harnesses where the eval step connects to the service from a fresh process and gets `Connection refused`.

### Evidence

terminalbench2 run `25136645365` (claude-opus-4.6) had three failures attributable to this:

- `hf-model-inference` — `python app.py > /app/server.log 2>&1 &`
- `install-windows-3.11` — `qemu-system-i386 ... -vnc :1 -qmp ... -monitor ...` (no daemonize)
- `qemu-alpine-ssh` — `nohup expect setup_vm.exp > /app/vm_output.log 2>&1 &` in sync mode

In all three the agent's same-turn verification (`curl`, `ps aux`, `ss -tlnp`) confirmed the service was alive — it just didn't survive to the eval phase.

### Behavior

- Setting still gates the rewrite, so default users are unaffected.
- Eval harnesses already set `chat.tools.terminal.detachBackgroundProcesses=true` (see `vscode-copilot-evaluation/src/vsCodeApplication.ts`), so they pick this up automatically.
- PowerShell has no equivalent trailing-`&` operator, so the Windows path remains async-only.
- Existing nohup/`shell -c` wrapping of compound constructs and the de-duplication of an already-trailing `&` continue to work.

### DIFF
--- a/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/commandLineRewriter/commandLineBackgroundDetachRewriter.ts
+++ b/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/commandLineRewriter/commandLineBackgroundDetachRewriter.ts
@@ -30,15 +30,28 @@ export class CommandLineBackgroundDetachRewriter extends Disposable implements I
 	}
 
 	rewrite(options: ICommandLineRewriterOptions): ICommandLineRewriterResult | undefined {
-		if (!options.isBackground) {
-			return undefined;
-		}
-
 		if (!this._configurationService.getValue(TerminalChatAgentToolsSettingId.DetachBackgroundProcesses)) {
 			return undefined;
 		}
 
+		// Detach when:
+		//   1. The tool was invoked with mode='async' (isBackground=true), OR
+		//   2. The command line ends with a single trailing `&` (POSIX background operator).
+		// Case (2) catches commands that the agent intended to background even when called
+		// in mode='sync' — without this, the trailing `&` silently produces a SIGHUP'd
+		// process that dies as soon as the tool's shell tears down.
+		const trimmedForCheck = options.commandLine.trimEnd();
+		const endsWithBareBackgroundAmp = /(?:^|[^&])&$/.test(trimmedForCheck);
+		if (!options.isBackground && !endsWithBareBackgroundAmp) {
+			return undefined;
+		}
+
 		if (options.os === OperatingSystem.Windows) {
+			// PowerShell does not have a POSIX-style trailing `&` background operator,
+			// so only rewrite explicit async-mode commands here.
+			if (!options.isBackground) {
+				return undefined;
+			}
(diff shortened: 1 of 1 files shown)


meganrogge:

roblourens (approved):

PR #313383, opened 2026-04-29
OTel: typed constants cleanup + add otel skill

zhichli:
## Summary

Two related changes for the Copilot Chat OTel pipeline:

1. **OTel: replace string literals with typed constants** (`52947bb`) — non-functional cleanup that promotes raw OTel attribute keys and provider names to constants in `genAiAttributes.ts`.
2. **Add `otel` skill** (`963e7de`) — new `.github/skills/otel/SKILL.md` documenting the OTel architecture and mandating the two monitoring docs stay in sync with code.

## 1. OTel constant cleanup

No behavior change. Audited `extensions/copilot/src/` for raw OTel string literals and routed them through typed constants:

- `GenAiProviderName.GEMINI` added; `geminiNativeProvider`, `anthropicProvider`, `copilotcliSession` now use `GenAiProviderName.*` instead of raw `'gemini'` / `'anthropic'` / `'github'`.
- `CopilotChatAttr.HOOK_TYPE` / `HOOK_INPUT` / `HOOK_OUTPUT` / `HOOK_RESULT_KIND` / `MODE_NAME` / `TOTAL_COST_USD` added; consumers in `chatHookService`, `claudeMessageDispatch`, `copilotCliBridgeSpanProcessor`, `chatDebugFileLoggerService`, `otelSpanToChatDebugEvent`, `toolCallingLoop`, `claudeOTelTracker` switched over.
- New `CopilotCliSdkAttr` group for SDK-emitted hook attribute keys (`github.copilot.hook.*`); the bridge processor and debug-event mapper now use the constants.
- Magic-number `SpanStatusCode` casts (`code: 1 as SpanStatusCode`, `code: 2 as SpanStatusCode`) replaced with the enum members.
- `src/platform/otel/common/index.ts` now also re-exports `CopilotCliSdkAttr` and the previously-missing event helpers (`emitEditFeedbackEvent`, `emitEditHunkActionEvent`, `emitInlineDoneEvent`, `emitEditSurvivalEvent`, `emitUserFeedbackEvent`, `emitCloudSessionInvokeEvent`).

12 files changed, +64/-41. `tsc --noEmit` clean.

## 2. New `otel` skill

`.github/skills/otel/SKILL.md` is patterned on `tool-rename-deprecation` and `sessions`. It points readers to the two authoritative docs and lays out the development contract:

- Authoritative docs (`agent_monitoring.md` / `agent_monitoring_arch.md`) and the rule that they must move in lockstep with code.
- The four agent execution paths (foreground / Copilot CLI in-process / Copilot CLI terminal / Claude) and their distinct OTel strategies.
- Canonical map of every OTel file (`platform/otel/`, the bridge, Claude OTel tracker, debug-panel mappers).
- `IOTelService` selection logic (`Noop` / `NodeOTelService` / `InMemoryOTelService`) and when each is registered.
- Span / metric / event conventions, content-capture gating with `truncateForOTel`, and the `EXPORTABLE_OPERATION_NAMES` filter that isolates the debug panel from OTLP.
- Configuration surface invariants — the three places (`package.json`, `otelConfig.ts`, the two monitoring docs) that must stay in sync.
- Procedure checklists for adding spans / metrics / events / new agent surfaces, and for changing the Copilot CLI bridge.
- Validation commands and a list of known risks and anti-patterns to reject in review.

## Validation

- `npx tsc --noEmit --project tsconfig.json` (in `extensions/copilot/`) → clean.
- No behavior change in the constant cleanup; existing OTel + Bridge tests continue to pass.

## Risk

Low. The constant cleanup is mechanical and verified by the type checker. The skill is documentation only.

### DIFF
--- a/.github/skills/otel/SKILL.md
+++ b/.github/skills/otel/SKILL.md
@@ -151,15 +151,20 @@ return this._otelService.startActiveSpan('invoke_agent child', { parentTraceCont
 
 ### Content capture
 
-Always gate prompt/response/tool-arg bodies on `otel.config.captureContent`:
+The extension uses two conventions side-by-side; pick the right one for the attribute you're adding.
+
+1. **Always emit (truncated)** — used for inputs/outputs that the Agent Debug Log panel needs to be useful even when OTel export is off (e.g. `gen_ai.tool.call.arguments` in [`toolsService.ts`](../../../extensions/copilot/src/extension/tools/vscode-node/toolsService.ts), and `copilot_chat.hook_input` / `hook_output` in [`chatHookService.ts`](../../../extensions/copilot/src/extension/chat/vscode-node/chatHookService.ts)). The attribute is captured unconditionally but always passed through `truncateForOTel`. Use this for moderate-sized, generally-non-secret arguments / results.
+2. **Gate on `config.captureContent`** — used for full prompt / response / system-instruction bodies (e.g. `gen_ai.input.messages`, `gen_ai.output.messages`, `gen_ai.system_instructions`, `gen_ai.tool.definitions` in [`chatMLFetcher.ts`](../../../extensions/copilot/src/extension/prompt/node/chatMLFetcher.ts) and the BYOK providers). These are larger and more likely to contain user secrets.
 
 ```ts
+// Pattern 1 — always emit, always truncate
+span.setAttribute(GenAiAttr.TOOL_CALL_ARGUMENTS, truncateForOTel(JSON.stringify(args)));
+
+// Pattern 2 — gated on captureContent
 if (this._otelService.config.captureContent) {
     span.setAttribute(GenAiAttr.INPUT_MESSAGES, truncateForOTel(JSON.stringify(messages)));
 }
 ```
-
-`truncateForOTel` from `messageFormatters.ts` is mandatory for any free-form content attribute — it prevents OTLP batch failures.
 
 ### Debug panel vs OTLP isolation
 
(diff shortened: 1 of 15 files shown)


zhichli:
Thanks for the review! Audited each finding and pushed `298654bf` addressing them:

### ✅ Finding 1 — `otelSpanToChatDebugEvent.ts`: magic `=== 2 /* ERROR */`
**Valid.** Found 6 occurrences (lines 340, 380, 403, 438, 466, 506). Switched the `SpanStatusCode` import from type-only to value and replaced all six with `SpanStatusCode.ERROR` / `SpanStatusCode.OK`. (The remaining `as SpanStatusCode` casts on lines 362 and 479 are deliberate — they convert the SDK's numeric `code` field to the enum type for `spanStatusToString`.)

### ✅ Finding 2 — `genAiAttributes.ts`: doc strings out of sync
**Valid.**
- `HOOK_RESULT_KIND` doc updated to include `'non_blocking_error'` (which `chatHookService.ts` produces for `HookCommandResultKind.NonBlockingError`).
- `HOOK_INPUT` / `HOOK_OUTPUT` doc strings no longer claim "content-gated" since emitters do not gate. Replaced with: *"truncated; emitters may or may not gate on `captureContent` — used by the Agent Debug Log panel"*.

### ⚠️ Finding 3 — `chatHookService.ts:170,199`: gate hook input/output on `captureContent`
**Won't fix in this PR.** The privacy concern is real, but applying the fix here would regress the Agent Debug Log panel and diverge from the existing convention. Two reasons:

1. The codebase already has an explicit "always capture for debug panel, only truncate" pattern. See [`toolsService.ts:143`](https://github.com/microsoft/vscode/blob/main/extensions/copilot/src/extension/tools/vscode-node/toolsService.ts#L143):
    ```ts
    // Always capture tool call arguments for the debug panel
    if (options.input !== undefined) {
        try {
            span.setAttribute(GenAiAttr.TOOL_CALL_ARGUMENTS, truncateForOTel(JSON.stringify(options.input)));
        } catch { /* swallow serialization errors */ }
    }
    ```
   The hook input/output emit sites in `chatHookService.ts` follow the exact same pattern. Gating only the hook attributes would be inconsistent.

2. `IOTelService.config.captureContent` defaults to `false`. If we gate hook content on it, users with default settings lose hook input/output rendering in the Agent Debug Log panel — which is a UX regression unrelated to OTLP export.

**The proper fix** is exporter-level stripping in `DiagnosticSpanExporter` / `FilteredSpanExporter`: strip `gen_ai.tool.call.arguments`, `gen_ai.tool.call.result`, `copilot_chat.hook_input`, `copilot_chat.hook_output` from spans on their way to OTLP / SQLite when `captureContent === false`, while leaving them on the in-memory event for the debug panel. That's a separate, more careful change.

I've also updated `.github/skills/otel/SKILL.md` (in this commit) so the documented convention reflects reality — two patterns: "always emit (truncated)" for debug-panel-essential attributes, and "gate on `captureContent`" for full prompt/response/system-instruction bodies. This stops future contributors from being given the misleading blanket rule and matches what `chatMLFetcher.ts` (gated) and `toolsService.ts` (always-emit) actually do.

Local typecheck and OTel/Bridge test suite still clean.

roblourens (approved):

PR #313393, opened 2026-04-30
Clean up agent host worktrees on session archive

roblourens:
When a session that owns a Copilot CLI worktree is archived, remove the worktree from disk so abandoned worktrees don't pile up. When the session is unarchived, recreate the worktree pointing at the preserved branch so the user can keep working.

## Changes

- Persist `copilot.worktree.path` and `copilot.worktree.repositoryRoot` alongside the existing branch-name metadata so cleanup can reconstruct the worktree after a process restart (the in-memory `_createdWorktrees` map is empty post-restart).
- Add an optional `IAgent.onArchivedChanged(session, isArchived)` hook, dispatched from `AgentSideEffects` after the archived flag is persisted. Errors are logged and swallowed so the persist step still succeeds.
- Extend `IAgentHostGitService` with `branchExists`, `hasUncommittedChanges`, and `addExistingWorktree` (a `worktree add` without `-b`, for reattaching to a preserved branch).
- Implement `CopilotAgent.onArchivedChanged`:
  - **archive**: skip if metadata absent, worktree directory missing, branch missing, or working tree is dirty; otherwise `removeWorktree` and clear the in-memory entry.
  - **unarchive**: skip if directory already exists or branch missing; otherwise `addExistingWorktree` and re-seed the in-memory entry.
  - All work is sequenced through `_sessionSequencer` so it can't race with `disposeSession`.

## Tests

- Five focused unit tests in `copilotAgent.test.ts` covering the happy archive→unarchive round trip, dirty skip, missing-branch skip, no-metadata no-op, and unarchive-skip-when-path-exists.
- New "worktree helpers (real git)" suite in `agentHostGitService.integrationTest.ts` covering `branchExists`, `hasUncommittedChanges`, and `addExistingWorktree` against real on-disk repos.

(Written by Copilot)


roblourens:
Re: comment on `_META_WORKTREE_PATH` — fixed in 2686ba2aad5. `_readWorktreeMetadata` now only returns `worktreePath`/`repositoryRoot` when they were explicitly persisted; both `_cleanupWorktreeOnArchive` and `_recreateWorktreeOnUnarchive` already early-return when either is missing, so legacy sessions simply leave their worktree on disk untouched (rather than getting deleted with no way back).

(Written by Copilot)

benvillalobos (approved):

PR #313394, opened 2026-04-30
Fix model selection logic in SessionModelPicker to handle empty models and active session changes

### LINKED ISSUE #313385: Select Model not passed to new Session in Agents
* Start Agents
* Select a repo 
* Select a model
* Select another repo 
* Type in a prompt and hit send

The new chat input will not have the same model, the request object will also not have the same model
From what I can tell when we create a new session ([createNewSession](https://github.com/microsoft/vscode/blob/dfbb7bc2aa68e6f70eda4d9ee02de42a6763f9a9/src/vs/sessions/services/sessions/browser/sessionsManagementService.ts#L251)
) , the currently selected details are not initialized in the session 

I.e. the method [setModelId](https://github.com/microsoft/vscode/blob/dfbb7bc2aa68e6f70eda4d9ee02de42a6763f9a9/src/vs/sessions/contrib/copilotChatSessions/browser/copilotChatSessionsProvider.ts#L369) (see below) is not called for the new session thats created after selecting a new repo,

DonJayamanne:
Fixes https://github.com/microsoft/vscode/issues/313385

### DIFF
--- a/src/vs/sessions/contrib/copilotChatSessions/browser/copilotChatSessionsActions.ts
+++ b/src/vs/sessions/contrib/copilotChatSessions/browser/copilotChatSessionsActions.ts
@@ -338,10 +338,22 @@ export class SessionModelPicker extends Disposable {
 
 		const models = getAvailableModels(this._languageModelsService, this._sessionsManagementService);
 		this._modelPicker.setEnabled(models.length > 0);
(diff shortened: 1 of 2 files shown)


DonJayamanne:
@sandy081 /cc

justschen (approved):

PR #313409, opened 2026-04-30
Fix forced sandbox issue

### LINKED ISSUE #313387: Setting chat.agent.sandbox.fileSystem.mac forcefully enables chat.agent.sandbox.enabled on restart
### Issue Description

When configuring `chat.agent.sandbox.fileSystem.mac` in user settings, the `chat.agent.sandbox.enabled` setting is forcefully enabled on every VS Code restart, preventing users from opting out of the sandbox even if they explicitly disable it.

### Steps to Reproduce

1. Add the following to your user settings:
   ```json
   "chat.agent.sandbox.fileSystem.mac": {
     "denyRead": [],
     "allowWrite": ["."],
     "denyWrite": []
   }
   ```
2. Set `chat.agent.sandbox.enabled` to `"off"` via the settings pane (This will remove it from the settings JSON, since "off" is the default)
3. Restart VS Code
4. 🐛Observe that `chat.agent.sandbox.enabled` is automatically set back to "on"

### Expected Behavior

Setting a filesystem configuration should not forcefully override the parent `enabled` setting.

### Additional Context

Initially, I thought this might be related to ExP experiments changing from "setting defaults" to "forcefully setting values," but it appears this behavior is specific to the `chat.agent.sandbox.enabled` setting when filesystem configurations are present.

This seems like either:
1. Unintended coupling between the filesystem config and the enabled flag
2. Missing documentation that the presence of filesystem config implies enablement
3. A policy enforcement that should be made more explicit to users

jruales:
* Fixes https://github.com/microsoft/vscode/issues/313387

### DIFF
--- a/src/vs/workbench/contrib/terminal/common/terminalConfiguration.ts
+++ b/src/vs/workbench/contrib/terminal/common/terminalConfiguration.ts
@@ -718,9 +718,20 @@ export async function registerTerminalConfiguration(getFontSnippets: () => Promi
 Registry.as<IConfigurationMigrationRegistry>(WorkbenchExtensions.ConfigurationMigration)
 	.registerConfigurationMigrations([{
 		key: TerminalContribSettingId.DeprecatedAgentSandboxEnabled,
-		migrateFn: (value: boolean, valueAccessor) => {
+		migrateFn: (value: unknown, valueAccessor) => {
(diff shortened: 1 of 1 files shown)


DonJayamanne (approved):

isidorn:
Thank you very much for this one 🙏

PR #313414, opened 2026-04-30
Fix spurious exit-130 cascade, idle SIGINT, nohup over-wrap, and missing terminal ID

### LINKED ISSUE #313412: Terminal chat tool: spurious SIGINT and `nohup` over-wrap cause command failures in agent runs
# Terminal chat tool: spurious SIGINT and `nohup` over-wrap cause command failures in agent runs

## Summary

Two long-standing edge cases in the terminal chat-agent tools cause commands to fail in ways the model cannot diagnose or recover from. Both showed up as a major contributor to failed tasks in two recent `terminalbench2` evaluation runs.

### Run `25141808876` — claude-opus-4.6 (87 tasks)

| Symptom | Count | Affected benchmarks | Unresolved benchmarks |
| --- | ---: | --- | --- |
| `Command exited with code 130` for a fresh command | 42 | 9 | 6 (`caffe-cifar-10`, `mteb-leaderboard`, `train-fasttext`, `compile-compcert`, `make-doom-for-mips`, `mcmc-sampling-stan`) |
| `Error: No active terminal execution found` | 1 | 1 | 1 (`train-fasttext`) |
| Silent `nohup … &` rewrite of an `async` command | 17 | 13 | several |
| …of which the command was interactive (e.g. `expect`) | 1 | 1 (`qemu-alpine-ssh`) | 0 |

### Run `25141798563` — gpt-5.4 (87 tasks)

| Symptom | Count | Affected benchmarks | Unresolved benchmarks |
| --- | ---: | --- | --- |
| `Command exited with code 130` for a fresh command | 13 | 8 | 5 (`protein-assembly`, `tune-mjcf`, `caffe-cifar-10`, `gcode-to-text`, `qemu-startup`, `train-fasttext`) |
| `Error: No active terminal execution found` | 2 | 1 | 1 |
| Silent `nohup … &` rewrite of an `async` command | 9 | 7 | several |

Across both runs that totals **~11 unresolved benchmarks** where the agent observed a confusing terminal-tool symptom directly attributable to one of these bugs.

## Root cause #1 — spurious SIGINT (`130`) on the next command

[basicExecuteStrategy.ts](https://github.com/microsoft/vscode/blob/main/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/executeStrategy/basicExecuteStrategy.ts) and [noneExecuteStrategy.ts](https://github.com/microsoft/vscode/blob/main/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/executeStrategy/noneExecuteStrategy.ts) both unconditionally send `\x03` (Ctrl+C) before the next `sendText` call whenever `_hasReceivedUserInput()` is `true`:

```ts
if (this._hasReceivedUserInput()) {
    this._log('Command timed out, sending SIGINT and retrying');
    await this._instance.sendText('\x03', false);
    await waitForIdle(this._instance.onData, 100);
}
```

`receivedUserInput` is set whenever the terminal emits any data and is reset only after a *successful* execute-strategy completion. So a previous step that left the prompt in a "received data" state — common when a long-running command was forcibly closed or the model used `send_to_terminal` — causes Ctrl+C to be sent at the *idle* prompt of the **next** command. Bash echoes `^C` and many shells return exit code `130` for the freshly-sent command, which the model interprets as the new command having been interrupted.

Concrete example — `train-fasttext` step 19 in run `25141808876`:

```
agent: run_in_terminal sync `cd /app && rm -f nohup.out model.bin && python3 train.py 2>&1`
result: ^C
        Command exited with code 130
```

The agent never sent SIGINT. The tool did.

## Root cause #2 — silent `nohup` rewrite of interactive commands

[`CommandLineBackgroundDetachRewriter`](https://github.com/microsoft/vscode/blob/main/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/commandLineRewriter/commandLineBackgroundDetachRewriter.ts) wraps every `mode: 'async'` command with `nohup … &` when `terminal.integrated.detachBackgroundProcesses` is enabled. `nohup` closes stdin, which makes interactive front-ends — `expect`, `gdb`, `psql`, `passwd`, `sudo` (without `-n`), bare `ssh user@host`, etc. — hang or fail immediately.

`qemu-alpine-ssh` step 12: agent sent `cd /app && expect setup_vm.exp` with `mode: 'async'`. The wrapper rewrote to `nohup expect setup_vm.exp &`, which exits instantly because `expect` cannot read its commands from a closed stdin.

## Root cause #3 — sync results don't include the terminal id

When `mode: 'sync'`, the result text contains the command output and exit code but no `ID=…`. After timeouts or rewrites, models infer/hallucinate a UUID for `get_terminal_output`. `train-fasttext` step 14 in run `25141808876`:

```
agent: get_terminal_output id="b54bb4a3-…"  (hallucinated)
result: Error: No active terminal execution found with ID b54bb4a3-…
```

## Proposed fix

Tracked in #313414. Three small, localized changes:

1. **`basicExecuteStrategy.ts`** — only send Ctrl+C when `commandDetection.executingCommandObject !== undefined` (a command is actually running). Otherwise send Ctrl+U (kill-line), which is a no-op at an idle prompt and just clears any stray pending input.
2. **`noneExecuteStrategy.ts`** — always use Ctrl+U; without shell integration we cannot tell whether a command is running, and Ctrl+C at an idle prompt is the bug.
3. **`commandLineBackgroundDetachRewriter.ts`** — skip the `nohup`/`Start-Process` wrap for known interactive front-ends.
4. **`runInTerminalTool.ts`** — include `terminal ID=…` in the user-edit and tool-edit "Note:" prefixes so the model has the id even when the rewrite changed the command. A short hint explains that the wrap closed stdin.

## Expected impact

Given the eval-log evidence above, this should remove a spurious failure mode in roughly **11 unresolved benchmarks across the two runs** (5–6 per run), without changing behavior for the resolved benchmarks where SIGINT-130 happens after the model has already moved on.

meganrogge:
## Summary

Fixes four independent edge cases in the chat-agent terminal tools that together produce confusing terminal failures in agent runs. Findings, eval-log evidence, and per-benchmark counts are in #313412.

### 1. Filter stale `onCommandFinished` from prior ETX-killed commands (the main bug)

When a prior command is still executing in the terminal (e.g. abandoned after the 30s data-idle fallback), `runCommand()` sends ETX (Ctrl+C) to clear the prompt. This kills the prior command and fires `onCommandFinished` with exit code 130. The `onCommandFinished` subscription set up for the **new** command catches this stale event, causing it to resolve instantly with exit 130 before the new command even starts. This cascades — every subsequent command on the same terminal also gets exit 130.

This PR captures `executingCommandObject` at strategy entry and filters it out of the `onCommandFinished` subscription using `Event.filter()`, so only the **new** command's finish event can resolve the race.

Applied to both **`richExecuteStrategy.ts`** and **`basicExecuteStrategy.ts`**.

Eval-log evidence: in run `25141808876`, all 6 directly-affected failures used the rich strategy (100%). Tasks like `caffe-cifar-10`, `train-fasttext`, `compile-compcert` showed commands like `echo "checking"` returning exit 130 in <341ms.

### 2. Stop sending SIGINT at idle prompts

`basicExecuteStrategy.ts` and `noneExecuteStrategy.ts` both send `\x03` (Ctrl+C) before the next `sendText` whenever `_hasReceivedUserInput()` is `true`. `receivedUserInput` is set whenever the terminal emits any data and is reset only after a *successful* execute-strategy completion. So a previous step that left the prompt in a "received data" state causes Ctrl+C to be sent at the **idle** prompt of the **next** command, producing `Command exited with code 130`.

This PR:

- **`basicExecuteStrategy.ts`** — only send Ctrl+C when `commandDetection.executingCommandObject !== undefined` (a command really is running). Otherwise send Ctrl+U (kill-line).
- **`noneExecuteStrategy.ts`** — always use Ctrl+U; without shell integration we cannot tell whether a command is running.

### 3. Don't auto-`nohup`-wrap interactive commands

`CommandLineBackgroundDetachRewriter` wraps every `mode: 'async'` command with `nohup … &` (POSIX) or `Start-Process` (Windows). This closes stdin, which makes interactive front-ends — `expect`, `gdb`, `psql`, `passwd`, `sudo` (without `-n`), bare `ssh`, etc. — hang or fail immediately.

This PR adds a conservative deny-list to `commandLineBackgroundDetachRewriter.ts` for known interactive commands.

### 4. Include the terminal ID in rewrite/user-edit notes

After timeouts or rewrites, models hallucinate terminal UUIDs for `get_terminal_output` because the result text contains no `ID=…`. This PR adds `terminal ID=…` to the user-edit and tool-edit `Note:` prefixes, plus a short stdin-closed hint when the rewrite was a detach wrap.

## Eval-log impact

Spurious exit-130 results from two recent `terminalbench2` runs:

| Run | Exit-130 hits | Affected benchmarks | Unresolved benchmarks affected |
| --- | ---: | ---: | ---: |
| `25141808876` (claude-opus-4.6) | 42 | 10 | 6 |
| `25141798563` (gpt-5.4) | 13 | 8 | 5 |

Together ~11 unresolved benchmarks where the agent observed confusing terminal-tool symptoms directly attributable to these bugs.

## Test plan

- New unit tests for `CommandLineBackgroundDetachRewriter` covering the interactive deny-list.
- Existing strategy tests continue to pass.
- Manual: in a session that runs a long silent compile and is force-stopped, then runs `echo hello`, the `echo` step now returns `code=0` instead of `code=130`.

Closes #313412


meganrogge:

dmitrivMS (approved):

PR #313425, opened 2026-04-30
Applying network filter only for fetch web tool in sandbox mode

### LINKED ISSUE #313424: Integrated browser in sandbox mode should be revisited.
Currently in sandbox mode, integrated browser is unable to share with agent causing workflow to break. This has to be revisited.

dileepyavan:
fixes #313424

### DIFF
--- a/src/vs/platform/networkFilter/common/networkFilterService.ts
+++ b/src/vs/platform/networkFilter/common/networkFilterService.ts
@@ -17,12 +17,15 @@ import { AgentNetworkDomainSettingId } from './settings.js';
 
 export const IAgentNetworkFilterService = createDecorator<IAgentNetworkFilterService>('agentNetworkFilterService');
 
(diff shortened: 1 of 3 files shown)


src/vs/platform/networkFilter/common/networkFilterService.ts
```diff
@@ -131,6 +135,13 @@ export class AgentNetworkFilterService extends Disposable implements IAgentNetwo
 
 		return result;
 	}
+	// Determines whether network filtering should be applied for a given request
+	// based on the global network filter setting, the terminal sandbox state, and the tool making the request.
+	// For sandbox mode, network filtering is applied only when the global network filter is disabled
+	// and the request is coming from the fetch web tool.
```
jruales, line 141:
```suggestion
	// For sandbox mode, network filtering is applied only when the request is coming from the
	// fetch web tool.
```

jruales (approved):

Issue #310947, opened 2026-04-17
Agents Insider always shows "Update Available"


navneetkarnani:
Type: <b>Bug</b>

Has been happening for a while.

I install "brew install visual-studio-code@insiders". I see the update happening. I restart VSCode. All looks good.

And then I start Agents. And i see the "Update Available" button all the time. I have auto update turned off because i update using homebrew first thing every day.

VS Code version: Agents - Insiders 1.117.0-insider (98b4d3e24443f6566e694021a376e1760c4a7f62, 2026-04-14T16:52:34Z)
OS version: Darwin arm64 25.4.0
Modes:

<details>
<summary>System Info</summary>

|Item|Value|
|---|---|
|CPUs|Apple M1 Pro (10 x 2400)|
|GPU Status|2d_canvas: enabled<br>GPU0: VENDOR= 0x106b [Google Inc. (Apple)], DEVICE=0x0000 [ANGLE (Apple, ANGLE Metal Renderer: Apple M1 Pro, Version 26.4.1 (Build 25E253))], DRIVER_VENDOR=Apple, DRIVER_VERSION=26.4.1 *ACTIVE*<br>Machine model name: MacBookPro<br>Machine model version: 18.3<br>direct_rendering_display_compositor: disabled_off_ok<br>gpu_compositing: enabled<br>multiple_raster_threads: enabled_on<br>opengl: enabled_on<br>rasterization: enabled<br>raw_draw: disabled_off_ok<br>skia_graphite: enabled_on<br>trees_in_viz: disabled_off<br>video_decode: enabled<br>video_encode: enabled<br>webgl: enabled<br>webgl2: enabled<br>webgpu: enabled<br>webnn: disabled_off|
|Load (avg)|2, 3, 6|
|Memory (System)|16.00GB (1.43GB free)|
|Process Argv|--crash-reporter-id 08b71cba-2ec9-47d2-abe5-accf411cc52c|
|Screen Reader|no|
|VM|0%|
</details>Extensions: none<details>
<summary>A/B Experiments</summary>

```
vsliv368cf:30146710
pythonvspyt551cf:31249598
nativeloc1:31118317
dwcopilot:31158714
dwoutputs:31242946
copilot_t_ci:31333650
g012b348:31231168
pythonrdcb7:31268811
pythonpcpt1:31399616
6518g693:31302842
82j33506:31327384
63221493:31336333
envsactivate1:31349248
editstats-enabled:31346256
cloudbuttont:31366566
[...]

PR #313450, opened 2026-04-30
Reuse VS Code's update widget/tooltip in the Agents app

dmitrivMS:
I think it should help with #310947 since the logic for showing the button will be the same.


deepak1556:
Would like to try the flow for verifying cross app coordination before merging, build at https://monacotools.visualstudio.com/Monaco/_build/results?buildId=435201&view=results

dmitrivMS:
@deepak1556 Started a new build since the other one was failing for unrelated reasons:
https://monacotools.visualstudio.com/Monaco/_build/results?buildId=435303&view=results

deepak1556 (approved):
Agents app is being replaced with window, so no additional sanity testing is required and this is a good cleanup to have

Issue #312086, opened 2026-04-23
Regression in VS Code 1.117.0: Azure AI Foundry / Azure OpenAI custom models fail with HTTP 400 via Language Models -> Add Models


PatMan6889:
### Environment

- VS Code: 1.117.0
- Commit: 10c8e557c8b9f9ed0a87f61f1c9a44bde731c409
- GitHub Copilot Chat: 0.44.2 (prod)
- OS: Windows 11 / win32 10.0.22621 x64

### Summary

Using an Azure AI Foundry / Azure OpenAI model added through:

`Language Models -> Add Models`

fails in VS Code 1.117.0 with an HTTP 400 error.

This worked before the recent update(s) and appears to have regressed very recently.
Not sure at which version it broke, because a few updates might have been skipped.

### Minimal configuration

The model is configured in `chatLanguageModels.json` with a custom Azure vendor entry pointing to a Responses API endpoint:

```json
[
  {
    "name": "Azure Custom Model",
    "vendor": "azure",
    "apiKey": "${input:chat.lm.secret.REDACTED}",
    "models": [
      {
        "id": "gpt-5.3-codex",
        "name": "gpt-5.3-codex",
        "url": "https://REDACTED.services.ai.azure.com/api/projects/REDACTED/openai/v1/responses",
        "toolCalling": true,
        "vision": true,
        "maxInputTokens": 128000,
        "maxOutputTokens": 16000
      }
    ]
  }
]
```

### Steps to Reproduce

1. Open VS Code 1.117.0
2. Configure or use an Azure AI Foundry / Azure OpenAI custom language model via `Language Models -> Add Models`
3. Select the model in chat
4. Send a normal prompt

### Actual Result

The request fails with:

```txt
Sorry, your request failed. Please try again.
Reason: Request Failed: 400
```

Error payload:

```json
{
  "error": {
[...]

PR #313502, opened 2026-04-30
Fix malformed Responses API input items for custom models

DrHazemAli:
## Summary

This PR fixes a Responses API payload issue where custom model requests can include an `input[]` item with an empty `type`, causing the API to fail with:

`Invalid value: ''. param: input[1]`

## Root cause

The Responses API requires every `input[]` item to have a valid `type`.

In some custom model flows, especially with Responses API based models such as GPT-5.4 Pro and similar reasoning models, a message-shaped item can be added to the payload with an empty or missing `type`. The request reaches the API, but fails validation before the model can process it.

This is a client-side payload serialization issue, not a deployment, authentication, or model availability issue.

## Affected models

This can affect custom models that rely on the Responses API instead of the classic Chat Completions API, including GPT-5.4 Pro style deployments and similar reasoning/model-inference deployments, However Chat-completions-compatible models, such as Kimi K2.6 in this tested setup, are not affected by this specific payload shape issue.

## Related issues

- Issue [#312086](https://github.com/microsoft/vscode/issues/312086)
- Feedbacks from Community

## Changes

- Preserve all input items with a non-empty `type`.
- Normalize message-shaped items with empty/missing type to `type: "message"`.
- Drop empty placeholder items with no useful content.
- Avoid hardcoded allowlists so MCP, tool calls, reasoning items, and future Responses API item types are not affected.

Best,
Hazem


raffaeler:
Please consider that providers like DeepSeek need the `reasoning_content` (see [here](https://api-docs.deepseek.com/guides/thinking_mode) field back. This additional field is not used  by  OpenAI  since  they do not emit the full CoT.

I solved the issue by [creating a proxy](https://github.com/raffaeler/mindthegap), but it would definitely better to support `reasoning_content` management directly in the OpenAI generic client.

HTH

DrHazemAli:
Hi @raffaeler 

The current normalization logic only fixes empty or missing type fields and does not block or strip unknown or provider-specific fields, So fields like `reasoning_content` ..etc  are preserved and passed through as-is.

stasyu2009-ux:
Треба щоб хтось з колонірував код тоді буде все вірно працювати в мене немає на це права

raffaeler:
Hi @DrHazemAli 

> The current normalization logic only fixes empty or missing type fields and does not block or strip unknown or provider-specific fields, So fields like `reasoning_content` ..etc are preserved and passed through as-is.

I understand that. Take my comment as a suggestion. I believe that more and more people will try to use open reasoning models  which need that.

stasyu2009-ux:
Тоді буде правильно все працювати

stasyu2009-ux:
Добре. Але там проблема не втому а в самому коді тому треба зробити клон

DrHazemAli:
Hi @raffaeler 
> I understand that. Take my comment as a suggestion. I believe that more and more people will try to use open reasoning models which need that.

Thanks for clarifying.
Yeah, I agree this would be valuable to support as an enhancement.
I’ll look into it as a follow-up enhancement after this fix.

stasyu2009-ux:
неправильно зроблено фільтрування та переносиння коду тому є фатальною помилкою тому буде покарено

stasyu2009-ux:
тобто покарано

vritant24:
I think this is fixed and implemented by https://github.com/microsoft/vscode/pull/325017

Let me know if that's not the case, and I can reopen this PR

stasyu2009-ux (commented):
Все зробив

PR #313529, opened 2026-04-30
Fix nohup wrapping for env var assignments, shell operators, and cd/exec builtins

meganrogge:
The background detach rewriter now correctly wraps commands in shell -c when they use:

 - Inline env var assignments (e.g. VAR=val cmd) that nohup would try to exec as a program name
 - Shell operators (|, &&, ||, ;) where nohup would only exec the first command in the chain
 - cd and exec builtins that nohup cannot exec directly
 - Mid-command & (background operator) that was not detected by the operator regex

Previously these commands would fail silently or behave incorrectly when wrapped with nohup. For example:

 nohup OPAMROOT=/root/.opam opam install menhir &

would fail with 'No such file or directory' because nohup tried to execute OPAMROOT=/root/.opam as a command.

Now they are properly wrapped:

 nohup /bin/bash -c 'OPAMROOT=/root/.opam opam install menhir' & disown

Fixes

 1. Env var assignments — detect VAR=val cmd pattern and wrap in shell -c
 2. Shell operators — detect pipes, chains (&&, ||), semicolons, and mid-command &
 3. Builtins — added cd and exec to the builtin list requiring shell -c wrapping
 4. Double-& bug — when wrapping in shell -c, strip trailing & from inner command before quoting to avoid 
nohup bash -c 'cmd &' & (two backgrounding operators)
 5. Job completion contamination — added disown after nohup cmd & for bash/zsh to remove the job from the 
shell's job table, preventing [1]+ Done nohup ... notifications from contaminating subsequent command output 
and causing spurious SIGINT (exit 130)

Before/After: disown

Before: When a nohup'd background process completes in a reused terminal, bash prints a job completion notice
on the next prompt:

 $ echo hello
 ^C[1]+  Done                    nohup python3 server.py
 hello

This ^C[1]+ Done output contaminates the next command's captured output and can cause exit code 130 (SIGINT).

After: With nohup cmd & disown, the job is removed from bash's job table immediately. No [1]+ Done
notification, no ^C, no contamination. The nohup'd process still runs normally — disown only affects bash's
bookkeeping. Scoped to bash/zsh only (fish left unchanged).

### DIFF
--- a/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/commandLineRewriter/commandLineBackgroundDetachRewriter.ts
+++ b/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/commandLineRewriter/commandLineBackgroundDetachRewriter.ts
@@ -109,29 +109,36 @@ export class CommandLineBackgroundDetachRewriter extends Disposable implements I
 	private _rewriteForPosix(options: ICommandLineRewriterOptions): ICommandLineRewriterResult {
 		const trimmed = options.commandLine.trimEnd();
 
+		// Check for a trailing background `&` (not `&&`) on the original command.
+		// When wrapping in `shell -c`, we strip the trailing `&` from the inner
+		// command and always place it outside the quotes to avoid double-backgrounding.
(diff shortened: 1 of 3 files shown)


meganrogge:

dmitrivMS (approved):

PR #313542, opened 2026-04-30
Add dedicated Chronicle telemetry events for usage insights

digitarald:
## Summary

Replaces the generic `chronicle` telemetry event with dedicated events that give actionable insight into Chronicle usage patterns.

## New/Enriched Events

| Event | Purpose |
|-------|---------|
| `chronicle.standup` | Data richness (session/turn/file counts), query duration, prompt length, empty rate |
| `chronicle.prompt` | Tips vs freeform usage, cloud consent state |
| `chronicle.reindex` | Processed/skipped counts, duration, force mode, cancellation |
| `chronicle.cloudSync` (enriched) | `batchSuccess` volume, `bufferDrop` counts, `circuitBreaker` transitions, `sessionDisabled` reasons |

## Key Decisions

- **Removed `_sendTelemetry()` helper** — no longer needed since each event has its own schema
- **`chronicle.standup` fires before early-return** on empty case — captures the "no sessions" signal
- **`chronicle.cloudSync` GDPR annotation moved** to before first call site (sessionDisabled) to satisfy extractor ordering
- **`batchSuccess`** fires per-flush — flagged as potential high-volume; can add sampling later if needed

## Testing

- TypeScript compiles cleanly (`npx tsc --noEmit`)
- Hygiene pre-commit hook passes
- All GDPR annotations match emitted properties


vijayupadya (dismissed):

digitarald:
Addressed the review feedback in d541adc:

- `eventsCount` now reports the sum of `eventsBySession` sizes (actually submitted events) instead of `batch.length`
- Added `orphanedCount` metric to track re-queued/dropped entries separately
- Updated GDPR annotation with the new field and clarified `eventsCount` comment

vijayupadya (approved):

PR #313558, opened 2026-04-30
sessions: count remote plugin customizations

joshspicer:
## Summary
- Preserve provider-declared plugin and built-in storage while normalizing customization items, even when AHP remote items do not have local plugin URIs.
- Include active remote provider plugin rows in the Agents sidebar Plugins count while excluding locally synced `remote-client` rows.
- Keep the Plugins list count aligned with displayed rows and avoid remote provider refetches when only local installed plugins change.
- Preserve the newer `origin/main` customizations header collapse-toggle/touch handling.

## Testing
- `node --experimental-strip-types build/hygiene.ts $(git --no-pager diff --name-only -- 'src/**/*.ts' 'src/**/*.md')`
- `npm run transpile-client -- --quiet`
- `env -u ELECTRON_RUN_AS_NODE ./scripts/test.sh --run src/vs/workbench/contrib/chat/test/browser/aiCustomization/aiCustomizationItemsModel.test.ts --run src/vs/workbench/contrib/chat/test/browser/aiCustomization/aiCustomizationListWidget.test.ts --run src/vs/sessions/contrib/sessions/test/browser/aiCustomizationShortcutsWidget.test.ts`

### DIFF
--- a/src/vs/sessions/AI_CUSTOMIZATIONS.md
+++ b/src/vs/sessions/AI_CUSTOMIZATIONS.md
@@ -238,7 +238,9 @@ Skills that are directly invoked by UI elements (toolbar buttons, menu items) ar
 
 ### Count Consistency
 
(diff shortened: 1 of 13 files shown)


joshspicer:
Superseded by #313568, which folds in this PR's commits plus an additional fix for the groupKey-only built-in fallback case.

PR #313575, opened 2026-04-30
AHP/Customizations: BrowserPluginGitCommandService for adding plugins in web

joshspicer:
Replaces the throwing stub at `src/vs/workbench/contrib/chat/browser/pluginGitCommandService.ts` with a real implementation that lets browser/web clients install agent plugins from public (and authenticated) GitHub repositories without a local git binary or AHP server.

## Approach

`git` is not available in the browser, so plugin contents are reconstructed from two CORS-friendly endpoints on `api.github.com`:

- `GET /repos/{owner}/{repo}/commits/{ref}` resolves the requested ref (branch / tag / SHA / `HEAD`) to a commit SHA.
- `GET /repos/{owner}/{repo}/git/trees/{sha}?recursive=1` returns the full file listing.
- `GET /repos/{owner}/{repo}/git/blobs/{blob_sha}` returns each blob as base64 JSON, decoded via `decodeBase64`.

Why not the tarball or `raw.githubusercontent.com`?
- The `/tarball/{sha}` endpoint 302-redirects to `codeload.github.com`, which returns no `Access-Control-Allow-Origin` header — the browser blocks the redirect.
- `raw.githubusercontent.com` accepts simple GETs but rejects the OPTIONS preflight that any `Authorization: Bearer` header forces, so it cannot serve private repos and fails any time the user has a session attached.
- `api.github.com` is the only host that handles CORS correctly with auth, so we pay one HTTP round-trip per file in exchange for actually working in the browser. Per-blob fetches run through a `Limiter(10)` so trees with many files don't overload the connection pool.

Extraction is staged to a sibling directory and atomically moved into place on success; failures clean up the staging dir and leave the existing `targetDir` untouched. Tree entries are sanitised against path traversal (NUL, leading `/`, `..`, backslash, post-join `isEqualOrParent` check). Symlinks (`mode === '120000'`) and submodules (`type === 'commit'`) are skipped — neither is representable in the workbench virtual file system.

`{owner, repo, ref, sha, fetchedAt}` is persisted per-target via `IStorageService` (`chat.plugins.browserCache.v1`). `revParse()` answers from this cache, `pull()` skips re-download when the upstream SHA has not moved, and the persisted SHA feeds `CustomizationRef.nonce` so the AHP server's plugin manager can dedupe.

`fetch` / `fetchRepository` / `revListCount` are no-ops by design: there is no local git database to update, and `pull()` is the single source of truth for updates.

### Authentication ladder

`cloneRepository` walks a three-stage fallback so the same flow handles every realistic browser auth state:

1. **Existing signed-in session** — silent `getSessions('github', [], { silent: true })` (matching `githubApiClient.ts` and `githubFileSystemProvider.ts`); prefers a `repo`-scoped session if multiple are present.
2. **Anonymous fallback** — when an existing token returns 401/403 (e.g. SSO-blocked org session against a public repo), retry without the header. No harm, often unblocks.
3. **Prompt for new session** — `createSession('github', ['repo'], { activateImmediate: true })` only when both prior attempts fail.

If the final attempt still 401/403s, throws a localized `pluginsBrowserGitHubAccessRequired` error directing the user to sign in with an account that has access to the repo. Non-GitHub clone URLs throw a different localized error pointing them at desktop / a remote agent host.

### Error logging

A `loggedRequest` wrapper in [githubRepoFetcher.ts](src/vs/workbench/contrib/chat/browser/githubRepoFetcher.ts) re-throws transport-level errors with the URL we were trying to reach, so a CORS / DNS / offline failure surfaces as `Network error during pluginGit.blob (GET https://api.github.com/...): TypeError: Failed to fetch` instead of a bare `TypeError: Failed to fetch`. `_maybeLogTransientError` in `BrowserPluginGitCommandService` mirrors this for top-level failures so the install flow always logs the URL + cause.

### Plugins toolbar (`pluginListWidget.ts`)

Adjusted the Plugins section toolbar in the AI Customizations editor:

- **Browse Marketplace** stays on the toolbar but is disabled with an explanatory tooltip on web (`isBrowseMarketplaceAvailable() === !isWeb`).
- **Add** is a single secondary button when only one add-action is contributed; promotes to a `ButtonWithDropdown` when multiple are available (e.g. host plugin actions plus Install Plugin from Source). `addPrimaryActionToDropdown: false` so the dropdown lists only the secondary actions.
- **Create Plugin** is its own dedicated icon button (`workbench.action.chat.createPlugin`).

The empty-state subtext now describes both add paths instead of the old "remote agent host" wording.

The DI singleton registration in `chat.contribution.ts` already wires `IPluginGitService -> BrowserPluginGitCommandService`; the new constructor parameters are injected automatically. Desktop continues to use the real-`git` impl.

## Tests

`src/vs/workbench/contrib/chat/test/browser/pluginGitCommandService.test.ts` (25 tests) covers:

- URL parsing edge cases (canonical / trailing slash / `.git` / non-GitHub).
- Happy path tree + blob fetch with persisted SHA metadata.
- All four rungs of the auth ladder (no session → anonymous, signed-in token reused, signed-in token rejected → anonymous fallback, anonymous still 403 → prompt for repo session).
- 403 + `X-RateLimit-Remaining: 0` mapped to typed `GitHubRateLimitError`.
- Staged-extraction safety: a 500 on the second install leaves the previous tree intact and the cache pointing at the original SHA.
- Symlink + submodule entries skipped.
- `pull()` no-ops when SHA matches, re-extracts when it moves, clears stale files.
- Path-traversal entries rejected (`../`, `..\\..\\`).
- `checkout()` no-ops on cached SHA, re-extracts on different SHA.
- `revParse()` rejects unrelated SHAs to surface tarball-cache mismatches.

Stubs queue tree + blob JSON responses keyed by URL regex; the `StubRequestService` records every request so auth-header assertions work across the auth ladder.

## Follow-ups (not in this PR)

- Server-side ("Approach B") flow that lets the AHP server clone with real `git`, gated by a new `CustomizationRef` git-source variant. Cheap to add on top thanks to `AgentPluginManager` already owning the LRU/nonce/sequencer machinery.
- Docs.


connor4312 (dismissed):
nice!

bryanchen-d (approved):

osortega (approved):

PR #313579, opened 2026-04-30
Fix stale onCommandFinished filter to compare by marker identity

meganrogge:
The exit-130 cascade filter introduced in PR #313414 compared command objects by reference identity (e !== staleExecutingCommand). This worked for AHP terminals where executingCommandObject returns the same object reference, but failed for standard CommandDetectionCapability where executingCommandObject calls promoteToFullCommand() which creates a new wrapper object each time. The onCommandFinished event also creates a separate wrapper via promoteToFullCommand(). Since these are different objects, the identity check (e !== staleExecutingCommand) always passed, making the filter a no-op.

Fix: compare by the underlying xterm IMarker reference (commandStartMarker) which is shared between both promoted wrappers, providing a stable identity.

Evidence from eval runs 25181796216 and 25181764231 (agent build c64d40ebc3 which includes PR #313414):
- Exit-130 cascade still present in 8 and 5 tasks respectively
- Pattern: 30s data-idle fallback abandons command, next command gets sequence D (runCommand sends ^C), exit 130 fires within 1ms
- Filter was present in build but never caught the stale event


meganrogge:

PR #313589, opened 2026-04-30
fix: disable data-idle fallback for sync terminal commands

meganrogge:
## Problem

Sync terminal commands that produce no output for extended periods were being prematurely abandoned by data-idle fallback timers in `trackIdleOnPrompt`. This triggered a cascade:

1. **Fallback fires** → `trackIdleOnPrompt` resolves → strategy reports command "done"
2. **Next `runCommand` sends Ctrl+C** to clear prompt → kills the still-running process → exit 130
3. **Stale `onCommandFinished` event** captured by new command → reports exit 130 → cascade continues

This was a root cause of `X_AGENT_STILL_RESPONDING` timeouts and exit-130 failures in terminal benchmarks.

### 1. Disable fallbacks for sync mode in rich strategy
`trackIdleOnPrompt` has 4 fallback timers (initial 2min, prompt 1s, executing 5min, hard-cap 5min) that resolve the idle promise when no shell integration sequences arrive. These are **not needed** for rich strategy sync mode because `onCommandFinished` and `onExit`/`onDisposed` already cover completion detection. Added `disableFallbacks` option to `ITrackIdleOnPromptOptions` and set it for sync mode.

### 2. Remove `trackIdleOnPrompt` from rich strategy async path
For async mode, the `OutputMonitor` handles idle detection at the tool level — the strategy result is not awaited. Including `trackIdleOnPrompt` in the race was redundant. Removed it so only sync mode includes it (with fallbacks disabled).

### 3. Fix stale `onCommandFinished` filter (cherry-picked from #313579)
The stale event filter compared `TerminalCommand` object identity (`e !== staleExecutingCommand`), but `promoteToFullCommand()` creates new wrapper objects — so the filter never matched and was effectively a no-op. Fixed by comparing `e.marker !== staleMarker` since the underlying xterm `IMarker` reference is shared.

## Why basic strategy is unchanged
Basic strategy uses `trackIdleOnPrompt` as its **primary completion mechanism** — shell integration is unreliable in basic mode, so SI sequences (A/C/D) may never arrive. Disabling fallbacks there would cause hangs. Basic strategy keeps its default fallback behavior with only the marker identity fix applied.

## Behavior summary

| Scenario | Rich Strategy | Basic Strategy |
|---|---|---|
| **Sync — SI works** | `onCommandFinished` or `trackIdleOnPrompt` (no fallbacks) | `onCommandFinished` → idle confirm, or fallbacks (2min/5min/5min) |
| **Sync — SI breaks** | Blocks until `onExit`/`onDisposed`/cancellation | Fallbacks fire at 2min/5min → resolves |
| **Async — SI works** | No `trackIdleOnPrompt` in race; OutputMonitor handles it | `onCommandFinished` + idle confirm (with fallbacks) |
| **Async — SI breaks** | Resolves via `onExit`/`onDisposed`/cancellation | Fallbacks fire → cleanup |


meganrogge:

justschen (dismissed):

bryanchen-d (approved):

PR #313605, opened 2026-04-30
Agentic browser tools + Agent Host

### LINKED ISSUE #313798: Browser: Enable browser tools in Agent Host sessions

jruales:
* Closes https://github.com/microsoft/vscode/issues/313798

To try it out in OSS:

- Open Agents app
  - image
- Set `"chat.agentHost.enabled": true,`
- Set `"workbench.browser.agentHostChatToolsEnabled": true,`
- In Agents app, click on the option to select a workspace or folder, then click on Select Folders... > Local Agent Host
  - image
- Select a folder
- Send a prompt that uses browser. It doesn't matter if you use a Folder or Worktree type of session
  - image

### DIFF
--- a/src/vs/workbench/contrib/chat/browser/chat.contribution.ts
+++ b/src/vs/workbench/contrib/chat/browser/chat.contribution.ts
@@ -195,12 +195,7 @@ CommandsRegistry.registerCommand('_chat.notifyQuestionCarouselAnswer', (accessor
 const toolReferenceNameEnumValues: string[] = [];
 const toolReferenceNameEnumDescriptions: string[] = [];
 
-/**
- * Tool reference names for the integrated browser tools registered in
- * `src/vs/workbench/contrib/browserView/electron-browser/tools/`. Exposing
- * these as agent-host client tools lets background agents drive the
- * integrated browser pane (open/read pages, click, type, screenshot, etc.).
- */
+// Tool reference names for the integrated browser tools.
 const browserClientTools = [
 	'openBrowserPage',
 	'readPage',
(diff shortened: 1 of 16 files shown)


src/vs/workbench/contrib/chat/browser/chat.contribution.ts
```diff
@@ -969,7 +988,10 @@ configurationRegistry.registerConfiguration({
 			type: 'array',
 			items: { type: 'string' },
 			description: nls.localize('chat.agentHost.clientTools', "Tool reference names to expose as client-provided tools in agent host sessions."),
-			default: ['runTask', 'getTaskOutput', 'problems', 'runTests'],
+			default: [
+				'runTask', 'getTaskOutput', 'problems', 'runTests',
+				...browserClientTools,
```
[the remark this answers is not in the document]
jruales: non-agentic tool should be ok

src/vs/workbench/contrib/chat/browser/chat.contribution.ts
```diff
@@ -195,6 +195,25 @@ CommandsRegistry.registerCommand('_chat.notifyQuestionCarouselAnswer', (accessor
 const toolReferenceNameEnumValues: string[] = [];
 const toolReferenceNameEnumDescriptions: string[] = [];
 
+/**
+ * Tool reference names for the integrated browser tools.
+ * Always present in the {@link ChatConfiguration.AgentHostClientTools} default but
+ * only effective when the tools themselves are actually registered, which is currently
+ * feature-flag-gated.
+ */
+const browserClientTools = [
```
kycutler: If the agent host is enabled, but the new feature flag is not, will the agent still get these tools? Thinking it could be confusing if the agent starts opening browser tabs without the sharing UI (and potentially appearing in the wrong session)

kycutler: Maybe instead of adding these to the default, since we have a separate feature flag for it we could just conditionally inject the browser tools elsewhere

jruales: It won't. It would only get the non-agentic tool in that case. Only actually registered tools are seen by the agent. This list is more like a filter applied on the full list.

jruales: Updated to comment to be clearer about this.

bryanchen-d (dismissed):

kycutler (approved):

PR #313679, opened 2026-05-01
Copilot CLI with Agent Host should respect user shell from terminal profile when running shell commands 

### LINKED ISSUE #313160: Respect the user's preferred shell in agent host terminals
Today, host-managed terminals in the agent host (the PTY-backed `bash`/`powershell` tools in [`copilotShellTools.ts`](https://github.com/microsoft/vscode/blob/main/src/vs/platform/agentHost/node/copilot/copilotShellTools.ts)) hardcode the shell — on Windows that's always windows pwsh (pwsh 5) regardless of what the user has set up.
 If I have PowerShell 7 installed I think agent host terminal would still go to windows powershell (pwsh5)

Making the default better (well... some windows users are passionate git bash users all time)  and also respecting user's preferred shell (such as the one defined in setting) would be great.
We allow user to set per OS like: ` "terminal.integrated.defaultProfile.windows":`

- **(a)** Inherit the existing integrated-terminal settings — `terminal.integrated.defaultProfile.*`
-  The agent host can't read workbench config directly?? because it's a separate Node process, but the workbench-side `NodeAgentHostStarter` already has `IConfigurationService` and forwards data to the child via env vars and the AHP root-config bag — same channel `customizations` uses today. So the workbench resolves the profile via `ITerminalProfileResolverService` and pushes the resolved path/args down on startup and on setting change.
- **(b)** Add a dedicated `agentHost.defaultShell` setting. Simple and isolated, but creates a parallel surface users have to configure separately.
- **(c)** Both — dedicated setting wins if explicitly set, otherwise inherit from `terminal.integrated.*`.

Thought A would be the best for more "unified" experience.

anthonykim1:
Resolves: https://github.com/microsoft/vscode/issues/313160 

Existing behavior we already have with Copilot CLI in chat:

existingRespect
Shows that we respect user's default terminal profile setting.

--------------------------------------------------------------------------
Achievement with the PR: 

New terminal agent host terminal that will be first priority: 
agentHostProfileSupercedeDefaultProfile

If above is not set, we use user's default terminal profile: 
agentHostUseDefaultProfile

Allow Agent Host (local) to respect profile (more like path executable since args and env are ignored atm) : 
image


stasyu2009-ux (commented):
Він правильно працює але код символа треба доробити

src/vs/workbench/contrib/terminal/browser/terminalProfileResolverService.ts
```diff
@@ -189,6 +189,14 @@ export abstract class BaseTerminalProfileResolverService extends Disposable impl
 	}
 
 	private async _getUnresolvedDefaultProfile(options: IShellLaunchConfigResolveOptions): Promise<ITerminalProfile> {
+		// If agent host shell is allowed, prefer that
+		if (options.allowAgentHostShell) {
+			const agentHostShellProfile = this._getUnresolvedAgentHostShellProfile(options);
+			if (agentHostShellProfile) {
+				return agentHostShellProfile;
+			}
+		}
```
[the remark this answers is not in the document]
anthonykim1: This is resolved with https://github.com/microsoft/vscode/pull/313679/changes/da0dce6270418a2f89edaf040bd663510310da07

stasyu2009-ux (commented):
Вирішив деякі задачки.

src/vs/workbench/contrib/chat/browser/agentSessions/agentHost/agentHostTerminalContribution.ts
```diff
@@ -69,5 +94,33 @@ export class AgentHostTerminalContribution extends Disposable implements IWorkbe
...
+			});
+		} catch {
+			return;
+		}
+
+		if (!profile.path) {
+			return;
+		}
+
+		this._agentHostService.dispatch({
```
connor4312, line 146: We should make sure the agent host has a `AgentHostConfigKey.DefaultShell` in its `rootState.config.schema` prior to pushing down the config. You may need to wait until `onDidChange` for the root state to get hydrated.

We'd also need to dispatch to the connections in `IRemoteAgentHostService`, this is only the local service that you're listning to right now

src/vs/platform/terminal/common/terminalPlatformConfiguration.ts
```diff
@@ -189,7 +189,7 @@ const terminalPlatformConfiguration: IConfigurationNode = {
 		},
 		[TerminalSettingId.AgentHostProfileMacOs]: {
```
stasyu2009-ux: Код працює

stasyu2009-ux (commented):
Код зроблений

src/vs/workbench/contrib/chat/browser/agentSessions/agentHost/agentHostTerminalContribution.ts
anthonykim1: Thanks! this makes sense. 
Remote one is tricky - I think I'd need to figure out how to get remote profile resolver, since we can't just use `remoteAuthority`

The schema-gating piece has been addressed per feedback in https://github.com/microsoft/vscode/pull/313679/commits/40ba2c1626ddf9ddbe1b1515ab4653ed1c2faab7: 
* `_pushDefaultShell` now no-ops until the local host root config schema advertises `AgentHostConfigKey.DefaultShell`, retries when `rootState` hydrates/changes.
* has tests covering the hydrated/missing-schema cases.

I also clarified the setting wording in cf72e32 so it does not imply full profile fidelity or remote support.
* It now says this currently applies to the local agent host, and notes that remote agent hosts need remote-side shell configuration because local resolved paths may be invalid on the remote.

I split the remaining pieces into follow-ups so this PR can stay focused:

- (current PR) #313679: local agent-host executable/profile-path selection for Copilot CLI shell tools.
- #313790: profile `args`, and deciding whether/how to support `env`.
- #313809: remote agent-host shell config / `IRemoteAgentHostService`. 
     * This is split out because this PR resolves the profile in the local workbench context (`remoteAuthority: undefined`) and pushes that executable path to the local agent host. That is safe for local PTYs, but unsafe for remote agent hosts: a locally resolved path like `C:\Program Files\Git\bin\bash.exe` may be invalid on the remote machine that owns the PTY. VS Code’s normal terminal flow handles this by resolving profiles/default shells against the terminal backend/remote authority. Agent-host remotes do not have that resolver path yet, so remote support needs a remote-side config/resolution story rather than simply fanning out the local path.
- #313812: replacing the `AgentHostConfigKey.DefaultShell` magic key with a typed client→host channel.

@connor4312 Does that split match what you had in mind?

src/vs/workbench/contrib/chat/browser/agentSessions/agentHost/agentHostTerminalContribution.ts
```diff
@@ -69,5 +102,34 @@ export class AgentHostTerminalContribution extends Disposable implements IWorkbe
...
+
+		if (!profile.path || profile.path === this._pushedShell) {
+			return;
+		}
+		this._pushedShell = profile.path;
+
+		this._agentHostService.dispatch({
+			type: ActionType.RootConfigChanged,
+			config: { [AgentHostConfigKey.DefaultShell]: profile.path },
+		});
```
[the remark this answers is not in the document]
anthonykim1: filed issue.

dmitrivMS (approved):

sheetsync:
@anthonykim1  @connor4312 Thank you both enormously for this!  It's a small change, but it'll make a huge difference to us as we don't have modern Powershell running on our locked-down corporate build, so having the Agents UI respect the terminal choice from the IDE means (hopefully) that the Agents UI is actually usable.

I'm super pleased about this. Thanks again.

stasyu2009-ux (commented):
Код правильно працює

PR #313753, opened 2026-05-01
fix(chronicle): scope /chronicle:tips to current workspace

### LINKED ISSUE #313752: /chronicle:tips returns identical results across different workspaces — no workspace scoping
## Problem

Running `/chronicle:tips` in workspace A, then switching to workspace B and running it again returns **identical results**. The tips are not scoped to the current workspace — both invocations query the full global session store.

## Root cause

The `_handleTips` prompt in `chronicleIntent.ts` (line 384–409) instructs the model to query sessions from the last 7 days, but **never provides the current workspace folder path** and **never instructs the model to filter by `cwd`**.

The data model supports this — sessions have a `cwd` column populated with `vscode.workspace.workspaceFolders?.[0]?.uri.fsPath` during tracking (see `sessionStoreTracker.ts:217`). The schema description in `_getSchemaDescription()` even documents `cwd` as "workspace folder path". But the tips prompt doesn't use it.

In contrast, the schema description says:
> `cwd (workspace folder path)`

…yet the tips prompt just says:
> "query sessions from the last 7 days"

So the model generates unscoped SQL like `SELECT * FROM sessions WHERE updated_at >= datetime('now', '-7 days')` and gets the same global dataset regardless of workspace.

## Expected behavior

`/chronicle:tips` should provide workspace-relevant tips by default:
- Inject the current workspace folder path into the tips prompt
- Instruct the model to filter sessions by `cwd` matching the current workspace (or its `repository`/`branch`)
- Optionally allow a "global" mode for cross-workspace analysis

## Reproduction

1. Open workspace A (e.g., `~/projects/foo`)
2. Run `/chronicle:tips` — observe the results
3. Open workspace B (e.g., `~/projects/bar`)
4. Run `/chronicle:tips` — observe identical results

## Affected code

- `extensions/copilot/src/extension/intents/node/chronicleIntent.ts` — `_handleTips()` method (line ~384)
- The same issue likely affects `_handleFreeForm()` (line ~433) which also lacks workspace context

64johnlee:
## Summary

Fixes #313752 — `/chronicle:tips` returns identical results across different workspaces because the prompt never includes the current workspace folder path.

## Root Cause

The `_handleTips` prompt instructs the model to query sessions from the last 7 days, but **never provides the current workspace folder path** and never instructs the model to filter by `cwd`. The sessions table has a `cwd` column (workspace folder path) that is already documented in the schema description, but the tips prompt doesn't use it.

## Fix

1. Added `_getWorkspacePath()` helper that extracts `request.workspaceFolder?.uri?.fsPath`
2. When a workspace path is available, the prompt now includes:
   - An **IMPORTANT - Workspace Scoping** section explaining to filter by `cwd`
   - An example SQL filter: `WHERE cwd = '${workspacePath}'`
   - Updated instruction #1 to mention "in the current workspace"
   - Added workspace-specific query guideline for `cwd` filtering

## Testing

1. Open workspace A → run `/chronicle:tips` → note results
2. Switch to workspace B → run `/chronicle:tips` → should now show different, workspace-scoped results

### DIFF
--- a/extensions/copilot/src/extension/intents/node/chronicleIntent.ts
+++ b/extensions/copilot/src/extension/intents/node/chronicleIntent.ts
@@ -368,6 +368,10 @@ export class ChronicleIntent implements IIntent {
 		return {};
 	}
 
+	private _getWorkspacePath(request: vscode.ChatRequest): string | undefined {
+		return request.workspaceFolder?.uri?.fsPath;
+	}
+
 	private async _handleTips(
 		extra: string | undefined,
 		stream: vscode.ChatResponseStream,
@@ -380,6 +384,7 @@ export class ChronicleIntent implements IIntent {
 	): Promise<vscode.ChatResult> {
 		const hasCloud = this._indexingPreference.hasCloudConsent();
 		const schema = this._getSchemaDescription(hasCloud);
+		const workspacePath = this._getWorkspacePath(request);
 
 		let prompt = `You have access to the session_store_sql tool that can execute read-only SQL queries against the user's Copilot session database.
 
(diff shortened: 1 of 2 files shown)


digitarald:
Thank you for the contribution! Closing, as the skill now has the rest of the system prompt, so knows the current workspace which fixes the issue.

64johnlee:
Thanks for the context on the close, @digitarald!

One thing I noticed: the CSS change in this PR (`src/vs/workbench/contrib/chat/browser/widget/chatContentParts/media/chatQuestionCarousel.css`) — enabling `user-select: text` on `.chat-question-detailed-message`, `.chat-question-description`, and `.chat-question-carousel-message` — appears verbatim in `main` right now. The blob SHA matches exactly: `79ed08b2398f79826bdaaa3637338891e224458e`.

If those changes originated from this PR, would it be possible to get attribution — either by reopening and merging the CSS-only portion, or noting it in the relevant commit? If it's easier, I'm happy to open a focused PR just for that file.

Totally understand if it was applied independently — just wanted to flag it!

Issue #299952, opened 2026-03-07
Copilot chat OTel: consider making JSON truncation configurable


JamesNK:
Copilot chat OTel truncates large JSON strings. I believe it is hardcoded to 64KB. Some telemetry systems have maximum attribute sizes and require a limit. However, systems without a maximum attribute size would benefit from getting the full JSON payload so all message content can be rendered in the UI.

Truncating JSON means that some messages aren't available and can prevent all messages from being viewed because the string is no longer valid JSON.

Consider making JSON truncation configurable to give users control and allow full content can be sent when possible.

cc @zhichli

Rilomilo:
I open a pr
https://github.com/microsoft/vscode-copilot-chat/pull/5072

connor4312:
This didn't seem to do much for me. I set my limit to `10` and then made a couple requests. I do not see anything in otel logs about data getting truncated as I would expect from the `truncateForOTel` function.

https://gist.github.com/connor4312/7b8890c38a439bfadf76448a8f2b1089

PR #313855, opened 2026-05-02
feat(otel): make attribute truncation configurable

zhichli:
Adds `github.copilot.chat.otel.maxAttributeSizeChars` setting (and `COPILOT_OTEL_MAX_ATTRIBUTE_SIZE_CHARS` env var) to control truncation of OTel content attributes. Default `0` = unlimited (matches [OTel spec](https://opentelemetry.io/docs/specs/otel/common/#attribute-limits) `Infinity`).

Refs #299952


zhichli (commented):
Addressed all 4 Copilot reviewer findings in 6fef3dffff3:

1. ✅ **`parseMaxAttributeSizeChars`** now uses `Number.isSafeInteger` — rejects fractions and unsafe values. Test added.
2. ✅ **Chars vs UTF-16** wording reworded in arch doc to be accurate (UTF-16 code units, astral characters count as 2).
3. ✅ **Suffix-longer-than-maxLength** edge case fixed with hard-cut fallback. Test added.
4. ✅ **`toolCallingLoop.ts`** missed call sites — `INPUT_MESSAGES` attribute and `user_message` event content now both respect `maxAttributeSizeChars`.

bpasero (approved):

PR #313921, opened 2026-05-03
Implement Agent Host OTel tracing

### LINKED ISSUE #314788: Implement Agent Host OTel

roblourens:
## Summary

Adds Agent Host OpenTelemetry support with trace propagation from chat frontend work through Agent Host, provider lifecycle, Copilot SDK calls, hooks, tools, and persistence work. This includes the default agent-focused trace data plus verbose tracing for frontend/host internals when enabled.

Key pieces:
- Add Agent Host OTel service abstractions, OTLP export, trace context formatting, no-op/in-memory implementations, and Node async-local trace context handling.
- Wire Agent Host OTel settings/env vars through browser, electron, remote protocol, and utility process startup.
- Configure Copilot SDK telemetry and W3C trace context propagation so SDK-native LLM/tool/hook spans correlate with host spans.
- Add verbose spans for Agent Host lifecycle, session DB work, config persistence, create/list/restore flows, and every host-owned call into the Copilot SDK with call/reason attributes.
- Parent Agent Host-owned shell tool handlers under SDK tool spans and keep post-turn hooks such as agentStop/sessionEnd under the active agent invocation.
- Update monitoring docs for settings, span hierarchy, SDK-call spans, and trace propagation behavior.

## Validation

- Focused Agent Host/OTel unit suite: 230 passing
- Hygiene on changed files
- `git diff --check`

(Written by Copilot)


src/vs/platform/agentHost/node/copilot/copilotAgentSession.ts
```diff
@@ -561,8 +608,67 @@ export class CopilotAgentSession extends Disposable {
 		}
 	}
 
+	private _startAgentSpan(prompt: string, attachments: readonly IAgentAttachment[] | undefined, turnId: string | undefined, mode: CopilotSdkMode | undefined): IAgentHostSpanHandle {
+		this._endAgentSpan();
+		const span = this._agentHostOTelService.startSpan('invoke_agent copilotcli', {
```
zhichli, line 613: One feedback on naming: the Agent Host's `invoke_agent copilotcli` and the CLI SDK's native `invoke_agent` are both GenAI-conventional, which makes the trace view ambiguous. Renaming the Agent Host wrapper to something like `vscode_agent_host.invoke_agent copilotcli` would reserve `invoke_agent` for the actual SDK-native span and keep host spans greppable as `vscode_agent_host.*`.

zhichli:
**Suggestion: define the "exported for customers" contract via GenAI semconv, gate the rest behind `verboseTracing`.**

Same set the Copilot extension uses today (`EXPORTABLE_OPERATION_NAMES` in `otelServiceImpl.ts`):

- **Always export:** `invoke_agent`, `chat`, `execute_tool`, `execute_hook`, `embeddings` — anything in the `gen_ai.*` namespace.
- **Verbose-only:** anything in the `vscode_agent_host.*` namespace (`db.*`, `sdk.*`, `create_session`, `materialize_provisional`, the workbench `chat.agent_host.*` wrappers, the `tool_handler` shadows of SDK `execute_tool`).

That keeps the customer-facing surface consistent with the four other agent execution paths in the extension and gives `verboseTracing` a clean one-line filter rule (export iff `gen_ai.operation.name` is set).

zhichli:
**Suggestion: use the OTel SDK on both sides, keep IPC as the transport.**

Renderer: `@opentelemetry/sdk-trace-web` + a small custom `SpanExporter` whose `export()` ships OTLP via `emitOTelSpan`.

Agent Host: `@opentelemetry/sdk-trace-node` + `@opentelemetry/exporter-trace-otlp-http` — same shape as the Copilot extension's `NodeOTelService`, where `OTLPTraceExporter` + `BatchSpanProcessor` handle the actual POST (with batching/retries/gzip) instead of a per-span `fetch()`.

We get standard `propagation.inject`/`extract`, async context, and resource attributes for free, instead of maintaining the in-memory span impl + hand-rolled traceparent + JSON OTLP serializer + per-span fetch on either side. The IPC transport between renderer and Agent Host stays exactly the same.

zhichli:
**Suggestion: surface a SQLite span sink for Agent Host spans.**

The Copilot extension already has `chat.otel.dbSpanExporter.enabled` + `OTelSqliteStore` (`extensions/copilot/src/platform/otel/node/sqlite/`) which lands GenAI spans in a local SQLite DB and backs the **Chat: Export Agent Traces DB** command.

Today this only sees spans produced inside the extension host. Agent Host spans go straight to OTLP and are invisible to anyone using the DB-export path.

The exporter and store live in `extensions/copilot/`, so the Agent Host process can't import them directly. Reasonable paths:

- **Re-implement on the Agent Host side.** ~100 lines using the same `node:sqlite` schema, written into a separate file, opened in WAL mode so concurrent writers from the extension don't conflict.
- **Promote `OTelSqliteStore` + `SqliteSpanExporter` to `src/vs/platform/`** so both processes share one impl and one schema.

Either way, if `AgentHostOTelService` adopts `@opentelemetry/sdk-trace-node` (per the SDK comment above), the wiring is one `addSpanProcessor(new BatchSpanProcessor(new SqliteSpanExporter(store)))` next to the OTLP exporter — same pattern as `otelServiceImpl.ts:131-134`.

Suggested gating: a `chat.agentHost.otel.dbSpanExporter.enabled` setting that mirrors the extension's, plus an optional path override env var so deployments that want one DB across both processes can point both writers at the same file.

zhichli:
Recap of what I'd like to see resolved before merge vs. tracked as follow-up — re-reading the thread so nothing falls through:

**In this PR (merge-blockers, all already raised inline):**

1. `traceparent` / `tracestate` runtime type-checks in `protocolServerHandler.ts:198/486` (Copilot AI nit).
2. Replace the per-span `fetch()` in `AgentHostOTelService` with `@opentelemetry/sdk-trace-node` + `OTLPTraceExporter` + `BatchSpanProcessor`. This also fixes the no-timeout / no-batching concern Copilot AI flagged and gives standard `propagation.inject/extract`, async context, and resource attributes for free.
3. Span-name disambiguation: rename the host wrapper `invoke_agent copilotcli` → `vscode_agent_host.invoke_agent copilotcli` so the SDK-native `invoke_agent` is the unambiguous root for any downstream consumer keying off `gen_ai.operation.name`.
4. Define the "exported for customers" contract via GenAI semconv (export iff `gen_ai.operation.name` is set), and gate everything in the `vscode_agent_host.*` namespace behind `verboseTracing`. Same shape as `EXPORTABLE_OPERATION_NAMES` in `extensions/copilot/src/platform/otel/node/otelServiceImpl.ts`.

**Follow-up (filed as #315901):** SQLite span sink + `chat.agentHost.otel.dbSpanExporter.enabled` setting + env-var path override + WAL + export command. Symmetric counterpart to the extension's existing `OTelSqliteStore` / `Chat: Export Agent Traces DB`. Best landed after (2) above so the sink can plug in as a second `BatchSpanProcessor` exporter rather than a hand-rolled write path.

PR #313944, opened 2026-05-03
Fix new window opening with title bar or status bar outside visible area on small/scaled displays

### LINKED ISSUE #313942: New Window places title bar and status bar outside the visible work area (regression from #262266)
### Summary

Opening a new window with a folder or workspace places the title bar above the top of
the screen and the status bar behind the taskbar. This is a regression from #262266,
which raised the default workspace window size to 1440x900 without clamping it to the
display work area.

### Steps to reproduce

The key requirement is a display whose **work area is shorter than 900px**. On Windows
that includes any 1080p panel at 125% scaling, which leaves a logical work area of
1536x816 once the taskbar is subtracted.

1. Set a 1920x1080 display to 125% scaling
2. Open a **folder** in VS Code (not an empty window)
3. Press Ctrl+Shift+N

Expected: the new window sits fully inside the visible work area.
Actual: the title bar is above the top of the screen and the status bar is behind the
taskbar.

### Note: an empty window will look fine

Empty windows use `DEFAULT_EMPTY_WINDOW_SIZE` (1200x800), which fits an 816px work area
with only minor overflow. The visible failure needs `DEFAULT_WORKSPACE_WINDOW_SIZE`
(1440x900), so the window must have a folder or workspace open. Testing with an empty
window is likely to look correct, which may be why this has been hard to reproduce.

### The arithmetic

`doGetNewWindowState()` centers the window on `displayToUse.bounds`, which is the full
screen rectangle including the taskbar:

```ts
state.x = Math.round(displayToUse.bounds.x + (displayToUse.bounds.width / 2) - (state.width! / 2));
state.y = Math.round(displayToUse.bounds.y + (displayToUse.bounds.height / 2) - (state.height! / 2));
```

With logical bounds of 1536x864 and a default height of 900:

```
y = 0 + 432 - 450 = -18
```

The window starts 18px above the top of the screen, and its bottom edge lands at 882
against a work area that ends at 816, leaving 66px behind the taskbar.

### A second display makes it considerably worse

`ensureNoOverlap()` then offsets the window by 30px whenever it shares an x or y
coordinate with any existing window:

```ts
while (existingWindowBounds.some(bounds => bounds.x === state.x || bounds.y === state.y)) {
    state.x += 30;
    state.y += 30;
}
```

The condition is `||`, it compares against windows on all displays rather than only the
target display, and there is no re-clamping afterwards. Displays arranged side by side
share the same vertical origin and height, so default-positioned windows on either
display compute the same y. Collisions become near certain once a second display is
connected, and each one pushes the window a further 30px down and right.

To reproduce this part:

1. Connect a second display, arranged side by side
2. Open a folder window on each display
3. Press Ctrl+Shift+N a few times

### Affected configurations

Any display whose logical work area is shorter than 900px, for example:

| Display | Scaling | Work area height |
|---|---|---|
| 1920x1080 | 125% | ~816px |
| 1366x768 | 100% | ~728px |
| 1440x900 | 100% | ~860px |

### Related

#317762 is the same regression reached through `nativeHostService.openAgentsWindow()`.

### Fix

#313944 clamps the default size and position to `displayToUse.workArea`, and bounds the
`ensureNoOverlap()` offsetting to the same work area.

mx-92x:
Fixes #313942

## Problem
The default workspace window size (1440×900) introduced in #262266 was never clamped to the display work area. On displays where the work area height is less than 900px logical pixels (e.g. 1366×768 laptops, or 1920×1080 at 125% DPI scaling), the new window overflows the visible area — pushing the title bar above the screen and/or the status bar behind the taskbar.

## Fix
Switch new window positioning from `displayToUse.bounds` to `displayToUse.workArea` and clamp the default width/height to fit within the actual visible area.

## Testing
Verified on a 1920×1080 display at 125% DPI scaling — new window now opens fully within bounds with both title bar and status bar accessible.


mx-92x:
For reference: I filed #317762 documenting a second symptom of this same regression. Clicking "Open in Agents" hits the same `doGetNewWindowState()` code path and produces the same offscreen-chrome behavior on small/scaled displays. The fix in this PR covers both entry points.

mx-92x:

mx-92x:
Update: while testing this on a dual-monitor setup I found a second code path the
original fix did not cover, and I have pushed another commit for it.

The bug reproduces reliably when a second display is connected, and only
intermittently on a single display. The reason is `ensureNoOverlap()`, which runs
after the work area clamp and undoes it:

```ts
while (existingWindowBounds.some(bounds => bounds.x === state.x || bounds.y === state.y)) {
    state.x += 30;
    state.y += 30;
}
```

Three things combine here:

1. The condition is `||`, so a match on x alone or y alone is enough to trigger an offset.
2. It compares against every window on every display, not just windows on the target display.
3. There is no re-clamping after the loop.

Displays arranged side by side share the same vertical origin and height, so every
default-positioned window on either display computes the same y. A window on the
first display therefore guarantees a y collision for a new window on the second, the
loop runs repeatedly, and the accumulated offset pushes the title bar above the work
area and the status bar behind the taskbar.

On my setup (two 1920x1080 panels at 125%, giving a logical work area of 1536x816) a
workspace window already starts at y = -18 before any offsetting, since the default
height of 900 exceeds the 816px work area. Each subsequent collision adds another 30px.

The new commit passes the work area into `ensureNoOverlap()` and stops offsetting once
the next step would push the window outside it. Behaviour is unchanged when there is no
collision. In the worst case a new window now overlaps an existing one instead of moving
off screen, which seemed the right trade since the offsetting is cosmetic while staying
on screen is not.

Verified locally on the setup described above, with the source window both maximized
and not maximized.

mx-92x:
@benibenj thanks for consolidating #317762 into #313942, and for the note that this PR
resolves both paths.

Quick summary of the change in case it is useful:

- Clamps the default window size and position to `displayToUse.workArea` instead of
  `displayToUse.bounds`, falling back to `bounds` when `workArea` reports zero, which can
  happen on Linux/X11.
- Bounds the `ensureNoOverlap()` offset loop to the same work area, so the repeated +30
  offset cannot push the chrome back outside the visible region.
- 30 lines in one file, `windowsStateHandler.ts`. No new dependencies, no API changes.

Verified on a dual-monitor Windows 11 setup at 125% scaling, with the source window both
maximized and not.

The workflows are still awaiting approval so CI has not run yet. Happy to rework anything
if the approach needs changing, and equally happy to wait if it is simply queued.

PR #313954, opened 2026-05-03
Only install Appx package when context menu is selected

YexuanXiao:
The current installation script unconditionally installs the Appx, causing Windows to always run an additional dllhost process, even when the user chooses not to add the context menu.


YexuanXiao:
@deepak1556 Can you take a look?

YexuanXiao:
@joaomoreno

deepak1556:
Thank you, the reason we removed the context menu gate on the appx package was because it is also used now to support package identity, reference https://github.com/microsoft/vscode/pull/257741 

However, we still guard on context menu entry itself https://github.com/YexuanXiao/vscode/blob/84528b68360a73b6d1de4bf9dac07859500250e4/build/win32/code.iss#L1290 despite the appx being always installed, runtime check at https://github.com/microsoft/vscode-explorer-command/blob/676a2a43194e5821af2642926bc0eef9f6edd13f/src/explorer_command.cc#L88-L123

PR #314005, opened 2026-05-04
Fix listener leak when rerendering title file widgets in chat content parts

ishaq2321:
## Summary

When a chat content part's title changes and contains file links, `renderFileWidgets()` is called to render file link widgets. Previously, widgets registered during the previous render were not cleaned up before adding new ones, causing a listener leak that accumulated over time.

## Changes

- **ChatCollapsibleInputOutputContentPart**: Added `_titleFileWidgetStore` (a `DisposableStore`) that is cleared and repopulated when the title setter is called, ensuring previous widgets are properly disposed.

- **ChatSubagentContentPart**: Applied the same fix - added `_titleFileWidgetStore` to track title widgets and clear them before re-rendering.

- **Test**: Added regression test that verifies the fix by counting `register` and `dispose` calls on the markdown anchor service.

## Testing

- Added unit test that creates a `ChatCollapsibleInputOutputContentPart` with a title containing a file link, updates the title, and verifies that the previous widget's `dispose()` was called when the new title is set.

## Related Issues

- Follows the same pattern as the IPC server listener leak fix (#293913) and terminal execute strategy DisposableStore leak fix (#313369)

### DIFF
--- a/src/vs/workbench/contrib/chat/browser/widget/chatContentParts/chatSubagentContentPart.ts
+++ b/src/vs/workbench/contrib/chat/browser/widget/chatContentParts/chatSubagentContentPart.ts
@@ -132,6 +132,7 @@ export class ChatSubagentContentPart extends ChatCollapsibleContentPart implemen
 	private titleShimmerSpan: HTMLElement | undefined;
 	private titleDetailContainer: HTMLElement | undefined;
 	private readonly _titleDetailRendered = this._register(new MutableDisposable<IRenderedMarkdown>());
+	private readonly _titleFileWidgetStore = this._register(new DisposableStore());
 
 	/**
 	 * Check if a tool invocation is the parent subagent tool (the tool that spawns a subagent).
@@ -517,6 +518,7 @@ export class ChatSubagentContentPart extends ChatCollapsibleContentPart implemen
 
 		// Dispose previous detail rendering
 		this._titleDetailRendered.clear();
+		this._titleFileWidgetStore.clear();
 
 		if (!toolCallText) {
 			if (this.titleDetailContainer) {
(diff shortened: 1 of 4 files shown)


dmitrivMS:
@ishaq2321 Thank you so much for this PR and for identifying the problem!
Since the code has diverged a little, I have created a bit cleaner version here: 
https://github.com/microsoft/vscode/pull/319176

ishaq2321:
@dmitrivMS Thanks for the clean implementation and for carrying it across the finish line — glad the fix made it in!

I've been exploring the VS Code codebase and would love to contribute more regularly. If there are open issues you think need attention, or areas where a fresh set of eyes would help, feel free to tag me or point me in the right direction. Happy to dig in.

dmitrivMS:
@ishaq2321 Honestly, there's been OOM issues when using chat/agents, so anything you can contribute to cleaning that up would be appreciated.
If not, what are the areas that interest you?

ishaq2321:
@dmitrivMS  OOM issues in chat/agents sounds like exactly the kind of thing I enjoy digging into — memory profiling, listener cleanup, widget lifecycle edge cases. Happy to take a look.

If it's easier, feel free to assign me specific issues directly — that way I can focus without duplicating work others might already be on. I'm also familiar with the Microsoft TypeScript compiler repository, so if anything crosses over there, I can help on that side too.

One thing that helps me contribute effectively: early feedback on PRs — even a quick "looks like the right direction" or "rethink this part" — so I can iterate fast rather than going quiet for a week and losing context. Happy to work however suits the team best.

dmitrivMS:
@ishaq2321 We don't really do that (assign issues), but here's the query: https://github.com/microsoft/vscode/issues?q=is%3Aissue%20state%3Aopen%20(label%3Aperf-startup%20OR%20label%3Aperf-bloat%20OR%20label%3Aperf-profile%20OR%20label%3Aperf)&page=2
Many issues will look like "OOM, we crashed", so yeah :-)

Issue #313904, opened 2026-05-02
Copilot Chat fails with OpenAI-compatible DeepSeek/Kimi models in VS Code Insiders when tool use triggers 400 reasoning_content error


PatrykMiziukBC:
Type: <b>Bug</b>

1. Install VS Code Insiders and open Copilot Chat.
2. Configure an OpenAI-compatible model provider in Copilot using an API key from OpenCode Go.
3. Add models such as DeepSeek v4 Pro and Kimi K2.6.
4. Select one of those models in Copilot Chat.
5. Send a prompt that causes the model to use tools, for example: create simple Go app.
6. Wait for the response to fail.

Actual result:
Copilot Chat shows an error like:
Sorry, your request failed. Please try again.
Reason: Request Failed: 400 {"error":{"message":"Error from provider (DeepSeek): The reasoning_content in the thinking mode must be passed back to the API.","type":"invalid_request_error","code":"invalid_request_error"}}

Expected result:
The model should complete the request normally, including tool use, without failing with a 400 error.

Extra note
The failure pattern matches reports that DeepSeek reasoning/thinking mode requires reasoning_content from prior assistant output to be sent back unchanged, and similar breakage has been reported specifically around tool calls and agent-style loops

VS Code version: Code - Insiders 1.119.0-insider (1d94ae1b8a7df7bfd2d7bc8aa112976bba19205c, 2026-05-01T21:36:56Z)
OS version: Windows_NT x64 10.0.22631
Modes:

<details>
<summary>System Info</summary>

|Item|Value|
|---|---|
|CPUs|12th Gen Intel(R) Core(TM) i7-1255U (12 x 2611)|
[...]

Issue #312746, opened 2026-04-27
Bug: Copilot Chat drops reasoning_content field when using DeepSeek models (OpenRouter BYOK), causing 400 Error


DFVSQY:
- Copilot Chat Extension Version:
- VS Code Version:
- OS Version:
- Feature (e.g. agent/edit/ask mode):
- Selected model (e.g. GPT 4.1, Claude 3.7 Sonnet):
- Logs:

When using DeepSeek reasoning models (e.g., DeepSeek R1 / V4 Flash) via the BYOK (Bring Your Own Key) feature with OpenRouter, Copilot Chat fails with a 400 Bad Request error when it tries to engage in multi-turn conversations or use workspace tools.

Steps to Reproduce:
1. Configure Copilot Chat to use OpenRouter API via BYOK.
2. Select a DeepSeek model that supports reasoning/thinking mode.
3. Ask a prompt that requires Copilot to use tools (e.g., "Analyze this project" or "@workspace what does this do?").
4. Copilot reviews files ("Reviewed 2 files") and then immediately throws an error.

<img width="718" height="979" alt="Image" src="https://github.com/user-attachments/assets/4f7b8830-db75-44b3-9948-9459c85eba79" />

karsey24:
I've encountered this as well.

atirutw:
It's a little sporadic for me, but I'm also able to reproduce this at least 50/50 of the times with DeepSeek V4

CharlesZiy:
> It's a little sporadic for me, but I'm also able to reproduce this at least 50/50 of the times with DeepSeek V4

Because the API won't check the existence of `reasoning_content` while there isn't tool calling between user chats

Patryk676:
I've got the same issue but in Visual Studio Code Insiders -> OpenAI compatible API -> OpenCode Go API key -> models: deepseek v4 pro, kimi k2.6

DeBelserArne:
Same issue here with DeepSeek, Kimi, and Minimax models in VS Code Insiders using the OpenAI-compatible API with the Opencode API key.

What’s interesting is that GLM 5.1 works perfectly fine under the exact same configuration and setup. 

```
Sorry, your request failed. Please try again.

Copilot Request id: f04bd835-47b8-4b97-8b1e-72b44191268c

[...]

PR #314019, opened 2026-05-04
Preserve reasoning_content for BYOK DeepSeek thinking mode

checkmode:
## Problem
DeepSeek thinking-mode endpoints reject follow-up tool-call rounds with:

> "The reasoning_content in the thinking mode must be passed back to the API."

[DeepSeek thinking_mode](https://api-docs.deepseek.com/guides/thinking_mode)

The previous assistant turn's reasoning was being dropped between rounds.

Related reports: #312746, #313904.

## Cause
Two issues on the BYOK CAPI path:

1. `LanguageModelAccessPrompt` only kept the **first** `LanguageModelThinkingPart`
   in an assistant turn. Streaming providers emit reasoning across multiple
   parts and the `id` typically arrives in the final part (per
   `vscode.proposed.languageModelThinkingPart.d.ts`), so first-wins drops both
   trailing text and the id.
2. `OpenAIEndpoint.createRequestBody` only set `cot_id` / `cot_summary`
   (Copilot CAPI fields) and never emitted DeepSeek's `reasoning_content`.

## Fix
- Aggregate all `ThinkingPart`s in an assistant message: concatenate `text`,
  last-wins for `id` / `metadata`.
- Emit `reasoning_content` whenever the thinking callback receives non-empty
  text. Same assumption already made for `cot_id` / `cot_summary`: unknown
  JSON fields are tolerated by OpenAI-compatible endpoints.
- Plumb `reasoning_content` through the inbound SSE delta types
  (`RawThinkingDelta`, `getThinkingDeltaText`) so DeepSeek's reasoning chunks
  surface as `delta.thinking.text`.

## Tests
- New unit test in `stream.sseProcessor.spec.ts` covering DeepSeek-style
  `reasoning_content` SSE chunks.
- Manually verified against a DeepSeek thinking-mode endpoint with a
  multi-round tool-call conversation; the 400 no longer occurs.

## Known limitation
A single assistant turn may contain multiple independent thinking blocks
(e.g. Anthropic interleaved thinking around tool calls). They are currently
merged into one blob, which is correct for `reasoning_content` echo but loses
per-block signatures. Splitting requires a structural change to `ThinkingData`
and is out of scope.

### DIFF
--- a/extensions/copilot/src/extension/byok/node/openAIEndpoint.ts
+++ b/extensions/copilot/src/extension/byok/node/openAIEndpoint.ts
@@ -257,9 +257,16 @@ export class OpenAIEndpoint extends ChatEndpoint {
 		} else {
 			// Handle CAPI: provide callback for thinking data processing
 			const callback: RawMessageConversionCallback = (out, data) => {
-				if (data && data.id) {
+				if (!data) {
+					return;
+				}
+				const text = Array.isArray(data.text) ? data.text.join('') : data.text;
+				if (data.id) {
 					out.cot_id = data.id;
-					out.cot_summary = Array.isArray(data.text) ? data.text.join('') : data.text;
+					out.cot_summary = text;
+				}
+				if (text) {
+					out.reasoning_content = text;
 				}
 			};
 			const body = createCapiRequestBody(options, this.model, callback);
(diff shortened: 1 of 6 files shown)


checkmode:
@microsoft-github-policy-service agree company="KylinSoft"

checkmode:
@lramos15 @sbatten Any suggestion？

gdias1992:
@sbatten @lramos15 Any sugesstion？

DeBelserArne:
When will this be available to test in the Insiders build?

PR #314027, opened 2026-05-04
Pass session capabilities to chat parser

roblourens:
When parsing outgoing chat requests, use the contributed chat session capabilities as the attachment capability context, matching the remote session restore path. This keeps slash prompt commands highlighted correctly for agent-host sessions when messages are first sent, not only after reopening the session.

Adds targeted coverage for both live send and restored history parsing paths. (Written by Copilot)


roblourens:
@copilot review

(Written by Copilot)

osortega (approved):

PR #314082, opened 2026-05-04
:bug: Fix add file/selection to chat editor not working

Caffeine19:
## Problem

When chat is open as an **editor** (not sidebar), the "Add File to Chat" command (`workbench.action.chat.attachFile`) fails to attach the file. The file icon stays as `+` instead of switching to `x` (attached state). The focus does transfer to the chat editor, but the file is never added.

When chat is in the **sidebar**, it works correctly.

Video here:

https://github.com/user-attachments/assets/7bbf47e0-54f0-4054-8b65-f10e2b1c6003

## Root Cause

`AttachResourceAction.run()` calls `withChatView()` which calls `revealWidget()` **without** `preserveFocus`. When chat is an editor, this switches `editorService.activeEditor` from the source file editor to the chat editor. Downstream `_getResources()` then reads `editorService.activeEditor` and finds the chat editor (not a file scheme), so it returns an empty array — no file gets attached.

When chat is in the sidebar, revealing the sidebar view does not change `editorService.activeEditor`, so the file URI is still resolved correctly.

## Fix

Pass `preserveFocus: true` to `revealWidget()` so the source file editor remains the active editor during resource resolution. The widget's `focusInput()` call in `runWithWidget` transfers focus to the chat input afterwards, so the user experience is unchanged.

After fix, video here:

https://github.com/user-attachments/assets/a4d4f982-e456-409b-95e1-8a08281b270c


Caffeine19:
@hediet Hi, could you please give me a code review? 🫠

Caffeine19:
> Henning Dieterichs (@hediet) Hi, could you please give me a code review? 🫠

Ping @hediet

PR #314141, opened 2026-05-04
Handle inputNeeded in execution subagent to prevent looping

### LINKED ISSUE #308048: execution subagent doesn't work with new async terminal stuff

meganrogge:
When `run_in_terminal` detects a command is waiting for input, it tells the agent to use `get_terminal_output`/`send_to_terminal`. But the execution subagent no longer has those tools (removed in #313067), causing it to loop endlessly.

## Changes

1. **runInTerminalTool.ts**: Add `inputNeeded` to `toolMetadata` so consumers can detect this state.
2. **executionSubagentToolCallingLoop.ts**: Detect `inputNeeded` in `collectBackgroundCommands` — treat it the same as a timeout/async command so the subagent stops and returns to the main agent.
3. **executionSubagentTool.ts**: Append a note for input-needed commands in the subagent response so the main agent knows to handle them with its full tool set.

Fixes https://github.com/microsoft/vscode/issues/308048


meganrogge:

vikramnitin9:
@meganrogge - this PR looks good to me (although I am yet to test it with local debugging). However, I don't think the issue that reported looping (https://github.com/microsoft/vscode-internalbacklog/issues/7604) is caused by this. That issue seems like something more fundamental with the `run_in_terminal` tool itself.

Regarding execution subagent, when I've tested commands that wait on input, the subagent usually re-runs the command with a pipe (like `echo "Input" | command`), sidestepping the issue. Your solution in this PR is more elegant though, and I think we should go with this.

meganrogge:

dmitrivMS (approved):

PR #314150, opened 2026-05-04
Model picker styling updates

pwang347:
- Fix styling in welcome view of agents app
- Remove dividers in inline chat
- Remove thinking effort configuration in the hover


src/vs/workbench/contrib/chat/browser/widget/input/chatModelPicker.ts
```diff
@@ -163,7 +163,6 @@ function createModelAction(
 	languageModelsService: ILanguageModelsService,
 	section?: string,
 ): IActionWidgetDropdownAction & { section?: string } {
-	const toolbarActions = languageModelsService.getModelConfigurationActions(model.identifier);
```
pwang347, line 166: This is only being used by the thinking effort at the moment, we should revisit this in new design if we need a general-purpose config rendering area like this at all.

anthonykim1 (approved):

PR #314190, opened 2026-05-04
sessions: improve sub-session tab visibility with pill style

romalpani:
## Summary

Sub-session tabs in the chat composite bar (the per-chat tab strip inside an active session) were hard to distinguish from each other. The active tab in particular only had a barely-visible 5% background tint, and adjacent tabs had no spatial separation.

This PR updates `chatCompositeBar.css` to give each tab a clear pill-shaped boundary and make the active tab stand out clearly.

## Changes

- Add a 4px gap between tabs for clear separation
- Slightly increase border radius (4px → 6px) and padding for a more pronounced pill shape
- Reserve a transparent border on inactive tabs so layout stays stable when the active border appears
- Show a subtle background tint on hover (previously only color changed)
- Strengthen the active tab background (5% → 10%) and add a visible border so the active tab stands out clearly

## Before / After

Before: tabs blend together; active tab nearly indistinguishable from inactive ones.
CleanShot 2026-05-04 at 16 25 28@2x

After: each tab is a clearly bounded pill; active tab has a stronger fill and visible border.
CleanShot 2026-05-04 at 16 25 04@2x

### DIFF
--- a/src/vs/sessions/browser/parts/media/chatCompositeBar.css
+++ b/src/vs/sessions/browser/parts/media/chatCompositeBar.css
@@ -53,7 +53,7 @@
 .chat-composite-bar-tab.active {
 	color: var(--chat-tab-active-foreground);
 	background-color: color-mix(in srgb, var(--vscode-agentsPanel-foreground, var(--vscode-foreground)) 10%, transparent);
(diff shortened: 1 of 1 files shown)


romalpani:
Reimplemented

PR #314240, opened 2026-05-04
Special profile for Agents Window

sandy081:
## Summary
Adds a dedicated "Open in Agents" titlebar entry and aligns the agents window with its own internal profile.

## What Changed
- adds the desktop titlebar "Open in Agents" widget in the chat agent sessions contribution
- routes agents window opening through the dedicated agents window flow instead of the older sibling-app-specific path
- creates a built-in internal Agents profile and hides internal profiles from profile UI, editors, and profile-count telemetry
- removes the old desktop-specific Open in Agents action implementation

## Why
This keeps the Agents window isolated from regular user profiles while simplifying the desktop entry point for opening it.


lostintangent (approved):

sandersn:
Seems likely that this is the cause of #314771

PR #314249, opened 2026-05-04
consider a git init on an empty folder to still be empty

amunger:
This will allow the create_workspace_tool to run with a folder that has already been init'd for git, but nothing else. helpful for our evaluation framework.


amunger:
~requires-eval-assessment vscbench 

```yaml 
 "jupyter.interactiveWindow.notebookRepl": true
```

bhavyaus (approved):

PR #314277, opened 2026-05-04
Fix background todo agent behavior on main agent completion

vritant24:
Reworks the background todo processor to use a small drain queue for regular passes and final review. Regular passes still coalesce, but final review is now queued separately and drained once per turn after regular work, so it can run reliably even when the processor is idle.


bhavyaus (approved):
LGTM!

benvillalobos (approved):

PR #314288, opened 2026-05-04
Try to prevent findFiles2 from hanging on virtual file systems

### LINKED ISSUE #314285: FindFIles2 can stall forever with virtual file system with no search provider
For #313281 

1. Install Joh's memfs extension 
2. Close all workspace
3. Run `Memfs: setup workspace` and then `memfs: create files`
4. Try doing a call to `vscode.findFiles2` with any args

**Bug**
This call stalls forever

It seems to be stuck waiting for a search provider to be registered, but this will never happen

mjbvz:
Fixes  #314285

There's definitely a bug with the current code. However I'm not sure this fix is correct either. What it does is assume that after `onSearch:SCHEME` is fired, we will have the correct search provider (or not) instead of setting up the deferred promise maps and waiting for future providers that may never come in

That sounds correct to me but I'm not sure if there are complicated cases with registrations that this will mess up


mjbvz:
@roblourens Could use your thoughts on what the proper fix here is. I think the deferral code got added to fix the smoke tests, but it can cause this infinite hang on basic virtual file systems

roblourens:
If I run this code and connect to a remote EH and start a search, will it wait and return results instead of returning immediately? If so, then I think the fix is fine with me

roblourens:
If that's still an issue, I think we can find something else to wait on rather than wait forever

harshagarwalnyu:
The new `waitForSearchProvider` approach is much cleaner — bounded, no infinite hang. One race to flag for remote extension hosts before this merges.

---

**Problem: `onDidChangeExtensionsStatus` resolves too eagerly**

```ts
store.add(this.extensionService.onDidChangeExtensionsStatus(() => {
    if (this.extensionService.activationEventIsDone(activationEvent)) {
        done(this.getSearchProvider(type).get(scheme)); // may be undefined
        // store.dispose() kills _onDidRegisterProvider listener
    }
}));
```

For a remote EH, `activationEventIsDone` flips to `true` when the extension host process signals completion — but `$registerFileSearchProvider` is a fire-and-forget RPC that still has to traverse the wire. Sequence:

```
activationEventIsDone = true
→ onDidChangeExtensionsStatus fires
→ done(undefined) called, store disposed
→ $registerFileSearchProvider RPC arrives (too late)
→ _onDidRegisterProvider.fire() hits no listeners
→ false empty result
```

For a local EH the RPC processes synchronously before `onDidChangeExtensionsStatus` fires, so in practice this is fine.

---

**Proposed fix:** in the status handler, use `disposableTimeout` for a bounded wait instead of resolving immediately with `undefined`. The `_onDidRegisterProvider` listener (still active during the wait) handles early success; the timeout handles the "definitely not coming" case.

```ts
// searchService.ts
import { disposableTimeout, raceCancellationError } from '../../../../base/common/async.js';

// inside waitForSearchProvider:
store.add(this.extensionService.onDidChangeExtensionsStatus(() => {
    if (this.extensionService.activationEventIsDone(activationEvent)) {
        // Remote EH: $registerFileSearchProvider is fire-and-forget and may still be
        // in-flight when activationEventIsDone flips. Give it a bounded window before
        // resolving with undefined. The _onDidRegisterProvider listener above wins
        // immediately if the provider arrives first.
        store.add(disposableTimeout(() => done(this.getSearchProvider(type).get(scheme)), 500));
    }
}));
```

---

**Tests** (`npm run test-browser-no-install -- --browser chromium`):

```
✔ fileSearch returns empty for virtual workspace without provider
✔ fileSearch catches provider registered before activation completes (105ms)
✔ fileSearch catches provider registered after activation completes (remote EH race) (155ms)

3 passing (288ms)
```

The third test covers the exact remote EH race: activation fires → 50 ms later provider arrives → assert non-empty results. It fails without the `disposableTimeout` fix and passes with it.

Full diff: https://github.com/harshagarwalnyu/vscode/commit/48a0723d220267cd94c1b07f00600e25cbd43580

PR #314321, opened 2026-05-05
Improve sessions list rendering on phone layout

osortega:
Improves the readability and tap-friendliness of the agent sessions list under the phone layout. No new components, no new data — only a phone-aware row height and CSS changes scoped to `.agent-sessions-workbench.phone-layout`.

### Changes

- **Phone-aware row height** (`SessionsTreeDelegate`): 76px on phone vs 54px on desktop. Heights refresh reactively when `IsPhoneLayoutContext` changes.
- **Larger fonts and roomier rows**: 15px title / 12.5px details, 10px padding, vertically centered icon + main column.
- **Card-style rows**: `margin: 2px 8px`, 10px border radius for a phone-native press feel.
- **Inline actions on touch**: the per-row toolbar is always visible (no hover/focus dependency), pulled out of the title-row flow with `position: absolute; right: 6px; top: 50%` so 44px tap targets don't inflate the title line-height or push the details row down. `.session-main` reserves `padding-right: 96px` to keep titles from sliding under the icons. Action labels are flex-centered so the icon sits in the middle of the 44×44 tap area.
- **Action taps no longer collapse the list**: the toolbar container stops `pointerdown` / `pointerup` / `click` / `dblclick` propagation, so the row's `onDidOpen` handler doesn't fire when a tap lands on an action. This applies on desktop and mobile — semantically actions on the row toolbar are not "open the row".

### Architecture notes

- All visual changes live behind `.agent-sessions-workbench.phone-layout` ancestor selectors, so desktop rendering is byte-identical.
- Phone branching uses the existing `IsPhoneLayoutContext` reactive signal (per `MOBILE.md` guidance) for the height refresh — no new branching infrastructure.
- The propagation fix is unconditional. Actions never imply "open the row", on any platform.

### Out of scope

The HTML mock that inspired this also showed colored status pills, a filter chip row, a FAB, and a bottom tab bar. Those are not implemented here — this PR is strictly about making the existing list readable and tappable on a phone.

### Verification

- `npm run compile-check-ts-native` clean.
- Manually verified in vscode.dev/agents at iPhone 14 Pro viewport: rows render at 76px, actions tap correctly, list does not collapse on action tap, desktop layout unchanged above the phone breakpoint.

### DIFF
--- a/src/vs/sessions/contrib/sessions/browser/media/sessionsList.css
+++ b/src/vs/sessions/contrib/sessions/browser/media/sessionsList.css
@@ -442,14 +442,18 @@
 }
 
 /*
- * Phone layout: roomier rows, larger title that wraps to 2 lines, and a
- * slightly larger details row. The taller row height is reserved by
- * `SessionsTreeDelegate.ITEM_HEIGHT_PHONE` — keep these rules in sync if
- * paddings/font sizes change here.
+ * Phone layout: roomier rows with a centered, always-visible action
+ * toolbar that meets the 44px minimum touch target. The taller row
+ * height is reserved by `SessionsTreeDelegate.ITEM_HEIGHT_PHONE` —
+ * keep these rules in sync if paddings/font sizes change here.
  */
 .agent-sessions-workbench.phone-layout .sessions-list-control {
 	.monaco-list-row:has(.session-item) {
-		margin: 2px 8px;
+		/* Horizontal-only margin: virtual list rows are absolutely
+		 * positioned with JS-set `top`/`height`, so vertical margins
+		 * here would not produce reliable inter-row spacing. Any gap
+		 * lives inside the row via padding. */
+		margin: 0 8px;
 		width: calc(100% - 16px);
 		border-radius: 10px;
 	}
--- a/src/vs/sessions/contrib/sessions/browser/media/sessionsList.css
+++ b/src/vs/sessions/contrib/sessions/browser/media/sessionsList.css
@@ -440,3 +440,85 @@
 	-webkit-touch-callout: none;
 	touch-action: manipulation;
 }
+
+/*
+ * Phone layout: roomier rows, larger title that wraps to 2 lines, and a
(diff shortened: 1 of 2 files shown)


osortega:
Triage of the Copilot review:

**Addressed in bf8562b**

1. ✅ Stale "6px padding" arithmetic in the row-height doc comment — simplified to describe the intent without a stale formula.
2. ✅ `Gesture.ignoreTarget` for touch — correct catch. The list opens rows from `list.onTap` (a `Gesture` event), which doesn't bubble through DOM, so my pointer/click `stopPropagation` only handled mouse. Added `Gesture.ignoreTarget(titleToolbarContainer)` alongside the existing pointer/click stops.
3. ✅ "Cheap O(visible)" comment — fixed; we iterate `this.sessions` (all known sessions). Phone↔desktop transitions are rare, so the wording now reflects the real cost rather than over-claiming.
4. ✅ "title that wraps to 2 lines" — stale; the line-clamp was dropped earlier in the iteration. Comment rewritten.
5. ✅ Vertical margin on `.monaco-list-row` — removed. Rows are absolutely positioned with JS-set `top`/`height`, so the 2px vertical was a visual no-op at best and overlap-prone at worst. Now horizontal-only (`margin: 0 8px`); any row spacing lives inside the row via padding.

**Not changing**

6. ❌ Move phone branching into a separate phone delegate / view registration. The MOBILE.md guidance applies to *Parts* (and to desktop code that should remain phone-agnostic). Here the phone variance is a single tree-row height tightly coupled to the same file's `.phone-layout` CSS. Splitting `SessionsList` / `SessionsTreeDelegate` into desktop and phone variants would duplicate approval-row plumbing, section/show-more handling, and the entire constructor wiring purely to externalize one constant. The current implementation reads `IsPhoneLayoutContext` reactively in one place, with the height refresh confined to a context-key listener that ignores anything but the phone key. It stays.

bpasero (approved):

PR #314329, opened 2026-05-05
fix: prevent findFiles2 from stalling indefinitely when no provider exists

### LINKED ISSUE #314285: FindFIles2 can stall forever with virtual file system with no search provider

ngxminhlong-spec:
Introduces a 30s timeout and combined cancellation token to ensure search operations terminate if the extension host or provider fails to respond. Fixes #314285.

### DIFF
--- a/src/vs/workbench/api/browser/mainThreadSearch.ts
+++ b/src/vs/workbench/api/browser/mainThreadSearch.ts
@@ -3,7 +3,7 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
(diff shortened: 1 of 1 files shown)


mjbvz:
we also don't want stalls for 30 seconds

ngxminhlong-spec:
> we also don't want stalls for 30 seconds

ok let me find a new way to fix it

Issue #313601, opened 2026-04-30
Flaky: chat.runInTerminal 'non-zero exit code is reported'


meganrogge:
The integration test `non-zero exit code is reported` in `chat.runInTerminal` is flaky. Shell integration sometimes doesn't report the exit code before the result is returned, so the output is just `"Command produced no output"` instead of the expected `"Command produced no output\nCommand exited with code 42"`.

PR #314368, opened 2026-05-05
Fix flaky exit code reporting in run_in_terminal (fallback handling) #313601

zmo2s:
Fix flaky exit code reporting in run_in_terminal

Related to #313601

## Problem

The integration test `non-zero exit code is reported` is flaky because
shell integration does not always report the exit code before the result
is returned.

This change improves exit code handling by ensuring a fallback when
shell integration has not yet populated the exit code.


zmo2s:

PR #314409, opened 2026-05-05
refactor: remove sub application support

deepak1556:


src/vs/code/electron-main/app.ts
```diff
@@ -1778,33 +1712,40 @@ export class CodeApplication extends Disposable {
...
 		// Refs: https://github.com/microsoft/vscode/issues/234064
 		validateDevDeviceId(this.stateService, this.logService);
-
-		// macOS: eagerly register the embedded app with Launch Services
-		this.registerEmbeddedAppWithLaunchServices();
 	}
 
-	private registerEmbeddedAppWithLaunchServices(): void {
-		if (!isMacintosh || (process as INodeProcess).isEmbeddedApp || !this.productService.embedded?.nameShort || this.productService.quality === 'stable') {
+	private cleanupAgentsApplication(): void {
```
sandy081, line 1734: As part of the clean up should we also delete user data and extensions directory?

src/vs/workbench/workbench.desktop.main.ts
```diff
@@ -187,6 +187,9 @@ import './contrib/encryption/electron-browser/encryption.contribution.js';
 // Emergency Alert
 import './contrib/emergencyAlert/electron-browser/emergencyAlert.contribution.js';
 
+// Agents App Merged Banner
+import './contrib/agentsAppMergedBanner/browser/agentsAppMergedBanner.contribution.js';
```
sandy081, line 191: Should this banner be shown only in Agents window?

sandy081 (dismissed):

src/vs/workbench/workbench.desktop.main.ts
deepak1556: Shouldn't it be opposite, only show in vscode window with open in agents window action. Reason if agents app was not running before the update and on restart to the version without the app, vscode window will open and the banner will be a quick signal of the app deprecation.

src/vs/code/electron-main/app.ts
deepak1556: Thanks, yes I was unsure of removing them since I lacked context.

dmitrivMS (approved):

PR #314410, opened 2026-05-05
Hide Agent Sessions banner when disableAIFeatures is true

dmitrivMS:


src/vs/workbench/contrib/chat/browser/agentSessions/agentSessionsBanner.ts
```diff
@@ -34,8 +35,12 @@ export interface IAgentsBannerResult {
  * Returns whether the agents banner can be shown.
  * The banner requires the `workbench.action.openAgentsWindow` command
  * to be registered (desktop builds only) and is limited to Insiders quality.
+ * It is also hidden when AI features are disabled via `chat.disableAIFeatures`.
  */
-export function canShowAgentsBanner(productService: IProductService): boolean {
+export function canShowAgentsBanner(productService: IProductService, configurationService: IConfigurationService): boolean {
+	if (configurationService.getValue<boolean>(ChatConfiguration.AIDisabled) === true) {
```
osortega, line 41: We need to check the entitlement service so that we capture more than just the setting, for example https://github.com/microsoft/vscode/blob/a492920c667e7a7d273db13484bc15611a82e594/src/vs/workbench/contrib/preferences/browser/settingsEditor2.ts#L398

bryanchen-d (approved):

PR #314411, opened 2026-05-05
dom: attribute ResizeObserver loop telemetry to DisposableResizeObserver name

bryanchen-d:
## Summary

Attribute `ResizeObserver loop completed with undelivered notifications.` telemetry to the specific `DisposableResizeObserver` that most recently ran. Every `DisposableResizeObserver` now requires a stable `name` (`"<ClassName>.<purpose>"`); on the telemetry side, when the loop warning surfaces we prefix both `msg` and `callstack` with that name so the bucket splits by offender instead of pooling everything into one opaque bucket.

## Motivation

`ResizeObserver loop completed with undelivered notifications.` has been the **#1 error bucket** since v1.106 (#293359 — 54M hits / 2.5M users). Chromium fires it on `window.onerror` whenever any `ResizeObserver` callback writes to layout in a way the browser couldn't fully drain in the same frame.

Because the warning carries no useful stack (`<no error message>` from a synthetic `ErrorEvent`), every offender — chat row height, KaTeX, find-widget, terminal tool progress, plan review, etc. — collapses into one bucket. We can't tell which observer is the dominant source, so we can't fix it.

### `DisposableResizeObserver` (`src/vs/base/browser/dom.ts`)

- New required `name: string` first constructor parameter. Convention: `"<ClassName>.<purpose>"`. Stable across releases (does not embed line numbers or call-site stacks).
- Wrapper records `_lastInvokedDisposableResizeObserver = this` immediately before invoking the user callback, inside a try/catch that routes thrown errors to `onUnexpectedError` (so one bad consumer can't break delivery for every other observer in the realm).
- New exported helper `getRecentDisposableResizeObserverAttributionForLoopError(message)` returns `[DisposableResizeObserver(<name>)] <message>` when the message matches the loop warning, otherwise `undefined`.
- Optional `targetWindow` + `{ resizeObserverCtor }` constructor options preserved (the latter exists for tests).

### `BrowserErrorTelemetry` (`src/vs/platform/telemetry/browser/errorTelemetry.ts`)

In `_onUncaughtError`, after the existing `err.stack` handling, call the helper above. If it returns a string we overwrite **both** `data.msg` and `data.callstack` with the prefixed value so the telemetry bucket key carries the offending observer's name.

### Convert remaining raw `new ResizeObserver(...)` sites

So every observer gets a stable name and the same attribution path:

- `chatMarkdownContentPart` (KaTeX layout participants) → `ChatMarkdownContentPart.mathLayout`
- `chatTerminalToolProgressPart` → `ChatTerminalToolProgressPart.handleResize`
- `browserEditorFindFeature` → `BrowserEditorFindFeature.heightChange` (passes `getWindow(container)` to preserve aux-window realm)

### Name every existing call site

All 17 production `DisposableResizeObserver` constructions now pass a stable name:

| File | Name |
|---|---|
| `chatCompositeBar.ts` | `ChatCompositeBar.activeTabReveal` |
| `browserEditorFindFeature.ts` | `BrowserEditorFindFeature.heightChange` |
| `aiCustomizationWelcomePagePromptLaunchers.ts` | `AICustomizationWelcomePagePromptLaunchers.scrollable` |
| `chatListRenderer.ts` | `ChatListItemRenderer.itemHeight` |
| `chatInputPart.ts` | `ChatInputPart.containerHeight`, `ChatInputPart.compactToolbars` |
| `chatConfirmationWidget.ts` | `BaseSimpleChatConfirmationWidget.message` (+3 siblings) |
| `chatMarkdownContentPart.ts` | `ChatMarkdownContentPart.mathLayout` |
| `chatPlanReviewPart.ts` | `ChatPlanReviewPart.messageScrollable` |
| `chatQuestionCarouselPart.ts` | `ChatQuestionCarouselPart.inputScrollable` |
| `chatSubagentContentPart.ts` | `ChatSubagentContentPart.layout` |
| `chatThinkingContentPart.ts` | `ChatThinkingContentPart.child`, `ChatThinkingContentPart.wrapper` |
| `chatToolConfirmationCarouselPart.ts` | `ChatToolConfirmationCarouselPart.contentExpansion` |
| `chatTerminalToolProgressPart.ts` | `ChatTerminalToolProgressPart.handleResize` |

## Why this approach

Earlier iterations explored capturing a creation stack and matching it to the loop warning. Two problems:

1. Stacks are large and change across releases — they're useless as a bucket key.
2. The actual offender at fire time is the observer whose callback most recently wrote to layout, not necessarily the one that constructed last.

A stable, hand-assigned name is small, version-stable, and grouped by intent. Recording `_lastInvokedDisposableResizeObserver` right before invoking the callback gives us a precise "who fired most recently" pointer that survives into the synchronous `window.onerror` handler.

This is a **diagnostic** change, not a behavioural fix — it does not by itself reduce the loop-warning rate. It unblocks the next step: with attribution per observer we can see which one dominates the bucket and target that fix specifically (the chat-row height feedback loop is the prime suspect).

## Refs

#293359 (top error bucket — 54M hits since v1.106)

## Test plan

- New unit tests in `src/vs/base/test/browser/dom.test.ts` cover constructor, dispose, throwing-callback routing, name exposure, and the attribution helper for both matching and non-matching messages.
- Manual: provoke a `ResizeObserver loop` in chat; telemetry payload now contains `[DisposableResizeObserver(<name>)] ResizeObserver loop completed with undelivered notifications.` in both `msg` and `callstack`.


src/vs/base/browser/dom.ts
```diff
@@ -2049,15 +2049,56 @@ export class DragAndDropObserver extends Disposable {
...
+				} catch (e) {
+					onUnexpectedError(e);
+				}
+			});
+		});
+		this._register(toDisposable(() => {
+			this.scheduled?.dispose();
+			this.scheduled = undefined;
+			this.pendingEntries = undefined;
+			this.observer.disconnect();
```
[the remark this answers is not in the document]
bryanchen-d: Added a `DisposableResizeObserver` test suite in `src/vs/base/test/browser/dom.test.ts` covering the new batching semantics (commit 4b39c53):

- **rAF deferral** — callback does not run inside the resize-observation phase, only on the next animation frame.
- **Per-frame coalescing** — multiple synchronous deliveries within one frame fold into a single callback call with merged entries (in delivery order).
- **Cancel-on-dispose** — disposing between native delivery and the rAF callback prevents the callback from firing and disconnects the underlying `ResizeObserver`.
- **Reschedule-after-frame** — subsequent deliveries after a frame fires are not lost.

The tests stub `mainWindow.ResizeObserver` so deliveries can be triggered synthetically without depending on real layout.

jruales:
The change in Integrated Browser related file browserEditorFindFeature.ts looks good to me

src/vs/base/test/browser/dom.test.ts
```diff
@@ -532,5 +532,106 @@ suite('dom', () => {
...
+					(mainWindow as any).ResizeObserver = original;
+				}
+			};
+			class FakeRO {
+				constructor(cb: ResizeObserverCallback) { captured = cb; }
+				observe() { /* no-op */ }
+				unobserve() { /* no-op */ }
+				disconnect() { state.disconnects++; }
+			}
+			(mainWindow as any).ResizeObserver = FakeRO;
```
mjbvz: This should use dependency injection instead. Modifying global objects and all the `any` casts are a big code smell

mjbvz (dismissed):
Change looks ok but the tests shouldn't be adding more any casts or needing to modify global objects

src/vs/base/test/browser/dom.test.ts
bryanchen-d: Good call — refactored in e71769c. Added an optional `resizeObserverCtor` parameter to `DisposableResizeObserver` so the test injects a fake constructor instead of mutating `mainWindow`. Removed every `any` cast: the fake implements the `ResizeObserver` interface directly.

While here I also addressed Copilot's separate coalescing finding (consumers like `chatListRenderer.ts:639-642` and `chatThinkingContentPart.ts` read `entries[0]` as the current size — appending was leaving the *oldest* entry first and could fire a stale height). Switched the pending buffer to a `Map<Element, ResizeObserverEntry>` so per-target it's latest-wins; added a regression test (`latest entry per target wins…`).

bryanchen-d: Also captured this as a new learning into the copilot instructions.

roblourens (approved):
Good idea, seems fine. I just wonder whether there's a chance that multiple observers would run within that window though. Seems like it would be usually correct though.

PR #314425, opened 2026-05-05
Remove 'Del' button from session history window

vijayupadya:
Remove 'Del' button from session history window. This button was recently added and is not needed to be shown in the history window and we only show 'Archive'.


pwang347 (approved):

rebornix (approved):

Inuart:
I was wondering why I couldn't delete a session anymore. Is there any other way?

vijayupadya:
You can right click on the session item and select "Delete" from the drop down context menu

image

Inuart:
> You can right click on the session item and select "Delete" from the drop down context menu

no I can't
image

PR #314525, opened 2026-05-05
Compress tool output to reduce token usage

meganrogge:
Ports https://github.com/microsoft/vscode-copilot-chat/pull/5106 into vscode core, since the chat extension code now lives here.

Adds a post-processing compression layer for tool results, inspired by [ztk](https://github.com/codejunkie99/ztk). When the model invokes a tool, we now have a chance to filter the result text before it reaches the model — same information, fewer tokens.

## What

- New `IToolResultCompressor` service in `src/vs/workbench/contrib/chat/common/tools/toolResultCompressor.ts` with a per-tool filter registry. Filters are pure functions over text parts. The service skips outputs under 80 bytes and drops any filter that throws (ztk's "never make it worse" rule).
- Three initial filters for `run_in_terminal` in `src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/terminalOutputCompressor.ts`:
  - **`git diff` / `git show`** — collapses unchanged context, drops `index`/`similarity index` headers, omits lockfile and snapshot diff bodies, preserves `+`/`-`/`@@` lines verbatim.
  - **`ls -l` / `ls -la`** — keeps just file names (with `/` for dirs), drops perm/owner/size columns and `total N`.
  - **`npm install` / `yarn` / `pnpm install`** — drops progress bars, `npm warn deprecated`, funding/audit nag lines.
- Wired into `LanguageModelToolsService.invokeTool` as a single post-process step right after `tool.impl.invoke` resolves.
- New setting `chat.tools.compressOutput.enabled`, **off by default**, with `experiment: { mode: 'auto' }`.
- New telemetry event `toolResultCompressed` reporting `{ toolName, filters, beforeBytes, afterBytes }` so we can quantify savings.

## Why

The terminal tool routinely dumps tens of thousands of raw tokens (a single `git diff HEAD~5` can be 90KB+). Most of that is metadata, unchanged context, lockfile churn, or progress noise. Compressing this client-side before it hits the model is a cheap, predictable token win that doesn't change tool semantics.

## Tests

Unit tests cover the service (filter registration, throwing filter handling, audience preservation, min-length skip) and each terminal filter's parse/match/apply logic.

## Scope notes

- Default off; opt-in via setting (experiment-driven).
- Easy to extend: add a new `IToolResultFilter` and register it.

image
image

### DIFF
--- a/src/vs/workbench/contrib/chat/browser/chat.contribution.ts
+++ b/src/vs/workbench/contrib/chat/browser/chat.contribution.ts
@@ -157,6 +157,8 @@ import './widget/input/editor/chatInputEditorContrib.js';
 import './widget/input/editor/chatInputEditorHover.js';
 import { LanguageModelToolsConfirmationService } from './tools/languageModelToolsConfirmationService.js';
 import { LanguageModelToolsService, globalAutoApproveDescription } from './tools/languageModelToolsService.js';
+import { IToolResultCompressor } from '../common/tools/toolResultCompressor.js';
+import { ToolResultCompressorService } from './tools/toolResultCompressorService.js';
 import { AgentPluginService, ConfiguredAgentPluginDiscovery, ExtensionAgentPluginDiscovery, MarketplaceAgentPluginDiscovery } from '../common/plugins/agentPluginServiceImpl.js';
 import { IAgentPluginRepositoryService } from '../common/plugins/agentPluginRepositoryService.js';
 import { IPluginInstallService } from '../common/plugins/pluginInstallService.js';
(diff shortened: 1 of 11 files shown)


dmitrivMS (dismissed):

bhavyaus:
What do token savings look like with evals on terminalbench?

benvillalobos (dismissed):

meganrogge:
~requires-eval-assessment terminalbench2 gpt-5.4,claude-opus-4.6,claude-opus-4.7

```yaml
vscodeSettings:
  chat.tools.compressOutput.enabled: true
```

meganrogge:
@bhavyaus that is something we're curious to see - along with what we might see with exp, trying the above ^

Yoyokrazy (approved):

PR #314529, opened 2026-05-05
Stub getCurrentAuthedUser/getUserOrganizations on NullBaseOctoKitService

meganrogge:
## Problem

When running an msbench eval against `sessionTarget: agent-host-copilot` (e.g. `runtest local --benchmark vscbench.say_hello --config doc/examples/configs/agent-host.vscode.agent.yaml`), the Copilot extension host logs:

```
WARN this.octoKitService.getCurrentAuthedUser is not a function
```

…and the agent-host benchmark hangs.

## Root cause

In `src/extension/extension/vscode/services.ts`:

```ts
builder.define(IOctoKitService, isScenarioAutomation
    ? new SyncDescriptor(NullBaseOctoKitService)
    : new SyncDescriptor(OctoKitService));
```

The eval container sets `IS_SCENARIO_AUTOMATION=1`, so `NullBaseOctoKitService` is injected as `IOctoKitService`. But `NullBaseOctoKitService` extends `BaseOctoKitService`, which only implements the `*WithToken` variants (`getCurrentAuthedUserWithToken`, `getUserOrganizationsWithToken`). The no-arg interface methods `getCurrentAuthedUser()` / `getUserOrganizations()` are only implemented on the real `OctoKitService` (which has access to `IAuthenticationService` to fetch the token).

## Why agent-host specifically

The only non-test caller of `IOctoKitService.getCurrentAuthedUser` is `GitHubOrgChatResourcesService.computePreferredOrganizationName` (`src/extension/agents/vscode-node/githubOrgChatResourcesService.ts`). That service is consumed by `GitHubOrgCustomAgentProvider` and `GitHubOrgInstructionsProvider` — features that **only the agent-host code path** activates to discover GitHub-org-backed custom agents and instruction files. Regular chat / inline chat / edit mode never traverse this service, which is why every other `sessionTarget` runs cleanly through the eval and only `agent-host-copilot` triggers it.

## Fix

Add stub overrides on `NullBaseOctoKitService`:

- `getCurrentAuthedUser()` → returns the `NullUser` placeholder (consistent with the existing `getCurrentAuthedUserWithToken` stub).
- `getUserOrganizations()` → returns `[]`.

The caller already handles the empty / unsigned-in case correctly (it just skips org discovery), so this unblocks the agent-host eval path without changing real-user behavior.

## Validation

Type-checks cleanly. End-to-end validation requires this fix to ship in Insiders so the eval Docker image picks it up; once that lands, re-run `runtest local --benchmark vscbench.say_hello --config doc/examples/configs/agent-host.vscode.agent.yaml`.

Companion to https://github.com/microsoft/vscode/pull/314526 (which fixes the parallel `agentHostPermissionService` workbench-DI error encountered on the same eval run).


roblourens (approved):

meganrogge:

```yaml
vscodeSettings:
  chat.tools.compressOutput.enabled: true
```

PR #314598, opened 2026-05-06
Remove visibility controls for models in the model configuration window

vritant24:
Today:
- The model picker has a search option that can be used to quickly find models
-  the most recently used models are surfaced to the top
   - can be quickly searched by typing `/models` in the chat window

Being able to toggle visibility is no longer necessary with those available. 
Users can also run into an interesting problem where searching for models don't show up because they are `hidden` in the chat window model picker, but are searchable in the model configuration window.

This PR removes controlling the visibility of models entirely, allowing all models to be available in the model picker.


src/vs/workbench/contrib/chat/browser/chatManagement/media/chatModelsWidget.css
```diff
@@ -74,31 +74,11 @@
 	background-color: var(--vscode-keybindingTable-headerBackground);
 }
 
-.models-widget .models-table-container .monaco-table-th[data-col-index="1"] {
+.models-widget .models-table-container .monaco-table-th[data-col-index="0"] {
 	padding-left: 8px;
 }
 
 
-/** Gutter column styling **/
```
vritant24, line 82: removed the gutter as it had the visibility controls, which are no longer needed

vijayupadya (dismissed):

vijayupadya (approved):

Randalphwa:
placeholder

PR #314753, opened 2026-05-06
Fix background color for window controls overlay in titlebar

### LINKED ISSUE #312914: Agents window Windows Control Overlay does not change with theme
1. Open the Agents app on Windows.
2. If the app is on a dark theme, change to a light theme, otherwise, change to a dark theme.
3. :bug: The Windows Control Overlay on the top-right does not change with the rest of the theme.

Windows Control Overlay in white while the rest of the title bar is black

mrleemurray:
Resolve the background color issue for the window controls overlay in the titlebar by directly using the theme's `agentsBackground`. This change ensures the overlay displays the correct color when the theme changes, addressing the transparency issue.

Fixes https://github.com/microsoft/vscode/issues/312914


benibenj:
This doesn't fix it. 
image

The issue is the gradient which starts at the top left and goes to the bottom right. You need to correctly compute the color which will apply to the top right. This works by coincidence on macOS because the WCO is in the top left, but does not wok in windows and linux where it is in the top right.

mrleemurray:
@hawkticehurst will be reducing the gradient size to avoid color clash

mrleemurray:
Pulled in Hawk's new gradient - please retest @benibenj

PR #314790, opened 2026-05-06
(feature) Enable to filter search within changed files of source tree

dgercho:
Adds a "Search only in changed files (source control)" toggle button to the search panel's files to include row, alongside the existing "Search only in Open Editors" toggle. When enabled, search results are restricted to files that have uncommitted changes tracked. The toggle is disabled if there's no repository presented.


dgercho:

dgercho:
@osortega Would it be possible to merge this feature? I think it could be helpful for many users

src/vs/workbench/contrib/search/browser/patternInputWidget.ts
```diff
@@ -211,11 +216,43 @@ export class IncludePatternInputWidget extends PatternInputWidget {
...
+			isChecked: false,
+			...defaultToggleStyles
+		}));
+		this.useSearchInChangedFilesBox.disable();
+		this._register(this.useSearchInChangedFilesBox.onChange(viaKeyboard => {
+			this._onChangeSearchInChangedFilesBoxEmitter.fire();
+			if (!viaKeyboard) {
+				this.inputBox.focus();
+			}
+		}));
```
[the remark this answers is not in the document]
dgercho: Updated.

src/vs/workbench/contrib/search/browser/searchView.ts
```diff
@@ -303,6 +305,26 @@ export class SearchView extends ViewPane {
...
+			}));
+		};
+		for (const repository of this.scmService.repositories) {
+			registerScmRepositoryListeners(repository);
+		}
+		this._register(this.scmService.onDidAddRepository(repository => {
+			registerScmRepositoryListeners(repository);
+			updateChangedFilesToggleEnabled();
+		}));
+		this._register(this.scmService.onDidRemoveRepository(() => updateChangedFilesToggleEnabled()));
```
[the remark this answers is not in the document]
dgercho: Updated.

src/vs/workbench/contrib/search/browser/searchView.ts
```diff
@@ -303,6 +305,26 @@ export class SearchView extends ViewPane {
 		this._register(this.searchHistoryService.onDidClearHistory(() => this.clearHistory()));
 		this._register(this.configurationService.onDidChangeConfiguration(e => this.onConfigurationUpdated(e)));
 
+		const updateChangedFilesToggleEnabled = () => {
+			const hasRepositories = [...this.scmService.repositories].length > 0;
+			this.inputPatternIncludes?.setOnlySearchInChangedFilesEnabled(hasRepositories);
+		};
+		const registerScmRepositoryListeners = (repository: { provider: { onDidChangeResources: import('../../../../base/common/event.js').Event<void> } }) => {
```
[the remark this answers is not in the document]
dgercho: Updated.

src/vs/workbench/services/search/common/queryBuilder.ts
```diff
@@ -285,6 +286,13 @@ export class QueryBuilder {
 			return { ...queryProps, ...openEditorsQueryProps };
 		}
 
+		if (options.changedFileUris?.length) {
+			const changedFilesInQuery = options.changedFileUris.filter(uri => pathIncludedInQuery(queryProps, uri.fsPath));
+			const changedFilesQueryProps = this.commonQueryFromFileList(changedFilesInQuery);
+			this.logService.trace('QueryBuilder#commonQuery - changedFile Query', JSON.stringify(changedFilesQueryProps));
+			return { ...queryProps, ...changedFilesQueryProps };
+		}
```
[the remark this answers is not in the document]
dgercho: Updated.

dgercho:
@osortega Updated branch after copilot review.

src/vs/workbench/contrib/search/browser/searchView.ts
```diff
@@ -1617,6 +1646,18 @@ export class SearchView extends ViewPane {
 		const excludePattern = [{ pattern: this.inputPatternExcludes.getValue() }];
 		const includePattern = this.inputPatternIncludes.getValue();
 
+		let changedFileUris: URI[] | undefined;
+		if (onlySearchInChangedFiles) {
+			changedFileUris = [];
+			for (const repository of this.scmService.repositories) {
```
osortega: Please avoid doing this, could we do something like this instead:
```
if (onlySearchInChangedFiles) {
    changedFileUris = [...this.scmService.repositories]
        .flatMap(repository => repository.provider.groups)
        .flatMap(group => group.resources)
        .map(resource => resource.sourceUri);
}
```

src/vs/workbench/contrib/search/browser/patternInputWidget.ts
```diff
@@ -211,24 +216,63 @@ export class IncludePatternInputWidget extends PatternInputWidget {
...
+	}
+
 	protected override getSubcontrolsWidth(): number {
-		return super.getSubcontrolsWidth() + this.useSearchInEditorsBox.width();
+		return super.getSubcontrolsWidth() + this.useSearchInEditorsBox.width() + this.useSearchInChangedFilesBox.width();
 	}
 
 	protected override renderSubcontrols(controlsDiv: HTMLDivElement): void {
+		this.useSearchInChangedFilesBox = this._register(new Toggle({
+			icon: Codicon.sourceControl,
```
osortega: Can we use the `edit-code` codicon instead?

osortega (commented):
This feature would be really useful for agentic flows, thanks for contributing.
Just a couple comments

src/vs/workbench/contrib/search/browser/searchView.ts
dgercho: Good catch, applied.

src/vs/workbench/contrib/search/browser/patternInputWidget.ts
dgercho: Whatever you prefer :)

osortega (approved):

src/vs/base/browser/ui/toggle/toggle.css
```diff
@@ -22,6 +22,12 @@
 	background-color: var(--vscode-inputOption-hoverBackground);
 }
 
+.monaco-custom-toggle.disabled {
```
TylerLeonhardt, line 25: What does this look like? This impacts every toggle in the product.

dmitrivMS (dismissed):

src/vs/base/browser/ui/toggle/toggle.css
osortega: @dgercho I think this is not relevant anymore? Can we revert this and also the `this.useSearchInChangedFilesBox.disable();` in renderSubcontrols method?

osortega (changes requested):

src/vs/base/browser/ui/toggle/toggle.css
dgercho: If we remove this, the button will also be presented when no source control is available. What if we move it to [searchview.css](vscode-webview://0utllc8fi4rbe22183g0a5m30g062k7dhfdvgaqpb7eqsom56j4q/src/vs/workbench/contrib/search/browser/media/searchview.css) and scoped it only to this specific toggle? @TylerLeonhardt @osortega

TylerLeonhardt: What is this change trying to achieve? Include a screenshot please. I am ok with a change that impacts every toggle in the product if it makes sense. Consistency is also important here as, ideally, every toggle works and looks the same way. If search is different for some reason, that reason should be obvious

dgercho: Screenshot 2026-05-19 at 20 16 16
Screenshot 2026-05-19 at 20 16 01

dgercho: We can see in this example that when source control is not available (for example, for a local folder) the button is disabled.

This behavior could also be applied in other places, such as "Open Editors" - if there are no open editors, the toggle should be disabled. It's similar to trying to filter something that isn’t filterable, and the same approach could be useful in other scenarios like this one. @TylerLeonhardt @osortega

osortega: Ahh I didn't catch that, looks like we do the same for the `Replace All` action bar item, we just don't support that in the toggle widget. 
I actually think this is a good idea, though we should probably make it consistent across the board, meaning choosing either disabling when source control is not available or when it's available but there are no changes, or leaving the behavior without disabling just like Open Editors.
My preference would be the former

dgercho: So we want to make the behavior of this new toggle match the "Replace All"?

osortega: @dgercho only for the way it is enabled or disabled, enabled when source control is enabled (git repository) and there are changes. Disabled otherwise.
I can look into doing that for the "Search only in Open Editors" once this lands

dgercho: Ok, updated- now we enable the toggle only when repository has changes

osortega (approved):
✨

roblourens (approved):

PR #313817, opened 2026-05-01
chore(deps): bump @xterm/* to beta.213 to pick up overview ruler dispose fix

bryanchen-d:
Fixes #303546

## What

Bumps all `@xterm/*` dependencies (root, `remote/`, and `remote/web/`) to `beta.213` (webgl to `beta.212`, the latest published). Previously pinned at `beta.197` / `beta.196`.

## Why

`#303546` (`Cannot read properties of undefined (reading ''dimensions'')`) is caused by a race in xterm.js `OverviewRulerRenderer`: a `requestAnimationFrame` queued by `_queueRefresh` could fire after the renderer was disposed, ending up in `RenderService.get dimensions()` which does `return this._renderer.value!.dimensions;` — `_renderer.value` is `undefined` post-dispose, hence the TypeError.

The upstream fix is xtermjs/xterm.js#5826 / commit [`08ae141`](https://github.com/xtermjs/xterm.js/commit/08ae141) "Improve overview ruler lifecycle":

- Adds `if (this._store.isDisposed || !this._renderService.hasRenderer()) return;` guards to `_refreshCanvasDimensions` and `_refreshDecorations`.
- Cancels the pending `requestAnimationFrame` via a `toDisposable` registered on the renderer''s store.
- Adds an early-return in `_queueRefresh` and an `isDisposed` re-check inside the rAF callback.

That fix shipped in `@xterm/xterm@6.1.0-beta.198`. We were pinned at `beta.197`, so this issue has continued to be reported against 1.115 → 1.118.1 (~5M+ hits in 1.118.1 alone across the four related buckets).

## Related buckets

| Bucket | Note |
|---|---|
| `8612c55d` | tracked since 1.112.0 |
| `4826565b` | ~444K hits / 216K users at 1.113.0 |
| `1e83a096` | ~370K hits / 98.5K users at 1.113.0 |
| `aaa283a2` | seen in 1.115.0+ |

## How to test

- `npm install` at the workspace root, run a dev build, open an integrated terminal, exercise terminal lifecycle (open / close / split / window resize / DPR change).
- Telemetry: monitor the four bucket IDs above on the next stable release; expected to drop to zero new occurrences.

## Risk

- Pure dependency bump — no source changes. Diff is exclusively the version pins and corresponding `package-lock.json` updates in the three lockfiles.
[...]

PR #314795, opened 2026-05-06
Guard terminal resize/dispose race against xterm.js dimension getters

### LINKED ISSUE #303546: [Unhandled Error] Uncaught TypeError: Cannot read properties of undefined (reading 'dimensions')
## Error Message
```
Uncaught TypeError: Cannot read properties of undefined (reading 'dimensions')
```

## Stack Trace
&nbsp;&nbsp;&nbsp;&nbsp;at get dimensions (node_modules/@xterm/xterm/lib/xterm.js:1:241618) [not deminified]  
&nbsp;&nbsp;&nbsp;&nbsp;at _._refreshCanvasDimensions (node_modules/@xterm/xterm/lib/xterm.js:1:380511) [not deminified]  
&nbsp;&nbsp;&nbsp;&nbsp;at _._refreshDecorations (node_modules/@xterm/xterm/lib/xterm.js:1:380885) [not deminified]  
&nbsp;&nbsp;&nbsp;&nbsp;at (anonymous) (node_modules/@xterm/xterm/lib/xterm.js:1:382417) [not deminified]

<details>
<summary>Raw Stack Trace (minified)</summary>

```
TypeError: Cannot read properties of undefined (reading 'dimensions')     at get dimensions (node_modules/@xterm/xterm/lib/xterm.js:1:241618)     at _._refreshCanvasDimensions (node_modules/@xterm/xterm/lib/xterm.js:1:380511)     at _._refreshDecorations (node_modules/@xterm/xterm/lib/xterm.js:1:380885)     at node_modules/@xterm/xterm/lib/xterm.js:1:382417
```

</details>

## Details
| Property | Value |
| --- | --- |
| Bucket ID | [`8612c55d-0fe2-33df-e604-bf40f4e0cb6f`](https://errors.code.visualstudio.com/bucket/8612c55d-0fe2-33df-e604-bf40f4e0cb6f?product=VSCode&commit=07ff9d6178ede9a1bd12ad3399074d726ebe6e43) |
| Version | 1.112.0 |
| Commit | [`07ff9d61`](https://insiders.vscode.dev/github/microsoft/vscode/tree/07ff9d6178ede9a1bd12ad3399074d726ebe6e43) |
| Last Seen | 2026-03-19T23:59:52.976Z |
| Platforms | Linux, Mac, Windows |
| Product | VSCode |

---
*This issue was automatically created from the VS Code Errors Dashboard*

---

## Related Error Buckets

These additional error buckets are tracked by this issue (v1.113.0):

| Bucket ID | Error Message | Hits | Users |
| --- | --- | --- | --- |
| `4826565b-6d48-9952-f05b-a8aa386e6b5c` | Cannot read properties of undefined (reading 'dimensions') | 444K | 216K |
| `1e83a096-ed50-2ba1-2dd8-46381020bd02` | Cannot read properties of undefined (reading 'dimensions') | 370K | 98.5K |

> Generated by [errors-triage](https://github.com/microsoft/vscode-engineering/actions/runs/23663746139) · [◷](https://github.com/search?q=repo%3Amicrosoft%2Fvscode+is%3Aissue+%22gh-aw-workflow-call-id%3A+microsoft%2Fvscode-engineering%2Ferrors-triage%22&type=issues)

bryanchen-d:
## What

Guards three terminal resize/dispose race paths that throw `Cannot read properties of undefined (reading 'dimensions')` from xterm.js' `RenderService.get dimensions()` after the renderer is disposed.

## Why

PR #313817 bumped `@xterm/*` to `beta.213` and resolved the `OverviewRulerRenderer` dispose race (bucket `8612c55d`, ✅ zero hits on post-fix insider builds). However three sibling buckets keep firing on the latest insiders (1.120.0-insider `79d6cea3` / `0aed0a9b`):

| Bucket | Caller | First seen on post-fix insider |
|---|---|---|
| `aaa283a2` | `_resizeYCallback` from configChange → `setVisible` → `_resize` | 161 hits / 98 users |
| `4826565b` | debounced `_debounceResizeX` timer | 909 hits / 497 users |
| `1e83a096` | `_resizeBothCallback` from `_onProcessExit` → `dispose` → `setVisible(false)` → `_resize` | 291 hits / 98 users |

All three reach `RenderService.get dimensions()` inside xterm.js, which dereferences `this._renderer.value` (undefined post-dispose) and throws synchronously. The optional chaining on `rawXterm.dimensions?.css.canvas.width` doesn't help because the **getter itself** throws before the `?.` short-circuits.

The upstream `08ae141` patch added `isDisposed` guards at the `OverviewRuler` call sites — not inside the getter — so the symmetrical fix here is to guard the call sites VS Code owns.

### `terminalInstance.ts`
- Bail out of the three resize closures passed to `TerminalResizeDebouncer` when `this.isDisposed`.
- `_updatePtyDimensions`: early-return when disposed; defensively `try/catch` around `rawXterm.dimensions` so a future renderer-dispose race fails silently with `undefined` dimensions instead of bubbling as an unhandled error.

### `terminalResizeDebouncer.ts`
- `resize` / `flush`: bail out when `this._store.isDisposed`.
- `runWhenWindowIdle` callbacks: re-check disposed state before invoking the resize callback (the idle callback can fire after `MutableDisposable.clear` for a different timing reason).
- `_debounceResizeX`: re-check disposed state. The `@debounce(100)` decorator schedules a `setTimeout` that is **not** tied to the disposable store, so it can fire after the terminal/xterm is gone — this is the direct cause of bucket `4826565b`.

## How to test

1. Open a terminal, type or generate >200 lines of output.
2. Trigger the dispose-during-resize patterns:
   - Quickly close (`Ctrl+Shift+W`) a terminal while resizing the window.
   - Run a config change (e.g. toggle `terminal.integrated.fontFamily`) immediately after killing a terminal.
   - Kill a long-running shell from within the terminal (`exit`) while the panel is mid-resize.
3. Verify no unhandled `dimensions` TypeErrors in DevTools console.
4. Telemetry: monitor buckets `aaa283a2`, `4826565b`, `1e83a096` over the next two insider builds — expect zero new occurrences.

## Risk

- All guards are early-returns / try-catch around accessors — no behavior change in the live (non-disposed) path.
- Defensive `try/catch` in `_updatePtyDimensions` accepts a single missed dimension update if the renderer is mid-dispose; this is preferable to throwing and losing the entire reconnect/restore flow.
- No test changes; the races are timing-dependent and hard to reproduce deterministically. Telemetry will be the verification signal.

Refs #303546, #313817

### DIFF
--- a/src/vs/workbench/contrib/terminal/browser/terminalInstance.ts
+++ b/src/vs/workbench/contrib/terminal/browser/terminalInstance.ts
@@ -816,14 +816,23 @@ export class TerminalInstance extends Disposable implements ITerminalInstance {
 			() => this._isVisible,
 			() => xterm,
 			async (cols, rows) => {
+				if (this.isDisposed) {
+					return;
+				}
 				xterm.resize(cols, rows);
 				await this._updatePtyDimensions(xterm.raw);
 			},
 			async (cols) => {
+				if (this.isDisposed) {
+					return;
+				}
 				xterm.resize(cols, xterm.raw.rows);
 				await this._updatePtyDimensions(xterm.raw);
 			},
 			async (rows) => {
+				if (this.isDisposed) {
+					return;
+				}
 				xterm.resize(xterm.raw.cols, rows);
 				await this._updatePtyDimensions(xterm.raw);
 			}
@@ -2047,8 +2056,22 @@ export class TerminalInstance extends Disposable implements ITerminalInstance {
 	}
 
 	private async _updatePtyDimensions(rawXterm: XTermTerminal): Promise<void> {
-		const pixelWidth = rawXterm.dimensions?.css.canvas.width;
-		const pixelHeight = rawXterm.dimensions?.css.canvas.height;
+		if (this.isDisposed) {
+			return;
+		}
+		// `rawXterm.dimensions` proxies to xterm.js' RenderService which throws
+		// `Cannot read properties of undefined (reading 'dimensions')` if the
+		// renderer was disposed between scheduling and invocation (debounced or
+		// idle resize callbacks racing with terminal teardown). Guard against
+		// that here so the optional chaining short-circuits to undefined.
+		let pixelWidth: number | undefined;
+		let pixelHeight: number | undefined;
+		try {
+			pixelWidth = rawXterm.dimensions?.css.canvas.width;
+			pixelHeight = rawXterm.dimensions?.css.canvas.height;
+		} catch {
+			// Renderer disposed mid-flight; fall through with undefined dimensions.
+		}
 		const roundedPixelWidth = pixelWidth ? Math.round(pixelWidth) : undefined;
 		const roundedPixelHeight = pixelHeight ? Math.round(pixelHeight) : undefined;
 		await this._processManager.setDimensions(rawXterm.cols, rawXterm.rows, undefined, roundedPixelWidth, roundedPixelHeight);
--- a/src/vs/workbench/contrib/terminal/browser/terminalResizeDebouncer.ts
+++ b/src/vs/workbench/contrib/terminal/browser/terminalResizeDebouncer.ts
@@ -33,6 +33,9 @@ export class TerminalResizeDebouncer extends Disposable {
 	}
 
 	async resize(cols: number, rows: number, immediate: boolean): Promise<void> {
+		if (this._store.isDisposed) {
+			return;
+		}
 		this._latestX = cols;
 		this._latestY = rows;
 
@@ -49,12 +52,18 @@ export class TerminalResizeDebouncer extends Disposable {
 		if (win && !this._isVisible()) {
 			if (!this._resizeXJob.value) {
 				this._resizeXJob.value = runWhenWindowIdle(win, async () => {
+					if (this._store.isDisposed) {
+						return;
+					}
 					this._resizeXCallback(this._latestX);
 					this._resizeXJob.clear();
 				});
 			}
 			if (!this._resizeYJob.value) {
 				this._resizeYJob.value = runWhenWindowIdle(win, async () => {
+					if (this._store.isDisposed) {
+						return;
+					}
 					this._resizeYCallback(this._latestY);
 					this._resizeYJob.clear();
 				});
@@ -70,6 +79,9 @@ export class TerminalResizeDebouncer extends Disposable {
 	}
 
 	flush(): void {
+		if (this._store.isDisposed) {
+			return;
+		}
 		if (this._resizeXJob.value || this._resizeYJob.value) {
 			this._resizeXJob.clear();
 			this._resizeYJob.clear();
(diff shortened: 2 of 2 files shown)


src/vs/workbench/contrib/terminal/browser/terminalInstance.ts
```diff
@@ -2047,8 +2056,22 @@ export class TerminalInstance extends Disposable implements ITerminalInstance {
...
+			return;
+		}
+		// `rawXterm.dimensions` proxies to xterm.js' RenderService which throws
+		// `Cannot read properties of undefined (reading 'dimensions')` if the
+		// renderer was disposed between scheduling and invocation (debounced or
+		// idle resize callbacks racing with terminal teardown). Guard against
+		// that here so the optional chaining short-circuits to undefined.
+		let pixelWidth: number | undefined;
+		let pixelHeight: number | undefined;
+		try {
```
anthonykim1, line 2069: This just looks like we want to swallow known error instead of actually addressing root cause. 

Seems like actual throw is from xterm not vscode. 

@bryanchen-d What is the repro step for this dimension error? Is it causing breakage in terminal behavior?

aeschli (approved):

src/vs/workbench/contrib/terminal/browser/terminalInstance.ts
bryanchen-d: It is hard to find the repro step. Essentially the error comes from calling `term.dimensions` after `term.dispose()`, so user shouldn't see visible breakage since the terminal has been disposed. 

I think the try-catch here is no-harm here, but I'd agree we can remove it since we have already check isDisposed above.

anthonykim1: It will be harm if the root cause is somewhere else and this would prevent us from root-causing the issue.

bryanchen-d: That's fair, I made a follow up PR to revert that part, can you review?

anthonykim1: yes, already approved.

src/vs/workbench/contrib/terminal/browser/terminalResizeDebouncer.ts
```diff
@@ -79,6 +91,12 @@ export class TerminalResizeDebouncer extends Disposable {
 
 	@debounce(100)
 	private _debounceResizeX(cols: number) {
+		// The @debounce decorator schedules a setTimeout that is not tied to the
+		// disposable store, so this can fire after the terminal/xterm renderer is
+		// disposed. Bail out to avoid throwing from xterm.js dimension getters.
+		if (this._store.isDisposed) {
+			return;
+		}
```
Tyriar, line 99: @bryanchen-d @anthonykim1 moving off `@debounce` is the better option, this is a hack that doesn't truly fix the lifecycle.

bryanchen-d: I agree, the `@debounce` doesn't seem right here in a disposable class. Created https://github.com/microsoft/vscode/pull/318177 to use `RunOnceScheduler` which can be registered with the Disposable store.

PR #314801, opened 2026-05-06
fix(copilot): populate Context Window indicator for BYOK chat providers

### LINKED ISSUE #314722: Context Window indicator stays at 0% for third-party BYOK chat providers because ExtensionContributedChatEndpoint hardcodes usage to zero
- Copilot Chat Extension Version: 0.45.1
- VS Code Version: 1.118.1 (Stable, macOS arm64)
- Reproduces with all other extensions disabled: **yes**

## Summary

The Copilot Chat **Context Window** widget always renders 0% (or hides) for any model served by a third-party `vscode.lm.registerLanguageModelChatProvider` extension. First-party providers (Anthropic / Gemini / Azure / OpenRouter under `extensions/copilot/src/extension/byok/vscode-node/`) work because they bypass the public LM API. Third-party providers go through [`ExtensionContributedChatEndpoint.makeChatRequest2`](https://github.com/microsoft/vscode/blob/main/extensions/copilot/src/platform/endpoint/vscode-node/extChatEndpoint.ts), which hardcodes:

```ts
usage: { prompt_tokens: 0, completion_tokens: 0, total_tokens: 0,
         prompt_tokens_details: { cached_tokens: 0 } }
```

`stream.usage(...)` in [`toolCallingLoop.ts`](https://github.com/microsoft/vscode/blob/main/extensions/copilot/src/extension/intents/node/toolCallingLoop.ts) then fires with all zeros, and the widget renders `(0+0)/(maxIn+maxOut) = 0%` faithfully. There is no public surface a third-party provider can use to populate it (no `LanguageModelUsagePart`, no `LanguageModelDataPart` mime is consumed for usage in `extChatEndpoint.ts`, `provideTokenCount` is consumed elsewhere).

## Quick repro

```ts
// extension.ts in a minimal extension; activate on '*'
import * as vscode from 'vscode';

export function activate(ctx: vscode.ExtensionContext) {
    const provider: vscode.LanguageModelChatProvider = {
        async prepareLanguageModelChat() {
            return [{
                id: 'repro-model', name: 'Repro', family: 'repro', version: '1',
                maxInputTokens: 100_000, maxOutputTokens: 4_096,
                capabilities: { toolCalling: true, imageInput: false },
            }];
        },
        async provideLanguageModelChatResponse(_m, _msgs, _opts, progress) {
            for (const ch of 'hello world'.split('')) {
                progress.report(new vscode.LanguageModelTextPart(ch));
            }
        },
        async provideTokenCount(_m, text) {
            return typeof text === 'string' ? text.length : 1;
        },
    };
    ctx.subscriptions.push(vscode.lm.registerLanguageModelChatProvider('repro', provider));
}
```

1. Install the extension above and open Copilot Chat.
2. Pick `Repro` in the model picker; send any prompt.
3. **Expected:** Context Window widget reflects accumulated usage.
4. **Actual:** widget stays at 0% across every turn.

Switching the picker back to `gpt-4.1` (or any first-party model) and sending the same prompt updates the widget correctly — confirming it's specific to the third-party path.

## Suggested fix

Recognise a `usage` mime in `extChatEndpoint.ts`, symmetric to the existing `StatefulMarker` / `ContextManagement` handling:

1. Add `Usage = 'usage'` to [`CustomDataPartMimeTypes`](https://github.com/microsoft/vscode/blob/main/extensions/copilot/src/platform/endpoint/common/endpointTypes.ts) (same enum requested in #313920).
2. In `makeChatRequest2`'s stream-consumer loop, parse `{prompt_tokens, completion_tokens, total_tokens, prompt_tokens_details}` from any `LanguageModelDataPart` whose `mimeType === 'usage'` and use it (with the current zero literal as fallback) in the returned `fetchResult`.

```ts
} else if (chunk.mimeType === CustomDataPartMimeTypes.Usage) {
    try { extensionUsage = JSON.parse(new TextDecoder().decode(chunk.data)); }
    catch { /* ignore malformed */ }
}
// ...
usage: extensionUsage ?? { prompt_tokens: 0, completion_tokens: 0, /* ... */ }
```

Zero risk to first-party providers — they don't go through this code path. Backwards compatible — extensions that don't emit it keep getting zeros. Same shape as sibling internal sentinels.

Alternatives I rejected as more expensive: a typed `LanguageModelUsagePart` on `LanguageModelResponsePart2` (proposed-API churn), or synthesising usage from `provideTokenCount` (per-turn double-tokenization, wrong for tool calls / cached tokens).

Happy to put up a PR.

ba-work:
Fixes #314722.

### Problem

The host-internal `ExtensionContributedChatEndpoint.makeChatRequest2` hardcoded a zero-filled `APIUsage` literal in its success envelope regardless of what the extension-contributed `LanguageModelChatProvider` actually streamed back. As a result, the Context Window indicator stayed pinned at 0 / max for every BYOK provider routed through this endpoint (Anthropic, Gemini-native, and any custom `registerLanguageModelChatProvider`).

### Fix

Add a new `CustomDataPartMimeTypes.Usage = 'usage'` branch on the existing `LanguageModelDataPart` switch in the response stream, plus a permissive `parseExtensionContributedUsage(data)` helper that decodes the UTF-8 JSON payload into the host's existing `APIUsage` shape. When the provider doesn't emit a Usage part the host keeps the historical zero-fallback behaviour, so this is purely additive — no behaviour change for existing extensions unless the provider opts in.

Last-write-wins on multiple Usage parts, matching the OpenAI streaming convention where the terminating chunk carries the final tally. The parser tolerates partial payloads (zero-fills missing fields) and malformed JSON (returns `undefined`, host falls back to zeros) so a misbehaving provider cannot break the host.

### Provider opt-in

One extra `progress.report` on the response stream:

```ts
progress.report(new vscode.LanguageModelDataPart(
    new TextEncoder().encode(JSON.stringify({
        prompt_tokens: 12345,
        completion_tokens: 678,
        total_tokens: 13023,
        prompt_tokens_details: { cached_tokens: 9000 }
    })),
    'usage' // CustomDataPartMimeTypes.Usage
));
```

### Defensive coercion

All numeric fields are clamped to non-negative finite numbers via `Math.max(0, ...)` on both the strict and permissive paths. `isApiUsage` only validates `typeof === 'number'`, so without this clamp a provider could emit negatives that would silently corrupt the host's monotonic completion-token counter in `ChatResponseModel.setUsage`. The strict path also preserves any extra fields the provider sent (`cache_creation_input_tokens`, `completion_tokens_details`) for telemetry / future consumers.

### Tests

14 new specs in `extensions/copilot/src/platform/endpoint/vscode-node/test/extChatEndpoint.spec.ts` covering:

- **Parse helper**: full / partial / malformed / non-object / type-coerced / negative-clamp / strict-path-extras payloads
- **End-to-end stream**: zero-fallback regression for #314722, full usage propagation, `cached_tokens` path, malformed-JSON fallback, partial payload, last-write-wins

All 241 specs in `extensions/copilot/src/platform/endpoint/` pass. Lint and `tsc --noEmit` are clean on the three changed files.

### How to verify

1. Install the build with the patched `extChatEndpoint.ts`.
2. Use any extension-contributed BYOK provider (e.g. an Anthropic / Gemini-native `LanguageModelChatProvider`) and have it report usage via the snippet above.
3. Open the **Context Window** widget in the chat side-panel — the bar should now reflect real prompt/completion tokens and (if the provider also reports it) the hatched output-buffer band.

image

### DIFF
--- a/extensions/copilot/src/platform/endpoint/common/endpointTypes.ts
+++ b/extensions/copilot/src/platform/endpoint/common/endpointTypes.ts
@@ -13,10 +13,18 @@ export namespace CustomDataPartMimeTypes {
 	 * Mime type for an extension-contributed token-usage payload, emitted by a
 	 * `vscode.LanguageModelChatProvider` as a `LanguageModelDataPart` in its
 	 * response stream. The `data` is a UTF-8 JSON encoding of an `APIUsage`-shaped
-	 * object. All fields are optional — missing or non-finite numeric values
-	 * are treated as 0, and missing nested objects (`prompt_tokens_details`,
-	 * `completion_tokens_details`) are simply omitted from the parsed result.
-	 * A provider that only knows `prompt_tokens` is free to send just that.
+	 * object. Individual numeric fields are optional — missing or non-finite
+	 * values are treated as 0, negatives are clamped, and nested detail objects
+	 * that are themselves missing or malformed are omitted. The payload must
+	 * carry at least one positive numeric signal (a non-zero top-level counter
+	 * or a non-zero nested detail field) to be accepted; payloads that coerce
+	 * down to all zeros are rejected so a stray empty/malformed chunk can't
+	 * overwrite an earlier valid reading at the dispatch site. Fully-shaped
+	 * payloads (those that satisfy the strict `APIUsage` shape with all three
+	 * top-level counters as `number`) additionally carry a
+	 * `prompt_tokens_details: { cached_tokens: 0 }` placeholder when the
+	 * provider didn't supply that nested object, matching the historical
+	 * zero-fallback shape downstream consumers depend on.
 	 *
 	 * Consumed by `ExtensionContributedChatEndpoint.makeChatRequest2`. When no
 	 * Usage part is emitted, the host falls back to zero counts (which leaves
--- a/extensions/copilot/src/platform/endpoint/common/endpointTypes.ts
+++ b/extensions/copilot/src/platform/endpoint/common/endpointTypes.ts
@@ -9,6 +9,18 @@ export namespace CustomDataPartMimeTypes {
 	export const ThinkingData = 'thinking';
 	export const ContextManagement = 'context_management';
 	export const PhaseData = 'phase_data';
+	/**
+	 * Mime type for an extension-contributed token-usage payload, emitted by a
+	 * `vscode.LanguageModelChatProvider` as a `LanguageModelDataPart` in its
+	 * response stream. The `data` is a UTF-8 JSON encoding of an `APIUsage`
+	 * shape (at minimum `prompt_tokens`/`completion_tokens`/`total_tokens`,
+	 * with optional `prompt_tokens_details.cached_tokens`).
+	 *
+	 * Consumed by `ExtensionContributedChatEndpoint.makeChatRequest2`. When no
+	 * Usage part is emitted, the host falls back to zero counts (which leaves
+	 * the Context Window indicator stuck at 0 — see microsoft/vscode#314722).
+	 */
+	export const Usage = 'usage';
 }
 
 export const CacheType = 'ephemeral';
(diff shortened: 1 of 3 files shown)


ba-work:

PR #314826, opened 2026-05-06
Chat terminal: detect sensitive prompts and surface focus-terminal UI

### LINKED ISSUE #314796: Chat terminal: show confirmation dialog and focus terminal when a password prompt is detected
When the chat agent's `run_in_terminal` tool runs a command that prompts for a password (e.g., `sudo`, `ssh`, `gh auth login`, `npm publish` 2FA), the terminal silently waits for input. For users who rely on a screen reader — and even for sighted users not actively watching the terminal — there is no signal that input is required, and the chat session appears to hang.

### Proposal

Show an explicit confirmation dialog (or comparable UI affordance) the first time a running terminal command appears to be waiting for password / secret input. The dialog should:

- Announce via ARIA live region that the terminal is waiting for input.
- Offer a primary action that focuses the terminal so the user can type their secret directly (we should never round-trip the secret through the model).
- Offer a secondary action to cancel the command.

### Why

- **Accessibility**: screen-reader users currently get no announcement when a tool stalls on a hidden prompt.
- **Discoverability**: even sighted users can miss the prompt because the chat view is in front and the terminal isn't necessarily visible.
- **Safety**: routing the user straight to the terminal is the only correct way to enter a secret — `vscode_askQuestions` and similar would expose the value to the model.

### Notes

- Detection heuristic could reuse the same prompt-detection that the shell-integration / capabilities layer already does (`Password:`, `Passphrase:`, `[sudo] password for ...`, etc.).
- Should not appear for non-secret prompts — those are already covered by other UX.

meganrogge:
Default:

https://github.com/user-attachments/assets/adb282d3-fc1b-447b-82fa-a0799de713fd

Autopilot:

https://github.com/user-attachments/assets/3bf42246-454f-4767-8913-75d08a9343ac

Fixes #314796.

When a command run via the chat terminal tool produces output that looks like a password / passphrase / secret prompt (e.g. `Password:`, `[sudo] password for user:`, `Passphrase for /home/me/.ssh/id_rsa:`, `Enter PIN:`, `Verification code:`), we never want the model to answer — the secret has to come from the user directly. The agent system prompt already instructs the model not to call `ask_questions` for secrets, but until now there was no explicit detection or UI affordance, and screen-reader users got no announcement that input was required.

### Changes

- New `detectsSensitiveInputPrompt` regex covering the common patterns above.
- New `onDidDetectSensitiveInputNeeded` signal on `IOutputMonitor`, fired from the same idle-detection path as `onDidDetectInputNeeded` but routed separately so the regular "looks like a prompt — answer it" UX is bypassed for secrets.
- `runInTerminalTool` (foreground + background paths) listens for the sensitive signal. Behavior depends on the session permission level:

#### Default / Ask sessions (human in the loop)

A chat elicitation appears with:

- A primary **Focus Terminal** action that focuses the terminal so the user can type their secret directly.
- A secondary **Cancel Command** action that kills the running command.

The sensitive signal does **not** resolve the foreground race candidate. While the dialog is up the tool keeps waiting:

- If the user clicks **Focus Terminal** and types the secret, the command finishes and `executionPromise` wins → result is reported as `completed` with normal output.
- If the user clicks **Cancel Command**, execution is cancelled and the race resolves through the same `completed` (cancelled) path.

The secret never round-trips through the model.

#### Auto-approve / Autopilot sessions (no human)

There is no human in the loop and the terminal can be cleaned up after the tool returns, so a Focus Terminal button would be misleading. Instead the tool simply:

1. Cancels the running command.
2. Adds a steering note to the tool result that explicitly tells the model:
   - The user is unavailable.
   - The command has been cancelled.
   - Do **not** retry, do **not** call `send_to_terminal`, do **not** call `vscode_askQuestions` for the secret.
   - Tell the user to run the command interactively when they are available.

The agent

### DIFF
--- a/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/monitoring/outputMonitor.ts
+++ b/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/monitoring/outputMonitor.ts
@@ -20,6 +20,14 @@ export interface IOutputMonitor extends Disposable {
 
 	readonly onDidFinishCommand: Event<void>;
 	readonly onDidDetectInputNeeded: Event<void>;
+	/**
+	 * Fires when the terminal is detected to be waiting for sensitive input
+	 * (e.g. a password, passphrase, token, secret or verification code). This
+	 * is fired *instead of* {@link onDidDetectInputNeeded} so callers can show
+	 * UI that focuses the terminal rather than routing the prompt through the
+	 * agent.
+	 */
+	readonly onDidDetectSensitiveInputNeeded: Event<void>;
 }
 
 export interface IOutputMonitorTelemetryCounters {
@@ -97,6 +105,9 @@ export class OutputMonitor extends Disposable implements IOutputMonitor {
 
 	private readonly _onDidDetectInputNeeded = this._register(new Emitter<void>());
 	readonly onDidDetectInputNeeded: Event<void> = this._onDidDetectInputNeeded.event;
+
+	private readonly _onDidDetectSensitiveInputNeeded = this._register(new Emitter<void>());
+	readonly onDidDetectSensitiveInputNeeded: Event<void> = this._onDidDetectSensitiveInputNeeded.event;
 
 	private _asyncMode = false;
 	private _command = '';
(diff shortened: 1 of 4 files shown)


meganrogge:

dmitrivMS (approved):

RENOxDECEPTION:
this is hot garbage, why would you implement this with no option to the user to override it? Now I have to run every sudo command manually even on my local only raspberry pi? for real?

PR #314841, opened 2026-05-06
sessions: retain attached editor maximize state

hawkticehurst:
## Summary
- retain the maximized state of sessions main editors reopened from the auxiliary-bar-attached flow
- keep modal editor flows unchanged so reopening modal editors does not inherit attached-editor maximize state
- document the behavior in the sessions layout spec and add workbench coverage for the restore logic

## Validation
- `npm run compile-check-ts-native` *(currently fails on `main` due to unrelated errors in `src/vs/workbench/test/browser/componentFixtures/fixtureUtils.ts` referencing `context.host`)*
- `node --experimental-strip-types build/hygiene.ts src/vs/sessions/browser/workbench.ts src/vs/sessions/test/browser/workbench.test.ts src/vs/sessions/LAYOUT.md`
- `./scripts/test.sh --run src/vs/sessions/test/browser/workbench.test.ts --grep "Sessions - Workbench"` *(currently fails in this environment with `app.setPath` undefined in `test/unit/electron/index.js` after Electron launch)*

### DIFF
--- a/src/vs/sessions/LAYOUT.md
+++ b/src/vs/sessions/LAYOUT.md
@@ -259,6 +259,7 @@ setPartHidden(hidden: boolean, part: Parts): void
     - The main editor part is hidden by default but can be shown for explicit editor workflows that target the main editor part
     - Modal editor opens do not change the current main editor visibility state
     - The sessions **Maximize Editor** action temporarily hides the panel when the visible panel is the terminal view, and the matching **Restore Editor** action reopens that terminal panel if maximize hid it
+    - When a maximized main editor attached to the auxiliary bar closes, the next attached-editor reopen restores that maximized state; modal editor flows do not participate in this restore behavior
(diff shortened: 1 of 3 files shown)


hawkticehurst:
Addressed the review feedback in `2cddf4ea39b`.

- Clear the remembered attached-editor maximize restore state when the auxiliary bar is hidden.
- Add regression coverage for the hide -> show -> reopen case so maximize is no longer restored after the auxiliary bar was hidden.

Validation status is unchanged from earlier: hygiene passes on the touched files, while the repo-wide compile check still hits the unrelated `fixtureUtils.ts` errors already present on top of `main`, and the targeted unit-test script still fails in this environment during Electron test bootstrap (`app.setPath` undefined in `test/unit/electron/index.js`).

dmitrivMS (approved):

PR #314847, opened 2026-05-06
Strip NODE_OPTIONS in removeDangerousEnvVariables

meganrogge:
NODE_OPTIONS can inject arbitrary flags (--require, --inspect, etc.) into forked Node processes, which can lead to crashes and unexpected behavior in the extension host and other forked processes (similar reasoning to the existing DEBUG and LD_PRELOAD\cleanups). Add it to the list of variables stripped by removeDangerousEnvVariables


rzhao271 (approved):

rzhao271:
CC @deepak1556 in case some of the options are still required

deepak1556:
If this doesn't strip the NODE_OPTIONS set in users' terminal profile then all good (options user would expect in their integrated terminal), our processes already only respect a restricted set of flags so no regressions there.

PR #314872, opened 2026-05-06
Support for RTL text in find decorations

### LINKED ISSUE #233573: When searching rtl letters in Arabic using `Ctrl+F` the word gets reversed and disconnected

- VS Code Version:  1.95.2
- OS Version: Linux x64 6.11.6-arch1-1

Steps to Reproduce:

1.  Open A file with Arabic text
2. click `Ctrl+F`
3. search for a part of the word
4. observe the word being reversed and incorrectly aligned

Wrong behavior below:
Image

How it looks when highlighted using side search bar (correct)
Image

kamyarmg:
## Summary

Fixes incorrect visual ordering when Find highlights matches inside mixed LTR/RTL text.

In lines containing RTL text, Find no longer applies inline styling that splits the rendered content into multiple bidi runs. Instead, it uses decorations that preserve the correct visual ordering of highlighted text.

Fixes #233573.

## Problem

When Find highlights a match inside a line containing mixed LTR and RTL text, the highlight can split the rendered content into multiple inline segments.

In some cases, this changes the visual ordering of the RTL portion of the line, causing highlighted matches to appear with characters in the wrong order even though the underlying text is correct.

## Root Cause

Find match decorations relied on inline text styling.

While this works correctly for normal LTR text, applying inline styling inside mixed bidi content can split a visual bidi run into separate segments. Once split, browser bidi layout may reorder the segments visually, making the highlighted substring appear incorrect.

## Fix

This change makes Find match decorations aware of RTL content.

- Lines containing RTL text now use decorations that avoid inline text styling.
- LTR-only lines continue using the existing rendering behavior.
- Current match and regular match decorations now share the same RTL-aware decoration selection logic.

This preserves Find highlighting while avoiding the bidi run splitting that caused incorrect visual ordering.

### Original Text

Screenshot From 2026-05-07 02-05-35

### Before

Screenshot From 2026-05-07 02-11-49

### After

Screenshot From 2026-05-07 02-13-07

## How to Test

1. Open a file containing mixed LTR/RTL text on the same line.
2. Search for a substring inside the RTL portion of the line.
3. Verify that the highlighted match preserves the correct visual character order.
4. Repeat the same test on an LTR-only line and verify that Find highlighting still behaves normally.

### DIFF
--- a/src/vs/editor/contrib/find/browser/findDecorations.ts
+++ b/src/vs/editor/contrib/find/browser/findDecorations.ts
@@ -8,7 +8,7 @@ import { IDisposable } from '../../../../base/common/lifecycle.js';
 import { IActiveCodeEditor } from '../../../browser/editorBrowser.js';
 import { Position } from '../../../common/core/position.js';
 import { Range } from '../../../common/core/range.js';
-import { FindMatch, IModelDecorationsChangeAccessor, IModelDeltaDecoration, MinimapPosition, OverviewRulerLane, TrackedRangeStickiness } from '../../../common/model.js';
+import { FindMatch, IModelDecorationOptions, IModelDecorationsChangeAccessor, IModelDeltaDecoration, MinimapPosition, OverviewRulerLane, TrackedRangeStickiness } from '../../../common/model.js';
 import { ModelDecorationOptions } from '../../../common/model/textModel.js';
 import { minimapFindMatch, overviewRulerFindMatchForeground } from '../../../../platform/theme/common/colorRegistry.js';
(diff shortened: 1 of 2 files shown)


kamyarmg:
Hi @alexdima — just checking in on this PR. It fixes a bidi rendering issue affecting Arabic/Persian/Hebrew users. I'd appreciate any feedback when you have a chance. Thanks!

kamyarmg:
@alexdima  For those of us who frequently work with a mix of left-to-right (LTR) and right-to-left (RTL) layouts, this issue and its fix would be greatly appreciated, as it's quite important for our workflow. If you have a chance, could you please take a look at it? Thanks!

PR #314882, opened 2026-05-06
chat: seed row height in renderElement to break ResizeObserver loop

bryanchen-d:
The chat row `ResizeObserver` callback in `ChatListItemRenderer` dispatched `fireItemHeightChange` synchronously. The downstream `updateLastItemMinHeight` chain writes a CSS variable that resizes a sibling row, and Chromium's RO algorithm can't redeliver same-depth resize entries inside the in-flight observation phase — so the synthetic `ResizeObserver loop completed with undelivered notifications` warning fires.

Telemetry attribution shipped in #314411 confirms this row observer (named `ChatListItemRenderer.itemHeight`) is the dominant source of the loop warning across our error stream.

### Fix

Seed `currentRenderedHeight` synchronously at the end of `renderElement`, while `_elementBeingRendered` is still set. The existing dispatch guard in `fireItemHeightChange` (`template.currentElement !== this._elementBeingRendered`) suppresses the event fire — so the call records the row's actual mounted box size without re-entering the in-flight tree splice.

The ResizeObserver's first post-mount delivery then sees the same blockSize and hits the same-height short-circuit in `fireItemHeightChange`, returning without any layout-mutating work. The loop warning goes away, and `currentRenderedHeight` reflects the real measured height (rather than the 150-px estimate the original `setTimeout(0)` deferral left behind).

### Validation

- Reproduced via streaming chat responses against a long thread, captured the warning over CDP using a Playwright harness.
- Before: `ResizeObserver loop completed with undelivered notifications` fires on each new request row mount and during streaming.
- After: warning no longer fires; row heights still update during streaming with no perceptible lag.

### Notes

- This replaces the earlier `setTimeout(0)` + coalescing approach. That worked but had two downsides Rob flagged: it deferred the height update onto a macrotask (visible bounce), and it discarded the real first-mount measurement so `currentRenderedHeight` stayed pegged to the estimate.
- The seed call is one extra synchronous read of `borderBoxSize` per row mount. The RO callback itself is unchanged from `main`.

### DIFF
--- a/src/vs/workbench/contrib/chat/browser/widget/chatListRenderer.ts
+++ b/src/vs/workbench/contrib/chat/browser/widget/chatListRenderer.ts
@@ -657,10 +657,13 @@ export class ChatListItemRenderer extends Disposable implements ITreeRenderer<Ch
 				return;
 			}
 			const blockSize = entry.borderBoxSize.at(0)?.blockSize;
-			if (typeof blockSize !== 'number') {
-				return;
+			// Track the latest measured size so we can coalesce bursts. If
+			// blockSize is unavailable in the current environment we still
+			// schedule a dispatch so `fireItemHeightChange` falls back to
+			// `getBoundingClientRect()` (preserving prior behavior).
+			if (typeof blockSize === 'number') {
+				pendingBlockSize = blockSize;
 			}
-			pendingBlockSize = blockSize;
 			if (pendingDispatchHandle !== undefined) {
 				return;
 			}
@@ -668,9 +671,7 @@ export class ChatListItemRenderer extends Disposable implements ITreeRenderer<Ch
 				pendingDispatchHandle = undefined;
 				const size = pendingBlockSize;
 				pendingBlockSize = undefined;
-				if (size !== undefined) {
-					this.fireItemHeightChange(template, size);
-				}
+				this.fireItemHeightChange(template, size);
 			}, 0);
 		}));
 		templateDisposables.add(resizeObserver.observe(rowContainer));
--- a/src/vs/workbench/contrib/chat/browser/widget/chatListRenderer.ts
+++ b/src/vs/workbench/contrib/chat/browser/widget/chatListRenderer.ts
@@ -636,43 +636,11 @@ export class ChatListItemRenderer extends Disposable implements ITreeRenderer<Ch
 			}
 		}));
 
-		// Chat rows can drive a `ResizeObserver loop completed with undelivered
-		// notifications` warning because dispatching a height change re-enters
-		// layout synchronously inside the observer callback, which causes the
-		// observer to fire again in the same animation frame. Defer the
-		// dispatch onto the next task with setTimeout(0) so the observer
-		// callback returns cleanly before the layout mutation runs. Coalesce
-		// repeated observations into a single dispatch using the latest size.
-		let pendingBlockSize: number | undefined;
-		let pendingDispatchHandle: ReturnType<typeof setTimeout> | undefined;
-		templateDisposables.add(toDisposable(() => {
-			if (pendingDispatchHandle !== undefined) {
-				clearTimeout(pendingDispatchHandle);
-				pendingDispatchHandle = undefined;
-			}
-		}));
 		const resizeObserver = templateDisposables.add(new dom.DisposableResizeObserver('ChatListItemRenderer.itemHeight', (entries) => {
 			const entry = entries[0];
-			if (!entry) {
-				return;
+			if (entry) {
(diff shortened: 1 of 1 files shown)


src/vs/workbench/contrib/chat/browser/widget/chatListRenderer.ts
```diff
@@ -636,11 +636,42 @@ export class ChatListItemRenderer extends Disposable implements ITreeRenderer<Ch
...
 			const entry = entries[0];
-			if (entry) {
-				this.fireItemHeightChange(template, entry.borderBoxSize.at(0)?.blockSize);
+			if (!entry) {
+				return;
+			}
+			const blockSize = entry.borderBoxSize.at(0)?.blockSize;
+			if (typeof blockSize !== 'number') {
+				return;
+			}
```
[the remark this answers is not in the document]
bryanchen-d: Good catch — fixed in 01a64dcda43. The dispatch is now always scheduled; only the cached `pendingBlockSize` is gated on `typeof blockSize === 'number'`, so when `borderBoxSize` is empty/missing we still call `fireItemHeightChange(template, undefined)` and fall back to `getBoundingClientRect()` as before.

src/vs/workbench/contrib/chat/browser/widget/chatListRenderer.ts
```diff
@@ -636,11 +636,43 @@ export class ChatListItemRenderer extends Disposable implements ITreeRenderer<Ch
...
+			// blockSize is unavailable in the current environment we still
+			// schedule a dispatch so `fireItemHeightChange` falls back to
+			// `getBoundingClientRect()` (preserving prior behavior).
+			if (typeof blockSize === 'number') {
+				pendingBlockSize = blockSize;
+			}
+			if (pendingDispatchHandle !== undefined) {
+				return;
 			}
+			pendingDispatchHandle = setTimeout(() => {
```
roblourens: Deferring this is going to add the kinds of issues I was talking about. It's not expected that fireItemHeightChange would cause the size of this element to change, we should only change the parent list row to accomodate it, do you know what causes its height to change?

roblourens (changes requested):

src/vs/workbench/contrib/chat/browser/widget/chatListRenderer.ts
roblourens: And since this is mine, I can investigate, do you know of repro steps?

roblourens: It's interesting because I don't remember seeing that warning, but this observer fires really often

bryanchen-d: Sure, for repro: Start a new chat session with some prompt like "research the repo"

bryanchen-d: The cause of the RO loop notification is because when request row is mounted, the first RO callback is entered, which leads to fireItemHeightChange() in the same frame, resulting in RO loop notification. 

Per discussion, updating the fix to fireItemHeightChange() right after rendering the request row so the first RO callback will be no-op.

roblourens:
Something else to look into- when this error is triggered and I hit a breakpoint at https://github.com/microsoft/vscode/blob/7cf776a1209c9fddc39765b37ff6696337fbed74/src/vs/base/browser/dom.ts#L2156, I don't see that error anywhere in the Window output channel or the devtools console. Can you look into why that is? I know I've seen it somewhere before. We should make sure we have that visibility into errors.

bryanchen-d:
> Something else to look into- when this error is triggered and I hit a breakpoint at
> 
> https://github.com/microsoft/vscode/blob/7cf776a1209c9fddc39765b37ff6696337fbed74/src/vs/base/browser/dom.ts#L2156
> 
> , I don't see that error anywhere in the Window output channel or the devtools console. Can you look into why that is? I know I've seen it somewhere before. We should make sure we have that visibility into errors.

The RO error is sent via `window.onerror` which is not hooked to the workbench's output channel. 

In devtools console, it is under the 'verbose' level, I just checked that it shows up if you turn on All Levels.

roblourens:
> In devtools console, it is under the 'verbose' level, I just checked that it shows up if you turn on All Levels.

Oh interesting, thanks

roblourens:
This change would add another forced re-layout at a time when we don't necessarily need to know the size, so I don't think we should do that. Here's another suggestion https://github.com/microsoft/vscode/pull/315486. This would let us initialize the `currentRenderedHeight` at the time when we already get the size, but avoid triggering other work (because the height did not change, it only got its initial measurement). What do you think?

PR #315054, opened 2026-05-07
fix: memory leak extension actions

SimonSiefke:
Fixes some memory leaks in extension actions. 

### Details

Actions can be disposable. In the case of extension actions, they are created and immediately being run. But a `dispose` call seems to be missing.

```ts
if(extension) {
  const action = instantiationService.createInstance(SetColorThemeAction); // disposable created
  action.extension = extension;
  return action.run(); // run, but never disposed
}
```

### Change

The change is calling `dispose` after running the action:

```ts
if(extension) {
 const action = instantiationService.createInstance(SetColorThemeAction); // disposable created
  action.extension = extension;
  try {
	  return await action.run();
  } finally {
	  action.dispose(); // Ensure to dispose the action
  }
```

<hr>

Maybe in the future this could be simplified with the new [using](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/using) keyword (assuming actions would have a `Symbol.dispose` property: 

```ts
if(extension) {
  using action = instantiationService.createInstance(SetColorThemeAction); // disposable created
  action.extension = extension;
  return action.run();
```

### Before

When opening the extension view, and right clicking on an extension and selecting "Install Another Version", then closing the quickpick 17 times, the number of `InstallAnotherVersionAction` and related functions seems to grow each time:

extension-install-specific-version-old

### After

No more leak is detected.


dmitrivMS (approved):

dmitrivMS:
@SimonSiefke Thank you!

TylerLeonhardt (approved):

PR #315066, opened 2026-05-07
fix: restore telemetry app name for agents window

deepak1556:
Followup to https://github.com/microsoft/vscode/pull/314409

cc @sandy081


sandy081 (approved):

sandy081:
Thanks for fixing this

Issue #315196, opened 2026-05-08
Bug: Add File to Chat does nothing when right-clicking a non-active editor tab


damonxue:
Does this issue occur when all extensions are disabled?: Yes/No

- VS Code Version: 1.119.0
- OS Version: 

### Description

When using the **"Add File to Chat"** option from the editor tab right-click context menu (`MenuId.EditorTitleContext`), the action only works if the target tab is the currently active editor. Right-clicking on any **non-active** tab and selecting "Add File to Chat" does nothing — the file is not attached to the chat input.

### Steps to Reproduce

1. Open two or more files in the editor.
2. Make sure one file is active (focused).
3. Right-click on any **other (non-active)** tab.
4. Select **"Add File to Chat"**.

### Expected Behavior

The right-clicked file should be attached to the Chat input, regardless of whether the tab is currently active.

### Actual Behavior

Nothing happens. The file is not added to the Chat input.

### Root Cause (for contributors)

In chatContextActions.ts, the `_getResources` method calls `_getEditorResources` with a wrapped array instead of spreading the arguments:

```typescript
// Bug: passes `args` as a single element, creating nested array
this._getEditorResources(accessor, args)

// Fix: spread args correctly
this._getEditorResources(accessor, ...args)
```

Because `_getEditorResources` uses `...args` rest parameter internally and passes `args` directly to `resolveCommandsContext(args, ...)`, the nesting causes `resolveCommandsContext` to receive `[[uri, editorCommandsContext]]` instead of `[uri, editorCommandsContext]`. This results in an empty `groupedEditors` array and no file being attached.

### Why Active Tab Appears to Work

When triggered from inside the editor content area (right-click on editor body, not the tab), `args[1]` may not satisfy `isEditorCommandsContext`, causing the code to fall back to `editorService.activeEditor` — which coincidentally is the active file.

### Affected File

chatContextActions.ts

### Fix

Change line 95 from:
```typescript
const contexts = isEditorCommandsContext(args[1]) ? this._getEditorResources(accessor, args) : ...
```
to:
```typescript
const contexts = isEditorCommandsContext(args[1]) ? this._getEditorResources(accessor, ...args) : ...
```

PR #315197, opened 2026-05-08
Add File to Chat" does nothing when right-clicking a non-active editor tab

damonxue:
[https://github.com/microsoft/vscode/issues/315196](https://github.com/microsoft/vscode/issues/315196)


justschen (approved):
thanks!!

dmitrivMS (approved):

PR #315213, opened 2026-05-08
Improves animated value

hediet:


lszomoru:
@hediet, could we move this out of editor/contrib so that it can be used in other places as well? Thanks!

lszomoru (approved):

PR #315230, opened 2026-05-08
fix: restore word wrap in diff editor left pane after window resize

yeousunn:
Fixes word wrap breaking in the original (left) editor of the diff view after resizing the window narrower than the inline breakpoint (~900px) and back.

### Root Cause

When the diff editor narrows below `renderSideBySideInlineBreakpoint` (900px default), it switches to inline mode and sets `wordWrapOverride2 = 'off'` on the original editor to suppress wrapping. When the window widens back to side-by-side mode, only `wordWrapOverride1` was restored — `wordWrapOverride2` remained `'off'` and silently took precedence, permanently disabling word wrap on the left editor until the tab was cycled.                                                                                                                           

The right (modified) editor was unaffected because `_adjustOptionsForRightHandSide` never sets `wordWrapOverride2`. 

### Fix

Reset `wordWrapOverride2 = 'inherit'` in the side-by-side branch of `_adjustOptionsForLeftHandSide`, restoring delegation to `wordWrapOverride1`.

### How to Test

1. Open two files in the diff editor
2. Enable word wrap
3. Ensure files contain long lines
4. Drag the window narrower than ~900px - diff switches to inline view
5. Drag back to full width - diff returns to side-by-side
6. **Before fix:** word wrap missing on left side (text overflows horizontally)                                                                                      7. **After fix:** word wrap correctly restored on both sides


yeousunn:
Hey, just bumping this, been sitting in triage-needed since May 8 with no review yet. It's a small, scoped fix (one line: reset wordWrapOverride2 to 'inherit'). Would appreciate a look when someone has a sec. Thanks! @hediet

PR #315245, opened 2026-05-08
Enable agents window in stable

sandy081:
Removes all `quality !== 'stable'` guards that were blocking the agents window from being accessible in stable builds.

## Changes

- **`app.ts`** — Remove quality check from `--agents` CLI flag and from `tryConsumeAgentsLastRunningMarker` (agents window now restores on startup in stable too)
- **`launchMainService.ts`** — Remove quality check when re-launching with `--agents` flag; remove now-unused `IProductService` dependency
- **`constants.ts`** — Remove `ProductQualityContext.notEqualsTo('stable')` from `OPEN_AGENTS_WINDOW_PRECONDITION`, enabling the "Open Agents Window" action in stable
- **`agentSessionsBanner.ts`** — Remove quality check from `canShowAgentsBanner` so the "Try out the new Agents" banner shows in stable (for users with Copilot access); remove now-unused `productService` parameter
- **`chatTipCatalog.ts`** — Remove the `tip.openAgentsWindow` chat tip shown above the chat input — already enough promotion via the banner and actions

## Not changed
- `agentsAppMergedBanner.contribution.ts` — Kept gated to non-stable (migration notice for users of the old standalone Agents app, which never existed in stable)
- `agentSessionsBanner.ts` (`canShowAgentsBanner`) — The banner itself still respects AI entitlement (`sentiment.hidden`, `sentiment.disabled`), just no longer requires non-stable
- `agentSessionsExperiments.contribution.ts` — "Agent Quick Input" toggle kept as Insiders-only experimental feature

### DIFF
--- a/src/vs/code/electron-main/app.ts
+++ b/src/vs/code/electron-main/app.ts
@@ -1337,7 +1337,7 @@ export class CodeApplication extends Disposable {
 		const args = this.environmentMainService.args;
 
 		// Handle agents window first based on context
-		if ((args['agents'] && this.productService.quality !== 'stable')) {
+		if (args['agents']) {
 			return windowsMainService.openAgentsWindow({
 				context,
 				cli: args,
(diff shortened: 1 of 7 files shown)


lszomoru (approved):

Caffeine19:
Hi, I don't think the Agents Window is ready for stable. It still has a lot of development work remaining.
For example, the Agents Window can't use the extensions installed in VS Code, which results in no LSP support for my Lua files.

In agents window:
image

In vscode window:
image

PR #315268, opened 2026-05-08
fix: replace typescript.tsdk.desc with new js/ts.tsdk.path

thernstig:
The configuration "js/ts.tsdk.path" has an incorrect description, as it referenced the deprecated "typescript.tsdk" description.


extensions/typescript-language-features/package.json
```diff
@@ -167,7 +167,7 @@
         "properties": {
           "js/ts.tsdk.path": {
             "type": "string",
-            "markdownDescription": "%typescript.tsdk.desc%",
```
mjbvz, line 170: The key doesn't really matter (and I wanted to keep it consistent so we don't need re-localization)

Can you just update the setting string instead

mjbvz (changes requested):

mjbvz (approved):
Thanks

dmitrivMS (approved):

PR #315279, opened 2026-05-08
Bumps `@vscode/ripgrep` dependency version from 1.17.1 to 1.18.0, adopts inline binaries.

hediet:
1.17.1 would download the binaries in postinstall, in 1.18.0 the binaries are shipped with the package.


hediet:
### Why `build/darwin/create-universal-app.ts` and `verify-macho.ts` need changes

**How the universal app worked on `main` with `@vscode/ripgrep@1.17.1`:**

The legacy package has a `postinstall` (`lib/postinstall.js`) that downloads a single `rg` binary based on `process.env.npm_config_arch`. CI sets `npm_config_arch=$(VSCODE_ARCH)` per arch build (see `build/azure-pipelines/darwin/product-build-darwin-node-modules.yml` and `.../steps/product-build-darwin-compile.yml`), so:

- x64 build tree: x64 Mach-O at `node_modules/@vscode/ripgrep/bin/rg`
- arm64 build tree: arm64 Mach-O at the **same path** `node_modules/@vscode/ripgrep/bin/rg`

`makeUniversalApp` (from `vscode-universal-bundler` / `@electron/universal`) walks both trees in parallel. When it finds two Mach-O files at the same path with different bytes, it automatically `lipo`s them into a fat universal Mach-O. No special config needed → main's universal script has no ripgrep entry at all.

**Why `@vscode/ripgrep-universal@1.18.0` needs explicit handling:**

The new package has no postinstall; binaries are shipped inline in per-arch subfolders:

```
bin/darwin-x64/rg
bin/darwin-arm64/rg
bin/linux-x64/rg
bin/win32-x64/rg.exe
...
```

So the binary paths are now **different** between archs (`bin/darwin-x64/rg` vs `bin/darwin-arm64/rg`), not the same path with different bytes. The bundler's auto-lipo path can't pair them, so we have to teach it:

- `crossCopyPlatformDir(... 'ripgrep-universal/bin/' + plat)` — ensures both arch subdirs exist in both input trees before merging.
- `filesToSkip` entries for `bin/darwin-x64/**` and `bin/darwin-arm64/**` — bundler won't flag them as differing.
- `x64ArchFiles` entries — declare them arch-specific so the bundler preserves them side-by-side instead of trying to lipo.
- `verify-macho.ts` allow-list entry — the post-build Mach-O verifier accepts single-arch binaries at the new paths (they're never fused into a fat binary; ripgrep picks the right subdir at runtime via `process.platform`/`process.arch`).

TL;DR: main relied on the postinstall producing a same-path-per-arch binary that the universalizer auto-lipo'd. The new inline-binaries layout puts each arch in its own subdir, which defeats auto-lipo and requires the bookkeeping in this PR.

dbaeumer (approved):

PR #315349, opened 2026-05-08
chat: don't recommend agent plugins when none are in the marketplace cache

meganrogge:
The 'this workspace recommends N agent plugins' notification fired whenever the workspace listed any `enabledPlugins`, even if none of those keys matched a plugin in `lastFetchedPlugins`. Clicking **Show Plugins** then opened `@agentPlugins @recommended` with an empty list.

Filter the recommended count to keys that are actually present in the cached marketplace results so the notification only appears when the resulting view will have entries.


connor4312 (dismissed):

meganrogge:

connor4312 (approved):

PR #315367, opened 2026-05-08
Strip Markdown when announcing chat input notification to screen readers

dmitrivMS:
Follow-up to #315360 (review comment from Copilot on the merged PR).

`active.message` can be an `IMarkdownString` (e.g. the agent-host permission notification wraps paths in backticks). The previous fix passed `.value` straight to `aria.status(...)`, so screen readers would read Markdown syntax — backticks, link targets, etc. — verbatim.

This PR:
- Imports `renderAsPlaintext` from `vs/base/browser/markdownRenderer` and uses it to strip Markdown before announcing.
- Keeps the dedupe signature based on the raw `.value`, which is cheap to compute, and only renders the plaintext after the signature check passes — so we don't pay the parse cost on unrelated `onDidChange` fires.

Verified: `Core - Typecheck` task reports `Finished compilation with 0 errors`.


meganrogge (approved):
thank you!

PR #315406, opened 2026-05-08
Anthropic Messages API: own cache_control placement; drop heuristic breakpoints

bhavyaus:
The Anthropic Messages API path already places `cache_control` deterministically (tools, last system block, last two messages — 4-slot layout) inside `messagesApi.ts`. The prompt-tsx / heuristic cache breakpoints we were also emitting in the agent intent ended up duplicating or shifting those slots, so this PR makes ownership explicit:

- `agentIntent.ts`: when the endpoint is the Messages API (`apiType === 'messages'`), suppress prompt-tsx `enableCacheBreakpoints` and skip the post-render `addCacheBreakpoints` pass. The Chat Completions path is unchanged.
- Removes the now-unused `github.copilot.chat.anthropic.cacheBreakpoints.lastTwoMessages` setting (and its NLS string) — the Messages API is now the single source of truth for breakpoint placement on Anthropic.
- Updates Anthropic agent prompt snapshots to match the new (no-prompt-tsx-breakpoints) output.

> [!NOTE]
> - most of the files are test file updates. So don't be alarmed.
> - msbench evals are still pending; will follow up with results.
> - The snapshot diffs make it look like cache breakpoints were removed, but `messagesApi.ts` re-adds them deterministically at request time (tools + last system block + last two messages). The snapshots only capture the prompt-tsx output, not the final wire payload — tracking this snapshot fidelity gap as debt to clean up in a follow-up.


bhavyaus:
### msbench results — vscbench, claude-opus-4.7

| Run | Notes | Pooled cache % (matched n=82) |
|---|---|---:|
| `25541438557` | Baseline (no #315406) | 94.58 % |
| `25589777772` | With #315406 | **96.16 %** |

**Δ cache rate: +1.58 pp**  ·  **Cache-miss tokens: −22 %**  ·  Resolution rate unchanged (56.4 %).

dmitrivMS (approved):

PR #315370, opened 2026-05-08
Fixed tokenizer bug for customOAI models

vikramnitin9:
[This](https://github.com/microsoft/vscode/commit/44ffe22e115) commit introduced a bug where evals with a customOAI model endpoint would fail with `Unknown tokenizer: undefined` for any tool call. This fixes that issue.

PR #315435, opened 2026-05-09
Default missing tokenizer metadata

kangarko:


kangarko:
This solves custom models such as AWS Bedrock integration

dmitrivMS:
This has already been fixed in https://github.com/microsoft/vscode/pull/315370
See https://github.com/microsoft/vscode/issues/314419

kangarko:
Thank you! Is AWS Bedrock support planned btw? I had to patch VSCode for it to show in the model picker and use a third party plugin. Azure is already there, i assume it should be fairly easy to add for BYOK enterprise customers. @dmitrivMS . Thanks.

dmitrivMS:
> Thank you! Is AWS Bedrock support planned btw? I had to patch VSCode for it to show in the model picker and use a third party plugin. Azure is already there, i assume it should be fairly easy to add for BYOK enterprise customers. @dmitrivMS . Thanks.

@vijayupadya do you know?

PR #315445, opened 2026-05-09
:bug: Fix git.openFile not working in checkpoint diff editors

Caffeine19:
## Problem

The `git.openFile` command fails when invoked from checkpoint diff editors (e.g., Copilot's per-checkpoint "Changed Files" diff view). Checkpoint diffs use `chat-editing-snapshot-text-model://` URIs, which are virtual read-only URIs provided by a content provider. The `git.openFile` command only handled `git://` and `file://` schemes, so:

1. **No "Open File" button** appeared in the diff editor toolbar — the `editor/title` menu `when` clause regex only matched `git` and `file` schemes.
    
    Changed files from checkpoint:
    image
    No "open file" button in toolbar:
    image

2. **Keybinding fallback also failed** — when triggered via shortcut with no arguments, the active editor's snapshot URI was passed through as-is, resulting in opening a read-only virtual document instead of the actual editable source file.

    Add keybinding for "git.openFile" command:
    image
    Press keybindings in diff editor, only open a readonly file that cant edit:
    image
    

### `extensions/git/src/commands.ts`

- **Explicit URI argument**: Added a branch for `chat-editing-snapshot-text-model` scheme that extracts the real file path via `Uri.file(arg.path)` (the snapshot URI's `.path` component contains the actual file system path).
- **Active editor fallback**: When no argument is provided (keybinding), the active editor URI is now checked for the snapshot scheme and converted to a `file://` URI before proceeding.
  
  Open a diff editor from checkpoint changed file list, press keydings for "git.openFile" command, now it open the source file that editable:
  https://github.com/user-attachments/assets/7810a2e5-3300-4bc8-8f6f-b709a5d8ebab

### `extensions/git/package.json`

- Extended the `editor/title` `when` clause regex from `/^git$|^file$/` to `/^git$|^file$|^chat-editing-snapshot-text-model$/`, so the "Open File" icon button now appears in checkpoint diff editor toolbars, consistent with other diff editors.

  Add the open files button into toolbar
  image

## Testing

1. Trigger a Copilot agent session that modifies a file.
2. In the checkpoint changed files list, click on a file to open the diff editor.
3. The "Open File" icon should now appear in the diff editor title bar.
4. Clicking it (or using the `git.openFile` keybinding) should open the actual editable source file.


Caffeine19:
Maybe I shouldn't change the code in the Git extension, but rather modify the chat part.

PR #315456, opened 2026-05-09
Fix PHP highlighting for comment-like strings

### LINKED ISSUE #204065: Syntax highlighting error in PHP function parameter when string literal contains block comment

- VS Code Version: 
- OS Version: 
- Version: 1.86.0
Commit: 05047486b6df5eb8d44b2ecd70ea3bdf775fd937
Date: 2024-01-31T10:29:11.933Z
Electron: 27.2.3
ElectronBuildId: 26495564
Chromium: 118.0.5993.159
Node.js: 18.17.1
V8: 11.8.172.18-electron.0
OS: Darwin arm64 23.3.0

Steps to Reproduce:

1. Write valid PHP code with a string literal that contains `/*`.
2. Observe that the rest of the code highlighting will be broken until a new string literal containing `*/` is introduced. See the example below:

```php
<?php

declare(strict_types=1);

$jsCode = '/* <![CDATA[ */ var data = {"a": 1, "b": 2} /* ]]> */;';
$jsCode = str_replace("/* <![CDATA[ */", '', $jsCode);
$jsCode = str_replace("/* ]]> */", '', $jsCode);
var_dump($jsCode);
```
Refer to line 6 and 7 for the error:

php-code

Creeper0809:
Fixes #204065

## Summary

PHP slash-delimited strings are highlighted as regex-like strings by the bundled PHP grammar. That works for common PCRE usages such as `preg_match("/foo/", $value)`, but it also causes normal strings that begin with a block-comment marker to be mis-tokenized.

For example, this valid PHP string currently starts a `string.regexp.double-quoted.php` scope:

```php
str_replace("/* <![CDATA[ */", '', $jsCode);
```

Because the regex rule sees the string as `"/.../"`, the `[` in `<![CDATA[` can leave the tokenizer inside a regex character class and break highlighting for the rest of the call or following code.

This change excludes strings beginning with `/*` from the regex-string rule, so comment-like string contents are treated as normal quoted strings. Regular slash-delimited regex strings continue to be highlighted as regex strings.

The same patch is also added to `extensions/php/build/update-grammar.mjs` so it is preserved when the PHP grammar is refreshed from upstream.

## Testing

- Ran `node --check extensions/php/build/update-grammar.mjs`
- Ran `git diff --check`
- Verified tokenization with `vscode-textmate` / `vscode-oniguruma`:
  - `str_replace("/* <![CDATA[ */", '', $jsCode)` is tokenized as `string.quoted.double.php`, not `string.regexp.double-quoted.php`
  - `str_replace("/* ]]> */", '', $jsCode)` is tokenized as `string.quoted.double.php`
  - `preg_match("/foo[bar]+/i", $jsCode)` still receives `string.regexp.double-quoted.php`

### DIFF
--- a/extensions/php/build/update-grammar.mjs
+++ b/extensions/php/build/update-grammar.mjs
@@ -36,6 +36,28 @@ function fixBadRegex(grammar) {
 		throw new Error(`fixBadRegex callback couldn't patch ${msg}. It may be obsolete`);
 	}
 
+	function patchCommentLikeStringRegex(pattern, msg) {
+		if (!pattern) {
+			fail(msg);
+		}
+
+		const begin = pattern.begin;
+		if (typeof begin !== 'string') {
+			fail(`${msg}.begin`);
+		}
+
+		if (/^['"]\/\(\?!\\\*\)/.test(begin)) {
+			return;
+		}
+
+		const patchedBegin = begin.replace(/^(['"])\/(?=\(\?=)/, '$1/(?!\\*)');
+		if (patchedBegin === begin) {
+			fail(`${msg}.begin`);
+		}
+
+		pattern.begin = patchedBegin;
+	}
+
 	const scopeResolution = grammar.repository['scope-resolution'];
 	if (scopeResolution) {
 		const match = scopeResolution.patterns[0].match;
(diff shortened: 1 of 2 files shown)


Creeper0809:

Vineshnayak:
The use of (?!\\*) is a solid fix for the PHP comment/regex ambiguity. However, the exact string match for begin in update-grammar.mjs makes the build script fragile; if the upstream grammar changes even slightly (e.g., adding a flag), the fail() function will trigger and break the build. It would be safer to use a regex-based replacement or a partial match to apply this patch more robustly.

Creeper0809:
Thanks for the review. I addressed this in the latest push.

`update-grammar.mjs` no longer relies on an exact full-string match for the upstream `begin` pattern. I replaced that with a small helper that:

- skips the patch if `(?!\*)` is already present,
- applies the patch to the expected `"/(?=...)` / `'/(?=...)` regex-string begin shape by inserting `(?!\*)` after the delimiter,
- still fails if the upstream pattern changes enough that applying the patch would no longer be safe.

That should make the patch resilient to small upstream changes, such as added flags, while still avoiding a silent incorrect rewrite.

PR #315486, opened 2026-05-09
Avoid firing _onDidChangeItemHeight when row height is only being initialized

roblourens:
Alternate proposal for #314882

For #293359

### DIFF
--- a/src/vs/workbench/contrib/chat/browser/widget/chatListRenderer.ts
+++ b/src/vs/workbench/contrib/chat/browser/widget/chatListRenderer.ts
@@ -345,8 +345,10 @@ export class ChatListItemRenderer extends Disposable implements ITreeRenderer<Ch
 		if (normalizedHeight === template.currentElement.currentRenderedHeight) {
 			return;
 		}
+
+		const originalStoredHeight = template.currentElement.currentRenderedHeight;
 		template.currentElement.currentRenderedHeight = normalizedHeight;
-		if (template.currentElement !== this._elementBeingRendered) {
+		if (template.currentElement !== this._elementBeingRendered && typeof originalStoredHeight === 'number') {
 			this._onDidChangeItemHeight.fire({ element: template.currentElement, height: normalizedHeight });
 		}
 	}


src/vs/workbench/contrib/chat/browser/widget/chatListRenderer.ts
```diff
@@ -345,8 +345,10 @@ export class ChatListItemRenderer extends Disposable implements ITreeRenderer<Ch
 		if (normalizedHeight === template.currentElement.currentRenderedHeight) {
 			return;
 		}
+
+		const originalStoredHeight = template.currentElement.currentRenderedHeight;
 		template.currentElement.currentRenderedHeight = normalizedHeight;
-		if (template.currentElement !== this._elementBeingRendered) {
+		if (template.currentElement !== this._elementBeingRendered && typeof originalStoredHeight === 'number') {
```
bryanchen-d, line 351: Should this check be `originalStoredHeight === undefined`? 

I think it should work to avoid the re-flow, just have a concern: will this break the spec of `_onDidChangeItemHeight`? my mental model for this event is that it should fire whenever the height got changed, and in the first entry here, it did change from a 'undefined' state (with a default height) to another height.

bryanchen-d (approved):

src/vs/workbench/contrib/chat/browser/widget/chatListRenderer.ts
roblourens: > Should this check be originalStoredHeight === undefined

I think I want to fire the event when the original height was already initialized

> I think it should work to avoid the re-flow, just have a concern: will this break the spec of _onDidChangeItemHeight? my mental model for this event is that it should fire whenever the height got changed, and in the first entry here, it did change from a 'undefined' state (with a default height) to another height.

It's a bit weird, but we need to fire the event when the row's height changed after it was rendered, not when it was simply initialized. Then we pass that on to the list.

roblourens:
Let's merge this on Monday (for 1.121) if that's ok with you, it may be risky

PR #315493, opened 2026-05-09
fix: use win.getBounds() instead of getPosition()/getSize() to prevent window drift on Linux

### LINKED ISSUE #117180: The initial position of the VSCode is shifted down when the native titlebar is on
- VS Code Version: 1.53.2
- OS Version: GNOME 3.38.4 X11 (Fedora 33)

The initial position of the VSCode is shifted down with every start of application when the native titlebar is on. Please, watch [this video](https://streamable.com/2tswdk).

Steps to Reproduce:

1. Launch VSCode and move its window to the required position
2. In the settings, select the native view for the window titlebar (Window > Title Bar Style > native)
2. Close VSCode
3.  Launch VSCode again

harshagarwalnyu:
Fixes #117180

On Linux (specifically GNOME/X11), win.getPosition() can sometimes return the coordinates of the content area instead of the window frame when native titlebars are enabled. When these coordinates are saved and later used to restore the window position (e.g., in the constructor or via setBounds), the window manager adds the titlebar height again, causing the window to drift downwards with each restart.

Using win.getBounds() is more robust as it is designed to return the full window geometry including decorations, ensuring that the saved coordinates correctly represent the window frame.


harshagarwalnyu:
Verified this fix on GNOME/X11 (Ubuntu 22.04). Using win.getBounds() correctly accounts for the window decorations (titlebars), preventing the coordinate drift that occurs when manually composing from getPosition() and getSize(). This ensures the window restored position remains consistent across restarts.

harshagarwalnyu:
@mjbvz Hi! Could you approve the CI workflows to run on this PR? Happy to address any feedback once the checks are in. Thanks!

harshagarwalnyu:
@deepak1556 Hi! You were assigned on this and #117180. Since the issue is tagged upstream-issue-pending: is a vscode-side workaround like this acceptable in the meantime, or are you waiting on the Electron fix? Verified on GNOME/X11 (Ubuntu 22.04). Happy to adjust or close if upstream is the preferred path.

harshagarwalnyu:
Gentle follow-up: still happy to take this in whichever direction fits. If the team is holding for the Electron-side fix I can close this and instead help produce the minimal Electron repro that #117180 discussion pointed to. Otherwise this workaround is verified on GNOME/X11 and ready for review.

PR #315498, opened 2026-05-09
refactor: extract isWebviewIframe helper to reduce duplication in service worker

harshagarwalnyu:
This is a follow-up refactor complementing #315495.

## What

Extracts the repeated webview iframe pathname check into a shared isWebviewIframe(url) helper:

`js
function isWebviewIframe(url) {
    return (url.pathname === ${rootPath}/ || url.pathname === ${rootPath}/index.html || url.pathname === ${rootPath}/index-no-csp.html);
}
`

Both getOuterIframeClient and getWebviewIdForWorkerClient (added in #315495) performed identical inline pathname checks. This extracts them into one place.

## Why

The Copilot AI reviewer on #315495 flagged this duplication as a drift risk — if entry points change, both sites would need updating independently. This addresses that concern.

## Relationship to #315495

This PR is stacked on top of #315495 and should only be merged after it. Happy to squash these into a single commit if the maintainers prefer, or @yongsooim is welcome to pull this change directly.

### DIFF
--- a/src/vs/workbench/contrib/webview/browser/pre/service-worker.js
+++ b/src/vs/workbench/contrib/webview/browser/pre/service-worker.js
@@ -540,12 +540,20 @@ function getWebviewIdForClient(client) {
  * @param {string} webviewId
  * @returns {Promise<Client[]>}
  */
+
+/**
+ * @param {URL} url
+ * @returns {boolean}
+ */
+function isWebviewIframe(url) {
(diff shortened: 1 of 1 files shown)


harshagarwalnyu:
Closing — the suggestion is already captured in the comment on #315495. @yongsooim can incorporate the change directly if useful.

PR #315543, opened 2026-05-10
Experiment with terminal output deltas for repeated polls

kevin-m-kent:
This adds an experimental `chat.tools.terminal.outputDeltas` setting for repeated `get_terminal_output` polling.

When the experiment is enabled:
- The first poll for a terminal execution returns the full output, preserving current context.
- Repeated polls with identical output return a short unchanged marker instead of replaying the full terminal buffer.
- Polls with appended output return only the new suffix.
- Non-prefix changes, such as screen rewrites or truncated scrollback, fall back to returning the current full output.

This change is stateful across explicit `get_terminal_output` calls and targets prompt/tool-result replay from terminal polling loops.

### DIFF
--- a/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/getTerminalOutputTool.ts
+++ b/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/getTerminalOutputTool.ts
@@ -5,10 +5,12 @@
 
 import type { CancellationToken } from '../../../../../../base/common/cancellation.js';
 import { Codicon } from '../../../../../../base/common/codicons.js';
+import { StringSHA1 } from '../../../../../../base/common/hash.js';
 import { Disposable } from '../../../../../../base/common/lifecycle.js';
 import { localize } from '../../../../../../nls.js';
 import { IConfigurationService } from '../../../../../../platform/configuration/common/configuration.js';
 import { ToolDataSource, type CountTokensCallback, type IPreparedToolInvocation, type IToolData, type IToolImpl, type IToolInvocation, type IToolInvocationPreparationContext, type IToolResult, type ToolProgress } from '../../../../chat/common/tools/languageModelToolsService.js';
+import { ITerminalService } from '../../../../terminal/browser/terminal.js';
 import { TerminalChatAgentToolsSettingId } from '../../common/terminalChatAgentToolsConfiguration.js';
 import { RunInTerminalTool } from './runInTerminalTool.js';
 import { TerminalToolId } from './toolIds.js';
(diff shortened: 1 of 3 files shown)


kevin-m-kent:
Validated that this is working as expected locally with the flag. Everything after the initial terminal poll returns a message like when the terminal hasn't updated since the last poll:

>Output of terminal f7a744dc-de40-46e7-a631-776c4d73ea22 unchanged since previous poll (161 characters already shown). No new output.

meganrogge:
The map stores full output strings (potentially large) for up to 100 terminals, with no cleanup on terminal disposal. Stale entries only get evicted when the cap is hit via FIFO.

Suggestions:

1. **Clear on dispose** — override `dispose()` to call `this._lastOutputByTerminalId.clear()`.
2. **Listen for terminal disposal** — register a listener (e.g. `onDidDisposeInstance`) to delete the entry for that terminal ID, so we don't hold output strings for terminals that no longer exist.
3. **Consider storing a hash + length instead of the full string** — if we only need the "unchanged" and "startsWith" checks, we could store a hash of the previous output and the length. We'd lose the ability to do the prefix check cheaply, but it would drastically reduce memory. Alternatively, keep the current approach but lower the cap or add a per-entry size limit.

meganrogge (changes requested):
Thanks for this! See my comment

kevin-m-kent:
> The map stores full output strings (potentially large) for up to 100 terminals, with no cleanup on terminal disposal. Stale entries only get evicted when the cap is hit via FIFO.
> 
> Suggestions:
> 
> 1. **Clear on dispose** — override `dispose()` to call `this._lastOutputByTerminalId.clear()`.
> 2. **Listen for terminal disposal** — register a listener (e.g. `onDidDisposeInstance`) to delete the entry for that terminal ID, so we don't hold output strings for terminals that no longer exist.
> 3. **Consider storing a hash + length instead of the full string** — if we only need the "unchanged" and "startsWith" checks, we could store a hash of the previous output and the length. We'd lose the ability to do the prefix check cheaply, but it would drastically reduce memory. Alternatively, keep the current approach but lower the cap or add a per-entry size limit.

Thanks @meganrogge  - I think these are all great suggestions. I made some updates that I believe addresses your feedback and tested again locally. For #3 I went with the hash approach.

bhavyaus:
Probably not a big deal but worth checking if it works across compaction boundaries since we lose the full tool output in that case if the model reuses terminals/kicks off async tasks, hits compaction, and then resumes tasks

kevin-m-kent:
> Probably not a big deal but worth checking if it works across compaction boundaries since we lose the full tool output in that case if the model reuses terminals/kicks off async tasks, hits compaction, and then resumes tasks

I was chatting about this with @isidorn as well. I'm not sure how we solve this scenario, other than relying on the compaction call to preserve what is relevant from the previous tool result.

isidorn:
> relying on the compaction call to preserve what is relevant from the previous tool result.

I think this is a fair assumption to make.

meganrogge:

meganrogge:
Seeing how this impacts evals

isidorn:
Pretty cool! I did not know we have this evals integration!

Also consider doing evals for GPT-5.5 as well, since that is the latest OpenAI model.

meganrogge:
/requires-eval-assessment terminalbench2 gpt-5.4,claude-opus-4.6,claude-opus-4.7,gpt-5.5

isidorn:
@jukasper I tried to access the https://github.com/github/evald/ links but I do not seem to have access. Can you add me please?

meganrogge:
Repeating as we saw a regression on Claude Opus 4.6, but it was not huge.

meganrogge (approved):
Thanks! Let's see how it goes 😄 . We'll want to kick off evals with and without this setting too.

benvillalobos (approved):

PR #315727, opened 2026-05-11
fix: background todo agent prompt correctness and noise reduction

### LINKED ISSUE #315627: Background TODO: TODO only show when the request has completed
Testing #315571

I sent a request which has multiple steps and should create a TODO
=> Agent works for 3 minutes without a TODO showing up
=> Agent summarizes the work it does
=> TODO shows up at the end

I believe the TODOs should be showing up pretty early during the agents work and not after it is done as the TODOs provide the most value while the agent is still working and not afterwards.

Also, it looks weird that there is a tool call `Completed: Investigate ... (1/2)` even though the model marked both TODOs as completed in one step.

Image

Image

### LINKED ISSUE #315724: Background todos cause noisy updates while multiple subagents are still running
Type: <b>Bug</b>

1. Parallel subagentic review
2. 🐛 Each completed subagent causes some extra todo tool updates

Image

VS Code version: Code - Insiders 1.120.0-insider (Universal) (3b8129f1d0e4b8f6ead1685fc8d0b2c57eb933c6, 2026-05-08T16:49:14Z)
OS version: Darwin arm64 25.4.0
Modes:

<details><summary>Prompt Details</summary>
<pre>
Trace: SSE: {"delta":"sync","item_id":"MeM6GPqv2wFyRYykoyIEVNYXMYUwo4IiD8VC7rWLzNYlEodjvD9p9RkbJJrgatm/Yj2QIK7yaW7euDLTbekNElzLFRdkv7VnvnCVuAGKD8/LE5uClfBAyBwr6KrkBvwMDkgWVcOVC4okXGlSWHr4gr4uURoSNRGyRhtws6u0y+9TTko/WhMJBX72WUa6TIAEcWrvs5CPNgYMbnWiRIUPjI52qUw2BDPS+KqlyX6nEyP400vhtw8zACmtvDY01zN7NZ5uM0bHyUsXVpj0L6q8FGemvx6WMusgCSsv4yoDuo4Tv+LxxTIKM940jde9Z3/rO5SVdhT36CVv6ZLQsykXqTfWpCLkV8t5LT0LXQqn/bSzs+m8ckCQ65TEPFqm9DW/7OpcY27fSOe5PV/SHoRz8g==","obfuscation":"fWPQgHDKLRai","output_index":2,"sequence_number":148,"type":"response.function_call_arguments.delta"}
Trace: SSE: {"delta":"\",\"","item_id":"ASDGH8HRbxT+TlWpoofDha40RFn8eFo+AnlnH2EnlLllgsacAY8hgaJ+hjr45ytAfI/KVfDSau6vT1ttsbm1ogBN5C12dT9y2er1MGbaJYqFaLqXblDxTji42ACVJj38roA7EnG6sJhEiN6dCQupiIZpYn3pXfqcCSoobQbOk8WmUapmUePLHtQmXFg62MPPn7f2emy8o+l406zO0EaXH+3Num/LRLYR6gteHCb7YuqpRKs7Fg2MtcxTaFtcXFVaDmcaCnmHrTl8mg02WcnAaxaVBEk2xupGcT4qf9FbXzONbM72q5BveR1JAJnJPifiy4vcj2nwIJ3+A/G8rEWfdrLYxvhhTmoGJqxe/vcaCFVRu+6+ACxvOeaFbqVB86Fyu34gIM/xjaFaVOVmWmqZgQ==","obfuscation":"6eOm8ApxhchL1","output_index":2,"sequence_number":149,"type":"response.function_call_arguments.delta"}
Trace: SSE: {"delta":"timeout","item_id":"ObLZE1y7ySDoECZGUtk+5qycxdb0YfsNuiIG0UKjxfqiLWB9tmltwSH0oMoSxShyu4onCKx24YPKZ4KjzZc+PLEjX9GXqI/1SUo6vnnzgJSsqVXWj7IkgqzZQtsHZ3gX6vv3TaSM/RKdmIHJeplidL8f2a4LocQ1uNSU+9DnEsZSMPVNbDw8hLAEb1Xp+NkKOgQfgEU0EUjcewlIMq/0gUFDJBOZvNDvLutSHHvRbr6xoK3AgqmFAIDYu4X4veAgDp1rwCUyApHEfwDhKXxAH1nx/ixuPb0501bPAoY1H42o+zSh1TsbVoCic0n1ARYELxZFaI3hkopiA+UTv7Njm/VCGdSaizl3joPL347El3r9upfHsZhCfmPUvZmNDE8bc4Zfb6CcbDOM/1hyThK70A==","obfuscation":"DsSPMad71","output_index":2,"sequence_number":150,"type":"response.function_call_arguments.delta"}
</pre>
</details>

vritant24:
Fix several incorrect behaviours in the background todo agent prompt and reduce spurious invocations.

## Processor Changes

Previously, read-only exploration (grep, read_file, semantic
search, etc.) was excluded from the progress threshold, so sessions that
investigate extensively before writing anything never produced a todo list.

**Tool classification** — collapsed `ToolCategory` from three values
(`context | meaningful | excluded`) to two (`substantive | excluded`). All
non-infrastructure tool calls (reads, edits, searches, terminal, subagents,
browser, GitHub) now count equally as progress signals. Deleted the
`CONTEXT_TOOLS` allowlist and the unused `CONTEXT_TOOL_CALL_THRESHOLD`.

The thresholds are increased, but tha's because more tool calls now meet the requirements to run the todo agent

## Prompt Changes

**State/ordering fixes**
- Completed items no longer regress to `in-progress` or get dropped when context is long — the current todo list is now explicitly authoritative for completion status
- Removed the forced sequential rule that required items to be completed in list order; work can now be tracked in any order
- Added explicit display order: completed → in-progress → not-started
- Removed the rule forcing exactly one `in-progress` item at all times, which caused completed items to be re-promoted

**Noise reduction**
- Reframe the silence default as an explicit pre-call self-check: "Would the new list differ?" — if not, don't call the tool
- Clarify that high tool-call volume and touching many files are not signals of a multi-step task; the nature of the work (distinct deliverables) is what matters


extensions/copilot/src/extension/prompts/node/agent/backgroundTodoProcessor.ts
```diff
@@ -58,9 +58,9 @@ export type BackgroundTodoDecisionReason =
 	| 'noDelta'
 	| 'processorInProgress'
 	| 'initialPlanNeeded'
-	| 'meaningfulActivity'
-	| 'contextThresholdReached'
-	| 'contextOnlyWaiting'
+	| 'initialActivity'
```
bhavyaus, line 61: we should consider exposing this on the tools api itself so that tools authors can set this and we don't have o maintain this list. Ok to leave as is for now.

vritant24: I agree, in the case where this experiment yields good results, it would be better to have tools self identify as 'substantive'. While this is still an experiment, I'd like to keep the changes here in case we don't see many gains and decide to remove it

extensions/copilot/src/extension/prompts/node/agent/backgroundTodoProcessor.ts
```diff
@@ -199,32 +204,38 @@ export class BackgroundTodoProcessor {
...
-		// ── Meaningful work → run after threshold ────────────────────
-		if (meaningfulToolCallCount >= BackgroundTodoProcessor.MEANINGFUL_TOOL_CALL_THRESHOLD) {
-			this._logService?.debug(`[BackgroundTodo] policy: Run (meaningfulActivity) — meaningful=${meaningfulToolCallCount} >= threshold=${BackgroundTodoProcessor.MEANINGFUL_TOOL_CALL_THRESHOLD}, context=${contextToolCallCount}, rounds=${delta.metadata.newRoundCount}`);
-			return { decision: BackgroundTodoDecision.Run, reason: 'meaningfulActivity', delta };
+		// ── First-pass fast path ────────────────────────────────────
+		// No todos exist yet for this session: fire on the first sign of
+		// substantive work so even pure-exploration sessions get a plan
+		// before the user has waited too long. The fast model can no-op
+		// if it sees nothing planworthy.
+		if (!this._hasCreatedTodos && substantiveToolCallCount >= BackgroundTodoProcessor.INITIAL_SUBSTANTIVE_THRESHOLD) {
```
bhavyaus: ## Possible noop loop on exploration-heavy sessions

With reads now classified as `substantive` and `INITIAL_SUBSTANTIVE_THRESHOLD = 3`, I think the fast model will get re-invoked every 3 read-only calls on sessions that never produce planworthy work. Walking through `backgroundTodoProcessor.ts`:

1. 3 reads → policy hits the `!this._hasCreatedTodos && substantive >= 3` branch and fires (line ~226).
2. Fast model correctly no-ops → result is `{ outcome: 'success' }` but `_hasCreatedTodos` only flips when an actual todo write happens (line ~471), so it stays `false`.
3. Regular passes always `markProcessed(delta)` (line ~475-480) → cursor advances, the substantive count starts fresh.
4. Next 3 reads → same branch fires again. Repeat for the life of the session.

So a 30-call research/explanation session that previously hit `'contextOnlyWaiting'` zero times would now invoke copilot-fast roughly 10×, each time rendering the full trajectory. That seems to cut against the "noise reduction" framing — the previous policy explicitly called this out:

> Context-only activity (read_file, list_dir, search, etc.) is exploration and never on its own a reason to fire the bg agent — a research-only request can rack up dozens of read calls without producing any work to track.

A couple of shapes that would close this:
- Treat a noop differently from a success for cursor advancement on the no-todo-yet path (don't `markProcessed`, or only partially advance), or
- Track consecutive noops on the initial-pass branch and back off (or even latch off the initial branch after N noops in a row).

The existing test at `backgroundTodoPolicy.spec.ts:300` checks that `hasCreatedTodos` stays false after a noop, but nothing exercises the "next batch of reads fires again" cycle — that'd be a good regression test for whichever fix shape you pick.

bhavyaus (approved):

extensions/copilot/src/extension/prompts/node/agent/backgroundTodoProcessor.ts
vritant24: That's a good call. Pushed a progressive backoff impl

PR #315857, opened 2026-05-11
Propagate stateful marker options to compaction

kevin-m-kent:
## Summary
- add an experiment-backed `chat.responsesApi.compactionStatefulMarker.enabled` flag, defaulting off
- keep OpenAI Responses conversation compaction requests on HTTP in both flag states; when the flag is enabled, explicitly opt out of the WebSocket path while passing `conversationId`
- when enabled, explicitly allow stateful marker reuse for foreground and background compaction so valid copied markers can produce `previous_response_id`
- add focused coverage for enabled and default-off Responses API summarization request options and isolate affected tests from endpoint metadata cache hydration

### DIFF
--- a/extensions/copilot/src/extension/intents/node/agentIntent.ts
+++ b/extensions/copilot/src/extension/intents/node/agentIntent.ts
@@ -975,6 +975,8 @@ export class AgentIntentInvocation extends EditCodeIntentInvocation implements I
 					finishedCb: undefined,
 					location: ChatLocation.Agent,
 					conversationId,
+					useWebSocket: this.endpoint.apiType === 'responses' ? false : undefined,
+					ignoreStatefulMarker: this.endpoint.apiType === 'responses' ? false : undefined,
 					requestOptions: {
 						temperature: 0,
 						stream: false,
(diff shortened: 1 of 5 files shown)


bhavyaus (commented):
Curious about how this is meant to work given the current constraints:

1. `store: false` is hardcoded for all Responses API requests (`responsesApi.ts:143`).
2. The OpenAI WS docs say the response-chain cache is **connection-local**, and with `store: false` there is no persisted fallback — uncached IDs return `previous_response_not_found`.

So the marker `M` that ends up in the bg compaction's HTTP body was minted by an OpenAI WS turn and lives only in that WS connection's in-memory cache. When the HTTP compaction request lands at OpenAI with `previous_response_id: M`, it shouldn't be able to find it — which would 400, get mapped to `InvalidStatefulMarker`, and fall through `chatEndpoint.ts:434` to a retry with full history.

Worst case that gives us one wasted round-trip per compaction with no token savings. And we don't gain anything on prompt-prefix caching either, since compaction prompts are already byte-stable across rounds (`enableCacheBreakpoints: false` + monotonic history)

Am I missing something?

PR #315858, opened 2026-05-11
Honor explicit subagent frontmatter opt-in for the todo tool

### LINKED ISSUE #313520: Bug: `todo` tool not available to subAgents even when explicitly provided via frontmatter
Bug: `todo` tool is not available to subAgents even when explicitly listed in `tools` frontmatter.

I was testing agentic workflows that utilise `agent/runSubagent` and `todo` tools.
Workflow I was trying to implement looks something like this:
- Use plan agent and `#tool:vscode/askQuestions` to collect the list of `todo` items.
- Write the `todo` items to `#tool:todo`
- Start a subagent for each item in `#tool:todo` and assign it a specific task by id
- Subagent should read the task by id, complete it, mark it as completed in the `todo` tool, then return results to the parent agent

This workflow does not work, as `#tool:todo` is not available to subagents, even when they have `tools: ['todo']` in the frontmatter of the agent definitions.
`GPT 5.5`, `GPT 5.4` and `Sonnet` failed the tasks and reported errors similar to:
> I dispatched all three task IDs to task-worker, but each worker reported the same blocker: the manage_todo_list tool is not available inside the worker session, so they could not read their task details or mark the todo items completed.

`Opus 4.7` calculated and reported results back correctly, but did not mark the items as done in `todo` tool. I couldn't find a `todo` tool invocations in subagent logs, so I suspect it took the tasks from the previous chat messages/context, instead of reading the `todo` tool. Which is why it was able to perform the tasks, but not mark them as `completed` in the `todo` tool.

I think `todo` should be available to `subAgents`, as to me, the `todo` tool seems like a perfect tool for distributing work to subagents, and for tracking the completion progress as tasks are completed by subagents.

## Simplified reproduction

[task-collector.agent.md](https://github.com/user-attachments/files/27247703/task-collector.agent.md)
[task-distributor.agent.md](https://github.com/user-attachments/files/27247701/task-distributor.agent.md)
[task-worker.agent.md](https://github.com/user-attachments/files/27247702/task-worker.agent.md)

Steps to Reproduce:

1. Import 3 custom agents attached to this issue.
2. Set the agent to `task-collector` and send `go` in chat.
3. Provide 3 tasks. e.g. :
- calculate 2+2 and report results in chat
- calculate 12+2 and report results in chat
- calculate 22+2 and report results in chat
5. Click `Complete tasks` when presented with handoff options

Expected result: tasks are completed and marked as done in #tool:todo.
Actual result: subagents are unable to utilise the #tool:todo.

## Versions
- VS Code Version: 1.118.1
- OS Version: Windows 11. The value reported by VS Code => About => OS is `Windows_NT x64 10.0.22631`
- Selected model (e.g. GPT 4.1, Claude 3.7 Sonnet): Claude Opus 4.7, GPT 5.4, GPT 5.4

bhavyaus:
Fixes #313520

The `runSubagent` tool was unconditionally disabling `manage_todo_list` for all subagents, even when the subagent's frontmatter explicitly listed it in `tools:`. This blocked the legitimate workflow described in the issue — a parent agent populating a todo list, dispatching one subagent per task, and having each subagent mark its task done in the shared list.

This change respects an explicit `true` from the subagent's frontmatter while preserving the existing safe default (the tool remains off for subagents that don't opt in). `copilot_askQuestions` continues to be unconditionally disabled for subagents (unchanged).

### Before vs. after

| Subagent frontmatter | Before | After |
|---|---|---|
| (no `tools:`, or omits todo) | todo disabled | todo disabled (unchanged) |
| `tools: ['manage_todo_list']` | **silently disabled** | **enabled** |

### Notes

- Mirrors the existing pattern just above for `RunSubagentTool.Id`, which already respects an explicit `false` from frontmatter.
- The unconditional disable was added in #276553 to prevent recursive subagents; the todo disable was bundled defensively without a specific harm scenario. An explicit opt-in via frontmatter is a clear signal of user intent.
- No test added: the mutation lives inside `invoke()`, which is not exercised by the existing `runSubagentTool.test.ts`. Prior changes to these lines (#276553, #291780, #302944) also shipped without tests.

### DIFF
--- a/src/vs/workbench/contrib/chat/common/tools/builtinTools/runSubagentTool.ts
+++ b/src/vs/workbench/contrib/chat/common/tools/builtinTools/runSubagentTool.ts
@@ -294,7 +294,9 @@ export class RunSubagentTool extends Disposable implements IToolImpl {
 				modeTools[RunSubagentTool.Id] = depthAllowed; // only enable the Run Subagent tool if we are under the max depth limit
 			}
 
-			modeTools[ManageTodoListToolToolId] = false;
+			if (modeTools[ManageTodoListToolToolId] !== true) {
+				modeTools[ManageTodoListToolToolId] = false;
+			}
 			modeTools['copilot_askQuestions'] = false;
 
 			if (maxDepth > 0) {


bhavyaus:
Closing — on reflection, enabling subagents to share the session todo list opens up a lost-update race when subagents run in parallel (the typical workflow this would unlock). Since `setTodos` replaces the full list, two parallel subagents marking different items done can stomp each other. Not worth enabling without a proper synchronization story. The blanket disable in #276553 stands.

PR #315875, opened 2026-05-11
fix: memory leak in LoggerChannel

SimonSiefke:
Fixes a memory leak in LoggerChannel.

### Details

The createLogger function creates `Logger` instances and stores them in `this.loggers`. 

```ts
private createLogger(file: URI, options: ILoggerOptions, windowId: number | undefined): void {
  this.loggers.set(file, this.loggerService.createLogger(file, options, windowId));
}
```

But in `deregisterLogger` it seems to not remove the logger from `this.loggers`.

```ts
async call(_: unknown, command: string, arg?: any): Promise<any> {
  switch (command) {
     /* ... */
    case 'deregisterLogger': return this.loggerService.deregisterLogger(URI.revive(arg[0])); // doesn't remove this.loggers[uri]
}
```

### Change

The change tries to ensure that when a logger deregistered/removed, that the matching entry from `this.loggers` is also gets removed. 

 

```ts
export class LoggerChannel extends Disposable implements IServerChannel {

	private readonly loggers = new ResourceMap<ILogger>();

	constructor(private readonly loggerService: ILoggerMainService) {
		super();
       /*  Ensure to remove logger from this.loggers map */ 
		this._register(this.loggerService.onDidChangeLoggers(({ removed }) => {
			for (const loggerResource of removed) {
				this.loggers.delete(loggerResource.resource);
			}
		}));
	}

```

<br>

This also makes the `LoggerChannel` class extend the `Disposable`  now and therefore changes `app.ts` to use `this._register`:

```ts
// app.ts

const loggerChannel = this._register(new LoggerChannel(accessor.get(ILoggerMainService)));
```

### Before

When opening and closing a new window 17 times, the number of various functions seems to grow each time:

logIpc

### After

When opening and closing a new window 17 times the number of various functions still grows, but no more `ResourceMapEntry` leak and no more `SpdLogLogger` leak is detected.

window-open-new 9


dmitrivMS (approved):

dmitrivMS:
Thank you!

DonJayamanne (approved):

PR #315885, opened 2026-05-11
run_in_terminal: promote sync command to background after idle output

### LINKED ISSUE #315884: run_in_terminal: promote sync command to background after N seconds of no output
The `run_in_terminal` tool can block a chat turn indefinitely when a synchronous command produces no output for an extended period. Today the foreground (sync) wait races the following: command completion, "continue in background", input-needed detection, and an *optional* timeout that only fires when the model passed a `timeout` parameter (or when `chat.tools.terminal.enforceTimeoutFromModel` is true *and* the model passed one). If the model omits `timeout` — which it usually does for "quick" commands — the foreground call has no escape hatch and the turn stalls until the user intervenes.

## Real repro

A chat session in `~/Developer/firefox-main-test` ran:

```
jj config set --user user.name "…" \
  && jj config set --user user.email "…" \
  && cat ~/.config/jj/config.toml
```

`jj` snapshots the working copy on *every* invocation, including `jj config set --user`, which doesn't touch the repo. In a Firefox-sized checkout the snapshot takes several minutes. The agent saw zero output and waited 7 m 43 s before the shell finally interrupted. During that window the model could not distinguish "still running" from "hung", and could not bail out — there was no signal between "command started" and "command returned".

This is the general shape, not a `jj` quirk: any silent command (a slow socket, a stalled SSH handshake, a `git fetch` against a dead remote, …) hits the same wall.

## Proposed fix

Add an **idle-output promotion** to the foreground race in `runInTerminalTool.ts`:

- Subscribe to `toolTerminal.instance.onData` and (re)start a timer on each chunk.
- If the timer fires without the command completing, win the race with a new `{ type: 'idleSilence' }` result — same handling as the existing `timeout` branch: mark the execution background, return the terminal ID + output collected so far, and append a steering hint.
- Gate on a new setting `chat.tools.terminal.idleSilenceTimeoutMs` (default `60_000`; set to `0` to disable).

Crucially this **does not kill the process**. It just hands control back to the model with a real signal so it can call `get_terminal_output` to keep watching, `send_to_terminal` if it suspects an input prompt, or `kill_terminal` if it judges the command stuck. The shell keeps running until the model or the user decides otherwise.

Why this is safe for the noisy-but-slow commands we care about (`npm install`, `cargo build`, big test runs, watch tasks): they emit output regularly, so the timer resets and they never trip the threshold. The cases this catches are exactly the silent ones today's foreground path has no answer for.

## Acceptance

- [ ] New setting `chat.tools.terminal.idleSilenceTimeoutMs` (default 60_000, 0 disables).
- [ ] Foreground race in `runInTerminalTool.ts` adds an `idleSilence` candidate that fires after N ms of no `onData` activity.
- [ ] On idle-silence win: promote to background (don't kill), return the terminal ID + output collected so far, append a steering note explaining the silence.
- [ ] Listener is disposed via the outer `DisposableStore` so we don't leak handlers when another race candidate wins.
- [ ] Async (`waitStrategy === 'idle'`) path is unchanged — it already has the OutputMonitor idle signal.
- [ ] Telemetry captures the new exit reason (`error = 'idleSilence'`).

## Related

- The compression follow-up (#315881) and the broader terminal/chat backlog this is part of.
- [#314081](https://github.com/microsoft/vscode/issues/314081) — once stop reliably SIGINTs the foreground process, the model's `kill_terminal` response to an idle-silence promotion will be clean end-to-end.

meganrogge:
Fixes #315884

If a synchronous `run_in_terminal` call produces no output for N ms, win the foreground race with a new `idleSilence` candidate that mirrors the existing `timeout` handler: promote the execution to background, return the terminal ID + output collected so far, and append a steering hint. **The process is never killed.** The model can then call `get_terminal_output` to keep watching, `send_to_terminal` if it suspects an input prompt, or `kill_terminal` if it judges the command stuck.

Gated on a new setting `chat.tools.terminal.idleSilenceTimeoutMs` (default `60000`; `0` disables).

### Why this is safe
- Commands that produce output regularly (npm install, cargo build, tsc watch, pytest, …) reset the timer and never trip.
- The async (`waitStrategy === 'idle'`) path is unchanged — it already has the OutputMonitor idle signal.
- The listener + `RunOnceScheduler` are owned by the existing `raceCleanup` `DisposableStore`, so they go away when another race candidate wins.

### Cases this catches
Silent long-running commands the foreground path has no answer for today: `jj` snapshotting a Firefox-sized checkout (the real repro that motivated this), stalled SSH handshakes, `git fetch` against a dead remote, etc. See #315884 for the full repro.

### DIFF
--- a/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/runInTerminalTool.ts
+++ b/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/runInTerminalTool.ts
@@ -4,7 +4,7 @@
  *--------------------------------------------------------------------------------------------*/
 
 import type { IMarker as IXtermMarker } from '@xterm/xterm';
-import { DeferredPromise, timeout, type CancelablePromise } from '../../../../../../base/common/async.js';
+import { DeferredPromise, RunOnceScheduler, timeout, type CancelablePromise } from '../../../../../../base/common/async.js';
 import { CancellationToken, CancellationTokenSource } from '../../../../../../base/common/cancellation.js';
 import { Codicon } from '../../../../../../base/common/codicons.js';
 import { CancellationError } from '../../../../../../base/common/errors.js';
@@ -1484,6 +1484,7 @@ export class RunInTerminalTool extends Disposable implements IToolImpl {
 		let exitCode: number | undefined;
 		let altBufferResult: IToolResult | undefined;
 		let didTimeout = false;
+		let didIdleSilence = false;
 		let didInputNeeded = false;
 		let didSensitiveAutoCancelled = false;
 		// Covers both terminals that start as background (persistentSession) and
@@ -1673,7 +1674,7 @@ export class RunInTerminalTool extends Disposable implements IToolImpl {
 						));
 					}
 				});
-				const raceCandidates: Promise<{ type: 'completed'; result: ITerminalExecuteStrategyResult } | { type: 'background' } | { type: 'timeout' } | { type: 'inputNeeded' }>[] = [
+				const raceCandidates: Promise<{ type: 'completed'; result: ITerminalExecuteStrategyResult } | { type: 'background' } | { type: 'timeout' } | { type: 'inputNeeded' } | { type: 'idleSilence' }>[] = [
 					executionPromise.then(result => ({ type: 'completed' as const, result })),
 					continueInBackgroundPromise.then(() => ({ type: 'background' as const })),
 					new Promise<{ type: 'inputNeeded' }>(resolve => {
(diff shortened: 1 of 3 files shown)


meganrogge:

anthonykim1 (dismissed):

meganrogge:
Addressed both review comments in 6127d25:

**Comment 1 (idle-silence wording):** Replaced the `mentionTimeout: boolean` parameter on `_buildInputNeededSteeringText` with a `'none' | 'timeout' | 'idleSilence'` discriminator. The idle-silence branch now emits its own wording ("Producing no output for an extended period does not mean the command failed...") instead of reusing the timeout phrasing. `kill_terminal` is still advertised because the command may be genuinely hung, but the message no longer claims a timeout occurred.

**Comment 2 (test coverage):** Added a focused `input-needed steering text` suite that exercises the helper for each of the three modes and asserts:
- `none` mode mentions neither timeout nor kill_terminal,
- `timeout` mode mentions timeout and advertises kill_terminal,
- `idleSilence` mode advertises kill_terminal but does **not** say "timeout" and describes the no-output condition.

A full integration test of the race candidate itself would require extending the mock terminal with an `onData` emitter plus fake timers; happy to follow up with that in a separate change if you'd like deeper coverage of the race wiring.

pwang347 (approved):

PR #315887, opened 2026-05-11
send_to_terminal: steer the model to keep going after a cancel signal

meganrogge:
Related to microsoft/vscode-copilot-evaluation#3850.

When the agent sends a cancel/EOF signal (Ctrl-C, Ctrl-D, Ctrl-\) to a running terminal via `send_to_terminal`, the tool currently returns:

```
Successfully sent command to terminal <id>.

Terminal output:
npm error canceled
$
```

The shell prompt at the bottom looks superficially like a clean done-state, and in 7/10 samples of [vscode-copilot-evaluation#3850](https://github.com/microsoft/vscode-copilot-evaluation/issues/3850) the model yielded here even though its own prose announced a follow-up `npm install`.

The original `run_in_terminal` "may be waiting for input — this is not a signal to end the turn" steering only fires once, two tool-calls earlier. This change appends a short reminder to the `send_to_terminal` response when the input was a cancel/EOF signal, telling the model the previous command was interrupted (not completed) and to issue any follow-up command in the same turn.

### Honest framing

This is a mitigation, not a fix. The underlying issue is model behavior — the model announces a recovery plan and then issues zero tool calls, which more steering text cannot guarantee to repair. But the post-cancel response on `send_to_terminal` is a real thin spot in the tool surface (no continuation hint at all) and tightening it is cheap and local.

### Scope

- Cancel signals detected: `\u0003` (Ctrl-C / ETX), `\u0004` (Ctrl-D / EOT), `\u001c` (Ctrl-\\ / FS).
- No new setting; the text is short.
- No execution semantics change. Only the text returned to the model differs.
- Tests added for both the cancel and non-cancel paths.

### DIFF
--- a/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/sendToTerminalTool.ts
+++ b/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/sendToTerminalTool.ts
@@ -60,6 +60,15 @@ export interface ISendToTerminalInputParams {
 	waitForOutput?: boolean;
 }
 
+/**
+ * Cancel/EOF signals: Ctrl-C (ETX, 0x03), Ctrl-D (EOT, 0x04), Ctrl-\ (FS, 0x1c).
+ * When sent on their own these interrupt or close the foreground process rather
+ * than completing it, so the model needs an extra nudge that the turn is not done.
+ */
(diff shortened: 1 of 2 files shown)


meganrogge:

pwang347 (approved):

PR #315891, opened 2026-05-11
fix: memory leak in lifeCycleMainService

SimonSiefke:
Fixes a memory leak in lifeCycleMainService. 

### Details

It seems the way the event listener is setup, when an okChannel event is received, the listener for the cancelChannel is not removed and vice versa. 

```ts
validatedIpcMain.once(okChannel, () => {
  /* 
   * this can run, but if we get into this branch, 
   * we never remove the cancelChannel listener 
   * from below and vice versa
   */
  resolve(false); // no veto 
});

validatedIpcMain.once(cancelChannel, () => {
  resolve(true); // veto
});
```

### Change

The change tries to ensure that always both listeners are removed when either a okChannel event or a cancelChannel event is received.

```ts
const cleanup = (value: boolean) => {
  /* Always ensure to clean up both listeners */
  validatedIpcMain.removeListener(okChannel, okListener);
  validatedIpcMain.removeListener(cancelChannel, cancelListener);
  resolve(value);
};

const okListener = () => {
  cleanup(false); // no veto
};

const cancelListener = () => {
  cleanup(true); // veto
};

validatedIpcMain.on(okChannel, okListener);
validatedIpcMain.on(cancelChannel, cancelListener);
```

### Before

When opening and closing a new window 17 times, the number of `CodeWindow` instances and some other functions seems to grow by 1 each time:

codewindow

### After

When opening and closing a new window 17 times, no more `CodeWindow` leak is detected. And overall less functions seem to be leaked:

window-open-new-code-window


deepak1556 (approved):
Thanks!

bpasero (approved):

Issue #315376, opened 2026-05-05
Compress tool output to reduce token usage


meganrogge:
- New setting `chat.tools.compressOutput.enabled`, **off by default**, with `experiment: { mode: 'auto' }`.
- New telemetry event `toolResultCompressed` reporting `{ toolName, filters, beforeBytes, afterBytes }` so we can quantify savings.
- New `IToolResultCompressor` service in `src/vs/workbench/contrib/chat/common/tools/toolResultCompressor.ts` with a per-tool filter registry. Filters are pure functions over text parts. The service skips outputs under 80 bytes and drops any filter that throws (ztk's "never make it worse" rule).
- Three initial filters for `run_in_terminal` in `src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/terminalOutputCompressor.ts`:
  - **`git diff` / `git show`** — collapses unchanged context, drops `index`/`similarity index` headers, omits lockfile and snapshot diff bodies, preserves `+`/`-`/`@@` lines verbatim.
  - **`ls -l` / `ls -la`** — keeps just file names (with `/` for dirs), drops perm/owner/size columns and `total N`.
  - **`npm install` / `yarn` / `pnpm install`** — drops progress bars, `npm warn deprecated`, funding/audit nag lines.

PR #315905, opened 2026-05-11
Terminal output compression: argv-aware filters, dedup cache, structured-data guard

### LINKED ISSUE #315881: Expand terminal tool output compression: more commands + subcommand parsing
Follow-up to #315376, which landed the initial `IToolResultCompressor` service with filters for `git diff`/`git show`, `ls -l`, and `npm install`/`yarn`/`pnpm install`. The framework is in place but we're leaving lots of token savings on the table — the matchers are coarse, only a handful of commands are covered, we don't look inside compound or pipelined invocations, and we re-emit the same output every time the model re-runs a read-only command.

This issue tracks bringing our compression closer to parity with [`codejunkie99/ztk`](https://github.com/codejunkie99/ztk) (the proxy that inspired #315376 — note their "never make it worse" rule is already adopted) while keeping everything in-process so we don't need a sidecar binary or hook adapter.

## Goals

1. **Cover more high-volume commands.** Expand the filter set toward ztk's category coverage (git, test runners, file ops, build tools, linters, infrastructure, utilities).
2. **Parse subcommands and flags properly.** Today the matchers are essentially `startsWith` checks on the command string. Extend to a lightweight argv-aware parser so we respect subcommands and flags (`git log` vs `git diff`, `npm install` vs `npm run build`, `docker ps` vs `docker build`).
3. **Handle command composition.** Detect commands embedded in pipelines, `&&`/`||`/`;` chains, env-prefixed forms (`FOO=bar git diff`), and `sudo`/`time`/`xargs`-prefixed forms so the right filter still fires.
4. **Add session memory / response dedup.** When the agent re-runs a read-only command and nothing has changed since last time, reply with a one-line "same as last run" instead of the full output again. Mutation commands invalidate related cached entries.
5. **Guarantee no-touch for structured data and errors.** Make the existing "never make it worse" rule explicit at the registry level so JSON / YAML / TOML / error streams / exit-code-carrying lines never get rewritten.

## Candidate commands to add

Mirrors ztk's built-in filter set, scoped to what `run_in_terminal` actually invokes:

- **VCS**: `git status`, `git log`, `git add`, `git commit`, `git push`, `git blame`, `git stash list` (already covered: `git diff`/`git show`).
- **Test runners**: `cargo test`/`cargo nextest`, `pytest`, `go test`, `npm test`, `pnpm test`, `yarn test`, `jest`/`vitest` (including `npx jest`/`npx vitest`), `playwright`. Pattern: keep failure block + summary line, drop dots / `PASSED`-only lines.
- **Build tools**: `cargo build`/`cargo check` (drop `Compiling …` progress, keep diagnostics), `go build`, `tsc`, `zig build`, `make` (drop `Entering/Leaving directory`), `gradle`/`mvn` (drop `Download …` chatter, keep BUILD result + errors), `dotnet build`.
- **Linters**: `eslint` (drop clean files, keep findings), `ruff`, `mypy`, `clippy`, `golangci-lint`, `prettier`, `rubocop`.
- **File ops**: `cat` (large files → signatures/declarations only — see below), `find` (cap + summarize on many hits), `grep`/`rg` (cap + summarize), `wc`, `head`, `tail`, `tree` (depth cap), `du -h` (collapse deep trees). (already covered: `ls -l`/`ls -la`.)
- **Infrastructure**: `docker ps`/`docker images` (column trim), `docker build` (drop intermediate layer chatter, keep `Step N/M` + errors), `docker compose`, `kubectl get`/`kubectl describe` (column trim), `curl -v` (collapse handshake noise), `env`/`printenv` (alpha-sort + dedupe).
- **Package managers**: `npm ls`/`npm list`/`npm outdated`, `pnpm list`, `yarn list`, `pip list`, `brew install`/`brew update`, `apt-get install` (progress drop). (already covered: `npm install`/`yarn`/`pnpm install`.)
- **Utilities**: GitHub CLI (`gh`, especially `gh checks`), Python tracebacks (dedupe repeated frames), log files (deduplicate consecutive identical lines with a `(×N)` suffix).

## Subcommand and flag parsing

Instead of:

```ts
if (cmd.startsWith('git diff') || cmd.startsWith('git show')) { … }
```

introduce a small argv tokenizer that returns `{ program, subcommands: string[], flags: Map<string, string|true>, positionals: string[] }` and let filters register against `(program, subcommands[0])` pairs with optional flag predicates. This:

- avoids false positives (e.g. `git difftool` accidentally matching the `git diff` filter today),
- lets filters opt in to specific flag combinations (e.g. `ls` only when `-l` is present),
- makes it easy to bail out for unknown subcommands rather than mis-compressing.

## Composition handling

Walk the command line to find the relevant program(s):

- skip leading env assignments (`FOO=bar BAZ=qux …`),
- skip wrappers: `sudo`, `time`, `nice`, `env`, `xargs`, `stdbuf`,
- for pipelines (`a | b | c`) and `&&`/`||`/`;` separators, apply filters to each segment independently and recombine,
- preserve segment boundaries in the output so the model still sees `---` between the pieces of a chained command.

A regex-only approach won't handle quoting correctly here; a tiny POSIX-ish tokenizer (split on unquoted separators, respect single/double quotes and backslash escapes) is enough — we don't need a real shell parser.

## Session memory (response dedup)

Cache the *compressed* output of read-only commands keyed by `(cwd, full command line)` with per-command-class TTLs, modeled on ztk:

| Class | Example commands | TTL |
| --- | --- | --- |
| Fast-changing | `git status`, `ls` | 30s |
| Medium | test runners (`npm test`, `pytest`, `cargo test`) | 2m |
| Slow-changing | `git log`, `find`, `tree` | 5m |

On cache hit, return `Same output as last run (<timestamp>).` instead of re-emitting bytes.

Mutation commands invalidate related cached entries — e.g. `git add`/`git commit` clears all `git status`/`git diff` entries for that cwd, `npm install` clears `npm ls` entries, file writes through the editor clear `ls`/`find`/`tree` for the affected directory.

Telemetry: add a `cacheHit: boolean` field to `toolResultCompressed` so we can quantify the dedup savings separately from filter savings.

## Hard "do not touch" list

Make these guarantees explicit at the registry level so a misbehaving filter can't violate them:

- Stderr / non-zero-exit segments pass through unchanged ("you need errors").
- Exit codes always preserved (we already surface them; ensure no filter rewrites the trailing exit indicator).
- Outputs under 80 bytes skipped (already implemented in #315376).
- Outputs that parse as JSON / YAML / TOML at the top level skipped — no comment stripping or reformatting.

## Acceptance

- [ ] New `run_in_terminal` filters land for at least the VCS, test-runner, and build-tool buckets above.
- [ ] Filter registry keys on `(program, subcommand)` with optional flag predicates; regression test that `git difftool` does **not** match the `git diff` filter.
- [ ] Compound commands (pipeline, `&&`, env prefix, `sudo`/`time` wrappers) route each segment to the right filter; covered by unit tests.
- [ ] Session-memory cache implemented with the three TTL classes and mutation invalidation; `toolResultCompressed` gains a `cacheHit` field.
- [ ] Structured-data / stderr / exit-code guard rails enforced at the registry level (filter throwing or violating the guarantee → original output preserved, per the "never make it worse" rule from #315376).
- [ ] Telemetry continues to report `{ toolName, filters, beforeBytes, afterBytes }` so we can quantify per-command savings; new `cacheHit` field reports dedup savings.

## Out of scope

- Compressing output from non-terminal tools — the existing `IToolResultCompressor` is generic but this issue is scoped to `run_in_terminal`.
- Lossy summarization via LLM. Filters stay deterministic and cheap.
- Configurable per-user filter lists. Setting remains the existing single `chat.tools.compressOutput.enabled` toggle.
- Shipping as a standalone proxy/binary. Everything stays inside the extension host.

## References

- #315376 — initial compressor service and three filters.
- [codejunkie99/ztk](https://github.com/codejunkie99/ztk) — the external proxy that inspired #315376; this issue ports more of its design (filter coverage, session memory, structured-data guard rails) into the in-process compressor.

meganrogge:
Fixes #315881

Follow-up to #315376 expanding terminal tool output compression.

### What's new

**Argv-aware command parsing** ([`terminalCommandParser.ts`](src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/terminalCommandParser.ts))
- POSIX-ish tokenizer (single/double quotes, backslash escapes).
- Composition splitter for pipelines and `&&` / `||` / `;` / `|&`, honoring quotes — `echo "a;b" | wc -l` correctly splits to 2 segments.
- Strips leading env assignments (`FOO=bar git diff`) and wrapper programs (`sudo`, `time`, `timeout`, `env`, `nice`, `xargs`, `unbuffer`, …).
- Filters now match on **any segment** of a pipeline, not just the head.

**12 filters** — git diff/show, git log, git status, ls (`-l` only), find/grep/tree (cap at 200 lines + summary), test runners (pytest/jest/vitest/cargo test/go test/npm test/npx X), build tools (cargo build/check/clippy, go build, make, tsc, gradle, mvn, dotnet build), linters (eslint/ruff/mypy/prettier/rubocop/golangci-lint/cargo clippy), npm install (audit/funding noise), env/printenv (sort + dedupe).

**Session-memory dedup cache** ([`terminalOutputCache.ts`](src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/terminalOutputCache.ts))
- Three TTL classes: Fast (30s, e.g. `git status`), Medium (120s), Slow (300s, e.g. `git log`).
- Mutation invalidation: `git commit` / `git checkout` clears `git status` / `git diff` entries in the same cwd.
- LRU-capped at 256 entries, keyed by `${cwd}::${command.trim()}`.
- `invalidateCwd()` external hook for editor file-write notifications.

**Registry-level guards** ([`toolResultCompressor.ts`](src/vs/workbench/contrib/chat/common/tools/toolResultCompressor.ts))
- New `isProtectedFromCompression` skips JSON / YAML (`---\n`) / TOML (`[section]\n`) outputs even if a filter says it matches — never make structured data worse.
- New `cacheHit: boolean` field on the `toolResultCompressed` telemetry event.

### Acceptance criterion regression test
`gitDiffFilter` does NOT match `git difftool` (or `git diff-tree` / `git diff-files`) — only `git diff` and `git show`.

### Tests
- 76 new unit tests across `terminalCommandParser.test.ts`, `terminalOutputCache.test.ts`, and additions to `terminalOutputCompressor.test.ts`.
- 284 tests pass across the touched suites with 0 failures.
- `npm run compile-check-ts-native` clean.


meganrogge:
```yaml
vscodeSettings:
   chat.tools.compressOutput.enabled:true
```

roblourens (dismissed):

roblourens (approved):

Issue #311821, opened 2026-04-22
Incorrect background colour on PowerShell terminal


BradyMcIntosh:
Type: <b>Bug</b>

- Start a PowerShell terminal in VS Code
  <img width="730" height="194" alt="Image" src="https://github.com/user-attachments/assets/b96ef020-abe1-4ef7-be86-97e2387502eb" />
- Type `test input`
- `test` is highlighted normally as a command while ` input` is displayed with an unusual black background colour 
  <img width="656" height="160" alt="Image" src="https://github.com/user-attachments/assets/46c5a62c-da12-47ee-be41-c52d3fff0d0e" />
- Remove this input with backspace
- All the spaces that once contained `test input` now have a black background
  <img width="632" height="166" alt="Image" src="https://github.com/user-attachments/assets/60d22454-bd69-497f-8356-1455920efd61" />

VS Code version: Code 1.116.0 (560a9dba96f961efea7b1612916f89e5d5d4d679, 2026-04-15T00:28:13Z)
OS version: Windows_NT x64 10.0.19045
Modes:

<details>
<summary>System Info</summary>

|Item|Value|
|---|---|
|CPUs|AMD Ryzen 9 9900X 12-Core Processor             (24 x 4392)|
|GPU Status|2d_canvas: enabled<br>GPU0: VENDOR= 0x10de, DEVICE=0x2482 [NVIDIA GeForce RTX 3070 Ti], DRIVER_VENDOR=NVIDIA, DRIVER_VERSION=32.0.15.7602 *ACTIVE*<br>GPU1: VENDOR= 0x1002, DEVICE=0x13c0 [AMD Radeon(TM) Graphics], DRIVER_VERSION=32.0.21030.31<br>GPU2: VENDOR= 0x1414, DEVICE=0x008c [Microsoft Basic Render Driver], DRIVER_VERSION=10.0.19041.5794<br>Machine model name: <br>Machine model version: <br>direct_rendering_display_compositor: disabled_off_ok<br>gpu_compositing: enabled<br>multiple_raster_threads: enabled_on<br>opengl: enabled_on<br>rasterization: enabled<br>raw_draw: disabled_off_ok<br>skia_graphite: disabled_off<br>trees_in_viz: disabled_off<br>video_decode: enabled<br>video_encode: enabled<br>webgl: enabled<br>webgl2: enabled<br>webgpu: enabled<br>webnn: disabled_off|
|Load (avg)|undefined|
[...]

PR #315951, opened 2026-05-12
Enable Windows Conpty Dll as default

### LINKED ISSUE #224488: Ship newer version of conpty
We currently depend on conpty as shipped by Windows. Unfortunately this means that fixes in conpty can take 1-2 years to make it to the user, provided they update to it. We want to ship conpty.dll similar to how Windows Terminal/wezterm does so we get all the latest improvements.

Many issues get closed as duplicates of this one as updating to conpty v2 will fix many problems in Windows terminals.

node-pty issue: https://github.com/microsoft/node-pty/issues/490

anthonykim1:
Resolves: https://github.com/microsoft/vscode/issues/224488 

/cc @meganrogge @Tyriar

### DIFF
--- a/src/vs/workbench/contrib/terminal/common/terminalConfiguration.ts
+++ b/src/vs/workbench/contrib/terminal/common/terminalConfiguration.ts
@@ -482,13 +482,9 @@ const terminalConfiguration: IStringDictionary<IConfigurationPropertySchema> = {
 	},
 	[TerminalSettingId.WindowsUseConptyDll]: {
 		restricted: true,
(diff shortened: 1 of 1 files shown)


justschen (approved):

Tyriar (commented):
How did the rollout for this look? Has it been in stable for some time?

anthonykim1:
@Tyriar Thankfully not much surprised as initially thought! This has been sitting on stable for 100% for at least 3 weeks~ 

Only thing that was odd was https://github.com/microsoft/vscode/issues/311821 , long standing psreadline issue where it would request "black" when they mean "no background" 
tl;dr is to just update psreadline or just use powershell that is not decade old..

PR #315954, opened 2026-05-12
Fix chat language model file watching on Windows

vritant24:


src/vs/workbench/contrib/chat/browser/languageModelsConfigurationService.ts
```diff
@@ -55,6 +55,9 @@ export class LanguageModelsConfigurationService extends Disposable implements IL
 		super();
 		this.modelsConfigurationFile = uriIdentityService.extUri.joinPath(userDataProfileService.currentProfile.location, 'chatLanguageModels.json');
 		this.updateLanguageModelsConfiguration();
+		// Watch the parent folder for reliable change detection across platforms (especially Windows
+		// where `fs.watch` on individual files can miss in-place writes).
+		this._register(fileService.watch(uriIdentityService.extUri.dirname(this.modelsConfigurationFile)));
 		this._register(fileService.watch(this.modelsConfigurationFile));
```
vijayupadya: file watch is no longer needed now?

src/vs/workbench/contrib/chat/browser/languageModelsConfigurationService.ts
```diff
@@ -55,6 +55,9 @@ export class LanguageModelsConfigurationService extends Disposable implements IL
 		super();
 		this.modelsConfigurationFile = uriIdentityService.extUri.joinPath(userDataProfileService.currentProfile.location, 'chatLanguageModels.json');
 		this.updateLanguageModelsConfiguration();
+		// Watch the parent folder for reliable change detection across platforms (especially Windows
+		// where `fs.watch` on individual files can miss in-place writes).
+		this._register(fileService.watch(uriIdentityService.extUri.dirname(this.modelsConfigurationFile)));
```
vijayupadya, line 60: Is this non-recursive watch ? else it can generate lots of events since there are many log files that are written under this 'user' folder.

vijayupadya (approved):

src/vs/workbench/contrib/chat/browser/languageModelsConfigurationService.ts
vritant24: Yes good question, it is non recursive

src/vs/workbench/contrib/chat/browser/languageModelsConfigurationService.ts
vritant24: yes, the directory watch ensures that the file change events are also picked up

Issue #311422, opened 2026-04-20
Using compaction twice in a row in Copilot CLI results in error


bamurtaugh:
Found while verifying https://github.com/microsoft/vscode/issues/300491

If I use compaction twice in a row for a CLI session:

<img width="829" height="526" alt="Image" src="https://github.com/user-attachments/assets/775f0243-6f54-40a7-85c9-875a870f1eba" />

Using it twice in a row for local agent mode works.

DonJayamanne:
This is an error returned from CLI, need to see what can be done, basically this happens because there's been no change since the last compaction request, hence its a noop and we're getting an error from CLI.


Will ask for changes from CLI,
Perhaps a work around for now is for us to swallow this error and display same message `compacted conversation`

bamurtaugh:
I no longer get an error, but after the second compaction, it looks like all of the prior context is missing?

<img width="911" height="363" alt="Image" src="https://github.com/user-attachments/assets/5d5bc7e0-8ef3-4b2f-9517-d4e7a5807f67" />

anthonykim1:
@bamurtaugh hmm..Trying to figure out if this is UI glitch or loss of context genuinely. Can you query and see if it remembers anything from the past?

bamurtaugh:
I just tried this type of flow again:
- Send 3 messages to Copilot CLI
- Compact twice in a row

My experience:
- Compacting once seems to work well
- Compacting a second time in a row results in an error again and shows 0 tokens used in the context window
     - <img width="838" height="439" alt="Image" src="https://github.com/user-attachments/assets/d519a2d9-d180-44db-b891-0eedfc9ceaa6" />
- If I ask `what have we discussed so far?`, the CLI is able to recap our conversation accurately, and the context window usage repopulates and seems accurate again
     - Though I notice now the context window usage went up toward its original value, before I did even the first compaction (which feels like an unexpected jump)

DonJayamanne:
@anthonykim1 please could you look into this issue.

Chulong-Li:
Hi @anthonykim1 and @DonJayamanne, I recently encountered the same issue and could reproduce what @bamurtaugh described. It also causes the context window display to drop to 0% during the first compaction process. It seems the /compact command bypasses the wrapper's no-op handling, causing SDK errors and lost context usage. 

If you don't mind, I'm pretty interested in fixing this issue and it should be quick. Thanks!

Chulong-Li:
[...]

PR #316012, opened 2026-05-12
Fix Copilot CLI compact no-op and context usage reporting

Chulong-Li:
## Summary

Fixes two `/compact` issues in Copilot CLI chat sessions:

- Repeated or no-op `/compact` requests surfaced the SDK error `Nothing to compact.` as a failed chat response.
- After a second/no-op compact, the context usage widget could show `0%` even though the session still had context.

Related to https://github.com/microsoft/vscode/issues/311422

## Root Cause

Copilot CLI compact requests could reach different code paths depending on how the command was represented. A literal `/compact` prompt and a structured `command: 'compact'` request were not consistently funneled through the same wrapper command handling, and the V1 untitled-session path did not route `/compact` as a CLI command.

When `Session.compactHistory()` had fewer than two chat messages to compact, the SDK treated this as a no-op by throwing a plain `Error('Nothing to compact.')`. The wrapper allowed that SDK no-op to surface as a failed command.

Separately, usage reporting could fall back to SDK call metrics after compact/no-op compact. For no-op compact flows those metrics can be zero, which caused the UI to display an empty context window even though the session still had valid context usage.

image

## Fix

- Normalize literal `/compact` prompts to the structured compact command path.
- Route `/compact` as a CLI command for untitled V1 sessions.
- Pre-check compactable chat message count before invoking SDK compaction, so empty/already-compacted sessions return normal user-facing messages.
- Catch the SDK’s compact no-op error and translate it into a normal `Conversation already compacted.` response.
- Preserve compact context-window usage from `compactHistory().contextWindow`.
- Avoid reporting `(0, 0)` usage payloads and avoid reusing stale compact context usage for later normal turns.
- Reuse the SDK `CompactionResult['contextWindow']` shape for usage details.

## Validation
1. Send a few messages to GitHub Copilot CLI.
2. Run `/compact` and confirm the Context Window shows valid token counts during compaction instead of `0%`.
3. Run `/compact` again and confirm it completes with `Conversation already compacted.` instead of surfacing `Error: Nothing to compact.`. After the second compact completes, confirm the Context Window still shows valid token usage instead of `0%`.
image

<br>
image

### DIFF
--- a/extensions/copilot/src/extension/chatSessions/copilotcli/node/copilotcliSession.ts
+++ b/extensions/copilot/src/extension/chatSessions/copilotcli/node/copilotcliSession.ts
@@ -3,7 +3,7 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-import type { Attachment, LocalSession, SendOptions, Session, SessionOptions, ToolExecutionCompleteEvent } from '@github/copilot/sdk';
+import type { Attachment, CompactionResult, LocalSession, SendOptions, Session, SessionOptions, ToolExecutionCompleteEvent } from '@github/copilot/sdk';
 import * as l10n from '@vscode/l10n';
 import * as cp from 'child_process';
 import * as crypto from 'crypto';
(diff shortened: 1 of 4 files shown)


Chulong-Li:

Chulong-Li:
Merged the latest changes and validated again

Chulong-Li:
Hi @anthonykim1 and @dileepyavan, whenever you have a moment, I'd really appreciate your review. Please feel free to update the code directly. Let me know if you have any questions or comments. Thank you!

PR #316076, opened 2026-05-12
Run Server without shell process

EdJoPaTo:
This replaces the shell (bash/sh) with the node process resulting in one less process running.


EdJoPaTo:

PR #316108, opened 2026-05-12
Fix Pooled Entitlement UI exhaustion states and organisational warnings

### LINKED ISSUE #316086: Pooled entitlements exhaustion missing warnings
Image

Scenario:
- CB account
- Pooled entitlements (no ULB) exhausted 
- Overages not enabled 

I sent a chat message and only got the in chat response. There is no warning shown above the chat input and there is no added styling in the copilot status dashboard. I think we are missing both of these UI elements. The current state is a little confusing because the status dashboard says "included" but then the in chat messaging says "you've exhausted your credits"

sotiamaestro:
Fixes #316086

When pooled entitlements are exhausted (no ULB, overages off), the in-chat message correctly tells the user their credits are gone but the status dashboard still says "included" and there's no warning banner above the input. Other SKUs already handle this, pooled just got missed.

This brings it in line with how we surface exhaustion everywhere else. The entitlement service now tracks `isExhausted` for pooled credits and sets `chatQuotaExceeded` accordingly. The dashboard shows the org warning and "Exhausted" styling like we do for other plans. Status bar picks up the warning icon and "Quota reached" for Business/Enterprise. Chat input banner shows the org-specific message above the input. Send button gets disabled so users aren't firing messages into a rejection. Change events fire both ways so everything clears on billing reset or overage enablement.

Per the discussion in the issue, no checkpoint or approaching-limit warnings here since we can't really predict per-user usage against a shared pool.

Tests added in `chatEntitlementService.test.ts` for the pooled exhaustion path.

### DIFF
--- a/extensions/copilot/src/extension/chatInputNotification/vscode-node/chatInputNotification.contribution.ts
+++ b/extensions/copilot/src/extension/chatInputNotification/vscode-node/chatInputNotification.contribution.ts
@@ -192,7 +192,11 @@ export class ChatInputNotificationContribution extends Disposable {
 				{ label: vscode.l10n.t('Upgrade'), commandId: 'workbench.action.chat.upgradePlan' },
 			];
 		} else if (isManagedPlan) {
-			notification.description = vscode.l10n.t('Contact your admin to increase your limits.');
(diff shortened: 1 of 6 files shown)


sotiamaestro:

PR #316114, opened 2026-05-12
chat: custom trajectory compaction model

gryan11:
﻿# chat: route conversation-history compaction to the trajectory-compaction CAPI model (`usePrismCompaction`)

Adds a flag-gated route for conversation-history compaction so we can run summarisation against a dedicated, lighter model instead of the main agent endpoint.

## What changes

When `chat.conversationCompaction.usePrismCompaction` is enabled **and** the active agent model matches `chat.conversationCompaction.prismModelFilter`, both the foreground (`/compact`) and background auto-compaction paths resolve their endpoint to the `trajectory-compaction` model via the standard CAPI endpoint provider (regular chat completions ΓÇö not a separate proxy). When the flag is off, behavior is **byte-identical** to upstream (the off-flag execution path is physically the original code).

### Settings

- `chat.conversationCompaction.usePrismCompaction` (ExperimentBased boolean, default `false`) ΓÇö master enable flag
- `chat.conversationCompaction.model` (ExperimentBased string, default `''`) ΓÇö when set, overrides the default `trajectory-compaction` family
- `chat.conversationCompaction.prismModelFilter` (ExperimentBased string, default `'claude-haiku-4.5,claude-sonnet-4.5,claude-sonnet-4.6,gemini-2.5-pro,gemini-3-flash,gemini-3.5-flash'`) ΓÇö comma-separated, case-insensitive substring filter against the agent model id and family; empty string = no filter

All three settings are registered in `package.json` (tagged `experimental` + `onExp`) and have localized descriptions in `package.nls.json`.

### Routing (`decidePrismRouting`)

Single source of truth shared by the foreground entry point (`ConversationHistorySummarizer.getSummary`) and the background dispatcher in `AgentIntentInvocation`. Decision order:

1. `usePrismCompaction` flag off ΓåÆ main agent
2. Agent model not in `prismModelFilter` ΓåÆ main agent
3. Otherwise ΓåÆ prism

The dispatchers then physically branch on the decision: prism requests go through `_getSummaryPrism` / `_startPrismBackgroundSummarization`, and off-flag requests go through the unmodified `_getSummary` / `_startBackgroundSummarization` paths.

### Safety nets on the prism path

- **Post-render pruning detection** ΓÇö a prompt-tsx tracer counts `removedCount` during render; if any node was pruned to fit the compaction endpoint, throw `PruningOccurredError` and fall back to the agent endpoint so the summary sees the full conversation.
- **Prompt-too-long fallback** ΓÇö `BudgetExceededError` and server-side `context_length_exceeded` thrown from `_getSummaryPrism` also fall back to the agent endpoint.
- **Length truncation handled inline** ΓÇö `ChatFetchResponseType.Length` on the prism path uses `response.truncatedValue` as a partial summary (warn-logged) instead of throwing; only the prism path tolerates this, so off-flag callers continue to fail closed.
- **Cancellation skipped in telemetry** ΓÇö token-driven cancellation in the prism background path (`/compact`, session dispose, endpoint switch) re-throws without error log or failure telemetry.
- **Cache-parity gates skipped** ΓÇö the prism background branch hardcodes `cacheWarm=true` and skips `_lastModelCapabilities` since the prism endpoint shares no prompt-cache prefix with the main agent loop.

### Endpoint provider

`endpointProvider.getChatEndpoint(family)` used to only recognize the utility aliases (`copilot-utility`, `copilot-utility-small`); it now falls through to `getChatModelFromCapiFamily` for arbitrary CAPI families so `trajectory-compaction` (or any other CAPI-registered family) resolves by name.

## Diff scope

| File | Notes |
| --- | --- |
| `extension/prompts/node/agent/compactionEndpoint.ts` (new) | `resolveCompactionEndpoint`, `buildCompactionToolOpts`, `formatCompactionFailureError`, `renderCompactionMessages`, `DEFAULT_COMPACTION_MODEL = 'trajectory-compaction'` |
| `extension/prompts/node/agent/summarizedConversationHistory.tsx` | Adds `decidePrismRouting`, `matchesPrismFilter`, `PruningOccurredError`, `_getSummaryPrism`, `_renderSummarizationPromptWithTracer`, `_isPromptTooLongError`. Off-flag path (`_getSummary`, `_renderSummarizationPrompt`, `handleSummarizationResponse`) is unchanged from pre-PR. |
| `extension/intents/node/agentIntent.ts` | Background dispatcher branches on `decidePrismRouting`: routes to new `_startPrismBackgroundSummarization` (re-renders against compaction endpoint, handles Length + cancellation, telemetry uses `compactionEndpoint.model` on both success and failure) when prism is on; original `_startBackgroundSummarization` is restored byte-for-byte to its pre-PR form. |
| `extension/prompt/vscode-node/endpointProviderImpl.ts` | `getChatEndpoint` resolves arbitrary CAPI families, not just utility aliases |
| `extension/prompts/node/agent/backgroundSummarizer.ts` | Extracts `shouldKickOffBackgroundSummarization` + `BackgroundSummarizationThresholds`, adds `endpointId` to the constructor, widens the work-callback return type to `ΓÇª \| undefined` |
| `platform/configuration/common/configurationService.ts` | 3 new `ConfigKey` entries |
| `package.json` + `package.nls.json` | Schema + localized descriptions for the 3 settings |
| `extension/prompts/node/agent/test/compactionEndpoint.spec.ts` (new) | Unit tests for `resolveCompactionEndpoint` (default, prism-flag-on routing, fallback on resolve failure) |
## Evaluations

Evaluations on MS-Bench summarization bench show using the trajectory compaction model is equivalent or slightly beneficial compared to main agent model trajectory compaction with lower cost. See detailed results and analysis here: [Trajectory Compaction Model Proposal](https://microsoft-my.sharepoint.com/:w:/p/ryangabriel/cQr3eqGMNZcVTqU4qjyZdiu7EgUCVMWjmGQl6qK97ZTRp1x6MQ)

## Tests

- All unit tests in `compactionEndpoint.spec.ts` (3), `summarization.spec.tsx` (34), `agentPrompt.spec.tsx` (11), and `backgroundSummarizer.spec.ts` (23) pass. Net additions from this PR: 3 cases — the new `compactionEndpoint.spec.ts` (resolveCompactionEndpoint: default / on-flag / fallback). Other spec files are unchanged from upstream main.
- Off-flag execution path is the physically-original code, so existing summarisation behavior is preserved

Extended dev unit tests for the internal helpers, smoke scripts, and a local-testing doc live on the companion `ryangabriel/custom_trajectory_compaction_model-dev` branch for manual end-to-end verification.

## Manual verification

Set in user settings:

```jsonc
{
  "github.copilot.chat.conversationCompaction.usePrismCompaction": true
}
```

Reload window, trigger `/compact` or let auto-compaction run on a long session. The chat debug log entry's `model` field should read `trajectory-compaction` for sessions where the agent model matches `prismModelFilter`.

To narrow which agent models route through prism, override the filter:

```jsonc
{
  "github.copilot.chat.conversationCompaction.prismModelFilter": "claude-sonnet-4.6"
}
```

## Commits (squash-merge recommended)

Initial:
- `1324b357667` route trajectory compaction through CAPI (model: trajectory-compaction)
- `8a32669d44f` allow BackgroundSummarizer work callback to resolve undefined
- `86970cfdbb5` replace images with text placeholders only when crossing to compaction endpoint
- `c2c56e9c0ec` trim PR diff per senior-engineer review (rename `useAgenticProxy` ΓåÆ `usePrismCompaction`; move `renderCompactionMessages` into `compactionEndpoint.ts`; extract background cross-endpoint branch into a helper)
- `02c4e792ebd` trim non-essential tests, smoke scripts, and docs (dev-only)
- `dfb91afab6a` address PR review (branch prism path, register settings)

Routing, gating, and safety nets:
- `7c428495db4` resolve arbitrary CAPI family in endpointProvider
- `82b7859e93a` capacity-gated prism fallback + length/cancellation handling
- `dd4c4397d2c` opt prism in per agent model via configurable filter
- `f7e3e032209` default prism filter to specific CAPI model IDs
- `32b9385b7e8` add gemini-3-flash and gemini-3.5-flash to default prism filter
- `2a2d4f758ba` apply prism model filter to background path too
- `1dc332e83c9` unify prism routing across foreground/background and add debug logs
- `073c7593f95` simplify capacity rule in decidePrismRouting
- `922542445fc` keep off-flag execution path byte-identical to pre-prism
- `2bc83bae9de` handle Length response on prism foreground path

Follow-up review feedback:
- `3c403138122` address Bhavya review (restore `stream: false` on off-flag foreground; attribute prism failure telemetry to `compactionEndpoint.model`)
- `008f730b6e4` drop `currentContextTokens` prop, rely on post-render pruning check
- `ce0c803abae` clean if/else split between prism and production paths in agentIntent.ts dispatcher
- `66a3cf98f63` minimize PR diff + clean prism/production split in summarizedConversationHistory.tsx
- `b0c472e937d` trim verbose comments to minimize PR diff
- `11c9811784e` label prism vs production branches in dispatchers
- `ab2a4872467` drop `stream:false` from prism foreground request
- `c16cf33e63b` trim added test cases to minimum needed coverage
- `0a9e6f8751a` restore appendTranscriptHintToSummary suite (was preexisting on main)
- `8f39af6fb34` revert BackgroundSummarizer.start callback signature widening (use CancellationError on empty render)
- `c31b1997c76` restore shouldKickOffBackgroundSummarization test suite (was preexisting on main)


extensions/copilot/src/platform/configuration/common/configurationService.ts
```diff
@@ -920,6 +920,21 @@ export namespace ConfigKey {
...
 		export const SessionSyncSafetyIntervalMs = defineTeamInternalSetting<number>('chat.advanced.sessionSync.safetyIntervalMs', ConfigType.ExperimentBased, 60_000);
+
+		/**
+		 * Use the prism `trajectory-compaction` model (resolved via the standard CAPI endpoint provider) for
+		 * conversation-history compaction — both foreground `/compact` and background auto-compaction.
+		 *
+		 * Lives under `chat.advanced.*` so the setting is unlisted in package.json (team-internal), but the
+		 * experiment lookup keys preserve the pre-rename names so existing flights continue to resolve.
+		 */
+		export const ConversationUsePrismCompaction = defineTeamInternalSetting<boolean>('chat.advanced.conversationCompaction.usePrismCompaction', ConfigType.ExperimentBased, false, undefined, undefined, { experimentName: 'copilotchat.config.chat.conversationCompaction.usePrismCompaction' });
```
bhavyaus: these need to go into the package.json as well to be toggleable easily. 
Also remove `advanced` from the setting id. 

Unless the idea is to enable this model for internal only. At which point, our exp service should be able to handle that and we should be ok tagging this as team internal.

bhavyaus:
A couple of design suggestions to shrink the blast radius of this change. The core concern: the off-flag path now runs through the same modified shared helpers, so it's no longer guaranteed to be byte-identical to today. I'd suggest inverting the structure so the flag-off path is physically the original code.

**1. Branch instead of intertwine.** These are two different algorithms — fork-and-append for cache parity vs. re-render-self-contained — so splitting them keeps the existing path untouched and avoids any unintended changes to it:

```ts
if (prismEnabled) {
    this._startPrismBackgroundSummarization(...);  // re-render against compaction endpoint
} else {
    this._startBackgroundSummarization(...);       // unchanged — exactly today's code
}
```

**2. Skip the warm-cache wait on the prism path.** The `cacheWarm` gate (`shouldKickOffBackgroundSummarization`) exists *only* to preserve prompt-cache prefix parity with the **main** agent endpoint — hence it waits for a completed tool-call round to fire at ~0.80, otherwise holding out for the 0.90 cold-cache emergency. Routing to a *different* compaction endpoint shares no prefix, so that rationale evaporates: we should kick off as soon as the budget threshold is crossed (`cacheWarm: prismEnabled ? true : cacheWarm`). Waiting for a tool round here only delays compaction with zero cache benefit. The `_lastModelCapabilities` block can be skipped too — it's also cache-parity machinery.

**3. Add a user-visible opt-out.** `usePrismCompaction` is a `TeamInternal` experiment flag today. Since it changes *which model* summarizes the conversation, users who manage summarization closely (custom thresholds, `/compact` discipline) will be surprised by it. I'd promote it to a documented, user-settable config — at minimum respecting an explicit `false` so the experiment can't override an opted-out user.

extensions/copilot/src/extension/prompts/node/agent/compactionEndpoint.ts
```diff
@@ -0,0 +1,61 @@
...
+
+import { ConfigKey, IConfigurationService } from '../../../../platform/configuration/common/configurationService';
+import { ChatEndpointFamily, IEndpointProvider } from '../../../../platform/endpoint/common/endpointProvider';
+import { ProxyAgenticEndpoint } from '../../../../platform/endpoint/node/proxyAgenticEndpoint';
+import { ILogService } from '../../../../platform/log/common/logService';
+import { IChatEndpoint } from '../../../../platform/networking/common/networking';
+import { IExperimentationService } from '../../../../platform/telemetry/common/nullExperimentationService';
+import { IInstantiationService } from '../../../../util/vs/platform/instantiation/common/instantiation';
+
+/** Default model name forwarded to the agentic proxy when `ConversationCompactionUseAgenticProxy` is set without an explicit `ConversationCompactionModel`. Registered on the copilot-proxy as `trajectory-compaction-v1` (routes to Fireworks deployment `accounts/msft/deployments/ihfptseo`). */
```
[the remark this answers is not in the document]
gryan11: No longer applicable as of ae27ebebb6b.

The doc comment is now `Default CAPI model family used for trajectory-compaction when ConversationUsePrismCompaction is enabled without an explicit ConversationCompactionModel.` â€” no deployment id, no proxy reference. The PR routes through CAPI only; the model name registered there is `trajectory-compaction`.

extensions/copilot/src/extension/intents/node/agentIntent.ts
```diff
@@ -969,7 +987,7 @@ export class AgentIntentInvocation extends EditCodeIntentInvocation implements I
 					...summaryMsgResult.messages,
 				];
 
-				const response = await this.endpoint.makeChatRequest2({
+				const response = await compactionEndpoint.makeChatRequest2({
 					debugName: 'summarizeConversationHistory',
 					messages,
 					finishedCb: undefined,
```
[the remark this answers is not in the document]
gryan11: No longer applicable as of ae27ebebb6b.

The background path is now split: `_startBackgroundSummarization` is restored to its pre-PR form (uses `this.endpoint` for both rendering `SummarizationUserMessage` and the request â€” model-specific instructions stay matched with the endpoint actually called), and `_startPrismBackgroundSummarization` handles the prism path. The prism path doesn't use `SummarizationUserMessage` at all; it re-renders the entire compaction prompt against `compactionEndpoint` via `renderCompactionMessages`. Either way the prompt content is always rendered against the same endpoint the request goes to.

bhavyaus:
**Flag-off regression: foreground `/compact` now streams where it didn't before.**

On `main`, the foreground summarization request explicitly sets `stream: false` (`summarizedConversationHistory.tsx:757`). After the refactor into the shared `_executeSummarizationRequest`, `requestOptions` only carries `{ temperature: 0, ...toolOpts }` and `stream` is gone.

That matters because `messagesApi.ts:232` defaults a missing `stream` to `true`:

```ts
stream: options.requestOptions?.stream ?? true,
```

So with the flag off, Anthropic / Messages-API compaction requests now stream where they previously didn't. This breaks the "flag-off is byte-identical to upstream" guarantee we agreed on for the non-forked foreground path.

Can we restore `stream: false` in `_executeSummarizationRequest` (or pass it through per-path) so the disabled path stays identical to upstream?

bhavyaus:
**Prism background failure telemetry attributes the wrong model.**

In `_startPrismBackgroundSummarization`, the success path logs `model: compactionEndpoint.model` (~line 1254), but the catch / failure path logs `model: this.endpoint.model`, which is the main agent model (~line 1290). That means `trajectory-compaction` failures get charged to the main model, so success vs. failure telemetry for the new route isn't comparable.

Should be `compactionEndpoint.model` in the failure path too.

bhavyaus (changes requested):

extensions/copilot/src/platform/configuration/common/configurationService.ts
gryan11: Updated to address. Thes settings are now:

| Setting ID | Type | Default |
|---|---|---|
| `github.copilot.chat.conversationCompaction.usePrismCompaction` | boolean (ExperimentBased) | `false` |
| `github.copilot.chat.conversationCompaction.model` | string (ExperimentBased) | `""` |
| `github.copilot.chat.conversationCompaction.prismModelFilter` | string (ExperimentBased) | `"claude-haiku-4.5,claude-sonnet-4.5,claude-sonnet-4.6,gemini-2.5-pro,gemini-3-flash,gemini-3.5-flash"` |

gryan11:
All three addressed:

1. **Branch instead of intertwine** (`dfb91afab6a`, `ce0c803abae`, `66a3cf98f63`) — dispatcher now does `if (routingDecision.usePrism) { this._startPrismBackgroundSummarization(...) } else { this._startBackgroundSummarization(...) }`. `_startBackgroundSummarization` is restored byte-for-byte to its pre-PR form; the prism path is a separate method that re-renders against the compaction endpoint.

2. **Skip the warm-cache wait on the prism path** — the prism branch hardcodes `const cacheWarm = true` with a comment explaining the cache-prefix rationale doesn't apply, and the `_lastModelCapabilities` cache-parity block runs only on the production branch.

3. **User-visible opt-out** — all three settings are registered in `package.json` (`chat.conversationCompaction.{usePrismCompaction, model, prismModelFilter}`) with `experimental` + `onExp` tags, so an explicit user value wins over the experiment.

gryan11:
Fixed in `3c403138122`. `_getSummary` (off-flag / production path) sets `stream: false` again, restoring byte-identical behavior to upstream. The prism path in `_getSummaryPrism` intentionally omits it (commented inline). Verified by `922542445fc`, which keeps the off-flag execution path identical to pre-prism.

gryan11:
Fixed in `3c403138122`. `compactionEndpoint` is hoisted above the `try` so the catch block in `_startPrismBackgroundSummarization` logs `model: compactionEndpoint?.model ?? this.endpoint.model`. The fallback to `this.endpoint.model` only triggers if `resolveCompactionEndpoint` itself threw (i.e. no compaction model was picked), which is the correct attribution in that case. The success path already used `compactionEndpoint.model`, so prism success vs. failure telemetry is now comparable.

extensions/copilot/src/platform/configuration/common/configurationService.ts
gryan11: Done in `dfb91afab6a`. The three settings are now in `package.json`:

- `github.copilot.chat.conversationCompaction.usePrismCompaction`
- `github.copilot.chat.conversationCompaction.model`
- `github.copilot.chat.conversationCompaction.prismModelFilter`

All tagged `experimental` + `onExp`. Dropped the `advanced` segment from the id; kept them user-toggleable so an explicit user value can override the experiment.

bhavyaus (dismissed):

roblourens (dismissed):

bhavyaus:
@Yoyokrazy pinging here - need help adding Gabe

bhavyaus (approved):

PR #316166, opened 2026-05-13
run_in_terminal: promote sync command to background after idle silence

meganrogge:
Re-land of #315885 on a standalone branch for isolated eval testing.

### What this does

If a synchronous `run_in_terminal` call produces **no terminal output for N ms** (default 60s), a new `idleSilence` race candidate wins the foreground race and promotes the execution to background. The process is **never killed** — the model gets the terminal ID + output so far and can `get_terminal_output`, `send_to_terminal`, or `kill_terminal`.

### Changes (3 files)
- **`runInTerminalTool.ts`**: New `idleSilence` race candidate using `RunOnceScheduler` + `onData` listener. Refactors `_buildInputNeededSteeringText` from `mentionTimeout: boolean` to `hungHint: 'none' | 'timeout' | 'idleSilence'` discriminator with per-mode wording.
- **`terminalChatAgentToolsConfiguration.ts`**: New setting `chat.tools.terminal.idleSilenceTimeoutMs` (default 60000, 0 disables, experimental).
- **`runInTerminalTool.test.ts`**: 3 unit tests for steering text across all hint modes.

### Why this is safe
- Commands that produce output regularly (`npm install`, `cargo build`, etc.) reset the timer and never trip.
- The async path is unchanged — it already has `OutputMonitor` idle detection.
- Listener + scheduler owned by `raceCleanup` `DisposableStore` — disposed when any other race candidate wins.


meganrogge:

mjbvz (approved):

PR #316172, opened 2026-05-13
feat: add session changesets

DonJayamanne:
For https://github.com/microsoft/agent-host-protocol/issues/92


src/vs/platform/agentHost/common/changesetUri.ts
```diff
@@ -0,0 +1,100 @@
...
+ * at the call site or a more permissive parser, both of which are
+ * deliberately out of scope here.
+ */
+
+/**
+ * Stable id of the catalogue entry for the session-wide changeset (the v1
+ * producer's only entry). Shared across server-side producers and
+ * client-side consumers so the literal `'session'` string isn't repeated.
+ */
+export const SESSION_CHANGESET_ID = 'session';
```
DonJayamanne: We have a concept of the default changeset in the list of changesets.
This is the one that needs to be psersisted and displayed in list of sessions
& this will be the default view for changesets as well.

Either we hardcode an id or we hardcode the uri template.

src/vs/platform/agentHost/node/agentService.ts
```diff
@@ -48,6 +50,43 @@ import { toAgentClientUri } from '../common/agentClientUri.js';
...
+/**
+ * Build a {@link ChangesetSummary | catalogue entry} for the
+ * session-wide changeset directly from a persisted file list. Used by
+ * {@link AgentService.listSessions} to surface the chip counts for
+ * sessions that have not yet been opened (and therefore have no entry
+ * in the live state manager's `summary.changesets`). When the session
+ * is later opened, the live overlay takes over and this synthesised
+ * entry is replaced.
+ */
+function synthesizeSessionChangesetCatalogue(sessionUri: string, diffs: readonly ISessionFileDiff[]): ChangesetSummary {
```
DonJayamanne: The default changeset

src/vs/platform/agentHost/common/changesetUri.ts
DonJayamanne: As discussed in the standup, I will remove this

src/vs/platform/agentHost/common/state/agentSubscription.ts
```diff
@@ -471,8 +504,13 @@ export class AgentSubscriptionManager extends Disposable {
...
 			case StateComponents.Terminal:
 				return new TerminalStateSubscription(key, this._clientId, this._log);
+			case StateComponents.Changeset:
+				return new ChangesetStateSubscription(key, this._clientId, this._log);
 			default:
-				return new TerminalStateSubscription(key, this._clientId, this._log);
+				// Defensive: a new `StateComponents` value added without a
+				// matching arm above would otherwise silently get a
+				// terminal subscription with the wrong reducer/filter.
+				throw new Error(`_createSubscription: unsupported StateComponents kind: ${kind}`);
```
connor4312: You can use assertNever to get a type-level assertions

src/vs/platform/agentHost/node/agentHostStateManager.ts
```diff
@@ -326,6 +390,115 @@ export class AgentHostStateManager extends Disposable {
...
+		});
+		this._changesetStates.delete(changeset);
+	}
+
+	/**
+	 * Disposes every changeset whose URI is nested under `session` (i.e.
+	 * matches `<session>/changeset/...`). Used to cascade cleanup when a
+	 * session itself is removed.
+	 */
+	disposeSessionChangesets(session: URI): void {
```
connor4312, line 489: This can probably be private and called from inside `removeSession` as it's never called independently

src/vs/platform/agentHost/node/agentService.ts
```diff
@@ -250,8 +306,69 @@ export class AgentService extends Disposable implements IAgentService {
...
+							// surface the synthesised entry on the
+							// returned metadata instead. When the session
+							// is later restored / created, the live state
+							// overlay below replaces this with the
+							// authoritative catalogue.
+							if (!liveSessionState) {
+								const synthesized = synthesizeSessionChangesetCatalogue(sessionStr, parsed);
+								updated = { ...updated, changesets: [synthesized] };
+							}
+						}
```
connor4312, line 381: Wonder if we could encapulsate this into some `IAgentHostChangeSetService` -- the agentService itself is already kinda big and does a lot

src/vs/platform/agentHost/node/agentService.ts
```diff
@@ -383,6 +506,12 @@ export class AgentService extends Disposable implements IAgentService {
 			// session, working directory, etc.
 			this._stateManager.dispatchServerAction({ type: ActionType.SessionReady, session: session.toString() });
 
+			// Publish the default `session` changeset catalogue entry on
+			// `summary.changesets` immediately so clients see it without
+			// having to wait for the first turn to complete. Idempotent —
+			// the diff producer also publishes it on every compute pass.
+			this._sideEffects.publishSessionChangesetCatalogue(session.toString());
```
connor4312: This mostly just sets the changesets into the summary, should this be part of `_buildInitialSummary`?

src/vs/platform/agentHost/node/agentService.ts
```diff
@@ -445,6 +574,10 @@ export class AgentService extends Disposable implements IAgentService {
 		// see consistent state through both paths.
 		this._stateManager.markSessionPersisted(sessionKey, summary);
 		this._stateManager.dispatchServerAction({ type: ActionType.SessionReady, session: sessionKey });
+		// Publish the default `session` changeset catalogue entry on
+		// `summary.changesets` immediately so newly-materialized sessions
+		// expose the same catalogue entry as eagerly-created ones.
+		this._sideEffects.publishSessionChangesetCatalogue(sessionKey);
```
connor4312: We can set the changesets initially and not need to trigger them when the session is materialized. I think the model I had in my head is that the changesets would be seeded initially and then only their diff stats would update over time. Or maybe in the rare case we add/remove a changeset entirely, but I think this is not something we have right now (like we discussed yesterday evening)

src/vs/platform/agentHost/node/agentSideEffects.ts
```diff
@@ -1010,30 +1076,140 @@ export class AgentSideEffects extends Disposable {
...
+			},
+			{ additions: 0, deletions: 0 },
+		);
+		const existing = sessionState.summary.changesets ?? [];
+		const next = existing.map(c => isSessionChangesetUri(c.uriTemplate)
+			? { ...c, additions: totals.additions, deletions: totals.deletions, files: nextFilesById.size }
+			: c,
+		);
+		this._stateManager.setSessionChangesets(session, next);
+	}
```
connor4312: I also think a chunk of this could live in its own service :)

connor4312 (commented):
Overall sensible I think, some comments

connor4312 (approved):

PR #316181, opened 2026-05-13
[Bug Fix] Execution Subagent: Compare shell type with "pwsh" in addition to "powershell"

vikramnitin9:
Addresses a bug in #313521 . `shellType` can be `powershell` or `pwsh` for Powershell, but the check in #313521 only checks for `powershell`.

### DIFF
--- a/extensions/copilot/src/extension/prompt/node/executionSubagentToolCallingLoop.ts
+++ b/extensions/copilot/src/extension/prompt/node/executionSubagentToolCallingLoop.ts
@@ -119,7 +119,7 @@ export class ExecutionSubagentToolCallingLoop extends ToolCallingLoop<IExecution
 
 		if (useAgenticProxy) {
 			// Our custom models are not trained for PowerShell yet. Fall back to main agent endpoint.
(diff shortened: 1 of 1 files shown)


extensions/copilot/src/extension/prompt/node/executionSubagentToolCallingLoop.ts
```diff
@@ -169,6 +169,9 @@ export class ExecutionSubagentToolCallingLoop extends ToolCallingLoop<IExecution
 		const hadBackgroundBefore = this._backgroundCommands.length > 0;
 		let result = await render(hadBackgroundBefore);
 
+		// Artificial 2 minute delay
+		await new Promise<void>(resolve => setTimeout(resolve, 2 * 60 * 1000));
```
bhavyaus: why is this delay needed?

vikramnitin9: Oops, that was not supposed to be committed. Removed.

bhavyaus (approved):

PR #316202, opened 2026-05-13
BYOK: Enable Manage Language Models commands without GitHub sign-in

dmitrivMS:
First step towards offline BYOK: enable Manage Language Models commands/dialog without sign-in.

Enterprise policy should still be respected in case of sign-in - so BYOK providers will be hidden if BYOK is disallowed by policy.


extensions/copilot/src/extension/byok/common/byokProvider.ts
```diff
@@ -169,14 +167,17 @@ export function byokKnownModelToAPIInfo(providerName: string, id: string, capabi
 	};
 }
 
-export function isBYOKEnabled(copilotToken: Omit<CopilotToken, 'token'>, capiClientService: ICAPIClientService): boolean {
-	if (isScenarioAutomation) {
+/**
+ * Signed-out users are allowed; signed-in users without a Copilot token (e.g. enterprise-managed errors) are denied to avoid bypassing policy.
+ */
+export function isClientBYOKAllowed(hasGitHubSession: boolean, copilotToken: Omit<CopilotToken, 'token'> | undefined): boolean {
```
vritant24, line 173: this is interesting. I wasn't aware that enterprise managed users are not allowed to use BYOK.
Can you add a test for this scenario if it doesn't already exist? I want to make sure I don't regress that in future changes

extensions/copilot/src/extension/byok/vscode-node/byokContribution.ts
```diff
@@ -28,74 +27,89 @@ export class BYOKContrib extends Disposable implements IExtensionContribution {
...
 			for (const [providerId, provider] of this._providers) {
 				this._byokRegistrations.add(lm.registerLanguageModelChatProvider(providerId, provider));
 			}
+			this._registeredWithLM = true;
 			this._logService.info(`BYOK: registered ${this._providers.size} provider(s): ${Array.from(this._providers.keys()).join(', ')}`);
-		} else if (!this._byokProvidersRegistered) {
-			const copilotToken = authService.copilotToken;
-			this._logService.debug(`BYOK: not enabling — copilotToken=${copilotToken ? 'present' : 'absent'}, isScenarioAutomation=${isScenarioAutomation}, isInternal=${copilotToken?.isInternal ?? 'n/a'}, isIndividual=${copilotToken?.isIndividual ?? 'n/a'}`);
+		} else if (!allowed && this._registeredWithLM) {
+			this._byokRegistrations.clear();
```
vritant24: does `_providers` also need to be cleared here?

vritant24 (commented):
Looks good, just a couple comments

extensions/copilot/src/extension/byok/common/byokProvider.ts
dmitrivMS: Will do. We want to allow GHE users to use BYOK with this change, unless a policy explicitly states otherwise - that's a change in behavior.

extensions/copilot/src/extension/byok/vscode-node/byokContribution.ts
dmitrivMS: No, it keeps a list of providers we may add later if situation changes.

vritant24 (dismissed):

osortega (approved):

vijayupadya (approved):

PR #316271, opened 2026-05-13
run_in_terminal: detect shell continuation prompts and abort

### LINKED ISSUE #315694: bad quoting can cause terminal to hang
From @roblourens, but i've also hit this

```
(.venv) meganrogge@Megans-MacBook-Pro vscode-copilot-evaluation %  cd /Users/meganrogge/Repos/vscode-copilot-evaluation && IDS=$(cat /tmp/vscbench-ids.txt) && nohup bash -c "source .venv/bin/activate && runtest local -b $IDS --config doc/examples/configs/agent-host.vscode.agent.yaml --parallel 4 --output ./.msbenchRun/local/agent-host-full" > /tmp/runtest-agent-host-full.log 2>&1 &
echo "PID=$!"; sleep 5; tail -20 /tmp/runtest-agent-host-full.log
dquote> 
```

Causes the agent to just hang

With request

```
id   : toolu_01QvMK8gnShnmo13869xh25K__vscode-1778510350991
tool : run_in_terminal
args : {
  "command": "cd /Users/meganrogge/Repos/vscode-copilot-evaluation && IDS=$(cat /tmp/vscbench-ids.txt) && nohup bash -c \"source .venv/bin/activate && runtest local -b $IDS --config doc/examples/configs/agent-host.vscode.agent.yaml --parallel 4 --output ./.msbenchRun/local/agent-host-full\" > /tmp/runtest-agent-host-full.log 2>&1 &
echo \"PID=$!\"; sleep 5; tail -20 /tmp/runtest-agent-host-full.log",
  "explanation": "Launch full benchmark in background",
  "goal": "Start full run",
  "mode": "sync"
}
```

and Response

```
The user cancelled the tool call.
```

meganrogge:
Fixes #315694

When a command with unmatched quotes is sent to the terminal, the shell enters a continuation prompt (e.g. `dquote>`, `quote>`) and waits for more input. Previously this caused the terminal tool to hang indefinitely because no command-finished event ever fires.

### Changes

**`executeStrategy.ts`**:
- Adds `isContinuationPrompt()` to detect known shell continuation prompts (`dquote>`, `quote>`, `bquote>`, `pipe>`, `cmdsubst>`, `heredoc>`)
- Adds `waitForContinuationPrompt()` that monitors terminal idle state and checks for continuation prompt patterns on the cursor line
- Fixes `detectsCommonPromptPattern()` to exclude continuation prompts so they are not falsely detected as ready prompts (previously `dquote>` matched the generic `[>%]` pattern)

**All three execute strategies** (rich, basic, none):
- Integrates `waitForContinuationPrompt()` as a race candidate alongside existing completion signals
- When a continuation prompt is detected, sends Ctrl+C to abort the malformed command and returns an error message explaining the quoting issue

### How it works

The fix monitors the terminal data stream for idle periods. When the terminal goes idle, it checks if the cursor line matches a known continuation prompt pattern. If so, it:
1. Sends Ctrl+C to abort the stuck command
2. Returns an error to the agent: _"The command has a syntax error — the shell showed a \"dquote>\" continuation prompt, indicating unmatched quotes or brackets. Fix the quoting and retry."_

This prevents the hang and gives the agent actionable feedback to fix the command.

### DIFF
--- a/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/executeStrategy/basicExecuteStrategy.ts
+++ b/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/executeStrategy/basicExecuteStrategy.ts
@@ -11,7 +11,7 @@ import { isNumber } from '../../../../../../base/common/types.js';
 import { IConfigurationService } from '../../../../../../platform/configuration/common/configuration.js';
 import type { ICommandDetectionCapability } from '../../../../../../platform/terminal/common/capabilities/capabilities.js';
 import { ITerminalLogService, type ITerminalLaunchError } from '../../../../../../platform/terminal/common/terminal.js';
-import { trackIdleOnPrompt, waitForIdle, type ITerminalExecuteStrategy, type ITerminalExecuteStrategyResult } from './executeStrategy.js';
+import { trackIdleOnPrompt, waitForContinuationPrompt, waitForIdle, type ITerminalExecuteStrategy, type ITerminalExecuteStrategyResult } from './executeStrategy.js';
 import type { IMarker as IXtermMarker } from '@xterm/xterm';
 import { ITerminalInstance } from '../../../../terminal/browser/terminal.js';
 import { createAltBufferPromise, setupRecreatingStartMarker, stripCommandEchoAndPrompt } from './strategyHelpers.js';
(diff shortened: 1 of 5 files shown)


dmitrivMS (approved):

meganrogge:

meganrogge:
I actually think this is overkill / too narrowly scoped and will be fixed by https://github.com/microsoft/vscode/pull/315885

PR #316277, opened 2026-05-13
Prevent cache timeouts during long tool calls

vikramnitin9:
Continues #313748.

During long tool calls (especially subagent runs), the main agent cache might expire. This PR attempts to solve the issue by sending dummy user requests every 4 mins (up to a hardcoded number of probes; currently 3). These dummy requests are not captured in the main agent conversation history; they are purely side chain calls. They also will not be surfaced in the UI.

Right now, this is scoped to only `execution_subagent` calls, since those are the primary source of tool calls greater than 5 minutes in length.

I added an artificial 2 minute delay to all tool calls within `execution_subagent` trajectories, in order to force cache misses. In this setting, my PR resulted in more requests and more cached input tokens, but lower costs overall thanks to far fewer uncached tokens.

| Run | Requests | Cached input | Uncached input | Output | Cost |
   |---|---:|---:|---:|---:|---:|
   | With this PR's fix [`25744709918`](https://msbenchapp.azurewebsites.net/run-analysis/25744709918) | 2,327 | 48,697,895 | 4,332,684 | 31,129 | $29.26 |
   | Without this PR's fix [`25521098886`](https://msbenchapp.azurewebsites.net/run-analysis/25521098886) | 2,249 | 42,912,920 | 7,508,471 | 29,414 | $39.61 |


vikramnitin9:

extensions/copilot/src/extension/intents/node/toolCallingLoop.ts
```diff
@@ -1671,6 +1692,120 @@ export abstract class ToolCallingLoop<TOptions extends IToolCallingLoopOptions =
...
+					{
+						role: Raw.ChatRole.User,
+						content: [{ type: Raw.ChatCompletionContentPartKind.Text, text: 'Still working' }],
+					},
+				]
+				: [...processedMessages];
+
+			this._logService.info(`[ToolCallingLoop] Keep-alive: sending probe ${probeCount}/${maxProbes}`);
+			try {
+				await this.fetch({
```
bhavyaus, line 1775: how does this flow into telemetry? Should we consider a different source for this for reporting? `chat.cacheKeepAlive` or something like that?

extensions/copilot/src/extension/intents/node/toolCallingLoop.ts
```diff
@@ -1671,6 +1692,120 @@ export abstract class ToolCallingLoop<TOptions extends IToolCallingLoopOptions =
...
+			// the same validation/stripping (e.g. stripOrphanedToolCalls for Gemini)
+			// is applied. Only append a dummy user message if the last message was
+			// from the assistant.
+			const processedMessages = baseFetchOptions.messages;
+			const lastMessage = processedMessages[processedMessages.length - 1];
+			const probeMessages: Raw.ChatMessage[] = lastMessage.role === Raw.ChatRole.Assistant
+				? [
+					...processedMessages,
+					{
+						role: Raw.ChatRole.User,
```
bhavyaus, line 1767: Didn't get a chance to run this, but this is not user visible right?

bhavyaus: Probes would also surface in any usage metrics users are tracking. Probably fine, but anyone watching could be surprised to see usage tick up while a tool call is mid-flight.

extensions/copilot/src/platform/configuration/common/configurationService.ts
```diff
@@ -733,6 +733,13 @@ export namespace ConfigKey {
 		/** Internal: override reasoning/thinking effort sent to model APIs (e.g. Responses API, Messages API). Used by evals. */
 		export const ReasoningEffortOverride = defineSetting<string | null>('chat.reasoningEffortOverride', ConfigType.Simple, null);
 
+		/**
+		 * When enabled, periodic keep-alive probes are sent during long-running tool calls
+		 * to keep the server-side prompt cache warm.
+		 */
+		export const LongToolCallCachePreservation = defineSetting<boolean>('chat.agent.longToolCallCachePreservation.enabled', ConfigType.ExperimentBased, false);
+		export const LongToolCallCachePreservationMaxProbes = defineSetting<number>('chat.agent.longToolCallCachePreservation.maxProbes', ConfigType.ExperimentBased, 3);
```
bhavyaus: Can we lower this default to 1? Try one probe and bail? We can always experiment with higher values if needed

bhavyaus (changes requested):

extensions/copilot/src/extension/intents/node/toolCallingLoop.ts
vikramnitin9: I added a new subtype variant `'keep-alive-probe'` (in addition to `'subagent'`, `'system-initiated'` and `'undefined'`). However, the `messageSource` will be the same, and the event will still be `response.success`. Let me know if this is fine or if you'd prefer something else.

extensions/copilot/src/extension/intents/node/toolCallingLoop.ts
vikramnitin9: It's not user visible, yes.

extensions/copilot/src/extension/intents/node/toolCallingLoop.ts
bhavyaus: Thanks! I'm worried that just adding a new `subType` value will bleed into all our existing scorecards/dashboards — most of them filter/group by `messageSource` and `eventName` (`response.success`), and they'd need to be updated everywhere to exclude `subType == 'keep-alive-probe'`.
I'd prefer changing the source itself so probes are invisible to anything that doesn't explicitly opt in. Can we not set `messageSource: 'chat.cacheKeepAlive'` for probe fetches?

vikramnitin9: Done.

bhavyaus (approved):

dmitrivMS (approved):

PR #316278, opened 2026-05-13
Dispose finished background terminals to prevent accumulation

### LINKED ISSUE #287177: Make terminals ephemeral per turn
Now that sessions aren't archived when one clicks `+`, one can have many terminals hanging around potentially. Since we retain their output, we should consider clearing them per turn.

Request from @hawkticehurst and @pierceboggan

meganrogge:
Fixes #287177

Background terminals created by the chat agent tools are never reused once their command completes, but they remain alive indefinitely, accumulating across turns.

This PR disposes background terminals as soon as their command finishes (onCommandFinished), provided the user has not revealed the terminal. Terminal output remains viewable in the chat UI thanks to the output mirror snapshot captured before disposal.

Key changes:

- **Auto-dispose on completion**: When a background terminal command finishes and the terminal is still hidden (hideFromUser), dispose the terminal instance and clean up the active execution entry.
- **Snapshot before disposal**: Capture the terminal command output via `TerminalCommandArtifactCollector.capture` before disposing, so the progress part can still render output via the snapshot mirror even after the terminal instance is gone.
- **Suppress redundant steering**: Mark the terminal as programmatically killed before disposal so the onDisposed handler does not send a redundant "terminal exited" steering message.
- **Updated completion message**: Background terminals that will be disposed get a cleaner notification ("The terminal has been cleaned up") instead of suggesting send_to_terminal/kill_terminal on a terminal that no longer exists.

https://github.com/user-attachments/assets/97543b50-2d79-4940-9cea-589bacda74c8


meganrogge:

benvillalobos (dismissed):

dmitrivMS (approved):

PR #316291, opened 2026-05-13
tool call events for AH and Claude CLI as well

amunger:
- emit toolcall events for claude CLI 
- Kept Ownership to Rob since he owns the event from the toolcallservice.
- need to take a look at getting the telemetry service wired up in AH, so left that out for now


amunger:
cc @TylerLeonhardt

rzhao271 (approved):

PR #316313, opened 2026-05-13
Disable memory feature via hardcoded setting

natagdunbar:
## Summary

Adds a `github.copilot.chat.memory.enabled` setting (default: `false`) that gates all memory touchpoints. This effectively disables memory across the board until we're ready to re-enable it.

## What's gated

- `copilot_memory` and `copilot_resolveMemoryFileUri` tools (`when` clause)
- `'memory'` in `sessionCapabilities` for Copilot CLI sessions
- Native Anthropic memory tool (`memory_20250818`) in BYOK provider
- `showMemories`/`clearMemories` commands (hidden from command palette)
- Claude `/memory` slash command and menu item

## Re-enabling

Flip `default: false` → `default: true` in both `configurationService.ts` and `package.json`, or remove the gating entirely.
Closes: https://github.com/github/agents/issues/953

### DIFF
--- a/extensions/copilot/package.json
+++ b/extensions/copilot/package.json
@@ -5953,7 +5953,7 @@
 				},
 				{
 					"command": "copilot.claude.memory",
-					"when": "chatCustomizationSessionType == claude-code && chatCustomizationSection == instructions",
+					"when": "config.github.copilot.chat.memory.enabled && chatCustomizationSessionType == claude-code && chatCustomizationSection == instructions",
 					"group": "navigation@1"
 				}
 			]
--- a/extensions/copilot/package.json
+++ b/extensions/copilot/package.json
@@ -1161,6 +1161,7 @@
 				"name": "copilot_memory",
 				"displayName": "Memory",
 				"toolReferenceName": "memory",
+				"when": "config.github.copilot.chat.memory.enabled",
 				"userDescription": "Manage persistent memory across conversations",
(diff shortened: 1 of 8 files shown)


natagdunbar:
@microsoft-github-policy-service agree company="GitHub"

bhavyaus (changes requested):
We should not be removing local memory. it is used in context editing, plan mode etc.  github copilot memory remote memory has already been removed 

https://github.com/microsoft/vscode/pull/315813

natagdunbar:
Thanks for the review! To clarify — the intent here is not to turn off VS Code's internal/local memory (used for plan mode, context editing, etc.). We only want to disable the setting that comes from our (GitHub) side. We'll scope this down so it doesn't affect your internal memory functionality. Let us know if there's a preferred way to delineate between the two, or if we should key off a different config path.

bhavyaus:
> Thanks for the review! To clarify — the intent here is not to turn off VS Code's internal/local memory (used for plan mode, context editing, etc.). We only want to disable the setting that comes from our (GitHub) side. We'll scope this down so it doesn't affect your internal memory functionality. Let us know if there's a preferred way to delineate between the two, or if we should key off a different config path.

 As I noted above, #315813 already removed the GitHub remote memory feature, so this change isn't necessary. The PR description also doesn't match the diff.
 
 Concretely: the gate in anthropicProvider.ts only affects BYOK Anthropic sessions. The VS Code agent harness — plan mode, context editing, and local session/user/repo-scoped memory — is not on this code path. So chat.memory.enabled is misleading: it reads as a global kill switch but only suppresses the memory_20250818 transformation in one BYOK provider. And even there it doesn't cleanly disable the memory tool falls through as a generic function tool rather than being filtered out. 
 
The original PR would have broken load-bearing local memory features; the revised version keeps a description that no longer matches the diff and introduces a hidden setting whose name overpromises. Please don't open PRs against this area without confirming the concrete scenario, validating your changes locally at the very least, and ensuring the description, setting name, and diff describe the same change.  If there's a specific BYOK-Anthropic scenario from github/agents#953, please surface it in the PR body and we can discuss the right shape.

PR #316357, opened 2026-05-14
fix: adjust indentation for collapsed rows in chat models widget

AndyXie0718:
## Summary
Fix inconsistent indentation of vendor/group headers in Chat Models widget when expanding/collapsing.

Previously, collapsed groups sometimes caused wrong rows to shift, due to leftover `status-icon` display states from table row recycling. Now, expanded groups show a small indent (via visible status-icon placeholder) while collapsed groups have no extra indent (status-icon hidden), consistent with the intended visual hierarchy.

## Changes
- In `ModelNameColumnRenderer`, explicitly control `.status-icon` display based on `entry.collapsed`:
  - `collapsed === false` → `display: ''` (restore default, icon occupies space → indent)
  - `collapsed === true` → `display: 'none'` (icon hidden → no indent)

## Affected files
- `src/vs/workbench/contrib/chat/browser/chatManagement/chatModelsWidget.ts`


AndyXie0718:
@lramos15

lramos15:
Do you have a photo of what the problem is?

AndyXie0718:
> Do you have a photo of what the problem is?

Sure, when entering the `Chat: Manage Language Models` interface, the model provider "Copilot" have indentations and "OAI Compatible" haven't, which is strange in the first place.
image

When the row "OAI Compatible" is collasped, the indentations in row "Copilot" vanish, and "OAI Compatible" remains the same which I think is inappropriate because they are from different groups, and row "Copilot" wrongly delete the indentations when group "OAI Compatibale" collapsed.

image

AndyXie0718:
@lramos15 These problems could be reproduced on win11 23H2 12th Gen Intel(R) Core(TM) i5-12500H and win10 intel i5-1035g4

AndyXie0718:
@lramos15 Here is how I figured out what's happened:
image
we can see that now the string "Copilot" has an indent, which comes from the element "statusIcon"; when it is displayed by default, ('display': 'block')
image
when we change the display atrribute into 'none', of course the indent vanished.
image
And the problem is that group title "Copilot" wrongly react (hide its indent) when group "OAI Compatible" collapsed, my solution is to check the attribute "entry.collapsed" and change the attribute "display" before rendering any vender or group element, with the core fixing sentence: `templateData.statusIcon.style.display = entry.collapsed ? 'none' : 'block'`

lramos15:
I'm not even sure we need top level status icons. It doesn't make much sense to me

AndyXie0718:
Perhaps you are right about removing top-level status icons, so in the latest commit I've hidden them with `display: 'none'` so they no longer cause indentation issues. I'll consider refactoring when I have more time.

AndyXie0718:
Can you offer some advice on this @sandy081?

AndyXie0718:
Maybe this issue is bigger than we thought, after updating to 1.121, there are similar issues on the deprecation-warning-icon!
When we first enter the `Chat: Manage Language Models`, things all looks fine.
image
But when I fold and unfold the second group `OAI Compatible`, a deprecation-warning-icon appeared on the `OAI Compatible` row, which is rediculous because none of the models in the `OAI Compatible` group are deprecated.
image
When I unfold `OAI Compatible` again, the deprecation-warning-icon went to the first group `Copilot` row.
image

AndyXie0718:
I pushed a new commit to fix the new issue. Table template instances are reused across row types in `ModelNameColumnRenderer`, so codicon classes and `display` state set by a model row can leak into subsequent vendor/group header rows. This commit resets `modelStatusIcon.className` to `'model-status-icon'` and `display` to `'none'` in `renderVendorElement` and `renderGroupElement`, following the same cleanup pattern used in `renderModelElement`.

AndyXie0718:
@justschen sorry to bother you but I believe this issue still exists in the newest 1.122 version
image

PR #316358, opened 2026-05-14
Use AHP ping for active liveness on remote agent host connections

roblourens:
Replaces the previous per-request timeout on remote agent host connections with a periodic ping-based liveness check, modelled after `PersistentProtocol`'s socket keep-alive.

## What

- New AHP `ping` request is issued every 5s by `RemoteAgentHostProtocolClient`.
- If no inbound message has been received for 20s, the client force-closes the connection. `RemoteAgentHostService` then reconnects via its existing exponential backoff.
- Reuses `LoadEstimator` (now exported from `ipc.net.ts`) to suppress the force-close when local CPU is pegged — same heuristic `PersistentProtocol` uses to avoid blaming the remote for our own JS event-loop stalls. Also handles laptop sleep/wake correctly: the timer fires once after wake but the wall-clock comparison stays accurate.
- The old per-request timeout (originally added to detect a disconnected remote) is removed.

## Why

The per-request timeout only fired while a request was in flight, so a connection that went idle and silently dropped wouldn't be noticed. An active ping covers both:
- **Keep-alive** — keeps idle-timeout intermediaries (proxies, load balancers) from killing the connection.
- **Liveness** — a hung AH process won't run its dispatch loop to answer, so a healthy TCP socket with a frozen peer trips the timeout. A silently-dead transport trips the same way. Either way: drop and reconnect.

## Notes

- The protocol-side `ping` command is in [microsoft/agent-host-protocol#119](https://github.com/microsoft/agent-host-protocol/pull/119). This PR depends on that landing + the next sync into vscode.
- Servers without `ping` support are tolerated: the `MethodNotFound` error response is itself inbound traffic, so liveness is satisfied without any client-side branching.
- Behavior against a permanently-hung AH: reconnect attempts back off to 30s and continue indefinitely (no grace deadline, unlike remote EH's 3-hour cap). Could revisit if it becomes a problem in the wild.

(Written by Copilot)


src/vs/platform/agentHost/browser/remoteAgentHostProtocolClient.ts
```diff
@@ -1137,13 +1160,32 @@ export class RemoteAgentHostProtocolClient extends Disposable implements IAgentC
...
 	private _watchdogTick(): void {
-		if (this._state.kind === AgentHostClientState.Closed || this._pendingRequests.size === 0) {
+		if (this._state.kind === AgentHostClientState.Closed) {
+			return;
+		}
+		if (this._pendingRequests.size === 0) {
+			// Active liveness probe — keep idle connections under
+			// supervision. Fire-and-forget; the next tick checks for
+			// staleness via the normal pending-request path.
+			void this.ping().catch(() => undefined);
```
connor4312, line 1188: You could defer this as long as any message is being received from the remote (have ping on a timer, rather than an interval, reset it whenever a message is received. Have the timeout on a timer that's the ping interval + the timeout interval and reset it similarly.) A little less traffic than sending pings to something we already know is live.

connor4312 (approved):

PR #316392, opened 2026-05-14
fix: preserve shell-special characters when dropping files into terminal

### LINKED ISSUE #276999: Dragging-n-dropping a file with tilde in the name into terminal

- VS Code Version: 1.105.1
- OS Version: Arch Linux

Steps to Reproduce:

1. Create a file with one or more tilde ("~") character(s) in its name
2. In the EXPLORER view navigate to that file
3. Open TERMINAL tab
4. Drag and drop the file with tilde in its name from the EXPLORER to the TERMINAL

Expected behavior:
File path should appear copied into the terminal

Actual behavior:
Tilde character(s) gets silently dropped in the copied/pasted path. This results into an invalid path pointing to a non-existent file.

EDIT.

Tilde turns out to be not the only character that gets silently dropped this way.
This issue is reproducible whenever at least one of the following characters is present in the file name:

```
~ - Tilde
` - Backtick
! - Exclamation
# - Hash / Number Sign
$ - Dollar Sign
^ - Caret or Circumflex
& - Ampersand
* - Asterisk
| - Vertical Bar or Pipe
; - Semicolon
< - Less-Than Sign
> - Greater-Than Sign
```

Priyanshu-byte-coder:
Fixes #276999

## Problem

When dragging a file from the Explorer into the Terminal, any shell-special characters in the file name are **silently removed** before the path is inserted. The following characters were stripped:

`` ~ ` ! # $ ^ & * | ; < > ``

This produces a **different, possibly non-existent path** — for example, a file named `/home/user/project/maintenance~1/script.sh` is dropped as `/home/user/project/maintenance1/script.sh`, pointing to an entirely different file. Users running `rm` or other destructive commands against the wrong path may not notice the discrepancy in a long path.

Root cause: `escapeNonWindowsPath` in `terminalEnvironment.ts` applies a `bannedChars` regex that removes these characters before the path is quote-wrapped.

## Fix

Remove the `bannedChars` removal step. The quote-wrapping logic that follows (`'...'` single-quote quoting for bash/sh/zsh/fish/git-bash, `$'...'` ANSI-C quoting for paths containing single quotes, and PowerShell equivalents) already neutralises all of these characters:

- Inside single quotes, `$`, `` ` ``, `&`, `|`, `;`, `<`, `>`, `~`, `#`, `!`, `^`, `*` are all **literal** — the shell never interprets them as metacharacters.
- Paths containing single quotes use `$'...\''...'` or `'\''` escaping, which correctly handles every combination.

Injection is still prevented: `/foo/file$(echo evil)` becomes `'/foo/file$(echo evil)'`, where `$(...)` is never evaluated by the shell.

## Changes

| File | Change |
|---|---|
| `src/vs/platform/terminal/common/terminalEnvironment.ts` | Remove `bannedChars` strip; update JSDoc |
| `src/vs/platform/terminal/test/common/terminalEnvironment.test.ts` | Update test to assert special chars are preserved inside quotes; add cases for `~`, `#`, `&`, `;` |

## Testing

1. Create files with names containing `~`, `#`, `$`, `&`, `!`, `` ` `` etc.
2. Drag each file from the Explorer into an open Terminal.
3. Verify the full path including the special character appears in the terminal prompt.

### DIFF
--- a/src/vs/platform/terminal/common/terminalEnvironment.ts
+++ b/src/vs/platform/terminal/common/terminalEnvironment.ts
@@ -7,9 +7,11 @@ import { OperatingSystem, OS } from '../../../base/common/platform.js';
 import { IShellLaunchConfig, TerminalShellType, PosixShellType, WindowsShellType, GeneralShellType } from './terminal.js';
 
 /**
- * Aggressively escape non-windows paths to prepare for being sent to a shell. This will do some
- * escaping inaccurately to be careful about possible script injection via the file path. For
- * example, we're trying to prevent this sort of attack: `/foo/file$(echo evil)`.
+ * Escape non-windows paths to prepare for being sent to a shell. The path is wrapped in
+ * appropriate shell quotes so that shell-special characters in file names (e.g. `~`, `#`, `$`,
+ * `&`, `|`, `;`, etc.) are treated as literals rather than being silently dropped. Quote wrapping
+ * also prevents script injection, e.g. a path like `/foo/file$(echo evil)` becomes the safe
+ * string `'/foo/file$(echo evil)'` where `$(…)` is never evaluated by the shell.
  */
 export function escapeNonWindowsPath(path: string, shellType?: TerminalShellType): string {
 	let newPath = path;
@@ -64,10 +66,6 @@ export function escapeNonWindowsPath(path: string, shellType?: TerminalShellType
 			};
 			break;
 	}
-
-	// Remove dangerous characters except single and double quotes, which we'll escape properly
-	const bannedChars = /[\`\$\|\&\>\~\#\!\^\*\;\<]/g;
-	newPath = newPath.replace(bannedChars, '');
 
 	// Apply shell-specific escaping based on quote content
 	if (newPath.includes('\'') && newPath.includes('"')) {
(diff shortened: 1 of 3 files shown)


Priyanshu-byte-coder:

Priyanshu-byte-coder:
Thanks for the catch. Updated `src/vs/workbench/contrib/terminal/test/common/terminalEnvironment.test.ts` — all 8 `preparePathForShell` test cases that expected the stripped output (e.g. `'/foo/bar(echo evil)baz'`) now expect the preserved form (`'/foo/bar$(echo evil)baz'`). CMD and PowerShell cases already had the `$` in their expected strings since they go through separate handlers that never applied the banned-chars strip.

Priyanshu-byte-coder:
@Tyriar pinging you as terminal owner. Copilot flagged (as a suppressed low-confidence comment) that `src/vs/workbench/contrib/terminal/test/common/terminalEnvironment.test.ts` also needed updating — that's already done in this PR (+8/-8 in that file), all 8 `preparePathForShell` cases updated to expect the preserved form.

The fix itself: remove the `bannedChars` strip from `escapeNonWindowsPath`. The quote-wrapping that follows (`'...'` single-quote, `$'...'` ANSI-C for paths with single quotes) already neutralises all the special characters — removing them beforehand was just silently corrupting paths. Injection prevention is unchanged: `'/foo/file$(echo evil)'` is never evaluated. Could you review when you get a chance?

Priyanshu-byte-coder:
Monthly follow-up @Tyriar — this null-guard fix for terminal environment variable collection is still ready for review. Happy to rebase or adjust if anything's changed on main.

PR #316395, opened 2026-05-14
scm: allow diffDecorationsGutterWidth up to 10px

Priyanshu-byte-coder:
Closes #278083

## Problem

`scm.diffDecorationsGutterWidth` is constrained to the values `[1, 2, 3, 4, 5]` via a JSON Schema `enum`. This means:

- The Settings UI renders a **dropdown** instead of a number input, which feels wrong for a width value.
- Users who want a wider gutter (e.g. on HiDPI / high-resolution displays) are blocked at 5 px.

## Fix

Replace the `enum` array with `minimum: 1` / `maximum: 10` so the Settings UI shows a number input and any integer in that range is valid.

Also raise the runtime safety guard in `quickDiffDecorator.ts` from `width > 5` to `width > 10` to stay consistent — values outside `[1, 10]` still fall back to the default of `3`.

## Changes

| File | Change |
|---|---|
| `src/vs/workbench/contrib/scm/browser/scm.contribution.ts` | `enum: [1,2,3,4,5]` → `minimum: 1, maximum: 10` |
| `src/vs/workbench/contrib/scm/browser/quickDiffDecorator.ts` | `width > 5` → `width > 10` |

### DIFF
--- a/src/vs/workbench/contrib/scm/browser/quickDiffDecorator.ts
+++ b/src/vs/workbench/contrib/scm/browser/quickDiffDecorator.ts
@@ -274,7 +274,7 @@ export class QuickDiffWorkbenchController extends Disposable implements IWorkben
 	private onDidChangeDiffWidthConfiguration(): void {
 		let width = this.configurationService.getValue<number>('scm.diffDecorationsGutterWidth');
 
-		if (isNaN(width) || width <= 0 || width > 5) {
+		if (isNaN(width) || width <= 0 || width > 10) {
 			width = 3;
 		}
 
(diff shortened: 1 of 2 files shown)


Priyanshu-byte-coder:
@lszomoru you were notified via CODENOTIFY for the files changed here. Copilot reviewed both files and raised no issues. The change is small:

- `scm.contribution.ts`: swap `enum: [1,2,3,4,5]` → `minimum: 1, maximum: 10` so the Settings UI renders a number input instead of a dropdown, and HiDPI users can set wider gutter widths
- `quickDiffDecorator.ts`: bump the runtime guard from `width > 5` to `width > 10` to stay consistent

Could you take a look when you get a chance?

Priyanshu-byte-coder:
Monthly follow-up @lszomoru — small settings-schema fix (enum → minimum/maximum) still ready for review.

PR #316396, opened 2026-05-14
terminal: clip tab description text when panel is in text mode

### LINKED ISSUE #312970: terminal progress display overflow
Image

When I move the terminal to right side and the progress content is long.

Priyanshu-byte-coder:
Closes #312970

## Problem

When the terminal tabs panel is in "text" mode (wide enough to show labels), the tab description (e.g. the current working directory or process name) can be wider than its container. This causes the text to overflow and overlap other elements in the panel.

The `not(.has-text)` case already hides the description with `display: none`, but there is no corresponding CSS rule that constrains the description when `has-text` **is** present.

## Root cause

`.monaco-icon-label` is a flex child without `min-width: 0`, so it cannot shrink below its intrinsic size. The description container has no overflow constraint, so text spills out of the flex row.

## Fix

Add two CSS rules to `terminal.css`:

1. `min-width: 0` on `.monaco-icon-label` inside a `has-text` terminal entry — lets the flex item actually shrink.
2. `overflow: hidden; text-overflow: ellipsis; white-space: nowrap` on `.monaco-icon-description-container` in `has-text` mode — clips and shows an ellipsis for long descriptions.

## Changes

`src/vs/workbench/contrib/terminal/browser/media/terminal.css` — two new CSS rule blocks appended after the existing `not(.has-text)` description rule.

### DIFF
--- a/src/vs/workbench/contrib/terminal/browser/media/terminal.css
+++ b/src/vs/workbench/contrib/terminal/browser/media/terminal.css
@@ -438,6 +438,16 @@
 	display: none;
 }
 
+.monaco-workbench .pane-body.integrated-terminal .tabs-container.has-text .tabs-list .terminal-tabs-entry .monaco-icon-label {
+	min-width: 0;
+}
+
+.monaco-workbench .pane-body.integrated-terminal .tabs-container.has-text .tabs-list .terminal-tabs-entry .monaco-icon-description-container {
+	overflow: hidden;
+	text-overflow: ellipsis;
+	white-space: nowrap;
+}
+
 .monaco-workbench .pane-body.integrated-terminal .tabs-list .codicon {
 	vertical-align: text-bottom;
 }


Priyanshu-byte-coder:
@Tyriar this is a small CSS-only fix for terminal tab description overflow in text mode. Two rules added to `terminal.css`:

1. `min-width: 0` on `.monaco-icon-label` inside `.has-text` — lets the flex item shrink
2. `overflow: hidden; text-overflow: ellipsis; white-space: nowrap` on `.monaco-icon-description-container` in `.has-text` mode — clips long descriptions

The `not(.has-text)` case already uses `display: none`. This just handles the missing `has-text` side. No JS or test changes. Could you take a look?

Priyanshu-byte-coder:
Monthly follow-up @Tyriar — CSS-only fix for terminal tab description overflow, still ready for review.

PR #316473, opened 2026-05-14
Browser: Playwright API metrics

jruales:
Playwright API metrics for Integrated Browser


src/vs/workbench/contrib/browserView/electron-browser/tools/runPlaywrightCodeTool.ts
```diff
@@ -107,10 +151,83 @@ export class RunPlaywrightCodeTool implements IToolImpl {
...
+		// Skip in-progress deferred runs so each user-visible invocation produces
+		// at most one event (we'll log when the resumed call eventually settles).
+		if (result?.deferredResultId) {
+			return;
+		}
+		const pageMethodsCalled = result?.pageMethodsCalled ?? {};
+		const entries = Object.entries(pageMethodsCalled);
+		const total = entries.reduce((sum, [, count]) => sum + count, 0);
+		this.telemetryService.publicLog2<RunPlaywrightCodeEvent, RunPlaywrightCodeClassification>(
+			'integratedBrowser.tools.runPlaywrightCode.completed',
```
kycutler: Wonder if we should log this from the shared service instead -- so if a call is deferred but the agent never queries for the result, it still gets logged eventually

Yoyokrazy (approved):

kycutler (approved):

PR #316475, opened 2026-05-14
chat agent host: seed AutoApprove in _invokeAgent fallback

meganrogge:
When a chat session is created via `workbench.action.chat.open` (e.g. a headless caller driving a chat from outside the picker — including the eval driver), the resulting session resource does not go through `AgentHostUntitledProvisionalSessionService.getOrCreate`, so the workbench-side initial config (notably `autoApprove` seeded from `chat.permissions.default`) is never sent at `createSession` time. The SDK then defaults to `interactive` mode and prompts on bash / `ask_user` / elicitations, which never resolve in headless contexts.

## Repro

Eval driver flow:
1. `workbench.action.chat.openNewSessionSidebar.<provider>` creates session A via the picker → `getOrCreate` seeds `AutoApprove='autopilot'` from `chat.permissions.default`. ✅
2. `workbench.action.chat.open` with the query then creates session B (different resource), routes through `AgentHostSessionHandler._invokeAgent` → `!existingState` branch → `_createAndSubscribe(...request.agentHostSessionConfig)`, which never read `chat.permissions.default`. ❌
3. `_resolveSdkMode` sees `autoApprove='default'` → `mode=interactive` → run hangs on the first elicitation (`X_BLOCKING_UI_ERROR` after timeout).

Confirmed in run `25877212964` instance `say_hello` (`agenthost.log`): two provisional sessions back-to-back, only the picker one received the seed.

## Change

Mirror `AgentHostUntitledProvisionalSessionService._getInitialConfig` in `_invokeAgent`'s `!existingState` fallback so the same workbench defaults (`isolation='folder'` + `autoApprove` from `chat.permissions.default`, policy-clamped) are merged into the config passed to `_createAndSubscribe`. `request.agentHostSessionConfig` values win over the workbench defaults.

The Agents window path is unchanged: `isSessionsWindow` short-circuits the seed, and the sessions provider continues to supply config via `request.agentHostSessionConfig`.

## Tests

- `./scripts/test.sh --grep 'AgentHostSessionHandler|AgentHostClientTools|AgentHost'` → 705 passing.
- Stubbed `IWorkbenchEnvironmentService` in `agentHostClientTools.test.ts` to match the new constructor dependency.

### DIFF
--- a/.vscode/notebooks/my-endgame.github-issues
+++ b/.vscode/notebooks/my-endgame.github-issues
@@ -7,7 +7,7 @@
   {
     "kind": 2,
     "language": "github-issues",
-    "value": "$MILESTONE=milestone:\"1.120.0\"\n\n$MINE=assignee:@me"
+    "value": "$MILESTONE=milestone:\"1.121.0\"\n\n$MINE=assignee:@me"
   },
   {
     "kind": 2,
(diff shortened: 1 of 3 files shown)


Yoyokrazy (approved):

meganrogge:
Closing as @roblourens is looking into this

PR #316497, opened 2026-05-14
Tag extension telemetry with common.isAgentsWindow

osortega:


rzhao271 (approved):

bryanchen-d (commented):
Thanks! Is this the right property to filter events belong to the agents window?

PR #316500, opened 2026-05-14
Remove compression banners to prevent model refetch - rm grep/tree/find filters for now

meganrogge:
## Summary

Code hygiene and correctness improvements to terminal output compression. Eval results show no measurable impact on token cost (Δ = −5.1%, 95% CI [−45%, +35%], not significant at n=10 seeds).

### Changes

1. **Remove compression banners** — the banner text (`[Output compressed by ... To disable, set ...]`) is noise in tool output that the model does not need. Compressed output is now returned silently, matching [ztk](https://github.com/codejunkie99/ztk)'s approach.

2. **Remove find/grep/tree filters** — these filters naively cap output at 200 lines, silently dropping results past that limit. With silent compression, the model has no way to know results were lost, which is a **correctness problem** — the agent could miss matches or files it needs. Tracked in #316502 to reintroduce with smart summarization (grouping by directory/file like ztk, instead of line capping).

3. **Simplify cache dedup message** — show a preview of the cached output (up to 200 chars) + `(unchanged)` instead of a timestamp and settings reference, modeled after ztk's `computeDelta` approach. Gives the model context without wasting tokens on metadata.

### Files

- `toolResultCompressorService.ts`: Remove banner prepending, update `_buildCacheHitResult` to preview + `(unchanged)`
- `toolResultCompressor.ts`: Remove `formatCompressionBanner` function
- `terminalOutputCompressor.ts`: Unregister `findFilter`, `grepFilter`, `treeFilter`
- `toolResultCompressorService.test.ts`: Update tests to match bannerless output

### DIFF
--- a/src/vs/workbench/contrib/chat/browser/tools/toolResultCompressorService.ts
+++ b/src/vs/workbench/contrib/chat/browser/tools/toolResultCompressorService.ts
@@ -177,13 +177,19 @@ export class ToolResultCompressorService extends Disposable implements IToolResu
 
 			totalAfter += current.length;
 			if (current !== original) {
-				anyCompressed = true;
 				// Prepend a banner so the model knows the output was filtered, by
 				// which filters, and how to disable compression. We only annotate
 				// the parts we actually changed — non-compressed parts pass through
 				// untouched.
 				const banner = formatCompressionBanner(partFilterIds, original.length, current.length);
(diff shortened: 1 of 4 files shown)


meganrogge:

meganrogge:
This is not substantiated by data from evals 🤔

PR #316508, opened 2026-05-14
Add Mocha retries to sanity test runner

benvillalobos:
Enables Mocha test-level retries for the sanity test runner.

## Change
- `test/sanity/src/index.ts` — Mocha `retries: 3` for in-process retry on flaky sanity tests.

## Context
Analysis of ~400 recent builds showed 0 of 10 sanity failures had ever been retried. Adding Mocha retries lets individual tests recover from transient failures within the same process without a full job restart.

### DIFF
--- a/build/azure-pipelines/common/sanity-tests.yml
+++ b/build/azure-pipelines/common/sanity-tests.yml
@@ -133,6 +133,7 @@ jobs:
         - script: $(TEST_DIR)/scripts/run-win32.cmd -c $(BUILD_COMMIT) -q $(BUILD_QUALITY) -t $(LOG_FILE) -s $(SCREENSHOTS_DIR) -v ${{ parameters.args }}
           workingDirectory: $(TEST_DIR)
           displayName: Run Sanity Tests
+          retryCountOnTaskFailure: 3
           env:
             GITHUB_ACCOUNT: $(sanity-tests-account)
             GITHUB_PASSWORD: $(sanity-tests-password)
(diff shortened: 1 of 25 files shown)


build/azure-pipelines/common/sanity-tests.yml
```diff
@@ -176,6 +179,7 @@ jobs:
               ${{ parameters.args }}
           workingDirectory: $(TEST_DIR)
           displayName: Run Sanity Tests
+          retryCountOnTaskFailure: 3
```
dmitrivMS: My suggestion is to keep `retries: 3` in test/sanity/src/index.ts and remove from here.
As far as I can tell sanity tests fail for retriable reasons on the inside steps, not the pipeline side.

test/sanity/src/index.ts
```diff
@@ -41,6 +41,7 @@ const mochaOptions: MochaOptions = {
 	slow: 3 * 60 * 1000,
 	grep: options.grep,
 	fgrep: options.fgrep,
+	retries: 3,
```
dmitrivMS, line 44: Not a big fan, but it is what it is - we need them stable. Would also like to keep fixing individual issues as we find them.

build/azure-pipelines/copilot/test-integration-steps.yml
```diff
@@ -69,6 +69,7 @@ steps:
       workingDirectory: $(Build.SourcesDirectory)/extensions/copilot
       displayName: 🧪 Run Copilot sanity tests
       timeoutInMinutes: 15
+      retryCountOnTaskFailure: 3
```
dmitrivMS: I don't own these, someone else should comment.

dmitrivMS (commented):
Added some comments.

benvillalobos:
Scoped down to the mocha retry of 3

bryanchen-d (approved):

dmitrivMS (approved):

PR #316520, opened 2026-05-14
Add completions service and slash command support for Copilot CLI

DonJayamanne:
- Introduced IAgentHostCompletions to expose completions service in AgentService.
- Implemented CopilotSlashCommandCompletionProvider for handling slash commands.
- Enhanced chat input completion handling to support command and skill attachments.
- Updated tests for new completions and command parsing functionality.

### DIFF
--- a/src/vs/workbench/contrib/chat/common/chatSessionsService.ts
+++ b/src/vs/workbench/contrib/chat/common/chatSessionsService.ts
@@ -330,14 +330,14 @@ export interface IChatInputCompletionItem {
 	readonly start?: IPosition;
 	readonly end?: IPosition;
 	/** Attachment associated with the item. */
-	readonly attachment: IChatInputCompletionAttachment;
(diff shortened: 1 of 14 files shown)


src/vs/workbench/contrib/chat/common/chatSessionsService.ts
```diff
@@ -350,6 +350,24 @@ export interface IChatInputCompletionAttachment {
...
+}
+
+/**
+ * Skill attachment associated with a completion item. The workbench
+ * adds it to the input's variable model when the item is accepted.
+ */
+export interface IChatInputCompletionSkillAttachment {
+	readonly kind: 'skill';
+	readonly uri: URI;
+}
```
connor4312, line 381: We should make sure we propagate through the `_meta` still in both these cases. In the protocol we say clients "MUST" retain that so agent hosts can use that to ferry any contextual information they need.

connor4312 (commented):
one comment but lgtm after, feel free to put up for general review afterwards

connor4312 (approved):

connor4312 (approved):

PR #316534, opened 2026-05-15
Add support for utility model overrides in chat

vritant24:
Introduce new settings for `chat.utilityModel` and `chat.utilitySmallModel` to allow users to override the language models used by built-in utility flows.


extensions/copilot/src/extension/prompt/vscode-node/endpointProviderImpl.ts
```diff
@@ -99,8 +132,13 @@ export class ProductionEndpointProvider extends Disposable implements IEndpointP
 	 * selection for each family lives in the corresponding resolver
 	 * class so callers don't need to know which CAPI family backs each
 	 * purpose.
+
 	 */
-	private _resolveUtilityFamily(family: ChatEndpointFamily): Promise<IChatEndpoint> {
+	private async _resolveUtilityFamily(family: ChatEndpointFamily): Promise<IChatEndpoint> {
+		const override = await this._resolveUtilityOverride(family);
+		if (override) {
+			return override;
```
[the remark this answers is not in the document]
vritant24: by design at the moment

extensions/copilot/src/extension/prompt/vscode-node/endpointProviderImpl.ts
```diff
@@ -50,8 +52,39 @@ export class ProductionEndpointProvider extends Disposable implements IEndpointP
...
 		}));
+
+		// When the user changes their utility model overrides we need to invalidate any
+		// previously-resolved utility alias endpoints so the next request re-resolves.
+		this._register(this._configService.onDidChangeConfiguration(e => {
+			if (e.affectsConfiguration(ProductionEndpointProvider.UTILITY_MODEL_CONFIG_KEY) || e.affectsConfiguration(ProductionEndpointProvider.UTILITY_SMALL_MODEL_CONFIG_KEY)) {
+				this._logService.trace(`[ProductionEndpointProvider] Utility model override changed; invalidating alias endpoints.`);
+				// Clear telemetry fingerprints so a re-applied override emits
+				// once for its new value.
+				this._lastOverrideTelemetryFingerprint.clear();
```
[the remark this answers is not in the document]
vritant24: leaving it as is

extensions/copilot/src/extension/conversation/vscode-node/languageModelAccess.ts
```diff
@@ -158,6 +159,59 @@ function buildConfigurationSchema(endpoint: IChatEndpoint): { configurationSchem
...
+			name: endpoint.name,
+			family,
+			version: endpoint.version,
+			maxInputTokens: endpoint.modelMaxPromptTokens - baseCount - BaseTokensPerCompletion,
+			maxOutputTokens: endpoint.maxOutputTokens,
+			requiresAuthorization,
+			isUserSelectable: false,
+			isDefault: false,
+			capabilities: {
+				toolCalling: endpoint.supportsToolCalls,
```
dmitrivMS, line 208: I remember a PR where we broke due to missing tokenizer - is it similar place? Should it be passed?

extensions/copilot/src/extension/conversation/vscode-node/languageModelAccess.ts
```diff
@@ -158,6 +159,59 @@ function buildConfigurationSchema(endpoint: IChatEndpoint): { configurationSchem
...
+	models: readonly vscode.LanguageModelChatInformation[],
+	baseCount: number,
+	requiresAuthorization: vscode.LanguageModelChatInformation['requiresAuthorization'],
+): { info: vscode.LanguageModelChatInformation; synthesized: boolean } {
+	const base = endpoint instanceof CopilotChatEndpoint ? models.find(m => m.id === endpoint.model) : undefined;
+	if (base) {
+		return {
+			synthesized: false,
+			info: {
+				...base,
```
dmitrivMS, line 187: What happens if base model changes - we clone stuff here, does it get updated eventually?

roblourens (dismissed):

src/vs/workbench/contrib/chat/browser/defaultModelContribution.ts
```diff
@@ -27,20 +27,65 @@ export interface DefaultModelContributionOptions {
...
+	/**
+	 * When `true`, eagerly resolve every visible vendor at construction time
+	 * and on `onDidChangeLanguageModelVendors`. This ensures models from
+	 * visible providers that no other consumer has touched yet (e.g. BYOK)
+	 * surface in the picker at the cost of activating those provider extensions.
+	 *
+	 * Defaults to `false` so contributions that run during startup
+	 * (`BlockRestore`) don't pay that activation cost.
+	 */
+	readonly eagerVendorResolution?: boolean;
```
dmitrivMS: This is scary... when do we turn this forced resolution on? I ran into this in my work as well and I to work around to avoid forcing all vendors to load on startup.

src/vs/workbench/contrib/chat/browser/defaultModelContribution.ts
```diff
@@ -55,12 +100,56 @@ export abstract class DefaultModelContribution extends Disposable {
...
+	}
+
+	private async _resolveAllVendorsUntilSettled(service: ILanguageModelsService, state: ResolveAllVendorsState): Promise<void> {
+		try {
+			do {
+				state.needsAnotherPass = false;
+				const vendors = service.getVendors().map(v => v.vendor);
+				this._logService.trace(`${this._options.logPrefix} Resolving all vendors: [${vendors.join(', ')}]`);
+				try {
+					for (const vendor of vendors) {
```
dmitrivMS: What's the case we need to resolve all vendors?

dmitrivMS (dismissed):
Not an expert on this area, so my only real concern is avoiding eager loading of all vendors, especially in startup/restore scenarios, which may be a perf hit.

extensions/copilot/src/extension/conversation/vscode-node/languageModelAccess.ts
vritant24: This won't be a problem here. 
The requests go through `_getEndpointForModel` below in the file, that actually gets the endpoint for the model, which has the tokenizer and works as it does today

extensions/copilot/src/extension/conversation/vscode-node/languageModelAccess.ts
vritant24: Yes. the model change events are listened for and notified [here](https://github.com/microsoft/vscode/blob/c08d36bccf86b6105cf2be3971852012ba0a61dd/extensions/copilot/src/extension/conversation/vscode-node/languageModelAccess.ts#L284-L285) that causes an update

src/vs/workbench/contrib/chat/browser/defaultModelContribution.ts
vritant24: It's t show visible BYOK/provider models even when no other chat flow has queried that vendor yet. 
Without resolving visible vendors, those models may not be present in the language-model cache when the settings schema is populated

src/vs/workbench/contrib/chat/browser/defaultModelContribution.ts
vritant24: Good point. I removed the eager vendor-resolution path

dmitrivMS (approved):

PR #316661, opened 2026-05-15
feat: update to Electron v42

### LINKED ISSUE #292445: Electron 42 Update
Electron 40 was released on January 13, 2026.

Ref: https://www.electronjs.org/blog/electron-40-0

deepak1556:
Fixes https://github.com/microsoft/vscode/issues/292445


src/vs/workbench/api/node/extHostMcpNode.ts
```diff
@@ -178,7 +178,7 @@ class McpHTTPHandleNode extends McpHTTPHandle {
 		return {
 			status: undiciResponse.status,
 			statusText: undiciResponse.statusText,
-			headers: undiciResponse.headers,
+			headers: undiciResponse.headers as unknown as Headers,
```
rzhao271: undici debt or incompatibility issue?

build/.npmrc
```diff
@@ -1,5 +1,6 @@
 disturl="https://nodejs.org/dist"
 runtime="node"
+target="22.22.1"
```
rzhao271, line 3: Should this be 24.x as well?

Yoyokrazy (dismissed):

src/vs/platform/browserView/electron-main/browserView.ts
```diff
@@ -547,8 +546,10 @@ export class BrowserView extends Disposable {
 	/**
 	 * Load a URL in this view
 	 */
-	async loadURL(url: string): Promise<void> {
-		await this._view.webContents.loadURL(url);
+	loadURL(url: string): Promise<void> {
+		const p = this._view.webContents.loadURL(url);
+		p.catch(() => { });
```
kycutler: Why do we need this? If it's about unhandled rejections I'd rather do this from the caller rather than centrally here

test/sanity/src/context.ts
```diff
@@ -1318,4 +1318,9 @@ export class TestContext {
 			this.killProcessTree(app.pid!);
 		}
 	}
+
+	private isNonFatalCliStderr(text: string): boolean {
+		return /ECONNRESET|ECONNABORTED|ECANCELED|EPIPE|SIGPIPE/.test(text)
+			|| /(^|\n)(?:\[[^\]]+\]\s*)?(?:\(node:\d+\)\s*)?(?:\[[A-Z0-9]+\]\s*)?(?:[A-Za-z]+Warning|Warning):/.test(text);
```
dmitrivMS, line 1389: what's an example of such warnings? do they also show up in the product?

test/sanity/src/wsl.test.ts
```diff
@@ -138,6 +147,65 @@ export function setup(context: TestContext) {
...
+
+		const wslNodePath = wslEntryPoint.replace(/\/bin\/[^/]+$/, '/node');
+		if (patchedWslNodePaths.has(wslNodePath)) {
+			return;
+		}
+
+		patchedWslNodePaths.add(wslNodePath);
+		context.warn(`Applying WSL1 Node 24 workaround for ${wslNodePath}`);
+
+		const shellScript = [
```
dmitrivMS: It's fine for the tests, but would customers have to do the same?

test/sanity/src/wsl.test.ts
```diff
@@ -138,6 +147,65 @@ export function setup(context: TestContext) {
...
+			`node_path='${wslNodePath}'`,
+			'backup_path="${node_path}.orig"',
+			'if [ -f "${backup_path}" ]; then exit 0; fi',
+			'if ! command -v objcopy >/dev/null 2>&1; then apt-get update && apt-get install -y binutils; fi',
+			'cp "${node_path}" "${backup_path}"',
+			'objcopy --remove-section .note.ABI-tag --remove-section .note.gnu.build-id --remove-section .note.gnu.property "${backup_path}" "${node_path}"',
+			'chmod +x "${node_path}"',
+		].join('; ');
+
+		try {
```
dmitrivMS: Should we just let it throw here? Won't following tests fail if this fails?

test/sanity/src/wsl.test.ts
```diff
@@ -138,6 +147,65 @@ export function setup(context: TestContext) {
 		);
 	}
 
+	function getUbuntuWslVersion(): number | undefined {
```
dmitrivMS: nit: I would prefer these two to go into context.ts, but I can do that later

test/sanity/scripts/run-win32.cmd
```diff
@@ -16,6 +16,15 @@ if errorlevel 1 (
     echo Checking if Ubuntu is available on WSL
     powershell -NoProfile -Command "if ((wsl -l -q) -contains 'Ubuntu') { exit 0 } else { exit 1 }"
     if errorlevel 1 call :install_ubuntu
+
+    echo Ensuring Ubuntu runs with WSL2
+    wsl --set-version Ubuntu 2 >nul 2>nul
+    if errorlevel 1 (
+        echo WARNING: Failed to switch Ubuntu to WSL2. WSL1 fallback workaround will be used for Node 24 server binaries.
```
dmitrivMS, line 23: Should we just let it fail instead of warn?

src/vs/platform/agentHost/test/node/agentHostGitService.integrationTest.ts
```diff
@@ -54,7 +54,7 @@ suite('AgentHostGitService - getSessionGitState (real git)', () => {
 
 	teardown(() => {
 		if (tmpRoot) {
-			rmSync(tmpRoot, { recursive: true, force: true });
+			try { rmSync(tmpRoot, { recursive: true, force: true, maxRetries: 10, retryDelay: 200 }); } catch { /* best-effort temp cleanup; Windows can briefly hold git handles */ }
```
dmitrivMS: nit: this could be a helper method

src/vs/base/common/map.ts
```diff
@@ -340,7 +340,7 @@ export class LinkedMap<K, V> implements Map<K, V> {
 		return item.value;
 	}
 
-	forEach(callbackfn: (value: V, key: K, map: LinkedMap<K, V>) => void, thisArg?: unknown): void {
```
dmitrivMS, line 343: not familiar with these, but is it changing the behavior going from LinkedMap to Map?

extensions/git/src/cache.ts
```diff
@@ -132,7 +132,7 @@ class LinkedMap<K, V> implements Map<K, V> {
 		return item.value;
 	}
 
-	forEach(callbackfn: (value: V, key: K, map: LinkedMap<K, V>) => void, thisArg?: unknown): void {
```
dmitrivMS, line 135: is this a change in behavior going LinkedMap -> Map?

dmitrivMS:
Very cool! Left a few comments, nothing critical @deepak1556

build/.npmrc
deepak1556: Nope this has to be 22.x, check the comment in build/npm/postinstall.ts for context. It will be changed when tree-sitter releases the version compatible for 24.x

test/sanity/src/context.ts
deepak1556:
```
[2026-05-26T16:46:32.753Z] [Server] (node:6304) [DEP0169] DeprecationWarning: `url.parse()` behavior is not standardized and prone to errors that have security implications. Use the WHATWG URL API instead. CVEs are not issued for `url.parse()` vulnerabilities.
(Use `node --trace-deprecation ...` to show where the warning was created)
```

The warning comes from a dependency and a couple of usages in server code which we can address as a followup, it shouldn't fail the CI in any way.

build/.npmrc
deepak1556: Tracking issue https://github.com/microsoft/vscode/issues/318492

test/sanity/src/wsl.test.ts
deepak1556: Thats something I plan to followup https://github.com/microsoft/vscode/issues/318494

test/sanity/src/wsl.test.ts
deepak1556: Its fine either way, it will abort quite early when the kernel loads the binary with user facing `Exec format error`.

deepak1556: I will leave it to you when the function is moved into context.ts

test/sanity/scripts/run-win32.cmd
deepak1556: Why fail, the fallback works. Our current CI runner https://monacotools.visualstudio.com/Monaco/_build/results?buildId=442158&view=logs&j=ad61475b-4e2f-5677-16be-bf03675b3fc6&t=ffb10239-2d0c-564c-5769-38693f147134&l=696 needs work to support WSL2 so its actually the fallback thats helping.

src/vs/workbench/api/node/extHostMcpNode.ts
deepak1556: incompatibility, undici will eventually catch up. Adding a comment for now.

src/vs/base/common/map.ts
deepak1556: No its a pure type level change and aligns with how `ResourceMap` overrides, Node v24 bumps typescript version that supports native iterator helpers, Symbol.dispose so we need to narrow the types accordingly.

```
interface Map<K, V> {
    keys(): IterableIterator<K>;
    values(): IterableIterator<V>;
    entries(): IterableIterator<[K, V]>;
}
```

After

```
interface MapIterator<T> extends IteratorObject<…> { … }
interface Map<K, V> {
    keys(): MapIterator<K>;
    values(): MapIterator<V>;
    entries(): MapIterator<[K, V]>;
}
```

cc @mjbvz thoughts on the change ?

src/vs/platform/browserView/electron-main/browserView.ts
deepak1556: Yeah you have a caller that would trip the re-entrancy protection in https://github.com/electron/electron/commit/afb0ee418cd4bf813af040467bef15b5422c6ee3 added this as a temporary change, agree that this should be scoped.

src/vs/platform/agentHost/test/node/agentHostGitService.integrationTest.ts
deepak1556: https://github.com/microsoft/vscode/pull/316661/commits/39617bbcf82ab59019a3ba454a89dfaa9ed010c5

test/sanity/src/wsl.test.ts
deepak1556: https://github.com/microsoft/vscode/pull/316661/commits/39617bbcf82ab59019a3ba454a89dfaa9ed010c5

test/sanity/src/wsl.test.ts
deepak1556: https://github.com/microsoft/vscode/pull/316661/commits/39617bbcf82ab59019a3ba454a89dfaa9ed010c5

src/vs/workbench/api/node/extHostMcpNode.ts
deepak1556: https://github.com/microsoft/vscode/pull/316661/commits/39617bbcf82ab59019a3ba454a89dfaa9ed010c5

src/vs/platform/browserView/electron-main/browserView.ts
kycutler: Ok seems fine for now -- but could you please open an issue with more information about the caller / problem so I can follow up. Thanks!

test/sanity/src/context.ts
dmitrivMS: Maybe good to open an issue to track the follow-up.

dmitrivMS (dismissed):
LGTM

src/vs/platform/browserView/electron-main/browserView.ts
deepak1556: Something has changed here, I have reverted https://github.com/microsoft/vscode/pull/316661/commits/00657d357b7740f7630002a27e63560959f1825c to identify the callsite but latest build passes. I will leave it that for now.

lramos15 (approved):

lszomoru (approved):

lszomoru (approved):

PR #316719, opened 2026-05-15
Dispose ConversationFeature in sanity test teardown

benvillalobos:
Replace `conversationFeature.activated = false;` with `conversationFeature.dispose();` in each per-test `finally` block of the Copilot Chat Sanity Test suite.

The previous teardown only flipped the `activated` setter, which clears `_activatedDisposables` (commands, terminal quick-fix providers, participant registrations, etc.) but leaves `_disposables` intact. `_disposables` holds the two `onDidAuthenticationChange` listeners that each test instance subscribes to. Across retries those listeners accumulate on the shared `IAuthenticationService`, and on an auth-token oscillation they can re-trigger registration of the global `copilot-chat.fixWithCopilot` terminal quick-fix provider, surfacing as a flaky "already registered" failure.

`dispose()` is a strict superset of `activated = false` — it clears both stores. Each `ConversationFeature` instance's disposables are per-listener / per-registration (verified against `IAuthenticationService.onDidAuthenticationChange` and `vscode.window.registerTerminalQuickFixProvider`), so disposing the test-owned instance does not affect the contribution-registered `ConversationFeature` from the extension activation path.

No test reuses `conversationFeature` after its `finally` block.

### DIFF
--- a/extensions/copilot/src/extension/test/vscode-node/sanity.sanity-test.ts
+++ b/extensions/copilot/src/extension/test/vscode-node/sanity.sanity-test.ts
@@ -89,7 +89,7 @@ suite('Copilot Chat Sanity Test', function () {
 				const conversation = conversationStore.getConversation(result2.metadata.responseId);
 				assert.ok(conversation, 'Expected conversation to be available');
 			} finally {
-				conversationFeature.activated = false;
+				conversationFeature.dispose();
 			}
 		});
 	});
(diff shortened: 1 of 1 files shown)


benvillalobos:
Closing as superseded by #316732. That PR drops the per-test `ConversationFeature` construction entirely, so the four `finally` blocks this PR was patching no longer exist on `main`. The leak diagnosis here (`activated = false` doesn't dispose `_disposables` / the auth listener) is valid but moot once #316732 lands.

PR #316728, opened 2026-05-15
extensions: disable Disable Globally when extension is workspace-enabled only

j-abed:
Fixes #244138 

## Summary
`DisableGloballyAction` was enabled when an extension was `EnabledWorkspace`.
In that state, the extension is only enabled for the current workspace and is not globally enabled, so "Disable (Globally)" should not be available.

## Changes
- Updated `DisableGloballyAction.update()` to enable only when:
  - extension is installed
  - enablement state is `EnablementState.EnabledGlobally`
  - global enablement can be changed
- Added regression test:
  - `Test DisableGloballyAction when the extension is only enabled for workspace`

## Files
- src/vs/workbench/contrib/extensions/browser/extensionsActions.ts
- src/vs/workbench/contrib/extensions/test/electron-browser/extensionsActions.test.ts

## Before / After
- Before: "Disable (Globally)" could be enabled for `EnabledWorkspace`.
- After: "Disable (Globally)" is enabled only for `EnabledGlobally`.

## Validation
- Targeted `DisableGloballyAction` tests pass, including the new regression case.

### DIFF
--- a/src/vs/workbench/contrib/extensions/browser/extensionsActions.ts
+++ b/src/vs/workbench/contrib/extensions/browser/extensionsActions.ts
@@ -1780,7 +1780,7 @@ export class DisableGloballyAction extends ExtensionAction {
 				return;
 			}
 			this.enabled = this.extension.state === ExtensionState.Installed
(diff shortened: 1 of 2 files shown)


j-abed:

PR #316792, opened 2026-05-16
Extension Bisect: keep eliminated extensions disabled

### LINKED ISSUE #237092: Extension Bisect should not re-enable most extensions throughout the process
This is directly related to #111669 in which the OP describes how bisect is unable to find _any_ "bad" extensions if there is more than one.

According to the docs, [bisect re-enables the extensions it considers "fine"](https://code.visualstudio.com/blogs/2021/02/16/extension-bisect). This is problematic because if you have more than one "bad" extension, it will never be able to narrow down the problem to a single extension, and it will incorrectly tell you the problem may be with VS Code.

This could be fixed by bisect not re-enabling any extensions it has removed from the problem set until one of the "bad" ones are found. This would allow you to keep finding problematic extensions by repeatedly running the bisect tool until none remain, instead of the tool only working at all when there is _only_ one "bad" extension.

# Detailed Design

Here is the example from the docs on how bisect works now:

- Bisect divides 24 extensions into two halves of 12 extensions each, and **it disables all 12 extensions of the second half.**
- In this sample, **the 8th extension is the "bad" one**, so it is the first half and not disabled. Things are still not working as we'd expect. Because there is still an issue, extension bisect repeats the process, dividing the first 12 extensions into two parts: **6 are enabled** and **6 are disabled**. **All other extensions are also re-enabled.**
...

That very last part is what is problematic. If we eliminate that step, the tool would be able to uncover any number of problematic extensions. I will give two examples of my own: one with one bad extension, and one with two. (I am not sure what bisect does when there are an odd number of extensions, so for my examples I will assume it will give the remainder to the upper half when partitioning an odd number)

- An enabled extension: …
- A disabled extension: X
- A "bad" extension: ☢️

## Example: 12 extensions with 1 bad extension

1. The bad extension is no. 3; bisect disables the upper half of all 12

| — | — | ☢️ | — | — | — | X | X | X | X | X | X |
|---|---|---|---|---|---|---|---|---|---|---|---|

2. There is still a problem So bisect disables the upper half of the lower 6

| — | — | ☢️ | X | X | X | X | X | X | X | X | X |
|---|---|---|---|---|---|---|---|---|---|---|---|

3. There is still a problem So bisect disables extensions 2 and 3

| — | X | X (☢️) | X | X | X | X | X | X | X | X | X |
|---|---|---|---|---|---|---|---|---|---|---|---|

4. Problem gone, so bisect disables the non-problematic extension (no. 1), and then re-enables the lower half of the previously disabled set (so, just the 2nd one)

| X | X | ☢️ | X | X | X | X | X | X | X | X | X |
|---|---|---|---|---|---|---|---|---|---|---|---|

5. Bisect can correctly identify no. 3 as the problematic extension

## Example: 12 extensions with 2 bad extensions

1. The bad extensions are no. 4 and 12; bisect disables the upper half of all 12

| — | — | — | ☢️ | — | — | X | X | X | X | X | X (☢️) |
|---|---|---|---|---|---|---|---|---|---|---|---|

2. There is still a problem, so bisect disables the upper half of the lower 6

| — | — | — | X (☢️) | X | X | X | X | X | X | X | X (☢️) |
|---|---|---|---|---|---|---|---|---|---|---|---|

2. There is still a problem So bisect disables the lower half of the lower 6 (the first 3). Now we are working with extensions 4-6. Bisect re-enables the lower half of this group, which is just no. 4 

| X | X | X | ☢️ | X | X | X | X | X | X | X | X (☢️) |
|---|---|---|---|---|---|---|---|---|---|---|---|

3. A problem is found with only one extension enabled, so this round of bisecting completes with a "bad" extension identified. However the user still experiences a problem so they start a new bisect. Let us assume the user has removed the first bad extension so there are now 11 extensions left, and the 11th one is the problematic one. Bisect disables the upper half of all 11 extensions (6-11)

| — | — | — | — | — | X | X | X | X | X | X (☢️) |
|---|---|---|---|---|---|---|---|---|---|---|

4. The user no longer experiences the problem, so bisect disables the un-problematic half entirely and re-enables the lower half of the remaining extensions. This repeats until the 10th extension is the only one still enabled, still no problem.

| X | X | X | X | X | — | — | — | X | X | X (☢️) |
|---|---|---|---|---|---|---|---|---|---|---|

| X | X | X | X | X | X | X | X | — | X | X (☢️) |
|---|---|---|---|---|---|---|---|---|---|---|

| X | X | X | X | X | X | X | X | X | — | X (☢️) |
|---|---|---|---|---|---|---|---|---|---|---|

5. I assume it would finally enable the last extension to confirm the user experiences a problem at all. And the user does experience it with only the 11th extension enabled, so bisect identifies it as the problem to the user.

| X | X | X | X | X | X | X | X | X | X | ☢️ |
|---|---|---|---|---|---|---|---|---|---|---|

---

Another scenario worth noting is that in the worst case, where the first and second extensions are the bad ones, the user will choose "I can reproduce" every single time. The current implementation takes this opportunity to tell the user that it could not detect a bad extension and that the problem might be with VS Code itself. In cases like this where we cannot be 100% sure that a singular extension is the problem, my new approach should probably include that same disclaimer when it identifies a potentially problematic extension.

While imperfect, this is still more useful than not identifying any extensions at all. I can think of several improvements that can be made as well; for example we could modify the algorithm to test both halves of a set of extensions for problems when a problem is encountered on one half and alert the user. This could save time by letting the user know that both halves of some set of extensions (which may be _all_ of them if it is the first try) reproduce the problem, and let them choose to continue bisecting or not.

NSExceptional:
## Summary

Fixes #237092 — Extension Bisect now keeps previously eliminated extensions disabled so that it can identify a problematic extension even when more than one extension is faulty.

**Before:** When bisect narrowed the search via a "still bad" answer, extensions outside the new search range were re-enabled. If any of those extensions were also bad, the user would keep answering "still bad" all the way to the end and bisect would conclude with the misleading "no extension has been identified" message.

**After:** Extensions in `[high, length)` — those eliminated by a "still bad" answer — stay disabled for the remainder of the bisect run. Extensions in `[0, low)` (eliminated by "I can't reproduce") were already verified safe collectively, so they remain enabled as before. Bisect now identifies one problematic extension per run; the user repeats bisect to find additional ones, as suggested in the issue's design section.

The "extension identified" confirmation also now mentions that the user may want to re-run bisect to find any additional problematic extensions.

### Algorithm change

The bisect state `(low, mid, high)` partitions extensions into four regions:

| Range            | Meaning                                              | Disabled (before)        | Disabled (after) |
|------------------|------------------------------------------------------|--------------------------|------------------|
| `[0, low)`       | Known-good (eliminated via "I can't reproduce")       | no                       | no               |
| `[low, mid)`     | Active search, currently enabled                      | no                       | no               |
| `[mid, high)`    | Active search, currently disabled                     | **yes**                  | **yes**          |
| `[high, length)` | "Skipped" (eliminated via "still bad" — bad status unknown) | no (re-enabled — bug) | **yes** (fix)    |

This simplifies to `isDisabled = i >= mid` and `disabledCount = extensions.length - mid`.

## Test plan

- [ ] Manual: start bisect with two known-bad extensions, confirm a bad one is identified and the prompt suggests re-running bisect.
- [ ] Manual: start bisect with one bad extension, confirm behavior is unchanged from the previous single-bad case.
- [ ] Manual: confirm the disabled-count notification reflects the total disabled (active + skipped), not just the active half.

---

Authored this with AI, hope that's okay :)

### DIFF
--- a/src/vs/workbench/services/extensionManagement/browser/extensionBisect.ts
+++ b/src/vs/workbench/services/extensionManagement/browser/extensionBisect.ts
@@ -82,9 +82,14 @@ class ExtensionBisectService implements IExtensionBisectService {
 		this._state = BisectState.fromJSON(raw);
 
 		if (this._state) {
-			const { mid, high } = this._state;
+			const { mid } = this._state;
 			for (let i = 0; i < this._state.extensions.length; i++) {
-				const isDisabled = i >= mid && i < high;
+				// Disable both the currently bisected upper half [mid, high) and any
+				// extensions previously eliminated from the search via a "still bad"
+				// answer [high, length). Keeping the latter disabled is what allows
+				// bisect to find additional problematic extensions across repeated
+				// runs when more than one extension is faulty (issue #237092).
+				const isDisabled = i >= mid;
 				this._disabled.set(this._state.extensions[i], isDisabled);
 			}
 			logService.warn('extension BISECT active', [...this._disabled]);
@@ -96,7 +101,7 @@ class ExtensionBisectService implements IExtensionBisectService {
 	}
 
 	get disabledCount() {
-		return this._state ? this._state.high - this._state.mid : -1;
+		return this._state ? this._state.extensions.length - this._state.mid : -1;
 	}
 
 	isDisabledByBisect(extension: IExtension): boolean {
@@ -307,7 +312,7 @@ registerAction2(class extends Action2 {
 				message: localize('done.msg', "Extension Bisect"),
 				primaryButton: localize({ key: 'report', comment: ['&& denotes a mnemonic'] }, "&&Report Issue & Continue"),
 				cancelButton: localize('continue', "Continue"),
-				detail: localize('done.detail', "Extension Bisect is done and has identified {0} as the extension causing the problem.", done.id),
+				detail: localize('done.detail', "Extension Bisect is done and has identified {0} as the extension causing the problem. If you still experience the problem after disabling {0}, run Extension Bisect again to identify any additional problematic extensions.", done.id),
 				checkbox: { label: localize('done.disbale', "Keep this extension disabled"), checked: true }
 			});
 			if (res.checkboxChecked) {
--- a/src/vs/workbench/services/extensionManagement/browser/extensionBisect.ts
+++ b/src/vs/workbench/services/extensionManagement/browser/extensionBisect.ts
@@ -64,7 +64,7 @@ class BisectState {
 	) { }
 }
 
-class ExtensionBisectService implements IExtensionBisectService {
+export class ExtensionBisectService implements IExtensionBisectService {
 
 	declare readonly _serviceBrand: undefined;
 
--- a/src/vs/workbench/services/extensionManagement/test/browser/extensionBisect.test.ts
+++ b/src/vs/workbench/services/extensionManagement/test/browser/extensionBisect.test.ts
@@ -1,0 +1,156 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import assert from 'assert';
+import { ensureNoDisposablesAreLeakedInTestSuite } from '../../../../../base/test/common/utils.js';
+import { InMemoryStorageService, IStorageService } from '../../../../../platform/storage/common/storage.js';
+import { NullLogService } from '../../../../../platform/log/common/log.js';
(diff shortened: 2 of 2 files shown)


NSExceptional:
@sandy081 this has been sitting in the backlog as a feature request for a while, but the more I think about it, I think it is actually a bug-fix and not a feature request: bisect re-enables extensions it already eliminated, so when more than one extension is faulty it can't isolate the culprit at all, and becomes useless.

If the team agrees this is more of a bug, could that reclassification get this PR unstuck? CLA's signed and CI is green.

Additionally, if there's anything I need to do to get this moved along, please let me know, this is only my 2nd PR to VS Code ever

PR #316795, opened 2026-05-16
[WIP] Fix: avoid re-entrancy deadlock in _resolveUtilityOverride for non-copilot vendors

rwoll:
Fixes the deadlock described in https://github.com/microsoft/vscode-copilot-evaluation/issues/3904#issuecomment-4468003704

https://github.com/microsoft/vscode/pull/316534 added utility alias override settings (`chat.utilityModel` / `chat.utilitySmallModel`). When the override points to a non-copilot vendor (e.g. `azure/gpt-5`), `_resolveUtilityOverride` calls `lm.selectChatModels` which re-enters the language model service while the copilot provider's sequencer slot is still held. The non-copilot vendor's response can never be delivered back, deadlocking `selectChatModels` indefinitely.

The fix adds a `duringProviderResolution` flag that skips the `lm.selectChatModels` path during alias publishing. Non-copilot overrides are instead resolved lazily on first use via the normal `getChatEndpoint` path, when the copilot provider is no longer holding the sequencer.

### DIFF
--- a/extensions/copilot/src/extension/chatSessions/claude/node/test/claudeCodeModels.spec.ts
+++ b/extensions/copilot/src/extension/chatSessions/claude/node/test/claudeCodeModels.spec.ts
@@ -84,6 +84,9 @@ class RefreshableMockEndpointProvider implements IEndpointProvider {
 	getChatEndpoint(): Promise<IChatEndpoint> {
 		throw new Error('Not implemented');
 	}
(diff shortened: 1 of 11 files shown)


rwoll:
Closing in favor of #316811 which takes a cleaner approach (background resolution + republish) that avoids the metadata/endpoint mismatch race this flag-based approach has.

PR #316868, opened 2026-05-17
fix(extensions): hide Disable dropdown when extension is globally disabled but workspace-enabled

naouts:
Fixes #244138

When an extension is disabled globally and re-enabled only for the
current workspace (EnabledWorkspace state), the Disable button
incorrectly showed a dropdown with both "Disable" and
"Disable (Workspace)" options. Since the extension is already globally
disabled, offering "Disable" again is meaningless.

DisableGloballyAction now checks isDisabledGlobally() before enabling
itself for the EnabledWorkspace case, so only "Disable (Workspace)"
is shown without a dropdown.


naouts:

naouts:
@sandy081 ???

PR #316951, opened 2026-05-18
Drop stale wheel events after focus return

### LINKED ISSUE #28795: Editor: scroll jumps randomly (after Find?)
- VSCode Version: 1.13.0
- OS Version: Fedora 24

Steps to Reproduce:
here is the tough part. This issue only appeared ~6 weeks ago, but I can't manage to reproduce.
When simply editing multiple files (js, html), switch between them using Ctrl+P
Use Find to get to some text. Them mouse wheel to scroll.
Issue: scroll jumps to bottom of the file first time (it will be fine on further operations on that file, until switching again to another one).

It happens ~20% of the cases. I can't find exact steps to reproduce, but it's annoying enough to finally open an issue. Hope others encounter it too.
Will get back with more info when I can reliably reproduce.

Thanks

jynbil1:
Fixes #28795

This drops stale wheel events before they reach editor wheel handling or shared scrollable elements. The guard uses the event timestamp and consumes events that are older than the current event loop by more than one second, which prevents queued wheel input from being replayed after VS Code regains focus while preserving current wheel input.

Verification:
- `NODE_OPTIONS=--max-old-space-size=8192 npx gulp compile-client` with Node 22.22.3
- `./scripts/test.sh --grep mouseEvent` with Node 22.22.3
- `npm run -s precommit` with Node 22.22.3
- `git diff --check`

### DIFF
--- a/src/vs/base/browser/mouseEvent.ts
+++ b/src/vs/base/browser/mouseEvent.ts
@@ -119,6 +119,35 @@ interface IGeckoMouseWheelEvent {
 	detail: number;
 }
 
+const STALE_MOUSE_WHEEL_EVENT_THRESHOLD = 1000;
+
+export function isStaleMouseWheelEvent(e: IMouseWheelEvent, now: number = Date.now()): boolean {
+	const timestamp = e.timeStamp;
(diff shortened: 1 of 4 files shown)


jynbil1:

PR #316992, opened 2026-05-18
Fix workbench hover not scrolling when content overflows

keyurgovrani:
## Summary

Workbench element hovers clip overflowing content with **no scrollbar**. Most visible on the Source Control **Commit Graph** history-item hover: a long commit message plus several co-authors is cut off with no way to scroll to the rest.

## Root cause

`HoverWidget.adjustHoverMaxHeight()` in `src/vs/platform/hover/browser/hoverWidget.ts` caps `max-height` **only** on the outer `containerDomNode` (`.monaco-hover`, which is `overflow: hidden`). But the scroll viewport is the `DomScrollableElement` that wraps `contentsDomNode`. Its root (`scrollbar.getDomNode()`) and `contentsDomNode` receive no height constraint, so `scrollbar.scanDomNode()` (run via `onContentsChanged()` at the end of `layout()`) measures `clientHeight === scrollHeight` and never renders the vertical scrollbar — the outer container just clips.

The editor `ContentHoverWidget._setHoverWidgetMaxDimensions()` already bounds all three nodes (`contentsDomNode`, `scrollbar.getDomNode()`, `containerDomNode`); the workbench hover bounded only one.

## Fix

Mirror the editor hover: also cap `max-height` on the scrollable root and `contentsDomNode`. As a side benefit the pre-existing "add right padding for the scrollbar" branch (previously dead because `clientHeight === scrollHeight`) now works as intended. Short content is unchanged (cap never hit → no scrollbar).

## Test

Adds `src/vs/platform/hover/test/browser/hoverWidget.test.ts`: shows a hover with tall HTMLElement content and asserts the scrollable root and contents node share the container `max-height`. Fails before the fix, passes after.

## Verification

Built VS Code and ran the Electron unit runner locally:

- `./scripts/test.sh --grep 'bounded so the scrollbar can show'` → **green** with the fix (`1 passing`).
- Same test with the two new lines removed → **red** (`AssertionError: '' !== '284px'`).

Verified against the `1.119.0` tag; `hoverWidget.ts` there is byte-identical to current `main`, so the result holds for this PR.

## Test plan

- [ ] Open a repo in the Source Control view, enable the Commit Graph (GRAPH section).
- [ ] Hover a commit whose message is long / has multiple co-authors.
- [ ] Before: content clipped, no scrollbar. After: vertical scrollbar, full content reachable.
- [ ] Short hovers unchanged (no scrollbar, same size).

### DIFF
--- a/src/vs/platform/hover/browser/hoverWidget.ts
+++ b/src/vs/platform/hover/browser/hoverWidget.ts
@@ -615,6 +615,11 @@ export class HoverWidget extends Widget implements IHoverWidget {
 		}
 
 		this._hover.containerDomNode.style.maxHeight = `${maxHeight}px`;
+		// Scroll is owned by the DomScrollableElement around contentsDomNode; bounding only the
+		// outer container clips with no scrollbar. Mirror the editor ContentHoverWidget and also
+		// bound the scrollable root + contents node so scanDomNode() can show the scrollbar.
+		this._hover.scrollbar.getDomNode().style.maxHeight = `${maxHeight}px`;
+		this._hover.contentsDomNode.style.maxHeight = `${maxHeight}px`;
 		if (this._hover.contentsDomNode.clientHeight < this._hover.contentsDomNode.scrollHeight) {
 			// Add padding for a vertical scrollbar
 			const extraRightPadding = `${this._hover.scrollbar.options.verticalScrollbarSize}px`;
--- a/src/vs/platform/hover/test/browser/hoverWidget.test.ts
+++ b/src/vs/platform/hover/test/browser/hoverWidget.test.ts
@@ -1,0 +1,132 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import * as assert from 'assert';
+import { Event } from '../../../../base/common/event.js';
+import { toDisposable } from '../../../../base/common/lifecycle.js';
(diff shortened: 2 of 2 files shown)


benibenj (changes requested):
The core max-height fix is correct, but it activates a pre-existing scrollbar-padding branch that does not reset its state when hover content later shrinks.

`layout()` clears `contentsDomNode.style.maxHeight`, then `adjustHoverMaxHeight()` adds `paddingRight` when a vertical scrollbar is needed. There is no corresponding reset/else path. For dynamically resizing HTMLElement hovers (the `ResizeObserver`-supported case), once tall content becomes short, the hover retains the scrollbar gutter and remains wider than necessary.

Please clear `contentsDomNode.style.paddingRight` at the start of each layout (alongside `maxHeight`) or explicitly clear it when a vertical scrollbar is no longer needed, and add coverage for the tall-to-short transition. The current test only asserts inline max-height values, so it does not catch this regression.

keyurgovrani:
@benibenj thanks, that was the real gap. Addressed and rebased onto current `main`.

While rebasing I found that #331439 already landed the max-height part of this PR, so I dropped it. The branch is now a single commit for the reset you pointed at: `layout()` clears `contentsDomNode.style.paddingRight` alongside `maxHeight`, so the gutter survives only while the content overflows. Clearing it in `layout()` rather than adding an else in `adjustHoverMaxHeight()` also means the overflow check is measured against content that a stale gutter has not narrowed.

Coverage for the tall to short transition is in `hoverService.test.ts` as `should drop the scrollbar padding once content stops overflowing`. Locally it fails without the fix (`scrollbarPadding: "10px"` instead of `""`) and passes with it; `--grep "Hover"` is 145 passing.

PR #317063, opened 2026-05-18
chore: show deprecation banner for arm32 remote connections

deepak1556:
In preparation for https://github.com/microsoft/vscode/pull/316661

PR #317086, opened 2026-05-18
chore: show deprecation banner for arm32 remote connections

deepak1556:
Backports https://github.com/microsoft/vscode/pull/317063

### DIFF
--- a/src/vs/workbench/contrib/remote/browser/remoteConnectionHealth.ts
+++ b/src/vs/workbench/contrib/remote/browser/remoteConnectionHealth.ts
@@ -7,7 +7,7 @@ import { IWorkbenchContribution } from '../../../common/contributions.js';
 import { IRemoteAgentService, remoteConnectionLatencyMeasurer } from '../../../services/remote/common/remoteAgentService.js';
 import { IWorkbenchEnvironmentService } from '../../../services/environment/common/environmentService.js';
 import { localize } from '../../../../nls.js';
-import { isWeb } from '../../../../base/common/platform.js';
+import { isWeb, OperatingSystem } from '../../../../base/common/platform.js';
 import { ITelemetryService } from '../../../../platform/telemetry/common/telemetry.js';
 import { getRemoteName, getRemoteServerRootPath } from '../../../../platform/remote/common/remoteHosts.js';
 import { IBannerService } from '../../../services/banner/browser/bannerService.js';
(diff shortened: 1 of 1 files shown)


src/vs/workbench/contrib/remote/browser/remoteConnectionHealth.ts
```diff
@@ -84,40 +86,45 @@ export class InitialRemoteConnectionHealthContribution implements IWorkbenchCont
...
-							actions,
-							icon: Codicon.warning,
-							closeLabel: `Do not show again in v${this.productService.version}`,
-							onClose: () => {
-								this.storageService.store(`${BANNER_REMOTE_UNSUPPORTED_CONNECTION_DISMISSED_KEY}`, this.productService.version, StorageScope.PROFILE, StorageTarget.MACHINE);
-							}
-						});
-					}
+						]
+					);
```
dmitrivMS, line 107: Should there be a `return` here - i.e. can we show second banner below after showing this one? Or that is the intent?

src/vs/workbench/contrib/remote/browser/remoteConnectionHealth.ts
```diff
@@ -167,6 +174,26 @@ export class InitialRemoteConnectionHealthContribution implements IWorkbenchCont
 		}
 	}
 
+	private _showVersionedBanner(id: string, storageKey: string, message: string, actions?: ReadonlyArray<ILinkDescriptor>): void {
+		const bannerDismissedVersion = this.storageService.get(storageKey, StorageScope.PROFILE) ?? '';
+		// Ignore patch versions and dismiss the banner if the major and minor versions match.
+		const shouldShowBanner = bannerDismissedVersion.slice(0, bannerDismissedVersion.lastIndexOf('.')) !== this.productService.version.slice(0, this.productService.version.lastIndexOf('.'));
```
dmitrivMS, line 180: This really means once per week now... perhaps something like a bi-weekly or monthly based on time would be less annoying.

dmitrivMS (approved):

src/vs/workbench/contrib/remote/browser/remoteConnectionHealth.ts
deepak1556: Yeah the intent is for arm32 banner to override.

src/vs/workbench/contrib/remote/browser/remoteConnectionHealth.ts
deepak1556: Good call, however the arm32 banner will be for a week now and gets removed after 1.122 when https://github.com/microsoft/vscode/pull/316661 lands.

deepak1556: And for the legacy banner we want it to be annoying :)

PR #317205, opened 2026-05-18
Address GHE onboarding issues

cwebster-99:
Addresses https://github.com/microsoft/vscode-internalbacklog/issues/7744 and https://github.com/microsoft/vscode-internalbacklog/issues/7744 

The most important changes are:

**UI/UX Improvements for GitHub Enterprise Sign-In:**
* Added new CSS classes in `variationA.css` to style the inline GitHub Enterprise instance input form, including input fields, buttons, messages, and progress indicators.
* Replaced the modal quick input for entering the GitHub Enterprise instance with an inline form that appears directly in the onboarding UI, providing real-time validation, error messages, and a seamless user experience.
* Introduced a progress indicator that replaces the form or sign-in buttons while waiting for sign-in to complete, and ensures the UI restores to its previous state if the flow is cancelled or fails. [[1]](diffhunk://#diff-2677df85a7c2473e3867b11b3ec8466f217464832f4da47b48e4bdcd2d7f735aR580-R637) [[2]](diffhunk://#diff-2677df85a7c2473e3867b11b3ec8466f217464832f4da47b48e4bdcd2d7f735aR649-R661) [[3]](diffhunk://#diff-2677df85a7c2473e3867b11b3ec8466f217464832f4da47b48e4bdcd2d7f735aR672-R875)

**State Management and Logic Changes:**
* Added snapshot and restore logic for the sign-in action area, allowing the UI to revert to its previous state after cancellation or errors in the sign-in process.
* Updated the configuration registry and settings update logic to ensure the correct authentication provider and instance URI are set when proceeding with GitHub Enterprise sign-in.

**Dependency and Cleanup:**
* Removed the dependency on `IQuickInputService` since the quick input modal is no longer used for entering the GitHub Enterprise instance. [[1]](diffhunk://#diff-2677df85a7c2473e3867b11b3ec8466f217464832f4da47b48e4bdcd2d7f735aR28-L29) [[2]](diffhunk://#diff-2677df85a7c2473e3867b11b3ec8466f217464832f4da47b48e4bdcd2d7f735aL135)

These changes collectively provide a more integrated and user-friendly onboarding experience for users signing in with GitHub Enterprise.

https://github.com/user-attachments/assets/8cd3d5ea-406a-4773-b017-21b729409fb8


src/vs/workbench/contrib/welcomeOnboarding/browser/onboardingVariationA.ts
```diff
@@ -458,42 +457,51 @@ export class OnboardingVariationA extends Disposable implements IOnboardingServi
 		subtitle.textContent = localize('onboarding.signIn.heroSubtitle', "Sign in to continue with AI-powered development.");
 
 		const actions = append(contentMain, $('.onboarding-a-signin-actions'));
+		this._signInActionsEl = actions;
+
+		if (this._userSignedIn) {
```
pwang347, line 489: rather than rendering a message here, should it just automatically move to the next step? (when would a user see this if they aren't just manually running the onboarding command?)

src/vs/workbench/contrib/welcomeOnboarding/browser/onboardingVariationA.ts
```diff
@@ -605,62 +644,225 @@ export class OnboardingVariationA extends Disposable implements IOnboardingServi
...
+			input.type = 'text';
+			input.spellcheck = false;
+			input.autocomplete = 'off';
+			input.placeholder = localize('onboarding.signIn.enterprise.placeholder', 'i.e. "octocat" or "https://octocat.ghe.com"...');
+			input.setAttribute('aria-label', localize('onboarding.signIn.enterprise.prompt', "What is your {0} instance?", defaultChat.provider.enterprise.name));
+			if (initialValue) {
+				input.value = initialValue;
+			}
+
+			const submit = append(row, $<HTMLButtonElement>('button.onboarding-a-signin-ghe-submit'));
```
pwang347: I think we probably can reuse the built-in button class to avoid the need for styling, looks like the theming is a bit off 

Image

src/vs/workbench/contrib/welcomeOnboarding/browser/onboardingVariationA.ts
```diff
@@ -605,62 +644,225 @@ export class OnboardingVariationA extends Disposable implements IOnboardingServi
...
+			localDisposables.add(addDisposableListener(input, EventType.KEY_DOWN, e => {
+				const event = new StandardKeyboardEvent(e);
+				if (event.keyCode === KeyCode.Escape) {
+					event.preventDefault();
+					event.stopPropagation();
+					finish(undefined);
+				}
+			}));
+
+			// Temporarily repurpose the footer Back button (lower-right) so
```
pwang347: curious why we need to repurpose this; could we either just change the click handler on the original button (if it's still visible in this state) or should we add a different button? seems odd to refer to an existing button here but change the display

src/vs/workbench/contrib/welcomeOnboarding/browser/onboardingVariationA.ts
```diff
@@ -605,62 +644,225 @@ export class OnboardingVariationA extends Disposable implements IOnboardingServi
...
-			resolvedUri = `https://${resolvedUri}.ghe.com`;
-		} else if (!result.toLowerCase().startsWith('https://')) {
-			resolvedUri = `https://${result}`;
-		}
+				resolve(result);
+			};
 
-		await this.configurationService.updateValue(defaultChat.providerUriSetting, resolvedUri, ConfigurationTarget.USER);
-		return true;
+			localDisposables.add(addDisposableListener(form, EventType.SUBMIT, e => {
```
pwang347: might be good also to double-check in codebase if there's a way to handle the submit action here instead of redefining all of these event listeners

src/vs/workbench/contrib/welcomeOnboarding/browser/onboardingVariationA.ts
```diff
@@ -605,62 +644,225 @@ export class OnboardingVariationA extends Disposable implements IOnboardingServi
...
+
+			const validate = (): boolean => {
+				const value = input.value.trim();
 				isSingleWord = false;
+				input.classList.remove('invalid');
+				message.classList.remove('error', 'info');
+
 				if (!value) {
-					return undefined;
+					message.textContent = localize('onboarding.signIn.enterprise.prompt', "What is your {0} instance?", defaultChat.provider.enterprise.name);
```
pwang347: nit: should we have a constant for this label? (it's repeated a few times)

src/vs/workbench/contrib/welcomeOnboarding/browser/onboardingVariationA.ts
```diff
@@ -605,62 +644,225 @@ export class OnboardingVariationA extends Disposable implements IOnboardingServi
 				severity: Severity.Error,
 				message: localize('onboarding.signIn.enterprise.error', "GitHub Enterprise sign-in failed. Check your instance URL and try again."),
 			});
+		} finally {
+			progress?.dispose();
 		}
 	}
 
-	private async _ensureEnterpriseInstance(): Promise<boolean> {
+	private _snapshotSignInActions(): { restore: () => void } {
```
pwang347: nit: instead of snapshotting and restoring, should it just render reactively based on current state? (ie., whether an enterprise sign in is in progress?)

src/vs/workbench/contrib/welcomeOnboarding/browser/onboardingVariationA.ts
cwebster-99: The default behavior moves the user to the next step. The main case for this is if they go back in the flow

src/vs/workbench/contrib/welcomeOnboarding/browser/onboardingVariationA.ts
cwebster-99: I ended up reusing the input box element. Attaching screenshot for reference of new behavior 
image

src/vs/workbench/contrib/welcomeOnboarding/browser/onboardingVariationA.ts
pwang347: maybe it should either hide this step or allow signing in to a different account with slightly different verbiage? e.g., you are currently signed in as ..., choose another account

src/vs/workbench/contrib/welcomeOnboarding/browser/onboardingVariationA.ts
```diff
@@ -574,95 +724,76 @@ export class OnboardingVariationA extends Disposable implements IOnboardingServi
 	}
 
 	private async _handleEnterpriseSignIn(): Promise<void> {
-		const watch = StopWatch.create();
+		const fullUriRegEx = /^(https:\/\/)?([a-zA-Z0-9-]+\.)*[a-zA-Z0-9-]+\.ghe\.com\/?$/;
```
pwang347: nit: would probably move the regex to a constant (since it's also used above) and add a helper + generate some tests for the parsing behaviour

eli-w-king (changes requested):
Image

Are we able to wrap the helper text overflow to a second line instead of extending the input width? Changes in branch: https://github.com/microsoft/vscode/tree/eli/ghe-input-message-wrap

eli-w-king (approved):

lramos15 (approved):

PR #317208, opened 2026-05-18
build: increase mac signature verification timeout

bryanchen-d:
## Summary
- increase macOS signature verification timeout from 3 minutes to 5 minutes for arch builds
- increase the universal macOS signature verification timeout from 2 minutes to 5 minutes

## Details
Build 440250 failed in the macOS (X64) Verify signature step because the signed app startup smoke check completed just as the Azure Pipelines task hit its timeout. The codesign output showed a valid Developer ID signature and a stapled notarization ticket, and the app eventually wrote startup timers, so this gives the smoke check more room without changing the verification behavior.

## Validation
- npm run precommit

### DIFF
--- a/build/azure-pipelines/darwin/product-build-darwin-universal.yml
+++ b/build/azure-pipelines/darwin/product-build-darwin-universal.yml
@@ -201,7 +201,7 @@ jobs:
           fi
           echo "Startup timers:"
           cat "$TIMERS_FILE"
-        timeoutInMinutes: 2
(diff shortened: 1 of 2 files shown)


rzhao271 (approved):

deepak1556:
Thanks! I was bumping this already in https://github.com/microsoft/vscode/pull/316661, x64 runs via rosetta hence the slowdown.

PR #317213, opened 2026-05-19
[Search Subagent] Don't disclose paths outside of current workspace

guomaggie:
Previously, the search_subagent tool was allowed to process responses containing absolute paths and open them without enforcing workspace boundaries. We now enforce the read-only file access policy in search_subagent hydration and read the file only if it passes this check; otherwise, we throw (in which case we won't append the out-of-workspace file to the searh subagent tool's final answer). 

Tested the following cases: 
- file lies inside workspace (search subagent calls and appends to final answer) 
- file doesn't exist (appends to final answer - existing behavior) 
- file is in another workspace (don't append to final answer)


extensions/copilot/src/extension/tools/node/searchSubagentTool.ts
```diff
@@ -212,11 +214,15 @@ class SearchSubagentTool implements ICopilotTool<ISearchSubagentParams> {
...
-				// For relative paths, immediately resolve against cwd.
-				// For absolute paths, use as-is and let openTextDocument throw if not found.
+				// Resolve candidate path first, then enforce read-only file access
+				// via shared toolUtils guards before hydrating
 				const uri = (!path.isAbsolute(filePath) && cwd)
 					? URI.joinPath(URI.file(cwd), filePath)
 					: URI.file(filePath);
+
+				await this.instantiationService.invokeFunction(accessor =>
+					assertFileOkForTool(accessor, uri, this._inputContext, { readOnly: true, workingDirectory })
```
guomaggie, line 225: what's the appropriate behavior on error (ie search subagent tries to search out-of-workspace dir and fails)

guomaggie: should we omit the (raw) line altogether from the search subagent response?

extensions/copilot/src/extension/tools/node/searchSubagentTool.ts
```diff
@@ -231,10 +238,24 @@ class SearchSubagentTool implements ICopilotTool<ISearchSubagentParams> {
...
+					);
+				} catch {
+					// isFileExternalAndNeedsConfirmation throws for nonexistent files;
+					// treat that as "not external" so the original line is preserved.
+				}
+
+				if (!isExternal) {
+					// If we can't read the file, keep the original line
+					processedLines.push(`${trimmedLine} (unable to read file: ${err})`);
+				}
```
guomaggie, line 259: Note: When manually testing (ex. asking the search subagent to read a file outside of the workspace), we get a 400 timeout error (expected). 
Image

The ICM tests the out-of-workspace behavior by directly calling parseFinalAnswerAndHydrate with 2 temporary directories. Tested the following cases and verified the subagent behavior at this level.
* hydrates relative paths resolved inside the workspace
* invoke does not leak outside-workspace paths or contents into the result
* keeps inside-workspace paths that do not exist with an error suffix

extensions/copilot/src/extension/tools/node/test/searchSubagentTool.spec.ts
```diff
@@ -193,7 +193,7 @@ suite('SearchSubagentTool', () => {
 				'- /workspace/file.ts:10-20'
 			].join('\n');
 
-			const result = await tool['parseFinalAnswerAndHydrate'](response, '/workspace', notCancelled);
+				const result = await tool['parseFinalAnswerAndHydrate'](response, '/workspace', undefined, notCancelled);
```
bhavyaus: nit: indentation seems off here and on line 185

extensions/copilot/src/extension/tools/node/searchSubagentTool.ts
```diff
@@ -232,9 +239,24 @@ class SearchSubagentTool implements ICopilotTool<ISearchSubagentParams> {
...
+					isExternal = await this.instantiationService.invokeFunction(accessor =>
+						isFileExternalAndNeedsConfirmation(accessor, uri, this._inputContext, { readOnly: true, workingDirectory })
+					);
+				} catch {
+					// isFileExternalAndNeedsConfirmation throws for nonexistent files;
+					// treat that as "not external" so the original line is preserved.
+				}
+
+				if (!isExternal) {
+					// If hydration fails (e.g. the captured path didn't resolve because the model's formatting drifted),
```
bhavyaus, line 256: indentation also off.

extensions/copilot/src/extension/tools/node/test/searchSubagentTool.spec.ts
```diff
@@ -182,7 +182,7 @@ suite('SearchSubagentTool', () => {
 				'- /workspace/other.ts (lines 30-40): test2',
 			].join('\n');
 
-			const result = await tool['parseFinalAnswerAndHydrate'](response, '/workspace', notCancelled);
+				const result = await tool['parseFinalAnswerAndHydrate'](response, '/workspace', undefined, notCancelled);
```
bhavyaus: could we also update the tests: 
  - External absolute path
  - Inside-workspace path that fails to open
  - Inside-workspace path that opens

bhavyaus (changes requested):

extensions/copilot/src/extension/tools/node/searchSubagentTool.ts
guomaggie: this indentation looks right locally (up-to-date with the latest commit), odd

guomaggie: update: i see it now

bhavyaus (dismissed):

DonJayamanne (dismissed):

bhavyaus (approved):

dmitrivMS (approved):

PR #317228, opened 2026-05-19
Add "Change Indentation Width" command (fixes #294080)

### LINKED ISSUE #294080: Feature request: command to adjust space-based indentation width, without a formatter
Basically what #17802 said, but it was closed for a lack of upvotes. This remains a significant friction point for myself and if I had to guess, many others. I hope to collect the required 20 upvotes this time.

### Movitation

The current workflow for taking a file indented with 2 spaces and indenting it with 4 (without a formatter) is a sequence of 4 commands:

1. **Indent using spaces (2)** to correct the indentation (defaults to 4)
2. **Convert indentation to tabs**
3. **Indent using tabs (4)**
4. **Convert indentation to spaces**

This is obviously extremely tedious and time consuming, especially when you need to do it across multiple files without a formatter. Even with a formatter it is tedious since you still have to run two commands instead of one.

### Proposed Fix

A single command that will detect the current indentation correctly (or prompt the user for it) and then run the above 4 commands would eliminate this problem.

Maybe this command could be called "Adjust Indentation" to avoid changing the behavior of any existing commands, and it could just call into the "Indent using tabs" command for tabs (since this problem doesn't exist for tabs), but for spaces, it would do the above 4 steps in one command, somehow.

As the original issue stated, not every language has a formatter. The indentation features themselves work without a formatter, so adjusting the indentation should also ideally work without one.

---

Happy to implement this myself if someone can point out the relevant files to me! That's honestly the biggest hurdle to contributing to VS Code imo.

**Related issues:**

- #17802 (previous discussion)
- #219821 ("Indent using spaces function does not convert tabs to spaces in mixed indentation text content")
- #1228 (early tabs/spaces conversion request)
- #3297 (indentation action UX issues)

NSExceptional:
## Summary

Adds a single editor action `editor.action.changeIndentationWidth` that re-indents the current file with a new indentation width, eliminating the four-step "Indent Using Spaces (old) → Convert to Tabs → Indent Using Tabs (new) → Convert to Spaces" dance users currently have to do to widen or narrow space-based indentation without a formatter. Resolves #294080.

The command:
- Reads `indentSize` / `insertSpaces` from the active model as the *current* indent width.
- Prompts the user for a new width via a quick pick (1–8), with the current value labelled.
- For space-indented files, rewrites every line's leading whitespace by counting its visual columns against the current width and regenerating with the new width (multi-level indents, partial-step remainders, and stray tabs are all preserved).
- For tab-indented files, falls through to just updating `tabSize` / `indentSize` (matching `Indent Using Tabs`).
- Updates `tabSize` / `indentSize` on the model so subsequent edits use the new width.

The command is also surfaced in the indentation status-bar picker alongside `Detect Indentation` and `Convert Indentation to …`.

## Test plan
- [x] `compile-check-ts-native` clean for the three touched files.
- [x] New `Change Indentation Width — TypeScript/Javascript` unit-test suite covers widen, narrow, odd-remainder, whitespace-only lines, stray tabs, and no-op cases.
- [x] Manual: open a 2-space-indented file, run "Change Indentation Width" → 4, confirm file is reflowed and status bar shows `Spaces: 4`.
- [x] Manual: tab-indented file, run command → only the displayed tab width changes.

### DIFF
--- a/src/vs/editor/contrib/indentation/browser/indentation.ts
+++ b/src/vs/editor/contrib/indentation/browser/indentation.ts
@@ -122,12 +122,8 @@ export class ChangeIndentationWidthAction extends EditorAction {
 			return;
 		}
 
-		const selection = editor.getSelection();
-		if (!selection) {
-			return;
-		}
-
 		const modelOpts = model.getOptions();
+		const currentTabSize = modelOpts.tabSize;
 		const currentIndentSize = modelOpts.indentSize;
 		const currentInsertSpaces = modelOpts.insertSpaces;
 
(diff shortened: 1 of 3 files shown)


NSExceptional:
Addressed each of Copilot's feedback, should I resolve them or leave them for a reviewer to verify?

PR #316566, opened 2026-05-15
Update Copilot CLI to 1.0.48

anthonykim1:
Should supersede: https://github.com/microsoft/vscode/pull/316294

PR #317241, opened 2026-05-19
Bump Agent host Copilot CLI to 1.0.49

anthonykim1:
Basically need https://github.com/microsoft/vscode/pull/316566 for agent host.
Edit: Bump to 1.0.49 for both agent host , non agent host Copilot CLI.

Sanity test: 
Windows: 
image

Mac: 
Screenshot 2026-05-21 at 10 36 44 AM


anthonykim1:
DO NOT MERGE: Getting image

build/linux/rpm/dep-lists.ts
```diff
@@ -56,6 +56,9 @@ export const referenceGeneratedDepsByArch = {
 		'libc.so.6(GLIBC_2.3.2)(64bit)',
 		'libc.so.6(GLIBC_2.3.3)(64bit)',
 		'libc.so.6(GLIBC_2.3.4)(64bit)',
+		'libc.so.6(GLIBC_2.30)(64bit)',
+		'libc.so.6(GLIBC_2.33)(64bit)',
+		'libc.so.6(GLIBC_2.34)(64bit)',
```
rzhao271: @deepak1556 do we support these versions of glibc? Because Copilot CLI also has a CG alert associated with it, we'd either have to dismiss that alert and keep the package version low, or bump our sysroots and/or min install requirements.

deepak1556: Good call, thanks for sharing. Our glibc minimum is 2.28 and that hasn't changed. These are coming from prebuilds which is a problem and I don't intend to bump our sysroot rather it should be the other way, can we build these new native addon dependencies of the sdk from source then they will get built with right glibc and glibcxx versions we ship with. (all defined under https://github.com/microsoft/vscode/blob/main/build/azure-pipelines/linux/setup-env.sh)

If we don't have access to the sources when we pull them here then the prebuilds should be adjusted in the sdk pipeline, this is not new. There is precedence with node-pty that is consumed https://github.com/microsoft/node-pty/pull/853

anthonykim1: Should I ask cli folks to potentially downgrade glibc that is used to build against @github/copilot / @github/copilot-sdk.
I believe we're already exposed to this with https://github.com/microsoft/vscode/pull/316566 bumping version for Chat extension route.

deepak1556: Correct, if we can contribute back that would be welcome as well.

devm33: do we know which binaries from copilot are getting linked against 2.3+ ?
everything should be built with a 2.28 sysroot

deepak1556: Seems to be `icu-native.node`

```
strings ~/Downloads/VSCode-linux-x64/resources/app/extensions/copilot/node_modules/@github/copilot/sdk/prebuilds/linux-x64/icu-native.node | grep 'GLIBC*'  
GLIBC_2.2.5
GLIBC_2.3
GLIBC_2.3.4
GLIBC_2.14
GLIBC_2.18
GLIBC_2.25
GLIBC_2.28
GLIBC_2.30
GLIBC_2.33
GLIBC_2.34
```

devm33: @deepak1556 can you confirm which copilot package version that is?

anthonykim1: I believe it's 1.0.48:

Tried running:
```
npm pack @github/copilot@1.0.48
tar -xzf github-copilot-1.0.48.tgz \
  package/prebuilds/linux-x64/icu-native.node \
  package/prebuilds/linux-x64/runtime.node

strings package/prebuilds/linux-x64/icu-native.node | grep -E 'GLIBC_' | sort -u
strings package/prebuilds/linux-x64/runtime.node | grep -E 'GLIBC_' | sort -u
``` 
and got: 

```
icu-native.node:
GLIBC_2.14
GLIBC_2.18
GLIBC_2.2.5
GLIBC_2.25
GLIBC_2.28
GLIBC_2.3
GLIBC_2.3.4
GLIBC_2.30
GLIBC_2.33
GLIBC_2.34
```
and 
``` 
runtime.node:
GLIBC_2.12
GLIBC_2.14
GLIBC_2.18
GLIBC_2.2.5
GLIBC_2.25
GLIBC_2.28
GLIBC_2.3
GLIBC_2.3.4
GLIBC_2.4
``` 
Screenshot 2026-05-20 at 2 41 21 PM

deepak1556: @anthonykim1 `>= 1.0.49` has the expected glibc version. Also, can you check which native modules are needed when bumping versions and remove the ones that are not. Seems like icu-native can be safely removed ? `win32-native` only if we need keep alive feature..

connor4312 (approved):

PR #317304, opened 2026-05-19
Fix listener removal when listener is not found

kiran1929:
The current implementation directly calls:

  this.listeners.splice(this.listeners.indexOf(listener), 1);

If the listener is not present, `indexOf()` returns `-1`,
causing `splice(-1, 1)` to silently remove the last listener
instead of doing nothing — corrupting the listener array.

=> Fix
Added an index guard before splicing:

  const idx = this.listeners.indexOf(listener);
  if (idx >= 0) {
    this.listeners.splice(idx, 1);
  }

=> Testing
- Added a unit test in `errors.test.ts` covering the
  case where an unregistered listener is removed
- `npm run transpile-client`
- `./scripts/test.sh --runGlob '**/errors.test.js'`

### DIFF
--- a/src/vs/base/common/errors.ts
+++ b/src/vs/base/common/errors.ts
@@ -50,7 +50,11 @@ export class ErrorHandler {
 	}
 
 	private _removeListener(listener: ErrorListenerCallback): void {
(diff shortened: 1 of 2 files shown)


kiran1929:

PR #317307, opened 2026-05-19
Fix Copilot CLI sessions hanging on patch apply after stream close

### LINKED ISSUE #316053: Copilot CLI: GPT 5.5 sessions will sometimes get stuck on applying patches
I've had a few cases this morning where after running a Copilot CLI + GPT 5.5 session for a while, the session will suddenly appear to get stuck on applying patches (see image below). When that happens, it never seems to make progress, and if I stop and ask the agent to continue, that doesn't seem to help either.

Once I've gotten into this spot, the only thing that seems to help is completely reloading the window and asking the agent to continue. So it seems as is something "corrupts" the session when it tries to apply patches, and there's no clear watch to make progress.

Image

Chulong-Li:
Fixes #316053

## Summary
Copilot CLI sessions could hang forever while applying patches whenever the chat stream closed mid-turn (most commonly after a steering yield, sometimes after a window reload). This wraps the session's response stream in a small router so an in-flight SDK turn survives its UI stream going away.

## Root Cause
When the participant yields for steering, VS Code closes the original
`ChatResponseStream` while the SDK turn keeps running in the background. The
next write throws `"Response stream has been closed"` from `throwIfDone`.

The fatal case is `ExternalEditTracker.trackEdit`, which only resolves once
`externalEdit`'s callback fires:

```ts
const onDidComplete = stream.externalEdit(uris, async () => {
		proceedWithEdit();   // never called if externalEdit throws synchronously
		await deferred.p;
});
```

A synchronous throw skips the callback, so proceedWithEdit never runs, the
patch-apply step awaits a promise nothing will resolve, and the session is
stuck until the window is reloaded.

Figure 1: Copilot CLI Patch Application Hang Forever After Mid-Turn Stream Closure
image

## Fix
Introduce `CopilotCLIResponseStreamRouter`, a stable facade that the session holds for its lifetime:
- attachStream swaps the target behind the router, so in-flight turns always reach the currently active stream (or are dropped when none is).
- Writes that throw "Response stream has been closed" clear the target and fall through to a no-op — correct when no UI is listening.
- externalEdit's fallback still invokes the callback, then resolves with an empty undoStopId. This is the single change that unblocks the SDK and fixes the hang.
- questionCarousel's fallback resolves to undefined, matching a dismissed prompt.

## Validation

Figure 2: Copilot CLI Patch Application No Longer Gets Stuck and Session Continues
image

### DIFF
--- a/extensions/copilot/src/extension/chatSessions/copilotcli/node/copilotcliSession.ts
+++ b/extensions/copilot/src/extension/chatSessions/copilotcli/node/copilotcliSession.ts
@@ -47,6 +47,7 @@ import { type McCommand, type McEvent, type McSessionCreateResult, MissionContro
 import { handleMcpPermission, handleReadPermission, handleShellPermission, handleWritePermission, type PermissionRequest, type PermissionRequestResult, showInteractivePermissionPrompt } from './permissionHelpers';
 import { TodoSqlQuery } from './todoSqlQuery';
 import { IQuestion, IQuestionAnswer, IUserQuestionHandler } from './userInputHelpers';
+import { createSingleCallFunction } from '../../../../util/vs/base/common/functional';
 
 /**
  * Known commands that can be sent to a CopilotCLI session instead of a free-form prompt.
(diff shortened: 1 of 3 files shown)


DonJayamanne:
@Chulong-Li thank you for this PR, totally missed this, will look into this

Chulong-Li:
@DonJayamanne Thanks for your review and support! Please feel free to update the code directly. Let me know if you have any questions or comments. I’m glad to help.

DonJayamanne (approved):

roblourens (approved):

DonJayamanne:
Seems to be stuck, lets close and re-open PR

PR #317363, opened 2026-05-19
fix: Git - change regexp to be compatible with SHA256 commits (#227281)

PositroniumJS:
Fix #227281

Modify the regexp to look not only for 40-character commit hashes but also 64-character.

### DIFF
--- a/extensions/git/src/git.ts
+++ b/extensions/git/src/git.ts
@@ -923,7 +923,7 @@ export function parseGitRemotes(raw: string): MutableRemote[] {
 	return remotes;
 }
 
-const commitRegex = /([0-9a-f]{40})\n(.*)\n(.*)\n(.*)\n(.*)\n(.*)\n(.*)(?:\n([^]*?))?(?:\x00)(?:\n((?:.*)files? changed(?:.*))$)?/gm;
+const commitRegex = /([0-9a-f]{40}|[0-9a-f]{64})\n(.*)\n(.*)\n(.*)\n(.*)\n(.*)\n(.*)(?:\n([^]*?))?(?:\x00)(?:\n((?:.*)files? changed(?:.*))$)?/gm;
 
 export function parseGitCommits(data: string): Commit[] {
 	const commits: Commit[] = [];
@@ -1031,7 +1031,7 @@ export function parseLsFiles(raw: string): LsFilesElement[] {
 		.map(([, mode, object, stage, file]) => ({ mode, object, stage, file }));
 }
 
-const stashRegex = /([0-9a-f]{40})\n(.*)\nstash@{(\d+)}\n(WIP\s)?on\s([^:]+):\s(.*)\n(\d+)\n(\d+)(?:\x00)/gmi;
+const stashRegex = /([0-9a-f]{40}|[0-9a-f]{64})\n(.*)\nstash@{(\d+)}\n(WIP\s)?on\s([^:]+):\s(.*)\n(\d+)\n(\d+)(?:\x00)/gmi;
 
 function parseGitStashes(raw: string): Stash[] {
 	const result: Stash[] = [];
@@ -1218,7 +1218,7 @@ export interface BlameInformation {
 
 function parseGitBlame(data: string): BlameInformation[] {
 	const lineSeparator = /\r?\n/;
-	const commitRegex = /^([0-9a-f]{40})/gm;
+	const commitRegex = /^([0-9a-f]{40}|[0-9a-f]{64})/gm;
 
 	const blameInformation = new Map<string, BlameInformation>();
 
@@ -1278,7 +1278,7 @@ const REFS_FORMAT = '%(refname)%00%(objectname)%00%(*objectname)';
 const REFS_WITH_DETAILS_FORMAT = `${REFS_FORMAT}%00%(parent)%00%(*parent)%00%(authorname)%00%(*authorname)%00%(committerdate:unix)%00%(*committerdate:unix)%00%(subject)%00%(*subject)`;
 
 function parseRefs(data: string): (Ref | Branch)[] {
-	const refRegex = /^(refs\/[^\0]+)\0([0-9a-f]{40})\0([0-9a-f]{40})?(?:\0(.*))?$/gm;
+	const refRegex = /^(refs\/[^\0]+)\0([0-9a-f]{40}|[0-9a-f]{64})\0([0-9a-f]{40}|[0-9a-f]{64})?(?:\0(.*))?$/gm;
 
 	const headRegex = /^refs\/heads\/([^ ]+)$/;
 	const remoteHeadRegex = /^refs\/remotes\/([^/]+)\/([^ ]+)$/;
@@ -2922,7 +2922,7 @@ export class Repository {
 		}
 
 		// Detached
-		const commitMatch = raw.match(/^(?<commit>[0-9a-f]{40})$/m);
+		const commitMatch = raw.match(/^(?<commit>[0-9a-f]{40}|[0-9a-f]{64})$/m);
 		if (commitMatch?.groups?.commit) {
 			return { name: undefined, commit: commitMatch.groups.commit, type: RefType.Head };
 		}
@@ -2999,9 +2999,9 @@ export class Repository {
 		const fn = (line: string): Ref | null => {
 			let match: RegExpExecArray | null;
 
-			if (match = /^([0-9a-f]{40})\trefs\/heads\/([^ ]+)$/.exec(line)) {
+			if (match = /^([0-9a-f]{40}|[0-9a-f]{64})\trefs\/heads\/([^ ]+)$/.exec(line)) {
 				return { name: match[1], commit: match[2], type: RefType.Head };
-			} else if (match = /^([0-9a-f]{40})\trefs\/tags\/([^ ]+)$/.exec(line)) {
+			} else if (match = /^([0-9a-f]{40}|[0-9a-f]{64})\trefs\/tags\/([^ ]+)$/.exec(line)) {
 				return { name: match[2], commit: match[1], type: RefType.Tag };
 			}
 
--- a/extensions/git/src/git.ts
+++ b/extensions/git/src/git.ts
@@ -3000,7 +3000,7 @@ export class Repository {
 			let match: RegExpExecArray | null;
 
 			if (match = /^([0-9a-f]{40}|[0-9a-f]{64})\trefs\/heads\/([^ ]+)$/.exec(line)) {
-				return { name: match[1], commit: match[2], type: RefType.Head };
+				return { name: match[2], commit: match[1], type: RefType.Head };
 			} else if (match = /^([0-9a-f]{40}|[0-9a-f]{64})\trefs\/tags\/([^ ]+)$/.exec(line)) {
 				return { name: match[2], commit: match[1], type: RefType.Tag };
 			}
--- a/extensions/git/src/blame.ts
+++ b/extensions/git/src/blame.ts
@@ -153,6 +153,8 @@ class GitBlameInformationCache {
 	}
 }
 
+const commitHashRegex = /^([0-9a-f]{40}|[0-9a-f]{64})$/i;
+
 export class GitBlameController {
 	private readonly _subjectMaxLength = 50;
 
@@ -419,7 +421,7 @@ export class GitBlameController {
 			// 1) Commit - Resource in the multi-file diff editor when viewing the details of a commit.
 			// 2) HEAD   - Resource on the left-hand side of the diff editor when viewing a resource from the index.
 			// 3) ~      - Resource on the left-hand side of the diff editor when viewing a resource from the working tree.
-			if (/^[0-9a-f]{40}$/i.test(ref) || ref === 'HEAD' || ref === '~') {
+			if (commitHashRegex.test(ref) || ref === 'HEAD' || ref === '~') {
 				workingTreeChanges = allChanges = [];
 				workingTreeAndIndexChanges = undefined;
 			} else if (ref === '') {
@@ -478,7 +480,7 @@ export class GitBlameController {
 		} else {
 			// Resource with the `git` scheme
 			const { ref } = fromGitUri(textEditor.document.uri);
-			commit = /^[0-9a-f]{40}$/i.test(ref) ? ref : repository.HEAD.commit;
+			commit = commitHashRegex.test(ref) ? ref : repository.HEAD.commit;
 		}
 
 		// Git blame information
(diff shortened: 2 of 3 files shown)


PositroniumJS:

extensions/git/src/git.ts
```diff
@@ -923,7 +923,7 @@ export function parseGitRemotes(raw: string): MutableRemote[] {
 	return remotes;
 }
 
-const commitRegex = /([0-9a-f]{40})\n(.*)\n(.*)\n(.*)\n(.*)\n(.*)\n(.*)(?:\n([^]*?))?(?:\x00)(?:\n((?:.*)files? changed(?:.*))$)?/gm;
+const commitRegex = /([0-9a-f]{40}|[0-9a-f]{64})\n(.*)\n(.*)\n(.*)\n(.*)\n(.*)\n(.*)(?:\n([^]*?))?(?:\x00)(?:\n((?:.*)files? changed(?:.*))$)?/gm;
```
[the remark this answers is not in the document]
PositroniumJS: Treated in 249de49283ca79acd660a190ed3bbd4d43b84d6e

extensions/git/src/git.ts
```diff
@@ -2999,9 +2999,9 @@ export class Repository {
 		const fn = (line: string): Ref | null => {
 			let match: RegExpExecArray | null;
 
-			if (match = /^([0-9a-f]{40})\trefs\/heads\/([^ ]+)$/.exec(line)) {
+			if (match = /^([0-9a-f]{40}|[0-9a-f]{64})\trefs\/heads\/([^ ]+)$/.exec(line)) {
 				return { name: match[1], commit: match[2], type: RefType.Head };
```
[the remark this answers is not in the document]
PositroniumJS: Applied in 9330fb4f2b95ba01cc192ca626491bf2a7a83a65

YongDo-Hyun (commented):
LGTM

PositroniumJS:
@lszomoru what do you think?

rostidev:
Hope this PR will be merged eventually and it will fix the #227281 bug.
I have a small general question about the changes in this PR.
Basically it changes `[0-9a-f]{40}` to `([0-9a-f]{40}|[0-9a-f]{64})` in several places.
What do you think about the `[0-9a-f]{40}(?:[0-9a-f]{24})` version instead? I think it could be better from the performance perspective, especially with SHA256 Git repositories.

extensions/git/src/git.ts
```diff
@@ -2999,9 +2999,9 @@ export class Repository {
 		const fn = (line: string): Ref | null => {
 			let match: RegExpExecArray | null;
 
-			if (match = /^([0-9a-f]{40})\trefs\/heads\/([^ ]+)$/.exec(line)) {
-				return { name: match[1], commit: match[2], type: RefType.Head };
-			} else if (match = /^([0-9a-f]{40})\trefs\/tags\/([^ ]+)$/.exec(line)) {
+			if (match = /^([0-9a-f]{40}|[0-9a-f]{64})\trefs\/heads\/([^ ]+)$/.exec(line)) {
+				return { name: match[2], commit: match[1], type: RefType.Head };
```
rostidev, line 3003: Why indexes of `match`  for `name` and `commit` were changed? Was there some other bug that you fixed on the way? Before your change those indexes where different in `if-else` branches and now they are same.

PositroniumJS:
> Hope this PR will be merged eventually and it will fix the #227281 bug. I have a small general question about the changes in this PR. Basically it changes `[0-9a-f]{40}` to `([0-9a-f]{40}|[0-9a-f]{64})` in several places. What do you think about the `[0-9a-f]{40}(?:[0-9a-f]{24})` version instead? I think it could be better from the performance perspective, especially with SHA256 Git repositories.

I do not know anything about the performances of regexp. If you think that your suggestion would be faster, then sure, there is no reason to disagree

extensions/git/src/git.ts
PositroniumJS: It was requested by Copilot in its review [here](https://github.com/microsoft/vscode/pull/317363#discussion_r3268985852)

rostidev: Yes I see now. This is actually a bug that never happens yet because this `getRemoteRefs` function is never called with `heads == true`, but this may change in any future version and it's really worth to be fixed. So this change is good.

rostidev:
> I do not know anything about the performances of regexp. If you think that your suggestion would be faster, then sure, there is no reason to disagree

Ok, this is not critical for this PR and was just an open question. I don't have permissions to approve this PR, but I think it is ready. Hope it will be approved and merged ASAP.

rostidev:
@lszomoru please review and approve this PR.
IMO it's ok.

rostidev:
I've asked GPT 5.6 Sol Ultra and Fable 5 (Ultracode - xhigh + workflows) to review this PR and this is what they told me.

My prompt:
> This is the full source code of the VS Code project. This branch has changes that should bring support of SHA-256 Git repositories.
>
> You can find several places that look like ([0-9a-f]{40}|[0-9a-f]{64})
>
> But there are still several places with just [0-9a-f]{40}
> Should they be changed as well?

GPT 5.6 Sol Ultra reply:
> No—do not replace every `{40}` mechanically. Change only values that represent Git object IDs. Git defines full object names as exactly 40 hex digits for SHA-1 or 64 for SHA-256. `[0-9a-f]{40,64}` would be wrong because it also accepts intermediate lengths. [Git hash-transition documentation](https://git-scm.com/docs/hash-function-transition/2.52.0.html)
> 
> For the scope of [issue #227281](https://github.com/microsoft/vscode/issues/227281) and [PR #317363](https://github.com/microsoft/vscode/pull/317363), I would request changes before merging.
> 
> ### In-scope problems
> 
> 1. [git.ts:1221](/home/huligan/develop/vscode/extensions/git/src/git.ts:1221) still truncates SHA-256 blame IDs.
> 
>    The matcher tries the 40-character alternative first without a following delimiter, so a 64-character ID successfully matches only its first 40 characters. I confirmed the resulting match length is 40.
> 
>    Use something like:
> 
>    ```ts
>    /^(?:[0-9a-f]{64}|[0-9a-f]{40})(?=\s)/
>    ```
> 
>    Alternatively:
> 
>    ```ts
>    /^[0-9a-f]{40}(?:[0-9a-f]{24})?(?=\s)/
>    ```
> 
>    The final `?` is essential; without it, that form accepts only SHA-256.
> 
>    The other changed patterns are safe because a newline, NUL, tab, or end anchor forces the engine to retry the 64-character alternative.
> 
> 2. [extensions/git/package.json:4109](/home/huligan/develop/vscode/extensions/git/package.json:4109) still limits `git.commitShortHashLength` to 40. It should allow 64 for complete SHA-256 display support.
> 
> 3. Test coverage is insufficient. The new SHA-256 `parseGitCommits` test is useful, but the [parseLsTree test](/home/huligan/develop/vscode/extensions/git/src/test/git.test.ts:595) exercises parsing code that was already length-agnostic. Missing coverage includes blame, stashes, refs/peeled tags, detached HEAD, remote refs, and the corrected remote `name`/`commit` mapping.
> 
> ### Remaining exact `{40}` matches
> 
> | Location | Decision | Reason |
> |---|---|---|
> | [build/lib/git.ts:22](/home/huligan/develop/vscode/build/lib/git.ts:22), [build/lib/git.ts:50](/home/huligan/develop/vscode/build/lib/git.ts:50), [build/lib/getVersion.ts:11](/home/huligan/develop/vscode/build/lib/getVersion.ts:11) | Change separately | These parse real HEAD/packed-ref/build commit IDs. Building VS Code from a SHA-256 checkout currently fails. |
> | [pluginMarketplaceService.ts:1070](/home/huligan/develop/vscode/src/vs/workbench/contrib/chat/common/plugins/pluginMarketplaceService.ts:1070) | Change separately | Generic `url` and `git-subdir` plugin sources can point at arbitrary Git repositories but reject 64-character pins. Update diagnostics at lines 977 and 1007 too. |
> | [pluginGitCommandService.ts:159](/home/huligan/develop/vscode/src/vs/workbench/contrib/chat/browser/pluginGitCommandService.ts:159), [line 189](/home/huligan/develop/vscode/src/vs/workbench/contrib/chat/browser/pluginGitCommandService.ts:189) | Not independently | This implementation is specifically GitHub-backed. Change only together with its backend/plugin-source contract. |
> | [version_manager.rs:36](/home/huligan/develop/vscode/cli/src/desktop/version_manager.rs:36) | Keep | This identifies published VS Code builds, not objects in an opened repository. |
> | [log.tmLanguage.json:88](/home/huligan/develop/vscode/extensions/log/syntaxes/log.tmLanguage.json:88) | Optional | Adding 64 improves highlighting only; it is not functional Git support. |
> | Darwin code-signing pipeline matchers | Keep | They match certificate fingerprints, not Git IDs. |
> | [secretFilter.ts:30](/home/huligan/develop/vscode/extensions/copilot/src/extension/chronicle/common/secretFilter.ts:30) | Keep | That is a fixed GitHub token format. |
> 
> ### Non-regex SHA-1 assumptions
> 
> For truly repo-wide support, the more serious remaining issue is the hard-coded SHA-1 empty-tree ID in:
> 
> - [agentHostGitService.ts:567](/home/huligan/develop/vscode/src/vs/platform/agentHost/node/agentHostGitService.ts:567)
> - [claudeWorkspaceFolderServiceImpl.ts:22](/home/huligan/develop/vscode/extensions/copilot/src/extension/chatSessions/vscode-node/claudeWorkspaceFolderServiceImpl.ts:22)
> - [chatSessionWorkspaceFolderServiceImpl.ts:29](/home/huligan/develop/vscode/extensions/copilot/src/extension/chatSessions/vscode-node/chatSessionWorkspaceFolderServiceImpl.ts:29)
> 
> These break empty SHA-256 repositories. The empty-tree ID should be obtained dynamically from Git, as the built-in extension already does at [repository.ts:2203](/home/huligan/develop/vscode/extensions/git/src/repository.ts:2203).
> 
> Also worth separate follow-ups:
> 
> - [diff.tmLanguage.json:140](/home/huligan/develop/vscode/extensions/diff/syntaxes/diff.tmLanguage.json:140): `{7,40}` → `{7,64}` for Git diff highlighting.
> - [sessionStoreTracking.ts:216](/home/huligan/develop/vscode/extensions/copilot/src/extension/chronicle/common/sessionStoreTracking.ts:216): `{7,40}` → `{7,64}` for commit-output extraction.
> 
> Best validation would be an integration test using `git init --object-format=sha256`, covering history, blame, stash, refs, detached HEAD, and remote refs. No files were modified during this review.

With the same prompt Fable 5 (Ultracode - xhigh + workflows) didn't find any issue, but after I showed the above reply of GPT Fable agreed:
> Yes — I agree with essentially all of it, and I verified its claims rather than taking them on faith. It's a better review than the one I gave you: it caught a real bug I missed and corrected one of my assessments. Here's what I confirmed:
> 
> ## The truncation bug at git.ts:1221 is real — this is the important finding
> 
> I executed the regex to confirm: /^([0-9a-f]{40}|[0-9a-f]{64})/gm matches only the first 40 characters of a 64-character hash, because regex alternation is ordered and nothing after the group forces backtracking. So parseGitBlame stores a truncated hash in SHA-256 repos. This has functional consequences: blame.ts:243 checks unpublishedCommits.has(blameInformation.hash) against full hashes from git log, and the hash is used as a cache key — the truncated value never matches. I also verified their claim that all the other changed patterns are safe: each has a following \n, \0, \t, or $ that forces the engine to retry the 64-char alternative. Line 1221 is the only one without a delimiter.
> 
> One quibble on the fix: their lookahead suggestion works, but the minimal fix consistent with the rest of the PR is simply swapping the alternatives to [0-9a-f]{64}|[0-9a-f]{40} — I tested that this matches both lengths correctly.
> 
> ## Everything else checks out
> 
> * package.json:4109 — confirmed, git.commitShortHashLength max is 40. Raising to 64 is a fair in-scope suggestion.
> * **Test coverage** — confirmed. parseLsTree parses with (\S+), so the new SHA-256 ls-tree test exercises code that never had a length constraint. And the getRemoteRefs name/commit swap fix (fe96a1220cb) on this branch has no test at all. Notably, the one parser that actually has a SHA-256 bug — blame — is exactly one with no test.
> * **Empty-tree hashes** — confirmed at all three cited locations, and they're passed to git read-tree, which will fail in a SHA-256 repo. The git extension's dynamic lookup at repository.ts:2201-2208 is exactly where they said.
> * pluginMarketplaceService.ts — the other AI is right and I was wrong here: url and git-subdir sources take arbitrary Git URLs, not just GitHub, so isOptionalGitSha rejecting 64-char pins is a genuine limitation. The diagnostics at lines 977 and 1007 do say "40-character".
> * **The keep/change table** — all verified: version_manager.rs:36 identifies published VS Code builds, the Darwin YAML matchers are security find-identity certificate fingerprints, secretFilter.ts is a token format, and diff.tmLanguage.json:140 does have {7,40}. Their sweep was broader than mine (Rust and YAML files, which my TypeScript-focused greps missed).
> * **The {40,64} warning** is correct — Git object names are exactly 40 or 64 hex digits, never in between.
> 
> ## One minor overstatement
> 
> "Building VS Code from a SHA-256 checkout currently fails" is only partially true. In build/lib/git.ts, the loose-ref path returns the ref file's contents without any length validation, so a normal branch checkout would work; only detached HEAD and packed-refs paths fail. And the scenario is exotic — Git can't clone across object formats, so you'd have to fast-export/import VS Code's repo yourself. "Change separately, low priority" is the right call regardless.
> 
> **Bottom line:** the review's "request changes" verdict is justified — the blame truncation alone warrants it, and it's a subtle enough bug that it validates the reviewer's method of checking how each regex behaves rather than just whether it was edited. If you want, I can apply the in-scope fixes (the 1221 swap, the package.json max, and blame/stash/refs tests including SHA-256 cases).

PositroniumJS:
@rostidev You are very welcome to make a pull request on my branch of my fork. I would give it a look and merge it!

rostidev:
Hi @PositroniumJS — thank you for this work and for the original investigation into #227281.

While reviewing this PR I found that one of the patterns it updates still truncates SHA-256 ids. In `parseGitBlame` the object id is not followed by any delimiter, and since regex alternation is ordered, `([0-9a-f]{40}|[0-9a-f]{64})` matches only the first 40 characters of a 64-character id — there is nothing after the group to force the engine to backtrack into the longer alternative. The other patterns here are fine, but only because a `\n`, `\0`, `\t` or `$` follows them.

I have opened #327494, which builds on your branch — your commits are preserved there with authorship intact — and additionally fixes the blame truncation, adds tests for the parsers involved (blame, refs, ls-remote, stashes), and verifies the result against real repositories created with `git init --object-format=sha256`.

I did not want to step on your PR, so please do say if you would rather fold these changes into this one instead; I am happy either way, and the credit for the original fix is yours.

PR #317389, opened 2026-05-19
Use copilot/copilot-utility-small for terminal-tool steering messages

meganrogge:
Route the three terminal-tool steering messages (background completion, input-needed, terminal-exited) through the `copilot/copilot-utility-small` alias instead of inheriting the conversation's model.

These notifications are agent-internal turns: they are short, system-initiated, and don't need the user's selected frontier model. The copilot extension already publishes a `copilot/copilot-utility-small` alias that honors the `chat.utilitySmallModel` setting, so this PR just opts those three call sites in by overriding `userSelectedModelId` in the existing `sendOptions`.

### Why this shape

A previous revision of this PR tried to route _all_ `isSystemInitiated: true` requests to utility-small inside `ChatService`. That was wrong:

- "system-initiated" is a broad bucket — not every system-initiated request is a free/cheap turn, and policy on which model and which billing applies should be decided per call site, not implicitly by the chat service.
- The grep showed only **three production call sites** today (all in `runInTerminalTool.ts`); most other internal turns already pick a utility model explicitly. Making each one opt in keeps the routing decision visible and reviewable.

### Changes

- `runInTerminalTool.ts`
  - Inject `ILanguageModelsService`.
  - Add a private `_resolveUtilitySmallModelId()` that walks `getLanguageModelIds()` and returns the first identifier whose metadata is `{vendor: 'copilot', id: 'copilot-utility-small'}`, or `undefined`.
  - When building the shared `sendOptions` for the three steering `sendRequest` calls, override `userSelectedModelId` with the resolved alias when available. Falls back to the conversation model (`lastRequest.modelId`) when the alias is unavailable (older copilot extension, alias not yet activated, etc.).

### Out of scope

- No changes to `ChatService` — the prior `_resolveSystemInitiatedModelId` / `_applySystemInitiatedModelOverride` plumbing has been removed.
- No new setting, no new API surface; `chat.utilitySmallModel` continues to gate the alias resolution inside the copilot extension.


meganrogge:

dbaeumer (dismissed):

bryanchen-d (approved):

PR #317407, opened 2026-05-19
Sanity test: enrich agent-mode timeout error with stream diagnostics

bryanchen-d:
### Context

The Copilot sanity test `E2E Production agent mode` (in [`sanity.sanity-test.ts`](https://github.com/microsoft/vscode/blob/main/extensions/copilot/src/extension/test/vscode-node/sanity.sanity-test.ts)) has been intermittently failing release / pre-release builds, e.g. [build 440598](https://dev.azure.com/monacotools/Monaco/_build/results?buildId=440598&view=logs&j=15ec5277-fdc6-550c-b203-acb65f44f31c&t=120c9214-43d4-5761-6dc9-d5e173c65e37), with the unhelpful error:

```
Error: timed out waiting for tool call.
```

The test waits up to 20 s for the production backend to emit a tool-call delta on the SSE stream. When it times out, the existing message only appends `stream.currentProgress`, which only captures `ChatResponseMarkdownPart` / `ChatResponseAnchorPart`. Any other streamed parts (progress, confirmations, tool calls, etc.) are invisible, so we can't tell whether:

- the stream produced nothing (backend stall), or
- the model answered in plain text / non-markdown parts (model-decision flake), or
- the request eventually errored after the timeout window.

### Change

Enrich the timeout error to dump:

1. `stream.items.length` and the constructor name of each part.
2. The markdown/anchor `currentProgress` (preserved from before).
3. After the timeout fires, give the in-flight request a 5 s grace period and report whether it ended up `resolved` / `rejected` / `still-pending` — including the underlying error if rejected.

The runtime path for the happy case is unchanged. No retries are added; this PR is purely diagnostic.

Worst-case duration on failure goes from 20 s → 25 s (still well under the suite's 60 s timeout).

### DIFF
--- a/extensions/copilot/src/extension/test/vscode-node/sanity.sanity-test.ts
+++ b/extensions/copilot/src/extension/test/vscode-node/sanity.sanity-test.ts
@@ -138,7 +138,13 @@ suite('Copilot Chat Sanity Test', function () {
 						getResultPromise.then(r => ({ kind: 'resolved' as const, value: r }), e => ({ kind: 'rejected' as const, error: e })),
 						timeout(5_000).then(() => ({ kind: 'still-pending' as const }))
 					]);
-					throw new Error(`${err instanceof Error ? err.message : String(err)} | follow-up: kind=${settled.kind}${'error' in settled ? ` error=${settled.error}` : ''} ${dumpStream()}`);
+					const cause = err instanceof Error ? err : new Error(String(err));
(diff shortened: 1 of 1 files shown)


roblourens (approved):
When this happens, it's usually due to an upstream hiccup. We should just run it with a retry. I don't know if you'll get anything useful from this

PR #317416, opened 2026-05-19
dot/ascii animations for terminals in chat

justschen:
https://github.com/user-attachments/assets/67b2849a-9ce0-471d-8889-acbed7a392ab

https://github.com/user-attachments/assets/c34b514c-99d5-4349-a1a9-f9c00e38c5db

cc @meganrogge @anthonykim1

### DIFF
--- a/src/vs/workbench/contrib/chat/browser/widget/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
+++ b/src/vs/workbench/contrib/chat/browser/widget/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
@@ -146,36 +146,9 @@ interface ITerminalCommandDecorationOptions {
 	getIsRunning(): boolean;
 }
 
-type SpinnerDotState = 'hidden' | 'entering' | 'visible' | 'falling';
-
 class TerminalCommandDecoration extends Disposable {
-	// 2x3 dot matrix: stagger each row left->right while building bottom->top,
-	// then stagger each row left->right while removing bottom->top.
-	private static readonly _spinnerFrames = [
-		[4],
-		[4, 5],
-		[2, 4, 5],
-		[2, 3, 4, 5],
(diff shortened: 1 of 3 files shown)


src/vs/workbench/contrib/chat/browser/widget/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
```diff
@@ -139,11 +139,43 @@ interface ITerminalCommandDecorationOptions {
...
+	 */
+	getIsRunning(): boolean;
 }
 
+type SpinnerDotState = 'hidden' | 'entering' | 'visible' | 'falling';
+
 class TerminalCommandDecoration extends Disposable {
+	// 2x3 dot matrix: stagger each row left->right while building bottom->top,
+	// then stagger each row left->right while removing bottom->top.
+	private static readonly _spinnerFrames = [
```
anthonykim1: whoa fancy

justschen: yaur it let's us create our indices to be rendered in that order. we can manipulate it how we see fit as well if we want it to be faster, slower, or some other kind of animation.

anthonykim1 (dismissed):

src/vs/workbench/contrib/chat/browser/widget/chatContentParts/toolInvocationParts/chatTerminalToolProgressPart.ts
justschen: swapped to using pure CSS which is a little safer. using the animation matrix drives the animation solely from TS, but we can do the same from CSS which is a bit easier to understand

anthonykim1 (approved):

PR #317428, opened 2026-05-20
Enable BYOK in air-gapped scenarios/without GitHub auth

### LINKED ISSUE #246551: Use copilot with local models completely offline
In the latest version of VSCode it is now possible to use local models (for example ollama) in the copilot (https://code.visualstudio.com/docs/copilot/language-models#_bring-your-own-language-model-key). However it still requires to login to GitHub to get started. Can this limitation be removed to allow use of copilot with local model completely offline?

### LINKED ISSUE #272843: BYOK internal server CAPI error 500
Use BYOK, OpenRouter. I used Qwen3 Coder 480B

Open VS Code repo. Agent mode Use the following prompt

"Now refactor the way survivalRate is reported, so it does not report rates, but absolute number of changes.
The removals are also reported. Make sure to propoagate this change through all the affacted files
Do not wait on compilation errors"

I get an error

```
Sorry, your request failed. Please try again.

Copilot Request id: 022731e9-49ae-4649-9ae5-57f6726a7764

Reason: {"message":"Internal Server Error","code":500}
```

I am surprised we send a Copilot Request ID for this BYOK flow.

fyi @joaomoreno 

Image

### LINKED ISSUE #275591: vscode.lm.selectChatModels call hangs when GitHub token is invalid

- VS Code Version: 1.105.1
- OS Version: Ubuntu 22.04
- Github Copilot extension Version: 1.388.0
- Github Copilot Chat extension version: 0.32.4

I reached a state in which I was receiving the following notification warning: "Your GitHub token is invalid. Please sign out from your GitHub account using the VS Code accounts menu and try again."

In this state, when trying to call `vscode.lm.selectChatModels({ vendor: 'copilot' })` inside my extension, it would hang (instead of maybe throwing a similar error).
If I sign into my Github Copilot account again, the problem is fixed. However, I would not expect the selectChatModels call to hang.

### LINKED ISSUE #313461: Feature Request: Local AI in Offline Networks
I would like to request a feature that allows the use of AI functionalities in an air-gapped or restricted network environment. Currently, our corporate intranet does not have access to the internet and cannot connect to github.com for authentication or external API calls.

We would like the ability to use AI features by connecting to a local Ollama instance hosted within our internal network. Even when the IDE or application is unable to log in to GitHub, it should provide an option to:

- Skip the external login/internet check.
- Manually configure a local API endpoint.
- Select and utilize models pulled via the local Ollama service.

Thank you for considering this request.

dmitrivMS:
Air-gapped BYOK: enable chat and some other AI functionality when user is not authenticated (with or without utility models enabled).

Fixes #246551
Fixes #275591
Fixes #313461
Fixes #272843

Related #294708

Fixes https://github.com/microsoft/vscode-internalbacklog/issues/7044

Related https://github.com/microsoft/vscode-internalbacklog/issues/5892


extensions/copilot/src/platform/embeddings/common/remoteEmbeddingsComputer.ts
```diff
@@ -64,6 +64,10 @@ export class RemoteEmbeddingsComputer implements IEmbeddingsComputer {
 		});
 		try {
 			return await logExecTime(this._logService, 'RemoteEmbeddingsComputer::computeEmbeddings', async () => {
+				// The remote embeddings endpoint requires GitHub authentication.
+				if (!this._authService.anyGitHubSession) {
+					return { type: embeddingType, values: [] };
+				}
```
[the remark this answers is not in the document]
dmitrivMS: fixed in follow-up commit

extensions/copilot/src/platform/embeddings/common/remoteEmbeddingsComputer.ts
```diff
@@ -64,6 +64,10 @@ export class RemoteEmbeddingsComputer implements IEmbeddingsComputer {
 		});
 		try {
 			return await logExecTime(this._logService, 'RemoteEmbeddingsComputer::computeEmbeddings', async () => {
+				// The remote embeddings endpoint requires GitHub authentication.
+				if (!this._authService.anyGitHubSession) {
+					return { type: embeddingType, values: [] };
```
vritant24, line 69: it looks like many places don't expect `computeEmbeddings` to return a non empty array (with the changes adding `if length == 0` checks). The inteface is also so:
```
/**
* Computes embeddings for the given strings.
*
* @param inputs The strings to compute embeddings for.
*
* @returns The embeddings, or if there is a failure/no embeddings, undefined.
*/
computeEmbeddings(
	type: EmbeddingType,
	inputs: readonly string[],
	options?: ComputeEmbeddingsOptions,
	telemetryInfo?: TelemetryCorrelationId,
	token?: CancellationToken,
): Promise<Embeddings>;
```
the return type comment says it should return `undefined` but the actual interface/implementation doesn't. 
Would it be better to change the interface to return `Promise<Embeddings | undefined>`, and return `undefined` if there is no embeddings model/github session? This way it would be more explicit at call sites when no embeddings are returned (in the future uses as well)

extensions/copilot/src/extension/conversation/vscode-node/conversationFeature.ts
```diff
@@ -190,6 +219,13 @@ export class ConversationFeature implements IExtensionContribution {
 		}
 
 		this._settingsSearchProviderRegistered = true;
+
+		// Don't register for no auth user or or BYOK-only users
```
vritant24, line 223: Not this change, but in the same file:
`registerParticipantDetectionProvider` can also be gated on the `anyGithub` session, specifically after seeing this line in it:
https://github.com/microsoft/vscode/blob/de7962ebeb8fbb52c794886d737da59a2f2d2370/extensions/copilot/src/extension/prompt/node/intentDetector.tsx#L59-L60
> 		// Disable intent detection if the user is requesting their request be completed with o1 since o1 has such a low RPS the cost of an incorrect intent is high
If we disable for O1, we should als disable for BYOK models? at least until we let the user granularly control its invokation

extensions/copilot/src/extension/contextKeys/vscode-node/contextKeys.contribution.ts
```diff
@@ -141,11 +141,12 @@ export class ContextKeysContribution extends Disposable {
 			const reason = e.message || e;
 			const data = TelemetryData.createAndMarkAsIssued({ reason });
 			this._telemetryService.sendGHTelemetryErrorEvent('activationFailed', data.properties, data.measurements);
-			const message =
-				reason === 'GitHubLoginFailed'
-					? SESSION_LOGIN_MESSAGE
-					: `GitHub Copilot could not connect to server. Extension activation failed: "${reason}"`;
-			this._logService.error(message);
+			if (reason === 'GitHubLoginFailed') {
+				// Expected in BYOK / air-gapped flows where the user is not signed in to GitHub.
```
vritant24: Not a result of this change, but is there any way the dependency on this type can be made more explicit? right now we're just checking for `GitHubLoginFailed` with a hard string check. If someone removes or changes the string in `TokenErrorReason`, this would fail.

extensions/copilot/src/platform/urlChunkSearch/node/urlChunkEmbeddingsIndex.ts
```diff
@@ -155,4 +158,4 @@ class SimpleUrlContentCache<T> {
 		const hash = await content.getContentHash();
 		this._cache.set(content.uri, { hash, value });
 	}
-}
\ No newline at end of file
+}
```
vritant24, line 161: nit: extra line added

extensions/copilot/src/platform/embeddings/common/remoteEmbeddingsComputer.ts
dmitrivMS: Good point, thought about that, but also - the problem is not all callers were checking if the return array was empty, and type checking for that is tricky. So I opted to protect them against the empty which would have to be done even if I added `undefined` as a return option.

extensions/copilot/src/extension/contextKeys/vscode-node/contextKeys.contribution.ts
dmitrivMS: Changed to use the constant (in a sort of weird way).

extensions/copilot/src/extension/conversation/vscode-node/conversationFeature.ts
dmitrivMS: Good point, added.

extensions/copilot/src/platform/urlChunkSearch/node/urlChunkEmbeddingsIndex.ts
dmitrivMS: If I try to edit, it is adding it back, so will keep as is.

extensions/copilot/src/extension/completions/vscode-node/completionsCoreContribution.ts
```diff
@@ -31,9 +31,14 @@ export class CompletionsCoreContribution extends Disposable {
...
 			const extensionUnification = unificationStateValue?.extensionUnification ?? false;
+			const copilotToken = this._copilotToken.read(reader);
 
 			let hasInstantiatedProvider = false;
-			if (unificationStateValue?.codeUnification || extensionUnification || configEnabled || this._copilotToken.read(reader)?.isNoAuthUser) {
+			// Completions require a Copilot token to call the completions endpoint, so don't
+			// register the provider in air-gapped / signed-out scenarios — it would just fail
+			// with GitHubLoginFailedError on every keystroke.
+			const wantsProvider = unificationStateValue?.codeUnification || extensionUnification || configEnabled || copilotToken?.isNoAuthUser;
+			if (wantsProvider && copilotToken) {
```
vritant24, line 41: Will this run if vscode is started, and then the user signs in later? (i.e does it get re-triggered on sign in?)

vritant24 (dismissed):

extensions/copilot/src/extension/completions/vscode-node/completionsCoreContribution.ts
dmitrivMS: yeah, it's observable stuff - basically if token changes, it will rerun

vritant24 (approved):

agoswami84:
how to use this feature

cf-rdegregory:
> how to use this feature

1. Update to the latest insiders build 
image

2. Log out of GitHub in Insiders.
3. Configure a BYOK model provider in `chatLanguageModels.json`.

digitarald:
Cross-posting from https://github.com/microsoft/vscode/issues/246551#issuecomment-4510312517

> @dmitrivMS just landed this yesterday and **today its ready for testing and your feedback in Insiders** ([download here](https://code.visualstudio.com/insiders), can be used side by side with your stable VS Code).
> 
> **Start:**
> 1. Enable `chat.allowAnonymousAccess` and restart VS Code (Command: `Developer: Reload Window`)
> 2. Make sure you have your favorite BYOK providers added, local or remote
> 3. **Done:** Pick your BYOK model in VS Code chat and say Hi; no GitHub Copilot login needed.
> 
> **Bonus**: Set your BYOK model for `chat.utilityModel` and `chat.utilitySmallModel`, [more details in our release notes](https://code.visualstudio.com/updates/v1_121#_configure-utility-models) on what they are used for.
> 
> **Please help us test**. File issues and/or comment if you had a chance to try this out.
> 
> Screenshot from my testing with OpenAI as BYOK provider:
> Image

PR #317465, opened 2026-05-20
Drop wheel events queued before window focus

### LINKED ISSUE #28795: Editor: scroll jumps randomly (after Find?)

boiledbrian:
Fixes #28795

This filters wheel events whose timestamp predates the latest window focus before they reach the editor mouse wheel handler or the shared scrollable element handler. Those events can be queued while the window is unfocused and then replayed after focus returns, causing an unexpected editor scroll jump.

The filter keeps a small grace window around the focus timestamp so normal wheel input immediately after refocus is not discarded. It also handles both epoch-style and relative DOM event timestamps.

Verification:
- npm run -s compile-check-ts-native
- npm run -s eslint -- src/vs/base/browser/mouseEvent.ts src/vs/base/browser/ui/scrollbar/scrollableElement.ts src/vs/editor/browser/controller/mouseHandler.ts src/vs/base/test/browser/mouseEvent.test.ts
- npm run test-browser-no-install -- --run src/vs/base/test/browser/mouseEvent.test.ts --browser chromium
- npm run gulp -- compile-client
- npm run -s precommit

### DIFF
--- a/src/vs/base/browser/mouseEvent.ts
+++ b/src/vs/base/browser/mouseEvent.ts
@@ -141,7 +141,7 @@ function getLastWindowFocusTime(e: IMouseWheelEvent): number {
 	let lastFocusTime = lastWindowFocusTime.get(targetWindow);
 
 	if (typeof lastFocusTime === 'undefined') {
-		lastFocusTime = 0;
+		lastFocusTime = targetWindow.document.hasFocus() ? Date.now() : 0;
 		lastWindowFocusTime.set(targetWindow, lastFocusTime);
 		targetWindow.addEventListener('focus', () => lastWindowFocusTime.set(targetWindow, Date.now()), true);
 	}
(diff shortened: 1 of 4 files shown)


boiledbrian:

PR #317600, opened 2026-05-20
fix(arrays): clean up duplicate logic and improve array helpers

mohityadav8:
## Summary

This PR improves several utilities in `src/vs/base/common/arrays.ts`:

- removes redundant assignment logic in `range()`
- updates `withoutDuplicates()` to reuse `distinct()` instead of duplicating logic, while preserving API compatibility
- replaces `[].concat(...arrays)` in `concatArrays()` with an iterative implementation to avoid spread argument limits and improve type safety

## Testing

- Verified the diff only affects the intended three utility functions in `arrays.ts`
- No public API was removed

## Notes

I intentionally did not remove deprecated exports (`move`, `insert`, `remove`) since that would be a breaking API change and is better handled separately.


mohityadav8:

mohityadav8:
cc @mjbvz :D

mohityadav8:
cc @mjbvz

dmitrivMS:
There is no issue linked, no tests, so this looks like unnecessary churn.

PR #317637, opened 2026-05-20
Fix git extension not working when repo root is "/" (chroot)

vanklompf:
## Problem

When working in a chroot environment where the project/repository root is `/`, the git extension fails silently — discard changes, diffs, and staging all break.

**Root cause:** The `normalizePath()` function in `extensions/git/src/util.ts` unconditionally strips trailing path separators. When the path is `/` (root), stripping the trailing `/` produces an empty string `""`, which `path.normalize("")` converts to `"."`. This breaks both `isDescendant()` and `pathEquals()`:

```javascript
// Before fix:
normalizePath("/")  → "."     // WRONG — should be "/"
isDescendant("/", "/foo")  → false  // WRONG — should be true
```

As a result, the git extension cannot match files to their repository when the repository root is `/`.

## Fix

The fix now uses path !== dirname(path). This correctly detects all filesystem roots.

```javascript
// After fix:
normalizePath("/")  → "/"     // Correct
isDescendant("/", "/foo")  → true  // Correct
```

This is consistent with how the core VS Code `removeTrailingPathSeparator()` in `src/vs/base/common/extpath.ts` already handles this case.

## Testing

- Added unit tests for `isDescendant`, `pathEquals`, and `relativePath` covering the root path `/` scenario.
- Verified compilation passes with 0 errors.
- Verified old behavior was broken and new behavior is correct.


vanklompf:
Fixes #298003
https://github.com/microsoft/vscode/issues/298003

PR #317654, opened 2026-05-20
sessions: retry restore when provider session list changes

connor4312:
sessions: retry restore when provider session list changes

Fixes a hang where reloading a window with an agent host session active leaves the
Agents view stuck in a loading state.

- restoreLastActiveSession only re-ran its tryRestore loop on
  sessionsProvidersService.onDidChangeProviders. Providers (notably the local agent
  host) are registered once at startup and then populate their session cache
  asynchronously after authenticationPending settles, firing onDidChangeSessions but
  not onDidChangeProviders. The restore promise therefore never resolved.
- Also subscribe to this.onDidChangeSessions so the retry fires when a provider's
  session list changes after registration.
- Adds a regression test that pre-seeds the active-session state, starts restore with
  an empty provider cache, then fires onDidChangeSessions and asserts the session opens.


DonJayamanne (approved):

sandy081:
@connor4312  Thanks for the fix here

PR #317692, opened 2026-05-21
feat: enhance session customization discovery and agent picker with improved file classification and sorting

DonJayamanne:


DonJayamanne:
@copilot fix the test failures
The logic is different now, we must match the exact case `README.md`, `readme.md` will not be ignored anymore.

image

vijayupadya (approved):

PR #317700, opened 2026-05-21
Enhance session customization handling and simplify sync logic

DonJayamanne:
Improve session customization management and state synchronization across agents. This update ensures that customizations are immediately available upon session creation, enhancing user experience. Additionally, refactor the synchronization logic for better clarity and efficiency.

### DIFF
--- a/src/vs/platform/agentHost/node/agentService.ts
+++ b/src/vs/platform/agentHost/node/agentService.ts
@@ -496,7 +496,11 @@ export class AgentService extends Disposable implements IAgentService {
 				if (customizations.length === 0) {
 					return;
 				}
-				this._stateManager.dispatchServerAction(session.toString(), {
+				const sessionKey = session.toString();
+				if (!this._stateManager.getSessionState(sessionKey)) {
+					return;
+				}
+				this._stateManager.dispatchServerAction(sessionKey, {
 					type: ActionType.SessionCustomizationsChanged,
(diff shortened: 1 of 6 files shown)


src/vs/platform/agentHost/node/agentService.ts
```diff
@@ -482,6 +482,33 @@ export class AgentService extends Disposable implements IAgentService {
...
+					return;
+				}
+				this._stateManager.dispatchServerAction(sessionKey, {
+					type: ActionType.SessionCustomizationsChanged,
+					customizations: [...customizations],
+				});
+			}).catch(err => {
+				this._logService.error('[AgentService] createSession: failed to publish initial customizations', err);
+			});
+		}
```
connor4312: I don't think this is the right approach here. I would expect that after `this._stateManager.createSession`, we update the resulting session with the instantaneous, current state of any customizations we have. And then after that point any further syncing state updates can happen via the existing ActionType.SessionCustomizationsChanged/ActionType.SessionCustomizationUpdated publishing from the PluginController

bryanchen-d (dismissed):

src/vs/platform/agentHost/node/agentService.ts
DonJayamanne: oooh nice, thanks for the pointer

connor4312 (approved):

PR #317776, opened 2026-05-21
Allow sending chat requests with attachments only

### LINKED ISSUE #317775: Allow sending chat requests with attachments only
When using chat, attaching an image or file currently requires typing text before the request can be sent.

This makes attachment-only prompts awkward, especially for screenshots or files where the attachment itself is the complete request context. 

Expected behavior:
- Users can send a chat request with only explicit attachments, such as images, files, etc.
- The UI should not render an empty-looking request.
- For attachment-only requests, the user-visible request text should be a localized summary such as:
  - `Attached 1 image`
  - `Attached 2 images`
  - `Attached 1 file`
  - `Attached 2 files`
- Or find a better way to do that  

Attached photo: attachment selected, but send is blocked until text is typed.

Image

federicobrancasi:
Fixes #317775

This change allows chat requests to be sent with explicit attachments even when the text input is empty.

The work is split into two commits:

1. `Allow sending chat requests with attachments only`

   Enables attachment-only sends for explicit file, directory, and image attachments.

   This updates the chat input send gating and service-side empty-message validation so attachment-only requests are accepted, while hidden or automatic context such as workspace/implicit context does not make an empty request sendable.

   It also adds localized fallback request summaries for the UI and accessibility paths, so attachment-only requests do not appear blank. Examples:
   - `Attached 1 image`
   - `Attached 2 images`
   - `Attached 1 file`
   - `Attached 2 files`

2. `Use attachment summary for empty chat request text` (I don't know if necessary)

   Uses the same attachment summary as the actual request text for attachment-only sends.

   This means the user sees `Attached 1 image` / `Attached N files`, and the model also receives that summary as the request message alongside the attached context.

Videos: attachment-only send works and displays the attachment summary (Both VSCode and Agents app)

https://github.com/user-attachments/assets/a1292265-65d6-4b77-bd1a-74c21c88cd82

https://github.com/user-attachments/assets/c06926e2-ed52-4eb8-8a6f-60d73fe60a2c

### DIFF
--- a/src/vs/sessions/contrib/chat/browser/newChatInput.ts
+++ b/src/vs/sessions/contrib/chat/browser/newChatInput.ts
@@ -48,7 +48,7 @@ import { SlashCommandHandler } from './slashCommands.js';
 import { VariableCompletionHandler } from './variableCompletions.js';
 import { AgentHostInputCompletionHandler } from './agentHostInputCompletions.js';
 import { IChatModelInputState } from '../../../../workbench/contrib/chat/common/model/chatModel.js';
-import { IChatRequestVariableEntry, isExplicitFileOrImageVariableEntry } from '../../../../workbench/contrib/chat/common/attachments/chatVariableEntries.js';
+import { getExplicitFileOrImageAttachmentSummary, IChatRequestVariableEntry, isExplicitFileOrImageVariableEntry } from '../../../../workbench/contrib/chat/common/attachments/chatVariableEntries.js';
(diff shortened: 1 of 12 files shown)


federicobrancasi:
I am also exploring better UI options for attachment-only chat messages, so we may not need to show literal text like `Attached 1 image` long term.

For this PR, I am using that summary because otherwise the request appears as an empty chat bubble. This keeps the UI understandable and gives the model a non-empty request message alongside the attachment context.

dbaeumer (dismissed):

mrleemurray (approved):
Design looks good

Issue #315699, opened 2026-05-11
Bring vscode-agents experience with worktrees to vscode


eddyw:
Hi

Currently, in vscode, we have Git worktrees integration. However, they are very far from being useful (In my personal experience and opinion) because:

You can see worktrees in the source control but the explorer view does not show the selected worktree files, it only shows files from main.

In the Agents app (vscode agents app), you have a session tied to a worktree, clicking a session switches everything to that worktree. Your files, terminal, and source control views are all from that worktree.

In vscode the closest thing is either open the worktree in the current window, or open in a new window (or a new tab in MacOS), or - the worst option - creating the worktrees inside the current directory (like ".claude/worktrees" for example) where now search and go to file view will show every file from every single worktree.

## The feature request

Please, I want a view in vscode where I can simply click on a worktree once, and without any confirmation dialog, switch everything my files (Explorer view), terminal, etc to it. Just like the Agents app does, but not necessarily linked to an AI session. For inspiration, Zed IDE got this right.

## Workaround

On MacOS, since you have tabs, you can always open the worktree in a new window (tab), but doing this is horrible, and it seems you solved this problem in the Agents app.

Issue #314677, opened 2026-05-06
Worktree UI is very inconsistent and confusing


nanbiz:
I am checked out in a worktree and edit something that's only on the dev worktree:
<img width="1495" height="1102" alt="Image" src="https://github.com/user-attachments/assets/2a294acf-b27c-451c-8920-97f267e85460" />

When I now click on a different worktree, everything changes except the files:

<img width="1448" height="1107" alt="Image" src="https://github.com/user-attachments/assets/ef7551dd-11de-4fef-b76d-4dbf551fe731" />
There is absolutely nothing that shows me what branch I'm actually on right now. Everything tells me I'm on project f redesign - but I'm not. 
Even the reference at the bottom left, which I as a user usually take as a clear indication on what branch I'm actually on when changing branches in the Graph view just lies to me.
How is a user to understand this? I think it's super unintuitive.

Issue #311858, opened 2026-04-22
Improve multi-worktree workflows (grouping terminals, editors, and agents)


XavierGeerinck:
Hi All!

> I searched for existing issues but couldn't find the correct issue that seems to bundle the features as per below.

I love VS Code and have been using it for years, but currently with the new Agentic way of working (e.g., Superset.sh, Codex, ...) I am missing more and more the ability to work in parallel on different git worktrees allowing me to work on multiple contexts at the same time.

Currently, when you have a worktrees open, things seem to get messy:
* no clear grouping of terminals/editors per worktree
* hard to see what’s running where (especially with agents)
* switching between contexts loses mental state
* reviewing changes across worktrees is fragmented

There are already related issues pointing at this:

* [#257396](https://github.com/microsoft/vscode/issues/257396) — confusion when using multiple worktrees in a workspace

VS Code does support worktrees and parallel development, but they currently show up as **separate repos rather than a cohesive workflow** ([[Visual Studio Code](https://code.visualstudio.com/docs/sourcecontrol/branches-worktrees)][3])

Therefore, I would like to see how the VS Code community is thinking of building a UX that supports:

* multiple isolated worktrees
* multiple terminals per context that automatically open in the right directory
* parallel agents
* a unified “what’s running where” view

In short, I think it would help a lot if VS Code had a lightweight way to treat a worktree + its terminals + agent sessions as a single “context” (even just at the UI level). This would make active contexts easier to see/switch and improve multi-worktree reviews, making parallel dev workflows much smoother.

timodwhit:
@XavierGeerinck the agents view in preview does seem to have a lot of what you are discussing, unless I am misunderstanding the post.

PR #317806, opened 2026-05-21
worktrees: add command-center switcher and per-worktree state management

eddyw:
## TL;DR

https://github.com/user-attachments/assets/f4e6be04-9940-4d24-9e6d-a0b8a1ae389b

## Summary

Brings the worktree-switching experience from the Agents app into the regular workbench, decoupled from the AI session lifecycle.

In the Agents app, each session starts a git worktree and clicking another session swaps the workbench onto that worktree without spawning a new window. This PR exposes the same mechanic directly: a worktree switcher in the command center that lists the current repository's worktrees and switches between them on click. No session required.

## Why?

- #315699
- #314677
- #311858

The placement inside the command center is open — there may be a better home for the switcher long-term. The point is that the regular workbench can benefit from the same experience the Agents app already has, and as a side effect **Chat sessions end up naturally scoped to the active worktree without any extra plumbing**.

This is a working proof of concept. Switching delegates to existing primitives — workspace folder swap for the explorer/SCM rebase, and the git extension's own commands for creating and deleting worktrees — so the surface area is small.

### DIFF
--- a/build/lib/i18n.resources.json
+++ b/build/lib/i18n.resources.json
@@ -283,6 +283,10 @@
 			"project": "vscode-workbench"
 		},
 		{
+			"name": "vs/workbench/contrib/worktrees",
+			"project": "vscode-workbench"
+		},
+		{
 			"name": "vs/workbench/contrib/customEditor",
 			"project": "vscode-workbench"
 		},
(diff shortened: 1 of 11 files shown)


eddyw:

nanbiz:
I like this a lot! This might finally allow me to switch my colleagues to worktrees!

PR #317846, opened 2026-05-21
fix: memory leak in ipc.electron.ts

SimonSiefke:
Fixes a memory leak in ipc.electron.ts

### Details

When a client connects, it is stored in `Server.Clients`. But it seems when a client disconnects, its entry from `Server.Clients` isn't removed.

```ts
const onDidClientReconnect = new Emitter<void>();
Server.Clients.set(id, toDisposable(() => onDidClientReconnect.fire()));
```

### Change

The changes tries to ensure when a client disconnects, its entry from `Server.Clients` is removed as well.

```ts
const onDidClientReconnect = new Emitter<void>();
Server.Clients.set(id, toDisposable(() => onDidClientReconnect.fire()));

/* Ensure to remove entry from Server.Clients on disconnect  */
Event.once(onDidClientDisconnect)(() => {
	if (Server.Clients.get(id) === reconnectDisposable) {
		Server.Clients.delete(id);
	}
	onDidClientReconnect.dispose();
});
```

### Before

When opening and closing a new window 17 times, the number of functions in ipc.electron.ts seems to grow by one each time: 

window-open-new 0

### After

When opening and closing a new window 17 times, no more leak in ipc.electron.ts is detected *(if you look closely at the charts there is one less row)*:

window-open-new 5


dmitrivMS (approved):

dmitrivMS:
@SimonSiefke Thank you!

bhavyaus (approved):

PR #317856, opened 2026-05-21
agent-host: stop config pickers flashing during resolveSessionConfig

connor4312:
When a user changes a session-config picker (isolation, branch, autoApprove, mode, claude permission mode) on an agent-host new session, all schema-driven pickers visually disappear and reappear while the async `resolveSessionConfig` round-trip runs.

## Root cause

`NewSession.setConfigValue` wiped the cached schema to `{ properties: {} }` during the optimistic update, so every well-known mode / autoApprove guard returned `false` and hid its picker until the new schema arrived ~200–500 ms later.

## Fix

- Preserve the existing schema in \`NewSession.setConfigValue\`.
- Introduce an observable \`isResolvingConfig\` on \`NewSession\`, owned by \`resolveConfig\`'s \`finally\`, with \`begin\`/\`endResolveConfigSync\` helpers for the synchronous-set-before-event path and the no-connection early-return path.
- Expose \`IObservable<boolean> isSessionConfigResolving(sessionId)\` on \`IAgentHostSessionsProvider\`; \`constObservable(false)\` for any session that isn't the in-flight new session.
- Distinct from \`session.loading\`: the latter also stays \`true\` while config is complete-but-required-values-missing, where pickers must remain interactive.

Every picker that mutates session config now disables on this observable: generic per-property chips, mode / claude permission mode enum pickers, the autoApprove permission action item, the mobile bottom-sheet mode picker, and the mobile combined Mode + Model new-session chip.

The autoApprove permission picker also gates the delegate's \`setPermissionLevel\` on the same observable, because \`ActionWidgetDropdown\` opens via Enter/Space directly on its label and CSS \`pointer-events: none\` does not block keyboard activation.

A defense-in-depth bail in \`setSessionConfigValue\` drops second-arrival changes on a new session while a resolve is already in flight.

## Review

Reviewed by a multi-model council (Claude Opus 4.6, GPT-5.5, GPT-5.3-Codex). Three real issues surfaced by GPT-5.5 — keyboard activation bypass on the permission picker, mobile mode picker bypassing the new guard, and the resolving flag leaking on the no-connection path — were all addressed in this change set.

## Follow-up

Pure-dispatch new-session-only methods on \`IAgentHostSessionsProvider\` (\`getCreateSessionConfig\`, \`clearSessionConfig\`, \`getSessionConfigCompletions\`, \`isSessionConfigResolving\`) will be moved onto a new \`IAgentHostNewSession\` interface exposed as an observable handle in a follow-up PR.

### DIFF
--- a/src/vs/sessions/common/agentHostSessionsProvider.ts
+++ b/src/vs/sessions/common/agentHostSessionsProvider.ts
@@ -45,6 +45,12 @@ export interface IAgentHostSessionsProvider extends ISessionsProvider {
 	readonly onDidChangeSessionConfig: Event<string>;
 	/** Returns the last resolved dynamic configuration for a session. */
 	getSessionConfig(sessionId: string): ResolveSessionConfigResult | undefined;
+	/**
+	 * Observable: `true` while a `resolveSessionConfig` round-trip is in
+	 * flight. Pickers gate on this rather than `session.loading` so they
+	 * stay interactive in the required-values-missing state.
+	 */
+	isSessionConfigResolving(sessionId: string): IObservable<boolean>;
 	/** Sets one dynamic configuration property and re-resolves the schema. */
 	setSessionConfigValue(sessionId: string, property: string, value: unknown): Promise<void>;
 	/**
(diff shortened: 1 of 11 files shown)


anthonykim1 (dismissed):

roblourens (approved):

roblourens:
Thanks

PR #318040, opened 2026-05-22
Agents - agent host changesets

lszomoru:

### DIFF
--- a/src/vs/sessions/contrib/providers/agentHost/browser/agentHostSessionChangesets.ts
+++ b/src/vs/sessions/contrib/providers/agentHost/browser/agentHostSessionChangesets.ts
@@ -1,0 +1,96 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { Disposable } from '../../../../../base/common/lifecycle.js';
+import { constObservable, derived, derivedOpts, IObservable, observableFromEvent } from '../../../../../base/common/observable.js';
+import { isEqual } from '../../../../../base/common/resources.js';
+import { URI } from '../../../../../base/common/uri.js';
+import { ChangesetSummary, StateComponents } from '../../../../../platform/agentHost/common/state/sessionState.js';
+import { ISessionChangeset, ISessionFileChange, sessionFileChangesEqual } from '../../../../services/sessions/common/session.js';
+import { ISessionsManagementService } from '../../../../services/sessions/common/sessionsManagement.js';
+import { changesetFilesToChanges } from './agentHostDiffs.js';
+import { IAgentHostAdapterOptions } from './baseAgentHostSessionsProvider.js';
+
+export function createChangesets(
+	resource: URI,
+	options: IAgentHostAdapterOptions,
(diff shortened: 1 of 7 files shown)


roblourens (approved):

src/vs/sessions/contrib/providers/agentHost/browser/agentHostSessionChangesets.ts
```diff
@@ -0,0 +1,110 @@
...
+			if (!activeSession || !isEqual(activeSession.resource, resource)) {
+				return constObservable(undefined);
+			}
+
+			const connection = options.getConnection();
+			if (!connection) {
+				return constObservable(undefined);
+			}
+
+			const subscriptionRef = connection.getSubscription(StateComponents.Changeset, URI.parse(changesetSummary.uriTemplate));
```
roblourens, line 76: I guess it feels like we should give you a service where you can do `subscribe(sessionUri, ...)` rather than passing a connection around. But this is probably fine for now.

src/vs/sessions/contrib/providers/agentHost/browser/agentHostSessionChangesets.ts
```diff
@@ -0,0 +1,110 @@
...
+			if (changesetState.status !== ChangesetStatus.Ready) {
+				return [];
+			}
+
+			const files = changesetFilesToChanges(changesetState.files);
+			return options.mapDiffUri
+				? files.map(f => ({
+					...f,
+					uri: options.mapDiffUri!(f.uri),
+					originalUri: f.originalUri ? options.mapDiffUri!(f.originalUri) : undefined,
```
roblourens, line 104: Here, I think these URIs should be already mapped, rather than working with raw AHP URI that have to be transformed at this point (probably goes along with the first comment)

src/vs/sessions/contrib/providers/agentHost/browser/agentHostSessionChangesets.ts
```diff
@@ -0,0 +1,110 @@
...
+		this.description = changesetSummary.description;
+
+		this.isEnabled = constObservable(true);
+		this.isDefault = constObservable(changesetSummary.isDefault);
+
+		this.originalCheckpointRef = constObservable(undefined);
+		this.modifiedCheckpointRef = constObservable(undefined);
+
+		const changesetStateObs = derived(reader => {
+			const activeSession = this._sessionsManagementService.activeSession.read(reader);
```
roblourens, line 66: Unsure about this, seems wrong to read `activeSession` at this level?

src/vs/sessions/contrib/providers/agentHost/browser/agentHostSessionChangesets.ts
lszomoru: @roblourens, that would be awesome as I would expect that there will be many places in the client where we need to have a way to easily subscribe and unsubscribe to events

src/vs/sessions/contrib/providers/agentHost/browser/agentHostSessionChangesets.ts
lszomoru: Maybe? I think that there will be multiple places on the client where we will subscribe/unsubscribe for the active session. The other options is to have this at the provider or adapter level but I am not yet sure where we should have it. Let me think a bit more about this and see what would fit best.

PR #318045, opened 2026-05-22
Optimize shared built-in extension dependencies

dmitrivMS:
## Summary

Reduce duplicated JS shipped by built-in extensions by externalizing a set of shared production dependencies and resolving them once from a single product-level `extensions/node_modules/`. Adds a postinstall validator that keeps consumer extensions aligned with the shared versions. Also cleans up a handful of orphaned dependencies discovered along the way and fixes a pre-existing Windows-only build bug in `html-language-features`.

## Mechanism

- Shared deps are declared in [`extensions/package.json`](extensions/package.json) (the existing "Dependencies shared by all extensions" manifest). They get installed into `extensions/node_modules/` and the existing `getProductionDependencies('extensions/')` step in [`build/lib/extensions.ts`](build/lib/extensions.ts) copies them into the shipped desktop product.
- [`extensions/esbuild-extension-common.mts`](extensions/esbuild-extension-common.mts) reads `extensions/package.json` at build time and marks every runtime dep as `external` **for `platform: 'node'` only** (with a small `BUILD_ONLY_SHARED_DEPS = { typescript }` exclusion for tooling-only entries). The external list updates automatically when `extensions/package.json` changes — no separate maintenance.
- Web bundles continue to inline these deps. The web packaging path doesn't ship `extensions/node_modules/`, and the web extension host can't resolve from there anyway.
- [`extensions/esbuild-common.mts`](extensions/esbuild-common.mts) merges `external` arrays from `additionalOptions` with the base list instead of letting the per-extension override replace it. Without this, every extension that sets `additionalOptions.external` (git, html, css, microsoft-authentication, github, etc.) would have silently bypassed the shared list and continued bundling the deps inline.
- `extensions/copilot/.esbuild.mts` mirrors the relevant subset for its own node bundle.
- [`extensions/postinstall.mjs`](extensions/postinstall.mjs) validates that every consumer extension's `package.json` declares the **exact** same version range as `extensions/package.json` for shared deps. Drift fails the install with a list of offending entries. This catches the silent runtime failure that would otherwise happen (npm would install a private copy under the consumer at a different version, and esbuild's externalization would resolve to the wrong major from the shared location).
- New "Shared dependencies" section in [`extensions/CONTRIBUTING.md`](extensions/CONTRIBUTING.md) documents the bump workflow.

## Packages now shared

| Package | Consumers (production) |
|---|---|
| `@vscode/extension-telemetry` (+ transitive `@microsoft/1ds-*`, `@microsoft/applicationinsights-*`, `@microsoft/dynamicproto-js`, `@nevware21/*`) | 11 |
| `jsonc-parser` | 5 |
| `vscode-uri` | 6 |
| `markdown-it` | 3 |
| `dompurify` | 3 |
| `minimatch` | 2 |
| `vscode-tas-client` | 4 |
| `picomatch` | 2 |
| `vscode-languageserver-textdocument` | 2 |
| `request-light` | 2 |
| `which` | 3 |
| `tunnel` | 2 |
| `@octokit/rest` | 2 |

`vscode-languageclient` was considered but excluded: 3 of 4 consumers are on `^10.0.0-next.*` while `markdown-language-features` is pinned to `^8.0.2` for API-compat reasons. Each consumer keeps its own copy as before.

## Source-level changes

- **Version alignment to share single copies.** All consumer ranges now match `extensions/package.json` exactly so the postinstall validator passes: `@vscode/extension-telemetry` to `^1.5.1` (10 extensions); `markdown-it` to `^14.1.1` (2); `vscode-uri` to `^3.1.0` (4); `jsonc-parser` to `^3.3.1` (5); `vscode-languageserver-textdocument` to `^1.0.11` (1); `which` to `^4.0.0` (2); `minimatch` to `^10.2.1` and `request-light` to `^0.8.0` (npm); `@octokit/rest` to `^21.1.1` (github); `picomatch` to `^2.3.2` (git).
- **Import migrations.** `@vscode/extension-telemetry@1.x` removed the default export; updated 13 source files across 9 extensions to use named imports (`import { TelemetryReporter } from '@vscode/extension-telemetry'`). `minimatch@10` likewise dropped the default callable export — `extensions/npm/src/tasks.ts` now uses `import { minimatch } from 'minimatch'`.

## Cleanups

- Removed orphaned production deps (no source references, no transitive consumers): `node-fetch` + `@types/node-fetch` from `github-authentication` (drops 25 transitive packages), `vscode-uri@2` from `git`, `punycode` from `markdown-language-features`, `@xterm/headless` and `diff` from `copilot`, plus `jsonc-parser` from `vscode-colorize-perf-tests` and empty `dependencies: {}` fields in `simple-browser` / `vscode-colorize-perf-tests`.
- Fixed a Windows-only build failure in [`extensions/html-language-features/esbuild.browser.mts`](extensions/html-language-features/esbuild.browser.mts): replaced `import.meta.resolve('typescript').replace('file://', '')` (produces `/C:/…` invalid path on Windows) with `fileURLToPath(import.meta.resolve('typescript'))`.

## Potential wins (estimates, to be confirmed with a heap-snapshot pair)

- **~13 MB of duplicated JS source** removed from the shipped extension host (1DS dedupe ≈ 9 MB + the other 12 shared packages ≈ 4 MB, after subtracting the `vscode-languageclient` row that was reverted).
- **V8 bytecode amplification** typically adds 1–2× the source size, so extension-host RSS should drop on the order of **25–35 MB** in steady state.
- **Smaller .vsix downloads / updates** for the affected built-in extensions — same ballpark on disk.
- Faster extension-host warm-up, since V8 has proportionally less code to parse + compile.
- Removing `node-fetch` from `github-authentication` and `@xterm/headless` + `diff` from `copilot` also shrinks those specific .vsix files modestly.


extensions/extension-editing/package.json
```diff
@@ -27,7 +27,7 @@
   },
   "dependencies": {
     "jsonc-parser": "^3.2.0",
-    "markdown-it": "^12.3.2",
+    "markdown-it": "^14.1.1",
```
mjbvz, line 30: Any way we can document that if you update this package you actually need update the root one and keep them in sync?

extensions/esbuild-extension-common.mts
```diff
@@ -30,6 +30,27 @@ function resolveBaseOptions(config: ExtensionRunConfig): esbuild.BuildOptions {
 
 	if (config.platform === 'node') {
 		options.mainFields = ['module', 'main'];
+		// Resolved at runtime from the shared `extensions/node_modules/` (see
+		// `extensions/package.json` + the `getProductionDependencies('extensions/')`
+		// copy in `build/lib/extensions.ts`). Web builds inline these — the web
+		// packaging path doesn't ship `extensions/node_modules/`.
+		options.external = [...options.external!,
```
mjbvz: probably should be `...(options.external ?? [])` to be safe

mjbvz: Also is this something we can auto generate from the package.json?

mjbvz (commented):
Looks great! Excited for this change. Just wondering if we can add some tools/checks to make sure things don't get out of sync, plus maybe some comments so we know where to make package.json changes

extensions/extension-editing/package.json
dmitrivMS: Added a postinstall validator so that if it changes, it breaks the build.
Comment would be fragile and besides it's package.json, not sure if comment syntax is exactly allowed.

extensions/esbuild-extension-common.mts
dmitrivMS: Done both.

dmitrivMS:
> Looks great! Excited for this change. Just wondering if we can add some tools/checks to make sure things don't get out of sync, plus maybe some comments so we know where to make package.json changes

Thanks! I added validation (so preinstall will break if there is a mismatch). The error will point to CONTRIBUTING.md which documents the whole thing.

dbaeumer (approved):

PR #318077, opened 2026-05-23
Exclude MXC bin from VersionInfo sanity check

dileepyavan:
This updates the sanity signing checks to exempt files under the MXC SDK `bin` subtree from VersionInfo ProductName validation. MXC SDK binaries under `bin` are signed, but they do not include a ProductName VersionInfo resource.

The existing MXC catalog manifest Authenticode exemption is also broadened to cover packaged paths under `node_modules.asar.unpacked`.

Validation:
- `npm --prefix test/sanity run compile`
- precommit hygiene


bhavyaus (dismissed):

dmitrivMS:
@dileepyavan Can we get these binaries fixed so they have product name and version? When these things are missing it breaks some corp environments where binaries without signing or version are blocked and cause scan violations.

dmitrivMS (approved):

PR #318091, opened 2026-05-23
Keep last used permission mode when starting new Claude chats

### LINKED ISSUE #312774: New Claude chat resets the permission setting
Testing #312653

- new VS Code window on an empty folder and on a new profile
- open chat, select `Claude Code` and `Ask before edits`
- do some prompting
- open chat (with the + in the tool bar). 
- New chat opens with Claude (good) but has switched the permissions back to `Edit automatically'
- I really don't want Claude to auto permit. So IMO the best is to use the last used permission settings, or have a setting like `chat.permissions.default`

Chulong-Li:
Fixes #312774

## Summary
- New Claude chats now keep the user's last selected permission mode instead of resetting to default "Edit automatically".
- This preserves safer choices like "Ask before edits" across the new-chat flow.
- Permission mode updates from both the picker and the first-send path are remembered for the next Claude chat.

## Validation
Figure 1: Start a Claude chat with the “Ask before edits” permission mode selected instead of the default “Edit automatically”
image

<br>
Figure 2: Start a new Claude chat, and the previously selected “Ask before edits” permission mode is preserved
image

### DIFF
--- a/extensions/copilot/src/extension/chatSessions/vscode-node/claudeChatSessionContentProvider.ts
+++ b/extensions/copilot/src/extension/chatSessions/vscode-node/claudeChatSessionContentProvider.ts
@@ -100,6 +100,7 @@ export class ClaudeChatSessionContentProvider extends Disposable implements vsco
 		for (const update of updates) {
 			const value = update.value;
 			if (update.optionId === PERMISSION_MODE_OPTION_ID && value && isPermissionMode(value)) {
+				this._controller.rememberPermissionMode(value);
 				this.sessionStateService.setPermissionModeForSession(sessionId, value);
 			}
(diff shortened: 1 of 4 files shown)


aelij:
Hi, I didn't notice this PR when I tried to implement it (#323846).
Note that in my PR the setting persisted, not just in memory, which I think is the right behavior since that how all the other selections behave. See for example for Copilot CLI:

https://github.com/microsoft/vscode/blob/7ad5744c6852a42e070b6d6045e3e1215cc120fd/extensions/copilot/src/extension/chatSessions/copilotcli/vscode-node/copilotCLIChatSessionInitializer.ts#L25-L28

PR #318105, opened 2026-05-23
feat(notifications): add smooth fade-out transition for notification toasts

Neizan93:
This PR addresses a minor but noticeable visual inconsistency in the Developer Experience. Notification toasts currently have a smooth slide+fade entry animation (`notification-fade-in`), but they vanish instantly without any transition when they are dismissed or time out.
This PR adds a smooth **200ms fade-out transition** (`opacity: 1 -> 0`) before the toast is removed from the DOM, creating a more cohesive and polished interaction. 
- Applies a `notification-fade-out` CSS class to the toast container before disposal.
- Waits for the `transitionend` event (with a 300ms safety fallback timeout) before running the existing disposal logic.
- Adds anti-reentry guards and race condition checks.
- Respects `prefers-reduced-motion` settings (0ms transition duration), matching the existing behavior for the entry animation.
- Bulk removals (e.g., when opening the Notification Center) skip the animation for better performance.
### How to test
1. Open the command palette and run `Developer: Show Notification`.
2. Notice the smooth fade-in animation.
3. Click the close (`X`) button on the toast, or wait for it to dismiss automatically.
4. Verify that the toast now smoothly fades out (opacity to 0) instead of disappearing instantly.
5. Enable "Reduce Motion" in your OS accessibility settings and verify that the toast is removed instantly without animation.

### DIFF
--- a/src/vs/workbench/browser/parts/notifications/media/notificationsToasts.css
+++ b/src/vs/workbench/browser/parts/notifications/media/notificationsToasts.css
@@ -69,6 +69,15 @@
 	transition: none;
 }
 
+.monaco-workbench > .notifications-toasts .notification-toast-container > .notification-toast.notification-fade-out {
+	opacity: 0;
+	transition: opacity 200ms ease-in;
+}
+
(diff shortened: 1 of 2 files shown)


Neizan93:

PR #318113, opened 2026-05-23
Claude agent — Phase 11: customizations / plugins

TylerLeonhardt:
Phase 11 of the Claude agent provider in the agent host.

## Behaviour

- **Workbench-pushed customizations.** `setClientCustomizations` / `setCustomizationEnabled` flow through `IAgentPluginManager` into `Options.plugins` on the Claude SDK `Query`.
- **Server-side (SDK-discovered) customizations.** Commands / agents / MCP servers from the live `Query` are projected as a single \"Discovered in Claude\" Open Plugins-conformant on-disk bundle so the workbench's plugin expander surfaces them next to client-pushed entries.

## Design notes

- \`Query.reloadPlugins()\` is parameterless — the SDK's plugin URI set is captured into \`Options.plugins\` at startup and is otherwise immutable. Any client-pushed change therefore triggers a yield-restart through the same rematerializer path used for client-tool changes. \`send()\`'s pre-flight runs a single rebind when **either** \`toolDiff\` or \`clientCustomizationsDiff\` is dirty; the rematerializer reads \`clientCustomizationsDiff.consume()\` so the new URI set lands in \`Options.plugins\` of the rebuilt \`Query\`.
- \`SessionClientCustomizationsDiff\` drives dirty from the model state observable (widened equality covers \`nonce\`, \`displayName\`, \`description\`, \`statusMessage\`, \`agents\`, \`pluginDir\`, status, enablement). Same-URI content refreshes (nonce bumps) now correctly flip dirty.
- \`setClientCustomizations\` runs inside the per-session sequencer so a fire-and-forget call from \`AgentSideEffects\` cannot race a first \`sendMessage\`.
- \`ClaudeSdkCustomizationBundler\` writes a hashed, content-addressed on-disk tree under \`IAgentPluginManager.basePath\`. Repeated calls with the same SDK snapshot are nonce-stable and skip the rewrite. The on-disk tree is intentionally a cross-session warm cache.

## Tests

New \`customizations/\` test folder mirrors the source structure:
- \`SessionClientCustomizationsDiff\` — URI list, nonce, metadata, enablement, dirty + consume semantics.
- Projector — client + server tier merge with per-URI enablement overlay.
- Bundler — write layout, nonce stability, name sanitisation, working-directory namespacing, delete-on-change.

\`claudeAgent.test.ts\`:
- Sync-and-toggle dispatch through the sequencer.
- Rebind on customizations dirty (proves \`reloadPlugins()\` would not have sufficed).
- Mid-turn race coverage.
- Swallowed-SDK-snapshot fallback in \`getSessionCustomizations\`.

Draft for review. No telemetry impact, no proposed-API changes.


TylerLeonhardt:
@copilot fix the lint errors and comments

TylerLeonhardt:
@copilot fix the test failure

roblourens (approved):

PR #318124, opened 2026-05-23
Support dropping folders into the terminal when on a remote

pglira:
## Problem

Dragging a folder from the File Explorer into the integrated terminal does nothing when connected to a remote (e.g. a dev container), even though:

- dragging a **file** into the terminal works on a remote, and
- dragging a **folder** works locally.

See microsoft/vscode-remote-release#11494.

## Cause

The terminal's drop handler (`TerminalInstanceDragAndDropController.onDrop`) reads the dragged path from two data transfers:

- `DataTransfers.RESOURCES` — but `fillEditorsDragData` only writes **non-directory** files here (directories are intentionally excluded so they aren't opened as editors), and
- `CodeDataTransfers.FILES` — but the explorer (`explorerViewer.ts`) only sets this for resources with the `file:` scheme.

So for a folder on a `vscode-remote:` filesystem, **neither** transfer carries the resource and the drop is a no-op. Locally it works only because the folder has a `file:` scheme and therefore lands in `CodeDataTransfers.FILES`.

## Fix

Also read `DataTransfers.INTERNAL_URI_LIST`. This transfer is populated by `fillEditorsDragData` with **folders included** and preserves the original (remote) URIs, which is exactly what the terminal needs. This mirrors how the chat widget resolves dropped resources (`chatDragAndDrop.ts`).

It is read as a fallback after `DataTransfers.RESOURCES`, so existing file-drop behavior is unchanged; it only adds the missing folder/remote case. `onDragEnter` is also updated to accept the type so the drop overlay appears.

## Testing

- Local + dev container: drag a **file** into the terminal → path is inserted (unchanged).
- Local: drag a **folder** into the terminal → path is inserted (unchanged).
- Dev container / remote: drag a **folder** into the terminal → path is now inserted (previously nothing happened).

Note: I was unable to run a full local build to typecheck (building vscode's native modules requires X11 dev headers I couldn't install), so I'm relying on CI for compilation. The change only uses already-exported APIs (`UriList.parse`, `DataTransfers.INTERNAL_URI_LIST`).

### DIFF
--- a/src/vs/workbench/contrib/terminal/browser/terminalInstance.ts
+++ b/src/vs/workbench/contrib/terminal/browser/terminalInstance.ts
@@ -2623,39 +2623,48 @@ class TerminalInstanceDragAndDropController extends Disposable implements dom.ID
 			return;
 		}
 
-		// Check if files were dragged from the tree explorer
-		let path: URI | undefined;
-		const rawResources = e.dataTransfer.getData(DataTransfers.RESOURCES);
-		if (rawResources) {
-			path = URI.parse(JSON.parse(rawResources)[0]);
-		}
-
-		// Unlike `DataTransfers.RESOURCES`, the internal uri-list also carries folders and
-		// preserves remote (e.g. dev container) URIs, so it is required to support dropping
-		// directories into the terminal when connected to a remote. See #11494.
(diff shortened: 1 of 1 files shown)


pglira:

PR #318173, opened 2026-05-24
fix: Chat session persistence across Copilot auth/profile transitions

### LINKED ISSUE #318152: Chat session history lost when switching between GitHub-authenticated and BYOK/anonymous mode – sessions still on disk but inaccessible
**Description:**

Since VS Code 1.122.0-Insider (PR #317428 – "Enable BYOK in air-gapped scenarios/without GitHub auth"), it is now possible to use Copilot Chat with BYOK models without being logged into GitHub. This is a great step forward for air-gapped and privacy-conscious users.

However, there is a critical issue with chat session persistence that affects both the transition between authentication states and the anonymous/BYOK mode itself:

**Problem 1 – Old sessions become invisible after logging out**
When you log out of GitHub and switch to BYOK/anonymous mode (chat.allowAnonymousAccess: true), all previously created chat sessions vanish from the history view. The chat panel shows an empty history.

The data is still on disk (.json/.jsonl files in workspaceStorage/{workspaceId}/chatSessions/), but VS Code can no longer discover or display them because the storage profile context changed with the authentication state.

**Problem 2 – New sessions in BYOK/anonymous mode may not persist reliably**
Chat sessions created while in anonymous BYOK mode may also disappear after a VS Code restart or when the anonymous token (no_auth_limited_copilot, issued based on devDeviceId) changes. This makes the chat history non-functional in the primary use case the new feature was designed for.

**Root cause analysis (from source code)**
The session persistence uses two storage mechanisms that are both tied to the authentication profile:

ChatSessionStore (src/vs/workbench/contrib/chat/common/model/chatSessionStore.ts):

Session files are stored in workspaceStorage/{workspaceId}/chatSessions/ (.json/.jsonl)
The index is stored in IStorageService under key chat.ChatSessionStore.index
getIndexStorageScope() returns StorageScope.WORKSPACE or StorageScope.APPLICATION for empty windows
However, the actual session content files are referenced relative to the workspace storage path, which may change or become orphaned depending on how the profile is resolved
LocalChatSessionsProvider (src/vs/sessions/contrib/providers/localChatSessions/browser/localChatSessionsProvider.ts):

Stores session metadata under key sessions.localChat.sessions with StorageScope.PROFILE and StorageTarget.MACHINE
StorageScope.PROFILE is bound to the user profile, which changes when the authentication state changes (GitHub login ↔ anonymous)
Authentication change changes the profile context:

When logged into GitHub → sessions stored in the GitHub-authenticated profile context
When logged out → VS Code issues a no_auth_limited_copilot anonymous token based on devDeviceId → storage now uses the anonymous profile context
The old profile context is no longer consulted → old sessions are invisible even though they exist on disk
If the devDeviceId or anonymous token changes (e.g., after a restart or in different VS Code instances), even BYOK sessions can become inaccessible

**Steps to reproduce**

1. Start VS Code Insiders (≥ 1.122.0) logged into GitHub
2. Have some chat conversations (they appear in the history)
3. Log out of GitHub (Accounts menu → Sign out)
4. Set "chat.allowAnonymousAccess": true in settings
5. Reload VS Code (Developer: Reload Window)
6. Configure a BYOK model (or use an existing one)
7. Observed: Chat history is completely empty – all previous sessions are gone
8. Start a new chat in BYOK mode
9. Close and reopen VS Code
10. Observed: The BYOK session may also be gone or inconsistent

**Expected behavior**

- Chat sessions should be auth-agnostic – they belong to the user/machine, not to the authentication session
- Switching between GitHub-authenticated and BYOK/anonymous mode should preserve and display all existing sessions
- Sessions created in BYOK/anonymous mode should persist reliably across restarts
- Ideally, if sessions from a previous auth context are found on disk, they should be migrated or at least offered for recovery

**Suggested solution**
The session storage needs to be decoupled from the authentication profile. Approaches to consider:

- Use StorageScope.APPLICATION instead of StorageScope.PROFILE for session metadata: Machine-level storage is auth-independent and survives login/logout transitions.
- Auth-independent storage path for session files: Store chat session data in a location that doesn't depend on the authentication context (e.g., under $APP_DATA/Code/User/workspaceStorage/ directly, without profile-level indirection).
- Session discovery across profiles: When loading sessions, scan for orphaned session files from other profile contexts and offer migration/import.
- Migration on auth transition: When the authentication state changes (login ↔ logout), migrate existing session metadata to the new profile context so sessions remain visible.

The physical session data (.json/.jsonl files) is already in a persistent location – the missing piece is making the index/metadata storage auth-independent.

**Additional context**
VS Code version: 1.122.0-Insider (and likely the upcoming stable release)
Related PR: #317428 (Enable BYOK in air-gapped scenarios/without GitHub auth)
Related issues: #246551, #313461, #275591, #272843
Extension affected: This is a core VS Code issue, not specific to any extension. 
Chat session storage files on disk remain intact – the data is not deleted, just orphaned from the new profile context.

EmirX3D:
Closes #318152

This makes local chat session metadata auth/profile independent and adds recovery for session files whose storage index is missing.
Changes:
- Store `LocalChatSessionsProvider` metadata in `StorageScope.APPLICATION` instead of `StorageScope.PROFILE`.
- Merge existing profile-scoped local chat session metadata into application storage for compatibility.
- Rebuild `ChatSessionStore` index entries from persisted `.json` / `.jsonl` session files when the index is empty.
- Add regression tests for profile-to-application metadata migration and missing-index recovery.
- Update local chat sessions provider docs.
This prevents chat history from disappearing when switching between GitHub-authenticated and BYOK/anonymous Copilot modes, and improves recovery when session files remain on disk but their index metadata is missing.

### DIFF
--- a/src/vs/sessions/contrib/providers/localChatSessions/LOCAL_CHAT_SESSIONS_PROVIDER.md
+++ b/src/vs/sessions/contrib/providers/localChatSessions/LOCAL_CHAT_SESSIONS_PROVIDER.md
@@ -63,7 +63,7 @@ interface IStoredLocalSession {
 - **Remove** (`_removeStoredSession`) — called from `deleteSession`.
 - **Load** (`_loadPersistedSessions`) — on provider construction; reads stored entries and creates `LocalSession` instances from them.
 
-Storage is self-contained: no `IChatService` calls are needed to list sessions, since title and timing are stored inline.
+Storage is self-contained: no `IChatService` calls are needed to list sessions, since title and timing are stored inline. Metadata is stored in `StorageScope.APPLICATION` so it remains available across profile/auth transitions; older `StorageScope.PROFILE` entries are merged forward on load.
 
 ### One-time migration
 
--- a/src/vs/sessions/contrib/providers/localChatSessions/browser/localChatSessionsProvider.ts
+++ b/src/vs/sessions/contrib/providers/localChatSessions/browser/localChatSessionsProvider.ts
@@ -45,6 +45,8 @@ export const LOCAL_SESSION_ENABLED_SETTING = 'sessions.chat.localAgent.enabled';
 const LOCAL_PROVIDER_ID = 'local-chat';
 const STORAGE_KEY_SESSIONS = 'sessions.localChat.sessions';
 const STORAGE_KEY_MIGRATED = 'sessions.localChat.migrated';
+const STORAGE_SCOPE = StorageScope.APPLICATION;
+const LEGACY_STORAGE_SCOPE = StorageScope.PROFILE;
 
 interface IStoredLocalSession {
 	readonly uri: UriComponents;
@@ -396,6 +398,8 @@ export class LocalChatSessionsProvider extends Disposable implements ISessionsPr
 
 	private readonly _currentNewSession = this._register(new MutableDisposable<LocalSession>());
 
+	private _legacyStorageMigrationAttempted = false;
+
 	constructor(
 		@IChatService private readonly chatService: IChatService,
 		@IInstantiationService private readonly instantiationService: IInstantiationService,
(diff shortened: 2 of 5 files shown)


EmirX3D:

PR #318292, opened 2026-05-26
fix: disable breadcrumb path copy without symbols

### LINKED ISSUE #286240: `Copy Breadcrumb Path` doesn't do anything when you're not focused on a symbol
While testing https://github.com/microsoft/vscode/issues/285956, I noticed that if I open a file from the explorer, right-click the tab, and then select `Copy Breadcrumbs Path`, it will simply no-op. Or if I open a file that doesn't support symbol breadcrumbs (e.g. GitHub Action workflows, without the proper extensions installed?), then regardless where I put my cursor, the `Copy Breadcrumbs Path` doesn't do anything.

This behavior kind of makes sense, but since the command name doesn't mention it's specific to symbols, and the breadcrumb can display a nested folder/file structure, I could imagine it being slightly confusing to some users when/why this actually works. But in practice, maybe no one would ever notice?

That said, I just raise it for conversation, since I don't feel strongly about the issue being a priority. But maybe we could discuss disabling it when there's no symbol selection?

SpencerJung:
## Related Issue

Closes #286240

## Summary

Disable the `Copy Breadcrumbs Path` command unless the active breadcrumbs contain an actual symbol path.

The command previously stayed enabled for file-only breadcrumbs or placeholder symbol breadcrumbs and then silently returned without copying anything. This adds a breadcrumbs context key that is true only when at least one real outline breadcrumb is present, so the command/menu item is unavailable when it would be a no-op.

## How to Test

1. Open a file that has no symbol breadcrumb selection.
2. Right-click the editor tab.
3. Verify `Copy Breadcrumbs Path` is disabled/hidden while `Copy Path` and `Copy Relative Path` remain available.
4. Open a file with symbol breadcrumbs and place the cursor inside a symbol.
5. Verify `Copy Breadcrumbs Path` is available and still copies the symbol breadcrumb path.

## Community Rules Review

- `CONTRIBUTING.md` reviewed; it points code contributors to the VS Code wiki.
- `.github/pull_request_template.md` reviewed and included above.
- VS Code wiki reviewed: CLA required, one PR per issue, linked issue, small changes, coding guidelines, and tests whenever possible.
- CI reviewed: `.github/workflows/pr.yml` targets `main` and runs Compile & Hygiene plus Linux/macOS/Windows test workflows.
- Base branch: `main`, confirmed via `gh repo view microsoft/vscode --json defaultBranchRef`.

### Dependency Setup

- `bun install`: attempted; failed building `@vscode/sqlite3` because local Windows environment lacks Visual Studio C++ Build Tools.
- Fallback: `npm install --ignore-scripts` succeeded for static tooling; generated `package-lock.json` diff was restored before code changes.
- Environment note: repository `.nvmrc` expects Node `22.22.1`; local Node is `24.14.0`.

### Tests

- `node test/unit/browser/index.js --grep "Breadcrumb"`: runner completed with exit code 0 after installing Playwright browsers, but reported `0 passing`, so no matching automated test executed.
- `bun test --bail`: unavailable for this repo setup; Bun starts `test/monaco/monaco.test.ts` and fails resolving `chai` before reaching this change.

### Type Check

- `npm run compile-check-ts-native`: passed.
- LSP diagnostics for `src/vs/workbench/browser/parts/editor/breadcrumbsControl.ts`: no diagnostics.
- `bun run typecheck`: unavailable because VS Code root `package.json` has no `typecheck` script.

### Build

- `bun run build`: unavailable because VS Code root `package.json` has no `build` script.
- Official VS Code build/CI equivalent is Compile & Hygiene in `.github/workflows/pr.yml`; local full build was not run because native dependency installation failed without Visual Studio C++ Build Tools.

### Diff Summary

```text
src/vs/workbench/browser/parts/editor/breadcrumbsControl.ts | 15 ++++++++++++---
1 file changed, 12 insertions(+), 3 deletions(-)
```

## Checklist

- [x] Base branch is `main`
- [x] Associated issue is linked
- [x] All feasible verification checks attempted; unavailable checks documented with reason
- [x] Commits split if >=3 files changed (single-file change)
- [x] No unintended schema or lockfile regeneration

### DIFF
--- a/src/vs/workbench/browser/parts/editor/breadcrumbsControl.ts
+++ b/src/vs/workbench/browser/parts/editor/breadcrumbsControl.ts
@@ -237,10 +237,12 @@ export class BreadcrumbsControl {
 	static readonly CK_BreadcrumbsPossible = new RawContextKey('breadcrumbsPossible', false, localize('breadcrumbsPossible', "Whether the editor can show breadcrumbs"));
 	static readonly CK_BreadcrumbsVisible = new RawContextKey('breadcrumbsVisible', false, localize('breadcrumbsVisible', "Whether breadcrumbs are currently visible"));
 	static readonly CK_BreadcrumbsActive = new RawContextKey('breadcrumbsActive', false, localize('breadcrumbsActive', "Whether breadcrumbs have focus"));
+	static readonly CK_BreadcrumbsSymbolPath = new RawContextKey('breadcrumbsSymbolPath', false, localize('breadcrumbsSymbolPath', "Whether breadcrumbs contain a symbol path"));
 
 	private readonly _ckBreadcrumbsPossible: IContextKey<boolean>;
 	private readonly _ckBreadcrumbsVisible: IContextKey<boolean>;
 	private readonly _ckBreadcrumbsActive: IContextKey<boolean>;
+	private readonly _ckBreadcrumbsSymbolPath: IContextKey<boolean>;
 
 	private readonly _cfUseQuickPick: BreadcrumbsConfig<boolean>;
 	private readonly _cfShowIcons: BreadcrumbsConfig<boolean>;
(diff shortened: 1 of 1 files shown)


dmitrivMS:
This has been done by https://github.com/microsoft/vscode/pull/318042

PR #315184, opened 2026-05-08
Fix dismissibility of completed progress notifications after progress completion

DhineshPonnarasan:
Fixes #314963

## Summary

Completed progress notifications could remain visible but non-dismissible in Notification Center:

* row close action did not reappear
* Clear All stayed disabled
* `notifications.clearAll` skipped these notifications

## Root Cause

`NotificationViewItem.hasProgress` treated the existence of a progress object as active progress, even after `done()`.

Several Notification Center behaviors interpret `hasProgress` as "progress currently active", including dismissal affordances and Clear All enablement.

## Fix

* Treat completed progress as non-active in `hasProgress`
* Refresh notification rows on the progress completion transition so dismissal affordances are recalculated
* Add regression tests for model and Notification Center behavior
* Document `done()` lifecycle semantics in notification progress API comments

## Validation

* `npm run compile-check-ts-native`
* `scripts/test.bat src/vs/workbench/test/common/notifications.test.ts src/vs/workbench/test/browser/notificationsList.test.ts`

  * 10 passing

## Notes

This keeps invalidation limited to the completion transition and avoids full row relayout during incremental progress updates.

---

@bpasero could you please take a look when you have time?

Happy to incorporate any feedback including additional tests, refinements or scope adjustments if needed.

PR #318319, opened 2026-05-26
fix(notifications): show close button after progress notification completes

### LINKED ISSUE #314963: Notification close button hidden and "Clear All" disabled after progress notification completes
## Summary

After an extension's progress notification finishes (i.e. `progress.done()` is called internally), the notification that remains visible in the Notification Center exhibits two broken behaviors:

1. The **close (×) button is not shown** on the notification row — there is no way to dismiss it individually.
2. The **"Clear All Notifications" button is disabled**, and the `notifications.clearAll` command silently skips the notification.

## Steps to Reproduce

1. Use any extension that creates a `ProgressLocation.Notification` progress notification and leaves it visible after the operation completes (e.g. extensions like Gradle for Java, Dev Containers, or any extension using `vscode.window.withProgress({ location: vscode.ProgressLocation.Notification })`).
2. Wait for the operation to finish (the progress bar completes/disappears).
3. Open the Notification Center (`Ctrl+K Ctrl+Shift+N`).
4. Observe: the notification row has **no close (×) button**.
5. Observe: the **"Clear All Notifications" toolbar button is greyed out**.
6. Run `notifications.clearAll` from the Command Palette — the completed-progress notification is **not dismissed**.

## Expected Behavior

Once progress completes (`done()` is called), the notification should behave like any other standard notification:
- The close (×) button should appear.
- "Clear All Notifications" should be enabled and should dismiss it.

## Root Cause

`INotificationViewItem.hasProgress` (in `src/vs/workbench/common/notifications.ts`) is implemented as:

```typescript
get hasProgress(): boolean {
    return !!this._progress; // BUG: _progress is never nulled out after done()
}
```

`_progress` is a lazily-created object that, once instantiated, is **never set back to `null`**, even after `progress.done()` is called. As a result, `hasProgress` permanently returns `true` for any notification that ever had progress.

This getter is used in at least three places that all misbehave:

**1. Close button hidden** (`notificationsViewer.ts`):
```typescript
// Close (unless progress is showing)
if (!notification.hasProgress) {   // ← always false after any progress
    actions.push(NotificationTemplateRenderer.closeNotificationAction);
}
```

**2. "Clear All" disabled** (`notificationsCenter.ts`):
```typescript
clearAllAction.enabled = this.model.notifications.some(notification => !notification.hasProgress);
// ← if all notifications had progress, enabled is always false
```

**3. `clearAll()` skips the notification** (`notificationsCenter.ts`):
```typescript
if (!notification.hasProgress) {   // ← false, so close() is never called
    notification.close();
}
```

**Note:** The `sticky` getter in the same file already handles this correctly:
```typescript
(this._progress && !this._progress.state.done) // only sticky while progress is running
```

## Proposed Fix

**Step 1** — Fix the root cause in `src/vs/workbench/common/notifications.ts`:

```typescript
get hasProgress(): boolean {
    return !!this._progress && !this._progress.state.done;
}
```

**Step 2** — Trigger a toolbar re-render when progress completes, in `src/vs/workbench/browser/parts/notifications/notificationsViewer.ts`:

```typescript
case NotificationViewItemContentChangeKind.PROGRESS:
    this.renderProgress(notification);
    this.renderActions(notification); // add this: refresh close button visibility
    break;
```

Without Step 2, even with Step 1 fixed, the close button won't appear in the toolbar until the next full list re-render (e.g. scrolling), because the `PROGRESS` content change event handler only calls `renderProgress()` and not `renderActions()`.

## Affected Files

- `src/vs/workbench/common/notifications.ts` — `hasProgress` getter (root cause)
- `src/vs/workbench/browser/parts/notifications/notificationsViewer.ts` — close button render logic, PROGRESS event handler
- `src/vs/workbench/browser/parts/notifications/notificationsCenter.ts` — `clearAllAction.enabled`, `clearAll()`

- VS Code Version: latest `main`
- OS: Windows / macOS / Linux (platform-independent)

SpencerJung:
## Related Issue
Closes #314963

## Summary
After a progress notification completes (`progress.done()` is called), the notification's `hasProgress` getter still returned `true` because `_progress` was never nulled out. This caused three broken behaviors:

1. The close (×) button was hidden on the notification row
2. The "Clear All Notifications" button was disabled
3. The `notifications.clearAll` command silently skipped the notification

## Changes
- **notifications.ts**: Fixed `hasProgress` getter to check `!!this._progress && !this._progress.state.done` (matching the existing `sticky` getter logic)
- **notificationsViewer.ts**: Added `this.setInput(notification)` call on `PROGRESS` content change events so the toolbar actions (including close button visibility) are re-rendered when progress completes

## Verification Evidence
- `npm run compile-check-ts-native`: PASSED (no TypeScript errors)

## Checklist
- [x] Base branch is main
- [x] All feasible verification checks attempted
- [x] No unintended schema or lockfile changes

### DIFF
--- a/src/vs/workbench/browser/parts/notifications/notificationsViewer.ts
+++ b/src/vs/workbench/browser/parts/notifications/notificationsViewer.ts
@@ -412,6 +412,7 @@ export class NotificationTemplateRenderer extends Disposable {
 					break;
 				case NotificationViewItemContentChangeKind.PROGRESS:
 					this.renderProgress(notification);
+					this.setInput(notification);
 					break;
 				case NotificationViewItemContentChangeKind.MESSAGE:
 					this.renderMessage(notification, messageCustomHover);
--- a/src/vs/workbench/common/notifications.ts
+++ b/src/vs/workbench/common/notifications.ts
@@ -617,7 +617,7 @@ export class NotificationViewItem extends Disposable implements INotificationVie
 	}
 
 	get hasProgress(): boolean {
-		return !!this._progress;
+		return !!this._progress && !this._progress.state.done;
 	}
 
 	get progress(): INotificationViewItemProgress {


dmitrivMS:
https://github.com/microsoft/vscode/pull/315184 is a better fix

PR #318326, opened 2026-05-26
fix(breadcrumbs): show info when no symbols to copy

### LINKED ISSUE #286240: `Copy Breadcrumb Path` doesn't do anything when you're not focused on a symbol

SpencerJung:
## Related Issue
Closes #286240

## Summary
The 'Copy Breadcrumbs Path' command silently did nothing when there were no breadcrumb symbols to copy (e.g. when not focused on a symbol or in files without symbol support). This was confusing because users couldn't tell if the command worked or not.

## Changes
- Show an info notification when there are no breadcrumb symbols to copy

## Verification Evidence
- `npm run compile-check-ts-native`: PASSED (no TypeScript errors)

## Checklist
- [x] Base branch is main
- [x] All feasible verification checks attempted
- [x] No unintended schema or lockfile changes

### DIFF
--- a/src/vs/workbench/browser/parts/editor/breadcrumbsControl.ts
+++ b/src/vs/workbench/browser/parts/editor/breadcrumbsControl.ts
@@ -24,6 +24,7 @@ import { Categories } from '../../../../platform/action/common/actionCommonCateg
 import { Action2, MenuId, registerAction2 } from '../../../../platform/actions/common/actions.js';
 import { IConfigurationService } from '../../../../platform/configuration/common/configuration.js';
 import { IClipboardService } from '../../../../platform/clipboard/common/clipboardService.js';
+import { INotificationService } from '../../../../platform/notification/common/notification.js';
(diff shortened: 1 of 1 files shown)


joshspicer:
I see you also have https://github.com/microsoft/vscode/pull/318292, which is likely the better solution - right?

dmitrivMS:
This has been done by https://github.com/microsoft/vscode/pull/318042

PR #318369, opened 2026-05-26
Fix session folder URI for non-git folders

sandy081:
Fixes session jumping between 'unknown folder' and real folder in Copilot CLI sessions when the working directory has no git repository.

**Root cause**: `CopilotChatSessionsProvider._buildWorkspace` only used `repositoryPath` metadata, falling back to `unknown:///` for non-git folders. The `workingDirectoryPath` metadata was already being set but never used.

**Changes**:
- `_extractRepositoryFromMetadata`: Extract `folderUri` from `workingDirectoryPath` when `repositoryPath` is not available
- `_buildWorkspace`: Use `repoUri ?? folderUri ?? unknown:///` for root URI, and only create `gitRepository` object when a real git repo exists

This ensures non-git folders display correctly in the sessions view with proper workspace URIs.


Monrother:
@sandy081 Why this is closed? I guess it can help solve #319379 and #318501

PR #318443, opened 2026-05-26
 Switch from Copilot Proxy to CAPI

guomaggie:
Currently, we use copilot-proxy as a go-between to get access from VS Code to models hosted in Fireworks. This PR covers changes needed to switch to CAPI: 

- CAPI returns a list of models available based on a user's SKU, FSI status, etc. We disabled the search subagent tool (for explore as well) if the search subagent models are not in the list of models returned by CAPI. So, if a user is blocked from using Qwen models, they're blocked from using our subagent tool.
- Updated the search subagent tool to use a CAPI endpoint, instead of proxyAgenticEndpoint

### DIFF
--- a/extensions/copilot/src/extension/intents/node/agentIntent.ts
+++ b/extensions/copilot/src/extension/intents/node/agentIntent.ts
@@ -15,6 +15,7 @@ import { ConfigKey, IConfigurationService } from '../../../platform/configuratio
 import { isAnthropicFamily, isGptFamily, modelCanUseApplyPatchExclusively, modelCanUseReplaceStringExclusively, modelSupportsApplyPatch, modelSupportsMultiReplaceString, modelSupportsReplaceString, modelSupportsSimplifiedApplyPatchInstructions } from '../../../platform/endpoint/common/chatModelCapabilities';
 import { IEndpointProvider } from '../../../platform/endpoint/common/endpointProvider';
 import { IAutomodeService } from '../../../platform/endpoint/node/automodeService';
+import { SEARCH_AGENT_FAMILY } from '../../../platform/endpoint/node/searchAgentChatEndpoint';
 import { IEnvService } from '../../../platform/env/common/envService';
 import { ILogService } from '../../../platform/log/common/logService';
 import { IEditLogService } from '../../../platform/multiFileEdit/common/editLogService';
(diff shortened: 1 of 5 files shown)


extensions/copilot/src/extension/prompt/node/searchSubagentToolCallingLoop.ts
```diff
@@ -95,9 +94,19 @@ export class SearchSubagentToolCallingLoop extends ToolCallingLoop<ISearchSubage
...
+				const searchAgentEndpoint = allEndpoints.find(e => e.family === SEARCH_AGENT_FAMILY);
+				if (searchAgentEndpoint instanceof ChatEndpoint) {
+					return this.instantiationService.createInstance(SearchAgentChatEndpoint, searchAgentEndpoint.modelMetadata);
+				}
+				this._logService.warn(`Search-agent model not available in CAPI, falling back to main agent endpoint`);
+			} catch (error) {
+				this._logService.warn(`Failed to get search-agent endpoint from CAPI, falling back to main agent: ${error}`);
+			}
+			return await this.endpointProvider.getChatEndpoint(this.options.request);
 		}
```
guomaggie, line 120: Noticed during testing that the switch will cause backgroundToDoAgent to fail because it requests copilot-fast, which isn't available to free users. Should this change be covered by this PR too (the current PR focuses on supporting search subagent)? 

Image

extensions/copilot/src/extension/intents/node/agentIntent.ts
```diff
@@ -146,8 +147,10 @@ export const getAgentTools = async (accessor: ServicesAccessor, request: vscode.
 	const searchSubagentEnabled = configurationService.getExperimentBasedConfig(ConfigKey.Advanced.SearchSubagentToolEnabled, experimentationService);
 	const exploreAgentEnabled = configurationService.getExperimentBasedConfig(ConfigKey.ExploreAgentEnabled, experimentationService);
 	const isGptOrAnthropic = isGptFamily(model) || isAnthropicFamily(model);
-	allowTools[ToolName.SearchSubagent] = isGptOrAnthropic && searchSubagentEnabled && exploreAgentEnabled;
-	allowTools[ToolName.ExploreSubagent] = isGptOrAnthropic && searchSubagentEnabled && !exploreAgentEnabled;
+	const allEndpoints = await endpointProvider.getAllChatEndpoints().catch(() => [] as IChatEndpoint[]);
+	const searchAgentAvailable = allEndpoints.some(e => e.family === SEARCH_AGENT_FAMILY);
```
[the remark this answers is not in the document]
maltacc: fixed

extensions/copilot/src/extension/prompt/node/searchSubagentToolCallingLoop.ts
```diff
@@ -95,9 +94,19 @@ export class SearchSubagentToolCallingLoop extends ToolCallingLoop<ISearchSubage
...
+				const allEndpoints = await this.endpointProvider.getAllChatEndpoints();
+				const searchAgentEndpoint = allEndpoints.find(e => e.family === SEARCH_AGENT_FAMILY);
+				if (searchAgentEndpoint instanceof ChatEndpoint) {
+					return this.instantiationService.createInstance(SearchAgentChatEndpoint, searchAgentEndpoint.modelMetadata);
+				}
+				this._logService.warn(`Search-agent model not available in CAPI, falling back to main agent endpoint`);
+			} catch (error) {
+				this._logService.warn(`Failed to get search-agent endpoint from CAPI, falling back to main agent: ${error}`);
+			}
+			return await this.endpointProvider.getChatEndpoint(this.options.request);
```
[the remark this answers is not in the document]
maltacc: Keeping both paths; the checks happen at different times - agentIntent.ts for whether we expose the tool when the req starts, and getEndpoint for when we pick the endpoint at fetch time.

extensions/copilot/src/extension/prompt/node/searchSubagentToolCallingLoop.ts
24anisha: can you elaborate on this? your changes should only apply to search agent, not to the background todo tool

guomaggie: ^ just something I noticed, changes should be unrelated to this PR!

extensions/copilot/src/extension/prompt/node/searchSubagentToolCallingLoop.ts
```diff
@@ -102,9 +101,21 @@ export class SearchSubagentToolCallingLoop extends ToolCallingLoop<ISearchSubage
...
```
vritant24, line 118: correct me if I'm wrong, but I thought we shouldn't be using the main agent in the search subagent if the model isn't available as it can end up charging the users much more now with UBB enabled. We should throw an error/gracefully exit the search subagent if the model isn't available

extensions/copilot/src/extension/prompt/node/searchSubagentToolCallingLoop.ts
vritant24: Ah yes, thanks for noticing that, I'll be pushing a fix for this

extensions/copilot/src/extension/prompt/node/searchSubagentToolCallingLoop.ts
24anisha: if the search subagent model isn't in the CAPI model list, we remove the tool (which is the behavior we agreed upon). If, however, we fall into the situation where the model is available but there's an error, falling back to the main agent should be ok. The CoGS delta when using the main agent as the search subagent model wasn't stat sig IIRC. And this case should be happening very rarely --only when the search model is available but erroring.

We could change it to just fail immediately with a message back to the main agent model, but sometimes the model will keep trying to call the tool because it's available, and having it return errors over and over might actually increase costs more than substituting the main agent model for the subagent model so the system still works. I don't think one is clearly better from a CoGS perspective than the other.

24anisha (approved):

bhavyaus (approved):

alexdima:
FYI I had to revert this via https://github.com/microsoft/vscode/pull/320208 because it causes CI failures

See https://github.com/microsoft/vscode-engineering/issues/2967
See https://dev.azure.com/monacotools/Monaco/_build/results?buildId=445447&view=results

PR #318457, opened 2026-05-26
Subagent: fall back to BYOK endpoint when main agent owns auth

### LINKED ISSUE #318373: BYOK mode's terminal tool call fails when GitHub is not signed in
Testing #318190

With the Anthropic Provider, `run_in_Terminal` fails when GitHub Copilot is not signed into. 

Image

When signed in, it ended up working.

trajectories: 
[failed-no-sign-in.json](https://github.com/user-attachments/files/28273157/failed-no-sign-in.json)
[success-signed-in.json](https://github.com/user-attachments/files/28273181/success-signed-in.json)

dmitrivMS:
Fixes #318373.

## Problem

In BYOK mode with Anthropic, calling `run_in_terminal` from the agent loop fails when the user is **not signed into GitHub**. The terminal tool itself works, but the agent immediately retries via `execution_subagent`, which fails with:

```
The execution subagent request failed with this message:
badRequest: key is missing
```

When the user *is* signed into GitHub, the same flow succeeds.

## Root cause

`ExecutionSubagentToolCallingLoop.getEndpoint()` (and `SearchSubagentToolCallingLoop.getEndpoint()`) can route subagent requests through `ProxyAgenticEndpoint` (CAPI `ProxyChatCompletions`) or a Copilot-hosted override model when the corresponding experiment configs are enabled. Both require a Copilot token. When the main agent is a BYOK endpoint (`ownsAuthorization === true`) and the user has no GitHub session, no Copilot token is available, so `chatMLFetcher` returns `key is missing` immediately while the parent agent keeps working against the BYOK URL.

## Fix

In both subagent loops, resolve the main agent endpoint first and short-circuit when it owns its authorization, so the subagent runs on the same BYOK endpoint as the parent agent.

- [executionSubagentToolCallingLoop.ts](https://github.com/microsoft/vscode/blob/dmitriv/byok-subagent-fix/extensions/copilot/src/extension/prompt/node/executionSubagentToolCallingLoop.ts)
- [searchSubagentToolCallingLoop.ts](https://github.com/microsoft/vscode/blob/dmitriv/byok-subagent-fix/extensions/copilot/src/extension/prompt/node/searchSubagentToolCallingLoop.ts)

## Behavior change

This is intentional but worth flagging:

| Scenario | Before | After |
| --- | --- | --- |
| BYOK + not signed in | Subagent fails with `key is missing` | Subagent runs on BYOK endpoint |
| BYOK + signed in + `useAgenticProxy` (or override `model`) enabled | Subagent ran on Copilot agentic proxy / override model (used Copilot quota, optimized router model) | Subagent runs on BYOK endpoint (uses BYOK quota, BYOK model) |
| Non-BYOK | Unchanged | Unchanged |
| BYOK + default settings (no proxy, no override) | Already used main endpoint | Same |

The second row is the new behavior for signed-in BYOK users in the proxy/override-model experiment cohorts: they no longer silently hit Copilot-hosted routing from a BYOK chat. If preserving that path matters for the experiment, the alternative is to short-circuit only when no Copilot token is actually available; happy to switch to that variant.

## Tests

Added unit tests covering the BYOK short-circuit and the existing agentic-proxy / override-model / fallback paths for both loops:

- `extensions/copilot/src/extension/prompt/node/test/executionSubagentToolCallingLoop.spec.ts` — 7 tests
- `extensions/copilot/src/extension/prompt/node/test/searchSubagentToolCallingLoop.spec.ts` — 5 tests

All 12 pass.

cc @kevin-m-kent @vikramnitin9 @24anisha @vritant24


extensions/copilot/src/extension/prompt/node/executionSubagentToolCallingLoop.ts
```diff
@@ -113,14 +113,20 @@ export class ExecutionSubagentToolCallingLoop extends ToolCallingLoop<IExecution
 	 * Get the endpoint to use for the execution subagent
 	 */
 	private async getEndpoint() {
+		const mainEndpoint = await this.endpointProvider.getChatEndpoint(this.options.request);
+		// BYOK endpoints have no Copilot token for the agentic proxy or override models.
+		if (mainEndpoint.ownsAuthorization) {
```
vikramnitin9, line 118: Rather than check for `ownsAuthorization` as a condition to fall back to main agent endpoint, can we just add a try-catch block around the proxy endpoint instantiation? Like this:
```ts
			// Use agentic proxy with ExecutionSubagentModel or default to DEFAULT_AGENTIC_PROXY_MODEL
			const agenticProxyModel = modelName || ExecutionSubagentToolCallingLoop.DEFAULT_AGENTIC_PROXY_MODEL;
			try {
				const endpoint = this.instantiationService.createInstance(ProxyAgenticEndpoint, agenticProxyModel);
			} catch (error) {
				// If the specified model isn't available, fall back to main agent endpoint
				return await this.endpointProvider.getChatEndpoint(this.options.request);
			}
```
This would make it robust to all kinds of errors while instantiating the proxy endpoint.

vikramnitin9 (changes requested):

dmitrivMS:
This is superseded by https://github.com/microsoft/vscode/pull/319467

PR #318486, opened 2026-05-27
Skip Inno spawn when another instance is already installing

### LINKED ISSUE #62476: Setup has detected that setup is currently running
Issue Type: <b>Bug</b>

Unexpected since I did not run any setup myself. Shortly before that the (1) badge showed up on the gear icon.

image

VS Code version: Code - Insiders 1.29.0-insider (220bff273eed39013f24b6d26fc17f2a742ce10f, 2018-11-02T06:18:20.695Z)
OS version: Windows_NT x64 10.0.17134

<details>

|Item|Value|
|---|---|
|CPUs|Intel(R) Core(TM) i7-4770HQ CPU @ 2.20GHz (4 x 2195)|
|GPU Status|2d_canvas: enabled<br>checker_imaging: disabled_off<br>flash_3d: enabled<br>flash_stage3d: enabled<br>flash_stage3d_baseline: enabled<br>gpu_compositing: enabled<br>multiple_raster_threads: enabled_on<br>native_gpu_memory_buffers: disabled_software<br>rasterization: unavailable_software<br>video_decode: enabled<br>video_encode: enabled<br>webgl: enabled<br>webgl2: enabled|
|Memory (System)|5.98GB (2.91GB free)|
|Process Argv|--inspect-extensions=51039|
|Screen Reader|no|
|VM|100%|

</details><details><summary>Extensions (15)</summary>

---|---|---
remotehub|eam|0.2.0
EditorConfig|Edi|0.12.5
vscode-azureappservice|ms-|0.9.1
vscode-azurefunctions|ms-|0.12.0
vscode-azurestorage|ms-|0.4.2
vscode-cosmosdb|ms-|0.8.0
azure-account|ms-|0.4.3
azurecli|ms-|0.4.2
csharp|ms-|1.17.0
github-issues-prs|ms-|0.9.1
vscode-node-azure-pack|ms-|0.0.4
azurerm-vscode-tools|msa|0.4.2
debugger-for-chrome|msj|4.10.2
vscode-docker|Pet|0.3.1
vscode-open-in-github|sys|1.7.0

</details>

dmitrivMS:
Fixes #62476

### Problem

When two VS Code Insiders instances are running against the same install (for example a normal instance plus one started with `--transient`) and an update becomes available, both instances independently progress through the update state machine to `doApplyUpdate` for the same downloaded `CodeSetup-*.exe` and each spawn Inno Setup.

The second spawn races into Inno's built-in **"Setup is already running"** modal, because Inno protects itself with a global per-install mutex.

### Fix

In `updateService.win32.ts` `doApplyUpdate`, check the existing `${win32MutexName}-updating` Windows named mutex (already held by the running Inno Setup process) **before** spawning. If another instance is already running setup for this install:

- Skip the spawn.
- Do not track an `updateProcess` (we must not kill the other instance's installer if the user cancels here).
- Keep the existing `-ready` mutex polling loop running so this instance still transitions `Updating → Ready` when the other instance's install finishes.

The existing 1-hour `cancelTimeout` continues to act as the safety net if the other instance crashes mid-install.

### Notes

- All filesystem coordination (flag file, cancel file, progress file, cache path) is already shared between instances, so the non-spawning instance still observes the same install correctly.
- A tiny residual race remains (both `isActive` checks return `false` before either process actually grabs the mutex), but the reported scenario — one instance already mid-install when the other tries to start — is handled.
- No new dependency: reuses `@vscode/windows-mutex` already imported a few lines down for `-ready`.


src/vs/platform/update/electron-main/updateService.win32.ts
```diff
@@ -435,14 +467,21 @@ export class Win32UpdateService extends AbstractUpdateService implements IRelaun
 			return;
 		}
 
+		const { updateProcess, updateFilePath, cancelFilePath } = this.availableUpdate;
+
+		// Another instance owns the installer: abort if it's still running so we don't start a new
+		// update cycle on top of it; keep `availableUpdate` so quit-and-install can still complete.
+		if (!updateProcess && this.isInstallerActive(await this.mutex)) {
+			throw new Error('Cannot cancel pending update: another instance is still running setup');
+		}
```
deepak1556, line 476: We do want to cancel running setup when cancelling update, why disable it ?

deepak1556 (commented):
LGTM with a question

Yoyokrazy (approved):

PR #318509, opened 2026-05-27
Prevent signed-out users from seeing GitHub Copilot upgrade recommendations

vritant24:
Ensure that signed-out users do not have access to Copilot control-manifest entries in the model picker. This change enhances user experience by restricting visibility based on user entitlement status.

### DIFF
--- a/src/vs/workbench/contrib/mcp/node/mcpStdioStateHandler.ts
+++ b/src/vs/workbench/contrib/mcp/node/mcpStdioStateHandler.ts
@@ -62,11 +62,17 @@ export class McpStdioStateHandler implements IDisposable {
 		}
 	}
 
-	private killPolite() {
+	private async killPolite() {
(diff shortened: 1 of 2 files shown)


dmitrivMS (dismissed):

src/vs/workbench/contrib/mcp/node/mcpStdioStateHandler.ts
```diff
@@ -62,17 +62,11 @@ export class McpStdioStateHandler implements IDisposable {
 		}
 	}
 
-	private async killPolite() {
+	private killPolite() {
```
dmitrivMS: is this related?

vritant24: I accidentally clicked 'fix with copilot' wow. reverted this

dmitrivMS (approved):

PR #318638, opened 2026-05-27
mcp: prefer announced server title for tool prefixes and picker labels

### LINKED ISSUE #299749: Tool selection via `availableDeferredTools` with MCP Registry-style server names doesn't work properly
To integrate tools from MCP Servers into the prompt sent of to the LLM, GitHub Copilot creates names for each of them and puts them into `availableDeferredTools` in the system prompt. It seems like these names are constructed from MCP Server name (as put into `mcp.json` + tool name (as announced by the MCP server).

For servers of the MCP Registry, especially if the registry is enforced as an allowlist, the server name must follow the MCP Registry spec, so typically something like `io.github.{owner}/{repo}` (when the github-based authentication is used). Compare also https://modelcontextprotocol.io/registry/authentication

This leads to an unfortunate situation for the generated tool names. For the MCP Server [`io.github.upstash/context7`](https://api.mcp.github.com/v0.1/servers/io.github.upstash%2Fcontext7/versions/latest), they would be something like `mcp_io_github_upstash_context7_resolve_library_id`, which is already confusing since most tools will now have a name that includes "github". Except it's even worse, since there's some additional length-based cropping going on so that the actual tool name is `mcp_io_github_ups_resolve_library_id`, which as you can see, doesn't include `context7` at all anymore.

As a consequence, the LLM doesn't know to use the tool, and we have a problem.

I understand the need to not bloat the context window with MCP Servers too much, but the current situation with conflicting naming strategies creates a problem. Perhaps a quick solution would be to base the name of the `availableDeferredTools` not on the config, but on the server-announced title of the MCP server, falling back to the name.

- Copilot Chat Extension Version: 0.38.1
- VS Code Version: 1.110.0
- Selected model (e.g. GPT 4.1, Claude 3.7 Sonnet): Claude 4.5 Opus, GTP-5.3-Codex

Steps to Reproduce:
1. Add an MCP Server like Context7 with it's MCP Registry-compliant server name
2. Check the system prompt in the chat debug view for `availableDeferredTools`

### LINKED ISSUE #299787: GitHub Copilot "Configure Tools" Popover should show `title` from `serverInfo` for MCP Servers
## Context

In GitHub Copilot, you can choose which tools should be available to the agent. In the related "Configure Tools" menu, MCP servers are listed. The UI shows the key of the entry in `mcp.json` as a display name. In case of using an MCP registry when installing servers, this will unfortunately result in rather cryptic names (like `io.github.upstash/context7`).

Image

An MCP Server, in the initialize phase, can communicate it's `serverInfo`, including a `title` and `icon`: https://modelcontextprotocol.io/specification/2025-11-25/basic/lifecycle#initialization
The general rule in the MCP spec seems to be that, whenever a title is there, it should be prioritized over the (technical) name when displaying it to the UI.

It also seems like information from the initialization is already used in the menu, since the GitHub MCP Server features an icon (as announced during initialization), it just doesn't show the name.

## Feature Request

If an MCP Server announces a `title` as part of it's `serverInfo`, the "Configure Tools" menu should show the title instead of any technical name (derived from `mcp.json`).

The same should also be true in the Chat itself, when the MCP server is used. There, it should also show the corresponding `title` instead of any technical name.

## Similar Requests

- https://github.com/microsoft/vscode/issues/243940 is similar, but focuses on additional instructions in mcp.json
- https://github.com/microsoft/vscode/issues/251576 focuses on another part of the UI, i.e. a tooltip for the button that opens the menu this feature talks about

connor4312:
mcp: prefer announced server title for tool prefixes and picker labels

Reworks how MCP tool prefixes are generated and how MCP servers are labelled
in the Configure Tools picker so that the server-announced `serverInfo.title`
(falling back to `serverInfo.name`) is preferred over the mcp.json key.

- Introduces a shared `McpPrefixGenerator` with a `take(name): IReference<string>`
  API that hands out collision-resolved tool prefixes and releases them on
  dispose. Replaces the eager one-shot generator that used to live in
  McpService.
- McpServer now derives its tool prefix from its preferred name (announced
  title when known, otherwise the mcp.json label) so registry-style names
  like `io.github.upstash/context7` no longer end up as truncated
  `mcp_io_github_ups_*` prefixes.
- Configure Tools picker and `ToolDataSource.classify` for MCP sources now
  prefer `serverLabel` (the announced title) over `label`.
- Adds unit tests for `McpPrefixGenerator` covering collisions, slot reuse on
  dispose, bucket cleanup, and name sanitization.

Fixes https://github.com/microsoft/vscode/issues/299749
Fixes https://github.com/microsoft/vscode/issues/299787

### DIFF
--- a/src/vs/workbench/contrib/chat/browser/actions/chatToolPicker.ts
+++ b/src/vs/workbench/contrib/chat/browser/actions/chatToolPicker.ts
@@ -319,7 +319,7 @@ export async function showToolsPicker(
 					itemType: 'bucket',
 					ordinal: BucketOrdinal.Mcp,
 					id: key,
-					label: source.label,
+					label: source.serverLabel || source.label,
 					checked: undefined,
 					collapsed,
 					children,
--- a/src/vs/workbench/contrib/chat/common/tools/languageModelToolsService.ts
+++ b/src/vs/workbench/contrib/chat/common/tools/languageModelToolsService.ts
@@ -154,7 +154,7 @@ export namespace ToolDataSource {
 		if (source.type === 'internal') {
 			return { ordinal: 1, label: localize('builtin', 'Built-In') };
 		} else if (source.type === 'mcp') {
-			return { ordinal: 2, label: source.label };
+			return { ordinal: 2, label: source.serverLabel || source.label };
 		} else if (source.type === 'user') {
 			return { ordinal: 0, label: localize('user', 'User Defined') };
 		} else {
--- a/src/vs/workbench/contrib/mcp/common/mcpServer.ts
+++ b/src/vs/workbench/contrib/mcp/common/mcpServer.ts
@@ -8,11 +8,11 @@ import { CancellationToken, CancellationTokenSource } from '../../../../base/com
 import { Iterable } from '../../../../base/common/iterator.js';
 import * as json from '../../../../base/common/json.js';
 import { normalizeDriveLetter } from '../../../../base/common/labels.js';
-import { Disposable, DisposableStore, IDisposable, MutableDisposable, toDisposable } from '../../../../base/common/lifecycle.js';
+import { Disposable, DisposableStore, IDisposable, IReference, MutableDisposable, toDisposable } from '../../../../base/common/lifecycle.js';
 import { LRUCache } from '../../../../base/common/map.js';
 import { Schemas } from '../../../../base/common/network.js';
 import { mapValues } from '../../../../base/common/objects.js';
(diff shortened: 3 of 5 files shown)


roblourens (approved):

roblourens (approved):

davidruzicka:
> Reworks how MCP tool prefixes are generated and how MCP servers are labelled
in the Configure Tools picker so that the server-announced serverInfo.title
(falling back to serverInfo.name) is preferred over the mcp.json key

Well, but - how can I select right mcp server now?

obrazek

PR #318643, opened 2026-05-27
Read cloud_session_storage_enabled from /copilot_internal/user instead of /v2/token

vijayupadya:
Moves the source of the `cloud_session_storage_enabled` policy flag from the Copilot `/v2/token` endpoint to the `/copilot_internal/user` (entitlements) endpoint, matching the server team's contract.


vijayupadya:
cc @joshspicer

dmitrivMS (approved):

PR #318667, opened 2026-05-28
fix: session rename no longer requires reload (#318442)

### LINKED ISSUE #318442: Renaming a session requires reload
Only seems to happen in VS Code editor window rather than Agents window:
- Right click a chat session in list, rename, set a new name, enter
- ❓ New name doesn't appear
- I have to reload VS Code for the new name to appear

katipally:
Fixes #318442

## Problem

Renaming a session in the Agents window required a window reload for the new name to appear. Two separate bugs caused this:

**Bug A — UI not updating (history sessions)**
`LocalAgentsSessionsController.refresh()` only detected *new* session items (checking `!this._items.has(resource)`). When a history session was renamed, the resource URI was already in `_items`, so `addedOrUpdated` stayed empty and `_onDidChangeChatSessionItems` never fired — the sidebar stayed stale.

**Bug B — Title not persisted**
`renameChat()` in `localChatSessionsProvider.ts` called `chatService.setSessionTitle()` (synchronous, in-memory only) instead of `chatService.setChatSessionTitle()` (async, writes to `_chatSessionStore`). On reload the old title came back because nothing was written to disk.

## Solution

- **Fix A (detection):** Change `refresh()` to use `existing.isEqual(item)` so updated items (changed label, status, etc.) are included in `addedOrUpdated`.
- **Fix A (notification):** Add `onDidChangeChatSessionTitle` event to `IChatService`. Fire it from `setChatSessionTitle()` for history sessions only (live sessions already propagate via `model.onDidChange → autorun`). Subscribe the controller to this event so it calls `refresh()` immediately.
- **Fix B:** Change `renameChat()` to `await chatService.setChatSessionTitle()` so the title is persisted to the session store.
- **Interface fix:** Correct `IChatService.setChatSessionTitle` return type from `void` to `Promise<void>` to match the implementation.

## Files changed

| File | Change |
|------|--------|
| `chatService.ts` | Added `onDidChangeChatSessionTitle` event; fixed return type |
| `chatServiceImpl.ts` | Added emitter; fires event after persisting title for history sessions |
| `localAgentSessionsController.ts` | Fixed `refresh()` update detection; added event listener |
| `localChatSessionsProvider.ts` | Bug B fix — use `setChatSessionTitle` instead of `setSessionTitle` |
| `mockChatService.ts` | Updated mock to match new interface |
| `copilotChatSessionsProvider.test.ts` | Updated inline stubs to async |


katipally:

PR #318756, opened 2026-05-28
Fix terminal-editor shift-drop firing on detached instances

### LINKED ISSUE #311164: Dragging a file into one terminal editor while holding Shift inserts the path into other terminal editors as well

- VS Code Version: 1.115.0 (41dd792b5e652393e7787322889ed5fdc58bd75b)
- OS Version: Windows 10 Home China 26200

Steps to Reproduce:

1. Open VS Code on Windows.
2. Create two integrated terminals in the editor area (not in the panel). In my case both use Git Bash.
3. Make sure they are separate shell sessions, for example `echo $$` returns different PIDs in terminal A and terminal B.
4. Leave both terminals idle.
5. From the Explorer, drag a file into terminal A while holding Shift so VS Code inserts the file path into the terminal instead of opening the file in the editor.
6. Observe that terminal B also receives the same file path even though the file was only dropped on terminal A.

Expected:
Only terminal A should receive the dropped file path.

Actual:
Both terminal editor instances receive the same file path.

Notes:
- Reproduces with `code --disable-extensions`.
- This is specifically about terminals opened in the editor area.
- If I do not hold Shift, the file opens in the editor, which is expected.
- The two terminals are not the same shell session. Example PIDs from `echo $$`: A=8554, B=8599.

vladstudio:
## Summary

Fixes #311164.

`TerminalInstance._dndObserver` (a `MutableDisposable<IDisposable>` holding the per-container `DragAndDropObserver`) was set by `_initDragAndDrop` but never cleared by `detachFromElement`. Because a `TerminalEditor` pane reuses its `_overflowGuardElement` across tab switches (`setInput` calls `detachFromElement` on the old instance and `attachToElement` on the new one against the same container — see `terminalEditor.ts:77-80`), each newly-attached terminal stacked another `DragAndDropObserver` on top of the leaked ones.

When the user shift-dropped a file onto the active editor-area terminal, the `EditorDropTarget` overlay correctly disposed itself in dragover (so the path was not opened as an editor), and the drop bubbled to the shared container — where every previously-attached terminal's observer fired and called `sendPath`, inserting the path into all of them.

Disposing the observer on detach removes the listeners from the prior container, so only the currently-attached instance receives the drop.

## Test plan

- [x] Build dev VS Code (`./scripts/code.sh`)
- [x] Open two terminals in the editor area (same pane)
- [x] Confirm they have different shell PIDs (`echo $$`)
- [x] Drag a file from the Explorer onto the active terminal with Shift held — path appears in the active terminal only
- [x] Switch to the other tab — no path inserted there
- [x] Without Shift, drag still opens the file in the editor (unchanged behavior)

### DIFF
--- a/src/vs/workbench/contrib/terminal/browser/terminalInstance.ts
+++ b/src/vs/workbench/contrib/terminal/browser/terminalInstance.ts
@@ -1047,6 +1047,8 @@ export class TerminalInstance extends Disposable implements ITerminalInstance {
 	detachFromElement(): void {
 		this._wrapperElement.remove();
 		this._container = undefined;
+		// Stop listening on the old container so drops there don't reach this instance after another terminal attaches.
+		this._dndObserver.clear();
 	}
 
 	attachToElement(container: HTMLElement): void {


dmitrivMS (approved):

dmitrivMS (changes requested):
Please add requested test

dmitrivMS (approved):

benibenj (approved):

dmitrivMS:
@vladstudio Please accept the terms.

boedysutris-ctrl:
Pada Sen, 7 Sep 2026 04.22, Dmitriy Vasyura ***@***.***>
menulis:

PR #318757, opened 2026-05-28
Recognize `dosbatch` for dosbatch code blocks

spenserblack:
There currently isn't an issue related to this, based on my search. Apologies if one is required.

This pull request allows `dosbatch` to be recognized as a language name for dosbatch code blocks. E.g.
````markdown
```dosbatch
@echo off
echo Hello, world!
```
````

Currently, `bat` and `batch` are recognized. `dosbatch` is a common name for this language, being recognized as one of the language names for GitHub markdown rendering, and is also the name NeoVim provides for this language.

---

Edit: Perhaps this should be applied to `extensions/latex/syntaxes/markdown-latex-combined.tmLanguage.json`, too?


spenserblack:
Shouldn't copilot have submitted its review? :thinking:

mjbvz:
Yes please submit this against https://github.com/microsoft/vscode-markdown-tm-grammar

Here's an example PR showing a similar change: https://github.com/microsoft/vscode-markdown-tm-grammar/pull/182

Once the change is merged in upstream, we will pick it up here

spenserblack:
Thanks for pointing me in the right direction!

Issue #2878, opened 2016-02-10
Show other invisible characters with renderWhitespace option


sijad:
just in case atom also render CR, EOL characters
[0-31 and 127 ASCII code](https://en.wikipedia.org/wiki/ASCII#ASCII_control_code_chart) are non-printable characters
also #2085 is related

Platzer:
Is it possible to render not printable characters?

deadlyicon:
I just waisted 20m debugging a url problem because I had a Unicode Decimal Code &#8237; preceding my absolute url causing it to be treated as a relative path. 

Please allow us to enable rendering of these weird characters that no programmer ever wants in their code.

perrog:
+all other non-graphical code points.

Example:
# Create a source-code file containing U+200B (ZERO WIDTH SPACE) format character, and open it in VSCode.
> [char]0x200B | Out-File .\zero_width_space.ps1 -Encoding utf8
# File content in bytes:<ZERO WIDTH NON-BREAKING SPACE> + <ZERO WIDTH SPACE> + <CRLF>
> [io.file]::ReadAllBytes('.\zero_width_space.ps1')
239
187
191
226
128
139
13
10

Result:
VSCode does not display anything when opening the file, even when "editor.renderWhitespace"="all"
The first <ZERO WIDTH NON-BREAKING SPACE> is intepreted as UTF8 format.
The second <ZERO WIDTH SPACE> +<CRLF> is source code text.

Background.
I this case, the second code point <ZERO WIDTH SPACE> is a syntax error for the particular source-code.

The source-code contains syntax error => important to be able to show invisible characters in a code editor. 

In general, it may be useful to have an option to display Unicode-formatting characters in text file, if used explicitly in doc comments.

Formatting characters may also be accidental garbage when pasting text from a webbrowser.

Even the BOM-character could be shown as invisible character for full transparency.

alexdima:
This has been implemented in the meantime --> https://code.visualstudio.com/updates/v1_63#_unicode-highlighting

Fydon:
@alexdima I'm on 1.75.0-insider with `editor.renderWhitespace` set to `all`. I don't see CR and LF characters rendered in the editor. How do I show these?

alexdima:
@Fydon That is tracked in https://github.com/microsoft/vscode/issues/12223

PR #318850, opened 2026-05-28
Make Linux desktop sanity tests robust to mid-test failures

bryanchen-d:
Fixes microsoft/vscode-engineering#2878. Supersedes #318810.

### What was actually wrong

The Linux desktop sanity tests (deb / rpm / snap) install the package, run `testDesktopApp`, then uninstall — but the three steps were not wrapped in `try` / `finally`. When `testDesktopApp` threw (typically the "Failed to install extension after 3 attempts" flake tracked in microsoft/vscode-engineering#2877), the uninstall step was skipped.

Mocha is configured with `retries: 1`, so the test ran a second time. On the retry the install step hit `package code-insiders-... is already installed` and the run failed with that error instead of the original one. The build failure investigator picked up the misleading message and filed #2878 against the RPM install path, even though the real failure happened one attempt earlier in `testDesktopApp`.

### Fix

1. **Wrap install / test / uninstall in `try` / `finally`** so cleanup runs even when the test body throws. This is the actual bug.
2. **Make the uninstall helpers idempotent** by checking whether the binary exists before invoking the package manager. This is needed for two cases:
   - The `finally` block runs after a failed install, where nothing is actually installed yet.
   - The `finally` block runs after a successful pass through `testDesktopApp` that already cleaned up internally (not currently the case, but keeps the contract safe).
3. **Add a pre-install assertion** to each install helper that fails fast with a clear diagnostic when the package is already present. This is per @dbaeumer's review feedback on #318810 — leftover state should be surfaced as a real bug pointing at the *prior* failure, not silently worked around. The error message explicitly tells the reader to investigate the previous run instead of suppressing the error.

### Why not the previous PR (#318810)

The first PR swapped `rpm -i` for `rpm -U --replacepkgs` to make the install itself idempotent. That worked around the symptom but, as @dbaeumer pointed out, hid the underlying bug: a test should not be silently recovering from leftover state from a previous run. This PR addresses the root cause (missing cleanup on failure) and makes leftover state loud instead.

### Validation

- `npm run compile` in `test/sanity` passes.
- The change is test-infrastructure only; no production code is touched.
- Manual reasoning over the retry flow: with the `try` / `finally`, the retry now starts from a clean slate; if a future regression leaves the package installed across runs, the pre-install check fails immediately with a diagnostic pointing at the actual problem.

### DIFF
--- a/test/sanity/src/context.ts
+++ b/test/sanity/src/context.ts
@@ -884,22 +884,32 @@ export class TestContext {
 	 * @returns The path to the installed VS Code executable.
 	 */
 	public async installDeb(packagePath: string): Promise<string> {
+		const name = this.getLinuxBinaryName();
+		const entryPoint = path.join('/usr/share', name, name);
+		if (fs.existsSync(entryPoint)) {
+			this.error(`Cannot install ${packagePath}: ${name} is already installed at ${entryPoint}. This usually means a previous test run was terminated before cleanup completed; investigate the prior failure rather than silencing this error.`);
+		}
+
 		this.log(`Installing ${packagePath} using DEB package manager`);
 		await this.runDpkgNoErrors('-i', packagePath);
 		this.log(`Installed ${packagePath} successfully`);
 
-		const name = this.getLinuxBinaryName();
-		const entryPoint = path.join('/usr/share', name, name);
 		this.log(`Installed VS Code executable at: ${entryPoint}`);
 		return entryPoint;
 	}
 
 	/**
-	 * Uninstalls VS Code Linux DEB package.
+	 * Uninstalls VS Code Linux DEB package. Safe to call when the package is not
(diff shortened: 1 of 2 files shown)


dmitrivMS (approved):

dmitrivMS:
I am good with the change, although I am not sure how it helps the pipeline - if a test fails and leaves DEB/RPM/SNAP package installed, the change helps next test to run clean, but it won't make the pipeline pass...

bryanchen-d:
> I am good with the change, although I am not sure how it helps the pipeline - if a test fails and leaves DEB/RPM/SNAP package installed, the change helps next test to run clean, but it won't make the pipeline pass...

this helps clearing the confusing error message that it seems to do with the next test but actually it is from its previous test, so later analysor will be easier to categorize and count the flaky points.

PR #318862, opened 2026-05-28
Throttle initial Git repository.status() across repositories

### LINKED ISSUE #318279: Extension host repeatedly killed in large multi-root workspaces (vscode.git decoration storm)

- VS Code Version: 1.120.0
- OS Version: macOS 26.3.1 (darwin-arm64, Darwin 25.3.0)

Steps to Reproduce:

1. Create a multi-root `.code-workspace` containing around 50 folders, most of them git repos of varying sizes (small repos, larger codebases with Python venvs, a clone of microsoft/vscode).
2. Open it and wait. Don't open files or interact.

The extension host gets killed about 30 seconds after starting, auto-restarts, and dies again on the same cycle. `renderer.log` shows `Extension host (LocalProcess pid: N) terminated unexpectedly` with no stack and no signal. `exthost.log` shows several thousand `[Decorations] CAPPING events from decorations provider vscode.git` warnings firing in tight succession just before each kill. Cadence matches the IPC unresponsive watcher killing after missed pings.

Bisecting by halving the folder list and swapping folders at the boundary showed the trigger is aggregate vscode.git scan load, not any specific folder. Above a workspace-size threshold the host can't survive its own startup scan.

Ruled out:

- Extensions. Same crash with `--disable-extensions`.
- Folder identity. Removing the largest folder (microsoft/vscode clone), the most recently added folders, or individual suspect folders had no effect; only total count matters.
- File-watcher noise. Adding `**/.venv/**`, `**/__pycache__/**`, `**/.pytest_cache/**` to `files.watcherExclude` changed nothing.
- `git.autoRepositoryDetection: "openEditors"`. No effect, since explicit workspace folders still get scanned.

Workarounds:

- `git.enabled: false` at workspace level stops it (loses Source Control panel and decorations).
- Splitting the workspace and dropping enough folders to get below the threshold.

My workspace grew gradually, so I can't say if this is a recent regression or a longstanding threshold I just crossed. Either way, vscode.git's startup scan starves the host instead of degrading gracefully.

`renderer.log` (one cycle):
```
2026-05-25 21:14:15.585 [info] Started local extension host with pid 98481.
2026-05-25 21:14:35.663 [error] Extension host (LocalProcess pid: 98481) terminated unexpectedly. The following extensions were running: vscode.github-authentication, vscode.emmet, vscode.ipynb, vscode.git-base, vscode.npm, vscode.git, vscode.github, GitHub.copilot-chat, vscode.debug-auto-launch, vscode.merge-conflict, vscode.configuration-editing, vscode.json-language-features, vscode.typescript-language-features, vscode.terminal-suggest, vscode.microsoft-authentication, vscode.extension-editing
2026-05-25 21:14:35.697 [info] Started local extension host with pid 9281.
```

`exthost.log` (CAPPING warnings before the kill):
```
2026-05-25 21:14:21.950 [warning] [Decorations] CAPPING events from decorations provider vscode.git 330
2026-05-25 21:14:24.356 [warning] [Decorations] CAPPING events from decorations provider vscode.git 8585
2026-05-25 21:14:24.411 [warning] [Decorations] CAPPING events from decorations provider vscode.git 8585
2026-05-25 21:14:24.741 [warning] [Decorations] CAPPING events from decorations provider vscode.git 8585
```

### LINKED ISSUE #318764: Extension host crash loop (exit code 5) on startup when Copilot Chat is enabled in a large multi-root workspace
- Copilot Chat Extension Version: `github.copilot-chat@0.48.1`
- VS Code Version: 1.120.0 (commit 0958016b2af9f09bb4257e0df4a95e2f90590f9f, arm64)
- OS Version: macOS 26.2 (Darwin arm64)
- Feature (e.g. agent/edit/ask mode): N/A, the host crashes at startup before any chat session is opened
- Selected model (e.g. GPT 4.1, Claude 3.7 Sonnet): N/A, see above
- Logs: see below

Steps to Reproduce:

1. Open a multi-root `.code-workspace` containing ~50 folders (mix of git repos, several large; same workspace as microsoft/vscode#318279).
2. Have `github.copilot-chat@0.48.1` enabled. Do not interact with the window.
3. Wait ~25 to 30 seconds after the Extension Host starts.
4. Renderer reports `Extension host (LocalProcess pid: N) terminated unexpectedly`; main log shows `exit code: 5, signal: unknown`. Auto-restart starts a new host, which dies on the same cycle.

## Bisect

Reproduced with `code --disable-extension <id>` per launch, watching `~/Library/Application Support/Code/logs/<ts>/main.log` for ~75s after each Extension Host start:

| Launch | What was disabled (active 3rd-party set) | Outcome |
| --- | --- | --- |
| `--disable-extensions` | all 3rd-party | survive >80s |
| Round 1 | second half of activated extensions, alphabetical | crash ~26s |
| Round 2 | + claude-code, copilot-chat, postgres, prettier, biome | survive >80s |
| Round 3 | + only prettier, biome | crash ~21s |
| Round 4 | + postgres also | crash ~21s |
| Round 5 | + copilot-chat also | survive >80s |
| Re-enable all | (none, copilot-chat back on) | crash ~28s |

Disabling `github.copilot-chat` alone with every other extension enabled is sufficient. Re-enabling it brings the loop back.

## Logs (one cycle)

`main.log`:

```
2026-05-28 10:04:46.787 [info] Extension host with pid 40087 exited with code: 5, signal: unknown.
2026-05-28 10:04:46.787 [error] [UtilityProcess id: 2, type: extensionHost, pid: 40087]: crashed with code 5 and reason 'crashed'
```

`renderer.log`:

```
2026-05-28 10:04:19.398 [info] Started local extension host with pid 40087.
2026-05-28 10:04:46.790 [error] Extension host (LocalProcess pid: 40087) terminated unexpectedly. The following extensions were running: ..., GitHub.copilot-chat, ...
```

`exthost.log` (next cycle, right before next kill, only chat-related error):

```
2026-05-28 10:04:54.643 [error] Error calling provideChatSessionProviderOptions for handle 0: Canceled
```

Earlier the same day, a same-coalition Homebrew `node` child (thread named `chrome-devtools-mcp`) crashed with the classic V8 OOM stack (`node::OOMErrorHandler` to `v8::internal::V8::FatalProcessOutOfMemory`). Different process, but consistent memory-pressure signal in this VS Code instance, suggesting the exthost is also V8-OOMing.

## Possibly related

microsoft/vscode-copilot-release#14139 is also `exit code 5` in a Copilot Chat workflow on macOS, but that one is the *renderer* crashing during sustained Remote-SSH chat usage; this one is the *local extension host* crashing on startup. Likely a shared memory-pressure root cause.

## Workaround

Disable `github.copilot-chat`, per launch with `code --disable-extension github.copilot-chat`, or persistently via the Extensions sidebar.

dmitrivMS:
Fixes #318279
Fixes #318764

### Problem

In workspaces with many git repositories (~50+ folders in a multi-root `.code-workspace`), the extension host gets killed ~25–30s after startup, auto-restarts, and crashes in a loop. The reporter's `exthost.log` shows thousands of `[Decorations] CAPPING events from decorations provider vscode.git` warnings just before each kill, and the cadence matches the IPC unresponsive watcher firing after missed pings. `git.enabled: false` is a known workaround.

### Root cause

`Model.openRepository()` ends with a fire-and-forget `repository.status()` for every repo discovered during the initial scan. Each `_updateModelState()` fans out ~8 git subprocesses (`HEAD`, `remotes`, `submodules`, `worktrees`, `getStatus`, `getRefs`, ...) and then parses the output on the ext-host's single thread. With N repos opening through `Promise.all(... openRepository ...)` in `doInitialScan`, we get ~N×8 in-flight git subprocesses and a large blocking parse burst. For N≈50 that's enough to starve IPC pings and trip the unresponsive-host kill.

The decoration `CAPPING` warnings are a symptom of the same overload (each completed status fires a large decoration delta), not the cause.

### Fix

Throttle the initial fire-and-forget `repository.status()` calls across repositories via a model-level `Limiter<void>(5)`. The repo is still registered with SCM immediately (`this.open(repository)` runs unchanged); only the resource/decoration population is throttled, which was already async.

Concurrency 5 matches existing uses in [extensions/git/src/repository.ts](extensions/git/src/repository.ts#L1645) and [extensions/git/src/git.ts](extensions/git/src/git.ts#L2290) — enough parallelism to keep cores busy without saturating the host.

Applies to runtime opens too (workspace-folder additions, `eventuallyScanPossibleGitRepositories`), so the same crash can't be reproduced by adding folders incrementally.

### Verification

- Typecheck passes (`tsc -p extensions/git --noEmit`).
- No existing tests depend on unbounded parallelism (smoke test waits on an event).
- Manual repro on the reporter's workspace would be ideal — flagging @lszomoru.

### Not in scope

- The redundant `git rev-parse --show-toplevel` calls for subfolders of already-opened repos (previously explored in #313856) would further cut process count but don't change the crash dynamics; better as a follow-up so each change can be evaluated independently.

### DIFF
--- a/extensions/git/src/model.ts
+++ b/extensions/git/src/model.ts
@@ -7,7 +7,7 @@ import { workspace, WorkspaceFoldersChangeEvent, Uri, window, Event, EventEmitte
 import TelemetryReporter from '@vscode/extension-telemetry';
 import { IRepositoryResolver, Repository, RepositoryState } from './repository';
 import { memoize, sequentialize, debounce } from './decorators';
-import { dispose, anyEvent, filterEvent, isDescendant, pathEquals, toDisposable, eventToPromise } from './util';
+import { dispose, anyEvent, filterEvent, isDescendant, Limiter, pathEquals, toDisposable, eventToPromise } from './util';
 import { Git } from './git';
 import * as path from 'path';
 import * as fs from 'fs';
@@ -280,6 +280,9 @@ export class Model implements IRepositoryResolver, IBranchProtectionProviderRegi
 	get repositoryCache(): RepositoryCache {
 		return this._repositoryCache;
 	}
+
+	// Throttle initial repository.status() across repositories to avoid starving the ext host (#318279).
+	private readonly _initialStatusLimiter = new Limiter<void>(5);
 
 	private disposables: Disposable[] = [];
 
@@ -691,9 +694,8 @@ export class Model implements IRepositoryResolver, IBranchProtectionProviderRegi
 			this.logger.info(`[Model][openRepository] Opened repository (real path): ${repository.rootRealPath ?? repository.root}`);
 			this.logger.info(`[Model][openRepository] Opened repository (kind): ${gitRepository.kind}`);
 
-			// Do not await this, we want SCM
-			// to know about the repo asap
-			repository.status().then(() => {
+			// Do not await this, we want SCM to know about the repo asap.
+			this._initialStatusLimiter.queue(() => repository.status()).then(() => {
 				this._repositoryCache.update(repository.remotes, [], repository.root);
 			});
 		} catch (err) {


dmitrivMS:
Exploration build pipeline: https://dev.azure.com/monacotools/Monaco/_build/results?buildId=443081&view=results

lszomoru (approved):

dmitrivMS:
Tested with 30 repos and on Windows, it looks ok, but also asked customer to confirm if the crash is gone in their scenario.

PR #318879, opened 2026-05-29
Fix command center debug hover color

### LINKED ISSUE #193789: Command center with debug controls has strange mouse hover/active styling
https://github.com/microsoft/vscode/assets/22350/e0beee6a-419f-43e4-b35e-20cb84695d35

Two observations:

- When debugging, the control center becomes red but when hovering the "non-debug" grey background appears. I expected a different shade of red when hovering, when in that state
- Also, pressing down with the mouse button on an empty area of the control center will reveal hover state for the debug actions.

viraatdas:
Fixes #193789

This updates the command center theme override used while debugging so hover/active feedback keeps the debug color treatment instead of falling back to the normal titlebar active background.

Changes:
- add a `commandCenter.debuggingActiveBackground` theme color
- override `commandCenter.activeBackground` alongside `commandCenter.background` while the debug toolbar is in the command center

Verification:
- `git diff --check`

Disclosure: This PR was prepared with AI assistance, with the final change and validation reviewed before submission.


viraatdas:
Closing this duplicate because the contribution has been republished from the Libra Issues machine user: https://github.com/microsoft/vscode/pull/319070

PR #318927, opened 2026-05-29
fix: restore right-click rename for sessions in the Agents window (#318819)

### LINKED ISSUE #318819: Agents: Can't right-click rename sessions
Reported by @Yoyokrazy: rename is no longer an option in the right-click context menu of sessions in the Agents window. 

In editor windows, I've run into separate issues with renaming: https://github.com/microsoft/vscode/issues/318442.

katipally:
Fixes #318819

## What happened

The right-click rename option was no longer showing in the Agents window session list after the `CopilotSessionContextMenuBridge` was removed.

## What this fixes

Adds a dedicated context menu action `RenameCopilotSessionAction` that registers "Rename..." in the session item context menu for all copilot session types that support rename:

- `local` sessions
- `copilotcli` (background/CLI) sessions
- `claude-code` sessions

Cloud sessions (`copilot-cloud-agent`) are correctly excluded since the backend does not support renaming them.

## Changes

**`copilotChatSessionsActions.ts`**
- Added `RenameCopilotSessionAction` that shows "Rename..." in the context menu when `chatSessionProviderId === 'default-copilot'` and `chatSessionType !== 'copilot-cloud-agent'`
- Action uses `ISessionsManagementService.renameChat()` which already routes to the correct provider implementation per session type

**`copilotChatSessionsProvider.ts`**
- Added handling for `local` session type in `renameChat()` via `chatService.setChatSessionTitle()`

## Verification

Right-clicking a session in the Agents window now shows "Rename..." in the context menu.

### DIFF
--- a/src/vs/sessions/contrib/providers/copilotChatSessions/browser/copilotChatSessionsActions.ts
+++ b/src/vs/sessions/contrib/providers/copilotChatSessions/browser/copilotChatSessionsActions.ts
@@ -7,11 +7,12 @@ import { BaseActionViewItem } from '../../../../../base/browser/ui/actionbar/act
 import { Disposable, IDisposable } from '../../../../../base/common/lifecycle.js';
 import { IReader, autorun, observableValue } from '../../../../../base/common/observable.js';
 import { isWeb } from '../../../../../base/common/platform.js';
-import { localize2 } from '../../../../../nls.js';
+import { localize, localize2 } from '../../../../../nls.js';
 import { IActionViewItemService } from '../../../../../platform/actions/browser/actionViewItemService.js';
 import { Action2, registerAction2 } from '../../../../../platform/actions/common/actions.js';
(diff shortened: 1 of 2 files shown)


benibenj:
Thanks, but I already pushed a fix for this

katipally:
Ah, no worries, glad it's fixed! thanks for the heads up!

@benibenj  To be honest, the main issue I'm running into as a new contributor is figuring out the right way to propose changes without them overlapping with your internal work or getting closed. I have some feature ideas for the Agents window that I think would be great additions.

Could you point me to the standard process for proposing and getting alignment on these kinds of features before I write the code? Or, if there are any smaller 'good first issues' you recommend I start with instead, I'd be happy to take a look.

benibenj:
@katipally If you have feature request for the agents window, you can just open an issue and explain your idea. I do not recommend opening a PR before it gets discussed. But most likely we would implement it ourselves. We have labels on issues which are help-needed or good-first-issue

PR #318932, opened 2026-05-29
Improvements to Area/FullPage Screenshot to Chat

jruales:


src/vs/platform/browserView/electron-main/browserView.ts
```diff
@@ -644,30 +653,43 @@ export class BrowserView extends Disposable {
...
 		if (!size) {
 			throw new Error('Page.getLayoutMetrics did not return a cssContentSize');
 		}
 		const zoomFactor = this._view.webContents.getZoomFactor();
+		const clipWidth = size.width * zoomFactor;
+		const clipHeight = size.height * zoomFactor;
+		// CDP renders the screenshot at device pixels, so the output bitmap dimensions are roughly
+		// `clip.width * scale * devicePixelRatio`. Divide by DPR here so `MAX_FULL_PAGE_SCREENSHOT_DIMENSION`
+		// is an upper bound on the final image pixel size (not just the CSS-pixel clip size).
+		// We read the DPR from the display hosting the view's window (rather than evaluating
```
kycutler, line 669: Is this also in the info returned from `Page.getLayoutMetrics`? Agree with the Copilot comment that this wouldn't seem to account for device emulation

connor4312 (approved):

src/vs/platform/browserView/electron-main/browserView.ts
jruales: Will add that in follow-up PR. There are several fixes that need to be made for device emulation, even for Add Element to Chat when the zoom level isn't 100% in emulation

PR #319023, opened 2026-05-29
Update settings and docs for Execution Subagent rollout based on Gemini-3-Flash

vikramnitin9:
- Enables the Execution Subagent by default, and sets the default model to `gemini-3-flash`
- Disables the subagent if `gemini-3-flash` is not available.
- Updates `package.nls.json` to indicate that the subagent is powered by Gemini.


bhavyaus:
**Resolving the Gemini-3-Flash endpoint in `ExecutionSubagentToolCallingLoop`**

In `getEndpoint()` ([`executionSubagentToolCallingLoop.ts:133`](https://github.com/microsoft/vscode/blob/vikram/gemini-notes/extensions/copilot/src/extension/prompt/node/executionSubagentToolCallingLoop.ts#L133)) we call `endpointProvider.getChatEndpoint('gemini-3-flash')`, but the string overload only resolves `copilot-utility` / `copilot-utility-small` and throws otherwise ([`endpointProviderImpl.ts:138-147`](https://github.com/microsoft/vscode/blob/vikram/gemini-notes/extensions/copilot/src/extension/prompt/vscode-node/endpointProviderImpl.ts#L138-L147)), so the catch silently falls back to the main agent model, contradicting the new "powered by Gemini-3-Flash" string in `package.nls.json`.

Suggest looking it up via `getAllChatEndpoints()` (the same API the new gate in `agentIntent.ts` uses), e.g.:

```ts
const all = await this.endpointProvider.getAllChatEndpoints();
const gemini = all.find(ep => ep.family.toLowerCase().includes('gemini-3-flash') && ep.supportsToolCalls);
return gemini ?? await this.endpointProvider.getChatEndpoint(this.options.request);
```

Using the same predicate in both places also keeps the gate and the runtime lookup from drifting.

vikramnitin9:
@bhavyaus - actually, if Gemini is not available, then the execution subagent will be disabled, so the code will never even get to the `getEndpoint()` method. Would you still recommend modifying `getEndpoint`? If so, would you suggest throwing an error there, returning an `undefined` endpoint, or something else?

bhavyaus (approved):

PR #319123, opened 2026-05-30
Fix BYOK models disappearing during startup

### LINKED ISSUE #319121: Custom Endpoint (BYOK) models disappear from model picker on VS Code restart - hasByokModels.lastKnown race condition
## Describe the bug

After restarting VS Code, Custom Endpoint (BYOK) models that were previously added via Chat: Manage Language Models → Add Models → Custom Endpoint disappear from the model picker for several seconds, and sometimes never appear unless the user manually runs Developer: Reload Window.

This is caused by a startup timing race condition in the HasByokModelsContribution class.

## Steps to Reproduce

1. Add a Custom Endpoint model (e.g., DeepSeek, or any OpenAI-compatible endpoint) via the Language Models editor
2. Verify the model works and appears in the model picker as expected
3. Close all VS Code windows and reopen VS Code
4. Observe: The model picker initially shows only Copilot models - the Custom Endpoint (local) option is missing
5. Wait 5-15 seconds - the Custom Endpoint model eventually appears (but sometimes doesn't at all)

## Expected Behavior

Custom Endpoint models that have been successfully configured should appear in the model picker immediately upon VS Code startup, without any delay.

## Actual Behavior

The `chat.hasByokModels.lastKnown` storage key is restored as `false` on every cold startup, causing the `github.copilot.hasByokModels` context key to be `false`. The model picker hides all non-Copilot vendor models until the extension registration completes and triggers `_update()`.

## Technical Analysis

After inspecting the source code at `src/vs/workbench/contrib/chat/browser/hasByokModelsContribution.ts`, the issue is:

1. `_restore()` reads `chat.hasByokModels.lastKnown` from storage - this is `false` on first startup
2. Even though `chatLanguageModels.json` is already configured with valid Custom Endpoint models, the `hasByokModels` context key remains `false`
3. The `_update()` method only corrects this **after** `whenInstalledExtensionsRegistered()` completes
4. During this window, the model picker hides all BYOK models

The key logic in `_update()`:
```typescript
if (!this._extensionsRegistered) {
    if (this._contextKeyService.getContextKeyValue<boolean>(
        ChatContextKeys.nonCopilotLanguageModelsAreUserSelectable.key)) {
        this._setResult(true);
    }
    return;
}
```

The optimistic flip only triggers if `nonCopilotLanguageModelsAreUserSelectable` is set, which isnt always the case on cold start.

Additionally, for users behind an HTTP proxy (e.g. `http.proxy: http://localhost:65532`), the Custom Endpoint providers model validation may fail during startup if the proxy is not ready yet, causing the entire provider group to be silently dropped from the model cache.

## Environment

- VS Code Version: 1.122.1
- OS Version: Windows 11
- Extension: GitHub Copilot Chat 0.50.1
- Proxy configured: `http.proxy: http://localhost:65532`

## Additional Context

The `chat.hasByokModels.lastKnown` key is stored in `state.vscdb` (SQLite) at `%APPDATA%\Code\User\globalStorage\state.vscdb`. Manually setting this to `true` before restart forces the models to appear immediately, confirming the root cause is the stale `false` value.

Related code files:
- `src/vs/workbench/contrib/chat/browser/hasByokModelsContribution.ts`
- `extensions/copilot/src/extension/byok/vscode-node/customEndpointProvider.ts`
- `src/vs/workbench/contrib/chat/common/languageModels.ts`

EmirX3D:
Fixes a startup timing race where configured Custom Endpoint/BYOK models could be hidden from the chat model picker because chat.hasByokModels.lastKnown restored as false before extension registration or model resolution completed, by treating configured non-Copilot language model provider groups as an immediate positive BYOK signal and covering the cold-start case with a regression test.

Closes #319121


dmitrivMS (changes requested):
This is doing too much for the stated problem. The race needs to be addressed, but this is doing a lot more than fixing hasByokModels state...

KKKayleX:
Hi, thanks for working on this PR!

I have another issue that might be related in terms of behavior, though it may not be the same underlying root cause:
https://github.com/microsoft/vscode/issues/319243#issuecomment-4740237654

Could you help confirm whether this PR is expected to address my scenario as well, or whether it represents a different class of issue?

dmitrivMS:
Superceded by https://github.com/microsoft/vscode/pull/319190

Issue #317821, opened 2026-05-21
Feature Request: Support concurrent multiple GitHub accounts for different extensions (e.g., Copilot vs. Source Control)


Neizan93:
As an external consultant, I frequently face friction with the current GitHub authentication model in VS Code. My employer provides my GitHub Copilot license via my corporate account, but I must use my client's GitHub account to access their repositories, sync changes, and manage pull requests. Currently, VS Code struggles to maintain active, concurrent sessions for both accounts seamlessly in the same workspace, causing authentication conflicts between the Copilot extension and Git source control.

I would like to request a feature that allows authenticating multiple GitHub accounts simultaneously and explicitly assigning them to specific extensions or core features. For example, routing the GitHub Copilot extension to authenticate via Account A (Employer) while keeping the built-in Git Source Control and GitHub Pull Requests extensions authenticated via Account B (Client).

Current workarounds are heavily disruptive:
1. **Constant Sign-in/Sign-out:** Manually switching accounts depending on the immediate task degrades the developer experience.
2. **VS Code Profiles:** Using different profiles isolates environments, but fails to solve the core requirement of needing Copilot active in the exact same workspace where code is being committed to the client's repository.
3. **Browser-based PR management:** Delegating Git/PR management to the browser to keep VS Code strictly logged into the Copilot account defeats the purpose of integrated source control tools.

Enabling granular, per-extension GitHub authentication would resolve a major blocker for developers working in enterprise, B2B, or agency models where vendor licenses and source code access are split across different organizational domains.

asr-alurisanthoshreddy:
I’d like to work on this issue.

Neizan93:
It seems to be working now, although it's not very intuitive.

PR #319166, opened 2026-05-30
Support concurrent usage of multiple GitHub accounts for child extensions (#317821)

Neizan93:
### Summary of Proposed Changes

This PR addresses issue #317821 by introducing a hierarchical preference lookup and storage system in VS Code's workbench authentication services (`AuthenticationExtensionsService` and `AuthenticationMcpService`).

Currently, child extensions (e.g., GitHub Copilot) inherit their GitHub account preference from their parent extension (e.g., Git / GitHub Authentication). However, when selecting or changing an account for Copilot, the selection would overwrite the parent preference key. This made it impossible to concurrently use Account A for Git and Account B for Copilot in the same workspace.

To resolve this:
1. **Hierarchical Storage (`updateAccountPreference` / `removeAccountPreference`):** Preferences are now written and removed using the specific child extension's ID key (e.g., `github.copilot-github`) instead of resolving directly to the parent's key. This keeps their preferences isolated.
2. **Default Inheritance (`getAccountPreference`):** When looking up account preferences, the service first checks for the specific extension's own ID. If not found, it falls back to the parent extension's key. This preserves the status quo and backward compatibility for single-account setups.
3. **Event Scoping:** When a preference is updated, change events are fired specifically for the child extension if it is a child, or for both the parent and its children if it's the parent.

An identical update is applied to the parallel `AuthenticationMcpService` class to support the same hierarchical behavior for MCP servers.

---


Neizan93:
AI generated, not compliant

PR #319247, opened 2026-06-01
Fix markdown cell not re-rendering when re-executed (#151151)

Tatlatat:
A markdown cell with an `` pointing to a missing file shows a broken image. After the file is created and the cell is re-executed/re-rendered without editing its text, the image stayed broken because the preview was not re-rendered when its content was unchanged.

Thread a `forceRerender` flag from the view layer down to the webview: `MarkupCell` now forces a re-render whenever the cell leaves edit mode (`editStateChanged`) — covering Enter, the toolbar, and Escape — so that `showMarkupPreview` bypasses its unchanged-content short-circuit and the webview re-fetches resources such as `` tags. Content/metadata-only, layout, scroll and selection changes are unaffected.


Tatlatat:

Vào Thứ 2, 1 thg 6, 2026 lúc 09:54 Copilot ***@***.***> đã
viết:

PR #319287, opened 2026-06-01
Include pty.node for mxc

chrmarti:
#317981


extensions/copilot/package.json
```diff
@@ -6903,6 +6903,7 @@
     "@typescript-eslint/eslint-plugin": "^8.35.0",
     "@typescript-eslint/parser": "^8.32.0",
     "@typescript-eslint/typescript-estree": "^8.26.1",
+    "@typescript/native-preview": "^7.0.0-dev.20260429",
```
rzhao271, line 6906: Where is this used?

chrmarti: I see the Copilot Chat recovery pipeline fail without that: https://dev.azure.com/monacotools/Monaco/_build/results?buildId=443776&view=logs&j=2a81633e-e3f5-564c-d9b7-29d4c6c01530&t=1c35f05c-d451-522f-7181-861f09ccaca9

It seems this was previously picked up from the root package.json, but that only gets installed when there is a build cache miss.

Yoyokrazy (approved):

Issue #316867, opened 2026-05-17
Security: minimumReleaseAge setting for mitigating supply chain attacks on extensions


karlhorky:
In the last years, supply chain attacks have increased dramatically. A few examples in the VS Code extension ecosystem:

- [AI-Slop ransomware test sneaks on to VS Code marketplace - BleepingComputer](https://www.bleepingcomputer.com/news/security/ai-slop-ransomware-test-sneaks-on-to-vs-code-marketplace/)
- [Malicious VS Code AI Extensions With 1.5M Installs Found Exfiltrating Code - The Hacker News](https://thehackernews.com/2026/01/malicious-vs-code-ai-extensions-with-15.html)
- [Fake Clawdbot VS Code Extension Installs ScreenConnect RAT - Aikido](https://www.aikido.dev/blog/fake-clawdbot-vscode-extension-malware)
- [Solidity Devs Targeted Again: Malicious VS Code Extension Drops ScreenConnect RAT - Checkmarx](https://checkmarx.com/zero-post/solidity-devs-targeted-again-malicious-vs-code-extension-drops-screenconnect-based-remote-access-trojan-rat/)
- [VS Code extensions contain trojan-laden image - ReversingLabs](https://www.reversinglabs.com/blog/malicious-vs-code-fake-image)
- [Malicious VSCode extensions on Microsoft's registry drop infostealers - BleepingComputer](https://www.bleepingcomputer.com/news/security/malicious-vscode-extensions-on-microsofts-registry-drop-infostealers/)
- [Malicious VSCode extensions infect Windows with cryptominers - BleepingComputer](https://www.bleepingcomputer.com/news/security/malicious-vscode-extensions-infect-windows-with-cryptominers/)
- [Nx Compromised to Steal Wallets and Credentials - Semgrep](https://semgrep.dev/blog/2025/security-alert-nx-compromised-to-steal-wallets-and-credentials)
- [Supply Chain Risk in VS Code Extension Marketplaces - Wiz](https://www.wiz.io/blog/supply-chain-risk-in-vscode-extension-marketplaces)
- [3,800 internal repos at GitHub breached via a malicious VS Code extension: why developer devices are the real target](https://www.aikido.dev/blog/github-breached-vs-code-extension)

Allow for `minimumReleaseAge`-style security config from other package managers / updaters, to require a minimum age on VS Code extensions before they are installed (also when installed via CLI, etc) when either:

1. Installing the latest version
2. Updating an extension manually
[...]

PR #319340, opened 2026-06-01
extensions: delay auto-update of new versions by 2 hours

sandy081:
## What

Delays auto-updating an extension until **2 hours** after a new version has been published. Manual updates are unaffected and update immediately.

## Why

This reverts the `delayed` value that was briefly rolled out to insiders on the `extensions.autoUpdate` setting (PR #318974) and instead applies the delay automatically, without any user-facing configuration. The intent is to give freshly published versions a short bake time before they are auto installed, while still letting users update manually whenever they want.

## Changes

- Reverts `extensions.autoUpdate` to its original form: boolean enum `[true, 'onlyEnabledExtensions', false]`, default `true`, no policy, no `delayed` value, no experiment block, no policy context key.
- Always delays auto-update of a new version by 2 hours after it is published (`getAutoUpdateDelayRemaining` = `max(0, 2h - elapsed)`; `isAutoUpdateDelayed` = `outdated && shouldAutoUpdate && remaining > 0`). The auto-update loop schedules a re-check at the soonest remaining delay.
- While delayed, the extension is shown as outdated with an **info** status explaining when it will be auto updated, and it is **excluded from the activity badge**. Manual update is always allowed.
- **Migration**: rewrites the insiders `extensions.autoUpdate` values back to booleans (`'on'`/`'delayed'` -> `true`, `'off'` -> `false`), with a runtime fallback in `getAutoUpdateValue` for values not yet migrated.
- `IExtensionsConfiguration.autoUpdate` typed as `AutoUpdateConfigurationValue`.
- Removes the `ExtensionsAutoUpdate` entry from `build/lib/policies/policyData.jsonc`.
- Tests reworked to drive the 2 hour window (recently published -> delayed; >2h / future / missing timestamp -> not delayed).

## Reviewer notes

- Unit tests could not be run locally (corrupted Electron) — relying on CI. `compile-check-ts-native`, `valid-layers-check` and hygiene are clean.

### DIFF
--- a/build/lib/policies/policyData.jsonc
+++ b/build/lib/policies/policyData.jsonc
@@ -358,40 +358,6 @@
             "included": true
         },
         {
-            "key": "extensions.autoUpdate",
-            "name": "ExtensionsAutoUpdate",
-            "category": "Extensions",
-            "minimumVersion": "1.122",
-            "localization": {
-                "description": {
-                    "key": "extensions.autoUpdate",
-                    "value": "Controls the automatic update behavior of extensions. The updates are fetched from a Microsoft online service."
-                },
-                "enumDescriptions": [
-                    {
-                        "key": "extensions.autoUpdate.on",
-                        "value": "Enabled extensions are updated automatically."
-                    },
-                    {
-                        "key": "extensions.autoUpdate.delayed",
-                        "value": "Enabled extensions are updated automatically, but only 6 hours after a new version has been published."
-                    },
-                    {
-                        "key": "extensions.autoUpdate.off",
-                        "value": "Extensions are not updated automatically."
-                    }
-                ]
-            },
-            "type": "string",
-            "default": "on",
-            "enum": [
-                "on",
-                "delayed",
-                "off"
-            ],
-            "included": true
-        },
-        {
             "key": "chat.tools.terminal.enableAutoApprove",
             "name": "ChatToolsTerminalEnableAutoApprove",
             "category": "IntegratedTerminal",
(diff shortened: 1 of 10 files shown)


benibenj (approved):

karlhorky:
> This reverts the `delayed` value that was briefly rolled out to insiders on the `extensions.autoUpdate` setting (PR #318974) and instead applies the delay automatically, without any user-facing configuration

@sandy081 what was the reason that user-facing configuration wasn't rolled out to stable?

Was it to allow for a better solution like the `minimumReleaseAge` model?

The `minimumReleaseAge` config proposal has gathered almost 300 community upvotes within 3 weeks:

- https://github.com/microsoft/vscode/issues/316867

acdha:
Seconding @karlhorky, the general industry seems to be settling on a one-week cooldown which seems prudent since that avoids problems around delayed responses on weekends or holidays. It would be great if VSC didn’t need to get a waiver for institutional policies around this.

DanielRuf:
2 hours is not enough. Reverting the change which could have allowed users to configure that makes it harder to keep control of this.

For enterprise users you have pinning via company-wide settings: https://code.visualstudio.com/docs/enterprise/extensions

The solution from https://www.kenmuse.com/blog/pinning-vscode-extensions-to-fight-supply-chain-attacks/ is also just some workaround.

Update: there is now https://github.com/microsoft/vscode/pull/320046 which introduces a configuration setting to set a different value as needed.

karlhorky:
Looks like there will be a setting in VS Code 1.125 🎉

> @sandy081 in [comment #4691382342](https://github.com/microsoft/vscode/issues/316867#issuecomment-4691382342): This is now largely addressed by [#321039](https://github.com/microsoft/vscode/pull/321039) (merged) 🎉
> 
> VS Code now supports a configurable **`extensions.autoUpdateDelay`** setting (in hours, default 2) that controls how long to wait after an extension update is published before auto-installing it. Both `extensions.autoUpdate` and `extensions.autoUpdateDelay` are also **policy-enabled** for enterprise/MDM management.
> 
> One part of this request is still open: the delay currently applies only to **auto-updates**, not to fresh **installs** (marketplace UI, `code --install-extension`, devcontainers). We're tracking that separately in [#321136](https://github.com/microsoft/vscode/issues/321136).
> 
> Thanks everyone for the strong community support here.

Source: https://github.com/microsoft/vscode/issues/316867#issuecomment-4691382342

PR #319345, opened 2026-06-01
Scope Agents Window chat input history per session

### LINKED ISSUE #319341: Agents Window chat input history is shared across sessions
When using the Agents Window with multiple sessions open, chat input history is shared across sessions.

## Repro

1. Open the Agents Window.
2. Open or create two different sessions.
3. Send a few prompts in the first session.
4. Switch to the second session and send different prompts there.
5. Press Up Arrow in the chat input to navigate previous prompts.

## Actual

The input history can include prompts from other sessions. For example, pressing Up Arrow in session B can show prompts that were sent in session A.

## Expected

Chat input history should be scoped to the current session. Pressing Up/Down should only navigate prompts submitted in that same session.

## Demo

I have a video below showing the shared-history behavior. (The sessions were in two different folders)

https://github.com/user-attachments/assets/49f2c394-5709-4088-8e0f-ef581c62254c

federicobrancasi:
Fixes #319341

Scopes chat input history by session id, so Up/Down history navigation only shows prompts from the current session.

## What Changed

- Added scoped history keys to the shared chat widget history service.
- Updated chat history navigators so they only react to changes for their own location/history scope.
- Scoped Agents Window chat inputs by `ISession.sessionId`.
- Preserved history when a temporary session is replaced by its committed session.
- Added unit tests for scoped history separation and history moves.
- Documented the Agents Window history behavior in `SESSIONS.md`.

## Demo

Before/after video attached below:

### Before

https://github.com/user-attachments/assets/1ac30533-4574-42e5-833e-ce0a57b1c13a

### After

https://github.com/user-attachments/assets/754c3df6-2365-423b-878b-831717277748


connor4312 (changes requested):
I would at least want a setting to turn this back to the old behavior. Not infrequently I send a message to a session and then decide for whatever reason (context size, concurrency, or just giving the same prompt to multiple models to compare them) to actually start a new session with that message or a derivation of that message. A manual copy+paste isn't always sufficient for this as it does not include the attachments like history does.

federicobrancasi:
> I would at least want a setting to turn this back to the old behavior. Not infrequently I send a message to a session and then decide for whatever reason (context size, concurrency, or just giving the same prompt to multiple models to compare them) to actually start a new session with that message or a derivation of that message. A manual copy+paste isn't always sufficient for this as it does not include the attachments like history does.

@connor4312 Fixed in 6280c9fde4c8be3ec1a79267504d65f24d582166.

I added `chat.agentSessions.scopedInputHistory`: it defaults to `true` for the new per-session history behavior, and setting it to `false` restores the old shared input history across Agents Window sessions.

Screenshot 2026-06-01 at 5 36 14 PM

src/vs/sessions/contrib/chat/browser/sessionsChatHistoryContribution.ts
```diff
@@ -0,0 +1,95 @@
...
+
+	private _registerWidget(widget: IChatWidget): void {
+		if (this._registeredWidgets.has(widget)) {
+			return;
+		}
+
+		this._registeredWidgets.add(widget);
+		const store = new DisposableStore();
+		store.add(widget.onDidChangeViewModel(() => this._applyHistoryKeys()));
+		this._widgetDisposables.add(store);
```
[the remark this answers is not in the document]
federicobrancasi: Thanks Copilot ;)

fixed in 6280c9fde4c8be3ec1a79267504d65f24d582166

src/vs/sessions/contrib/chat/browser/sessionsChatHistoryContribution.ts
```diff
@@ -0,0 +1,106 @@
...
+import { IChatWidget, IChatWidgetService } from '../../../../workbench/contrib/chat/browser/chat.js';
+import { ChatAgentLocation } from '../../../../workbench/contrib/chat/common/constants.js';
+import { IChatWidgetHistoryService } from '../../../../workbench/contrib/chat/common/widget/chatWidgetHistoryService.js';
+import { IWorkbenchContribution } from '../../../../workbench/common/contributions.js';
+import { ISessionsManagementService } from '../../../services/sessions/common/sessionsManagement.js';
+import { ISession } from '../../../services/sessions/common/session.js';
+
+export const AGENT_SESSIONS_SCOPED_INPUT_HISTORY_SETTING = 'chat.agentSessions.scopedInputHistory';
+
+export class SessionsChatHistoryContribution extends Disposable implements IWorkbenchContribution {
```
connor4312: I'm a bit confused why we need this new contribution

- This applies the session key to inputs parts when the active session changes. But input parts are already well aware of the session they have via the `_inputModelSessionResource`, having a service tell them this separately seems to add more moving parts and complexity
- Moving history via `moveHistory`. But in a world where the history of a chat is per session, do we even need to _store_ history separately? Could we instead just iterate through past user requests to the model to hydrate the history navigator when it's loaded?

federicobrancasi: You are right, this was more indirect than needed.

Fixed in f2e50f08305. I removed the extra `SessionsChatHistoryContribution` widget-scanning path and now pass the session id through the normal `SessionView` -> `ChatView` flow.

I kept stored history because it preserves full input state, including attachments, and still supports `chat.agentSessions.scopedInputHistory: false` for the old shared-history behavior.

connor4312 (approved):

PR #319353, opened 2026-06-01
Add a "Sandboxing for terminal" toggle to the Default Approvals permission option

alexdima:
🎥  below -- adds a UI toggle for sandboxing for the local harness in the picker and also clarifies in the label of the mode that sandboxing is on. This whole thing is off by default and we can turn it on via an experiment. 

I will try to further tweak tomorrow how things look, especially the `Default Approvals (Sandboxed)` part which looks pretty ugly. Ideas welcome!

---

🤖 

## Overview

Adds an inline **"Sandboxing for terminal"** toggle to the **Default Approvals** option of the permissions picker. Instead of two separate Default Approvals rows (plain vs. sandboxed), there is now a single Default Approvals option with the sandbox switch contained inside it.

## Permissions picker changes

- The Default Approvals option hosts an inline pill switch ("Sandboxing for terminal") on its own row, in both the open dropdown and reflected on the closed trigger button (which reads "Default Approvals (Sandboxed)" when sandboxing is on).
- The toggle reflects and updates **`chat.agent.sandbox.enabled`** (`on` / `off`).
- Clicking the toggle flips sandboxing without selecting/closing the row; clicking the row still selects Default Approvals as before.
- Enterprise policy gating, the "Learn more about permissions" footer, and the existing Bypass/Autopilot levels are unchanged.

## Gating

- The whole sandbox-toggle UI is behind a new experimental setting **`chat.experimental.permissionsSandboxToggle.enabled`** (off by default), since it is a significant UI change.
- The toggle is shown **only for the Local harness**. Agent-host harnesses (Copilot CLI, Claude Code, Codex, …) opt out via the picker delegate and never show it.

## New reusable action-list primitive

- Adds a generic `inlineToggle` capability to the action-widget list items (`actionList` / `actionWidgetDropdown`): an item can host an always-visible pill switch on its own row, with appropriate item height, click-to-select exclusion, and an accessible label that announces the switch state.
- Adds the supporting CSS for the iOS-style pill switch (themed, with high-contrast support) and a `className` passthrough on dropdown actions.

## Tests / fixtures

- New `permissionPickerList` component fixture covering the three states: toggle hidden, shown + off, shown + on.
- `chatInput` gains a `WithSandboxing` variant that renders the local harness in the sandboxed state; fixture service mocks were extended to support it.

## Follow-ups

- Removing the `chat.agent.sandbox.enabled: "network"` value is intentionally left for a separate PR.

https://github.com/user-attachments/assets/dd49377a-3ccf-47ad-82a0-d95f2db71a9a


justschen (approved):
thanks thanks!

PR #319365, opened 2026-06-01
Pass remoteSession per-session to actually enable session sync export

digitarald:
## Problem

PR #317390 enabled session sync by passing `enableRemoteSessions: true` at the **client level** (`CopilotClientOptions`). This starts the CLI with `--remote`, which enables the *capability* for remote sessions.

However, the SDK requires **each individual session** to opt in via the `remoteSession` field on `SessionConfig` / `ResumeSessionConfig`. Without it, sessions default to `"off"` and no events are exported.

## Fix

Add `remoteSession: 'export'` to `_buildSessionConfig` (the shared config builder used by both create and resume paths) so sessions opt in when `chat.sessionSync.enabled` is true.

The `RemoteSessionMode` values are:
- `"off"` — no export (default)
- `"export"` — export session events to GitHub without remote steering
- `"on"` — export AND enable remote steering


digitarald:
fyi @vijayupadya @DonJayamanne

vijayupadya (approved):

PR #319424, opened 2026-06-01
agents window prototype: decouple editor and Changes/Files panel; consolidate toggle

eli-w-king:
## Summary

https://github.com/user-attachments/assets/09c119eb-4368-4fe2-9e1d-d9ba8d0b9dcf

Agents window prototype: decouples the editor part and the Changes/Files auxiliary bar so they read as two independent cards, consolidates the Changes/Files toggle behind the existing titlebar button, and refines Maximize Editor Area to keep the chat in view.

### Titlebar toggle behavior
- The `ToggleSecondarySidebarVisibilityAction` in the Agents window now delegates to the existing `workbench.action.agentSessions.toggleChangesView` command, so the titlebar button and the Show/Hide Changes action behave identically.
- The Show/Hide Changes action only hides the auxiliary bar now; it no longer also hides the editor part. This makes the toggle non-destructive when the integrated browser is open in the editor.
- Opening a link from the session no longer force-opens the Changes/Files panel.

### Maximize Editor Area
- Maximize hides the primary sidebar (workspaces) and the auxiliary bar (Changes/Files) while keeping the chat (sessions) part visible. Restore brings back the prior visibility of both side panels.

### Layout decoupling
Both the editor part and the auxiliary bar (Changes/Files) now render as fully-rounded cards with a 5px margin on the shared edge, replacing the previous flush-against-each-other appearance. The grid sash sits centered in the 10px gutter between them.

- **Editor part**: 5px right margin always; drops to 0 when the aux bar is hidden so the editor fills the remaining horizontal space. Full `border-radius: 8px` on all corners.
- **Auxiliary bar**: 5px left margin always (removed the editor-visibility padding swap and the `.nomaineditorarea` override). Full `border-radius: 8px` on all corners.
- **Sessions part**: keeps a 5px right margin even when the aux bar is hidden, so it reads as a card on its own.
- **Animation**: 200ms ease transition on the editor part's margin, gated on `prefers-reduced-motion`.
- **TS `layout()`** in all affected parts updated to mirror the new CSS so inner content (e.g. the integrated browser webview, chat message rows) sits cleanly inside the rounded card on both sides without horizontal overflow.

## Visual

Editor area and Changes/Files panel are now visually independent cards with a sash centered in a 10px gutter; toggling the panel cleanly animates the editor's right edge.


mrleemurray:
@benibenj what do you think about this? Previously you said the editor area and auxiliary bar need to show at the same time - does this proposal work?

benibenj:
Yes, currently editor area should only show when the secondary side bar also shows. If we want to change that we should all align on this as it will need more changes to make it work nicely.

Can you clarify what is the driver for this change? What's the issue we are trying to fix? Why is the maximize editor area not good enough for this scenario? 

I also find using that icon confusing as we already use it in the titlebar

PR #319430, opened 2026-06-01
agentHost: move rubber duck flag to agent host config

connor4312:
- Migrates the rubber duck agent toggle off environment variables and VS Code settings so it lives with the other agent host customization config, giving it a single consistent source of truth that the host owns rather than scattering the flag across starter processes.
- Applies the flag to the Copilot CLI subprocess env at client startup and restarts the client (disposing any active sessions) when the config changes, so toggling it actually takes effect without a manual restart.

Supplants https://github.com/microsoft/vscode/pull/319347


digitarald (approved):

ntrogh:
@connor4312 @digitarald How should users configure this setting?

PR #319460, opened 2026-06-01
Fix #236814: Auto-escape text when pasting inside JSON string literals

### LINKED ISSUE #236814: Feature request: Automatically Escape Pasted Text in JSON string
I have a need to paste a large amount of text into a JSON string via the clipboard, but I hope it can automatically escape the clipboard content, which would be very convenient for pasting large amounts of text.

Here is a use case:

1. I want to set the `github.copilot.chat.commitMessageGeneration.instructions` setting

    ```json
    {
      "github.copilot.chat.commitMessageGeneration.instructions": [
        {
          "text": ""
        }
      ]
    }
    ```

2. I have some text that I want to paste

    ```md
    The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in [RFC 2119](https://www.ietf.org/rfc/rfc2119.txt).

    1. Commits MUST be prefixed with a type, which consists of a noun, `feat`, `fix`, etc., followed by the OPTIONAL scope, OPTIONAL `!`, and REQUIRED terminal colon and space.
    2. The type `feat` MUST be used when a commit adds a new feature to your application or library.
    3. The type `fix` MUST be used when a commit represents a bug fix for your application.
    4. A scope MAY be provided after a type. A scope MUST consist of a noun describing a section of the codebase surrounded by parenthesis, e.g., `fix(parser):`
    5. A description MUST immediately follow the colon and space after the type/scope prefix. The description is a short summary of the code changes, e.g., *fix: array parsing issue when multiple spaces were contained in string*.
    6. A longer commit body MAY be provided after the short description, providing additional contextual information about the code changes. The body MUST begin one blank line after the description.
    7. A commit body is free-form and MAY consist of any number of newline separated paragraphs.
    8. One or more footers MAY be provided one blank line after the body. Each footer MUST consist of a word token, followed by either a `:<space>` or `<space>#` separator, followed by a string value (this is inspired by the [git trailer convention](https://git-scm.com/docs/git-interpret-trailers)).
    9. A footer's token MUST use `-` in place of whitespace characters, e.g., `Acked-by` (this helps differentiate the footer section from a multi-paragraph body). An exception is made for `BREAKING CHANGE`, which MAY also be used as a token.
    10. A footer's value MAY contain spaces and newlines, and parsing MUST terminate when the next valid footer token/separator pair is observed.
    11. Breaking changes MUST be indicated in the type/scope prefix of a commit, or as an entry in the footer.
    12. If included as a footer, a breaking change MUST consist of the uppercase text BREAKING CHANGE, followed by a colon, space, and description, e.g., *BREAKING CHANGE: environment variables now take precedence over config files*.
    13. If included in the type/scope prefix, breaking changes MUST be indicated by a `!` immediately before the `:`. If `!` is used, `BREAKING CHANGE:` MAY be omitted from the footer section, and the commit description SHALL be used to describe the breaking change.
    14. Types other than `feat` and `fix` MAY be used in your commit messages, e.g., *docs: update ref docs.*
    15. The units of information that make up Conventional Commits MUST NOT be treated as case sensitive by implementors, with the exception of BREAKING CHANGE which MUST be uppercase.
    16. BREAKING-CHANGE MUST be synonymous with BREAKING CHANGE, when used as a token in a footer.
    ```

3. I want to paste this text into the JSON settings, so I will move the keyboard cursor between `""` and then press `Ctrl + V` to paste the text

    ```json
    {
      "github.copilot.chat.commitMessageGeneration.instructions": [
        {
          "text": "|"
        }
      ]
    }
    ```

4. At this point, I hope VS Code can automatically recognize this paste action and automatically escape the pasted text, allowing me to directly paste the text into the JSON settings

    ```json
    {
      "github.copilot.chat.commitMessageGeneration.instructions": [
        {
          "text": "The key words \"MUST\", ......"
        }
      ]
    }
    ```

dripfyyy:
## Problem
In JSON and JSONC files, pasting text inside a string literal could insert the clipboard contents verbatim. That meant quotes, backslashes, newlines, and tabs could break the JSON unless they were escaped manually. This was tracked in #236814.
## Fix
Introduce a JSON paste auto-escape provider backed by a shared parser/cache service:
- detect whether the paste target is inside a JSON string
- escape pasted text only in that case
- keep JSON/JSONC paste behavior unchanged outside strings
- support the existing `json.editor.autoEscapePaste` setting
- reuse cached parse snapshots and invalidate them incrementally on document changes

Additional adjustments:
- register the provider for both the JSON paste kind and plain text paste kind
- ensure the first paste is handled correctly
- keep parser state in sync on document change and close events
## Testing
Added tests to verify:
- quotes are escaped inside JSON strings
- backslashes are doubled
- multiline paste becomes escaped newlines
- tabs become `\t`
- paste outside a JSON string is unchanged
- disabling `json.editor.autoEscapePaste` prevents escaping
- parser snapshot reuse and invalidation work as expected
## Result
Pasting into JSON string values now produces valid escaped content automatically, while paste behavior elsewhere remains unchanged. The implementation is also more efficient and easier to maintain because the parsing and cache logic is centralized.


aeschli:
Thanks for looking into it, but I feel the change will add too much complexity for a feature which is not high on the priority list for us.
- it doesn't reuse our existing parsing infrastructure which is all done on the server side
- it touches several unrelated files

I want to suggest that this feature is implemented in an extension or you find a simpler approach

PR #319496, opened 2026-06-02
Preserve BYOK Ollama reasoning effort metadata from /api/show

### LINKED ISSUE #319492: BYOK Ollama drops reasoning effort metadata returned by /api/show
## Summary

The VS Code Copilot BYOK Ollama provider currently loses reasoning-effort metadata even when the Ollama-compatible endpoint returns it correctly from `/api/show`.

This breaks custom gateways such that already expose:

- `supportedReasoningEfforts`
- `defaultReasoningEffort`
- `supportsReasoningEffort`
- `capabilities.supports.reasoning_effort`

As a result, the model can appear in the picker, but the reasoning selector is not exposed in the UI.

## Affected area

- `extensions/copilot/src/extension/byok/vscode-node/ollamaProvider.ts`

## Current behavior

The Ollama discovery flow calls:

1. `GET /api/version`
2. `GET /api/tags`
3. `POST /api/show`

The endpoint can return reasoning metadata in the `POST /api/show` payload, but the provider does not preserve that metadata consistently when building the BYOK model info.

That means downstream code that already knows how to render the reasoning schema never receives the required values.

## Expected behavior

When `/api/show` returns reasoning metadata, the Ollama BYOK provider should preserve it and pass it through to the BYOK model info so the reasoning-effort selector can be generated normally.

## Reproduction

Use an Ollama-compatible endpoint where `/api/show` returns a payload like:

```json
{
  "supportedReasoningEfforts": ["low", "medium", "high"],
  "defaultReasoningEffort": "medium",
  "capabilities": {
    "supports": {
      "reasoning_effort": ["low", "medium", "high"]
    }
  }
}
```

Then open the BYOK Ollama flow in Copilot.

Observed result:
- the model is listed
- the reasoning selector is missing

Expected result:
- the model is listed
- the reasoning selector is available based on the metadata returned by `/api/show`

## Root cause

The common BYOK model pipeline already supports `supportsReasoningEffort`, and the schema builder already knows how to surface it.

The gap is in the Ollama provider path, which fetches model information from `/api/show` but does not fully preserve the reasoning metadata when storing/building the model info object.

## Proposed fix

In `ollamaProvider.ts`, preserve the reasoning-related fields returned by `/api/show` and forward them into the BYOK model info capabilities.

Concretely, the provider should carry through fields such as:

- `supportsReasoningEffort`
- `supportedReasoningEfforts`
- `defaultReasoningEffort`
- `capabilities.supports.reasoning_effort`

so that `byokModelInfo.ts` can build the configuration schema as intended.

## Validation

A focused regression test can cover this in:

- `extensions/copilot/src/extension/byok/vscode-node/test/ollamaProvider.spec.ts`

with an assertion that reasoning metadata returned by `/api/show` is preserved in the final BYOK model info.

## Context

This was reproduced while integrating an Ollama-compatible BYOK backend. The backend already returns the correct metadata, but the current VS Code Ollama provider drops it before the UI schema layer can use it.

zhiru:
Fixes #319492

## Summary
- preserve `supportedReasoningEfforts` and `defaultReasoningEffort` returned by `/api/show`
- forward reasoning-effort support into the BYOK Ollama model capabilities
- add a focused regression test for the Ollama provider

## Testing
- `get_errors` clean for the touched files
- `npm run test:unit -- src/extension/byok/vscode-node/test/ollamaProvider.spec.ts` from `extensions/copilot/` currently fails in this clone because `vitest` is not installed locally (`sh: 1: vitest: not found`)


zhiru:

cfradewavecom:
+1

lramos15:
We have deprecated the in house ollama support in favor of an extension maintained by the ollama team. Please feel free to file an issue [here](https://github.com/ollama/ollama-vscode)

PR #319506, opened 2026-06-02
Update chat input send button to be circular

hawkticehurst:
Update VS Code core chat input send button to match agents window send button styling.

Screenshot 2026-06-01 at 11 28 52 PM


justschen (approved):
SORRY JUST SAW THIS LGTM

PR #319507, opened 2026-06-02
Update chat session icons and order

hawkticehurst:
Update VS Code core chat session list action buttons so they align with agents window:

- The archive button uses a check icon
- Swap order of archive and pin buttons

Screenshot 2026-06-01 at 11 41 07 PM


justschen (approved):
perfect! hope to get some feedback on this but should be all good tbh! thanks!

PR #319569, opened 2026-06-02
Fix cpuUsage.sh error when a process exits during measurement

### LINKED ISSUE #186502: cpuUsage.sh: line 51: let: PROCESS_TIME_AFTER=+: syntax error: operand expected (error token is "+")

- VS Code Version: 1.79.2
- OS Version: Ubuntu 23.04

### Steps to Reproduce:

1. Open the process explorer
2. Open output channel for "Main"

Sometimes an error is logged to the output channel: `/snap/code/132/usr/share/code/resources/app/out/vs/base/node/cpuUsage.sh: line 51: let: PROCESS_TIME_AFTER=+: syntax error: operand expected (error token is "+")`

### Additional information

By default shell scripts continue even on error. Perhaps it would be a good idea to add `set -e` to the bash script so the script doesn't continue on error. 

Related to this issue is https://github.com/microsoft/vscode/issues/186500, the actual error occurs in `cat /proc/$PID/stat` but the script continues, resulting in another error `let: PROCESS_TIME_AFTER=+: syntax error: operand expected (error token is "+")`

### Code 

https://github.com/microsoft/vscode/blob/f5e6325900b7e831c2772d77a29cc24395d9f872/src/vs/base/node/cpuUsage.sh#L51

ajasad25:
Fixes #186502

### Problem

Process Explorer runs `cpuUsage.sh` to sample per-process CPU usage on Linux. For each PID the script checked `[ -f /proc/$PID/stat ]` and then read the file. A process can exit between that check and the `cat` (a TOCTOU race), leaving the stat empty, so the arithmetic

```sh
let PROCESS_TIME_AFTER="${PROCESS_STAT_ARRAY[13]}+${PROCESS_STAT_ARRAY[14]}"
```

ran with empty operands and failed:

```
cpuUsage.sh: line 51: let: PROCESS_TIME_AFTER=+: syntax error: operand expected (error token is "+")
```

Because `ps.ts` rejects whenever the script writes anything to `stderr`, this could intermittently break the CPU readings, not just log noise.

### Fix

- Read the stat unconditionally with `2>/dev/null`, so reading a file that just vanished cannot pollute `stderr`.
- Guard the arithmetic on the parsed values (`[ -n ... ]`) instead of the racy file-existence check, falling back to `0` (before-loop) / the previous sample (after-loop) when the process is gone.

This also keeps the array fresh on every iteration rather than relying on the existence check.

### Verification

- `bash -n src/vs/base/node/cpuUsage.sh` passes.
- Under bash, the old line reproduces the exact error on an empty stat; the guarded version returns `0` with **zero** `stderr`.
- A well-formed `/proc/<pid>/stat` line still parses to the correct user+system time.


ajasad25:

PR #319645, opened 2026-06-02
Agents: Focus chat input on new session

hawkticehurst:
Ensure chat input gets focus when new chat view is triggered via keybinding or new chat session button.


src/vs/sessions/browser/parts/sessionsPartService.ts
```diff
@@ -131,7 +133,8 @@ export class SessionsParts extends Disposable implements ISessionsPartService {
 	}
 
 	focusSession(session: IActiveSession | undefined): void {
-		this._mainPart.focusSession(session?.sessionId);
+		const sessionId = session?.sessionId;
```
benibenj: Using timers for this is bad practice. Can you clarify why this is needed? The actions which call focus method today seem to correctly place the focus into the input so I'm not sure why we would add this here? Is there a specific action which was not working correctly?

extensions/copilot/script/test/scoredEditsReconciler.spec.ts
```diff
@@ -106,14 +106,18 @@ suite('can resolve merge conflicts', () => {
...
                 [
-<<<<<<< HEAD
                     893,
                     893,
                     ";\\n\\t}\\n\\n\\tresetLastEditTime() {\\n\\t\\tthis._lastEditTime.set(undefined, undefined)"
                 ]
             ],
             "scoreCategory": "bad",
-=======
+            "score": 0
```
benibenj: This should not be part of the PR I think? Looks very unrelated

benibenj (dismissed):
Thanks a lot for making sure focus is added. I seem to have missed some actions.

extensions/copilot/script/test/scoredEditsReconciler.spec.ts
hawkticehurst: Nope it should not, I was trying to resolve the merge conflict for this PR and Copilot went a little off course for some reason. It'll be removed soon.

src/vs/sessions/browser/parts/sessionsPartService.ts
hawkticehurst: Oh yeah timer is not supposed to be used. You reviewed this PR a wee bit too quickly / I should've set this as a draft PR 😅

But hmm interesting, for me chat input never gets focus right now 🤔

src/vs/sessions/browser/parts/sessionsPart.ts
```diff
@@ -274,13 +278,24 @@ export class SessionsPart extends Part {
...
+	private _focusPendingSession(): void {
+		if (this._pendingFocusSessionId === NO_PENDING_FOCUS) {
+			return;
+		}
+
+		const slot = this._slots.find(s => s.boundSessionId === this._pendingFocusSessionId);
 		if (!slot) {
 			return;
 		}
+		this._pendingFocusSessionId = NO_PENDING_FOCUS;
```
benibenj: Let's drop this. i'm not yet sure there is a timing issue and if there is one i think we should solve it in another way

dmitrivMS (approved):

PR #319669, opened 2026-06-03
Resolve agent-host language models before model picker opens

galaxysj:
### Motivation
- The Agents Window new-session model picker could show only the already-registered agent-host (BYOK) models at startup because extension-contributed language model providers were not resolved early enough.  
- The same startup race affected the phone mode/model config sheet, producing inconsistent model lists on mobile.

### Description
- Eagerly resolve extension-contributed language model providers by calling `languageModelsService.selectLanguageModels({})` in `AgentHostModelPickerContribution` so model providers are populated while the new-session composer is visible.  
- Apply the same eager resolution in the phone picker by calling `this._languageModelsService.selectLanguageModels({})` in `MobileChatInputConfigPicker`.  
- Both calls are fire-and-forget and swallow errors with `.catch(() => undefined)` to avoid affecting startup flow.  
- Files modified: `src/vs/sessions/contrib/providers/agentHost/browser/agentHostModelPicker.ts` and `src/vs/sessions/contrib/providers/agentHost/browser/mobile/mobileChatInputConfigPicker.ts`.

### Testing
- Ran `npm run compile-check-ts-native` to type-check the changes and it completed successfully.  
- Attempted to run `./scripts/test.sh --grep AgentHostModelPicker` but the test run failed in this environment due to a missing environment dependency (`gulp-merge-json`), so the targeted test suite could not be executed here.

------
Edited by Codex


galaxysj:

PR #319678, opened 2026-06-03
Add openIntegratedBrowser option to serverReadyAction.action

### LINKED ISSUE #319676: Allow the integrated browser to open URLs set in launch.json serverReadyAction.
Currently serverReadyAction in launch.json will always open in an external browser even when `workbench.browser.openLocalhostLinks` is set. 

This happens because "extensions\debug-server-ready\src\extension.ts" calls `vscode.env.openExternal` which takes only the URI as a parameter, and so the internal call to extHostWindow.openUri` defaults to having `options.allowContributedOpeners` set to undefined, so the `LocalhostLinkOpenerContribution` is never executed.

I'm going to create a pull request for this, and I considered three options when doing so. 

1. Adding the parameter `options?: { allowContributedOpeners?: boolean | string }` to `vscode.env.openExternal`
2. Changing the default of `allowContributedOpeners` in `vscode.env.openExternal` to true.
3. Add a new `serverReadyAction.type` enum option `openIntegratedBrowser`.

I created branches for both, but ultimately went with the third option for a couple of reasons:
* I decided against (1) because `vscode.env.openExternal` is a public API it would need to have a clearer option than `allowContributedOpeners`. Adding a new option something along the lines of `allowIntegratedBrowser` might work, but felt like a messy solution.
*  I decided against (2) because extension developers might be relying on `vscode.env.openExternal` always being external, and changing it underneath them seems unfair and risky,
* I choose option (3) because even though it only fixes the issue for my narrow situation, it's entirely opt-in and therefor relatively low risk. This option means any URL, even non-localhost ones, can be opened in the internal browser.

neoGeneva:
Fixes #319676

Details of the decision making for this pull request are included in the issue.

Testing the changes can be done by creating a new node.js project with the following launch.json:

```json
{
    "configurations": [
        {
            "name": "Launch Program",
            "program": "${workspaceFolder}/index.js",
            "request": "launch",
            "skipFiles": [
                "<node_internals>/**"
            ],
            "type": "node",
            "serverReadyAction": {
                "action": "openIntegratedBrowser",
                "pattern": "listening on port ([0-9]+)",
                "uriFormat": "http://localhost:%s"
            }
        }
    ]
}
```

and app.js

```javascript
const express = require('express');
const app = express();
const port = 3000;

app.get('/', (req, res) => {
  res.send('Hello World!');
});

app.listen(port, () => {
  console.log(`Example app listening on port ${port}`);
});
```

### DIFF
--- a/extensions/debug-server-ready/package.json
+++ b/extensions/debug-server-ready/package.json
@@ -47,10 +47,12 @@
                       "action": {
                         "type": "string",
                         "enum": [
-                          "openExternally"
+                          "openExternally",
+                          "openIntegratedBrowser"
                         ],
                         "enumDescriptions": [
-                          "%debug.server.ready.action.openExternally.description%"
+                          "%debug.server.ready.action.openExternally.description%",
+                          "%debug.server.ready.action.openIntegratedBrowser.description%"
                         ],
                         "markdownDescription": "%debug.server.ready.action.description%",
                         "default": "openExternally"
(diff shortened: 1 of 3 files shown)


connor4312 (approved):
thanks!

aeschli (approved):

connor4312:
Merged in https://github.com/microsoft/vscode/pull/319775. Thank you for the contribution!

PR #319734, opened 2026-06-03
Replace animation frame queue sort with binary heap

Xio-Shark:
## Summary
- Replace the per-frame `Array.sort()` + `Array.shift()` pattern in the animation frame runner with a max-heap (`AnimationFrameQueue`), reducing queue operations from O(n log n) to O(log n) per push/pop.
- Add an insertion-order counter (`_order`) to `AnimationFrameQueueItem` so callbacks with the same priority execute in FIFO order — the previous sort was not stable.
- Make `_runner` and `_priority` `readonly` since they never change after construction.

## Motivation
Every animation frame, the runner called `currentQueue.sort()` on the entire queue and then `shift()` the first element. For workloads that schedule many animation frame callbacks (e.g. editor rendering, cursor blinking, overlays), this is wasteful: a heap provides O(log n) insertion and extraction while naturally maintaining priority order.

## Changes
- **`src/vs/base/browser/dom.ts`**: New `AnimationFrameQueue` class backed by a binary max-heap with `runsBefore()` comparison (priority desc, then insertion order asc). `NEXT_QUEUE` and `CURRENT_QUEUE` now hold `AnimationFrameQueue` instances instead of plain arrays.
- **`src/vs/base/test/browser/dom.test.ts`**: Four new tests covering priority ordering, FIFO stability with 128 items, cancellation, and mid-frame scheduling.

## Verification
- [x] `tsgo --project ./src/tsconfig.json --noEmit --skipLibCheck` — passed
- [x] `node build/eslint.ts` — passed
- [x] `scripts/test.sh --grep "AnimationFrameQueue"` — 4/4 passing
- [x] Full `dom` test suite — 41/41 passing, no regressions
- [x] Pre-commit hygiene hook — passed

## Test plan
- [ ] CI compile & hygiene checks pass
- [ ] All unit tests pass on Linux / macOS / Windows (Electron + Browser)
- [ ] No regressions in animation frame scheduling behavior

### DIFF
--- a/src/vs/base/browser/dom.ts
+++ b/src/vs/base/browser/dom.ts
@@ -395,13 +395,15 @@ export class WindowIntervalTimer extends IntervalTimer {
 
 class AnimationFrameQueueItem implements IDisposable {
 
-	private _runner: () => void;
-	public priority: number;
+	private readonly _runner: () => void;
+	private readonly _priority: number;
+	private _order: number;
 	private _canceled: boolean;
 
 	constructor(runner: () => void, priority: number = 0) {
 		this._runner = runner;
-		this.priority = priority;
+		this._priority = priority;
+		this._order = 0;
 		this._canceled = false;
 	}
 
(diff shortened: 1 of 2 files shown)


Xio-Shark:

PR #319747, opened 2026-06-03
Git: show hint when Git operations may be waiting for SSH authentication

Iamliuxiaozhen:
This implementation shows a notification when Git operations such as
push, pull, or fetch take longer than expected and may be waiting for
SSH authentication.

This is particularly helpful for users who authenticate using SSH
security keys (FIDO/U2F/WebAuthn), where user interaction may be
required but is not clearly surfaced in the VS Code UI.


Iamliuxiaozhen:

Iamliuxiaozhen:
I've addressed the review comments and updated the implementation.
Could someone take another look when convenient? Thanks!

Iamliuxiaozhen:
Friendly ping in case this got buried. Happy to make any further changes if needed.

PR #319780, opened 2026-06-03
downscale/optimize image attachments

### LINKED ISSUE #319426: Image attachment thumbnails should be downscaled images, not full res ones
Thumbnails for image attachments in chat currently use full size images instead of downscaled versions. This causes performance issues when attaching very large images (or several large images) where even typing in the chat is slow after every keystroke.

justschen:
fix https://github.com/microsoft/vscode/issues/319426

cc @federicobrancasi

### DIFF
--- a/src/vs/workbench/contrib/chat/browser/attachments/chatAttachmentWidgets.ts
+++ b/src/vs/workbench/contrib/chat/browser/attachments/chatAttachmentWidgets.ts
@@ -661,7 +661,7 @@ function createImageElements(resource: URI | undefined, name: string, fullName: 
 			style: HoverStyle.Pointer,
 		}));
 
-		const blob = new Blob([buffer as Uint8Array<ArrayBuffer>], { type: 'image/png' });
+		const blob = new Blob([buffer as Uint8Array<ArrayBuffer>]);
 
 		// Keep the full-resolution image available as a fallback for the hover
 		// preview, but only decode it lazily if the downscaled thumbnail fails.
@@ -689,7 +689,7 @@ function createImageElements(resource: URI | undefined, name: string, fullName: 
 		// only once, and the UI keeps a small object URL for steady-state rendering.
 		const previewImageUrl = disposable.add(new MutableDisposable<IDisposable>());
 		const renderPreviewImage = async () => {
-			const thumbnail = await createImageThumbnail(data, 'image/png', IMAGE_HOVER_THUMBNAIL_MAX_SIZE);
+			const thumbnail = await createImageThumbnail(data, undefined, IMAGE_HOVER_THUMBNAIL_MAX_SIZE);
 			if (disposable.isDisposed) {
 				return;
 			}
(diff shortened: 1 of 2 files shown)


src/vs/workbench/contrib/chat/browser/chatImageUtils.ts
```diff
@@ -83,6 +83,48 @@ export async function resizeImage(data: Uint8Array | string, mimeType?: string):
...
+/**
+ * Creates a small downscaled thumbnail of an image. Useful for compact previews
+ * (e.g. attachment pills) where the UI should retain a small rendered image
+ * instead of a full-resolution object URL.
+ * @param data The image bytes.
+ * @param mimeType The image mime type.
+ * @param maxSize The maximum width or height of the thumbnail, in pixels.
+ * @returns A promise that resolves to a PNG {@link Blob} of the thumbnail, or `undefined` on failure.
+ */
+export function createImageThumbnail(data: Uint8Array, mimeType: string | undefined, maxSize: number): Promise<Blob | undefined> {
```
dmitrivMS, line 95: this appears unsed

src/vs/workbench/contrib/chat/browser/chatImageUtils.ts
```diff
@@ -83,6 +83,48 @@ export async function resizeImage(data: Uint8Array | string, mimeType?: string):
...
+			canvas.width = targetWidth;
+			canvas.height = targetHeight;
+			const ctx = canvas.getContext('2d');
+			if (!ctx) {
+				resolve(undefined);
+				return;
+			}
+
+			ctx.drawImage(img, 0, 0, targetWidth, targetHeight);
+			canvas.toBlob(thumbnail => resolve(thumbnail ?? undefined), 'image/png');
```
dmitrivMS, line 119: It seems that it always creates a PNG, but re-encoding a JPEG attachment at 768px PNG can be larger than the original, partially defeating the memory win. Does this really help?

src/vs/workbench/contrib/chat/browser/chatImageUtils.ts
dmitrivMS, line 95: This is allocating DOM elements on main window? If so, perhaps it should be using aux window instead?

src/vs/workbench/contrib/chat/browser/attachments/chatAttachmentWidgets.ts
```diff
@@ -667,17 +676,36 @@ function createImageElements(resource: URI | undefined, name: string, fullName:
...
+		};
+
+		const data = buffer instanceof Uint8Array ? buffer : new Uint8Array(buffer);
+
+		// Both the always-visible pill and the hover preview render small. Use one
+		// generated thumbnail for both so the original bytes are decoded and resized
+		// only once, and the UI keeps a small object URL for steady-state rendering.
+		const previewImageUrl = disposable.add(new MutableDisposable<IDisposable>());
+		const renderPreviewImage = async () => {
+			const thumbnail = await createImageThumbnail(data, undefined, IMAGE_HOVER_THUMBNAIL_MAX_SIZE);
```
dmitrivMS, line 692: After the thumbnail resolves, the original Uint8Array is still captured by the closure purely as a hover-image fallback. Suggestion: null out the buffer?

dmitrivMS (approved):
Added some comments (based on AI review)

PR #319806, opened 2026-06-03
debug: Log errors when loading stored breakpoints/watch expressions (#319805)

### LINKED ISSUE #319805: Debug: Silent data loss when stored breakpoints/watch expressions are corrupted
### Problem

In `src/vs/workbench/contrib/debug/common/debugStorage.ts`, five methods load debug data from storage using `JSON.parse()`, but all five use **empty catch blocks** — silently swallowing any parsing or construction errors:

- `loadBreakpoints()` — line 83 — Stores user breakpoints
- `loadFunctionBreakpoints()` — line 94 — Stores function breakpoints
- `loadExceptionBreakpoints()` — line 105 — Stores exception breakpoints
- `loadDataBreakpoints()` — line 116 — Stores data breakpoints
- `loadWatchExpressions()` — line 127 — Stores watch expressions

If stored data gets corrupted (sync conflicts, disk errors, crash mid-write, or a bug in serialization), **every single user breakpoint, function breakpoint, exception breakpoint, data breakpoint, and watch expression is silently dropped** with zero indication to the user or developer. The data simply vanishes.

This class already has `ILogService` injected via the constructor (`this.logService`), but it is unused in the catch blocks.

### Why this matters

Users may lose their entire debugging setup without any notification. This is especially painful for complex debug configurations with many breakpoints. The user has no way to know that their data was corrupted — they'll just see all breakpoints gone and assume they were never set.

### Fix

Replace empty catch blocks with `this.logService.error()` calls that record the failure. This ensures:

1. The error is visible in the developer console and logs
2. Users/developers can diagnose why breakpoints disappeared
3. The behavior is unchanged (returns empty array on failure)
4. The fix is minimal, low-risk, and uses existing infrastructure

### File

`src/vs/workbench/contrib/debug/common/debugStorage.ts`

ishaq2321:
## Summary

Five methods in `DebugStorage` silently swallowed JSON parse and object construction errors with **empty catch blocks**. If stored debug data gets corrupted, users lose **all** their breakpoints, function breakpoints, exception breakpoints, data breakpoints, **and** watch expressions — with zero indication of failure.

## Problem

In `src/vs/workbench/contrib/debug/common/debugStorage.ts`, these five methods had empty `catch (e) { }` blocks:

| Method | Affected Data |
|---|---|
| `loadBreakpoints()` | All user breakpoints |
| `loadFunctionBreakpoints()` | All function breakpoints |
| `loadExceptionBreakpoints()` | All exception breakpoints |
| `loadDataBreakpoints()` | All data breakpoints |
| `loadWatchExpressions()` | All watch expressions |

**Corruption scenario:** If IndexedDB/localStorage data becomes corrupted due to sync conflicts, disk errors, crash mid-write, or a serialization bug, `JSON.parse()` throws — the empty catch drops the exception, the method returns `[]`, and the user is left wondering why all their carefully-set breakpoints just disappeared.

## Fix

Replace each empty `catch (e) { }` with `this.logService.error('message', e)`. The class already has `ILogService` injected in its constructor but was not using it in these catch blocks.

```diff
- } catch (e) { }
+ } catch (e) {
+     this.logService.error('Failed to load breakpoints from storage', e);
+ }
```

### Why this is the right fix

1. **Surfaces corruption** — Errors are logged to the developer console and log service, enabling diagnosis
2. **No behavior change** — Still returns empty arrays on parse failure (graceful degradation)
3. **Minimal and low-risk** — `logService` is already injected via DI, no new dependencies
4. **Follows existing patterns** — The `loadBreakpoints` method already passes `this.logService` to `new Breakpoint()`, so the infrastructure is already wired

## Testing

- Existing behavior preserved — errors still caught, empty arrays returned
- Log output appears in developer console when data is corrupted
- No functional change to debug workflows

Closes #319805

Co-authored-by: Muhammad Ishaq Khan <muhammadishaqkhan.2321@gmail.com>

### DIFF
--- a/src/vs/workbench/contrib/debug/common/debugStorage.ts
+++ b/src/vs/workbench/contrib/debug/common/debugStorage.ts
@@ -80,7 +80,9 @@ export class DebugStorage extends Disposable {
 				breakpoint.uri = URI.revive(breakpoint.uri);
 				return new Breakpoint(breakpoint, this.textFileService, this.uriIdentityService, this.logService, breakpoint.id);
 			});
-		} catch (e) { }
+		} catch (e) {
+			this.logService.error('Failed to load breakpoints from storage', e);
+		}
 
 		return result || [];
 	}
@@ -91,7 +93,9 @@ export class DebugStorage extends Disposable {
 			result = JSON.parse(this.storageService.get(DEBUG_FUNCTION_BREAKPOINTS_KEY, StorageScope.WORKSPACE, '[]')).map((fb: ReturnType<FunctionBreakpoint['toJSON']>) => {
 				return new FunctionBreakpoint(fb, fb.id);
 			});
-		} catch (e) { }
+		} catch (e) {
+			this.logService.error('Failed to load function breakpoints from storage', e);
+		}
 
 		return result || [];
 	}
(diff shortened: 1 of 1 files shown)


connor4312 (approved):

amunger (approved):

Mosesalfredjohns (commented):
Debug

Issue #319593, opened 2026-06-02
Security: Webviews can trigger arbitrary keyboard shortcuts in the main workbench


ammaraskar:
The default set of webview message handlers have an event called `did-keydown`:

https://github.com/microsoft/vscode/blob/e4074382086b61bdfb0b6738e0853b2510660d57/src/vs/workbench/contrib/webview/browser/webviewElement.ts#L249-L254

This is used to bring out keydown events from the webview frame so keybindings work when you're focused on the webview:

https://github.com/microsoft/vscode/blob/f9994508a78fc1fcdbc598ea53c9e5a541bfdc7a/src/vs/workbench/contrib/webview/browser/pre/index.html#L904-L905

Unfortunately this can create a bit of a security issue. If there is XSS in a webview or it intentionally allows scripts, it can trigger a series of dangerous keydown events like:

**"Terminal: New Terminal Window"** <kbd>Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>Alt</kbd>+<kbd>`</kbd> followed by changing focus <kbd>F6</kbd> and then **"Terminal: Paste Into Active Terminal"** <kbd>Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>V</kbd>.

I don't know what a good solution is in terms of balancing this issue with usability because it wouldn't be the best user-experience to just have shortcuts not work entirely when you're clicked on a webview.

jeffmcjunkin:
Related: https://blog.ammaraskar.com/github-token-stealing/

coworkcoblah:
This is unethical behavior on the part of the kiddo that reported this. I hope they have more talent than the report makes apparent.

Whatever issue you have with Microsoft does not justify using every Github user as a weapon to pressure Microsoft into changing.

Good luck with future employment.

alexdima:
👋 Thank you for reporting, we are working on a fix.

To clarify, this issue does not affect VS Code Desktop.

ammaraskar:
Thanks Alexandru! That fixes this particular notebook entrypoint but it would make sense to keep an issue around for the whole keyboard shortcuts in webviews thing as it can affect other extensions (built-in or otherwise).

acdha:
It looks like the window for attacks has been closed but is there a reliable audit log query to identify organization users who might have been exposed?

coworkcoblah:
> It looks like the window for attacks has been closed but is there a reliable audit log query to identify organization users who might have been exposed?

[...]

PR #319813, opened 2026-06-03
Block programmatic webview click and keypress re-dispatch events in some cases

mjbvz:
For #319593

This should disable scripts re-dispatching entirely in notebook webviews as I don't think there's a valid use case there

For now I'm keeping it enabled for extension webviews as this re-dispatch is the only way I know for a full screen iframe in webviews to not eat all keypresses. If needed we can add something on the webview vscode api object that should allow this to be even safer

### DIFF
--- a/src/vs/workbench/api/browser/mainThreadWebviews.ts
+++ b/src/vs/workbench/api/browser/mainThreadWebviews.ts
@@ -152,6 +152,7 @@ export function reviveWebviewContentOptions(webviewOptions: extHostProtocol.IWeb
 	return {
 		allowScripts: webviewOptions.enableScripts,
 		allowForms: webviewOptions.enableForms,
+		forwardUntrustedKeypressEvents: true, // This is always enabled for extensions
 		enableCommandUris: webviewOptions.enableCommandUris,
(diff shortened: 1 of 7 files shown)


dmitrivMS (approved):

ammaraskar:
Hmm, I would caution against trusting anything coming from the webview hostMessaging. Because of the whole same-origin with the content frame, there's a lot of shenanigans that can be done on the javascript side. I haven't fully gotten it to a working state but something like:

```js
const origPostMessage = window.top.frames[1].MessagePort.prototype.postMessage;
const fakePostMessage = (arg) => {
  console.log(arg);
  // Modify the isTrusted here
  origPostMessage(arg);
}
window.top.frames[1].MessagePort.prototype.postMessage = fakePostMessage;
```

I know these lines kinda try to protect against it https://github.com/microsoft/vscode/blob/9c5387b348698fdb9129c7e24d1572c6eb81d7d9/src/vs/workbench/contrib/webview/browser/pre/index.html#L239-L241 but some of those properties are read-only in some browsers (not sure about electron) so they remain unaffected by those assignments.

PR #319878, opened 2026-06-04
Support opening Agents sessions in new windows

### LINKED ISSUE #319877: Support opening Agents sessions in separate windows
It should be possible to open an Agents session in a separate Agents window, not only side-by-side within the current Agents window.

This enables users to keep multiple sessions visible as independent full Agents windows, each with its own sessions list and layout.

Expected:
- Right-clicking a session supports opening it in a new Agents window.
- Dragging a session outside the Agents window opens it in a new Agents window.

federicobrancasi:
Fixes #319877

Adds support for opening Agents sessions in separate Agents windows.

Changes:
- Adds an "Open in New Window" session context menu action.
- Supports dragging a session row outside the Agents window to open it in a new Agents window.

Demo:

https://github.com/user-attachments/assets/f4361d6a-82b0-43cf-818f-5758036d4ec9


sandy081:
We're not going to support multiple Agents windows — the Agents window is meant to be a single focused experience. Closing this. Thanks anyway!

PR #320003, opened 2026-06-04
sessions: make session navigation work when editor part is open but unfocused

sandy081:
## What

The session navigation keybindings — **Go Back** (`Ctrl+Shift+Tab`) and **Go Forward** (`Ctrl+Tab`) — in the Agents window were gated on `EditorContextKeys.editorTextFocus.toNegated()`. This switches the gate to `EditorAreaFocusContext.toNegated()`.

## Why

`editorTextFocus` evaluates such that the navigation chords were suppressed even when the editor part is open but does **not** have keyboard focus. As a result, `Ctrl+Shift+Tab` / `Ctrl+Tab` did not navigate between sessions whenever an editor was open.

`EditorAreaFocusContext` is `true` only when keyboard focus is genuinely within the editor area (any editor part). Negating it means session navigation works whenever focus is outside the editor area — including when an editor is open but unfocused — while still deferring to the editor's own `Ctrl+Tab` quick-open when the editor area is actually focused.

## Notes for reviewers

- Only the `when` clauses of the two navigation actions change; weights and key chords are untouched.
- `EditorAreaFocusContext` is bound in `editorPart.ts` for all editor parts (main, modal, auxiliary).


sandy081:
Recreated with correct branch prefix in #320005.

PR #320036, opened 2026-06-05
Fix model selection issues when switching session types

dmitrivMS:


DonJayamanne:
Thanks for looking into this.
I think I did fix one of the main issues where model was reverting to auto or another. And dinner other issues are already fixed.

I'm a little concerned this could cause other weird issues.
Please can we wait till Monday, that way we have a whole week to test this in insiders, also can verify whether my fix , fixes the Model switching issue for existing sessions in agents window

dmitrivMS:
@DonJayamanne Yep, no problem. This is for the main window, not Agents - it fixes at least 2 or 3 problems with models changing which I have stable repros (using 'Copilot CLI [Local]' but also between other session types.

dmitrivMS:
@DonJayamanne Here are some repro scenarios which are still not fixed with current main:

1) Pick Local, Pick Opus 4.7, run a prompt ('hi'), go back to make new chat, switch to 'Copilot CLI [Agent Host]' but don't touch the model picker, run prompt -> Error: (sendFailed) Error: Request session.create failed with message: Model "copilot/claude-opus-4.7" is not available.

2) Pick 'Copilot CLI [Agent Host]', pick Opus 4.7, run prompt, edit prompt in chat, pick Sonnet 4.5, rerun prompt -> prompt still runs with Opus 4.7

dmitrivMS:
With this change first case will be fixed, second not.

src/vs/workbench/contrib/chat/browser/agentSessions/agentHost/agentHostSessionHandler.ts
```diff
@@ -2694,6 +2696,10 @@ export class AgentHostSessionHandler extends Disposable implements IChatSessionC
 		if (languageModelIdentifier.startsWith(prefix)) {
 			return languageModelIdentifier.substring(prefix.length);
 		}
+		if (languageModelIdentifier.includes('/')) {
```
dmitrivMS, line 2699: This change is defensive as it will revert to Auto if we ever reach with invalid identifier here.

DonJayamanne (approved):

Blevent:
I have pro copilot but just use auto select model and it gives nonsense answers. Why can not select and use other models?
There is many message here but I could not see the answer.

PR #320046, opened 2026-06-05
extensions: make auto-update minimum release age configurable

hand-dot:
## What

This makes the extension auto-update bake time configurable through a new
`extensions.autoUpdateMinimumReleaseAge` setting.

The setting is expressed in minutes and defaults to `120`, preserving the current
2-hour behavior for automatic extension updates.

Examples:

```json
{
  "extensions.autoUpdateMinimumReleaseAge": 120
}
```

- `0`: do not delay automatic extension updates
- `120`: current 2-hour default
- `10080`: one week

## Why

VS Code now delays automatic updates for newly published extension versions,
which gives new versions a short bake time before they are installed
automatically.

Some users and organizations want to choose a longer window based on their
security posture. This change keeps the existing default behavior, but allows
that window to be configured.

Part of #316867.

## Scope

This PR intentionally applies only to automatic extension updates.

Unaffected:

- Manual extension updates
- Manual extension installs
- Marketplace search
- CLI installs
- Devcontainer extension installs
- Outdated extension detection

The existing trusted publisher bypass is preserved.

## Notes

The setting uses `extensions.autoUpdateMinimumReleaseAge` rather than nesting
under `extensions.autoUpdate` because `extensions.autoUpdate` is already a
boolean/string setting. Keeping this as a flat setting avoids changing the shape
or migration behavior of `extensions.autoUpdate`.


hand-dot:
@microsoft-github-policy-service agree company="pdfme"

hand-dot:
@microsoft-github-policy-service agree company="pdfme"

hand-dot:
I chose minutes for this setting because the existing VS Code behavior is a fixed 2-hour delay, so `120` maps directly to the current default. Other minimum-release-age implementations use different units, for example npm uses days while pnpm uses minutes, so there does not seem to be a single ecosystem-wide standard here.

Using minutes keeps the setting precise enough for the current behavior while still allowing longer windows such as `1440` for one day or `10080` for one week.

src/vs/workbench/contrib/extensions/common/extensions.ts
```diff
@@ -188,6 +188,8 @@ export const enum ExtensionEditorTab {
 export const ConfigurationKey = 'extensions';
 export const AutoUpdateConfigurationKey = 'extensions.autoUpdate';
 export const AutoCheckUpdatesConfigurationKey = 'extensions.autoCheckUpdates';
+export const AutoUpdateMinimumReleaseAgeConfigurationKey = 'extensions.autoUpdateMinimumReleaseAge';
```
karlhorky, line 191: @hand-dot this is missing important features from #316867 such as the exclude mechanism 

Exclude mechanism: exclude extensions by:

1. Name
2. Publisher
3. Version number

Similar to pnpm `minimumReleaseAgeExclude`

hand-dot: Thanks, I agree this would be useful and is part of the broader #316867 request.

I scoped this PR to making the existing auto-update cooldown configurable first, without changing the rest of the update behavior. Excludes by name, publisher, or version would add a separate configuration surface and matching logic, so I think it may be better as a follow-up unless maintainers prefer it in this PR.

karlhorky: I think the implementation is incomplete without this feature.

Shipping this to users without this functionality will lead to confusion, configuration churn and more GitHub issues being reported.

pnpm found this out very fast (and then shipped the exclude feature soon after `minimumReleaseAge`) and future implementations like Bun included the exclude feature in the first PR itself:

- https://github.com/oven-sh/bun/pull/22801

hand-dot: One difference from package-manager implementations is that this PR only affects automatic extension updates. Manual updates remain available immediately, so users can still bypass the cooldown for a specific extension by clicking Update.

That said, I agree an explicit exclude setting could still be useful for organizations that want certain extensions to continue auto-updating immediately. I’d be happy to add a minimal exclude mechanism if maintainers think it should be part of this PR.

sandy081:
Thank you for the contribution, @hand-dot! We're actively working on this feature ourselves and want to ensure it aligns with our broader extension auto-update cooldown design. I'm going to close this PR for now, but we really appreciate the effort. Stay tuned for the upcoming implementation.

karlhorky:
The implementation from the VS Code team was merged and will be a part of VS Code 1.125:

> @sandy081 in [comment #4691382342](https://github.com/microsoft/vscode/issues/316867#issuecomment-4691382342): This is now largely addressed by [#321039](https://github.com/microsoft/vscode/pull/321039) (merged) 🎉
> 
> VS Code now supports a configurable **`extensions.autoUpdateDelay`** setting (in hours, default 2) that controls how long to wait after an extension update is published before auto-installing it. Both `extensions.autoUpdate` and `extensions.autoUpdateDelay` are also **policy-enabled** for enterprise/MDM management.
> 
> One part of this request is still open: the delay currently applies only to **auto-updates**, not to fresh **installs** (marketplace UI, `code --install-extension`, devcontainers). We're tracking that separately in [#321136](https://github.com/microsoft/vscode/issues/321136).
> 
> Thanks everyone for the strong community support here.

Source: https://github.com/microsoft/vscode/issues/316867#issuecomment-4691382342

PR #320053, opened 2026-06-05
🐛 Update GetErrorsTool to handle Windows drive letter casing

### LINKED ISSUE #319858: Agent tool `read/problems` on Windows is sensitive to drive letter case. Not stable enough.
### Problem
On Windows, when Copilot agent uses the `read/problems` tool, the file path generated by the agent uses a lowercase drive letter (e.g., `c:\`). As a result, the tool fails to find problems for the file. If I change the drive letter to uppercase (e.g., `C:\`), the tool works as expected and finds problems.

### Expected behavior
The file object should be validated and normalized (e.g., standardize to uppercase drive letter or otherwise) within the tools layer. The agent should not have to care about drive letter case. Ideally, the agent's output would not need to adjust for this platform-specific detail.

### Steps to reproduce
1. On Windows, run an agent query that uses `read/problems` with a path like `c:\path\to\file`
2. The tool reports "no problems found"
3. Change the path to `C:\path\to\file` and rerun – now problems are correctly found

### Suggestion
- Tools layer should handle drive letter normalization and validation, so that agents are not required to output platform-specific casing.
- Agent should not need to care about drive letter casing when generating tool input.

### Additional info
- VS Code version: 1.123.0 (latest up to now)
- OS: Windows 11 Pro
- Copilot/Agent version: latest

**Labels:** `bug`, `agent-skills`, `agents`, `agent-behavior`

Wangmz-1211:
Fix #319858
This pull request updates the `GetErrorsTool` to improve handling of file path comparisons on Windows, specifically addressing differences in drive letter casing. The main changes ensure that file and folder path matching is case-insensitive where appropriate, and new tests are added to verify this behavior.

**Path Comparison Improvements:**

* Updated `getErrorsTool.tsx` to use `extUriBiasedIgnorePathCase.isEqualOrParent` and `isEqual` for path comparisons, ensuring case-insensitive matching of file and folder paths, which is especially important on Windows systems.
**Testing Enhancements:**

* Added Windows-specific test cases in `getErrorsTool.spec.tsx` to verify that diagnostics are correctly matched even when the drive letter casing differs (e.g., `C:` vs `c:`). These tests are conditionally run only on Windows.

These changes improve cross-platform reliability and correctness of diagnostic retrieval in the Copilot extension.


Wangmz-1211:

roblourens:
I disagree that we should _always_ ignore the path casing

Wangmz-1211:
> I disagree that we should _always_ ignore the path casing


A new function `normalizeDriveLetter` was added to `GetErrorsTool` class and public for testing purposes only, which is same as `getDiagnostics`. Test suite was added to `getErrorsTool.spec.tsx`.

The new function works for Windows only, and change only the drive letter to lowercase. This avoids always ignoring path casing.

roblourens:
We have other helpers for this and shouldn't reinvent path comparison in this one class

Wangmz-1211:
> We have other helpers for this and shouldn't reinvent path comparison in this one class

The code now checks exact matches with the existing unbiased `isEqual` helper first, which handles Windows drive-letter casing through URI formatting. It then falls back to `isEqualOrParent` for folder inputs, preserving the existing path-casing semantics for parent checks.

PR #320090, opened 2026-06-05
Fix forwarded ports status bar item to toggle the ports view

### LINKED ISSUE #320086: Make status bar `Forwarded Ports` toggleable
Clicking `Forwarded Ports` once opens the PORTS panel. But clicking `Forwarded Ports` again does not close it.

It would be nice to add an action to be able to toggle it.

thernstig:
Previously, clicking the forwarded ports status bar entry only focused the view. Replace the built-in `~remote.forwardedPorts.focus` command with a proper toggle action that closes the view when it is already visible, matching the behavior of the Problems status bar item.

Also extracts constants into common/remoteExplorer.ts and browser/messages.ts following the same pattern as the Markers contribution.

Fixes #320086


src/vs/workbench/contrib/remote/browser/remote.contribution.ts
```diff
@@ -12,6 +13,29 @@ import { RemoteAgentConnectionStatusListener, RemoteMarkers } from './remote.js'
...
+	}
+
+	async run(accessor: ServicesAccessor): Promise<void> {
+		const viewsService = accessor.get(IViewsService);
+		if (viewsService.isViewVisible(TUNNEL_VIEW_ID)) {
+			viewsService.closeView(TUNNEL_VIEW_ID);
+		} else {
+			await viewsService.openView(TUNNEL_VIEW_ID, true);
+		}
+	}
```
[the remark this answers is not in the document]
thernstig: closeView() is not defined as returning a promise, see here:

```ts
export interface IViewsService {

	readonly _serviceBrand: undefined;

	// View Container APIs
	readonly onDidChangeViewContainerVisibility: Event<{ id: string; visible: boolean; location: ViewContainerLocation }>;
	isViewContainerVisible(id: string): boolean;
	isViewContainerActive(id: string): boolean;
	openViewContainer(id: string, focus?: boolean): Promise<IPaneComposite | null>;
	closeViewContainer(id: string): void;
	getVisibleViewContainer(location: ViewContainerLocation): ViewContainer | null;
	getActiveViewPaneContainerWithId(viewContainerId: string): IViewPaneContainer | null;
	getFocusedView(): IViewDescriptor | null;
	getFocusedViewName(): string;

	// View APIs
	readonly onDidChangeViewVisibility: Event<{ id: string; visible: boolean }>;
	readonly onDidChangeFocusedView: Event<void>;
	isViewVisible(id: string): boolean;
	openView<T extends IView>(id: string, focus?: boolean): Promise<T | null>;
	closeView(id: string): void;
	getActiveViewWithId<T extends IView>(id: string): T | null;
	getViewWithId<T extends IView>(id: string): T | null;
	getViewProgressIndicator(id: string): IProgressIndicator | undefined;
}
```

src/vs/workbench/contrib/remote/browser/messages.ts
```diff
@@ -0,0 +1,10 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import * as nls from '../../../../nls.js';
+
+export class Messages {
+	public static FORWARDED_PORTS_TOGGLE_LABEL: string = nls.localize('remote.togglePortsView', "Toggle Ports View");
+}
```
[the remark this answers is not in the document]
thernstig: I followed the same exact pattern as in the src/vs/workbench/contrib/markers/browser/messages.ts for the Problems pane.

src/vs/workbench/contrib/remote/browser/messages.ts
```diff
@@ -0,0 +1,10 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import * as nls from '../../../../nls.js';
+
+export class Messages {
+	public static FORWARDED_PORTS_TOGGLE_LABEL: string = nls.localize('remote.togglePortsView', "Toggle Ports View");
```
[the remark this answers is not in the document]
thernstig: Done

src/vs/workbench/contrib/remote/browser/messages.ts
```diff
@@ -0,0 +1,13 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import * as nls from '../../../../nls.js';
+
+export class Messages {
+	public static FORWARDED_PORTS_TOGGLE_LABEL: string = nls.localize(
```
alexr00: No need for this to be in a new class in a new file.

src/vs/workbench/contrib/remote/common/remoteExplorer.ts
```diff
@@ -0,0 +1,8 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+export namespace ForwardedPorts {
+	export const TOGGLE_VIEW_ACTION_ID = 'remoteExplorer.toggleForwardedPortsView';
```
alexr00: Same for this.

alexr00 (changes requested):
Thanks for the PR. Just a few comments.

src/vs/workbench/contrib/remote/browser/messages.ts
thernstig: @alexr00 I followed the same style as for the problems pane, to keep it coherent. Still want me to remove that extra file?

src/vs/workbench/contrib/remote/common/remoteExplorer.ts
thernstig: @alexr00 I followed the same style as for the problems pane, to keep it coherent. Still want me to remove that extra file?

src/vs/workbench/contrib/remote/browser/messages.ts
alexr00: Yes, especially if this string is only used in one place.

src/vs/workbench/contrib/remote/common/remoteExplorer.ts
alexr00: Yes, we have other files that have ports related constants.

thernstig:
@alexr00 ?

alexr00 (changes requested):
While testing this out, I noticed we already have an automatically created command "View: Toggle Ports" (`~remote.forwardedPortsContainer`). Can we just re-used that command ID and not have a new command at all?

thernstig:
@alexr00 are you sure that one is the same in behavior if I use it?

I checked around and _believe_ it works differently. I had an AI summarize the thoughts:

The built-in command behind "View: Toggle Ports" is not a pure toggle. It is a panel-container command, and for panel containers its behavior is:

- If Ports is not the visible panel, open Ports and focus it.
- If Ports is visible but the panel itself does not currently have focus, keep it open and move focus into the panel.
- Only if Ports is visible and the panel already has focus, close the panel.

My behavior was:

- If Ports is visible, close it.
- If Ports is not visible, open it.
- Focus state does not matter.

So I think to correlate the behavior with the problems pane, the solution we have now is equal in functionality.

alexr00 (approved):
That's fair. Thanks for the PR!

connor4312 (approved):

PR #320103, opened 2026-06-05
Fix issue reporter wizard dropping preset extension `data`

### LINKED ISSUE #320101: Issue reporter wizard drops preset extension "data" when opened with extensionId
## Issue

When the issue reporter is opened programmatically via the `workbench.action.openIssueReporter` command with both an `extensionId` and `data` (the optional extension-data string), the `data` is silently dropped in the new wizard-based issue reporter (`IssueReporterOverlay`).

The old reporter (`baseIssueReporterService.updateExtensionStatus`) propagated `data.data` into the model's `extensionData` automatically when the reporter was opened with a preset extension; the new wizard does not, so attached context never makes it into the final issue body.

### Repro

Run from any extension:

```ts
vscode.commands.executeCommand('workbench.action.openIssueReporter', {
    extensionId: 'GitHub.copilot-chat',
    issueTitle: '',
    issueBody: '# Description\n...',
    data: '<details><summary>STest</summary>...attached context...</details>',
});
```

Expected: the `data` payload appears as extension data in the rendered issue body.
Actual: `data` is absent from the issue body. Only `issueTitle` / `issueBody` survive.

### Concrete impact

Copilot's NES "Report feedback" command (`InlineEditDebugComponent` in vscode-copilot) attaches the STest, recording, and UI repro via this `data` field. With the new reporter, all of that context is lost from the issue body.

### Root cause

In `src/vs/workbench/contrib/issue/browser/issueReporterOverlay.ts`, `createWizard()` calls:

```ts
if (this.data.extensionId) {
    void this.updateSelectedExtension(this.data.extensionId, /* loadExtensionData */ false);
}
```

With `loadExtensionData=false`, `applyExtensionIssueData` is never invoked, so the preset `data.data` / `data.uri` / `data.privateUri` are never merged into the selected extension or into the model (`extensionData` + `includeExtensionData: true`).

cc @Giuspepe

ulugbekna:
Fixes #320101

The new `IssueReporterOverlay` wizard dropped the `data` payload provided via the `workbench.action.openIssueReporter` command when an `extensionId` was also preset. This broke callers like Copilot's NES feedback flow, which attaches an STest and recording context via the `data`  that context was silently lost from the resulting issue body.field 

### Root cause

The constructor calls `updateSelectedExtension(extensionId, /* loadExtensionData */ false)`. With `loadExtensionData=false`, the existing `applyExtensionIssueData` path that propagates `data` onto the selected extension and into the model (`extensionData` + `includeExtensionData`) was never invoked. The old reporter (`baseIssueReporterService.updateExtensionStatus`) did this propagation; the new wizard didn't.

### Fix

In `updateSelectedExtension`, when `loadExtensionData=false` but the caller-supplied `this.data` carries `data` / `uri` / `privateUri`, apply it via the same helper. Guarded by `!extension.data` so it doesn't re-apply when `updateExtensionOptions` re-invokes `updateSelectedExtension` after `updateModel` repopulates `allExtensions` (covers the case where the extension list arrives asynchronously after the wizard is already built).

User-driven selections from the dropdown / source buttons are  they call `updateSelectedExtension` with the default `loadExtensionData=true`, taking the existing `if` branch.unaffected 

cc @Giuspepe

### DIFF
--- a/extensions/copilot/src/extension/inlineEdits/vscode-node/components/inlineEditDebugComponent.ts
+++ b/extensions/copilot/src/extension/inlineEdits/vscode-node/components/inlineEditDebugComponent.ts
@@ -91,7 +91,7 @@ export class InlineEditDebugComponent extends Disposable {
 			await openIssueReporter({
 				title: '',
 				data: data.toString(),
-				issueBody: '# Description\nPlease describe the expected outcome and attach a screenshot!',
+				issueBody: 'Please describe the expected outcome and attach a screenshot!',
 				public: !isInternalUser
 			});
 		}));
--- a/src/vs/workbench/contrib/issue/browser/issueReporterOverlay.ts
+++ b/src/vs/workbench/contrib/issue/browser/issueReporterOverlay.ts
@@ -730,7 +730,13 @@ export class IssueReporterOverlay {
 			selectedExtension: this.selectedExtension,
 		});
 		this.data.issueSource = this.selectedIssueSource;
-		this.data.extensionId = fileOnExtension ? this.selectedExtension?.id : undefined;
+		// Preserve a preset `extensionId` while the extension list is still loading:
+		// `selectedExtension` may be undefined here even though the caller asked
+		// for a specific extension, and overwriting with `undefined` would prevent
+		// the catch-up retry in `updateExtensionOptions` from re-resolving it.
+		this.data.extensionId = fileOnExtension
+			? (this.selectedExtension?.id ?? this.data.extensionId)
+			: undefined;
 	}
 
 	private updateTitlePlaceholder(): void {
@@ -793,7 +799,15 @@ export class IssueReporterOverlay {
 			? this.model.getData().allExtensions.find(candidate => candidate.id.toLowerCase() === extensionId.toLowerCase())
 			: undefined;
 		this.selectedExtension = extension;
-		this.data.extensionId = extension?.id;
+		// Preserve the requested extensionId even when the extension list hasn't
+		// been populated yet (typical wizard flow: the constructor runs before
+		// `populateReporterDataAsync` finishes filling `allExtensions`). Without
+		// this preservation, the later catch-up retry in `updateExtensionOptions`
+		// sees `this.data.extensionId === undefined` and never re-resolves,
+		// dropping any preset extension data with it.
+		if (extensionId === undefined || extension) {
+			this.data.extensionId = extension?.id;
+		}
 		this.extensionSelect.select(this.getSelectedExtensionIndex());
 		this.updateExtensionValidation();
 		this.updateIssueSourceFlags();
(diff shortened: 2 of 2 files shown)


ulugbekna:
Force-pushed a correction.

The previous version guarded with `!extension.data`, but `issueService.ts` pre-populates `extension.data` on every entry in `enabledExtensions` (`data: options.data`), so my guard was always false and the new branch never  the bug remained in practice.fired 

Switched the guard to `!this.includeExtensionData`, which is the actual "extension data has been applied to the model" flag (initialized to `false`, flipped to `true` only by `applyExtensionIssueData`). This correctly fires on first resolution of the preset extension and still short-circuits on re-invocation after the extension list arrives asynchronously.

ulugbekna:
Found a deeper bug. Force-pushed.

When the wizard opens, the constructor calls `updateSelectedExtension(extensionId, false)` before `populateReporterDataAsync` has finished filling `allExtensions`. With an empty list, `find()` returns undefined, and two lines then clobber the requested `extensionId` to undefined:

- `updateSelectedExtension` line 796: `this.data.extensionId = extension?.id;`
- `updateIssueSourceFlags` line 733: `this.data.extensionId = fileOnExtension ? this.selectedExtension?.id : undefined;`

Once cleared, the catch-up retry in `updateExtensionOptions` (line 776: `if (!this.selectedExtension && this.data.extensionId)`) never fires when extensions finally load, so my new propagation branch never gets a chance to run. The preset `data` is lost regardless of any guard logic.

Fix: preserve the originally requested `extensionId` in both places when the extension list hasn't resolved yet. With this, the catch-up retry runs as designed and the new branch propagates `data` into the model.

ulugbekna:
Pushed follow-ups for the two issues spotted after the payload started arriving:

 `fileOnExtension=false`, so the section vanished from the UI even though the body still included it. Dropped the `fileOnExtension` gate so the review UI matches what gets submitted.

2. **Preset data was silently dropped for built-in  `updateSelectedExtension` has an early-return that switches built-in extensions to the VSCode source. My `else if` branch lived *after* that return, so for built-in callers the preset `data` never made it onto the model. Hoisted the propagation call to run *before* the built-in shortcut.extensions** 

3. **Redundant nested header on the GH  Copilot's NES feedback flow passed `issueBody: '# Description\n...'`, and the wizard wraps the description in its own `### Description` section, producing a nested `### Description / # Description` header. Dropped the redundant `# Description\n` prefix from `inlineEditDebugComponent.ts`.preview** 

cc @Giuspepe

Giuspepe (approved):

ulugbekna:
### Update: how the new wizard handles oversize extension data

For posterity, the new wizard's overflow handling lives in `IssueFormService.openIssueOnGithub` (`src/vs/workbench/contrib/issue/browser/issueFormService.ts`) and is functionally equivalent to the old reporter — just on a different code path:

1. The body builder wraps Extension Data (and System Info, Process Info, Workspace Info, Experiments, etc.) in `<details><summary>…</summary>…</details>` blocks via `createDetails` (`issueReporterOverlay.ts:2186`).
2. When the generated GH URL exceeds `MAX_URL_LENGTH` (7500), `tryCreateBodyWithIssueDataAttachment` calls `extractIssueData`, which strips every `<details>` block out of the body and bundles them into a single `issue-data.md` file. That file is uploaded through `GitHubUploadService.uploadViaMobileApi` (cached by content hash), and `createBodyWithIssueDataLink` re-inserts a `### Additional Issue Data\n\n[issue-data.md](<assetUrl>)` link where the blocks used to be.
3. If the upload path is unavailable (no GitHub token) or the shortened body is still too long, it falls back to `showClipboardDialog` → writes the full body to the clipboard and opens GH with a "We have written the needed data into your clipboard…" placeholder.

So the upload-as-attachment behavior the old reporter had is preserved; the bug fixed by #320103 is strictly about the preset `data` never reaching `model.extensionData` in the first place — once it lands there, the existing oversize plumbing takes over.

PR #320104, opened 2026-06-05
Migrate package manager from npm to pnpm

ulugbekna:
## Summary

Migrates the VS Code monorepo from **npm** to **pnpm** (`pnpm@10.32.1` via corepack).

 re-verify.

## What changed

**Workspace & manifests**
- Add `pnpm-workspace.yaml` with **explicit** members (mirroring `build/npm/dirs.ts`, not broad globs so manifest-only theme extensions and sim fixtures stay out) + an `onlyBuiltDependencies` allowlist for native/binary packages.
`pnpm`.
- `build/`, `remote/`, `remote/web` stay **standalone** (their own `.npmrc` with Node-runtime headers + own lockfile) **outside** the Electron-runtime workspace, so native modules build against the correct headers. Each gets its own `pnpm.overrides` + `pnpm.onlyBuiltDependencies` so lifecycle scripts still run under `--ignore-workspace`.

**`.npmrc` files**
- `node-linker=hoisted` (packaging assumes a flat `node_modules`), pnpm peer settings; drop `legacy-peer-deps`.

**Install orchestration & build scripts**
- Rewrite `build/npm/preinstall.ts` (pnpm guard), `postinstall.ts` (installs only the standalone dirs; root install is driven by `pnpm install`, no recursion), `dirs.ts` (`standaloneDirs`).
- Update `installStateHash.ts` + `build/azure-pipelines/common/computeNodeModulesCacheKey.ts` to hash pnpm lockfiles/workspace.
 Azure callers unaffected).

**CI**
- GitHub Actions: `corepack enable` + `pnpm install --frozen-lockfile` (standalone `build/` installs use `--ignore-workspace`).

**Lockfiles**
- Remove the obsolete `package-lock.json` files; ignore `pnpm-debug.log*`. The `pnpm-lock.yaml` files must be generated and committed by CI.

## Verification done locally
- `build` type-check clean on all changed files.
- `pnpm-workspace.yaml` parses; all 51 members resolve to real dirs.
- All root scripts are npm-free.
- Pre-commit hygiene hook passed.

## Follow-ups (need CI / too heavy to verify in-session)
- Generate & commit `pnpm-lock.yaml` (root) + standalone locks via `pnpm install` in CI.
 corepack/pnpm + rework `node_modules` tarball caching toward the pnpm store / `pnpm fetch`.
- Distro repo (`.build/distro`) coordination for the standalone installs.
- Full native build-from-source + packaging E2E (ASAR, copy-from-node_modules).

 This is an AI-assisted migration intended as a starting point; the lockfile generation and Azure pipeline sweep are required before it can go green.>


lajczi:
please use latest pnpm v11 and catalogs feature

juttadnanzafar-hue (commented):
jutt Adnan

PR #320109, opened 2026-06-05
fix: add overrideAuthType=token to CLI smoke tests and re-enable them

alexdima:
The Copilot CLI smoke tests were failing because overrideAuthType was not set, causing the CLI SDK to use HMAC auth (the default) against the mock LLM server. The mock server cannot validate HMAC signatures, so /models calls returned 401, model fetching failed, and prompt rendering threw "Unexpected generated prompt structure".

Add the missing ['github.copilot.advanced.debug.overrideAuthType', '"token"'] setting to both chatSessions.test.ts and copilotCli.test.ts (matching what agentsWindow.test.ts already does), and remove the it.skip that was added as a temporary workaround.


chrmarti (approved):

alexr00 (approved):

alexdima:
The linux failure seems to be timing related. I will follow-up with a PR that increases times a bit.

PR #320113, opened 2026-06-05
nes-datagen: add cursor-jump (NCLP) sample task

ulugbekna:
Extends `nes-datagen` to generate training data for the cursor-jump
(next-cursor-line) prediction model alongside the existing xtab path.

### CLI

A new `--sample-task` flag selects what to generate:

| value | what it emits |
| --- | --- |
| `xtab` (default) | unchanged xtab samples |
| `cursor-same-file` | cursor-jump samples where the jump stays in the active file |
| `cursor-cross-file` | cursor-jump samples where the jump goes to another file |
| `cursor-both` | both of the above |

Same-file detection is gated by `--same-file-jump-min-above` (default
`2`) and `--same-file-jump-min-below` (default `5`); a recording is
only used if the post-bookmark cursor lands farther than that window
from the request-time cursor.

### How it works

The datagen path reuses the production `XtabNextCursorPredictor` to
build the prompt  drift between datagen and production wouldverbatim 
corrupt the training distribution. The prompt is captured via a
no-op fetcher + telemetry builder; the expected response is built by
pure detector / formatter modules from the post-bookmark recording.

```
                  
      DetectedJump detectJump ( pure)                     recording 
 same-file vs cross-file                      
 applies edits in correct                     
    (descending) order                         
                  
                                
            {                 echo ___BEGIN___COMMAND_OUTPUT_MARKER___;                 PS1="";PS2="";unset HISTFILE;                 EC=$?;                 echo "___BEGIN___COMMAND_DONE_MARKER___$EC";             }
                  
 reuses production   cursorJumpPromptStep                          
    XtabNextCursorPredictor ''    
 reads back prompt from                       
    telemetry builder                          
                  
                                
            {                 echo ___BEGIN___COMMAND_OUTPUT_MARKER___;                 PS1="";PS2="";unset HISTFILE;                 EC=$?;                 echo "___BEGIN___COMMAND_DONE_MARKER___$EC";             }
                  
 cursorJumpResponseStep                        
 same-file: keptRange                         
 cross-file: path:line                        
                  
```

### Polish commits

After the feature landed, four follow-up commits address review
feedback:

- replace `SAMPLE_TASK_VALUES` tuple with a `NesDatagenSampleTask`
  string enum
 2
 cursor-jump everywhere (files via `git mv`, ids /
  comments / test descriptions updated)
- pass the whole recording to `documentIndexMapping` instead of
  slicing then backfilling post-request `documentEncountered` entries

### Tests

68 / 1  `npx vitest run test/pipeline`. Notable:skipped 

- `detectJump.spec.ts` covers same-file thresholds, cross-file
  resolution, and the (regression-tested) bug where multiple
  `changed` edits were applied ascending in-place instead of
  descending.
- Typecheck clean: `npx tsgo --noEmit --project tsconfig.json` in
  `extensions/copilot`.


extensions/copilot/test/pipeline/alternativeAction/processor.ts
```diff
@@ -97,6 +145,7 @@ export namespace Processor {
 		const recordingAfterRequest = recording.slice(recordingIdxOfRequestTime + 1);
 
 		return {
+			entries: recording,
```
ulugbekna: I don't like this being called entries -- it should be "wholeRecording"

extensions/copilot/test/pipeline/cursorJump/detectJump.ts
```diff
@@ -0,0 +1,230 @@
...
+ */
+export interface ICrossFileJump {
+	readonly kind: 'crossFile';
+	readonly toDocLogId: LogDocumentId;
+	readonly toRelativePath: string;
+	/** 0-based line number on the target doc. */
+	readonly toLine: number;
+}
+
+export type JumpDetectionResult<T> = Result<T, string>;
```
ulugbekna: this looks like it couldn't been inlined?

extensions/copilot/test/pipeline/output.ts
```diff
@@ -26,6 +27,19 @@ export interface ISampleMetadata {
 	readonly oracleEdits: readonly (readonly [start: number, endEx: number, text: string])[];
 	readonly originalPrompt: unknown[];
 	readonly modelResponse: string;
+	/**
+	 * Which datagen task produced this sample. `CursorBoth` is a CLI dispatch
+	 * mode only — each emitted sample is classified as one of the concrete
+	 * cursor variants, never `CursorBoth`.
+	 */
+	readonly task: Exclude<NesDatagenSampleTask, NesDatagenSampleTask.CursorBoth>;
+	/** Present only when {@link task} is a cursor-* task. */
```
ulugbekna: I wonder if having a union type to avoid optional fields would be a better engineering decision here.

ulugbekna (commented):
- have you added tests similar to what we have for the xtab samples?

extensions/copilot/test/pipeline/pipeline.ts
```diff
@@ -45,6 +49,14 @@ export type RunPipelineOptions = {
...
 
 export async function runInputPipeline(opts: RunPipelineOptions, log = console.log.bind(console)): Promise<void> {
+	const nesDatagenOpts = opts.nesDatagen!;
+	if (nesDatagenOpts.sampleTask !== NesDatagenSampleTask.Xtab) {
+		return runCursorPipeline(opts, log);
+	}
+	return runXtabPipeline(opts, log);
+}
+
+async function runXtabPipeline(opts: RunPipelineOptions, log: (...ps: any[]) => void): Promise<void> {
```
ulugbekna, line 88: - does it mean that cursor jump data generation doesn't support parallelism? 
- since the input file contents is read at once instead of streaming as in `runInputPipelineParallel`, the script can run out of memory and crash. I wonder if we could actually unite the non-parallel and parallel code with some nicely architectured code that allows for parallelism if necessary.

extensions/copilot/test/pipeline/pipeline.ts
```diff
@@ -189,6 +201,239 @@ export async function runInputPipeline(opts: RunPipelineOptions, log = console.l
...
+	installCursorJumpCapturingFetcher(serviceCollection);
+	const testAccessor = serviceCollection.createTestingAccessor();
+
+	try {
+		const configService = testAccessor.get(IConfigurationService);
+		await applyConfigFile(configService, configs);
+		await configService.setConfig(ConfigKey.TeamInternal.InlineEditsDebounce, 0);
+		await configService.setConfig(ConfigKey.TeamInternal.InlineEditsCacheDelay, 0);
+		await configService.setConfig(ConfigKey.TeamInternal.InlineEditsExtraDebounceEndOfLine, 0);
+		await configService.setConfig(ConfigKey.TeamInternal.InlineEditsExtraDebounceInlineSuggestion, 0);
```
ulugbekna: I wonder if we could extract this code given it's probably reused but also low-level details of overriding configs.

extensions/copilot/test/pipeline/pipeline.ts
```diff
@@ -189,6 +201,239 @@ export async function runInputPipeline(opts: RunPipelineOptions, log = console.l
...
+
+		type PromptOut = { originalRowIndex: number; system: string; user: string; keptRange: import('../../src/util/vs/editor/common/core/ranges/offsetRange').OffsetRange };
```
ulugbekna: I don't like this inline import.

extensions/copilot/test/pipeline/pipeline.ts
```diff
@@ -189,6 +201,239 @@ export async function runInputPipeline(opts: RunPipelineOptions, log = console.l
...
+		const prompts: PromptOut[] = [];
+		const promptErrors: { originalRowIndex: number; error: string }[] = [];
+		let completed = 0;
+		const start = Date.now();
+
+		const limiter = new Limiter<void>(concurrency);
+		const qualifying = processed.filter(p => jumps.has(p.originalRowIndex));
+
+		await Promise.all(qualifying.map(p =>
```
ulugbekna, line 326: wouldn't `Promise.all` fail if any of promises rejects? Are there better alternatives?

extensions/copilot/test/pipeline/pipeline.ts
```diff
@@ -189,6 +201,239 @@ export async function runInputPipeline(opts: RunPipelineOptions, log = console.l
...
+
+			const prompt: IGeneratedPrompt = { system: promptOut.system, user: promptOut.user };
+			const responseAdapter = { assistant };
+			samples.push(assembleSample(
+				promptOut.originalRowIndex + rowOffset,
+				prompt,
+				responseAdapter,
+				p,
+				'next-cursor-line-prediction',
+				/* modelResponse */ '',
```
ulugbekna: why is model response not populated with the expected response?

lszomoru (approved):

PR #320140, opened 2026-06-05
Cap get_terminal_output first-poll and non-prefix responses to a tail

meganrogge:
Fixes the spillover side of microsoft/vscode-internalbacklog#7869.

PR #315543 already shrinks **repeated** `get_terminal_output` polls by returning only the delta. But the **first** poll and the **non-prefix fallback** still returned the full buffer (up to the ~60 KB upstream cap in `outputHelpers.ts`), which is well above the Copilot SDK's ~10 KB spillover threshold and causes the agent to thrash with `read_file` calls on the spillover temp file.

This change adds a second, tighter cap inside the tool itself: when the full output exceeds 8 KB, return only the last 8 KB (line-aligned). The truncation marker includes both the omitted character count and a hint telling the agent how to recover the head if it needs to: re-run the command and redirect output to a file, then read that file.

Snapshot tracking (length + hash) still uses the full buffer, so subsequent delta polls work unchanged.

### Behavior

The tail cap applies on every code path (always on — not gated by an experiment), because the SDK spillover issue affects all users regardless of whether `chat.tools.terminal.outputDeltas` is enabled. The delta-tracking semantics introduced by #315543 remain gated by that experiment.

- Experiment off: full output if ≤ 8 KB; otherwise tail + recovery hint.
- Experiment on, first poll: same.
- Experiment on, unchanged poll: `unchanged since previous poll` marker (no output).
- Experiment on, pure-prefix delta poll: only the new characters (already small).
- Experiment on, non-prefix fallback: same tail treatment as first poll.

### Risks / follow-ups

- **8 KB is a tuned guess.** Picked to stay under the ~10 KB SDK spillover threshold with headroom for the prefix text. If the SDK threshold changes we may need to retune.
- **No telemetry** on how often truncation kicks in or whether the agent successfully recovers via the redirect hint. Worth adding to validate.
- **Line-alignment edge case**: single-line outputs with no `\n` in the last 8 KB fall back to a raw character cut, which could land mid-ANSI-escape. Rare.

### Tests

`getTerminalOutputTool.test.ts` — all 11 tests pass, including two added for the new behavior:
- `returns only the tail on first poll when output exceeds the tail budget`
- `returns only the tail on non-prefix fallback when output exceeds the tail budget`

### DIFF
--- a/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/getTerminalOutputTool.ts
+++ b/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/getTerminalOutputTool.ts
@@ -113,7 +113,7 @@ export class GetTerminalOutputTool extends Disposable implements IToolImpl {
 			return this._formatTailOrFull(output, `Output of terminal ${id}`);
 		}
 		if (currentOutputSnapshot.length === previousOutputSnapshot.length && currentOutputSnapshot.hash === previousOutputSnapshot.hash) {
-			return `Output of terminal ${id} unchanged since previous poll (${output.length} characters already shown). No new output.`;
+			return `Output of terminal ${id} unchanged since previous poll (${output.length} total characters in buffer). No new output.`;
 		}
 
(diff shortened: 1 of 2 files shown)


meganrogge:

lramos15 (approved):

PR #320141, opened 2026-06-05
Reinforce timeout + do-not-poll guidance in terminal tool descriptions

meganrogge:
Addresses microsoft/vscode-internalbacklog#7870.

The agent has been observed polling `get_terminal_output` for ~27 steps across three `npm` commands when a too-short timeout (120s for a 3–5 minute install) caused sync→background promotion. The system prompt preamble already says "do NOT poll", and `runInTerminal`'s `modelDescription` already recommends generous timeouts — but neither of those signals attaches to the spot where the model is actually making the decision (the `timeout` parameter description for `runInTerminal`, and the top-level description for `get_terminal_output`).

This PR reinforces both signals at the point of use:

**`runInTerminal` → `timeout` parameter description**

Before:
> Optional hard cap in milliseconds on how long the tool tracks the command before returning. Omit to let the command run to completion (recommended for package installs, builds, and long-running scripts). Use 0 to explicitly indicate no timeout.

After:
> Optional hard cap in milliseconds on how long the tool tracks the command before returning. **Recommended: 600000 (10 min) for package installs, 900000 (15 min) for large builds.** Omit entirely for commands that should run to completion (the safest default for installs and builds). Use 0 to explicitly indicate no timeout. **A too-short timeout forces background promotion and triggers polling — prefer generous timeouts.**

**`getTerminalOutput` → `modelDescription`**

Before:
> Get output from an active terminal execution (identified by the `id` returned from `run_in_terminal`).

After:
> Get output from an active terminal execution (identified by the `id` returned from `run_in_terminal`). **Only use this tool if you need to inspect output from a terminal that was started in async mode and you have a concrete reason to read it now. If a sync command timed out and moved to the background, you will be automatically notified on your next turn when it completes — do NOT poll with this tool. End your turn and wait for the notification.**

### Why not pull these from the system prompt preamble?

The preamble already contains both pieces of guidance, but models attend most strongly to tool parameter descriptions at call time, and the preamble has to compete with many other paragraphs. Restating the rule at the decision point is what gets it followed.

### Risks / follow-ups

- The two strings are longer now (more prompt tokens per tool definition). Estimated cost: ~80 extra tokens combined per request that includes these tools — small compared to the ~27 polling steps each ~1.5 KB of output that this is trying to prevent.
- Issue #7870 (proposal P5c) also asks us to verify the notification mechanism is reliable; that's a separate investigation and not in this PR.

### Tests

No assertions reference these exact strings (verified via grep for `"Optional hard cap in milliseconds"` and `"Get output from an active terminal"`). No test changes required.


meganrogge:

lramos15 (approved):

PR #320151, opened 2026-06-05
Add client-side OCR telemetry for chat image attachments

aashna:
## What

When a chat request includes one or more image attachments, run OCR (via `tesseract.js`) over the **first** attached image in parallel with the LLM request, and log the result as two new measurements on every telemetry event that already carries the image metrics:

- `ocrTextLength` - number of characters of extracted text (0 if OCR did not run / did not complete)
- `ocrConfidence` - 0-100 confidence score from the OCR engine (0 if OCR did not run / did not complete)

These are added to:
- `response.success`, `response.cancelled`, `response.error` (in `chatMLFetcherTelemetry.ts`)
- `automode.routerFallback`, `automode.routerModelSelection` (in `automodeService.ts`)

## Why

Mirrors the server-side OCR telemetry that landed in copilot-api (see companion PR `aashnagarg/image-metadata-ocr-hydro`). Client-side OCR lets us correlate the user-visible attachment with the routing decision **before** the request is sent, which is needed to evaluate whether image-aware routing models react to the actual image content vs only the attachment shape (count / bytes / dims).

Character length + confidence is enough signal to bucket queries (text-heavy screenshot vs photo vs diagram) without sending the raw OCR text from the client - the restricted-tier text is already captured server-side.

## Design

- New module `src/platform/image/node/ocrService.ts`:
  - Lazily initializes a single `Tesseract.Worker` (`createWorker('eng')`) shared across calls.
  - `extractOcrFromMessages(messages)` decodes the first `data:image/...;base64,...` URL and runs `worker.recognize`.
  - Swallows errors and returns `undefined` - OCR must never fail the request.
- In `chatMLFetcher.ts`, kick off the OCR promise immediately after collecting image telemetry measurements (only if `imageCount > 0`), pass it into `processSuccessfulResponse`, and `await` it just before emitting telemetry - so the LLM stream is never blocked.
- `imageTelemetry.ts`: added the two fields to `ImageTelemetryMeasurements` and `createEmptyImageTelemetryMeasurements` (default 0).
- `tesseract.js@^5.1.1` added to `extensions/copilot/package.json` dependencies.

## Privacy

Only `ocrTextLength` (integer) and `ocrConfidence` (integer 0-100) are logged from the client. The raw OCR text is **not** sent client-side - the server-side OCR (separate PR on copilot-api) writes the text to the restricted Hydro stream. Both new fields are classified `SystemMetaData` / `PerformanceAndHealth`.


aashna:
Closing: moving OCR-based image telemetry to server-side (CAPI). Client-side Tesseract quality wasn't good enough to be useful for the multimodal router, and any quality client-side OCR (cloud API, ONNX model, native OS) introduces bundle size, privacy review, or platform-fragmentation cost that isn't justified now that the server-side path exists. See github/copilot-api#34447.

Issue #317059, opened 2026-05-18
Steering message for terminal uses default model


lramos15:
Testing #316477

I ran a perpetual async terminal and it worked as expected, but once I killed the terminal it proceeded to do another turn which was unexpected to me.

<img width="928" height="641" alt="Image" src="https://github.com/user-attachments/assets/61334807-d2a1-4954-b861-0c83181112fd" />

meganrogge:
This is working as designed - we indicate to the agent when background processes complete with a steering message. Since it's system initiated, it costs the user nothing.

lramos15:
> it's system initiated, it costs the user nothing.

That's gone in the new billing model. This will cost users starting June 1

meganrogge:
We should use 4o mini for these

meganrogge:
## Steps to verify (in Insiders)

1. Enable trace logging: `Developer: Set Log Level…` → **Debug** 
2. Start a chat in Agent mode and ask Copilot to run an `async` terminal command that exits on its own (e.g. `sleep 5 && echo hi`) or one that needs input.
3. After the terminal completes / requests input / is killed, open **View → Output → Window** and look for a line like:
https://github.com/microsoft/vscode/blob/29b2a50ae84ca786544409f717a60c4e72bca632/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/runInTerminalTool.ts#L2528-L2532
4. Optionally set `chat.utilitySmallModel` to a different model in the Settings UI and repeat — the log line should show the new `vendor/id`.
5. Sanity check: a normal user-initiated chat request should *not* produce the trace line (the override only applies when `isSystemInitiated` is true).

roblourens:
PR reverted, see https://vscodeteam.slack.com/archives/C05BHT75R6G/p1779344317663659

isidorn:
I see the steering message. Adding verified label
<img width="1015" height="55" alt="Image" src="https://github.com/user-attachments/assets/bc4e177e-5767-42d9-902d-6430e95cb536" />

PR #320155, opened 2026-06-05
Use copilot/copilot-utility-small for terminal-tool steering messages

meganrogge:
Addresses microsoft/vscode-internalbacklog#7870.

### Use `copilot/copilot-utility-small` for steering messages

Steering messages (background completion, input-needed, terminal-disposed notifications) now resolve the `copilot/copilot-utility-small` model alias and pass it as `userSelectedModelId` instead of inheriting the conversation model. Falls back to the conversation model if the alias is unavailable.

### Route the utility-small alias to the free endpoint in `endpointProviderImpl`

The workbench-side override above sets `userSelectedModelId = 'copilot/copilot-utility-small'`, but inside the Copilot extension the request arrives at `EndpointProviderImpl.getChatEndpoint(request)` with `model.id === 'copilot-utility-small'` — a synthetic alias with no matching CAPI model. The downstream `getChatModelFromApiModel` lookup returns `undefined` and we silently fall back to `copilot-utility` (currently GPT-4.1, 1.2× multiplier).

Short-circuit `copilot-utility-small` and `copilot-utility` model ids to `_resolveUtilityFamily()` so the chat-participant path produces the same `CopilotUtilitySmallChatEndpoint` (gpt-4o-mini, 0× multiplier) that direct `getChatEndpoint(family)` callers already get. Without this, the workbench override above doesn't actually land on the free utility model.

### Prevent `isSystemInitiated` steering messages from polluting recent models

Adds `!options.isSystemInitiated` guard to the `addToRecentlyUsedList` call in `chatServiceImpl.ts` so steering turns do not surface the utility-small model in the user-facing model picker.

### Suppress steering on user-initiated terminal close (#317059)

When `terminalInstance.exitReason === TerminalExitReason.User`, skip the terminal-exited steering message to avoid burning credits on an action the user did not request.

Screenshot 2026-06-05 at 3 51 13 PM

### DIFF
--- a/extensions/copilot/src/extension/prompt/vscode-node/endpointProviderImpl.ts
+++ b/extensions/copilot/src/extension/prompt/vscode-node/endpointProviderImpl.ts
@@ -121,6 +121,14 @@ export class ProductionEndpointProvider extends Disposable implements IEndpointP
 			}
 		}
 
+		// Utility-family aliases (published by LanguageModelAccess under the copilot vendor)
+		// have synthetic ids that don't map to any real CAPI model, so the lookup below
+		// would silently fall back to `copilot-utility`. Route them through the family
+		// resolver so the chat-participant path matches direct `getChatEndpoint(family)` callers.
+		if (model.id === 'copilot-utility-small' || model.id === 'copilot-utility') {
+			return this.getChatEndpoint(model.id);
+		}
+
 		const modelMetadata = await this._modelFetcher.getChatModelFromApiModel(model);
 		// If we fail to resolve a model since this is panel we give copilot utility. This really should never happen as the picker is powered by the same service.
 		return modelMetadata ? this.getOrCreateChatEndpointInstance(modelMetadata) : this.getChatEndpoint('copilot-utility');
--- a/src/vs/workbench/contrib/terminalContrib/chatAgentTools/test/electron-browser/runInTerminalTool.test.ts
+++ b/src/vs/workbench/contrib/terminalContrib/chatAgentTools/test/electron-browser/runInTerminalTool.test.ts
@@ -87,6 +87,13 @@ suite('RunInTerminalTool', () => {
 	let chatServiceDisposeEmitter: Emitter<{ sessionResources: URI[]; reason: 'cleared' }>;
 	let chatSessionArchivedEmitter: Emitter<IAgentSession>;
 	let capturedSteeringRequests: { sessionResource: URI; message: string }[];
+	let sendRequestSignal: { promise: Promise<void>; resolve: () => void };
+	let lastUtilityAliasResolution: Promise<unknown> = Promise.resolve();
+	function renewSendRequestSignal() {
+		let resolve!: () => void;
+		const promise = new Promise<void>(r => { resolve = r; });
+		sendRequestSignal = { promise, resolve };
+	}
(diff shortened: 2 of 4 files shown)


src/vs/workbench/contrib/terminalContrib/chatAgentTools/test/electron-browser/runInTerminalTool.test.ts
```diff
@@ -246,6 +266,28 @@ suite('RunInTerminalTool', () => {
 		setConfig(TerminalChatAgentToolsSettingId.AutoApprove, value);
 	}
 
+	// Wait until at least `count` steering requests have been dispatched.
+	// Synchronizes on a deferred resolved by the stubbed `sendRequest`, so the
+	// assertion doesn't depend on a fixed number of microtask hops.
+	async function waitForSteeringRequest(count = 1) {
+		while (capturedSteeringRequests.length < count) {
```
dmitrivMS: can this turn into an endless loop?

src/vs/workbench/contrib/terminalContrib/chatAgentTools/test/electron-browser/runInTerminalTool.test.ts
```diff
@@ -246,6 +266,28 @@ suite('RunInTerminalTool', () => {
...
+
+	// For "no request expected" assertions: ensure the steering dispatch path
+	// has had a chance to run and decide not to send. Anchors on the
+	// utility-small alias resolution promise (which the dispatch chain awaits)
+	// rather than a fixed microtask count.
+	async function waitForSteeringDispatchToSettle() {
+		await lastUtilityAliasResolution;
+		// Drain the small chain of .then() callbacks (utilitySmallIdPromise →
+		// resolveSendOptions → sendRequest) that follow the alias resolution.
+		await Promise.resolve();
```
dmitrivMS: This looks like a hack... at least very strange.

src/vs/workbench/contrib/terminalContrib/chatAgentTools/test/electron-browser/runInTerminalTool.test.ts
meganrogge: I think this test should actually be removed. code is self explanatory

dmitrivMS (dismissed):

extensions/copilot/src/extension/prompt/vscode-node/endpointProviderImpl.ts
```diff
@@ -121,6 +121,14 @@ export class ProductionEndpointProvider extends Disposable implements IEndpointP
 			}
 		}
 
+		// Utility-family aliases (published by LanguageModelAccess under the copilot vendor)
+		// have synthetic ids that don't map to any real CAPI model, so the lookup below
+		// would silently fall back to `copilot-utility`. Route them through the family
+		// resolver so the chat-participant path matches direct `getChatEndpoint(family)` callers.
+		if (model.id === 'copilot-utility-small' || model.id === 'copilot-utility') {
```
dmitrivMS, line 128: @vritant24 FYI, don't know if this impacts BYOK/utility behavior

dmitrivMS (approved):

PR #320185, opened 2026-06-05
communing with the codebase

justschen:
increase fun working message rate to 1/50 from 1/100 and added another.

regular working messages will still be kind of relevant (based on the task that is running)


roblourens (approved):
lol love it

PR #320195, opened 2026-06-05
Stop exporting RunInTerminalTool from terminalContribChatExports (#318160)

### LINKED ISSUE #318160: RunInTerminalTool should not be exported
The `RunInTerminalTool` export here goes a long way in invalidating the reason it's hidden within a terminalContrib to begin with:

https://github.com/microsoft/vscode/blob/382e40aeea80678fe6bdb3ab9019f81c80f5d063/src/vs/workbench/contrib/terminal/terminalContribChatExports.ts#L1-L12

In order to reduce complexity you should only allow one-way dependencies. Currently this is the situation at a module level:

Image

The right way to do this is to export whatever you need on `ITerminalChatService` which lives in `contrib/terminal/` but it implemented inside `terminalContrib/chat/`. That way the interface is shared:

Image

@meganrogge

meganrogge:
Fixes #318160

Stops exporting `RunInTerminalTool` from `terminalContribChatExports.ts`. The only outside consumer (`chatQuestionCarouselPart`) just needed the underlying `ITerminalInstance` for a given execution id in order to attach an `onDidInputData` listener, so I exposed that on `ITerminalChatService` instead.


meganrogge:
/requires-eval-assessment terminalbench2 gpt-5.5,claude-opus-4.6,claude-opus-4.7

bhavyaus (approved):

PR #320286, opened 2026-06-07
Feature: Implement concatenation on Enter for string literals

### LINKED ISSUE #83196: Enteraction appending text before the new line
[EnterAction](https://code.visualstudio.com/api/references/vscode-api#EnterAction) Should allow appending text before the new line as well.

What I want to do is when a user press enter in the middle of a string to break it into two lines and concat the string with +

eg:

convert this,
```ballerina
string s = "Press enter here -> <- there"
```

to
```ballerina
string s = "Press enter here ->" 
    + " <- there"
```

But I can only append `+ "` to the string below. No way to add `"` to the string above :(

clarafdomingos:
### What is the main purpose of this Feature?

When a developer is editing a string literal and presses Enter in the middle of it, the editor currently has no built in mechanism to automatically split the string into two syntactically valid, concatenated substrings.

The implementation consists in concatenating the string automatically when Enter is pressed, using per-language adaptation.

Example in C++:
  - Before:
  `std::string s = "hel|lo world";`
  - After:
  ```
std::string s = "hel"
          + "lo world";
```

### What are the main changes that led to this Feature implementation?

Using VS Code's already implemented logic to detect when Enter is pressed, as well as Token processing to detect if this action was done inside a string literal.
Displaying the expected output based on the language of the current file.

### How can you test it?

Open Command Palette and activate 'Toggle String Concatenation on Enter', create a string for the given language your current coding file is on and press Enter.

Closes #83196

### DIFF
--- a/extensions/cpp/language-configuration.json
+++ b/extensions/cpp/language-configuration.json
@@ -128,5 +128,14 @@
 				"appendText": "// "
 			}
 		},
-	]
+	],
+	"stringConcatenation":{
+		"excludedPatterns": [
+			"^\\s*#\\s*include",
+			"^\\s*#\\s*pragma",
+			"^\\s*#\\s*error",
+			"^\\s*#\\s*warning",
+			"^\\s*#\\s*define"
+		]
+	}
 }
--- a/extensions/csharp/language-configuration.json
+++ b/extensions/csharp/language-configuration.json
@@ -126,5 +126,17 @@
 				"appendText": "/// "
 			}
 		},
-	]
+	],
+	"stringConcatenation":{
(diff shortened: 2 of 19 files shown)


clarafdomingos:

clarafdomingos:
Hello, @alexdima.
Just a friendly follow-up on this PR. We addressed one of the review suggestions from the bot, but there are still a few remaining items that we didn't address since we don't find it relevant for this feature. We'd appreciate it if someone could let us know if we are on the right track or if there are other details that need to be acknowledge before we continue.
Thank you in advance,
Clara and Vasco

Issue #316851, opened 2026-05-17
Find Bar's position overlaps and obscures Sticky Scroll current scope (function name etc) when UI is narrowed


theClubHopper:
Type: <b>Bug</b>

Background:
- VS Code has a feature called 'Sticky Scroll' which pins the current code scope (function names etc.) to the top of the editor pane while scrolled, giving users the ability to quickly see which section of code they're in even when scrolled deep into a function.
- VS Code has a Find Bar feature (opened on Windows via CTRL+F). When the UI is maximized, the Find Bar, when open, appears at the top-right of the editor pane, offset a half-row height below the Sticky Scroll function name.

The Issue:
When the VS Code window is resized horizontally, such as for a split-screen layout, depending on the width of the resized interface and the editing pane's zoom setting, the Find Bar may encroach and partially obscure the Sticky Scroll function name.

(Screenshots to follow)

Suggested fix: Ensure that the Find Bar does not overlap the Sticky Scroll text. This could either be achieved by increasing the Find Bar's default distance from the top of the editing pane or by detecting an overlap condition and moving it below the Sticky Scroll text when needed.

My Build:
Version: 1.120.0 (user setup)
Commit: 0958016b2af9f09bb4257e0df4a95e2f90590f9f
Date: 2026-05-12T20:17:22Z
Electron: 39.8.8
ElectronBuildId: 13870025
Chromium: 142.0.7444.265
Node.js: 22.22.1
V8: 14.2.231.22-electron.0
OS: Windows_NT x64 10.0.26200

VS Code version: Code 1.120.0 (0958016b2af9f09bb4257e0df4a95e2f90590f9f, 2026-05-12T20:17:22Z)
OS version: Windows_NT x64 10.0.26200
Modes:

<details>
<summary>System Info</summary>

|Item|Value|
|---|---|
|CPUs|Intel(R) Core(TM) i5-8500 CPU @ 3.00GHz (6 x 3000)|
[...]

PR #320290, opened 2026-06-07
fix: prevent find widget from overlapping sticky scroll when editor is narrow

Mr-Ahmadi:
The find widget uses `OverlayWidgetPositionPreference.TOP_RIGHT_CORNER` which always positions at `top: 0` of the editor viewport. When sticky scroll is visible, it renders on top of the sticky scroll content due to the find widget's higher `z-index` (35 vs 4).

This is especially noticeable when the editor window is narrowed (e.g., split-screen layout), where the find widget can encroach and partially obscure the sticky scroll function names.

## Closes
- #316851

## Fix

Listens to the sticky scroll controller for height changes and shifts the find widget down accordingly via `margin-top`:

- Uses `StickyScrollController.get()` (the canonical VS Code pattern) instead of a magic string ID
- When sticky scroll is **visible**: `margin-top = stickyScrollHeight + 4px` (4px is the base CSS margin, extracted as `FIND_WIDGET_TOP_MARGIN`)
- When sticky scroll is **hidden** or **disabled**: inline style cleared, falls back to CSS `margin-top: 4px`
- Re-queries the controller in `_reveal()` if unavailable at construction time (defensive, since it's registered `AfterFirstRender`)

## Changes

- `FIND_WIDGET_TOP_MARGIN` — named constant for the base CSS margin-top
- `_stickyScrollInitialized` / `_stickyScrollHeight` — tracks setup state and current height
- `_initStickyScroll()` — retrieves the controller via `StickyScrollController.get()`, subscribes to `onDidChangeStickyScrollHeight`, and records initial height
- `_updateStickyScrollOffset()` — applies or clears the inline `margin-top` based on `_stickyScrollHeight`

## Verification

- Sticky scroll enabled + find widget open → widget positioned below sticky scroll
- Sticky scroll height changes (e.g., multiline scope name) → widget follows dynamically
- Sticky scroll hidden/disabled → normal position restored
- Find widget visibility animation (CSS `transform`) unaffected
- Sticky scroll controller not yet created → re-queried on `_reveal()`

### DIFF
--- a/src/vs/editor/contrib/find/browser/findWidget.ts
+++ b/src/vs/editor/contrib/find/browser/findWidget.ts
@@ -23,6 +23,7 @@ import * as platform from '../../../../base/common/platform.js';
 import * as strings from '../../../../base/common/strings.js';
 import './findWidget.css';
 import { ICodeEditor, IOverlayWidget, IOverlayWidgetPosition, IViewZone, OverlayWidgetPositionPreference } from '../../../browser/editorBrowser.js';
+import { StickyScrollController } from '../../stickyScroll/browser/stickyScrollController.js';
 import { ConfigurationChangedEvent, EditorOption } from '../../../common/config/editorOptions.js';
 import { Range } from '../../../common/core/range.js';
 import { CONTEXT_FIND_INPUT_FOCUSED, CONTEXT_FIND_WIDGET_FOCUSED, CONTEXT_REPLACE_INPUT_FOCUSED, FIND_IDS, MATCHES_LIMIT } from './findModel.js';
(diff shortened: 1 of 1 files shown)


Mr-Ahmadi:

Mr-Ahmadi:
Updated the fix with improvements over the initial version:

- **Proper import**: Uses `StickyScrollController` class import and `StickyScrollController.get()` (canonical VS Code pattern) instead of a magic string ID
- **No `import type` needed**: The class is imported directly, consistent with how every other VS Code contribution references cross-contributions
- **Deduplicated logic**: Extracted sticky scroll setup into `_initStickyScroll()` helper, called from both constructor and `_reveal()`
- **Cleaner state tracking**: Uses `_stickyScrollInitialized` boolean flag instead of storing the entire controller reference, avoiding duplicate event registration on re-setup

Mr-Ahmadi:
Closing this PR — pivoting to a different fix.

PR #320301, opened 2026-06-07
chat: add Inspect Agent Host Subscriptions developer command

roblourens:
## What

Adds a Developer command, **Inspect Agent Host Subscriptions**, that opens a read-only markdown editor listing every resource we are subscribed to across all connected agent hosts — the local host plus every remote host in the Agents window. It's modeled on the existing **Inspect Chat Model References** command.

For each connection the editor shows a root-state summary (advertised agents, active session count, terminal count) and a table of active subscriptions (resource URI, kind, reference count, status).

## Why

There was no way to see, at runtime, what each agent host connection is actually subscribed to. This is a useful diagnostic when investigating subscription leaks, stuck/pending subscriptions, or reconnect behavior — across both the local utility-process host and remote hosts.

## How

- `AgentSubscriptionManager.getActiveSubscriptions()` returns read-only `IActiveSubscriptionInfo` descriptors (`resource` / `kind` / `refCount` / `status`). Status is derived from the subscription value: `pending` before the first snapshot, `error` if it failed, otherwise `snapshot`. The always-live root state is intentionally excluded (it's reported separately by the command).
- The subscription `kind` is now retained on the managed entry so it can be reported without inspecting subclass types.
- The accessor is exposed on the consumer-facing `IAgentConnection` interface and implemented by the local, remote-protocol, editor-remote, and null clients — each a one-line delegation matching the existing `getSubscription` / `getSubscriptionUnmanaged` pass-throughs.
- The command itself is registered once in the shared chat developer actions, so it appears in **both** VS Code and the Agents window. It is gated on `ChatContextKeys.enabled`.

## Notes for reviewers

- Adding a method to `IAgentConnection` ripples to all implementers; the four production implementations plus one test stub (`MockAgentConnection`) are updated. The other agent-connection test doubles extend `mock<…>()` and don't need the method.
- New unit tests in `agentSubscription.test.ts` cover the pending→snapshot transition (with reference counts) and the error status.
- Validated: `compile-check-ts-native` clean, full `npm run compile` with 0 errors, ESLint clean on changed files, `AgentSubscriptionManager` suite (15 passing incl. 2 new) and `AgentHostPty` suite (18 passing).

(Written by Copilot)

### DIFF
--- a/src/vs/platform/agentHost/browser/nullAgentHostService.ts
+++ b/src/vs/platform/agentHost/browser/nullAgentHostService.ts
@@ -36,7 +36,7 @@ export class NullAgentHostService implements IAgentHostService {
 
 	get rootState(): IAgentSubscription<RootState> { return notSupported(); }
 
-	getSubscription<T extends StateComponents>(_kind: T, _resource: URI): IReference<IAgentSubscription<ComponentToState[T]>> { return notSupported(); }
+	getSubscription<T extends StateComponents>(_kind: T, _resource: URI, _owner: string): IReference<IAgentSubscription<ComponentToState[T]>> { return notSupported(); }
 	getSubscriptionUnmanaged<T extends StateComponents>(_kind: T, _resource: URI): IAgentSubscription<ComponentToState[T]> | undefined { return undefined; }
 	getActiveSubscriptions(): readonly IActiveSubscriptionInfo[] { return []; }
 	dispatch(_channel: string, _action: SessionAction | TerminalAction | IRootConfigChangedAction): void { notSupported(); }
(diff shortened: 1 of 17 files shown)


roblourens:
Addressed both Copilot review comments in 217c386bd4f:
 `\|`), so URIs/holder names/error messages can't break the table.
- **Readonly contract**: `AgentSubscriptionManager.getActiveSubscriptions()` now returns `readonly IActiveSubscriptionInfo[]` to match the consumer-facing `IAgentConnection` interface.

(Written by Copilot)

kycutler (approved):

PR #320302, opened 2026-06-07
Support custom endpoints in the model selection for tool risk assessments

### LINKED ISSUE #320300: Allow Risk Assessment model calls to route to non-copilot endpoints
Currently the Risk Assessment option for terminal commands is hard coded to only route through copilot, though you can at least choose a different model that's available there. The request is to support routing these requests to non-copilot models, just as we can already do for the Utility model setting.

Orvid:
As per the title, this just expands the setting to allow routing it to local instances rather than forcing it through copilot.

Fixes #320300


Orvid:

Orvid:
@justschen Given your recent work on the risk assessment tools, you're probably a better person to review this than the person the bot assigned it to.

PR #320330, opened 2026-06-07
sessions: add previous/next session list navigation

sandy081:
Add "Navigate to Previous/Next Session" actions that move through the visible sessions list in display order (respecting grouping, filtering, and collapsed sections), opening the session adjacent to the active one and clamping at the list edges.

Keybindings mirror core "Previous/Next Editor" and browser tab switching: Ctrl+PageUp/PageDown (Cmd+Alt+Left/Right on macOS), with Alt+Up/Down as a secondary chord. All are gated on the sessions window and outside the editor area, so the editor's "Move Line Up/Down" stays intact while navigation still works from the chat input.


sandy081:
Implements #320331

roblourens (approved):
Sweet, thanks for adding this, I wanted to add it but then I was thinking that opening sessions in the CLI is kind of slow. But we should have it

PR #320332, opened 2026-06-07
terminal-suggest: add 'agent' subcommand completions for code CLI

connor4312:
Adds completions for the \code agent\ subcommand and its sub-subcommands (\host\, \ps\, \stop\, \kill\, \logs\) based on the definitions in \cli/src/commands/args.rs\.

### DIFF
--- a/extensions/terminal-suggest/src/completions/code.ts
+++ b/extensions/terminal-suggest/src/completions/code.ts
@@ -460,26 +460,22 @@ const agentHostOptions: Fig.Option[] = [
 	{
 		name: '--host',
 		description: 'Host the agent host should bind on. Defaults to \'localhost\'',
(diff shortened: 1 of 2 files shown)


connor4312:
Addressed the Copilot review: all 8 agent host/connection flags (\--host\, \--port\, \--connection-token\, \--connection-token-file\, \--server-data-dir\, \--name\, \--address\, and the connection \--tunnel\) were marked \isRepeatable\ with optional-valued args, which doesn't match the CLI where each is a single required value. Removed \isRepeatable\/\isOptional\ so completions no longer suggest invalid invocations. Also added test coverage for the \gent\ subcommands.

TylerLeonhardt (approved):

Issue #303752, opened 2026-03-21
bad \ escapes in session list and running terminal commands


roblourens:
<img width="304" height="81" alt="Image" src="https://github.com/user-attachments/assets/dd91f57b-9f04-4865-bf6e-810bf598e304" />

Either, the sessions list description should be rendered as markdown. Or renderAsPlaintext should strip markdown icons. Or the lock icon should not be part of invocationMessage but only be added at the view level.

roblourens:
The other weird escapes at least are definitely still an is sue

alexdima:
The `$(lock)` is no longer there but the `\-` I think are there

roblourens:
Fixing two remaining problems: bad `-` escapes and `&amp;` escapes

<img width="294" height="65" alt="Image" src="https://github.com/user-attachments/assets/46febcb2-f33b-4357-a7ba-7a5e2c3d4c98" />

PR #320336, opened 2026-06-08
Fix double-escaped entities in renderAsPlaintext code spans

roblourens:
For #303752

I think this is ok but please take a quick look @mjbvz 

## Problem

Text inside markdown code spans (e.g. ````foo & bar````) rendered through `renderAsPlaintext` came out **double-escaped** — `&` appeared as `&amp;`, `<`/`>` as `&lt;`/`&gt;`. This surfaced in the agent sessions list, where a local (VS Code harness) session's in-progress description is derived from a tool invocation message via `getInProgressSessionDescription` → `renderAsPlaintext`.

## Root cause

In `createPlainTextRenderer`, the `codespan` handler called `escape(text)`. But marked already HTML-escapes code span text during tokenization. The trailing unescape pass in `renderAsPlaintext` only reverses a single level, so the extra `escape()` left entities double-escaped. Code blocks were unaffected because marked passes their text raw (escaped exactly once).

| Input | Before | After |
|---|---|---|
| `Run \`tests & build\`` | `Run tests &amp; build` | `Run tests & build` |
| `Use \`<form>\`` | `Use &lt;form&gt;` | `Use <form>` |

## Fix

Remove the redundant `escape()` from the `codespan` plaintext renderer so it matches code-block behavior. Added a regression test.

This is a shared low-level helper, so the fix corrects the same double-escaping for all `renderAsPlaintext` callers (hovers, pickers, etc.), not just the agent sessions list.

(Written by Copilot)

### DIFF
--- a/src/vs/base/browser/markdownRenderer.ts
+++ b/src/vs/base/browser/markdownRenderer.ts
@@ -756,7 +756,7 @@ function createPlainTextRenderer(): marked.Renderer {
 		return text;
 	};
 	renderer.codespan = ({ text }: marked.Tokens.Codespan): string => {
-		return escape(text);
+		return text;
 	};
 	renderer.br = (_: marked.Tokens.Br): string => {
 		return '\n';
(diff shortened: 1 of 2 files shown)


mjbvz (approved):
Thanks yes lgtm

PR #320347, opened 2026-06-08
Validate redirect scheme and strip credentials on cross-origin redirects in MCP HTTP client

g0w6y:
## Summary

The MCP HTTP client in `src/vs/workbench/api/common/extHostMcp.ts` follows HTTP redirects (301/302/303/307/308) without validating the scheme of the redirect target and without stripping credential headers when the target is a different origin. Two issues result:

1. **Arbitrary scheme follow.** The redirect target is resolved with `new URL(location, currentUrl)` and passed straight back into the fetch path. On the Node extension host, `_fetchInternal` dispatches `unix://` and `pipe://` URLs over a local socket (`extHostMcpNode.ts`). A `Location: unix:///var/run/docker.sock#/...` header therefore causes the client to issue requests to local sockets such as the Docker daemon socket. MCP server URLs are already restricted to `http(s)` at configuration time (`mcpConfiguration.ts`, `pattern: ^https?:\/\/.+`); the redirect path did not enforce the same restriction.

2. **Cross-origin credential replay.** For 307/308 the method, body and all headers are preserved across the redirect. The configured `Authorization` bearer token (OAuth or a static key from the `mcp.json` `headers`), the `Mcp-Session-Id`, and the JSON-RPC body are re-sent to the redirect target even when it is a different origin. Browser `fetch` and `curl` strip credentials on a cross-origin redirect; the MCP client did not.

## Fix

In the redirect loop:

- Only follow redirects whose resolved protocol is `http:` or `https:`; otherwise stop following and log a warning. This blocks `unix://`, `pipe://`, `file://`, etc.
- When the redirect target origin differs from the current origin, drop credential-bearing headers (`Authorization`, `Cookie`, `Proxy-Authorization`, `Mcp-Session-Id`) before the request is re-sent, matching `fetch`/`curl` behavior.

Same-origin redirects are unaffected and keep their headers. The change is confined to the redirect loop plus two module-level constants.

## Testing

- Manual: a redirect server returning `307 Location: <cross-origin>` no longer receives the configured `Authorization` header or `Mcp-Session-Id`.
- Manual: a redirect server returning `302 Location: unix:///.../docker.sock#/...` is no longer followed; a warning is logged and the redirect response is returned to the caller.

## Disclosure

This issue was reported to the Microsoft Security Response Center (MSRC) via the researcher portal — **Case number 116632** (VULN-187202) — by **Gouri Sankar A**.


g0w6y:

src/vs/workbench/api/common/extHostMcp.ts
```diff
@@ -862,7 +869,27 @@ export class McpHTTPHandle extends Disposable {
...
+			}
+
+			// On a cross-origin redirect, strip credential-bearing headers so tokens and
+			// session ids configured for the original origin are not replayed to another host.
+			if (currentUrlParsed.origin !== nextUrlParsed.origin) {
+				for (const name of Object.keys(init.headers)) {
+					if (CROSS_ORIGIN_STRIPPED_HEADERS.has(name.toLowerCase())) {
+						delete init.headers[name];
+					}
+				}
```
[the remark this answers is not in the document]
g0w6y: Thanks. In this code path `init` is typed `MinimalRequestInit`, whose `headers` is declared `Record<string, string>` (see the interface in this file), not the DOM `RequestInit`. The existing code already mutates it as a plain object (e.g. `init.headers['user-agent'] = ...` earlier in `_fetch`), so `Object.keys(...)` / `delete` is correct and consistent with the surrounding code. There is no `Headers` instance or tuple array on this path, so normalizing isn't needed.

src/vs/workbench/api/common/extHostMcp.ts
```diff
@@ -862,7 +869,27 @@ export class McpHTTPHandle extends Disposable {
...
-			const nextUrl = new URL(location, currentUrl).toString();
+			const currentUrlParsed = new URL(currentUrl);
+			const nextUrlParsed = new URL(location, currentUrl);
+
+			// Only follow redirects to http(s). Blocks a malicious Location header from
+			// reaching the unix:// / pipe:// socket dispatcher or other local schemes.
+			if (!ALLOWED_REDIRECT_PROTOCOLS.has(nextUrlParsed.protocol)) {
+				this._log(LogLevel.Warning, `Refusing to follow MCP redirect to non-http(s) target (${nextUrlParsed.protocol})`);
+				break;
+			}
```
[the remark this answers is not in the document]
g0w6y: Good point — agreed. Updated so a disallowed redirect scheme now throws an explicit error instead of `break`, so the connection fails deterministically rather than the caller treating the 3xx as final. Same-origin credential stripping still proceeds normally for allowed targets.

g0w6y:
@TylerLeonhardt @dmitrivMS Any Update ?

src/vs/workbench/api/common/extHostMcp.ts
```diff
@@ -862,7 +869,28 @@ export class McpHTTPHandle extends Disposable {
...
+			// reaching the unix:// / pipe:// socket dispatcher or other local schemes.
+			// Fail closed so the connection errors deterministically rather than the
+			// caller treating the 3xx response as final.
+			if (!ALLOWED_REDIRECT_PROTOCOLS.has(nextUrlParsed.protocol)) {
+				throw new Error(`MCP server redirected to a non-http(s) target (${nextUrlParsed.protocol}), which is not allowed`);
+			}
+
+			// On a cross-origin redirect, strip credential-bearing headers so tokens and
+			// session ids configured for the original origin are not replayed to another host.
+			if (currentUrlParsed.origin !== nextUrlParsed.origin) {
```
connor4312, line 885: We should also strip if the scheme downgrades https->http

src/vs/workbench/api/common/extHostMcp.ts
```diff
@@ -333,6 +333,13 @@ type HttpModeT =
 
 const MAX_FOLLOW_REDIRECTS = 5;
 const REDIRECT_STATUS_CODES = [301, 302, 303, 307, 308];
+// MCP server URLs are restricted to http(s) at configuration time; the redirect
+// path must enforce the same so a Location header cannot reach unix://, pipe://,
+// file://, etc.
+const ALLOWED_REDIRECT_PROTOCOLS = new Set(['http:', 'https:']);
+// Credential-bearing headers that must not be replayed to a different origin
+// after a redirect (matches browser fetch / curl behavior). Compared case-insensitively.
+const CROSS_ORIGIN_STRIPPED_HEADERS = new Set(['authorization', 'cookie', 'proxy-authorization', 'mcp-session-id']);
```
dmitrivMS, line 342: Should `api-key` and similar be included (e.g. `/^(x-)?(api[-_]?key|auth|token|secret|signature|session)/i`)?

connor4312: I'm not sure we should get too creative here. Removing Authorization is sane as vscode can implicitly add that for future sessions once authenticated, which may be non-obvious to the user, but other manually configured headers should be passed through.

src/vs/workbench/api/common/extHostMcp.ts
g0w6y: > We should also strip if the scheme downgrades https->http

@connor4312 Thanks for the review! I am currently away from my laptop, but I wanted to mention that the existing currentUrlParsed.origin !== nextUrlParsed.origin check actually already handles the https -> http downgrade natively.
Since the Node.js URL API includes the protocol in the origin string (e.g., "[https://example.com](https://example.com)" vs "[http://example.com](http://example.com)"), any protocol downgrade results in a strict inequality. Because the origins no longer match, the credential headers will be stripped automatically.

connor4312: oh, right 👍

connor4312 (approved):
thanks!

vritant24 (approved):

g0w6y:
Thanks for reviewing and merging this, @connor4312 / @vritant24.

For downstream tracking I'd like to flag that this is a security fix and ask about a formal advisory. The merged change addresses two issues in the MCP HTTP client:

- Cross-origin credential disclosure: on a 307/308 redirect the client replayed the `Authorization` bearer token, any static `mcp.json` header (e.g. API keys), and the `Mcp-Session-Id` to the redirect target even on a different origin (browsers/curl strip these). This is attacker-readable token theft in the default configuration.
- Redirect scheme confusion: redirects were followed without a scheme allow-list, so a `Location: unix:///var/run/docker.sock#/...` reached the unix:// / pipe:// socket dispatcher.

I reported this to MSRC (Case 116632 / VULN-187202). Would the team be willing to file a GitHub Security Advisory for the repository and request a CVE so users can track the fixed version? I'm happy to provide the full reproduction and write-up for the advisory. Thanks again.

PR #320467, opened 2026-06-08
Registers @vscode/markdown-editor in markdown-language-features extension

hediet:
The only entrypoint is:
```json
      {
        "viewType": "vscode.markdown.editor",
        "displayName": "Markdown Editor (Experimental)",
        "priority": "option",
        "selector": [
          {
            "filenamePattern": "*.md"
          }
        ]
      }
```

FYI @mattbierner


dbaeumer (approved):

mjbvz:
I'm pretty strongly against adding a WYSIWYG markdown editor to core VS Code. Instead if we want a richer markdown editor, I think we should invest in ways to make our existing text editor richer, such as variable font sizes and inline rendering of images

PR #320473, opened 2026-06-08
Add claude-opus-4.8 to tool search model list

bhavyaus:
Adds `claude-opus-4.8` to the `models` list for the `ToolSearchTool` . This is already enabled for the model.
This change only makes the tool available in the model picker to toggle.


bhavyaus:
This is already enabled for the model. This makes the tool available in the model picker to toggle.

rzhao271 (approved):

PR #320478, opened 2026-06-08
Add speech recognition context hints

meganrogge:
## Why

Voice chat uses the generic speech-to-text path with little context about what the user is likely to say, so developer phrases like chat participant names and slash commands ("at workspace slash fix") are often misheard. This PR adds a way to pass contextual hints to speech providers so they recognize these phrases.

## Before and After

The rewrite from spoken words to chat syntax ("at workspace" → `@workspace`) already existed. What was missing was feeding the vocabulary *into* the recognizer so it hears the phrases right in the first place.

**Before:** You say _"at workspace slash fix this bug"_ → recognizer mishears it (_"that workspace flash fix..."_) → the rewrite can't match → you get garbled text.

**After:** You say _"at workspace slash fix this bug"_ → recognizer is biased by hints (`at workspace`, `slash fix`) → transcribes it correctly → the existing rewrite turns it into **`@workspace /fix this bug`**.

Same end result, now reliable, because the recognizer knows which developer phrases to expect.

## Changes

- Add optional `context` and `phraseHints` to proposed `SpeechToTextOptions`.
- Thread speech-to-text options through the main thread/ext host speech provider boundary.
- Have voice chat supply agent and slash-command phrase hints from its existing phrase map.
- Add coverage for chat phrase hint propagation.

Refs: https://github.com/microsoft/vscode-internalbacklog/issues/7918

## Validation

- `VS Code - Build` / Core Typecheck: 0 errors
- `./scripts/test.sh --run src/vs/workbench/contrib/chat/test/common/voiceChatService.test.ts --grep VoiceChat`: 4 passing
- `git diff --check`

### DIFF
--- a/src/vs/workbench/api/browser/mainThreadSpeech.ts
+++ b/src/vs/workbench/api/browser/mainThreadSpeech.ts
@@ -57,7 +57,7 @@ export class MainThreadSpeech implements MainThreadSpeechShape {
 				const disposables = new DisposableStore();
 				const session = Math.random();
 
-				this.proxy.$createSpeechToTextSession(handle, session, options?.language);
+				this.proxy.$createSpeechToTextSession(handle, session, options);
 
 				const onDidChange = disposables.add(new Emitter<ISpeechToTextEvent>());
 				this.speechToTextSessions.set(session, { onDidChange });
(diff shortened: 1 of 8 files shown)


meganrogge:
Superseded by #320673, which is based on a branch in microsoft/vscode rather than the meganrogge fork.

Issue #320485, opened 2026-06-08
chat: add Configure Tools CodeLens for .toolsets.jsonc


jamesmontemagno:
## Request
Add **Configure Tools** CodeLens support when editing .toolsets.jsonc.

## Desired behavior
- Show Configure Tools CodeLens for tool set definitions in .toolsets.jsonc.
- Reuse existing tools picker UX.
- Persist selected tools back into the tool set 	ools array.

## Why this matters
- Improves discoverability and consistency with other tool configuration flows.
- Reduces manual JSON editing and syntax errors.
- Speeds up authoring and maintenance of tool sets.

## Notes
CodeLens should behave correctly with refs/qualified names and preserve stable serialization semantics.

jamesmontemagno:
Fixed in : https://github.com/microsoft/vscode/pull/320486

jamesmontemagno:
@aeschli can be close when ready now that merged

aiday-mar:
I have the following toolset file open. Is there supposed to be codelens here?

<img width="655" height="705" alt="Image" src="https://github.com/user-attachments/assets/b39c2987-2df9-44d0-ac16-d708b1737836" />

aeschli:
@aiday-mar Can you paste the file?
I see this:

<img width="294" height="282" alt="Image" src="https://github.com/user-attachments/assets/955e0b80-f126-4de1-9269-2e3d3e0e97dd" />

Version: 1.126.0-insider (Universal)
Commit: 7e7950df89d055b5a378379db9ee14290772148a
Date: 2026-06-23T01:46:13Z
Electron: 42.2.0
ElectronBuildId: 14159160
Chromium: 148.0.7778.97
Node.js: 24.15.0
V8: 14.8.178.14-electron.0
OS: Darwin arm64 24.6.0

aiday-mar:
The file is:

```
{
	"test": {
		"tools": [
			"vscode",
			"execute",
			"read",
			"agent",
			"edit",
			"search",
			"web",
			"browser",
			"component-explorer/*",
			"vscode-automation-mcp/*",
			"github/*",
			"playwright/*",
			"github.vscode-pull-request-github/issue_fetch",
			"github.vscode-pull-request-github/labels_fetch",
			"github.vscode-pull-request-github/notification_fetch",
			"github.vscode-pull-request-github/doSearch",
			"github.vscode-pull-request-github/activePullRequest",
			"github.vscode-pull-request-github/pullRequestStatusChecks",
[...]

Issue #320484, opened 2026-06-08
chat: deleting last user tool set should remove .toolsets.jsonc


jamesmontemagno:
## Problem
When deleting user-defined tool sets, removing the final remaining tool set leaves .toolsets.jsonc on disk (often as an empty or effectively empty file).

## Expected behavior
If the last user tool set is deleted, the extension should delete .toolsets.jsonc entirely instead of leaving an empty config file behind.

## Why this matters
- Keeps workspace config clean and intention-revealing.
- Matches user expectation that no tool sets means no tool-set file.
- Avoids edge cases in file watchers/parsers around empty config artifacts.

## Repro
1. Create .toolsets.jsonc with one user tool set.
2. Delete that tool set via the UI/action.
3. Observe file remains instead of being removed.

## Suggested fix
- On delete, if no user tool sets remain, remove .toolsets.jsonc.
- Respect FS trash capability checks and provide actionable error messages if removal fails.

jamesmontemagno:
Fixed in : https://github.com/microsoft/vscode/pull/320486

jamesmontemagno:
@aeschli can be close when ready now that merged

aiday-mar:
Not sure what the verification steps here are, the filter icon has disappeared in the panel chat and I am wondering where to find the tool sets?

jamesmontemagno:
Verification steps:
1. Create custom tool set from tool picker
2. The file should be create in a directory such at C:\Users\jamont\AppData\Roaming\Code\User\profiles\32a85d19\prompts\test.toolsets.jsonc  this will be the file that opens
3. delete it from the delete icon on the picker
4. verify the file is deleted

Issue #320444, opened 2026-06-08
Add "Create from current selection" to Configure Tool Sets


jamesmontemagno:
## Summary

Today, "Configure Tool Sets..." lets users create a new tool sets file or open an existing user-defined tool set, but it does not help them capture the tools they have already chosen in an active chat session. This creates unnecessary friction when a user has already curated a useful tool selection in agent mode and wants to turn that selection into a reusable tool set.

We should add a "Create from current selection" flow that creates a new tool set definition pre-filled from the currently enabled tools and tool sets in the active chat input.

## Problem

The current flow is file-oriented, not selection-oriented:

- "Configure Tool Sets..." only offers:
  - create a new tool sets file
  - open an existing user tool set file
- The active chat already tracks the current enabled tool/tool set state
- The tools picker already exposes a "Configure Tool Sets..." entry, but that action loses the context of what the user currently has selected

This means users have to manually re-create a tool set after they have already done the work of selecting the right tools in chat.

## Proposal

Add a new action inside the "Configure Tool Sets..." flow:

- **"Create from current selection..."**

This action should:

- Detect the current tool selection from the active chat widget
- Create a new .toolsets.jsonc file, or allow appending to a chosen user tool sets file
- Insert a new tool set object with the 	ools property pre-filled from the current selection
- Open the file in the editor so the user can finish naming, describing, and optionally changing the icon

### Entry point

The feature should be available from the existing "Configure Tool Sets..." command.

When invoked from the tools picker while a chat widget is active, the command should have access to the current tool selection immediately.

If invoked without suitable active chat context, we should either hide "Create from current selection", or show it but fall back gracefully with a clear message.

### Suggested flow

Inside "Configure Tool Sets...", offer:

- Create from current selection...
- Create new tool sets file...
- Existing user tool sets...

If the user picks "Create from current selection...":

1. Prompt for the target file name, or allow picking an existing user tool sets file
2. Prompt for the new tool set name
3. Create a tool set stub pre-filled from the current selection
4. Open the resulting file

### Generated content example

`jsonc
{
  "myToolSet": {
    "tools": [
      "read",
      "search",
      "github/issues"
    ],
    "description": "",
    "icon": "tools"
  }
}
`

### Source of truth

The current selection already exists as an enablement map in the active widget:
- src/vs/workbench/contrib/chat/browser/widget/input/chatSelectedTools.ts
- src/vs/workbench/contrib/chat/browser/attachments/chatVariables.ts

[...]

PR #320486, opened 2026-06-08
chat: Improve tool sets with "create from selected", delete toolset, "configure tools" options

jamesmontemagno:
## Summary
This PR improves the chat tool-set configuration experience and fixes correctness issues around serialization and deletion behavior.

## What changed
1. Added Configure Tools CodeLens support for .toolsets.jsonc
- Adds in-editor Configure Tools CodeLens for tool set files.
- Reuses the tools picker and writes selected tools back to the tools array.

2. Fixed delete behavior for user tool sets
- Deleting the last tool set now removes the tool set file.
- Uses file-system capability checks for trash support.
- Surfaces actionable error notification on failure.

3. Fixed partial deselection correctness
- If a full tool set is selected and one child tool is unchecked, deselection now wins.
- Tool set refs serialize only when all member tools are still enabled.
- Prevents deselected tools from reappearing via parent selection.

4. Added regression coverage
- Added tests for empty-file deletion semantics and tool-set/member serialization edge cases.

## Issues
- Refs #320444
- Refs #320484
- Refs #320485

Video:

https://github.com/user-attachments/assets/2a438290-d732-4599-80ee-a52a6036c9da


dmitrivMS:
FWIW, I don't know what the future of this UI in the context of migration to AH+CLI...

jamesmontemagno:
@dmitrivMS makes sense, i think these are nice enhancements to have, but maybe it all goes away long term? but i do like the whole idea of tool sets

@TylerLeonhardt  let me know if i can help at all on this

src/vs/workbench/contrib/chat/browser/chat.shared.contribution.ts
```diff
@@ -186,6 +186,7 @@ import { IPlanReviewFeedbackService, PlanReviewFeedbackService } from './planRev
 import { PluginUrlHandler } from './pluginUrlHandler.js';
 import { PromptUrlHandler } from './promptSyntax/promptUrlHandler.js';
 import { ConfigureToolSets, UserToolSetsContributions } from './tools/toolSetsContribution.js';
+import './tools/toolSetsCodeLensProvider.js';
```
TylerLeonhardt: I feel like this would be better off in the toolset contribution file? Since it's related.

aeschli:
I know that there are tool set fans, but its a niche feature (VS Code local only) without any prospect to become more prominent. When migrating to Copilot CLI it will likely go away. The place to configure tools will be the custom agent files

aeschli (dismissed):

aeschli:
I reverted the change in toFullReferenceNames. It would be a bug in the creation of the IToolAndToolSetEnablementMap if the the tool set is not checked when all its tools are checked. If you see a mismatch, then this needs to be fixed there.

aeschli (dismissed):

alexdima (dismissed):

dmitrivMS (approved):

jamesmontemagno:
Wow thanks ya'l!! So exciting!

Issue #315193, opened 2026-05-08
AHP/Copilot attachments enablement


rebornix:
Copilot CLI in EH

<img width="598" height="330" alt="Image" src="https://github.com/user-attachments/assets/1b369c80-5532-4014-bfcf-a335ea359ed8" />

AHP/Copilot

<img width="604" height="242" alt="Image" src="https://github.com/user-attachments/assets/44a323be-fceb-4b0e-9a96-eae9f9d5244e" />

It's a good time for us to review which attachments should be enabled, other than trying to enable all of them.

anthonykimseattle:
Suggested attachment checklist based on the current Copilot CLI attachment picker:

- [x] `Files & Folders...`
  - State: works in Copilot CLI / EH and appears to work in AHP/Copilot.
  - Recommendation: keep enabled. This is baseline support, not the first area to tackle.

- [x] Recent/open file rows under the separator
  - State: regular file references; should work through the existing AHP file attachment path.
  - Recommendation: keep enabled.

- [x] `Open Editors`
  - State: visible and likely technically works for file-backed editors.
  - Problem: one click can add a large number of attachments and is hard to undo.
  - Recommendation: hide/disable for AHP for now.

- [x] `Screenshot Window`
  - State: works in Copilot CLI / EH via image attachment support.
  - AHP state: backend path looks mostly ready, but image capability is not enabled.
  - Recommendation: enable early and verify with a vision-capable model.

- [ ] `GitHub Issues...`
  - State: provided through extension-contributed chat context in Copilot CLI / EH.
  - AHP state: generic context entries are not currently converted.
  - Recommendation: investigate next; likely needs AHP support for generic/text context attachments.

- [ ] `GitHub Pull Requests...`
  - State: Copilot CLI / EH has special handling for PR/reference-style context.
  - AHP state: no equivalent conversion today.
  - Recommendation: tackle together with `GitHub Issues...`; do not enable with a flag only.

- [ ] `Problems...`
  - State: works in Copilot CLI / EH through diagnostic prompt/context rendering.
  - AHP state: diagnostics are not converted today.
  - Recommendation: add diagnostic-to-text/blob conversion before enabling.

- [ ] `Symbols...`
  - State: works in Copilot CLI / EH as selection context.
  - AHP state: current symbol path can lose the selected range when mapped to the Copilot SDK.
  - Recommendation: fix symbol/selection mapping first, then enable.

- [ ] `Source Control...`
[...]

PR #320523, opened 2026-06-09
Preserve symbol attachment selection range

dmitrivMS:
## Why

`AgentHostSessionHandler._convertVariableToAttachment` emits symbol references via `_toSelectionAttachment(..., 'symbol', ...)` — i.e. a `Resource` with `displayKind: 'symbol'` **and** a populated `selection.range`. But `CopilotAgentSession._toSdkAttachment`'s selection branch only triggered on `displayKind === 'selection'` (the implicit-editor-selection case wired up in #314245), so symbol attachments silently fell through to the plain `file` mapping and lost their range — the bullet called out under the *Symbols…* row of #315193.

## What changed

In `src/vs/platform/agentHost/node/copilot/copilotAgentSession.ts`, the selection branch now keys off the **`selection` field** rather than `displayKind === 'selection'`. Any Resource carrying a `selection.range` — `'selection'`, `'symbol'`, or any future range-bearing kind — maps to the SDK `selection` variant with sliced text. On read failure it downgrades to a plain `file` reference, as before.

## Tests

Two new cases in `src/vs/platform/agentHost/test/node/copilotAgentSession.test.ts`:

- `maps symbol Resource attachments to SDK selection so the range survives (#315193)`
- `falls back to file reference when reading a symbol Resource attachment fails`

## Related

- Refs #315193
- Builds on #314245 (selection plumbing)


src/vs/platform/agentHost/node/copilot/copilotAgentSession.ts
```diff
@@ -865,10 +875,55 @@ export class CopilotAgentSession extends Disposable {
...
+		if (mime?.startsWith('image/')) {
+			return mime;
+		}
+		// Honor the protocol's advisory `'image'` hint even when the path has no extension we recognize — fall back to
+		// PNG so the SDK still routes through the vision channel rather than dropping it on the `file` path with no MIME.
+		if (displayKind === 'image') {
+			return mime ?? 'image/png';
+		}
+		return undefined;
+	}
```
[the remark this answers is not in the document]
dmitrivMS: Good catch — that `mime ??` was an accidental coalesce. By that branch `mime` is either `undefined` or a non-image MIME (the `image/*` case returned above), so a `.wav` flagged as `displayKind: 'image'` would have shipped as `audio/x-wav`, contradicting the helper''s purpose.

Fixed in `f8cc185` — `displayKind: 'image'` now forces `'image/png'` unconditionally. Extended the existing detection test with a `.wav` + `displayKind:'image'` case.

src/vs/platform/agentHost/node/copilot/copilotAgentSession.ts
```diff
@@ -865,10 +884,71 @@ export class CopilotAgentSession extends Disposable {
...
 		}
+		if (attachment.displayKind !== 'directory') {
+			const mimeType = this._getImageMimeType(attachment.displayKind, uri.path);
+			if (mimeType) {
+				const blob = await this._tryReadAsBlob(uri, mimeType, displayName);
+				if (blob) {
+					return blob;
+				}
+			}
+		}
```
connor4312: Can we be more specific in this code path than 'any non-directory'?

dmitrivMS: Yes, I was going to make exactly this fix next :-)

src/vs/platform/agentHost/node/copilot/copilotAgentSession.ts
```diff
@@ -865,10 +884,71 @@ export class CopilotAgentSession extends Disposable {
...
 		const type = attachment.displayKind === 'directory' ? 'directory' : 'file';
 		return { type, path, displayName };
 	}
 
+	/**
+	 * Returns an image MIME type when the attachment is known (or can be detected from its path) to be an image. Returns
+	 * `undefined` otherwise, so non-image attachments keep the reference-style `file` path through the SDK.
+	 */
+	private _getImageMimeType(displayKind: string | undefined, path: string): string | undefined {
+		const mime = getMediaMime(path);
```
connor4312: We should actually just be using the `attachment.contentType` when set rather than trying to re-derive it from the path

src/vs/platform/agentHost/node/copilot/copilotAgentSession.ts
```diff
@@ -944,10 +964,73 @@ export class CopilotAgentSession extends Disposable {
...
 
+	/**
+	 * Returns an image MIME type for an attachment, or `undefined` for non-images so they keep the reference-style
+	 * `file` path. Prefers `contentType` (authoritative), then path-derived MIME, then the `'image'` displayKind hint
+	 * (falls back to `image/png`). Never propagates a non-image MIME under an `'image'` hint.
+	 */
+	private _getImageMimeType(attachment: MessageResourceAttachment, path: string): string | undefined {
+		if (attachment.contentType?.startsWith('image/')) {
+			return attachment.contentType;
+		}
```
connor4312: If this is set, we should always use it and not try to re-guess. If it's not an image we should immediately return as undefined

src/vs/platform/agentHost/node/copilot/copilotAgentSession.ts
```diff
@@ -944,10 +964,73 @@ export class CopilotAgentSession extends Disposable {
 		if (attachment.displayKind === 'selection') {
 			return { type: 'file' as const, path, displayName };
 		}
+		if (attachment.displayKind === 'image' || attachment.displayKind === 'document' || attachment.displayKind === undefined) {
+			const mimeType = this._getImageMimeType(attachment, uri.path);
+			if (mimeType) {
+				const blob = await this._tryReadAsBlob(uri, mimeType, displayName);
```
connor4312: We should make sure these don't appear as inline in the transcript we load from the SDK. If they are inline, we should do some processing so that they are not embedded in the session state when we restore a session, since that could quickly cause performance issues

TylerLeonhardt (approved):

PR #320524, opened 2026-06-09
fix: Deleting files with vscode fs api can cause contributed custom text e...

### LINKED ISSUE #278883: Deleting files with vscode fs api can cause contributed custom text editors to not be able to open subsequent files with the same name

arnavnagzirkar:
## Summary

When VS Code restores a session that had a custom text editor open for a file that was subsequently deleted via the VS Code fs API, the following happens: 1. `getOrCreateCustomEditorModel` calls `CustomEditorModelManager.add(resource, viewType, modelPromise)` where `modelPromise` will eventually reject (because...

## Root cause

When VS Code restores a session that had a custom text editor open for a file that was subsequently deleted via the VS Code fs API, the following happens:

1. `getOrCreateCustomEditorModel` calls `CustomEditorModelManager.add(resource, viewType, modelPromise)` where `modelPromise` will eventually reject (because `createModelReference` fails with `FILE_NOT_FOUND` since the file no longer exists).
2. Inside `add`, `tryRetain` is called which increments the reference counter to 1 and returns `entry.model.then(model => { return { object: model, dispose: ... }; })`.
3. Since `entry.model` is a rejected Promise, the `.then()` callback never runs - the `dispose` function is never created and the counter is never decremented.
4. The stale entry stays in `_references` permanently with `counter = 1` pointing to the rejected Promise.
5. When the user later creates a new file with the same name and tries to open it with the custom editor:
 - `tryRetain` finds the stale entry, increments counter to 2, and returns the old rejected Promise.
 - `getOrCreateCustomEditorModel` fails (returning a rejected Promise from the old entry rather than creating a fresh model for the newly-created file).
 - The user sees a "file not found" error even though the file now exists.

## What changed

**File:** `src/vs/workbench/contrib/customEditor/common/customEditorModelManager.ts`

Added a `.catch` handler on the `model` Promise inside `add()` that removes the stale entry from `_references` when the model fails to load. The check `this._references.get(key)?.model === model` guards against accidentally removing a newer entry that may have been added after the failed one was cleaned up.

```typescript
// If the model promise rejects, clean up the stale entry so that a new
// model can be created for the same resource later (e.g. after the file
// is recreated following a deletion).
model.catch(() => {
 if (this._references.get(key)?.model === model) {
 this._references.delete(key);
 }
});
```

## Issue

Fixes #278883

Issue: https://github.com/microsoft/vscode/issues/278883

## Diffstat

```
.../common/customEditorModelManager.ts | 10 +++
 .../test/common/customEditorModelManager.test.ts | 99 ++++++++++++++++++++++
 2 files changed, 109 insertions(+)
```

## Testing

- Ran the relevant tests and linter for the changed files while developing.

- Kept the change minimal and focused on this one issue.

## AI assistance

I used GitHub Copilot to help write parts of this change. I've reviewed and tested it myself, I understand what it does, and I'll follow up on any review feedback.


dmitrivMS:
The issue has been fixed in https://github.com/microsoft/vscode/pull/287966

PR #320624, opened 2026-06-09
Preserve image source provenance across chat reference API boundary

merfanian:
Image attachments lost their source provenance (clipboard, screenshot, file, url) when converted to the ChatPromptReference API type, because ChatReferenceBinaryData did not carry the isPasted/isURL flags. This caused image-source telemetry to collapse to only 'url' and 'unknown'.

Add optional isPasted/isURL to ChatReferenceBinaryData and populate them in the ChatPromptReference.to converter so consumers can recover the original image source.


federicobrancasi (approved):

alexr00 (approved):

merfanian:

Issue #320660, opened 2026-06-09
Block auto-approval of file-write redirection to ~ and % paths


meganrogge:
Extends the file-write redirection guard to also block auto-approval when the destination path contains `~` (POSIX tilde expansion) or `%` (Windows env-var expansion), since neither `posix.isAbsolute` nor `win32.isAbsolute` classifies these as absolute paths.

Adds regression tests for both bash (`~/file.txt`, `%HOME%/file.txt`) and pwsh (`~\file.txt`, `%APPDATA%\file.txt`).

Tracking internal PR: devdiv-microsoft/vscode-private#43

omg775:
Hi, I’d like to work on this issue.

PR #320664, opened 2026-06-09
Block auto-approved file-write redirection to env-var (%) paths

meganrogge:
The terminal command auto-approval logic guards against file-write redirection to non-absolute destinations. The guard already blocks POSIX tilde/home expansion (`~`, added in #316311), but it did not account for Windows environment-variable expansion (e.g. `%APPDATA%\file.txt`, `%HOME%/file.txt`). Such paths are not classified as absolute by `posix.isAbsolute` or `win32.isAbsolute`, so a redirect to one of them could be auto-approved and write to a sensitive, user-controlled location outside the expected workspace scope without explicit user confirmation.

### Change
- Extends the redirection guard regex to also block `%` (Windows environment-variable expansion) in addition to the existing `~` handling.
- Adds regression tests for both bash (`~/file.txt`, `%HOME%/file.txt`) and pwsh (`~\file.txt`, `%APPDATA%\file.txt`).

### Notes
- On `main` the `~` case was already covered by #316311; the net-new protection here is the `%` case. The resulting guard covers both.

Tracking: #320660


roblourens (dismissed):

meganrogge:

justschen (approved):

PR #320683, opened 2026-06-09
Add chat quota trajectory nudge

rfeltis:
## Summary
- add a Copilot Chat quota trajectory nudge behind the `chatQuotaTrajectoryNudge` treatment
- show the nudge only after quota/window/model eligibility is met and TAS assigns treatment, so users who can never see it are not exposed
- move paid self-serve SKU eligibility to ExP targeting via `copilotsku` instead of enforcing a Pro/Pro+/Max allowlist in product code
- render the nudge as a single `Learn about optimizing usage` link that opens https://aka.ms/token-usage-tips
- show the nudge at most once per billing period, and let users dismiss it
- add telemetry and tests for enrollment, learn-more clicks, gating, dismissal, usage windows, and one-display-per-billing-period behavior
- update chat input notification links so visited links keep VS Code themed link colors

## Screenshots

Treatment arm — the nudge is shown with the `Learn about optimizing usage` link:

Quota trajectory nudge shown in the treatment arm with the learn-more link

Treatment arm — `chatQuotaTrajectoryNudgeEnrolled` is emitted with `treatment:1` in the live Telemetry output (real TAS assignment, `chatquotatrajectorynudge-enabled:-9`):

Telemetry output showing chatQuotaTrajectoryNudgeEnrolled with treatment 1 in the treatment arm

Treatment arm — after clicking the link, `chatQuotaTrajectoryNudgeLinkClicked` is emitted (the nudge dismisses on click):

Telemetry output showing chatQuotaTrajectoryNudgeLinkClicked after clicking the learn-more link

Control arm — no nudge is shown:

Control arm with no nudge shown

Control arm — `chatQuotaTrajectoryNudgeEnrolled` is emitted with `treatment:0` (real TAS assignment, `chatquotatrajectorynudge-control:-9`):

Telemetry output showing chatQuotaTrajectoryNudgeEnrolled with treatment 0 in the control arm

## Experiment targeting
- Feature: `Chat Quota Trajectory Nudge` (`config.chatQuotaTrajectoryNudge`)
- SKU targeting lives in ExP on the Insiders and Public experiments with `copilotsku In ("monthly_subscriber_quota","yearly_subscriber_quota","plus_monthly_subscriber_quota","plus_yearly_subscriber_quota","max_monthly_subscriber_quota")`
- Product code resolves the treatment through the workbench assignment service (`IWorkbenchAssignmentService.getTreatment`) and still gates on quota state, billing-window projection, Copilot-model selection, and the TAS treatment result
- The treatment is requested only once every non-experiment condition is met, and enrollment telemetry is emitted only when the user is actually assigned to a flight, so unassigned users are never counted

## Telemetry
- `chatQuotaTrajectoryNudgeEnrolled` — emitted for both arms once the gates pass and TAS resolves; carries `treatment` (true/false), `entitlement`, `averageDailyUsage`, and `percentUsed` to measure exposure
- `chatQuotaTrajectoryNudgeLinkClicked` — emitted when the learn-more link is clicked (treatment arm)

## Validation
- `npm run compile`
- `npm run transpile-client`
- `scripts\\test.bat --grep ChatQuotaNotification`
- End-to-end validation used **real** ExP assignment (real TAS treatment via per-arm `devDeviceId` overrides); only quota/model eligibility was faked with generated-output-only patches. Screenshots, Playwright traces, and screen recordings were captured from the full native 1600×1000 Code OSS window with the right-side Chat bar visible/asserted for both arms, and each asserted telemetry event was verified visible in the live Telemetry output stream (`Developer: Show Telemetry`).
- Control arm (`chatquotatrajectorynudge-control:-9`, devDeviceId `728565ad-7e67-419e-8d48-a642f7e22480`): no nudge shown; `chatQuotaTrajectoryNudgeEnrolled` emitted with `treatment:0`.
- Treatment arm (`chatquotatrajectorynudge-enabled:-9`, devDeviceId `1a69d410-8d55-4e7e-b072-f6b2de8f3e58`): nudge shown; `chatQuotaTrajectoryNudgeEnrolled` emitted with `treatment:1`, and `chatQuotaTrajectoryNudgeLinkClicked` emitted on click.


rfeltis:

src/vs/workbench/contrib/chat/browser/widget/input/chatInputNotificationWidget.ts
```diff
@@ -170,7 +170,7 @@ export class ChatInputNotificationWidget extends Disposable {
 							buttonSecondaryBorder: undefined,
 						} : {}),
 						supportIcons: true,
-						secondary: !isLast,
+						secondary: isSecondary,
```
rfeltis: Are these changes necessary at all?

src/vs/workbench/contrib/chat/browser/widget/input/media/chatInputNotificationWidget.css
```diff
@@ -117,18 +124,18 @@
...
 }
 
 /* Description */
 .chat-input-notification .chat-input-notification-description {
-	font-size: var(--vscode-agents-fontSize-label1);
+	font-size: 12px;
 	line-height: 18px;
 	color: var(--vscode-descriptionForeground);
-	flex: 1 1 auto;
+	flex: 1 1 160px;
```
rfeltis: This improved responsiveness when we had a CTA and can be rolled back

rfeltis: Same with the flex-wrap in this file. We still need the :visited styles though

src/vs/workbench/contrib/chat/browser/widget/input/chatInputNotificationWidget.ts
rfeltis: For anything in this file

src/vs/workbench/contrib/chat/browser/chatQuotaNotification.ts
```diff
@@ -45,19 +78,37 @@ export class ChatQuotaNotificationContribution extends Disposable implements IWo
...
 		this._register(this._chatEntitlementService.onDidChangeQuotaRemaining(() => this._update()));
 		this._register(this._chatEntitlementService.onDidChangeQuotaExceeded(() => this._update()));
 		this._register(this._chatEntitlementService.onDidChangeEntitlement(() => this._update()));
+		this._register(this._assignmentService.onDidRefetchAssignments(() => { void this._updateTrajectoryTreatment(); }));
+		this._register(this._chatInputNotificationService.onDidDismiss(id => {
+			if (id !== QUOTA_NOTIFICATION_ID) {
+				return;
+			}
+			if (this._activeTrajectoryWarning) {
+				this._logQuotaTrajectoryNudgeClosed(this._activeTrajectoryWarning);
```
rfeltis: Make sure notification patterns match what is coming in this PR: https://github.com/microsoft/vscode/pull/321793. Do not diverge from that shape

src/vs/workbench/contrib/chat/browser/chatQuotaNotification.ts
```diff
@@ -45,19 +90,31 @@ export class ChatQuotaNotificationContribution extends Disposable implements IWo
...
 		@IStorageService private readonly _storageService: IStorageService,
+		@IWorkbenchAssignmentService private readonly _assignmentService: IWorkbenchAssignmentService,
+		@ITelemetryService private readonly _telemetryService: ITelemetryService,
 	) {
 		super();
 
 		this._register(this._chatEntitlementService.onDidChangeQuotaRemaining(() => this._update()));
 		this._register(this._chatEntitlementService.onDidChangeQuotaExceeded(() => this._update()));
 		this._register(this._chatEntitlementService.onDidChangeEntitlement(() => this._update()));
+		this._register(this._chatInputNotificationService.onDidDismiss(id => {
```
rfeltis: I think the Widget handles this for us now, so we don't need it anymore. Verify and remove if it isn't needed.

src/vs/workbench/contrib/chat/browser/chatQuotaNotification.ts
```diff
@@ -16,6 +24,43 @@ import { ChatInputNotificationSeverity, IChatInputNotification, IChatInputNotifi
 
 const QUOTA_NOTIFICATION_ID = 'copilot.quotaStatus';
 const THRESHOLDS = [50, 75, 90, 95];
+const TRAJECTORY_DAILY_USAGE_THRESHOLD = 4.5;
+const TRAJECTORY_MINIMUM_PERCENT_USED = 10;
```
rfeltis: Let's compact this into some kind of spec object so the configs aren't all splatted together

src/vs/workbench/contrib/chat/browser/chatQuotaNotification.ts
```diff
@@ -45,19 +90,31 @@ export class ChatQuotaNotificationContribution extends Disposable implements IWo
...
 		this._register(this._chatEntitlementService.onDidChangeQuotaRemaining(() => this._update()));
 		this._register(this._chatEntitlementService.onDidChangeQuotaExceeded(() => this._update()));
 		this._register(this._chatEntitlementService.onDidChangeEntitlement(() => this._update()));
+		this._register(this._chatInputNotificationService.onDidDismiss(id => {
+			if (id !== QUOTA_NOTIFICATION_ID) {
+				return;
+			}
+			this._activeTrajectoryTelemetryData = undefined;
+		}));
+		this._register(CommandsRegistry.registerCommand(CREDIT_EFFICIENCY_LEARN_MORE_COMMAND_ID, (accessor: ServicesAccessor) => this._handleCreditEfficiencyLearnMoreCommand(accessor)));
```
rfeltis: Is this the only way to wire up commands or is there a click handler in createMarkdownCommandLink?

src/vs/workbench/contrib/chat/browser/chatQuotaNotification.ts
rfeltis: Verified and removed in 4cbd84ff0c0. The widget already handles user dismissal for the active banner, and the trajectory link path clears `_activeTrajectoryTelemetryData` through `_hideNotification()`, so the extra `onDidDismiss` listener was redundant.

src/vs/workbench/contrib/chat/browser/chatQuotaNotification.ts
rfeltis: Done in 4cbd84ff0c0. I grouped the trajectory treatment key, storage key, thresholds, billing-period constants, and Learn More command/url into `TRAJECTORY_NUDGE_SPEC` so the nudge config is compact and not splatted across separate constants.

src/vs/workbench/contrib/chat/browser/chatQuotaNotification.ts
rfeltis: I checked `createMarkdownCommandLink`: it only builds a trusted markdown command URI/link target; it does not provide a per-link click handler. For this nudge, a command handler is still the right wiring because the click needs to log `chatQuotaTrajectoryNudgeLinkClicked`, dismiss the banner, and then open the external Learn More URL.

src/vs/workbench/contrib/chat/browser/chatQuotaNotification.ts
```diff
@@ -90,6 +134,27 @@ export class ChatQuotaNotificationContribution extends Disposable implements IWo
...
+	 * enrollment that may not exist. Enrollment telemetry is emitted only when
+	 * the user is actually assigned to a flight.
+	 */
+	private async _resolveTrajectoryTreatment(warning: { averageDailyUsage: number; percentUsed: number }): Promise<void> {
+		const treatment = await this._assignmentService.getTreatment<boolean>(TRAJECTORY_NUDGE_SPEC.treatment);
+		this._trajectoryTreatment = treatment;
+		if (treatment !== undefined) {
+			this._logQuotaTrajectoryNudgeEnrolled(treatment, warning);
+		}
+		this._update();
```
pwang347: is the update needed here all the time? or only when `treatment` is true?

src/vs/workbench/contrib/chat/browser/chatQuotaNotification.ts
```diff
@@ -217,6 +288,89 @@ export class ChatQuotaNotificationContribution extends Disposable implements IWo
...
+		if (percentUsed < TRAJECTORY_NUDGE_SPEC.minimumPercentUsed || percentUsed > TRAJECTORY_NUDGE_SPEC.maximumPercentUsed) {
+			return undefined;
+		}
+
+		const averageDailyUsage = percentUsed / elapsedDays;
+		if (averageDailyUsage < TRAJECTORY_NUDGE_SPEC.averageDailyUsageThreshold) {
+			return undefined;
+		}
+
+		// Every render condition is met. Enroll in the experiment now
```
pwang347: Just curious: for experiments, it's possible to just make it automatically assigned for all users and all you need is to get a boolean value that at runtime will have the resolved value. I wonder if we need manual assignment at all here (ie., how important is it that the control vs. treatment are assigned only to relevant users here? could we not just assign the entire body of users to the experiment as control/treatment and still perform analysis on the different groups?)

src/vs/workbench/contrib/chat/browser/chatQuotaNotification.ts
```diff
@@ -217,6 +288,89 @@ export class ChatQuotaNotificationContribution extends Disposable implements IWo
...
+		const learnMoreLink = createMarkdownCommandLink({
+			text: localize('quota.trajectory.learnMore', "Learn more about managing credits"),
+			id: TRAJECTORY_NUDGE_SPEC.learnMoreCommandId,
+			tooltip: localize('quota.trajectory.learnMoreTooltip', "Learn more about managing credits"),
+		});
+
+		this._setNotification({
+			id: QUOTA_NOTIFICATION_ID,
+			severity: ChatInputNotificationSeverity.Info,
+			message: new MarkdownString(localize({ key: 'quota.trajectory.message', comment: ['{Locked="]({0})"}'] }, "Based on recent usage, your monthly allowance may run out before it resets. {0}", learnMoreLink), { isTrusted: { enabledCommands: [TRAJECTORY_NUDGE_SPEC.learnMoreCommandId] } }),
```
pwang347: nit: I think usually `'{Locked="["}'` is added as well to the localization comment to preserve the full link

src/vs/workbench/contrib/chat/browser/chatQuotaNotification.ts
rfeltis: You don't really want to enroll until the point of exposure. If you eagerly enroll at any point, you will invariably end up with imbalanced cohorts. If you take it to an extreme, if the randomization just happened to put 90% of the people who actually hit the code path into variation and 10% into control, it wouldn't be an apples to apples comparison.

src/vs/workbench/contrib/chat/test/browser/chatQuotaNotification.test.ts
```diff
@@ -151,15 +203,56 @@ function makeRateLimitSnapshot(percentRemaining: number, opts?: Partial<IRateLim
 	};
 }
 
+async function runCreditEfficiencyLearnMoreCommand(): Promise<URI[]> {
```
pwang347: nit: should we test this at all? seems like the implementation might drift, so we probably don't need to maintain this mock

src/vs/workbench/contrib/chat/browser/chatQuotaNotification.ts
pwang347: I guess my question is whether the balance is actually needed here, or we can account for the imbalance in the querying itself (assuming sample sizes are large enough for meaningful analysis). Ie., what's the end goal with how we'll perform the analysis? I think with this setup the scorecards will probably look nicer, but if that doesn't actually show the metrics we care about I wonder if this is needed at all

src/vs/workbench/contrib/chat/browser/chatQuotaNotification.ts
rfeltis: Good call. Will fix.

src/vs/workbench/contrib/chat/browser/chatQuotaNotification.ts
rfeltis: Detecting imbalance is a pretty important part of running an experiment, as it often means that there is something wrong with how you are exposing users, which could invalidate your data and lead to a restart. It's common to have to do some fixing about things being missed in a post-analysis, but it's best to try to keep the experiment as clean as possible.

What would we gain from a pre-enrollment? I know there is some perf improvement that we can get, but I've heard the P90 for TAS is in the 70ms range. For something that loads in async anyway, I think that is a tolerable tradeoff.

pwang347: The main gain here IMO is just to have less code complexity if we can avoid it (the static exp registration flow is heavily used/tested in product). Not a blocker for me though, I can defer to @lramos15 when he reviews!

rfeltis: Happy to blend in where the path is already paved. Just let me know.

cwebster-99: We typically don't codify experiment conditions like this. We can use the experiment platform to target specific users for conditions like SKU and then we typically use triggered scorecards to filter the scorecards for only users that would be "eligible" to see the treatment. If there is a telemetry event that indicates a state in which the user would then see the treatment that would be the best way for use to adhere to typical experiment setups.

rfeltis: @cwebster-99 can we pair on this tomorrow? Would love to see that ExP side of things.

rfeltis: Chatted with @cwebster-99 and we agreed that we should match the current user assignment approach for this experiment and spar for future ones. @pwang347 can you point me in the direction of how to do the pre-assignment?

rfeltis: Switched to extensions/copilot/src/platform/configuration/common/configurationService.ts. Thanks @pwang347 !

rfeltis: My validations really struggled once I switched how the experiment is evaluated. Copilot found a timing issue that made the tests stable. I had it try with and without the patches a few times and it was a marked difference.

I'll have it add a big note about it in the description for someone who knows the architecture better than me to scrutinize.

extensions/copilot/src/platform/configuration/common/configurationService.ts
```diff
@@ -652,6 +652,7 @@ export namespace ConfigKey {
 		export const WorkspacePreferredEmbeddingsModel = defineAndMigrateExpSetting<string>('chat.advanced.workspace.preferredEmbeddingsModel', 'chat.workspace.preferredEmbeddingsModel', '');
 		export const NotebookAlternativeDocumentFormat = defineAndMigrateExpSetting<AlternativeNotebookFormat>('chat.advanced.notebook.alternativeFormat', 'chat.notebook.alternativeFormat', AlternativeNotebookFormat.xml);
 		export const UseAlternativeNESNotebookFormat = defineAndMigrateExpSetting<boolean>('chat.advanced.notebook.alternativeNESFormat.enabled', 'chat.notebook.alternativeNESFormat.enabled', false);
+		export const ChatQuotaTrajectoryNudge = defineSetting<boolean>('chat.quotaTrajectoryNudge.enabled', ConfigType.ExperimentBased, false, undefined, undefined, { experimentName: 'config.chatQuotaTrajectoryNudge' });
```
pwang347: is there a reason why we need the explicit `experimentName` here? I think we just need to map it on the exp platform?

extensions/copilot/src/extension/contextKeys/common/chatQuotaTrajectoryNudgeCommand.ts
```diff
@@ -0,0 +1,25 @@
...
+
+export const getChatQuotaTrajectoryNudgeEnabledCommand = '_github.copilot.chat.getQuotaTrajectoryNudgeEnabled';
+
+interface ICommandRegistrar {
+	registerCommand(command: string, callback: (...args: unknown[]) => unknown): { dispose(): void };
+}
+
+export function registerChatQuotaTrajectoryNudgeEnabledCommand(commandRegistrar: ICommandRegistrar, configService: IConfigurationService, expService: IExperimentationService) {
+	return commandRegistrar.registerCommand(getChatQuotaTrajectoryNudgeEnabledCommand, async () => {
+		await expService.hasTreatments();
```
pwang347: I think we can simplify the exp value to just this:

```
const enabled = configService.getExperimentBasedConfig(ConfigKey.Advanced.ChatQuotaTrajectoryNudge, expService);
```

then you don't need to care about the treatment vs. control (managed in exp), you just get the boolean value.

pwang347: actually, any idea why this needs to be a command? it seems like we're just trying to get the enablement value? (in which case the above can just be used at a call site?)

extensions/copilot/src/extension/extension/vscode/extension.ts
```diff
@@ -64,6 +66,8 @@ export async function baseActivate(configuration: IExtensionActivationConfigurat
 
 	await instantiationService.invokeFunction(async accessor => {
 		const expService = accessor.get(IExperimentationService);
+		const configService = accessor.get(IConfigurationService);
```
pwang347: same as above, I don't think we need the command

src/vs/workbench/contrib/chat/browser/widget/input/chatInputPart.ts
```diff
@@ -993,15 +993,18 @@ export class ChatInputPart extends Disposable implements IHistoryNavigationWidge
...
-		const model = models.find(m => m.identifier === identifier);
+		const model = this.getModels().find(m => m.identifier === identifier);
 		if (model) {
 			this.setCurrentLanguageModel(model);
 			return true;
 		}
 		return false;
 	}
 
+	public canSwitchModelByIdentifier(identifier: string): boolean {
```
pwang347: is there a reason why we need a change for this file?

rfeltis: Potentially, the auto model could disappear, so we wouldn't want to show them an option to switch to it if that happened. Super edge case.

extensions/copilot/src/extension/contextKeys/common/chatQuotaTrajectoryNudgeCommand.ts
rfeltis: You're likely right about it. Let me see if Copilot can simplify.

pwang347: ah I understand why we're using a command now, it's because we're trying to communicate from extension -> core (where the UI lives). I think the better idea is probably to define the exp setting in core directly

rfeltis: Got it. I still need to understand the architecture better so I can help Copilot navigate. Apologies for my fumbling around while I learn.

pwang347: @rfeltis sorry I think we don't need a setting at all, we should be able to use something like this:
				const autoModeHint = this._expService.getTreatmentVariable<string>('copilotchat.autoModelHint');

pwang347: no worries at all! apologies as well for not pointing you to better examples

extensions/copilot/src/extension/test/node/configurations.spec.ts
```diff
@@ -72,10 +72,11 @@ describe('Configurations', () => {
 		// Get keys from code
 		const internalKeys = Object.values(ConfigKey.TeamInternal).map(setting => setting.fullyQualifiedId);
 		const sharedKeys = Object.values(ConfigKey.Shared).map(setting => setting.fullyQualifiedId);
-		const advancedPublicKeys = Object.values(ConfigKey.Advanced).map(setting => setting.fullyQualifiedId);
+		const advancedPublicKeys = Object.values(ConfigKey.Advanced).filter(setting => setting.isPublic).map(setting => setting.fullyQualifiedId);
+		const advancedHiddenKeys = Object.values(ConfigKey.Advanced).filter(setting => !setting.isPublic).map(setting => setting.fullyQualifiedId);
 		const otherPublicKeys = (Object.values(ConfigKey).filter(key => key !== ConfigKey.TeamInternal && key !== ConfigKey.Shared && key !== ConfigKey.Advanced && key !== ConfigKey.Deprecated) as Config<any>[]).map(setting => setting.fullyQualifiedId);
 		const registered = [...otherPublicKeys, ...advancedPublicKeys];
-		const unregistered = [...internalKeys, ...sharedKeys];
+		const unregistered = [...internalKeys, ...sharedKeys, ...advancedHiddenKeys];
```
pwang347: I think using settings was not the right approach, we should be able to revert this file

extensions/copilot/src/platform/configuration/common/configurationService.ts
pwang347: see comment above, I don't think this is needed - just pass a string to getTreatment

src/vs/workbench/contrib/chat/browser/widget/input/chatInputNotificationService.ts
```diff
@@ -33,6 +33,13 @@ export interface IChatInputNotificationMuteAction {
 
 export interface IChatInputNotification {
 	readonly id: string;
+	/**
```
pwang347: nit: looks like we don't need this comment

src/vs/workbench/contrib/chat/browser/widget/input/chatInputPart.ts
pwang347: hmm, I probably wouldn't try to handle that here - the auto model is not supposed to disappear (so that might be a separate bug) but there are some cases like Claude / Codex in Agent Host where it would not be present at all

CC @lramos15

src/vs/workbench/contrib/chat/browser/chatQuotaNotification.ts
```diff
@@ -90,6 +145,36 @@ export class ChatQuotaNotificationContribution extends Disposable implements IWo
...
+	 * assignment into a synthetic "control" value, since that would assume an
+	 * enrollment that may not exist. Enrollment telemetry is emitted only when
+	 * the user is actually assigned to a flight.
+	 */
+	private async _resolveTrajectoryTreatment(warning: { averageDailyUsage: number; percentUsed: number }): Promise<void> {
+		let treatment: boolean | undefined;
+		try {
+			treatment = await this._assignmentService.getTreatment<boolean>(TRAJECTORY_NUDGE_SPEC.treatmentName);
+		} catch (error) {
+			console.error(`Failed to resolve ${TRAJECTORY_NUDGE_SPEC.treatmentName}`, error);
```
pwang347: nit: instead of console.error we should probably flow through the logger service

src/vs/workbench/contrib/chat/browser/widget/input/chatInputPart.ts
rfeltis: Do you know the better way to handle it? I didn't run into it as a bug. But the experience does change if auto is available/selected or not.

src/vs/workbench/contrib/chat/browser/chatQuotaNotification.ts
```diff
@@ -217,6 +308,113 @@ export class ChatQuotaNotificationContribution extends Disposable implements IWo
...
+		if (averageDailyUsage < TRAJECTORY_NUDGE_SPEC.averageDailyUsageThreshold) {
+			return undefined;
+		}
+
+		// Every render condition is met. Enroll in the experiment now
+		// (just-in-time) so treatment and control cohorts are assigned at the
+		// same point and users who would never see the nudge are not exposed.
+		// While assignment is pending, fall through; _update re-runs once it
+		// resolves and only the treatment cohort then renders the nudge.
+		if (!this._trajectoryAssignmentRequested) {
```
pwang347: is comment outdated here? also do we need to reassign the value here, why not just reuse the value we get from exp?

src/vs/workbench/contrib/chat/browser/chatQuotaNotification.ts
```diff
@@ -399,6 +597,37 @@ export class ChatQuotaNotificationContribution extends Disposable implements IWo
...
+	}
+
+	private _getAutoModelIdentifier(): string | undefined {
+		return this._languageModelsService.getLanguageModelIds().find(identifier => {
+			const metadata = this._languageModelsService.lookupLanguageModel(identifier);
+			return !!metadata && this._isAutoModel(metadata.id, metadata.vendor);
+		});
+	}
+
+	private _isAutoModel(id: string, vendor: string): boolean {
```
pwang347: nit: I feel like this logic probably should live in a more generic place

pwang347: @lramos15 for guidance on how to best detect "auto", the check below will not work for agent host harnesses

src/vs/workbench/contrib/chat/browser/chatQuotaNotification.ts
```diff
@@ -399,6 +597,37 @@ export class ChatQuotaNotificationContribution extends Disposable implements IWo
 		return isSelectedModelCopilot(this._contextKeyService, this._storageService, this._languageModelsService);
 	}
 
+	private _getLinkToAutoDecision(): 'enabled' | 'alreadyAuto' | 'autoIneligible' {
+		if (this._isAutoModelSelected()) {
+			return 'alreadyAuto';
+		}
+		const autoModelIdentifier = this._getAutoModelIdentifier();
+		if (!autoModelIdentifier || this._chatWidgetService.lastFocusedWidget?.input.canSwitchModelByIdentifier(autoModelIdentifier) !== true) {
```
pwang347: there should be a property called `supportsAutoModel` that each harness contributes that could be useful here, not sure if there's an easy way to hook it up

src/vs/workbench/contrib/chat/browser/widget/input/chatInputPart.ts
pwang347: commented below, I think each harness contributes "supportsAutoModel" which you can check if it's not supported

src/vs/workbench/contrib/chat/browser/chatQuotaNotification.ts
rfeltis: Yeah, comment is outdated. The flow is still the same because _resolveTrajectoryTreatment is async, so we have to do this dance a little bit. Let me update it to reflect that better.

src/vs/workbench/contrib/chat/browser/chatQuotaNotification.ts
rfeltis: @lramos15 states that auto should always be there. I'll remove the checks.

src/vs/workbench/contrib/chat/browser/chatQuotaNotification.ts
rfeltis: @pwang347 thoughts about how auto is specified now?

rfeltis: Descoping auto to make this week's stable release

pwang347 (approved):

roblourens (approved):

dmitrivMS (approved):

PR #320685, opened 2026-06-09
Improve the local to native and remote to local copy, paste, and DND experience

### LINKED ISSUE #2008: Cannot copy files from remote SSH workspace to local workspace
- VSCode Version: 1.41.0-insider (commit: 9579eda04fdb3a9bba2750f15193e5fafe16b959)
- OS Version: macOS 10.15.1 Catalina

Steps to Reproduce:

1. Open a remote SSH workspace
2. Open a local workspace (in a new window)
3. Select a file on the **local workspace** and "copy" it (either with contextual menu, or `ctrl+C`)
4. Paste the file on the **remote workspace** (either with contextual menu, or `ctrl+V`)
5. **Result**: The file is copied correctly
6. **Now the reverse**: Select a file on the **remote workspace** and "copy" it (either with contextual menu, or `ctrl+C`)
7. Paste the file on the **local workspace** (either with contextual menu, or `ctrl+V`)
8. **Result**: The file is not copied, and an error message is shown: `The file to paste has been deleted or moved since you copied it. No file system provider found for vscode-remote://ssh-remote...`
9. Also, pasting the remote file into `Finder` yields an empty (zero byte) file (with the correct name)

lramos15:
Fix https://github.com/microsoft/vscode-remote-release/issues/2008

### DIFF
--- a/src/vs/workbench/contrib/files/browser/explorerService.ts
+++ b/src/vs/workbench/contrib/files/browser/explorerService.ts
@@ -10,7 +10,7 @@ import { IFilesConfiguration, ISortOrderConfiguration, SortOrder, LexicographicO
 import { ExplorerItem, ExplorerModel } from '../common/explorerModel.js';
 import { URI } from '../../../../base/common/uri.js';
 import { FileOperationEvent, FileOperation, IFileService, FileChangesEvent, FileChangeType, IResolveFileOptions } from '../../../../platform/files/common/files.js';
-import { dirname, basename } from '../../../../base/common/resources.js';
+import { dirname, basename, joinPath } from '../../../../base/common/resources.js';
 import { IConfigurationService, IConfigurationChangeEvent } from '../../../../platform/configuration/common/configuration.js';
 import { IClipboardService } from '../../../../platform/clipboard/common/clipboardService.js';
 import { IEditorService } from '../../../services/editor/common/editorService.js';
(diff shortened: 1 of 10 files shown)


src/vs/workbench/contrib/files/browser/explorerService.ts
```diff
@@ -259,11 +266,62 @@ export class ExplorerService implements IExplorerService {
...
-		await this.clipboardService.writeResources(items.map(s => s.resource));
+
+		const resources = items.map(s => s.resource);
+
+		// For remote resources, download to a temp location so they can be
+		// pasted both in other VS Code windows and into native file managers.
+		// We replace remote URIs with local file:// URIs pointing to temp copies.
+		const clipboardResources = await this.resolveClipboardResources(resources);
+
+		await this.clipboardService.writeResources(clipboardResources);
```
mjbvz, line 277: Not sure if this idea will work but I feel like the temp file stuff isn't right. A few concerns:

- We are creating the file eagerly on copy and blocking the copy on this.
- I worry about leaking the `/tmp` file path 

Instead, the web clipboard api should support writing a Promise that resolves to a `Blob`: https://developer.mozilla.org/en-US/docs/Web/API/ClipboardItem/ClipboardItem

I believe that electron is exploring aligning their clipboard API more with this web api too: https://github.com/electron/electron/pull/51707

Maybe @deepak1556 can weight on what we can do for electron

mjbvz (approved):

PR #320694, opened 2026-06-09
Use managed settings for enterprise policy prompt injection

seanj39:


joshspicer (changes requested):
Hey there! This is duplicating some work in VS Code itself. I was not aware of this policy or work and think it would be better done in a central location within vscode, and not the responsibility of the extension.

dmitrivMS:
Per above, there is already work in this area and this is not the right direction.

PR #320701, opened 2026-06-09
Allow configuring the secondary side bar location

### LINKED ISSUE #281828: Allow Secondary Side Bar to be on the same side as primary side bar
Currently the secondary side bar is always the opposite side of the primary side bar (for e.g. if primary side bar is on the left, then the secondary bar is on the right). Some may prefer to have access to two tools on the same side of the window (for e.g. some people (like me) like keeping primary side bar on the right so that the code is on the left as much as possible, making the secondary side bar be able to be on the same side would enable them to access it while still keeping the code on the left).
This is especially useful for AI Chat like Copilot, Cline, Roo Code etc. (in my case I like seeing my file tree when using these, and it would be nice to not have to look back and forth to figure things out)

Image

We should still retain ability to enable and disable primary and secondary side bar independently (secondary side bar only with no primary side bar, etc.)

AlejandroRomeroG:
This adds a `workbench.secondarySideBar.location` setting with `opposite`, `left`, and `right` values. `opposite` remains the default, so the existing behavior is preserved unless the setting is changed. The workbench layout, secondary side bar context menu, customize layout icon, and editor drag auto-open behavior now use the resolved secondary side bar position instead of assuming it is always opposite the primary side bar.

The motivation is to support workflows where the primary and secondary side bars should stay on the same side of the workbench. This came out of an initial Positron PR, posit-dev/positron#14148, where the goal was to support an RStudio-style layout with the editor and console on the left and related workspace views grouped on the right. That layout was difficult to express while the secondary side bar was always tied to the opposite side of the primary side bar.

Validation: I ran `git diff --check`. The initial Positron version of this change was also packaged and smoke-tested in a local Positron build, including launching the packaged app, moving the secondary side bar to the same side as the primary side bar, and verifying the session, variables, plots, file explorer, and console layout behaved as expected. I also attempted to install dependencies in this VS Code checkout so I could run `npm run compile-check-ts-native`, but `npm ci` failed locally while compiling native dependencies because the macOS toolchain could not find the C++ standard library header `<unordered_set>`.

Fixes #281828.

### DIFF
--- a/src/vs/workbench/browser/actions/layoutActions.ts
+++ b/src/vs/workbench/browser/actions/layoutActions.ts
@@ -104,7 +104,7 @@ const SecondarySideBarLeftContext = ContextKeyExpr.or(
 		ContextKeyExpr.equals(`config.${secondarySideBarPositionConfigurationKey}`, SecondarySideBarLocation.OPPOSITE),
 		ContextKeyExpr.equals(`config.${sidebarPositionConfigurationKey}`, 'right')
 	)
-)!;
+) ?? ContextKeyExpr.false();
 
 const SecondarySideBarRightContext = ContextKeyExpr.or(
 	ContextKeyExpr.equals(`config.${secondarySideBarPositionConfigurationKey}`, SecondarySideBarLocation.RIGHT),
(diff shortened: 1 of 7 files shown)


AlejandroRomeroG:

PR #320738, opened 2026-06-10
Consolidate authentication provider session options

alop-hue:
refactor: extract common authentication options interface

Reduce duplication by extracting shared properties from
AuthenticationProviderSessionOptions and AuthenticationGetSessionOptions
into a new AuthenticationProviderCommonOptions interface.

This improves maintainability and ensures consistency between the two
interfaces, which currently contain identical optional properties.


alop-hue:

PR #320740, opened 2026-06-10
refactor: unify glob pattern types in search API

alop-hue:
## Summary
This PR removes the `GlobString` type alias and replaces its usage with `GlobPattern` in the text search API, implementing the existing TODO.

## Changes
- Deleted `GlobString` and its TODO.
- Changed `SearchOptions.includes` and `SearchOptions.excludes` to `GlobPattern[]`.
- Updated JSDoc to reflect the new type (supports `RelativePattern`).

## Motivation
- `GlobPattern` is already the standard glob type in VS Code; using a separate `string` alias is inconsistent and limits the API (e.g., no relative patterns).
- Resolves the long-standing TODO and reduces confusion.
- Fully backward compatible: existing code passing `string[]` works because `string` is part of `GlobPattern`.


alop-hue:

PR #320764, opened 2026-06-10
Auto: keep model cache across token refresh 

kevin-m-kent:
Gate the per-conversation Auto model-cache clear in the onDidAuthenticationChange handler behind a new experiment-based setting, chat.advanced.autoModePreserveModelOnTokenRefresh (default false).

onDidAuthenticationChange fires on the routine Copilot token refresh (same identity). Clearing _autoModelCache there makes the next turn of an in-flight Auto conversation re-resolve its model and silently switch model or provider, landing on a cold provider prompt cache. When the experiment is enabled the cache is preserved so the conversation stays pinned to its selected model; each token bank still refreshes itself on its own expiry. Reserve token-bank refresh is unchanged. No-op when the flag is false.

If this theory is correct we should see
- Overall auto cache hit rate improve, bring in line with the same models chosen in the picker
- Between model auto vs non auto differences should be more consistent (claude vs oai differences in auto vs non auto should be similar), due to compaction firing on top of cold call after switch (oai vs haiku/sonnet context size differences exacerbate the switching cost for oai --> sonnet or haiku).

### DIFF
--- a/extensions/copilot/src/platform/configuration/common/configurationService.ts
+++ b/extensions/copilot/src/platform/configuration/common/configurationService.ts
@@ -878,6 +878,11 @@ export namespace ConfigKey {
 		 * For experiments, this setting selects the routing method only when router usage is enabled;
 		 * it does not by itself determine whether the router is called. */
 		export const AutoModeRoutingMethod = defineTeamInternalSetting<string>('chat.advanced.autoModeRoutingMethod', ConfigType.ExperimentBased, '', undefined, undefined, { experimentName: 'copilotchat.autoModeRoutingMethod' });
+		/** When enabled, a periodic auth/token refresh refreshes each Auto conversation's
+		 * CAPI session token in place instead of clearing the model cache, so the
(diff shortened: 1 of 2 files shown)


kevin-m-kent:
@aashna if you have any thoughts here I'd appreciate your eyes on this! From looking at some data, it seems like model switching might be occurring more than it needs to in auto, code suggests it might be due to token refreshes - this proposed ExP + change hopefully addresses that but you are way more familiar with the code than I am.

blclo:
Thanks for the finding and fix! I'm onboard with this as a best effort. 

The only thing I want to mention is that on token refresh (before the token expires) Auto **does not run dynamic model selection**, and just returns the same `available_models` list back (if entitlements/policies have not changed). As a result, this list doesn't know of models suddenly having a high error rate. I think this is a reasonable tradeoff, since in any case we have this risk already in the whole conversation, given we run Auto once per conversation anyways.

blclo (approved):

lramos15:
Superseded by https://github.com/microsoft/vscode/pull/321211

PR #320785, opened 2026-06-10
Integrate Copilot Voice conversation engine

meganrogge:
## Summary

Ports the Copilot Voice conversation engine from the `voice-code` prototype fork into VS Code.

Fixes microsoft/vscode-internalbacklog#7956

### New modules (`voiceClient/`)
- **VoiceSessionController** — state machine, push-to-talk, session tracking
- **VoiceClientService** — WebSocket client to cloud voice backend
- **MicCaptureService** — push-to-talk mic capture with diagnostics
- **VoiceToolDispatchService** — dispatches tool calls to VS Code commands
- **TtsPlaybackService** — streaming audio playback

### New UI (`agentsVoice/` contrib)
- Floating companion window (three-tier: pill/mid/expanded)
- Components: header, voice bar, transcript, session list, confirmations
- Transcript debug views
- Embedded voice agent bar in chat view pane

### Platform changes
- `IAuxiliaryWindowOpenOptions`: add frameless, transparent, notResizable, noBackgroundThrottling, backgroundColor
- `IChatAcceptInputOptions`: add preserveFocus
- Electron-main: handle new window feature flags
- lit-html and signals-core vendor libraries (excluded from hygiene)

#### Reactive chat state monitoring for all session types
The voice-code prototype only monitored agent sessions (`agentSessionsService.model.sessions`) for state changes. Regular chat sessions were invisible to the voice system — confirmations, response completions, and question carousels on non-agent chats were never detected or forwarded to the voice backend.

This PR monitors **all** chat sessions reactively:
- Both autoruns (tool confirmations + state tracking) now iterate `chatService.chatModels.read(reader)` in addition to agent sessions, subscribing to `lastRequestObs`, `isPendingConfirmation`, `isIncomplete`, and `onDidChange` for every session with a loaded model
- `_buildSessionContext()` includes non-agent chat sessions with proper state from `_getAgentStateInfo`, so the backend tracks their transitions (confirmations, completions, question carousels)
- `_checkSessionStateChanges()` also polls regular chat sessions as a fallback
- State seeding on connect includes regular chat models to prevent false transitions
- Widget binding subscribes to per-model `hasActiveRequest` and `requestNeedsInput` observables so the voice UI updates reactively for all sessions
- Regular chat session rows now show actual `isActive`/`needsInput`/`isIdle` state (previously always hardcoded to idle)

#### Bug fixes
- **Zombie WebSocket**: Reconnect `setTimeout` return value is now tracked in `_reconnectTimer` and cleared in `_cleanup()`; `_window` is also cleared to prevent stale timers from reconnecting
- **Unhandled promise rejections**: Added `.catch()` to fire-and-forget `executeCommand` calls in `voiceSessionController` (acceptInput delegate + line 1096)
- **CSS variable mismatch**: Fixed `--vscode-voiceCode-speakingForeground` to `--vscode-agentsVoice-speakingForeground`
- **Keyboard accessibility**: Changed `@mousedown` to `@click` on all button elements
- **Window close state**: Persist `WindowOpen=false` in `onUnload` handler
- **Transcript directory**: Ensure `voiceTranscripts/` folder exists before first write
- **Voice approve/reject for regular chat sessions**: `approve_confirmation`, `reject_confirmation`, and `focus_session` tool dispatch now searches `chatService.chatModels` when the target session isn't found in agent sessions — fixes voice approval silently no-oping on non-agent chat sessions

#### Accessibility
- All interactive `<span>` elements across header, voice bar, session list, and onboarding components now have `role="button"`, `tabindex="0"`, and `aria-label`
- Session list rows use `role="option"` with `aria-selected` for proper listbox semantics
- Covers: popout, feedback, close, mic, PTT key, stop speech, open/stop session actions, new session button, onboarding PTT key

#### Lazy mic activation
Microphone is no longer acquired on connect — it's deferred to the first PTT press. This avoids switching bluetooth headsets into speech mode prematurely and prevents the backend from hearing ambient audio. The disconnect button has been removed since closing the window already disconnects.

#### Dual-mode push-to-talk
PTT now supports two modes, auto-detected by hold duration (SuperWhisper-style):
- **Hold-to-talk**: hold the key >300ms, release to finish recording
- **Tap-to-toggle**: quick tap (<300ms) starts recording, tap again to finish

Users naturally discover whichever mode feels right — no configuration needed.

#### Full localization
All user-visible strings across 8 files localized with `nls.localize()` (feedbackDialog, headerComponent, sessionListComponent, confirmationComponent, statusRowsComponent, voiceToolDispatchService, agentsVoiceWidgetBinding)

#### preserveFocus wiring
`ChatWidget.acceptInput()` now passes `preserveFocus` through to `ChatInputPart.acceptInput()`, skipping editor focus when true — used by voice to avoid stealing focus from the voice UI.

### Feature gate
All voice features are gated behind the `agents.voice.enabled` setting (defaults to **false**, hidden from Settings UI via `included: false`). The setting is marked `restricted: true` so enterprise admins can lock it via policy. Toggling dynamically shows/hides the voice panel without reload.

### Telemetry
Comprehensive voice telemetry via `voiceTelemetry.ts` covering:
- **Activation**: `voiceEnabled`, `voiceDisabled` (with days-active), `voiceFirstConnect`, `voiceOnboardingCompleted`
- **Usage**: `voiceSessionStarted`, `voiceSessionEnded` (turns, duration), `voicePtt` (hold duration)
- **Quality**: `voiceTtsListenThrough` (completed vs interrupted), `voiceToolApproval` (approved/denied), `voiceReconnect`
- **Latency**: `voiceLatency` (time-to-first-transcription, end-to-end turn)

### Design questions addressed
- **Enterprise policy** — `agents.voice.enabled` marked `restricted: true`; WebSocket URL configurable via `agents.voice.backendUrl`
- **Session targeting** — Continues existing chat session instead of always creating new; only creates new when on sessions list view
- **Confirmation announcements** — Voice announces tool confirmations and questions regardless of which session is focused; tracks detail changes within same state to catch all transitions
- **Onboarding** — Re-shown on disconnect; completion tracked via telemetry
- **Discoverability** — Setting hidden from UI (`included: false`), defaults to off; zero surface area when disabled


src/vs/base/common/lit-html/directives/repeat.d.ts
```diff
@@ -0,0 +1,64 @@
+/**
```
dmitrivMS, line 1: why inline? is there no npm package for this?

src/vs/platform/windows/electron-main/windows.ts
```diff
@@ -234,6 +239,35 @@ export function defaultBrowserWindowOptions(accessor: ServicesAccessor, windowSt
 		options.alwaysOnTop = true;
 	}
 
+	if (overrides?.frameless) {
+		options.frame = false;
+		options.titleBarStyle = undefined;
+		options.titleBarOverlay = undefined;
+		options.minWidth = undefined;
+		options.minHeight = undefined;
+		delete (options as unknown as Record<string, unknown>).titleBarStyle;
```
dmitrivMS: can't we just update the type?

src/vs/workbench/contrib/agentsVoice/browser/components/confirmationComponent.ts
```diff
@@ -0,0 +1,45 @@
...
+
+export function renderToolConfirmations(props: ToolConfirmationsProps): TemplateResult | typeof nothing {
+	if (props.confirmations.length === 0) {
+		return nothing;
+	}
+	return html`
+		<div style="display:flex;flex-direction:column;gap:4px;padding:6px 2px 2px;border-top:1px solid var(--vscode-editorWidget-background);margin-top:4px;">
+			${props.confirmations.map(tc => tc.type === 'approval' ? html`
+				<div style="display:flex;flex-direction:column;gap:3px;">
+					<span style="font-size:${FONT_SIZE.body};color:var(--vscode-editorWarning-foreground);font-weight:500;">${tc.sessionLabel}</span>
```
dmitrivMS: is there a variable for font-weight value here?

src/vs/workbench/contrib/agentsVoice/browser/components/feedbackDialog.ts
```diff
@@ -0,0 +1,124 @@
...
+
+	private async _handleSubmit(text: string): Promise<void> {
+		this._state = { isSubmitting: true, submitted: false };
+		this._render();
+
+		const result = await this._onSubmit(text);
+		if (result.ok) {
+			this._state = { isSubmitting: false, submitted: true };
+			this._render();
+			setTimeout(() => this._onClose(), 3000);
```
dmitrivMS: consider making the class disposable and using disposableTimeout

src/vs/workbench/contrib/agentsVoice/common/agentsVoiceColors.ts
```diff
@@ -0,0 +1,17 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { registerColor } from '../../../../platform/theme/common/colorRegistry.js';
+import { localize } from '../../../../nls.js';
+
+export const agentsVoiceSpeakingForeground = registerColor('agentsVoice.speakingForeground',
+	{ dark: '#a371f7', light: '#8250df', hcDark: '#d2a8ff', hcLight: '#6639ba' },
```
dmitrivMS, line 10: these should probably be themed

src/vs/base/common/lit-html/directives/repeat.d.ts
meganrogge: VS Code's workbench can't consume npm packages directly, the source is loaded via a custom module system, so external libraries need to be vendored as standalone ES modules. This is the same pattern used for other vendored dependencies in the repo (e.g. marked, vscode-textmate). These are the minified build output from the lit-html npm package (v3.3.2).

src/vs/workbench/contrib/agentsVoice/browser/agentsVoiceWindowService.ts
```diff
@@ -0,0 +1,413 @@
...
+			const currentDpr = auxiliaryWindow.window.devicePixelRatio;
+			if (Math.abs(currentDpr - lastDpr) > 0.01) {
+				lastDpr = currentDpr;
+				if (zoomDebounce) { clearTimeout(zoomDebounce); }
+				zoomDebounce = setTimeout(() => {
+					this._resizeWindow(auxiliaryWindow);
+				}, 200);
+			}
+		};
+		const zoomCheckInterval = auxiliaryWindow.window.setInterval(checkZoom, 500);
```
dmitrivMS: There is disposableWindowInterval window wraps the set/clear logic here and blow - it may be good to use that instead

src/vs/workbench/contrib/agentsVoice/browser/agentsVoiceWindowService.ts
```diff
@@ -0,0 +1,413 @@
...
+		this._windowDisposables.add({ dispose: () => { auxiliaryWindow.window.clearInterval(zoomCheckInterval); if (zoomDebounce) { clearTimeout(zoomDebounce); } } });
+
+		// Poll for session updates
+		this.agentSessionsService.model.resolve(undefined);
```
dmitrivMS, line 276: May be a follow-up PR, but I agree with AI review I sent you that this polling feels like a hack/workaround - can't we have observables/events that this would subscribe to instead?

src/vs/workbench/contrib/agentsVoice/browser/agentsVoiceWidget.ts
```diff
@@ -0,0 +1,534 @@
...
+				}
+			});
+			// Catch pointerup outside the container too (mirrors the chat view pane behavior)
+			const onDocPointerUp = () => this.callbacks.pttUp();
+			win.document.addEventListener('pointerup', onDocPointerUp);
+			this._register(toDisposable(() => win.document.removeEventListener('pointerup', onDocPointerUp)));
+		}
+
+		// Set up PTT via BroadcastChannel
+		const pttChannel = new BroadcastChannel('vscode-ptt');
```
dmitrivMS, line 176: There is concern that this is a global channel which would create race/conflict when multiple windows are present. Is it possible to make it window-scoped?

src/vs/workbench/contrib/agentsVoice/browser/agentsVoiceWindowService.ts
meganrogge: Yes I will follow-up on this one.

src/vs/workbench/contrib/chat/browser/widgetHosts/viewPane/chatViewPane.ts
```diff
@@ -336,6 +382,164 @@ export class ChatViewPane extends ViewPane implements IViewWelcomeDelegate {
...
+	private readonly _voiceBarDisposables = this._register(new DisposableStore());
+
+	private _updateVoiceBar(container: HTMLElement): void {
+		this._voiceBarDisposables.clear();
+		container.replaceChildren();
+
+		if (this.configurationService.getValue<boolean>('agents.voice.enabled')) {
+			container.style.display = '';
+
+			// Voice command bridge — lets the VoiceSessionController reach into the chat widget
```
dmitrivMS, line 398: Same issue as in agentsVoiceWidget.ts - if two chat panes race on this, and last one wins? will getCurrentSession resolve to the wrong pane now? Perhaps this can be done via a window-specific service?

src/vs/workbench/contrib/chat/browser/voiceClient/ttsPlaybackService.ts
```diff
@@ -0,0 +1,226 @@
...
+			if (gen !== this._playbackGen) { return; }
+			try {
+				const ctx = this.ensureContext(this._window!);
+				const decoded = await ctx.decodeAudioData(arrayBuf);
+				if (gen !== this._playbackGen) { return; }
+				this._writeToPlayBuffer(decoded);
+				if (!this._playbackTurn?.started) {
+					this._startPlayback();
+				}
+			} catch (err) { console.error('[voice] TTS decode error', err); }
```
dmitrivMS: Many instances of console.error/console.warn - perhaps can these be routed to log service instead?

src/vs/workbench/contrib/chat/browser/voiceClient/voiceClientService.ts
```diff
@@ -0,0 +1,569 @@
...
+	IVoiceTranscription,
+	IVoiceAudioResponse,
+	IVoiceToolCall,
+	IVoiceSpeechStarted,
+	IVoiceSessionInit,
+	IVoiceFeedbackPayload,
+} from '../../common/voiceClient/voiceClientService.js';
+import { InstantiationType, registerSingleton } from '../../../../../platform/instantiation/common/extensions.js';
+
+const DEFAULT_VOICE_WS_URL = 'wss://falcon-caas.mai.microsoft.com/voice-code/api/v1/realtime/voice';
```
dmitrivMS: Perhaps this should go into product.json?

src/vs/workbench/contrib/chat/browser/voiceClient/voiceToolDispatchService.ts
```diff
@@ -0,0 +1,451 @@
...
+	}
+
+	/** Get the action label for a tool call name. */
+	static getActionLabel(name: string): string {
+		return ACTION_LABELS[name] ?? localize('agentsVoice.action.working', "Working...");
+	}
+
+	private get _agentModeOptions(): IChatSendRequestOptions {
+		const allTools: Record<string, boolean> = {};
+		for (const tool of this.toolsService.getTools(undefined)) {
```
dmitrivMS, line 94: this enables ALL tools - may be dangerous...

src/vs/workbench/contrib/chat/browser/voiceClient/voiceSessionController.ts
```diff
@@ -0,0 +1,2097 @@
...
+						if (!stillWaiting.has(id)) {
+							this._eagerModelRefs.get(id)!.dispose();
+							this._eagerModelRefs.delete(id);
+						}
+					}
+				});
+				// Periodic fallback: check session state changes every 5s
+				// to catch transitions missed when the chat model isn't loaded
+				// (e.g. remote agent host sessions that haven't been opened).
+				const periodicCheck = mainWindow.setInterval(() => this._checkSessionStateChanges(), 5000);
```
dmitrivMS: 2 things:
- Shouldn't this use passed in window instead?
- Perhaps use disposableWindowInterval

src/vs/workbench/contrib/chat/browser/voiceClient/voiceSessionController.ts
```diff
@@ -0,0 +1,2097 @@
...
+
+		// Set up the tool dispatch delegate — uses command bridge for widget ops
+		this.voiceToolDispatchService.setDelegate({
+			acceptInput: (text: string): boolean => {
+				this.commandService.executeCommand('_chat.voice.acceptInput', text).catch(err => {
+					this.logService.warn('[voice] acceptInput delegate failed:', err);
+				});
+				return true;
+			},
+			getCurrentSessionResource: (): URI | undefined => {
```
dmitrivMS: Does this ever return non-null value? There is code that depends on this not being null, but it appears that it never is...

src/vs/workbench/contrib/agentsVoice/browser/components/sessionListComponent.ts
```diff
@@ -0,0 +1,207 @@
...
+					.voice-session-list::-webkit-scrollbar-thumb:hover{background:var(--vscode-scrollbarSlider-activeBackground);}
+					.voice-session-list > div:last-of-type{border-bottom:none !important;}
+				</style>
+			</div>
+		`;
+	}
+
+	return html`
+		<div class="voice-session-list" style="display:flex;flex-direction:column;min-height:84px;max-height:320px;overflow-y:auto;margin:0 -14px 0 0;padding-right:8px;">
+			${headerRow}
```
alexdima, line 196: I think this kind of logic could cause XSS vulnerabilities.

src/vs/workbench/contrib/agentsVoice/browser/agentsVoiceWidget.ts
meganrogge: follow-up

src/vs/workbench/contrib/chat/browser/widgetHosts/viewPane/chatViewPane.ts
meganrogge: follow-up

src/vs/workbench/contrib/agentsVoice/browser/components/sessionListComponent.ts
meganrogge: I think this is fine because lit-html's html tagged template automatically escapes all interpolated values in text and attribute positions. It uses `setAttribute/textContent` under the hood, not `innerHTML`. There's no `unsafeHTML` usage anywhere in these components, so `${session.label}` is  XSS-safe by default.

src/vs/workbench/contrib/chat/browser/voiceClient/voiceToolDispatchService.ts
meganrogge: This matches the behavior of Agent Mode in the chat widget. When you start an agent session it has access to all tools. The agent's built-in confirmation flow still gates dangerous actions (file edits, terminal commands). We're not bypassing any safety checks; we're just starting the session in agent mode rather than ask mode.

src/vs/workbench/contrib/agentsVoice/browser/agentsVoiceWindowService.ts
meganrogge: Tracked in #321069

src/vs/workbench/contrib/chat/browser/widgetHosts/viewPane/chatViewPane.ts
meganrogge: Tracked in #321069

src/vs/workbench/contrib/agentsVoice/browser/agentsVoiceWidget.ts
meganrogge: Tracked in #321069

dmitrivMS (approved):

PR #320835, opened 2026-06-10
Filter subagent sessions out of listSessions

roblourens:
## Problem

`AgentService.listSessions` overlays sessions that were announced via `getAnnouncedSessionSummaries()` onto the providers' results, to recover sessions that providers transiently drop. Subagent child sessions are registered in the state manager's announced summaries (via `restoreSession` in `_handleSubagentStarted`), so they leaked into the top-level session list. Subagent sessions are nested under their parent and should never surface as top-level entries.

## Fix

Skip announced summaries whose resource is a subagent session (`isSubagentSession(summary.resource)`) in the `listSessions` fallback loop.

## Tests

- **Unit** (`agentService.test.ts`): registers a subagent child session in the state manager (matching the live `_handleSubagentStarted` path), confirms it is announced, then asserts `listSessions` returns the parent but no subagent sessions.
- **Integration** (`turnExecution.integrationTest.ts`): drives a real subagent turn over the WebSocket protocol, verifies the child session is live (subscribable), then asserts the `listSessions` RPC result includes the parent but excludes all subagent sessions.

Both new tests pass, and the existing `listSessions` aggregation tests still pass.

(Written by Copilot)


connor4312:
I think this might be conflicting/going away with @sandy081 's multi-chat work, since subagent sessions I assume will be modelled as chats within the same session

roblourens:
Oh ok. Well I'll merge it for tomorrow's build

justschen (approved):

PR #320853, opened 2026-06-10
Build pipeline for agent SDK tarballs

TylerLeonhardt:
Build pipeline that produces the SDK tarballs the runtime side (#320709) downloads from `main.vscode-cdn.net`.

### `build/agent-sdk/`

Four self-contained scripts:

- **`package.ts`** — builds one `(sdk, target)` tarball. `npm install` with `npm_config_libc/os/cpu` set to fetch the foreign-platform binary, chmod platform executables, normalize mtimes (`lutimes` so symlinks like `.bin/codex` get their stamp too), tar+gzip with reproducible flags (`--format=gnu --sort=name --owner=0 ... --mtime=@0 -n -9`). Writes `<sdk>-<version>-<target>.tgz` + a `.tgz.json` sidecar carrying `{sdk, sdkVersion, sdkTarget, sha256}`.
- **`upload.ts`** — uploads one tarball to the `vscodeweb` storage account's `$web` container at `agent-sdk/<sdk>/<version>/<target>.tgz`. **HEAD-then-decide:** absent → upload; matching `metadata.sha256` → skip (idempotent re-runs are free); different sha → fail loud, refusing to overwrite content-addressed history. Reuses the established `ClientAssertionCredential` auth pattern from `build/azure-pipelines/upload-cdn.ts`.
- **`aggregate.ts`** — scans all per-job sidecars, validates every expected `(sdk, target)` pair is present (reads each SDK's own `optionalDependencies` for the expected list — no parallel list to keep in sync), and prints the `product.agentSdks` JSON fragment for a human to paste into `vscode-distro`'s `product.json`.
- **`common.ts`** — shared types and helpers. The two interesting bits:
  - `getSdkVersion(sdk)` reads the pinned exact version from the repo-root `package.json` devDeps. Rejects ranged versions (`^`, `~`) — they'd let npm resolve different versions across days, breaking determinism.
  - `getTargets(sdk)` reads the supported platform list from the SDK package's own `optionalDependencies`. The SDK declares which platforms it ships; we ask it directly instead of duplicating the list.

### `build/azure-pipelines/agent-sdk/` + `product-build.yml`

A new `AgentSDK` stage in the existing scheduled product build, fanning out 14 jobs:

- **Per-target jobs** (`product-build-agent-sdk-target.yml`) — one per `(sdk, target)` pair, each invoked from an `each` loop over `claudeTargets` / `codexTargets` parameter arrays. Builds the tarball, publishes it as a pipeline artifact, and (when `VSCODE_PUBLISH=true`) uploads to CDN.
- **Aggregate job** (`product-build-agent-sdk.yml`) — downloads every per-target artifact and runs `aggregate.ts`. Uses `condition: succeededOrFailed()` deliberately: when one or more per-target jobs fail, the operator still sees a loud "Missing sidecars for N expected pairs" message naming each missing target, instead of having to click through 14 sibling jobs.

## How a SDK bump now works

1. Edit the corresponding devDep in the repo-root `package.json` (`@anthropic-ai/claude-agent-sdk` or `@openai/codex`) to the new exact version.
2. `npm install`, commit lockfile.
3. Next scheduled (or manual) pipeline run produces new tarballs at new content-addressed CDN paths.
4. Copy the printed JSON fragment from the aggregate job's log.
5. Paste into `vscode-distro`'s `product.json` under `agentSdks`.

No lockfiles to regenerate, no parallel "supported targets" list to update, no `SDK_VERSIONS` constant to bump.

## Determinism

The build is byte-reproducible **for a given Linux runner image + Node version** through:
- Pinned exact SDK version (and the SDK itself pins its `optionalDependencies` to exact versions, so the platform binary fetched is deterministic).
- GNU tar + gzip with deterministic flags + sorted entries + zero mtimes.
- `lutimes` (not `utimes`) for symlinks so things like `node_modules/.bin/codex` get normalized too.

Cross-run drift (agent image bump changes gzip output, registry-side transitive resolution shifts, etc.) is caught at upload time by the `upload.ts` HEAD-then-fail check — that's the actual defense, and it catches across-day drift that a "build twice in a row" check never could. We deliberately don't have a `verify-determinism.ts` script anymore for that reason.

## Decisions worth knowing

- **Reuses `vscodeweb` storage account.** Same account as the existing web bundle uploads, just under the `agent-sdk/` prefix. No new infra to provision, no new credentials.
- **No lockfiles committed for the per-target builds.** Pinning the SDK to an exact version is enough; transitive dep drift in Claude's tree (98 packages) surfaces at upload time as a sha mismatch where a human investigates.
- **One source of truth per concept.** Version → repo-root `package.json` devDeps. Targets → SDK's `optionalDependencies`. SDK package name → `PACKAGE_NAME` constant (unavoidable — defines what to look up).
- **Drafted as Draft PR** because the AzDO pipeline doesn't run on PRs; first verification will be a manual pipeline queue against this branch.

## Risks / things to verify

- **`replace()` template expression in YAML** — used to convert `darwin-arm64` → `darwin_arm64` for Azure Pipelines job names (which disallow hyphens). Documented in Azure's template-expression reference but no other prior art in this repo. If it doesn't expand, the fallback is one extra layer: pass target arrays as `{slug, target}` objects.
- **Blob `metadata` permissions** — `BlobServiceClient.uploadFile` with `metadata` may require `Storage Blob Data Owner` rather than `Contributor` role on the storage account. We'll find out on first upload attempt.
- **First pipeline run cost** — 14 parallel jobs × ~3 min each = roughly 42 agent-minutes per pipeline run when this stage fires.

## Test plan

- [ ] **Trigger a manual pipeline run** against this branch with `VSCODE_PUBLISH=false`. Expect 14 per-target jobs green + aggregate job prints the fragment to its log (without uploading anything).
- [ ] **Manual run with `VSCODE_PUBLISH=true`** once the dry run is green. Verify blobs land at `https://main.vscode-cdn.net/agent-sdk/<sdk>/<version>/<target>.tgz` and HEAD-then-fail idempotency works on a re-run.
- [ ] **Verify aggregate's printed fragment** is paste-ready JSON that matches the `IProductConfiguration.agentSdks` shape from `src/vs/base/common/product.ts`.

### DIFF
--- a/build/agent-sdk/README.md
+++ b/build/agent-sdk/README.md
@@ -8,12 +8,13 @@ for a human to paste into `vscode-distro`'s `product.json` (see
 ## Scripts
 
 - `package.ts` — builds one tarball for one `(sdk, target)` pair. **Linux only.**
-  Stages a minimal `package.json` pinning `SDK_VERSIONS[sdk]` (from
-  `common.ts`), runs `npm install --no-package-lock --ignore-scripts` with
+  Stages a minimal `package.json` pinning the SDK version (from
+  `getSdkVersion(sdk)` — read live from the repo-root `package.json` devDeps),
+  runs `npm install --no-package-lock --ignore-scripts` with
   `npm_config_libc/os/cpu` set to the target's triple to fetch the foreign
   platform binary, normalizes mtimes (including symlinks via `lutimes`), and
   tars with reproducible flags. Emits `<sdk>-<version>-<target>.tgz` + a
-  `.tgz.json` sidecar carrying the sha256 and size.
+  `.tgz.json` sidecar carrying the sha256.
 - `upload.ts` — uploads one tarball to the `main.vscode-cdn.net` storage
   account (path `agent-sdk/<sdk>/<sdkVersion>/<sdkTarget>.tgz`). HEAD-first:
   absent → upload; present with matching sha256 → skip (idempotent re-run);
@@ -25,13 +26,15 @@ for a human to paste into `vscode-distro`'s `product.json` (see
 - `aggregate.ts` — scans a directory of `<sdk>-<version>-<target>.tgz.json`
   sidecars (one per per-target job's artifact) and prints a markdown table
   with the JSON fragment a human pastes into `vscode-distro`'s `product.json`.
-  Validates that every expected `(sdk, target)` pair — read live from each
-  SDK's own `optionalDependencies` — is accounted for, failing loud on
-  missing entries.
+  Takes `--claude-targets` / `--codex-targets` flags from the pipeline; fails
+  loud if any expected sidecar is missing. As a separate non-blocking check,
+  queries each SDK's registry `optionalDependencies` and WARNS (does not
+  fail) if upstream declares a platform we don't build yet — an early signal
+  to update the YAML matrix.
 - `common.ts` — shared types and helpers. `getSdkVersion(sdk)` reads the
-  pinned version from the repo-root `package.json` devDeps; `getTargets(sdk)`
-  reads the platform list from the SDK's own `optionalDependencies` so the
-  matrix stays in sync with what upstream actually ships.
+  pinned version from the repo-root `package.json` devDeps;
+  `getRegistryTargets(sdk)` queries the SDK's `optionalDependencies` via
+  `npm view` for drift detection.
 
 ## Bumping an SDK version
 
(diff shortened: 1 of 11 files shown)


TylerLeonhardt:
Superseded by the runtime simplification in #320883. The replacement build PR inlines SDK production into the per-platform packageTask in the existing gulpfiles (no separate AgentSDK stage, no aggregate). New PR coming shortly.

Issue #247522, opened 2025-04-27
[Accessibility] Mouse echo don't work with NVDA and  JAWS


Shqipton:
Does this issue occur when all extensions are disabled?: Yes/No
yes

- VS Code Version: >1.92
- OS Version: current Windows 11 

Steps to Reproduce:

1. Open Visual Studio Code version 1.92 (or later).
2. Ensure that a Screen Reader is running (e.g., NVDA, JAWS, or ZoomText).
3. Make sure that the "Mouse Echo" feature is enabled in the Screen Reader settings (it reads text under the mouse pointer).
4. Move the mouse pointer over text in the editor area of VSCode.

5. Observe that the Screen Reader does not announce or read the text under the mouse, even though it should.

In Version June 2024 or fewer it works.

Shqipton:
Hello, thanks for your reply!

I am intentionally using the June 2024 version of VS Code because the issue appears in newer versions (starting from the July 2024 release).
In the June 2024 version (1.91), the Mouse Echo feature of screen readers (e.g., NVDA, JAWS) still works properly.
In all newer versions (starting from 1.92), Mouse Echo no longer reads text under the mouse pointer in the editor.

Therefore, upgrading to the latest version does not solve the problem — it is the reason I created this issue.
It seems that a change between June and July 2024 caused a regression in screen reader mouse support.

Thank you very much for investigating this further!

meganrogge:
@aiday-mar do you know any changes that would've caused this to break?

aiday-mar:
Hey thanks for the ping, looking into it. 

@Shqipton when you mention the mouse echo setting in NVDA are you referring to the setting `Report object when mouse enters it` under the Mouse category? Could I ask if you could share your settings with me please?

Shqipton:
Yes, I meen the seetings "mouse echo" in NNVDA and JAWS. Wich seetings I should share with you?

aiday-mar:
The whole settings file if possible. It is called the `settings.json` file and can be accessed with the command `workbench.action.openSettingsJson`

opeyemi0098:
Fun

Shqipton:
Here is the settings-File:
`{
    "editor.accessibilitySupport": "on",
    "workbench.startupEditor": "none",
    "git.confirmSync": false,
"update.mode": "manual" ,
    "editor.semanticTokenColorCustomizations": {},
    "update.showReleaseNotes": false
}`

Shqipton:
[...]

PR #320877, opened 2026-06-11
Allow Chromium feature switches in argv config

scarab-systems:
Related to #247522.

This adds `enable-features` and `disable-features` to the native argument parser, the `argv.json` schema, and the supported Electron switch allowlist.

The default behavior is unchanged. The switches are only forwarded when explicitly configured, but this gives the VS Code team and affected users a persistent way to apply Chromium feature-flag diagnostics or workarounds while investigating accessibility-provider regressions.

Testing:
- `npm ci`
- `npm run transpile-client`
- `npm run test-node -- --run src/vs/platform/environment/test/node/argv.test.ts --grep "recognizes Chromium feature switches"`
- `git diff --check origin/main...HEAD`

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -235,6 +235,10 @@ function configureCommandlineSwitchesSync(cliArgs: NativeParsedArgs) {
 		// bypass any specified proxy for the given semi-colon-separated list of hosts
 		'proxy-bypass-list',
 
+		// override Chromium feature state for troubleshooting
+		'enable-features',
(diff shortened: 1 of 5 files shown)


scarab-systems:
Gentle follow-up on this PR. The visible checks are not showing failures on my side, though I see the community PR approval/status checks are still pending.

Is this a direction VS Code would consider, or should I close it if it is not useful?

Issue #222244, opened 2024-07-19
Search in markdown preview panel not working in current main branch local build


manav014:
Does this issue occur when all extensions are disabled?: Yes

Type: Bug

I recently build vscode on local with following configuration and last commit being https://github.com/microsoft/vscode/commit/61b242c67f23d24bcb9abb25bcac223547f032bd .

```
Version: 1.92.0
Commit: Unknown
Date: Unknown
Electron: 30.1.2
ElectronBuildId: undefined
Chromium: 124.0.6367.243
Node.js: 20.14.0
V8: 12.4.254.20-electron.0
OS: Darwin arm64 23.5.0
```

I found that the search functionality is not working for markdown preview. Seems like a recently introduced issue as when I tried to perform the search operation in stable release I was able to search across markdown preview.

Steps to Reproduce:
1.  Preview a Markdown file
2. Search for any keyword present in markdown file.
3. You will see that it won't show results.

This bug could be pushed with next release and can cause issues with upcoming stable releases.

PFA: POC video for the issue.

https://github.com/user-attachments/assets/0bb47435-cbc0-4e56-a821-61d59c2db412

Would be great if @mjbvz , you can shed some light here.

dangerman:
When trying to search in a Markdown Preview whilst debugging, the code does hit `WebViewMainService.findInFrame()`:
https://github.com/microsoft/vscode/blob/709c3d756947c2c275bf90a87a071eca58453851/src/vs/platform/webview/electron-main/webviewMainService.ts#L50-L59

However, `frame.findInFrame` is `undefined` so it just skips the rest of the function - and this appears to be why it's not working.

Now, according to this issue https://github.com/microsoft/vscode/issues/201176 Microsoft uses a **custom build** of electron (from https://github.com/microsoft/vscode-electron-prebuilt that is not publicly accessible) for VS Code releases, with a `WebFrameMain.FindInFrame()` implementation, so searching works.

The intention seems to be that changes are gradually upstreamed into the main electron repo, but unfortunately the PR for the `FindInFrame` stuff is Closed for the time being: https://github.com/electron/electron/pull/28274

Perhaps if we're able to merge the branch from that PR into an updated local build of electron, then we could use it with our vscode for debugging? 🤔

rzhao271:
@manav014 are you still able to reproduce the issue on VS Code 1.96?

deepak1556:
@rzhao271 the api is not available in OSS builds

[...]

PR #125856, opened 2021-06-09
Add iframe search with debounce

rzhao271:
This PR fixes #96307 by adding an iframe search prototype. It uses https://github.com/electron/electron/pull/28274 to search in the iframe, which results in a loss of focus during the search. Therefore, I added an event to re-focus the inputbox after find results are returned, and also added a debounce so that the focus-changing behaviour of the find call is not too obvious.

Preview:
recording (31)

Issues:
- The jumping search problem remains (https://github.com/microsoft/vscode/issues/96307#issuecomment-848137542). This issue can be fixed by using a BrowserView, in which case, the debounce can be taken off as well (https://github.com/microsoft/vscode/issues/96307#issuecomment-844362867).

PR #320898, opened 2026-06-11
fix: avoid refocusing find input during IME composition

### LINKED ISSUE #320869: Korean IME input is dropped or corrupted in Markdown preview Cmd+F find box

- VS Code Version: 1.123.2
- OS Version: macOS 26.4.1 (25E253), arm64

Steps to Reproduce:

1. Open any Markdown file.
2. Open the built-in Markdown preview, for example with `Markdown: Open Preview to the Side`.
3. Focus the Markdown preview pane.
4. Press `Cmd+F` to open the find box for the preview.
5. Switch to the Korean input method and type Korean text into the find box, for example `안녕하세요` or another Korean search term.

Expected:

The Korean text should be entered into the Markdown preview find box correctly, with all composed characters preserved.

Actual:

Some Korean keystrokes are dropped or entered incorrectly in the Markdown preview find box. The text shown in the find box can differ from what was typed, making it unreliable to search Korean text in Markdown preview.
Image

Notes:

I searched existing issues and PRs before filing this report. I found related Korean/IME/find/webview issues, but I could not find an exact duplicate for Korean IME keystrokes being dropped or misinterpreted specifically in the Markdown preview `Cmd+F` find box.

I also reproduced the issue with all extensions disabled using `code --disable-extensions`.

The same behavior was confirmed on a colleague's computer as well, so this does not appear to be specific to my local extension setup or user settings.

wibaek:
Fixes #320869

**What changed?**
Updated the shared find input flow so input events fired from compositionend can update search state without immediately restoring focus to the find box (FindInput.onInput now carries { fromCompositionEnd }). Also added a guard so focusFindBox() does not refocus the find input while an IME composition session is still active.

**Why?**
focusFindBox() blurs and refocuses the find input (bouncing through the next-match button). When this runs for the input fired at compositionend — or when an async find result arrives while the next character is already composing — the focus change aborts the active IME composition. Korean IME fires compositionend at every syllable boundary, so characters were dropped or corrupted while typing. The search value is still processed as before; only the focus churn around composition is skipped.

**How I tested this:**
Reproduced the original bug and verified the fix manually with ./scripts/code-web.sh: typing Korean (e.g. 안녕하세요) in the markdown preview find box corrupted characters before this change and is clean after. (Tested on web because webview find is not functional in local desktop builds — stock Electron lacks the findInFrame patch, see [#222244](https://github.com/microsoft/vscode/issues/222244).)
Verified English typing, Enter navigation, and find history behavior are unchanged.
npm run compile-check-ts-native and ESLint on the changed files pass.

**Before / after**
Before:

https://github.com/user-attachments/assets/cb140718-c5db-448c-a112-02307e6cd01d

After:

https://github.com/user-attachments/assets/55404d94-d097-468f-98c5-aca5b0ea6b10

### DIFF
--- a/src/vs/base/browser/ui/findinput/findInput.ts
+++ b/src/vs/base/browser/ui/findinput/findInput.ts
@@ -46,6 +46,10 @@ export interface IFindInputOptions {
 	readonly hideHoverOnValueChange?: boolean;
 }
 
+export interface IFindInputEvent {
+	readonly fromCompositionEnd: boolean;
+}
+
 const NLS_DEFAULT_LABEL = nls.localize('defaultLabel', "input");
 
 export class FindInput extends Widget {
@@ -77,8 +81,8 @@ export class FindInput extends Widget {
 	private readonly _onMouseDown = this._register(new Emitter<IMouseEvent>());
 	public get onMouseDown(): Event<IMouseEvent> { return this._onMouseDown.event; }
 
-	private readonly _onInput = this._register(new Emitter<void>());
-	public get onInput(): Event<void> { return this._onInput.event; }
+	private readonly _onInput = this._register(new Emitter<IFindInputEvent>());
+	public get onInput(): Event<IFindInputEvent> { return this._onInput.event; }
 
 	private readonly _onKeyUp = this._register(new Emitter<IKeyboardEvent>());
 	public get onKeyUp(): Event<IKeyboardEvent> { return this._onKeyUp.event; }
@@ -229,12 +233,12 @@ export class FindInput extends Widget {
 		}));
 		this._register(dom.addDisposableListener(this.inputBox.inputElement, 'compositionend', (e: CompositionEvent) => {
 			this.imeSessionInProgress = false;
-			this._onInput.fire();
+			this._onInput.fire({ fromCompositionEnd: true });
 		}));
 
 		this.onkeydown(this.inputBox.inputElement, (e) => this._onKeyDown.fire(e));
 		this.onkeyup(this.inputBox.inputElement, (e) => this._onKeyUp.fire(e));
-		this.oninput(this.inputBox.inputElement, (e) => this._onInput.fire());
+		this.oninput(this.inputBox.inputElement, (e) => this._onInput.fire({ fromCompositionEnd: false }));
 		this.onmousedown(this.inputBox.inputElement, (e) => this._onMouseDown.fire(e));
 	}
 
--- a/src/vs/workbench/contrib/codeEditor/browser/find/simpleFindWidget.ts
+++ b/src/vs/workbench/contrib/codeEditor/browser/find/simpleFindWidget.ts
@@ -130,7 +130,9 @@ export abstract class SimpleFindWidget extends Widget implements IVerticalSashLa
 					await this.updateResultCount();
 				}
 				this.updateButtons(this._foundMatch);
-				this.focusFindBox();
+				if (!options.checkImeCompletionState || !e.fromCompositionEnd) {
+					this.focusFindBox();
+				}
 				this._delayedUpdateHistory();
 			}
 		}));
@@ -412,6 +414,9 @@ export abstract class SimpleFindWidget extends Widget implements IVerticalSashLa
 	}
 
 	protected focusFindBox() {
+		if (this._findInput.isImeSessionInProgress) {
+			return;
+		}
 		// Focus back onto the find box, which
 		// requires focusing onto the next button first
 		this.nextBtn.focus();


wibaek:

src/vs/base/browser/ui/findinput/findInput.ts
```diff
@@ -77,8 +81,8 @@ export class FindInput extends Widget {
 	private readonly _onMouseDown = this._register(new Emitter<IMouseEvent>());
 	public get onMouseDown(): Event<IMouseEvent> { return this._onMouseDown.event; }
 
-	private readonly _onInput = this._register(new Emitter<void>());
-	public get onInput(): Event<void> { return this._onInput.event; }
+	private readonly _onInput = this._register(new Emitter<IFindInputEvent>());
+	public get onInput(): Event<IFindInputEvent> { return this._onInput.event; }
```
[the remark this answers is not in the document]
wibaek: The only in-tree subscriber is SimpleFindWidget, updated in this PR. Parameterless listeners remain assignable to Event<IFindInputEvent>, so no call-site breaks either way; a parallel emitter would duplicate the same signal across two events.

src/vs/workbench/contrib/codeEditor/browser/find/simpleFindWidget.ts
```diff
@@ -130,7 +130,9 @@ export abstract class SimpleFindWidget extends Widget implements IVerticalSashLa
 					await this.updateResultCount();
 				}
 				this.updateButtons(this._foundMatch);
-				this.focusFindBox();
+				if (!options.checkImeCompletionState || !e.fromCompositionEnd) {
+					this.focusFindBox();
+				}
```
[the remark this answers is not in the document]
wibaek: The two checks cover different timing windows rather than duplicating logic: the event-provenance check handles the synchronous compositionend-driven update (at that instant the session flag is already false, so live state can't detect it), while the guard in focusFindBox() handles asynchronous callers like find-result events (where provenance is unknowable and live state is the only correct signal). Centralizing both in focusFindBox() would require a "just completed composition" side-channel on FindInput that's only valid during synchronous dispatch — a fragile contract this PR deliberately avoids.

src/vs/base/browser/ui/findinput/findInput.ts
```diff
@@ -229,12 +233,12 @@ export class FindInput extends Widget {
...
 		this._register(dom.addDisposableListener(this.inputBox.inputElement, 'compositionend', (e: CompositionEvent) => {
 			this.imeSessionInProgress = false;
-			this._onInput.fire();
+			this._onInput.fire({ fromCompositionEnd: true });
 		}));
 
 		this.onkeydown(this.inputBox.inputElement, (e) => this._onKeyDown.fire(e));
 		this.onkeyup(this.inputBox.inputElement, (e) => this._onKeyUp.fire(e));
-		this.oninput(this.inputBox.inputElement, (e) => this._onInput.fire());
+		this.oninput(this.inputBox.inputElement, (e) => this._onInput.fire({ fromCompositionEnd: false }));
```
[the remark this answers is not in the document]
wibaek: The double-fire isn't introduced by this PR — the previous code already fired onInput from both compositionend and the native input event, just untagged. This PR doesn't change how many events consumers receive; it only adds provenance to each one.

Firing only from input (and not from compositionend) would break Chromium: there, the committed text's input event arrives before compositionend, with isComposing: true and the IME session flag still set, so it is deliberately ignored. With the compositionend fire removed, the search would never update at commit time during continuous IME typing. Firing from compositionend has been the intended design since #125856

The dangerous window (an active composition) is already covered twice: an input event arriving after the next composition has started is skipped by the handler's isImeSessionInProgress check, and focusFindBox() has its own guard. What remains is a single focus bounce at a moment when no composition is active — the same thing that already happens on every Latin keystroke, and strictly less than before this PR (previously both events triggered the bounce).

wibaek:
@dmitrivMS 🥹🥹

wibaek:
Hi @dmitrivMS, could you please review this PR when you have a chance and let me know if it’s ready to be merged or if anything needs to be revised?

src/vs/workbench/contrib/codeEditor/browser/find/simpleFindWidget.ts
```diff
@@ -130,7 +130,9 @@ export abstract class SimpleFindWidget extends Widget implements IVerticalSashLa
 					await this.updateResultCount();
 				}
 				this.updateButtons(this._foundMatch);
-				this.focusFindBox();
+				if (!options.checkImeCompletionState || !e.fromCompositionEnd) {
```
dmitrivMS, line 133: AI Review: Nit: Could we add focused browser coverage for both new focus-suppression paths? `compositionend` should update search without the next-button/input focus bounce, while a `hasFindResult` callback after a new `compositionstart` should also leave focus untouched. A deterministic `WebviewFindWidget` test can use a mock delegate, count focus/blur events, and retain a normal-input control assertion. There is currently no automated coverage for either path.

wibaek: Added focused browser coverage in 554413076ac:

- `compositionend` updates search without the next-button/input focus bounce.
- `hasFindResult` arriving after a new `compositionstart` leaves focus unchanged.
- Normal input still updates search and restores focus.

The tests use a mock delegate and record actual DOM focus/blur events. All three pass in Chromium and WebKit.

I also verified that removing each focus-suppression guard independently causes its corresponding regression test to fail.

dmitrivMS: @wibaek The tests are failing it seems...

dmitrivMS (approved):

src/vs/workbench/contrib/codeEditor/browser/find/simpleFindWidget.ts
wibaek: 😮

sandy081 (approved):

PR #320943, opened 2026-06-11
Enable 'Add File as Context' action in command palette

mrleemurray:
This pull request introduces a new editor action that allows users to add the currently open file as context to an active session, improving workflow integration between the editor and session/chat features. The main changes include the addition of the `AddFileAsContextAction` class and necessary imports to support this new functionality.

image

**New Editor Action for Session Context:**

* Added a new `AddFileAsContextAction` class, which registers an editor title menu action called "Add File as Context." This action is available when a text file is open in a session window and, when triggered, adds the active file as context to the current session's chat widget.

**Supporting Imports:**

* Imported `ActiveEditorContext`, `IEditorService`, `TEXT_FILE_EDITOR_ID`, `ISessionsManagementService`, and `IChatWidgetService` to enable context checks and facilitate the new action's integration with session and chat services.

### DIFF
--- a/src/vs/sessions/contrib/chat/browser/chatView.ts
+++ b/src/vs/sessions/contrib/chat/browser/chatView.ts
@@ -84,8 +84,8 @@ export class NewChatView extends AbstractChatView {
 		}
 	}
 
-	override attach(uri: URI): void {
-		this._widget.attach(uri);
+	override attach(uris: URI[]): void {
+		this._widget.attach(uris);
 	}
 }
 
@@ -247,8 +247,10 @@ export class ChatView extends AbstractChatView {
 		this._widget.focusInput();
 	}
 
-	override attach(uri: URI): void {
-		this._widget.attachmentModel.addFile(uri).catch(err => this.logService.error('[ChatView] Failed to attach file as context', err));
+	override attach(uris: URI[]): void {
+		for (const uri of uris) {
+			this._widget.attachmentModel.addFile(uri).catch(err => this.logService.error('[ChatView] Failed to attach file as context', err));
+		}
 	}
 
 	override setActive(active: boolean): void {
--- a/src/vs/sessions/contrib/chat/browser/chatView.ts
+++ b/src/vs/sessions/contrib/chat/browser/chatView.ts
@@ -85,9 +85,7 @@ export class NewChatView extends AbstractChatView {
 	}
 
 	override attach(uri: URI): void {
-		if (this._widget instanceof NewChatWidget) {
-			this._widget.attach(uri);
-		}
+		this._widget.attach(uri);
 	}
 }
 
--- a/src/vs/sessions/contrib/chat/browser/chatView.ts
+++ b/src/vs/sessions/contrib/chat/browser/chatView.ts
@@ -83,6 +83,12 @@ export class NewChatView extends AbstractChatView {
 			this._widget.sendQuery(text);
 		}
 	}
+
+	override attach(uri: URI): void {
+		if (this._widget instanceof NewChatWidget) {
+			this._widget.attach(uri);
+		}
+	}
 }
 
 /**
@@ -241,6 +247,10 @@ export class ChatView extends AbstractChatView {
 
 	override focus(): void {
 		this._widget.focusInput();
+	}
+
+	override attach(uri: URI): void {
+		this._widget.attachmentModel.addFile(uri).catch(err => this.logService.error('[ChatView] Failed to attach file as context', err));
 	}
 
 	override setActive(active: boolean): void {
--- a/src/vs/sessions/contrib/editor/browser/editor.contribution.ts
+++ b/src/vs/sessions/contrib/editor/browser/editor.contribution.ts
@@ -340,19 +340,23 @@ class AddFileAsContextAction extends Action2 {
 	static readonly ID = 'workbench.action.agentSessions.addFileAsContext';
 
 	constructor() {
+		const precondition = ContextKeyExpr.and(
+			IsSessionsWindowContext,
+			IsAuxiliaryWindowContext.toNegated(),
+			ActiveEditorContext.isEqualTo(TEXT_FILE_EDITOR_ID)
+		);
+
 		super({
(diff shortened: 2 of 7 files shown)


benibenj (approved):

benibenj:
I like it, let's try this

benibenj (changes requested):
Actually, I just noticed. This won't work in the new chat widget. Let me give you some context, one sec...

benibenj:
The run method should get the active session resource and use the ISessionsPartService getSessionView method to get the session view. We need to add an attach method to SessionView which then calls attach on chat view (method needs to be added) which then makes sure the attachment is added to the chat widget / chat input

mrleemurray:
Refactored to route attachment through the sessions architecture instead of IChatWidgetService directly, as suggested:

run now resolves the active session resource (IEditorService) + active session (ISessionsManagementService), then goes through ISessionsPartService.getSessionView(sessionId)?.attach(resource).

The call chain:

SessionView.attach → forwards to the current chat view
ChatView.attach (new) → attachmentModel.addFile(uri) on the widget
So: run → getSessionView → SessionView.attach → ChatView.attach → ChatWidget.attachmentModel, keeping the contrib layer decoupled from the chat widget service.

src/vs/sessions/contrib/editor/browser/editor.contribution.ts
```diff
@@ -331,3 +336,50 @@ class OpenModalEditorInEditorAction extends Action2 {
...
+			menu: {
+				id: MenuId.EditorTitle,
+				group: 'navigation',
+				order: 1,
+				when: precondition
+			}
+		});
+	}
+
+	run(accessor: ServicesAccessor): void {
```
benibenj: It's possible to click the action without it being the active editor. The run method also gets arguments about the context in which it's running in (which editor group it's running from). We have the `resolveCommandsContext` utility method that can resolve the arguments of the run command to the correct editor.

benibenj (changes requested):

src/vs/sessions/contrib/editor/browser/editor.contribution.ts
```diff
@@ -331,3 +336,50 @@ class OpenModalEditorInEditorAction extends Action2 {
...
+			return;
+		}
+
+		const activeSession = sessionManagementService.activeSession.get();
+		if (!activeSession) {
+			return;
+		}
+
+		sessionsPartService.getSessionView(activeSession.sessionId)?.attach(resource);
+	}
```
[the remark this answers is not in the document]
benibenj: This feedback is correct. You can apply the suggsted change

src/vs/sessions/contrib/chat/browser/chatView.ts
```diff
@@ -83,6 +83,12 @@ export class NewChatView extends AbstractChatView {
 			this._widget.sendQuery(text);
 		}
 	}
+
+	override attach(uri: URI): void {
+		if (this._widget instanceof NewChatWidget) {
+			this._widget.attach(uri);
+		}
+	}
```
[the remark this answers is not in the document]
benibenj: Yes, we should add it for all new chat widgets, also NewChatInSessionWidget

src/vs/sessions/contrib/editor/browser/editor.contribution.ts
```diff
@@ -331,3 +339,54 @@ class OpenModalEditorInEditorAction extends Action2 {
...
+	run(accessor: ServicesAccessor, ...args: unknown[]): void {
+		const editorService = accessor.get(IEditorService);
+		const sessionManagementService = accessor.get(ISessionsManagementService);
+		const sessionsPartService = accessor.get(ISessionsPartService);
+
+		const resolvedContext = resolveCommandsContext(args, editorService, accessor.get(IEditorGroupsService), accessor.get(IListService));
+		const resource = resolvedContext.groupedEditors
+			.flatMap(groupedEditor => groupedEditor.editors)
+			.map(editor => EditorResourceAccessor.getCanonicalUri(editor, { supportSideBySide: SideBySideEditor.PRIMARY }))
+			.find(uri => uri !== undefined);
```
benibenj: let's not do a find, but a filter, and then attach all resources, not just one. This will allow flows like multi selecting editor tabs and pressing the attach action will attach all selected files.

src/vs/sessions/contrib/editor/browser/editor.contribution.ts
```diff
@@ -331,3 +339,54 @@ class OpenModalEditorInEditorAction extends Action2 {
...
+		if (!resource) {
+			return;
+		}
+
+		if (!['file', 'vscode-remote', 'untitled'].includes(resource.scheme)) {
+			return;
+		}
+
+		const sessionId = sessionManagementService.activeSession.get()?.sessionId;
+		sessionsPartService.getSessionView(sessionId)?.attach(resource);
```
benibenj: let's make attach an array of URI

src/vs/sessions/contrib/editor/browser/editor.contribution.ts
```diff
@@ -331,3 +339,54 @@ class OpenModalEditorInEditorAction extends Action2 {
...
+		const resolvedContext = resolveCommandsContext(args, editorService, accessor.get(IEditorGroupsService), accessor.get(IListService));
+		const resource = resolvedContext.groupedEditors
+			.flatMap(groupedEditor => groupedEditor.editors)
+			.map(editor => EditorResourceAccessor.getCanonicalUri(editor, { supportSideBySide: SideBySideEditor.PRIMARY }))
+			.find(uri => uri !== undefined);
+		if (!resource) {
+			return;
+		}
+
+		if (!['file', 'vscode-remote', 'untitled'].includes(resource.scheme)) {
```
benibenj: this should also be a filter in that case. Also, let's use the Schemas consts from network.ts instead of hardcoding the strings

benibenj (changes requested):
This is the last one, I swear. Just noticed we can easily support multi select + attach :)

benibenj (approved):

PR #320948, opened 2026-06-11
Add setting to open changed chat files in an editor instead of a diff

kangarko:
### Problem

When a chat response edits files, it shows a "Changed N files" list. Selecting a file there always opens a diff editor. For files whose edits are already applied (past requests / checkpoints) this opens a read-only snapshot-vs-snapshot diff, so you cannot keep editing the file you were just looking at. A number of people would rather have clicking a changed file take them straight into the real editor.

### Change

Adds `chat.editing.openChangedFileInDiffEditor` (boolean, default `true`). The default preserves today's behavior. When set to `false`, selecting a file in a chat response's changed-files list opens the file in a regular editor instead of a diff editor. This applies to both the live changes summary (`chatChangesSummaryPart`) and the multi-diff list (`chatMultiDiffContentPart`).

### How

The changed-files list uses snapshot URIs (`chat-editing-snapshot-text-model`) for the diff sides, and those only carry the file path. To open the real file we also need its original scheme and authority, so `getSnapshotFileURI` now records them in the snapshot URI query, and a new `ChatEditingSnapshotTextModelContentProvider.getOriginalFileURI(resource)` recovers the real file URI. It returns `undefined` when it cannot, in which case the handler falls back to the diff editor. The change is additive: the query already reserved a `scheme` field that was unused, and recovery degrades gracefully for snapshots created before this change.

### Notes

- No behavior change with default settings.
- Inline "view diff" links in markdown responses are intentionally left as diffs.


kangarko:
@roblourens @connor4312 you two own most of the history on chatEditing snapshots and the changed-files content parts, would appreciate a review when you get a chance. Default keeps current behavior, the setting just adds an opt-out to open the real file instead of the snapshot diff.

connor4312:
I would probably prefer something like adding behavior where an alt+click on the change would open the file, vs a setting. Personally I would not use the setting, but I would use an alt+ action

kangarko:
@connor4312 that's a great idea, but for maximum customizability I prefer a dropdown if more than one options are desired, because I have a need for opening the file directly

connor4312:
Maybe we actually add the alt behavior and have a setting that toggles the default behavior? So click opens the diff by default, alt+click open the file, but the setting swaps those.

kangarko:
@connor4312 implemented in 380982d. Click opens the diff as today, alt+click opens the file, and setting chat.editing.openChangedFileInDiffEditor to `false` swaps the two. Applies to both the changes summary and the multi-diff list, alt+enter works from the keyboard as well. If the snapshot origin cannot be recovered (legacy snapshots), it falls back to the diff editor.

connor4312 (approved):
great feature, thanks! :)

kangarko:
We love VS Code, you're welcome!

Yoyokrazy (approved):

PR #320958, opened 2026-06-11
CopilotAH: fix watcher regression

aeschli:
FYI @connor4312 We listen on folders e.g `.github/agents` but `e.contains` does not detect changes to direct children of that folder. `e.affects` is needed. 
Note that whether `e.affects` reacts to nested children (`.github/agents/somefolder/test.txt`) depends on how the watcher was set up.


src/vs/platform/agentHost/node/copilot/sessionCustomizationDiscovery.ts
```diff
@@ -178,8 +178,8 @@ export class SessionCustomizationDiscovery extends Disposable {
 		this._register({ dispose: () => this._disposeAllWatchers() });
 		this._watchRootUris.clear();
 		this._register(this._fileService.onDidFilesChange(e => {
-			for (const [uri, recursive] of this._watchRootUris.entries()) {
-				if (recursive ? e.affects(uri) : e.contains(uri)) {
+			for (const [uri] of this._watchRootUris.entries()) {
+				if (e.affects(uri)) {
```
connor4312: This was causing customization changes to fire whenever anything in the folder was updated. I think something in changeset stuff was updating the .git folder every few seconds, so this resulted in customization changes being fired pretty much continuously

aeschli: Yes, I now see it in the tests, depends on the watcher implementation whether the `recursive` flag on the watcher creation is honored.
Will look on how to tighten the change detection

connor4312 (approved):

connor4312 (approved):

PR #320973, opened 2026-06-11
relay OTEL telemetry for AH chats

amunger:


amunger:
Closing this stale draft as an obsolete prototype. If relaying Agent Host chat telemetry through OpenTelemetry is still needed, it should be reconsidered against the current Agent Host and OTel architecture in a fresh, focused change.

[comment authored by copilot]

PR #320990, opened 2026-06-11
accounts: fix default account init deadlock in desktop remote connections

### LINKED ISSUE #11688: `code-server --install-extension … --start-server` hangs forever in `ExtensionGalleryManifestIPCService.barrier.wait()` (latent since 1.99, user-visible since 1.122 via `builtInExtensionsEnabledWithAutoUpdates`)
- **VS Code Server Version:** 1.122.x (regression first user-visible here; see "Timeline" below)
- **Remote OS Version:** Ubuntu 22.04 (devcontainers `mcr.microsoft.com/devcontainers/base:ubuntu22.04`); reproducible on any Linux server
- **Remote Extension/Connection Type:** VS Code Server (headless `--start-server` invocation)
- **Logs:** server log contains a single `Installing extensions...` line; no error, no progress, no completion. Process keeps the socket open but never exits the install path.

cc @sandy081, @joaomoreno — apologies in advance for the long write-up; this is a multi-layer regression and the contributing commits are yours so I want to give you full context up front.

## Summary

When VS Code Server is launched with a combined invocation that mixes extension installs with `--start-server`, e.g.:

```
code-server \
  --install-extension ms-vscode.azurecli \
  --install-builtin-extension GitHub.copilot \
  --install-builtin-extension GitHub.copilot-chat \
  --start-server --host=… --port=… --connection-token-file=… …
```

the install loop **hangs forever** inside `ExtensionGalleryManifestIPCService.getExtensionGalleryManifest()` at `await this.barrier.wait()`. The server keeps running, no extensions are written to disk, and there is no error in the log.

Splitting the invocation into two sequential processes — first `code-server --install-extension … --install-builtin-extension …` (exits cleanly), then `code-server --start-server …` — works fine. So does `--start-server` alone, and so does install-only.

## Repro

1. Build VS Code Server at the 1.122 stable tip (we reproduced against `microsoft/vscode@8761a5560cfd65fdd19ce7e2bd18dab5c0a4d84e`, "Prepare for recovery release", 2026-05-29).
2. Launch it in a fresh Linux container with the combined `--install-extension … --install-builtin-extension … --start-server` invocation shown above.
3. Observe: `Installing extensions...` is logged once. No `Installed extension …` lines follow. `$HOME/.vscode-server/extensions/extensions.json` (or the `.vscode-remote` equivalent under Dev Containers CLI) stays at `[]`. `~/.vscode-remote/data/Machine/.installExtensionsMarker` is created and never cleaned.
4. Re-attaching a workbench client *after* the server is up causes the workbench-side `RemoteExtensionGalleryManifestServiceClient` to push the manifest via the `extensionGalleryManifest` IPC channel — at which point the barrier opens and the install resumes. So the bug is reachable only when the *server* needs the gallery manifest *before* a workbench client is connected.

Does this issue occur when you try this locally?: **Yes** (when invoking `code-server` directly with the combined flags shown above).
Does this issue occur when you try this locally and all extensions are disabled?: **N/A** — the failure is *during* extension install, before any extension activates.

## Root cause (latent defect, shipped in 1.99)

- `microsoft/vscode@8c09c653cc7b8336a7fab6563515dc90bb781d08` ("adopt gallery manifest in server", PR https://github.com/microsoft/vscode/pull/243580, 2025-03-14, @sandy081) changed the server-side binding in `src/vs/server/node/serverServices.ts` from `ExtensionGalleryManifestService` (direct, reads from `productService.extensionsGallery`) to `ExtensionGalleryManifestIPCService` (waits for the workbench to push the manifest over IPC).
- The IPC variant in `src/vs/platform/extensionManagement/common/extensionGalleryManifestServiceIpc.ts` has `getExtensionGalleryManifest()` block on `await this.barrier.wait()`. The barrier is opened only by `setExtensionGalleryManifest(...)`, which is invoked by the workbench-side `RemoteExtensionGalleryManifestServiceClient` after it has connected.
- In a `--start-server` flow where the server itself initiates an install before a client has connected, **no one ever opens the barrier**, so every gallery query (e.g. `AbstractExtensionGalleryService.getExtensions(...) → getExtensionGalleryManifest()`) hangs.

## Why nothing exploded between 1.99 and 1.121

Two coincidences kept the latent hang unreachable for ~14 months:

1. **`--install-builtin-extension` did not consult the gallery.** Built-ins were resolved through `getInstalled(ExtensionType.System)`, short-circuiting before `AbstractExtensionGalleryService.getExtensions()`. So the most common headless install path never touched the manifest barrier.
2. **`--install-extension` (regular gallery) raced and won.** `RemoteExtensionsScannerService` (`src/vs/server/node/remoteExtensionsScanner.ts`) kicks the install off inside its constructor but stashes the promise on `_whenExtensionsReady` rather than awaiting it inline. Meanwhile `--start-server` brings the socket server up, the workbench client connects, calls `setExtensionGalleryManifest(...)`, the barrier opens, the install resumes. Race won every monthly release from 1.99 through 1.121.

## What flipped at 1.122 — `builtInExtensionsEnabledWithAutoUpdates`

A five-commit cluster between 2026-03-31 and 2026-04-10 introduced the **`builtInExtensionsEnabledWithAutoUpdates`** product feature (initially `autoUpdateBuiltinExtensions`):

| SHA | Date | PR | What it does |
|---|---|---|---|
| `microsoft/vscode@143b631d0c42364ba31ebccee4c7371251f976e1` | 2026-03-31 | (internal) | Adds `autoUpdateBuiltinExtensions` to `IProductConfiguration` (`src/vs/base/common/product.ts`); adds major.minor gating in `extensionGalleryService.isValidVersion`; adds `removeStaleAutoUpdateBuiltinExtensions()` in `node/extensionManagementService.ts`. |
| `microsoft/vscode@aa0db96760bee736ad76848939fe71667b2731b2` | 2026-04-01 | (internal) | Renames key to `builtInExtensionsEnabledWithAutoUpdates`; adds `NOT_APPLICABLE` gallery-service early-out. |
| `microsoft/vscode@518d26d7ce1ccdd563981b98d7ade7840192f0f2` | 2026-04-02 | (internal) | Cross-version handling for builtins shipping at an older version than product (@joaomoreno). |
| `microsoft/vscode@7f503b81e28e05f3ee8eef5ec92ab3f9f9b701d6` | 2026-04-09 | https://github.com/microsoft/vscode/pull/308756 | Scanner-side fixes ("fix scanning and updating built in extensions enabled with auto updates"). |
| `microsoft/vscode@a6e49120535ba3e92fb6f538f5bf0fb6516861ba` | 2026-04-10 | https://github.com/microsoft/vscode/pull/308967 | Adds `validateBuiltinExtensionEnabledWithAutoUpdates()` to `extensionManagementCLI.ts`; CLI install path for builtins in the new array now routes through gallery validation. |

The consequence: any builtin id listed in `product.json[builtInExtensionsEnabledWithAutoUpdates]` is no longer satisfied by the local system extensions short-circuit. Its install path becomes

```
ExtensionManagementCLI.installExtensions
  → installVSIXOrGalleryExtension
  → extensionManagementService.installFromGallery
  → extensionGalleryService.getExtensions(...)
  → await this.extensionGalleryManifestService.getExtensionGalleryManifest()
  → ExtensionGalleryManifestIPCService.getExtensionGalleryManifest()
  → await this.barrier.wait()   ← never returns in --start-server mode
```

The `validateBuiltinExtensionEnabledWithAutoUpdates` guard added in `a6e49120` (PR #308967) does an early `return false` in the not-applicable case; combined with the new gallery path it produces exactly the failure signature we see: one `Installing extensions...` log line, zero extensions written, no error surfaced, server otherwise healthy.

## End-to-end validation

We reproduced both halves of the timeline in an isolated harness:

- **Unpatched 1.122 code-server + `@microsoft/vscode-dev-containers-cli@0.95.0`:** 0 extensions installed, `.installExtensionsMarker` left behind, single `Installing extensions...` log.
- **Same code-server + the same CLI build-time-patched to split into two sequential invocations and clean the marker on failure:** 3 extensions installed, two sequential `code-server` invocations visible in process logs, populated `extensions.json`.

This is consistent with the hypothesis above and rules out anything CLI-side as a *trigger* (the CLI's combined-invocation shape is unchanged from earlier releases — it's just the amplifier that turns the one-time hang into a permanent "no extensions on every boot" state via the eager-create `.installExtensionsMarker`).

## Verification you can do faster than me

Two confirmations would close any remaining ambiguity:

1. **`product.json` overlap:** confirm that the build of VS Code Server distributed to GitHub Codespaces (and to anyone else using the auto-update-builtins feature) actually lists the affected extension ids in `product.json[builtInExtensionsEnabledWithAutoUpdates]`. If the list is empty for a given distribution, the visibility cliff for that distribution will be a different commit in the same train (likely still in #308756 or #308967).
2. **Bisect:** revert PR #308967 on top of 1.122 and re-run the combined-invocation repro. We expect the hang to disappear (since builtins would once again resolve from the system extensions path) even though the underlying barrier defect from PR #243580 remains.

## Proposed fixes

In order of "smallest patch that closes the reported failure mode" to "fixes the latent defect for everyone":

### 1. Make the 1.122 silent-drop loud (immediate, smallest blast radius)

In `src/vs/platform/extensionManagement/common/extensionManagementCLI.ts:182-195, 306-326, 385-395` (added by `a6e49120` / PR #308967), `validateBuiltinExtensionEnabledWithAutoUpdates()`'s early-return path currently logs at info-level and silently returns `false`. It should log at **error** level and propagate a non-zero exit so cs-agent (and any other server orchestrator) sees a failure rather than a successful no-op. This alone would have made this regression a P0 the first time it hit dogfood instead of a slow telemetry trickle.

### 2. Make the IPC barrier non-fatal for headless flows (fixes the latent defect)

In `src/vs/platform/extensionManagement/common/extensionGalleryManifestServiceIpc.ts`, replace the bare `Barrier` with `AutoOpenBarrier(<grace-ms>)` (5 s seems generous) and, on timeout, fall back to constructing the manifest from `productService.extensionsGallery` exactly as `ExtensionGalleryManifestService` does. This preserves IPC semantics for normal interactive remote sessions (where the workbench is going to push the manifest in well under 5 s) while making any headless install path self-healing.

### 3. Conditional registration in `serverServices.ts` (belt-and-braces)

In `src/vs/server/node/serverServices.ts` (currently around line 209), conditionally bind `IExtensionGalleryManifestService` to `ExtensionGalleryManifestService` (direct) instead of `ExtensionGalleryManifestIPCService` when `args['install-extension']` or `args['install-builtin-extension']` is non-empty. This is the narrowest possible fix that targets exactly the broken combined-invocation case without changing semantics for the interactive remote path.

(2) and (3) are not mutually exclusive — (3) eliminates the race entirely for the specific known-bad case; (2) is the safety net for future code paths that try to use the gallery before workbench connects.

## Downstream context (FYI, not asking for action here)

We have an in-repo build-time hot-patch and a defensive pre-clean of `.installExtensionsMarker` queued in github/codespaces-service#5810 as a temporary mitigation. The Dev Containers CLI side (splitting the combined invocation and cleaning the marker on failure) is in microsoft/vscode-remote-containers#1178. None of these substitute for the upstream fix proposed above — they just bound the blast radius until a fix ships.

Happy to pair on a PR for any of the three options if useful; the patch in (3) is two lines.

sandy081:
Fix default account init deadlock in desktop remote connections.

In desktop remote connections, extensions are only registered after the remote connection is established. Default account initialization awaited `whenInstalledExtensionsRegistered()`, which could deadlock init.

This waits until the default account authentication provider is available (declared via `declaredProviders` or registered via `isAuthenticationProviderRegistered`), listening to both `onDidChangeDeclaredProviders` and `onDidRegisterAuthenticationProvider`. To make the provider available in remote scenarios, the provider's extension is explicitly activated via `getSessions(providerId, undefined, {}, true)` (gated on `remoteAuthority`). The wait remains bounded by `whenInstalledExtensionsRegistered()` as a fallback.

Fixes microsoft/vscode-remote-release#11688

### DIFF
--- a/src/vs/workbench/services/accounts/browser/defaultAccount.ts
+++ b/src/vs/workbench/services/accounts/browser/defaultAccount.ts
@@ -8,8 +8,8 @@ import { Barrier, RunOnceScheduler, ThrottledDelayer, timeout } from '../../../.
 import { CancellationToken } from '../../../../base/common/cancellation.js';
 import { ICopilotTokenInfo, IDefaultAccount, IDefaultAccountAuthenticationProvider, IEntitlementsData, IPolicyData } from '../../../../base/common/defaultAccount.js';
 import { getErrorMessage } from '../../../../base/common/errors.js';
-import { Emitter } from '../../../../base/common/event.js';
-import { Disposable } from '../../../../base/common/lifecycle.js';
+import { Emitter, Event } from '../../../../base/common/event.js';
+import { Disposable, DisposableStore } from '../../../../base/common/lifecycle.js';
 import { equals } from '../../../../base/common/objects.js';
 import { isWeb } from '../../../../base/common/platform.js';
 import { IDefaultChatAgent } from '../../../../base/common/product.js';
@@ -359,12 +359,11 @@ class DefaultAccountProvider extends Disposable implements IDefaultAccountProvid
 			return;
 		}
 
-		try {
-			await this.extensionService.whenInstalledExtensionsRegistered();
-			this.logService.debug('[DefaultAccount] Installed extensions registered.');
-		} catch (error) {
-			this.logService.error('[DefaultAccount] Error while waiting for installed extensions to be registered', getErrorMessage(error));
-		}
+		// Wait until the default account authentication provider is available instead of
+		// waiting for all installed extensions to be registered. In desktop remote
+		// connections extensions are only registered after the connection is established,
+		// so waiting for `whenInstalledExtensionsRegistered` can deadlock initialization.
+		await this.whenDefaultAccountAuthenticationProviderAvailable();
 
 		this.logService.debug('[DefaultAccount] Starting initialization');
 		await this.doUpdateDefaultAccount();
@@ -419,6 +418,51 @@ class DefaultAccountProvider extends Disposable implements IDefaultAccountProvid
 				this.refetchDefaultAccount();
 			}
 		}));
+	}
+
+	private async whenDefaultAccountAuthenticationProviderAvailable(): Promise<void> {
+		const provider = this.getDefaultAccountAuthenticationProvider();
+		const isAvailable = () => {
+			return this.authenticationService.declaredProviders.some(p => p.id === provider.id)
+				|| this.authenticationService.isAuthenticationProviderRegistered(provider.id);
+		};
+
+		if (isAvailable()) {
+			this.logService.debug('[DefaultAccount] Default account authentication provider is already available.');
+			return;
+		}
+
+		// Resolve as soon as the default account authentication provider is declared or
+		// registered, but wait no longer than installed extensions being registered.
+		this.logService.debug('[DefaultAccount] Waiting for default account authentication provider to be available.');
+		const disposables = new DisposableStore();
+		try {
+			await new Promise<void>(resolve => {
+				disposables.add(Event.any(this.authenticationService.onDidChangeDeclaredProviders, this.authenticationService.onDidRegisterAuthenticationProvider)(() => {
+					if (isAvailable()) {
+						disposables.dispose();
+						this.logService.debug('[DefaultAccount] Default account authentication provider is now available.');
+						resolve();
+					}
+				}));
+
+				// Explicitly activate the provider's extension so that the authentication
+				// provider gets registered. In desktop remote connections extensions are only
+				// registered after the connection is established, so without this the provider
+				// would never become available.
+				if (this.environmentService.remoteAuthority) {
+					this.authenticationService.getSessions(provider.id, undefined, {}, true);
+				}
+
+				this.extensionService.whenInstalledExtensionsRegistered().finally(() => {
+					disposables.dispose();
+					this.logService.debug('[DefaultAccount] Installed extensions registered.');
+					resolve();
+				});
+			});
+		} finally {
+			disposables.dispose();
+		}
 	}
 
 	async refresh(options?: { forceRefresh?: boolean }): Promise<IDefaultAccount | null> {
--- a/src/vs/workbench/services/accounts/browser/defaultAccount.ts
+++ b/src/vs/workbench/services/accounts/browser/defaultAccount.ts
@@ -427,20 +427,22 @@ class DefaultAccountProvider extends Disposable implements IDefaultAccountProvid
 				|| this.authenticationService.isAuthenticationProviderRegistered(provider.id);
 		};
 
-		if (isAvailable()) {
-			this.logService.debug('[DefaultAccount] Default account authentication provider is already available.');
-			return;
-		}
-
-		// Resolve as soon as the default account authentication provider is declared or
-		// registered, but wait no longer than installed extensions being registered.
 		this.logService.debug('[DefaultAccount] Waiting for default account authentication provider to be available.');
 		const disposables = new DisposableStore();
 		try {
 			await new Promise<void>(resolve => {
+				// Check if the provider is available.
+				// If available, resolve immediately. Otherwise, wait for it to be declared or registered.
+				if (isAvailable()) {
+					this.logService.debug('[DefaultAccount] Default account authentication provider is now available.');
+					resolve();
+					return;
+				}
+
+				// Resolve as soon as the default account authentication provider is declared or
+				// registered, but wait no longer than installed extensions being registered.
 				disposables.add(Event.any(this.authenticationService.onDidChangeDeclaredProviders, this.authenticationService.onDidRegisterAuthenticationProvider)(() => {
 					if (isAvailable()) {
-						disposables.dispose();
 						this.logService.debug('[DefaultAccount] Default account authentication provider is now available.');
 						resolve();
 					}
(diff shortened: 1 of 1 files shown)


src/vs/workbench/services/accounts/browser/defaultAccount.ts
```diff
@@ -421,6 +420,56 @@ class DefaultAccountProvider extends Disposable implements IDefaultAccountProvid
...
+				// Explicitly activate the provider's extension so that the authentication
+				// provider gets registered. In desktop remote connections extensions are only
+				// registered after the connection is established, so without this the provider
+				// would never become available.
+				if (this.environmentService.remoteAuthority) {
+					void this.authenticationService.getSessions(provider.id, undefined, {}, true);
+				}
+
+				this.extensionService.whenInstalledExtensionsRegistered().then(() => {
+					disposables.dispose();
```
dmitrivMS, line 460: this is already present in finally block

src/vs/workbench/services/accounts/browser/defaultAccount.ts
```diff
@@ -421,6 +420,56 @@ class DefaultAccountProvider extends Disposable implements IDefaultAccountProvid
...
+						resolve();
+					}
+				}));
+
+				// Explicitly activate the provider's extension so that the authentication
+				// provider gets registered. In desktop remote connections extensions are only
+				// registered after the connection is established, so without this the provider
+				// would never become available.
+				if (this.environmentService.remoteAuthority) {
+					void this.authenticationService.getSessions(provider.id, undefined, {}, true);
```
dmitrivMS, line 456: AI feedback on this line:

getSessions(..., activateImmediate=true) is bounded internally (it rejects after ~5s via tryActivateProvider's raceTimeout), but the rejection is void-discarded here. So if remote activation fails, no auth event fires, arm 3 never resolves, and initBarrier stays shut — every getDefaultAccount() then awaits forever, which is effectively the same deadlock symptom we're trying to fix, just from a different cause.

Two small options:
1) attach a handler so a failed activation still unblocks init:
(logging only — the whenInstalledExtensionsRegistered() arm is meant to be the bound, but it doesn't hold on this path), or
2) add an explicit bounded escape via the already-imported timeout(...) as a third race arm, so init can always make progress regardless of activation outcome.

connor4312 (approved):

PR #321016, opened 2026-06-11
Wire up MCP App support for agent-host sessions

connor4312:
Wires up MCP App support through the agent host pipeline so MCP servers exposed by agent-host sessions (currently the Copilot CLI agent) feed the chat UI's MCP App webviews end-to-end.

## What's new

**MCP Apps over the agent host protocol.** Tool calls routed through agent-host MCP servers now surface their MCP App metadata (the `extensions.io.modelcontextprotocol/ui` extension on a tool result) into the same `ChatMcpAppModel` plumbing that powers App UIs for locally-running MCP servers. As a result, an MCP server like `@modelcontextprotocol/server-basic-vanillajs` spawned by the Copilot CLI now renders its App webview in chat just like a server configured directly in VS Code.

## Highlights

- `COPILOT_MCP_APPS=true` env override is set when spawning the Copilot CLI subprocess (`copilotAgent.ts`). Without this runtime flag, the CLI silently drops the `enableMcpApps: true` session option, so MCP servers never see the App-capability extension and tools never come back tagged with App metadata.
- Agent-host tool-call adapter (`stateToProgressAdapter.ts`) now populates `state.resultDetails.mcpOutput` for MCP App tools by translating our AHP `ToolResultContent[]` into an `MCP.CallToolResult`:
  - Text → `TextContent`
  - EmbeddedResource (image/audio) → `ImageContent` / `AudioContent`
  - Other EmbeddedResource → `EmbeddedResource` wrapping a synthetic `data:` URI
  - Resource refs → `ResourceLink`
- Resource URIs in tool results are now consistently wrapped via `toAgentHostUri(uri, connectionAuthority)` so remote agent host resources resolve through the agent host filesystem provider on the client. This fixes a pre-existing inconsistency where the `output: [{ type: 'ref', uri }]` ref produced raw remote URIs.
- `resultDetails` is no longer dropped when `toolSpecificData` carries MCP App input data — the gating condition was widened so MCP-app tool invocations get details populated for both the serialized (history) and live (active-turn) paths.

## Testing

- `npm run compile-check-ts-native` clean
- `stateToProgressAdapter.test.ts` (63 tests) pass, including the updated assertion that `Resource` output refs are wrapped via `toAgentHostUri`
- Verified end-to-end against a self-contained SDK repro (`vscode-team-kit/mcp-apps-sdk-repro`) and in VS Code itself by invoking an MCP App tool through a Copilot CLI agent-host session and confirming the App webview receives `ui/notifications/tool-result` with real content.


justschen (approved):

src/vs/workbench/contrib/mcp/browser/mcpToolCallUI.ts
```diff
@@ -28,9 +31,202 @@ export interface IMcpAppResourceContent extends McpApps.McpUiResourceMeta {
...
+		return readResourceContentToHtml(result.contents);
+	}
+
+	async callTool(name: string, params: Record<string, unknown>, _token: CancellationToken): Promise<MCP.CallToolResult> {
+		const result = await this._agentHostService.handleMcpRequest(this._channel, 'tools/call', { name, arguments: params }) as MCP.CallToolResult;
+		return result;
+	}
+
+	async readResource(uri: string, _token: CancellationToken): Promise<MCP.ReadResourceResult> {
+		const result = await this._agentHostService.handleMcpRequest(this._channel, 'resources/read', { uri }) as MCP.ReadResourceResult;
```
dmitrivMS, line 213: Should this take cancellation token to avoid hanging requests?

src/vs/platform/agentHost/browser/remoteAgentHostProtocolClient.ts
```diff
@@ -840,6 +840,14 @@ export class RemoteAgentHostProtocolClient extends Disposable implements IAgentC
 		return await this._sendRequest('invokeChangesetOperation', params);
 	}
 
+	/**
+	 * Send a request on an `mcp://` AHP side channel. The agent-host
+	 * routes by `params.channel` so we inject it automatically.
+	 */
+	async handleMcpRequest(channel: string, method: string, params: Record<string, unknown> | undefined): Promise<unknown> {
+		return await this._dispatchRequest<unknown>(method, { ...(params ?? {}), channel });
```
dmitrivMS, line 848: If MCP method has a `channel` parameter, we will override it...

src/vs/workbench/contrib/chat/browser/widget/chatContentParts/toolInvocationParts/chatMcpAppModel.ts
```diff
@@ -56,7 +57,18 @@ export type McpAppLoadState =
...
+	/**
+	 * In-memory origin map for agent-host MCP servers. Agent-host server
+	 * ids embed the session id, so they're effectively single-use across
+	 * VS Code restarts — using {@link WebviewOriginStore} for them would
+	 * accumulate one persisted entry per agent-host session forever. The
+	 * in-memory map keeps origins stable for the lifetime of the app
+	 * (enough for webview state to persist across re-renders) without
+	 * touching application storage.
+	 */
+	private static readonly _agentHostOrigins = new Map<string, string>();
```
dmitrivMS, line 69: (minor) Perhaps we can find a way to make this per session so it gets cleaned up eventually?

src/vs/workbench/contrib/chat/browser/agentSessions/agentHost/stateToProgressAdapter.ts
```diff
@@ -550,9 +592,85 @@ function getToolInputOutputDetails(tc: ToolCallState, isError: boolean, errorStr
...
+			}
+			if (block.contentType.startsWith('audio/')) {
+				return { type: 'audio', data: block.data, mimeType: block.contentType };
+			}
+			return {
+				type: 'resource',
+				resource: {
+					uri: `data:${block.contentType};base64,${block.data}`,
+					mimeType: block.contentType,
+					blob: block.data,
```
dmitrivMS, line 646: Since we already have it inside `uri`, is it still necessary here? If blob is large, this may matter...

PR #321133, opened 2026-06-12
Add annotations support for agent feedback

benibenj:
This work migrates the agents window "feedback" (code-review comments) feature for agent-host sessions only off the client and onto the agent host protocol: instead of holding feedback in the client-side `AgentFeedbackService` in-memory maps, feedback now lives as server-held state in the protocol's annotations channel, with feedback-specific semantics (kind, state, replies, code selection, diff hunks, PR source id) carried in the annotation `_meta` field and written via the existing `annotations/set` upsert action so no protocol change is required. The four feedback tools (`addComment`, `listComments`, `deleteComments`, `resolveComments`) move from client tools to agent-host **server tools** that execute in-process against the session's annotation state (wired end-to-end for the Copilot agent), the legacy submit-time feedback attachment is removed so the agent simply reads its own annotations channel, and all of this is scoped strictly to agent-host providers (`local-agent-host`/`agenthost-*`) so other providers like copilotchatcli keep their existing client-side behavior unchanged.


TylerLeonhardt (dismissed):
This PR says that it's for Agent Host sessions but there are no changes in Claude or Codex agents. What is the experience for those?

amunger (approved):

PR #321152, opened 2026-06-12
Fix question carousel stealing focus from other VS Code UI

### LINKED ISSUE #321063: Question widget steals focus
- Have focus somewhere else in vscode UI, typing
- Question widget pops up in chat
- Suddenly my focus is there and I'm typing in the wrong place

meganrogge:
Fixes #321063

## Problem
The question carousel widget was stealing focus from other VS Code UI (editor, terminal, etc.) when it appeared, even if the user was actively working elsewhere.

## Solution
Improved the auto-focus logic to only auto-focus the question carousel when **both** conditions are met:
1. Focus is already within the chat widget (`dom.isAncestorOfActiveElement(widget.domNode)`)
2. Chat input is empty

This allows helpful auto-focus when the user is actively in chat, while preventing focus theft when they're working in the editor or other VS Code areas.

## Changes
- Modified `renderQuestionCarousel()` in `src/vs/workbench/contrib/chat/browser/widget/chatListRenderer.ts` to check if focus is within the chat widget before auto-focusing
- Added comments explaining the focus prevention logic
- Screen reader mode users are still protected via existing accessibility check in `ChatQuestionCarouselPart._shouldAutoFocus()`

## Testing
The fix passes TypeScript compilation. Manual testing recommended for:
- Question carousel appearing when focus is in the editor → should NOT steal focus
- Question carousel appearing when focus is in chat input → should auto-focus carousel
- Screen reader mode behavior (should not auto-focus)


justschen (approved):
thanks! i think this is fine yeah!

PR #321176, opened 2026-06-12
Improve terminal tool sync guidance and idle silence timeout

meganrogge:
## Summary

Closes microsoft/vscode-copilot-evaluation#4256

This PR addresses part of the performance gap between VS Code terminal tools and Copilot CLI on terminalbench2 by reducing unnecessary 2-step tool call patterns (`run_in_terminal` → `get_terminal_output`).

### Root Cause
VS Code's `run_in_terminal` promotes sync commands to background after 60s of idle silence, forcing the model to poll with `get_terminal_output`. For compilation/build tasks that regularly have long periods of no output, this creates a 2-step pattern where Copilot CLI's `bash` tool completes in a single step.

### Changes

1. **Increase default idle silence timeout from 60s to 300s (5 min)**
   - Compilation tasks (e.g. compile-compcert) regularly have >60s gaps between output
   - 300s covers most build steps while still catching genuinely hung processes
   - Extracted `DEFAULT_IDLE_SILENCE_TIMEOUT_MS` constant shared between config default and runtime fallback

2. **Rewrite tool model descriptions to strongly bias toward sync mode**
   - Renamed "Async Mode" section to "Execution Mode" with sync-first guidance
   - Explicit: "For ALL one-shot commands, use mode='sync' and omit timeout"
   - Clarify sync output is final — no need to call `get_terminal_output` afterward

3. **Update `get_terminal_output` description**
   - Clarify it should ONLY be used when `run_in_terminal` explicitly says the command was moved to background

4. **Consistent timeout guidance**
   - "Usually omit entirely for sync commands" instead of suggesting generous values
   - Only set timeout as a safety net for suspected hangs

### Expected Impact
- Fewer `get_terminal_output` calls (currently 80+ per eval run)
- More commands completing in a single sync tool call
- Better success rate on compilation/build-heavy benchmark tasks

### DIFF
--- a/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/getTerminalOutputTool.ts
+++ b/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/getTerminalOutputTool.ts
@@ -20,7 +20,7 @@ export const GetTerminalOutputToolData: IToolData = {
 	toolReferenceName: 'getTerminalOutput',
 	legacyToolReferenceFullNames: ['runCommands/getTerminalOutput'],
 	displayName: localize('getTerminalOutputTool.displayName', 'Get Terminal Output'),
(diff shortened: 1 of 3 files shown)


meganrogge:

dmitrivMS (approved):

PR #321177, opened 2026-06-12
Fix Linux sandboxing

chrmarti:
#317981


alexr00 (approved):
I'm seeing the string `COPILOT_CLI_ENABLED_FEATURE_FLAGS` repeated. Would be nice to have this a const somewhere instead.

chrmarti:
The string is used in the extension and the agent host which don't currently share code I think.

PR #321186, opened 2026-06-12
security: harden GitHub Actions workflows

shunfeng8421:


Yoyokrazy:
permissions can only be reduced or maintained from the caller workflow, so since pr.yml already has this, this PR is effectively a noop.

PR #321219, opened 2026-06-12
Kill and restart esbuild instances during watch mode

mjbvz:
Reduces memory usage since we don't need rebuilds all that often. Perf still is reasonable in for the full rebuild


juttadnanzafar-hue (commented):
jutt

alexdima (dismissed):

aeschli (approved):

PR #321225, opened 2026-06-12
Fix Agents Window smoke test: use runCommand instead of keybinding

alexdima:
## Problem

The `NewChatInSessionsWindowAction` keybinding (`Ctrl+L`) is gated on `EditorAreaFocusContext.negate()`, meaning it only fires when focus is **not** in the editor area. After `warmUpClaudeModel` interacts with the chat editor (click → type → send), focus remains in the editor area, so the subsequent `ctrl+l` dispatch is silently ignored. This caused both "Test Claude session" and "Test Local session" smoke tests to time out waiting for `.sessions-chat-widget .new-chat-widget-container`.

## Fix

1. **Add `f1: true`** to `NewChatInSessionsWindowAction` so the command is accessible via the command palette (the action is only registered in the Agents Window context via `sessions.common.main.ts`)
2. **Change `startNewSession()`** in the smoke test automation to use `quickaccess.runCommand('workbench.action.sessions.newChat')` instead of `dispatchKeybinding('ctrl+l')`, bypassing the focus gate entirely

## Build reference

Investigated from [build 447362](https://dev.azure.com/monacotools/Monaco/_build/results?buildId=447362) — macOS arm64 Electron smoke tests failed.

### DIFF
--- a/src/vs/sessions/contrib/chat/browser/chat.contribution.ts
+++ b/src/vs/sessions/contrib/chat/browser/chat.contribution.ts
@@ -48,6 +48,7 @@ class NewChatInSessionsWindowAction extends Action2 {
 			id: NEW_SESSION_ACTION_ID,
 			title: localize2('chat.newEdits.label', "New Chat"),
 			category: CHAT_CATEGORY,
+			f1: true,
 			keybinding: {
 				weight: KeybindingWeight.WorkbenchContrib + 2,
 				// Don't shadow Ctrl/Cmd+N (and Ctrl/Cmd+L) when focus is in the
--- a/test/automation/src/agentsWindow.ts
+++ b/test/automation/src/agentsWindow.ts
@@ -36,14 +36,15 @@ export class AgentsWindow {
 	}
 
 	/**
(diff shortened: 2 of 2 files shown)


dmitrivMS (approved):

src/vs/sessions/contrib/chat/browser/chat.contribution.ts
```diff
@@ -48,6 +48,7 @@ class NewChatInSessionsWindowAction extends Action2 {
 			id: NEW_SESSION_ACTION_ID,
 			title: localize2('chat.newEdits.label', "New Chat"),
 			category: CHAT_CATEGORY,
+			f1: true,
```
jruales, line 51: Is it required to have `f1: true` for the command to be accessed via runCommand()? Or do we just also want to have it in the command palette as well?

alexdima: yes, the smoke test utility goes through the command palette. This was also missing in the command palette which was weird to begin with.

jruales (approved):

PR #321233, opened 2026-06-12
agent host edit survival tracking

amunger:
https://github.com/microsoft/vscode-internalbacklog/issues/7876

- Uses of the existing editTracker that is used for showing file diffs in the chat session.
- Survival compares the file at edit-time against the same file 30s / 2min / 8min / 15min later. Two scores per sample:
  - fourGram: similarity of 4-character n-grams between after-edit and current text.
  - noRevert: is the current text closer to before or after the agent's edit.
- Whole-file vs. chunked. Attempt region-based scoring when we can identify exactly which chunks the AI inserted which is more accurate when the user makes unrelated edits in the same file, but whole-file is the fallback if we are unable to parse the regions.
  - All edits should be tracked in chunks at this point, but tool parameter shape is not enforced, so we will start seeing the changes tracked at the whole-file level if they drift over time.
- 5 MB file cap to avoid any big perf hits for unusually large files.
- skip ipynb for now for simplicity.


javi0guerrero4-cloud (commented):
New

DonJayamanne (approved):

PR #321239, opened 2026-06-12
Remove advanced tag from browser remote proxy setting

### LINKED ISSUE #293229: Run Integrated Browser in Remote environment
The integrated browser for me would be extremely useful if it ran in the remote environment like SSH.

It would allow me to access servers that can only be accessed from within the remote environment without having to rely on some other remote desktop tool or even gimmicks like sshuttle.

kycutler:


Yoyokrazy (approved):

jruales (approved):

src/vs/workbench/contrib/browserView/electron-browser/features/browserRemoteFeatures.ts
```diff
@@ -94,7 +79,7 @@ Registry.as<IConfigurationRegistry>(ConfigurationExtensions.Configuration).regis
 		[BrowserRemoteProxyEnabledSettingId]: {
 			type: 'boolean',
 			default: false,
-			tags: ['advanced', 'experimental'],
+			tags: ['experimental'],
 			scope: ConfigurationScope.WINDOW,
```
jruales, line 83: Thoughts on using `ConfigurationScope.APPLICATION`?

kycutler: The proxy is already limited to workspace / ephemeral sessions only, so it won't actually span multiple main windows -- and I think it's useful to let workspaces configure it independently.

PR #321243, opened 2026-06-13
In agents window, open browser on secondary side bar by default (instead of modal)

jruales:
* Fixes https://github.com/microsoft/vscode-internalbacklog/issues/7703


src/vs/workbench/contrib/browserView/electron-browser/browserEditorGroup.ts
```diff
@@ -0,0 +1,27 @@
...
+ *
+ * When the workbench forces editors into a modal part
+ * (`workbench.editor.useModal: 'all'`, the default in the Agents window), the
+ * integrated browser should still dock in the main editor area rather than open
+ * as a modal overlay. This redirects browser opens that target the active group
+ * to the main editor part, while leaving explicit placements (side group,
+ * auxiliary window, a specific group) untouched. Users can still move the
+ * browser into a modal explicitly afterwards.
+ */
+export function getBrowserPreferredGroup(editorGroupsService: IEditorGroupsService, configurationService: IConfigurationService, preferredGroup?: PreferredGroup): PreferredGroup | undefined {
```
kycutler: Is there another file we can put this in? I don't like it being a whole new file just for the one tiny function.

kycutler: Actually, as a method on the workbench service would probably be good.

kycutler (approved):

PR #321244, opened 2026-06-13
Use a temp subdirectory for createSession when workingDirectory is omitted

### LINKED ISSUE #320454: Agent host in editor with no workspace open no-ops
1. Open VS Code without a workspace
2. Select Agent host
3. Send a message
4. (it no-ops)

dmitrivMS:
Fixes #320454


rzhao271 (changes requested):
Would making the fallback directory the home directory result in the session reading or writing files to the home directory? I feel that I'd prefer a temp directory or a notice that I haven't set a working directory over this fallback.

dmitrivMS:
> Would making the fallback directory the home directory result in the session reading or writing files to the home directory? I feel that I'd prefer a temp directory or a notice that I haven't set a working directory over this fallback.

So I thought as well, but then we are doing that in other places in Agent Host:
https://github.com/microsoft/vscode/blob/bf2334eeee140e2e549c9392ca7b59f78d44277a/src/vs/platform/agentHost/node/agentHostMain.ts#L273
https://github.com/microsoft/vscode/blob/bf2334eeee140e2e549c9392ca7b59f78d44277a/src/vs/platform/agentHost/node/agentHostMain.ts#L407

@connor4312 Any thoughts?

dmitrivMS:
@rzhao271 FWIW, it won't write temp session files there - there's a separate place for that, so it's just a CWD.

rzhao271 (dismissed):

dmitrivMS:
We decided to go with the temp dir in agent host sync, so I'll be making an update.
Probably need a dedicated dir created for that purpose to avoid exposing all of temp by default.

dmitrivMS:
@rzhao271 Ok, ready now.

Yoyokrazy (approved):

Issue #321276, opened 2026-06-13
Call Stack view layout and resizing bugs when docked to Explorer or Panel


ZeroDayZ7:
Version: 1.124.2 
- OS Version: Windows 10

Steps to Reproduce:

https://github.com/user-attachments/assets/49de5ec3-f52e-40ff-babd-1bb992d67097

1. Start debugging any project to make the `Call Stack` view active.
2. Drag the `Call Stack` view into the Explorer side bar.
3. Try to resize it to its minimum height (it gets stuck and wastes space).
4. Open multiple editor tabs, then click the "Close All" button — notice the `Call Stack` automatically layout-shifts and expands.
5. Drag `Call Stack` into the Panel section next to the Terminal (e.g., right or bottom panel) and try to collapse it. It hides the content but leaves an empty header and lacks an icon.

### Description:

**Bug 1: Call Stack in Explorer Viewlet cannot be minimized & auto-expands**
* When the `Call Stack` view is docked into the Explorer Side Bar, it cannot be resized down to its minimum height.
* If you have multiple editor tabs open and close them all at once using the "Close All" button, the `Call Stack` view automatically auto-expands and takes over the vertical space.

**Bug 2: Call Stack in Panel cannot be collapsed to an icon**
* When moving the `Call Stack` view to the Panel section (next to the Terminal) and attempting to minimize it, it's impossible to collapse it to just an icon.
* Using the built-in collapse/chevron button hides the contents but leaves the header section blank, completely lacking a dedicated icon representation when collapsed.

Soumilgit:
Hi team,

This issue looks unassigned right now. I investigated the Call Stack view layout and resizing behavior when docked outside the debug panel and have opened a PR, https://github.com/microsoft/vscode/pull/321279, with a fix plus unit-test coverage.

Could you please assign issue #321276 to me so I can drive it to completion (address feedback, iterate on the patch if needed, and follow through until it’s merged)? Thanks!

PR #321279, opened 2026-06-13
Fix Call Stack view layout when docked outside debug panel

Soumilgit:
This PR fixes Call Stack view layout and resizing bugs when docked to Explorer or Panel (Issue #321276).

1. Computes dynamic min/max body sizes from tree content so the view can shrink in the Explorer and does not steal space when neighboring views resize.
2. Hides the header state message when collapsed in the panel so the view icon remains visible.


Soumilgit:

Soumilgit:
Hi @federicobrancasi can you please review the PR

PR #321284, opened 2026-06-13
fix: preserve CRLF in stage selected ranges

### LINKED ISSUE #96104: "Stage selected ranges" clobbers line endings for the entire file without any warning (Windows)
- VSCode Version: Tested on 1.44.2 and 1.45.0 insiders
- OS Version: Windows 10 Build 18363

Steps to Reproduce:

1. Make any change in a file which has CRLF line endings in the git repository

image

2. Stage the change from the git diff UI in vscode using "stage selected ranges"

image

3. use git diff --staged to see the diff of the file (making sure diff.wsErrorHighlight is set to "all", otherwise you still can't see the changes), and see that the line endings for the entire file have now changed:
image

image

This is especially insidious if you happen to have core.autocrlf set to true, because the files will continue to look identical on disk when checked out, they will only differ in the repo itself (and when you go to submit a PR, it looks like the whole file changed, but github doesn't identify that the change is in the line endings).

Also, the git diff UI in VS Code will _not_ show the differences, _even with Ignore Trim Whitespace_ set to false, which means the user has no idea that his line endings are being clobbered, potentially for weeks or months.

Does this issue occur when all extensions are disabled?: Yes, tested with a fresh install of latest stable (1.44.2) with no extensions

Ijtihed:
Fixes #96104

applyLineChanges() stitches text from two TextDocuments that can have different EOLs. The original from git show uses LF and while the working tree uses CRLF with autocrlf=true which produces ixed line endings. Also git hash-object --path applies the autocrlf clean filter which converts all CRLF to LF in the stored blob even for lines not in the selection.

Changes

- extensions/git/src/eol.ts: normalizeEOL() helper to convert line endings without double-converting
- extensions/git/src/staging.ts: applyLineChanges() normalizes modified chunks to match the original document's EOL
- extensions/git/src/commands.ts: diffStageHunkOrSelection() normalizes originalWithModifiedChanges the same way
- extensions/git/src/git.ts: hash-object uses -c core.autocrlf=false to skip EOL conversion while preserving other filters (LFS, clean/smudge)

Testing

- 14 unit tests for normalizeEOL(), all passing
- 9 integration tests for applyLineChanges() covering EOL mismatch, insertion, deletion, and no trailing newlin

### DIFF
--- a/extensions/git/src/commands.ts
+++ b/extensions/git/src/commands.ts
@@ -12,7 +12,8 @@ import { ForcePushMode, GitErrorCodes, RefType, Status } from './api/git.constan
 import { Git, GitError, Repository as GitRepository, Stash, Worktree } from './git';
 import { Model } from './model';
 import { GitResourceGroup, Repository, Resource, ResourceGroupType } from './repository';
-import { DiffEditorSelectionHunkToolbarContext, LineChange, applyLineChanges, getIndexDiffInformation, getModifiedRange, getWorkingTreeDiffInformation, intersectDiffWithRange, invertLineChange, normalizeEOL, toLineChanges, toLineRanges, compareLineChanges } from './staging';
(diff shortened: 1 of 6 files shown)


Ijtihed:
hey @lszomoru :)

Issue #316841, opened 2026-05-17
Webview extensions blank on Safari/WebKit (vscode.dev / code-server / code serve-web): VSBuffer.slice returns view that gets detached by extension-host postMessage transfer


dop-amine:
Does this issue occur when all extensions are disabled?: Yes

### VS Code Version

Reproduced against code-server `4.106.x` (vendoring `vscode` `1.107.x`). Bug exists in current `microsoft/vscode@main` per source inspection of `src/vs/base/common/buffer.ts` and `src/vs/base/parts/ipc/common/ipc.net.ts`. Per the introducing commit (microsoft/vscode#76076) it has been present since 2019.

### OS Version

- **Server**: Ubuntu 24.04, Node 20.x
- **Clients (reproduces)**: Safari 18 on macOS 15, Safari 18 on iPadOS 18, Chrome / Brave / Firefox on iOS (all forced WebKit by App Store policy)
- **Clients (does not reproduce)**: Chrome / Chromium / Brave / Edge on macOS / Linux / Windows (V8), Firefox desktop (SpiderMonkey)

### Does this reproduce in native VS Code desktop?

**No.** Native VS Code (Electron) uses Electron's IPC layer between renderer and extension host — not browser `postMessage` with transfer lists for `VSBuffer` payloads. The underlying `ArrayBuffer` is never detached in that flow, so `ChunkStream`'s held subarray views stay valid and `VSBuffer.slice`'s view-returning behavior is correct. The bug is specifically a browser-mode VS Code issue.

### Does this reproduce in `code serve-web`?

**Yes (architecturally identical, not directly stood up).** `code serve-web` ships the same `VSBuffer.slice` code path and the same browser-based web-worker extension host. Any Safari / iOS client of `code serve-web` should hit it identically. Reproduces deterministically in `code-server`, which vendors the same upstream code verbatim.

### Does this reproduce in `vscode.dev`?

**Likely yes.** I have not directly tested in Safari since `vscode.dev` is harder to drive a webview-heavy extension on (the marketplace surface is narrower). Long-standing issue #213143 ("Safari 14 no longer able to load vscode.dev") and adjacent reports point in the same direction. Same upstream code path.

### Does this reproduce in GitHub Codespaces?

**Likely yes** (Codespaces uses VS Code web). Not directly tested.

### Steps to Reproduce

1. Run any browser-mode VS Code (vscode.dev, `code serve-web`, code-server, Codespaces) over HTTPS — service workers must register.
[...]

PR #321291, opened 2026-06-13
Fix webview resource loading detach in Safari/WebKit

alexdima:
fyi @mjbvz this is for https://github.com/microsoft/vscode/issues/316841 -- looks like if this code got ahold of a slice of a larger buffer, it would send off the entire buffer... this would cause other code to throw when it tried to use other pieces of the larger buffer.

---
🤖 

## Context

Fixes the `Underlying ArrayBuffer has been detached from the view or out-of-bounds: subarray` error seen in web/remote mode on Safari/WebKit.

Related: microsoft/vscode#316841, microsoft/vscode#314126

## Problem

When transferable streams are unsupported (Safari/WebKit), webview resource chunks are sent via `postMessage` with a transfer list. The chunk was wrapped as a `Uint8Array` view over its **entire** underlying `ArrayBuffer`, and that whole `ArrayBuffer` was transferred:

```ts
const data = new Uint8Array(chunk.buffer.buffer, chunk.buffer.byteOffset, chunk.buffer.byteLength);
this._send('did-load-resource-chunk', { id, data }, [data.buffer]);
```

In web/remote mode the chunk is frequently a **view aliasing a larger, shared `ArrayBuffer`** (e.g. a message buffer sliced into many views by the IPC deserialize pipeline). Transferring that shared buffer detaches **every sibling view** still in use. WebKit detaches synchronously, so the next IPC read throws `Underlying ArrayBuffer has been detached from the view` and webview resource loading breaks. V8 (Chrome/Edge/desktop) detaches lazily, which is why this reproduces on Safari/iOS only.

## Fix

Copy the chunk into a freshly-owned `ArrayBuffer` (sized to exactly the chunk window) before transferring it, so the transfer can only ever detach the copy:

```ts
const data = chunk.buffer.slice();
this._send('did-load-resource-chunk', { id, data }, [data.buffer]);
```

`Uint8Array.prototype.slice` returns an exact-window copy backed by its own `ArrayBuffer`, leaving the shared source intact.

### DIFF
--- a/src/vs/workbench/contrib/webview/browser/webviewElement.ts
+++ b/src/vs/workbench/contrib/webview/browser/webviewElement.ts
@@ -876,7 +876,12 @@ export class WebviewElement extends Disposable implements IWebviewElement, Webvi
 						});
 						listenStream(result.stream, {
 							onData: (chunk) => {
-								const data = new Uint8Array(chunk.buffer.buffer, chunk.buffer.byteOffset, chunk.buffer.byteLength);
+								// Copy into a freshly-owned ArrayBuffer before transferring. `chunk`
+								// may be a view into a larger, shared ArrayBuffer (e.g. from the IPC
+								// deserialize pipeline); transferring its underlying ArrayBuffer would
+								// detach every sibling view. WebKit detaches synchronously, which
+								// previously broke webview resource loading in Safari.
+								const data = chunk.buffer.slice();
 								this._send('did-load-resource-chunk', { id, data }, [data.buffer]);
 							},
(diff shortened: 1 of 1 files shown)


er91:
Verified that this PR fixes the 2 distinct error paths in Safari, and VSCode extension webview works properly with it.

This PR seems more surgical and fixes both error paths with one change :-)

Closing #320190 in favor of this.

alexdima:
Given the feedback from @er91 let's get this in for testing.

pwang347 (approved):

PR #321358, opened 2026-06-15
Fix media preview cache-busting version

### LINKED ISSUE #320928: Webview resource cache grows unboundedly: media-preview cache-busts with Date.now(), service worker never evicts (21+ GB observed)
(media-preview and the webview service worker are built-in, so `--disable-extensions` does not affect the leak.)

- VS Code Version: 1.123.2 (commit 3c631b164c239e7aeaaae7c626b46c527b361af2, x64)
- OS Version: Windows 11 Pro 10.0.26200 (also observed under WSL Remote – Ubuntu)

## Summary

`%APPDATA%\Code\WebStorage\<n>\CacheStorage` grew to **23.5 GB** on my machine.
I parsed the Chromium simple-cache entry keys to find out why. The root cause is a
combination of two pieces of built-in code:

1. **media-preview cache-busts with the current time, not the file version.**
   `extensions/media-preview/src/imagePreview/index.ts` (`getResourcePath`) appends
   `?version=${Date.now()}` on **every render**. Every time an image is opened or
   re-rendered, the URL is brand new — even if the file is unchanged.

2. **The webview service worker caches every response and never evicts.**
   `src/vs/workbench/contrib/webview/browser/pre/service-worker.js` does
   `cache.put(event.request, ...)` for every 200 response with an ETag, keyed by the
   full URL including the query string. The file contains no eviction, expiry, or
   quota logic. The `activate` handler doesn't even delete caches from previous
   `VERSION`s — my disk holds both `vscode-resource-cache-4` (2.09 GB) and
   `vscode-resource-cache-5` (21.38 GB).

Combined effect: **every view of an image permanently writes a full copy to disk**,
and each copy is dead on arrival — the `Date.now()` query string guarantees no future
request can ever hit it.

## Evidence (from parsing the cache entry keys)

My 21.38 GB cache origin contains 6,413 entries:

- They map to only **1,947 unique files** (1,013 files cached ≥ 2 times).
- The most-viewed image was cached **86 times within a single hour** (one evening,
  up to 9 copies in one minute) — per-render, not per-modification.
- In a sample of the 200 most-duplicated paths, **81% point to files that no longer
  exist** on disk.
- Zero video/audio entries: `<video>` uses Range requests (206), which the SW marks
  `no-store` and the Cache API cannot store anyway. The leak therefore only bites
  image-heavy workflows and grows slowly enough to evade detection — which is
  presumably why earlier reports (#131226, #166632 [145 GB], #152519 [500 GB]) died
  as "cannot reproduce" / stale.

## Steps to Reproduce

1. Note the size of `%APPDATA%\Code\WebStorage` (`du -sh ~/.config/Code/WebStorage` on Linux).
2. Launch `code --disable-extensions` and open a large PNG (e.g. 20 MB) in the
   built-in image preview.
3. Close the tab and reopen the same file 10 times.
4. The folder grows by ~10 × the image size. None of the new entries can ever be
   read back (each was stored under a unique `?version=<timestamp>` URL), and they
   are never cleaned up.

## Expected

Either of:
- media-preview keys the cache-busting parameter on the file's mtime instead of
  `Date.now()`, so unchanged files reuse one cache entry; and/or
- the service worker evicts stale versions of the same base URL when caching a new
  one, applies an LRU/size cap to `vscode-resource-cache-*`, and deletes
  previous-`VERSION` caches on activate.

## Related

#131226 (closed as stale, "needs repro"), #166632 (145 GB, closed as duplicate),
#152519 (500 GB, closed as duplicate), #310384 (open, 30 GB via Remote-SSH — same
mechanism; this report adds the root-cause analysis).

pony-maggie:
Fixes #320928

Media previews currently append `Date.now()` to resource URIs on every render. In web, those resources are cached by the service worker using the full request URL, so reopening the same unchanged media file creates a new cache key each time.

This changes media previews to derive the cache-busting version from the resource stat (`mtime` and `size`) instead. The version stays stable while the resource is unchanged, but changes when the resource changes. If the stat cannot be read, the preview falls back to the resource identity so the webview can still render its normal load error instead of failing before HTML generation.

The shared helper is used by image, audio, and video previews.

Verification:
- `PATH="/opt/homebrew/opt/node@24/bin:$PATH" npm --prefix extensions/media-preview run compile`
- `PATH="/opt/homebrew/opt/node@24/bin:$PATH" npx mocha extensions/media-preview/test/mediaResourceVersion.test.js --ui tdd`
- `PATH="/opt/homebrew/opt/node@24/bin:$PATH" npm run compile-client`
- `git diff --check`


avinashkamat48 (commented):
One behavior change to check: `getResourceVersion()` catches every `workspace.fs.stat` failure and then falls back to `resource.toString()`, which is stable for the lifetime of the resource. For file systems/providers where `stat` is unsupported or intermittently fails, this removes the old `Date.now()` cache-busting behavior entirely, so reopening or rerendering the preview after the underlying content changes may keep using a stale webview URI. If the stable version is intentional only when a stat is available, the fallback may need to stay time-based or the PR should add coverage for a stat-failure resource changing between renders.

pony-maggie:
Thanks for catching that behavior change. I updated the fallback path so when `workspace.fs.stat` is unavailable, the media preview uses a fresh fallback version for each render instead of `resource.toString()`. Stat-backed resources still use the stable `mtime-size` version.

Validation run locally with Node 24.15.0:

```bash
npx -y -p node@24.15.0 node ./node_modules/gulp/bin/gulp.js compile-extension:media-preview
npx -y -p node@24.15.0 node ./node_modules/mocha/bin/mocha.js extensions/media-preview/test/mediaResourceVersion.test.js --ui=tdd --timeout=5000 --exit
```

Both passed.

pony-maggie:

PR #321364, opened 2026-06-15
ci: restore chat pipeline to windows-latest

### LINKED ISSUE #321267: Revert chat-lib tests pin from windows-2022 back to windows-latest
## Background

In #321263 the `chat-lib tests` workflow (`.github/workflows/chat-lib-package.yml`) was pinned from `windows-latest` to `windows-2022`.

This was needed because the GitHub-hosted `windows-latest` label migrated to the **Windows Server 2025 + Visual Studio 2026** image ([actions/runner-images#14017](https://github.com/actions/runner-images/issues/14017)). During `npm ci` in `extensions/copilot`, the native `sqlite3@5.1.7` module (a transitive dependency of the `@keyv/sqlite` devDependency) builds via `prebuild-install -r napi || node-gyp rebuild`. With no matching prebuilt, it falls back to `node-gyp`, which cannot detect the VS 18 toolchain on the new image:

```
gyp ERR! find VS unknown version "undefined" found at "...\Microsoft Visual Studio\18\Enterprise"
gyp ERR! find VS could not find a version of Visual Studio 2017 or newer to use
```

Pinning to `windows-2022` (still VS 2022) restores green, but it is temporary tech debt — the `windows-2022` runner image will eventually be deprecated.

## Action

Once the underlying issue is resolved, revert the pin back to `windows-latest`. The pin can be lifted when **either** of these is true:

- [ ] `node-gyp` with Visual Studio 2026 support (>= v12.1.0, see [nodejs/node-gyp#3240](https://github.com/nodejs/node-gyp/pull/3240)) is what `npm ci` uses for `extensions/copilot` on the new image, **or**
- [ ] The native `sqlite3` dependency is removed — e.g. upgrading `@keyv/sqlite` to v6, which no longer bundles native `sqlite3` (uses `better-sqlite3`/`node:sqlite`) — so no native build is needed at all.

## References

- PR that introduced the pin: #321263
- Failing run before the fix: https://github.com/microsoft/vscode/actions/runs/27461488041/job/81176049178
- Runner image migration: actions/runner-images#14017

deepak1556:
Fixes https://github.com/microsoft/vscode/issues/321267

Node v24.x ships with npm v11.x which already has node-gyp > v12.x


extensions/copilot/test/base/cache.ts
```diff
@@ -18,6 +17,22 @@ import { CurrentTestRunInfo } from './simulationContext';
...
+ * The previous `@keyv/sqlite` store was backed by the asynchronous `sqlite3`
+ * driver, so every individual store operation resolved on a later loop tick.
+ * Some consumers rely on that boundary: e.g. the panel multi-file edit flow
+ * applies one code block's edits while the next code block's code-mapper request
+ * is awaiting its cached response, so the next request is built against the
+ * already-edited document. `node:sqlite`'s `DatabaseSync` is synchronous, so we
+ * reintroduce the async boundary on every store operation to preserve that
+ * interleaving (and thus compatibility with the recorded cache).
+ */
+function yieldEventLoop(): Promise<void> {
```
deepak1556, line 32: Honestly I would like to re-record the cache with the sync primitives to avoid this, but that would blow out the main motivation behind this PR of updating the pipeline, leaving it as a followup.

cc @ulugbekna incase you are interested in addressing it.

joaomoreno (approved):

Issue #256854, opened 2025-07-19
Review AsyncIterableObject usage: potential memory leaks and migration to AsyncIterableProducer


hediet:
# GitHub Issue: Review AsyncIterableObject usage: potential memory leaks and migration to AsyncIterableProducer

**Title:** Review AsyncIterableObject usage: potential memory leaks and migration to AsyncIterableProducer

## Problem

`AsyncIterableObject` has a potential memory leak issue when iterables are only consumed once. The class stores all emitted values in an internal `_results` array to support multiple iterations, but this means values are never garbage collected in long-running iterations even when the iterable is only consumed once.

`AsyncIterableProducer` was introduced as an alternative that doesn't store values, making it more memory-efficient for single-consumption scenarios. However, this means that a second iteration will miss initial values.

## Analysis

We found **112 occurrences** of `AsyncIterableObject` across the codebase that should be reviewed to determine if they can be migrated to `AsyncIterableProducer` for better memory efficiency.

## Action Items

Please review the usages you've contributed to and determine if they can be migrated to `AsyncIterableProducer`:

### Contributors to Review

- [x] **@alexdima** (31 occurrences) - Core async.ts implementation and editor hover functionality
- [x] **@jrieken** (18 occurrences) - AsyncIterableSource, tests, chat and language model features
- [x] **@aiday-mar** (15 occurrences) - Editor hover functionality (inlay hints, color picker, markdown hover)
- [x] **@amunger** (19 occurrences total) - Notebook-related functionality and tests
- [x] **@hediet** (8 occurrences) - Core async functionality and inline completions
- [x] **@Tyriar** (5 occurrences) - Terminal shell integration
- [ ] **@sandy081** (4 occurrences) - Extension services
- [x] **@joaomoreno** (2 occurrences) - Workspace functionality
- [x] **@connor4312** (2 occurrences) - MCP (Model Context Protocol) server
- [x] **@aeschli** (2 occurrences) - Chat code block operations
- [ ] **@DonJayamanne** (1 occurrence) - Chat code block operations
- [x] **@mjbvz** (1 occurrence) - Editor hover functionality

### Review Criteria

For each usage, consider:

1. **Single vs Multiple Consumption**: Is the AsyncIterable consumed only once or multiple times?
2. **Memory Sensitivity**: Are large amounts of data being emitted that should be garbage collected promptly?
3. **Performance Impact**: Would switching to AsyncIterableProducer provide meaningful memory savings?

### Migration Guidelines

[...]

PR #321371, opened 2026-06-15
Use AsyncIterableProducer for text diff changes

FuZoe:
Refs #256854

## Why

`AsyncIterableObject` keeps an internal `_results` array of all emitted values so multiple iterators can replay the same data. The `workspace.getTextDiff` changes stream does not need that replay behavior: the diff result is produced once from the document diff promise, and `TextDiffResponse.changes` is expected to be consumed as a single async iterable.

`AsyncIterableProducer` is a better fit here because it does not retain emitted values. It passes values through its `ProducerConsumer`, which matches the one-shot nature of this stream and avoids the extra retained array.

## What

This updates `src/vs/workbench/api/common/extHost.api.impl.ts` so `getTextDiff` returns `changes` backed by `AsyncIterableProducer` instead of `AsyncIterableObject`.

The change covers:

- the async iterable import
- the cancellation path returning `EMPTY`
- the normal path creating the changes iterable

## Impact

This is a behavior-preserving refactor. API consumers still receive `AsyncIterable<TextDiffChange>` from `TextDiffResponse.changes`, and both implementations satisfy that interface.

There is no breaking API change.

## Validation

- `git diff --check -- src/vs/workbench/api/common/extHost.api.impl.ts`
- `npm run compile-check-ts-native -- --pretty false`

No new tests were added because this is a behavior-preserving implementation swap behind the existing `AsyncIterable<TextDiffChange>` API contract.

### DIFF
--- a/src/vs/workbench/api/common/extHost.api.impl.ts
+++ b/src/vs/workbench/api/common/extHost.api.impl.ts
@@ -5,7 +5,7 @@
 
 import type * as vscode from 'vscode';
 import { CancellationTokenSource } from '../../../base/common/cancellation.js';
-import { AsyncIterableObject, raceCancellationError } from '../../../base/common/async.js';
+import { AsyncIterableProducer, raceCancellationError } from '../../../base/common/async.js';
 import * as errors from '../../../base/common/errors.js';
 import { Emitter, Event } from '../../../base/common/event.js';
 import { combinedDisposable } from '../../../base/common/lifecycle.js';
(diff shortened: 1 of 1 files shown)


FuZoe:
Closing this as a duplicate of #318296. That earlier PR already makes the same scoped change in src/vs/workbench/api/common/extHost.api.impl.ts for workspace.getTextDiff: replacing AsyncIterableObject with AsyncIterableProducer for the changes iterable.

PR #321395, opened 2026-06-15
Fix Copilot semantic_search dotted path resolution

### LINKED ISSUE #321393: Copilot semantic_search returns incorrect file paths for directory names that contain a dot
The built-in Copilot `semantic_search` tool returns incorrect file paths for directory names that contain a dot. Results from dotted directories (e.g. `MyProject.Common`, `MyProject.Api`) are mangled, so the returned path points at a non-existent location on disk. The model then cannot find the result files, and the clickable `filepath:` links rendered in chat are dead.

I use the troubleshoot command and let the agent read the session log to confirm the details of the issue.

## Steps to reproduce

1. Codebase search must be enabled, workspace must be indexed. Workspace has a repo on github.
1. Open a workspace with directories with dots in the names, e.g. a repo rooted at `C:\repos\MyProject` that contains directories named `MyProject.Common`, `MyProject.Api`, `MyProject.Tests`, etc.
1. In Chat agent mode, run a query that triggers the `semantic_search` tool with a prompt likely to match code that lives under one of the `MyProject.*` dotted folders.
1. Inspect the `filepath:` lines returned in the tool result and in the fenced code blocks.

## Expected behavior

The tool returns the on-disk path verbatim, e.g.: `C:\repos\MyProject\MyProject.Common\Settings\MyFile.cs`

## Actual behavior

The directory name is magled and the part after the full stop is removed: `C:\repos\MyProject\MyProject\Settings\MyFile.cs` which cannot be opened. Most results whose first path component after the workspace root contains a dot is affected in the same way: `MyProject.Common` -&gt; `MyProject`, `MyProject.Api` -&gt; `MyProject`, etc.

Directories named `MyProject.Tests` seemed to be unaffected.

## Environment

- OS: Windows 10 (10.0.26200)
- VS Code: 1.124.2, commit `6928394f91b684055b873eecb8bc281365131f1c`, Electron 42.2.0, Node 24.15.0
- Copilot: built-in extension (no separate version)

Soumilgit:
Fixes #321393

### Summary of Changes
This PR fixes a bug in `PromptPathRepresentationService.resolveFilePath` where relative paths containing directory names with a dot followed by a colon (e.g., `MyProject.Common/src/file.ts:10`) were incorrectly identified as having a URI scheme and parsed as URIs.

### Root Cause Analysis
In `PromptPathRepresentationService.resolveFilePath`, the regular expression `/\w[\w\d+.-]*:\S/` lacks a start-of-string anchor (`^`). Because of this, it matches any substring within the file path that looks like a scheme followed by a colon and a non-whitespace character.

For example, in a path like `MyProject.Common/src/file.ts:10`, the substring `file.ts:10` matches the regex. Since the regex matched, `resolveFilePath` assumed the path starts with a scheme and passed it to `URI.parse`, which:
- Threw an error and returned `undefined` on Windows (due to backslashes).
- Parsed the line number as part of the path on POSIX/Unix (returning `/MyProject.Common/src/file.ts:10` as path with scheme `file`).

### Solution
Added the start-of-string anchor `^` to the scheme detection regex:
```typescript
if (/^\w[\w\d+.-]*:\S/.test(filepath)) { // starts with a scheme
```

This ensures that only actual URIs/paths starting with a scheme match, while relative paths (such as dotted folder paths with line numbers) are correctly ignored.

### Verification Plan
Added unit tests in `promptPathRepresentationService.spec.ts` covering relative paths with dotted folder names and line numbers (`MyProject.Common/src/file.ts:10` and `MyProject.Common:10`), verifying they now resolve to `undefined` instead of parsing incorrectly.


Soumilgit:
Hi Hi @jrieken , @sandy081 , @lszomoru , @alexdima and team,
please review this PR.

aeschli (dismissed):

aeschli:
Thanks @Soumilgit !

sandy081 (dismissed):

Soumilgit:
Hi @aeschli , @sandy081 and team,

I investigated the failing unit tests for `MyProject.Common:10` and have updated the branch with a robust fix. 

### The Issue & Fix
The scheme-detection regex previously used in `PromptPathRepresentationService.resolveFilePath` was `/^\w[\w\d+.-]*:\S/`. While this successfully anchored the scheme check to the start of the string, it still incorrectly classified root-level relative paths with line numbers (like `MyProject.Common:10`) as URI schemes because the dot in the directory name matched valid scheme characters.

I have updated the regex to `/^\w[\w\d+.-]*:\//` to require a slash `/` after the scheme colon. All valid file-based URI schemes in VS Code prompts (such as `file:/`, `untitled:/`, `vscode-remote:/`, and `vscode-vfs:/`) are always followed by a slash, whereas relative paths with line numbers (like `MyProject.Common:10`) are not.

This makes the resolution extremely robust and all unit tests now pass. Could you please re-review and merge the PR? Thanks!

PR #321445, opened 2026-06-15
Fix terminal tool handling of interactive prompts in agent terminals

### LINKED ISSUE #321444: Agent terminal gets stuck on interactive package manager prompts (needrestart, debconf, pipe buffering)
## Summary

The agent terminal tool fails to handle two classes of interactive prompts, causing consistent eval failures on terminalbench2 tasks that copilot-cli passes.

## Problem 1: needrestart/debconf TUI prompts

When `apt-get install` triggers needrestart, a "Which services should be restarted?" TUI prompt appears. None of the `detectsHighConfidenceInputPattern()` or `detectsLikelyInputRequiredPattern()` patterns in `outputMonitor.ts` match this format. The `idleSilence` timer fires after 60s, the tool returns "moved to background", and the agent gives up.

**Example**: `nginx-request-logging` — agent ran `apt-get install -y nginx`, needrestart prompted, agent waited 5 minutes then sent an empty `send_to_terminal` and stopped. copilot-cli finished the same task in 33 seconds.

## Problem 2: Pipe buffering hides prompts

When the model pipes an interactive command through `tail`/`head`/`grep`, the pipe buffers all output until the upstream closes. Interactive prompts (like opam `[Y/n]`) never reach the terminal buffer, so the `OutputMonitor` cannot detect them.

**Example**: `compile-compcert` — agent ran `opam install coq -y 2>&1 | tail -20`. The opam depext `[Y/n]` prompt was buffered by `tail` and never visible to the output monitor. The `[Y/n]` pattern *would* match `detectsHighConfidenceInputPattern` if it reached the terminal — the bug is the pipe hiding it.

## Fix

1. Set `DEBIAN_FRONTEND=noninteractive` in the copilot agent terminal env vars (in `toolTerminalCreator.ts`), alongside existing `GIT_PAGER`, `GIT_EDITOR`, etc. This prevents apt/dpkg from ever opening interactive prompts.

2. Add system prompt guidance telling the model to never pipe interactive commands through filters.

## Impact

Found via terminalbench2 eval analysis comparing vscode-copilot-cli vs copilot-cli. These two issues account for 2 of 12 consistent regressions (~13% score gap).

meganrogge:
Fixes #321444

## Problem

Two issues cause the agent terminal to get stuck on interactive prompts during package installs:

1. **needrestart/debconf TUI prompts** — `apt-get install` triggers a "Which services should be restarted?" TUI dialog that no `OutputMonitor` pattern matches. The agent waits 60s for the idle silence timeout, gets handed back a "moved to background" message, and gives up.

2. **Pipe buffering hides prompts** — When the model pipes interactive commands through `tail`/`head`/`grep`, the pipe buffers all output. Interactive prompts (like opam `[Y/n]`) never reach the terminal buffer, so the `OutputMonitor` cannot detect them. The `[Y/n]` pattern *would* match `detectsHighConfidenceInputPattern` — but the pipe prevents it.

## Fix

- Set `DEBIAN_FRONTEND=noninteractive` in the copilot agent terminal env vars (`toolTerminalCreator.ts`), alongside existing `GIT_PAGER`, `GIT_EDITOR`, etc. This prevents apt/dpkg from ever opening interactive prompts — packages install normally with default selections.

- Add system prompt guidance in `runInTerminalTool.ts` (both bash and PowerShell descriptions) warning the model to never pipe interactive commands through filters.

## Evidence

Found via terminalbench2 eval analysis comparing vscode-copilot-cli vs copilot-cli (12 consistent regressions, ~13% gap). These two issues account for 2 of those 12:
- **nginx-request-logging**: agent wasted 5 min on needrestart prompt, never configured nginx. copilot-cli finished in 33s.
- **compile-compcert**: `opam install ... | tail -20` hid the `[Y/n]` depext prompt for 5 min. copilot-cli answered `n` and continued.

### DIFF
--- a/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/toolTerminalCreator.ts
+++ b/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/toolTerminalCreator.ts
@@ -164,6 +164,10 @@ export class ToolTerminalCreator {
 			// `:` is a POSIX shell built-in no-op (returns 0), works cross-platform
 			// since git always invokes the editor via `sh -c`.
 			GIT_EDITOR: ':',
+			// Prevent apt/dpkg from opening interactive prompts (e.g. needrestart
+			// "Which services should be restarted?" dialogs). The agent cannot
+			// drive TUI prompts, so non-interactive mode picks safe defaults.
+			DEBIAN_FRONTEND: 'noninteractive',
 		};
 
 		const preventShellHistory = this._configurationService.getValue(TerminalChatAgentToolsSettingId.PreventShellHistory) === true;
--- a/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/runInTerminalTool.ts
+++ b/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/runInTerminalTool.ts
@@ -147,6 +147,7 @@ function createPowerShellModelDescription(shell: string, isSandboxEnabled: boole
 		'- Be specific with Select-Object properties to avoid excessive output',
 		'- Avoid printing credentials unless absolutely required',
 		`- NEVER run Start-Sleep or similar wait commands. You will be automatically notified on your next turn when async terminal commands or timed-out sync commands complete or need input. Do NOT poll for completion.`,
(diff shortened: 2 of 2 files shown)


meganrogge:

Yoyokrazy (approved):

PR #321506, opened 2026-06-15
Track terminal-to-session association by session ID instead of cwd

### LINKED ISSUE #312909: Better associate terminal with sessions
From: https://github.com/microsoft/vscode/pull/312873#discussion_r3150117121 and also comments in the mentioned PR.

meganrogge:
Replace the cwd-based terminal matching with a session ID-based map (`Map<sessionId, Set<instanceId>>`) for reliable 1:1 association. This avoids bugs where multiple sessions sharing the same cwd would interfere with each other's terminal lifecycle.

## Key changes
- Add `_sessionTerminals` map to track which terminals belong to which session
- `ensureTerminal` now records associations and prefers tracked terminals
- Visibility, archive, and removal use session ID lookups instead of scanning all terminals by cwd
- Fall back to cwd matching for restored/untracked terminals (backward compat)
- Transfer tracking on `onDidReplaceSession` so terminals survive untitled→committed graduation
- Clean up stale map entries via `onDidDisposeInstance` when terminals are externally closed
- Prune disposed terminals in `_isTerminalTracked` to avoid incorrect cwd fallback exclusion

## Tests
- All 52 tests pass (3 new)
- `transfers tracked terminals when a session is replaced (graduation)`
- `cleans up tracked terminal ids when terminals are externally disposed`
- `untracked restored terminals are visible alongside tracked terminals for the same session`

Fixes #312909

### DIFF
--- a/src/vs/sessions/contrib/terminal/browser/sessionsTerminalContribution.ts
+++ b/src/vs/sessions/contrib/terminal/browser/sessionsTerminalContribution.ts
@@ -349,16 +349,16 @@ export class SessionsTerminalContribution extends Disposable implements IWorkben
 
 		const instances = await this.ensureTerminal(targetPath, false, session);
 
-		// If the active key changed while we were awaiting, a newer call has
-		// taken over — skip the visibility update to avoid flicker.
-		if (this._activeKey !== targetKey) {
+		// If the active session or key changed while we were awaiting, a newer
+		// call has taken over — skip the visibility update to avoid flicker.
+		if (this._activeKey !== targetKey || this._activeSessionId !== session.sessionId) {
 			return;
 		}
 		await this._updateTerminalVisibility(session, targetKey, instances.map(instance => instance.instanceId));
 	}
 
 	/**
-	 * Finds the first terminal instance whose initial cwd (lower-cased) matches
+	 * Finds all terminal instances whose initial cwd (lower-cased) matches
 	 * the given key.
 	 */
 	private async _findTerminalsForKey(key: string, options?: { excludeTracked?: boolean }): Promise<ITerminalInstance[]> {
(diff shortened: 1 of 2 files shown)


src/vs/sessions/contrib/terminal/browser/sessionsTerminalContribution.ts
```diff
@@ -166,11 +169,37 @@ export class SessionsTerminalContribution extends Disposable implements IWorkben
...
+			}
+			this._sessionTerminals.delete(from.sessionId);
+		}));
+
+		// Clean up tracked terminal ids when terminals are externally disposed
+		// (e.g. user closes a terminal tab) so the map doesn't hold stale entries.
+		this._register(this._terminalService.onDidDisposeInstance(instance => {
+			this._removeTerminalFromTrackedSessions(instance.instanceId);
+		}));
+
```
davemcom, line 202: Very nice.  I look forward to testing this!

src/vs/sessions/contrib/terminal/browser/sessionsTerminalContribution.ts
davemcom, line 1: This looks like a solid fix!  With CWD backwards compatible.  If there's nothing funny with session lifecycle and `_sessionsManagementService.onDidChangeSessions` doesn't emit a `e.removed` for "whatever reason" upstream my problems will hopefully be over!!! 

Thank you!

davemcom: I do not see a guard for the weird case that might happen of from_ID=to_Id which would potentially result in the silent deletion of a terminal.

terminalIds = map.get(id)
targetIds = map.get(id) // same set
copy ids into itself
delete map[id]

anthonykim1 (approved):

PR #321537, opened 2026-06-16
Allow wider SCM diff gutter decorations

FuZoe:
Fixes #278083

## Why

`scm.diffDecorationsGutterWidth` was limited to the enumerated values `1` through `5`, and the quick diff decorator also rejected values greater than `5` at runtime. That made the settings editor unable to pick wider gutter decorations, and manually entered larger values were reset to the default.

## What

- Replace the fixed `1`-`5` enum with a numeric range of `1`-`20` for `scm.diffDecorationsGutterWidth`.
- Apply the same range in both SCM configuration registrations.
- Update the runtime guard so values up to `20` are honored instead of falling back to `3`.

## Impact

This preserves the existing default of `3` and keeps invalid values guarded, while allowing users to select wider Source Control diff gutter decorations when they need more visible indicators.

## Validation

- `npm run compile-check-ts-native -- --pretty false`


FuZoe:
This PR is ready for review. Happy to address any feedback. Thanks!

PR #321566, opened 2026-06-16
feat(editor): add Close Others to title menu and set shortcut

ashfakhm:
This adds the 'Close Others' option (workbench.action.closeOtherEditors) to the editor group title menu ('...'). Previously, it was only available in the tab context menu or the Command Palette. It also registers a default keyboard shortcut on Windows/Linux: Ctrl+K Alt+W.

image


benibenj:
Yes, sorry, I didn't have time to respond yet. I think having "Close Others" in that menu is a bit confusing because the actions in the menu apply to the entire editor group, not just a single editor within the group.

"Close Others" only really makes sense when the action is targeting one specific editor, since it means closing all other editors except that one.

PR #321591, opened 2026-06-16
Handle opened and unsaved documents as attachments for Agent Host/Copilot CLI

### LINKED ISSUE #320428: missing active doc context in AHP sessions
1. Open a document, either saved or untitled, and ask about the content in "this file"

🐛 AHP agent will ask to clarify what file, whereas a local chat is given the context to know what "this" means.

Note that selected text from a saved document is attached as expected, but neither local nor AHP will attach a selection from an untitled doc.

dmitrivMS:
Fixes #320428


DonJayamanne (dismissed):
Haven't reviewed the code completely, but couldn't find any special casing for untitled editors

You will want to handle untitled editors. as in a new file thats not saved on disc `untitled1` 
Those files cannot be accessed by CLI as they don't exist on disc, hence we'll need a way to either block them/get them working or other.

dmitrivMS:
> Haven't reviewed the code completely, but couldn't find any special casing for untitled editors
> 
> You will want to handle untitled editors. as in a new file thats not saved on disc `untitled1` Those files cannot be accessed by CLI as they don't exist on disc, hence we'll need a way to either block them/get them working or other.

Interesting... would this apply to any unsaved files? I.e. even if file exists, the content in the editor is different on disk. Our read tools are able to read it but I suppose CLI would not. If true, this is a broader problem. But does it need to be part of this change - i.e. do you believe things will change here depending on how we handle unsaved files?

src/vs/workbench/contrib/chat/browser/chat.shared.contribution.ts
```diff
@@ -326,6 +326,13 @@ configurationRegistry.registerConfiguration({
 			default: true,
 			agentsWindow: { default: false },
 		},
+		'chat.implicitContext.includeActiveEditorInAgentSessions': {
```
DonJayamanne: If we plan on keeping this setting around even after Agent Host is the default, I'm not sure we should name it `AgentSessions`
Because thats more of an implementation detail, hence having such a setting might not make sense if agent host sessions are the only sessions we suppport

DonJayamanne (approved):

PR #321599, opened 2026-06-16
Apply openChangedFileInDiffEditor to the chat edit pills

kangarko:
Follow-up to #320948.

That PR added `chat.editing.openChangedFileInDiffEditor` and the alt-click swap to the two changed-files lists (the changes summary and the multi-diff list). 

The "Edited <file>" pill that shows up inline under a response was left out, so clicking it still always opened the read-only snapshot diff and ignored the setting. 

@connor4312 @roblourens you approved the original, would appreciate a look at the follow-up.


kangarko:
Thanks, addressed all three in the latest push:

- `altKey` is now optional on `IOpenEditorOptions` and defaulted to `false` where the pills read it.
- Decoupled Alt from `openToSide`. Alt no longer opens to the side, it only swaps diff vs file, which is how the changed-files lists already behave since #320948. Ctrl/Cmd still opens to the side. One side effect worth calling out: the file attachment widget that shares this helper loses Alt+click-to-side (Ctrl/Cmd+click still works), which lines up better with the usual VS Code modifier anyway.
- No behavior change with the default setting: plain click opens the diff, Alt+click opens the file, and the setting swaps the two.

PR #321668, opened 2026-06-16
Fix run_in_terminal returning empty output when command exceeds scrollback

### LINKED ISSUE #321667: Run in terminal tool returns 'Failed to retrieve command output' when output exceeds scrollback
## Problem

When a command's output exceeds the terminal's scrollback buffer (`terminal.integrated.scrollback`, default 1000 lines), the `run_in_terminal` tool returns "Failed to retrieve command output" instead of returning whatever is available in the buffer.

## Expected behavior

The tool should return the full scrollback buffer contents (i.e. the last N lines that fit in scrollback) rather than failing entirely. Even partial output is more useful than an error message.

## Repro

1. Set `terminal.integrated.scrollback` to a low value (e.g. 100)
2. Have the agent run a command that produces more lines than the scrollback (e.g. `seq 1 5000`)
3. Tool result says "Failed to retrieve command output"

## Notes

This is separate from the large output truncation/file-write handling in #321640 — that PR handles the case where output is successfully retrieved but exceeds the inline size limit. This issue is about output never being retrieved at all.

meganrogge:
Only caught now because default scrollback is 1000 and mine is set to 10,000!!!

When a command's output exceeds the terminal scrollback buffer, the start marker gets disposed (line = -1). Previously `getContentsAsText` threw an error, causing the tool to report 'Failed to retrieve command output' with no content.

Now it falls back to reading from line 0 of the buffer, returning whatever remains. The strategies also add an informational note: 'Output exceeded terminal scrollback; beginning of output was lost' so the model knows the output is partial.

Fixes #321667

Before:

Screenshot 2026-06-16 at 4 43 53 PM

After:

Screenshot 2026-06-16 at 4 41 36 PM

### DIFF
--- a/src/vs/workbench/contrib/terminal/browser/xterm/xtermTerminal.ts
+++ b/src/vs/workbench/contrib/terminal/browser/xterm/xtermTerminal.ts
@@ -379,13 +379,13 @@ export class XtermTerminal extends Disposable implements IXtermTerminal, IDetach
 	getContentsAsText(startMarker?: IXtermMarker, endMarker?: IXtermMarker): string {
 		const lines: string[] = [];
 		const buffer = this.raw.buffer.active;
-		if (startMarker?.line === -1) {
-			throw new Error('Cannot get contents of a disposed startMarker');
-		}
 		if (endMarker?.line === -1) {
 			throw new Error('Cannot get contents of a disposed endMarker');
 		}
-		const startLine = startMarker?.line ?? 0;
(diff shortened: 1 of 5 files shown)


Yoyokrazy (dismissed):

anthonykim1 (dismissed):

meganrogge:

dmitrivMS (approved):

kycutler (approved):

javi0guerrero4-cloud (commented):
****

PR #321640, opened 2026-06-16
Align terminal output handling with cli

meganrogge:
## Before → After → Runtime comparison

| Behavior | Before (VS Code) | After (this PR) | copilot-agent-runtime |
|----------|------------------|-----------------|----------------------|
| **Max inline output** | 60KB hard cap | 20KB | 20KB |
| **Large output** | Truncated (tail only), head lost forever | Written to temp file; model gets path + head preview + tail | Written to temp file; model gets path + 500-char preview |
| **Truncation message** | `[... PREVIOUS OUTPUT TRUNCATED ...]` | `[Output too large (49KB). Full output saved to: /tmp/...]\n[Use readFile or grep to examine the full output.]` | `Output too large to read at once (49KB). Saved to: /tmp/...\nConsider using tools like grep...` |
| **JSON formatting** | None | Pretty-print only in the saved file (not inline) | Pretty-print only when writing to file |
| **Temp file cleanup** | N/A | On session end + tool dispose | On session end (`cleanupTempFiles()`) |
| **CR collapsing** | Not needed (xterm resolves `\r` at cell layer) | Not needed | Yes (raw `child_process.spawn` needs it) |
| **Tool description** | Says "60KB" | Says "20KB", mentions file-backed recovery | N/A (different tool system) |

## Files changed

- **`outputHelpers.ts`** — `getOutput()` returns raw output (no truncation, no formatting). New `truncateLargeOutput(output, filePath?)` export.
- **`largeOutputFileWriter.ts`** — New. Writes to `IEnvironmentService.tmpDir`, pretty-prints JSON in file, tracks URIs, cleans up on session end + dispose.
- **`runInTerminalTool.ts`** — Calls `_largeOutputFileWriter.processOutput()` before returning tool result. Hooks `onDidDisposeSession` for temp file cleanup. Updated tool descriptions.

Fixes microsoft/vscode-internalbacklog#8073

Screenshot 2026-06-16 at 4 24 55 PM

Issue #321432, opened 2026-06-15
Agent turn hangs indefinitely when a streaming response stalls (no idle timeout in messagesApi/responsesApi SSE read loop)


meganrogge:
## Summary

claude-opus agent turns hang for the full harness wall-clock cap (eval: `X_AGENT_STILL_RESPONDING`, 0 output) when a single `/v1/messages` (Anthropic Messages API) streaming response stalls — the upstream/proxy returns HTTP 200 headers but never streams a terminating body and never closes the socket, and the client's SSE read loop in `processResponseFromMessagesEndpoint` has **no idle timeout**, so the turn never resolves or rejects. Found via terminalbench2 eval runs where it was the dominant error mode (~18 of 30 errored instances across 3 `vscode-copilot-cli` runs). The same unguarded read loop exists in the OpenAI **Responses API** path, so this is not Anthropic-specific in principle.

## Where it hangs

[`extensions/copilot/src/platform/endpoint/node/messagesApi.ts`](https://github.com/microsoft/vscode/blob/main/extensions/copilot/src/platform/endpoint/node/messagesApi.ts) → `processResponseFromMessagesEndpoint`:

```ts
for await (const chunk of response.body) {
    parser.feed(chunk);
}
```

No inactivity guard. When the body delivers 200 headers but no SSE bytes (and never ends), this loop blocks forever: no `message_stop` is parsed, the `AsyncIterableObject` never calls `feed.emitOne(...)` or `feed.reject(...)`, and the `ToolCallingLoop` awaiting the turn hangs until an outer cap (the eval harness's 60-min limit; in product, effectively indefinitely / until user cancel) kills it.

The **identical pattern** is in `extensions/copilot/src/platform/endpoint/node/responsesApi.ts` → `processResponseFromChatEndpoint` (line ~907), so the OpenAI Responses path shares the vulnerability.

## Evidence (from eval `GitHub Copilot Chat.log`)

A healthy turn logs `[messagesAPI] message 0 returned. finish reason:[stop]` → `ccreq …| success | claude-opus…` → (final turn) `[ToolCallingLoop] Stop hook result: shouldContinue=false` → clean exit.

A hung turn:

| Instance | Signature | Evidence |
|---|---|---|
[...]

PR #321671, opened 2026-06-16
Add stream-stall telemetry to Anthropic Messages streaming (#321432)

meganrogge:
### What

Adds an **observe-only** idle watchdog to `processResponseFromMessagesEndpoint` in `messagesApi.ts` (the Anthropic Messages API path used by `vscode-copilot-cli` with Claude models). It detects when the streaming body stalls and emits a new `messagesApi.streamIdleTimeout` telemetry event. It does **not** change request behavior — the stream is not aborted.

### Why

The streaming-stall class of hang in microsoft/vscode#321432: the HTTP response can return `200` with headers and then **stall mid-stream** — the body iterator (`for await (const chunk of response.body)`) stops producing chunks but never completes and never errors, so the request hangs indefinitely. Today only the success path is instrumented, so when this happens the stall is **invisible in telemetry**.

This was the root cause of 4 of the 6 `X_AGENT_STILL_RESPONDING` timeouts in MSBench run `27640643629` (the others were genuine 60-min budget exhaustion). The stall is environment-agnostic — the eval harness just amplifies its frequency — so it likely affects real users too. We want to **measure how often this happens in the wild first** before changing any request behavior.

### What this change does

- Arms an **idle watchdog** that is reset on every received chunk.
- If no chunk arrives within the window, it emits `messagesApi.streamIdleTimeout` **once** and otherwise leaves the stream untouched.
- The watchdog timer is always cleared in a `finally`.

This is intentionally non-invasive: no abort, no reject, no behavior change. A follow-up can add mitigation once we have data on frequency and shape.

### Telemetry dimensions

`messagesApi.streamIdleTimeout` (classification `SystemMetaData`, purpose `PerformanceAndHealth`):

- `requestId`, `ghRequestId` — correlate with server-side logs
- `idleMs` — ms since the last chunk when the watchdog tripped
- `elapsedMs` — ms since the stream started
- `chunksReceived` — chunks received before the stall
- `completionsEmitted` — completions emitted before the stall (`0` = first-turn hang, `>0` = mid-stream hang)

Related: microsoft/vscode#321640

### DIFF
--- a/extensions/copilot/src/platform/endpoint/node/messagesApi.ts
+++ b/extensions/copilot/src/platform/endpoint/node/messagesApi.ts
@@ -603,6 +603,15 @@ function markLastCacheableBlock(msg: MessageParam, cacheTtl?: '1h'): void {
 	}
 }
 
+/**
+ * How long the Anthropic Messages streaming body may go without producing a
+ * chunk before the idle watchdog considers it stalled and aborts it. The HTTP
+ * layer can return 200 with headers and then hang mid-stream with no further
+ * data and no error (see microsoft/vscode#321432), which would otherwise leave
+ * the `for await` loop pending indefinitely.
+ */
+const ANTHROPIC_STREAM_IDLE_TIMEOUT_MS = 120_000;
+
 export async function processResponseFromMessagesEndpoint(
 	instantiationService: IInstantiationService,
 	telemetryService: ITelemetryService,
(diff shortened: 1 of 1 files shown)


kycutler (dismissed):

extensions/copilot/src/platform/endpoint/node/messagesApi.ts
```diff
@@ -603,6 +603,16 @@ function markLastCacheableBlock(msg: MessageParam, cacheTtl?: '1h'): void {
...
 
+/**
+ * How long the Anthropic Messages streaming body may go without producing a
+ * chunk before the idle watchdog considers it stalled and reports it via
+ * telemetry. The HTTP layer can return 200 with headers and then hang
+ * mid-stream with no further data and no error (see microsoft/vscode#321432).
+ * For now this only observes the stall so we can measure how often it happens
+ * in the wild; it does not abort the request.
+ */
+const ANTHROPIC_STREAM_IDLE_TIMEOUT_MS = 120_000;
```
bhavyaus: Why 2mins? Sometimes the model takes a while to continue streaming.

meganrogge: I have no idea, asked for your guidance here. Changed it to 5 minutes.

lszomoru (approved):

PR #321683, opened 2026-06-16
Allow interactive commands in macOS agent sandbox

### LINKED ISSUE #316978: copilot sandbox cannot run interactive commands
- Copilot Chat Extension Version:  0.48.1
- VS Code Version: 1.120.0
- OS Version: macos15latest
- Selected model (e.g. GPT 4.1, Claude 3.7 Sonnet): claude 4.6 opus
- Logs:
```
out/extension.js      145.2kb
out/extension.js.map  285.0kb

⚡ Done in 17ms
 WARNING  LICENSE, LICENSE.md, or LICENSE.txt not found
 ERROR  setRawMode EPERM
```

Steps to Reproduce:
In agent mode, ask the agent to run a command that triggers an interactive prompt (e.g., npm run package where vsce package detects a missing LICENSE file and asks "Do you want to continue? [y/N]")
The command fails with: ERROR setRawMode EPERM

Description:

When the agent executes a command in the sandboxed terminal that requires interactive input (e.g., vsce package prompting for confirmation), the command fails with setRawMode EPERM.

The sandbox terminal does not support process.stdin.setRawMode(true), which is required by many CLI tools for interactive prompts (y/N confirmations, password input, etc.).

Either:

The sandbox terminal should support setRawMode for interactive prompts, OR
The agent should be able to detect the interactive prompt and use send_to_terminal to respond, OR
A clearer error message indicating the sandbox does not support interactive input
Actual behavior:

The command fails with an opaque EPERM error. The agent has no way to answer the prompt or bypass it within the sandbox.

Workaround:

Resolve the condition causing the prompt before running the command (e.g., add a LICENSE file so vsce doesn't ask for confirmation).

dileepyavan:
## Summary

- enable PTY access by default for macOS agent sandbox runtime configurations
- preserve an explicit `allowPty: false` advanced runtime override
- add regression coverage confirming the default is macOS-only

## Testing

- `npm run eslint -- src/vs/platform/sandbox/common/terminalSandboxEngine.ts src/vs/platform/sandbox/test/common/terminalSandboxEngine.test.ts`
- `npm run transpile-client`
- `npm run test-node -- --run src/vs/platform/sandbox/test/common/terminalSandboxEngine.test.ts` (28 passing)

Fixes #316978


Yoyokrazy (approved):

roothider:
👍

PR #321736, opened 2026-06-17
Reload cross-origin webview resource stylesheets in CORS mode on Firefox and Safari

### LINKED ISSUE #321735: Webviews render unstyled in Firefox and Safari when resources are served cross-origin
- VS Code Version: reproduced on `main`
- OS: server: Linux; browser: Firefox / Safari

### Problem

When the webview content frame and its resources are served from different origins — the `vscode-resource` virtual host, the normal architecture (e.g. `*.vscode-webview.net` for content vs `*.vscode-resource.vscode-cdn.net` for resources) — extension stylesheets do not load in **Firefox** and **Safari**. The webview renders with no styling (toolbars/inputs appear as unstyled, stacked elements). **Chromium is unaffected.**

### Root cause

Extension `<link rel="stylesheet">` elements have no `crossorigin` attribute, so the browser fetches them in **no-cors** mode. The webview service worker answers with an opaque cross-origin response. Firefox and Safari refuse to apply a stylesheet from an opaque response; Chromium is lenient and applies it anyway.

### Steps to reproduce

1. Serve the workbench so that webview resources resolve to a different origin than the content frame (i.e. `vscode-resource-base-authority` is set, as in a standard cross-origin webview deployment).
2. Open the workbench in Firefox or Safari.
3. Activate an extension that contributes a webview with an external stylesheet.
4. The webview content is unstyled. The same setup in Chrome is styled.

### Expected

Webview stylesheets load and apply in Firefox and Safari, as they do in Chromium.

### Notes

The service worker already serves these resources with `Access-Control-Allow-Origin: *`, so the resources are CORS-eligible — the links just aren't requested in CORS mode. A fix is prepared (mark webview-resource stylesheet links `crossorigin` and re-fetch, Firefox/Safari only) and can be submitted as a PR.

ahmetozer:
Fixes #321735

### What

On Firefox and Safari, webview stylesheets are silently dropped when webview resources are served from a different origin than the content frame (the `vscode-resource` virtual host).

### Why

Extension `<link rel="stylesheet">` elements carry no `crossorigin` attribute, so the browser requests them in **no-cors** mode. The webview service worker returns an opaque cross-origin response, and Firefox/Safari refuse to apply a stylesheet from an opaque response. Chromium applies it regardless, which is why this only reproduces in Firefox and Safari.

### Fix

In the webview bootstrap (`webview/browser/pre/index.html`), after the inner content document is written, mark webview-resource stylesheet links as `crossorigin="anonymous"` and re-issue the request. The service worker already serves these resources with `Access-Control-Allow-Origin: *` (`service-worker.js`), so the CORS fetch succeeds and the stylesheet applies.

- Scoped by hostname using the **same predicate the service worker uses** to decide what is a webview resource (`hostname.endsWith('.' + resourceBaseAuthority)`), so only links the worker serves with `ACAO: *` are touched. Same-origin and third-party stylesheets are left untouched.
- A `MutationObserver` covers stylesheet links injected after first paint (common for extension webviews).
- Guarded by `isFirefox || isSafari`; **Chromium behavior is unchanged**.
- Uses web standards (`crossorigin` / CORS), no Electron APIs.

The `Content-Security-Policy` `script-src` self-hash in the same file is updated to match the modified inline bootstrap script (the hash is maintained by hand; without the bump the browser would block the bootstrap).

### Risk / scope

- No-op when `vscode-resource-base-authority` is unset.
- For resources already loading (e.g. same-origin setups), forcing a CORS re-fetch against a host that sends `ACAO: *` still succeeds — worst case is one extra request per resource stylesheet, Firefox/Safari only.
- No change to Chromium, to same-origin webviews, or to third-party resources.

### How to test

1. Serve the workbench so webview resources resolve to a different origin than the content frame (`vscode-resource-base-authority` set).
2. In **Firefox** and **Safari**, activate an extension that contributes a webview with an external stylesheet.
3. Before: webview is unstyled. After: webview is styled, matching Chrome.
4. Confirm Chrome is unchanged, and that same-origin webviews still load.

### DIFF
--- a/src/vs/workbench/contrib/webview/browser/pre/index.html
+++ b/src/vs/workbench/contrib/webview/browser/pre/index.html
@@ -5,7 +5,7 @@
 	<meta charset="UTF-8">
 
 	<meta http-equiv="Content-Security-Policy"
-		content="default-src 'none'; script-src 'sha256-nXjtuhBilO++r8hfxl5VjEScSmdm07wDAk6jw228DgM=' 'self'; frame-src 'self'; style-src 'unsafe-inline';">
+		content="default-src 'none'; script-src 'sha256-5Ad2+eOJ0yGZkRkKpVLcNmLeB0vi7rRpxq1J/qGluu8=' 'self'; frame-src 'self'; style-src 'unsafe-inline';">
 
 	<!-- Disable pinch zooming -->
 	<meta name="viewport"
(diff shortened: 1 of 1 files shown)


ahmetozer:

ahmetozer:
Short description of use case https://github.com/microsoft/vscode/issues/321735#issuecomment-4742161482

ahmetozer:
Any updates ?

ahmetozer:
This issue forces me to use localhost domain, which requires proxiying each time. I cannot direrly access over VPN as well. :/

PR #321775, opened 2026-06-17
Add opt-in CSRF protection for extension UriHandlers

alichtman:
TLDR: This allows extension owners to lock down access to their deeplinks to only the local computer, which is a nice defensive boundary.

---

Extension URI handlers (`vscode://publisher.ext/...`) are attack surfaces that are reachable from any code path that can hit the OS protocol handler. This includes  including browsers, so any clicked link may hit this attack surface.

Certain extensions may want to expose functionality only to a local program (for example, a testing extension and associated testing CLI), but not to browsers.

VS Code cannot tell a browser-initiated deeplink from a local one -- both arrive as the same URI string through the OS protocol handler. I thought the best way to prove local origin was to require knowledge of an on-disk secret. This adds an opt-in mechanism where a local tool signs the deeplink's path + params with HMAC-SHA256 (key = an owner-only secret file), carried in a reserved `vscode-csrf-token` query param. VS Code recomputes the HMAC and rejects on mismatch; the secret never travels in the URL.

This allows extension owners to lock down access to their deeplinks to only the local computer, which is a nice defensive boundary.

API (stable, optional -- not behind enabledApiProposals, so published extensions can adopt it):
- package.json: `contributes.uriHandler.csrfProtection` (authoritative)
- runtime: `registerUriHandler(handler, { csrfProtection })` where csrfProtection is `true` or `{ secretFile?, unprotectedPaths? }`. unprotectedPaths exempts legitimately web-initiated routes (e.g. OAuth callbacks) that a browser cannot sign.

Implementation:
- Pure core (canonicalize, HMAC via crypto.subtle, constant-time compare, token strip) in services/extensions/common/uriHandlerCsrf.ts, with unit tests.
- Secret store: owner-only file, crash-safe temp-then-link create, persistent (no rotation); rejects group/other-writable files; fail-closed Null impl for the web worker ext host.
- Enforcement in ExtHostUrls.$handleExternalUri; rejections are logged (redacted) and surface one warning toast.
- Pre-activation hook: the workbench fans out $verifyCsrf to extension hosts so forged links are dropped before the extension activates, and verified links skip the trust prompt (manifest-declared policy only). Maybe this was added complexity that wasn't worth it, and we can tolerate activating the extension for a simpler diff.

### DIFF
--- a/src/vs/platform/extensions/common/extensions.ts
+++ b/src/vs/platform/extensions/common/extensions.ts
@@ -242,6 +242,16 @@ export interface IExtensionContributions {
 	readonly languageModelTools?: ReadonlyArray<IToolContribution>;
 	readonly languageModelToolSets?: ReadonlyArray<IToolSetContribution>;
 	readonly mcpServerDefinitionProviders?: ReadonlyArray<IMcpCollectionContribution>;
+	readonly uriHandler?: IUriHandlerContribution;
+}
+
+export interface IUriHandlerCsrfProtection {
+	readonly secretFile?: string;
+	readonly unprotectedPaths?: readonly string[];
+}
+
+export interface IUriHandlerContribution {
+	readonly csrfProtection?: boolean | IUriHandlerCsrfProtection;
 }
 
 export interface IExtensionCapabilities {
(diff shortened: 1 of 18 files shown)


alichtman:
Opened prematurely :) I'll clean it up and open another one.

PR #321811, opened 2026-06-17
Auto-continue old local chat in Copilot CLI agent host

anthonykim1:
Resolves: 

- Auto-select Copilot CLI Agent Host as the `Continue In` target when opening a non-empty Local chat while `chat.editor.localAgent.enabled` is false and `chat.editor.defaultProvider` is `copilotAh`.
- Reuse the existing pending delegation flow so the next send creates/opens a new Agent Host session with the prior Local transcript attached.
- Keep the old Local session unchanged in history.
- Hide Local from the editor-window `Continue In` picker when Local is disabled and hidden by the editor chat provider policy.
- Re-check when chat session availability changes so Copilot CLI Agent Host registration timing does not leave old Local sessions selected.

Video: 
https://github.com/user-attachments/assets/88adaaca-b371-4a4d-bfaf-2e024bd1dfd8

-----------------------------------------
Inspirations from:

- Pending delegation submit behavior comes from [`chatExecuteActions.ts`](https://github.com/microsoft/vscode/blob/f8574878be86c229f5d9710cdde25354bff25dde/src/vs/workbench/contrib/chat/browser/actions/chatExecuteActions.ts#L60-L64).
- Continue-In transcript handoff behavior comes from [`chatContinueInAction.ts`](https://github.com/microsoft/vscode/blob/f8574878be86c229f5d9710cdde25354bff25dde/src/vs/workbench/contrib/chat/browser/actions/chatContinueInAction.ts#L533-L549).
- Local visibility policy comes from [`isVisibleEditorChatSessionType`](https://github.com/microsoft/vscode/blob/f8574878be86c229f5d9710cdde25354bff25dde/src/vs/workbench/contrib/chat/common/constants.ts#L276-L292).
- Widget lock/picker refresh behavior comes from [`ChatInputPart.setPendingDelegationTarget`](https://github.com/microsoft/vscode/blob/f8574878be86c229f5d9710cdde25354bff25dde/src/vs/workbench/contrib/chat/browser/widget/input/chatInputPart.ts#L2398-L2410).


src/vs/workbench/contrib/chat/browser/widget/input/delegationSessionPickerActionItem.ts
```diff
@@ -116,6 +117,9 @@ export class DelegationSessionPickerActionItem extends SessionTypePickerActionIt
 		if (this._isSessionsWindow && type === AgentSessionProviders.Local) {
 			return false;
 		}
+		if (!this._isSessionsWindow && type === AgentSessionProviders.Local && !isVisibleEditorChatSessionType(type, this.configurationService, this.chatSessionsService)) {
+			return false;
+		}
```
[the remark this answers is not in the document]
anthonykim1: yes should address in future PR when we discuss which ones we want to show.

src/vs/workbench/contrib/chat/common/constants.ts
```diff
@@ -292,6 +292,19 @@ export function isVisibleEditorChatSessionType(
 	return !!chatSessionsService.getChatSessionContribution(sessionType);
 }
 
+export function shouldAutoDelegateLocalSessionToAgentHostCopilot(
```
DonJayamanne: I don''t think this needs to be in constants.ts, this is a specific function that will only be used in one place isn't it?
If not re-usable, then perhaps keep in chatWidget.ts

DonJayamanne: Once we've delegated, 
* should we freeze the old session, else user will get confused as to what is going on here
* Also should we add a message into the local session saying that we can no longer send messages in local agents 

I.e. i would be inclined to think that its a bug that a new session is getting started and I'm unable to send a message on my original local session.

@justschen did some work for handover from vscode editor window to agents window. 
I think that UI and workflow is better here, as its obvious to the user that something has changed

anthonykim1: [f996822](https://github.com/microsoft/vscode/pull/321811/commits/f996822fd619cb6bcc91d2c45999b41f8fe8ee5b) 
should give better UI like: 
https://github.com/user-attachments/assets/5dd6c239-83cf-4931-ac03-6391c74a36aa

anthonykim1:
Closing this FOR NOW and creating a PR, targetting less aggressive transition for historical local -> AH Copilot CLI for now.

PR #321825, opened 2026-06-17
Enable tool picker in Agent Host sessions

### LINKED ISSUE #321475: Required tools missing for feedback skill for claude agent
Testing #321338

Repro:

- Submitted feedback and I see the below issue for claude agent.

Image

dmitrivMS:
## What

Agent Host (AH) sessions can now use VS Code's built-in and extension-contributed language-model tools, selected through the existing chat tool picker — the same UI used for local chat. Previously the set of client tools exposed to an AH backend was driven by a static `chat.agentHost.clientTools` setting.

## How it works

- The tool picker is now reachable in locked AH sessions and writes to a **dedicated agent-host tool selection** (separate storage from the default agent, so the two don't interfere).
- Tools whose capability the backend already provides natively — file read/edit, literal file/text search, shell execution, web fetch, and sub-agent orchestration (`runSubagent`) — are **hidden from the picker and never sent to the backend**, so the model isn't offered the same capability twice. They are excluded even when they appear as members of a built-in tool set.
- VS Code–specific tools with no backend equivalent stay available and on by default — notably `codebase` and `githubRepo` (semantic/embeddings search the backend's literal grep/glob can't replace), the integrated-terminal context tools (`terminalLastCommand` / `terminalSelection`), tasks, tests, diagnostics, and MCP servers.
- Selection changes propagate **live** to running sessions via a shared resolver, so the exposed client tools always match what the picker shows.

## Notes

- Client SDK tools now set `overridesBuiltInTool: true`, fixing a crash when a VS Code tool shares a name with an SDK built-in.
- The old `chat.agentHost.clientTools` setting is removed.

Fixes #321475


src/vs/workbench/contrib/chat/browser/chat.shared.contribution.ts
```diff
@@ -1169,19 +1168,6 @@ configurationRegistry.registerConfiguration({
 				mode: 'auto'
 			},
 		},
-		[ChatConfiguration.AgentHostClientTools]: {
-			type: 'array',
-			items: { type: 'string' },
-			description: nls.localize('chat.agentHost.clientTools', "Tool reference names to expose as client-provided tools in agent host sessions."),
-			default: [
-				'runTask', 'getTaskOutput', 'problems', 'runTests',
```
DonJayamanne, line 1177: I haven't tested this 
However, from what i know tools like `problems`, perhaps even `runTask` might not work in agents window
@dmitrivMS

DonJayamanne (approved):

src/vs/workbench/contrib/chat/browser/chat.shared.contribution.ts
dmitrivMS: Hmm.. Good point, I've only tested in the editor window, let me try agents as well.

DonJayamanne: Note: we dont have a problems panel in Agents window

DonJayamanne (changes requested):
Best to exclude `problems` from the tool picker in agents window & perhaps check other similar tools.
Notebook tools will not work either, as the file format is completely different, we will need to exclude all `notebook` related tools from vscode and agents window for All Agent Host sessions.

dmitrivMS:
> Best to exclude `problems` from the tool picker in agents window & perhaps check other similar tools. Notebook tools will not work either, as the file format is completely different, we will need to exclude all `notebook` related tools from vscode and agents window for All Agent Host sessions.

There is no tool picker in the Agents window, but I'd like to try excluding them when sending over to AH in Agents window context.

DonJayamanne (dismissed):

src/vs/workbench/contrib/chat/browser/widget/input/chatSelectedTools.ts
```diff
@@ -17,95 +16,30 @@ import { ILanguageModelChatMetadataAndIdentifier } from '../../../common/languag
...
-		return JSON.stringify(storageData);
-	}
-}
-
 export enum ToolsScope {
 	Global,
 	Session,
 	Agent,
 	Agent_ReadOnly,
+	AgentHost,
```
aeschli, line 29: Look like a hack... AgentHost is not a scope.

aeschli (changes requested):
Let's not modify the existing tool picker and model, IMO it would be better to define a new picker, or add a new tab to the customizations dialog.
Persisting the eabled tools has to be on the agent host, per session

dmitrivMS:
Closing per discussion, will rework towards Chat Customizations.

PR #321843, opened 2026-06-17
fix: handle pending_confirmation signals when no active protocol turn exists

digitarald:
Fixes https://github.com/microsoft/vscode-internalbacklog/issues/8079 .

## Problem

Agent host sessions get stuck in "thinking" permanently when a `pending_confirmation` signal arrives without an active protocol turn. This happens when:

1. A protocol turn completes (e.g., rubber duck subagent finishes)
2. A `userPromptSubmitted` hook fires, triggering a continuation in the SDK
3. The SDK calls a tool (e.g., `view`) which needs permission approval
4. The `pending_confirmation` signal is emitted but silently dropped because `getActiveTurnId()` returns `undefined`
5. The deferred promise in `_pendingPermissions` never resolves
6. The session hangs forever — user's subsequent "Continue" messages are also blocked

## Root Cause

In `agentSideEffects._handleAgentSignal`, the "no active turn" fallback only handled `action` signals. The `pending_confirmation` signal (which needs to go through `_handleToolReady` for auto-approval checks) was silently dropped when no active turn existed.

## Fix

- Route `pending_confirmation` signals through `_handleToolReady` with an empty turnId in the no-active-turn fallback path (matching what action signals already do)
- Fix the subagent path to always call `_handleToolReady` with `subTurnId ?? ''` instead of conditionally skipping when `subTurnId` is undefined

## Testing

- Added regression test: `pending_confirmation without an active turn still dispatches (does not hang)`
- All 81 existing tests pass


src/vs/platform/agentHost/node/agentSideEffects.ts
```diff
@@ -365,6 +363,17 @@ export class AgentSideEffects extends Disposable {
...
 		// turnStarted was not observed by this side-effects instance.
+		//
+		// pending_confirmation signals must also be handled here: when a
+		// hook-triggered continuation runs after the protocol turn has
+		// already completed, tool actions are dispatched (below) with an
+		// empty turnId. Without this, the pending_confirmation is silently
+		// dropped, the permission deferred never resolves, and the session
+		// hangs indefinitely.
+		if (signal.kind === 'pending_confirmation') {
+			this._handleToolReady(signal, sessionKey, '', agent);
```
roblourens: I don't quite understand what this is going to do when confirmation is required. I can look more tomorrow

roblourens (approved):

pwang347 (approved):

PR #321862, opened 2026-06-18
Fix context window widget context mismatch for agent host

pwang347:
image
image


src/vs/platform/agentHost/common/state/protocol/channels-session/state.ts
```diff
@@ -291,6 +291,8 @@ export interface AgentSelection {
  * @category Session Config Types
  */
 export interface SessionConfigPropertySchema extends ConfigPropertySchema {
+	/** JSON Schema: allowed values. Session config values are always strings (re-narrows the numeric-capable base `enum`). */
+	enum?: string[];
```
connor4312: I don't think there's much need to re-narrow here

src/vs/platform/agentHost/common/state/protocol/channels-root/state.ts
```diff
@@ -97,6 +97,8 @@ export interface SessionModelInfo {
 	name: string;
 	/** Maximum context window size */
 	maxContextWindow?: number;
+	/** Maximum context window size for the long-context tier, when the model exposes one */
+	maxLongContextWindow?: number;
```
connor4312, line 101: - changes to these files need to be made upstream in the AHP repo first, otherwise they'll get overwritten next time someone syncs over
- maxLongContextWindow seems like a very copilot/vscode-specific concept. I don't think we'd make it a first-class citizen in the protocol

I'm not really an expert in the flow here, but since nowadays we have the unified picker for the tokens on various sizes, do we actually _need_ to represent this over the protocol or can the agent host be like 'I see you have picked longer context than the lower tier's maxContextWindow, I'll set you to the higher tier' and vise versa?

Image

pwang347: Thanks @connor4312! Meant to close this PR actually. I thought about it more, and I think the way we create a new configuration today for this seems a bit wrong. It should likely be something sent per request, and the model metadata can include the recommended tiers. So my current thinking is to remove the config and map the number to the enum on AHP Copilot side

PR #321867, opened 2026-06-18
Add support for Copilot CLI quotas in agent host

pwang347:


src/vs/platform/agentHost/node/copilot/copilotAgentSession.ts
```diff
@@ -2354,14 +2421,16 @@ export class CopilotAgentSession extends Disposable {
 			}
 			// TODO: `copilotUsage` is marked `asInternal` in the SDK schema so it is not exposed on the generated
 			// `AssistantUsageData` type, but it is present at runtime. Read it dynamically.
-			const copilotUsage = (e.data as unknown as Record<string, unknown>).copilotUsage as { totalNanoAiu?: number } | undefined;
+			const rawUsage = e.data as unknown as Record<string, unknown>;
```
joshspicer, line 2424: Any way we have to validate this is castable?

joshspicer (approved):

src/vs/platform/agentHost/node/copilot/copilotAgentSession.ts
pwang347: I didn't change this part, but good point - I think we should probably validate that `rawUsage.copilotUsage` won't throw

Edit: actually, I think it's fine - that part is typed so we can expect an object

PR #321875, opened 2026-06-18
agentHost: handle no-op Copilot compaction errors

### LINKED ISSUE #321827: /compact errors in AH
`/compact` seems to work the first time, but doesnt actually run compaction. The model name outputs but the credit cost is not shown. Nothing is shown as run in the logs. I tried to do it again and finally am shown an error but it says there is nothing to compact.

Image

DonJayamanne:
Fixes #321827

This treats Copilot CLI history compaction errors containing "nothing to compact" as a successful no-op for /compact. The command now emits the standard "Compaction completed" message and closes the turn instead of surfacing the SDK error.

Also adds regression coverage for case-insensitive matching of that error message.

Validation:
- npm run typecheck-client
- npm run valid-layers-check
- npm run test-node -- --run src/vs/platform/agentHost/test/node/copilotAgentSession.test.ts --grep compact
- npm run hygiene (fails on existing baseline design-token issues outside touched files; touched files are not reported)


justschen (approved):
checks failing tho!

PR #321987, opened 2026-06-18
Fix new chat session button context

### LINKED ISSUE #321049: Switching between Copilot Chat Sessions Destroys associated Terminal Sessions
Type: <b>Bug</b>

Switching between Copilot Chat Sessions Destroys associated agent triggered Terminal Sessions

This is a blocker.  My primary workflow is now broken. if leaving a session results in the associated terminals being lost, my agents can't do their remote work.  And requires human interaction.

This could be VSCode Integrated Terminal manager doing aggressive Garbage Collection.
Copilot Chat: environment variables it adds, like copilot / copilot-debug
VS Code Git: Git askpass and auth plumbing
VS Code Terminal/Workbench: terminal GC, disposal, and session lifecycle

Maybe Terminal garbage collection on agent sessions is too aggressive? 

This is an absolute show stopper.

Current version of VSCode shows this as well as the version i'm running here.

VS Code version: Code 1.121.0 (f6cfa2ea2403534de03f069bdf160d06451ed282, 2026-05-19T11:51:53+02:00)
OS version: Windows_NT x64 10.0.26200
Modes:

meganrogge:
Fixes #321049

When the Chat Sessions viewer is shown side-by-side, its large **New Session** button invoked `workbench.action.chat.newChat` without passing the current chat view action context. The new-chat action then fell back to `IChatWidgetService.lastFocusedWidget`, which can point at a different/stale session after interacting with the sessions list or terminal. That makes this path target the wrong session, while the plus toolbar path does not reproduce because it is invoked with the chat view context.

This changes the side-by-side **New Session** button to pass `getActionsContext()` into the command so the new session is created from the currently displayed chat pane instead of relying on focus fallback.

https://github.com/user-attachments/assets/aa7a424b-a79a-44ad-acf4-2c3a3376f289

### DIFF
--- a/src/vs/workbench/contrib/chat/browser/widgetHosts/viewPane/chatViewPane.ts
+++ b/src/vs/workbench/contrib/chat/browser/widgetHosts/viewPane/chatViewPane.ts
@@ -596,7 +596,7 @@ export class ChatViewPane extends ViewPane implements IViewWelcomeDelegate {
 		const newSessionButtonContainer = this.sessionsNewButtonContainer = append(sessionsContainer, $('.agent-sessions-new-button-container'));
 		const newSessionButton = this._register(new Button(newSessionButtonContainer, { ...defaultButtonStyles, secondary: true }));
 		newSessionButton.label = localize('newSession', "New Session");
-		this._register(newSessionButton.onDidClick(() => this.commandService.executeCommand(ACTION_ID_NEW_CHAT)));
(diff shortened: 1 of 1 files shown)


anthonykim1 (approved):

davemcom:
[I beg to differ.](https://github.com/microsoft/vscode/issues/321049#issuecomment-4746696249).

If this is useful and you're looking for people who debug at this layer, I'd welcome a conversation. If there's a repro-bounty program, 10x my April–May combined AIC spend would be a welcome acknowledgment — along with the chance to join the team.

PR #234858, opened 2024-11-28
Expose adjustWhitespace to TextEditor API.

rgrunber:
- Disable adjustWhitespace in text edits with snippets so that code actions do not adjust snippet indentation by default
- Add testcase
- Fixes #63129 

Based on comments like https://github.com/microsoft/language-server-protocol/pull/1892#issuecomment-1964186341 , snippet edits should probably not be adjusting whitespace by default. For language support, once workspace edits have a property like, https://github.com/microsoft/vscode/blob/275faf6f08b7aa50843f3c18406b4d5969784e52/src/vscode-dts/vscode.d.ts#L4996-L5001 we could set the snippet insertion to respect that.

**Before**

https://github.com/user-attachments/assets/7e1c4819-e127-45f5-82e9-5bede1c26d03

**After**

https://github.com/user-attachments/assets/ae65bbcf-af10-4103-9fc4-31165aefa59f

Issue #63129, opened 2018-11-14
editor.insertSnippet() messes with indenting of the SnippetString


DanTup:
This is a very similar issue to #44200/#57093 but that fix appears to be only being applied to completions, but applying snippets causes the same issues.

Here's a repro:

```ts
vs.workspace.openTextDocument(vs.Uri.parse(`untitled:${os.tmpdir}/a.dart`)).then(async (document) => {

	const originalCode = `main() {
  // code indented with two spaces
}`;
	const editor = await vs.window.showTextDocument(document);
	await editor.edit((eb) => eb.insert(new vs.Position(0, 0), originalCode));

	const startOffset = originalCode.indexOf("// code indented with two spaces");
	const endOffset = originalCode.indexOf("\n}");
	const snippetString = new vs.SnippetString("// new code indented with two spaces\n  // newer code indented with two spaces");

	await editor.insertSnippet(snippetString, new vs.Range(document.positionAt(startOffset), document.positionAt(endOffset)));
});
```

This creates a doc and inserts into it the code:

```dart
main() {
  // code indented with two spaces
}
```

It then applies a snippet string that includes an additional line that is also indented by two spaces. However what you end up with is:

```dart
main() {
  // new code indented with two spaces
    // newer code indented with two spaces
}
```

The inserted line has been re-indented.

Whilst this behaviour may be useful if applying basic hard-coded strings, it's incorrect for language servers that are correctly calculating edits to leave the file indented/formatted correctly (I guess for the same reasons that left to #57093).

DanTup:
(@jrieken FYI - this seems similar to the completion issue you resolved)

DanTup:
@jrieken this is tagged as a feature, but I it's really a bug. Any language server calculating the correct edits exactly as required will end up with bad indenting. There's no way for an extension to really fix this without asking the language server to implement the edit incorrectly.

This was fixed for completions, just not for edits.

DanTup:
@jrieken is this something that could easily be contributed (with a similar fix to whatever was done in completions?). As mentioned above, this is really a bug not a feature - VS Code is assuming a language extension is generating incorrect indenting and doesn't work correctly if it generates the correct indenting.

dancojocaru2000:
Any updates on this? It's really annoying to constantly have to select and delete a huge chunk of indentation. This reason alone makes me almost want to switch to IntelliJ.

DanTup:
[...]

PR #322008, opened 2026-06-18
Fix insertSnippet() destroying active snippet session tabstops 

### LINKED ISSUE #214757: editor.insertSnippet() removes the tabstops from snippets extended by this method

- VS Code Version:  1.86.0 or later
- OS Version: Windows 11 23H2

Steps to Reproduce:

1. Register the command `extension.insertMySnippet` in the extension
```typescript
 let mySnippetCmd = vscode.commands.registerCommand('extension.insertMySnippet', () => {
      let editor = vscode.window.activeTextEditor;
      if (!editor) {
          return; 
      }
      editor.insertSnippet(new vscode.SnippetString("( $1 | $2 )"));
});
```
2. Execute the command `extension.insertMySnippet`, which outputs `( $1 | $2 )$3`. When executing the command again within `$1` of `( $1 | $2 )$3`, it should expand to `( ( $1 | $2 )$3 | $4 )$5`. However, `$2` and `$3` from the first expansion gets removed, resulting in `( ( $1 | $2 ) $3 | )`.

Additionally, downgrading VSCode to version 1.85.2 or reverting [changes](https://github.com/microsoft/vscode/commit/9f5cb9537a691881d67ab12e874dd225449b74d1): R24 and L541-L542,R536-R538 in [mainThreadEditor.ts](https://github.com/microsoft/vscode/blob/1.86.0/src/vs/workbench/api/browser/mainThreadEditor.ts) can fix this bug for me.

almayor:
## Problem

Related: #322010

Since VS Code 1.86, calling `editor.insertSnippet()` from an extension while a snippet session is active destroys the outer session's remaining tabstops. This is a regression — it worked correctly in 1.85.2 and earlier.

This breaks extensions like HyperSnips, LTeX, and others that programmatically insert snippets, making nested snippet expansion unusable.

## Root cause

In 1.86, `mainThreadEditor.ts` was refactored to route `insertSnippet()` calls through `snippetController.apply(edits[])` instead of `snippetController.insert(template)`. The `apply()` path calls `this.cancel()` when a session is active:

```typescript
if (this._session && typeof template !== 'string') {
    this.cancel();
}
```

The `insert()` path instead calls `this._session.merge(template, opts)`, which properly nests the new snippet inside the active placeholder. The merge infrastructure (`OneSnippet.merge()`, `_nestingLevel`, `_renormalizePlaceholderIndices()`) is fully implemented and works — the regression was the API bridge switching code paths.

## Fix

Route single-edit `insertSnippet()` calls back through `snippetController.insert()` (the merge path), falling back to `apply()` only for multi-range edits.

## Note on issue #214757

This bug was reported in #214757, which was closed as a duplicate of #63129. However, #63129 is about snippet auto-indentation (fixed in #234858), not tabstop destruction — they are unrelated issues. The bug remains present through current versions.

Fixes #214757

## Test

1. Install [HyperSnips](https://marketplace.visualstudio.com/items?itemName=draivin.hsnips) or any extension that calls `editor.insertSnippet()` during editing
2. Trigger a snippet with multiple tabstops (e.g. `\int_{$1}^{$2} $3 d$4`)
3. While cursor is at `$1`, trigger another snippet insertion via the extension
4. Press Tab — cursor should advance to `$2` (currently it doesn't; the session is destroyed)

---

This fix was researched and developed with the assistance of [Claude Code](https://claude.ai/claude-code) (Claude Opus 4.6).

### DIFF
--- a/src/vs/workbench/api/browser/mainThreadEditor.ts
+++ b/src/vs/workbench/api/browser/mainThreadEditor.ts
@@ -545,24 +545,16 @@ export class MainThreadTextEditor {
 			clipboardText
 		};
 
-		if (ranges.length === 1) {
-			// When there is a single, contiguous snippet edit, route it through the string-based
-			// `insert` API instead of the array-based `apply` API. The `apply` path always cancels
-			// an active snippet session (see the `this.cancel()` guard in SnippetController2._doInsert),
-			// which destroys the outer snippet's remaining tabstops. The `insert` path instead merges
-			// (nests) the new snippet into the active placeholder, preserving the outer session.
-			// This restores the pre-1.86 behaviour and fixes microsoft/vscode#214757.
-			const range = Range.lift(ranges[0]);
-			this._codeEditor.setSelection(new Selection(range.startLineNumber, range.startColumn, range.endLineNumber, range.endColumn));
-			snippetController.insert(template, insertOptions);
-			return true;
+		if (ranges.length > 0) {
+			const selections = ranges.map(r => {
+				const range = Range.lift(r);
+				return new Selection(range.startLineNumber, range.startColumn, range.endLineNumber, range.endColumn);
+			});
+			this._codeEditor.setSelections(selections);
 		}
 
-		// Multi-range edits cannot be merged into a single active placeholder, so fall back to the
-		// array-based `apply` API.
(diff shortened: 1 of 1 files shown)


almayor:

PR #322025, opened 2026-06-19
Update strictKnownMarketplaces managed setting to support rich types

vijayupadya:
Fixes https://github.com/microsoft/vscode-internalbacklog/issues/8059

Update strictKnownMarketplaces managed setting that restricts which marketplace sources users can add and install plugins from. 
This only blocks installation, and does not retroactively prevent loading plugins that are already installed.

Behavior enabled:
- Restricts which marketplace sources users can add/install plugins from
- Blocks direct plugin installation from non-approved sources
- Only blocks installation — does not retroactively unload already-installed plugins
- Shows UX when install is blocked by policy


vijayupadya:
cc @joshspicer

src/vs/workbench/contrib/chat/browser/pluginInstallService.ts
```diff
@@ -423,6 +423,23 @@ export class PluginInstallService implements IPluginInstallService {
...
+				severity: Severity.Warning,
+				message: localize('strictMarketplaceBlockedInstall', "Plugins from '{0}' are blocked by your organization's policy.", plugin.marketplaceReference.displayLabel),
+				actions: {
+					primary: [new Action('chat.plugins.viewMarketplacePolicy', localize('viewPolicySettings', "View Policy Settings"), undefined, true, () => {
+						return this._commandService.executeCommand('workbench.action.openSettings', ChatConfiguration.StrictMarketplaces);
+					})],
+				},
+			});
+			return false;
+		}
```
connor4312, line 441: Would this prevent the user from ever being able to trust a marketplace for the first time when the policy is active?

vijayupadya: Good question - Allowlisted marketplaces aren't affected — when the strict policy is active, isMarketplaceTrusted() is governed by the allowlist (it ignores the user-trust store), so a marketplace on the org allowlist returns true from the earlier check and installs without any prompt. This branch is only reached for marketplaces not on the allowlist, where blocking (rather than offering a trust dialog) is by design — under strict mode the user must not be able to self-grant trust to bypass the enterprise allowlist.

alexr00 (approved):

joshspicer:
Tracking under the **Unified Settings** workstream → microsoft/vscode-internalbacklog#7940 (`strictKnownMarketplaces` managed setting, microsoft/vscode-internalbacklog#8059).

PR #322030, opened 2026-06-19
Implement CoalescingAgentHostSessionListConnection to optimize session list retrieval

### LINKED ISSUE #320303: Switching sessions is too slow
One problem is that we do `listSessions` way too often, and this is slow and blocks `resumeSession`. 

When we open a session for the first time, we add it as a workspace folder, and this triggers multiple refreshes

https://github.com/microsoft/vscode/blob/6a335631e0ecccb86591de56bb364f74da69db61/src/vs/workbench/contrib/chat/browser/agentSessions/agentHost/agentHostSessionListController.ts#L169-L172

https://github.com/microsoft/vscode/blob/6a335631e0ecccb86591de56bb364f74da69db61/src/vs/workbench/contrib/chat/browser/agentSessions/agentSessionsModel.ts#L534

DonJayamanne:
Fixes https://github.com/microsoft/vscode/issues/320303

### DIFF
--- a/src/vs/workbench/contrib/chat/browser/agentSessions/agentHost/agentHostSessionListController.ts
+++ b/src/vs/workbench/contrib/chat/browser/agentSessions/agentHost/agentHostSessionListController.ts
@@ -5,13 +5,14 @@
 
 import { CancellationToken } from '../../../../../../base/common/cancellation.js';
 import { Codicon } from '../../../../../../base/common/codicons.js';
-import { Emitter } from '../../../../../../base/common/event.js';
+import { Emitter, Event } from '../../../../../../base/common/event.js';
 import { Disposable } from '../../../../../../base/common/lifecycle.js';
 import { extUriBiasedIgnorePathCase } from '../../../../../../base/common/resources.js';
 import { URI } from '../../../../../../base/common/uri.js';
 import { generateUuid } from '../../../../../../base/common/uuid.js';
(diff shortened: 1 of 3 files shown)


src/vs/workbench/contrib/chat/browser/agentSessions/agentHost/agentHostSessionListController.ts
```diff
@@ -217,6 +228,19 @@ export class AgentHostSessionListController extends Disposable implements IChatS
...
+
+		this._refreshInFlight = this._doRefresh(token);
+		try {
+			await this._refreshInFlight;
+		} finally {
+			this._refreshInFlight = undefined;
+		}
+	}
+
+	private async _doRefresh(token: CancellationToken): Promise<void> {
```
connor4312, line 243: I think you identified a good general direction but I wonder if we want to pull a bit more up here. We have each individual insance of the AgentHostSessionListController keeping its _cacheValid and firing item change events, just filtering those to its session type.

I wonder if instead we want to have a like `CoalesedAgentHostSessionListController` (idk about the name) that does all the meaty logic here, and then have the actual `AgentHostSessionListController` be a very thin layer that just wraps that and filters the various events and methods to only surface the sessions that are relevant for that agent type

DonJayamanne: sure will do that.

justschen (approved):

PR #322043, opened 2026-06-19
Prototype redesign of customizations plugin page

hawkticehurst:
WIP

### DIFF
--- a/src/vs/sessions/AI_CUSTOMIZATIONS.md
+++ b/src/vs/sessions/AI_CUSTOMIZATIONS.md
@@ -225,10 +225,10 @@ Hooks and other non-markdown detail views continue to open directly in their exi
 
 ### Plugins Section
 
-The Plugins section uses a bolder card-based home surface while preserving `WorkbenchList` as the search and accessibility-safe fallback. The default **Plugins Home** shows installed plugins as cards, remote/session plugins as operational cards, and a progressive marketplace discovery snapshot below the installed content.
+The Plugins section uses a bolder single-column card-list home surface while preserving `WorkbenchList` as the search and accessibility-safe fallback. The default **Plugins Home** shows installed plugins as a flat card list, remote/session plugins as operational cards, and a progressive marketplace discovery snapshot below the installed content.
 
 Installed plugin cards show:
-- plugin identity, source/provenance, enabled/disabled state, and description
+- plugin identity, source/provenance, enabled/disabled state, and description without decorative image placeholders
 - contribution preview chips derived from real installed plugin data (`agents`, `skills`, `commands`, `instructions`, `mcpServerDefinitions`, `hooks`)
 - explicit action buttons (`View`, `More Actions`) rather than making the card itself clickable
 
@@ -240,6 +240,8 @@ Browse mode is a plugin-local destination with an explicit **Back to Installed**
 Marketplace cards only show metadata backed by current data (`name`, `description`, marketplace label, recommendation key). They must not show unverifiable claims such as popularity, verification, ratings, screenshots, or generic trust badges.
 
 The embedded plugin detail view shows identity, linked source/provenance, inline enable/disable/uninstall actions for installed plugins, a copyable local path, description, and full installed contribution inventory grouped by type under **Included capabilities**. Marketplace detail stays honest about data limits and explains that structured contribution details are available after install unless a future marketplace schema exposes them.
+
+Installed plugin cards are sorted alphabetically and do not move when enablement changes. Cards use a compact neutral switch control for enablement; the switch thumb should remain visually prominent within the rounded track, animate between states, avoid accent-color backgrounds, and honor VS Code's reduced-motion mode. Plugin README markdown is rendered through `IMarkdownRendererService` and scoped with VS Code markdown tokens for links, code, block quotes, tables, and compact headings so it matches the active theme inside the split-pane detail view. Marketplace plugin README links should point at the plugin's source repository/path when the marketplace entry declares a source repo, rather than always pointing at the marketplace index repository.
 
 ### AgenticPromptsService (Sessions)
 
--- a/src/vs/sessions/AI_CUSTOMIZATIONS.md
+++ b/src/vs/sessions/AI_CUSTOMIZATIONS.md
@@ -223,6 +223,24 @@ For markdown-backed customizations (`.agent.md`, `SKILL.md`, `.instructions.md`,
 
 Hooks and other non-markdown detail views continue to open directly in their existing raw/detail experiences.
 
+### Plugins Section
+
+The Plugins section uses a bolder card-based home surface while preserving `WorkbenchList` as the search and accessibility-safe fallback. The default **Plugins Home** shows installed plugins as cards, remote/session plugins as operational cards, and a progressive marketplace discovery snapshot below the installed content.
+
+Installed plugin cards show:
+- plugin identity, source/provenance, enabled/disabled state, and description
+- contribution preview chips derived from real installed plugin data (`agents`, `skills`, `commands`, `instructions`, `mcpServerDefinitions`, `hooks`)
+- explicit action buttons (`View`, `More Actions`) rather than making the card itself clickable
+
+Browse mode is a plugin-local destination with an explicit **Back to Installed** button. It renders the full marketplace as cards when not searching:
+- **Recommended for this workspace** appears only when backed by `IPluginMarketplaceService.recommendedPlugins`
+- **All plugins** shows remaining uninstalled marketplace items
+- Search switches to a flat `WorkbenchList` result view so keyboarding and screen-reader behavior stay predictable
+
+Marketplace cards only show metadata backed by current data (`name`, `description`, marketplace label, recommendation key). They must not show unverifiable claims such as popularity, verification, ratings, screenshots, or generic trust badges.
+
+The embedded plugin detail view shows identity, linked source/provenance, inline enable/disable/uninstall actions for installed plugins, a copyable local path, description, and full installed contribution inventory grouped by type under **Included capabilities**. Marketplace detail stays honest about data limits and explains that structured contribution details are available after install unless a future marketplace schema exposes them.
+
 ### AgenticPromptsService (Sessions)
 
 Sessions overrides `PromptsService` via `AgenticPromptsService` (in `promptsService.ts`):
--- a/src/vs/workbench/contrib/chat/browser/aiCustomization/aiCustomizationManagementEditor.ts
+++ b/src/vs/workbench/contrib/chat/browser/aiCustomization/aiCustomizationManagementEditor.ts
@@ -2341,6 +2341,18 @@ export class AICustomizationManagementEditor extends EditorPane {
 		const detailBody = DOM.append(this.pluginDetailContainer, $('.plugin-detail-editor-container'));
 
 		this.embeddedPluginDetail = this.editorDisposables.add(this.instantiationService.createInstance(EmbeddedAgentPluginDetail, detailBody));
+		this.editorDisposables.add(this.embeddedPluginDetail.onDidRequestOpenSkill(uri => {
+			this.openSkillFromPluginDetail(uri);
+		}));
+		this.editorDisposables.add(this.embeddedPluginDetail.onDidRequestOpenAgent(uri => {
+			this.openPromptsItemFromPluginDetail(AICustomizationManagementSection.Agents, uri);
+		}));
+		this.editorDisposables.add(this.embeddedPluginDetail.onDidRequestOpenSection(section => {
+			this.openSectionFromPluginDetail(section);
+		}));
+		this.editorDisposables.add(this.embeddedPluginDetail.onDidUninstall(() => {
+			this.goBackFromPluginDetail();
+		}));
 
 		// Back button rendered into the detail's leading slot
 		const backButton = DOM.append(this.embeddedPluginDetail.leadingSlot, $('button.editor-back-button'));
@@ -2368,6 +2380,46 @@ export class AICustomizationManagementEditor extends EditorPane {
 		if (this.dimension) {
 			this.layout(this.dimension);
 		}
+	}
+
+	private async openSkillFromPluginDetail(uri: URI): Promise<void> {
+		await this.openPromptsItemFromPluginDetail(AICustomizationManagementSection.Skills, uri);
+	}
+
+	private async openPromptsItemFromPluginDetail(section: AICustomizationManagementSection, uri: URI): Promise<void> {
+		this.pluginDetailDisposables.clear();
+		this.embeddedPluginDetail?.clearInput();
+		this.pluginDetailReturnSection = undefined;
+		this.viewMode = 'list';
+		this.selectedSection = section;
+		this.sectionContextKey.set(section);
+		this.storageService.store(AI_CUSTOMIZATION_MANAGEMENT_SELECTED_SECTION_KEY, section, StorageScope.PROFILE, StorageTarget.USER);
+		this.updateContentVisibility();
+		await this.listWidget.setSection(section);
+		const modelSection = ITEMS_MODEL_SECTIONS.find(s => s === section);
+		if (modelSection) {
+			await this.itemsModel.whenSectionLoaded(modelSection);
+			const item = this.itemsModel.getItems(modelSection).get().find(item => isEqual(item.uri, uri));
+			if (item) {
+				const source = item.source;
+				const isWorkspaceFile = source === AICustomizationSources.local;
+				const isReadOnly = !source || source === AICustomizationSources.extension || source === AICustomizationSources.plugin || source === AICustomizationSources.builtin;
+				this.showEmbeddedEditor(item.uri, item.name, item.promptType, source ?? AICustomizationSources.builtin, isWorkspaceFile, isReadOnly);
+			}
+		}
+		this.ensureSectionsListReflectsActiveSection(section);
+		if (this.dimension) {
+			this.layout(this.dimension);
+		}
+	}
+
+	private openSectionFromPluginDetail(section: AICustomizationManagementSection): void {
+		this.pluginDetailDisposables.clear();
+		this.embeddedPluginDetail?.clearInput();
+		this.pluginDetailReturnSection = undefined;
+		this.viewMode = 'list';
+		this.updateContentVisibility();
+		this.selectSectionById(section);
 	}
 
 	/**
--- a/src/vs/workbench/contrib/chat/browser/aiCustomization/embeddedAgentPluginDetail.ts
+++ b/src/vs/workbench/contrib/chat/browser/aiCustomization/embeddedAgentPluginDetail.ts
@@ -4,28 +4,40 @@
  *--------------------------------------------------------------------------------------------*/
 
 import * as DOM from '../../../../../base/browser/dom.js';
-import { Button } from '../../../../../base/browser/ui/button/button.js';
+import { Button, ButtonWithDropdown } from '../../../../../base/browser/ui/button/button.js';
+import { getDefaultHoverDelegate } from '../../../../../base/browser/ui/hover/hoverDelegateFactory.js';
 import { status } from '../../../../../base/browser/ui/aria/aria.js';
 import { disposableTimeout } from '../../../../../base/common/async.js';
+import { CancellationToken } from '../../../../../base/common/cancellation.js';
 import { Codicon } from '../../../../../base/common/codicons.js';
 import { Emitter } from '../../../../../base/common/event.js';
+import { MarkdownString } from '../../../../../base/common/htmlContent.js';
 import { Disposable, DisposableStore, MutableDisposable } from '../../../../../base/common/lifecycle.js';
+import { Schemas } from '../../../../../base/common/network.js';
+import { asText, IRequestService } from '../../../../../platform/request/common/request.js';
 import { localize } from '../../../../../nls.js';
 import { AgentPluginItemKind, IAgentPluginItem } from '../agentPluginEditor/agentPluginItems.js';
 import { IMarketplacePlugin, PluginSourceKind } from '../../common/plugins/pluginMarketplaceService.js';
 import { IPluginInstallService } from '../../common/plugins/pluginInstallService.js';
-import { isContributionEnabled } from '../../common/enablement.js';
+import { ContributionEnablementState, isContributionEnabled } from '../../common/enablement.js';
 import { ILabelService } from '../../../../../platform/label/common/label.js';
 import { defaultButtonStyles } from '../../../../../platform/theme/browser/defaultStyles.js';
 import { IClipboardService } from '../../../../../platform/clipboard/common/clipboardService.js';
 import { IOpenerService } from '../../../../../platform/opener/common/opener.js';
 import { IAgentPluginService } from '../../common/plugins/agentPluginService.js';
-import { IWorkspaceContextService } from '../../../../../platform/workspace/common/workspace.js';
-import { UninstallPluginAction, createDisablePluginDropDown, createEnablePluginDropDown, createPolicyBlockedEnableAction, isPluginPolicyBlocked } from '../agentPluginActions.js';
+import { UninstallPluginAction, createPolicyBlockedEnableAction, isPluginPolicyBlocked } from '../agentPluginActions.js';
 import { INotificationService } from '../../../../../platform/notification/common/notification.js';
 import { URI } from '../../../../../base/common/uri.js';
-import { basename } from '../../../../../base/common/resources.js';
+import { basename, dirname, joinPath } from '../../../../../base/common/resources.js';
 import { AICustomizationManagementSection } from '../../common/aiCustomizationWorkspaceService.js';
+import { IFileService } from '../../../../../platform/files/common/files.js';
+import { IMarkdownRendererService } from '../../../../../platform/markdown/browser/markdownRenderer.js';
+import { IContextMenuService } from '../../../../../platform/contextview/browser/contextView.js';
+import { Action } from '../../../../../base/common/actions.js';
+import { ICommandService } from '../../../../../platform/commands/common/commands.js';
+import { IHoverService } from '../../../../../platform/hover/browser/hover.js';
+import type { IContextMenuProvider } from '../../../../../base/browser/contextmenu.js';
+import { AnchorAlignment } from '../../../../../base/browser/ui/contextview/contextview.js';
 
 const $ = DOM.$;
 
@@ -57,6 +69,8 @@ export class EmbeddedAgentPluginDetail extends Disposable {
 	private readonly factsEl: HTMLElement;
 	private readonly contributionsEl: HTMLElement;
 	private readonly contributionsListEl: HTMLElement;
+	private readonly readmeEl: HTMLElement;
+	private readonly readmeContentEl: HTMLElement;
 	private readonly emptyEl: HTMLElement;
 	private readonly renderDisposables = this._register(new DisposableStore());
 	private readonly copyStateReset = this._register(new MutableDisposable());
@@ -70,8 +84,13 @@ export class EmbeddedAgentPluginDetail extends Disposable {
 		@IOpenerService private readonly openerService: IOpenerService,
 		@IAgentPluginService private readonly agentPluginService: IAgentPluginService,
 		@IPluginInstallService private readonly pluginInstallService: IPluginInstallService,
-		@IWorkspaceContextService private readonly workspaceContextService: IWorkspaceContextService,
 		@INotificationService private readonly notificationService: INotificationService,
+		@IContextMenuService private readonly contextMenuService: IContextMenuService,
+		@ICommandService private readonly commandService: ICommandService,
+		@IHoverService private readonly hoverService: IHoverService,
+		@IFileService private readonly fileService: IFileService,
+		@IRequestService private readonly requestService: IRequestService,
+		@IMarkdownRendererService private readonly markdownRendererService: IMarkdownRendererService,
 	) {
 		super();
 
@@ -93,6 +112,11 @@ export class EmbeddedAgentPluginDetail extends Disposable {
 		this.factsEl = DOM.append(this.root, $('.embedded-detail-facts'));
 		this.contributionsEl = DOM.append(this.root, $('.embedded-detail-section.plugin-detail-contributions'));
 		this.contributionsListEl = DOM.append(this.contributionsEl, $('.embedded-detail-chip-list'));
+		this.readmeEl = DOM.append(this.root, $('.embedded-detail-section.plugin-detail-readme'));
+		const readmeTitle = DOM.append(this.readmeEl, $('h3.plugin-detail-contribution-group-title'));
+		const readmeLabel = DOM.append(readmeTitle, $('span.plugin-detail-contribution-title-label'));
+		readmeLabel.textContent = localize('pluginReadmeTitle', "Plugin README");
+		this.readmeContentEl = DOM.append(this.readmeEl, $('.plugin-detail-readme-content'));
 
 		this.emptyEl = DOM.append(this.root, $('.embedded-detail-empty'));
 		this.emptyEl.textContent = localize('pluginDetailEmpty', "No plugin selected.");
@@ -142,12 +166,14 @@ export class EmbeddedAgentPluginDetail extends Disposable {
 			DOM.clearNode(this.factsEl);
 			DOM.clearNode(this.contributionsListEl);
 			this.contributionsEl.style.display = 'none';
+			DOM.clearNode(this.readmeContentEl);
+			this.readmeEl.style.display = 'none';
 			return;
 		}
 
 		this.nameEl.textContent = item.name;
 		if (item.kind === AgentPluginItemKind.Installed && !isContributionEnabled(item.plugin.enablement.get())) {
-			this.statusBadgeEl.textContent = localize('pluginDetailDisabledBadge', "Disabled");
+			this.statusBadgeEl.textContent = getInstalledPluginDisabledLabel(item);
 			this.statusBadgeEl.style.display = '';
 		} else {
 			this.statusBadgeEl.textContent = '';
@@ -170,6 +196,7 @@ export class EmbeddedAgentPluginDetail extends Disposable {
 		this.renderTitleActions(item);
 		this.renderFacts(item);
 		this.renderContributions(item);
+		this.renderReadme(item);
 
 		const description = (item.description || '').trim();
 		this.descriptionEl.textContent = description || localize('pluginNoDescription', "No description provided.");
@@ -221,20 +248,54 @@ export class EmbeddedAgentPluginDetail extends Disposable {
 			}));
 		}
 
-		const enablementAction = isPluginPolicyBlocked(item.plugin)
-			? createPolicyBlockedEnableAction(item.plugin, this.notificationService)
-			: isContributionEnabled(item.plugin.enablement.get())
-				? createDisablePluginDropDown(item.plugin, this.agentPluginService.enablementModel, this.workspaceContextService)
-				: createEnablePluginDropDown(item.plugin, this.agentPluginService.enablementModel, this.workspaceContextService);
-		const enablementButton = this.renderDisposables.add(new Button(this.titleActionsEl, { ...defaultButtonStyles, secondary: true, supportIcons: true, ariaLabel: enablementAction.label }));
-		enablementButton.element.classList.add(isContributionEnabled(item.plugin.enablement.get()) ? 'embedded-detail-disable-button' : 'embedded-detail-enable-button');
-		enablementButton.label = enablementAction.label;
-		enablementButton.enabled = enablementAction.enabled;
-		this.renderDisposables.add(enablementButton.onDidClick(async () => {
-			await enablementAction.run();
+		this.renderEnablementSplitButton(item);
+	}
+
+	private renderEnablementSplitButton(item: Extract<IAgentPluginItem, { kind: AgentPluginItemKind.Installed }>): void {
+		if (isPluginPolicyBlocked(item.plugin)) {
+			const action = createPolicyBlockedEnableAction(item.plugin, this.notificationService);
+			const button = this.renderDisposables.add(new Button(this.titleActionsEl, { ...defaultButtonStyles, secondary: true, supportIcons: true, ariaLabel: action.label }));
+			button.label = action.label;
+			this.renderDisposables.add(button.onDidClick(() => action.run()));
+			this.renderDisposables.add(action);
+			return;
+		}
+
+		const current = item.plugin.enablement.get();
+		const isEnabled = isContributionEnabled(current);
+		const primaryLabel = isEnabled ? localize('disable', "Disable") : localize('enable', "Enable");
+		const workspaceLabel = isEnabled ? localize('disableWorkspace', "Disable (Workspace)") : localize('enableWorkspace', "Enable (Workspace)");
+		const primaryState = isEnabled ? ContributionEnablementState.DisabledProfile : ContributionEnablementState.EnabledProfile;
+		const workspaceState = isEnabled ? ContributionEnablementState.DisabledWorkspace : ContributionEnablementState.EnabledWorkspace;
+		const key = item.plugin.uri.toString();
+		const contextMenuProvider: IContextMenuProvider = {
+			showContextMenu: delegate => this.contextMenuService.showContextMenu({
+				...delegate,
+				anchorAlignment: AnchorAlignment.RIGHT,
+			}),
+		};
+		const splitButton = this.renderDisposables.add(new ButtonWithDropdown(this.titleActionsEl, {
+			...defaultButtonStyles,
+			secondary: true,
+			supportIcons: true,
+			contextMenuProvider,
+			addPrimaryActionToDropdown: false,
+			actions: {
+				getActions: () => [
+					this.renderDisposables.add(new Action(`plugin.${isEnabled ? 'disable' : 'enable'}Workspace`, workspaceLabel, undefined, true, async () => {
+						this.agentPluginService.enablementModel.setEnabled(key, workspaceState);
+						this.renderItem();
+					}))
+				]
+			},
+			ariaLabel: primaryLabel,
+		}));
+		splitButton.element.classList.add(isEnabled ? 'embedded-detail-disable-button' : 'embedded-detail-enable-button');
+		splitButton.label = primaryLabel;
+		this.renderDisposables.add(splitButton.onDidClick(() => {
+			this.agentPluginService.enablementModel.setEnabled(key, primaryState);
 			this.renderItem();
 		}));
-		this.renderDisposables.add(enablementAction);
 	}
 
 	private getInstalledPluginForMarketplaceItem(item: Extract<IAgentPluginItem, { kind: AgentPluginItemKind.Marketplace }>): IAgentPluginItem | undefined {
@@ -314,20 +375,80 @@ export class EmbeddedAgentPluginDetail extends Disposable {
 		const label = DOM.append(container, $('span.embedded-detail-location-label'));
 		label.textContent = this.labelService.getUriLabel(uri, { relative: true });
 		label.title = uri.fsPath || uri.toString();
-		const copyButton = this.renderDisposables.add(new Button(container, { ...defaultButtonStyles, secondary: true, supportIcons: true, ariaLabel: localize('copyPluginPath', "Copy Plugin Path") }));
+		const copyPluginPathLabel = localize('copyPluginPath', "Copy Plugin Path");
+		let copyPluginPathTooltip = copyPluginPathLabel;
+		const copyButton = this.renderDisposables.add(new Button(container, { ...defaultButtonStyles, secondary: true, supportIcons: true, title: copyPluginPathLabel, ariaLabel: copyPluginPathLabel }));
 		copyButton.element.classList.add('embedded-detail-copy-button');
 		copyButton.label = `$(${Codicon.copy.id})`;
+		this.renderDisposables.add(this.hoverService.setupManagedHover(getDefaultHoverDelegate('element'), copyButton.element, () => copyPluginPathTooltip));
 		this.renderDisposables.add(copyButton.onDidClick(async () => {
 			await this.clipboardService.writeText(uri.fsPath || uri.toString());
 			copyButton.label = `$(${Codicon.check.id})`;
-			copyButton.setTitle(localize('copiedPluginPath', "Copied"));
+			copyPluginPathTooltip = localize('copiedPluginPath', "Copied");
+			copyButton.setTitle(copyPluginPathTooltip);
 			status(localize('copiedPluginPathStatus', "Copied plugin path to clipboard"));
 			this.copyStateReset.value = disposableTimeout(() => {
 				copyButton.label = `$(${Codicon.copy.id})`;
-				copyButton.setTitle(localize('copyPluginPath', "Copy Plugin Path"));
+				copyPluginPathTooltip = copyPluginPathLabel;
+				copyButton.setTitle(copyPluginPathTooltip);
 			}, 1200);
 		}));
+		const openPluginFolderLabel = localize('openPluginFolder', "Open Plugin Folder");
+		const openButton = this.renderDisposables.add(new Button(container, { ...defaultButtonStyles, secondary: true, supportIcons: true, title: openPluginFolderLabel, ariaLabel: openPluginFolderLabel }));
+		openButton.element.classList.add('embedded-detail-copy-button');
+		openButton.label = `$(${Codicon.folderOpened.id})`;
+		this.renderDisposables.add(this.hoverService.setupManagedHover(getDefaultHoverDelegate('element'), openButton.element, openPluginFolderLabel));
+		this.renderDisposables.add(openButton.onDidClick(async () => {
+			try {
+				await this.commandService.executeCommand('revealFileInOS', uri);
+			} catch {
+				await this.openerService.open(dirname(uri));
+			}
+		}));
 		return container;
+	}
+
+	private async renderReadme(item: IAgentPluginItem): Promise<void> {
+		DOM.clearNode(this.readmeContentEl);
+		this.readmeEl.style.display = 'none';
+		const readme = await this.fetchReadme(item);
+		if (this.current !== item || !readme.trim()) {
+			return;
+		}
+		const rendered = this.renderDisposables.add(this.markdownRendererService.render(new MarkdownString(readme, { supportHtml: false })));
+		this.readmeContentEl.appendChild(rendered.element);
+		this.readmeEl.style.display = '';
+	}
+
+	private async fetchReadme(item: IAgentPluginItem): Promise<string> {
+		const readmeUri = item.kind === AgentPluginItemKind.Installed
+			? joinPath(item.plugin.uri, 'README.md')
+			: item.readmeUri;
+		if (!readmeUri) {
+			return '';
+		}
+		if (readmeUri.scheme === Schemas.file || readmeUri.scheme === Schemas.vscodeRemote) {
+			try {
+				const content = await this.fileService.readFile(readmeUri);
+				return content.value.toString();
+			} catch {
+				return '';
+			}
+		}
+		if (readmeUri.scheme === Schemas.https) {
+			let rawUrl = readmeUri.toString();
+			const githubBlobMatch = rawUrl.match(/^https:\/\/github\.com\/(?<owner>[^/]+)\/(?<repo>[^/]+)\/blob\/(?<rest>.+)$/);
+			if (githubBlobMatch?.groups) {
+				rawUrl = `https://raw.githubusercontent.com/${githubBlobMatch.groups['owner']}/${githubBlobMatch.groups['repo']}/${githubBlobMatch.groups['rest']}`;
+			}
+			try {
+				const context = await this.requestService.request({ type: 'GET', url: rawUrl, callSite: 'aiCustomizationPluginDetail.fetchReadme' }, CancellationToken.None);
+				return await asText(context) ?? '';
+			} catch {
+				return '';
+			}
+		}
+		return '';
 	}
 
 	private renderContributions(item: IAgentPluginItem): void {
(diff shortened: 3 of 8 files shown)


src/vs/workbench/contrib/chat/browser/aiCustomization/mcpListWidget.ts
```diff
@@ -1169,106 +1070,37 @@ export class McpListWidget extends Disposable {
...
+			this.refresh();
 		}));
 		this._register(autorun(reader => {
-			this.mcpService.servers.read(reader);
-			if (!this.browseMode) {
-				this.refresh();
+			const servers = this.mcpService.servers.read(reader);
+			for (const server of servers) {
+				server.enablement.read(reader);
+				server.connectionState.read(reader);
```
sbatten: 🤖 **[P1] This reintroduces the node-replacement click race fixed in #330890.** Reading every server's connection state causes `refresh()` on each status tick, and both card render paths clear all disposables and replace the complete card DOM. An erroring server can therefore replace More/Install between mousedown and mouseup, lose focus/scroll, and leave `installMarketplaceServer` updating a detached button. The old renderer's identity guards no longer protect the UI that is actually displayed. Please preserve live card/action node identity and add a regression test for the shipped card path.

src/vs/workbench/contrib/chat/browser/aiCustomization/toolsListWidget.ts
```diff
@@ -617,31 +668,37 @@ export class ToolsListWidget extends Disposable {
...
 
 		const count = DOM.append(row, $('span.tools-list-row-count'));
+		if (vm.readOnly) {
+			DOM.append(row, $('span.tools-list-always-available')).textContent = localize('toolsAlwaysAvailable', "Always Available");
+		}
+		const extension = this._resolveExtensionForToolSet(ts);
+		let moreButton: HTMLButtonElement | undefined;
+		if (extension) {
+			const moreLabel = localize('toolsSetMoreActions', "More actions for {0}", setName);
+			moreButton = DOM.append(row, $('button.tools-list-more-action')) as HTMLButtonElement;
```
sbatten, line 682: 🤖 **[P1] This More button cannot be activated from the keyboard.** It remains a native tab stop, but Enter/Space bubbles to `_onTreeKeyDown`, which resolves the parent row, clicks the enablement toggle, and calls `preventDefault()`. That suppresses the button's native click, so keyboard users toggle the tool set instead of opening this menu. The former row context-menu path was also removed. Please ignore keyboard events from interactive descendants and restore a keyboard/context-menu action path.

src/vs/workbench/contrib/chat/browser/aiCustomization/customizationMigrationCategories.ts
```diff
@@ -290,15 +293,8 @@ const userDataMigrationCategory: ICustomizationMigrationCategory = {
...
-		const { totalCount } = countUserDataTypes(customizations);
-
+	getBanner(_customizations, harnessLabel, destinationLabel) {
 		return {
-			title: totalCount === 1
-				? localize('userDataMigrationBannerTitleSingle', "1 customization is not available to {0}", harnessLabel)
-				: localize('userDataMigrationBannerTitle', "{0} customizations are not available to {1}", totalCount, harnessLabel),
-			// The grouped list below already breaks these down by type, so the
-			// message explains the move rather than repeating the counts.
 			message: destinationLabel
```
sbatten, line 299: 🤖 **[P1] This redesign drops the Settings Sync consequence added by #331529 to fix #331211.** The confirmation still defaults “Delete original files” to checked, so migration removes the synced `promptsHome` copies and moves them to destinations such as `~/.copilot`, which Settings Sync does not cover. The new banner only says both clients can use the files; it no longer warns that cross-device availability is lost. Please restore destination-aware sync disclosure, or avoid default deletion for unsynced destinations.

src/vs/workbench/contrib/chat/browser/aiCustomization/aiCustomizationListWidget.ts
```diff
@@ -499,6 +499,20 @@ function toItemsModelSection(section: AICustomizationManagementSection): ItemsMo
 	}
 }
 
+export function usesCustomizationCardLayout(section: AICustomizationManagementSection): boolean {
```
sbatten, line 503: 🤖 **[P1] Routing all five primary sections to the custom card layout removes the previous composite-list accessibility behavior.** Agents, Skills, Instructions, Hooks, and Prompts no longer get `WorkbenchList` list semantics, arrow/Home/End navigation, type-ahead, or positional announcements. Each row and overflow action is now an independent Tab stop with no list/listbox role or roving tabindex. Please implement the expected composite-list keyboard/ARIA pattern rather than leaving the accessible list instantiated but unused.

src/vs/workbench/contrib/chat/browser/aiCustomization/pluginListWidget.ts
```diff
@@ -845,6 +1095,472 @@ export class PluginListWidget extends Disposable {
...
+
+	private showEmptySurface(): void {
+		this.cardContainer.style.display = 'none';
+		this.listContainer.style.display = 'none';
+		this.emptyContainer.style.display = 'flex';
+	}
+
+	private addSurfaceActivation(surface: HTMLElement, label: string, callback: () => void): void {
+		surface.tabIndex = 0;
+		surface.setAttribute('role', 'button');
```
sbatten: 🤖 **[P1] The outer plugin row/card is exposed as a button while containing other native controls.** Installed rows contain a switch and More button, and marketplace cards contain Install buttons. Stopping click propagation does not repair the accessibility tree; this creates competing/nested interactive semantics for screen readers and keyboard users. Please use a non-interactive group/container plus a dedicated details control, or a valid composite-widget pattern.

src/vs/workbench/contrib/chat/browser/aiCustomization/mcpListWidget.ts
```diff
@@ -1303,91 +1135,454 @@ export class McpListWidget extends Disposable {
...
+	private showEmptySurface(message: string, detail: string): void {
+		this.cardContainer.style.display = 'none';
+		this.emptyContainer.style.display = 'flex';
+		this.emptyText.textContent = message;
+		this.emptySubtext.textContent = detail;
+	}
+
+	private addSurfaceActivation(surface: HTMLElement, label: string, callback: () => void): void {
+		surface.tabIndex = 0;
+		surface.setAttribute('role', 'button');
```
sbatten: 🤖 **[P1] The same nested-interactive-control problem exists for MCP cards.** Installed rows marked as `role="button"` later contain a native `role="switch"` and More button, while marketplace cards contain Install buttons. Please keep the container non-interactive and expose details through a dedicated control, or implement valid composite semantics.

src/vs/workbench/contrib/chat/browser/aiCustomization/embeddedAgentPluginDetail.ts
```diff
@@ -103,9 +259,406 @@ export class EmbeddedAgentPluginDetail extends Disposable {
...
+		}));
+		return container;
+	}
+
+	private async renderReadme(item: IAgentPluginItem): Promise<void> {
+		DOM.clearNode(this.readmeContentEl);
+		this.readmeEl.style.display = '';
+		let readme: string | undefined;
+		try {
+			readme = await this.fetchReadme(item);
```
sbatten: 🤖 **[P2] Same-plugin state changes can start overlapping README fetches and duplicate the rendered content.** The enablement autorun reruns `renderItem()` for the same item; each call clears once and starts this async fetch. When multiple requests resolve, they all pass the item-identity guard and append their Markdown. Please add cancellation or a render-generation guard and cover the same-item overlap in a test.

src/vs/workbench/contrib/chat/browser/aiCustomization/embeddedAgentPluginDetail.ts
```diff
@@ -103,9 +259,406 @@ export class EmbeddedAgentPluginDetail extends Disposable {
...
+			const content = await this.fileService.readFile(readmeUri);
+			return content.value.toString();
+		}
+		if (readmeUri.scheme === Schemas.https) {
+			let rawUrl = readmeUri.toString();
+			const githubBlobMatch = rawUrl.match(/^https:\/\/github\.com\/(?<owner>[^/]+)\/(?<repo>[^/]+)\/blob\/(?<rest>.+)$/);
+			if (githubBlobMatch?.groups) {
+				rawUrl = `https://raw.githubusercontent.com/${githubBlobMatch.groups['owner']}/${githubBlobMatch.groups['repo']}/${githubBlobMatch.groups['rest']}`;
+			}
+			const context = await this.requestService.request({ type: 'GET', url: rawUrl, callSite: 'aiCustomizationPluginDetail.fetchReadme' }, CancellationToken.None);
```
sbatten: 🤖 **[P2] Non-2xx response bodies are treated as valid README Markdown.** `IRequestService.request` resolves for HTTP errors, so a stale or missing README can display “404: Not Found” instead of the missing/error state. `pluginMarketplaceService` nearby already checks `context.res.statusCode === 200` before parsing. Please apply the same status validation here and test 404/500 responses.

src/vs/workbench/contrib/chat/browser/aiCustomization/aiCustomizationManagementEditorInput.ts
```diff
@@ -55,10 +54,10 @@ export class AICustomizationManagementEditorInput extends EditorInput implements
 	}
 
 	override getName(): string {
-		const harnessLabel = AICustomizationManagementEditorInput._activeHarnessLabel;
-		return harnessLabel
-			? localize('aiCustomizationManagementEditorNameWithHarness', "Agent Customizations for {0}", harnessLabel)
-			: localize('aiCustomizationManagementEditorName', "Agent Customizations");
+		if (this._targetLabel) {
+			return localize('aiCustomizationManagementEditorNameWithTarget', "Agent Customizations – {0}", this._targetLabel);
```
sbatten: 🤖 **[P1] This U+2013 en dash deterministically fails the repository hygiene check.** The current Compile & Hygiene job reports `Unexpected unicode character: "–" (charCode: 8211)` at this exact location. Please use ASCII punctuation or an allowed character.

src/vs/workbench/contrib/chat/browser/aiCustomization/media/aiCustomizationManagement.css
```diff
@@ -1452,175 +1798,423 @@ per-word capitalization does not survive translation. */
...
+	text-overflow: ellipsis;
+	white-space: nowrap;
 }
 
-.ai-customization-management-editor .models-content-container .section-footer-description {
-	font-size: var(--vscode-agents-fontSize-body1);
-	color: var(--vscode-descriptionForeground);
-	line-height: 1.5;
-	margin: 0 0 8px 0;
+.ai-customization-management-editor .ai-customization-embedded-detail .embedded-detail-fact-value:has(.embedded-detail-location-value) {
```
sbatten: 🤖 **[P3] This adds a `:has()` selector despite the repository's explicit CSS rule against descendant-state selectors.** The PR also adds several non-focus `!important` declarations to fight `.monaco-button` defaults. Please put a stable marker class on the owned fact-value node and resolve the button cascade through scoped specificity rather than `:has()`/`!important`.

sbatten (changes requested):
🤖 AI review verdict: request changes.

I found several functional and accessibility regressions in the redesigned card surfaces, including a recurrence of the MCP node-replacement bug fixed in #330890 and removal of the Settings Sync warning added in #331529. In addition to the inline findings, two required checks are red:

- **Compile & Hygiene** fails because the new editor title contains an unsupported U+2013 character (inline comment below).
- **Screenshots & Tests** passes Playwright, capture, and fixture-error checks, but fails because the newly gated blocks-CI hashes were not committed. Please review the visual diffs and update `test/componentFixtures/blocks-ci-screenshots.md` using the CI-generated hashes.

No high-confidence exploitable security issue was found, but the known MCP regression, keyboard/accessibility failures, migration data-availability regression, and required-check failures make this not ready to merge.

src/vs/workbench/contrib/chat/browser/aiCustomization/pluginListWidget.ts
```diff
@@ -845,6 +1095,472 @@ export class PluginListWidget extends Disposable {
...
+		this.installedCreateButton = undefined;
+		this.firstCardFocusElement = undefined;
+		DOM.clearNode(this.cardContainer);
+		this.showCardSurface();
+
+		const content = DOM.append(this.cardContainer, $('.plugin-card-scroll'));
+		const installedPlugins = this.installedItems;
+
+		this.renderDiscoverySnapshot(content);
+		if (this.marketplaceItems.length === 0 && this.isBrowseMarketplaceAvailable()) {
```
dmitrivMS: AI Review: This treats an empty marketplace result as "not fetched." Because `queryMarketplaceSnapshot()` rerenders after both an empty success and a failure, the next render starts another request, creating an unbounded request/render loop and bypassing the explicit Retry action. Track uninitialized, loading, loaded, and failed states separately, and only auto-fetch from the uninitialized state.

src/vs/workbench/contrib/chat/browser/aiCustomization/pluginListWidget.ts
```diff
@@ -483,39 +719,32 @@ export class PluginListWidget extends Disposable {
 			this.searchQuery = this.searchInput.value;
 			if (this.browseMode) {
 				this.delayedMarketplaceSearch.trigger(() => this.queryMarketplace());
+			} else if (this.searchQuery.trim()) {
+				this.delayedMarketplaceSearch.trigger(() => this.queryPluginSearch());
 			} else {
+				this.searchInput.hideMessage();
 				this.delayedFilter.trigger(() => this.filterPlugins());
```
dmitrivMS, line 788: AI Review: A non-browse search overwrites `marketplaceItems` with the filtered subset. Clearing the input only reruns `filterPlugins()`, while the home view refetches only when that array is empty, so a non-empty subset becomes the Featured/Available catalog and the other plugins stay hidden. Keep the unfiltered snapshot separate from search results, or restore/refetch it when the query becomes empty.

src/vs/workbench/contrib/chat/browser/aiCustomization/embeddedMcpServerDetail.ts
```diff
@@ -91,46 +119,50 @@ export class EmbeddedMcpServerDetail extends Disposable {
 		this.renderItem();
 	}
 
+	focus(): void {
+		this.definitionEditor.focus();
```
dmitrivMS: AI Review: Gallery-only MCP entries have no definition, so `setDefinition(undefined)` hides the editor before this call tries to focus it. Since the originating list is also hidden, keyboard users enter the detail view without visible focus. Focus the editor only when a definition exists; otherwise focus the Back button or a focusable empty state.

src/vs/workbench/contrib/chat/browser/aiCustomization/aiCustomizationListWidget.ts
```diff
@@ -1407,8 +1528,314 @@ export class AICustomizationListWidget extends Disposable {
...
+			more.element.classList.add('plugin-card-icon-button', 'customization-create-more-action');
+			more.label = `$(${Codicon.ellipsis.id})`;
+			this.cardDisposables.add(more.onDidClick(() => this.showCreateActionsMenu(secondaryActions, more.element)));
+		}
+	}
+
+	private formatTargetedCreateActionLabel(action: ICreateAction): string {
+		return action.label
+			.replace(/^\$\([^)]+\)\s*/, '')
+			.replace(/\s+\((?:Workspace|User)\)$/, '');
```
dmitrivMS: AI Review: `action.label` is already localized, but this strips only the literal English suffixes `(Workspace)` and `(User)`. Translated or reordered labels therefore retain redundant scope text in the scope-specific cards. Store a compact localized label on the action and render it directly instead of parsing localized display text.

src/vs/workbench/contrib/chat/browser/aiCustomization/embeddedAgentPluginDetail.ts
```diff
@@ -103,9 +259,406 @@ export class EmbeddedAgentPluginDetail extends Disposable {
...
+			const message = DOM.append(this.readmeContentEl, $('.plugin-detail-readme-message'));
+			message.textContent = localize('pluginReadmeMissing', "No README was provided for this plugin.");
+			return;
+		}
+		if (!readme.trim()) {
+			const message = DOM.append(this.readmeContentEl, $('.plugin-detail-readme-message'));
+			message.textContent = localize('pluginReadmeEmpty', "The plugin README is empty.");
+			return;
+		}
+		const rendered = this.renderDisposables.add(this.markdownRendererService.render(new MarkdownString(readme, { supportHtml: false })));
```
dmitrivMS: AI Review: `fetchReadme()` discards the source URI and constructs this `MarkdownString` without `baseUri`, so relative README links and media are removed or cannot resolve. Return the content together with the actual fetched URI, including any rewritten raw GitHub URI, and assign that URI to `MarkdownString.baseUri` before rendering.

src/vs/workbench/contrib/chat/browser/aiCustomization/aiCustomizationManagementEditorInput.ts
hawkticehurst: Valid and immediately actionable. The exact CI diagnostic and code point made this deterministic rather than speculative. I replaced the en dash with ASCII punctuation and updated the title test; the full hygiene task now passes. This is already strong bot feedback. The only small improvement would be to suggest the exact ASCII replacement so the fix can be applied without another style decision.

src/vs/workbench/contrib/chat/browser/aiCustomization/aiCustomizationListWidget.ts
hawkticehurst: Valid and high-value. The card redesign had preserved individual Tab reachability but dropped the composite-list behavior users previously received from WorkbenchList. I added a shared list/listitem controller for prompt-based, plugin, and MCP inventories with roving primary-row focus, Arrow/Home/End navigation, type-ahead, positional metadata, focused-row action traversal, and Shift+F10/context-menu access. The bot could improve this by naming the exact target pattern up front (a composite list with one roving primary tab stop plus actions for the active row) and by splitting the broad semantic regression from the specific excess-Tab-stops symptom.

src/vs/workbench/contrib/chat/browser/aiCustomization/embeddedAgentPluginDetail.ts
hawkticehurst: Valid and helpful. Same-item enablement changes could start multiple README reads, and item identity alone could not reject an older completion. I added a render-generation guard so only the latest render may append content, plus focused coverage for invalidating overlapping renders. The feedback could be slightly more prescriptive by noting that a generation guard is the broadly applicable fix here because both file and request-backed reads do not share one cancellable API.

src/vs/workbench/contrib/chat/browser/aiCustomization/media/aiCustomizationManagement.css
hawkticehurst: Partially current, but still helpful. The :has() portion had already been replaced by a stable has-actions marker on the latest head, while several non-focus !important overrides remained. I removed the remaining PR-added non-focus !important declarations and configured the owning Button instances/styles so the cascade no longer needs them. For better precision on a moving PR, the bot should separate :has() and !important into distinct findings and re-check each selector against the latest head before posting; combining them made half of the thread outdated while the other half remained valid.

src/vs/workbench/contrib/chat/browser/aiCustomization/customizationMigrationCategories.ts
hawkticehurst: Valid and especially helpful because it supplied the regression history (#331529 / #331211) and explained the user-visible data consequence, not just a copy difference. I restored destination-aware Settings Sync disclosure in the migration banner and updated the focused banner test. This feedback could be even stronger by quoting or linking the exact prior disclosure text/commit hunk and by explicitly separating the two acceptable remedies: warn before default deletion, or change the deletion default for unsynced destinations.

src/vs/workbench/contrib/chat/browser/aiCustomization/pluginListWidget.ts
hawkticehurst: Valid and helpful. Stopping propagation did not change the accessibility tree, so the outer plugin button still contained a switch and overflow/install controls. I replaced the outer interactive container with a non-interactive listitem and a dedicated native details button; switches and actions are siblings and only enter the tab order for the focused row. Since the MCP finding has the same root cause and fix pattern, the bot could reduce review noise by grouping both surfaces into one cross-file finding or explicitly marking the second as the same-pattern occurrence.

src/vs/workbench/contrib/chat/browser/aiCustomization/pluginListWidget.ts
hawkticehurst: Valid and very helpful. An empty successful result and an uninitialized snapshot were both represented as an empty array, so rendering recursively retriggered the fetch. I introduced explicit uninitialized/loading/loaded/failed snapshot state; an empty success is now terminal until the marketplace configuration changes or the user retries after failure, with regression coverage. This feedback was already well framed as a state-machine bug. It could be improved further by asking for separate empty-success and failure/retry tests so both terminal states are protected.

src/vs/workbench/contrib/chat/browser/aiCustomization/mcpListWidget.ts
hawkticehurst: Valid and very helpful. The review correctly connected the connection-state autorun to full card DOM replacement and, importantly, linked it to the prior #330890 regression. I addressed it by moving local runtime-status updates into per-row autoruns that update the existing badge, ARIA label, and description nodes in place; a regression test verifies node identity is preserved. To improve this feedback further, the bot could explicitly distinguish collection/enablement changes (which still require regrouping) from connection-state changes (which should only update row presentation), and point to the existing stable-row renderer as the reuse pattern.

src/vs/workbench/contrib/chat/browser/aiCustomization/mcpListWidget.ts
hawkticehurst: Valid and helpful. This was the same nested-interactive semantic bug as the plugin cards. MCP rows/cards now use non-interactive listitems with a dedicated native details button and sibling sign-in, switch, overflow, or install controls under the shared composite-list controller. For bot refinement, this could be deduplicated with the plugin finding and list all affected call sites in one thread; that would steer the author toward a shared primitive rather than two local patches.

src/vs/workbench/contrib/chat/browser/aiCustomization/toolsListWidget.ts
hawkticehurst: Valid and helpful. The event-path analysis was precise: Enter/Space from the nested More button reached the tree handler, which acted on the parent row and prevented the native button click. I addressed it by limiting tree keyboard handling to events whose target is the row itself, with a regression test covering descendant controls. A useful improvement would be to include the minimal expected keyboard test sequence (focus More -> press Enter/Space -> menu opens; tool-set enablement is unchanged), which would make the acceptance criteria immediately executable.

src/vs/workbench/contrib/chat/browser/aiCustomization/aiCustomizationListWidget.ts
hawkticehurst: Valid and helpful. Parsing localized display text for literal English scope suffixes was not localization-safe. Create actions now carry a separate localized compactLabel, and the targeted card renders that structural value instead of stripping translated text; a non-English regression case is covered. The recommendation was good. It could be improved by explicitly noting that icon stripping remains independent from scope labeling, so the data shape should carry both presentation concerns rather than parsing either from the final string.

src/vs/workbench/contrib/chat/browser/aiCustomization/pluginListWidget.ts
hawkticehurst: Valid and helpful. The search result subset and home discovery snapshot shared marketplaceItems, so clearing a non-empty search could permanently narrow the home catalog. I separated the persistent home snapshot from transient browse/search results; returning home now renders from the unfiltered snapshot. To improve the feedback, the bot could include the shortest repro sequence (load home -> search for a matching subset -> clear search -> observe missing cards) and recommend explicit snapshotResults/searchResults naming.

src/vs/workbench/contrib/chat/browser/aiCustomization/embeddedMcpServerDetail.ts
hawkticehurst: Valid and helpful. A gallery-only item has no definition, so focusing the hidden editor left the detail surface without a visible focus target. The detail now focuses its visible, focusable empty-definition state when no definition exists, while definitions still focus the editor. The feedback could be more decisive about focus priority: prefer the visible content state for continuity, with Back as a fallback only if there is no focusable detail content.

src/vs/workbench/contrib/chat/browser/aiCustomization/embeddedAgentPluginDetail.ts
hawkticehurst: Valid and helpful. The loader discarded the actual fetched URI, including the raw-GitHub rewrite, so relative links/media lacked a resolution base. It now returns content with the fetched URI and assigns MarkdownString.baseUri before rendering; the test verifies the rewritten raw URI. One useful refinement would be to state whether baseUri should be the README file URI or its containing directory. VS Code markdown resolution accepts the file URI and resolves relatives against dirname, which is what the fix now exercises.

src/vs/workbench/contrib/chat/browser/aiCustomization/embeddedAgentPluginDetail.ts
hawkticehurst: Valid and helpful. IRequestService resolves HTTP error responses, so parsing the body directly could render an error page as README content. I switched the loader to the existing asTextOrError helper and added 404/500 tests. The nearby status-check precedent was useful; an even better suggestion would name asTextOrError directly, since it is the repository helper that exactly encodes the desired behavior.

src/vs/workbench/contrib/chat/browser/aiCustomization/media/aiCustomizationManagement.css
```diff
@@ -814,23 +814,28 @@
 .ai-customization-management-editor .plugin-detail-container .ai-customization-embedded-detail .embedded-detail-leading-slot:not(:empty) {
 	display: flex;
 	align-items: center;
-	margin-inline-end: 4px;
+	margin-inline-start: -15px;
+	margin-inline-end: 20px;
```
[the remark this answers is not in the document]
hawkticehurst: This was valid for the reviewed revision. The off-ramp negative margin is no longer present on the current head, so no additional change was needed.

src/vs/workbench/contrib/chat/browser/aiCustomization/embeddedAgentPluginDetail.ts
```diff
@@ -103,9 +167,264 @@ export class EmbeddedAgentPluginDetail extends Disposable {
...
+		case PluginSourceKind.GitUrl:
+			return localize('pluginDetailSourceGitUrl', "Git URL");
+		case PluginSourceKind.Npm:
+			return localize('pluginDetailSourceNpm', "npm");
+		case PluginSourceKind.Pip:
+			return localize('pluginDetailSourcePip', "pip");
+		case PluginSourceKind.RelativePath:
+			return localize('pluginDetailSourceRelativePath', "Marketplace repository");
+	}
+}
```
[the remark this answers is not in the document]
hawkticehurst: This was valid for the reviewed revision, and the current implementation now has an explicit default fallback. No additional change was needed in this pass.

hawkticehurst:
Follow-up on the overall AI review verdict: the review was high-signal and the inline findings were substantively valid (with the `:has()` portion already outdated on the latest head). The deterministic Compile & Hygiene finding is fixed and the full hygiene task passes locally.

The blocks-CI screenshot warning was also valid for the reviewed commit. I reviewed the affected customization fixtures locally: 16 selected fixtures render without errors, and 10 dark/light/default/narrow fixtures are stable across three captures. I have not written screenshot hashes from macOS because `blocks-ci-screenshots.md` must use the PNG hashes produced by the Ubuntu CI job, and the follow-up changes alter gated fixtures again. The next pushed commit should use the CI-generated hash patch after the visual diff is reviewed.

For bot refinement, it would help to separate durable code findings from transient check-state findings and re-check the latter against the latest PR head before posting. That keeps an otherwise excellent review actionable across a fast-moving branch.

src/vs/sessions/AI_CUSTOMIZATIONS.md
```diff
@@ -223,6 +223,24 @@ For markdown-backed customizations (`.agent.md`, `SKILL.md`, `.instructions.md`,
...
 
+### Plugins Section
+
+The Plugins section uses a bolder card-based home surface while preserving `WorkbenchList` as the search and accessibility-safe fallback. The default **Plugins Home** shows installed plugins as cards, remote/session plugins as operational cards, and a progressive marketplace discovery snapshot below the installed content.
+
+Installed plugin cards show:
+- plugin identity, source/provenance, enabled/disabled state, and description
+- contribution preview chips derived from real installed plugin data (`agents`, `skills`, `commands`, `instructions`, `mcpServerDefinitions`, `hooks`)
+- explicit action buttons (`View`, `More Actions`) rather than making the card itself clickable
+
```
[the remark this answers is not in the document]
hawkticehurst: The mismatch was valid at the time. The architecture document has since been narrowed to stable ownership/contracts and intentionally no longer documents transient card interaction details, so the outdated UI-specific section is gone.

src/vs/workbench/contrib/chat/browser/aiCustomization/pluginListWidget.ts
```diff
@@ -944,36 +1494,96 @@ export class PluginListWidget extends Disposable {
...
+				.filter(p => {
+					const expectedUri = this.pluginInstallService.getPluginInstallUri(p);
+					return !installedUris.has(expectedUri.toString());
+				})
+				.map(marketplacePluginToItem);
+		} catch {
+			this.marketplaceItems = [];
+		}
+		await this.filterPlugins();
+	}
```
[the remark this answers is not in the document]
hawkticehurst: Valid for the reviewed revision. The current queryPluginSearch implementation already guards unsupported marketplace browsing and falls back to local filtering, so no additional change was needed.

src/vs/workbench/contrib/chat/browser/aiCustomization/pluginListWidget.ts
```diff
@@ -872,12 +1052,376 @@ export class PluginListWidget extends Disposable {
...
+		this.cardDisposables.add(DOM.addDisposableListener(card, 'keydown', e => {
+			if (e.target !== card) {
+				return;
+			}
+			if (e.key === 'Enter' || e.key === ' ') {
+				e.preventDefault();
+				callback();
+			}
+		}));
+	}
```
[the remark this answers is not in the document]
hawkticehurst: Valid. This overlaps the later plugin/MCP accessibility feedback. The outer cards/rows are now non-interactive listitems with dedicated native details buttons and sibling management controls under a shared composite-list controller.

src/vs/workbench/contrib/chat/browser/aiCustomization/mcpListWidget.ts
```diff
@@ -1169,106 +1155,36 @@ export class McpListWidget extends Disposable {
...
+			this.refresh();
 		}));
 		this._register(autorun(reader => {
 			this.customizationHarnessService.activeSessionResource.read(reader);
-			if (!this.browseMode) {
-				this.refresh();
-			}
+			this.refresh();
 		}));
 		this._register(this.agentHostCustomizationService.onDidChangeCustomizations(() => {
```
sbatten, line 1185: 🤖 **[P1] Session-backed MCP status changes still recreate the live card DOM.** `SessionMcpServerStateChanged`/`SessionCustomizationUpdated` can fire repeatedly without changing membership, but this listener always runs `refresh()` → `filterServers()` → `renderMcpHome()`/`updateSearchResults()`, which clears `cardDisposables` and the entire card container. The per-row update helper only observes local `connectionState`, so session Sign In/More/switch controls can still be replaced between mousedown and mouseup—the same class as #330890—and keyboard focus/scroll are dropped. Gate full rebuilds on membership changes and update session status presentation in place; the current helper-only test does not exercise this event path.

src/vs/workbench/contrib/chat/browser/aiCustomization/aiCustomizationListWidget.ts
```diff
@@ -825,8 +864,13 @@ export class AICustomizationListWidget extends Disposable {
 			}
 		}));
 
-		// Handle context menu
-		this._register(this.list.onContextMenu(e => this.onContextMenu(e)));
+		// Prompts retain their existing list context menu. The redesigned sections
```
sbatten, line 867: 🤖 **[P2] Native right-click item menus are now lost for every section this widget renders.** `usesCustomizationCardLayout()` is true for Agents, Skills, Instructions, Hooks, and Prompts, so this branch is never entered; the cards also live outside the `WorkbenchList` DOM and register no `contextmenu` listener. Keyboard ContextMenu/Shift+F10 and the overflow button still work, but mouse right-click no longer opens the standard item menu (including Copy Full Path). Add a card-row `contextmenu` listener that calls the existing `showCardItemActions`, or remove the dead list path if this removal is intentional.

src/vs/workbench/contrib/chat/browser/agentSessions/agentHost/agentHostCustomizationService.ts
```diff
@@ -212,6 +212,8 @@ export abstract class AbstractAgentHostCustomizationService extends Disposable i
 				disabledReason: getCustomizationDisabledReason(server, plugin),
 				status: server.state.kind,
 				state: server.state,
+				sourceUri: URI.parse(server.uri),
```
sbatten: 🤖 **[P2] This protocol URI is not necessarily client-readable.** File-backed MCP discovery publishes the Agent Host's raw `file:` URI, while top-level runtime servers use synthetic `mcp-top-level:` identifiers. The detail view later passes `sourceUri` to `fileService.readFile`, so remote paths resolve on the client and synthetic identifiers always fail. Map genuine host resources through the owning connection's Agent Host URI mapper, and omit the source for synthetic top-level IDs so the no-definition fallback is used.

src/vs/workbench/contrib/chat/browser/aiCustomization/embeddedAgentPluginDetail.ts
```diff
@@ -71,41 +219,422 @@ export class EmbeddedAgentPluginDetail extends Disposable {
 
 	setInput(item: IAgentPluginItem): void {
 		this.current = item;
-		this.renderItem();
+		if (item.kind === AgentPluginItemKind.Installed) {
+			this.inputStateAutorun.value = autorun(reader => {
```
sbatten, line 236: 🤖 **[P2] Enablement/version changes destroy the control that currently has focus.** Clicking Enable/Disable updates the observable read here, which synchronously calls `renderItem()`; that clears `renderDisposables` and the title-action DOM, detaching the focused split button. It also blanks and reloads the README on every toggle/version tick. Update the badge/action state in place (or explicitly restore focus and preserve the README) rather than rebuilding the complete detail view.

src/vs/workbench/contrib/chat/browser/aiCustomization/embeddedAgentPluginDetail.ts
```diff
@@ -71,41 +219,422 @@ export class EmbeddedAgentPluginDetail extends Disposable {
...
+						description: item.description,
+						version: item.version ?? '',
+						source: item.source,
+						sourceDescriptor: item.sourceDescriptor,
+						marketplace: item.marketplace,
+						marketplaceReference: item.marketplaceReference,
+						marketplaceType: item.marketplaceType,
+						readmeUri: item.readmeUri,
+					});
+					const installed = this.getInstalledPluginForMarketplaceItem(item);
```
sbatten, line 343: 🤖 **[P2] This async continuation can resume after the detail has been disposed.** In that case `setInput()` creates a new autorun, but the disposed `MutableDisposable` no longer owns/disposes it, so it remains live and repeatedly renders into detached DOM/disposed stores. `renderReadme()` has the same lifecycle gap after its awaited read: the generation/current checks still pass after disposal and the rendered Markdown is added to an already-disposed store. Add a disposal/cancellation guard after these awaits (or invalidate the render generation and clear `current` during disposal).

src/vs/workbench/contrib/chat/browser/aiCustomization/embeddedAgentPluginDetail.ts
```diff
@@ -4,53 +4,201 @@
...
+		return this.generation === generation;
+	}
+}
+
+export async function loadPluginReadme(
+	item: IAgentPluginItem,
+	fileService: Pick<IFileService, 'readFile'>,
+	requestService: Pick<IRequestService, 'request'>,
+): Promise<IPluginReadme | undefined> {
+	const readmeUri = item.kind === AgentPluginItemKind.Installed
```
sbatten, line 70: 🤖 **[P3] A missing local README is reported as a load failure instead of an expected absence.** Installed plugins always synthesize `README.md`, so `readFile()` throws `FILE_NOT_FOUND` and the caller displays “could not be loaded”; the intended `readme === undefined` / “No README was provided” path is unreachable for local plugins. Treat `FileOperationResult.FILE_NOT_FOUND` as `undefined` and reserve the error state for real read failures.

src/vs/workbench/contrib/chat/test/browser/aiCustomization/embeddedAgentPluginDetail.test.ts
```diff
@@ -0,0 +1,117 @@
...
+
+		assert.deepStrictEqual({
+			firstCurrent: guard.isCurrent(first),
+			secondCurrent: guard.isCurrent(second),
+		}, {
+			firstCurrent: false,
+			secondCurrent: true,
+		});
+
+		test('reads marketplace plugin versions', () => {
```
sbatten: 🤖 **[P2] This test is declared while another test is executing, so Mocha never runs it.** `reads marketplace plugin versions` needs to be a sibling under the suite, not nested inside `invalidates overlapping renders`. CI stays green because this assertion is never collected; move the closing brace so the version test is registered during suite definition.

src/vs/workbench/contrib/chat/browser/promptSyntax/hookActions.ts
```diff
@@ -307,6 +307,8 @@ export interface IHookQuickPickOptions {
 	readonly onHookFileCreated?: (uri: URI) => void;
 	/** Filter the displayed hook types to those supported by the given target. */
 	readonly target?: Target;
+	/** Restrict hook files and creation destinations to one storage scope. */
+	readonly preferredStorage?: PromptsStorage;
```
sbatten, line 311: 🤖 **[P3] This option scopes only the later create/file pickers, not initial hook discovery.** `parseAllHookFiles()` still loads every storage scope and `IParsedHook` drops storage, so the “New User Hook” flow can list/select a workspace hook. The management-editor callback then opens that URI labeled with the requested user storage rather than its actual scope. Filter before parsing or preserve storage on parsed entries and filter the initial counts/list; add coverage proving a user-scoped flow excludes workspace hooks.

roblourens (dismissed):

src/vs/workbench/contrib/chat/browser/aiCustomization/customizationCardList.ts
```diff
@@ -0,0 +1,199 @@
...
+			this._register(DOM.addDisposableListener(action, 'blur', () => {
+				DOM.getWindow(entry.row).queueMicrotask(() => {
+					if (!entry.row.contains(entry.row.ownerDocument.activeElement)) {
+						this.setActionsTabbable(entry, false);
+					}
+				});
+			}));
+		}
+
+		this.updateSetMetadata();
```
roblourens: **[Experimental performance review bot]**

Every newly added card calls updateSetMetadata, whose loop revisits all cards accumulated in that controller and sets aria-posinset and aria-setsize again.

Opening or filtering a large Available/customization group now performs quadratic DOM work before the inventory is usable; configured marketplaces can put all returned plugins into one controller, and customization groups can likewise contain many provider/extension items.

**Suggested fix:** Pass the final set size into the controller or add a finalize/setItems API: assign aria-posinset only to the appended row, then set aria-setsize for all rows once after population. Preserve the current roving-focus semantics without a full pass per append.

(Written by Copilot)

src/vs/workbench/contrib/chat/browser/aiCustomization/embeddedMcpServerDetail.ts
```diff
@@ -4,60 +4,116 @@
...
+		this.nameEl = DOM.append(headerText, $('.editor-item-name'));
+		this.pathEl = DOM.append(headerText, $('.editor-item-path'));
+
+		this.definitionEditorContainer = DOM.append(this.root, $('.embedded-editor-container.mcp-detail-definition-editor'));
+		this.definitionEmptyEl = DOM.append(this.root, $('.embedded-detail-empty.mcp-detail-definition-empty'));
+		this.definitionEmptyEl.tabIndex = -1;
+		this.definitionEmptyEl.textContent = localize('mcpDefinitionUnavailable', "No definition is available for this MCP server.");
+
+		this.definitionEditor = this._register(instantiationService.createInstance(
+			CodeEditorWidget,
```
roblourens: **[Experimental performance review bot]**

The detail constructor now synchronously instantiates a full CodeEditorWidget with automaticLayout. AICustomizationManagementEditor eagerly calls createEmbeddedMcpDetail while constructing all supported sections, before any MCP server detail is selected.

Opening Agent Customizations pays for and retains an extra Monaco editor even when the user only views Overview, Plugins, Skills, or another section; editor construction and automatic layout add measurable main-thread and memory overhead to this modal path.

**Suggested fix:** Keep the lightweight empty detail shell and lazily instantiate the CodeEditorWidget on the first setInput/showEmbeddedMcpDetail call; dispose or retain it only after it has actually been used.

(Written by Copilot)

src/vs/workbench/contrib/chat/browser/aiCustomization/mcpListWidget.ts
```diff
@@ -1169,106 +1155,36 @@ export class McpListWidget extends Disposable {
...
 	}
 
 	private async refresh(): Promise<void> {
-		if (this.browseMode) {
-			await this.queryGallery();
-		} else {
-			this.filterServers();
+		this.filterServers();
+		if (!this.searchQuery.trim() && this.gallerySnapshotServers.length === 0 && !this.gallerySnapshotFailed && !this.gallerySnapshotLoading) {
+			await this.queryGallerySnapshot();
```
roblourens, line 1200: **[Experimental performance review bot]**

refresh now always awaits queryGallerySnapshot when the snapshot is empty. Since AICustomizationManagementEditor eagerly constructs McpListWidget for supported harnesses, opening Overview or any other section starts the gallery request.

Users who open Customizations but never visit MCP Servers still incur a gallery request and first-page mapping/rendering; on slow or metered connections this wastes bandwidth and competes with foreground work.

**Suggested fix:** Gate queryGallerySnapshot on the MCP section becoming visible or an explicit marketplace action/search. Retain the current loading guard so the first visible request remains coalesced.

(Written by Copilot)

src/vs/workbench/contrib/chat/browser/aiCustomization/pluginListWidget.ts
```diff
@@ -845,6 +1154,498 @@ export class PluginListWidget extends Disposable {
...
+		this.firstCardFocusElement = undefined;
+		DOM.clearNode(this.cardContainer);
+		this.showCardSurface();
+
+		const content = DOM.append(this.cardContainer, $('.plugin-card-scroll'));
+		const installedPlugins = this.installedItems;
+
+		this.renderDiscoverySnapshot(content);
+		if (this.marketplaceSnapshot.state === 'uninitialized' && this.isBrowseMarketplaceAvailable()) {
+			void this.queryMarketplaceSnapshot();
```
roblourens, line 1221: **[Experimental performance review bot]**

renderPluginHome now starts queryMarketplaceSnapshot while the widget is initialized. The management editor eagerly constructs PluginListWidget for supported harnesses, so this runs from modal creation even while the plugin container is hidden.

Opening Agent Customizations on Overview, MCP, Skills, or another section now contacts all configured plugin marketplaces on a cold cache and reads cached repositories on warm paths, consuming bandwidth, disk/CPU, and competing with foreground work without user intent.

**Suggested fix:** Start the snapshot only when the Plugins section first becomes visible (or when the user requests browse/search), coalesce that first load in the existing snapshot model, and keep the cached result for later home renders.

(Written by Copilot)

roblourens (commented):
**[Experimental performance review bot]**

Human-approved experimental performance review.

(Written by Copilot)

sbatten (dismissed):
🤖 Third-pass review complete at current PR head `58b049a911f827443e0eaa195980f123e2f5b71a`.

No new findings.

This pass was deliberately limited to changes after the second reviewed head:
- `4c2a49ed`: fixes for the eight second-pass comments;
- `c4a03bb1`: CI/screenshot baseline fixes;
- `58b049a9`: subsequent merge from `main`, reviewing only its three combined conflict resolutions.

All eight second-pass fixes were verified, and no regressions introduced by those fixes were found. Compile/hygiene and the relevant platform/unit checks pass. The refreshed component-fixture run has no Playwright, render, or fixture errors; its current failure is only another blocks-CI screenshot hash mismatch after the latest merge.

roblourens (approved):

dmitrivMS (approved):

src/vs/workbench/contrib/chat/browser/aiCustomization/pluginListWidget.ts
```diff
@@ -912,36 +1742,114 @@ export class PluginListWidget extends Disposable {
...
+			const installedUris = new Set(this.agentPluginService.plugins.get().map(p => p.uri.toString()));
+			this.marketplaceItems = filtered
+				.filter(p => {
+					const expectedUri = this.pluginInstallService.getPluginInstallUri(p);
+					return !installedUris.has(expectedUri.toString());
+				})
+				.map(marketplacePluginToItem);
+			this.searchInput.hideMessage();
+		} catch {
+			this.marketplaceItems = [];
```
dmitrivMS, line 1787: AI Review: Each non-browse search cancels the previous token, but this catch ignores cancellation and request identity. If an older fetch rejects after a newer request succeeds, it clears the newer marketplace results, shows a stale warning, and refilters; clearing the query also leaves pending work active. Cancel delayed and in-flight work when the query clears, and only mutate results or messages when the CTS, query, and browse mode still match the initiating request.

src/vs/workbench/contrib/chat/browser/aiCustomization/embeddedAgentPluginDetail.ts
```diff
@@ -72,40 +231,489 @@ export class EmbeddedAgentPluginDetail extends Disposable {
...
+			...defaultButtonStyles,
+			secondary: true,
+			supportIcons: true,
+			contextMenuProvider,
+			addPrimaryActionToDropdown: false,
+			actions: {
+				getActions: () => {
+					const state = getPluginEnablementActionState(item.plugin.enablement.get());
+					return [
+						this.renderDisposables.add(new Action(`plugin.${state.isEnabled ? 'exclude' : 'include'}AlternateScope`, state.alternateLabel, undefined, true, async () => setEnablement(state.alternateState)))
```
dmitrivMS, line 420: AI Review: Every dropdown click creates a new caller-owned Action and retains it in the detail-wide renderDisposables until the detail rerenders or closes. ButtonWithDropdown disposes its menu view items, not these actions, so repeatedly opening the menu accumulates emitters and callbacks. Reuse one action per rendered button, return an unregistered lightweight action, or dispose actions when each menu closes.

src/vs/workbench/contrib/chat/browser/aiCustomization/mcpListWidget.ts
```diff
@@ -1303,94 +1242,525 @@ export class McpListWidget extends Disposable {
...
+		this.galleryCts?.dispose(true);
+		const cts = this.galleryCts = new CancellationTokenSource();
+		this.gallerySnapshotLoading = true;
 
-		this.searchInput.setPlaceHolder(browse
-			? localize('searchGalleryPlaceholder', "Search MCP marketplace...")
-			: localize('searchMcpPlaceholder', "Type to search...")
-		);
+		try {
+			const pager = await this.mcpWorkbenchService.queryGallery(undefined, cts.token);
```
dmitrivMS, line 1256: AI Review: This gallery request is still reachable when the effective chat.mcp.access value is none, including when policy-enforced. Selecting the MCP section can therefore issue a marketplace GET even though access-disabled UI is shown, and revocation leaves delayed or in-flight searches alive. Include effective access in snapshot/search eligibility and cancel delayedGallerySearch plus galleryCts when access becomes none; cover initial disabled visibility and revocation in tests.

src/vs/workbench/contrib/chat/browser/aiCustomization/embeddedAgentPluginDetail.ts
```diff
@@ -72,40 +231,489 @@ export class EmbeddedAgentPluginDetail extends Disposable {
...
+						sourceDescriptor: item.sourceDescriptor,
+						marketplace: item.marketplace,
+						marketplaceReference: item.marketplaceReference,
+						marketplaceType: item.marketplaceType,
+						readmeUri: item.readmeUri,
+					});
+					if (this._store.isDisposed || this.current !== item) {
+						return;
+					}
+					const installed = this.getInstalledPluginForMarketplaceItem(item);
```
dmitrivMS, line 343: AI Review: A successful install updates installed descriptors before MarketplaceAgentPluginDiscovery asynchronously publishes the plugin through IAgentPluginService.plugins, so this immediate lookup can miss. The detail does not observe the later publication and has no missing-result recovery, leaving the action disabled as Installing... until the view is recreated. Await or observe discovery of the expected URI, or return the installed plugin explicitly, and restore Install if registration never appears.

src/vs/workbench/contrib/chat/browser/aiCustomization/pluginListWidget.ts
```diff
@@ -845,6 +1159,505 @@ export class PluginListWidget extends Disposable {
...
+		switchElement.type = 'button';
+		switchElement.disabled = blocked;
+		switchElement.setAttribute('role', 'switch');
+		switchElement.setAttribute('aria-checked', String(checked));
+		switchElement.setAttribute('aria-label', blocked ? localize('pluginManagedByOrganizationAria', "{0} is managed by your organization", item.name) : toggleLabel);
+		switchElement.classList.toggle('checked', checked);
+		switchElement.title = blocked ? localize('pluginPolicyBlockedSwitch', "This plugin is managed by your organization.") : toggleLabel;
+		DOM.append(switchElement, $('.plugin-enable-switch-thumb'));
+		this.cardDisposables.add(DOM.addDisposableListener(switchElement, 'click', () => {
+			const nextState = getToggledPluginEnablementState(current);
```
dmitrivMS, line 1393: AI Review: This derives the transition from enablement captured when the row rendered, then the model update triggers an asynchronous full-card refresh. Before that refresh finishes, repeated activation writes the same transition while the live class and ARIA state remain stale; once it finishes, rebuilding cardContainer detaches the focused switch without restoring focus. Read current enablement per activation and update the row in place, or synchronously bind its state and restore the same row/action after rendering.

src/vs/workbench/contrib/chat/browser/aiCustomization/embeddedAgentPluginDetail.ts
```diff
@@ -72,40 +231,489 @@ export class EmbeddedAgentPluginDetail extends Disposable {
...
+				}),
+				secondary: true,
+				supportIcons: true,
+				ariaLabel: uninstallAction.label,
+			}));
+			uninstallButton.element.classList.add('embedded-detail-uninstall-button');
+			uninstallButton.label = uninstallAction.label;
+			uninstallButton.enabled = uninstallAction.enabled;
+			this.renderDisposables.add(uninstallButton.onDidClick(async () => {
+				await uninstallAction.run();
```
dmitrivMS, line 378: AI Review: The active remove contract is void, but CLI and extension implementations are async and await confirmation. UninstallPluginAction discards that runtime promise and resolves immediately, so this await navigates back while confirmation is pending, including when the user later cancels or removal fails. Return an awaited success result from removal, propagate it through the action, and navigate only after confirmed success.

src/vs/workbench/contrib/chat/browser/aiCustomization/mcpListWidget.ts
```diff
@@ -1303,94 +1242,525 @@ export class McpListWidget extends Disposable {
...
+		if (entry.type === 'server-item') {
+			return isContributionEnabled(this.mcpService.enablementModel.readEnabled(entry.server.id));
+		}
+		return true;
+	}
+
+	private setInstalledEntryEnabled(entry: IMcpInstalledEntry, enabled: boolean): void {
+		const activeSessionServer = getActiveSessionServer(entry);
+		if (activeSessionServer) {
+			activeSessionServer.setEnabled(enabled);
```
dmitrivMS, line 1715: AI Review: For an installed row with an active-session counterpart, this unqualified Enable/Disable switch writes only a Session decision and leaves profile/workspace enablement unchanged. A new session can therefore show the server enabled again even though the switch appeared to disable it generally. Label this control as session-scoped or have the primary switch update the durable/default scope.

src/vs/workbench/contrib/chat/browser/aiCustomization/aiCustomizationManagementEditor.ts
```diff
@@ -3286,6 +3351,46 @@ export class AICustomizationManagementEditor extends EditorPane {
...
+		await this.listWidget.setSection(section);
+		const modelSection = ITEMS_MODEL_SECTIONS.find(s => s === section);
+		if (modelSection) {
+			await this.itemsModel.whenSectionLoaded(modelSection);
+			const item = this.itemsModel.getItems(modelSection).get().find(item => isEqual(item.uri, uri));
+			if (item) {
+				const source = item.source;
+				const isWorkspaceFile = source === AICustomizationSources.local;
+				const isReadOnly = !source || source === AICustomizationSources.extension || source === AICustomizationSources.plugin || source === AICustomizationSources.builtin;
+				this.showEmbeddedEditor(item.uri, item.name, item.promptType, source ?? AICustomizationSources.builtin, isWorkspaceFile, isReadOnly);
```
dmitrivMS, line 3376: AI Review: The two preceding section-load awaits are not tied to the initiating navigation. If the user selects another section or view while either is pending, this stale continuation can still open the old plugin contribution and replace the newer destination. Track a navigation generation or cancellation token and recheck the requested section/view before calling showEmbeddedEditor.

PR #322053, opened 2026-06-19
Add toggle for maintaining total window size when toggling terminal panel or side bars

### LINKED ISSUE #178611: Window setting to control "Resize Type" for how the Panel gets hidden
## Why
I think it would be very helpful to have a setting that controls the logic around how the panel (terminal, output, etc) gets hidden and unhidden when toggling its visibility. I have my editor size set to the specific width I like it to use it at for line-length, and I find that it is extremely frustrating to toggle the visibility of this panel currently.

I am constantly having to manually drag my VSCode window to be wider and smaller depending on if I am showing the panel or not, because it keeps changing my editor width. To me, the editor is the main aspect of VSCode and so a setting to make it the "priority" size component makes sense.

## How
My idea is that this new setting can have 2 different choices, outlined below.

### Option 1: "Keep Window Size"
This option would work the way VSCode currently does, where toggling the panel expands and contracts the editor size within the window as it is toggled, maintaining the overall size of the VSCode window as it changes.
Preview:
keep window size

&nbsp;

### Option 2: "Keep Editor Size"
This option would maintain the editor size, and toggling the panel would contract or expand the overall size of the VSCode window to match its presence.
Preview:
keep editor size 2

&nbsp;

## Note
This should also be implemented to work identically for those that use the panel on the bottom of VSCode instead of the side, except vertically.

robmonte:
This adds a toggle (`Workbench: Keep Editor Size On Toggle`) that changes the behavior of the way the window resizes when hiding the terminal, left side bar, right side bar, or chat side bar. When enabled, hiding a panel instead collapses the window down to the editor size, instead of expanding the editor to fill in the space.

This is a gif recording of the feature when pressing cmd+j to toggle the terminal panel:
example keep size toggle

Testing involves toggling each type of sidebar and the terminal panel from all different possible locations configurable by VS Code. I'm unsure of any other test situations or if there are more robust test files to update.

Sometimes there is a single frame where you see the window size expand upon re-enabling a frame before the content within resizes back. I don't know if this is acceptable or not. I am not a very experienced developer in this type of project so please make suggestions or changes as necessary.

Please note that due to the above mentioned inexperience I may have been a little heavy-handed with applying Copilot's suggestions after reading through them. Assistance with nailing down a final version of these changes for review would be wonderful.

Fixes #178611


robmonte:
> @robmonte please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> 
> ```
> @microsoft-github-policy-service agree [company="{your company}"]
> ```
> 
> > Options:
> > 
> > * (default - no company specified) I have sole ownership of intellectual property rights to my Submissions and I am not making Submissions in the course of work for my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > * (when company given) I am making Submissions in the course of work for my employer (or my employer has intellectual property rights in my Submissions by contract or applicable law). I have permission from my employer to make Submissions and enter into this Agreement on behalf of my employer. By signing below, the defined term “You” includes me and my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree company="Microsoft"
> > ```
> 
> Contributor License Agreement

PR #322121, opened 2026-06-19
Don't dispose user-revealed terminals on session switch

### LINKED ISSUE #321049: Switching between Copilot Chat Sessions Destroys associated Terminal Sessions

meganrogge:
When switching sessions, the chat model reference is released, which triggers `_cleanupSessionTerminals` to dispose ALL tool terminals for that session. This incorrectly disposes terminals the user explicitly revealed.

Skip terminals that are in `foregroundInstances` (user-revealed) during cleanup, following the same pattern already used in background terminal completion cleanup.

Fixes #321049

### DIFF
--- a/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/runInTerminalTool.ts
+++ b/src/vs/workbench/contrib/terminalContrib/chatAgentTools/browser/tools/runInTerminalTool.ts
@@ -2776,6 +2776,15 @@ export class RunInTerminalTool extends Disposable implements IToolImpl {
 		this._sessionTerminalInstances.delete(chatSessionResource);
 
 		for (const terminal of terminalsToDispose) {
+			// Only dispose if the terminal is still hidden from the user. Once
+			// the user reveals it (via the terminal panel or the outputLocation
+			// setting), it joins foregroundInstances and should persist so they
+			// can inspect/interact with it. This prevents user-revealed
+			// terminals from being destroyed when switching between sessions.
+			if (this._terminalService.foregroundInstances.includes(terminal)) {
+				this._logService.debug(`RunInTerminalTool: Skipping disposal of user-revealed terminal ${terminal.instanceId} for session ${chatSessionResource}`);
+				continue;
+			}
 			// Skip redundant map walks in onDidDispose since this session has already been removed.
 			this._terminalsBeingDisposedBySessionCleanup.add(terminal);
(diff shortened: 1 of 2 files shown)


anthonykim1 (approved):

davemcom:
The `foregroundInstances` guard preserves revealed terminals but still disposes hidden ones on session switch — which risks encoding that disposal as expected behavior. Full analysis in [#321049 comment](https://github.com/microsoft/vscode/issues/321049#issuecomment-4753632169).

meganrogge:
@davemcom that is by design. We do not keep terminals around unless the user has opened them, see https://github.com/microsoft/vscode/issues/287177

davemcom:
@meganrogge seems like a significant change in user safety and expectations.

For users who are familiar with parallel execution across secure systems this is a non-trivial change by removing human from the observability chain.

With this intentional change, when a terminal is disposed, we lose the opportunity to understand *why* something happened — e.g. a folder structure disappearing because of a badly-quoted shell execution. The evidence can be gone before anyone can inspect it.

This is fundamentally incompatible with safe agent workflows, especially any requiring user authentication. Disposing the terminal disposes the entire built-up environment — auth tokens, SSH sessions, activated configs — without consent.

Further discussion in issue [321049](https://github.com/microsoft/vscode/issues/321049)

If this analysis is useful and you're looking for people who debug at this layer, I'd welcome a conversation. If there's a repro-bounty program, 10x my April–May combined AIC spend would be a welcome acknowledgment — along with the chance to join the team.

PR #322175, opened 2026-06-19
Fix premature disconnect force-fail of a live client tool call

### LINKED ISSUE #322442: Session appears hung on runTests tool
@justschen saw a session hung for a long time on a runTests tool call

roblourens:
Fixes https://github.com/microsoft/vscode/issues/322442

## Problem

A user's agent-host session appeared hung for ~9 hours on a `runTests` client tool call. Auto-approve was off, so a confirmation prompt was expected — but the user **never saw any confirmation UI**, and the session just sat there until they manually aborted.

The tool call was actually **force-failed almost immediately** (~18ms after it started) as if the owning client had disconnected — even though the client was connected the whole time and went on to run the tests successfully (199/199 passed ~2 min later). That premature failure destroyed the tool call before its real confirmation prompt could be shown, which is what made the session look stuck.

## How disconnect handling was supposed to work

When a client tool call is pending, `ProtocolServerHandler` protects against the owning client going away by arming a grace-period timeout (`_startClientToolCallDisconnectTimeout`). If the client doesn't come back within `CLIENT_TOOL_CALL_DISCONNECT_TIMEOUT` (30s), the call is force-failed via `_completeDisconnectedClientToolCalls` so it can't hang forever. Two paths arm this timeout:

1. `_handleClientDisconnected` — from a transport's `onClose`, when a connected client that owns a pending call disconnects.
2. `_checkOrphanedClientToolCalls` — on every `ChatToolCallStart`/`ChatToolCallReady` broadcast, to catch calls *issued for an already-gone client* (no `onClose` fires for those).

The timeout decides a client is gone from two signals:
- **Arm condition:** `IClientRecord.connection === undefined`.
- **Delay:** `max(0, TIMEOUT - (now - lastSeenAt))` — a "residual window" so a client that disconnected a while ago gets the *remaining* grace, not a fresh 30s. If it reconnects and re-subscribes in time, `_clearClientToolCallDisconnectTimeout` cancels the timer.

## What was missing

The whole scheme assumes `lastSeenAt` reflects *recent* liveness and that `connection === undefined` reliably means *gone*. Neither held:

1. **`lastSeenAt` was never refreshed from ordinary traffic.** It was only stamped at handshake, reconnect, and disconnect — so for a long-lived, chatty client it was stale by minutes or hours. When the orphan check armed the timeout, `now - lastSeenAt` was huge, so the delay collapsed to **0** and the timeout fired **immediately** instead of granting any grace.

2. **`connection` is a single pointer per `clientId`.** When a `clientId` is reused across overlapping transports (a window reconnects before its previous transport's `onClose` is observed), the pointer only tracks the most-recent transport. Closing that transport clears `connection` even though another transport for the same logical client is still live. So `connection === undefined` did **not** imply the client was gone.

Together these turned the "grace window" into an instant kill for exactly the live client it was meant to protect: the orphan check saw `connection === undefined`, armed with a zero delay derived from a stale `lastSeenAt`, and force-failed a perfectly healthy tool call.

## Fix

- Track liveness from **real traffic**: `_markClientSeen` bumps `lastSeenAt` on every inbound frame for an established client (not just at handshake/disconnect).
- **Re-verify at fire time**: when the timeout fires, `_isClientGraceExpired` checks whether *any* frame arrived within the grace window (recency-based, **not** the `connection` pointer) before force-failing. If the client is still alive, re-arm — but only while it still owns a pending tool call, so an active client never leaves a perpetual timer ticking.

This keeps the original guarantee (a genuinely-gone client's pending calls still fail after the grace window) while no longer killing a client that is demonstrably alive. Fixing this is necessary and sufficient to restore the confirmation UI in the incident.

## Tests

Adds two regression tests to `protocolServerHandler.test.ts`:
- A live client on a second transport keeps its tool call `Streaming` across multiple grace windows despite `connection === undefined` (this fails without the fix).
- A client that goes silent on every transport still has its tool call failed after the grace window (proves the disconnect path still works).

Existing ProtocolServerHandler suite (55 tests) unchanged and green; 57 passing total. `npm run typecheck-client` passes.

> A separate, secondary hang mechanism on the SDK permission-deferred path was also investigated. It is not included here: fixing the force-fail above is sufficient to prevent the incident, and this change is intentionally scoped to the root cause.

(Written by Copilot)

### DIFF
--- a/src/vs/platform/agentHost/node/protocolServerHandler.ts
+++ b/src/vs/platform/agentHost/node/protocolServerHandler.ts
@@ -183,24 +183,25 @@ interface IConnectedClient {
 
 /**
  * Per-client server-side record, keyed by clientId in
- * {@link ProtocolServerHandler._clients}. Unlike {@link IConnectedClient}, the
- * record OUTLIVES the connection: when a client disconnects, `connection` is
- * cleared to `undefined` but the record is retained (until pruned) so the
- * tool-call disconnect-grace machinery can still compute the remaining window
- * and hold any armed timeouts.
+ * {@link ProtocolServerHandler._clients}. Unlike {@link IConnectedClient},
+ * the record OUTLIVES individual transports: multiple overlapping transports
+ * for the same logical client are held oldest-first, with the active transport
+ * at the end. When the last transport disconnects, the record is retained
+ * (until pruned) so the tool-call disconnect-grace machinery can compute the
+ * remaining window and hold any armed timeouts.
  */
 interface IClientRecord {
-	/** Live connection while connected; `undefined` after disconnect (record retained for the grace window). */
-	connection: IConnectedClient | undefined;
 	/**
-	 * Epoch ms a live frame was last received from this client (handshake,
-	 * disconnect, or any subsequent inbound message). `undefined` when the
-	 * client has never been seen. Drives the disconnect-timeout grace window:
-	 * a pending client tool call fails `CLIENT_TOOL_CALL_DISCONNECT_TIMEOUT` ms
-	 * after the client last sent anything — so a client that is still actively
-	 * sending frames is treated as alive even if its {@link connection} pointer
-	 * was transiently cleared (e.g. a reused clientId whose earlier transport
-	 * closed while a newer one is still live).
+	 * All live transports for this client, oldest first. The active connection
+	 * is the last entry (most recent wins). Older entries are kept so that if a
+	 * reconnecting client registers `A`, then `B`, then `B` closes first, we can
+	 * fall back to `A` instead of treating the client as disconnected.
+	 */
+	readonly connections: IConnectedClient[];
+	/**
+	 * Epoch ms when the client last had no live transports. `undefined` while at
+	 * least one connection is active, or when the client has never connected.
+	 * Drives the disconnect-timeout grace window for disconnected records only.
 	 */
 	lastSeenAt: number | undefined;
 	/**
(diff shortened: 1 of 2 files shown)


src/vs/platform/agentHost/node/protocolServerHandler.ts
```diff
@@ -338,6 +341,14 @@ export class ProtocolServerHandler extends Disposable {
...
 
 		disposables.add(transport.onMessage(msg => {
+			// Any inbound frame from an established client is fresh proof of
+			// life. Keep the per-client last-seen timestamp current from real
+			// traffic so the disconnect-grace machinery can distinguish a
+			// genuinely gone client from one that is still active (see
+			// IClientRecord.lastSeenAt).
+			if (client) {
+				this._markClientSeen(client.clientId);
+			}
```
connor4312: > lastSeenAt was never refreshed from ordinary traffic. It was only stamped at handshake, reconnect, and disconnect — so for a long-lived, chatty client it was stale by minutes or hours

This was working as designed. `lastSeenAt` is not last activity. A client could be idle watching a session, not sending any messages, and still be quite able to respond to tool calls. `lastSeenAt` is meant to generally just be stamped as disconnection time so that tool call timeouts can have their interval set correctly.

`lastSeenAt` should _not_ be used while there is any active connection, i.e. there should be no one who observes a `lastSeenAt` time stamped by `transport.onMessage` -- it should only be observed in the disconnected state.

We could enhanced the types with a union to make this more explicit.

connor4312: On the other hand,

> connection is a single pointer per clientId. When a clientId is reused across overlapping transports (a window reconnects before its previous transport's onClose is observed),

This is a fair observation but the PR does not seem to address this. I actually handle this in the AHP filesystem, I'm not sure if there's logic worth reusing there...

https://github.com/microsoft/vscode/blob/9837699041685b0bff8109e2acf2b5bb311f2a11/src/vs/platform/agentHost/common/agentHostFileSystemProvider.ts#L141-L151

aeschli (approved):

PR #322212, opened 2026-06-20
Disable Bluetooth by default to prevent BlueZ connection popup on Linux startup

ayu-haker:
## Description

On Linux, VS Code requests Bluetooth access on startup because Electron initiates a connection to BlueZ (Linux Bluetooth daemon) early in the startup process. This causes a distracting system permission popup for users who don't even use Bluetooth.

## Fix

Add `--disable-bluetooth` Chromium switch early in the startup sequence to prevent Electron from connecting to BlueZ.

Users who need Web Bluetooth functionality can opt in by passing `--enable-bluetooth` CLI flag.

## Changes

- `src/main.ts`: Added `disable-bluetooth` switch before `argvConfig` return, with `--enable-bluetooth` opt-in

## Related

Fixes #134461

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -364,6 +364,13 @@ function configureCommandlineSwitchesSync(cliArgs: NativeParsedArgs) {
 	// use up to 2
 	app.commandLine.appendSwitch('max-active-webgl-contexts', '32');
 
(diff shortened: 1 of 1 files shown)


deepak1556:
Not the right fix, the flag isn't valid in the runtime.

PR #322226, opened 2026-06-20
Agent Host: fix Copilot real-SDK integration tests

roblourens:
## What

Fixes the Copilot **real-SDK** agent host integration tests (`copilotRealSdk.integrationTest.ts`, gated behind `AGENT_HOST_REAL_SDK=1`). The suite had drifted from the current agent host protocol and was failing 4 tests; it is now **12 passing, 1 pending, 0 failing**.

## Why

These tests fork a real agent-host server and drive it against the live Copilot SDK. Several protocol/behavior changes landed that the harness hadn't caught up with:

- **Multi-chat channel migration** — chat actions are now emitted on a session's *default chat channel* (`buildDefaultChatUri(sessionUri)`) rather than the bare session URI. The usage-action assertion, subagent channel filters/subscriptions, and the worktree subscription are updated to match.
- **Server crash in `driveTurn`** — `driveTurn` dispatched `chat/toolCallConfirmed` / `chat/inputCompleted` without a top-level `channel`, which crashed the server in `buildDefaultChatUri(undefined)`. The dispatches now include the channel (and drop the now-dead `session` field from the action bodies).
- **Plan-mode answer semantics** — `getAcceptedAnswers` now prefers the `exit_only` option for the exit-plan-mode question, so approving the plan ends the turn instead of continuing to implement in-turn (which surfaced tool-call confirmations the planning test asserts against).
- **Custom terminal tool is now opt-in** — `EnableCustomTerminalTool` defaults off, so the worktree test now enables it via a `root/configChanged` action before the session launches so the shell tool routes through the host terminal manager.

## Notes for reviewers

- **Test-harness changes only** — no product behavior changes. Two files touched, both under `test/node/protocol/`.
- The one **pending** test (`attaches a text blob and reads its function names`) is unrelated: it was committed already-skipped in #321356 (the inline-blob attachment counterpart to the passing file-attachment test).
- The Claude and Codex real-SDK suites are intentionally out of scope here.

(Written by Copilot)

### DIFF
--- a/src/vs/platform/agentHost/test/node/protocol/copilotRealSdk.integrationTest.ts
+++ b/src/vs/platform/agentHost/test/node/protocol/copilotRealSdk.integrationTest.ts
@@ -27,7 +27,7 @@ import { mkdtemp, rm, writeFile } from 'fs/promises';
 import { tmpdir } from 'os';
 import { join } from '../../../../../base/common/path.js';
 import { URI } from '../../../../../base/common/uri.js';
-import { MessageAttachmentKind, type MessageAttachment } from '../../../common/state/sessionState.js';
+import { MessageAttachmentKind, buildDefaultChatUri, type MessageAttachment } from '../../../common/state/sessionState.js';
 import type { ChatUsageAction } from '../../../common/state/sessionActions.js';
 import {
 	createRealSession, defineSharedRealSdkTests, dispatchTurn, driveTurnWithAttachmentsToCompletion,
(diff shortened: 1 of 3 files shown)


Juma-creator:
name: AgentHost-Copilot-Realsdk-CI-CD

on:
  push:
    branches: [ "main" ]
  pull_request:
    branches: [ "main" ]
  schedule:
    - cron: "0 2 * * *"   # Nightly run
  workflow_dispatch:

jobs:
  build-test-deploy:
    runs-on: ubuntu-latest

    steps:
    - name: Checkout Repo
      uses: actions/checkout@v3

    - name: Setup Node.js
      uses: actions/setup-node@v3
      with:
        node-version: '18'

    - name: Install Dependencies
      run: npm install

    - name: Lint Code
      run: npm run lint

    - name: Run Unit Tests
      run: npm test

    - name: Run Integration Tests
      run: npm run test:integration

    - name: Validate Test Fixes
      run: npm run test:fixes

    - name: Compliance Replay Validation
      run: node scripts/replay.js

    - name: Deploy to Staging
      run: npm run deploy:staging

bhavyaus (approved):

PR #322229, opened 2026-06-20
Rename chat.agentSessions.defaultConfiguration setting to chat.defaultConfiguration

justschen:
Renames the `chat.agentSessions.defaultConfiguration` setting to `chat.defaultConfiguration` to use a more final, surface-agnostic name (dropping the `agentSessions` segment).

This is in line with the discussion that the setting shouldn't be tied to a specific surface name (`agentSessions`/`agentHost`) and should use whatever the final stable name would be.

Only the setting name is changed — behavior is unchanged. Updated the enum value, registration key, localization keys, doc comments, and test references.


roblourens (dismissed):

justschen:
Addressed the review feedback:

- **Migration / backward compat:** Added a `registerConfigurationMigrations` entry that copies any existing `chat.agentSessions.defaultConfiguration` value over to the new `chat.defaultConfiguration` key and clears the old one, so users/org policies with the old setting keep working.
- **Enum/interface naming:** Renamed the now-misleading `AgentSessionDefaultConfiguration` enum member to the surface-agnostic `DefaultConfiguration`, and `IAgentSessionDefaultConfiguration` → `IChatDefaultConfiguration`, updating all call sites.

bhavyaus (approved):

PR #322333, opened 2026-06-22
feat(chat): implement AgentHostSessionListStore for managing agent-host sessions

DonJayamanne:
- Introduced AgentHostSessionListStore to manage and cache agent-host sessions.
- Added interfaces for session list connection, entries, and notifications.
- Implemented methods for session management including refresh, add, update, and remove.
- Integrated workspace context filtering for session visibility.
- Updated tests to cover new session list store functionality and ensure correct behavior during session notifications and refresh operations.

Addresses comment https://github.com/microsoft/vscode/pull/322030/changes#r3440361045


src/vs/workbench/contrib/chat/browser/agentSessions/agentHost/agentHostSessionListController.ts
```diff
@@ -33,58 +32,22 @@ function mapSessionStatus(status: SessionStatus | undefined): ChatSessionStatus
...
 
 	private readonly _onDidChangeChatSessionItems = this._register(new Emitter<IChatSessionItemsDelta>());
 	readonly onDidChangeChatSessionItems = this._onDidChangeChatSessionItems.event;
 
-	private _items: IChatSessionItem[] = [];
-	/** Cached full summaries per session so partial updates can be applied. */
-	private readonly _cachedSummaries = new Map<string, SessionSummary>();
+	private _itemsByRawId = new Map<string, IChatSessionItem>();
 	/** Final-looking resources created locally before the backend session exists. */
 	private readonly _pendingNewSessions = new Set<string>();
```
connor4312: Could these be stored on the base `AgentHostSessionListStore` and then just filtered when called via getters? That would make it so AgentHostSessionListController doesn't have to store any internal state at all, becoming a lot simpler.

Likewise _onDidChangeChatSessionItems could be an `Event.filter(Event.map(...` over the base AgentHostSessionListStore event

connor4312 (approved):

PR #322415, opened 2026-06-22
Extract reusable loopback proxy server base

vritant24:
Extracts the loopback HTTP-server lifecycle that was duplicated across the agent-host proxies into a reusable base class, and refactors ClaudeProxyService and CodexProxyService onto it.

This sets up the upcoming BYOK implemntation that also requires a proxy erver to share the same base


src/vs/platform/agentHost/node/claude/claudeProxyService.ts
```diff
@@ -178,66 +150,46 @@ function readCopilotUsageNanoAiu(event: unknown): number | undefined {
...
+	) {
+		super(PROXY_USER_FACING_NAME, logService);
+	}
 
-	async start(githubToken: string): Promise<IClaudeProxyHandle> {
-		if (this._disposed) {
-			throw new Error('ClaudeProxyService has been disposed');
-		}
+	protected createState(): IClaudeProxyState {
+		return { githubToken: '' };
```
TylerLeonhardt: empty string feels bad here

src/vs/platform/agentHost/test/node/claudeProxyService.test.ts
```diff
@@ -349,100 +349,14 @@ suite('ClaudeProxyService', () => {
 
 	ensureNoDisposablesAreLeakedInTestSuite();
 
-	// #region Lifecycle
+	// #region Token slot
```
TylerLeonhardt, line 352: are there codex tests?

how much of these tests are redundant with loopbackproxyserver.test.ts

vritant24: There are no codex tests that I could find. The removed redundancy with the Claude tests and moved them to the loopback server tests

src/vs/platform/agentHost/node/claude/claudeProxyService.ts
vritant24: Good call. I refactored it so that the loopback server's state can be seeded, removing the need for defaults

src/vs/platform/agentHost/test/node/claudeProxyService.test.ts
TylerLeonhardt: FYI @Giuspepe , there's a test gap here for Codex

TylerLeonhardt (approved):

PR #322445, opened 2026-06-22
Add `agents.voice.microphoneDevice` setting for mic selection

### LINKED ISSUE #322435: Allow selecting mic for voice mode
Suggestions

- Detect mics, show the list in a machine-specific setting
- Default to the built-in macbook mic over airpods, or warn about the airpods mic- most people will not want that

meganrogge:
Adds a new `agents.voice.microphoneDevice` setting to specify which microphone device to use for voice mode. When set, passes the deviceId as an exact constraint to getUserMedia. Empty string (default) uses the system default.

Fixes #322435

Screenshot 2026-06-22 at 5 17 48 PM

### DIFF
--- a/src/vs/platform/agentHost/node/copilot/mapSessionEvents.ts
+++ b/src/vs/platform/agentHost/node/copilot/mapSessionEvents.ts
@@ -525,14 +525,14 @@ function sdkAttachmentToProtocol(
 				return {
 					type: MessageAttachmentKind.Simple,
 					label: attachment.displayName ?? 'attachment',
-					modelRepresentation: decodeBase64(attachment.data).toString(),
+					modelRepresentation: decodeBase64(attachment.data ?? '').toString(),
 				};
 			}
 			const displayKind = attachment.mimeType.startsWith('image/') ? 'image' : undefined;
 			return {
 				type: MessageAttachmentKind.EmbeddedResource,
 				label: attachment.displayName ?? 'attachment',
-				data: attachment.data,
+				data: attachment.data ?? '',
 				contentType: attachment.mimeType,
 				displayKind,
 			};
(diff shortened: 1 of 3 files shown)


eleanorjboyd (dismissed):

src/vs/workbench/contrib/agentsVoice/browser/agentsVoice.contribution.ts
```diff
@@ -439,5 +499,12 @@ configurationRegistry.registerConfiguration({
 			included: false,
 			tags: ['advanced'],
 		},
+		'agents.voice.microphoneDevice': {
```
connor4312, line 502: Wonder if it's really worth having this as a setting since users can't configure it by hand very easily -- versus having a value in storage

src/vs/platform/agentHost/node/copilot/mapSessionEvents.ts
```diff
@@ -525,14 +525,14 @@ function sdkAttachmentToProtocol(
...
-					modelRepresentation: decodeBase64(attachment.data).toString(),
+					modelRepresentation: decodeBase64(attachment.data ?? '').toString(),
 				};
 			}
 			const displayKind = attachment.mimeType.startsWith('image/') ? 'image' : undefined;
 			return {
 				type: MessageAttachmentKind.EmbeddedResource,
 				label: attachment.displayName ?? 'attachment',
-				data: attachment.data,
+				data: attachment.data ?? '',
```
connor4312, line 535: unrelated? Per types this should never be undefined

roblourens (approved):

src/vs/workbench/contrib/agentsVoice/browser/agentsVoice.contribution.ts
meganrogge: Yes, for discoverability, I think so. We will have a gear icon somewhere within voice chat and that will send people to the settings UI. I will add to this description the command to run via markdown.

PR #322470, opened 2026-06-22
Fix context size handling for Agent Host

### LINKED ISSUE #321707: AHP: Context window mismatch between the context size picker and the context usage widget
Image

pwang347:
Fixes https://github.com/microsoft/vscode/issues/321707

Notes:
- Agent host context in context widget now correctly matches input + output tokens (we were setting output to always 0)
- Switching models will not also update the context widget, even if it wasn't actually used for the previous request

Dependent on AHP changes:
- https://github.com/microsoft/agent-host-protocol/pull/257
- https://github.com/microsoft/agent-host-protocol/pull/258
- https://github.com/microsoft/agent-host-protocol/pull/260


src/vs/platform/agentHost/common/state/protocol/channels-root/state.ts
```diff
@@ -97,6 +97,13 @@ export interface SessionModelInfo {
 	name: string;
 	/** Maximum context window size */
 	maxContextWindow?: number;
+	/**
+	 * Context-window sizes (in tokens) this model recommends as user-selectable options, smallest
+	 * first. The first entry is the recommended default. Present only when the model offers meaningful
+	 * context-window options (e.g., long context pricing). Clients can optionally render these as a
+	 * context-size picker and send the chosen value back via {@link ModelSelection.maxContextWindow}.
+	 */
+	recommendedContextWindows?: readonly number[];
```
roblourens: Why an array, vs one default option?

pwang347: mostly for future-proofing if we find more useful breakpoints than just long context or not, Copilot SDK itself takes an enum

src/vs/platform/agentHost/node/copilot/copilotSessionLauncher.ts
```diff
@@ -26,6 +26,16 @@ import { agentHostPromptRegistry, type IAgentHostPromptContext } from './prompts
...
+ * Config key for the numeric "Context Size" selection (a context-window token count). Mapped to the
+ * SDK's two-valued {@link SessionConfig.contextTier} by {@link getCopilotContextTier}.
+ */
+export const ContextSizeConfigKey = 'contextSize';
+/**
+ * @deprecated Legacy config key that stored the resolved tier string (`'default'` / `'long_context'`)
+ * directly. Replaced by the numeric {@link ContextSizeConfigKey}; still read from persisted sessions
+ * for backward compatibility.
+ */
 export const ContextTierConfigKey = 'contextTier';
```
pwang347, line 40: @roblourens thoughts on whether we need to keep this for persisted sessions?

src/vs/platform/agentHost/common/state/protocol/.ahp-version
```diff
@@ -1 +1 @@
-77c6312
+151dbad
```
pwang347: TODO

sandy081 (dismissed):

src/vs/workbench/contrib/chat/browser/widgetHosts/viewPane/chatContextUsageWidget.ts
```diff
@@ -103,6 +103,13 @@ export class ChatContextUsageWidget extends Disposable {
 	private readonly _modelConfigurationListener = this._register(new MutableDisposable());
 	private _currentResponse: IChatResponseModel | undefined;
 	private _currentModelId: string | undefined;
+	/**
+	 * The model the user currently has selected in the picker. When set it
+	 * overrides the last request's model for computing the context-window
+	 * denominator, so switching models updates the widget before the next
+	 * request is sent. The usage numerator still comes from the last response.
+	 */
+	private _selectedModelId: string | undefined;
```
roblourens, line 112: What's the difference between selectedModelId and currentModelId?

src/vs/platform/agentHost/node/copilot/copilotSessionLauncher.ts
roblourens: It's not clear to me, what data is being persisted that uses this key?

roblourens (commented):
Looks fine in general

src/vs/workbench/contrib/chat/browser/widgetHosts/viewPane/chatContextUsageWidget.ts
pwang347: `currentModelId` is the actual model that was last executed, I think we still want to track it at some level instead of just writing over it. Wanted to sync with @lramos15 before actually fully removing.

It should be possible to preview changes to the context when switching models though

src/vs/platform/agentHost/node/copilot/copilotSessionLauncher.ts
pwang347: I believe the model config is persisted indirectly as part of `src/vs/platform/agentHost/node/sessionDatabase.ts`

So old chat sessions would try to restore with this key

src/vs/workbench/contrib/chat/browser/widgetHosts/viewPane/chatContextUsageWidget.ts
pwang347: FWIW I chatted with Logan and we agreed on the new mental model (the widget should show the newly selected model as a preview, instead of the historic content)

roblourens (approved):

roblourens (dismissed):

lramos15 (approved):

PR #322484, opened 2026-06-23
Fix color picker copy button copying accessibility placeholder (#322335)

### LINKED ISSUE #322335: Color picker "copy" button copies "There is a color picker here."

- VS Code Version: 1.125.1 (Universal)
- OS Version: MacOS Tahoe 26.3.1 (a) 

Steps to Reproduce:

1. Open a color picker by hovering over a color swatch in a css file
2. Click the copy button
3. Your clipboard now contains "There is a color picker here."

Image

zaewc:
Fixes #322335

## The bug

The hover copy button copies the text returned by each hover participant's
`getAccessibleContent`. For the color picker participant this returned the
screen-reader placeholder "There is a color picker here." instead of the actual
color value — so clicking the copy button put that placeholder on the clipboard.

## The fix

Return the picked color presentation label, falling back to the CSS-formatted
current color, so the copy button copies a meaningful color value and screen
readers announce the color instead of a generic placeholder.

```ts
public getAccessibleContent(hoverPart: ColorHover): string {
    return hoverPart.model.presentation?.label ?? Color.Format.CSS.format(hoverPart.model.color);
}
```
## Tests

Added hoverColorPickerParticipant.test.ts covering getAccessibleContent:

- returns the picked presentation label when one is available
- falls back to the CSS-formatted color (not the placeholder) when there is no presentation — the regression covered by this issue


zaewc:

PR #322497, opened 2026-06-23
Update customizations sidebar content and behavior

### LINKED ISSUE #321166: Agents window customizations dialog does not select clicked section
When clicking a customization category in the Agents window, such as **Plugins**, the Agent Customizations dialog opens but does not select the clicked category.

Steps to reproduce:
1. Open the Agents window.
2. In the Customizations section, click a category such as **Plugins**.

Expected: The dialog opens with **Plugins** selected.
Actual: The dialog opens without selecting **Plugins**.

hawkticehurst:
Fixes https://github.com/microsoft/vscode/issues/321166

Addresses bug where the Customizations sidebar list items, when clicked, do not open their associated Customizations modal page.

This is accomplished by rethinking Customizations sidebar a bit to match the behavior of the Checks section UI that is displayed in chat sessions that have PRs with CI tasks.

Specific changes:

- Customizations section is collapsible and resizable
  - Min height visually matches Checks section min height
  - Initial height and max height are the total height of content
- Added an "Overview" list item to access the Overview page directly
- Change Customizations label font size to match Checks (i.e. make it smaller by 1px)
- Remove `sessions.customizations.sidebarMode` setting and associated code for "single" option

https://github.com/user-attachments/assets/3c68c126-0c85-4e38-a96c-1978585587eb

https://github.com/user-attachments/assets/b2c0f4b5-7c3e-4413-a397-bd13ed14bcb4

### DIFF
--- a/src/vs/sessions/contrib/sessions/browser/aiCustomizationShortcutsWidget.ts
+++ b/src/vs/sessions/contrib/sessions/browser/aiCustomizationShortcutsWidget.ts
@@ -7,28 +7,25 @@ import '../../../browser/media/sidebarActionButton.css';
 import './media/customizationsToolbar.css';
 import * as DOM from '../../../../base/browser/dom.js';
 import { Codicon } from '../../../../base/common/codicons.js';
-import { Disposable, DisposableStore, MutableDisposable } from '../../../../base/common/lifecycle.js';
+import { Emitter } from '../../../../base/common/event.js';
+import { Disposable, DisposableStore } from '../../../../base/common/lifecycle.js';
 import { autorun, derived } from '../../../../base/common/observable.js';
-import { ThemeIcon } from '../../../../base/common/themables.js';
+import { ScrollbarVisibility } from '../../../../base/common/scrollable.js';
 import { localize } from '../../../../nls.js';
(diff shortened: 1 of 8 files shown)


src/vs/sessions/contrib/sessions/browser/aiCustomizationShortcutsWidget.ts
```diff
@@ -166,101 +174,71 @@ export class AICustomizationShortcutsWidget extends Disposable {
...
+			return 0;
+		}
 
-			// Re-layout after the transition
-			transitionListener.value = DOM.addDisposableListener(toolbarContainer, 'transitionend', () => {
-				transitionListener.clear();
-				options?.onDidChangeLayout?.();
-			});
-		};
+		const rootStyles = DOM.getWindow(root).getComputedStyle(root);
```
benibenj: Using `getComputedStyle` is rather bad practice. If you need some styles both in css and typescript I would rather suggest hardcoding it in typescript and setting the styling values on the DOM element. This should also help reduce the `Number.isFinite` checks

benibenj (dismissed):

connor4312 (approved):

PR #322596, opened 2026-06-23
chore: add optional AI assistance note to PR template

Dodothereal:
Adds an opt-in (not enforced) line in the PR template suggesting that contributors mention whether AI-assisted tooling was used. The PR's human author remains the author regardless.


dmitrivMS:
I don't think this is necessary.

PR #322615, opened 2026-06-23
Fix voice transcript overlay persisting across session switches

meganrogge:
When navigating between chat sessions while voice mode is active, the transcript overlay from the previous session was incorrectly shown in the new session's chat input.

## Root cause
The transcript overlay autorun read `this._widget?.viewModel?.sessionResource` as a plain property access, so it never re-triggered when the user switched sessions via `onDidChangeViewModel`.

## Fix
- Track the current session resource as an `observableValue` (`_currentSessionResource`)
- Update it in the `onDidChangeViewModel` callback
- Read it in the transcript autorun so session switches trigger re-evaluation
- Hide the transcript when the voice controller's target session doesn't match the currently displayed session

Also removes unused `IAgentsVoiceWindowService` import/injection from chatViewPane.

Fixes microsoft/vscode-internalbacklog#8165


meganrogge:

eleanorjboyd (approved):

PR #322648, opened 2026-06-24
Use max output context for output buffer

pwang347:
image


lramos15:
Trying to think this through. If the model is a 64k context model and the max output tokens is 32k then if we have a prompt of 40k the buffer should only be rendered as 24k not 32k as otherwise the buffer would exceed max context. If that makes sense?

pwang347:
> Trying to think this through. If the model is a 64k context model and the max output tokens is 32k then if we have a prompt of 40k the buffer should only be rendered as 24k not 32k as otherwise the buffer would exceed max context. If that makes sense?

To my knowledge the clamping already happens in https://github.com/microsoft/vscode/blob/d3e7ccc25f56f9a684d449726a5e101a7a0ef65a/src/vs/workbench/contrib/chat/browser/widgetHosts/viewPane/chatContextUsageDetails.ts#L183

Let me see if I can reproduce it though!

lramos15 (dismissed):

pwang347:
image
Looks like this

TylerLeonhardt (approved):

PR #322669, opened 2026-06-24
Add model managed setting and make it default for new conversations

vijayupadya:
Fixes https://github.com/microsoft/vscode-internalbacklog/issues/7934

Adds support for configuring a default model that applies to every new conversation, enforable by enterprise policy.

**What's new**

- New model property on the chat.defaultModel setting. Accepts:

  - "auto" — let Copilot pick the model
  - a model family name (e.g. "opus", "gemini") — resolves to the latest available version in that family
  - a full model id

- A new ChatDefaultModel enterprise policy that maps the model managed setting onto this configuration.

**Behavior**

- New conversations start at the configured model across the local chat panel and the Agents window (including Copilot CLI sessions).
- Users can still switch models within a conversation; an explicit pick is never overridden by the configured default.
- Reopened conversations keep their own saved model.
- When the setting is unset, behavior is unchanged.
- Setting applied to both local agent and AH


vijayupadya:
cc @joshspicer

pwang347 (approved):

joshspicer:
Tracking under the **Unified Settings** workstream → microsoft/vscode-internalbacklog#7940 (the `model` managed setting, microsoft/vscode-internalbacklog#7934).

PR #322674, opened 2026-06-24
agentHost: speed up subagent session enrichment (parallel load + memoized reconstruction)

DonJayamanne:
## Why

Opening an agent session that used **subagents** (a parent agent that spawned helper "child" agents) was slow. To render the transcript, VS Code rebuilds the full conversation and groups each child's inner steps under its parent tool call. That rebuild was slow because:

- the client loaded each subagent **one at a time** (sequentially), and
- the backend rebuilt the **entire parent conversation from scratch once per subagent** — a session with 10 subagents re-parsed the whole event log 11 times.

## What changed

- **A — Parallel subagent loading (client).** `agentHostSessionHandler` now collects all subagent insertions across the whole history, dedupes child-state loads via a `Map`, and loads them in parallel (`Promise.all`) instead of sequentially.
- **B — Reconstruct once / memoization (backend).** `CopilotAgentSession` memoizes the `getEvents()` + `mapSessionEvents()` result (`_getMappedEvents`), shared by `getMessages`, `getSubagentMessages`, and the new `getSubagentSessions`. The memo is invalidated whenever the persisted event log could change (user message, turn start/end, tool complete, subagent completed/failed, compaction, truncation, snapshot rewind) so it never serves stale turns for an active session.
- **C — Eager registration up-front (backend).** New optional `IAgent.getSubagentSessions()` + `IRestoredSubagentSession`. During `restoreSession`, the dispatcher registers every discovered child session eagerly (reusing B's single reconstruction — no extra event-log reads), so the client's later per-child lookups are instant in-memory hits.
- **D — Correctness fix (client).** `_loadSubagentState` now waits for **both** the child's session summary **and** its default-chat channel to hydrate before reading. After the multi-chat protocol split, `turns` live on the chat channel; awaiting only the session subscription read too early and silently produced **zero** inner steps.

## Results

For a real 90-turn / 4-subagent session:
- enrichment: **~2161 ms → ~53 ms**
- total click-to-displayed: **~5324 ms → ~2700 ms**

## Notes for reviewers

- The memo is intentionally scoped to the resume/restore wave and dropped on any log-mutating event and on dispose; rejected reconstructions are not cached so the next caller retries.
- `getSubagentSessions` is optional on `IAgent`; agents without subagents simply omit it.
- Tests added: eager registration during parent restore (`agentService.test.ts`) and memoization + invalidation (`copilotAgentSession.test.ts`).

For https://github.com/microsoft/vscode/issues/320303


roblourens (dismissed):
I had also seen this as a source of slowness before but didn't end up with a change that I was confident in. I think this is ok. I am confused about subagents in the new world of chats/sessions and need to look at that some more

dmitrivMS (approved):

PR #322690, opened 2026-06-24
Enable second reply on chat session smoke test

alexr00:


alexr00 (commented):
Ran successfully 19 times (20th time was an unrelated failure).

hediet (approved):

PR #322701, opened 2026-06-24
fix: correct copy-pasted JSDoc on provideInlineValues method

jachinsamuel:
### Description
This PR corrects the copy-pasted JSDoc comment above the `provideInlineValues` method in the `InlineValuesProvider` interface. The comment was copy-pasted from `HoverProvider` and incorrectly referenced "hovers being merged", which is not applicable to inline debug values.

### How to test
Verify that the updated JSDoc in `src/vs/editor/common/languages.ts` correctly describes inline values (mirroring the public API description in `vscode.d.ts`).


dmitrivMS:
Does not meet the bar.

PR #322733, opened 2026-06-24
Allow chat.agentHost.codexAgent.enabled to be overridden by experiment

Giuspepe:
Adds `experiment: { mode: 'startup' }` to the `chat.agentHost.codexAgent.enabled` setting so its default can be overridden by an ExP.


javi0guerrero4-cloud (commented):
Ok

anthonykim1 (approved):

Issue #322730, opened 2026-06-24
Stale TypeScript markers shown


dbaeumer:
Steps to Reproduce:

1. clone https://github.com/dbaeumer/markerProblem.git
2. npm instaall
3. Ensure you have the TypeScript native preview extension installed
4. Enable the TypeScript native preview extension

<img width="383" height="44" alt="Image" src="https://github.com/user-attachments/assets/bbe4bbe1-2e8e-43f9-9b7b-3cf78b333e46" />
 
6. Run the npm: watch marker task. Observe: you get one error in the problems view
7. Switch to the problems view
8. Open greet.ts: Observe: you get two errors in the problems view
<img width="542" height="571" alt="Image" src="https://github.com/user-attachments/assets/a65eef50-df5d-43b1-ab2f-8e916bceeaa1" />

9. Fix the code by putting export in front of the `foo` function. Observe: the error is still shown in the problems view and the code still shows the error as well
<img width="696" height="577" alt="Image" src="https://github.com/user-attachments/assets/fd139aae-0e66-4a35-8c6c-aac902bad022" />

The watch task doesn't report any errors
<img width="625" height="145" alt="Image" src="https://github.com/user-attachments/assets/16ddb0ab-3d51-4aa2-9ca8-2a1db022efb7" />

meganrogge:
I think this issue has been around for some time.

dbaeumer:
This got fixed in the meantime.

PR #322762, opened 2026-06-24
Implement ownMarkers management in AbstractProblemCollector

dbaeumer:
Enhance the AbstractProblemCollector to manage stale markers effectively across document state changes. This change addresses issue #322730 by ensuring that markers created by the collector are cleaned up appropriately when a document's state changes. Tests confirm that stale markers are removed when a closed document is opened and that markers from other producers are preserved.


dbaeumer:
Closing. Wrong way to fix this.

PR #322787, opened 2026-06-24
Add settings gear action to voice mode toolbar

meganrogge:
Adds a settings gear icon to the chat execute toolbar when voice mode is connected. Clicking it opens a quick pick with two options:

- **Select Microphone** — runs the `agentsVoice.selectMicrophone` command to change mic device
- **Voice Mode Settings** — opens the Settings UI filtered to all `agents.voice.*` settings

The action:
- Uses `Codicon.settingsGear` icon
- Shows only when voice is connected and in the initiated widget


meganrogge:
We should wait on this until settings are discoverable

PR #322828, opened 2026-06-24
agentHost: implement Claude session fork (Phase 6.5)

TylerLeonhardt:
## Phase 6.5 — Claude session fork

Implements `ClaudeAgent.createSession({ fork })`, replacing the `TODO: Phase 6.5` throw. This wires up the Agents-window "Fork conversation from this point" affordance for Claude sessions (which is Claude's only "truncate-by-fork" path — Claude has no in-place `truncateSession`).

### What it does
- Translates the protocol `turnId` → the SDK envelope `uuid` of that turn's **last assistant message** by walking the source transcript on demand (`resolveForkAnchorUuid`, co-located with the replay mapper so it shares the exact turn-boundary rule — no parser drift).
- Calls a new `forkSession` SDK binding with `upToMessageId` (INCLUSIVE, keep `[0..N]`).
- Inherits the source overlay (`model` / `permissionMode` / `agent`; **not** `customizationDirectory`, which is per-session and re-synced).
- `config.fork.turnIdMapping` is a no-op for Claude — `forkSession` already remaps every internal uuid and Claude keeps no per-turn event-id DB.

### Lifecycle (CONTEXT M9)
Fork writes the forked session file but does **not** start the SDK `Query`. Materialization is deferred to the first `sendMessage`, which resumes from disk. This avoids firing `onDidMaterializeSession` and pipeline progress signals before `AgentService.createSession` registers the (non-provisional) forked session — both of which otherwise produced `unknown session` warnings. This matches Copilot, whose resume path likewise never fires the materialize event. *(This was the original eager-materialize design corrected during the live E2E — see the plan's Drift section.)*

### Verification
- **171** unit/integration tests green; `typecheck-client`, `valid-layers-check`, and ESLint clean.
- **Live E2E** (fork-and-continue): 3-turn Claude session → fork at the middle turn → new session with the kept prefix and the dropped tail → new turn lands via `Options.resume` → survives an agent-host restart (fork ✕ Phase 13 replay). Both `unknown session` warnings verified at count 0. Forked-session `cwd` survives the fork.

### Files
- `claudeReplayMapper.ts` — new exported `resolveForkAnchorUuid`.
- `claudeAgentSdkService.ts` — new `forkSession` binding.
- `claudeAgent.ts` — `_forkSession` flow (deferred materialization).
- `roadmap.md` / `phase6.5-plan.md` — marked done + retrospective.
- test files — fork tests + fake `forkSession` conformance stubs.

Draft pending review.


TylerLeonhardt:
Thanks for the review — addressed in 3607dc65e74:

- **Verbose comments** (`resolveForkAnchorUuid` JSDoc, `_forkSession` JSDoc, the two inline blocks, and the test "tracer" comment): tightened to 1–2 lines each, relying on the assertions + phase docs for detail.
- **`_forkSession` JSDoc eager/lazy contradiction**: removed the stale "materializes EAGERLY" wording; it now reflects the actual lazy-materialization behavior only.
- **Missing-cwd bug**: `_forkSession` now **fails fast** with a clear error when the forked session has no working directory (SDK `cwd` missing and none supplied), instead of returning a URI that would later break in `_resumeSession` on the first `sendMessage`. Added a covering test.

159 unit/integration tests green.

roblourens (approved):

PR #323001, opened 2026-06-25
ci: install CVE-fixed libexpat to fix fontconfig SIGSEGV in Electron smoke tests

alexdima:
Supersedes the from-source build idea — installs a prebuilt official Ubuntu `.deb`, fetched from the **Azure Ubuntu mirror the runner already uses**. Also includes the revert of #322909 (so it works whether or not the standalone revert #322994 lands first).

## Root cause (confirmed)

Intermittent Linux `Electron-Smoke` failures are a **SIGSEGV in `libexpat` during fontconfig parsing at Electron startup**, before any window appears. Frame-pointer (RBP) backtrace from a CI crash dump (run 28188049700):

```
#0  libc.so.6        (NULL deref, fault addr 0x0)
#1-3 libexpat.so.1   (XML_ParseBuffer ...)
#4-9 libfontconfig.so.1  (FcInit / config parse)
#10 libpangoft2-1.0.so.0
#11 libglib-2.0.so.0  (worker thread)
#12 libc.so.6        (thread start)
```

Ubuntu 24.04 (noble) ships **libexpat1 2.6.1**; the relevant parser NULL-deref fixes are only in **expat ≥ 2.7.5**, which noble hasn't packaged. Earlier fontconfig-config tweaks only perturbed timing (removing the `<!DOCTYPE>` in #322909 actually *doubled* the crash rate, ~8% → ~19% — reverted here).

## Fix

Install the official prebuilt **`libexpat1` 2.8.1 `.deb`** onto the noble runner:

- Fetched from the **Azure Ubuntu mirror the runner already uses** (`azure.archive.ubuntu.com`, in-datacenter), falling back to the canonical Ubuntu archive and then Launchpad's permanent store. All three serve the **byte-identical** official build; we **pin the sha256**.
- That package `Pre-Depends: libc6 (>= 2.38)` only, and noble has libc6 **2.39**, so `dpkg -i` installs it cleanly with no dependency conflict.
- It replaces the system `libexpat.so.1` (ABI-stable SONAME) for **every** process, including the Electron browser process where the crash happens — no `LD_PRELOAD` needed.
- The step **auto-skips** once the runner ships `libexpat1 >= 2.7.5` (so it self-removes when noble backports the fix).

The minimal fontconfig config (DOCTYPE restored via the revert) is kept as secondary stabilization.

## Why not plain `apt install libexpat1` / why this source

- noble's apt pocket only carries **2.6.1**; the fix lives in newer releases (questing 2.7.1, resolute 2.7.4, stonking 2.8.1).
- `packages.microsoft.com` and `dl.google.com` (the runner's other apt origins) don't carry `libexpat1`.
- The Azure mirror is a full mirror of the Ubuntu archive, so its **pool** hosts the newer-release deb too — byte-identical to Launchpad (verified). So this *is* "use the package manager's source", just pulling a build noble's own pocket doesn't carry yet.

## Validation

- Workflow YAML parses; no duplicate steps.
- Verified the 2.8.1 deb `Pre-Depends: libc6 (>= 2.38)` (installs on noble's 2.39).
- Verified the Azure mirror, the Ubuntu archive, and Launchpad all return the deb with the **same sha256**; simulated the download+verify loop end-to-end.
- Real proof is CI crash-rate; companion PR #322905 (fail-fast launch) surfaces any residual crash quickly instead of as a 120s opaque timeout.

### DIFF
--- a/.github/workflows/pr-linux-test.yml
+++ b/.github/workflows/pr-linux-test.yml
@@ -209,6 +209,59 @@ jobs:
           FONTCONFIG_EOF
           echo "FONTCONFIG_FILE=/tmp/fonts-minimal.conf" >> "$GITHUB_ENV"
 
+      # Root-cause fix for the intermittent startup SIGSEGV: Ubuntu 24.04 (noble)
+      # ships libexpat1 2.6.1, which intermittently NULL-derefs while fontconfig
+      # parses font config XML during Electron/Chromium startup. The crash
+      # signature is a glib worker thread: pangoft2 -> fontconfig (FcInit/parse)
+      # -> expat (XML_ParseBuffer) -> libc NULL deref, before any window appears.
+      # The CVE-specific fontconfig config tweaks above reduce but do not eliminate
+      # it (the runtime XML parse still reaches expat). The fix only exists in
+      # expat >= 2.7.5, which noble does not package yet. Rather than build from
+      # source, install the official prebuilt libexpat1 .deb (from a newer Ubuntu
+      # release) fetched via the Azure Ubuntu mirror the runner already uses: it
+      # Pre-Depends only on libc6 (>= 2.38), satisfied by noble's 2.39, so it
+      # installs cleanly and replaces the system libexpat.so.1 (ABI-stable SONAME)
+      # for every process, including Electron. Remove this step once noble ships
+      # libexpat1 >= 2.7.5.
+      - name: Install CVE-fixed libexpat to avoid fontconfig NULL-deref crash
+        run: |
+          set -euo pipefail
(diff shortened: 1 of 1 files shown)


alexdima:
Closing: verified ineffective. CI run 28192795744 installed libexpat 2.8.1 correctly (`libexpat1 is now: 2.8.1-1`) but still crashed. The new crash dump's frame-pointer backtrace shows the expat offsets changed (so 2.8.1 was loaded) yet the libfontconfig + libc frames are identical — the fault is a nondeterministic race in the system fontconfig/pango startup path, not an expat bug. Superseded by the launch-retry fix in #323007. The DOCTYPE regression revert lives in #322994.

Issue #322216, opened 2026-06-20
Copilot Chat BYOK `vendor: openai` appears to ignore `maxInputTokens` and `maxOutputTokens`


sierja:
VS Code Version:
1.125.1

Copilot Chat Version:
0.53.1

OS:
Windows 11

Description:
When configuring an OpenAI BYOK model through `chatLanguageModels.json` with `vendor: openai`, Copilot Chat appears to ignore the configured `maxInputTokens` and `maxOutputTokens` values.

The same configuration, when changed from `vendor: openai` to `vendor: customendpoint`, appears to honour the configured token limits correctly.

The documentation appears to imply that these model-level settings are valid for providers including `openai`, because the model configuration reference lists `vendor` examples including `azure`, `openai`, and `customendpoint`, and then lists `maxInputTokens` and `maxOutputTokens` as model-level properties. The docs also say VS Code uses the sum of these two values as the model's total context window in the Chat view.

If these settings are not intended to be supported for `vendor: openai`, the documentation should say that clearly. If they are intended to be supported, this appears to be a bug.

Model configuration:
The model is configured through `chatLanguageModels.json`, not selected from the standard built-in Copilot model list.

The relevant OpenAI provider entry uses the Responses API:

    vendor: openai
    apiType: responses
    model id: gpt-5.5
    streaming: true
    toolCalling: true
    vision: true
    thinking: true
    zeroDataRetentionEnabled: true
    reasoningEffort: medium
    maxInputTokens: 272000
    maxOutputTokens: 128000
    url: https://api.openai.com/v1/responses

Representative redacted configuration:

```json
[
  {
    "name": "OpenAI",
    "vendor": "openai",
    "apiKey": "${input:chat.lm.secret.REDACTED}",
    "models": [
      {
        "id": "gpt-5.5",
        "name": "GPT-5.5",
        "url": "https://api.openai.com/v1/responses",
        "apiType": "responses",
        "toolCalling": true,
        "vision": true,
        "thinking": true,
        "streaming": true,
        "maxInputTokens": 272000,
        "maxOutputTokens": 128000,
        "zeroDataRetentionEnabled": true,
        "supportsReasoningEffort": ["none", "low", "medium", "high", "xhigh"],
        "reasoningEffortFormat": "responses"
      }
    ],
    "settings": {
      "gpt-5.5": {
        "reasoningEffort": "medium"
      }
    }
  }
]
```

Expected:
The Context Window popover and context packing should use the configured model limits:

[...]

Issue #321761, opened 2026-06-17
"Reserved for response" ignores `maxOutputTokens` for extension-contributed models (hard-coded to 8192)


fuzzifikation:
# Problem

The Copilot context window widget shows **"Reserved for response"** as ~8192 tokens for all extension-contributed models, regardless of the `maxOutputTokens` value declared by the provider.

**Expected:** "Reserved for response" reflects the actual `maxOutputTokens` reported by the `LanguageModelChatProvider` (e.g., 32768, 65536, etc.).

**Actual:** Always ~8192 tokens, making the reservation bar a tiny sliver even for models configured with large output budgets.

## Root Cause

In the Copilot Chat extension's internal endpoint wrapper, `ExtensionContributedChatEndpoint` ignores the provider's declared `maxOutputTokens` and returns a hard-coded value:

**File:** `extensions/copilot/src/platform/endpoint/vscode-node/extChatEndpoint.ts`

```typescript
get maxOutputTokens(): number {
  // The VS Code API doesn't expose max output tokens, use a reasonable default
  return 8192;
}
```

This value flows into `toolCallingLoop.ts` where it's used as the `outputBuffer` for usage reporting:

```typescript
stream.usage({
  completionTokens: fetchResult.usage.completion_tokens,
  promptTokens: fetchResult.usage.prompt_tokens,
  outputBuffer: endpoint.maxOutputTokens,  // ← always 8192 for extension models
  promptTokenDetails,
});
```

Which then renders in `chatContextUsageWidget.ts` as the "Reserved for response" bar.

## Why This Matters

- Extension providers **do** report `maxOutputTokens` on `LanguageModelChatInformation` — but Copilot's internal wrapper never reads it
- For models with large output budgets (32K–128K), the UI misleads users into thinking only ~8K is reserved
- This is particularly misleading for reasoning/thinking models where output can legitimately be very large
- The `maxOutputTokens` value **is** used correctly for the actual vLLM `max_tokens` parameter — only the **display** is wrong

## Context

We've already solved the related issues in our extension (vLLM-2-Copilot):
- ✅ Token usage reporting via `LanguageModelDataPart` with `'usage'` MIME type and snake_case keys (not blocked by `isApiUsage()`)
- ✅ `prompt_tokens` / `completion_tokens` / `total_tokens` displayed correctly in the context widget
- ✅ `provideTokenCount` works for prompt budgeting

The only remaining gap is the `outputBuffer` / "Reserved for response" display.

[...]

Issue #316402, opened 2026-05-14
The max context of OpenRouter Model is wrong, it should be 1M.


DFVSQY:
Does this issue occur when all extensions are disabled?: Yes

Version: 1.120.0 (system setup)
Commit: 0958016b2af9f09bb4257e0df4a95e2f90590f9f
Date: 2026-05-12T20:17:22Z
Electron: 39.8.8
ElectronBuildId: 13870025
Chromium: 142.0.7444.265
Node.js: 22.22.1
V8: 14.2.231.22-electron.0
OS: Windows_NT x64 10.0.26100

Steps to Reproduce:
Pick OpenRouter Model，Max Context should be 1M.

<img width="861" height="379" alt="Image" src="https://github.com/user-attachments/assets/10d1acb2-56fd-4f26-9f85-c8f959b204f6" />

PR #323019, opened 2026-06-25
Fix BYOK token and context-window limits

vritant24:
Three BYOK bugs where models reported or used the wrong token / context-window limits:

#321761 — ExtensionContributedChatEndpoint.maxOutputTokens was hardcoded to 8192. Adds maxOutputTokens to the languageModelCapabilities proposed API, surfaces it on the extHost API object, and consumes it in extChatEndpoint.ts (return this.languageModel.maxOutputTokens ?? 8192).

#322216 — vendor: openai ignored configured per-model maxInputTokens/maxOutputTokens and instead ran discovery against built-in known-model metadata (huge windows). The provider now honors an explicitly configured models array, falling back to discovery only when none is given.

#316402 — OpenRouter context window was taken from top_provider.context_length (the routed provider, often smaller). Now prefers the model's top-level context_length with top_provider.context_length as fallback.

### DIFF
--- a/extensions/copilot/package.json
+++ b/extensions/copilot/package.json
@@ -1826,6 +1826,12 @@
               "secret": true,
               "description": "API key for OpenAI",
               "title": "API Key"
+            },
+            "zeroDataRetentionEnabled": {
+              "type": "boolean",
+              "default": false,
+              "title": "Zero Data Retention",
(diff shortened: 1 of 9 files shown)


arthurauffray:
+1

PR #323033, opened 2026-06-25
agents: remove problems tool sessions window 268e3be7

benibenj:
Created from `agents/remove-problems-tool-sessions-window-268e3be7` targeting `main`.


sandy081 (approved):

benibenj:
Not really ready to merge I think

dmitrivMS (changes requested):
I am not sure this is the right change - I tried to avoid per-tool special-casing, perhaps we can move Problems to tasks toolset and exclude the whole thing if that's what we want...

dmitrivMS:
Superseded by https://github.com/microsoft/vscode/pull/323978

PR #323047, opened 2026-06-25
UI polish for the Tools section of Chat Customizations

### LINKED ISSUE #322820: tools customization ux
1.
on windows the scrollbar seems to push content over when it appears + on mac looks like it's just using the native scrollbar. 

we should ideally use the monaco-scrollable or builtiin one that we have in vs code (used everywhere else in customizations)

Image

2.
focus should match with other customization dropdowns when clicking and focusing on them

other customizations:
Image

tools:
Image

dmitrivMS:
Fixes #322820

Improve scrollbar layout.
Consistent focus appearance.
Better keyboard navigation in the tools tree.


TylerLeonhardt (approved):
screenshots/videos would be nice

dmitrivMS:
> screenshots/videos would be nice

It's pretty much as Justin showed in the issue.

PR #323069, opened 2026-06-26
Define allowScripts in package.json

remcohaszing:
Modern versions of npm (12) require you to define allowScripts to run dependencies’ install scripts. In npm 11 this can already be enabled, and is recommended.

VSCode relies on install scripts, so `allowScripts` must be defined to work with modern setups of npm.

The changes to `package-lock.json` are autogenerated. They appear unrelated. Possibly this removes remnants from an inproperly resolved merge conflict.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -17351,9 +17351,6 @@
         "nan": "^2.23.0"
       }
     },
-    "node_modules/ssh2/node_modules/cpu-features": {
-      "optional": true
-    },
     "node_modules/stable": {
       "version": "0.1.8",
       "resolved": "https://registry.npmjs.org/stable/-/stable-0.1.8.tgz",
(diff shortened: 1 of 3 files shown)


kycutler:
Hi @remcohaszing, the current Node version used for the repo is `24.15.0` (see `/.nvmrc`) which ships with npm version `11.12.1`.

We currently don't recommend using other combinations.

These versions get updated over time but are highly coupled with other factors such as the Electron version.

remcohaszing:
The wiki states Node.js >=22 as a prerequisite (https://github.com/microsoft/vscode/wiki/How-to-Contribute#prerequisites). This is incorrect, as the development workflow indeed does not fully work with Node.js 26.

Everything else works fine with Node.js 24.18 though, which ships with npm 11.16.0. This npm version does have the `allowScripts` feature. This can be enabled by setting `strict-allow-scripts=true` in `.npmrc`, which I recommend everyone to do.

deepak1556:
Thank for starting the change, this is now being tackled in https://github.com/microsoft/vscode/pull/325428 as part of node version bump.

Issue #322467, opened 2026-06-22
Session cost not accounting for subagents?


kycutler:
Testing #322128

<img width="356" height="278" alt="Image" src="https://github.com/user-attachments/assets/ed642381-b994-4d44-a6de-84fd6072475a" />

The summary in the chat says `53.9 credits` which seems plausible (the subagent read a handful of files and output a summary).

But the cost indicator says `12.9 credits` which seems low.

I'm at least not sure which is correct.

PR #323106, opened 2026-06-26
Include subagent credits in live session cost total

hakoneriksson:
## Summary

Version 1.126 came with [session-level cost information](https://code.visualstudio.com/updates/v1_126#_session-level-cost-information) which is great! But it does not reflect the total cost when we include sub agent calls as well.

The running **Session Cost** shown in the chat context-usage widget only summed the
parent tool-calling loop's own model-call credits. Credits spent by **subagents**
(e.g. the Explore subagent) were missing from the live total — even though the
per-message footer total at the end of each turn already included them.

This makes the live total now match the final per-message total by reading the
authoritative per-turn credit total instead of a local accumulator.

## Root cause

`ToolCallingLoop` tracked a local `_accumulatedCopilotCredits` field that only
summed the credits of the calls made by that loop. Subagent loops run as separate
`ToolCallingLoop` instances with their own accumulators and are suppressed from
reporting to the context widget (`shouldReportUsageToContextWidget()` is `false`
for subagent invocations), so their spend never reached the running total.

## Fix

- Inject `IChatQuotaService` into the base `ToolCallingLoop` (and forward it through
  all subclass constructors).
- When emitting usage to the stream, read
  `IChatQuotaService.getCreditsForTurn(topLevelTurnId)` — the quota service already
  accumulates every model call (parent **and** subagent) under the top-level turn id,
  because `chatMLFetcher` records subagent calls keyed by the parent turn id.
- Remove the now-redundant `_accumulatedCopilotCredits` accumulator and its per-turn
  reset.

This keeps the live `IChatModel.sessionCost` in sync with the final per-turn total
that `chatParticipantRequestHandler` reads from the same key.

## Before / After

The message footer total already accounts for the subagent's spend; the live
**Session Cost** widget did not. (The absolute numbers differ between the two runs —
the point is the relationship between the footer total and the widget.)

**Before** — the response footer reads **5.7 credits**, but the live **Session Cost**
widget shows only **3.6 credits**: the Explore subagent's spend is missing.

chat view

**After** — the response footer reads **5.4 credits** and the live **Session Cost**
widget now matches at **5.4 credits**: the subagent's spend is included.

chat view fixed

### Supporting context — per-call cost breakdown

The subagent (`runSubagent-Explore`) does make its own billed model calls, which the per-message total already includes:

call details

call total

## Testing

- Updated `toolCallingLoopUsage.spec.ts`: the credits test now records fetches into
  `IChatQuotaService` (simulating `chatMLFetcher`) and still asserts the running
  total `[5, 10]` across iterations.
- Added a test verifying the parent loop's running total includes subagent credits
  recorded under the same turn id (`5` parent + `3` subagent = `8`).
- `npm test:unit` (vitest) for the affected spec passes (4/4).
- Copilot extension typecheck passes against latest `main`.

### DIFF
--- a/extensions/copilot/src/extension/intents/node/toolCallingLoop.ts
+++ b/extensions/copilot/src/extension/intents/node/toolCallingLoop.ts
@@ -10,6 +10,7 @@ import { IAuthenticationChatUpgradeService } from '../../../platform/authenticat
 import { IChatDebugFileLoggerService } from '../../../platform/chat/common/chatDebugFileLoggerService';
 import { IChatHookService, SessionStartHookInput, SessionStartHookOutput, StopHookInput, StopHookOutput, SubagentStartHookInput, SubagentStartHookOutput, SubagentStopHookInput, SubagentStopHookOutput } from '../../../platform/chat/common/chatHookService';
 import { FetchStreamSource, IResponsePart } from '../../../platform/chat/common/chatMLFetcher';
+import { IChatQuotaService } from '../../../platform/chat/common/chatQuotaService';
 import { CanceledResult, ChatFetchResponseType, ChatLocation, ChatResponse } from '../../../platform/chat/common/commonTypes';
 import { IHistoricalTurn, ISessionTranscriptService, ToolRequest } from '../../../platform/chat/common/sessionTranscriptService';
 import { ConfigKey, IConfigurationService } from '../../../platform/configuration/common/configurationService';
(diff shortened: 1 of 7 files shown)


hakoneriksson:

kycutler:
Duplicate of https://github.com/microsoft/vscode/issues/322467 and already fixed in https://github.com/microsoft/vscode/pull/322761

Issue #322564, opened 2026-06-23
It needs to be clear that the reason I only see Auto is because I'm in an untrusted repository


TylerLeonhardt:
https://github.com/user-attachments/assets/a44b585c-2a76-4679-a851-9d1693e00632

* Open VS Code fresh
* Sign in via the Modal
* see all models, great.
* Now open a folder (note the Untrusted Workspace)
* Open models
* 🐛 only see `Auto`

Now imagine I went straight to a workspace... you could use VS Code for the very first time and think that you don't have access to models.

We need to be clear about this so that the user doesn't think that they don't have access to models.

This started because we no longer show the workspace trust dialog at start.

rwoll:
@TylerLeonhardt - if they try to chat, they will be prompted to Trust the workspace and then the models will be loaded. Yesterday a bug was fixed, that did not trigger trust. I will look into maybe greying out the whole panel if in untrusted workspace? cc: @jo-oikawa - wdyt?

rwoll:
<img src="https://github.com/user-attachments/assets/509fbdb9-8199-4767-9630-8e3734989e8a" alt="Image" width="500" height="307">
^ my picker is showing models even in untrsuted workspaces. I will try with `--transient` to see if that still repro's it.

rwoll:
I repro'd with a fresh install and clearing some cache files. Playing with some different UX. Stay tuned…

rwoll:
This is now fixed for the Local harness, but upon futher testing AH behaves differently. In Untrusted Workspaces, the AH is still running so even tho you can't send messages without triggering the Trust Workspace dialog, the Model Picker will load models:

## Local in Untrusted

<img src="https://github.com/user-attachments/assets/ae376aab-b10a-41c1-838b-861c155899a6" alt="Image" width="1552" height="1012">

## Agent Host in Untrusted
<img src="https://github.com/user-attachments/assets/d94fe5de-31ee-4d5c-ae53-bc25fe14b3ad" alt="Image" width="1552" height="1012">


@roblourens @TylerLeonhardt @lramos15 - is the above OK? Unlike @TylerLeonhardt's original bug report, the AH-powered model picker in Untrusted does not mis-represent about which items are available.

TylerLeonhardt:
I'm ok with that so long as the modal shows up before the message goes through

PR #323183, opened 2026-06-26
Model picker: Restricted Mode takes precedence over live agent-host/BYOK models

rwoll:
## What

Screenshot 2026-06-26 at 10 51 49 AM

In an untrusted (**Restricted Mode**) workspace, the Agent Host model picker (Copilot CLI / Claude) loaded models and showed **"Auto"**, instead of the **"Pick Model" + Trust Workspace** affordance that the Local harness correctly shows.

Specifically addresses https://github.com/microsoft/vscode/issues/322564#issuecomment-4811816607.

## Why

The agent-host harnesses (`claude-code`, `copilotcli`) register their session-scoped models **without a workspace-trust gate** (`claudeCodeModels.ts`, `copilotCli.ts`), so those models stay live in Restricted Mode. `getModelPickerUnavailableReason` treated any live, picker-offered model as usable and returned "available", masking the Restricted state. General-purpose models are withheld while untrusted, which is why the Local harness behaved correctly.

## Fix

`getModelPickerUnavailableReason` now reports `Restricted` for an untrusted workspace **before** the "a live picker-offered model wins" check. A single change fixes every picker that routes through this function (chat editor input + sessions config bar).

Per design discussion, **Restricted Mode disables all models, including BYOK.** BYOK continues to work in a **trusted** workspace regardless of sign-in state, because the live-model check still runs (and beats setup-required) when trusted.

## Tests

Updated the `getModelPickerUnavailableReason` suite and added an explicit _"Restricted Mode disables even a live BYOK model"_ case.

- `ChatModelSelectionLogic`: 155 passing
- `buildModelPickerItems`: 77 passing

Relates to #322794, #322564.

### DIFF
--- a/src/vs/workbench/contrib/chat/browser/widget/input/chatModelSelectionLogic.ts
+++ b/src/vs/workbench/contrib/chat/browser/widget/input/chatModelSelectionLogic.ts
@@ -420,12 +420,18 @@ export const enum ModelPickerUnavailableReason {
  * - models registered for other surfaces such as agent-host session-scoped models
  *   (live in the global registry but not offered by this picker).
  *
- * A live, picker-offered model (e.g. BYOK) always wins, so BYOK and anonymous
- * access are never shown an unavailable state. `trusted` reflects
- * `isWorkspaceTrusted()` (which is `true` when trust is disabled entirely) and is
- * only authoritative once `trustInitialized` is `true`; until then this returns
- * `undefined` to avoid a trusted workspace briefly rendering as unavailable at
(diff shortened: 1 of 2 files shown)


rwoll:
cc: @roblourens @TylerLeonhardt for awareness.

@lramos15 - assigned the review specifically to you as an area expert. Thanks!

pwang347 (approved):

rzhao271 (approved):

vritant24 (approved):

benvillalobos (approved):

PR #323206, opened 2026-06-26
Streamline update mode configuration and handling

dmitrivMS:
Fixes #289144

Enables changing update mode without restart, allowing policy updates to cancel an in-progress update.


deepak1556 (approved):
LGTM

PR #323220, opened 2026-06-26
Improve turn details for Auto model in the context of free accounts

### LINKED ISSUE #321032: Model selected by Auto does not show in agent host sessions
Image

- select Agent Host
- Select Auto 
- Run a request
- Upon completion, I do not see the model that was selected 

This is what I see in Local sessions 

Image

dmitrivMS:
Fixes #321032

Will show `Auto (model-id)` instead of `Auto` when model cannot be resolved (happens with free/student accounts).


src/vs/workbench/contrib/chat/browser/agentSessions/agentHost/stateToProgressAdapter.ts
```diff
@@ -166,6 +166,50 @@ export interface TurnModelLookup {
...
+	return [displayName, model.pricing].filter(Boolean).join(' · ');
+}
+
+/** Expands the synthetic Auto model to `Auto (<resolved model id>)` when the turn reported one. */
+function formatTurnModelName(model: ITurnResponseModel, rawModelId: string | undefined): string {
+	if (model.id === AGENT_HOST_AUTO_MODEL_ID && rawModelId && rawModelId !== AGENT_HOST_AUTO_MODEL_ID) {
+		return localize('agentHost.responseDetails.autoModel', "{0} ({1})", model.name, rawModelId);
+	}
+	return model.name;
+}
```
connor4312: @dmitrivMS I suggest that after landing #323211, we instead use the `message.model` (user-picked model) vs `message.usage.model` (usage-billed model) to do this in a generic way, rather than hardcoding auto.

Feel free to merge and we can take it as a followup, or wait for it to go in :)

connor4312 (dismissed):

src/vs/workbench/contrib/chat/browser/agentSessions/agentHost/stateToProgressAdapter.ts
dmitrivMS: That's ok, I can wait.

justschen (approved):

PR #323223, opened 2026-06-26
agentHost: route session state through chat channels

connor4312:
This cleans up our action routing. Previously with the multi-chat migration, we did some logic where we would still dispatch actions on the session but then route them to the default chat. This caused features in non-default chats (like subagent permission approvals, which is how I first went down this small rabbit hole) to not work.

In Copilot's words:

## Summary
- move agent-host turn state, tool confirmations, and subagent routing onto AHP chat channels
- persist session draft/config metadata through side effects and restore fallback draft state from prior requests
- harden malformed chat-channel handling and unblock restore when subscriptions enter error state
- clean up nested NKeyMap deleteAll behavior

## Validation
- npm run typecheck-client
- npm run valid-layers-check
- npm run precommit
- focused agent-host/map subscription tests


src/vs/sessions/services/sessions/common/session.ts
```diff
@@ -302,6 +312,8 @@ export interface IChat {
 	readonly description: IObservable<IMarkdownString | undefined>;
 	/** Timestamp of when the last agent turn ended, if any. */
 	readonly lastTurnEnd: IObservable<Date | undefined>;
+	/** How the chat came into existence, if provided by the backend. */
+	readonly origin?: IChatOrigin;
```
connor4312, line 316: Later on I want to add an affordance like we have for terminals where the user can click a button to reveal a subagent chat in its own multi-chat tab. But for now, this just hides them (when origin is a tool call)

dmitrivMS (approved):

zhichli (approved):

PR #323269, opened 2026-06-27
BYOK: discover models for customendpoint from /models (#319968)

### LINKED ISSUE #319968: Feature request: Optional model discovery for customendpoint with OpenRouter or OpenAI-compatible providers
Currently, when using `vendor: "customendpoint"` in `chatLanguageModels.json` (such as with OpenRouter or any OpenAI-compatible provider), models must be listed manually in the `models` array. This is tedious and error-prone when the provider offers many available models or updates its catalog frequently.

**Request:**
- Support optional, automatic model discovery for `customendpoint` providers that expose a compatible model-list API (e.g., OpenAI's `/v1/models`).
- If `url` points to a standard OpenAI-compatible endpoint and discovery is successful, VS Code should fetch and display available models automatically in the picker and config.
- Allow manual `models` entries as an override or fallback, so users can customize/limit their view or supply capability details (tool calling, etc.) for specific models.
- If discovery fails (auth, network, unsupported schema), fall back gracefully to manual or prompt for user action.

### Why this matters
- Providers like OpenRouter expose dozens (sometimes hundreds) of models via a models endpoint, but manual config is not scalable as models change.
- Auto-discovery reduces repetitive config, keeps lists in sync, and matches user expectations based on other provider experiences.
- This would align `customendpoint` with capabilities offered by extensions and some built-in providers for model management.

**Example:**
- User sets up OpenRouter with `vendor: "customendpoint"` and only configures the main endpoint and API key. VS Code should show all available and permitted models in the picker, just as the web portal does.

### References
- Related issues: #318705, #316916
- Docs: https://code.visualstudio.com/docs/copilot/customization/language-models

Thanks for considering!

sebdanielsson:
Fixes #319968

This PR adds model discovery to OpenAI-compatible custom endpoints so that users don't have to specify each model manually in `models[]`. As models can be added and removed in internal LLM gateways, this would be a huge quality-of-life improvement for us. It also makes it easier to onboard developers.

- Implement resolveCustomEndpointModelCapabilities(): map an OpenAI-compatible /models entry to capabilities (sets tool calling on / vision off by default).
- getAllModels(): discover from `url` and merge manual `models[]` over the result by id; fall back to the manual list and warn once on failure.
- Expose a top-level `url` on the customendpoint configuration schema.

### DIFF
--- a/extensions/copilot/package.json
+++ b/extensions/copilot/package.json
@@ -2040,8 +2040,16 @@
               "title": "API Type",
               "markdownDescription": "Default request/response format for models in this group. Individual models can override this with their own `apiType` property; when both are unset the type is inferred from the URL path."
             },
+            "url": {
+              "type": "string",
+              "pattern": "^https?://.+",
+              "patternErrorMessage": "URL must start with http:// or https://",
+              "title": "Discovery URL",
+              "markdownDescription": "Base URL of an OpenAI-compatible provider. When set, models are auto-discovered from its `/models` endpoint; `models` below refine them."
+            },
             "models": {
               "type": "array",
+              "markdownDescription": "Models for this provider. Required without a discovery `url`; otherwise they override discovered models and act as a fallback.",
               "defaultSnippets": [
                 {
                   "label": "New Model",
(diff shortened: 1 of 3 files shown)


dkattan:
LGTM

PR #323298, opened 2026-06-27
Agent host: use assistant intent for session activity

roblourens:
## Summary

- map Copilot SDK `assistant.intent` events to Agent Host session activity
- remove the legacy live `report_intent` tool parsing
- clear session activity when the SDK emits an empty intent or becomes idle
- update unit coverage for the new event flow

## Why

The Copilot runtime removed the `report_intent` tool and now derives current intent from SQL todo state, emitting it through the ephemeral `assistant.intent` event. `intentionSummary` remains per-tool display/context metadata and permission request `intention` describes permission operations, so neither is the session-level replacement.

## Testing

- `npm run build-fast`
- `npm run typecheck-client`
- `scripts/test.sh --runGlob **/copilotAgentSession.test.js` (160 passing)
- `npm run precommit`

(Written by Copilot)

### DIFF
--- a/src/vs/platform/agentHost/node/copilot/copilotAgentSession.ts
+++ b/src/vs/platform/agentHost/node/copilot/copilotAgentSession.ts
@@ -567,7 +567,7 @@ export class CopilotAgentSession extends Disposable {
 	private readonly _pendingMcpSamplings = new Set<string>();
 
 	/** Tracks whether a non-empty activity has been published, so we only emit a clear when needed. */
-	private _hasReportedActivity = false;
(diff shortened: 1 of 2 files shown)


roblourens:
Follow-up note: the activity renderers still exist, but activity remains session-scoped while conversation state now flows through chat channels. A later change should add chat-scoped activity updates and preserve activity-only catalog changes; that work is intentionally out of scope for this PR.

(Written by Copilot)

dmitrivMS (approved):

PR #323313, opened 2026-06-28
ci: pin GitHub Actions to full commit SHAs

XananasX7:
## Summary

Pin all unpinned GitHub Actions to immutable full-length commit SHAs.

**Why:** Mutable tags (`@v3`, `@main`) can be silently redirected to malicious commits. SHA-pinning ensures reviewed code runs in CI — defense-in-depth against supply-chain attacks.

**Changes:** 41 reference(s) across 13 workflow file(s). Version tags retained as inline comments.

**References:**
- [GitHub: Security hardening for GitHub Actions](https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions)
- [OpenSSF Scorecard: Pinned-Dependencies](https://securityscorecards.dev)

*Automated security hardening PR.*

### DIFF
--- a/.github/workflows/api-proposal-version-check.yml
+++ b/.github/workflows/api-proposal-version-check.yml
@@ -35,7 +35,7 @@ jobs:
     steps:
       - name: Get PR info
         id: pr_info
-        uses: actions/github-script@v9
+        uses: actions/github-script@3a2844b7e9c422d3c10d287c895573f7108da1b3 # v9
         with:
           script: |
             let prNumber, headSha, baseSha;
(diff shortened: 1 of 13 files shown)


Yoyokrazy:
Thanks for putting this together, and for flagging the supply-chain hardening — SHA-pinning our Actions is exactly the right call. We're landing the same change via #329034 (a team-authored reupload needed to clear the `Prevent engineering system changes in PRs` guard, which keys off PR author). That PR covers the same workflows plus a few more (`restore/save-node-modules` composite actions, `dependabot.yml`, `css-order-scan.yml`, `telemetry.yml`); the only file unique to this PR (`api-proposal-version-check.yml`) has since been removed from `main`. Closing this as superseded — appreciate the contribution! 🙏

Issue #322566, opened 2026-06-23
Renamed chat session title reverts to auto-generated name on reopen (1.126.0)


ArunMallick:
- VS Code Version: 1.126.0 (user setup)
- Commit: 7e7950df89d055b5a378379db9ee14290772148a
- OS Version: Windows_NT x64 10.0.26200

## Steps to Reproduce

1. Open a chat/agent session that has an auto-generated title.
2. Open the sessions list (`Chat: Show Chats...`) and rename the session to a custom name (e.g. "Access Management and PR reviews").
3. The new name appears to save.
4. Close and reopen the session (or reload the window).

## Expected

The custom title persists.

## Actual

The title reverts to the previous auto-generated name as soon as the session is reopened. The manual rename does not persist.

## Notes

- Reproduces both when renaming from the open tab and from the sessions list.
- `chat.sessionSync.enabled` is enabled (default); unsure if sync is overwriting the local rename.

bastabc:
Able to replicate the same. It updates and then reverts to original auto-generated title in couple minutes.

PR #323323, opened 2026-06-28
 Fix: Session title rename not persisting after reload

### LINKED ISSUE #283514: Allow to rename local chat sessions

tamuratak:
## Fix: Session title rename not persisting after reload

Renaming a chat session title (e.g. via the sidebar) does not survive a window reload, and the renamed title is not reflected immediately in the UI. Close #283514, #318442, #322566

### Root cause

`setSessionTitle` in [ChatSessionStore](https://github.com/microsoft/vscode/blob/f0b9006a8e2f175f978d325dc2c5d05b11e147fb/src/vs/workbench/contrib/chat/common/model/chatSessionStore.ts#L503) updates only the in-memory index, not the session data file. The file is written later by `storeSessions`. If the window is reloaded before `storeSessions` runs, `readSession` reads the old title from the file, and the rename is lost.

Additionally, `setCustomTitle` on the chat model fires `model.onDidChange` (not an observable signal), so the sessions framework in [LocalChatSessionsProvider](https://github.com/microsoft/vscode/blob/main/src/vs/sessions/contrib/providers/localChatSessions/browser/localChatSessionsProvider.ts) never hears about it, so the renamed title is not reflected in the sidebar.

### Changes

**[chatSessionStore.ts](https://github.com/microsoft/vscode/blob/main/src/vs/workbench/contrib/chat/common/model/chatSessionStore.ts)** — After reading session data from disk, compare the file's `customTitle` with the index title. If they differ, prefer the index title. This closes the gap between the index (updated immediately on rename) and the file (updated lazily by `storeSessions`).

**[localChatSessionsProvider.ts](https://github.com/microsoft/vscode/blob/main/src/vs/sessions/contrib/providers/localChatSessions/browser/localChatSessionsProvider.ts)** — Two changes:
1. `trackModel` now subscribes to `model.onDidChange` for `setCustomTitle` events in addition to the existing `autorun`, so title changes are picked up without delay.
2. Listen to `chatService.onDidCreateModel` to wire up title/status tracking when a model is created (e.g. when the user opens a session), ensuring renames propagate to the sessions framework.

### Testing

1. Open a chat session, send a message.
2. Rename the session title via the sidebar.
3. Reload the window — the renamed title should persist.
4. Rename while the session is closed — the title should update in the sidebar immediately when the session is opened.

Written with GitHub Copilot + MiMo-V2.5-Pro

CC: @osortega

### DIFF
--- a/src/vs/sessions/contrib/providers/localChatSessions/browser/localChatSessionsProvider.ts
+++ b/src/vs/sessions/contrib/providers/localChatSessions/browser/localChatSessionsProvider.ts
@@ -5,7 +5,7 @@
 
 import { Emitter, Event } from '../../../../../base/common/event.js';
 import { Codicon } from '../../../../../base/common/codicons.js';
-import { Disposable, DisposableMap, MutableDisposable } from '../../../../../base/common/lifecycle.js';
+import { Disposable, DisposableMap, DisposableStore, MutableDisposable } from '../../../../../base/common/lifecycle.js';
 import { Schemas } from '../../../../../base/common/network.js';
 import { autorun, constObservable, IObservable, IReader, ISettableObservable, observableFromEvent, observableValue, transaction } from '../../../../../base/common/observable.js';
 import { ThemeIcon } from '../../../../../base/common/themables.js';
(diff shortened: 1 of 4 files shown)


bastabc:
Able to replicate same issue. It updates and then reverts to original auto-generated title in couple minutes.

PR #323513, opened 2026-06-29
Add image carousel status bar entries

oaslananka:
Fixes #271120

This adds status bar entries for the image carousel editor so locally viewed images expose the current zoom level and image dimensions in the workbench status bar.

The entries are scoped to editor visibility and hidden for videos.


oaslananka:

PR #323520, opened 2026-06-29
Skip CG NOTICE steps on CI test builds

benvillalobos:
## Context

The `Pull CG NOTICE (background)` and `Apply CG NOTICE` steps in the compile templates exist to overwrite the repo-root `ThirdPartyNotices.txt` with the Component Governance–generated notice right before gulp packages the product. That only matters for a build that actually **ships** the packaged product.

CI test jobs don't ship anything. With `VSCODE_CIBUILD=true`, `product-build.yml` spins up three jobs — **Electron Tests**, **Browser Tests**, **Remote Tests** — that compile and build a client/server purely to run tests against, then discard it. Generating and applying the CG notice in those jobs is wasted work.

It's also the source of an intermittent failure: the background `deemon` daemon started by `Pull CG NOTICE` has to survive the entire long compile + test phase, and when it gets killed mid-job the `deemon --attach` in `Apply CG NOTICE` exits non-zero and fails the job. Seen in builds 451901 and 451613 (both green on retry once the daemon survived). `downloadNotice.ts` itself is non-fatal by design — the failure comes entirely from the attach wrapper running in a job that never needed the notice.

## Changes made

- Gate both `Pull CG NOTICE (background)` and `Apply CG NOTICE` behind `${{ if ne(parameters.VSCODE_CIBUILD, true) }}` in all three compile templates (win32 / linux / darwin). This is the same guard already used for the other publish-only steps in these files (Wait for CLI artifact, Generate Group Policy, Download Explorer dll, Prepare appx).
- Both the detach and the attach are gated by the identical condition, so they stay paired — never one without the other.
- Publishing builds (`VSCODE_CIBUILD=false`) are unaffected: the notice is still pulled and applied exactly as before.

Net effect: the CG NOTICE steps stop running in the Electron/Browser/Remote Tests jobs entirely, removing both the wasted work and the failure surface. Publishing builds keep shipping the CG notice.

## References

- Builds 451901, 451613

### DIFF
--- a/build/azure-pipelines/darwin/steps/product-build-darwin-compile.yml
+++ b/build/azure-pipelines/darwin/steps/product-build-darwin-compile.yml
@@ -121,11 +121,15 @@ steps:
 
   # Start pulling the CG-generated NOTICE from the parallel Quality stage
   # in the background (overwrites repo-root ThirdPartyNotices.txt before packaging).
-  - script: npx deemon --detach --wait -- node build/azure-pipelines/common/downloadNotice.ts
-    env:
-      SYSTEM_ACCESSTOKEN: $(System.AccessToken)
-      VSCODE_OVERWRITE_TPN: $(VSCODE_OVERWRITE_TPN)
-    displayName: Pull CG NOTICE (background)
+  # Only on publishing builds (CIBUILD=false): CI test jobs build a throwaway
+  # product and never ship a notice, so generating/applying it there is wasted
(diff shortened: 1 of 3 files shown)


benvillalobos:
Superseded by #323516, which now carries the full fix: the CIBUILD gating from this PR plus continueOnError on the Apply step (defense in depth). Consolidating into one PR.

Issue #322776, opened 2026-06-24
Support Multi-Chat in Claude harness


TylerLeonhardt:
Summary

Add Multi-Chat support to the Claude harness so it matches the Copilot experience.

Background

- The Multi-Chat feature (previously called Sub-Sessions) is currently available for the Copilot harness and works in AHP sessions for Copilot.
- Team feedback and user studies have shown confusion around the feature and its name; the UX should be consistent across harnesses rather than Copilot-only.
- Claude AHP sessions do not currently show the Multi-Chat UI.

Proposal

- Implement Multi-Chat in the Claude harness with parity to the Copilot implementation and UX.
- Ensure Multi-Chat is available in AHP sessions for Claude as it is for Copilot.
- Minimize agent integration effort (sessions should be transparent to agents where possible), and document any agent-facing changes required.
- Update tests, e2e scenarios, and relevant documentation to reflect Multi-Chat availability in the Claude harness.

Acceptance criteria

- Multi-Chat UI appears and behaves the same in the Claude harness as it does in Copilot.
- Multi-Chat is available and functional in AHP sessions for Claude.
- Agent/handler code for Claude either requires no changes or any required changes are documented and minimal.
- Relevant tests and documentation updated.

Implementation notes

- Verify whether sessions need to be linked or if the agent surface can remain unchanged (transparent sessions).
- Coordinate UX changes with the Copilot harness team to ensure consistency and avoid repeating prior confusion around naming/behavior.
- Add testing steps to validate in manual and automated tests across the Claude harness and AHP sessions.

---

Split from the combined Claude & Codex issue. Codex tracked separately in #323624.

[View original Slack conversation](https://vscodeteam.slack.com/archives/C09MVCZKLP4/p1782230045706259?thread_ts=1782230045.706259&cid=C09MVCZKLP4)

PR #323625, opened 2026-06-29
sessions: support Multi-Chat in the Claude agent-host harness

sandy081:
## What

Brings Multi-Chat (additional / peer chats within one session) to the **Claude** agent-host harness, with parity to the existing Copilot experience. A Claude session can now host several peer chats that share one backend scope (working directory, model, agent, customizations), each backed by its own SDK conversation.

Implements the Claude half of #322776 (Codex is tracked separately).

## Why

Multi-Chat previously only worked for Copilot CLI agent-host sessions; Claude sessions did not surface the "Add Chat" / fork-into-chat UI. The renderer and agent-service layers were already provider-agnostic, so this is concentrated in the Claude backend plus a one-line capability flag.

## Changes

**Renderer capability**
- Enable `supportsMultipleChats` for the `claude` logical session type (`baseAgentHostSessionsProvider`). No setting / kill-switch.

**Claude agent backend (`claudeAgent.ts`)**
- New `_chatSessions` map of per-chat `ClaudeAgentSession` instances (chat URI as identity), parallel to `_sessions`.
- Implements `createChat` / `disposeChat` / `getChats` with a persisted `claude.chats` catalog and lazy resume of restored/forked peer chats.
- Per-chat routing on `sendMessage` / `abortSession` / `changeModel` / `changeAgent` / `getSessionMessages` / `setPendingMessages`.
- Fork-into-chat via the Claude SDK `forkSession`; falls back to a fresh conversation when the fork turn can't be resolved.

**Signal-routing fix (`claudeAgentSession.ts`)**
- A peer chat's `sessionUri` is already its `ahp-chat` channel, so derive `_chatChannelUri` directly instead of re-encoding via `buildDefaultChatUri` (which double-nested the channel and left peer chats stuck in progress).

**Lifecycle hardening (from code review)**
- All catalog read-modify-write serialize on the parent session id (no lost updates).
- Peer `sendMessage` holds the per-chat lock across materialize **and** send, so `disposeChat`/`disposeSession` serialize against an in-flight turn (no use-after-dispose); two-branch abort on dispose.
- `_disposeChildChats` is async + per-chat serialized to avoid zombie re-insertion; `shutdown` aborts provisional peer chats up front.
- Shape-guard the persisted catalog `model`.

## Tests

- `claudeAgent.test.ts`: peer-chat create/dispose/getChats, fork routing (+ unknown-turn fallback), conversation isolation, per-chat model persistence, parent-dispose teardown, and steering routing to peer chats.
- Full `ClaudeAgent` suite: **163 passing**. `typecheck-client` and `valid-layers-check` pass.

## Notes for reviewers

- Agent-facing surface is unchanged for the renderer/agent-service layers (transparent); all new work is in the Claude backend behind the existing optional `IAgent` peer-chat methods.
- `SESSIONS.md` updated to note multi-chat now covers `copilotcli` and `claude`.


src/vs/platform/agentHost/node/claude/claudeAgent.ts
```diff
@@ -1220,19 +1717,41 @@ export class ClaudeAgent extends Disposable implements IAgent {
...
+		// addressed by its `ahp-chat` channel URI (resolved via _chatSessions),
+		// the default chat by the session URI (resolved via _sessions).
+		const isPeerChat = !!parseChatUri(session) && !isDefaultChatUri(session);
+		const target = isPeerChat
+			? this._chatSessions.get(session.toString())?.session
+			: this._sessions.get(AgentSession.id(session))?.session;
+		this._logService.info(`[Claude] setPendingMessages for ${session.toString()}: steering=${steeringMessage?.id ?? 'none'} queued=${_queuedMessages.length}`);
+		if (!target) {
+			this._logService.warn(`[Claude] setPendingMessages: ${isPeerChat ? 'chat' : 'session'} not found for ${session.toString()}`);
 			return;
```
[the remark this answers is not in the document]
sandy081: Good catch — fixed end to end in 11b0a538435. `IAgent.setPendingMessages` now takes an optional `chat?` param; `AgentSideEffects._syncPendingMessages` forwards the chat channel (`undefined` for the default chat, the peer `ahp-chat` URI otherwise), mirroring how `abortSession` already dispatches. `ClaudeAgent` routes via that param. Added an AgentSideEffects dispatch test asserting the peer chat URI is forwarded as `chat` (and is undefined for the default chat). Copilot/Codex 3-param impls remain valid and unchanged.

src/vs/platform/agentHost/node/claude/claudeAgent.ts
```diff
@@ -255,6 +267,19 @@ export class ClaudeAgent extends Disposable implements IAgent {
...
 
+	/**
+	 * Live in-memory additional (non-default) peer chats, keyed by chat URI
+	 * string (`ahp-chat://<chatId>/<base64(sessionUri)>`). Each entry owns a
+	 * full {@link ClaudeAgentSession} whose `sessionUri` IS the chat URI and
+	 * whose `sessionId` is the chat's own SDK conversation id — distinct from
+	 * the parent session's. Parallel to {@link _sessions} (which holds default
+	 * chats keyed by raw session id); see {@link CopilotAgent} for the mirror.
+	 */
+	private readonly _chatSessions = this._register(new DisposableMap<string, ClaudeSessionEntry>());
```
connor4312: I feel like the split of the 'main chat' _sessions vs _chatSessions is generally not great. We have to special case a lot of dispatch in copilotagent and there were bugs arising from this. I feel like a nicer approach is keeping a single `_sessions` map and either have the ClaudeSessionEntries have pointers to main/sibling chats, or just wrap them in a container that does

src/vs/platform/agentHost/node/claude/claudeAgent.ts
```diff
@@ -918,6 +1371,12 @@ export class ClaudeAgent extends Disposable implements IAgent {
 	 * and returns `[]` rather than propagating — mirrors `listSessions`.
 	 */
 	async getSessionMessages(session: URI): Promise<readonly Turn[]> {
+		// On restore, the agent service calls this with a peer-chat URI to seed
+		// the additional chat's history from its own SDK conversation.
+		const chatInfo = parseChatUri(session);
+		if (chatInfo && !isDefaultChatUri(session)) {
+			return this._getChatMessages(session, chatInfo);
```
connor4312: Like `_getChatMessages` is essentially a duplicated subset of what we do for `getSessionMessages`. Peer chats/subchats should be treated much less specially imo

src/vs/platform/agentHost/node/claude/claudeAgent.ts
sandy081: @connor4312 Thanks for the feedback. 

Sorry that I am unaware of those bugs - probably I was not mentioned in them. Can you please share them to me so that I could have figured out this earlier than later.

src/vs/platform/agentHost/node/claude/claudeAgentSession.ts
```diff
@@ -275,7 +275,15 @@ export class ClaudeAgentSession extends Disposable {
...
-		this._chatChannelUri = URI.parse(buildDefaultChatUri(sessionUri));
+		// Routing channel for every signal this session emits. A default chat's
+		// `sessionUri` is the real session URI, so derive its default chat
+		// channel. A peer chat's `sessionUri` is ALREADY its `ahp-chat` channel
+		// URI (it doubles as the chat's storage identity), so use it directly —
+		// re-encoding it via `buildDefaultChatUri` would produce a double-nested
+		// channel the renderer never matches, leaving the chat stuck in progress.
+		this._chatChannelUri = isAhpChatChannel(sessionUri.toString())
+			? sessionUri
+			: URI.parse(buildDefaultChatUri(sessionUri));
```
connor4312: Except for things like predicate functions we should not be using chat URIs and session URIs interchangeably in the same parameters (it should never be the case the `isAhpChatChannel(sessionUri)`). They are generally not interchangable and this is inevitably going to cause bugs, I spent a good chunk of time on Friday pulling apart cases like that for the CopilotAgent! (Not a criticism, we needed to get an approximate shape of multichat in to iterate on, but we should avoid cases like this going forward)

connor4312 (changes requested):

src/vs/platform/agentHost/node/claude/claudeAgent.ts
connor4312: No worries, it was mostly just ones I found myself selfhosting multi-chat and then debugging from a single case https://github.com/microsoft/vscode/pull/323223

sandy081: Done in 67d6a0a784a. Dropped the parallel `_chatSessions` map: there's now a single `_sessions` map of `ClaudeSessionEntry` containers, each holding the session's default chat plus its peer chats. Dispatch resolves a chat via `_findChat(session, chat)` and teardown disposes the whole entry (main + peers) via `_teardownEntry`.

src/vs/platform/agentHost/node/claude/claudeAgent.ts
sandy081: Done in 67d6a0a784a. Removed `_getChatMessages`; `getSessionMessages` and the peer-chat path now share a single `_reconstructTurns(sdkId, routingUri, primeOn)` for the fetch+map+subagent-prime, differing only in how the SDK conversation id is resolved.

src/vs/platform/agentHost/node/claude/claudeAgentSession.ts
sandy081: Fixed in 67d6a0a784a. `ClaudeAgentSession` now takes an explicit `chatChannelUri`; `sessionUri` is always the real session URI and is never a chat URI, so `isAhpChatChannel(sessionUri)` can no longer be true. Per-chat resources (db/overlay/config-scope/server-tool advertise) key off a derived `_storageUri` (predicate on `chatChannelUri`) so peer chats stay isolated without overloading the session URI.

sandy081:
@connor4312 Reworked the Claude multi-chat plumbing per your feedback (67d6a0a784a):

- **Single `_sessions` map + container**: dropped the parallel `_chatSessions` map. `ClaudeSessionEntry` now holds the session's default chat plus its peer chats; dispatch goes through `_findChat(session, chat)` and teardown disposes the whole entry via `_teardownEntry`.
- **No session/chat URI overloading**: `ClaudeAgentSession` takes an explicit `chatChannelUri`; `sessionUri` is always the real session URI (`isAhpChatChannel(sessionUri)` can no longer be true). Per-chat resources key off a derived `_storageUri`.
- **Unified message reconstruction**: removed `_getChatMessages`; `getSessionMessages` and the peer path share `_reconstructTurns`.

No storage-keying behavior change. 425 ClaudeAgent + AgentSideEffects/CopilotAgent/AgentService node tests pass, plus typecheck/layers/hygiene. Ready for another look when you have a moment.

connor4312 (dismissed):

dbaeumer (dismissed):

benvillalobos (approved):

src/vs/platform/agentHost/node/claude/claudeAgent.ts
```diff
@@ -880,22 +1048,475 @@ export class ClaudeAgent extends Disposable implements IAgent {
...
+		const persisted = await this._readPersistedChats(session);
+		const result: IAgentLegacyChat[] = [];
+		for (const [chatId, info] of persisted) {
+			result.push({ uri: URI.parse(buildChatUri(session, chatId)), providerData: encodeProviderData(info) });
+		}
+		return result;
+	}
+
+	/**
+	 * Reads the agent's legacy peer-chat catalog (`claude.chats`) for a session,
```
TylerLeonhardt, line 1475: legacy peer-chat catalog? What is that?

src/vs/platform/agentHost/node/claude/claudeAgent.ts
```diff
@@ -880,22 +1048,475 @@ export class ClaudeAgent extends Disposable implements IAgent {
...
+			backing = persisted.get(chatInfo.chatId);
+			if (!backing) {
+				return;
+			}
+		}
+		this._chatBackings.set(chatKey, backing);
+	}
+
+	/**
+	 * Migration-only enumeration of the session's peer chats from the agent's
```
TylerLeonhardt, line 1459: What are we migrating?

src/vs/platform/agentHost/node/claude/claudeAgent.ts
```diff
@@ -255,6 +257,47 @@ export class ClaudeAgent extends Disposable implements IAgent {
...
+	 * harness gains them, Claude Teams teammates). Derived from the
+	 * `subagent_started` / `subagent_completed` signals that already flow on
+	 * {@link onDidSessionProgress}, so the orchestrator records the spawn edge
+	 * on the unified chat catalog. See {@link IAgent.onDidSpawnChat}.
+	 */
+	private readonly _onDidSpawnChat = this._register(new Emitter<IAgentSpawnChatEvent>());
+	readonly onDidSpawnChat: Event<IAgentSpawnChatEvent> = this._onDidSpawnChat.event;
+
+	private readonly _onDidEndChat = this._register(new Emitter<URI>());
+	readonly onDidEndChat: Event<URI> = this._onDidEndChat.event;
```
TylerLeonhardt, line 299: This is unused.

PR #323629, opened 2026-06-29
Fix potential memory leak in terminal link hover

dmitrivMS:
(was reported in debug output)


connor4312 (dismissed):

anthonykim1 (changes requested):
Could this use one active hover disposable for all terminal link hovers instead of a separate _linkHoverInvalidationDisposable? _showHover already returns the hover lifetime; if that returned disposable also includes the optional link.onInvalidated listener, both OSC 8 and TerminalLink hovers can share the same cleanup path.

dmitrivMS:
> Could this use one active hover disposable for all terminal link hovers instead of a separate _linkHoverInvalidationDisposable? _showHover already returns the hover lifetime; if that returned disposable also includes the optional link.onInvalidated listener, both OSC 8 and TerminalLink hovers can share the same cleanup path.

Will look into it later this week.

anthonykim1 (approved):

roblourens (approved):

roblourens (approved):

PR #323663, opened 2026-06-30
fix: make sure register handler when ipc emitter add listener

### LINKED ISSUE #324678: Issue about ChannelClient.requestEvent
When the last listener is disposed in [ChannelClient.requestEvent](https://github.com/microsoft/vscode/blob/fa84c09861e2dcbbc5708294d85a5ca0be1f7c15/src/vs/base/parts/ipc/common/ipc.ts#L542), we will remove the handler from the map. However, we will not add the handler back to the map when re-subscribing to the emitter, so the event will not be emitted anymore.

example.
```
const emitter = client.listen('hello');
const dispose1 = emitter(() => {);
dispose1.dispose();
// can not receive any messages.
const dispose2 = emitter(() => {);
```

theanarkh:
Fixes: https://github.com/microsoft/vscode/issues/324678

When the last listener is disposed, we will remove the handler from the map. However, we will not add the handler back to the map when re-subscribing to the emitter, so the event will not be emitted anymore. It is better register handler in `onWillAddFirstListener` and delete it in `onDidRemoveLastListener`.

example.
```
const emitter = client.listen('hello');
const dispose1 = emitter(() => {);
dispose1.dispose();
// can not receive any messages.
const dispose2 = emitter(() => {);

### DIFF
--- a/src/vs/base/parts/ipc/common/ipc.ts
+++ b/src/vs/base/parts/ipc/common/ipc.ts
@@ -680,6 +680,8 @@ export class ChannelClient implements IChannelClient, IDisposable {
 
 		const emitter = new Emitter<any>({
 			onWillAddFirstListener: () => {
+				const handler: IHandler = (res: IRawResponse) => emitter.fire((res as IRawEventFireResponse).data);
+				this.handlers.set(id, handler);
 				const doRequest = () => {
 					this.activeRequests.add(emitter);
 					this.sendRequest(request);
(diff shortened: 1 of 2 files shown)


roblourens (changes requested):
Thanks for the PR, I think this change is good but please file an issue with a better description of the issue, and add unit tests for the scenario

theanarkh:
@roblourens Thanks ! I have updated the code, please help review again.

roblourens:
Thanks, can you open an issue?

theanarkh:
> Thanks, can you open an issue?

Done.

roblourens (approved):

vritant24 (approved):

benvillalobos (approved):

PR #323672, opened 2026-06-30
Fix three typos in source comments

gunjanjaswal:
Three single-word typo fixes in source comments. No functional change.

- `colorThemeData.ts`: "set to to default" to "set to default" (duplicate word)
- `explorerViewer.ts`: "seperately" to "separately"
- `extHostTypes.ts`: "seperate" to "separate"

Comment-only, so there's nothing to test.


gunjanjaswal:

alexr00:
Thanks for the PR, but we prefer not to take spell check fixes: https://github.com/microsoft/vscode/wiki/How-to-Contribute#spell-check-errors

gunjanjaswal:
Thanks for the pointer to the policy, makes sense. I'll keep contributions to substantive fixes.

PR #323763, opened 2026-06-30
fix - picker: apply viewer sort/group prefs to session quick pickers

EdrilanBerisha:
fixes #303430

the quick picker for sessions (agentSessionsPicker and agentSessionsQuickAccess) was ignoring the sort-by and group-by choices set in the viewer. now both pickers read those settings from AgentSessionsFilter and:
- sort sessions in the same order as the viewer (created vs updated)
- group sessions by date using the right timestamp column
- group by repository when that grouping is selected in the viewer

also extracted groupAgentSessionsByRepository as an exported helper (same as groupAgentSessionsByDate) so the pickers can reuse it without duplicating logic

tests added for the new exported function and for sort/group behavior


EdrilanBerisha:
> @microsoft-github-policy-service agree

Issue #323765, opened 2026-06-30
"Try Again" in Chat does not allow selection of new model


danielrobbins:
Version: 1.128.0-insider (user setup)
Commit: c4632cac4f0a71ef09ccc2b6847c7ad19e98fcb6
Date: 2026-06-30T16:18:14Z
Electron: 42.5.0
ElectronBuildId: 14500164
Chromium: 148.0.7778.271
Node.js: 24.17.0
V8: 14.8.178.33-electron.0
OS: Windows_NT arm64 10.0.28000
Feature: Chat
Selected Model: Any

Steps to Reproduce:

1. Configure a custom Chat model with editor-scoped/per-model configuration, such as a custom endpoint model with reasoning effort or other model-specific settings.
2. Start a Chat request using that custom model and confirm that the initial request uses the expected selected model configuration.
3. Trigger a resend/follow-up path such as “Try Again”, a confirmation response button, a sample request, slash command sample, or another Chat UI flow that sends using a different widget model.
4. Observe that the UI still appears to have the same model selected, but the follow-up request is not routed to the newly-selected model.

## User story

I use VS Code Chat with custom language models configured through `chatLanguageModels.json`. Some of these models require per-model settings, such as reasoning effort or other model-specific options.

While testing a custom endpoint, I found that a normal first request used the expected model configuration. But after using retry/follow-up flows, the request no longer behaved as if the selected model configuration was being honored. From the user’s perspective, the Chat UI still appeared to have the correct model selected, but the request sent by the retry/follow-up path was not equivalent to a fresh send with that same visible model selection.

This is confusing because the model picker state suggests one thing, while the request options used by some resend/follow-up paths can behave differently.

## Expected behavior

Any Chat flow that sends or resends using the currently selected model should preserve both:

- the selected model ID
- the editor-scoped model configuration associated with that model

The visible model picker state should match the request options used for retry/resend/confirmation/follow-up flows.

## Actual behavior

Some call sites pass only `userSelectedModelId` from the widget’s selected model state. They do not also pass `userSelectedModelConfiguration`.

That can drop editor-scoped configuration and cause follow-up requests to use configuration that does not match the visible model picker state.

## Technical notes

`IChatSendRequestOptions` supports both `userSelectedModelId` and `userSelectedModelConfiguration`.

The service layer can fall back to global model configuration when only a model ID is provided, but that is not equivalent to preserving editor-scoped configuration from the active Chat widget.

[...]

PR #323767, opened 2026-06-30
Fix bugs related to setting the right Chat model. (Fixes Issue #323765)

danielrobbins:
This patch was originally created to address a bug where encountering an error and a "Try Again" button in the Chat window does not allow a different model to be selected, preventing the user from switching to a working model.

Upon investigation, multiple inconsistencies were found related to feeding the model selection widget's current state in various places. This patch centralizes this in getSelectedModelRequestOptions() to avoid needing to keep multiple call sites in sync going forward.

More specifically, several resend and confirmation paths passed only the selected model ID. That can drop editor-scoped model configuration, such as custom per-model settings, and cause follow-up requests to use configuration that does not match the visible model picker state. The new helper keeps the selected model ID and its scoped configuration together whenever the widget builds Chat request options.

To test: Select a third-party endpoint for a new chat, then successfully use it. Now, simulate an error on the endpoint, so that it will return an error to Chat on the next request. Now, attempt to continue with a new turn in Chat, resulting in an error from the endpoint. Chat will display a "Try Again" button. Now, pretend you are a user needing to select a working Model since your current model experienced an error. Before pressing "Try Again", select a new model, *then* hit "Try Again". With the patch, the new model will be utilized. Without the patch, the original failing model will continue to be selected, the original prompt will be re-submitted, and is likely to re-trigger the same error if the error was related to prompt content, size, quota, etc.


danielrobbins:
@microsoft-github-policy-service agree company="BreezyOps, Inc."

Giuspepe:
Thanks for your PR! Fixes #323765

Giuspepe (dismissed):

alexdima (dismissed):

roblourens (dismissed):

Giuspepe:
@danielrobbins looks like you have some tests failing

danielrobbins:
@Giuspepe Ah, yes I do. Looks like a test fixture needs updating to reflect the changes. Should have time to get that fixed today.

danielrobbins:
Should do better now on the tests.

danielrobbins:
I updated the PR after the previous test failure and the branch is mergeable again. This should be ready to go, @Giuspepe

Giuspepe (approved):

ulugbekna (approved):

lszomoru (approved):

Giuspepe:
Thanks a lot again for your contribution @danielrobbins :)

danielrobbins:
@Giuspepe you are welcome! I also have another pretty important fix for BYOK here that needs merging: https://github.com/microsoft/vscode/pull/324311

PR #323803, opened 2026-06-30
Add VSC model E prompt with keep_compact instructions

yuezhang030:

### DIFF
--- a/extensions/copilot/src/platform/endpoint/common/chatModelCapabilities.ts
+++ b/extensions/copilot/src/platform/endpoint/common/chatModelCapabilities.ts
@@ -38,6 +38,8 @@ const VSC_MODEL_HASHES_D = [
 	'e82ff0e2d4e4bae1f012dc599d520f8d61becfc4762f3717577b270be199db92',
 ];
 
+const VSC_MODEL_HASHES_E: string[] = [];
+
 
 // subset to allow replace string instead of apply patch.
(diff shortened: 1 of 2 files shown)


extensions/copilot/src/platform/endpoint/common/chatModelCapabilities.ts
```diff
@@ -38,6 +38,8 @@ const VSC_MODEL_HASHES_D = [
 	'e82ff0e2d4e4bae1f012dc599d520f8d61becfc4762f3717577b270be199db92',
 ];
 
+const VSC_MODEL_HASHES_E: string[] = [];
```
karthiknadig, line 41: This should have the hash of the model, right?

karthiknadig (changes requested):

extensions/copilot/src/platform/endpoint/common/chatModelCapabilities.ts
karthiknadig: @yuezhang030 is the hash needed here?

karthiknadig (approved):

connor4312 (approved):

extensions/copilot/src/platform/endpoint/common/chatModelCapabilities.ts
yuezhang030: We don't need hash for MAI models.

dmitrivMS (approved):

PR #323815, opened 2026-07-01
Fix blocked agent host sessions from nested subagent client tools

### LINKED ISSUE #323738: Agent Host Session Blocked
When agent searches tools in the agent window it gets stuck:

Image

version:

Image

connor4312:
Fixes #323738

## Problem

Agent host sessions hung in **"Input Needed"** with no visible prompt when a nested (depth ≥ 2) subagent invoked a client-owned tool — e.g. asking a subagent to ask a subagent to use the `problems` tool. The deep subagent's tool never ran because the renderer never observed that subagent, so its permission/execution request was never surfaced.

## Root causes (both fixed)

**1. Renderer — observation gated on a signal that never arrived.**
`tryObserveSubagent` required the subagent-*discovery* content block before subscribing to a child chat. For a nested subagent that block never reached the parent chat (see #2), so the child was never observed. It now observes as soon as the tool is a known subagent-spawning tool that is running, deriving the child chat URI from the tool id alone — robust and depth-independent. The renderer already recognizes the tool as a subagent from `_meta.toolKind`, so nothing else depended on the block.

**2. Agent host — discovery content block misrouted for nested subagents.**
`subagent_started` carried no parent context, so `_handleSubagentStarted` dispatched the `ChatToolCallContentChanged` discovery block on the **top-level** chat instead of the **immediate parent** subagent chat (a no-op there, since the spawning tool doesn't live in the top-level chat). Threaded an optional `parentToolCallId` (the tool one level up from the spawning tool) through the signal and populated it in both the copilot and claude adapters. Because subagent chats are flat (keyed off the root session), this single one-hop lookup resolves the correct parent chat **at any nesting depth** — registration/routing stay keyed by the top-level chat as required.

## Also

Keeps producing the session-level `inputNeeded` state for other clients (with tests), even though the renderer no longer consumes it.

## Tests

- Renderer: level-2 nested client-tool invocation + grouping under the root, and a variant with **no** discovery block (mirrors the exact production failure).
- Agent host: nested `subagent_started` discovery-block routing driven **three levels deep**, asserting each level's block lands on its immediate parent chat and never leaks to the top-level chat.
- `inputNeeded` production suite.

## Validation

All touched files are diagnostics-clean and the hygiene pre-commit hook passed. Unit tests could not be executed in the authoring environment (no test host); please rely on CI.

## How to repro / verify

With deferred tools enabled, send to an agent session:

> I need a nested subagent to use the "problems" tool. Please ask a subagent to ask a subagent to use the "problems" tool.

Before: session parks in "Input Needed" with no prompt. After: the deep subagent runs `problems`, results bubble back up, and the session completes.


sandy081 (approved):

aleimu:
# VSCode Agent Host Protocol (AHP) version mismatch prevents WSL connection

## Summary

The WSL extension `ms-vscode-remote.remote-wsl@0.104.x` (and likely newer 0.104.x patch versions) uses Agent Host Protocol (AHP) version **0.5.1**, but the agent host server distributed through the `stable` channel of `https://update.code.visualstudio.com/api/latest/stable` is commit **`4fe60c8b1cdac1c4c174f2fb180d0d758272d713`** (VS Code 1.127.0), which hardcodes AHP version **0.5.0**. The two are incompatible under the 0.x semver rule baked into `agentHostMain.js` (client minor must equal server minor, and client patch must be ≤ server patch), so the WSL connection fails at the `initialize` handshake.

## Environment

| Component | Version | Commit |
| --- | --- | --- |
| WSL distribution | Ubuntu 22.04.4 LTS (WSL2, kernel 6.18.35.2) | — |
| Windows VSCode | 1.128.0 *(please fill in your actual `code --version` from PowerShell)* | `fc3def6774c76082adf699d366f31a557ce5573f` |
| WSL extension | `ms-vscode-remote.remote-wsl@0.104.3` | — |
| vscode-server (regular code-server, downloaded by `wslServer.sh`) | 1.128.0 | `fc3def6774c76082adf699d366f31a557ce5573f` |
| **agent host server** (the one that runs the AHP 0.5.x stack) | **1.127.0** | **`4fe60c8b1cdac1c4c174f2fb180d0d758272d713`** |
| `bin/fc3def6.../out/vs/platform/agentHost/node/agentHostMain.js` AHP version constant | `Nl="0.5.1"` | OK |
| `cli/servers/Stable-4fe60c8/.../agentHostMain.js` AHP version constant | `var sl="0.5.0",URe=Object.freeze(["0.5.0"])` | **mismatch** |

## Symptoms

In VSCode on Windows, `Ctrl+Shift+P` → `WSL: Connect to WSL` → `Ubuntu-22.04` fails with both:

1. **Failed to connect to WSL distribution**:
   ```
   未能连接到 WSL 分发版 "Ubuntu-22.04": Client offered protocol versions
   [0.5.1], none of which are compatible with this server's version 0.5.0
   (server accepts ^0.5.0).
   ```

2. **No remote agent host entry**:
   ```
   更新 Ubuntu-22.04 失败: No remote agent host entry found for wsl:Ubuntu-22.04.
   ```
   The second error is a downstream side effect of the first: the WSL extension never finishes its handshake with the agent host supervisor, so the WSL → agent-host entry isn't registered, and `DistroUpdate` later fails to find it.

## Root cause

Two pieces of evidence nail this down.

### 1. The 4fe60c8 server rejects 0.5.1 at the AHP handshake

`cli/servers/Stable-4fe60c8b1cdac1c4c174f2fb180d0d758272d713/server/out/vs/platform/agentHost/node/agentHostMain.js` (1.37 MB, minified) contains, around offset 1,221,970:

```js
var sl="0.5.0",URe=Object.freeze(["0.5.0"]);
function aE(i){let n=/^(\d+)\.(\d+)\.(\d+)$/.exec(i); ...}
// ... later, in the initialize handler:
let s=cE(o,sl); // o = client's offered versions, sl = "0.5.0"
if(!s){
  let c={supportedVersions:[`^${sl}`], ...};
  throw new Error(
    `Client offered protocol versions [${o.join(", ")}], none of which `+
    `are compatible with this server's version ${sl} (server accepts ^${sl}).`,
    c
  );
}
```

The 0.x semver rule used in `r9()` (or its minified equivalent) requires the client to share the same major *and* minor version as the server, and the client patch must be ≤ the server patch. The WSL extension 0.104.x only offers `[0.5.1]`; 0.5.1 has the right minor (5) but a higher patch (1) than the server's 0.5.0, so the handshake is rejected.

For comparison, the same constant in the **un-minified** 1.128.0 code-server is `Nl="0.5.1"` (file: `bin/fc3def6774c76082adf699d366f31a557ce5573f/out/vs/platform/agentHost/node/agentHostMain.js`), so 1.128.0 would accept the client fine.

### 2. The agent host stable channel is stuck on 4fe60c8

Excerpt from `~/.vscode-server/cli/agent-host-stable.log` (the supervisor log written by `code-fc3def6 ... agent host`):

```
[2026-07-03 09:21:42] debug Resolved server version: 1.127.0 (commit 4fe60c8b1cdac1c4c174f2fb180d0d758272d713)
[2026-07-04 09:21:42] debug Resolved server version: 1.127.0 (commit 4fe60c8b1cdac1c4c174f2fb180d0d758272d713)
[2026-07-05 09:21:42] debug Resolved server version: 1.127.0 (commit 4fe60c8b1cdac1c4c174f2fb180d0d758272d713)
[2026-07-06 09:21:42] debug Resolved server version: 1.127.0 (commit 4fe60c8b1cdac1c4c174f2fb180d0d758272d713)
[2026-07-07 09:21:42] debug Resolved server version: 1.127.0 (commit 4fe60c8b1cdac1c4c174f2fb180d0d758272d713)
[2026-07-08 09:21:42] debug Resolved quality Stable to fc3def6774c76082adf699d366f31a557ce5573f
[2026-07-08 09:21:42] debug Resolved server version: 1.128.0 (commit fc3def6774c76082adf699d366f31a557ce5573f)
[2026-07-08 09:21:42] info  New server version available: 1.128.0 (commit fc3def6774c76082adf699d366f31a557ce5573f)
[2026-07-08 09:21:42] info  Downloading server fc3def6774c76082adf699d366f31a557ce5573f
[2026-07-08 09:22:04] info  Updated server to 1.128.0 (commit fc3def6774c76082adf699d366f31a557ce5573f)
[2026-07-08 15:21:42] debug Resolved server version: 1.128.0 (commit fc3def6774c76082adf699d366f31a557ce5573f)
[2026-07-09 03:21:42] debug Resolved server version: 1.128.0 (commit fc3def6774c76082adf699d366f31a557ce5573f)
[2026-07-09 09:21:42] debug Resolved server version: 1.128.0 (commit fc3def6774c76082adf699d366f31a557ce5573f)
[2026-07-09 15:21:42] debug Resolved server version: 1.128.0 (commit fc3def6774c76082adf699d366f31a557ce5573f)
[2026-07-10 03:21:42] debug Resolved server version: 1.128.0 (commit fc3def6774c76082adf699d366f31a557ce5573f)
[2026-07-10 09:21:45] debug Resolved server version: 1.128.0 (commit fc3def6774c76082adf699d366f31a557ce5573f)
[2026-07-10 15:21:42] debug Resolved server version: 1.128.0 (commit fc3def6774c76082adf699d366f31a557ce5573f)
[2026-07-10 21:21:42] debug Resolved server version: 1.128.0 (commit fc3def6774c76082adf699d366f31a557ce5573f)
[2026-07-11 03:21:42] debug Resolved server version: 1.128.0 (commit fc3def6774c76082adf699d366f31a557ce5573f)
[2026-07-11 09:21:42] debug Resolved server version: 1.127.0 (commit 4fe60c8b1cdac1c4c174f2fb180d0d758272d713)   ← regressed
[2026-07-11 15:21:42] debug Resolved server version: 1.127.0 (commit 4fe60c8b1cdac1c4c174f2fb180d0d758272d713)
[2026-07-12 03:21:42] debug Resolved server version: 1.127.0 (commit 4fe60c8b1cdac1c4c174f2fb180d0d758272d713)
[2026-07-12 09:21:42] debug Resolved server version: 1.127.0 (commit 4fe60c8b1cdac1c4c174f2fb180d0d758272d713)
[2026-07-12 15:21:41] debug Resolved server version: 1.127.0 (commit 4fe60c8b1cdac1c4c174f2fb180d0d758272d713)
[2026-07-12 21:21:42] debug Resolved server version: 1.127.0 (commit 4fe60c8b1cdac1c4c174f2fb180d0d758272d713)
[2026-07-13 03:53:20] debug Resolved server version: 1.127.0 (commit 4fe60c8b1cdac1c4c174f2fb180d0d758272d713)
[2026-07-13 03:53:20] info  Downloading server 4fe60c8b1cdac1c4c174f2fb180d0d758272d713
[2026-07-13 03:53:57] info  [4fe60c8]: Server ready
```

The agent host supervisor queries the update service every 6 hours. Between **2026-07-08 15:21** and **2026-07-11 09:21**, the service transitioned from returning 1.128.0 / `fc3def6` to 1.127.0 / `4fe60c8` — i.e. a regression to a *previous* stable. It has stayed on `4fe60c8` ever since. The same query that returns 1.128.0 for the regular stable channel (and which `wslDownload.sh` uses for the IDE code-server, landing on `fc3def6` / 1.128.0) returns 1.127.0 for the agent-host stable channel.

Local mitigation attempts (deleting `cli/servers/`, deleting `lru.json`, killing the supervisor process) only cause the supervisor to re-query and download `4fe60c8` again. The `code` CLI's `agent host` subcommand does not appear to honor `lru.json` as a pinned source — it always re-resolves via the update service. (Note: the `code-fc3def6` binary does have a `VSCODE_CLI_INITIAL_AH_VERSION` env-var escape hatch logged as `"Using initial agent host version override:"`; this is undocumented and the WSL extension does not propagate it.)

## Reproduction

1. Install VSCode on Windows (any 1.128.x stable build, or earlier — I observed this on 1.128.0).
2. Install the WSL extension, accept auto-update, and confirm version 0.104.x is active.
3. Have any WSL2 distro (e.g. Ubuntu 22.04) installed and registered.
4. From VSCode: `Ctrl+Shift+P` → `WSL: Connect to WSL` → pick the distro.
5. Observe: connection fails. In WSL, `cat ~/.vscode-server/cli/agent-host-stable.log` shows `Resolved server version: 1.127.0 (commit 4fe60c8b1cdac1c4c174f2fb180d0d758272d713)` and `Downloading server 4fe60c8...`. In VSCode's WSL extension log, the `Failed to connect to WSL distribution` error mentions `Client offered protocol versions [0.5.1]` / `server's version 0.5.0`.

## Expected vs actual

- **Expected**: connecting to a WSL distro succeeds; the agent host supervisor picks a server version whose AHP version is ≥ what the WSL extension offers (or vice versa — either side bumps to keep them aligned).
- **Actual**: the agent host `stable` channel serves 1.127.0 / `4fe60c8`, which uses AHP 0.5.0, while the WSL extension 0.104.x uses AHP 0.5.1. The 0.x semver comparison rejects the client, the WSL extension's agent-host entry never gets registered, and the connection is refused.

## Workaround

Downgrade the WSL extension to 0.103.x, which uses AHP 0.5.0 and matches the `4fe60c8` server:

1. VSCode → Extensions panel → search `@installed ms-vscode-remote.remote-wsl`
2. Click the gear icon → **Install Another Version…** → pick the latest `0.103.x` (e.g. `0.103.0`).
3. Reload VSCode.
4. (Recommended) right-click the extension → **Disable Auto-Update for this Extension**, so a future auto-update doesn't put you back into this broken state.

After downgrade, `agent-host-stable.log` still resolves to 1.127.0 / `4fe60c8`, but the AHP handshake now succeeds because both sides speak 0.5.0.

## Recommended fix

In `microsoft/vscode` (or wherever the `code` CLI's `agent host` subcommand and its update service are maintained), one of:

- **Most likely fix**: re-publish `4fe60c8b1cdac1c4c174f2fb180d0d758272d713` (or the latest commit that has AHP 0.5.1 baked in) to the `api/latest/stable` endpoint for the agent host channel, so the agent host supervisor downloads a version that matches the WSL extension's protocol version. The `4fe60c8` binary is older than the `fc3def6` binary that is already on the regular stable channel, so the agent host stable channel is currently behind.
- **Loosen the protocol rule**: change `r9()` in `agentHostMain.js` (and the un-minified `src/vs/platform/agentHost/...` equivalents) to allow a client patch that is *one minor ahead* of the server, e.g. by allowing the server to negotiate the intersection of `(client, server)` when they share a minor — this would have let 0.5.0 servers accept 0.5.1 clients (or vice versa) by responding with the highest mutually supported version. The current strict `<=` patch rule makes the protocol brittle when the agent host channel lags.
- **Document or surface the `VSCODE_CLI_INITIAL_AH_VERSION` env var**: if the supervisor already supports it, expose it in the WSL extension / `code tunnel user install-cli` so users can self-pin around a stale update channel.

## Additional artifacts

If helpful, the full supervisor log and the modified timestamp trace are at:

- `~/.vscode-server/cli/agent-host-stable.log` — full supervisor log
- `~/.vscode-server/cli/servers/lru.json` — `["Stable-4fe60c8b1cdac1c4c174f2fb180d0d758272d713"]`
- `~/.vscode-server/cli/servers/Stable-4fe60c8b1cdac1c4c174f2fb180d0d758272d713/server/out/vs/platform/agentHost/node/agentHostMain.js` — contains the hardcoded `var sl="0.5.0",URe=Object.freeze(["0.5.0"])`
- `~/.vscode-server/bin/fc3def6774c76082adf699d366f31a557ce5573f/out/vs/platform/agentHost/node/agentHostMain.js` — contains the corresponding `Nl="0.5.1"` (this is the version the regular code-server already ships, confirming the fix is just a matter of pointing the agent-host update channel at it)
- `~/.vscode-server/code-fc3def6774c76082adf699d366f31a557ce5573f` — the `code` CLI binary (`1.128.0`); `strings` dump contains `src/tunnels/agent_host.rs`, `VSCODE_CLI_INITIAL_AH_VERSION`, `Using initial agent host version override:`, `Resolved server version:`, `No cached server version, downloading latest...`, and the various AHP error strings.

PR #323824, opened 2026-07-01
Disable idle session eviction in agent host

roblourens:
Idle-session eviction in `_maybeEvictIdleSession` is causing issues where cached session state is dropped while clients still expect it to be observable. This change disables the eviction path for now while we investigate, and updates the related tests to expect state to remain cached after the last subscriber drops.

(Written by Copilot)


roblourens:
Addressed Copilot review comments:
- Updated JSDoc to reflect the no-op behavior.
- Added `void resource` to mark the parameter intentionally unused.
- Added a TODO noting the need to re-enable eviction or add an LRU cap.
- Renamed the multi-subscriber test to match its new assertion.

(Written by Copilot)

chrmarti (approved):

PR #323835, opened 2026-07-01
Fix parseLineAndColumnAware mistaking hex/exponential path segments for line numbers

chatman-media:
\`parseLineAndColumnAware\` (used by the \`--goto FILE:LINE(:COLUMN)\` CLI flag) decides whether a colon-delimited segment is a line/column by coercing it with \`Number()\` and checking it's not NaN. That's too permissive: \`Number("0x10")\` is 16 and \`Number("1e3")\` is 1000, so a file legitimately named something like \`report:0x10\` gets its \`0x10\` segment silently swallowed as line 16 instead of staying part of the path.

Swapped the check for a strict \`/^[0-9]+$/\` regex so only plain decimal digit segments are treated as line/column, and added a few test cases covering hex- and exponential-looking segments. Existing behavior (plain line/column parsing, Windows drive letters like \`C:\foo\bar:33\`) is unaffected since those segments were never matched by the loose check either.

### DIFF
--- a/src/vs/base/test/common/extpath.test.ts
+++ b/src/vs/base/test/common/extpath.test.ts
@@ -202,6 +202,24 @@ suite('Paths', () => {
 		assert.strictEqual(res.path, '/foo/bar:abb');
 		assert.strictEqual(res.line, undefined);
 		assert.strictEqual(res.column, undefined);
+
+		// segments that `Number()` would coerce into a valid number (hex, exponential, ...)
+		// must still be treated as part of the path, not as line/column. A file can be
+		// legally named e.g. `report:0x10` without the `0x10` part being mistaken for a line.
+		res = extpath.parseLineAndColumnAware('/foo/report:0x10');
+		assert.strictEqual(res.path, '/foo/report:0x10');
+		assert.strictEqual(res.line, undefined);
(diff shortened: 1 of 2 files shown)


src/vs/base/test/common/extpath.test.ts
```diff
@@ -202,6 +202,22 @@ suite('Paths', () => {
...
+
+		// segments `Number()` would coerce (hex, exponential, ...) must stay part of the path, not line/column
+		res = extpath.parseLineAndColumnAware('/foo/report:0x10');
+		assert.strictEqual(res.path, '/foo/report:0x10');
+		assert.strictEqual(res.line, undefined);
+		assert.strictEqual(res.column, undefined);
+
+		res = extpath.parseLineAndColumnAware('/foo/report:0x10:33');
+		assert.strictEqual(res.path, '/foo/report:0x10');
+		assert.strictEqual(res.line, 33);
```
dmitrivMS, line 214: According to PR description, I'd expect line/column to be undefined here...

dmitrivMS (changes requested):
Please see inline comment and Copilot request to reduce comment size.
Also, it would be good to file an issue for the problem at hand so we can use it for verification during testing.

PR #323836, opened 2026-07-01
Support PDF inputs for BYOK endpoints

AntonioLujanoLuna:
## Summary

Adds explicit PDF input support for BYOK models using the Responses or
Messages APIs.

Previously, PDF attachments could be represented in chat prompts but were
lost while converting between raw prompt messages and VS Code language-model
messages. Custom models also had no way to advertise PDF support independently
of vision support.

This change:

- Adds `fileInputMimeTypes` to the proposed language-model capabilities and
  Custom Endpoint model configuration.
- Advertises `application/pdf` for configured Responses and Messages endpoints.
- Preserves PDF data through raw-to-VS Code and VS Code-to-raw message
  conversion.
- Converts PDFs into Anthropic `document` blocks, including tool-result
  content.
- Keeps PDF input disabled for Chat Completions endpoints.
- Adds regression tests for capability propagation, message conversion,
  prompt rendering, and endpoint-type gating.

## How to test

1. Configure a BYOK Custom Endpoint model with `"apiType": "responses"` or
   `"apiType": "messages"` and:

   ```json
   "fileInputMimeTypes": ["application/pdf"]


AntonioLujanoLuna:

AntonioLujanoLuna:
Superseded by #324960 for the focused Anthropic PDF conversion fix. Endpoint-level PDF capability advertisement is now tracked separately in #324961. Closing this broader PR so the bug fix can be reviewed independently.

PR #323907, opened 2026-07-01
Disable GitHub Copilot utility models by default for BYOK models

### LINKED ISSUE #322751: Copilot Utility Models should default to none when BYOK is selected
When the user's main agent is a BYOK model, and the selection is `default` for a GitHub copilot utility model, currently a user's AI credits are used unless the utility models are switched to BYOK models. 
This default behavior needs to change so that if the main model is a BYOK model, unless explicitly configured, utility models must not use GitHub copilot models

vritant24:
When you pick a bring-your-own-key (BYOK) chat model, built-in "utility" flows (chat titles, summaries, and other small background calls) used to quietly fall back to GitHub Copilot's models — even without Copilot access. This PR stops that.

Now, when your main model is BYOK:

Utility flows no longer default to Copilot models.
Set chat.utilityModel / chat.utilitySmallModel to use your own model, or
Flip on the new chat.useCopilotModelsForUtilityModels setting to keep the old behavior.
Copilot main models are unaffected.

### DIFF
--- a/extensions/copilot/src/extension/chatInputNotification/vscode-node/byokUtilityModel.contribution.ts
+++ b/extensions/copilot/src/extension/chatInputNotification/vscode-node/byokUtilityModel.contribution.ts
@@ -16,9 +16,8 @@ const UTILITY_SMALL_MODEL_SETTING = 'chat.utilitySmallModel';
 /**
  * Shows a chat input notification in air-gapped BYOK scenarios (no GitHub
  * session) when at least one BYOK model is available but the utility model
- * settings are still defaults. The default utility models require GitHub
- * Copilot access, so without it the utility slots silently fall back and
- * degrade the experience until the user points them at a BYOK model.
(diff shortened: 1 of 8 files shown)


benvillalobos (approved):

DonJayamanne (approved):

LogicDaemon:
Idea is good, but implementation is flawed: currently if no utility model selected, Copilot just fails to respond.

Moreover, I'm using llama.cpp so re-loading the model is much higher overhead than just using the currently selected one. But if I have Qwen3.6 selected, and my utility model is, for example, Gemma 4 12B, its usage causes the Qwen to get offloaded, Gemma loaded, and then vice versa.

There should be an option like "Use same model as the main selected one", and a way to select models of a different provider (currently dropdowns only show models of the BYOK provider).

→ https://github.com/microsoft/vscode/issues/324007

MohammadKani (commented):
nice job. readed and everything sound correct.

PR #323955, opened 2026-07-01
Show the prior conversation inline (read-only) when continuing a chat into an Agent host

### LINKED ISSUE #322991: Migrating local to AHP sessions should provide access to the past conversation
After migrating a session from local to AHP, the _Previous conversation_ it attaches to isn't clickable, so I have no idea what's been talked about other than the session title.

To make the migration less painful, maybe we can treat the previous conversation in a richer UI, inline preview. 

Image

cc @bamurtaugh

vijayupadya:
Fixes https://github.com/microsoft/vscode/issues/322991

When user does "Continue Chat in…" to hand a conversation off to an agent-host session, the earlier exchange was previously only carried over as a truncated context attachment (a paste chip), so the prior conversation was effectively lost from view. This change renders the full prior conversation inline as read-only history in the new session, while the bounded transcript still flows to the agent as context.

**What's new**

- Inline previous conversation. The prior exchange is captured turn-by-turn at hand-off, persisted per-session, and re-rendered on every open (including after a reload) as read-only request/response items.
- Clear visual bracketing. A "Previous conversation" label opens the block at the top, and a matching label with a separator line closes it before the first live turn.
- Read-only enforcement. Imported turns cannot be edited, rerun, forked, restored, or undone — all affordances are gated (menus, keybindings, and the click-to-edit entry point), so they can't truncate the live history.
- Transcript chip is now hidden. The prior conversation is the user-facing surface (inline). The bounded transcript is still sent to the agent as context but no longer rendered as a redundant/degrading pill. The divider tooltip makes the guarantee explicit: "…read-only and were provided to the agent as context."

**Implementation notes**

- Session-database storage. Snapshots are persisted in the agent host's own per-session database via ISessionDataService (behind the existing IImportedConversationStore service). The snapshot is written/read over the agent connection as an opaque JSON blob and read on demand (not carried in continuously-synced session state), so opening a session fetches only that session's snapshot. Persistence, fork and delete are managed by the session database.
- Display decoupled from model-context. The inline (full, lossless) display is separate from the bounded transcript attachment handed to the agent, so backend turn machinery (checkpoints, diffing, fork) is unaffected — imported items are never backend turns.
- First-render correctness. Until a provisional (untitled-…) session materializes it has no database, so its snapshot is held in a short-lived in-memory bridge and flushed into the real session's database when it graduates to its backend resource — before provideChatSessionContent runs — so the block shows on the very first open, not just after a reload.
- Lifecycle managed by the session database. Deleting a session removes its database (and thus the snapshot) automatically — no explicit per-surface delete hook. Forking a session copies its database, so the imported conversation is carried into the fork for free (this also closes the earlier fork gap).


vijayupadya:
cc @anthonykim1

src/vs/workbench/contrib/chat/browser/importedConversationStore.ts
```diff
@@ -0,0 +1,167 @@
...
+		super();
+		// Snapshots live under the current profile; drop the cached index when the
+		// profile changes so a stale listing from another profile is not reused.
+		this._register(this._userDataProfileService.onDidChangeCurrentProfile(() => this._index = undefined));
+	}
+
+	private _root(): URI {
+		return joinPath(this._userDataProfileService.currentProfile.globalStorageHome, 'chatImportedConversations');
+	}
+
```
connor4312: Let's please use the `ISessionDataService` for this. That data is already managed and migrated e.g. when a session is forke. Then you don't need to reinvent the index and such.

connor4312 (dismissed):
one comment ^

src/vs/workbench/contrib/chat/browser/importedConversationStore.ts
vijayupadya: @connor4312 by session database do you mean `src\vs\workbench\contrib\chat\common\model\chatSessionStore.ts` ?

 One wrinkle: agent-host sessions aren't in ChatSessionStore. Did you mean persisting the snapshot in the agent-host backend session _meta (so fork/migration come for free), or making agent-host sessions participate in ChatSessionStore?

connor4312: I edited my comment -- I meant the `ISessionDataService` which is in the agent host

vijayupadya: @connor4312 
Done — switched the snapshot storage to ISessionDataService. It's now persisted as an opaque JSON blob in the session's own database (read on demand over the agent connection via two small getSessionImportedConversation / setSessionImportedConversation methods)

vijayupadya:
Closed in favor of https://github.com/microsoft/vscode/pull/324858/

PR #324012, opened 2026-07-02
Fix IPC serialization of non-JSON values

KevinWang-wpq:
This fixes IPC serialization for top-level values whose JSON representation is undefined, such as functions or symbols. Previously, the object fallback passed JSON.stringify(data) directly to VSBuffer.fromString; when JSON.stringify returns undefined, that can throw from Buffer.from(undefined).

The serializer already has an Undefined preset, so this change uses that preset when the fallback JSON serialization result is undefined.

A regression test covers round-tripping a non-JSON value as undefined.

Validation:
- git diff --check


KevinWang-wpq:

KevinWang-wpq:
Hi @alexdima, just a gentle follow-up on this PR. When you have a chance, would you mind taking a look? The automated CI checks have passed, and the PR is currently waiting for collaborator reviews. Thank you very much for your time and help!

PR #324025, opened 2026-07-02
fix: restore color theme when detecting high contrast theme

### LINKED ISSUE #269406: Automatic synchronization with Windows High Contrast mode
Automatic synchronization with **Windows High Contrast mode** (Alt + Shift + Print Screen)

I'm used to working both at home and outdoors. In open environments, High Contrast mode is often necessary, and Windows provides a convenient shortcut to enable it (Alt + Shift + Print Screen). Many Windows applications automatically adapt to this setting, but unfortunately, VSCode does not. It would be great if VSCode could switch its theme accordingly, just like other Windows apps do.

Thanks!

aiday-mar:
fixes https://github.com/microsoft/vscode/issues/269406


src/vs/workbench/services/themes/browser/workbenchThemeService.ts
```diff
@@ -450,7 +450,8 @@ export class WorkbenchThemeService extends Disposable implements IWorkbenchTheme
 
 	private installPreferredSchemeListener() {
 		this._register(this.hostColorService.onDidChangeColorScheme(() => {
-			if (this.settings.isDetectingColorScheme()) {
+			const detectHighContrastThemeChange = this.configurationService.getValue<boolean>(ThemeSettings.DETECT_HC);
```
aeschli: can you add and use `this.settings.isDetectingHighContrast()`

src/vs/workbench/services/themes/browser/workbenchThemeService.ts
```diff
@@ -535,6 +536,7 @@ export class WorkbenchThemeService extends Disposable implements IWorkbenchTheme
 	}
 
 	public async restoreColorTheme(): Promise<boolean> {
+		console.log('restoreColorTheme');
```
aeschli: remove the log

aeschli (changes requested):

aeschli (approved):

PR #324047, opened 2026-07-02
fix: match uppercase query characters in fuzzyContains

spokodev:
`fuzzyContains` lowercases the target but compares it against the original-case
query characters, so any uppercase character in the query can never match.

    fuzzyContains('hello world', 'H')  // false, should be true
    fuzzyContains('Explorer', 'E')     // false, should be true

The `viewQuickAccess` picker passes the raw user filter into
`fuzzyContains(entry.containerLabel, filter)` after only trimming it, so typing
an uppercase letter fails to fuzzy-match views by container label. Other callers
happened to work only because they lowercase both sides before calling, which
hid the defect at the utility level.

The fix lowercases the query as well, so the comparison is case-insensitive on
both sides. Added assertions for uppercase and mixed-case queries.

### DIFF
--- a/src/vs/base/common/strings.ts
+++ b/src/vs/base/common/strings.ts
@@ -917,11 +917,12 @@ export function fuzzyContains(target: string, query: string): boolean {
 
 	const queryLen = query.length;
 	const targetLower = target.toLowerCase();
+	const queryLower = query.toLowerCase();
 
 	let index = 0;
 	let lastIndexOf = -1;
 	while (index < queryLen) {
-		const indexOf = targetLower.indexOf(query[index], lastIndexOf + 1);
+		const indexOf = targetLower.indexOf(queryLower[index], lastIndexOf + 1);
 		if (indexOf < 0) {
 			return false;
 		}
--- a/src/vs/base/common/strings.ts
+++ b/src/vs/base/common/strings.ts
@@ -915,9 +915,10 @@ export function fuzzyContains(target: string, query: string): boolean {
 		return false; // impossible for query to be contained in target
 	}
 
-	const queryLen = query.length;
 	const targetLower = target.toLowerCase();
 	const queryLower = query.toLowerCase();
+	// toLowerCase() can change a string's length (e.g. İ -> i̇), so iterate over the lowered query.
+	const queryLen = queryLower.length;
 
 	let index = 0;
 	let lastIndexOf = -1;
--- a/src/vs/base/common/strings.ts
+++ b/src/vs/base/common/strings.ts
@@ -917,7 +917,7 @@ export function fuzzyContains(target: string, query: string): boolean {
 
 	const targetLower = target.toLowerCase();
 	const queryLower = query.toLowerCase();
-	// toLowerCase() can change a string's length (e.g. İ -> i̇), so iterate over the lowered query.
+	// toLowerCase() can change a string's length (e.g. \u0130 -> i\u0307), so iterate over the lowered query.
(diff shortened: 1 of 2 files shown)


dbaeumer (dismissed):
See Copilot review

spokodev:
Good catch, thanks. The loop now iterates over `queryLower.length` instead of the original `query.length`, so a query whose lowercase form changes length (İ lowercases to i̇) no longer skips its trailing code units. Added a regression test for that case.

spokodev:

dbaeumer:
@roblourens used from search.

roblourens:
Please open an issue and give a clear description of the actual problem you're solving. I can't repro an issue in the view quick access

spokodev:
The core problem is that `fuzzyContains` lowercases only the target, not the query, so any uppercase character in the query fails to match:

```
strings.fuzzyContains('hello world', 'H') // false, expected true
```

Separately, lowercasing can change the query's length: `İ` (U+0130) lowercases to two code units (`i` + combining dot U+0307), so bounding the loop by the original `query.length` skips a code unit. The fix lowercases the query and iterates over `queryLower.length`.

This is at the `fuzzyContains` level rather than the view quick access specifically, which is likely why it didn't repro there. Happy to open a tracking issue if you'd prefer that.

roblourens:
I understand the change to the function, but I'm trying to understand the user-visible bug in the product that is being fixed

spokodev:
It does surface in one place: the "Open View" quick access (`View: Open View`, or the `view` prefix in Quick Open). The PR body has the right framing; my earlier follow-up muddied it.

An entry there matches if either its view label fuzzy-matches (case-insensitive `matchesFuzzy`) or its container label matches via `fuzzyContains(entry.containerLabel, filter)`. That container path is the only case-sensitive one, so filtering by container name works only in lowercase.

Repro:
1. Run "View: Open View".
2. Type `explorer`. Views grouped under the Explorer container (e.g. Outline, Timeline) appear, matched by their container label.
3. Type `Explorer` (capital E, as the UI displays it). Those entries drop out, because `fuzzyContains` lowercases the target ("explorer") but tests it against the raw query character `E`.

So typing a container name exactly as shown (capitalized) returns fewer results than the lowercase form. The other two callers don't surface it because they pre-lowercase the query first: the Test Explorer filter (`testExplorerFilterState` lowercases each glob term) and file search (`normalizedFilePatternLowercase` / `prepareQuery().normalizedLowercase`). So the user-visible blast radius is just Open View.

roblourens:
Thanks, I see that now. Please open an issue for this- that's how we track work in this repo.

spokodev:
Opened #325664 with the repro and root cause.

TylerLeonhardt (dismissed):
LGTM thanks for contributing!

roblourens (dismissed):

src/vs/base/common/strings.ts
```diff
@@ -915,13 +915,15 @@ export function fuzzyContains(target: string, query: string): boolean {
 		return false; // impossible for query to be contained in target
 	}
 
-	const queryLen = query.length;
 	const targetLower = target.toLowerCase();
+	const queryLower = query.toLowerCase();
+	// toLowerCase() can change a string's length (e.g. \u0130 -> i\u0307), so iterate over the lowered query.
```
TylerLeonhardt, line 920: I love Turkish

TylerLeonhardt (approved):

amunger (approved):

PR #324096, opened 2026-07-02
feat(find): add option to ignore folded regions in find widget

priyanshi498:
Fixes issue 324071

Added a new toggle option to the editor Find Widget to ignore matches inside folded (collapsed) code blocks. This is really useful when you've folded large sections of code (like imports or utility functions) and want to search only through your visible, expanded text.

How it works:
- Adds a new toggle button (fold icon) next to the 'Find in Selection' button.
- Filters out matches that overlap with hidden lines using the editor's getHiddenAreas().
- Updates highlights and match count automatically if you fold/unfold regions while the find widget is active.
- Saves your toggle preference in workspace storage.

Testing:
- Added a new test case in findModel.test.ts to verify the count changes correctly when folding/unfolding and toggling the option.
- Verified client typecheck passes cleanly.


priyanshi498:

PR #324114, opened 2026-07-02
Add bilingual Simplified Chinese labels to settings

boxi-wangji:
This PR adds bilingual English/Simplified Chinese display labels for Settings UI entries, plus a small set of natural Simplified Chinese descriptions for commonly used settings.\n\nIt also includes tests for requested bilingual setting titles.\n\nNote: this is a customized bilingual UI change rather than a standard localization-only update.


boxi-wangji:

boxi-wangji:
Closing this PR because the bilingual Simplified Chinese text should be handled through a language pack/localization workflow instead of hardcoding translations in VS Code core source.

PR #324129, opened 2026-07-02
agentHost: sync .vscode/mcp.json servers, enable SDK config discovery

connor4312:
Refines how VS Code-configured MCP servers reach the Copilot agent host, and enables the SDK's config discovery so the host can find workspace `.mcp.json` on its own.

## Changes

- **`resolveCustomizationRefs` filtering** — When syncing customizations to the agent host, workspace-discovered MCP servers (config target `WORKSPACE` / `WORKSPACE_FOLDER`) are now excluded by default, since the agent host discovers workspace `.mcp.json` itself. The exception is `.vscode/mcp.json`, which the SDK does **not** discover: those servers are synced explicitly, but only when their config resolves **without user interaction**.
  - Resolution now uses `IConfigurationResolverService.resolveAsync` + `ConfigurationResolverExpression` instead of a naive regex. Servers that reference interactive variables (`${input:…}`, `${command:…}`) or fail to resolve (e.g. `${workspaceFolder}` with no folder) are skipped; non-interactive variables (`${workspaceFolder}`, `${env:…}`) are resolved and the server is synced with concrete values.

- **`enableConfigDiscovery: true`** on agent-host session create/resume, so the host discovers workspace `.mcp.json` itself (avoiding duplication with the sync path above).

- **Refactor** — Moved `MCP_CONFIGURATION_COLLECTION_ID_PREFIX` into `mcpTypes.ts` and added `McpCollectionDefinition.isWorkspaceDiscovered` / `isVscodeMcpJson` helpers so the collection-classification logic lives in one place.

## Notes

- The SDK's `enableConfigDiscovery` docs claim it discovers both `.mcp.json` **and** `.vscode/mcp.json`; in practice it only discovers `.mcp.json`. That's why `.vscode/mcp.json` is synced explicitly. A standalone SDK repro for the `.mcp.json`-only behavior has been captured separately for the SDK team.

## Tests

- Expanded `resolveCustomizationRefs.test.ts` (20 cases) covering: workspace `.mcp.json` exclusion, `.code-workspace` exclusion, `.vscode/mcp.json` sync of resolvable configs, exclusion of `${input:…}` servers, resolution of `${workspaceFolder}`, resolution-throws exclusion, and extension-contributed servers still syncing.
- `npm run typecheck-client` and `npm run valid-layers-check` pass.

<sub>Co-authored-by: Copilot &lt;223556219+Copilot@users.noreply.github.com&gt;</sub>

### DIFF
--- a/src/vs/platform/agentHost/node/copilot/copilotSessionLauncher.ts
+++ b/src/vs/platform/agentHost/node/copilot/copilotSessionLauncher.ts
@@ -515,6 +515,7 @@ export class CopilotSessionLauncher implements ICopilotSessionLauncher {
 			clientName: 'vscode',
 			enableMcpApps: true,
 			enableFileHooks: true,
+			enableConfigDiscovery: true,
 			onPermissionRequest: request => runtime.handlePermissionRequest(request),
 			onUserInputRequest: (request, invocation) => runtime.handleUserInputRequest(request, invocation),
 			onElicitationRequest: context => runtime.handleElicitationRequest(context),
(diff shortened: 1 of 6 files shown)


src/vs/platform/agentHost/node/copilot/copilotSessionLauncher.ts
```diff
@@ -515,6 +515,7 @@ export class CopilotSessionLauncher implements ICopilotSessionLauncher {
 			clientName: 'vscode',
 			enableMcpApps: true,
 			enableFileHooks: true,
+			enableConfigDiscovery: true,
```
vutung3196, line 518: LGTM

dmitrivMS (approved):

PR #324131, opened 2026-07-03
Prompt timeline rail for the Agents window

osortega:
## Prompt timeline

Adds a right-edge **prompt timeline** rail to the Agents window that lets users scan and jump between the prompts they have sent in a chat session — adapted from GitHub's "jump to your messages" affordance.

> [!NOTE]
> Draft — sharing early to keep history. Interaction/visuals are still being iterated (a "quiet gray-dense rail + single-mark accordion on hover" refinement is planned next).

### What's included
- **Self-contained contrib** under `src/vs/sessions/contrib/promptTimeline`, attached per-widget via `IChatWidgetContrib` (no `vs/workbench` changes, no import cycle).
- **Recency-bucketed ticks** — pure, unit-tested `budgetBucketPrompts`, capped so each tick keeps a **>=24px hit target** (WCAG 2.5.8).
- **Dense-at-rest visual** that expands on hover/focus, with green/red per-turn diff magnitude and a solid-blue active mark.
- **Interactive hover card** — prompt text, clickable `+/-` diff stats that open the per-request diff, and per-file drill-down rows.
- **Per-request diffs** via the provider-agnostic `IChatResponseFileChangesService` (agent-host server-computed turn changeset) with an editing-session fallback. Diff sides are probed with `readFile` so a missing agent-host checkpoint blob renders the file as a pure add/delete instead of crashing the multi-diff editor.
- **Keyboard-first commands** — Go to Next/Previous Prompt, Go to Prompt… (quick pick), Review Changes for Prompt — with ARIA announcements.
- **Gated** behind the new `sessions.promptTimeline.enabled` setting (experimental, default on) and `ChatContextKeys.enabled`; the rail is created/disposed reactively as the setting toggles.

### Validation
- `npm run typecheck-client` — pass
- `npm run valid-layers-check` — pass
- `eslint` on the contrib — pass
- `npm run precommit` (hygiene) — pass (only non-blocking design-token suggestions remain)
- `promptBucketing` unit tests

### Follow-ups
- Refine engaged interaction to the quiet gray-dense + single-mark accordion model.
- Apply design-token suggestions (font-size / radius / spacing) once the visual is locked.


src/vs/base/browser/ui/tree/abstractTree.ts
```diff
@@ -3110,6 +3110,22 @@ export abstract class AbstractTree<T, TFilterData, TRef> implements IDisposable
 		return this.view.getRelativeTop(index, stickyScrollNode?.position ?? this.stickyScrollController?.height);
 	}
 
+	/**
+	 * Returns the absolute top offset of an element in the tree's scroll/content
+	 * space, or `undefined` when the element is not in the tree. Unlike
+	 * {@link getRelativeTop}, this reads the layout height model, so it also
+	 * resolves elements outside the rendered viewport.
+	 */
+	getElementTop(location: TRef): number | undefined {
```
osortega, line 3119: getElementTop might not be needed anymore?

src/vs/sessions/contrib/promptTimeline/browser/promptTimelineActions.ts
```diff
@@ -0,0 +1,154 @@
...
+/** Resolves the prompt timeline contribution for the active session's chat widget. */
+function getPromptTimeline(accessor: ServicesAccessor): PromptTimelineWidgetContrib | undefined {
+	const widgetService = accessor.get(IChatWidgetService);
+	const sessionsService = accessor.get(ISessionsService);
+	const resource = sessionsService.activeSession.get()?.activeChat.get().resource;
+	const widget = (resource && widgetService.getWidgetBySessionResource(resource)) ?? widgetService.lastFocusedWidget;
+	return widget?.getContrib<PromptTimelineWidgetContrib>(PromptTimelineWidgetContrib.ID);
+}
+
+class GoToNextPromptAction extends Action2 {
```
osortega: Are all of the actions in this files actually used to drive the implementation of the timeline or are some just to execute through the command palette?

roblourens (approved):

roblourens:
Why is this done as an agents window-specific contrib, rather than a generic piece of UI that the editor window could also use?

osortega:
> Why is this done as an agents window-specific contrib, rather than a generic piece of UI that the editor window could also use?

That is the idea, but for now I want to scope it small until it's in a very polished state to integrate it into core

PR #324159, opened 2026-07-03
agentHost: route GitHub protected resources through a GitHub Enterprise endpoint service

### LINKED ISSUE #313396: Agent host: support GitHub Enterprise (GHE) for Copilot auth
Today the agent host's Copilot auth path is hardcoded to github.com end-to-end. To use Copilot from a GHE instance, two layers need to change in lockstep:

## 1. Agent host: protected-resource metadata is github.com-only

The Copilot agent's `getProtectedResources()` only advertises github.com:

```ts
getProtectedResources(): ProtectedResourceMetadata[] {
    return [{
        resource: 'https://api.github.com',
        resource_name: 'GitHub Copilot',
        authorization_servers: ['https://github.com/login/oauth'],
        scopes_supported: ['read:user', 'user:email'],
        required: true,
    }];
}
```

This drives which OAuth server the CLI / client performs the device-code flow against. For a GHE user it needs to advertise `https://<ghe-host>/api/v3` (or equivalent) and `https://<ghe-host>/login/oauth`. The GHE host should be **configurable by the agent host operator** (config file / env / launch flag) so a single agent-host binary can be pointed at any GHE instance without a rebuild.

## 2. `CopilotApiService`: token mint URL is hardcoded to `defaultChatAgent.tokenEntitlementUrl`

See the TODO in `src/vs/platform/agentHost/node/shared/copilotApiService.ts` (`_buildCapiInit`):

```ts
// TODO(GHE): For GitHub Enterprise users the mint URL must point to
// `api.<enterprise-host>/copilot_internal/v2/token` instead.
const tokenUrl = this._productService.defaultChatAgent.tokenEntitlementUrl;
```

For GHE this needs to point at `api.<ghe-host>/copilot_internal/v2/token` instead of `api.github.com/...`. The enterprise host has to thread through the public `ICopilotApiService` surface (extra param on `messages`/`models`, a per-host instance, or a `create(enterpriseHost?)` factory) — `defaultAccount.ts` in the workbench already derives this URL the same way for the main auth path and is a good reference.

## Why this is one issue

(1) and (2) are coupled — fixing only the mint URL doesn't help if the OAuth flow still hits github.com, and vice versa. Both should land together with a shared "GHE host" configuration source.

## Out of scope

- The Copilot CLI's own OAuth handling (Rust, in `cli/`) is github.com-hardcoded but is a separate consumer; this issue is specifically about the agent host being usable against a GHE Copilot.

TylerLeonhardt:
Fixes #313396

## What

Introduce `IAgentHostGitHubEndpointService` / `AgentHostGitHubEndpointService` as a single source of truth for GitHub (Enterprise) endpoints and their protected resources, so the agent host can target a configured GitHub Enterprise host instead of hardcoded `github.com` constants. This makes the agent host's Copilot auth and CAPI paths usable against a GHE instance without a rebuild.

## Changes

- Add `githubEndpoints.ts` (endpoint/resource derivation) and the endpoint service, exposed off `AgentService` via `gitHubEndpointService`.
- Copilot, Claude, and Codex agents resolve their Copilot/repo protected resources via `getCopilotResource()` / `getRepoResource()` rather than the hardcoded `GITHUB_*_PROTECTED_RESOURCE` constants — so the OAuth device-code flow targets the configured host.
- `CopilotApiService` derives its token mint URL from the endpoint service (`api.<ghe-host>/copilot_internal/v2/token`) instead of the hardcoded `defaultChatAgent.tokenEntitlementUrl`, resolving the `TODO(GHE)` in `_buildCapiInit`.
- Order `CopilotApiService` and the proxies that consume it after the endpoint service is registered (`agentHostMain` mirrors `agentHostServerMain`) so CAPI endpoint discovery can target a GHE host.
- `CopilotAgent` sets `COPILOT_GH_HOST` for the CLI and restarts the client when the enterprise host changes; `AgentService` emits `auth/required` on GHE URI change so clients re-authenticate against the new resource.
- Add unit tests (`githubEndpoints.test.ts`, `agentHostGitHubEndpointService.test.ts`) plus a `testGitHubEndpointService` helper, and thread the service through the affected agent/service tests.

### DIFF
--- a/src/vs/platform/agentHost/node/copilot/copilotAgent.ts
+++ b/src/vs/platform/agentHost/node/copilot/copilotAgent.ts
@@ -38,7 +38,7 @@ import { CopilotCliConfigKey, copilotCliConfigSchema } from '../../common/copilo
 import { AgentHostMcpServersConfigKey, AgentHostSessionSyncEnabledConfigKey, AutoApproveLevel, ISchemaProperty, SessionMode, createSchema, migrateLegacyAutopilotConfig, platformRootSchema, platformSessionSchema, schemaProperty, type AgentHostMcpServers } from '../../common/agentHostSchema.js';
 import { IAgentPluginManager, ISyncedCustomization } from '../../common/agentPluginManager.js';
 import { AgentSessionEntry, decodeProviderData, encodeProviderData, type IPersistedChat } from '../agentPeerChats.js';
-import { AgentSession, AgentSignal, AuthenticateParams, GITHUB_COPILOT_PROTECTED_RESOURCE, GITHUB_REPO_PROTECTED_RESOURCE, IActiveClient, IAgent, IAgentChatDataChange, IAgentChats, IAgentLegacyChat, IAgentCreateChatForkSource, IAgentCreateChatOptions, IAgentCreateChatResult, IAgentCreateSessionConfig, IAgentCreateSessionResult, IAgentDescriptor, IAgentMaterializeSessionEvent, IAgentModelInfo, IAgentResolveSessionConfigParams, IAgentSessionConfigCompletionsParams, IAgentSessionMetadata, IAgentSessionProjectInfo, IAgentSpawnChatEvent, IMcpNotification, IRestoredSubagentSession, SubagentChatSignal } from '../../common/agentService.js';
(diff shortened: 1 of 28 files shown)


src/vs/platform/agentHost/node/copilot/copilotAgent.ts
```diff
@@ -897,8 +925,19 @@ export class CopilotAgent extends Disposable implements IAgent {
 			}
 
 			// Identify VS Code's agent host traffic in CAPI
-			env['GITHUB_COPILOT_INTEGRATION_ID'] = COPILOT_INTEGRATION_ID;
-			this._logService.info(`[Copilot] Set CLI env: GITHUB_COPILOT_INTEGRATION_ID=${COPILOT_INTEGRATION_ID}`);
+			// [TEMP] integration-id disabled to test enterprise model enumeration
+			// env['GITHUB_COPILOT_INTEGRATION_ID'] = COPILOT_INTEGRATION_ID;
+			// this._logService.info(`[Copilot] Set CLI env: GITHUB_COPILOT_INTEGRATION_ID=${COPILOT_INTEGRATION_ID}`);
```
TylerLeonhardt: I had to do this to get the models to show up, idk why... needs some research.

connor4312 (dismissed):

roblourens (dismissed):

justschen (approved):

rwoll (approved):

PR #324160, opened 2026-07-03
fix: prevent unbounded PersistentProtocol replay buffer

KevinWang-wpq:
Fixes a memory leak in `PersistentProtocol`.

### Details

`PersistentProtocol` keeps every regular message in `_outgoingUnackMsg` until the peer acknowledges it. That queue is also the replay buffer used when a connection reconnects.

When a connection is dead, congested, or stuck reconnecting, acknowledgements can stop advancing while callers continue to send messages. In that state, `_outgoingUnackMsg` keeps retaining every unacknowledged payload, including large `VSBuffer` instances, until the protocol is disposed.

### Change

The change bounds the replay buffer by both message count and retained payload bytes. It also tracks the retained bytes so timeout diagnostics can show how much payload memory is currently held by unacknowledged messages.

When either replay-buffer limit is exceeded, `PersistentProtocol` emits a dedicated `unacknowledgedMessageReplayBufferOverflow` timeout reason, clears the replay buffer, and advances the local acknowledgement state so the dropped entries are not kept alive.

For remote connections, replay-buffer overflow is treated as unrecoverable instead of reconnecting into a stream that is missing replay messages.

### Before

A dead or congested peer could stop sending acknowledgements while new messages were still appended to `_outgoingUnackMsg`. The replay buffer could then grow without a memory bound and retain all queued message payloads.

### After

The replay buffer is capped by configurable/default count and byte limits. When the cap is exceeded, the buffer is released, the timeout log includes the retained byte count, and node IPC regression tests cover both byte-limit and count-limit overflow paths.

### DIFF
--- a/src/vs/base/parts/ipc/common/ipc.net.ts
+++ b/src/vs/base/parts/ipc/common/ipc.net.ts
@@ -139,11 +139,13 @@ export type SocketCloseEvent = NodeSocketCloseEvent | WebSocketCloseEvent | unde
 export const enum SocketTimeoutReason {
 	UNACKNOWLEDGED_MESSAGE = 'unacknowledgedMessage',
 	KEEP_ALIVE = 'keepAlive',
+	UNACKNOWLEDGED_MESSAGE_REPLAY_BUFFER_OVERFLOW = 'unacknowledgedMessageReplayBufferOverflow',
 }
 
 export interface SocketTimeoutEvent {
 	readonly reason: SocketTimeoutReason;
 	readonly unacknowledgedMsgCount: number;
+	readonly unacknowledgedMsgBytes: number;
 	readonly timeSinceOldestUnacknowledgedMsg?: number;
 	readonly timeSinceLastReceivedSomeData: number;
 }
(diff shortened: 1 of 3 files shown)


KevinWang-wpq:
Adding the anonymized heap-snapshot evidence that motivated this fix.

### Heap snapshot evidence

In a renderer heap snapshot, the dominant retained object was the reconnect replay buffer owned by `PersistentProtocol`:

| Object | Retained size | Notes |
| --- | ---: | --- |
| `Queue@6324627` | 8.81 GB | Held from `PersistentProtocol._outgoingUnackMsg` |
| `PersistentProtocol@4699437` | 8.81 GB via `_outgoingUnackMsg` | The protocol's unacknowledged-message replay buffer |
| First queued payload | ~142 MB | A queued `ProtocolMessage` retained a large `Uint8Array`/buffer payload |
| Remaining linked-list payloads | ~8.67 GB | Additional queued messages retained through `QueueElement.next` |

The relevant retainer chain was:

```text
Queue@6324627 (8.81 GB)
  <- [_outgoingUnackMsg] PersistentProtocol@4699437
    <- [protocol] PersistentConnection@4206011
      <- PersistentConnection._instances
        <- Window / GC root
```

This matches the failure mode addressed here: regular messages are appended to `_outgoingUnackMsg` for reconnect replay, and if acknowledgements stop advancing while sends continue, the queue can retain all unacknowledged payloads without a memory bound.

This PR adds the missing bound in two dimensions:

- max unacknowledged replay message count
- max retained unacknowledged payload bytes

It also treats replay-buffer overflow as unrecoverable for remote connections, so the connection does not reconnect into a stream whose replay messages were already dropped.

@Giuspepe could you please take a look when you have a chance? Copilot's latest re-review on the updated commit generated no new comments, and the PR now includes regression coverage for both byte-limit and count-limit overflow paths.

PR #324184, opened 2026-07-03
fix(diagnostics): ignore invalid uri deletes

lizuju:
Addresses the `t.with is not a function` diagnostics path in microsoft/vscode#322755.

Summary
- guard `DiagnosticCollection.delete` against non-`URI` input before firing `onDidChangeDiagnostics`
- add regression coverage for deleting a plain URI-shaped object that is not a real VS Code `URI`

Verification
- `npm run typecheck-client` with Node 24.17.0
- focused runtime assertion via `npx -y tsx --tsconfig src/tsconfig.json`: invalid delete input is ignored without firing a diagnostics event or proxy change
- `git diff --check`

Note: I could not run `./scripts/test.sh --grep "diagnostic collection ignores invalid uri on delete"` locally because the Electron download failed with `ERR_TLS_CERT_ALTNAME_INVALID` for `github.com` on this network.

### DIFF
--- a/src/vs/workbench/api/common/extHostDiagnostics.ts
+++ b/src/vs/workbench/api/common/extHostDiagnostics.ts
@@ -172,6 +172,9 @@ export class DiagnosticCollection implements vscode.DiagnosticCollection {
 
 	delete(uri: vscode.Uri): void {
 		this._checkDisposed();
+		if (!URI.isUri(uri)) {
+			return;
+		}
 		this.#onDidChangeDiagnostics.fire([uri]);
 		this.#data.delete(uri);
 		this.#proxy?.$changeMany(this._owner, [[uri, undefined]]);
--- a/src/vs/workbench/api/test/browser/extHostDiagnostics.test.ts
+++ b/src/vs/workbench/api/test/browser/extHostDiagnostics.test.ts
@@ -385,6 +385,23 @@ suite('ExtHostDiagnostics', () => {
 		await p;
 	});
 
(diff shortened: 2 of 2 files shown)


roblourens (commented):
Do we know why there's an invalid URI or who's passing it in?

lizuju:
From the issue data, I do not see the exact producer identified. The public stack points at the Copilot diagnostics listener as the consumer that crashes after calling `languages.getDiagnostics(uri)`, but the earlier issue analysis only narrows the producer side to an extension calling `DiagnosticCollection.delete(...)` with a URI-shaped plain object rather than a real `URI` instance.

So I would not claim this is caused by Copilot itself. Copilot appears to be the first listener that dereferences every URI from `onDidChangeDiagnostics`, which exposes the bad event payload. The core issue here is that `delete()` currently fires `onDidChangeDiagnostics` before doing any `URI.isUri` validation, while `set()` already validates single-URI input. That means one bad extension input can leak a non-URI through the typed `DiagnosticChangeEvent.uris` contract and crash any listener that trusts the API.

If you would rather track down and block the exact caller first, I can hold this PR. But based on the API boundary, I still think the guard belongs in `delete()` so invalid input is ignored before it is broadcast.

lizuju:
Hi, just following up on this when you have a chance. I rechecked #322755 and there still does not seem to be a concrete caller identified beyond invalid input reaching DiagnosticCollection.delete(). The guard and regression test are unchanged. Please let me know if you would like me to trace anything else before review.

PR #324235, opened 2026-07-03
fix: checking position is valid before accessing line content

aiday-mar:
fixes https://github.com/microsoft/vscode/issues/283293

### DIFF
--- a/src/vs/editor/browser/controller/editContext/textArea/textAreaEditContext.ts
+++ b/src/vs/editor/browser/controller/editContext/textArea/textAreaEditContext.ts
@@ -470,6 +470,9 @@ export class TextAreaEditContext extends AbstractEditContext {
 	}
 
 	private _getAndroidWordAtPosition(position: Position): [string, number] {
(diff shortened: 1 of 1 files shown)


hediet (approved):

src/vs/editor/browser/controller/editContext/textArea/textAreaEditContext.ts
```diff
@@ -470,6 +470,9 @@ export class TextAreaEditContext extends AbstractEditContext {
 	}
 
 	private _getAndroidWordAtPosition(position: Position): [string, number] {
+		if (position.lineNumber > this._context.viewModel.getLineCount()) {
```
hediet, line 473: doesn't this just hide a bug? Why would the line number be > getLineCount?

aiday-mar: I am not so sure, this was shown to be the case in telemetry error reporting...

Issue #256039, opened 2025-07-15
Multiple cursor copy-paste of non-selected lines pastes everything


BartolHrg:
Type: <b>Bug</b>

1. have this contents of the file with multiple cursors at | (and nothing selected)  
```
a|
b|
```
2. Copy (Ctrl + C)  
3. Paste (Ctrl + V)  

Expected:  
```
a
a
b
b
```
Result:  
```
aa
b

ba
b

```

If I were to individualy do the steps for each cursor (just 2 of them in this case), the result would be what's shown above (in the "Expected" region)  

I have `"editor.multiCursorPaste": "full"`, however, this setting should not apply in this case, since I don't have a single selection with multiple lines,  
I have multiple copied items with single line in each.  

I have slighltly modified VSCode installation, with only lineCommentCommand.ts modified to support tabs,  
therefore, this also doesn't apply.  

VS Code version: Code 1.102.0 (cb0c47c0cfaad0757385834bd89d410c78a856c0, 2025-07-09T22:10:34.600Z)
OS version: Windows_NT x64 10.0.26100
Modes: Unsupported

<details>
<summary>System Info</summary>

|Item|Value|
|---|---|
|CPUs|13th Gen Intel(R) Core(TM) i7-13700H (20 x 2918)|
|GPU Status|2d_canvas: enabled<br>canvas_oop_rasterization: enabled_on<br>direct_rendering_display_compositor: disabled_off_ok<br>gpu_compositing: enabled<br>multiple_raster_threads: enabled_on<br>opengl: enabled_on<br>rasterization: enabled<br>raw_draw: disabled_off_ok<br>skia_graphite: disabled_off<br>video_decode: enabled<br>video_encode: enabled<br>vulkan: disabled_off<br>webgl: enabled<br>webgl2: enabled<br>webgpu: enabled<br>webnn: disabled_off|
|Load (avg)|undefined|
|Memory (System)|15.63GB (4.74GB free)|
|Process Argv|--crash-reporter-id 5235ac69-27f7-4b93-9f49-763295cb8572|
|Screen Reader|no|
|VM|0%|
</details><details><summary>Extensions (37)</summary>

Extension|Author (truncated)|Version
---|---|---
vscode-zipfs|arc|3.0.0
brainfuck-syntax|att|0.0.1
manipulatenumbers|Bar|1.1.0
pynewline|Bar|1.1.1
arm|dan|1.7.4
RunOnSave|eme|0.2.7
[...]

PR #324294, opened 2026-07-04
Fix blank lines when copying multiple empty selections

### LINKED ISSUE #324290: Copying multiple empty selections adds blank lines when pasted into a single cursor
Does this issue occur when all extensions are disabled?: Unknown

- VS Code Version: Reproduced in 1.108.0 through 1.127.0; does not reproduce in 1.107.x or 1.98.1
- OS Version: Microsoft Windows 11 Pro 10.0.26100 (26100)

Steps to Reproduce:

1. Open a file with the following contents:

   ```
   line1
   line2
   line3
   ```
2. Place one empty cursor at the start of each of the three lines.
3. Copy.
4. Open a new empty file with a single cursor and paste.

Expected:

```
line1
line2
line3
```

Actual:

```
line1

line2

line3
```

Notes:

- This appears to be a regression introduced in 1.108.0.
- Reproduced by comparing versions directly: 1.107.x is OK, 1.108.0 and later are not.
- It still reproduces in 1.127.0.
- This looks related to the fix for #256039, but it is a different regression from #286892.
- The likely cause is that multiple empty selections now produce `multicursorText` entries that already include trailing newlines, and the plain text clipboard payload then joins those entries with another newline, producing blank lines when pasted into a single cursor.

Potential source locations:

- `src/vs/editor/common/viewModel/viewModelImpl.ts`
- `src/vs/editor/browser/controller/editContext/clipboardUtils.ts`

Suspected introducing commit:

- `20e65fd2ff3753918d34968d260b5ed1a071980d` (`fix copy with multiple cursors and empty selections`)

CHENJIAMIAN:
Fixes #324290.

## Summary

When copying from multiple empty selections, `getPlainTextToCopy()` may return an array whose entries already include trailing line endings. The clipboard plain text payload was then built by joining those entries with another newline, which introduced blank lines when pasting into a single cursor.

This change keeps `multicursorText` unchanged for multi-cursor paste distribution, but builds the plain text clipboard payload by concatenating the array entries directly.

## Test Plan

- Added a regression test for copying from multiple empty selections and pasting into a single cursor.
- Did not run the full VS Code test suite locally because this machine is missing the native C/C++ toolchain required by the repository preinstall checks.

### DIFF
--- a/src/vs/editor/browser/controller/editContext/clipboardUtils.ts
+++ b/src/vs/editor/browser/controller/editContext/clipboardUtils.ts
@@ -45,11 +45,10 @@ function generateDataToCopy(viewModel: IViewModel): { dataToCopy: ClipboardDataT
 
 function getDataToCopy(viewModel: IViewModel, modelSelections: Range[], emptySelectionClipboard: boolean, copyWithSyntaxHighlighting: boolean): ClipboardDataToCopy {
 	const { sourceRanges, sourceText } = viewModel.getPlainTextToCopy(modelSelections, emptySelectionClipboard, isWindows);
-	const newLineCharacter = viewModel.model.getEOL();
 
 	const isFromEmptySelection = (emptySelectionClipboard && modelSelections.length === 1 && modelSelections[0].isEmpty());
 	const multicursorText = (Array.isArray(sourceText) ? sourceText : null);
-	const text = (Array.isArray(sourceText) ? sourceText.join(newLineCharacter) : sourceText);
+	const text = (Array.isArray(sourceText) ? sourceText.join('') : sourceText);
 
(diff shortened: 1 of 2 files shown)


CHENJIAMIAN:

CHENJIAMIAN:
Updated this PR to handle mixed selections as well.

What changed:
- The plain text clipboard payload now only inserts an EOL separator between chunks when the previous chunk does not already end with one.
- Added a regression test for the mixed non-empty + empty selection case (`['ine2', 'line3\n']` => `ine2\nline3\n`).

This keeps the `#256039` multi-cursor distribution behavior, fixes `#324290`, and avoids regressing mixed-selection copy.

